PR #2194, opened 2022-11-02
refactor(longitudinal_controller): add debug code velocity controller

TakaHoribe:
## Description

- refactor: remove unnecessary arg (controller_state) from some functions
- refactor: add debug message for state transition

The issue report like "the target speed seems to be positive but the vehicle does not move" comes a lot.
The cause is mainly the state change condition (driving distance is too short, the velocity controller is waiting for steering control convergence, etc.).

So add a print for state change in the velocity controller. 

The state transition worked correctly.

image

### DIFF
--- a/control/trajectory_follower/include/trajectory_follower/pid_longitudinal_controller.hpp
+++ b/control/trajectory_follower/include/trajectory_follower/pid_longitudinal_controller.hpp
@@ -268,11 +268,9 @@ private:
 
   /**
    * @brief update control state according to the current situation
-   * @param [in] current_control_state current control state
    * @param [in] control_data control data
    */
-  ControlState updateControlState(
-    const ControlState current_control_state, const ControlData & control_data);
+  void updateControlState(const ControlData & control_data);
 
   /**
    * @brief calculate control command based on the current control state
--- a/control/trajectory_follower/include/trajectory_follower/pid_longitudinal_controller.hpp
+++ b/control/trajectory_follower/include/trajectory_follower/pid_longitudinal_controller.hpp
@@ -107,6 +107,14 @@ private:
   // control state
   enum class ControlState { DRIVE = 0, STOPPING, STOPPED, EMERGENCY };
   ControlState m_control_state{ControlState::STOPPED};
+  std::string toStr(const ControlState s)
+  {
+    if (s == ControlState::DRIVE) return "DRIVE";
+    if (s == ControlState::STOPPING) return "STOPPING";
+    if (s == ControlState::STOPPED) return "STOPPED";
+    if (s == ControlState::EMERGENCY) return "EMERGENCY";
+    return "UNDEFINED";
+  };
 
   // control period
   float64_t m_longitudinal_ctrl_period;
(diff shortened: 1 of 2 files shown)


control/trajectory_follower/include/trajectory_follower/pid_longitudinal_controller.hpp
```diff
@@ -107,6 +107,14 @@ class TRAJECTORY_FOLLOWER_PUBLIC PidLongitudinalController : public Longitudinal
   // control state
   enum class ControlState { DRIVE = 0, STOPPING, STOPPED, EMERGENCY };
   ControlState m_control_state{ControlState::STOPPED};
+  std::string toStr(const ControlState s)
```
takayuki5168, line 110: reference

control/trajectory_follower/src/pid_longitudinal_controller.cpp
```diff
@@ -539,45 +554,70 @@ PidLongitudinalController::ControlState PidLongitudinalController::updateControl
...
+        return print_and_return(ControlState::STOPPING);
       }
     } else {
       if (stopped_condition && !departure_condition_from_stopped) {
-        return ControlState::STOPPED;
+        return print_and_return(ControlState::STOPPED);
       }
     }
-  } else if (current_control_state == ControlState::STOPPING) {
+    return print_and_return(ControlState::DRIVE);
```
takayuki5168, line 564: This seems always called in the drive state.

IMO, better to have prev_control_state_ and compare it with current one.

control/trajectory_follower/include/trajectory_follower/pid_longitudinal_controller.hpp
takayuki5168: add const to the function

"`toStr` of pid_longitudinal_controller" does not make sense to me. `convertControlStateToString` or `getStringControlState` sounds better.

TakaHoribe: `toStr(enum)` still sounds odd? It receives the enum and return as a string. Sounds straightforward to me.

takayuki5168: Alright. This may be kind of preference, so if you think it's okay, forget my comment.

(I feel that
- "toStr of pid_longitudinal_controller class" does not make sense to me. We cannot understand what to convert to string only with the function name, until seeing the function's implementation.
- For example, "toStr of ControlState class" makes sense to me. We can almost understand what to convert to string only with the function name.

control/trajectory_follower/src/pid_longitudinal_controller.cpp
TakaHoribe: change implementation. Please check it.

takayuki5168 (approved):
LGTM

control/trajectory_follower/include/trajectory_follower/pid_longitudinal_controller.hpp
TakaHoribe: I agree. Now the ControlState is implemented in the longitudinal_controller class as an enum, so the "toStr of ControlState class" can not be implemented. I hope an enum class could have methods.

>  We cannot understand what to convert to string only with the function name,

`toStr()` is definitely vague, but I feel, with an arg expression `toStr(controlState)` looks obvious for its role.

takayuki5168: Okay!

PR #2202, opened 2022-11-02
feat(interpolation): add supporting methods to make the spline interpolators work with Eigen and NMPC

boyali:
## Related links

The current spline interpolation methods do not support some methods used in the nonlinear MPC. We add additional methods and overloaded functions to make the spline methods work with the NMPC.

### DIFF
--- a/common/interpolation/include/interpolation/interpolation_utils.hpp
+++ b/common/interpolation/include/interpolation/interpolation_utils.hpp
@@ -80,7 +80,7 @@ inline bool isStrictlyMonotonic(const std::vector<double> & x)
   return is_monotonic;
 }
 
-inline void validateKeys(
+inline std::vector<double> validateKeys(
   const std::vector<double> & base_keys, const std::vector<double> & query_keys,
   const bool & extrapolate_end_points = false)
 {
@@ -102,12 +102,20 @@ inline void validateKeys(
 
   // When query_keys is out of base_keys (This function does not allow extrapolation when
   // extrapolation boolean is false  // ).
+  constexpr double epsilon = 1e-3;
   if (
-    (query_keys.front() < base_keys.front() || base_keys.back() < query_keys.back()) &&
+    (query_keys.front() < base_keys.front() - epsilon ||
+     base_keys.back() < query_keys.back() + epsilon) &&
     !extrapolate_end_points) {
     throw std::invalid_argument(
       "The query_keys is out of the range of base_keys, consider to extrapolate option");
   }
+
+  auto validated_query_keys = query_keys;
+  validated_query_keys.front() = std::max(validated_query_keys.front(), base_keys.front());
+  validated_query_keys.back() = std::min(validated_query_keys.back(), base_keys.back());
+
+  return validated_query_keys;
 }
 
 template <class T>
--- a/common/interpolation/include/interpolation/interpolation_utils.hpp
+++ b/common/interpolation/include/interpolation/interpolation_utils.hpp
@@ -24,7 +24,6 @@
 
 namespace interpolation_utils
 {
-constexpr double EPS = std::numeric_limits<double>::epsilon();
 
 template <class T, typename std::enable_if_t<std::is_floating_point_v<T>, bool> * = nullptr>
 bool isEqual(T a, T b)
--- a/common/interpolation/include/interpolation/interpolation_utils.hpp
+++ b/common/interpolation/include/interpolation/interpolation_utils.hpp
@@ -18,82 +18,110 @@
 #include <array>
 #include <stdexcept>
 #include <vector>
+#include <algorithm>
 
 namespace interpolation_utils
 {
-inline bool isIncreasing(const std::vector<double> & x)
+
+// Strictly monotonic increasing
+inline bool isIncreasing(const std::vector<double> &x)
 {
-  if (x.empty()) {
+  if (x.empty())
+  {
     throw std::invalid_argument("Points is empty.");
   }
 
-  for (size_t i = 0; i < x.size() - 1; ++i) {
-    if (x.at(i) >= x.at(i + 1)) {
-      return false;
-    }
-  }
+  // x(i) <= x(i+1) condition check
+  auto const &is_increasing_it = std::adjacent_find(x.begin(), x.end(), std::greater_equal<>());
 
-  return true;
+  return is_increasing_it == x.cend();
 }
 
-inline bool isNotDecreasing(const std::vector<double> & x)
+// Strictly increasing and monotonic
+inline bool isNotDecreasing(const std::vector<double> &x)
 {
-  if (x.empty()) {
+  if (x.empty())
+  {
     throw std::invalid_argument("Points is empty.");
   }
 
-  for (size_t i = 0; i < x.size() - 1; ++i) {
-    if (x.at(i) > x.at(i + 1)) {
-      return false;
-    }
+  // x(i) < x(i+1) condition check ( x(i) == x(i+1) is allowed)
+  auto const &is_strictly_increasing_it = std::adjacent_find(x.begin(), x.end(), std::greater<>());
+
+  return is_strictly_increasing_it == x.cend();
+}
+
+inline bool isStrictlyMonotonic(const std::vector<double> &x)
+{
+  if (x.empty())
+  {
+    throw std::invalid_argument("Points is empty.");
   }
 
-  return true;
+  // x(i) <= or >= x(i+1) condition check
+  auto const &is_strictly_increasing_it = std::adjacent_find(x.begin(), x.end(), std::greater_equal<>());
+  auto const &is_strictly_decreasing_it = std::adjacent_find(x.begin(), x.end(), std::less_equal<>());
+
+  // if cannot find <= or >= conditions, the vector is strictly monotonic.
+  bool const is_strictly_increasing = is_strictly_increasing_it == x.cend();
+  bool const is_strictly_decreasing = is_strictly_decreasing_it == x.cend();
+
+  bool const is_monotonic = (is_strictly_increasing && !is_strictly_decreasing) ||
+                            (!is_strictly_increasing && is_strictly_decreasing);
+
+  return is_monotonic;
 }
 
 inline void validateKeys(
-  const std::vector<double> & base_keys, const std::vector<double> & query_keys)
+  const std::vector<double> &base_keys, const std::vector<double> &query_keys)
 {
   // when vectors are empty
-  if (base_keys.empty() || query_keys.empty()) {
+  if (base_keys.empty() || query_keys.empty())
+  {
     throw std::invalid_argument("Points is empty.");
   }
 
   // when size of vectors are less than 2
-  if (base_keys.size() < 2) {
+  if (base_keys.size() < 2)
+  {
     throw std::invalid_argument(
       "The size of points is less than 2. base_keys.size() = " + std::to_string(base_keys.size()));
   }
 
   // when indices are not sorted
-  if (!isIncreasing(base_keys) || !isNotDecreasing(query_keys)) {
+  if (!isIncreasing(base_keys) || !isNotDecreasing(query_keys))
+  {
     throw std::invalid_argument("Either base_keys or query_keys is not sorted.");
   }
 
   // when query_keys is out of base_keys (This function does not allow exterior division.)
-  if (query_keys.front() < base_keys.front() || base_keys.back() < query_keys.back()) {
+  if (query_keys.front() < base_keys.front() || base_keys.back() < query_keys.back())
+  {
     throw std::invalid_argument("query_keys is out of base_keys");
   }
 }
 
-template <class T>
+template<class T>
 void validateKeysAndValues(
-  const std::vector<double> & base_keys, const std::vector<T> & base_values)
+  const std::vector<double> &base_keys, const std::vector<T> &base_values)
 {
   // when vectors are empty
-  if (base_keys.empty() || base_values.empty()) {
+  if (base_keys.empty() || base_values.empty())
+  {
     throw std::invalid_argument("Points is empty.");
   }
 
   // when size of vectors are less than 2
-  if (base_keys.size() < 2 || base_values.size() < 2) {
+  if (base_keys.size() < 2 || base_values.size() < 2)
+  {
     throw std::invalid_argument(
       "The size of points is less than 2. base_keys.size() = " + std::to_string(base_keys.size()) +
       ", base_values.size() = " + std::to_string(base_values.size()));
   }
 
   // when sizes of indices and values are not same
-  if (base_keys.size() != base_values.size()) {
+  if (base_keys.size() != base_values.size())
+  {
     throw std::invalid_argument("The size of base_keys and base_values are not the same.");
   }
 }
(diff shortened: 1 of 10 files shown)


boyali:
I have added isStrictlyMonotonic checks and tests. I also put minor makeups in isIncreasing and isNotDecreasing methods' inner workflow by replacing for..loops to STL adjacent_find (please let me know if this does not fit your design). The related tests are untouched.

boyali:
I will add Eigen supports on top of these minor modifications.

boyali:
I have added optional extrapolation into the LERP methods. There could be cases where the end-point values are slightly bigger or smaller than the base key values. These values, in some cases, should not be rejected but extrapolated. I put the extrapolation option as a default false which does not affect the code usage by the other methods (compatible). 

I also replaced linear quer_key search in the lerp scalar with the binary search values. Tests are not affected.

boyali:
@rej55 -san
I am switching to use this package in the NMPC. I would like to report an issue in the logic of getSplineInterpolationValues. In this method we check if the both base and query keys are sorted. But, we actually do not need any query key to be sorted. I implemented binary search tree in lerp methods, I also added a binary search in the interpolatePoint method of spline class. I  use single point interpolation along the modules of the NMPC. 

I wonder your opinions, do we really need to enforce sorted query keys?

TakaHoribe:
@boyali This is still a draft. Are you still working on this? Or is it already ready for review?

Here is my comment:
- The format is broken. Please apply the pre-commit again and check if the modification is what you expect in [files changed tab](https://github.com/autowarefoundation/autoware.universe/pull/2202/files).
- having an extrapolation option in existing functions seems to be ok to me.

boyali:
> @boyali This is still a draft. Are you still working on this? Or is it already ready for review?
> 
> Here is my comment:
> 
> * The format is broken. Please apply the pre-commit again and check if the modification is what you expect in [files changed tab](https://github.com/autowarefoundation/autoware.universe/pull/2202/files).
> * having an extrapolation option in existing functions seems to be ok to me.

I am not working on it, just waited for the reviews. I will apply pre-commit and change it to PR ready. Thank you for reviewing.

common/interpolation/src/spline_interpolation.cpp
```diff
@@ -114,10 +114,12 @@ std::vector<double> splineByAkima(
     if (i == 0) {
       s_values.push_back(m_values.front());
       continue;
-    } else if (i == base_keys.size() - 1) {
+    }
+    if (i == base_keys.size() - 1) {
```
takayuki5168, line 118: Any reason to replace `else if` with `if` ?
I'm not sure yours is better than the current one.

common/interpolation/include/interpolation/interpolation_utils.hpp
```diff
@@ -15,48 +15,78 @@
...
-    throw std::invalid_argument("Points is empty.");
+    throw std::invalid_argument("The vector x is empty.");
   }
 
-  for (size_t i = 0; i < x.size() - 1; ++i) {
-    if (x.at(i) > x.at(i + 1)) {
-      return false;
-    }
+  // x(i) < x(i+1) condition check ( x(i) == x(i+1) is allowed)
+  auto const & is_strictly_increasing_it = std::adjacent_find(x.begin(), x.end(), std::greater<>());
```
takayuki5168, line 56: Usually, we use `auto const` not `const auto` in autoware.universe.
Any reason to use `auto const`?

common/interpolation/include/interpolation/interpolation_utils.hpp
```diff
@@ -15,48 +15,78 @@
...
+                            (!is_strictly_increasing && is_strictly_decreasing);
+
+  return is_monotonic;
 }
 
-inline void validateKeys(
-  const std::vector<double> & base_keys, const std::vector<double> & query_keys)
+inline std::vector<double> validateKeys(
+  const std::vector<double> & base_keys, const std::vector<double> & query_keys,
+  const bool & extrapolate_end_points = false)
```
takayuki5168, line 85: For primitive type (int, bool, double, ...), we do not have to add reference.

common/interpolation/include/interpolation/interpolation_utils.hpp
```diff
@@ -70,10 +100,27 @@ inline void validateKeys(
     throw std::invalid_argument("Either base_keys or query_keys is not sorted.");
   }
 
-  // when query_keys is out of base_keys (This function does not allow exterior division.)
-  if (query_keys.front() < base_keys.front() || base_keys.back() < query_keys.back()) {
-    throw std::invalid_argument("query_keys is out of base_keys");
+  // When query_keys is out of base_keys (This function does not allow extrapolation when
+  // extrapolation boolean is false  // ).
```
takayuki5168, line 104: The second `//` on the line 104 is not required.

common/interpolation/include/interpolation/interpolation_utils.hpp
```diff
@@ -15,48 +15,78 @@
...
 #include <vector>
 
 namespace interpolation_utils
 {
+constexpr double EPS = std::numeric_limits<double>::epsilon();
+
+template <class T, typename std::enable_if_t<std::is_floating_point_v<T>, bool> * = nullptr>
+bool isEqual(T a, T b)
+{
+  return abs(a - b) < std::numeric_limits<T>::epsilon();
```
takayuki5168, line 32: better to use std::abs instead of abs

common/interpolation/src/spline_interpolation.cpp
```diff
@@ -260,7 +292,7 @@ std::vector<double> SplineInterpolation::getSplineInterpolatedDiffValues(
       ++j;
     }
 
-    const double ds = query_key - base_keys_.at(j);
+    const double & ds = query_key - base_keys_.at(j);
```
takayuki5168, line 295: No need to add reference to primitive type?

common/interpolation/src/spline_interpolation.cpp
```diff
@@ -236,13 +243,38 @@ std::vector<double> SplineInterpolation::getSplineInterpolatedValues(
...
+  // k1 is the index query_key <=base_keys[k1]
+  auto const & k1 = std::distance(base_keys_.cbegin(), lower_bound_it);
+  auto const & k0 = k1 == 0 ? k1 : k1 - 1;  // if upper bound coincides with the first element
+
+  // Get the interpolated value
+  const double & ds = query_key - base_keys_.at(k0);
+  auto const & query_value = d.at(k0) + (c.at(k0) + (b.at(k0) + a.at(k0) * ds) * ds) * ds;
+
+  return query_value;
+}
```
takayuki5168, line 276: This function's name should be the same as `getSplineInterpolatedValues`.

In addition, to remove duplicated implementation, this function should be like
```
std::vector<double> query_keys(query_key);
return getSplineInterpolatedValues(query_keys.front()).front();
```

What do you think.

common/interpolation/src/linear_interpolation.cpp
```diff
@@ -52,8 +62,41 @@ std::vector<double> lerp(
...
+  // if k0 and k1 are the first index of the base_key
+  if (k0 == k1) {
+    return base_values[k0];
+  }
+
+  auto const & ratio = (query_key - base_keys.at(k0)) / (base_keys.at(k1) - base_keys.at(k0));
+
+  // Get terminal data items.
+  auto const & query_value = base_values.at(k0) + ratio * (base_values.at(k1) - base_values.at(k0));
+  return query_value;
```
takayuki5168, line 100: To remove duplicated implementations, IMO, it would be better to implement like
```
double lerp(
  const std::vector<double> & base_keys, const std::vector<double> & base_values, double query_key,
  bool const & extrapolate_end_points)
{
  return lerp(base_keys, base_values, std::vector<double>{query_key}, extrapolate_end_points).front();
}
```

common/interpolation/include/interpolation/interpolation_utils.hpp
```diff
@@ -15,48 +15,78 @@
...
-    if (x.at(i) > x.at(i + 1)) {
-      return false;
-    }
+  // x(i) < x(i+1) condition check ( x(i) == x(i+1) is allowed)
+  auto const & is_strictly_increasing_it = std::adjacent_find(x.begin(), x.end(), std::greater<>());
+
+  return is_strictly_increasing_it == x.cend();
+}
+
+inline bool isStrictlyMonotonic(const std::vector<double> & x)
```
takayuki5168, line 61: Could you tell me why this condition is required please.

common/interpolation/include/interpolation/interpolation_utils.hpp
```diff
@@ -97,6 +144,38 @@ void validateKeysAndValues(
...
+    auto const & tn = base_keys.rbegin()[0];
+    auto const & tn_1 = base_keys.rbegin()[1];
+
+    auto const & yn = base_values.rbegin()[0];
+    auto const & yn_1 = base_values.rbegin()[1];
+
+    auto const & ratio = (query_key - tn) / (tn - tn_1);
+    query_value = yn + ratio * (yn - yn_1);  // extrapolation and lerp have different semantics.
+  }
+}
```
takayuki5168, line 177: Is this implementation required?

https://github.com/autowarefoundation/autoware.universe/blob/a02830a142c6e8ad4b51c5174ba1a72d8237b829/common/interpolation/src/linear_interpolation.cpp#L21-L24
This lerp function allows extrapolation.

common/interpolation/src/spline_interpolation.cpp
boyali: clang recommentation do not use else if after continue.

common/interpolation/include/interpolation/interpolation_utils.hpp
boyali: Before this modification, there were only monotonically increasing checks. However, it does not matter whether the base coordinate is increasing or decreasing. This method is the mathematical description of monotonicity. Sometimes we might need to interpolate decreasing monotonic coordinates (0: -1, -2, ... -5, for example), sometimes increasing.

common/interpolation/include/interpolation/interpolation_utils.hpp
boyali: Yes, it is the original implementation and is used within the code. The second lerp I have added has different arguments called by the nonlinear MPC.

common/interpolation/include/interpolation/interpolation_utils.hpp
boyali: My development environment setting like this, and this is more natural to me. Is this rule strict and compulsory? If so, I make change.

common/interpolation/src/spline_interpolation.cpp
boyali: This breaks other codes. I tried. 
/mnt/WD/Codes/ROS2_AWF/src/universe/autoware.universe/common/interpolation/src/spline_interpolation_points_2d.cpp:131:69: error: request for member ‘at’ in ‘((const SplineInterpolationPoints2d*)this)->SplineInterpolationPoints2d::spline_y_.SplineInterpolation::getSplineInterpolatedValues(whole_s)’, which is of non-class type ‘double’
  131 |   const double y = spline_y_.getSplineInterpolatedValues({whole_s}).at(0);

boyali:
I think it is all done. Would you please check?

common/interpolation/include/interpolation/interpolation_utils.hpp
takayuki5168: I mean that `lerp` already considers extrapolation, so I guess we do not have to add `lerp_extrapolate`.

What you wanna do in NMPC is linear interpolation allowing extrapolation with the following function, right?
```
std::vector<double> lerp(
  const std::vector<double> & base_keys, const std::vector<double> & base_values,
  const std::vector<double> & query_keys, bool const & extrapolate_end_points = false);
```
Then I'll add a change to your PR later.

common/interpolation/include/interpolation/interpolation_utils.hpp
takayuki5168: Currently, the key is `s` of the path, but as you said, we can use decreasing values. I got it.

common/interpolation/include/interpolation/interpolation_utils.hpp
boyali: I haven't seen any extrapolation. In the MPC implementations, we sometimes add extra time or distance to guarantee MPC prediction length. At least the NMPC requires it.

common/interpolation/include/interpolation/interpolation_utils.hpp
```diff
@@ -15,48 +15,78 @@
...
+#include <algorithm>
 #include <array>
+#include <functional>
+#include <limits>
 #include <stdexcept>
 #include <vector>
 
 namespace interpolation_utils
 {
+constexpr double EPS = std::numeric_limits<double>::epsilon();
```
takayuki5168, line 27: IMO, this global epsilon should be removed.
How small the epsilon should be depends on each algorithm.
Therefore, it will be confusing if there is a global epsilon variable.

common/interpolation/include/interpolation/interpolation_utils.hpp
```diff
@@ -15,48 +15,78 @@
...
+    throw std::invalid_argument("The vector x is empty.");
   }
 
-  for (size_t i = 0; i < x.size() - 1; ++i) {
-    if (x.at(i) >= x.at(i + 1)) {
-      return false;
-    }
-  }
+  // x(i) <= x(i+1) condition check
+  const auto & is_increasing_it = std::adjacent_find(x.begin(), x.end(), std::greater_equal<>());
```
takayuki5168, line 43: Any reason to add reference here?

takayuki5168: In addition, how about
```
const bool is_increasing = std::adjacent_find(x.begin(), x.end(), std::greater_equal<>()) == x.cend();
return is_increasing;
```
or
```
return std::adjacent_find(x.begin(), x.end(), std::greater_equal<>()) == x.cend();
```
since `is_increasing_it` does not make sense to me.

Ignore my comment if you don't think it's better.

common/interpolation/include/interpolation/interpolation_utils.hpp
boyali: Thank you very much, we can remove it, after your commits I remove it.

common/interpolation/include/interpolation/interpolation_utils.hpp
boyali: Yes you are right. I put _it formiterator.

common/interpolation/include/interpolation/interpolation_utils.hpp
takayuki5168: I finished my change. Please modify!

boyali: I couldn’t have done it without you. I have removed EPS.

common/interpolation/include/interpolation/interpolation_utils.hpp
boyali: Changed to clear naming.

boyali:
I think this was all.

boyali:
I think I addressed all the review issues.

takayuki5168 (approved):
LGTM

xmfcx:
@boyali could you rebase this PR to main? There seems to be a conflict in `common/interpolation/package.xml`.

Issue #2235, opened 2022-11-07
Refactor conditional ROS_DISTRO_GALACTIC include with __has_include


ralwing:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I've agreed with the maintainers that I can plan this task.

### Description

Replace 
```
#ifdef ROS_DISTRO_GALACTIC
```
with
```
#if __has_include
```
where applicable

### Purpose

This will release code from environmentall variables and also make it more readable and easier to configure, and still galactic/humble compatible.

### Possible approaches

Refactor code

### Definition of done

All conditional includes replaced with suggested change.

kenji-miyake:
Wrote my concerns here: https://github.com/autowarefoundation/autoware.universe/pull/2236#issuecomment-1305785887

> This will release code from environmentall variables and also make it more readable and easier to configure,

@ralwing Could you explain the reason why you felt so? I feel it's not clear.

PR #2236, opened 2022-11-07
refactor: use __has_include for backward compatibility

ralwing:
This PR implements changes from #2235


takayuki5168:
@wep21 Does this PR make sense to you?

yukkysaito:
LGTM
I think this is make sense.

wep21:
Yes, it makes sense to me, but cppcheck caused this error in `galactic` when I tried the change.
https://sourceforge.net/p/cppcheck/discussion/development/thread/8596f26bbb/?limit=25
I don't know how to fix the error.

ralwing:
No idea what's wrong with DCO:
```
Grzegorz Głowacki [gglowacki@autonomous-systems.pl](mailto:gglowacki@autonomous-systems.pl) is not a valid email address.
```

ralwing:
@wep21 Seems like [the issue in cppcheck](https://github.com/danmar/simplecpp/issues/229)  is far from resolving it, is the inability to use cppcon a blocker?

kenji-miyake:
@ralwing Thank you for the PR. Since I have several concerns, I'll comment on them.

First of all, as explained in https://github.com/orgs/autowarefoundation/discussions/2622, we'll move to Humble soon.
Are these changes really worth merging now? They'll be removed soon.

And considering the transitioning tasks, I'm concerned about that this PR makes the tasks a little difficult. I'll explain the reason below.

The `ROS_DISTRO_` macro is also used for other than include files.
For example, it's used for `to_yaml`.

```
src/universe/autoware.universe/common/component_interface_utils/include/component_interface_utils/rclcpp/service_client.hpp
80:#ifdef ROS_DISTRO_GALACTIC
81-    using rosidl_generator_traits::to_yaml;
82-#endif
--
91:#ifdef ROS_DISTRO_GALACTIC
92-    return client_->async_send_request(request, wrapped);
93-#else
```

Let's think about when we remove the code for ROS distro compatibility.

If there is only the `ROS_DISTRO_` macro, the worker only needs to `grep ROS_DISTRO_`, and then all related parts code blocks are found.

But if `ROS_DISTRO_` and `__has_include` are mixed, the worker has to also `grep _has_include`, which increases the costs.
And then, it might be difficult to distinguish whether the code is for ROS distro compatibility or not.

@xmfcx What do you think? I'd like to hear your opinion.

kenji-miyake:
(In addition, even though it's a personal preference, I don't feel it's clean if the header name is written twice.)

image

xmfcx:
I don't mind header file being written twice, because it is still easier than writing a macro variable and the header.

But like @kenji-miyake pointed, when we want to end galactic support, it will be hard to know what to remove.

# Example

```cpp
#ifdef ROS_DISTRO_GALACTIC
#include <tf2_geometry_msgs/tf2_geometry_msgs.h>
#else
#include <tf2_geometry_msgs/tf2_geometry_msgs.hpp>
#endif
```

turns into:

```cpp
#if __has_include(<tf2_geometry_msgs/tf2_geometry_msgs.hpp> )
#include <tf2_geometry_msgs/tf2_geometry_msgs.hpp>
#else
#include <tf2_geometry_msgs/tf2_geometry_msgs.h>
#endif
```

To make this easy to show that we are keeping `.h` version for backwards compatibility, we can make it:

```cpp
#if __has_include(<tf2_geometry_msgs/tf2_geometry_msgs.hpp> )
// humble and later
#include <tf2_geometry_msgs/tf2_geometry_msgs.hpp>
#else
// for galactic support
#include <tf2_geometry_msgs/tf2_geometry_msgs.h>
#endif
```

so that people who wants to deprecate the galactic can search for "for galactic support" to see the references.

But this makes it longer. And if `__has_include` is used for some other reasons like different os, compatibility for multiple libraries, maybe cuda or not, it gets hard to keep track of the reason why each `__has_include` stands for.

I agree it helps us remove dependency for a variable (`ROS_DISTRO_GALACTIC`) but it also makes the macro ambiguous. I think we shouldn't make this change.

xmfcx:
@ralwing Even if we are to make this change, please suggest and add an identifier keyword comment like `//for galactic support` so people can search for it while deprecating old parts.

kenji-miyake:
@xmfcx Thanks for the detailed comment.

And there is another reason.
Since the person who will do the transition tasks would be me, I wouldn't like to merge this and increase my tasks, unless there are any rationale for this.

ralwing:
If i may @kenji-miyake, in my opinion refactor would be even easier, because You could simply `sed` all those changes, i've refactored in such a way that they all have the same structure, before there was many different methods of veryfing it:
```
#ifdef ROS_DISTRO_GALACTIC
#include <tf2_eigen/tf2_eigen.h>
#include <tf2_sensor_msgs/tf2_sensor_msgs.h>
#else
#include <tf2_eigen/tf2_eigen.hpp>

#include <tf2_sensor_msgs/tf2_sensor_msgs.hpp>
#endif
```
vs:
```
#ifdef ROS_DISTRO_GALACTIC
#include <tf2_eigen/tf2_eigen.h>
#include <tf2_sensor_msgs/tf2_sensor_msgs.h>
#else
#include <tf2_eigen/tf2_eigen.hpp>
#include <tf2_sensor_msgs/tf2_sensor_msgs.hpp>
#endif
```
This slight difference could have complicated the refactor in the future.  
But saying that, so I've got a compromise, let's meet somewhere in the middle:
instead of my change i'm offering this:
```
#if !defined(ROS_DISTRO_GALACTIC) && __has_include ( <tf2_sensor_msgs/tf2_sensor_msgs.hpp> )
#include <tf2_sensor_msgs/tf2_sensor_msgs.hpp>
#else
#include <tf2_sensor_msgs/tf2_sensor_msgs.h>
#endif
```
Does this satisfy Your needs? If so I'll update this branch.
And also... Thanks a lot for being so polite to me :bow: 
Grzegorz

kenji-miyake:
@ralwing Thank you for your comment. My proposal is, as I wrote first, to wait for the Humble transition. Then we don't need `ifdef` or `__has_include` anymore.
Of course, if you have any issues with the `ROS_DISTRO_` macro, then I can consider merging this now. In that case, please explain your issues.

> First of all, as explained in https://github.com/orgs/autowarefoundation/discussions/2622, we'll move to Humble soon.
Are these changes really worth merging now? They'll be removed soon.

kenji-miyake:
Also, CI build fails. I don't want to use much time for this now if there is no issue.
https://github.com/autowarefoundation/autoware.universe/actions/runs/3411438565/jobs/5675705119#step:7:1352

kenji-miyake:
> in my opinion refactor would be even easier, because You could simply sed all those changes, and also I've refactored code in such a way that they all have the same structure, before there were many different methods of inclusions:

@ralwing However, for future works, I'd appreciate it if you could explain the following things:
- Exactly what commands/steps did you execute for this PR?
- What commands/steps are you planning to execute to remove all `__has_include` after we move to Humble?

ralwing:
ok, i'm closing this.

kenji-miyake:
@ralwing Anyway, thank you again for the PR and patience! I really appreciate your contribution.

PR #2287, opened 2022-11-15
fix(behavior_path_planner): fix lane change logic on the edge case 

zulfaqar-azmi-t4:
While experimenting in Odaiba, due to the safety reason, it is difficult to test the lane change function when there is object parked at the end of the road. To verify this, the confirmation is executed via PSIM. r

Screenshot from 2022-11-17 12-48-42

Once tested in PSIM, when longitudinal check is performed, it is found out that the lane change is not successful, due to the collision check in prepare phase. To allow user to be able to relax the collision check when the `enable_collision_check_at_prepare_phase` is true, the parameter that ignore target which speed falls under some threshold should be introduced.

Screenshot from 2022-11-18 14-18-23

## Description

The PR solves several issue.
1. When setting `enable_collision_check_at_prepare_phase = true`, for both stationary object and moving object, the collision check will be conducted for the duration of prepare phase + lane changing phase. But it is nice if we can ignore prepare phase check for stationary object only. Therefore new parameter is added `prepare_phase_ignore_target_speed_thresh` to ignore target with certain speed.   
2. When performing the collision check, ego vehicle's trajectory is converted into predicted path, and for each pose in the predicted path, ego's polygon will be generated. However, currently ego's polygon is generated only at current pose and not the predicted path. Therefore to improve this, ego polygon will be generated at the estimated pose. 
3. In addition, the debug message is improve to allow more expressive debug message.
4. Added default value to `LaneChangeParameter`'s member variable to ensure no undefined behavior related to the parameter declaration.

### Before change
image21801

### After change
improved

## Tests performed

Test lane change scenario with stationary and moving target.

## Notes for reviewers

None


TakaHoribe:
> However, since this is parked vehicle, we don't really need to include prepare phase in the collision check.

I don't get this idea well. Why the "collision check in prepare phase" is not necessary for a parked vehicle?

In my understanding, even if it is a parked vehicle, the collision check for lane change must be performed in all phases.
However, it is reasonable that the collision check for parked vehicles does not have to be as strict as for moving vehicles. We could set the collision margin very small in the collision check for parked vehicles.

TakaHoribe (approved):
LGTM

TakaHoribe:
>I don't get this idea well. Why the "collision check in prepare phase" is not necessary for a parked vehicle?
In my understanding, even if it is a parked vehicle, the collision check for lane change must be performed in all phases.
However, it is reasonable that the collision check for parked vehicles does not have to be as strict as for moving vehicles. We could set the collision margin very small in the collision check for parked vehicles.

Discussion with @zulfaqar-azmi-t4 : the collision check in the prepare phase is to restrict the lane change plans overtaking a vehicle, not critical to the safety.

PR #2293, opened 2022-11-15
fix(gyro_odometer): use all data

kminoda:
Signed-off-by: kminoda <koji.minoda@tier4.jp>

## Description
I would like to change the implementation of gyro_odometer.

Current gyro_odometer had an issue when the IMU frequency is higher than the velocity frequency. For example when IMU freq = 200 Hz and velocity freq = 50 Hz, the current gyro_odometer will concatenate both data and output twist in 200 Hz. Since the velocity frequency is not that high, it upsamples velocity observation by simply publishing one observed data **four times**. 

This may be a problem because, assuming that the covariance of IMU and velocity are properly estimated, EKF may be overly confident on the velocity observation. 

To be accurate on probability theory, I would like to change the implementation to the following:
- publish twist for every dt [sec] (e.g. dt=0.02[sec] or 50 Hz. This is also a parameter)
- if either velocity or IMU is not observed in the last dt [sec], do not publish twist
- if two or more data is observed for either velocity or IMU in the last dt [sec], publish a mean value of all the observed data (and lower the covariance based on central limit theorem)

#### Normal scenario test
Run Autoware with tutorial data and confirmed that the localization works fine and the covariance estimation is also reasonable

#### Edge-case test
- Run Autoware without velocity (and also IMU) input and confirmed that the warning is output in log
- Run Autoware, wait for a few seconds and kill velocity (and also IMU) topic, and confirmed that the error is output in log

### DIFF
--- a/localization/gyro_odometer/src/gyro_odometer_core.cpp
+++ b/localization/gyro_odometer/src/gyro_odometer_core.cpp
@@ -31,8 +31,7 @@ geometry_msgs::msg::TwistWithCovarianceStamped get_twist_from_velocity_buffer(
 {
   geometry_msgs::msg::TwistWithCovarianceStamped twist_with_cov;
   if (velocity_buffer.empty()) {
-    twist_with_cov.twist.covariance[0 * 6 + 0] = 100000.0;
-    return twist_with_cov;
+    throw std::domain_error("Should not receive an empty vector");
   }
 
   int n = static_cast<int>(velocity_buffer.size());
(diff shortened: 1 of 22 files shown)


localization/gyro_odometer/src/gyro_odometer_core.cpp
```diff
@@ -187,3 +271,23 @@ bool GyroOdometer::getTransform(
   }
   return true;
 }
+
+void GyroOdometer::checkTimeout(
```
YamatoAndo, line 282: this function is not used

kminoda:
Let me make this PR a draft for now.
I encountered a possible degradation. The twist rate drops to 23-25 Hz even though `output_rate = 50`, IMU freq is 28-30Hz, wheel odometry freq is 28-30Hz.

Issue #2213, opened 2022-11-03
Vehicle stops in front of non-existing obstacle on the path


WJaworskiRobotec:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I'm convinced that this is not my fault but a bug.

### Description

While testing current Autoware in AWSIM for ITRI BusODD scenarios, the following problem was discovered :

The vehicle stops because the obstacle on the path is reported, but when detected objects are visualized, it seems that the objects are not really blocking the path.

- the only detected object is far from the vehicle, and nnot on the path
![Screenshot from 2022-11-03 11-50-28](https://user-images.githubusercontent.com/85557834/199752475-4e30d34b-d0a2-4c0f-99b5-329eb6d975f3.png)

- detected object is on the side of the vehicle, and is not blocking the movement

![Screenshot from 2022-11-03 11-32-13](https://user-images.githubusercontent.com/85557834/199753080-ea2bb522-fca2-4631-ba94-3dcdcdca3ee0.png)
![Screenshot from 2022-11-03 11-32-39](https://user-images.githubusercontent.com/85557834/199752934-f9e10b17-1617-4272-a46e-d6c19367232c.png)
![Screenshot from 2022-11-03 11-33-36](https://user-images.githubusercontent.com/85557834/199752939-7cebbf8d-90ec-4dd5-8ff7-f013c3b622f6.png)

Location of the observed problems:

![Screenshot from 2022-11-07 10-34-49](https://user-images.githubusercontent.com/85557834/200276662-e39a7978-e74a-44e1-8e1a-204dde9d37ee.png)

### Expected behavior

 The vehicle should not stop if there are no objects detected on the path

### Actual behavior

The vehicle stops with no objects on the path

### Steps to reproduce

Run AWSIM binary with Autoware - description to be updated once binary is publicly shared

maxime-clem:
An "obstacle on the path" can also be caused by an obstacle _point_.
Have you checked the pointcloud in rviz ? I am afraid the points from the tree leaves are what is causing your issue.
Here is the related open discussion: https://github.com/orgs/autowarefoundation/discussions/2981

shmpwk:
@WJaworskiRobotec 
This problem may comes from [this PR](https://github.com/autowarefoundation/autoware.universe/pull/1899/files).
[...]

PR #2296, opened 2022-11-15
fix(obstacle_stop_planner): Add checking of point height.

lchojnack:
## Description

This PR adds checking of a point height. If a point is lower than the vehicle's height, then the point is in collision with the vehicle. This allows the EGO to drive when an obstacle is above the EGO e.g. tree branches and leaves.

Related to:
- issue https://github.com/autowarefoundation/autoware.universe/issues/2213 
- discussion https://github.com/orgs/autowarefoundation/discussions/2981


maxime-clem (commented):
The idea seems good, expect for the case where ego is about to drive up a slope.
I believe in that case we can encounter an obstacle that is _currently_ at a higher `z` than the vehicle height, but that will still collide with the ego vehicle if it follows its trajectory.
image

What do you think ?

planning/obstacle_stop_planner/src/planner_utils.cpp
```diff
@@ -441,8 +441,10 @@ bool withinPolygon(
     Point2d point(candidate_points_ptr->at(j).x, candidate_points_ptr->at(j).y);
     if (bg::distance(prev_point, point) < radius || bg::distance(next_point, point) < radius) {
       if (bg::within(point, boost_polygon)) {
-        within_points_ptr->push_back(candidate_points_ptr->at(j));
-        find_within_points = true;
+        if (candidate_points_ptr->at(j).z < vehicle_height) {
```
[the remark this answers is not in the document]
lchojnack: As I understand from the code, the function `searchPointcloudNearTrajectory()` transforms the received point cloud to the trajectory frame, and then the transformed point cloud is used to detect collision between the trajectory and a point from the point cloud (`withinPolygon()` function).

So when the point cloud and the trajectory are in the same frame, the case with a slope is taken into consideration. 
Taking your example, I draw how trajectory would look from the side view in my opinion.
trajectory

Is my reasoning correct or did I miss something?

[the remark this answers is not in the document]
badai-nguyen: I also think this is good idea. To cover driving up a slope case, a dynamic z threshold might be better. How do you thing?
image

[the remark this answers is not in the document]
maxime-clem: Sorry I have not had time to check in details yet but I started testing this PR.
It is not easy to test in the simple planning simulator but I managed to create 2 situations that look wrong.

- First, obstacle points that are under the trajectory are causing the obstacle stop planner to stop.
image

- Second, obstacle points that are on the elevated part of the trajectory are ignored if the ego vehicle is at a similar elevation. If the ego vehicle is at a lower elevation they correctly cause a stop.
image
image

I think this highlight some issues with comparing the `z` values from trajectory and from the pointcloud.

lchojnack:
I've updated the PR. I've added a polyhedron for visualization (topic: `/planning/scenario_planning/lane_driving/motion_planning/obstacle_stop_planner/debug/marker`) and colliding point cloud (topic: `/planning/scenario_planning/lane_driving/motion_planning/obstacle_stop_planner/output/point_cloud/debug`). 

I've tested the PR in some scenarios, and I think it works now. Printscreens below
Screenshot_20221129_101853
Screenshot_20221128_153850
Screenshot_20221128_153753
Screenshot_20221128_153701

maxime-clem (commented):
Thank you for the update. I tested and it is working well.

I think the main remaining issue is that the polyhedrons are not rotated to match the road inclination. Since this is probably difficult to implement (and expensive to compute), how about adding a `z_buffer` parameter to enlarge the vehicle footprint on the z axis ? By increasing this buffer, a user could compensate for inclined roads.

By the way, did you profile the current version ? I am curious about the performance difference between checking for collision in 2D and 3D.

lchojnack:
@maxime-clem Thank you very much for your review. 
 
> I think the main remaining issue is that the polyhedrons are not rotated to match the road inclination. Since this is probably difficult to implement (and expensive to compute), how about adding a `z_buffer` parameter to enlarge the vehicle footprint on the z axis ? By increasing this buffer, a user could compensate for inclined roads.

I will add `z_buffer` parameter. 

> By the way, did you profile the current version ? I am curious about the performance difference between checking for collision in 2D and 3D.

No, I didn't. I will profile the performance in terms of time computation and provide a report.

planning/obstacle_stop_planner/src/planner_utils.cpp
```diff
@@ -449,6 +449,32 @@ bool withinPolygon(
...
+    boost_polygon.outer().push_back(bg::make<Point2d>(point.x, point.y));
+  }
+  boost_polygon.outer().push_back(bg::make<Point2d>(cv_polygon.front().x, cv_polygon.front().y));
+
+  for (size_t j = 0; j < candidate_points_ptr->size(); ++j) {
+    Point2d point(candidate_points_ptr->at(j).x, candidate_points_ptr->at(j).y);
+    if (bg::distance(prev_point, point) < radius || bg::distance(next_point, point) < radius) {
+      if (bg::within(point, boost_polygon)) {
+        if (candidate_points_ptr->at(j).z < z_max && candidate_points_ptr->at(j).z > z_min) {
+          within_points_ptr->push_back(candidate_points_ptr->at(j));
```
maxime-clem, line 469:
```suggestion
  for (const auto & p : *candidate_points_ptr) {
    Point2d point(p.x, p.y);
    if (bg::distance(prev_point, point) < radius || bg::distance(next_point, point) < radius) {
      if (bg::within(point, boost_polygon)) {
        if (p.z < z_max && p.z > z_min) {
          within_points_ptr->push_back(p);
```

maxime-clem (commented):
Thank you for the update and for providing profiling results.

I tested and it works as expected. The code looks good and I only have some small suggestions.
I think the last thing to do before merging is update the documentation of the `obstacle_stop_planner` for the new feature. Do you know if some tests (either in simulation or on a real vehicle) are planned to confirm that the original issue (https://github.com/autowarefoundation/autoware.universe/issues/2213) is solved by this PR ?

planning/obstacle_stop_planner/src/planner_utils.cpp
WJaworskiRobotec: @maxime-clem We are planning to run the test in AWSIM to confirm that the original issue is solved and record some videos. But to do that we need a bit more time to create a new ITRI binary based on humble instead of galactic. I will be working on it next week.

lchojnack: @maxime-clem 
Thank you for your review. I've applied your hints.

planning/obstacle_stop_planner/src/node.cpp
```diff
@@ -388,26 +407,50 @@ void ObstacleStopPlannerNode::searchObstacle(
...
+        planner_data.found_collision_points = withinPolyhedron(
+          one_step_move_vehicle_polygon, stop_param.stop_search_radius, prev_center_point,
+          next_center_point, slow_down_pointcloud_ptr, collision_pointcloud_ptr, z_axis_min,
+          z_axis_max);
+        auto after_size = collision_pointcloud_ptr->size();
+        if ((after_size - before_size) > 0) {
+          auto obstacle_ros_pointcloud_debug_ptr = std::make_shared<PointCloud2>();
+          pcl::toROSMsg(*collision_pointcloud_ptr, *obstacle_ros_pointcloud_debug_ptr);
+          obstacle_ros_pointcloud_debug_ptr->header.frame_id = trajectory_header.frame_id;
+          pub_point_cloud_->publish(*obstacle_ros_pointcloud_debug_ptr);
```
maxime-clem, line 432: Should this code be removed ?

maxime-clem (changes requested):
Thanks for the updates. I think there are some debugging code remaining in the code that need to be removed, otherwise everything looks good.
One thing still missing is an update to the `obstacle_stop_planner` documentation but this can be updated in a separate PR.
I am now testing the PR with some scenarios to make sure there are no big issue. Once they finish and once the remaining minor code issues are fixed, I will approve the PR.

planning/obstacle_stop_planner/src/node.cpp
lchojnack: @maxime-clem Thank you for your review. I leave this debug code in order to visualize the obstacle points as a point cloud. What do you think about adding a parameter enabling debugging visualization of an obstacle instead of removing it?

maxime-clem: The main issue currently is that the code does not build (`pub_point_cloud_` is not declared).

Otherwise it is fine to leave this code but if it is heavy to run, rather than adding parameter I recommend having a condition on the number of subscribers to the debug topic.
```C++
if(debug_pub_->get_subscription_count() + debug_pub_->get_intra_process_subscription_count() > 0) {
  /* calculate and publish debug pointclould */
}
```

lchojnack: > The main issue currently is that the code does not build (pub_point_cloud_ is not declared).

I'm really sorry that the code does not build. I was checking the code several times, and I did not notice the issue. 

> Otherwise it is fine to leave this code but if it is heavy to run, rather than adding parameter I recommend having a condition on the number of subscribers to the debug topic.

Ok, It is definitely a better idea to publish point cloud depending on registered subscribers to the topic instead add another parameter. I will do it that way.

maxime-clem (approved):
The code looks good and the behavior in simulation is as expected.

xmfcx:
@lchojnack could you rebase and resolve the conflicts with the `planning/obstacle_stop_planner/src/node.cpp` please? Then we can proceed to merge it.

lchojnack:
@xmfcx I've resolved conflicts and I've also tested the implemented functionality. The feature is working as expected. I've tested it using planning simulator. 

I think that would be good to test the feature implemented in PR https://github.com/autowarefoundation/autoware.universe/pull/2647 in order to be sure that during resolving a conflict in `planning/obstacle_stop_planner/src/node.cpp`  the  `chattering prevention mechanism` wasn't broken.

xmfcx:
@satoshi-ota could you check if this PR breaks any functionality that was added in your previous PR https://github.com/autowarefoundation/autoware.universe/pull/2647 ?

planning/obstacle_stop_planner/src/node.cpp
```diff
@@ -385,16 +405,26 @@ void ObstacleStopPlannerNode::searchObstacle(
...
-      const auto found_collision_points = withinPolygon(
-        one_step_move_vehicle_polygon, stop_param.stop_search_radius, prev_center_point,
-        next_center_point, slow_down_pointcloud_ptr, collision_pointcloud_ptr);
+      const auto found_collision_points =
+        node_param_.enable_z_axis_obstacle_filtering
+          ? withinPolyhedron(
+              one_step_move_vehicle_polygon, stop_param.stop_search_radius, prev_center_point,
+              next_center_point, slow_down_pointcloud_ptr, collision_pointcloud_ptr, z_axis_min,
+              z_axis_max)
+          : withinPolygon(
```
satoshi-ota, line 425: @lchojnack @xmfcx  In my understanding, we can choose `withinPolyhedron` or `withinPolygon` by option `enable_z_axis_obstacle_filtering`. So, I think we should add `withinPolyhedron` here.

https://github.com/autowarefoundation/autoware.universe/blob/7ac34852ea37ec3084060004f505bd61698b65f0/planning/obstacle_stop_planner/src/node.cpp#L480-L483

lchojnack: @satoshi-ota Thank you for your review. I've applied the hint and I've added if `enable_z_axis_obstacle_filtering` option to another part of the code.

satoshi-ota: @lchojnack Yes, it is :+1: Thanks :smile:

satoshi-ota:
@lchojnack @xmfcx It looks good to me from the point of view of chattering functionality.

xmfcx (approved):
will resolve the last simple conflicts and merge

lchojnack:
@xmfcx Thank you for resolving the last simple conflicts. 

This is just a friendly reminder. This PR should be merged together with parameters PR https://github.com/autowarefoundation/autoware_launch/pull/150

brkay54:
Hello @lchojnack @maxime-clem , sorry to disturb you. I am working on z-axis filtering for predicted objects so I want to ask you that how to arrange the z-axis coordinate of the obstacle in planning simulator? I set the value in dummy objects but it is not working 😞

lchojnack:
Hello @brkay54 ,
I've just checked on the latest autoware, and the feature works as expected.

Please make sure:
1. [This](https://github.com/autowarefoundation/autoware_launch/blob/c1da7a1973c31cf8c719eddb9ad847385a04c67a/autoware_launch/config/planning/scenario_planning/lane_driving/motion_planning/obstacle_stop_planner/obstacle_stop_planner.param.yaml#L7) parameter is True 
2. There are green box markers on topic 
`/planning/scenario_planning/lane_driving/motion_planning/obstacle_stop_planner/debug/marker`
and collision point cloud on topic 
`/planning/scenario_planning/lane_driving/motion_planning/obstacle_stop_planner/debug/collision_pointcloud` as in the picture below
image

I've used the following command to test
`
ros2 launch autoware_launch planning_simulator.launch.xml map_path:=$HOME/autoware_map/elevation_test/ vehicle_model:=sample_vehicle sensor_model:=sample_sensor_kit
`

brkay54:
Thank you for your answer. If I understood correctly, you used awsim for these tests (setting dummy obstacle). I did not use awsim, just I am using planning simulator. 
image

When I set this value to higher, obstacle position on Z axis did not change. 

Also, if you don't mind, can you share your test map with me?

Thank you! @lchojnack

lchojnack:
@brkay54 
> If I understood correctly, you used awsim for these tests (setting dummy obstacle).

No, I use planning simulation for testing using the command
`ros2 launch autoware_launch planning_simulator.launch.xml map_path:=$HOME/autoware_map/elevation_test/ vehicle_model:=sample_vehicle sensor_model:=sample_sensor_kit`

 I just have ready to run 'clean' autoware (without any local changes) on `awsim-stable` branch hash `984e35c`. 
 I use `2D Dummy Bus` for testing, and sometimes I modify its `H vehicle height` as below
 
image

> Also, if you don't mind, can you share your test map with me?

Please use the map from [the tutorial](https://autowarefoundation.github.io/autoware-documentation/main/tutorials/ad-hoc-simulation/planning-simulation/). I've just tested and it also works as expected using command
`ros2 launch autoware_launch planning_simulator.launch.xml map_path:=$HOME/autoware_map/sample-map-planning vehicle_model:=sample_vehicle sensor_model:=sample_sensor_kit`

PR #2301, opened 2022-11-16
ci(labeler): fix labeler target

shmpwk:
Signed-off-by: Shumpei Wakabayashi <shumpei.wakabayashi@tier4.jp>

## Description
- Exclude `"**/*.txt"` from documentation label since this always indicates CMakeLists.txt and is also categorized as component label. 
- Change the label name from `document` to `documentation` to deply mkdocs.

### DIFF
--- a/.github/labeler.yaml
+++ b/.github/labeler.yaml
@@ -6,9 +6,8 @@ ci:
   - .clang-format
   - .gitignore
   - .prettierignore
-document:
+documentation:
   - docs/**/*
-  - "**/*.txt"
   - "**/*.md"
   - "**/*.rst"
   - "**/*.jpg"


.github/labeler.yaml
```diff
@@ -6,9 +6,8 @@ ci:
   - .clang-format
   - .gitignore
   - .prettierignore
-document:
+documentation:
```
kenji-miyake, line 9: If you add the `documentation` label, the `deploy-docs` workflow may run, which might become a security hole.
Do you think it's okay?

kenji-miyake: If you want to use the `documentation` label, I believe it's better to use a different `label` for `deploy-docs` to avoid unnecessary risks.
https://github.com/autowarefoundation/autoware.universe/blob/7e37a6c65a6a72ddf537ac673bf76cd3ff7ae581/.github/workflows/deploy-docs.yaml#L24

shmpwk: > If you want to use the documentation label, I believe it's better to use a different label for deploy-docs to avoid unnecessary risks.

Done by https://github.com/autowarefoundation/autoware.universe/pull/2302#pullrequestreview-1181967226

kenji-miyake (approved):

Issue #2044, opened 2022-10-11
Using a de-grounded point cloud for NDT matching gives a more reasonable match score


cyn-liu:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I've agreed with the maintainers that I can plan this task.

### Description

Autoware uses NDT matching for localization. The NDT input point cloud contains ground, which is less helpful for localization, and the ground is very similar in many locations, so the ground point cloud contributes a non-negligible score in NDT matching, resulting in a still high NDT score when localization is lost, so using a de-grounded point cloud in NDT matching for localization may achieve better results.

Using bag1 for point cloud matching, both the point cloud with ground and the point cloud without ground have high matching scores when the localization is successful.
The following are the test videos:
[bag1: Positioning success - NDT input point cloud with ground](https://www.bilibili.com/video/BV1Sg411Y7oc/?vd_source=1a0865e27a7bbd9e74e940639d66bbb7)
[bag1: Positioning success - NDT input without ground point cloud](https://www.bilibili.com/video/BV1Ve4y1q7Pt/?vd_source=1a0865e27a7bbd9e74e940639d66bbb7)
The following is a comparison of the NDT matching scores with and without ground point clouds:
![图片](https://user-images.githubusercontent.com/104069308/195001701-9260b2a8-bc69-42f8-90bf-fe799976d1d9.png)

Using bag2 for point cloud matching, both the point cloud with ground and the point cloud without ground have high matching scores when the localization is successful.
The following are the test videos:
[bag2: Positioning success - NDT input point cloud with ground](https://www.bilibili.com/video/BV1gT411N7QM/?vd_source=1a0865e27a7bbd9e74e940639d66bbb7)
[bag2: Positioning success - NDT input without ground point cloud](https://www.bilibili.com/video/BV1yG4y1H7wv/?vd_source=1a0865e27a7bbd9e74e940639d66bbb7)
The following is a comparison of the NDT matching scores with and without ground point clouds:
![图片](https://user-images.githubusercontent.com/104069308/195002017-1d064713-05f6-4b13-8e74-db23cb3d8a1f.png)
**This proves that de-grounded point clouds can achieve good localization results.**

Using bag2 for point cloud matching, the point cloud with ground scored high when the localization failed and the localization failure was not detected in time.
The following is the test video:
[...]

Issue #1375, opened 2022-07-19
Adding NDT reinitalization statement and reinitialization while moving


ataparlar:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I've agreed with the maintainers that I can plan this task.

### Description

NDT is initializes the position once in the beginning while vehicle is motionless. While moving if a topic used by autoware_launch stops streaming and starts streaming again, NDT reinitialization never starts. Eventually, vehicle starts to fly or starts turning around.
There is no reinitialization statement of NDT in the code. There must be a recovery mechanism for localization.

[Here](https://youtu.be/MffXnseMeg0) is a video of the issue.
[https://youtu.be/MffXnseMeg0](https://youtu.be/MffXnseMeg0) 

### Purpose

Making reinitalization of NDT correctly while moving.

### Possible approaches

Writing a reinitialization statement to NDT node. 

### Definition of done

This is done when NDT reinitializes itself correctly without flying or turning.

YamatoAndo:
I think the reinitalization is necessary.
However, I think it would be more versatile to put this function in a pose_initilizer node, etc., rather than in an NDT node, because it can be used even when the NDT node is not used.

I also think that this feature will be one of the most important features for Autoware, so we need to discuss the architecture in detail. @mitsudome-r @yukkysaito

cyn-liu:
I have encountered a similar problem of how to reinitialize the NDT when it failed to match and localization failed, the vehicle starts turning around.
The following figure is a screenshot of the simulation of the localization module when the car passes the intersection and the point cloud matching fails, the blue line in the figure is the trajectory of the car. 
![图片](https://user-images.githubusercontent.com/104069308/179902445-d5c216ea-13a1-410a-a71d-b7080d135253.png)

We believe that a reasonable solution in this case would be to first bring the vehicle to a slow stop, and then reinitialize the NDT.
However, we have the following questions:
1. what flags are used to monitor localization failure. (my idea is to determine whether the localization fails based on whether  score of the NDT match is greater than the threshold value of 2.3 )
2. How to make the car stop slowly when the localization fails.(the current localization error monitoring module does not notify other modules to take emergency action after the localization failure.)

isamu-takagi:
[...]

PR #2254, opened 2022-11-09
refactor(ndt_scan_matcher): modularize pose initialization

kminoda:
## Description
Aiming at integrating dynamic map loading feature in ndt_scan_matcher smoothly, I would like to modularize ndt_scan_matcher into the following modules:

- module that load pointcloud map (map_module) (Done)
- module that execute the pose initialization (initial_pose_module) (This PR)

## Pre-review checklist for the PR author

The PR author **must** check the checkboxes below when creating the PR.

- [x] I've confirmed the [contribution guidelines].
- [x] The PR follows the [pull request guidelines].

## In-review checklist for the PR reviewers

The PR reviewers **must** check the checkboxes below before approval.

- [ ] The PR follows the [pull request guidelines].

## Post-review checklist for the PR author

The PR author **must** check the checkboxes below before merging.

- [ ] There are no open discussions or they are tracked via tickets.

After all checkboxes are checked, anyone who has write access can merge the PR.

[contribution guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/
[pull request guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/pull-request-guidelines/

PR #2312, opened 2022-11-17
fix(ndt_scan_matcher): using de-ground points calculate ndt match score

cyn-liu:
Signed-off-by: Cynthia Liu <cynthia.liu@autocore.ai>
## Description

related issue: [2044](https://github.com/autowarefoundation/autoware.universe/issues/2044).
As discussed in issue 2044, using de-grounded point cloud in NDT matching for localization may achieve better results, so I have made some changes to the code:
- NDT matched positioning using the original point cloud(/localization/util/downsample/pointcloud).
- Calculates NDT match scores using de-grounded point cloud(/localization/util_no_ground/downsample/pointcloud).

In the ndt_scan_matcher node I subscribe to both topics in a callback function, one topic is used to locate and the other is used to calculate the scores(transform_probability & nearest_voxel_transformation_likelihood).

## Related links

[issue2044](https://github.com/autowarefoundation/autoware.universe/issues/2044).
[issue1375](https://github.com/autowarefoundation/autoware.universe/issues/1375)

## Tests performed

Test done.
Test results refer to the previous issue link.


launch/tier4_localization_launch/launch/localization.launch.xml
```diff
@@ -13,12 +14,26 @@
...
+      <include file="$(find-pkg-share tier4_localization_launch)/launch/util/util.launch.xml">
+        <arg name="input/pointcloud" value="$(var input/no_ground_pointcloud)"/>
+        <arg name="use_pointcloud_container" value="$(var use_pointcloud_container)"/>
+        <arg name="pointcloud_container_name" value="$(var pointcloud_container_name)"/>
+        <arg name="tier4_localization_launch_param_path" value="$(var tier4_localization_launch_param_path)"/>
+      </include>
+    </group>
+    <!-- remove ground module -->
+    <include file="$(find-pkg-share ground_segmentation)/launch/scan_ground_filter.launch.py">
+      <arg name="input/pointcloud" value="$(var input/orign_pointcloud)"/>
```
KYabuuchi, line 35: This `orign_pointcloud` is a mistake for `origin_pointcloud`, right?

KYabuuchi (commented):
Thank you for your cooperation!
I found a mistake that causes localization_launch to fail. Please check it.

localization/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp
```diff
@@ -393,7 +418,7 @@ void NDTScanMatcher::callback_sensor_points(
       transformation_msg_array, oscillation_threshold_, inversion_vector_threshold_);
   }
   bool is_ok_converged_param = validate_converged_param(
-    ndt_result.transform_probability, ndt_result.nearest_voxel_transformation_likelihood);
+    no_ground_transform_probability, no_ground_nearest_voxel_transformation_likelihood);
```
kminoda, line 421: We should leave `transform_probability` and `nearest_voxel_transformation_likelihood` as it is right now, since other Autoware users (including TIER IV) are still using them. I believe its better to leave it as an option whether to use `no_ground_transform_probability` so that the users can choose on their own.

localization/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp
```diff
@@ -164,9 +164,14 @@ NDTScanMatcher::NDTScanMatcher()
     "ekf_pose_with_covariance", 100,
     std::bind(&NDTScanMatcher::callback_initial_pose, this, std::placeholders::_1),
     initial_pose_sub_opt);
-  sensor_points_sub_ = this->create_subscription<sensor_msgs::msg::PointCloud2>(
-    "points_raw", rclcpp::SensorDataQoS().keep_last(points_queue_size),
-    std::bind(&NDTScanMatcher::callback_sensor_points, this, std::placeholders::_1), main_sub_opt);
+  sensor_points_sub_ = std::make_shared<message_filters::Subscriber<sensor_msgs::msg::PointCloud2>>(
```
kminoda, line 167: (Not related to the code review!)

Do you have any evaluation results on `no_ground_*` scores? I understood that the data posted on the previous issue (https://github.com/autowarefoundation/autoware.universe/issues/2044) is a bit different from this PR's algorithm (since it is performing NDT using scans without ground removal). It would be great if you would share with us some of your results to show that this change is reasonable enough!

IMHO posting and discussing the evaluation report in this PR is a bit inconvenient, so I propose to open a new [GitHub Discussion](https://github.com/orgs/autowarefoundation/discussions) for this topic :smile:

kminoda (commented):
Thank you for your contribution!! Basically, I agree that the score with ground removal is worth considering as an alternative scan matching performance indicator.
I left some general comments.

launch/tier4_localization_launch/launch/localization.launch.xml
cyn-liu: Thanks

localization/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp
cyn-liu: share evaluation results, [discussion3056](https://github.com/orgs/autowarefoundation/discussions/3056)

kminoda: Thank you!

localization/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp
```diff
@@ -308,8 +319,9 @@ void NDTScanMatcher::callback_regularization_pose(
   regularization_pose_msg_ptr_array_.push_back(pose_conv_msg_ptr);
 }
 
-void NDTScanMatcher::callback_sensor_points(
-  sensor_msgs::msg::PointCloud2::ConstSharedPtr sensor_points_sensorTF_msg_ptr)
+void NDTScanMatcher::callback_points(
```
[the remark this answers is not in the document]
kminoda: How about making another module like `map_module` or `pose_initialization_module`? e.g. a module that only have the functionality of
- subscribe de-grounded scans
- calculate the proposed score and publish

However, separating the subscription may be tricky in terms of synchronization (since it may not make sense to use de-grounded scans whose timestamps are not close enough to the original scans that is used in NDT scan matching). I would also like to here opinions from others, if you have any. @KYabuuchi @YamatoAndo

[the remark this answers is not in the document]
YamatoAndo: What about implementing a ground filter in the ndt_scan_matcher node?
For example, how about considering LiDAR point clouds with z-coordinate values below a certain value as ground point clouds?
I don't think an exact ground filter is necessary, so I think it is no problem to use an approximated ground filter.
This would also reduce the computational cost and solve the synchronization problem since there is no need to subscribe to a de-grounded scan.

(As an aside, it would also be interesting to look at the ND voxels generated from the map point cloud and extract the ground area to classficate the lidar point cloud.  Another good idea would be to use RANSAC.)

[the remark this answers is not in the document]
cyn-liu: I already make rough modifications to the code in order to ensure that both mechanisms work as follows.

1. Localization and calculate scores only using original point cloud.
2. Use the original point cloud for localization and use the de-grounded point cloud to calculate the score

The way it is to use a flag to determine whether to generate and use a de-grounded point cloud. If the flag is false, NDT matches using the original `callback_sensor_points` function, and if the flag is true, the `callback_points function` (I defined) is used and **both point clouds must be received simultaneously**.

The disadvantage of this modification is that two similar callback functions are defined and there is a lot of code duplication.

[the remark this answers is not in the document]
cyn-liu: I think my modification method is bad.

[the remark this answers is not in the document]
cyn-liu: @kminoda  Thanks your advice, but synchronization are very important.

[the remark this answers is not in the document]
cyn-liu: @YamatoAndo Thanks, good idea, I forgot this method.
The first idea was to use the ground-removed point cloud for localization and score calculation, so to get a good de-grounded point cloud, the **raw point cloud** is processed in `localization.xml`.
Now just using the de-grounded point cloud to calculate the score and the ground should be removed directly in the NDT node(for example use the RANSAC method) .


However, there is a potential risk that using **pre-processed point cloud** for de-grounding may not yield satisfactory results, and may not reflect the advantage of de-grounded point clouds in the NDT matching calculation score, so the experiment must be redone to compare the score difference between the two point clouds.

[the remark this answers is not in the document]
kminoda: Thank you! I also think @.YamatoAndo's idea makes sense. Maybe we can use even more simple method than RANSAC (for example, removing all the points that satisfies `point.z < base_link.z + vehicle_height` ). 

It is true that the validation must be done since using pre-process pcd for degrounding may increase the risk as you've mentioned. But IMO we can merge the proposed score with some quick evaluation so that the Autoware users can optionally use it and check if the score is useful enough.

KYabuuchi (commented):
It looks like the modification was complete, so I added some comments. 
Please check them. :smile:

localization/ndt_scan_matcher/include/ndt_scan_matcher/ndt_scan_matcher_core.hpp
```diff
@@ -134,6 +137,7 @@ class NDTScanMatcher : public rclcpp::Node
     regularization_pose_sub_;
 
   rclcpp::Publisher<sensor_msgs::msg::PointCloud2>::SharedPtr sensor_aligned_pose_pub_;
+  rclcpp::Publisher<sensor_msgs::msg::PointCloud2>::SharedPtr no_ground_aligned_pose_pub_;
```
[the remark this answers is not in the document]
KYabuuchi: Sorry. I misinterpreted the variable name.  Perhaps you named it based on `sensor_aligned_pose_pub_`, right? 
It is hard to know what it publishes, so it would be better to include "points" in the publisher name.

localization/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp
```diff
@@ -546,6 +587,26 @@ void NDTScanMatcher::publish_initial_to_result_distances(
     make_float32_stamped(sensor_ros_time, initial_to_result_distance_new));
 }
 
+void NDTScanMatcher::remove_ground_point_cloud(
+  const pcl::shared_ptr<pcl::PointCloud<PointSource>> & sensor_points_mapTF_ptr,
+  const pcl::shared_ptr<pcl::PointCloud<PointSource>> & no_ground_points_mapTF_ptr)
```
kminoda, line 592:
```suggestion
  const pcl::shared_ptr<pcl::PointCloud<PointSource>> & no_ground_points_mapTF_ptr) const
```
I suggest adding const for this function

(or maybe you can make this as a separate function in `utils.cpp`, since it does not seem necessary to include it in NDTScanMatcher class.)

launch/tier4_localization_launch/config/ndt_scan_matcher.param.yaml
```diff
@@ -65,3 +65,9 @@
 
     # regularization scale factor
     regularization_scale_factor: 0.01
+
+    # whether use de-grounded pointcloud calculate score
```
kminoda, line 69:
```suggestion
    # A flag for using scan matching score based on de-grounded LiDAR scan
```

localization/ndt_scan_matcher/include/ndt_scan_matcher/ndt_scan_matcher_core.hpp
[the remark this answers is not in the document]
cyn-liu: OK! thanks.

launch/tier4_localization_launch/config/ndt_scan_matcher.param.yaml
```diff
@@ -65,3 +65,9 @@
...
+    use_no_ground_pointcloud: true
```
kminoda, line 70: `use_no_ground_pointcloud` may be tricky for beginners. (For example they might take it as a flag for using degrounded scans for NDT scan matching)
How about something like as follow?
- `estimate_scores_for_degrounded_scan`
- `calculate_degrounded_scan_matching_score`

YamatoAndo:
@cyn-liu Thank you for the update!
I have one question I would like to discuss.
If the ground is not included in the LiDAR point cloud, there is a risk that RANSAC will detect building walls, etc.
So, I think it would be better to simply filter by z-coordinate values.
What do you think?

localization/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp
cyn-liu: Thank you very much.
If I remove the ground directly using Z-axis information, I will remove this function.
Now I am considering how to modify the algorithm for removing the ground.

cyn-liu:
> @cyn-liu Thank you for the update! I have one question I would like to discuss. If the ground is not included in the LiDAR point cloud, there is a risk that RANSAC will detect building walls, etc. So, I think it would be better to simply filter by z-coordinate values. What do you think?

Thanks.
Removing the ground using the plane detection algorithm (for example RANSAC) is convenient for the user, who does not need to pay attention to the parameter configuration, because the ground information is automatically detected and filtered by the algorithm, but the accuracy of the algorithm cannot be guaranteed.

Using Z-axis information to remove the ground is the most efficient and accurate method, but it requires us to configure the vehicle height information and lidar to baselink TF information correctly, if the user doesn't configure it, the result will be very bad.

I will update the program to remove the ground by filtering the point cloud with `z <base_link.z + vehicle_height`.

cyn-liu:
@YamatoAndo @kminoda Hello!

I have some questions:
I think filtering the ground point cloud by z coordinate uses `z<base_link.z + wheel_radius` instead of `z<base_link.z + vehicle_height`.
What do you think?

launch/tier4_localization_launch/config/ndt_scan_matcher.param.yaml
kminoda: @cyn-liu This is just a little comment, but instead of resolving the conversation by yourself, would you post a commit which addresses the suggestion in a thread? It would be much easier for the reviewer to check if the issue is correctly addressed :smile: 
e.g.: https://github.com/autowarefoundation/autoware.universe/pull/2254#discussion_r1018578029

kminoda:
@cyn-liu Both would make sense. Your suggestion also seems good.
Ultimately, it depends on what kind of ground we would take into account. `z < base_link.z + wheel_radius` would be better to leave more non-ground points, while it may fail to remove ground points when the ground is not flat enough along z-axis (for example a steep slope or a valley-like road). However, at this right moment, I think you can just go for your idea. We can modify  it if the issue arises.

YamatoAndo:
Instead of using the values of vehicle_height and wheel_radius, we should create a new parameter for this.

cyn-liu:
> Instead of using the values of vehicle_height and wheel_radius, we should create a new parameter for this.


You mean to design a parameter to remove the ground and this parameter does not use either vehicle information or TF information from lidar to base_link, and only manually sets a value through experiments???

YamatoAndo:
> You mean to design a parameter to remove the ground and this parameter does not use either vehicle information or TF information from lidar to base_link, and only manually sets a value through experiments???

Yes. 
For example, `(lidar_point.z in map_frame) - (base_link.z in map_frame)  < (threshold)`.
And the threshold is determined experimentally.

cyn-liu:
> Yes. For example, `(lidar_point.z in map_frame) - (base_link.z in map_frame) < (threshold)`. And the threshold is determined experimentally.

Why not simplify it to `lidar_point. z in map_frame<threshold` ?

kminoda:
@cyn-liu 
I think `lidar_point. z in map_frame < threshold` is not a good idea since we cannot tell the actual altitude of the ground beforehand. (Do you mean `lidar_point. z in base_link < threshold`)

cyn-liu:
@kminoda 
Yes, I mean `lidar_point. z in base_link < threshold` or `lidar_point. z in sensor_link < threshold`.

kminoda:
@cyn-liu I think both are fine, given that the two statements
- `(lidar_point.z in map_frame) - (base_link.z in map_frame) < (threshold)`
- `lidar_point. z in base_link < threshold`
are mathematically the same. 

I am not sure about `lidar_point. z in sensor_link < threshold` though, since the height of lidar with respect to the base_link may vary.

launch/tier4_localization_launch/config/ndt_scan_matcher.param.yaml
```diff
@@ -65,3 +65,9 @@
 
     # regularization scale factor
     regularization_scale_factor: 0.01
+
+    # A flag for using scan matching score based on de-grounded LiDAR scan
+    estimate_scores_for_degrounded_scan: false
+
+    # If lidar_point.z - base_link.z <= this threshold , will be removed
+    ground_removal_z_threshold: 0.8
```
YamatoAndo, line 73: If the threshold is 0.8m, it seems to include objects other than the ground.
So I think it would be better to change the word to mean something of low height instead of saying ground.

YamatoAndo: I'm not a native English speaker, so I can't think of a good word for it.

cyn-liu: I just want to remove the ground, I found that if this threshold is set too small can not remove the ground.

kminoda: I think what he meant was not about the value but about the variable name. Still cannot come up with a right fit vocabulary, but how about something like this?
- `margin_from_base_link`
- `z_margin_for_ground_removal`

localization/ndt_scan_matcher/include/ndt_scan_matcher/ndt_scan_matcher_core.hpp
```diff
@@ -195,6 +200,9 @@ class NDTScanMatcher : public rclcpp::Node
   std::shared_ptr<Tf2ListenerModule> tf2_listener_module_;
   std::unique_ptr<MapModule> map_module_;
   std::unique_ptr<PoseInitializationModule> pose_init_module_;
+
+  bool estimate_scores_for_degrounded_scan_;
+  double z_margin_for_ground_removal_ = 1.0;
```
kminoda, line 205:
```suggestion
  double z_margin_for_ground_removal_;
```
I think it is better to avoid providing default value here

cyn-liu:
@kminoda Hello!
There is always a problem with my PR signature.
I add a signature every time I push, but I don't know why it still doesn't meet the requirements.
Can you help me solve this problem? Thank you very much.

```console
git commit -s -m "fix(ndt_scan_matcher): using de-ground points calculate ndt match score"
git push -f origin use-deground-points-calculate-ndt-match-score
```
My `git log`:
MicrosoftTeams-image

kminoda:
@cyn-liu I think it's due to this commit: https://github.com/autowarefoundation/autoware.universe/pull/2312/commits/ce5e6335abc69b1240a8696543dbb1b145c32d9a
I manually set `DCO` CI to pass for now.

kminoda (approved):
Sorry for being late. LGTM

YamatoAndo (approved):

mitsudome-r (commented):
Please update the README for the added parameters. https://github.com/autowarefoundation/autoware.universe/tree/5d5f091f035ac1efce07baed6a9fd80ab7b876f9/localization/ndt_scan_matcher

cyn-liu:
I modify the `README.md`  in `ndt_scan_matcher` package.

kminoda:
@cyn-liu Since the parameter management configuration has been changed, please modify launch/tier4_localization_launch/config/ndt_scan_matcher.param.yaml
 instead of `launch/tier4_localization_launch/config/ndt_scan_matcher.param.yaml`.

cyn-liu:
@kminoda Modified.

liuXinGangChina:
Hi,Mitsudome-san. Since Yamato-san and kminoda-san have approved this pr, will you help us to merge it?

cyn-liu:
> @cyn-liu Since the parameter management configuration has been changed, please modify launch/tier4_localization_launch/config/ndt_scan_matcher.param.yaml instead of `launch/tier4_localization_launch/config/ndt_scan_matcher.param.yaml`.
 
Hello @kminoda , I have a question:
The file `launch/tier4_localization_launch/config/ndt_scan_matcher.param.yaml` has been transferred to  `autoware_launch/config/localization/ndt_scan_matcher.param.yaml` , and I have updated this change in `autoware.universe` repo, If I want to add my parameters in `autoware_launch` repo, should I need create a new PR in this repo?

kminoda:
@cyn-liu Yes, your understanding is correct! 
(We are also considering adding a workflow based on GitHub Actions that syncs parameters in `autoware.universe` and `autoware_launch`, but for now, please update both parameters manually :pray: )

kminoda:
@liuXinGangChina It seems that one CI (DCO) failed, and it kept this PR from being able to merge.
Would you check it out again? I think now this PR can be merged (I set the DCO to pass)

kminoda:
I've also merged this PR https://github.com/autowarefoundation/autoware_launch/pull/155

PR #1276, opened 2022-07-07
feat(pointcloud_preprocessor): use point_cloud_msg_wrapper

esteve:
## Description

This PR replaces the standard point cloud iterators from the `sensor_msgs` package with [Point Cloud Message Wrapper](https://gitlab.com/ApexAI/point_cloud_msg_wrapper)

## Related links

#1199

## Pre-review checklist for the PR author

The PR author **must** check the checkboxes below when creating the PR.

- [x] I've confirmed the [contribution guidelines].
- [x] The PR follows the [pull request guidelines].

## In-review checklist for the PR reviewers

The PR reviewers **must** check the checkboxes below before approval.

- [x] The PR follows the [pull request guidelines].
- [x] The PR has been properly tested.
- [x] The PR has been reviewed by the code owners.

## Post-review checklist for the PR author

The PR author **must** check the checkboxes below before merging.

- [ ] There are no open discussions or they are tracked via tickets.
- [ ] The PR is ready for merge.

After all checkboxes are checked, anyone who has write access can merge the PR.

[contribution guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/
[pull request guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/pull-request-guidelines/

Issue #2306, opened 2022-11-16
Distortion correction would not work using point_cloud_msg_wrapper


miursh:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I'm convinced that this is not my fault but a bug.

### Description

Distortion correction node does not compensate ego vehicle motion.

### Expected behavior

The xy position of `/sensing/lidar/top/rectified/pointcloud_ex` must have some offset based on its vehicle velocity like this image below.
(Yellow: rectified/pointcloud_ex, Dark gray: and pointcloud_raw)

![image](https://user-images.githubusercontent.com/37187849/202206503-04392861-4521-4894-8cdc-f2377137c4c6.png)

### Actual behavior

Distortion correction node outputs warning on the terminal.
```
[component_container-8] [WARN 1668606722.543281569] [sensing.lidar.rear.distortion_corrector_node]: Required field time stamp doesn't exist in the point cloud.
[component_container-6] [WARN 1668606722.947966127] [sensing.lidar.left.distortion_corrector_node]: Required field time stamp doesn't exist in the point cloud.
[component_container-7] [WARN 1668606722.950508348] [sensing.lidar.right.distortion_corrector_node]: Required field time stamp doesn't exist in the point cloud.
[component_container_mt-5] [WARN 1668606723.075554512] [sensing.lidar.top.distortion_corrector_node]: Required field time stamp doesn't exist in the point cloud.
```

The xy position of `/sensing/lidar/top/rectified/pointcloud_ex` is exact same as `/sensing/lidar/top/pointcloud_raw_ex`. 
This image shows rectified/pointcloud_ex in yellow, and pointcloud_raw in dark gray.

![image](https://user-images.githubusercontent.com/37187849/202204658-ae8a93de-7764-41b8-a105-fbef3bdaf360.png)

### Steps to reproduce

1. run logging_simulator.launch.xml with certain args.
2. play sample rosbag

miursh:
I think because of [this PR](https://github.com/autowarefoundation/autoware.universe/pull/1276), some pointcloud preprocessor does not work as expected.
1. Distortion corrector could not parse timestamp field.
2. If concatenate_data subscribes pointcloud topic other than XYZI type, the crashes with runtime error.

@esteve Have you try one of those?

yukkysaito:
[...]

PR #2317, opened 2022-11-18
revert: "feat(pointcloud_preprocessor): use point_cloud_msg_wrapper"

miursh:
Reverts autowarefoundation/autoware.universe#1276

### DIFF
--- a/common/autoware_auto_common/include/common/types.hpp
+++ b/common/autoware_auto_common/include/common/types.hpp
@@ -109,46 +109,6 @@ struct COMMON_PUBLIC PointXYZI
   }
 };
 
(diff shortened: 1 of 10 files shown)


yukkysaito (approved):
LGTM

xmfcx:
What is the reason for reverting this change?

xmfcx:
I see it's for https://github.com/autowarefoundation/autoware.universe/issues/2306

xmfcx:
I think it's ok, it's better to add point_cloud_msg_wrapper feature once we finalize the point types architecture in [here](https://github.com/autowarefoundation/autoware-documentation/pull/180) and allow different point types too.

Issue #2252, opened 2022-11-08
Autoware does not check for collision on the vehicle actual trajectory


VRichardJP:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I'm convinced that this is not my fault but a bug.

### Description

Today, I observed autoware does not want to stop for an obstacle right in front of the ego vehicle if everything happens outside the path.

I guess a screenshot is worth a thousand words (situation reproduced with the planning simulator):
![Screenshot from 2022-11-08 20-22-23](https://user-images.githubusercontent.com/18645627/200577835-b38e284a-a802-4f4d-9f48-19fc43de3935.png)

R.I.P
![Screenshot from 2022-11-08 20-26-36](https://user-images.githubusercontent.com/18645627/200577825-639a9ce0-dc08-46d1-b9dd-4a1547e584c5.png)

### Expected behavior

Autoware should not hit obstacles on its trajectory

### Actual behavior

Better not be a human in a simulation.

### Steps to reproduce

See the screenshot

brkay54:
You can enable [obstacle_collision_checker](https://github.com/autowarefoundation/autoware_launch/blob/main/autoware_launch/launch/planning_simulator.launch.xml#L13) to check collision on `predicted_trajectory`. It should solve your problem.

VRichardJP:
Wow! That's life saving!

I'll try it.

VRichardJP:
If I understand correctly, the node should create a diagnostic error if there is something on the predicted trajectory? Then with the right configuration, autoware should trigger emergency when this happens, correct?

I have backported the latest `control.launch.py` to my tree to add the `obstacle_collision_checker`. When I start autoware, I can see the module is running and is generating diagnostics. 

I have reproduced my first screenshot scenario with a golf cart and traffic cones. However, in that situation, the diagnostic is always "OK". Even when the cone is right in front of the car (and detected as obstacle) the collision checker still generates "OK" diagnostics.

brkay54:
Hmm I see, I tried too but module created "ERROR" diagnostic if there is an object in braking distance. But after vehicle stops, you will see "OK" because object is not in braking distance of vehicle (because velocity of vehicle is 0). 

[...]

PR #2341, opened 2022-11-21
feat(obstacle_stop_planner): add optional check for predicted trajectory

brkay54:
Signed-off-by: Berkay Karaman <brkay54@gmail.com>

## Description
Related issue: https://github.com/autowarefoundation/autoware.universe/issues/2252

In current planning control design, autoware does not check the predicted trajectory for collision. Because trajectory is not draw from base_link of ego vehicle for some situations, we need to check the predicted trajectory in addition to trajectory output of planning module. 

This PR checks the predicted trajectory and motion planning trajectory for collision, then compare them. It makes decision for the nearest collision or slowdown points. 

It is optional, in default it will be disabled.

## Tests performed

I tested changes in planning simulator. It works good but there is a problem for some scenarios:

https://user-images.githubusercontent.com/45468306/203044604-08a24063-2cf5-4ad1-9136-b5891f482539.mp4

Controllers depend trajectory velocities while calculating the predicted trajectory. If there is a object on predicted trajectory, obstacle_stop_planner adds stop point to trajectory. Then, this affect the predicted trajectory so on.

Except this, it works as we wanted.

### DIFF
--- a/launch/tier4_planning_launch/config/scenario_planning/lane_driving/motion_planning/obstacle_stop_planner/obstacle_stop_planner.param.yaml
+++ b/launch/tier4_planning_launch/config/scenario_planning/lane_driving/motion_planning/obstacle_stop_planner/obstacle_stop_planner.param.yaml
@@ -4,6 +4,7 @@
     lowpass_gain: 0.9                        # gain parameter for low pass filter [-]
     max_velocity: 20.0                       # max velocity [m/s]
     enable_slow_down: False                  # whether to use slow down planner [-]
+    check_predicted_trajectory: False        # if it is "True", planner will take into account predicted trajectory of controller [-]
 
     stop_planner:
       # params for stop position
--- a/launch/tier4_planning_launch/launch/scenario_planning/lane_driving/motion_planning/motion_planning.launch.py
+++ b/launch/tier4_planning_launch/launch/scenario_planning/lane_driving/motion_planning/motion_planning.launch.py
@@ -202,6 +202,10 @@ def launch_setup(context, *args, **kwargs):
             ("~/input/objects", "/perception/object_recognition/objects"),
             ("~/input/odometry", "/localization/kinematic_state"),
             ("~/input/trajectory", "obstacle_velocity_limiter/trajectory"),
+            (
+                "~/input/predicted_trajectory",
+                "/control/trajectory_follower/lateral/predicted_trajectory",
+            ),
         ],
         parameters=[
             nearest_search_param,
--- a/planning/obstacle_stop_planner/README.md
+++ b/planning/obstacle_stop_planner/README.md
@@ -13,14 +13,15 @@
 
 ### Input topics
 
-| Name                        | Type                                            | Description         |
-| --------------------------- | ----------------------------------------------- | ------------------- |
-| `~/input/pointcloud`        | sensor_msgs::PointCloud2                        | obstacle pointcloud |
-| `~/input/trajectory`        | autoware_auto_planning_msgs::Trajectory         | trajectory          |
-| `~/input/vector_map`        | autoware_auto_mapping_msgs::HADMapBin           | vector map          |
-| `~/input/odometry`          | nav_msgs::Odometry                              | vehicle velocity    |
-| `~/input/dynamic_objects`   | autoware_auto_perception_msgs::PredictedObjects | dynamic objects     |
-| `~/input/expand_stop_range` | tier4_planning_msgs::msg::ExpandStopRange       | expand stop range   |
+| Name                           | Type                                            | Description          |
+| ------------------------------ | ----------------------------------------------- | -------------------- |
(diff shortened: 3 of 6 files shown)


satoshi-ota:
@brkay54 Thanks for your PR, but I would rather not feedback control outputs into planning module for the following reason:

- Planning module and Control module would be tightly coupled, and decrease modularity.
- Not all vehicles use MPC and this method cannot be used when using pure pursuite.

On the other hand, other solution is proposed in [discussion](https://github.com/autowarefoundation/autoware/discussions/2859), so could you check that?

TakaHoribe:
@brkay54 Thank you for this PR. 

We must be careful about adding this functionality: return predicted path information computed in control to the planning module. First, this PR will significantly change the interface between modules. It also changes the role of each module. Additionally, having FB loop between modules also brings up complex issues, exactly as in the video you posted.

There are reasons for separating control and planning, mainly to maintain high modularity as an OSS. However some performance is sacrificed by separating the modules, for example, path planning has to be done without consideration of control performance.

However, it is clear that we want to do the stop planning from the actual ego position. [A solution using object_collision_checke](https://github.com/autowarefoundation/autoware.universe/issues/2252)r is proposed in the issue, but is this not enough? Using this module will not degrade modularity and will allow stop planning for the predictive path output by the control while maintaining the existing interfaces between the modules.

In addition, we can suggest that instead of getting the predicted path from control, the predicted path be calculated in the planning stage ([It was suggested before](https://github.com/autowarefoundation/autoware/discussions/2859#discussioncomment-3714275)). There are several calculation methods. For instance,
- To integrate the vehicle's speed and angular velocity over a certain period of time (the simplest one).
- To create a model that converges to the target path from ego (e.g., lateral deviation decreases at a constant rate).
- The duplicated MPC (or the same controller used in the control module) runs in the planning module (if the computational resources are available).


What do you think?

@xmfcx @mitsudome-r If you have any comments.

brkay54:
Thank you for your comment @TakaHoribe san.
 

> However, it is clear that we want to do the stop planning from the actual ego position. [A solution using object_collision_checke](https://github.com/autowarefoundation/autoware.universe/issues/2252)r is proposed in the issue, but is this not enough? Using this module will not degrade modularity and will allow stop planning for the predictive path output by the control while maintaining the existing interfaces between the modules.
> 
Actually, it can not solve our problem exactly. `obstacle_collision_checker` just send emergency if the object inside of braking distance. If `emergency_hold` not true, It will cause unwanted behavior (like move - stop - move - stop and go on).

> In addition, we can suggest that instead of getting the predicted path from control, the predicted path be calculated in the planning stage ([It was suggested before](https://github.com/autowarefoundation/autoware/discussions/2859#discussioncomment-3714275)). There are several calculation methods. For instance,
> 
>     * To integrate the vehicle's speed and angular velocity over a certain period of time (the simplest one).
> 
>     * To create a model that converges to the target path from ego (e.g., lateral deviation decreases at a constant rate).
> 
>     * The duplicated MPC (or the same controller used in the control module) runs in the planning module (if the computational resources are available).
> 

Of course we can calculate an estimated path which start from ego base_link, however all of these methods will produce worse estimated path than controller produce. And also these methods cause a lot of false positive collisions (because we will use just current state of vehicle for prediction). Therefore, in my opinion, the most reliable predicted path will be output of controller.

> We must be careful about adding this functionality: return predicted path information computed in control to the planning module. First, this PR will significantly change the interface between modules. It also changes the role of each module. Additionally, having FB loop between modules also brings up complex issues, exactly as in the video you posted.
> 
> There are reasons for separating control and planning, mainly to maintain high modularity as an OSS. However some performance is sacrificed by separating the modules, for example, path planning has to be done without consideration of control performance.

I agree with you, separating the control and planning module is better to avoid this type conflicts. But if one module of planning or control will do collision checking in vehicle's actual path, it shouldn't be planning because control has better and more reliable estimated path and we should use it.

Alternative to this PR, I want to propose that we can add new functionalities to `obstacle_collision_checker`. It is currently checking obstacles in path for emergency stop and also it can be used for smooth stop too. It can communicate with the `vehicle_cmd_gate` to stop the vehicle if there is an obstacle on footprints of predicted path. For example it can use [pause_interface](https://github.com/autowarefoundation/autoware.universe/blob/main/control/vehicle_cmd_gate/src/vehicle_cmd_gate.cpp#L422) to stop vehicle or it can calculate the stop accelerations and send them to vehicle_cmd_gate etc. We can design the structure of how we stop. 

Of course it will cause conflict between `obstacle_collision_checker` and `obstacle_stop_planner`. To avoid this, `obstacle_collision_checker` check only some part of predicted_path that is separated from motion_planning trajectory output.

With these, we can separate the modules and we have more reliable predicted path for collision. What do you think about it?

yukkysaito:
Can we discuss it [here](https://github.com/orgs/autowarefoundation/discussions/2859) since it will be an architecture issue?

IMO : https://github.com/orgs/autowarefoundation/discussions/2859#discussioncomment-4235577

Issue #2310, opened 2022-11-17
The  behavior_path_planner can not plan the right path in some positions as the target


shulanbushangshu:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I'm convinced that this is not my fault but a bug.

### Description

When I set the different positions of target in the lane,the behavior_path_planner can not plan the right path in some positions.
case 1:

![lane_follow_normal](https://user-images.githubusercontent.com/102840938/202373400-c4fa7907-9b87-444b-962c-3652d2fd65ab.png)

case 2:

![Screenshot from 2022-11-16 16-27-25](https://user-images.githubusercontent.com/102840938/202373503-010ee0ac-ee9c-43d4-9314-09614c48402b.png)

case 3:
![Screenshot from 2022-11-17 14-35-22](https://user-images.githubusercontent.com/102840938/202374493-e4c7f533-d48c-4635-b272-c5c68621420e.png)

case 4:

![Screenshot from 2022-11-17 14-37-25](https://user-images.githubusercontent.com/102840938/202374591-db59890e-5f00-4547-a91d-fec04fabd627.png)

### Expected behavior

The behavior_path_planner shall plan the right path in all positions of the lanes.

### Actual behavior

The behavior_path_planner can not plan the right path when I set some different positions  as the target in the lanes.

### Steps to reproduce

1.Start autoware planning simulator;
2. Set initial position 
3. Set target in the different positions in the lane(far from the initial position)

### Versions

1.OS: Ubuntun 20.04
2.ROS2: Galactic
3.Autoware: Autoware.universe:main

ismetatabay:
Hi, can you please share map (pcd and lanelet) files? The lanelets which highlighted as I have shown below picture show that the route has been successfully generated by the mission_planner, I didn't see any problem in this case. If you're worried about not seeing the green path in rviz, the path will move forward as the vehicle move.  The parameter that is ```forward_path_length``` limits the path . You can also increase the generated path by changing this parameter.

https://github.com/autowarefoundation/autoware.universe/blob/29e8b435a4416e5db0eb6c7da0c4b0457e6c529b/launch/tier4_planning_launch/config/scenario_planning/lane_driving/behavior_planning/behavior_path_planner/behavior_path_planner.param.yaml#L4

[...]

PR #2351, opened 2022-11-22
fix(behavior_path_planner): fix-lane-follow-in-behavior-path-planner

shulanbushangshu:
Signed-off-by: jack.song <jack.song@autocore.ai>

## Description
When the global route includes the changing lane,the function of "checkLaneIsInIntersection" will return true(because the distance is not enough by calculation) by the current code in some positions as the target and the path by "lane_following" will be wrong. But the distance of the lanes is actual enough.
The simulation is shown as below:

image

image

## Related links
https://github.com/autowarefoundation/autoware.universe/issues/2310

## Tests performed
By changing the judgment of the "checkLaneIsInIntersection",the simulation is shown as below:
case1:

lane_fowwing

case2:

lane_follow_2

### DIFF
--- a/planning/behavior_path_planner/src/utilities.cpp
+++ b/planning/behavior_path_planner/src/utilities.cpp
@@ -1830,7 +1830,7 @@ bool checkLaneIsInIntersection(
       return false;
     }
 
-    auto back_lanelet_length = lanelet::utils::getLaneletLength2d(lanelet_sequence.back());
+    const auto back_lanelet_length = lanelet::utils::getLaneletLength2d(lanelet_sequence.back());
     if (back_lanelet_length > min_lane_change_distance) 
     {
       return false;
--- a/planning/behavior_path_planner/src/utilities.cpp
+++ b/planning/behavior_path_planner/src/utilities.cpp
@@ -1830,8 +1830,8 @@ bool checkLaneIsInIntersection(
       return false;
     }
 
-    auto backlanelet_length = lanelet::utils::getLaneletLength2d(lanelet_sequence.back());
-    if (backlanelet_length > min_lane_change_distance) 
+    auto back_lanelet_length = lanelet::utils::getLaneletLength2d(lanelet_sequence.back());
+    if (back_lanelet_length > min_lane_change_distance) 
     {
       return false;
     }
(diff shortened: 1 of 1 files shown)


planning/behavior_path_planner/src/utilities.cpp
```diff
@@ -1829,6 +1829,11 @@ bool checkLaneIsInIntersection(
     if (get_signed_distance > min_lane_change_distance) {
       return false;
     }
+
+    auto back_lanelet_length = lanelet::utils::getLaneletLength2d(lanelet_sequence.back());
```
zulfaqar-azmi-t4, line 1833: Should be `const` here.

zulfaqar-azmi-t4:
@shulanbushangshu 

Thank you very much for your contribution!

I think the solution should be fine, however, at the moment, I am not sure why the result differs depending on the location of the goal. I've check #2310 and the discussion seems to be on going.

Based on your current analysis, may i ask your help in explaining the reason?

shulanbushangshu:
@zulfaqar-azmi-t4 
I show the process :
1.When I set the target position near the start of the lane and the start position is near the intersection,the goal route by mission_planner includes has the change lane in my map.
2.In the lane_following_module, the checkLaneIsInIntersection() is used in the  getReferencePath().
By these conditions, the checkLaneIsInIntersection() will get the true as the result(by current codes).
3.And checkLaneIsInIntersection() will caculate the "additional_length_to_add"(it is a big value by current codes) and the length will be used in the calculation of getCenterLinePath().
4.In the getCenterLinePath(), the isDeadEndLanelet() is used and the result is true because of the change lane.
The s_forward will be calculated by "std::min(s_forward, lane_length - lane_change_buffer);".Because of the  lane_change_buffer(is the additional_length_to_add),the  s_forward will small than the s_backward.So the route_handler.getCenterLinePath() will calculate  the wrong path.

So I think that the checkLaneIsInIntersection() has the problem in judging the route with change lane by just using the goal position and last lane.





intersection_sforword

intersection_length_2

liuXinGangChina:
Hi,Mitsudome-san. @mitsudome-r 
Please help us to accelarate the viewing process.

Thanks,
Lucas

xmfcx:
@zulfaqar-azmi-t4 could you review again?

mitsudome-r:
@shulanbushangshu Is it possible to share the map so that I can also check the issue on my machine?

zulfaqar-azmi-t4 (approved):
Sorry I have overlooked this.
LGTM

zulfaqar-azmi-t4:
@shulanbushangshu 
I shall close this PR because the function is no longer used in the current version.

Issue #2354, opened 2022-11-23
Refactor mission planning to support when is goal behind start on the same lane


VRichardJP:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I've agreed with the maintainers that I can plan this task.

### Description

As pointed in #2291 and #1715, the `mission_planner` and `behavior_path_planner` modules cannot generate neither a route nor a path when the goal pose is behind the start pose on the same lanelet.

This limitation comes from the way paths are designed in Autoware modules. Basically, all planning modules decribe paths as lists of lanelets (`ConstLanelets` or `std::vector<ConstLanelet>`). This only works as long as Autoware does not need to go through the same lanelet twice (which happens when the goal pose is behind the start pose).

To support the latter case, path definition should be refined. For example, a path could be:
- a list of lanelets (`ConstLanelets`)
- a start point on the first lanelet (if any)
- a goal point on the last lanelet (if any)

With this new definition, the first and last lanelet could be the same, yet the starting "segment" and finishing "segment" could be different. In the case the goal point is behind the start point, the first segment would be from the start pose to the end of the lanelet, while the last segment would be from the beginning of the lanelet to the goal pose.

### Purpose

- add support for goal behind start scenario

### Possible approaches

1. In `route_handler`, introduce a new `LaneletPath` class, composed of:
- `ConstLanelets lanelets`: the list of lanelets composing the path
- `double start_arc_length`: the arc length of the start point on the first lanelet (if any)
- `double goal_arc_length`: the arc length of the goal point on the last lanelet (if any)

2. Any planning function using `ConstLanelets` as path should use the new type instead. `ConstLanelets` could still be used for non-path purposes (e.g. to list neighbor lanes). As a consequence, the following modules would need to be heavily reworked:
- `mission_planning`
- `route_handler`
- `behavior_path_planner`

3. Introduce Helper functions to implement route related logic (e.g. extend a path forward)

4. Clean old code 

### Definition of done

- [ ] new `mission_planner` supports all existing use cases + goal behind start on the same lanelet
- [ ] new `behavior_path_planenr` supports all existing use case + goal behind start on the same lanelet
- [ ] all route related logic is handled directly by `route_handler` module

Obviously, a set of OSM maps is needed to make sure Autoware still behaves the same in the supported cases.

[...]

Issue #1715, opened 2022-08-29
Can not give a new goal behind in the same lane


beyzanurkaya:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I'm convinced that this is not my fault but a bug.

### Description

if you give EGO vehicle an initial pose and a goal pose in a lane with the same `lane_id`, but there is no lane with a different `lane_id` between the initial and goal pose, a regular route is created. However, if you give EGO vehicle an initial pose and a goal pose in a lane with the same `lane_id`, but this time there is a lane with a different `lane_id` between the initial pose and the goal pose, a strange route occurs. Moreover, it is possible for EGO to reach its goal pose from the point where it was in the second case.

### Expected behavior

Actually it is possible to reach the goal pose given -according to the lane directions- of EGO from the point where it is. Therefore, a route should have been created as follows:

![Screenshot from 2022-08-29 13-58-43](https://user-images.githubusercontent.com/32412808/187186638-d11d693d-5ebf-4f2e-b45a-5341724f234f.png)

### Actual behavior

There is no realistic route that the vehicle can follow. A route is created as follows:

![Screenshot from 2022-08-29 13-58-54](https://user-images.githubusercontent.com/32412808/187186661-7c70d6a0-53ef-4bca-ac79-62ebb9c4e574.png)

### Steps to reproduce

1. Download [1715.txt](https://github.com/autowarefoundation/autoware.universe/files/9469249/1715.txt) and  [1715-2.txt](https://github.com/autowarefoundation/autoware.universe/files/9469262/1715-2.txt)

2. Download sample-map in [here](https://drive.google.com/file/d/1499_nsbUbIeturZaDj7jhUownh5fvXHd/view)
3. Change file extention from txt to yaml
4. Change pcd and lanelet2 maps file path in yaml files
5. run `ros2 launch scenario_test_runner scenario_test_runner.launch.py sensor_model:=sample_sensor_kit vehicle_model:=sample_vehicle \
scenario:=/scenario/path/1715.yaml \
architecture_type:=awf/universe launch_rviz:=false launch_autoware:=true`

6. Add ` /planning/scenario_planning/lane_driving/trajectory` topic in rviz

7. When you run the scenario, path occurred correctly.

[...]

PR #2424, opened 2022-12-01
feat(route_handler) support when goal is behind pose

VRichardJP:
* Refactor route_handler package: add new class LaneletPoint, LaneletPath and LaneletRoute to handle route related logic
* Use LaneletPath instead of ConstLanelets for path related logic
* Add documentation

Signed-off-by: Vincent Richard <richard-v@macnica.co.jp>

## Description

See https://github.com/autowarefoundation/autoware.universe/issues/2354

## Tests performed

TBD


VRichardJP:
If anyone is checking this PR, I still want to modify the design a bit.

In current PR state, it is "too" easy to get a list of lanelets from a `LaneletPath`, which should be avoided (because a list of lanelets is not a good way to describe a path). Instead, I want `LaneletPath` to be a `std::vector<LaneletSection>`, where `LaneletSection` is a type that describe a section within a lanelet (i.e. `{lanelet, start_arc_length, end_arc_length}`)

Although it requires slightly more changes to the `behavior_path_planner` code, I think it will make the code more robust. In particular, corner cases such as "the goal is behind start" would work without the need for the planning function engineer to think about it.

VRichardJP:
I think my path API has all the functionality the planning modules ever need. What is left is to use it so that corner cases such as "goal is behind start" is never a problem again. Unfortunately, there are maybe thousands LoC to change... I can't do it alone...

In particular, it is getting difficult to keep my work stable when other people are changing the planner modules at the same time.

Currently, there is a strange zigzag in near goal pose when a lane change is required:

Screenshot from 2022-12-15 14-50-58

Except that, I think it is working good. So I guess @mitsudome-r you could already review it.

VRichardJP:
Fixed:
Screenshot from 2022-12-15 17-19-59
Screenshot from 2022-12-15 17-19-43
Screenshot from 2022-12-15 17-19-21

mitsudome-r:
@VRichardJP Thanks for creating the PR.
The description with images helps me a lot to understand the implementation.

Although we really appreciate your PR I have discussed with @TakaHoribe on this, and had several concerns:
* refactoring of the package takes quite a lot of time for us to review.
* We are also considering updating the behavior path planner on the background https://github.com/autowarefoundation/autoware/discussions/3097 which also makes us difficult to merge this at the moment.

Also, we were thinking of different approach to solve the issue by modifying mainly mission_planner rather than adding new Lanelet classes.
* Whenever mission_planner detects a loop in a path, it will only send route to behavior planners until the lane just before the loop.
* As vehicle drives along the route and no longer have a loop, it will update the route until the goal.

Although we have to check if the behavior planner can handle update of route while driving, it would make implementation much simpler. Also, we can also handle route with multiple loops, not only the route with goal point behind starting point.

What do you think about the approach.

VRichardJP:
Hi, I totally get it. I know to change the API may be tricky and cumbersome (and add many bugs).

I never thought about sending partial routes. It could be a thing, although changing the route while the vehicle is driving also comes with its own problems. I could poke this after the holidays.

Although I first started to work on this PR to fix the "goal behind start pose" issue, the proposed changes go way beyond that "simple" case. In particular, it is also a way to write `behavior_path_planner` functions in a simpler and (hopefully) safer way. The more I worked on the planning functions, the more I am convinced the modules would benefit from a better abstraction than just a list of lanelets.

Take it as a message in a bottle. The proposed classes could improve the `route_handler`, or some parts salvaged and reused for new planning functions, or just ignored and thrown in the bin 😅

VRichardJP:
@mitsudome-r 
So if I understand your idea, this is how it could be implemented:

* Modify `route_handler`:
  * Add support for "goal behind start" use case (like in https://github.com/autowarefoundation/autoware.universe/pull/2425 or https://github.com/autowarefoundation/autoware.universe/issues/1715#issuecomment-1318155064). This route may contains loops (e.g. `1->2->3->4->1`).
* Modify `mission_planner`:
  * Move the route publishing logic to a new timer callback function (from `MissionPlanner::change_route`. The new function will:
    * Check the position of the vehicle on the route
    * Remove from the route all sections before current lanelet.
    * If the route has been modified, publish trimmed route (e.g. `1->2->3->4`).

Example:
Screenshot from 2023-01-11 08-23-07

In the situation above:
* `route_handler` generates `1->2->3->4->1`
* timer callback in `mission_planner` first send `1->2->3->4`
* after the ego vehicle has reached `2`, timer callback send `2->3->4->1`
* after the ego vehicle has reached `3`, the timer callback send `3->4->1`
* ...
* after the ego vehicle has reached `1` again, the timer callback send `1`

Although it should work, I have several concerns regarding this method:
* At the end of the day, it is a hacky solution to cope with `motion_planner` not supporting looped lanelet ids.
* This assumes the global route can be changed freely while driving. For example, if the ego vehicle is on lanelet `2`, does the planner generates exactly the same path when the route is `1->2->3` and `2->3`? Other modules may also react on a route change (e.g. new route = new mission).
* Currently, Autoware requires a (remote) operator to validate the route and engage the vehicle. But with this change, the full route is not known anymore (even though its just 1 lanelet missing here). In other words, Autoware could then change the route without any human validation.
* The "Check the position of the vehicle on the route" is actually tricky to implement correctly. For instance, on which route lanelet is the vehicle when several lanelets overlap within the route? For example in the situation below, the ego vehicle is slightly out of the route (e.g. because the operator moved the vehicle manually or autoware moved the vehicle off-road to avoid an obstacle), and lanelet `1` and `2` within the path are overlapped:
Screenshot from 2023-01-11 09-15-28
What if the `(out of path)` is a `3`, and the route is something like `1->2->3`? The ego vehicle is initially on `1`, but while avoiding an obstacle, it gets close to `3` and the thinks it is on `3` so skips the `1->2` route entirely. There are countless corner cases like this.

For most lanelet maps, there should not be any problem. But if there is, Autoware will just break. What do you think?

mitsudome-r:
@VRichardJP Thanks for the comment.

> This assumes the global route can be changed freely while driving.

Currently, TIER IV is trying to support goal modification at run time due to other needs so I believe we will make the change to the other modules as needed.

> Currently, Autoware requires a (remote) operator to validate the route and engage the vehicle.

At least, mission_planner should be aware of the whole route anyways so it can still publish visualization marker of the full route regardless of the route. It can also publish global route as a different topic if it is needed for debug purpose.

> The "Check the position of the vehicle on the route" is actually tricky to implement correctly. For instance, on which route lanelet is the vehicle when several lanelets overlap within the route? For example in the situation below, the ego vehicle is slightly out of the route (e.g. because the operator moved the vehicle manually or autoware moved the vehicle off-road to avoid an obstacle), and lanelet 1 and 2 within the path are overlapped

I believe this issue already exists with the current implementation anyways regardless of the looped route issue(There could be overlapping lanes at intersections). Therefore, I think we should be handling this as a separate issue.

VRichardJP:
@mitsudome-r 
Thank you for the details. Should I wait for the goal modification at runtime feature to be available or should I have a head start on this?

mitsudome-r:
@TakaHoribe Do we have any timeline about when we support mission_planner goal modification at runtime?

xmfcx:
We will wait for goal modification implementation to finish, discussed in [Proposal for handling goal modification](https://github.com/autowarefoundation/autoware/discussions/2983)
It is in the final stage for the implementation. Then this PR will be reworked, hopefully it will be much simpler after the discussion above is resolved.

mitsudome-r:
We should test if this issue still remains in the latest Autoware. (@mitsudome-r or any one from Planning team in TIER IV cc. @TakaHoribe @maxime-clem @takayuki5168 )

takayuki5168:
@VRichardJP 
Currently, this PR has a lot of conflicts, but do you have any idea about how to proceed the PR?

VRichardJP:
@takayuki5168 This PR should have been split into multiple. I initially used the current route handle and planning function as dog food to design and test the new lanelet library.
The library part is pretty much standalone and could be moved to a new independent package. The integration is another beast and could be done step by step, for example by duplicating API (old/new) and converting between methods. But these could be smaller PRs.

Unfortunatelly I don't have much time to spend on Autoware recently so I would not be able to provide much help.

takayuki5168:
@VRichardJP 
Thank you for the quick reply.

Now I read the PR's changes and found it very helpful to avoid the low-level lanelet operation in the nodes' implementation.
I (I mean, the Planning team at TIER IV) will try to take this PR and merge it to the main branch though it may take time.

Thank you very much for the cool PR, and I really appreciate all of your contributions to the Autoware. :pray: 

@mitsudome-r 
The Planning team at TIER IV will take this PR, so please make this still open and draft.

Issue #1543, opened 2022-08-08
Vehicle starts move before the steering angle converged


brkay54:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I'm convinced that this is not my fault but a bug.

### Description

Hi, I limited the `steer_rate_lim_dps` to 5.0 dps in lateral controller. When re-initialize the vehicle, vehicle starts move before steering angle converged.
[video](https://user-images.githubusercontent.com/45468306/183460119-8300f1ed-08e9-4791-8d72-5d2c7a7e5d06.mp4)

### Expected behavior

Vehicle should wait until steering angle converged before move.

### Actual behavior

Vehicle starts move before steering angle converged.

### Steps to reproduce

- set `steer_rate_lim_dps` to 5.0
- re-initialize the vehicle

takayuki5168:
cc @TakaHoribe

kosuke55:
@brkay54 
How about enabling the following options?
- [keep_steer_control_until_converged](https://github.com/autowarefoundation/autoware.universe/blob/main/launch/tier4_control_launch/config/trajectory_follower/lateral_controller.param.yaml#L60)
- [enable_keep_stopped_until_steer_convergence](https://github.com/autowarefoundation/autoware.universe/blob/main/launch/tier4_control_launch/config/trajectory_follower/longitudinal_controller.param.yaml#L9)

takayuki5168:
@brkay54 How's it going

brkay54:
Hi, I tested in planning_simulator with following parameters: 
[keep_steer_control_until_converged](https://github.com/autowarefoundation/autoware.universe/blob/main/launch/tier4_control_launch/config/trajectory_follower/lateral_controller.param.yaml#L60)
    [enable_keep_stopped_until_steer_convergence](https://github.com/autowarefoundation/autoware.universe/blob/main/launch/tier4_control_launch/config/trajectory_follower/longitudinal_controller.param.yaml#L9)

[...]

PR #2441, opened 2022-12-04
feat(trajectory_follower): pub steer converged marker

kosuke55:
Signed-off-by: kosuke55 <kosuke.tnp@gmail.com>

## Description

pub steeer conveged info as debug marker array.

image

## Pre-review checklist for the PR author

The PR author **must** check the checkboxes below when creating the PR.

- [x] I've confirmed the [contribution guidelines].
- [x] The PR follows the [pull request guidelines].

## In-review checklist for the PR reviewers

The PR reviewers **must** check the checkboxes below before approval.

- [ ] The PR follows the [pull request guidelines].
- [ ] The PR has been properly tested.
- [ ] The PR has been reviewed by the code owners.

## Post-review checklist for the PR author

The PR author **must** check the checkboxes below before merging.

- [ ] There are no open discussions or they are tracked via tickets.
- [ ] The PR is ready for merge.

After all checkboxes are checked, anyone who has write access can merge the PR.

[contribution guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/
[pull request guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/pull-request-guidelines/

PR #2432, opened 2022-12-02
refactor(trajectory_follower_nodes): refactor the controller framework

takayuki5168:
Signed-off-by: Takayuki Murooka <takayuki5168@gmail.com>

## Description
What I refactored is mainly as follows.
- refactor the main function of controller_node.cpp
    - create input data -> check if the controller is ready -> initialize if required -> run -> sync
https://github.com/autowarefoundation/autoware.universe/blob/e72ab74718492259ad6f635f73fa83e32d839391/control/trajectory_follower_nodes/src/controller_node.cpp#L187-L210
    - We can initialize the controller as we want depending on the controller's algorithm at that time.
- make each variable in input_data not a pointer so that lat/lon controllers do not need to check each variable in set_input is nullptr or not
image

- make isReady function in lat/lon controllers so that run() function will not return boost::none;
    - Previously, in run() function, we check the controller is ready to run. So it has to return boost variable.

- remove setInputData which is not necessary, and directly pass input_data to run function.

Also, I added a test for the longitudinal controller to check if the lateral control is converged.

## Pre-review checklist for the PR author

The PR author **must** check the checkboxes below when creating the PR.

- [x] I've confirmed the [contribution guidelines].
- [x] The PR follows the [pull request guidelines].

## In-review checklist for the PR reviewers

The PR reviewers **must** check the checkboxes below before approval.

- [ ] The PR follows the [pull request guidelines].

## Post-review checklist for the PR author

The PR author **must** check the checkboxes below before merging.

- [ ] There are no open discussions or they are tracked via tickets.

After all checkboxes are checked, anyone who has write access can merge the PR.

[contribution guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/
[pull request guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/pull-request-guidelines/

PR #2454, opened 2022-12-05
fix(trajectory_follower_nodes): mpc_follower does not send proper converged data under low steering rate limit

brkay54:
Signed-off-by: Berkay Karaman <brkay54@gmail.com>

## Description

Closes: https://github.com/autowarefoundation/autoware.universe/issues/1543

In current Autoware.universe, mpc_follower sends is_steer_converged data "true" to longitudinal controller without reaching the desired steering angle value.

The problem is that when reinitialize the vehicle under low steering angle rate limit, solver can not solve the problem and sends the previous command to vehicle rather than the target value.
Like before reinitialize previous cmd = 30.0 degree,
After we reinitialize vehicle, previous_cmd still 30.0 degree but target_value is maybe 0.0 degree ideally but because steering_angle_rate_limit so small, mpc can not solve the problem and sends 30.0 degree again and vehicle starts moving (because before re initialization, current steering was near to 30.0 degree.)

To solve this, I set mpc's steering_rate_limit large as possible if the vehicle in stop state (otherwise it will be what we declared), so It send the (expected) target steering without steering_rate_limit constraint.

Also I changed the logic of the `isSteerConverged()`. Like @kosuke55 said [here](https://github.com/autowarefoundation/autoware.universe/issues/1543#issuecomment-1323623075), mpc_follower will wait for the (prev_cmd == current_cmd) to be sure output will be the desired angle. Because mpc sends prev_cmd if mpc can not solve the problem, we are also checking is_mpc_calculated. 

### DIFF
--- a/control/mpc_lateral_controller/README.md
+++ b/control/mpc_lateral_controller/README.md
@@ -34,12 +34,15 @@ For the optimization, a Quadratic Programming (QP) solver is used and two option
 <!-- cspell: ignore ADMM -->
 
 - unconstraint_fast : use least square method to solve unconstraint QP with eigen.
-- [osqp](https://osqp.org/): run the [following ADMM](https://web.stanford.edu/~boyd/papers/admm_distr_stats.html) algorithm (for more details see the related papers at the [Citing OSQP](https://web.stanford.edu/~boyd/papers/admm_distr_stats.html) section):
+- [osqp](https://osqp.org/): run the [following ADMM](https://web.stanford.edu/~boyd/papers/admm_distr_stats.html)
+  algorithm (for more details see the related papers at
+  the [Citing OSQP](https://web.stanford.edu/~boyd/papers/admm_distr_stats.html) section):
 
 ### Filtering
 
 Filtering is required for good noise reduction.
-A [Butterworth filter](https://en.wikipedia.org/wiki/Butterworth_filter) is used for the yaw and lateral errors used as input of the MPC as well as for
+A [Butterworth filter](https://en.wikipedia.org/wiki/Butterworth_filter) is used for the yaw and lateral errors used as
+input of the MPC as well as for
 the output steering angle.
 Other filtering methods can be considered as long as the noise reduction performances are good
 enough.
@@ -105,6 +108,7 @@ AutonomouStuff Lexus RX 450h for under 40 km/h driving.
 | keep_steer_control_until_converged           | bool   | keep steer control until steer is converged                                                                                                       | true          |
 | new_traj_duration_time                       | double | threshold value of the time to be considered as new trajectory                                                                                    | 1.0           |
 | new_traj_end_dist                            | double | threshold value of the distance between trajectory ends to be considered as new trajectory                                                        | 0.3           |
+| mpc_converged_threshold_rps                  | double | threshold value to be sure output of the optimization is converged, it is used in stopped state                                                   | 0.3           |
 
 (\*1) To prevent unnecessary steering movement, the steering command is fixed to the previous value in the stop state.
 
@@ -143,11 +147,13 @@ AutonomouStuff Lexus RX 450h for under 40 km/h driving.
 ### How to tune MPC parameters
 
 1. Set appropriate vehicle kinematics parameters for distance to front and rear axle and max steer angle.
-   Also check that the input `VehicleKinematicState` has appropriate values (speed: vehicle rear-wheels-center velocity [km/h], angle: steering (tire) angle [rad]).
+   Also check that the input `VehicleKinematicState` has appropriate values (speed: vehicle rear-wheels-center
+   velocity [km/h], angle: steering (tire) angle [rad]).
    These values give a vehicle information to the controller for path following.
    Errors in these values cause fundamental tracking error.
 
-2. Set appropriate vehicle dynamics parameters of `steering_tau`, which is the approximated delay from steering angle command to actual steering angle.
+2. Set appropriate vehicle dynamics parameters of `steering_tau`, which is the approximated delay from steering angle
+   command to actual steering angle.
 
 3. Set `weight_steering_input` = 1.0, `weight_lat_error` = 0.1, and other weights to 0.
    If the vehicle oscillates when driving with low speed, set `weight_lat_error` smaller.
@@ -156,7 +162,8 @@ AutonomouStuff Lexus RX 450h for under 40 km/h driving.
    One of the simple way for tuning is to increase `weight_lat_error` until oscillation occurs.
    If the vehicle is unstable with very small `weight_lat_error`, increase terminal weight :
    `weight_terminal_lat_error` and `weight_terminal_heading_error` to improve tracking stability.
-   Larger `prediction_horizon` and smaller `prediction_sampling_time` is effective for tracking performance, but it is a trade-off between computational costs.
+   Larger `prediction_horizon` and smaller `prediction_sampling_time` is effective for tracking performance, but it is a
+   trade-off between computational costs.
    Other parameters can be adjusted like below.
 
 - `weight_lat_error`: Reduce lateral tracking error. This acts like P gain in PID.
@@ -165,8 +172,10 @@ AutonomouStuff Lexus RX 450h for under 40 km/h driving.
 - `weight_steering_input`: Reduce oscillation of tracking.
 - `weight_steering_input_squared_vel_coeff`: Reduce oscillation of tracking in high speed range.
 - `weight_lat_jerk`: Reduce lateral jerk.
-- `weight_terminal_lat_error`: Preferable to set a higher value than normal lateral weight `weight_lat_error` for stability.
-- `weight_terminal_heading_error`: Preferable to set a higher value than normal heading weight `weight_heading_error` for stability.
+- `weight_terminal_lat_error`: Preferable to set a higher value than normal lateral weight `weight_lat_error` for
+  stability.
+- `weight_terminal_heading_error`: Preferable to set a higher value than normal heading weight `weight_heading_error`
+  for stability.
 
 ## References / External links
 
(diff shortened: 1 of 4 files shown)


brkay54:
@kosuke55 Sorry to bother you, would you try this? Do you think this implementation can solve our problem?

control/trajectory_follower/src/mpc_lateral_controller.cpp
```diff
@@ -338,33 +360,11 @@ MpcLateralController::getInitialControlCommand() const
...
-  const size_t nearest = motion_utils::findFirstNearestIndexWithSoftConstraints(
-    m_current_trajectory_ptr->points, m_current_kinematic_state_ptr->pose.pose,
-    m_ego_nearest_dist_threshold, m_ego_nearest_yaw_threshold);
-
-  const double current_vel = m_current_kinematic_state_ptr->twist.twist.linear.x;
-  const double target_vel =
-    m_current_trajectory_ptr->points.at(static_cast<size_t>(nearest)).longitudinal_velocity_mps;
-
-  const auto latest_published_cmd = m_ctrl_cmd_prev;  // use prev_cmd as a latest published command
-  if (m_keep_steer_control_until_converged && !isSteerConverged(latest_published_cmd)) {
```
kosuke55, line 361: It's better to select to whether to control steer when stopping.
`m_keep_steer_control_until_converged` is the option, and we would like to keep it.

control/trajectory_follower/include/trajectory_follower/sync_data.hpp
```diff
@@ -31,6 +31,7 @@ struct LateralSyncData
 struct LongitudinalSyncData
 {
   bool is_velocity_converged{false};
+  int state_of_longitudinal_controller{2};
```
kosuke55, line 34: It's ok for draft, but we want change to use
```
enum class ControlState { DRIVE = 0, STOPPING, STOPPED, EMERGENCY };
```
and move it to `longitudinal_contrller_base`
https://github.com/autowarefoundation/autoware.universe/blob/main/control/trajectory_follower/include/trajectory_follower/pid_longitudinal_controller.hpp#L101-L111

kosuke55:
@brkay54 
thanks very much for your PR!

> Also I changed the logic of the isSteerConverged(). Like @kosuke55 said https://github.com/autowarefoundation/autoware.universe/issues/1543#issuecomment-1323623075, mpc_follower will wait for the (prev_cmd == current_cmd) to be sure output will be the desired angle. Because mpc sends prev_cmd if mpc can not solve the problem, we are also checking is_mpc_calculated.

In this part, I meant like
```
  const double steer_cmd_diff = std::abs(cmd.steering_tire_angle - prev_cmd.steering_tire_angle);
  const doulbe steer_cmd_rate = steer_cmd_diff / m_mpc.m_ctrl_period;
  const bool is_cmd_changing = steer_cmd_rate > m_steer_rate_lim_rps / 2;  // need to consider which thresh is reasonable
```
I thought it would be good to check the value of `is_cmd_changing` in `isSteerConverged()`.

> The problem is that when reinitialize the vehicle under low steering angle rate limit, solver can not solve the problem and sends the previous command to vehicle rather than the target value.

Did you check the value of `is_mpc_solved` and did you mean that it sends previous commands when it is `false`?
I think that before using `is_mpc_solved` `isStoppedState()` is checked with previous_cmd and current_steer_angle and if stopped it sends previous commands. So `is_mpc_solved` does not seem related to convergence status.

kosuke55:
@brkay54 
And sorry to bother you, some PRs related to mpc will be merged so this PR will be conflicted. After they are merged could you please resolve the conflict and clean up(keep only necessary diff) your PR?:pray:
- https://github.com/autowarefoundation/autoware.universe/pull/2441
- https://github.com/autowarefoundation/autoware.universe/pull/2432

brkay54:
> Did you check the value of `is_mpc_solved` and did you mean that it sends previous commands when it is `false`?

Yes, if `is_mpc_solved` = false, controller calls `getStopControlCommand()` function which returns previous control command. If it returns the prev_cmd, isChanging() function returns false and vehicle may start to move.

> I think that before using `is_mpc_solved` `isStoppedState()` is checked with previous_cmd and current_steer_angle and if stopped it sends previous commands. So `is_mpc_solved` does not seem related to convergence status.

I changed the structure, it do not send prev_cmd anymore when in stopped state. (Because we do not want to prev_cmd in stopped state, we need desired steering cmd.) If the vehicle in stopped state, it just resets the input buffer with prev_cmd, do not send any control command. After it resets the input buffer, it creates new control command. (Btw I changed the order in run() function, please also consider this.)

brkay54:
@kosuke55 
FYI: I am gonna fix the CI&CD check problems, if this approach is fine for you. 

Also this PR includes some structural changes of mpc_follower. 
@maxime-clem @TakaHoribe @takayuki5168 you may want to take a look.

kosuke55:
@brkay54 
Sorry for the delay. I discussed with @TakaHoribe.
First of all, we think it is not a good idea for longitudinal and lateral to share the same stop state because it would create too many dependencies.
Also, if the essential problem is that mpc can not solve, we want to fix that. If it is still unsolvable, we eliminate the constraint condition. Can you please explain again how to reproduce when is_mpc_solved is false, or can we get a rosbag?

brkay54:
The problem is not whether mpc can solve or not. The problem is that controller does not wait until the mpc create proper control command. Let me explain:

For example;
steering_angle_rate_lim: 0.0872665 rad/sec (5.0 degree/sec)

Before re-initialize the vehicle:
steering_sent: 0.20 rad
actual_steering: 0.19 rad

When re-initialize the vehicle in straight road ->
our desired steering angle should be 0.0 .
actual_steering: 0.19 rad

if mpc can not solve -> lateral controller will create the previous control command which is 0.20. (it is near to actual steering so `isSteerConverged()` function would return true and vehicle starts to move.)

if mpc can solve -> Because the steering_angle_rate_lim = 0.0872665 rad/sec (5.0 degree/sec), lateral_controller will create control command which is near to previous steering command such as 0.195 rad. (Again it is near to actual steering so `isSteerConverged()` function would return true and vehicle starts to move.)

My proposal is:

* if longitudinal_controller is in stopped state make the steering_angle_rate constraint large to get proper steering angle. if it is in driving state make it default value.
* check the differences between the steering values which are output when mpc is solved. If difference < threshold and steering_cmd - actual_steering < converged_steer_rad, make isSteeringConverged = true.

kosuke55:
@brkay54 cc @TakaHoribe @mitsudome-r 
Sorry for the delay and thank you for your detailed explanation.
Basically, I agreed with your proposals.

> ・ if longitudinal_controller is in stopped state make the steering_angle_rate constraint large to get proper steering angle. if it is in driving state make it default value.

This seems good as a workaround. I think it would be better to have a bool parameter(like `no_stopped_steering_angle_rate_lim`, `disable_steer_rate_limit_when_stopped` etc..)(by default `false`) and set the `steering_angle_rate_lim` to max only when it is enabled.
And it would also be better to set it just before using it in
https://github.com/autowarefoundation/autoware.universe/blob/main/control/mpc_lateral_controller/src/mpc.cpp/#L686-L691
```
const double steer_rate = no_stopped_steering_angle_rate_lim && is_stopped ? m_steer_rate_lim : numerical_limit<double>::max();
```
(And since the root cause seems that mpc can not solve under the constraint, so we will attempt to investigate further and fix in another PR.)

>・ check the differences between the steering values which are output when mpc is solved. If difference < threshold and steering_cmd - actual_steering < converged_steer_rad, make isSteeringConverged = true.

This also seems good. And in case the difference is very small, it would be better to look at it for a few seconds for better performance. For example, instead of making a judgment based on a change of 0.20 -> 0.195 in one step, it is better to use 1-second difference and determine that it changes when the value changes 0.2 -> 0.1.

> What do you mean when you said "it would create too many dependencies.".

Sorry for my lack of explanation.
One of autoware's policies is the modularity, and each module needs to work independently as possible. We would like to minimize the number of longitudinal controller information which is used in the lateral controller as much as possible. We want to determine the stopped state only in the lateral controller.

And `isStoppedState()` of mpc is originally the flag about whether to keep steer control, not about velocity state.
If the reason why you need to use stopped state of longitudinal controller in lateral controller is to prevent large difference commands with no constraints from being sent while driving because the lateral stop state which checks the steer angle does not match the longitudinal one, then how about check only velocity in mpc? As I mentioned abouve, how about `const bool is_stopped = std::abs(v) < eps` in mpc.cpp?

mitsudome-r:
We had another call on this PR. In conclusion:
* We will allow larger steering rate limit when the vehicle is at stop (so that we have less case that mpc fails to solve)
* We create new issue to solve the issue with MPC not being able to solve with a given constraint as a long term solution
* We improve isSteerConverged() function to look longer period of time to detect if steering is properly converged or not.

mehmetdogru:
@brkay54 will update soon.

brkay54:
@kosuke55 sorry for being late, I updated the PR, please check it. 
Also, created another PR which is related is_steering_converged info: https://github.com/autowarefoundation/autoware.universe/pull/3644

mehmetdogru:
@kosuke55 friendly ping to review.

kosuke55:
sorry to be late.

I checked psim and tier4 [internal scenario test](
https://evaluation.tier4.jp/evaluation/reports/7e93224f-887c-58c0-bc9d-419cf720d66e?project_id=prd_jt) (1326/1372->1329/1372), and it worked fine.

kosuke55 (approved):
LGTM

kosuke55:
I send launcher PR https://github.com/autowarefoundation/autoware_launch/pull/378/files
We need merge them at the same time.

kosuke55:
@TakaHoribe 
The code and behavior look good to me, if it is ok for you please approved this as a code owner.

Issue #2097, opened 2022-10-18
Improve behavior_velocity_planner's latency


soblin:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I've agreed with the maintainers that I can plan this task.

### Description

Report from @veqcc.

Currently the latency of `onTrigger` function in behavior_velocity_planner has a big influence in Autoware's entire latency. Below figures show (1) the latency of behavior_path_planner and (2) behavior_velocity_planner respectively, measured in [planning_simulation tutorial](https://autowarefoundation.github.io/autoware-documentation/main/tutorials/ad-hoc-simulation/planning-simulation/).

![image](https://user-images.githubusercontent.com/28677420/196393196-1bbd8906-aa26-4282-b864-8d89372e6290.png)

![image](https://user-images.githubusercontent.com/28677420/196393214-fb9a2d70-88c3-47f3-a69e-60c07c4dfe2f.png)

![image](https://user-images.githubusercontent.com/28677420/196441534-78802da8-3ca3-4844-bb6b-3857addb4a19.png)

Especially the latency of behavior_velocity_planner is constantly not negligible (about 200ms ) given the behavior_path_planner's publication rate of 10Hz, which indicates its latency is twice the subscription rate.

![image](https://user-images.githubusercontent.com/28677420/196397399-a2bdf497-b6fa-4df8-b6e9-3ea277630c79.png)

@veqcc and @soblin are working to lower this computation time as much as possible. I'm not sure about the lower bound of this latency, but it should be at least below the behavior_path_planner's publication rate of 10Hz.

### Possible approaches

- get statistics on the functions calls
  - measure elapsed time by binary search
  - get FlameGraph
- refactor current implementation
  -  check the logic flow and return instantly as much as possible
  - avoid and delay heavy computation
  - cache lanelet stuff because this data is expected to be almost the same in each unit of behavior_velocity module.
    - define a cache class for lanelet that has `CompoundPolygon3d` and `tier4_autowawe_utils::Polygon2d` in it
- do calculation *in-place*

### Definition of done

- measure the performance in realistic cases with a massive map, various & numerous obstacles, with multiple lane elements and characteristics (turn_direction, walkway, turn_signal, occlusion spot, blind spot, etc.)
- 2x ~ 3x performance improvement would be good.

[...]

PR #1635, opened 2022-08-19
feat(behavior_velocity): publish internal debug path

taikitanaka3:
## Description

- publish internal behavior velocity path as a debug message.

image

To test this use planning debug tool trajectory analyzer

## Pre-review checklist for the PR author

The PR author **must** check the checkboxes below when creating the PR.

- [x] I've confirmed the [contribution guidelines].
- [x] The PR follows the [pull request guidelines].

## In-review checklist for the PR reviewers

The PR reviewers **must** check the checkboxes below before approval.

- [ ] The PR follows the [pull request guidelines].

## Post-review checklist for the PR author

The PR author **must** check the checkboxes below before merging.

- [ ] There are no open discussions or they are tracked via tickets.

After all checkboxes are checked, anyone who has write access can merge the PR.

[contribution guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/
[pull request guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/pull-request-guidelines/

PR #2462, opened 2022-12-06
perf(behavior_velocity_planner): remove unnecessary debug data

veqcc:
Signed-off-by: Ryuta Kambe <ryuta.kambe@tier4.jp>

## Description
cc: @taikitanaka3 @soblin 
In behavior_velocity_planner node, the debug data `path_raw` and `spline_path` (or `path_iunterpolate`) has no longer be used since this commit https://github.com/autowarefoundation/autoware.universe/pull/1635 was merged.  
Thus I deleted those debug data from intersection, blind_spot and occlusion_spot modules.

## Related links
https://github.com/autowarefoundation/autoware.universe/issues/2097

## Performance
The graph shoes the processing time of behavior_velocity planner node with the rosbag that records the test running in Odaiba.
The blue line the one after this commit is applied, which is more than 10msec faster at maximum than the original. 
image


taikitanaka3 (approved):
LGTM

soblin (approved):

PR #2467, opened 2022-12-07
refactor(tvm_utility): redefine network node and inference config

angry-crab:
Signed-off-by: Xinyu Wang <xinyu.wang@tier4.jp>

## Description

Relates to #2308 
According to [Template config.hpp input and output type](https://github.com/autowarefoundation/modelzoo/issues/81), individual input and output node might have different data type. `NetworkNode` needs to be refactored. 

## Pre-review checklist for the PR author

The PR author **must** check the checkboxes below when creating the PR.

- [x] I've confirmed the [contribution guidelines].
- [x] The PR follows the [pull request guidelines].

## In-review checklist for the PR reviewers

The PR reviewers **must** check the checkboxes below before approval.

- [ ] The PR follows the [pull request guidelines].

## Post-review checklist for the PR author

The PR author **must** check the checkboxes below before merging.

- [ ] There are no open discussions or they are tracked via tickets.

After all checkboxes are checked, anyone who has write access can merge the PR.

[contribution guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/
[pull request guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/pull-request-guidelines/

Issue #2308, opened 2022-11-17
TVM copy overhead


angry-crab:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I've agreed with the maintainers that I can plan this task.

### Description

In a heterogeneous context, multi-state inference introduces copy overhead. 
https://github.com/liuzf1988/autoware.universe/blob/0fcaf94d7836f42eaf4a256fca97e5b63d2487d0/perception/lidar_centerpoint_tvm/lib/centerpoint_tvm.cpp#L225-L228
For example, the inference above first copies data from cpu to gpu, then gpu to cpu for `scatter`, and gpu back to cpu for the final stage inference. Data tranfer between devices should be avoided. 

### Purpose

Avoid data copy between devices and improve performance. 

### Possible approaches

Implement `scatter` with TVM script. 

### Definition of done

- [x] implement `scatter`
- [x] modify ci pipleline to include pre/postprocessing.

angry-crab:
@ambroise-arm 
Regarding the `scatter` compilation, it is something like this
https://github.com/angry-crab/tvm_example/blob/d4d146bdc657915553ebf510e044e45743a6a32b/scatter_tir.py#L8-L21
To push it to `modelzoo`, I think we have two options:
1. Compile it into one of the onnx model with a predefined name such as `util`
2. Consider it as another "network" and compiled into a different runtime lib. 

I guess the second seems better because there might be multiple such operators. 

As for `modelzoo` pipeline, do you have any suggestion where I should put it?

ambroise-arm:
> I guess the second seems better because there might be multiple such operators.
> As for modelzoo pipeline, do you have any suggestion where I should put it?

I like option 1. better, unless you have examples of the same `scatter` function being used in other processing pipelines. 

Instead of the `util` name, I would go for `preprocess` (if we consider it is the preprocessing for the backbone).
And then this could be made a general design rule that any network could benefit from `preprocess` and `postprocess` utilities like this. So the layout of a network archive would become:
```
deploy_graph.json
deploy_lib.so
deploy_param.params
inference_engine_tvm_config.hpp
preprocess.so (optional)
postprocess.so (optional)
```
[...]

PR #2468, opened 2022-12-07
feat(tvm_utility, lidar_centerpoint_tvm): tvm pipeline extension

angry-crab:
## Description

Resovles #2308 

#2467 has to be merged before this PR. 

Tested locally for both llvm and opencl. 


angry-crab:
@ambroise-arm 
I added myself to he maintainer list of `tvm_utility` so that we don't have to ask Fatih to merge the PR. 
Another change is that I have to download and unpack the packages first to see if there is any preprocessing lib.

perception/lidar_centerpoint_tvm/lib/network/scatter.cpp
```diff
@@ -62,7 +62,7 @@ void scatterFeatures_worker(
   }
 }
 
-void scatterFeatures(
+[[maybe_unused]] void scatterFeatures(
```
angry-crab, line 65: I think we want to keep this function as an example to rewrite cuda kernels using cpp, even though it is not being used anymore.

ambroise-arm: I generally don't like to have unused code laying around. It gets hard to remember either what it did or how it was called and it can become hard to interface with the rest as code evolves. What about removing the function but putting a comment, either in the documentation or somewhere relevant in the codebase, pointing to the commit hash that was using it and saying it is a way to to the implementation in CPP?

common/tvm_utility/tvm_utility-extras.cmake
```diff
@@ -70,15 +69,17 @@ function(get_neural_network MODEL_NAME MODEL_BACKEND DEPENDENCY)
     set(DOWNLOAD_DIR "${DATA_PATH}/downloads")
     set(SOURCE_DIR "${DATA_PATH}/models/${MODEL_NAME}")
     set(INSTALL_DIRECTORY "${DATA_PATH}/models/${MODEL_NAME}")
+    file(DOWNLOAD ${URL} "${DOWNLOAD_DIR}/${ARCHIVE_NAME}")
```
ambroise-arm, line 72: That's a bit shaky. It works fine for the first build: it downloads from the bucket. But for the second build the archive will exist, so what I guess happens is that `URL` becomes the name of the archive (from line 60) and then cmake downloads the file into itself. On my computer this results in the `.tar.gz` file having a size of 0 on disk. But because the files had already been extracted on the first build then it still works. I guess it will break if going back and forth between target versions.
Maybe remove the first `if` case at line 60, and instead do the download conditionally here? I don't know if it will solve everything though.

perception/lidar_centerpoint_tvm/lib/centerpoint_tvm.cpp
```diff
@@ -39,135 +42,113 @@ namespace perception
...
+  std::ifstream module(network_module_path);
+  if (!module.good()) {
+    throw std::runtime_error("File " + network_module_path + " specified in config.hpp not found");
+  }
+  module.close();
+  tvm::runtime::Module runtime_mod = tvm::runtime::Module::LoadFromFile(network_module_path);
 
-  return {output};
-}
+  f = runtime_mod.GetFunction(function_name);
```
ambroise-arm, line 61: nit: please use something more evocative and with more letters than `f`. `tvm_module`? idk

perception/lidar_centerpoint_tvm/lib/centerpoint_tvm.cpp
```diff
@@ -39,135 +42,113 @@ namespace perception
...
 }
 
-TVMArrayContainerVector BackboneNeckHeadPreProcessor::schedule(const MixedInputs & pillar_inputs)
+TVMArrayContainerVector VoxelEncoderPreProcessor::schedule(const MixedInputs & voxel_inputs)
 {
-  // do scatter to convert pillar_features to spatial_features
-  scatterFeatures(
-    pillar_inputs.features, pillar_inputs.coords, pillar_inputs.num_voxels, config_detail,
-    spatial_features);
+  /// generate encoder_in_features from the voxels
```
ambroise-arm, line 101: nit: three `/`?

perception/lidar_centerpoint_tvm/include/lidar_centerpoint_tvm/centerpoint_tvm.hpp
```diff
@@ -55,6 +55,22 @@ struct MixedInputs
   std::vector<int32_t> coords;
 };
 
+class LIDAR_CENTERPOINT_TVM_LOCAL TVMScatterIE : public tvm_utility::pipeline::InferenceEngine
```
ambroise-arm, line 58: What do you think of having this class in `tvm_utility` instead? I don't see anything specific to scatter and it could be reused by other packages. Could be renamed to refer to tvm modules generally.

ambroise-arm (changes requested):
It doesn't scale well if we have to define `ThreeStagePipeline` and so on. Let's leave it like this for now, if we do need more stages in the future it could be good to see if it possible to pass an array of classes or references to them and then iteratively chain their `schedule` functions.

perception/lidar_centerpoint_tvm/include/lidar_centerpoint_tvm/centerpoint_tvm.hpp
ambroise-arm: Ah nevermind, it doesn't make sense because of `coords`.

common/tvm_utility/tvm_utility-extras.cmake
angry-crab: I'm taking days off this week. Will come back and fix it next week.

xmfcx (approved):

PR #2476, opened 2022-12-08
fix(emergency_handler): fix mrm handling when mrm behavior is none

mkuri:
Signed-off-by: Makoto Kurihara <mkuri8m@gmail.com>

## Description

mrm-behavior-state-machine
When each MRM behavior is called from the MRM None behavior, a process to cancel the None behavior is called.
But there was no corresponding process to cancel the None behavior, so a warning was displayed.

Add a handler for MRM None behavior.

### DIFF
--- a/system/emergency_handler/src/emergency_handler/emergency_handler_core.cpp
+++ b/system/emergency_handler/src/emergency_handler/emergency_handler_core.cpp
@@ -224,6 +224,10 @@ void EmergencyHandler::callMrmBehavior(
   auto request = std::make_shared<tier4_system_msgs::srv::OperateMrm::Request>();
   request->operate = true;
 
+  if (mrm_behavior == MrmState::NONE) {
+    RCLCPP_WARN(this->get_logger(), "MRM behavior is None. Do nothing.");
+    return;
+  }
   if (mrm_behavior == MrmState::COMFORTABLE_STOP) {
     auto result = client_mrm_comfortable_stop_->async_send_request(request).get();
     if (result->response.success == true) {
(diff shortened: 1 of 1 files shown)


system/emergency_handler/src/emergency_handler/emergency_handler_core.cpp
```diff
@@ -224,6 +224,10 @@ void EmergencyHandler::callMrmBehavior(
   auto request = std::make_shared<tier4_system_msgs::srv::OperateMrm::Request>();
   request->operate = true;
 
+  if (mrm_behavior == MrmState::NONE) {
+    RCLCPP_WARN(this->get_logger(), "MRM behavior is None. Do nothing.");
```
kenji-miyake, line 228: No need to use THROTTLE?

mkuri: I don't think it is necessary because it is not the kind of warning that would display continuously.
Also, the timing of each warning is tied to the vehicle behavior, so I would like to display the warnings as they occer.

If they display continuously at unintended timing, I will change it to use THROTTLE.

kenji-miyake: Okay, thank you!

shmpwk: I agree it is not too much output.
I previously watched `invalid MRM behavior` which may be nearly same frequency as that warning, and it looks adequate amount.

shmpwk (approved):

Issue #2439, opened 2022-12-04
freespace_planning_algorithms-test fails


kosuke55:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I'm convinced that this is not my fault but a bug.

### Description

- Sometimes a timeout occurs without solving for a certain goal.
  -  ex) https://github.com/autowarefoundation/autoware.universe/actions/runs/3510437832/jobs/5880231292#step:7:11240

```
1: [ RUN      ] RRTStarTestSuite.Update
1: measuring average performance ...
1: plan success : 201.148[msec], solution cost : 43.6063
1: [INFO] [1668999471.297964168] [rosbag2_storage]: Opened database '/tmp/fpalgos-rrtstar_update-case0/fpalgos-rrtstar_update-case0_0.db3' for READ_WRITE.
1: measuring average performance ...
1/6 Test #1: freespace_planning_algorithms-test ...***Timeout  60.00 sec
test 2
    Start 2: rrtstar_core_informed-test
```

- `malloc(): invalid size (unsorted)`  error occurs in `AstarSearchTestSuite.SingleCurvature`
  - ex) https://github.com/tier4/autoware.universe/actions/runs/3597700641/jobs/6061717527#step:7:11186

```
1: [ RUN      ] AstarSearchTestSuite.SingleCurvature
1: plan success : 598.159[msec], solution cost : 34.2057
1: malloc(): invalid size (unsorted)
1: -- run_test.py: return code -6
1: -- run_test.py: generate result file '/__w/autoware.universe/autoware.universe/build/freespace_planning_algorithms/test_results/freespace_planning_algorithms/freespace_planning_algorithms-test.gtest.xml' with failed test
1: -- run_test.py: verify result file '/__w/autoware.universe/autoware.universe/build/freespace_planning_algorithms/test_results/freespace_planning_algorithms/freespace_planning_algorithms-test.gtest.xml'
1/6 Test #1: freespace_planning_algorithms-test ...***Failed    1.10 sec
```

### Expected behavior

all tests pass

### Actual behavior

tests fails

### Steps to reproduce

run tests

kosuke55:
cc @HiroIshida 
If you know anything about this, I would appreciate it if you could let me know.

HiroIshida:
[...]

PR #2505, opened 2022-12-14
fix(freespace_planning_algorithms): throw error when no waypoints stored

HiroIshida:
### 1. safer access to waypoints
Current implementation of `compute_length` is problematic because it can invalidly access vector element and cause the following error. 
```
terminate called after throwing an instance of 'std::out_of_range'
  what():  vector::_M_range_check: __n (which is 0) >= this->size() 
```
This is mainly happens when solver failed to solve, which means no waypoints are set.  So, also to fix the root cause, I add check if the waypoints is empty in `getWaypoints` function.

### 2. update test
related to 1, I updated the test so that if planning fails, don't call `getWaypoints`

### note
This may related to https://github.com/autowarefoundation/autoware.universe/issues/2439 but if invalid access in `compute_length()` causes the error, that's mean it takes more than 10 seconds to solve a single problem by astar, and the chance is quite low. So, probably this PR will not fix https://github.com/autowarefoundation/autoware.universe/issues/2439 


planning/freespace_planning_algorithms/include/freespace_planning_algorithms/abstract_algorithm.hpp
```diff
@@ -110,7 +110,14 @@ class AbstractPlanningAlgorithm
   virtual bool makePlan(
     const geometry_msgs::msg::Pose & start_pose, const geometry_msgs::msg::Pose & goal_pose) = 0;
   virtual bool hasObstacleOnTrajectory(const geometry_msgs::msg::PoseArray & trajectory) const;
-  const PlannerWaypoints & getWaypoints() const { return waypoints_; }
+  const PlannerWaypoints & getWaypoints() const
+  {
+    if (waypoints_.waypoints.size()) {
```
TakaHoribe, line 115:
```suggestion
    if (!waypoints_.waypoints.empty()) {
```

planning/freespace_planning_algorithms/include/freespace_planning_algorithms/abstract_algorithm.hpp
```diff
@@ -110,7 +110,14 @@ class AbstractPlanningAlgorithm
...
+      return waypoints_;
+    }
+    throw std::runtime_error("waypoints has length 0. Did you check that problem solved?");
```
TakaHoribe, line 118: It is natural to return an empty waypoint even if it is empty. The error handling should be done outside of this function

TakaHoribe (commented):
@HiroIshida Thank you for this PR. 
Question: What will happen when the solver fails to compute the waypoint? Before it results in an invalid index access exception, and with this PR the planner still dies with another exception?

planning/freespace_planning_algorithms/include/freespace_planning_algorithms/abstract_algorithm.hpp
HiroIshida: Sorry this was a typo

planning/freespace_planning_algorithms/include/freespace_planning_algorithms/abstract_algorithm.hpp
HiroIshida: OK, I should remove this

HiroIshida:
@TakaHoribe 

> Before it results in an invalid index access exception, and with this PR the planner still dies with another exception?
Yes. In the previous implementation, when user and developer mistakenly call `compute_length` against empty waypoint, it the program finish with the following error. 
```
terminate called after throwing an instance of 'std::out_of_range'
  what():  vector::_M_range_check: __n (which is 0) >= this->size() 
```
And although the error is simple, typically user need to compile it with debug flag and have to use debugger to detect this error. But if we throw a error with explicit message, user can easily detect the reason of bug without any debugger.

TakaHoribe:
@HiroIshida 

> mistakenly call compute_length against empty waypoint

Does this operation happen accidentally? For example, in the case of path search failure?

TakaHoribe (approved):
LGTM

Issue #2182, opened 2022-10-31
Trajectory planning for long vehicles not working properly


WJaworskiRobotec:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I'm convinced that this is not my fault but a bug.

### Description

While testing of ITRI Bus on ITRI campus in AWSIM, I've discovered that the planner does not work properly for planning of turns for long vehicles. As can be noticed on the screenshot below, the bus is trying to "cut" the right turn by driving on the sidewalk. I know that there were seperate issues/discussions about expanding drivable area, but in this case we should think about different trajectory planning algorithm considering the dimensions of the vehicle. In reality, long vehicles like buses drive more to the outer part of the road, to avoid driving on the sidewalk.

![ITRI_AWSIM](https://user-images.githubusercontent.com/85557834/199016405-6614e6f6-c2c7-4c7b-87d0-62c3685c1799.png)

The same behavior was observed in the planning_simulator, so it's not related to AWSIM:

![ITRI_planning_sim](https://user-images.githubusercontent.com/85557834/199016885-44528a45-ba8d-4cb7-98c1-05c6984c5fda.png)

Used vehicle model :
https://github.com/autowarefoundation/cbus_vehicle_launch

Location of the observed problem:

![Screenshot from 2022-11-07 10-25-50](https://user-images.githubusercontent.com/85557834/200275813-f9b4b972-be43-442a-b897-9715415eb89b.png)

EDIT: 
After more investigation, it looks like the problem can be solved by modifying drivable area (https://github.com/orgs/autowarefoundation/discussions/2851). Anyway, I believe that to make it possible for longer vehicles to behave properly, in some cases the drivable area should be even moved away from inner turn radius (in current logic the negative value of bound_offset) 

### Expected behavior

Bus properly driving on the entire map of ITRI campus. 

### Actual behavior

Bus gets stuck on the right turns.

### Steps to reproduce

1. Run planning simulator with ITRI map and cbus_vehicle
2. Set the goal poses to follow demo route 2 defined for ITRI project : [link](https://github.com/autowarefoundation/autoware-projects/issues/45#issuecomment-1274550309)

yukkysaito:
I think the current solution was to extend the drivable area to the opposite lane.
@satoshi-ota Can you explain the details?

[...]

PR #2609, opened 2023-01-04
feat(behavior_path_planner): expand the drivable area based on the vehicle footprint

maxime-clem:
## Description

This PR adds the option to expand the drivable area based on the vehicle footprint.
Extra options are also added to prevent expanding the drivable area into the predicted paths of dynamic objects, or beyond some map linestrings with specific types. 

## Related links

https://github.com/autowarefoundation/autoware.universe/issues/2182
https://github.com/orgs/autowarefoundation/discussions/3013
https://github.com/orgs/autowarefoundation/discussions/2851

## Tests performed

Simple test in the planning simulator

### Performances

<details>
  <summary>Flamegraphs</summary>
For a busy scene with many dynamic objects, the expansion of the drivable area has a significant impact on the performance of the `behavior_path_planner` (right side of the flamegraph).

image

Most of the time taken to generate the drivable area is now spent on the expansion.

image

</details>

## Notes for reviewers

**To test this PR**, please also checkout the corresponding launcher branch: https://github.com/autowarefoundation/autoware_launch/tree/behavior_path_planner/drivable-area-expander


mehmetdogru:
@maxime-clem 

I was testing the PR with gtu campus map & isuzu bus and I encountered something that I want to share. The case:

- there are path footprints go beyond the right bound however no space to expand it because of the road_border
- there is a neighbor lane and enough space for mpt to plan a trajectory on the left but no path footprints go beyond the left bound
- There is not enough space between the right and left bounds (area) for mpt to plan a trajectory inside it.
---

**lanelet structure & road borders:**
- 2023-01-05-18-24-47_road_border

**drivable_area bounds:**


**path_with_lane_id footprints (not accurate):**


**mpt footprints:**


---

As the current algorithm, a bound is expanded only if the path footprints go beyond it until the road_border. In a case like this one maybe there should be a way to expand the left bound. As an idea maybe if the conditions below are met, the opposite bound can be expanded: 

- If path footprints exceed a bound but it can not be expanded at all or can not be expanded enough
- Determine the related section (including the curve)
- If the related path section has enough curvature so we wouldn't do this on a straight lane
- If there is enough space over the opposite bound

At the end it would be something like this:

- 2023-01-05-18-25-38_result

What do you think about this idea?

I know that as a workaround the related lane shape can be annotated wider on the lanelet2 map and we wouldn't have this problem but I don't think HAD maps should be vehicle specific.

maxime-clem:
> What do you think about this idea?

@mehmetdogru Thank you for the feedback. I understand the problem and I believe your idea is good.

I think I can simplify the current implementation to integrate your idea as follows:
- Calculate the ego path footprint by considering the ego footprint at each path point (_same as current implementation_).
- If the ego path footprint goes outside of the drivable area then we extend the drivable area (_new_).
   - Create an extended drivable area using some buffer distance on the left/right bound (_new but similar to the static drivable area extension_).
      - The buffer distance can be a parameter or can be calculated based on how much the ego path footprint goes out of the drivable area.
      - Option to only extend the drivable area between the first and last points that go out of the drivable area, with new parameters like you described to also extend ahead/behind these points.
   - [OPTIONAL] Cut the extended drivable area using uncrossable lines and dynamic objects' predicted paths (_same as current implementation_).
   - Add the extended drivable area to the original drivable area (_same as current implementation_).

In summary, I would simplify the current implementation to only use the ego path footprint to _check_ if we need to extend the drivable area or not. Then I would use a simpler way to extend the drivable area (buffer distance on the left/right bounds).
I believe this will solve the problem your reported, fix some issues with the current implementation, and also improve performances.

maxime-clem:
@mehmetdogru 
I have not been able to implement the simplification I had in mind so I will continue using the current polygon based implementation for now.
In the latest commit ([59f9b68](https://github.com/autowarefoundation/autoware.universe/pull/2609/commits/59f9b68d1b4edd19fdd45568ccbc8f1a94ff3ed6)), I add an option to "compensate" when the drivable area cannot be expanded due to a blocking linestring. "Compensate" means that we will create an additional footprint shifted in the opposite direction of the blocking linestring. I think this achieves what you need, although the current implementation with polygons does not work well due to the low path resolution. I added an option to use circle footprints which helps with that, but it still needs improvements.

mehmetdogru:
> I have not been able to implement the simplification I had in mind so I will continue using the current polygon based implementation for now. In the latest commit ([59f9b68](https://github.com/autowarefoundation/autoware.universe/pull/2609/commits/59f9b68d1b4edd19fdd45568ccbc8f1a94ff3ed6)), I add an option to "compensate" when the drivable area cannot be expanded due to a blocking linestring. "Compensate" means that we will create an additional footprint shifted in the opposite direction of the blocking linestring. I think this achieves what you need, although the current implementation with polygons does not work well due to the low path resolution. I added an option to use circle footprints which helps with that, but it still needs improvements.

@maxime-clem 

Thank you very much for the implementation. I test it out. There are couple of things I would like share (I can share more details if necessary):

- Main issue is that the calculations are very heavy and it takes up to 700 ms per cycle, is it also like this on your end? (related to pc specs etc.) `filterFootprint` func takes %95 percent of the time spent on expansion.
- For the same reason maybe expansion for da can be done just for mpt traj length (not for all the path length)
- Related to the compensation, I am not sure `extra_compensate_shift` works as intended. Even though I increase the value could't see any difference. If we use the circle_footprints when we drive on the straight lane and our actual footprints are not out of the drivable area, expansion is gonna be done anyways.

I believe what we should have at the end of the day is that enough drivable area when we are turning, and no expansion when we drive straight (no path footprint out). Roughly something like the one I shared before:

- 2023-01-05-18-25-38_result

I created [a scenario](https://drive.google.com/file/d/1eVYvZwtSsdxn5M6MZd1_eARQxy9kh7yG/view?usp=sharing) for this turn as a reference if you would like to test this PR as well. If you are gonna test this turn with Isuzu bus please make sure that:

- you have the changes in [autoware](https://github.com/mehmetdogru/autoware/tree/test/isuzu-config) to have Isuzu vehicle & sensor model.
- you have [isuzu mpt configurations](https://github.com/mehmetdogru/autoware_launch/tree/test/isuzu-config)
- when you are launching `scenario_test_runner` specify `isuzu_vehicle` and `isuzu_sensor` as `vehicle_model` and `sensor_model`

launch/tier4_planning_launch/config/scenario_planning/lane_driving/behavior_planning/behavior_path_planner/drivable_area_expansion.param.yaml
```diff
@@ -24,3 +25,25 @@
...
+          rear: 0.5 # [m] extra length to add to the rear of the ego footprint
+          left: 0.5 # [m] extra length to add to the left of the ego footprint
+          right: 0.5 # [m] extra length to add to the rear of the ego footprint
+      dynamic_objects:
+        avoid: true # if true, the drivable area is not expanded in the predicted path of dynamic objects
+        extra_footprint_offset:
+          front: 0.5 # [m] extra length to add to the front of the ego footprint
+          rear: 0.5 # [m] extra length to add to the rear of the ego footprint
+          left: 0.5 # [m] extra length to add to the left of the ego footprint
+          right: 0.5 # [m] extra length to add to the rear of the ego footprint
```
takayuki5168, line 44: "dynamic object" instead of "ego"?

planning/behavior_path_planner/src/util/drivable_area_expansion/footprints.cpp
```diff
@@ -0,0 +1,76 @@
...
+#include <tf2_geometry_msgs/tf2_geometry_msgs.hpp>
+
+#include <boost/assign.hpp>
+#include <boost/geometry.hpp>
+
+#include <tf2/utils.h>
+
+namespace drivable_area_expansion
+{
+polygon_t rotatePolygon(const polygon_t & polygon, const double angle)
```
takayuki5168, line 28: FYI: The similar function exists here. The argument type is not the same though.
You can move your util function used in other package here. For me, it's not necessary to do that in this PR.
https://github.com/autowarefoundation/autoware.universe/blob/ded1d94524e673d871f60b7fd4268d7a417ce304/common/tier4_autoware_utils/src/geometry/boost_polygon_utils.cpp#L93-L94

planning/behavior_path_planner/include/behavior_path_planner/util/drivable_area_expansion/parameters.hpp
```diff
@@ -0,0 +1,111 @@
...
+
+#include <string>
+#include <vector>
+
+namespace drivable_area_expansion
+{
+
+struct DrivableAreaExpansionParameters
+{
+  static constexpr auto ENABLED_PARAM = "dynamic_expansion.enabled";
```
takayuki5168, line 31: Can I ask why you hold a parameter name as a member variable.

maxime-clem: So I do not have to write the parameter twice (declaring and updating)
https://github.com/autowarefoundation/autoware.universe/blob/f1d53ddc66efe35cd61c3f2bf710dec39c13dc70/planning/behavior_path_planner/src/behavior_path_planner_node.cpp#L1100-L1102

takayuki5168: Got it. Thanks.

maxime-clem:
@mehmetdogru 
The newest version can solve your scenario but requires an additional change in the parameters of the `obstacle_avoidance_planner`:
 ```yaml
lat_error_weight: 10.0
```
In situations where the path is close to one of the drivable area bound, even if the other bound is extended such that there is enough space to pass, the motion planner will try to stick to the path and will keep colliding with the drivable area bound.  Changing the `lat_error_weight` allows the motion planner to deviates from the reference path.

I will make a few more changes this week and then will open the PR for review. Thank you for your patience.

takayuki5168:
LGTM

@satoshi-ota I think you should take care of this PR. Could you approve if the PR looks good.

mitsudome-r:
@mehmetdogru won't be able to review for 3 weeks.

We can merge this if code-owner approves the PR, and create issue if @mehmetdogru notice any issue.
@takayuki5168 Could you approve the PR if you are fine with merging?

takayuki5168 (approved):

PR #2613, opened 2023-01-05
Fix for freespace planner memleak issue.

Narendarselva:
Freespace planner application crashes when "replan_when_obstacle_found" is enabled. Issue being collision_indexes_table_ is populated contineously without resetting for every setMap which leads to contineous memory allocation for the above mentioned vector table. Its curcial while using the application on a low resource compute environment.

PFA for a video on memleak issue.
https://user-images.githubusercontent.com/48509412/210761182-7edfed4f-88aa-4d13-98ba-603e1ad10e16.mp4


Narendarselva:
This is not just the memleak issue , it also relates to performance will fix both and will create a new pull request.

Narendarselva:
Created a new PR.
[https://github.com/autowarefoundation/autoware.universe/pull/2621](url)

PR #2618, opened 2023-01-07
fix(pointcloud_preprocessor): fix output intensity value of  ring outlier filter 

miursh:
## Description
* Correct the output intensity of ring-outlier-filter since its current output seems always zero like value.
* Fix point fileds error mentioned [here](https://github.com/autowarefoundation/autoware.universe/pull/3014#issuecomment-1565621541)

## Tests performed

I checked the output intensity is reasonable value.
(before) 
image

(after: obviously intensity is included)
image

I checked the output pointcloud position is exactly the same in the before and after, but only the intensity is different.
(True in image below means "same", False means "not same")
image

Also I checked the processing time.
The average time is almost the same; 10.88 ms (before) and  10.95 ms (after) in my desktop PC.
The graph below shows the processing time for before and after applying this PR. 
image

### DIFF
--- a/sensing/pointcloud_preprocessor/src/outlier_filter/ring_outlier_filter_nodelet.cpp
+++ b/sensing/pointcloud_preprocessor/src/outlier_filter/ring_outlier_filter_nodelet.cpp
@@ -69,6 +69,8 @@ void RingOutlierFilterComponent::faster_filter(
     input->fields.at(static_cast<size_t>(autoware_point_types::PointIndex::Azimuth)).offset;
   const auto distance_offset =
     input->fields.at(static_cast<size_t>(autoware_point_types::PointIndex::Distance)).offset;
+  const auto intensity_offset =
+    input->fields.at(static_cast<size_t>(autoware_point_types::PointIndex::Intensity)).offset;
 
   std::vector<std::vector<size_t>> ring2indices;
   ring2indices.reserve(max_rings_num_);
@@ -132,10 +134,12 @@ void RingOutlierFilterComponent::faster_filter(
             output_ptr->x = p[0];
             output_ptr->y = p[1];
             output_ptr->z = p[2];
-            output_ptr->intensity = input_ptr->intensity;
           } else {
             *output_ptr = *input_ptr;
           }
+          const float & intensity =
+            *reinterpret_cast<const float *>(&input->data[indices[i] + intensity_offset]);
+          output_ptr->intensity = intensity;
 
           output_size += output.point_step;
         }
@@ -159,10 +163,12 @@ void RingOutlierFilterComponent::faster_filter(
           output_ptr->x = p[0];
           output_ptr->y = p[1];
           output_ptr->z = p[2];
-          output_ptr->intensity = input_ptr->intensity;
         } else {
           *output_ptr = *input_ptr;
         }
+        const float & intensity =
+          *reinterpret_cast<const float *>(&input->data[indices[i] + intensity_offset]);
+        output_ptr->intensity = intensity;
 
         output_size += output.point_step;
       }
@@ -180,6 +186,8 @@ void RingOutlierFilterComponent::faster_filter(
     input->fields.begin(),
     input->fields.begin() + static_cast<size_t>(autoware_point_types::PointIndex::Intensity) + 1,
     output.fields.begin());
+
+  output.fields[static_cast<size_t>(autoware_point_types::PointIndex::Intensity)].offset = 12;
 
   output.height = 1;
   output.is_bigendian = input->is_bigendian;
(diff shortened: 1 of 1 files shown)


VRichardJP:
Is is the same problem than for `pcl::PointXYZI`?
https://pcl.readthedocs.io/projects/tutorials/en/latest/adding_custom_ptype.html#id3

Where the XYZ is stored on 4 floats instead of 3, so the XYZ data is SSE-aligned. As describe, operation such as dot product can leverage SSE optimization by setting the 4-th value to 0. Like this:
```C++
float dot(const T& p1, const T& p2) {
  p2.data[0] = 0;
  // optimized with SIMD instructions
  float sum0 = p1.data[0] * p2.data[0];
  float sum1 = p1.data[1] * p2.data[1];
  float sum2 = p1.data[2] * p2.data[2];
  float sum3 = p1.data[3] * p2.data[3]; // must be 0
  float res = sum0 + sum1 + sum2 + sum3;
  return res;
}
```

VRichardJP:
What is the state of this PR?

Just to clarify the issue, the root cause is that `velodyne_convert_node` output data layout is incompatible with `autoware_point_types::PointXYZIRADRT`. It seems that `velodyne_convert_node` use the same memory layout than PCL, which adds a 4 byte padding after xyz fields (see my previous comment).

```C++
struct PointXYZIRADRT
{
  float x{0.0F};           // offset: 0
  float y{0.0F};           // offset: 4
  float z{0.0F};           // offset: 8
  float intensity{0.0F};   // offset: 12
  uint16_t ring{0U};       // offset: 16
  float azimuth{0.0F};     // offset: 20
  float distance{0.0F};    // offset: 24
  uint8_t return_type{0U}; // offset: 28
  double time_stamp{0.0};  // offset: 32
}                          // sizeof(PointXYZIRADRT) == 40
```

We can see the extra padding here:
```
$ ros2 topic echo /sensing/lidar/top/pointcloud_raw --field fields
[sensor_msgs.msg.PointField(name='x', offset=0, datatype=7, count=1), 
sensor_msgs.msg.PointField(name='y', offset=4, datatype=7, count=1), 
sensor_msgs.msg.PointField(name='z', offset=8, datatype=7, count=1), 
sensor_msgs.msg.PointField(name='intensity', offset=16, datatype=7, count=1), <-- jump here
sensor_msgs.msg.PointField(name='ring', offset=20, datatype=4, count=1)]
```

On the extended cloud, this cause paddings because of `double` alignment requirements for `time_stamp` field:
```
$ ros2 topic echo /sensing/lidar/top/pointcloud_raw_ex --field fields
[sensor_msgs.msg.PointField(name='x', offset=0, datatype=7, count=1), 
sensor_msgs.msg.PointField(name='y', offset=4, datatype=7, count=1), 
sensor_msgs.msg.PointField(name='z', offset=8, datatype=7, count=1), 
sensor_msgs.msg.PointField(name='intensity', offset=16, datatype=7, count=1), <-- jump here
sensor_msgs.msg.PointField(name='ring', offset=20, datatype=4, count=1), 
sensor_msgs.msg.PointField(name='azimuth', offset=24, datatype=7, count=1), 
sensor_msgs.msg.PointField(name='distance', offset=28, datatype=7, count=1), 
sensor_msgs.msg.PointField(name='return_type', offset=32, datatype=2, count=1), 
sensor_msgs.msg.PointField(name='time_stamp', offset=40, datatype=8, count=1)] <-- second jump here
```

Because of the incompatible data layout, it is not possible to do things like:
```C++
PointXYZIRADRT p;
// extract point i
std::memcpy(&p, &input->data[i * input->point_step], sizeof(PointXYZIRADRT));
```
(Without this PR you can still do so but you will just end up with garbage data past the XYZ fields.)

miursh:
@VRichardJP I'm sorry I haven't been able to handle this PR due to a lack of time.
I think the root cause is that this filter is forcibly casting from PointXYZIRADRT to PointXYZIR without reading the PointField in the header. 
Since autoware_point_types/types.hpp does not guarantee the memory alignment/order of the member (xyzir...) in the pointcloud2, it should read PointField and parse the offset of each variables.
Even this PR does not guarantee the the alignment of variables but I believe that adding padding, as done in this PR, can resolve the issue because this memory layout is commonly observed in many environments.

miursh:
@VRichardJP 
We recently are re-designing the [pointcloud fields](https://autowarefoundation.github.io/autoware-documentation/main/design/autoware-architecture/sensing/data-types/point-cloud/#point-cloud-fields) in autoware and discussing [here](https://github.com/orgs/autowarefoundation/discussions/3442).
Once it's finalized, I think all pointcloud nodes must be modified to follow this design. 
When we do this, you may propose your "pointcloud formatter" to be in ther first position of the pipeline. Or maybe we don't need formatter anymore? :thinking:

yukkysaito:
I'll review later🙏

yukkysaito (approved):
LGTM

PR #2621, opened 2023-01-09
feat(freespace_planning_algorithms): fix performance issue in freespace planning with respect to collision table initialization using A-Search algo

Narendarselva:
### **Description**
Collison and vertex tables are populated to infer the vehicle boundaries to detect the collison with any static or dynamic objects on the occupancy_grid. And these tables are just inference tables with base index and theta size {0,0,144*}. Which means these are references to generate the boundaries on the local co-oridantes based on the individual trajectory point's orientation. Populating the collision and vertex index table for every setMap affects the whole freespace planning algo performance.Added fix to populate these index tables only once when we do setMap. Which this change was able to see upto 50% increase in freespace planning performance.

### **Test Performed**
Performed the UT testcases for freespace_planning_algorithms.PFA for the reports.

plot-rrtstar_update-case3
plot-rrtstar_update-case2
plot-rrtstar_update-case1
plot-astar_multi-case3
plot-astar_multi-case2
plot-astar_multi-case1
plot-astar_multi-case0
plot-rrtstar_update-case0
plot-rrtstar_informed_update-case3
plot-rrtstar_informed_update-case2
plot-rrtstar_informed_update-case1
plot-rrtstar_informed_update-case0
plot-rrtstar_fastest-case3
plot-rrtstar_fastest-case2
plot-rrtstar_fastest-case1
plot-rrtstar_fastest-case0
plot-astar_single-case3
plot-astar_single-case2
plot-astar_single-case1
plot-astar_single-case0
plot-astar_multi-case2


TakaHoribe:
@Narendarselva Thank you for your PR. I saw your description and code, but still don't know why we can get rid of the update of the collision table for the new costmap. If all obstacles in the costmap are static, then theoretically we can omit the update because the table in local coordinate does not change. But how about the case where dynamic obstacles are moving in the cost map?

Narendarselva:
@TakaHoribe Thanks for the comment. If you look at the "computeCollisionIndexes" function which is use to populate the coll_index_table , this table is not generated based out of costmap. This function just creates a reference boudaries for discretized theta angles for the ego vehicle with reference to x=0 , y=0 co-ordinate. So based on the theta size it creates collision boundaries for ego vehicle with different discretized angle (subjected to pose and turining radius(theta size)) but x & y co-ordinates fixed to (0,0). Just for an example I just simulated the "computeCollisionIndexes" and plotted the same with x=10 & y=10 but with different orientations.
Screenshot from 2023-01-10 18-17-59

So how this is inferred with detecting the collisions ???
For that if we look at the "detectCollision" function , the input to the collision function is individual pose in the trajectory that is created out of freespace planning algo. So these individual poses are converted to local base_index and then the collision boundaries are calculated using the "coll_index_table which is created with x=0 y=0"
    **// must slide to current base position
    coll_index.x += base_index.x;
    coll_index.y += base_index.y;**
Just by adding the trajectory's pose x & y to 0 ,0 it slides the collision boundary to respective co-ordinates.
Kindly correct if my understanding is wrong.

TakumIto:
Your explanation seems to be correct.
The "computeCollisionIndexes" function does not use obstacle data, only vehicle information.

I have confirmed that this PR works correctly the case where dynamic obstacles are moving and Pullover feature.
See the attached videoes.
[pr2621_goalplanner.webm](https://user-images.githubusercontent.com/61740530/236781277-69cde1f8-8c07-470f-95a4-e313c49de48a.webm)
[pr2621_parking.webm](https://user-images.githubusercontent.com/61740530/236781310-acd09e64-e426-4b1c-82d6-035ed983d110.webm)

kosuke55:
I think it is ok because the `coll_indexes_table_` is just a mapping to the occupancy grid at a certain posture, so it should be calculated only once. `is_obstacle_table_` determines if there are any obstacles, and only this need to be updated at each `setMap`.

kosuke55 (approved):

kosuke55:
@Narendarselva @TakumIto 
Sorry for being so late and thank you for your contribution.

Narendarselva:
> @Narendarselva @TakumIto Sorry for being so late and thank you for your contribution.

No worries @TakumIto Happy that I made contributions.

PR #2656, opened 2023-01-16
refactor(ekf_localizer): remove info structs

IshitaTakeshi:
## Description

Structs inhibit flat software design. I removed the pose / twist structs to flatten the software structure. 

### DIFF
--- a/localization/ekf_localizer/include/ekf_localizer/ekf_localizer.hpp
+++ b/localization/ekf_localizer/include/ekf_localizer/ekf_localizer.hpp
@@ -51,14 +51,12 @@ struct PoseInfo
 {
   geometry_msgs::msg::PoseWithCovarianceStamped::SharedPtr pose;
   int counter;
-  int smoothing_steps;
 };
 
 struct TwistInfo
 {
   geometry_msgs::msg::TwistWithCovarianceStamped::SharedPtr twist;
   int counter;
-  int smoothing_steps;
 };
 
 class Simple1DFilter
--- a/localization/ekf_localizer/include/ekf_localizer/ekf_localizer.hpp
+++ b/localization/ekf_localizer/include/ekf_localizer/ekf_localizer.hpp
@@ -174,7 +174,8 @@ private:
     current_twist_queue_;  //!< @brief current measured twist
   std::queue<int> current_twist_count_queue_;
 
-  std::queue<PoseWithCovarianceStamped::SharedPtr> current_pose_queue_;  //!< @brief current measured pose
+  std::queue<PoseWithCovarianceStamped::SharedPtr>
+    current_pose_queue_;  //!< @brief current measured pose
   std::queue<int> current_pose_count_queue_;
 
   geometry_msgs::msg::PoseStamped current_ekf_pose_;  //!< @brief current estimated pose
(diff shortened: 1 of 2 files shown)


localization/ekf_localizer/include/ekf_localizer/ekf_localizer.hpp
```diff
@@ -47,19 +47,8 @@
...
-  int smoothing_steps;
-};
-
-struct TwistInfo
-{
-  geometry_msgs::msg::TwistWithCovarianceStamped::SharedPtr twist;
-  int counter;
-  int smoothing_steps;
-};
+using PoseWithCovarianceStamped = geometry_msgs::msg::PoseWithCovarianceStamped;
```
kminoda, line 50: I think it is safer to declare this inside EKFLocalizer

kminoda: Also, what do you think about declaring PoseInfo using `std::pair` like follows?
```
using PoseInfo = std::pair<geometry_msgs::msg::PoseWithCovarianecStamped, int>
```

I agree that it is better avoiding using original struct, but I also want to make `pose` and its `pose_count` stored in one variable instead of two separate variables to make it clear that those two values are tightly-coupled and must be updated synchronously.

IshitaTakeshi: Thank you for pointing out this. I moved the type aliases.

IshitaTakeshi: I think the first thing we should do is to reduce the variable scope. 
The variables themselves are not the major cause of the confusion. The large variable scopes are. 
We need to make a separated class to limit the scope.

kminoda: (sorry for being late)
This PR does NOT change the variable scope, does it? (Maybe I am misunderstanding the term "scope")

IshitaTakeshi: Yes this PR itself does not change the scope, but if we do it here the code difference is going to be large so I'd like to do in another PR

kminoda: I see, thank you!
But then, I doubt the necessity of adding a new test script just for this simple inequality. Of course it will be nice in terms of coverage, but then we may going to have a number of test scripts.

kminoda: Ok, I see, thank you!

localization/ekf_localizer/include/ekf_localizer/ekf_localizer.hpp
```diff
@@ -117,6 +103,9 @@ class EKFLocalizer : public rclcpp::Node
   EKFLocalizer(const std::string & node_name, const rclcpp::NodeOptions & options);
 
 private:
+  using PoseWithCovarianceStamped = geometry_msgs::msg::PoseWithCovarianceStamped;
```
kminoda, line 106: This is just a minor comment, but maybe we can incorporate this literal in another PR, since currently this literal is used only in two parts (and within this script and not in .cpp)

IshitaTakeshi: Fixed in [109ffa7](https://github.com/autowarefoundation/autoware.universe/pull/2656/commits/109ffa77c0dd8523f1cfad0f45b88874b4e62238)

kminoda (approved):
LGTM

PR #2657, opened 2023-01-16
refactor(ekf_localizer): add a function to check the delay step

IshitaTakeshi:
## Description

Checking the delay step in the callback function is redundant. I added a function to do this and improved the readability. 
Note that the warning message is modified to reduce the number of arguments. 

## Tests performed

Smoke tests to check the delay step boundary


localization/ekf_localizer/src/check.cpp
```diff
@@ -0,0 +1,31 @@
...
+
+#include "ekf_localizer/check.hpp"
+
+#include "ekf_localizer/warning.hpp"
+
+#include <fmt/core.h>
+
+#include <string>
+
+bool checkDelayStep(const Warning & warning, const int delay_step, const int extend_state_step)
```
kminoda, line 23: How about something like this?
```suggestion
bool isDelayStepBelowThreshold(const Warning & warning, const int delay_step, const int delay_step_threshold)
```

`delay_step_threshold` can also be something like `maximum_state_step` or `maximum_delay_step`.

IshitaTakeshi: Updated [38426bc](https://github.com/autowarefoundation/autoware.universe/pull/2657/commits/38426bc828f44a8619954193b7448b49dbc08b67), [296b6bb](https://github.com/autowarefoundation/autoware.universe/pull/2657/commits/296b6bb403b0c528e061af59d792e7800672210c)

localization/ekf_localizer/src/ekf_localizer.cpp
```diff
@@ -428,13 +430,8 @@ void EKFLocalizer::measurementUpdatePose(const geometry_msgs::msg::PoseWithCovar
...
 
   int delay_step = std::roundf(delay_time / ekf_dt_);
-  if (delay_step > params_.extend_state_step - 1) {
-    warning_.warnThrottle(
-      fmt::format(
-        "Pose delay exceeds the compensation limit, ignored. delay: %f[s], limit = "
-        "extend_state_step * ekf_dt : %f [s]",
-        delay_time, params_.extend_state_step * ekf_dt_),
-      1000);
+  if (!delayStepIsSufficientlySmall(delay_step, params_.extend_state_step)) {
```
kminoda, line 433: Do we still need this function? How about this?
```suggestion
  if (delay_step >= params_.extend_state_step) {
```

localization/ekf_localizer/src/check.cpp
kminoda: Thank you!

localization/ekf_localizer/src/ekf_localizer.cpp
IshitaTakeshi: That's true. So only the difference is whether we can test it or not.

kminoda: I see, thank you!
But then, I doubt the necessity of adding a new test script just for this simple inequality. Of course it will be nice in terms of coverage, but then we may going to have a number of test scripts.

IshitaTakeshi: How about keeping it this time and modify when we need to change or when we can cover it in another test?

kminoda: > How about keeping it this time

Is you suggestion to
- keep this line as `if (delay_step >= params_.extend_state_step) {` in this PR
- isolate this line as another function to cover it in a test, if needed in future (e.g. the function gets more complicated)

? Then yes, I agree

IshitaTakeshi: I mean I want coverage so keeping `delayStepIsSufficientlySmall` and its test, then in the future if we can isolate some parts in `measurementUpdatePose` including `delayStepIsSufficientlySmall`, we can safely remove `delayStepIsSufficientlySmall` and its test.

kminoda: Hmm, I believe we should revert the `delayStepIsSufficientlySmall` function, since 
- the advantage of testing the inequality operator of int seems small
- it decreases the readability IMHO (we should look into another file just to see the simple inequality operation)

I understand the importance of test coverage, but in this particular case I think we don't need an additional test.

IshitaTakeshi: Resolved in [78875d0](https://github.com/autowarefoundation/autoware.universe/pull/2657/commits/78875d0008c9132af525cd87ac4bd1d7e95fc68f)

kminoda: Thank you!

localization/ekf_localizer/CMakeLists.txt
```diff
@@ -46,7 +47,8 @@ if(BUILD_TESTING)
     test/test_mahalanobis.cpp
     test/test_measurement.cpp
     test/test_numeric.cpp
-    test/test_state_transition.cpp)
+    test/test_state_transition.cpp
+    test/test_warning_message.cpp)
```
kminoda, line 51: IMHO this test is better not to be added, either.

This is too dependent on the implementation, and thus would increase a cost of modifying them (e.g. changing the error messages). Also, I see little advantage of adding this test, since it is essentially a test for `fmt::format`.

What do you think?

kminoda: (Same argument applies to https://github.com/autowarefoundation/autoware.universe/pull/2694)

IshitaTakeshi: Let's say we modify the warning message in the future, and with some mistake we forget to remove `{}` in the format string.

```cpp
const std::string s = "The delay step {} should be less than the maximum state step {}. Some forgotten comment {}";
```

And then we run this code, C++ throws the error at the **runtime**, not the compile time.

```
C++ exception with description "argument not found" thrown in the test body.
```

To avoid this runtime error, we need this test.

kminoda: I see, that make sense. Thank you :+1:

localization/ekf_localizer/src/ekf_localizer.cpp
```diff
@@ -428,13 +429,8 @@ void EKFLocalizer::measurementUpdatePose(const geometry_msgs::msg::PoseWithCovar
...
   int delay_step = std::roundf(delay_time / ekf_dt_);
-  if (delay_step > params_.extend_state_step - 1) {
-    warning_.warnThrottle(
-      fmt::format(
-        "Pose delay exceeds the compensation limit, ignored. delay: %f[s], limit = "
-        "extend_state_step * ekf_dt : %f [s]",
-        delay_time, params_.extend_state_step * ekf_dt_),
-      1000);
+  if (delay_step >= params_.extend_state_step) {
+    warning_.warnThrottle(delayStepWarningMessage(delay_step, params_.extend_state_step), 2000);
```
kminoda, line 433: As well as https://github.com/autowarefoundation/autoware.universe/pull/2694#discussion_r1084867985, I think the message should provide pose/twist information.

IshitaTakeshi: Updated in [880dc3f](https://github.com/autowarefoundation/autoware.universe/pull/2657/commits/880dc3fd335d53238e5d3be497a92a5e7b9eca5c)

localization/ekf_localizer/include/ekf_localizer/warning_message.hpp
```diff
@@ -0,0 +1,25 @@
+// Copyright 2022 Autoware Foundation
```
kminoda, line 1: I will approve after the conflict is resolved!

localization/ekf_localizer/include/ekf_localizer/warning_message.hpp
```diff
@@ -0,0 +1,25 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef EKF_LOCALIZER__WARNING_MESSAGE_HPP_
+#define EKF_LOCALIZER__WARNING_MESSAGE_HPP_
+
+#include <string>
+
+std::string poseDelayStepWarningMessage(
+  const double delay_step, const double extend_state_step, const double ekf_dt);
```
kminoda, line 21: It's better to make `extended_state_step` as `int`, given that it is defined so in `hyper_parameters.hpp`.
```suggestion
  const double delay_step, const int extend_state_step, const double ekf_dt);
```

localization/ekf_localizer/include/ekf_localizer/warning_message.hpp
IshitaTakeshi: Resolved the conflicts

kminoda (approved):
LGTM

PR #2694, opened 2023-01-19
refactor(ekf_localizer): add a function to generate a delay time warning message

IshitaTakeshi:
## Description

To simplify the code, I added a function to generate a delay time warning message

## Tests performed

Smoke tests to check the generated message


localization/ekf_localizer/CMakeLists.txt
```diff
@@ -46,7 +47,8 @@ if(BUILD_TESTING)
     test/test_mahalanobis.cpp
     test/test_measurement.cpp
     test/test_numeric.cpp
-    test/test_state_transition.cpp)
+    test/test_state_transition.cpp
+    test/test_warning_message.cpp)
```
kminoda, line 51: I personally think this test for warning messages is not necessary. What do you think?
(same as https://github.com/autowarefoundation/autoware.universe/pull/2657#discussion_r1080865663)

kminoda (commented):
Thank you :+1: I left a comment about the test

localization/ekf_localizer/CMakeLists.txt
IshitaTakeshi: I think we need the test. The message function is rarely executed when running the logging simulator, so it is hard to detect a bug if we wrongly set the string format.

IshitaTakeshi: The basic concept is "do not believe in human eyes." 
Humans can often make mistakes in very easy and obvious parts. Even if it looks unnecessary, we should add tests.

kminoda: This is too dependent on the implementation, and thus would increase a cost of modifying them in the future (e.g. changing the error messages). Also, I see little advantage of adding this test, since it is essentially a test for fmt::format.

But this is just my opinion, so I will leave the decision up to @YamatoAndo @TakaHoribe .

localization/ekf_localizer/src/warning_message.cpp
```diff
@@ -0,0 +1,22 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "ekf_localizer/warning_message.hpp"
+
+#include <string>
+
+std::string delayTimeWarningMessage(const double delay_time)
+{
+  return fmt::format("Delay time {} cannot be less than zero.", delay_time);
```
kminoda, line 21: The error message is changed from the original one. I think it is better to leave the message as it was.

IshitaTakeshi: The reason I changed the error message is because the original error message says `Pose time stamp is inappropriate` so it cannot be used for both pose and twist. 
Also, `delay_time = max(delay_time, 0);` is not included in the `delayTimeWarningMessage` function, so if we want to notify that `set delay to 0[s]`, we should add another `warning_.warnThrottle` in `ekf_localizer.cpp`.

localization/ekf_localizer/CMakeLists.txt
YamatoAndo: As kminoda-san said, I don't think it is necessary to make a function to test if the message content is as intended.
But since IshitaTakeshi-san has already made it, I think we can merge this PR.

localization/ekf_localizer/src/warning_message.cpp
kminoda: I got the argument about `set delay to 0[s]` , but I think we should leave the pose/twist message as it could be useful.

IshitaTakeshi: Updated in [d83f293](https://github.com/autowarefoundation/autoware.universe/pull/2694/commits/d83f293dae87f1bed33d7651320be8ed1f5bc34b)

kminoda (approved):
LGTM

PR #2702, opened 2023-01-20
refactor(ekf_localizer): add a function to erase a slash from the beginning of a string

IshitaTakeshi:
## Description

Add a function to erase a slash '/' from a beginning of a string 

## Tests performed

Check if the slash can be removed


localization/ekf_localizer/src/ekf_localizer.cpp
```diff
@@ -297,12 +298,9 @@ bool EKFLocalizer::getTransformFromTF(
   tf2::BufferCore tf_buffer;
   tf2_ros::TransformListener tf_listener(tf_buffer);
   rclcpp::sleep_for(std::chrono::milliseconds(100));
-  if (parent_frame.front() == '/') {
```
kminoda, line 300: It seems a bit out of scope of this PR, but how about using this? https://github.com/autowarefoundation/autoware.universe/blob/main/common/tier4_autoware_utils/include/tier4_autoware_utils/ros/transform_listener.hpp

This getTransformFromTF function is a legacy function according to @.YamatoAndo -san, so I think we can completely replace it with the above library class.

IshitaTakeshi: If we don't repeat calling lookupTransform, or if we just show the warning message and do nothing more than this, I can use the proposed function.

kminoda: @IshitaTakeshi The `EKFLocalizer::getTransformFromTF` function has some tricky code such as `rclcpp::sleep_for(std::chrono::milliseconds(100));`, so I think using the one from tier4_autoware_utils would make much more sense than leaving this function as it is. Given that this tier4_autoware_utils::transform_listener is already used in several packages, I think it is OK to replace it. (not necessarily in this PR, though)

IshitaTakeshi: OK then I'm going to do it or continue the discussion on another PR so could you approve this PR first?

localization/ekf_localizer/include/ekf_localizer/string.hpp
```diff
@@ -0,0 +1,29 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef EKF_LOCALIZER__STRING_HPP_
+#define EKF_LOCALIZER__STRING_HPP_
+
+#include <string>
+
+inline std::string eraseBeginSlash(const std::string & s)
```
kminoda, line 20: How about `eraseLeadingSlash` or `eraseLeadingForwardSlash`?

kminoda (approved):

localization/ekf_localizer/src/ekf_localizer.cpp
kminoda: OK :+1:

localization/ekf_localizer/include/ekf_localizer/string.hpp
kminoda: (Sorry I approved mistakenly, so dismissed it)

IshitaTakeshi: Updated in [e95c689](https://github.com/autowarefoundation/autoware.universe/pull/2702/commits/e95c68914665df278eda5eddfce8904383789f4b)

kminoda (approved):
LGTM

PR #2718, opened 2023-01-23
fix(behavior_velocity_planner): fix misc concurrency issues

### LINKED ISSUE #2706: Misc concurrency issues in `behavior_velocity_planner`
### Checklist

### Description

Similar to https://github.com/autowarefoundation/autoware.universe/issues/2495, I have found several concurrency issues in [src/universe/autoware.universe/planning/behavior_velocity_planner/src/node.cpp](https://github.com/autowarefoundation/autoware.universe/blob/main/planning/behavior_velocity_planner/src/node.cpp):
- `BehaviorVelocityPlannerNode::onParam`: missing mutex lock for `planner_data_`
- `BehaviorVelocityPlannerNode::onExternalVelocityLimit`: missing mutex lock for `planner_data_`
- `BehaviorVelocityPlannerNode::onTrigger`: could make use of `std::lock_guard` instead of manually locking/unlocking the mutex. Also, it suffers from the same issue described in https://github.com/autowarefoundation/autoware.universe/issues/2495: making a shallow copy of `planner_data_` does not solve concurrency problems: when we copy `planner_data_`, we duplicate the member pointers, not pointed data. As a consequence, the "copied" `planner_data` actually points to the same data than `planner_data_`, and concurrent read/write may occur.

### Expected behavior

No concurrent access of protected data.

### Actual behavior

Concurrent access of protected data.

### Steps to reproduce

N/A

### Versions

_No response_

### Possible causes

_No response_

### Additional context

_No response_

VRichardJP:
Signed-off-by: Vincent Richard <richard-v@macnica.co.jp>

## Description

Fix #2706

### DIFF
--- a/planning/behavior_velocity_planner/src/node.cpp
+++ b/planning/behavior_velocity_planner/src/node.cpp
@@ -324,9 +324,10 @@ void BehaviorVelocityPlannerNode::onAcceleration(
 
 void BehaviorVelocityPlannerNode::onParam()
 {
+  // Note(vrichard): mutex lock is not necessary as onParam is only called once in the constructed. It would be required if it was a callback.
+  // std::lock_guard<std::mutex> lock(mutex_);
   planner_data_.velocity_smoother_ =
     std::make_unique<motion_velocity_smoother::AnalyticalJerkConstrainedSmoother>(*this);
-  return;
 }
 
 void BehaviorVelocityPlannerNode::onLaneletMap(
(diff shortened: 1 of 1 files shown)


soblin:
@VRichardJP Thanks for your PR and sorry for my late response :)
For the most part I agree with this change:

Now behavior_velocity_planner assigns different callback groups to each topic, each spinner thread (of type MutuallyExclusive) run in parallel so we have to lock/unlock the `planner_data_` in each callback if it has access to `planner_data_` as you pointed out. Using RAII style locking in `onTrajectory` is also a good for avoiding careless mistake.

I think the lock in `onParam` is unnecessary because actually this function is not a callback but is just used in the constructor only once. I'm sure this additional lock does not have any side effects like dead locking, so can you replace it with some comment ?

VRichardJP:
@soblin I overlooked the `onParam`. Indeed the mutex lock is not necessary, but I added a memo as you suggested in case someone would make it a callback.

soblin:
@VRichardJP thanks, I'll approve this.

soblin (approved):
LGTM

Issue #2728, opened 2023-01-24
Improve motion_velocity_smoother's latency


veqcc:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I've agreed with the maintainers that I can plan this task.

### Description

cc: @soblin 

The latency of motion_velocity_smoother node is currently very big and has a negative influence on other nodes in Autoware.
The following figure is the latency of motion_velocity_smoother node.
![image](https://user-images.githubusercontent.com/26951562/214265004-54bae8dc-1ac0-44be-80b1-8191067e5e7c.png)

### Purpose

Our mission is to improve the latency by refactoring the implemenration.
This optimization would benefit for not only this node but also other nodes using the same QP solver.

### Possible approaches

The main cause of the large latencies is considered to be `MotionVelocitySmootherNode::smoothVelocity` function.
This function includes OSQP (one of QP solvers) and a lot of resampling procedures.
Thus the possible approaches are:
 - solve QP faster
   - use another QP solver
   - use OSQP parameters which Autoware is not currently using
 - refactor resampling implementation
   - Autoware give the same parameters to resampling functions several times. We assume we can omit the redundant executions

### Definition of done

 - measure the performance in realistic cases with various trajectories and with other QP solvers (or other OSQP options)
 - 2x ~ 3x performance improvement would be good

PR #2744, opened 2023-01-25
perf(motion_utils): improve performance of zero_order_hold

veqcc:
Signed-off-by: Ryuta Kambe [ryuta.kambe@tier4.jp](mailto:ryuta.kambe@tier4.jp)

# Description
There are two changes in this Pull Request:
 - when `closest_segment_idx` is found in backward loop in `zero_order_hold` function, immediatly break the loop. The previous implementation was forward and continue updating the idx value, which is unnecessary.
 - pre-calculate the `closest_segment_idx` with base and query keys. In the several calls of  `zero_order_hold` from `resample.cpp` in the form of `zoh`, the two parameters `base_keys` and `query_keys` are the same. Thus the calculation of `closest_segment_idx` is also the same, which includes double for loops and has a heavy computational cost. 

The following figure is the latency of [applyLateralAccelerationFilter](https://github.com/autowarefoundation/autoware.universe/blob/11ce12b63c9067b1c8843a3a05cc8c902e438faa/planning/motion_velocity_smoother/src/motion_velocity_smoother_node.cpp#L539)  ~ [resampleTrajectory](https://github.com/autowarefoundation/autoware.universe/blob/11ce12b63c9067b1c8843a3a05cc8c902e438faa/planning/motion_velocity_smoother/src/motion_velocity_smoother_node.cpp#L558), which calls `resampleTrajectory` in `resample.cpp` three times.
image

# Related links
https://github.com/autowarefoundation/autoware.universe/issues/2728

### DIFF
--- a/common/interpolation/include/interpolation/zero_order_hold.hpp
+++ b/common/interpolation/include/interpolation/zero_order_hold.hpp
@@ -21,36 +21,57 @@
 
 namespace interpolation
 {
+inline std::vector<size_t> calc_closest_segment_indices(
+  const std::vector<double> & base_keys, const std::vector<double> & query_keys,
+  const double overlap_threshold = 1e-3)
+{
+  // throw exception for invalid arguments
+  const auto validated_query_keys = interpolation_utils::validateKeys(base_keys, query_keys);
+
+  std::vector<size_t> closest_segment_indices(validated_query_keys.size(), 0);
+  for (size_t i = 0; i < validated_query_keys.size(); ++i) {
+    // Check if query_key is closes to the terminal point of the base keys
+    if (base_keys.back() - overlap_threshold < validated_query_keys.at(i)) {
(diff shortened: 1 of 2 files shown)


TakaHoribe:
Nice work!

soblin:
It looks like still the behavior_planning node dies in some scenarios.

https://evaluation.tier4.jp/evaluation/reports/9078a3a2-704d-566b-80eb-832d62e74e67?project_id=prd_jt

purewater0901 (approved):
LGTM

soblin (approved):
I ran the CI again for the failed cases above [here](https://evaluation.tier4.jp/evaluation/reports/436cb3e0-de36-58df-a215-a94d163c346f?project_id=prd_jt), and although some of them did not start due to state transition error at that time, they were successful on my local computer with the latest rebase version.
I'm sure we can merge this PR.

takayuki5168 (approved):

Issue #2782, opened 2023-01-30
bag_time_manager_rviz_plugin fails to build -Werror=deprecated-declarations


esteve:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I'm convinced that this is not my fault but a bug.

### Description

The following lines use a deprecated API (`rmw_qos_profile_t`) which makes `bag_time_manager_rviz_plugin` fail to build:

https://github.com/autowarefoundation/autoware.universe/blob/main/common/bag_time_manager_rviz_plugin/src/bag_time_manager_panel.cpp#L69-L74

### Expected behavior

`bag_time_manager_rviz_plugin` builds successfully with `-Werror=deprecated-declarations` enabled

### Actual behavior

`bag_time_manager_rviz_plugin` fails to build with `-Werror=deprecated-declarations` enabled

### Steps to reproduce

Add `-Werror=deprecated-declarations` as a compiler flag

taikitanaka3:
@esteve 
There are many places we should change to support rolling. So I think we should discuss whether to support rolling or not.

https://github.com/autowarefoundation/autoware.universe/blob/72470412a319238f84e781799f133d832a8c8f5c/common/tier4_state_rviz_plugin/src/autoware_state_panel.cpp

@kenji-miyake @yukkysaito 
what do you think about this?

isamu-takagi:
> There are many places we should change to support rolling. So I think we should discuss whether to support rolling or not.

@taikitanaka3 @esteve 
There are probably other changes that need to be made to work with rolling. If we support rolling, it is good to create an rolling branch first. And confirming that it works, then checking at the differences to decide whether to merge.

But my opinion is not to support rolling yet as it increases maintenance cost. Is there any reason you want to support rolling?

xmfcx:
@esteve I've made some changes to make maintenance costs minimum for this PR on following commit: https://github.com/autowarefoundation/autoware.universe/commit/c79c2b103c63adf684de037bc5521cd2deb49888 @isamu-takagi @taikitanaka3 do you think this would be acceptable?

The point is to make it easier to update for the future and at the same time have small maintenance costs in general.

taikitanaka3:
@xmfcx 
[...]

PR #2783, opened 2023-01-30
fix(bag_time_manager_rviz_plugin): update create_client API to use rc…

esteve:
## Description

`rclcpp::create_client` in `rolling` deprecated `rmw_qos_profile_t` in favor of the `rclcpp::QoS` API

Fixes https://github.com/autowarefoundation/autoware.universe/issues/2782

### DIFF
--- a/common/bag_time_manager_rviz_plugin/src/bag_time_manager_panel.cpp
+++ b/common/bag_time_manager_rviz_plugin/src/bag_time_manager_panel.cpp
@@ -66,10 +66,8 @@ BagTimeManagerPanel::BagTimeManagerPanel(QWidget * parent) : rviz_common::Panel(
 void BagTimeManagerPanel::onInitialize()
 {
   raw_node_ = this->getDisplayContext()->getRosNodeAbstraction().lock()->get_raw_node();
-  client_pause_ =
-    raw_node_->create_client<Pause>("/rosbag2_player/pause", rclcpp::ServicesQoS);
-  client_resume_ =
-    raw_node_->create_client<Resume>("/rosbag2_player/resume", rclcpp::ServicesQoS);
+  client_pause_ = raw_node_->create_client<Pause>("/rosbag2_player/pause", rclcpp::ServicesQoS);
+  client_resume_ = raw_node_->create_client<Resume>("/rosbag2_player/resume", rclcpp::ServicesQoS);
   client_set_rate_ =
     raw_node_->create_client<SetRate>("/rosbag2_player/set_rate", rclcpp::ServicesQoS);
 }
--- a/common/bag_time_manager_rviz_plugin/src/bag_time_manager_panel.cpp
+++ b/common/bag_time_manager_rviz_plugin/src/bag_time_manager_panel.cpp
@@ -19,6 +19,7 @@
 #include <qt5/QtWidgets/QHBoxLayout>
 #include <qt5/QtWidgets/QLabel>
 #include <qt5/QtWidgets/QWidget>
+#include <rclcpp/version.h>
 #include <rviz_common/display_context.hpp>
 
 namespace rviz_plugins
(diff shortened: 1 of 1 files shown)


taikitanaka3:
thanks for pr, is this also tested by humble version？

esteve:
> @esteve
> thanks for pr, is this also tested by humble version？

This PR should work with Humble, but I'm building it right now to check, I'll report with the results, don't approve it yet. Thanks.

esteve:
@taikitanaka3 I fixed the latest commit because the changes were not building for Humble and lower.

taikitanaka3:
@esteve 
can you take a look at this comment https://github.com/autowarefoundation/autoware.universe/issues/2782 ?

xmfcx:
@esteve I've made some changes to make maintenance costs minimum for this PR on following commit: https://github.com/autowarefoundation/autoware.universe/commit/c79c2b103c63adf684de037bc5521cd2deb49888

esteve:
@xmfcx looks good, can you push it here? Thanks.

xmfcx (approved):

common/bag_time_manager_rviz_plugin/src/bag_time_manager_panel.cpp
```diff
@@ -66,12 +68,17 @@ BagTimeManagerPanel::BagTimeManagerPanel(QWidget * parent) : rviz_common::Panel(
...
   raw_node_ = this->getDisplayContext()->getRosNodeAbstraction().lock()->get_raw_node();
-  client_pause_ =
-    raw_node_->create_client<Pause>("/rosbag2_player/pause", rmw_qos_profile_services_default);
-  client_resume_ =
-    raw_node_->create_client<Resume>("/rosbag2_player/resume", rmw_qos_profile_services_default);
-  client_set_rate_ =
-    raw_node_->create_client<SetRate>("/rosbag2_player/set_rate", rmw_qos_profile_services_default);
+
+// APIs taking rclcpp::QoS objects are only available in ROS 2 Iron and higher
+#if RCLCPP_VERSION_MAJOR >= 18
```
isamu-takagi, line 73: As with switching to galactic, using `ROS_DISTRO_ROLLING` is more useful for maintenance.

esteve: `ROS_DISTRO_ROLLING` does not point to a specific release, but to the latest commit of the ROS repositories, so if I use `ROS_DISTRO_ROLLING` and once `iron` is released, this code would become invalid. It's more precise to use the version in a package dependency, as it'll be more correct no matter what happens to `rolling`.

taikitanaka3 (approved):

common/bag_time_manager_rviz_plugin/src/bag_time_manager_panel.cpp
isamu-takagi: It's the same with `RCLCPP_VERSION_MAJOR >= 18`. It needs to be `==` to use a specific release.

esteve: It's different, with `>=` we make sure that it'll work with `iron` and higher, if we only check `rolling`, then when `iron` is released, the code will fail because `iron` will no longer be the same as `rolling` and we will have to update the code. With this change, nothing needs to be updated when `iron` is released.

isamu-takagi: There is no guarantee that this interface will work with rclcpp version 19. Both `RCLCPP_VERSION_MAJOR >= 18` and `ROS_DISTRO_ROLLING` can error when the code is updated.

isamu-takagi: Actually this is a small problem, so we should consider it when we support iron.

PR #2786, opened 2023-01-31
fix(obstacle_avoidance_planner): add a margin for vehicle to stop before the end of drivable area boundary

### LINKED ISSUE #2664: Add a margin for vehicle to stop before the end of drivable area boundary
### Checklist

- [X] I've agreed with the maintainers that I can plan this task.

### Description

Currently in mpt a stop point is calculated for the first footprint which goes out of drivable area. I would like to propose to add a longitudinal margin along the traj for vehicle to stop if it goes out of drivable area. So that way vehicle wouldn't stop exactly where it would be out of drivable area and simply a safer approach would be implemented.

Current state:

stop_w_a_margin_for_end_of_da

### Purpose

Add a margin to stop before the point where it goes out drivable area

### Possible approaches

- Define a margin
- Calculate longitudinal distance as much as the margin backwards from the first traj point which goes out of the drivable area
- Insert the stop point there instead

### Definition of done

- [x] Implement the possible approach
- [x] Test it

ahmeddesokyebrahim:
## Description
closes #2664

Fixing the [enhancement issue # 2664](https://github.com/autowarefoundation/autoware.universe/issues/2664) for adding a margin for stop point when the the stop point is calculated for the footprint out of drivable area.

The following is screenshots with the stop point outside the drivable area and with the margin added

Screenshot_20230127_225813

Screenshot_20230127_225835

For testing this PR,  parameter change from [autoware_launch](https://github.com/autowarefoundation/autoware_launch/pull/192) is needed as well.


planning/obstacle_avoidance_planner/src/node.cpp
```diff
@@ -1295,17 +1299,37 @@ void ObstacleAvoidancePlanner::insertZeroVelocityOutsideDrivableArea(
...
+      size_t stop_idx = i;
+      for (size_t j = i - 1; i > 0; --j) {
+        auto stop_position_arc_length = motion_utils::calcSignedArcLength(traj_points, 0, j);
+        if (
+          stop_position_arc_length <
+          (position_outside_drv_area_arc_length - vehicle_stop_margin_)) {
+          stop_idx = j;
+          break;
+        }
+      }
```
mehmetdogru, line 1315: Also of course you could use:

```suggestion
const auto & stop_idx = i - std::floor(vehicle_stop_margin_/ delta_arc_length_for_mpt_points) - 1
```

But given that distance between the traj points are not precisely `delta_arc_length_for_mpt_points` we can go with the idea as above.

planning/obstacle_avoidance_planner/config/obstacle_avoidance_planner.param.yaml
```diff
@@ -43,10 +43,15 @@
       enable_clipping_fixed_traj: false
       non_fixed_trajectory_length: 5.0 # length of the trajectory merging optimized mpt trajectory to original(not optimized) trajectory
 
+      vehicle_stop_margin_outside_drivable_area: 1.5 # vehicle stop margin to let the vehicle stop before the calculated stop point if it is calculated outside the drivable area [m] .
+                                                     # This margin will be realized with delta_arc_length_for_mpt_points m precision.
+
+
```
mehmetdogru, line 49: lets delete this line

planning/obstacle_avoidance_planner/config/obstacle_avoidance_planner.param.yaml
```diff
@@ -43,10 +43,15 @@
...
     object:  # avoiding object
       max_avoiding_objects_velocity_ms: 0.5 # maximum velocity for avoiding objects [m/s]
       max_avoiding_ego_velocity_ms: 6.0 # maximum ego velocity when avoiding objects [m/s]
 
+
```
mehmetdogru, line 54: lets delete this line

planning/obstacle_avoidance_planner/src/utils/debug_utils.cpp
```diff
@@ -850,6 +858,22 @@ visualization_msgs::msg::MarkerArray getDebugVisualizationWallMarker(
...
+      getVirtualWallMarkerArray(
+        virtual_wall_pose_with_margin, "virtual_wall_with_margin", 0.0, 1.0, 0),
+      &vis_marker_array);
+
+    appendMarkerArray(
+      getVirtualWallTextMarkerArray(
+        virtual_wall_pose_with_margin, "virtual_wall_text_with_margin", 0.0, 0.0, 1.0),
+      &vis_marker_array);
+  }
+
```
mehmetdogru, line 876: we dont need extra markers. Instead lets rearrange this part with `stop_pose_by_drivable_area_with_margin` so the virtual wall would be only where we actually stop:

https://github.com/autowarefoundation/autoware.universe/blob/231f70c1d4920fe3c5bd822bf969a7669d284aad/planning/obstacle_avoidance_planner/src/utils/debug_utils.cpp#L844-L846

planning/obstacle_avoidance_planner/src/utils/debug_utils.cpp
```diff
@@ -834,6 +834,14 @@ visualization_msgs::msg::MarkerArray getDebugVisualizationMarker(
...
   }
 
+  if (debug_data.stop_pose_by_drivable_area_with_margin) {
+    appendMarkerArray(
+      getFootprintByDrivableAreaMarkerArray(
+        *debug_data.stop_pose_by_drivable_area_with_margin, vehicle_param,
+        "footprint_by_drivable_area_with_margin", 0.0, 1.0, 0.0),
+      &vis_marker_array);
+  }
+
```
mehmetdogru, line 844: we don't need extra markers

planning/obstacle_avoidance_planner/src/node.cpp
```diff
@@ -1295,17 +1302,34 @@ void ObstacleAvoidancePlanner::insertZeroVelocityOutsideDrivableArea(
...
+    // only insert zero velocity to the first point outside drivable area (minus stop margin)
     if (is_outside) {
-      traj_points[i].longitudinal_velocity_mps = 0.0;
+      size_t stop_idx = i;
+      const auto & op_target_point = motion_utils::calcLongitudinalOffsetPoint(
+        traj_points, traj_point.pose.position, -1.0 * vehicle_stop_margin_outside_drivable_area_);
+
+      if (op_target_point) {
+        const auto target_point = op_target_point.get();
+        stop_idx = motion_utils::findNearestSegmentIndex(traj_points, target_point);
```
mehmetdogru, line 1313: I guess we should check if target point overlaps with nearest traj point or not. Otherwise even the margin is 0, nearest segment idx will nearest_point-1

```suggestion
        // confirm that target point doesn't overlap with the nearest trajectory point
        const size_t nearest_target_idx =
          motion_utils::findNearestIndex(traj_points, op_target_point.get());
        const auto dist = tier4_autoware_utils::calcDistance2d(
          traj_points.at(nearest_target_idx), op_target_point.get());
        const double overlap_threshold = 1e-3;

        if (dist > overlap_threshold) {
          const auto target_point = op_target_point.get();
          stop_idx = motion_utils::findNearestSegmentIndex(traj_points, target_point);
        }
```

planning/obstacle_avoidance_planner/src/node.cpp
```diff
@@ -1295,17 +1302,34 @@ void ObstacleAvoidancePlanner::insertZeroVelocityOutsideDrivableArea(
     const bool is_outside = drivable_area_utils::isOutsideDrivableAreaFromRectangleFootprint(
       traj_point, left_bound, right_bound, vehicle_param_, is_considering_footprint_edges_);
 
-    // only insert zero velocity to the first point outside drivable area
+    // only insert zero velocity to the first point outside drivable area (minus stop margin)
```
mehmetdogru, line 1305: sounds better?:

```suggestion
    // only insert zero velocity to the point with longitudinal offset margin from the first point outside drivable area
```

planning/obstacle_avoidance_planner/config/obstacle_avoidance_planner.param.yaml
```diff
@@ -43,10 +43,15 @@
       enable_clipping_fixed_traj: false
       non_fixed_trajectory_length: 5.0 # length of the trajectory merging optimized mpt trajectory to original(not optimized) trajectory
 
+      vehicle_stop_margin_outside_drivable_area: 1.5 # vehicle stop margin to let the vehicle stop before the calculated stop point if it is calculated outside the drivable area [m] .
+                                                     # This margin will be realized with delta_arc_length_for_mpt_points m precision.
```
brkay54, line 47: Is this comment necessary? Maybe we can remove it?

planning/obstacle_avoidance_planner/src/node.cpp
```diff
@@ -1295,17 +1302,34 @@ void ObstacleAvoidancePlanner::insertZeroVelocityOutsideDrivableArea(
...
+
         debug_data_.stop_pose_by_drivable_area->position.z =
-          planner_data.path.points.at(path_idx).pose.position.z;
+          planner_data.path.points.at(path_idx_1).pose.position.z;
+
+        const size_t path_idx_2 = motion_utils::findNearestIndex(
+          planner_data.path.points, traj_points[stop_idx].pose.position);
+
+        debug_data_.stop_pose_by_drivable_area_with_margin->position.z =
+          planner_data.path.points.at(path_idx_2).pose.position.z;
```
brkay54, line 1332: Could you use more descriptive variable names?

brkay54 (commented):
Thank you for your work! I tested in planning simulator, and I worked as expected.
Please consider the minor changes what I and Mehmet commented. Also please remove the extra markers.

planning/obstacle_avoidance_planner/config/obstacle_avoidance_planner.param.yaml
ahmeddesokyebrahim: This explanation was added based on a previous comment from @mehmetdogru . It is very beneficial to leave it for anyone would like to change it to understand it will realized by the other parameter

planning/obstacle_avoidance_planner/src/node.cpp
ahmeddesokyebrahim: I returned it back to original code, as Mehmet suggested to have only 1 vistual wall where the vehicle stops

takayuki5168 (approved):

PR #2846, opened 2023-02-08
perf(detection_area): point and polygon inclusion check by minimum enclosing circle

veqcc:
Signed-off-by: veqcc <ryuta.kambe@tier4.jp>

## Description
Currently, the module `detection_area` needs a lot of computational cost for searching witch point is included in a detection_area by `bg::within`. This is too heavy function and the size of points (pointCloud) is often over 10,000.
Thus, I introduce using the minimum enclosing circle of a detection_area. If a point is NOT included in the circle, then of course the point don't to be checked by `bg::within`, witch would reduce the latency.

The following graph shows how the detection_area is optimized by this method. The red graph is the original latency, and the blue one is my proposal.
image

## Questions
 - Could you check the algorithm carefully? This is somewhat complicated.
 - I heard this kind of implementaion should not be in behavior_velocity module, but in common module. However, due to my lack of understanding of types of points and polygons, I could not implement it under common module (ex: `lanelet` cannot be used in common. really?). Could you give me some advice?
 - What kind of tests should be added to this module?
Thank you

### DIFF
--- a/planning/behavior_velocity_planner/src/scene_module/detection_area/scene.cpp
+++ b/planning/behavior_velocity_planner/src/scene_module/detection_area/scene.cpp
@@ -211,7 +211,7 @@ std::pair<lanelet::BasicPoint2d, double> calcSmallestEnclosingCircle(
   lanelet::BasicPoint2d center(0.0, 0.0);
   double radius_squared = 0.0;
 
-  auto cross = [](lanelet::BasicPoint2d p1, lanelet::BasicPoint2d p2) -> double {
+  auto cross = [](const lanelet::BasicPoint2d &p1, const lanelet::BasicPoint2d &p2) -> double {
     return p1.x() * p2.y() - p1.y() * p2.x();
   };
 
--- a/planning/behavior_velocity_planner/src/scene_module/detection_area/scene.cpp
+++ b/planning/behavior_velocity_planner/src/scene_module/detection_area/scene.cpp
@@ -230,7 +230,7 @@ std::pair<lanelet::BasicPoint2d, double> calcSmallestEnclosingCircle(
 
   auto make_circle_2 =
     [&](const lanelet::BasicPoint2d & p1, const lanelet::BasicPoint2d & p2) -> void {
-    center = (p1 + p2) / 2.0;
+    center = (p1 + p2) * 0.5;
     radius_squared = (center - p1).squaredNorm() + eps;
   };
 
--- a/planning/behavior_velocity_planner/src/scene_module/detection_area/scene.cpp
+++ b/planning/behavior_velocity_planner/src/scene_module/detection_area/scene.cpp
@@ -204,6 +204,7 @@ bool DetectionAreaModule::modifyPathVelocity(PathWithLaneId * path, StopReason *
 }
 
 // calc smallest enclosing circle with average O(N) algorithm
+// reference: https://erickimphotography.com/blog/wp-content/uploads/2018/09/Computational-Geometry-Algorithms-and-Applications-3rd-Ed.pdf
 std::pair<lanelet::BasicPoint2d, double> calcSmallestEnclosingCircle(
   const lanelet::ConstPolygon2d & poly)
 {
--- a/planning/behavior_velocity_planner/src/scene_module/detection_area/scene.cpp
+++ b/planning/behavior_velocity_planner/src/scene_module/detection_area/scene.cpp
@@ -218,6 +218,7 @@ std::pair<lanelet::BasicPoint2d, double> calcSmallestEnclosingCircle(
   auto make_circle_3 = [&](
                          const lanelet::BasicPoint2d & p1, const lanelet::BasicPoint2d & p2,
                          const lanelet::BasicPoint2d & p3) -> void {
+    // reference for circumcenter vector https://en.wikipedia.org/wiki/Circumscribed_circle
     const double A = (p2 - p3).squaredNorm();
     const double B = (p3 - p1).squaredNorm();
     const double C = (p1 - p2).squaredNorm();
--- a/planning/behavior_velocity_planner/src/scene_module/detection_area/scene.cpp
+++ b/planning/behavior_velocity_planner/src/scene_module/detection_area/scene.cpp
@@ -223,7 +223,7 @@ std::pair<lanelet::BasicPoint2d, double> calcSmallestEnclosingCircle(
     const double B = (p3 - p1).squaredNorm();
     const double C = (p1 - p2).squaredNorm();
     const double S = cross(p2 - p1, p3 - p1);
-    if (S < eps) return;
+    if (std::abs(S) < eps) return;
     center = (A * (B + C - A) * p1 + B * (C + A - B) * p2 + C * (A + B - C) * p3) / (4 * S * S);
     radius_squared = (center - p1).squaredNorm() + eps;
   };
--- a/planning/behavior_velocity_planner/src/scene_module/detection_area/scene.cpp
+++ b/planning/behavior_velocity_planner/src/scene_module/detection_area/scene.cpp
@@ -203,9 +203,9 @@ bool DetectionAreaModule::modifyPathVelocity(PathWithLaneId * path, StopReason *
   return true;
 }
 
-// calc smallest enclosing circle with average O(N) algorithm 
+// calc smallest enclosing circle with average O(N) algorithm
 std::pair<lanelet::BasicPoint2d, double> calcSmallestEnclosingCircle(
-  const lanelet::ConstPolygon2d &poly)
+  const lanelet::ConstPolygon2d & poly)
 {
   const double eps = 1e-5;
   lanelet::BasicPoint2d center(0.0, 0.0);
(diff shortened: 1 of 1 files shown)


planning/behavior_velocity_planner/src/scene_module/detection_area/scene.cpp
```diff
@@ -203,6 +203,67 @@ bool DetectionAreaModule::modifyPathVelocity(PathWithLaneId * path, StopReason *
...
 
+// calc smallest enclosing circle with average O(N) algorithm
+std::pair<lanelet::BasicPoint2d, double> calcSmallestEnclosingCircle(
+  const lanelet::ConstPolygon2d & poly)
+{
+  const double eps = 1e-5;
+  lanelet::BasicPoint2d center(0.0, 0.0);
+  double radius_squared = 0.0;
+
+  auto cross = [](lanelet::BasicPoint2d p1, lanelet::BasicPoint2d p2) -> double {
```
taikitanaka3, line 214: I think it's  better to use const
```suggestion
  auto cross = [](const lanelet::BasicPoint2d &p1, const lanelet::BasicPoint2d &p2) -> double {
```

taikitanaka3:
@veqcc 
 > I heard this kind of implementaion should not be in behavior_velocity module, but in common module. However, due to my lack of understanding of types of points and polygons, I could not implement it under common module (ex: lanelet cannot be used in common. really?). Could you give me some advice?
 
 you can add this kind of common function to util.cpp first, if it can be used in other packages.
 It it can be generalized to use in other planning packages then you had better to move this to common
 https://github.com/autowarefoundation/autoware.universe/blob/main/planning/behavior_velocity_planner/src/utilization/util.cpp
 
 > What kind of tests should be added to this module?
 I think you can add [this kind of functional](https://github.com/autowarefoundation/autoware.universe/blob/ff9d064492f44ef51e048a7f41341eed344e0139/planning/behavior_velocity_planner/test/src/test_arc_lane_util.cpp#L64) unit test to what you changed. right now it's depend on PR auther

planning/behavior_velocity_planner/src/scene_module/detection_area/scene.cpp
```diff
@@ -203,6 +203,67 @@ bool DetectionAreaModule::modifyPathVelocity(PathWithLaneId * path, StopReason *
...
+    return p1.x() * p2.y() - p1.y() * p2.x();
+  };
+
+  auto make_circle_3 = [&](
```
taikitanaka3, line 218: I think you had better to add reference links to show what A B C D means or you had better to add more concrete names to them.

planning/behavior_velocity_planner/src/scene_module/detection_area/scene.cpp
veqcc: To apply this suggestion, is it need to change the `cross` parameters to have memory allocation?
```
const auto p4 = p2 - p1;
const auto p5 = p3 - p1;
const double S = cross(p4, p5);
```

veqcc: Or, I would appreciate it if you come up with another algorithm for calculating the center coordinate of the circle through three point.

taikitanaka3: @veqcc 
I changed this suggestion to const

planning/behavior_velocity_planner/src/scene_module/detection_area/scene.cpp
taikitanaka3: LGTM
for behavior anyway
image

planning/behavior_velocity_planner/src/scene_module/detection_area/scene.cpp
```diff
@@ -203,6 +203,67 @@ bool DetectionAreaModule::modifyPathVelocity(PathWithLaneId * path, StopReason *
...
+  double radius_squared = 0.0;
+
+  auto cross = [](const lanelet::BasicPoint2d & p1, const lanelet::BasicPoint2d & p2) -> double {
+    return p1.x() * p2.y() - p1.y() * p2.x();
+  };
+
+  auto make_circle_3 = [&](
+                         const lanelet::BasicPoint2d & p1, const lanelet::BasicPoint2d & p2,
+                         const lanelet::BasicPoint2d & p3) -> void {
+    const double A = (p2 - p3).squaredNorm();
```
veqcc, line 221:
```suggestion
    // reference for circumcenter vector https://en.wikipedia.org/wiki/Circumscribed_circle
    const double A = (p2 - p3).squaredNorm();
```

planning/behavior_velocity_planner/src/scene_module/detection_area/scene.cpp
veqcc: Thank you for your review.
I just added a reference for calculating the circumcenter vector.

taikitanaka3:
@veqcc  I ran [CI](https://evaluation.tier4.jp/evaluation/reports/5c7336d3-88ca-5cb2-bb50-262026e51571?project_id=x2_dev) with this PR please wait for result

planning/behavior_velocity_planner/src/scene_module/detection_area/scene.cpp
```diff
@@ -203,6 +203,68 @@ bool DetectionAreaModule::modifyPathVelocity(PathWithLaneId * path, StopReason *
...
+    const double C = (p1 - p2).squaredNorm();
+    const double S = cross(p2 - p1, p3 - p1);
+    if (S < eps) return;
+    center = (A * (B + C - A) * p1 + B * (C + A - B) * p2 + C * (A + B - C) * p3) / (4 * S * S);
+    radius_squared = (center - p1).squaredNorm() + eps;
+  };
+
+  auto make_circle_2 =
+    [&](const lanelet::BasicPoint2d & p1, const lanelet::BasicPoint2d & p2) -> void {
+    center = (p1 + p2) / 2.0;
```
taikitanaka3, line 233: @veqcc 
I prefer multiply than division
```suggestion
    center = (p1 + p2) * 0.5;
```

planning/behavior_velocity_planner/src/scene_module/detection_area/scene.cpp
```diff
@@ -203,6 +203,68 @@ bool DetectionAreaModule::modifyPathVelocity(PathWithLaneId * path, StopReason *
   return true;
 }
 
+// calc smallest enclosing circle with average O(N) algorithm
+std::pair<lanelet::BasicPoint2d, double> calcSmallestEnclosingCircle(
```
taikitanaka3, line 207: @maxime-clem 
can you check this algorithm too?

maxime-clem: @veqcc can you add a reference to the algorithm you implemented ?
I only searched a little but it seems the state of art algorithm is [Megiddo's prune-and-search algorithm](https://www.cs.mcgill.ca/~cs507/projects/1998/jacob/solutions.html) with O(n) complexity.

maxime-clem: I found the reference (https://people.inf.ethz.ch/emo/PublFiles/SmallEnclDisk_LNCS555_91.pdf).
This algorithm is called "Welzl's algorithm".
```
Welzl, Emo. "Smallest enclosing disks (balls and ellipsoids); New Results and New Trends in Computer Science, ed. H. Maurer; LNCS 555." (1991).
```

veqcc: Thank you for your review!
Actually, this implementation is referenced from Computer Geometry (https://amzn.asia/d/0HX8Tla), and no original implementation on the Internet. There are many someone's reference implementations.
I have not strictly compared which algorithm is the best in our case. Do you think the comparison is needed?

planning/behavior_velocity_planner/src/scene_module/detection_area/scene.cpp
```diff
@@ -203,6 +203,68 @@ bool DetectionAreaModule::modifyPathVelocity(PathWithLaneId * path, StopReason *
...
+  };
+
+  auto make_circle_2 =
+    [&](const lanelet::BasicPoint2d & p1, const lanelet::BasicPoint2d & p2) -> void {
+    center = (p1 + p2) * 0.5;
+    radius_squared = (center - p1).squaredNorm() + eps;
+  };
+
+  auto in_circle = [&](const lanelet::BasicPoint2d & p) -> bool {
+    return (center - p).squaredNorm() <= radius_squared;
```
taikitanaka3, line 238: @veqcc 
is this sqrt radius or not?
```suggestion
    return (center - p).squaredNorm() <= std::sqrt(radius_squared);
```

planning/behavior_velocity_planner/src/scene_module/detection_area/scene.cpp
```diff
@@ -203,6 +203,68 @@ bool DetectionAreaModule::modifyPathVelocity(PathWithLaneId * path, StopReason *
   return true;
 }
 
+// calc smallest enclosing circle with average O(N) algorithm
+std::pair<lanelet::BasicPoint2d, double> calcSmallestEnclosingCircle(
+  const lanelet::ConstPolygon2d & poly)
+{
+  const double eps = 1e-5;
```
taikitanaka3, line 210: @veqcc 
can you add description of what eps means?

veqcc:
```suggestion
  // The `eps` is used to avoid precision bugs in circle inclusion checkings.
  // If the value of `eps` is too small, this function doesn't work well. More than 1e-10 is recommended.
  const double eps = 1e-5;
```

taikitanaka3 (approved):
LGTM

taikitanaka3:
image

PR #2883, opened 2023-02-14
perf(map_based_prediction): performance improvement of interpolateReferencePath with cumulative sum

veqcc:
Signed-off-by: veqcc <ryuta.kambe@tier4.jp>

### Description
The node `map_based_prediction` is too slow to publish `/perception/object_recognition/objects` topic when there are a lot of vehicles around the ego. This is due to the inefficient implementation of `interpolateReferencePath`, especially the calculation of the length of paths with `calcSignedArcLength`. It calls the `calcSignedArcLength` function for EVERY index of the path. This can be optimized with the idea cumulative sum, in which the path length of 0 to i-th index can be easily calculated with the value of 0 to (i-1)th.
The following figure shows how latencies are optimized with the idea. The red graph is the original latency, and the blue graph is the latency of the proposed implementation.
image


tkimura4 (approved):
LGTM
It's a great job!

PR #2891, opened 2023-02-15
fix(pure_pursuit): return zero curvature when neighboring idx isn't found

kyoichi-sugahara:
## Description

When candidate point for the curvature calculation isn't found, 0 is returned to avoid unnecessary warnings.

But there is a still space to solve this probelem by 
- shifting closest point so that enough distance can be gotten.
- changing distane from the closest point between forward direction and backward direction.

### DIFF
--- a/control/pure_pursuit/src/pure_pursuit/pure_pursuit_lateral_controller.cpp
+++ b/control/pure_pursuit/src/pure_pursuit/pure_pursuit_lateral_controller.cpp
@@ -186,16 +186,19 @@ double PurePursuitLateralController::calcCurvature(const size_t closest_idx)
   if (static_cast<size_t>(closest_idx) >= idx_dist) {
     prev_idx = closest_idx - idx_dist;
   } else {
-    // return zero curvature because prev_idx can't be found
+    // return zero curvature when forward distance is not long enough in the trajectory
     return 0.0;
   }
 
   if (trajectory_resampled_->points.size() - 1 >= closest_idx + idx_dist) {
     next_idx = closest_idx + idx_dist;
   } else {
-    // return zero curvature because next_idx can't be found
+    // return zero curvature when backward distance is not long enough in the trajectory
     return 0.0;
   }
+  // TODO: shift the center point of the curvature calculation to allow sufficient distance,
+  // because if sufficient distance cannot be obtained in front or behind, the curvature will be
+  // zero in the current implementation.
 
   // Calculate curvature assuming the trajectory points interval is constant
   double current_curvature = 0.0;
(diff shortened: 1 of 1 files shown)


brkay54 (approved):
Thank you for your work! LGTM!

control/pure_pursuit/src/pure_pursuit/pure_pursuit_lateral_controller.cpp
```diff
@@ -185,9 +185,16 @@ double PurePursuitLateralController::calcCurvature(const size_t closest_idx)
 
   if (static_cast<size_t>(closest_idx) >= idx_dist) {
     prev_idx = closest_idx - idx_dist;
+  } else {
+    // return zero curvature because prev_idx can't be found
```
taikitanaka3, line 189: @kyoichi-sugahara
Thank you for PR.
could you tell me which case meets condition below? 
> // return zero curvature because prev_idx can't be found

> // return zero curvature because next_idx can't be found

control/pure_pursuit/src/pure_pursuit/pure_pursuit_lateral_controller.cpp
```diff
@@ -185,9 +185,16 @@ double PurePursuitLateralController::calcCurvature(const size_t closest_idx)
...
+    return 0.0;
   }
+
   if (trajectory_resampled_->points.size() - 1 >= closest_idx + idx_dist) {
     next_idx = closest_idx + idx_dist;
+  } else {
+    // return zero curvature because next_idx can't be found
```
taikitanaka3, line 196: @kyoichi-sugahara 
IMO
I think it's better to use `prev_idx` and `next_idx` a certain distance far from closest_idx with threshold.
What do you think?
```
 prev_index,<--->closest index <--->next_idx
```

taikitanaka3:
@kyoichi-sugahara 
TODO: write down TODOs for calculating curvature.

kyoichi-sugahara:
checked warnings are not outputted

https://user-images.githubusercontent.com/32741405/219276568-85de6b75-d05e-42f7-b601-bb3b0cc0043e.mp4

control/pure_pursuit/src/pure_pursuit/pure_pursuit_lateral_controller.cpp
kyoichi-sugahara: @taikitanaka3 
how about this?
>return zero curvature when backward/forward distance is not long enough in the trajectory

control/pure_pursuit/src/pure_pursuit/pure_pursuit_lateral_controller.cpp
kyoichi-sugahara: @taikitanaka3 
how about this?
>return zero curvature when backward/forward distance is not long enough in the trajectory

taikitanaka3 (approved):
LGTM

TakaHoribe (approved):
thank you!

PR #2895, opened 2023-02-15
perf(map_based_prediction): calculate yaw without calling getLaneletAngle

veqcc:
Signed-off-by: veqcc [ryuta.kambe@tier4.jp](mailto:ryuta.kambe@tier4.jp)

### Description
The motivation to improve the latency of map_based_prediction node is the same as https://github.com/autowarefoundation/autoware.universe/pull/2883 .
How I optimized `getPredictedReferencePath` is to remove the `getLanletAngle` function calls. In `getLanletAngle`, we search which segment is the closest to the point in parameter, where the segments are created from two adjacent centerline points in lanelet. These points in parameter, however, are exactly the points of centerline in lanelet (as seen in `convertPathType`). Thus, we don't have to walk through the lanelet centerline points again and to calculate the yaw from the previous point is enough.

The following figure shows how latencies are optimized with the idea. The red graph is the original latency, and the blue graph is the latency after https://github.com/autowarefoundation/autoware.universe/pull/2883 is merged, and the green graph is the latency of this proposal.
image

By the way, the current implementation of skipping the first point is not smart I think. If you have better implementation ideas, please add a comment.

### DIFF
--- a/perception/map_based_prediction/src/map_based_prediction_node.cpp
+++ b/perception/map_based_prediction/src/map_based_prediction_node.cpp
@@ -1173,33 +1173,53 @@ std::vector<PosePath> MapBasedPredictionNode::convertPathType(
       lanelet::ConstLanelets prev_lanelets = routing_graph_ptr_->previous(path.front());
       if (!prev_lanelets.empty()) {
         lanelet::ConstLanelet prev_lanelet = prev_lanelets.front();
+        bool init_flag = true;
+        geometry_msgs::msg::Pose prev_p;
         for (const auto & lanelet_p : prev_lanelet.centerline()) {
           geometry_msgs::msg::Pose current_p;
           current_p.position = lanelet::utils::conversion::toGeomMsgPt(lanelet_p);
-          const double lane_yaw = lanelet::utils::getLaneletAngle(prev_lanelet, current_p.position);
+          if (init_flag) {
+            init_flag = false;
+            prev_p = current_p;
+            continue;
+          }
+
+          const double lane_yaw = std::atan2(
+            current_p.position.y - prev_p.position.y, current_p.position.x - prev_p.position.x);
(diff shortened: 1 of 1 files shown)


perception/map_based_prediction/src/map_based_prediction_node.cpp
```diff
@@ -1173,33 +1173,53 @@ std::vector<PosePath> MapBasedPredictionNode::convertPathType(
...
           current_p.position = lanelet::utils::conversion::toGeomMsgPt(lanelet_p);
-          const double lane_yaw = lanelet::utils::getLaneletAngle(prev_lanelet, current_p.position);
+          if (init_flag) {
+            init_flag = false;
+            prev_p = current_p;
+            continue;
+          }
+
+          const double lane_yaw = std::atan2(
+            current_p.position.y - prev_p.position.y, current_p.position.x - prev_p.position.x);
```
tkimura4, line 1188: Could you check if this change doesn't affect other modules such as cruise_planner.
I think that calculated yaw may be slightly inaccurate where lane-points are dense, such as lanes with high curvature.

tkimura4, line 1188: @purewater0901 @takayuki5168 
Could you check if this change doesn't affect other modules such as cruise_planner.
I think that calculated yaw may be slightly inaccurate where lane-points are dense, such as lanes with high curvature.

veqcc: Thank you for your review.
For your information, @purewater0901 told me that this yaw calculations do not require too much precision. This PR may result in slightly different `converted_path`. For example, the head point in `lanelet` used to be added to `converted_path`, but I skip it.

tkimura4 (approved):
LGTM

PR #2930, opened 2023-02-22
feat: add vehicle status api

tkhmy:
## Description
- add vehicle status api
-  ~~add publish mgrs_grip topic in the map_loader~~ -> move to #3200
- This PR contains 4 api
    - /api/vehicle/kinematics
    - /api/vehicle/status
    ~~ - /api/vehicle/door ~~

## Related links
- https://github.com/autowarefoundation/autoware/issues/3232
- https://github.com/autowarefoundation/autoware_adapi_msgs/pull/24
- https://github.com/autowarefoundation/autoware-documentation/pull/312

## Tests performed
- makesure all those api output the correct data

1. /api/vehicle/kinematic
    - geographic_pose
        - use a map that contain correct MGRS or UTM
        - check whether the latitude and longitude output correctly
        - the altitude should be the same as the pose z value
     - make sure the following have the correct output with the related topic
         - pose = `/localization/kinematic_state` ->pose
         - twist =  `/localization/kinematic_state` ->twist
         - accel = `/localization/acceleration` -> accel

2. /api/vehicle/state
     - make sure the following have the correct output with the related topic
         - steering_tire_angle = `/vehicle/status/steering_status` ->steering_tire_angle
         - gear =  `/vehicle/status/gear_status` ->report
         - turn_indicator = /vehicle/status/turn_indicators_status` -> report
         - hazard_light = /vehicle/status/hazard_lights_status` -> report
         - energy_level = `/vehicle/status/battery_charge` -> energy_level
             - can be test by using following command
             ```bash
             ros2 topic pub /vehicle/status/battery_charge tier4_vehicle_msgs/msg/BatteryStatus energy_level:\ 20
             ```


tkhmy:
@isamu-takagi @KYabuuchi 
Do you think is better to separate the map_loader publish mgrs_grid function into another PR?

KYabuuchi:
@tkhmy Since those functions will be reviewed by different people, it would be better to separate them if you can.

tkhmy:
> @tkhmy Since those functions will be reviewed by different people, it would be better to separate them if you can.

@KYabuuchi 
Thanks for your commend, I moved the change to this PR, please review
https://github.com/autowarefoundation/autoware.universe/pull/3200

common/autoware_ad_api_specs/include/autoware_ad_api_specs/vehicle.hpp
```diff
@@ -0,0 +1,56 @@
+// Copyright 2022 TIER IV, Inc.
```
kminoda, line 1: Better to make this the latest?
```suggestion
// Copyright 2023 TIER IV, Inc.
```

tkhmy: updated [7c12946](https://github.com/autowarefoundation/autoware.universe/pull/2930/commits/7c1294671f974715967f10b73426e219bbcafdf0)

tkhmy:
@isamu-takagi I fixed the naming and it should be working now.
Can you have a look?

I think the autoware document currently do not have a link to the vehicle api page. can I create a PR for that?
About the door layout, will it be handle in vehicle interface and the api need to read the layout and door status from the vehicle interface?

isamu-takagi:
> I think the autoware document currently do not have a link to the vehicle api page. can I create a PR for that?

@tkhmy Thank you for checking document. I updated pages in this PR https://github.com/autowarefoundation/autoware-documentation/pull/406. Now the vehicle status page is available from side menu.

>About the door layout, will it be handle in vehicle interface and the api need to read the layout and door status from the vehicle interface?

Yes. The vehicle driver needs to support a interface for multiple doors. So for now, how about returning an empty array for status and layout?

tkhmy:
> Thank you for checking document. I updated pages in this PR https://github.com/autowarefoundation/autoware-documentation/pull/406. Now the vehicle status page is available from side menu.

@isamu-takagi Thank you, I just saw the update, I think it is good

> Yes. The vehicle driver needs to support a interface for multiple doors. So for now, how about returning an empty array for status and layout?

Actually I notice that vehicle interface do not have the message type for publishing multiple message, the current available is this [tier4_api_msgs/msg/DoorStatus](https://github.com/tier4/tier4_autoware_msgs/blob/tier4/universe/tier4_api_msgs/msg/DoorStatus.msg).

I think we can ask define vehicle_interface to use the [DoorStatusArray](https://github.com/autowarefoundation/autoware_adapi_msgs/blob/main/autoware_adapi_v1_msgs/vehicle/msg/DoorStatusArray.msg) directly, then the api will just relay the topic.

How do you think?

isamu-takagi:
>I think we can ask define vehicle_interface to use the [DoorStatusArray](https://github.com/autowarefoundation/autoware_adapi_msgs/blob/main/autoware_adapi_v1_msgs/vehicle/msg/DoorStatusArray.msg) directly, then the api will just relay the topic.

@tkhmy It sounds good. So we need to define interface for vehicle driver. It may be better to separate the PR for the door.

tkhmy:
> It sounds good. So we need to define interface for vehicle driver. It may be better to separate the PR for the door.

Okay I will separate the door from this PR

common/component_interface_specs/include/component_interface_specs/vehicle.hpp
```diff
@@ -0,0 +1,106 @@
...
+};
+
+struct Acceleration
+{
+  using Message = geometry_msgs::msg::AccelWithCovarianceStamped;
+  static constexpr char name[] = "/localization/acceleration";
+  static constexpr size_t depth = 1;
+  static constexpr auto reliability = RMW_QOS_POLICY_RELIABILITY_RELIABLE;
+  static constexpr auto durability = RMW_QOS_POLICY_DURABILITY_VOLATILE;
+};
```
isamu-takagi, line 48: Please move these to [localization.hpp](https://github.com/autowarefoundation/autoware.universe/blob/main/common/component_interface_specs/include/component_interface_specs/localization.hpp) and kinematic_state is already defined there.

common/component_interface_specs/include/component_interface_specs/vehicle.hpp
```diff
@@ -0,0 +1,106 @@
...
+
+struct MapProjectorInfo
+{
+  using Message = tier4_map_msgs::msg::MapProjectorInfo;
+  static constexpr char name[] = "/map/map_projector_type";
+  static constexpr size_t depth = 1;
+  static constexpr auto reliability = RMW_QOS_POLICY_RELIABILITY_RELIABLE;
+  static constexpr auto durability = RMW_QOS_POLICY_DURABILITY_TRANSIENT_LOCAL;
+};
```
isamu-takagi, line 93: Please move this to map.hpp (new file).

system/default_ad_api/src/vehicle.cpp
```diff
@@ -0,0 +1,140 @@
...
+{
+  if (hash_map.find(input) == hash_map.end()) {
+    return default_value;
+  } else {
+    return hash_map[input];
+  }
+}
+
+void VehicleNode::kinematic_state(
+  const vehicle_interface::KinematicState::Message::ConstSharedPtr msg_ptr)
```
isamu-takagi, line 50: The message callback is enough just to save it. Conversions are done in timer callbacks to skip processing data that will not be published.

system/default_ad_api/src/vehicle.hpp
```diff
@@ -0,0 +1,113 @@
...
+private:
+  using GearReport = vehicle_interface::GearStatus::Message;
+  using ApiGear = autoware_adapi_v1_msgs::msg::Gear;
+  using TurnIndicatorsReport = vehicle_interface::TurnIndicatorStatus::Message;
+  using ApiTurnIndicator = autoware_adapi_v1_msgs::msg::TurnIndicators;
+  using HazardLightsReport = vehicle_interface::HazardLightStatus::Message;
+  using ApiHazardLight = autoware_adapi_v1_msgs::msg::HazardLights;
+  using MapProjectorInfo = vehicle_interface::MapProjectorInfo::Message;
+
+  std::unordered_map<uint8_t, uint8_t> gear_type_ = {
```
isamu-takagi, line 53: These mapping are not needed in header file and can be moved to cpp file.

system/default_ad_api/src/vehicle.hpp
```diff
@@ -0,0 +1,113 @@
...
+  void steering_status(const vehicle_interface::SteeringStatus::Message::ConstSharedPtr msg_ptr);
+  void gear_status(const GearReport::ConstSharedPtr msg_ptr);
+  void turn_indicator_status(const TurnIndicatorsReport::ConstSharedPtr msg_ptr);
+  void map_projector_info(const MapProjectorInfo::ConstSharedPtr msg_ptr);
+  void hazard_light_status(const HazardLightsReport::ConstSharedPtr msg_ptr);
+  void energy_status(const vehicle_interface::EnergyStatus::Message::ConstSharedPtr msg_ptr);
+  uint8_t mapping(
+    std::unordered_map<uint8_t, uint8_t> hash_map, uint8_t input, uint8_t default_value);
+  void on_timer();
+  Eigen::Vector3d toBasicPoint3dPt(const geometry_msgs::msg::Point src);
```
isamu-takagi, line 108: These functions are not needed in header file and can be moved to cpp file.

system/default_ad_api/src/vehicle.hpp
```diff
@@ -0,0 +1,113 @@
...
+  Sub<vehicle_interface::SteeringStatus> sub_steering_;
+  Sub<vehicle_interface::GearStatus> sub_gear_state_;
+  Sub<vehicle_interface::TurnIndicatorStatus> sub_turn_indicator_;
+  Sub<vehicle_interface::HazardLightStatus> sub_hazard_light_;
+  Sub<vehicle_interface::MapProjectorInfo> sub_map_projector_info_;
+  Sub<vehicle_interface::EnergyStatus> sub_energy_level_;
+  rclcpp::TimerBase::SharedPtr timer_;
+
+  MapProjectorInfo::ConstSharedPtr map_projector_info_;
+  autoware_ad_api::vehicle::VehicleKinematics::Message vehicle_kinematics_;
```
isamu-takagi, line 94: For kinematics, it seems better to save kinematic_state and acceleration and convert them in a timer callback.

common/component_interface_specs/include/component_interface_specs/vehicle.hpp
tkhmy: fixed in [c5d7a70](https://github.com/autowarefoundation/autoware.universe/pull/2930/commits/c5d7a70861bacad66d454f63a8dfbed210af96db)

common/component_interface_specs/include/component_interface_specs/vehicle.hpp
tkhmy: fixed in [c5d7a70](https://github.com/autowarefoundation/autoware.universe/pull/2930/commits/c5d7a70861bacad66d454f63a8dfbed210af96db)

system/default_ad_api/src/vehicle.cpp
tkhmy: Do you mean the conversion for the GPS position?
If so, I will need to declare a pointer for pose

system/default_ad_api/src/vehicle.hpp
tkhmy: so I will just save all the msg as pointer and then only process in the callback?

tkhmy:
@isamu-takagi I changed the stuff that you request and also fixed the conflict with main, can you review again?

system/default_ad_api/src/vehicle.cpp
```diff
@@ -0,0 +1,178 @@
...
+    lanelet::Origin origin{position};
+    lanelet::projection::UtmProjector projector{origin};
+    lanelet::GPSPoint projected_gps_point =
+      projector.reverse(toBasicPoint3dPt(kinematic_state_msgs_->pose.pose.position));
+    vehicle_kinematics.geographic_pose.header = kinematic_state_msgs_->header;
+    vehicle_kinematics.geographic_pose.header.frame_id = "global";
+    vehicle_kinematics.geographic_pose.position.latitude = projected_gps_point.lat;
+    vehicle_kinematics.geographic_pose.position.longitude = projected_gps_point.lon;
+    vehicle_kinematics.geographic_pose.position.altitude = projected_gps_point.ele;
+  }
```
isamu-takagi, line 168: If the projection type is not supported, set header and set pose to NaN.

system/default_ad_api/src/vehicle.cpp
```diff
@@ -0,0 +1,178 @@
...
+  const vehicle_interface::EnergyStatus::Message::ConstSharedPtr msg_ptr)
+{
+  vehicle_status_.energy_percentage = msg_ptr->energy_level;
+}
+
+void VehicleNode::on_timer()
+{
+  vehicle_status_.stamp = now();
+  autoware_ad_api::vehicle::VehicleKinematics::Message vehicle_kinematics;
+  vehicle_kinematics.pose.header = kinematic_state_msgs_->header;
```
isamu-takagi, line 143: It needs null check for `kinematic_state_msgs_` and `acceleration_msgs_`.

system/default_ad_api/src/vehicle.cpp
```diff
@@ -0,0 +1,178 @@
...
+    vehicle_kinematics.geographic_pose.header = kinematic_state_msgs_->header;
+    vehicle_kinematics.geographic_pose.header.frame_id = "global";
+    vehicle_kinematics.geographic_pose.position.latitude = projected_gps_point.lat;
+    vehicle_kinematics.geographic_pose.position.longitude = projected_gps_point.lon;
+    vehicle_kinematics.geographic_pose.position.altitude = projected_gps_point.ele;
+  }
+  vehicle_kinematics.accel.header = acceleration_msgs_->header;
+  vehicle_kinematics.accel.accel = acceleration_msgs_->accel;
+  pub_kinematics_->publish(vehicle_kinematics);
+  pub_status_->publish(vehicle_status_);
```
isamu-takagi, line 172: Vehicle status may be published with some fields not assigned.

system/default_ad_api/src/vehicle.cpp
tkhmy: fixed in [813b048](https://github.com/autowarefoundation/autoware.universe/pull/2930/commits/813b048edd79b610e773195a2ff7d3720f19f6a4)

system/default_ad_api/src/vehicle.cpp
tkhmy: @isamu-takagi I think the default will be 0, can you advice me how to do the null check?

system/default_ad_api/src/vehicle.cpp
tkhmy: do you mean the vehicle status publish first then only do the conversion for kinematics?

system/default_ad_api/src/vehicle.cpp
isamu-takagi: Something like:
```cpp
void VehicleNode::on_timer()
{
  publish_kinematics();
  publish_status();
}

void VehicleNode::publish_kinematics()
{
  if (!kinematic_state_msgs_ || !acceleration_msgs_) return;

  autoware_ad_api::vehicle::VehicleKinematics::Message vehicle_kinematics;
  vehicle_kinematics.pose.header = kinematic_state_msgs_->header;
  ...
}

void VehicleNode::publish_status()
{
  vehicle_status_.stamp = now();
  ...
}
```

It looks like the process died at `kinematic_state_msgs_->header`.

system/default_ad_api/src/vehicle.cpp
isamu-takagi: No, it means needs to check if steering, gear, turn_indicator, hazard_lights, energy_status have been received.
For example, having `SteeringStatus::ConstSharedPtr, GearReport::ConstSharedPtr, etc.` as a member variable instead of `VehicleStatus` and publish status if all pointers is valid.

system/default_ad_api/src/vehicle.cpp
tkhmy: fixed in [39bd585](https://github.com/autowarefoundation/autoware.universe/pull/2930/commits/39bd5852c76c25fa787c49096bb2f4dac98f2bfb)

system/default_ad_api/src/vehicle.cpp
isamu-takagi: > do we need to force them to publish all the data?

I think so. The API doesn't know if the module doesn't support it or module is not ready.

tkhmy: fixed at [9f12741](https://github.com/autowarefoundation/autoware.universe/pull/2930/commits/9f127414bb4b656f16f3c7d3fecf0bcd9c76a13a)

isamu-takagi (approved):
LGTM

Issue #2495, opened 2022-12-13
`planner_data` access outside mutex lock inside `behavior_path_planner_node`


VRichardJP:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I'm convinced that this is not my fault but a bug.

### Description

Found the exact same issue than #2422 here:
https://github.com/autowarefoundation/autoware.universe/blob/f1a9a9608559a5b89f631df3dc2fadd037e36ab4/planning/behavior_path_planner/src/behavior_path_planner_node.cpp#L657

The data is accessed outside the `mutex_pd_` lock guard

While working on my PR, I reset the `route_handler` internal data after a new route has just arrived. The data is only left uninitialized while the new route is being checked/loaded (which happens in the same callback). However I got the following segfault, showing that another thread tried to access the data at the same time:
```
Thread 14 "component_conta" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffe27d4640 (LWP 79951)]
0x00007fffe0175c8b in lanelet::utils::query::getClosestLanelet(std::vector<lanelet::ConstLanelet, std::allocator<lanelet::ConstLanelet> > const&, geometry_msgs::msg::Pose_<std::allocator<void> > const&, lanelet::ConstLanelet*) () from /home/sig/autoware/install/lanelet2_extension/lib/liblanelet2_extension_lib.so
(gdb) up
#1  0x00007fffd115efe6 in route_handler::LaneletRoute::getClosestLaneletPointWithinRoute (this=0x0, pose=...) at /home/sig/autoware/src/universe/autoware.universe/planning/route_handler/src/lanelet_route.cpp:394
394	  if (!lanelet::utils::query::getClosestLanelet(route_lanelets_, pose, &closest_lanelet)) {
(gdb) p route_lanelets_
Cannot access memory at address 0x20
(gdb) p this
$1 = (const route_handler::LaneletRoute * const) 0x0
(gdb) up
#2  0x00007fffd29e05f0 in behavior_path_planner::util::getDrivableAreaForAllSharedLinestringLanelets (planner_data=std::shared_ptr<const behavior_path_planner::PlannerData> (use count 10, weak count 0) = {...}) at /home/sig/autoware/src/universe/autoware.universe/planning/behavior_path_planner/src/utilities.cpp:1647
[...]

PR #2933, opened 2023-02-22
fix(behavior_path_planner): concurrency issue

VRichardJP:
solves #2495

## Description

Properly lock mutex for `planner_data_` and `bt_manager_` read/write.

### DIFF
--- a/planning/behavior_path_planner/include/behavior_path_planner/behavior_path_planner_node.hpp
+++ b/planning/behavior_path_planner/include/behavior_path_planner/behavior_path_planner_node.hpp
@@ -149,14 +149,13 @@ private:
 
   TurnSignalDecider turn_signal_decider_;
 
-  std::mutex mutex_pd_;  // mutex for planner_data_
-  std::mutex mutex_bt_;  // mutex for bt_manager_
+  std::mutex mutex_pd_;       // mutex for planner_data_
+  std::mutex mutex_manager_;  // mutex for bt_manager_ or planner_manager_
+  std::mutex mutex_map_;      // mutex for has_received_map_ and map_ptr_
+  std::mutex mutex_route_;    // mutex for has_received_route_ and route_ptr_
 
   // setup
   bool isDataReady();
-
-  // update planner data
-  std::shared_ptr<PlannerData> createLatestPlannerData();
 
   // parameters
   std::shared_ptr<AvoidanceParameters> avoidance_param_ptr_;
@@ -195,7 +194,8 @@ private:
    * @brief Modify the path points near the goal to smoothly connect the lanelet and the goal point.
    */
   PathWithLaneId modifyPathForSmoothGoalConnection(
-    const PathWithLaneId & path) const;  // (TODO) move to util
+    const PathWithLaneId & path,
+    const std::shared_ptr<PlannerData> & planner_data) const;  // (TODO) move to util
   OnSetParametersCallbackHandle::SharedPtr m_set_param_res;
 
   /**
@@ -206,8 +206,15 @@ private:
   /**
    * @brief extract path from behavior tree output
    */
+#ifdef USE_OLD_ARCHITECTURE
   PathWithLaneId::SharedPtr getPath(
-    const BehaviorModuleOutput & bt_out, const std::shared_ptr<PlannerData> planner_data);
+    const BehaviorModuleOutput & bt_out, const std::shared_ptr<PlannerData> & planner_data,
+    const std::shared_ptr<BehaviorTreeManager> & bt_manager);
+#else
+  PathWithLaneId::SharedPtr getPath(
+    const BehaviorModuleOutput & bt_out, const std::shared_ptr<PlannerData> & planner_data,
+    const std::shared_ptr<PlannerManager> & planner_manager);
+#endif
 
   /**
    * @brief skip smooth goal connection
@@ -242,27 +249,34 @@ private:
   /**
    * @brief publish debug messages
    */
-  void publishSceneModuleDebugMsg();
+#ifdef USE_OLD_ARCHITECTURE
+  void publishSceneModuleDebugMsg(
+    const std::shared_ptr<SceneModuleVisitor> & debug_messages_data_ptr);
+#endif
 
   /**
    * @brief publish path candidate
    */
 #ifdef USE_OLD_ARCHITECTURE
   void publishPathCandidate(
-    const std::vector<std::shared_ptr<SceneModuleInterface>> & scene_modules);
+    const std::vector<std::shared_ptr<SceneModuleInterface>> & scene_modules,
+    const std::shared_ptr<PlannerData> & planner_data);
 #else
   void publishPathCandidate(
-    const std::vector<std::shared_ptr<SceneModuleManagerInterface>> & managers);
+    const std::vector<std::shared_ptr<SceneModuleManagerInterface>> & managers,
+    const std::shared_ptr<PlannerData> & planner_data);
 
   void publishPathReference(
-    const std::vector<std::shared_ptr<SceneModuleManagerInterface>> & managers);
+    const std::vector<std::shared_ptr<SceneModuleManagerInterface>> & managers,
+    const std::shared_ptr<PlannerData> & planner_data);
 #endif
 
   /**
    * @brief convert path with lane id to path for publish path candidate
    */
   Path convertToPath(
-    const std::shared_ptr<PathWithLaneId> & path_candidate_ptr, const bool is_ready);
+    const std::shared_ptr<PathWithLaneId> & path_candidate_ptr, const bool is_ready,
+    const std::shared_ptr<PlannerData> & planner_data);
 };
 }  // namespace behavior_path_planner
 
--- a/planning/behavior_path_planner/src/behavior_path_planner_node.cpp
+++ b/planning/behavior_path_planner/src/behavior_path_planner_node.cpp
@@ -1498,8 +1498,7 @@ SetParametersResult BehaviorPathPlannerNode::onSetParam(
     update_param(
       parameters, "avoidance.publish_debug_marker", avoidance_param_ptr_->publish_debug_marker);
     update_param(
-      parameters, "lane_change.publish_debug_marker",
-      lane_change_param_ptr_->publish_debug_marker);
+      parameters, "lane_change.publish_debug_marker", lane_change_param_ptr_->publish_debug_marker);
     // Drivable area expansion parameters
     using drivable_area_expansion::DrivableAreaExpansionParameters;
     const std::lock_guard<std::mutex> lock(mutex_pd_);  // for planner_data_
--- a/planning/behavior_path_planner/src/behavior_path_planner_node.cpp
+++ b/planning/behavior_path_planner/src/behavior_path_planner_node.cpp
@@ -136,7 +136,7 @@ BehaviorPathPlannerNode::BehaviorPathPlannerNode(const rclcpp::NodeOptions & nod
     RCLCPP_INFO(get_logger(), "use behavior tree.");
 
     const std::string path_candidate_name_space = "/planning/path_candidate/";
-    mutex_bt_.lock();
+    const std::lock_guard<std::mutex> lock(mutex_manager_);  // for bt_manager_
 
     bt_manager_ = std::make_shared<BehaviorTreeManager>(*this, getBehaviorTreeManagerParam());
 
@@ -188,8 +188,6 @@ BehaviorPathPlannerNode::BehaviorPathPlannerNode(const rclcpp::NodeOptions & nod
     bt_manager_->registerSceneModule(pull_out_module);
 
     bt_manager_->createBehaviorTree();
-
-    mutex_bt_.unlock();
   }
 #else
   {
@@ -198,7 +196,7 @@ BehaviorPathPlannerNode::BehaviorPathPlannerNode(const rclcpp::NodeOptions & nod
     const std::string path_candidate_name_space = "/planning/path_candidate/";
     const std::string path_reference_name_space = "/planning/path_reference/";
 
-    mutex_bt_.lock();
+    const std::lock_guard<std::mutex> lock(mutex_manager_);  // for planner_manager_
 
     const auto & p = planner_data_->parameters;
     planner_manager_ =
@@ -284,8 +282,6 @@ BehaviorPathPlannerNode::BehaviorPathPlannerNode(const rclcpp::NodeOptions & nod
       path_reference_publishers_.emplace(
         "avoidance", create_publisher<Path>(path_reference_name_space + "avoidance", 1));
     }
-
-    mutex_bt_.unlock();
   }
 #endif
 
@@ -987,13 +983,21 @@ bool BehaviorPathPlannerNode::isDataReady()
     return missing("scenario_topic");
   }
 
-  if (!route_ptr_) {
-    return missing("route");
+  {
+    std::lock_guard<std::mutex> lk_route(mutex_route_);
+    if (!route_ptr_) {
+      return missing("route");
+    }
   }
 
-  if (!map_ptr_) {
-    return missing("map");
+  {
+    std::lock_guard<std::mutex> lk_map(mutex_map_);
+    if (!map_ptr_) {
+      return missing("map");
+    }
   }
+
+  const std::lock_guard<std::mutex> lock(mutex_pd_);  // for planner_data_
 
   if (!planner_data_->dynamic_object) {
     return missing("dynamic_object");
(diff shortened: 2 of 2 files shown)


satoshi-ota:
@xmfcx @purewater0901 Could you review this PR? Thanks in advance :pray:

mitsudome-r:
@satoshi-ota @purewater0901 
This PR is probably better to be reviewed by the code-owner who knows the code very well so @purewater0901 would be a better reviewer than @xmfcx

VRichardJP:
I don't know much about the new planner manager, but if it is not thread-safe like the old behavior tree manager, then the concurrency issues raised in #2495 still hold on the new API.

I have updated the code accordingly.

xmfcx:
@VRichardJP could you rebase this PR?

Then, @purewater0901 could you review this PR?

purewater0901:
@xmfcx Sorry for being late in response. I'll take a look at it later.

purewater0901:
@xmfcx Could you tell me what is the problem in the current code?

VRichardJP:
@purewater0901 This PR fixes the concurrency issue described here https://github.com/autowarefoundation/autoware.universe/issues/2495#issuecomment-1439549862

purewater0901:
@VRichardJP Thank you for your comment. My concern is that if you lock the `planner_data_` in the `run()` function, we cannot update the route and map until it finishes the process. That's why I separate `planner_data_` from `planner_data`

VRichardJP:
@purewater0901 And this is exactly how things should be, no?
If the map is changed while the planner modules are working on it, the code behavior is impredictible. For example, consider a routine going through all the lanelets of the route:
```c++
for (const &llt: route_lanelets)
{
  doSomething(llt);
}
```
Imagine the route is updated during this routine. This could happen:
```
llt = route_lanelets[0];
doSomething(llt);
llt = route_lanelets[1];
doSomething(llt);
llt = route_lanelets[2];
doSomething(llt);
// route_lanelets updated!!! 
// New route has only 2 lanelets
llt = route_lanelet[3];
doSomething(llt)
-> Segmentation Fault
```

purewater0901:
@VRichardJP 
https://github.com/autowarefoundation/autoware.universe/blob/main/planning/behavior_path_planner/src/behavior_path_planner_node.cpp#L978
https://github.com/autowarefoundation/autoware.universe/blob/main/planning/behavior_path_planner/src/behavior_path_planner_node.cpp#L1025
I think your idea is incorrect. Since the route update is only performed right before each module uses the route, the above issues you mentioned would not happen in the current code. Note that when `behavior_path_planner` gets new route, it does not update the route in the `planner_data_`. You can find the code [here](https://github.com/autowarefoundation/autoware.universe/blob/main/planning/behavior_path_planner/src/behavior_path_planner_node.cpp#L1378-L1383)

VRichardJP:
Yes, the current implementation does not crash at the moment because no one else is modifying the map/route. But if any piece of code, somewhere, was to lock the mutex and modify the map/route, then the code would have totally unpredictable behavior (this is what I show [here](https://github.com/autowarefoundation/autoware.universe/issues/2495#issuecomment-1439549862)). And this is a problem.

I understand your concern regarding the lock in `onMap` and `onRoute`. I could remove the `mutex_pd_` lock by moving the `planner_data_->route_handler->setMap(*map_ptr_)` and `planner_data_->route_handler->setRoute(*route_ptr_)` in the `run()` (at the same place than the old `createLatestPlannerData()`). However, it does not mean we can get away with no locking mechanism. Since `map_ptr_` is the only pointer holding the data, as soon as you do `map_ptr_ = msg;` the destructor of the object pointed by `map_ptr_` is called (because the shared_ptr counter becomes 0), so you cannot have someone else doing `planner_data_->route_handler->setMap(*map_ptr_)` at the same time. In other words, if you want to remove the `mutex_pd_` lock in `onMap()` and `onRoute()`, you need to add a `mutex_map_` and a `mutex_route_`. These mutexes would only be locked shortly for the `planner_data_->route_handler->setMap(*map_ptr_)` and `planner_data_->route_handler->setRoute(*route_ptr_)`.

What do you think?

VRichardJP:
Just to make sure I am clear, here is a toy example that shows why updating the pointer without mutex is not safe:
```C++
#include <chrono>
#include <iostream>
#include <memory>
#include <thread>
#include <vector>

using namespace std::chrono_literals;

class Foo {
public:
  Foo(size_t n, int x)
    : data(n, x) 
  {
  }

  ~Foo() {
    std::cout << "deleting data..." << std::endl;
  }

  std::vector<int> data;
};

static std::shared_ptr<Foo> foo_ptr;

void updateFoo() {
  static int x = 0;
  std::cout << "updating ptr!" << std::endl;
  foo_ptr = std::make_shared<Foo>(100, x++);
}

void accessFoo(const Foo &foo) {
  for (int i=0; i < foo.data.size(); ++i) {
    std::cout << "foo[" << i << "] = " << foo.data[i] << std::endl; 
    std::this_thread::sleep_for(1s);
  }
}

int main(int argc, char** argv) {

  // init
  updateFoo();

  // like onMap() callback
  std::thread update_routine([] {
    for (;;) {
      std::this_thread::sleep_for(5s);
      updateFoo();
    }
  });

  // like run() timer callback
  for (;;) {
    // concurrent unprotected access
    accessFoo(*foo_ptr);
  }

  update_routine.join();

  return 0;
}
```

The code is fairly simple:
- there is a shared ptr to a `Foo` object (like map/route).
- a routine is updating the ptr with a object every 5s (like `onMap` and `onRoute`)
- the main thread read the pointed data (like `run()` callback), and slowly print it to stdout
- to make the concurrency obvious, I added a few sleeps but I would also happen without

If you compile and run you see this:
```bash
$ g++ main.cc -o main && ./main                                                                                
updating ptr!
foo[0] = 0
foo[1] = 0
foo[2] = 0
foo[3] = 0
foo[4] = 0
updating ptr!
deleting data...
foo[5] = 0
foo[6] = 0
foo[7] = 0
foo[8] = 0
foo[9] = 0
updating ptr!
deleting data...
foo[10] = 2
foo[11] = 2
foo[12] = 2
foo[13] = 2
foo[14] = 2
updating ptr!
deleting data...
foo[15] = 2
foo[16] = 2
foo[17] = 2
foo[18] = 2
foo[19] = 2
updating ptr!
deleting data...
foo[20] = 4
foo[21] = 4
foo[22] = 4
``` 

What happens? While the main thread is only accessing the data of the initial foo object (`x = 0`), which is a 0-vector of size 100, it does not always print 0. This because the data the thread is reading is being overwritten by the other thread. Why? Because when the `foo_ptr` is updated, there is no other `shared_ptr` pointing to the old data, so the Foo object is destroyed. Meaning that the memory the old pointer was pointing is free to be used by something else.

On the other hand, if the `foo_ptr` is duplicated before it is reasigned, then there is no issue. Do we need a lock for that? From the pure C++ standard point of view you have no guarantee the pointer swaping is done atomicall. On x86 it might be true? Still, the compiler is free to optimize out the duplicated pointer if no explicit lock mechanism is used. So there are actually 2 solutions of the map/route pointers: either use a mutex, or we could wrap the pointers with `std::atomic<std::shared_ptr<T>>`.

VRichardJP:
@purewater0901 Sorry I messed a lot with the rebase, I think the code is ok now.

purewater0901:
@VRichardJP 
First of all, thank you for your detailed explanations and examples. They helped me a lot to understand your point. 

>EDIT: actually my last sentence is overkill, it would just be necessary to lock the map/route mutex to copy the shared_ptr, so that the pointed data destructor would not be called. Once we have an handle to the old data, the map_ptr_ or route_ptr_ could point to something else without problem. What do you think?

I agree with your idea that creating a new pointer to copy the old data (a handler to the old data). It seems that since the route handler has a pointer to the map and route in its data, the data in the 'planner_data' is modified through `planner_data_'. 

Anyway, to protect the data from being overwritten by other threads, your current implementation seems reasonable to me.

purewater0901:
@VRichardJP Just to make sure that in order to prevent illegal access from other threads, you added a new pointer like 

```C++
#include <chrono>
#include <iostream>
#include <memory>
#include <thread>
#include <vector>

using namespace std::chrono_literals;

class Foo {
public:
    Foo(size_t n, int x)
            : data(n, x)
    {
    }

    ~Foo() {
        std::cout << "deleting data..." << std::endl;
    }

    std::vector<int> data;
};

static std::shared_ptr<Foo> foo_ptr;

void updateFoo() {
    static int x = 0;
    std::cout << "updating ptr!" << std::endl;
    foo_ptr = std::make_shared<Foo>(100, x++);
}

void accessFoo(const Foo &foo) {
    for (int i=0; i < foo.data.size(); ++i) {
        std::cout << "foo[" << i << "] = " << foo.data[i] << std::endl;
        std::this_thread::sleep_for(1s);
    }
}

int main(int argc, char** argv) {

    // init
    updateFoo();

    // like onMap() callback
    std::thread update_routine([] {
        for (;;) {
            std::this_thread::sleep_for(5s);
            updateFoo();
        }
    });

    // like run() timer callback
    for (;;) {
        // concurrent unprotected access
        std::shared_ptr<Foo> planner_data = foo_ptr;  // new pointer
        accessFoo(*planner_data);
    }

    update_routine.join();

    return 0;
}
```

I run this code and it seems okay 
```
updating ptr!
foo[0] = 0
foo[1] = 0
foo[2] = 0
foo[3] = 0
foo[4] = 0
updating ptr!
foo[5] = 0
foo[6] = 0
foo[7] = 0
foo[8] = 0
foo[9] = 0
updating ptr!
deleting data...
foo[10] = 0
foo[11] = 0
foo[12] = 0
foo[13] = 0
foo[14] = 0
updating ptr!
deleting data...
foo[15] = 0
foo[16] = 0
foo[17] = 0
foo[18] = 0
foo[19] = 0
updating ptr!
deleting data...
foo[20] = 0
foo[21] = 0
foo[22] = 0
foo[23] = 0
foo[24] = 0
updating ptr!
deleting data...
foo[25] = 0
foo[26] = 0
foo[27] = 0
foo[28] = 0
foo[29] = 0
updating ptr!
deleting data...
foo[30] = 0
foo[31] = 0
foo[32] = 0
foo[33] = 0
foo[34] = 0
updating ptr!
deleting data...
foo[35] = 0
foo[36] = 0
foo[37] = 0
foo[38] = 0
foo[39] = 0
updating ptr!
deleting data...
foo[40] = 0
foo[41] = 0
foo[42] = 0
foo[43] = 0
foo[44] = 0
updating ptr!
deleting data...
foo[45] = 0
foo[46] = 0
foo[47] = 0
```

VRichardJP:
@purewater0901 
Yes, this snippet works on x86:
```C++
    for (;;) {
        // concurrent unprotected access
        std::shared_ptr<Foo> planner_data = foo_ptr;  // new pointer
        accessFoo(*planner_data);
    }
```
However I don't think it's C++ compliant. There are CPU architectures for which a memory fence would be mandatory to read the new `foo_ptr` value (e.g. typically if the CPU cores have separate cache).

purewater0901:
@VRichardJP 
So you mean if the `foo_ptr` is overwritten as
```C++
foo_ptr = std::make_shared<Foo>(100, x)
```
in some places, the new pointer `planner_data` also points to new values?

VRichardJP:
If the `shared_ptr` is duplicated before it is reassigned to something (like `foo_ptr = std::make_shared<Foo>(100, x)`), the old data stays alive (with `planner_data` pointing to it). As long as `planner_data` exists, no one will destroy it.

With this updated example, I think the issue is clearer:
```C++
#include <chrono>
#include <iostream>
#include <memory>
#include <thread>
#include <vector>

using namespace std::chrono_literals;

class Foo {
public:
    Foo(size_t n, int x)
            : data(n, x)
            , x_(x)
    {
        std::cout << "Creating data: x = " << x_ << std::endl;
    }

    ~Foo() {
        std::cout << "deleting data: x = " << x_ << std::endl;
    }

    std::vector<int> data;
    int x_;
};

static std::shared_ptr<Foo> foo_ptr;

void updateFoo() {
    static int x = 0;
    foo_ptr = std::make_shared<Foo>(100, x++);
}

void accessFoo(const Foo &foo) {
    for (int i=0; i < foo.data.size(); ++i) {
        std::cout << "foo[" << i << "] = " << foo.data[i] << std::endl;
        std::this_thread::sleep_for(1s);
    }
}

int main(int argc, char** argv) {

    // init
    updateFoo();

    // like onMap() callback
    std::thread update_routine([] {
        for (;;) {
            std::this_thread::sleep_for(5s);
            updateFoo();
        }
    });

    // like run() timer callback
    for (;;) {
        // concurrent unprotected access
        #if 0
        std::shared_ptr<Foo> planner_data = foo_ptr;  // new pointer
        accessFoo(*planner_data);
        #else
        accessFoo(*foo_ptr);
        #endif
    }

    update_routine.join();

    return 0;
}
```

The only difference is I show what is the value of x when the data is created/deleted:

```bash
$ g++ -O3 main.cc -o main && ./main
Creating data: x = 0
foo[0] = 0
foo[1] = 0
foo[2] = 0
foo[3] = 0
foo[4] = 0
Creating data: x = 1
deleting data: x = 0   <--- deleted!
foo[5] = 0
foo[6] = 0
foo[7] = 0
foo[8] = 0
foo[9] = 0
Creating data: x = 2
deleting data: x = 1
foo[10] = 2
foo[11] = 2
foo[12] = 2
foo[13] = 2
foo[14] = 2
Creating data: x = 3
deleting data: x = 2
foo[15] = 2
foo[16] = 2
```

Here you can clearly see the destructor for `x = 0`, while we are reading its content (through `*foo_ptr`). Note that it would be different if we passed `foo_ptr` by copy.

If the pointer is duplicated first, you see the `x=0` is never deleted while the main thread is running:

```bash
$ g++ -O3 main.cc -o main && ./main
Creating data: x = 0
foo[0] = 0
foo[1] = 0
foo[2] = 0
foo[3] = 0
foo[4] = 0
Creating data: x = 1
foo[5] = 0
foo[6] = 0
foo[7] = 0
foo[8] = 0
foo[9] = 0
Creating data: x = 2
deleting data: x = 1
foo[10] = 0
foo[11] = 0
foo[12] = 0
```

purewater0901:
@VRichardJP Okay, but once `planner_data` is destroyed (after the accessFoo function has finished), the x=0 data is also going to be destroyed right? If so, do we have any serious problems?

VRichardJP:
Yes, it is deleted right after the `accessFoo(*planner_data)`, when the shared_ptr planner_data is destroyed.

VRichardJP:
I ran a couple more test, and I confirm a mutex lock is definitely required:
```C++
#include <chrono>
#include <iostream>
#include <memory>
#include <mutex>
#include <thread>
#include <vector>

using namespace std::chrono_literals;

class Foo {
public:
    Foo(size_t n, int x)
            : data(n, x)
            , x_(x)
    {}

    std::vector<int> data;
    int x_;
};

static std::mutex mtx;
static std::shared_ptr<Foo> foo_ptr;

void updateFoo() {
    static int x = 0;
    std::lock_guard<std::mutex> lk(mtx);
    foo_ptr = std::make_shared<Foo>(100, x++);
}

void accessFoo(const Foo &foo) {
    for (int i=0; i < foo.data.size(); ++i) {
      if (foo.data[i] != foo.x_) {
        throw std::runtime_error("Failed!!");
      }
    }
}

int main(int argc, char** argv) {

    // init
    updateFoo();

    // like onMap() callback
    std::thread update_routine([] {
        for (;;) {
            std::this_thread::sleep_for(1ms);
            updateFoo();
        }
    });

    // like run() timer callback
    for (;;) {
        // Case 1: throw "Failed!!"
        // accessFoo(*foo_ptr);

        // Case 2: SIGABRT
        std::shared_ptr<Foo> planner_data = foo_ptr;  // new pointer
        accessFoo(*planner_data);

        // Case 3:
        // std::shared_ptr<Foo> planner_data{nullptr};
        // {
        //   std::lock_guard<std::mutex> lk(mtx);
        //   planner_data = foo_ptr;
        // }
        // accessFoo(*planner_data);
    }

    update_routine.join();

    return 0;
}
```

This is the same code, just no more print/sleep.

In case 1, I have no lock, no copy. The main detects the data has been modified and throw an error:
```bash
$ g++ -O3 main.cc -o main && ./main
terminate called after throwing an instance of 'std::runtime_error'
  what():  Failed!!
fish: Job 1, './main' terminated by signal SIGABRT (Abort)
```

In case 2, I have no lock, but I copy the pointer. The program crashes due to double free:
```
double free or corruption (top)
fish: Job 1, './main' terminated by signal SIGABRT (Abort)
```

In case 3, I lock ptr access, and make a copy. The program never crash.

purewater0901:
@VRichardJP I also run the code and encountered the exact same issue. Thank you for your experiment, and I get a deeper insight into multithreading.

purewater0901:
For case2, do you know why we have the double free problem?

VRichardJP:
Because I added optimization `-O3` ^^
To be precise, without optimizations，case 1 and 2 both fail because of the runtime error.
```bash
$ g++ main.cc -o main && ./main
terminate called after throwing an instance of 'std::runtime_error'
  what():  Failed!!
Aborted (core dumped)
```
When the optimizations are turned on, the case 1 seems to always fail with the runtime error, but the case 2 has a random memory error (double free is one of them):
```sh
$ g++ -O2 main.cc -o main && ./main
free(): too many chunks detected in tcache
Aborted (core dumped)
$ g++ -O2 main.cc -o main && ./main
double free or corruption (out)
Aborted (core dumped)
$ g++ -O2 main.cc -o main && ./main
Segmentation fault (core dumped)
```

This illustrates the undefined behavior I pointed out earlier: in both cases 1 and 2, nothing says `foo_ptr` could be modified by someone else, so the compiler may assume the pointer will never change and optimize things (reorder, move data, skip planner_data, etc), and totally break the code logic doing so.

purewater0901:
@VRichardJP I see. Thank you so much. I'll approve this PR

purewater0901 (approved):
LGTM

VRichardJP:
yes

PR #1821, opened 2022-09-09
feat(trtexec_vendor): add vendor package for trtexec

wep21:
## Description

a vendor package for `trtexec`

## Tests performed
Please check if you can build tensorrt engine from onnx file with `trtexec`
```
colcon build --cmake-args -DCMAKE_BUILD_TYPE=Release  --symlink-install --continue-on-error --packages-up-to trtexec_vendor lidar_centerpoint
source install/setup.bash
trtexec --onnx=$(ros2 pkg prefix lidar_centerpoint --share)/data/pts_backbone_neck_head_centerpoint.onnx --shapes=spatial_features:1x32x560x560 --saveEngine=$(ros2 pkg prefix lidar_centerpoint --share)/data/pts_backbone_neck_head_centerpoint.engine --buildOnly
```

## Pre-review checklist for the PR author

The PR author **must** check the checkboxes below when creating the PR.

- [x] I've confirmed the [contribution guidelines].
- [x] The PR follows the [pull request guidelines].

## In-review checklist for the PR reviewers

The PR reviewers **must** check the checkboxes below before approval.

- [ ] The PR follows the [pull request guidelines].
- [ ] The PR has been properly tested.
- [ ] The PR has been reviewed by the code owners.

## Post-review checklist for the PR author

The PR author **must** check the checkboxes below before merging.

- [x] There are no open discussions or they are tracked via tickets.
- [x] The PR is ready for merge.

After all checkboxes are checked, anyone who has write access can merge the PR.

[contribution guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/
[pull request guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/pull-request-guidelines/

PR #2982, opened 2023-03-02
feat(tensorrt_yolox): add build only option for tensorrt engine

yukke42:
## Description

- the option to build tensorrt engine without inference is required for test on cloud
- add depend package
- fix documentation

### DIFF
--- a/perception/tensorrt_yolox/README.md
+++ b/perception/tensorrt_yolox/README.md
@@ -29,20 +29,21 @@ Zheng Ge, Songtao Liu, Feng Wang, Zeming Li, Jian Sun, "YOLOX: Exceeding YOLO Se
 
 ### Core Parameters
 
-| Name              | Type  | Default Value | Description                                                                           |
-| ----------------- | ----- | ------------- | ------------------------------------------------------------------------------------- |
-| `score_threshold` | float | 0.3           | If the objectness score is less than this value, the object is ignored in yolo layer. |
-| `nms_threshold`   | float | 0.7           | The IoU threshold for NMS method                                                      |
+| Name              | Type  | Default Value | Description                                                                            |
+| ----------------- | ----- | ------------- | -------------------------------------------------------------------------------------- |
+| `score_threshold` | float | 0.3           | If the objectness score is less than this value, the object is ignored in yolox layer. |
+| `nms_threshold`   | float | 0.7           | The IoU threshold for NMS method                                                       |
 
 **NOTE:** These two parameters are only valid for "plain" model (described later).
 
 ### Node Parameters
 
-| Name         | Type   | Default Value | Description                                                        |
-| ------------ | ------ | ------------- | ------------------------------------------------------------------ |
-| `onnx_file`  | string | ""            | The onnx file name for yolo model                                  |
-| `label_file` | string | ""            | The label file with label names for detected objects written on it |
-| `mode`       | string | "fp32"        | The inference mode: "fp32", "fp16", "int8"                         |
+| Name            | Type   | Default Value | Description                                                        |
+| --------------- | ------ | ------------- | ------------------------------------------------------------------ |
+| `model_path`    | string | ""            | The onnx file name for yolox model                                 |
+| `label_path`    | string | ""            | The label file with label names for detected objects written on it |
+| `trt_precision` | string | "fp32"        | The inference mode: "fp32", "fp16", "int8"                         |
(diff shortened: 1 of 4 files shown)


wep21:
@yukke42 Why don't you use `trtexec`?

perception/tensorrt_yolox/package.xml
```diff
@@ -23,6 +23,7 @@
   <depend>cuda_utils</depend>
   <depend>cv_bridge</depend>
   <depend>image_transport</depend>
+  <depend>image_transport_decompressor</depend>
```
wep21, line 26: I guess this is execution dependency.

yukke42: changed at [4536ca8](https://github.com/autowarefoundation/autoware.universe/pull/2982/commits/4536ca87c2835d12efbdc43671087b6847c211c5)

yukke42:
> @yukke42 Why don't you use `trtexec`?

@wep21 
If the tensorrt_yolox package becomes more complicated in terms of cuda/tensorrt options, s.t. int8, calibration, ... etc, 
trtexec can be used by develpers who knows both trtexec and this package. But, by this option, they don't have to know them and just use the same commnd as autoware uses .

wep21:
> they don't need know them and just add the build_only arg to launch command.
But they do need to know launch options such as int8, calibrator defined in each package(not only this package but other packages which use tensorrt), right?
I thin `trtexec` is a generized tool which covers most use cases to build tensorrt engine file, so we don't have to add build only option in each package.

yukke42:
> But they do need to know launch options such as int8, calibrator defined in each package(not only this package but other packages which use tensorrt), right?
I think trtexec is a generalized tool which covers most use cases to build tensorrt engine file, so we don't have to add build only option in each package.

@wep21 Yes, we can do the same thing with this option and trtexec. But it's hard to reproduce the same engine file automatically.
tensorrt_yolox uses `optimizationProfile` and I think it's `--optShapes` of trtexec. It requires the magic number like [this](https://github.com/autowarefoundation/autoware.universe/pull/1821#issuecomment-1246151813) and needs another scirpt to get values from onnx file for it to delete these magic number. And, traffic_light_classifier don't use  `optimizationProfile` , the package don't need opt option.
If some contrbutes change the tensorrt option/implementation, we need to catch up all the chages and change trtexec options as well. This is for not only this package but also other packages require tensorrt.

wep21:
It may be out of scope from this PR, but I think the problem is that each package has own implementation to build tensorrt engine, while they do almost the same process. I think it is better to use `trtexec` to build tensorrt engine in launch file like
```
<launch>
  <executable cmd="trtexec --onnx=$(var onnx) --saveEngine=$(var enigne) --optShapes=input:1x3x$(var input_h)x$(var input_w)" name="trt_build" output="screen" shell="true"/>

  <node pkg="yolo" exec="yolo" name="yolo">
    <param name="engine" value="$(var engine)" />
  </node>
<launch>
```
and then remove engine build function from each package. What do you think?
Anyway, I won't block this PR if you need this feature as soon as possible for cloud test.

yukke42:
@wep21 I see, Let me make a new discussion topic about it. 
I hope you allow this pr to merge and then I'll chnage it later.

perception/tensorrt_yolox/README.md
```diff
@@ -29,20 +29,21 @@ Zheng Ge, Songtao Liu, Feng Wang, Zeming Li, Jian Sun, "YOLOX: Exceeding YOLO Se
...
-| ------------ | ------ | ------------- | ------------------------------------------------------------------ |
-| `onnx_file`  | string | ""            | The onnx file name for yolo model                                  |
-| `label_file` | string | ""            | The label file with label names for detected objects written on it |
-| `mode`       | string | "fp32"        | The inference mode: "fp32", "fp16", "int8"                         |
+| Name            | Type   | Default Value | Description                                                        |
+| --------------- | ------ | ------------- | ------------------------------------------------------------------ |
+| `model_path`    | string | ""            | The onnx file name for yolox model                                 |
+| `label_path`    | string | ""            | The label file with label names for detected objects written on it |
+| `trt_precision` | string | "fp32"        | The inference mode: "fp32", "fp16", "int8"                         |
+| `build_only`    | bool   | false         | shutdown node after TensorRT engin file is built                   |
```
wep21, line 46:
```suggestion
| `build_only`    | bool   | false         | shutdown node after TensorRT engine file is built                  |
```
typo

wep21 (approved):
LGTM

Issue #2970, opened 2023-03-01
`ring_outlier_filter` output is sometimes empty


VRichardJP:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I'm convinced that this is not my fault but a bug.

### Description

Following the [sample sensor kit](https://github.com/autowarefoundation/sample_sensor_kit_launch/blob/06330e6b07449a3609ced295c0a9ce6a1805ef99/common_sensor_launch/launch/velodyne_node_container.launch.py#L154) I have enabled the ring outlier filter on our car.

After that the `ndt_scan_matcher` started to crash on startup on receiving empty input cloud:
![2023-03-01-17-09-07-209](https://user-images.githubusercontent.com/18645627/222082549-c1e0a32b-2be4-442d-9601-939285548d06.jpg)

I found `ring_outlier_filter` is the node responsible for this. On the picture below, I printed on the left the size of `ring_outlier_filter` output cloud (`$ ros2 topic echo /sensing/lidar/top/outlier_filtered/pointcloud --field width`) and on the right the size of its input cloud (`$ ros2 topic echo /sensing/lidar/top/rectified/pointcloud_ex --field width`):
![Inked2023-03-01-17-06-10-959_LI](https://user-images.githubusercontent.com/18645627/222082538-6cc4ef4e-c886-4dd4-8a23-d8045910d923.jpg)

As you can see, the output is sometimes empty. 

I could also confirm with Rviz the rectified cloud looks totally normal while the the outlier cloud sometimes "blinks".

Besides the output being empty, I think it is also annoying `ndt_scan_matcher` crashes because of that.  A small guard could be added to `NDTScanMatcher::callback_sensor_points`.

### Expected behavior

No empty output

### Actual behavior

Sometimes empty output

### Steps to reproduce

use the outlier filter?

kminoda:
I also observed the similar issue (where the number of points becomes zero occasionally) in the latest awf/autoware with sample rosbag provided in the Autoware tutorial. 

@sykwer Do you think the above phenomena is related to this PR? https://github.com/autowarefoundation/autoware.universe/pull/2908/
(Anyway, we need to have some sort of error-catch algorithm in ndt_scan_matcher to accept 0 point inputs)

YamatoAndo:
Thank you for your report!

[...]

PR #2908, opened 2023-02-18
perf(crop_box_filter): performance tuning with filter() API change

sykwer:
## Description
This PR makes `crop_box_filter` faster without changing the logical output. The tail latency of the `crop_box_filter_self` node gets about x5 faster with the introduction of the TLSF allocator (see [this Autoware Discussion](https://github.com/autowarefoundation/autoware/discussions/3274)) and this PR merge.

crop_box_filter_self_performance_improvement_page-0001
Measurement Condition
- Ubuntu22.04 + ROS2 Humble + [Autoware Universe rosbag simulation](https://autowarefoundation.github.io/autoware-documentation/main/tutorials/ad-hoc-simulation/rosbag-replay-simulation/)
- Core Isolated
- Core Frequency Fixed (2.6GHz)
- L3 Cache: 12MB

This PR contains two controversial changes.
- In the future, `filter(const PointCloud2ConstPtr & input, const IndicesPtr & indices, PointCloud2 & output)` will be changed to `filter(const PointCloud2ConstPtr & input, const IndicesPtr & indices, PointCloud2 & output, const TransformInfo & transform_info)` and classes inheriting from `Filter` class will be changed to follow this new API.
  - This unaesthetic change in design is due to performance reasons, which will be explained later.
  - Since there are several child classes inheriting from the `Filter` class, and this time only the `CropBoxFilter` class should support the new API, a temporary `faster_filter()` function is prepared and used for the transition period until all child classes support the new API.
- Some internal implementations in `pcl_ros::transformPointCloud()` are ported into the Autoware source code and we stop using this library function.
  - It's also for performance reasons.

The first API change is needed in order to defer the coordinate transformation of the point clouds. If a point cloud is transformed into its coordinate in the `Filter` parent class before passing to a child class, it's inevitable to consume one more buffer, leading to one more copy and extra memory space. Therefore, I'd like to change the design so that only the transformation matrix of the coordinates is passed to the child class, and the transformation is applied e.g. just before a point in the input point cloud is referenced.

The ported part from `pcl_ros::transformPointCloud()` is where a matrix of the coordinate transformation is calculated.

The `faster_filter()` API will be removed in the following steps.
1. The implementations of all the child classes conform to the new API `faster_filter(const PointCloud2ConstPtr & input, const IndicesPtr & indices, PointCloud2 & output, const TransformInfo & transform_info)`
2. Remove `faster_filter()` API (= renamed to `filter()`)

## Related links
[...]

PR #2992, opened 2023-03-03
fix(crop_box_filter): fix bug

sykwer:
## Description
Fix bug in https://github.com/autowarefoundation/autoware.universe/pull/2908.
Resolve https://github.com/autowarefoundation/autoware.universe/issues/2970.

### DIFF
--- a/sensing/pointcloud_preprocessor/src/crop_box_filter/crop_box_filter_nodelet.cpp
+++ b/sensing/pointcloud_preprocessor/src/crop_box_filter/crop_box_filter_nodelet.cpp
@@ -143,7 +143,12 @@ void CropBoxFilterComponent::faster_filter(
                            point[1] > param_.min_y && point[1] < param_.max_y &&
                            point[0] > param_.min_x && point[0] < param_.max_x;
     if ((!param_.negative && point_is_inside) || (param_.negative && !point_is_inside)) {
-      memcpy(&output.data[output_size], &point, input->point_step);
+      memcpy(&output.data[output_size], &input->data[global_offset], input->point_step);
+
+      *reinterpret_cast<float *>(&output.data[output_size + x_offset]) = point[0];
+      *reinterpret_cast<float *>(&output.data[output_size + y_offset]) = point[1];
+      *reinterpret_cast<float *>(&output.data[output_size + z_offset]) = point[2];
+
       output_size += input->point_step;
     }
   }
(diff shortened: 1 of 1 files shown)


yukkysaito (approved):
LGTM

sensing/pointcloud_preprocessor/src/crop_box_filter/crop_box_filter_nodelet.cpp
```diff
@@ -143,7 +143,12 @@ void CropBoxFilterComponent::faster_filter(
                            point[1] > param_.min_y && point[1] < param_.max_y &&
                            point[0] > param_.min_x && point[0] < param_.max_x;
     if ((!param_.negative && point_is_inside) || (param_.negative && !point_is_inside)) {
-      memcpy(&output.data[output_size], &point, input->point_step);
+      memcpy(&output.data[output_size], &input->data[global_offset], input->point_step);
```
VRichardJP, line 146: The old `memcpy` was certainly wrong, because `point` is a `Eigen::Vector4f`, whose memory layout is not necessarily the same than `output.data`. However with the fixed `memcpy` why do we need to copy points xyz manually after? The `memcpy` line should be enough no?

VRichardJP (changes requested):

VRichardJP:
Although it is not related to the bug, I think there is another problem with the implementation: the last point of the point cloud is ignored:
https://github.com/autowarefoundation/autoware.universe/blob/4c9868953c4f88bf3e8e422f9036ce740472f63e/sensing/pointcloud_preprocessor/src/crop_box_filter/crop_box_filter_nodelet.cpp#L125-L126

Let's take a simple example: input cloud is 1D and has 1 point. Then: 
- `input->point_step = 1`
- `input->data.size() = 1`
So when the loop starts (`global_offset = 0`), the clause `global_offset + input->point_step < input->data.size()` is false and we immediately leave the loop, ignoring the first (last) point.

Could you fix that as well?

sensing/pointcloud_preprocessor/src/crop_box_filter/crop_box_filter_nodelet.cpp
sykwer: This is because `input` contains the data before the coordinate transformation, so the data after the transformation must be overwritten later.

VRichardJP: In that case would it be faster if we add the copy in a `if (transform_info.need_transform) {}` ?
the "fast" version would be something like:
```c++
if (transform_info.need_transform) {
  for (...) {
    Eigen::Vector4f point(...);
    // transform then memcpy and transformed xyz copy if inside
  }
} else {
  for (...) {
    Eigen::Vector4f point(...);
    // memcpy only if inside
  }
}
```

PR #2994, opened 2023-03-03
fix(emergency_handler): fix bug of takeover_request

veqcc:
Signed-off-by: veqcc [ryuta.kambe@tier4.jp]

@mkuri @asana17 

## Description
The parameter `is_takeover_requested_` was not properly used. It was changed `true` when `MRMState::NORMAL`. However, the timeout process was implemented only in the state `MRMState::OPERATING`.

### DIFF
--- a/system/emergency_handler/src/emergency_handler/emergency_handler_core.cpp
+++ b/system/emergency_handler/src/emergency_handler/emergency_handler_core.cpp
@@ -380,14 +380,32 @@ void EmergencyHandler::updateMrmState()
   // State Machine
   if (mrm_state_.state == MrmState::NORMAL) {
     // NORMAL
-    if (is_auto_mode && is_emergency) {
-      if (param_.use_takeover_request) {
-        takeover_requested_time_ = this->get_clock()->now();
-        is_takeover_request_ = true;
+
+    if (is_takeover_request_) {
+      const auto time_from_takeover_request = this->get_clock()->now() - takeover_requested_time_;
+      if (time_from_takeover_request.seconds() > param_.timeout_takeover_request) {
+        transitionTo(MrmState::MRM_OPERATING);
+        is_takeover_request_ = false;
         return;
       } else {
-        transitionTo(MrmState::MRM_OPERATING);
-        return;
+        if (!is_emergency || is_takeover_done) {
+          is_takeover_request_ = false;
+        } else {
+          // do nothing
+        }
+      }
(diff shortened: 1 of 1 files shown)


isamu-takagi:
The takeover state is removed when changing from EmergencyState to MrmState. What is this parameter currently used for?

mkuri:
@isamu-takagi 
> The takeover state is removed when changing from EmergencyState to MrmState. What is this parameter currently used for?

The takeover state on the officially supported API was removed when MrmState was created, but the takeover state in the emergency handler that existed in the previous design remains.

The current behavior is that when the above parameter is true, the system waits for a certain period of time after a failure, and if the operator does not takeover, the system will switch to MRM.
The following two options are available.

- Remove all features related to takeover (maybe some products use the takeover feature)
- Leave it as a base for when takeover will be formally designed

This PR is based on the latter policy, and is a PR to fix a bug found in the existing takeover state.

isamu-takagi:
@mkuri 
I got it. I understood that old `TAKEOVER_REQUETING` state is now `NONE` state. However, the `operateMrm` function clears the `behavior` when the state is `NONE`. I think it should be retained.

isamu-takagi:
@mkuri @veqcc 
[MRM state](https://autowarefoundation.github.io/autoware-documentation/main/design/autoware-interfaces/ad-api/features/fail-safe/) expects `{state: MRM_OPERATING, behavior: NONE}` during takeover requesting. Neither the current nor fixed implementation fits this. So I think it's better to remove this processing. The takeover request will be a new topic and are being discussed in https://github.com/autowarefoundation/autoware/issues/3487.

mkuri:
@isamu-takagi 
Yes, I agree with removing the take_over process.

@veqcc 
How about we close this PR and remove the process related to take over in the new PR to solve the following issue?

> The parameter is_takeover_requested_ was not properly used. It was changed true when MRMState::NORMAL. However, the timeout process was implemented only in the state MRMState::OPERATING.

Reference
https://github.com/autowarefoundation/autoware.universe/pull/2994#issuecomment-1464849767
> The following two options are available.
> Remove all features related to takeover (maybe some products use the takeover feature)
> Leave it as a base for when takeover will be formally designed

I will adopt the former policy.

veqcc:
> @veqcc
How about we close this PR and remove the process related to take over in the new PR to solve the following issue?
I see. I will close it and make a new one.

PR #3014, opened 2023-03-06
perf(ring_outlier_filter): performance tuning

sykwer:
## Description

This PR makes `ring_outlier_filter` faster without changing the logical output. The tail latency of the `ring_outlier_filter` node gets about x3 faster with the introduction of the TLSF allocator (see [this Autoware Discussion](https://github.com/autowarefoundation/autoware/discussions/3274)) and this PR merged.

ring_outlier_filter_performance_improvement_page-0001

Measurement Condition
- Ubuntu22.04 + ROS2 Humble + [Autoware Universe rosbag simulation](https://autowarefoundation.github.io/autoware-documentation/main/tutorials/ad-hoc-simulation/rosbag-replay-simulation/)
- Core Isolated
- Core Frequency Fixed (2.6GHz)
- L3 Cache: 12MB
## Related links
- [Introduction of a library for the replacement of heap allocation in Autoware (Autoware Discussion)](https://github.com/autowarefoundation/autoware/discussions/3274)
- [Background, Implementation Detail and Performance Analysis (TIER Ⅳ INTERNAL LINK)](https://tier4.atlassian.net/wiki/spaces/~5ed0b4584824b20c18371c06/pages/2610726095/awf.tutorial+ring+outlier+filter)

## Tests performed

Check if `ring_outlier_filter` publishes the same logical output as before (using [Autoware Universe rosbag simulation](https://autowarefoundation.github.io/autoware-documentation/main/tutorials/ad-hoc-simulation/rosbag-replay-simulation/))

### DIFF
--- a/sensing/pointcloud_preprocessor/docs/ring-outlier-filter.md
+++ b/sensing/pointcloud_preprocessor/docs/ring-outlier-filter.md
@@ -22,11 +22,13 @@ This implementation inherits `pointcloud_preprocessor::Filter` class, please ref
 
 ### Core Parameters
 
-| Name                      | Type   | Default Value | Description |
-| ------------------------- | ------ | ------------- | ----------- |
-| `distance_ratio`          | double | 1.03          |             |
-| `object_length_threshold` | double | 0.1           |             |
-| `num_points_threshold`    | int    | 4             |             |
+| Name                      | Type    | Default Value | Description |
+| ------------------------- | ------- | ------------- | ----------- |
+| `distance_ratio`          | double  | 1.03          |             |
+| `object_length_threshold` | double  | 0.1           |             |
+| `num_points_threshold`    | int     | 4             |             |
+| `max_rings_num`           | uint_16 | 128           |             |
+| `max_points_num_per_ring` | size_t  | 2000          |             |
 
 ## Assumptions / Known limits
 
--- a/sensing/pointcloud_preprocessor/docs/ring-outlier-filter.md
+++ b/sensing/pointcloud_preprocessor/docs/ring-outlier-filter.md
@@ -22,13 +22,13 @@ This implementation inherits `pointcloud_preprocessor::Filter` class, please ref
 
 ### Core Parameters
 
-| Name                      | Type    | Default Value | Description |
-| ------------------------- | ------- | ------------- | ----------- |
-| `distance_ratio`          | double  | 1.03          |             |
-| `object_length_threshold` | double  | 0.1           |             |
-| `num_points_threshold`    | int     | 4             |             |
-| `max_rings_num`           | uint_16 | 128           |             |
-| `max_points_num_per_ring` | size_t  | 2000          |             |
+| Name                      | Type    | Default Value | Description                                                                         |
+| ------------------------- | ------- | ------------- | ----------------------------------------------------------------------------------- |
+| `distance_ratio`          | double  | 1.03          |                                                                                     |
+| `object_length_threshold` | double  | 0.1           |                                                                                     |
+| `num_points_threshold`    | int     | 4             |                                                                                     |
+| `max_rings_num`           | uint_16 | 128           |                                                                                     |
+| `max_points_num_per_ring` | size_t  | 4000          | Set this value large enough such that `HFoV / resolution < max_points_num_per_ring` |
 
 ## Assumptions / Known limits
 
--- a/sensing/pointcloud_preprocessor/include/pointcloud_preprocessor/outlier_filter/ring_outlier_filter_nodelet.hpp
+++ b/sensing/pointcloud_preprocessor/include/pointcloud_preprocessor/outlier_filter/ring_outlier_filter_nodelet.hpp
@@ -53,15 +53,26 @@ private:
   rcl_interfaces::msg::SetParametersResult paramCallback(const std::vector<rclcpp::Parameter> & p);
 
   bool isCluster(
-    const PointCloud2ConstPtr & input, int first_data_idx, int last_data_idx, int walk_size)
+    const PointCloud2ConstPtr & input, std::pair<int, int> data_idx_both_ends, int walk_size, const TransformInfo & transform_info)
   {
-    auto first_point = reinterpret_cast<const PointXYZI *>(&input->data[first_data_idx]);
-    auto last_point = reinterpret_cast<const PointXYZI *>(&input->data[last_data_idx]);
-    const auto x = first_point->x - last_point->x;
-    const auto y = first_point->y - last_point->y;
-    const auto z = first_point->z - last_point->z;
-    return (walk_size > num_points_threshold_) ||
-           (x * x + y * y + z * z >= object_length_threshold_ * object_length_threshold_);
+    if (walk_size > num_points_threshold_) return true;
+
+    auto first_point = reinterpret_cast<const PointXYZI *>(&input->data[data_idx_both_ends.first]);
+    auto last_point = reinterpret_cast<const PointXYZI *>(&input->data[data_idx_both_ends.second]);
+
+    Eigen::Vector4f p1(first_point->x, first_point->y, first_point->z, 1);
+    Eigen::Vector4f p2(last_point->x, last_point->y, last_point->z, 1);
+
+    if (transform_info.need_transform) {
+      p1 = transform_info.eigen_transform * p1;
+      p2 = transform_info.eigen_transform * p2;
+    }
+
+    auto x = p1[0] - p2[0];
+    auto y = p1[1] - p2[1];
+    auto z = p1[2] - p2[2];
+
+    return x * x + y * y + z * z >= object_length_threshold_ * object_length_threshold_;
   }
 
 public:
--- a/sensing/pointcloud_preprocessor/include/pointcloud_preprocessor/outlier_filter/ring_outlier_filter_nodelet.hpp
+++ b/sensing/pointcloud_preprocessor/include/pointcloud_preprocessor/outlier_filter/ring_outlier_filter_nodelet.hpp
@@ -33,6 +33,11 @@ protected:
   virtual void filter(
     const PointCloud2ConstPtr & input, const IndicesPtr & indices, PointCloud2 & output);
 
+  // TODO(sykwer): Temporary Implementation: Remove this interface when all the filter nodes conform to new API
+  virtual void faster_filter(
+      const PointCloud2ConstPtr & input, const IndicesPtr & indices, PointCloud2 & output,
+      const TransformInfo & transform_info);
+
 private:
   double distance_ratio_;
   double object_length_threshold_;
@@ -44,19 +49,13 @@ private:
   /** \brief Parameter service callback */
   rcl_interfaces::msg::SetParametersResult paramCallback(const std::vector<rclcpp::Parameter> & p);
 
-  bool isCluster(
-    sensor_msgs::msg::PointCloud2::SharedPtr & input_ptr,
-    const std::vector<std::size_t> & tmp_indices)
-  {
-    PointXYZI * front_pt = reinterpret_cast<PointXYZI *>(&input_ptr->data[tmp_indices.front()]);
-    PointXYZI * back_pt = reinterpret_cast<PointXYZI *>(&input_ptr->data[tmp_indices.back()]);
-
-    const auto x_diff = front_pt->x - back_pt->x;
-    const auto y_diff = front_pt->y - back_pt->y;
-    const auto z_diff = front_pt->z - back_pt->z;
-    return static_cast<int>(tmp_indices.size()) > num_points_threshold_ ||
-           (x_diff * x_diff) + (y_diff * y_diff) + (z_diff * z_diff) >=
-             object_length_threshold_ * object_length_threshold_;
+  bool isCluster(const PointCloud2ConstPtr & input, int first_data_idx, int last_data_idx, int walk_size) {
+    auto first_point = reinterpret_cast<const PointXYZI*>(&input->data[first_data_idx]);
+    auto last_point = reinterpret_cast<const PointXYZI*>(&input->data[last_data_idx]);
+    const auto x = first_point->x - last_point->x;
+    const auto y = first_point->y - last_point->y;
+    const auto z = first_point->z - last_point->z;
+    return (walk_size > num_points_threshold_) || (x*x + y*y + z*z >= object_length_threshold_ * object_length_threshold_);
   }
 
 public:
--- a/sensing/pointcloud_preprocessor/include/pointcloud_preprocessor/outlier_filter/ring_outlier_filter_nodelet.hpp
+++ b/sensing/pointcloud_preprocessor/include/pointcloud_preprocessor/outlier_filter/ring_outlier_filter_nodelet.hpp
@@ -33,10 +33,11 @@ protected:
   virtual void filter(
     const PointCloud2ConstPtr & input, const IndicesPtr & indices, PointCloud2 & output);
 
-  // TODO(sykwer): Temporary Implementation: Remove this interface when all the filter nodes conform to new API
+  // TODO(sykwer): Temporary Implementation: Remove this interface when all the filter nodes conform
+  // to new API
   virtual void faster_filter(
-      const PointCloud2ConstPtr & input, const IndicesPtr & indices, PointCloud2 & output,
-      const TransformInfo & transform_info);
+    const PointCloud2ConstPtr & input, const IndicesPtr & indices, PointCloud2 & output,
+    const TransformInfo & transform_info);
 
 private:
   double distance_ratio_;
@@ -49,13 +50,16 @@ private:
   /** \brief Parameter service callback */
   rcl_interfaces::msg::SetParametersResult paramCallback(const std::vector<rclcpp::Parameter> & p);
 
-  bool isCluster(const PointCloud2ConstPtr & input, int first_data_idx, int last_data_idx, int walk_size) {
-    auto first_point = reinterpret_cast<const PointXYZI*>(&input->data[first_data_idx]);
-    auto last_point = reinterpret_cast<const PointXYZI*>(&input->data[last_data_idx]);
+  bool isCluster(
+    const PointCloud2ConstPtr & input, int first_data_idx, int last_data_idx, int walk_size)
+  {
+    auto first_point = reinterpret_cast<const PointXYZI *>(&input->data[first_data_idx]);
+    auto last_point = reinterpret_cast<const PointXYZI *>(&input->data[last_data_idx]);
     const auto x = first_point->x - last_point->x;
     const auto y = first_point->y - last_point->y;
     const auto z = first_point->z - last_point->z;
-    return (walk_size > num_points_threshold_) || (x*x + y*y + z*z >= object_length_threshold_ * object_length_threshold_);
+    return (walk_size > num_points_threshold_) ||
+           (x * x + y * y + z * z >= object_length_threshold_ * object_length_threshold_);
   }
 
 public:
--- a/sensing/pointcloud_preprocessor/include/pointcloud_preprocessor/outlier_filter/ring_outlier_filter_nodelet.hpp
+++ b/sensing/pointcloud_preprocessor/include/pointcloud_preprocessor/outlier_filter/ring_outlier_filter_nodelet.hpp
@@ -20,6 +20,7 @@
 
 #include <point_cloud_msg_wrapper/point_cloud_msg_wrapper.hpp>
 
+#include <utility>
 #include <vector>
 
 namespace pointcloud_preprocessor
--- a/sensing/pointcloud_preprocessor/include/pointcloud_preprocessor/outlier_filter/ring_outlier_filter_nodelet.hpp
+++ b/sensing/pointcloud_preprocessor/include/pointcloud_preprocessor/outlier_filter/ring_outlier_filter_nodelet.hpp
@@ -53,7 +53,8 @@ private:
   rcl_interfaces::msg::SetParametersResult paramCallback(const std::vector<rclcpp::Parameter> & p);
 
   bool isCluster(
-    const PointCloud2ConstPtr & input, std::pair<int, int> data_idx_both_ends, int walk_size, const TransformInfo & transform_info)
+    const PointCloud2ConstPtr & input, std::pair<int, int> data_idx_both_ends, int walk_size,
+    const TransformInfo & transform_info)
   {
     if (walk_size > num_points_threshold_) return true;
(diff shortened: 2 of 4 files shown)


sensing/pointcloud_preprocessor/src/outlier_filter/ring_outlier_filter_nodelet.cpp
```diff
@@ -44,82 +44,116 @@ RingOutlierFilterComponent::RingOutlierFilterComponent(const rclcpp::NodeOptions
...
+  ring2indices.reserve(128);
+
+  for (size_t i = 0; i < 128; i++) {
+    ring2indices.push_back(std::vector<size_t>());
+    ring2indices.back().reserve(2000);  // TODO(sykwer): Make it parameter
+  }
+
+  for (size_t data_idx = 0; data_idx < input->data.size(); data_idx += input->point_step) {
+    const uint16_t ring = *reinterpret_cast<const uint16_t *>(&input->data[data_idx + ring_offset]);
+    ring2indices[ring].push_back(data_idx);
```
VRichardJP, line 79: Out-of-bound access if `ring` is more than 127. Even if there is no Lidar sensor with more than 128 channels, I don't think this is future-proof. We could at least have a `max_channel` as parameter?

sykwer: yes

sensing/pointcloud_preprocessor/src/outlier_filter/ring_outlier_filter_nodelet.cpp
```diff
@@ -44,82 +44,116 @@ RingOutlierFilterComponent::RingOutlierFilterComponent(const rclcpp::NodeOptions
...
-  PointCloud2Modifier<PointXYZI> output_modifier{output, input->header.frame_id};
-  output_modifier.reserve(input->width);
-
-  std::vector<std::size_t> tmp_indices;
-  tmp_indices.reserve(input->width);
-
   const auto azimuth_offset =
     input->fields.at(static_cast<size_t>(autoware_point_types::PointIndex::Azimuth)).offset;
   const auto distance_offset =
     input->fields.at(static_cast<size_t>(autoware_point_types::PointIndex::Distance)).offset;
```
VRichardJP, line 67: Just matter of readability: I think you could define a local helper lambda `getDistance` to get of all the verbose reinterpret cast after, like this:
```c++
auto getDistance = [&](const size_t idz) { return *reinterpret_cast<const float *>(&input->data[idx + distance_offset]); }
```
And the same for azimut. It would also limit the risk to misindexing.

sensing/pointcloud_preprocessor/src/outlier_filter/ring_outlier_filter_nodelet.cpp
```diff
@@ -44,82 +44,116 @@ RingOutlierFilterComponent::RingOutlierFilterComponent(const rclcpp::NodeOptions
...
-          output_modifier.push_back(
-            std::move(*reinterpret_cast<PointXYZI *>(&input_ptr->data[tmp_idx])));
+
+      if (isCluster(
+            input, indices[walk_first_idx], indices[walk_last_idx],
+            walk_last_idx - walk_first_idx + 1)) {
+        for (int i = walk_first_idx; i <= walk_last_idx; i++) {
+          auto output_ptr = reinterpret_cast<PointXYZI *>(&output.data[output_size]);
+          *output_ptr = *reinterpret_cast<const PointXYZI *>(&input->data[indices[i]]);
+          output_size += output.point_step;
```
VRichardJP, line 124: This part assume data layout is `PointXYZI` or at least that input memory layout starts the same than `PointXYZI`. In practice it will just copy the first 4 fields. Is there a reason to ignore the other fields from input (ring, azimut, etc?) I know today Autoware use the ring filter as the last filter that needs "extended" fields, but who knows if it will still be true in the future?

sensing/pointcloud_preprocessor/src/outlier_filter/ring_outlier_filter_nodelet.cpp
```diff
@@ -44,82 +44,116 @@ RingOutlierFilterComponent::RingOutlierFilterComponent(const rclcpp::NodeOptions
...
+        *reinterpret_cast<const float *>(&input->data[current_data_idx + distance_offset]);
+      const float & next_distance =
+        *reinterpret_cast<const float *>(&input->data[next_data_idx + distance_offset]);
 
       if (
-        std::max(current_pt_distance, next_pt_distance) <
-          std::min(current_pt_distance, next_pt_distance) * distance_ratio_ &&
+        std::max(current_distance, next_distance) <
+          std::min(current_distance, next_distance) * distance_ratio_ &&
         azimuth_diff < 100.f) {
```
VRichardJP, line 114: should `azimuth_diff < 100.f` be a parameter?

sensing/pointcloud_preprocessor/src/outlier_filter/ring_outlier_filter_nodelet.cpp
```diff
@@ -44,82 +44,116 @@ RingOutlierFilterComponent::RingOutlierFilterComponent(const rclcpp::NodeOptions
...
+
+  for (const auto & indices : ring2indices) {
+    if (indices.size() < 2) continue;
+
+    walk_first_idx = 0;
+    walk_last_idx = -1;
+
+    for (size_t idx = 0U; idx < indices.size() - 1; ++idx) {
+      const size_t & current_data_idx = indices[idx];
+      const size_t & next_data_idx = indices[idx + 1];
```
VRichardJP, line 94: out-of-bound access when `idx = indices.size()-1`

sensing/pointcloud_preprocessor/src/outlier_filter/ring_outlier_filter_nodelet.cpp
```diff
@@ -44,82 +44,116 @@ RingOutlierFilterComponent::RingOutlierFilterComponent(const rclcpp::NodeOptions
...
+      walk_last_idx = idx;
```
VRichardJP, line 95: I think we are skipping the last point with this logic.
For example, let's imagine a perfect circle ring with N points.

When we first enter the first loop we have :
```
idx = 0
walk_first_idx = 0
walk_last_idx = idx = 0
current_data_idx = indices[0]
next_data_idx= indices[1]
```

The ring is a perfect circle, so all the points are part of the same walk. So we always continue here:
```c++
      if (
        std::max(current_distance, next_distance) <
          std::min(current_distance, next_distance) * distance_ratio_ &&
        azimuth_diff < 100.f) {
        continue;  // Determined to be included in the same walk
      }
```

On the last loop we have:
```
idx = N-2
walk_first_idx = 0
walk_last_idx = idx = N-2
current_data_idx = indices[N-2]
next_data_idx= indices[N-1]
```

So after the loop, we have these lines:
```c++
  if (isCluster(
          input, indices[walk_first_idx], indices[walk_last_idx],
          walk_last_idx - walk_first_idx + 1)) {
      for (int i = walk_first_idx; i <= walk_last_idx; i++) {
        auto output_ptr = reinterpret_cast<PointXYZI *>(&output.data[output_size]);
        *output_ptr = *reinterpret_cast<const PointXYZI *>(&input->data[indices[i]]);
        output_size += output.point_step;
      }
    }
  }
```
which basically copy points from `walk_first_idx = 0` to `walk_last_idx = N-2`, thus skipping the last point.

sensing/pointcloud_preprocessor/src/outlier_filter/ring_outlier_filter_nodelet.cpp
```diff
@@ -44,82 +44,116 @@ RingOutlierFilterComponent::RingOutlierFilterComponent(const rclcpp::NodeOptions
...
+          input, indices[walk_first_idx], indices[walk_last_idx],
+          walk_last_idx - walk_first_idx + 1)) {
+      for (int i = walk_first_idx; i <= walk_last_idx; i++) {
+        auto output_ptr = reinterpret_cast<PointXYZI *>(&output.data[output_size]);
+        *output_ptr = *reinterpret_cast<const PointXYZI *>(&input->data[indices[i]]);
+        output_size += output.point_step;
       }
     }
-    tmp_indices.clear();
   }
```
VRichardJP, line 142: Another approximation I have found with this implementation is that the ring end is not "connected" to its start.

For example, let's imagine this ring:
```
ring : xxxx..xxxxxxxxxxx..xxxx.xx
index: abcdefghijklmnopqrstuvwxyz
``` 
Where chains of `x` represent points from the same "walk", and `.` the noisy points ment to be filtered. The index alphabet is just for the explanation. So in the situation here, there are 3 rings:
1. from 'g' to 'q'
2. from 't' to 'w'
3. from 'y' to 'd' (wrapped around)

However in your implementation, we start at index 0 ('a') and the start/end of the ring are not considered connected. So you would find the rings:
1. from 'a' to 'd'
2. from 'g' to 'q'
3. from 't' to 'w'
The algorithm would consider the walk 'y' -> 'z', but as it is too short it would ignore the points.

With the default parameters `num_points_threshold_ = 4`, up to 3 points could be missing because of that. When using bigger `num_points_threshold_` value, the filter would sometimes produce some big blinspot in the data.

sensing/pointcloud_preprocessor/src/outlier_filter/ring_outlier_filter_nodelet.cpp
```diff
@@ -44,82 +44,116 @@ RingOutlierFilterComponent::RingOutlierFilterComponent(const rclcpp::NodeOptions
...
+
+  output.data.resize(output_size);
+
+  // Note that `input->header.frame_id` is data before converted when `transform_info.need_transform
+  // == true`
+  output.header.frame_id = !tf_input_frame_.empty() ? tf_input_frame_ : tf_input_orig_frame_;
+
+  output.height = 1;
+  output.fields = input->fields;
```
VRichardJP, line 151: Comming back to my previous comment about the `PointXYZI`data copy. Here we say input and output has the same fields, yet we don't copy "azimut" and "ring" fields from input.

VRichardJP (commented):
Sorry I'm annoying with all my comments ;-)

sensing/pointcloud_preprocessor/src/outlier_filter/ring_outlier_filter_nodelet.cpp
```diff
@@ -44,82 +44,116 @@ RingOutlierFilterComponent::RingOutlierFilterComponent(const rclcpp::NodeOptions
...
-      const auto & current_idx = ring_indices.second.at(idx);
-      const auto & next_idx = ring_indices.second.at(idx + 1);
-      tmp_indices.emplace_back(current_idx);
+  std::vector<std::vector<size_t>> ring2indices;
+  ring2indices.reserve(128);
+
+  for (size_t i = 0; i < 128; i++) {
+    ring2indices.push_back(std::vector<size_t>());
+    ring2indices.back().reserve(2000);  // TODO(sykwer): Make it parameter
+  }
```
VRichardJP, line 75: I am wondering if building `ring2indices` is necessary.
The whole algorithm assumes that ring points are ordered by azimut (modulo 36000). What about ring order? are the input points ordered by ring as well? (for example, all the points of ring 0, then all the points of ring 1, etc)
If not, it would not be necessary to build `ring2indices`.

sensing/pointcloud_preprocessor/src/outlier_filter/ring_outlier_filter_nodelet.cpp
VRichardJP: I think it would be important to print an error if the `ring` index is bigger than `max_channel` (and just return output = input?). Out-of-bound read/write are quite difficult to detect as they don't necessarily make the program crash when optimizations are used.

sensing/pointcloud_preprocessor/src/outlier_filter/ring_outlier_filter_nodelet.cpp
sykwer: At the moment, it is fine with no changes (cf. https://github.com/autowarefoundation/autoware.universe/blob/c7d84c64aadb6dc86a95920f392c5036225d76a0/common/autoware_point_types/include/autoware_point_types/types.hpp#L59-L63)

But we have a bug and are dealing with it in https://github.com/autowarefoundation/autoware.universe/pull/2618.

In the long-term perspective, we will make chagne according to [this design](https://github.com/autowarefoundation/autoware-documentation/blob/main/docs/design/autoware-architecture/sensing/data-types/point-cloud.md).

VRichardJP: I see. Thank you for the link. I find the manual 4-bytes padding quite strange... but that's another topic.
Still, I think this way of using `reinterpret_cast` is UB. For example, since the the compiler is not told not to pad the 2 structs with annotations such as `__attribute__((packed))`, the compiler is "free" to pad the structs as it sees fit.

sensing/pointcloud_preprocessor/src/outlier_filter/ring_outlier_filter_nodelet.cpp
sykwer: There is no guarantee that the point clouds are arranged in the order of `ring`.

sensing/pointcloud_preprocessor/src/outlier_filter/ring_outlier_filter_nodelet.cpp
sykwer: This number has been hard-coded for a long period of time and should, of course, be parameterised, but this is not necessary.

sensing/pointcloud_preprocessor/src/outlier_filter/ring_outlier_filter_nodelet.cpp
sykwer: It is a universal problem in Autoware and not related to this Pull Request, so please raise another Issue.

sensing/pointcloud_preprocessor/src/outlier_filter/ring_outlier_filter_nodelet.cpp
sykwer: same as below

sensing/pointcloud_preprocessor/src/outlier_filter/ring_outlier_filter_nodelet.cpp
sykwer: Since this problem has been around for a while and has nothing to do with this Pull Request, can you please create a new Issue?

sensing/pointcloud_preprocessor/include/pointcloud_preprocessor/outlier_filter/ring_outlier_filter_nodelet.hpp
```diff
@@ -45,18 +54,27 @@ class RingOutlierFilterComponent : public pointcloud_preprocessor::Filter
...
+    auto first_point = reinterpret_cast<const PointXYZI *>(&input->data[data_idx_both_ends.first]);
+    auto last_point = reinterpret_cast<const PointXYZI *>(&input->data[data_idx_both_ends.second]);
+
+    Eigen::Vector4f p1(first_point->x, first_point->y, first_point->z, 1);
+    Eigen::Vector4f p2(last_point->x, last_point->y, last_point->z, 1);
+
+    if (transform_info.need_transform) {
+      p1 = transform_info.eigen_transform * p1;
+      p2 = transform_info.eigen_transform * p2;
+    }
```
VRichardJP, line 71: I don't think the transformation is necessary here:
Distance between 2 points is unchanged by translation/rotation:
`|T*p2 - T*p1| = |p2 - p1|`

sensing/pointcloud_preprocessor/include/pointcloud_preprocessor/outlier_filter/ring_outlier_filter_nodelet.hpp
```diff
@@ -45,18 +54,27 @@ class RingOutlierFilterComponent : public pointcloud_preprocessor::Filter
...
+
+    auto x = p1[0] - p2[0];
+    auto y = p1[1] - p2[1];
+    auto z = p1[2] - p2[2];
+
+    return x * x + y * y + z * z >= object_length_threshold_ * object_length_threshold_;
```
VRichardJP, line 77: Further optimization idea:
Another way to compute the distance between the 2 points is to use the [law of cosines](https://en.wikipedia.org/wiki/Law_of_cosines): `dist2(p1,p2) = d1^2 + d2^2 - 2*d1*d2*cos(a)`. Where `d1` and `d2` are the `distance` attribute of `p1` and `p2`, and `a` the `azimut` diff between the 2 points.

What is interesting with this approach is that `a` is always small. Indeed, since `isCluster` has early returns when walk has more than `num_points_threshold_` (3 by default), it means the current walk has 3 points or less. But to be added to the walk, the azimut diff between 2 points must be [less than 1 degree](https://github.com/autowarefoundation/autoware.universe/blob/7213591028bde203b6587b428db51b2159de01ff/sensing/pointcloud_preprocessor/src/outlier_filter/ring_outlier_filter_nodelet.cpp#L115). So the `a` value is at most 3 degrees. Thus `cos(a) ~= 1`.

The distance calculation can be simplified to: `dist2(p1,p2) = (d1 - d2)^2`. 

-> we don't need to fetch the xyz points anymore, just the 2 `distance` attributes are necessary.

sykwer: This proposal changes the logical output by approximation. It is beyond the scope of this Pull Request to speed up the processing without changing the logical output. It would be appreciated if you could start a new issue for discussion and agreement.

sensing/pointcloud_preprocessor/include/pointcloud_preprocessor/outlier_filter/ring_outlier_filter_nodelet.hpp
sykwer: I made an ordinary mistake 💦  Thanks!

sensing/pointcloud_preprocessor/docs/ring-outlier-filter.md
```diff
@@ -22,11 +22,13 @@ This implementation inherits `pointcloud_preprocessor::Filter` class, please ref
...
-| `distance_ratio`          | double | 1.03          |             |
-| `object_length_threshold` | double | 0.1           |             |
-| `num_points_threshold`    | int    | 4             |             |
+| Name                      | Type    | Default Value | Description |
+| ------------------------- | ------- | ------------- | ----------- |
+| `distance_ratio`          | double  | 1.03          |             |
+| `object_length_threshold` | double  | 0.1           |             |
+| `num_points_threshold`    | int     | 4             |             |
+| `max_rings_num`           | uint_16 | 128           |             |
+| `max_points_num_per_ring` | size_t  | 2000          |             |
```
Shin-kyoto, line 31: How did you set the default value of `max_points_num_per_ring` as 2000 ?
Could I get the reason?

sensing/pointcloud_preprocessor/src/outlier_filter/ring_outlier_filter_nodelet.cpp
```diff
@@ -37,89 +37,150 @@ RingOutlierFilterComponent::RingOutlierFilterComponent(const rclcpp::NodeOptions
...
       }
     }
-    tmp_indices.clear();
   }
+
+  output.data.resize(output_size);
+
+  // Note that `input->header.frame_id` is data before converted when `transform_info.need_transform
+  // == true`
+  output.header.frame_id = !tf_input_frame_.empty() ? tf_input_frame_ : tf_input_orig_frame_;
```
miursh, line 173: Since this outlier filter algorithm works in "tf_input_frame_" coordinates, the output pointcloud at this point is in the same  "tf_input_frame_" coordinates.
```suggestion
  output.header.frame_id = tf_input_frame_;
```

sensing/pointcloud_preprocessor/src/outlier_filter/ring_outlier_filter_nodelet.cpp
```diff
@@ -37,89 +37,150 @@ RingOutlierFilterComponent::RingOutlierFilterComponent(const rclcpp::NodeOptions
...
+
+  output.fields.resize(4);  // x, y, z, intensity
+  std::copy(input->fields.begin(), input->fields.begin() + 4, output.fields.begin());
```
miursh, line 176: Could you avoid using magic number ` + 4` here?
`PointIndex::Intensity + 1` might be better?

sensing/pointcloud_preprocessor/docs/ring-outlier-filter.md
sykwer: It corresponds to figures that work well with the rosbag simulation in the AWF Tutorial. Are there any recommended default values? @Shin-kyoto

Shin-kyoto: @sykwer 
If there is no particular significance, I think it's best to set the value of `max_points_num_per_ring` according to the LiDAR. Let's set a value of 4000 as a large value. 
Also, please update the README and show that if this value is large enough such that FoV * resolution > `max_points_num_per_ring`, it is OK.

sykwer:
@Shin-kyoto 
I've confirmed by your tests that the number of points is unchanged before and after this PR.

Shin-kyoto (approved):
I have also confirmed that the number of points and the x,y,z value of each point is unchanged before and after this PR.
I checked whether the number of point clouds per topic and the values (x, y, z) of each point in the point clouds remained unchanged before and after the pull request.
The results are as follows:
- Number of point clouds per topic: No difference before and after the pull request.
- Values (x, y, z) of each point in the point clouds within the topic: No difference before and after the pull request for each of x, y, and z.

LGTM

miursh (approved):

miursh:
I just found a small bug in output pointcloud filed.
The output says the offset of intensity is **16** in the pointfiled even the point_step is 16, which causes the next point's x value becomes the previous point's intensity. 
```
height: 1
width: 192641
fields:
- name: x
  offset: 0
  datatype: 7
  count: 1
- name: y
  offset: 4
  datatype: 7
  count: 1
- name: z
  offset: 8
  datatype: 7
  count: 1
- name: intensity
  offset: 16
  datatype: 7
  count: 1
is_bigendian: false
point_step: 16
```

PR #3033, opened 2023-03-08
fix(avoidance): fix unexpected behavior in yield scenario

satoshi-ota:
## Description

In current logic, only when the front (nearest) target object should be avoid and the avoidance path is unsafe, the module executes YIELD maneuver. But, it should execute YIELD maneuver in following scenarios, too.

- Ego doesn't have to avoid the front object (passable without avoidance path).
- Ego has to avoid the behind object.
- Avoidance path is unsafe.

I found two bugs in the avoidance module:

1. The module never execute YILED maneuver if the front object is **NOT** `avoid_require = true`.
2. The module never cancel approved avoidance path if the front object is **NOT** `avoid_require = true`.

:arrow_down: Ego collides with NPC because of above bugs.

https://user-images.githubusercontent.com/44889564/223875419-f575247a-d020-40aa-843b-98c1dfb2d974.mp4

---

### Diff1

> The module never execute YILED maneuver if the front object is **NOT** `avoid_require = true`.

Search nearest object that should be avoid.

```c++
  for (const auto & o : data.target_objects) {
    if (o.avoid_required) {
      data.avoid_required = true;
      data.stop_target_object = o;
    }
  }
```

### Diff2

> The module never cancel approved avoidance path if the front object is **NOT** `avoid_require = true`.

Transit yield maneuver to cancel approved when the don't have to avoid.

```c++
  if (!data.safe && registered) {
    data.yield_required = true;
    data.candidate_path = toShiftedPath(data.reference_path);
    RCLCPP_WARN_THROTTLE(
      getLogger(), *clock_, 5000,
      "found safe avoidance path, but it is not safe. canceling avoidance path...");
  }
```

### Expected behavior

- Execute YIELD manauver even if the front object is **NOT** `avoid_require = true`.
- Keep enough distance in stopping behavior.

https://user-images.githubusercontent.com/44889564/223874980-660b1dac-f4cb-4e26-91f5-e519b62ed1b8.mp4

### DIFF
--- a/planning/behavior_path_planner/src/scene_module/avoidance/avoidance_module.cpp
+++ b/planning/behavior_path_planner/src/scene_module/avoidance/avoidance_module.cpp
@@ -752,6 +752,14 @@ void AvoidanceModule::fillShiftLine(AvoidancePlanningData & data, DebugData & de
       getLogger(), *clock_, 5000, "not found safe avoidance path. transit yield maneuver...");
   }
 
+  if (!data.safe && registered) {
+    data.yield_required = true;
+    data.candidate_path = toShiftedPath(data.reference_path);
+    RCLCPP_WARN_THROTTLE(
+      getLogger(), *clock_, 5000,
+      "found safe avoidance path, but it is not safe. canceling avoidance path...");
+  }
+
   /**
    * TODO(Satoshi OTA) Think yield maneuver in the middle of avoidance.
    * Even if it is determined that a yield is necessary, the yield maneuver is not executed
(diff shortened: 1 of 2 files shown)


satoshi-ota:
[[TIER IV INTERNAL LINK]](https://evaluation.tier4.jp/evaluation/reports/bd1f3f3c-2e4c-5b8c-a4aa-283610da6ecb/tests/4dfdc12c-2099-56f4-b71a-79d297f224b5?project_id=prd_jt) 352/362

**NOTE**: UC-A-04-00306-*, UC-A-04-00307-* is out of scope.

planning/behavior_path_planner/src/scene_module/avoidance/avoidance_module.cpp
```diff
@@ -736,9 +736,11 @@ void AvoidanceModule::fillShiftLine(AvoidancePlanningData & data, DebugData & de
    * Check whether the ego should avoid the front object. When the ego follows reference path,
```
TakaHoribe, line 736: Should this description be fixed as well?

satoshi-ota: Yes, I should. I'll fix in this PR.

satoshi-ota: I fixed in https://github.com/autowarefoundation/autoware.universe/pull/3033/commits/9b52f29a649d4bcc0441790990bbf10cd027c81d

planning/behavior_path_planner/src/scene_module/avoidance/avoidance_module.cpp
```diff
@@ -752,6 +754,18 @@ void AvoidanceModule::fillShiftLine(AvoidancePlanningData & data, DebugData & de
       getLogger(), *clock_, 5000, "not found safe avoidance path. transit yield maneuver...");
   }
 
+  /**
+   * Even if data.avoid_required is false, the module cancels registerd shift point when the
```
satoshi-ota, line 758:
```suggestion
   * Even if data.avoid_required is false, the module cancels registered shift point when the
```

PR #3038, opened 2023-03-09
feat(pose_initializer): enable pose initialization while running (only for sim)

kminoda:
## Description
Current issue: Pose initialization does not start when the vehicle is running, which is acceptable when running autoware.launch.xml, but is troublesome when using simulator.

Solution: Accept pose initialization while running when using logging_simulator.

- https://github.com/autowarefoundation/autoware_launch/pull/243

## Tests performed

Launch logging_simulator and planning_simulator and confirmed both works.
- logging simulator: localization works for sample rosbag
- planning simulator: the vehicle runs from start to goal position along the calculated path


taikitanaka3:
@YamatoAndo @KYabuuchi 
Could you review this PR?

KYabuuchi:
@taikitanaka3  Yes. I will review later.

launch/tier4_simulator_launch/launch/simulator.launch.xml
```diff
@@ -97,10 +98,7 @@
   <!-- Dummy localization -->
   <group if="$(var launch_dummy_localization)">
     <include file="$(find-pkg-share pose_initializer)/launch/pose_initializer.launch.xml">
-      <arg name="gnss_enabled" value="false"/>
-      <arg name="ndt_enabled" value="false"/>
-      <arg name="ekf_enabled" value="false"/>
-      <arg name="stop_check_enabled" value="false"/>
+      <arg name="config_file" value="$(var pose_initializer_param_simulator_path)"/>
```
KYabuuchi, line 101: In `autoware_launch/config/localization/pose_initializer.simulator.param.yaml`,     both of `ndt_enabled` and `ekf_enabled` are  true. 
However, neither EKF nor NDT will run in planning_simulation mode.

I guess this will NOT allow planning_simulator to complete the initial position estimation.

KYabuuchi (changes requested):

launch/tier4_simulator_launch/launch/simulator.launch.xml
kminoda: @KYabuuchi Thank you! Resolved here: https://github.com/autowarefoundation/autoware.universe/pull/3038/commits/ad3e61b6eab4567ef64efe1f3813818311c6d806
Also confirmed that both planning simulator and logging simulator works

VRichardJP:
I am curious about this change. I test Autoware a lot with the planning simulator and I don't have issue setting the initial pose while the vehicle is driving. However I remember it is the case for the logging simulator.

kminoda:
@VRichardJP This PR does not change anything about `planning_simulator`, only for logging_simulator (Anyway thanks for pointing it out, I modified the description of PR as it was kind of confusing)

launch/tier4_simulator_launch/launch/simulator.launch.xml
KYabuuchi: Great!

KYabuuchi:
I believe all YAML files must be placed in autoware_launch and in each packages.
The PR on the autoware_launch will add 3 new YAML files, but doesn't this PR add the same files?
(Please let me know if my point is incorrect. I am not familiar with how to manage config files)

kminoda:
@KYabuuchi Thanks for pointing it out! I think the package should only have the default parameter (=`planning_simulator.param.yaml`), so I think we don't need to add the other simulator parameter files in the `pose_initailizer/config` directory. But it's just my opinion and open to discussion if anyone else have any comments on this.

KYabuuchi (commented):
LGTM :+1:
(I'm not a code owner, though.)

KYabuuchi (approved):

launch/tier4_localization_launch/launch/localization.launch.xml
```diff
@@ -6,10 +6,13 @@
   <arg name="random_downsample_filter_param_path"/>
   <arg name="ndt_scan_matcher_param_path"/>
   <arg name="localization_error_monitor_param_path"/>
+  <arg name="pose_initializer_param_path"/>
+  <arg name="pose_initializer_param_logging_simulator_path"/>
```
isamu-takagi, line 10: Why are you passing two config file paths? It is unnecessary if you specify the config file path according to the system run mode.

launch/tier4_localization_launch/launch/util/util.launch.xml
```diff
@@ -13,11 +13,10 @@
...
   <!-- pose_initializer -->
   <group>
+    <let name="config_file" value="$(var pose_initializer_param_path)" if="$(eval &quot;'$(var run_mode)'=='online'&quot;)"/>
+    <let name="config_file" value="$(var pose_initializer_param_logging_simulator_path)" if="$(eval &quot;'$(var run_mode)'=='logging_simulation'&quot;)"/>
     <include file="$(find-pkg-share pose_initializer)/launch/pose_initializer.launch.xml">
-      <arg name="gnss_enabled" value="true"/>
-      <arg name="ndt_enabled" value="true"/>
-      <arg name="ekf_enabled" value="true"/>
-      <arg name="stop_check_enabled" value="true"/>
+      <arg name="config_file" value="$(var config_file)"/>
```
YamatoAndo, line 19: Is it possible to have the default settings loaded in a yaml file and then override the parameters?

like this.
```
<arg name="config_file" value="$(var pose_initializer_param_path)"/>
<arg name="stop_check_enabled" value="false" if="$(eval &quot;'$(var run_mode)'=='logging_simulation'&quot;)/>
```

If it's possible, I think this would be better.

YamatoAndo: @isamu-takagi What do you think?

isamu-takagi: @YamatoAndo 
I think it's a good solution for logging simulator problems. Ideally it would be nice to be able to set it in a parameter file from autoware_launch repository, but that can be done in a separate PR.

launch/tier4_localization_launch/launch/localization.launch.xml
kminoda: Thanks. There's no particular reason. Resolved: https://github.com/autowarefoundation/autoware.universe/pull/3038/commits/99b590e8aca2fb7dd109e94c26949e0f1d2bde88

launch/tier4_localization_launch/launch/util/util.launch.xml
YamatoAndo: @isamu-takagi Thank you for the reply.
But after talking with Minoda-san, we decided to follow Takagi-san's suggestion since we want to write everything in parameter files.

isamu-takagi (approved):
LGTM

YamatoAndo (approved):

kminoda:
@asana17 @ito-san @kenji-miyake Could you review this PR? (For `tier4_system_launch`)

kenji-miyake (approved):

kminoda:
@KeisukeShima Could you review this PR?

KeisukeShima (approved):

PR #3067, opened 2023-03-13
perf(behavior_path_palnner): improve performance of isPathInLanelets

TakanoTaiga:
Signed-off-by: Taiga Takano taiga.takano@tier4.jp

# Description
Changed the order in which covered_by functions are executed. This change improves performance without changing output.

This function is called in the getSafePath function, which is heavily used in the lane_change module and has become a bottleneck.

スクリーンショット 2023-03-13 19 31 43

### DIFF
--- a/planning/behavior_path_planner/src/util/lane_change/util.cpp
+++ b/planning/behavior_path_planner/src/util/lane_change/util.cpp
@@ -139,12 +139,14 @@ bool isPathInLanelets(
   const PathWithLaneId & path, const lanelet::ConstLanelets & original_lanelets,
   const lanelet::ConstLanelets & target_lanelets)
 {
-  const auto current_lane_poly_2d = lanelet::utils::to2D(
-    lanelet::utils::getPolygonFromArcLength(original_lanelets, 0, std::numeric_limits<double>::max())
-    ).basicPolygon();
-  const auto target_lane_poly_2d = lanelet::utils::to2D(
-    lanelet::utils::getPolygonFromArcLength(target_lanelets, 0, std::numeric_limits<double>::max())
-    ).basicPolygon();
+  const auto current_lane_poly_2d =
+    lanelet::utils::to2D(lanelet::utils::getPolygonFromArcLength(
+                           original_lanelets, 0, std::numeric_limits<double>::max()))
+      .basicPolygon();
+  const auto target_lane_poly_2d =
+    lanelet::utils::to2D(lanelet::utils::getPolygonFromArcLength(
+                           target_lanelets, 0, std::numeric_limits<double>::max()))
+      .basicPolygon();
 
   for (const auto & pt : path.points) {
     const lanelet::BasicPoint2d ll_pt(pt.point.pose.position.x, pt.point.pose.position.y);
--- a/planning/behavior_path_planner/src/util/lane_change/util.cpp
+++ b/planning/behavior_path_planner/src/util/lane_change/util.cpp
@@ -139,21 +139,19 @@ bool isPathInLanelets(
   const PathWithLaneId & path, const lanelet::ConstLanelets & original_lanelets,
   const lanelet::ConstLanelets & target_lanelets)
 {
-  const auto current_lane_poly = lanelet::utils::getPolygonFromArcLength(
-    original_lanelets, 0, std::numeric_limits<double>::max());
-  const auto target_lane_poly =
-    lanelet::utils::getPolygonFromArcLength(target_lanelets, 0, std::numeric_limits<double>::max());
-  const auto current_lane_poly_2d = lanelet::utils::to2D(current_lane_poly).basicPolygon();
-  const auto target_lane_poly_2d = lanelet::utils::to2D(target_lane_poly).basicPolygon();
+  const auto current_lane_poly_2d = lanelet::utils::to2D(
+    lanelet::utils::getPolygonFromArcLength(original_lanelets, 0, std::numeric_limits<double>::max())
+    ).basicPolygon();
+  const auto target_lane_poly_2d = lanelet::utils::to2D(
+    lanelet::utils::getPolygonFromArcLength(target_lanelets, 0, std::numeric_limits<double>::max())
+    ).basicPolygon();
+
   for (const auto & pt : path.points) {
     const lanelet::BasicPoint2d ll_pt(pt.point.pose.position.x, pt.point.pose.position.y);
-    const auto is_in_current = boost::geometry::covered_by(ll_pt, current_lane_poly_2d);
-    if (is_in_current) {
-      continue;
-    }
-    const auto is_in_target = boost::geometry::covered_by(ll_pt, target_lane_poly_2d);
-    if (!is_in_target) {
-      return false;
+    if (!boost::geometry::covered_by(ll_pt, target_lane_poly_2d)) {
+      if (!boost::geometry::covered_by(ll_pt, current_lane_poly_2d)) {
+        return false;
+      }
     }
   }
   return true;


VRichardJP:
Not that I doubt the numbers, but I don't see how this change could improve performance.

Writing:
```C++
  const auto current_lane_poly_2d = lanelet::utils::to2D(
    lanelet::utils::getPolygonFromArcLength(original_lanelets, 0, std::numeric_limits<double>::max())).basicPolygon();
  const auto target_lane_poly_2d = lanelet::utils::to2D(
    lanelet::utils::getPolygonFromArcLength(target_lanelets, 0, std::numeric_limits<double>::max())).basicPolygon();
```

Is essentially the same as:
```C++
  // 1
  const auto current_lane_poly = lanelet::utils::getPolygonFromArcLength(original_lanelets, 0, std::numeric_limits<double>::max());
  // 2
  const auto current_lane_poly_2d = lanelet::utils::to2D(current_lane_poly).basicPolygon();
  // 3
  const auto target_lane_poly = lanelet::utils::getPolygonFromArcLength(target_lanelets, 0, std::numeric_limits<double>::max());
  // 4
  const auto target_lane_poly_2d = lanelet::utils::to2D(target_lane_poly).basicPolygon();
```

So it is the same than the original code, except line 2 and 3 are exchanged. But that should not have any impact since the compiler is free to "reorganize" the code as it sees fit when optimisations are turned on. It is the same for the two `const bool` that are not assigned later.

I am wondering if the performance gain you measured does not come from something else, such as: cpu load at the time of the test, lower memory fragmentation or just the linker was kind with you this time but next time you make a change somewhere the performance gain will disappear, etc.

TakanoTaiga:
@VRichardJP 

Certainly, that change is not necessary.
The change in the order of execution of the coverd_by function contributes to the performance improvement shown in this PR.

VRichardJP:
Sorry, I still don't buy it. I reproduced and tested the change on my machine.
First I pulled the latest main. Then, to measure the performance, I have prepared this simple setup:
```diff
diff --git a/launch/tier4_planning_launch/launch/scenario_planning/lane_driving/behavior_planning/behavior_planning.launch.py b/launch/tier4_planning_launch/launch/scenario_planning/lane_driving/behavior_planning/behavior_planning.launch.py
index 4616590fac..0c517e8587 100644
--- a/launch/tier4_planning_launch/launch/scenario_planning/lane_driving/behavior_planning/behavior_planning.launch.py
+++ b/launch/tier4_planning_launch/launch/scenario_planning/lane_driving/behavior_planning/behavior_planning.launch.py
@@ -229,6 +229,7 @@ def launch_setup(context, *args, **kwargs):
             behavior_velocity_planner_component,
         ],
         output="screen",
+        prefix="gnome-terminal --",
     )
 
     # This condition is true if run_out module is enabled and its detection method is Points
diff --git a/planning/behavior_path_planner/src/util/lane_change/util.cpp b/planning/behavior_path_planner/src/util/lane_change/util.cpp
index 3bd7549bf3..0c2db75809 100644
--- a/planning/behavior_path_planner/src/util/lane_change/util.cpp
+++ b/planning/behavior_path_planner/src/util/lane_change/util.cpp
@@ -31,6 +31,7 @@
 #include <tf2_ros/transform_listener.h>
 
 #include <algorithm>
+#include <chrono>
 #include <limits>
 #include <memory>
 #include <string>
@@ -257,9 +258,16 @@ std::optional<LaneChangePath> constructCandidatePath(
     point.lane_ids = lane_changing_segment.points.at(*nearest_idx).lane_ids;
   }
 
+  auto start = std::chrono::high_resolution_clock::now();
   if (!isPathInLanelets(shifted_path.path, original_lanelets, target_lanelets)) {
+    auto end = std::chrono::high_resolution_clock::now();
+    auto duration = std::chrono::duration_cast<std::chrono::nanoseconds>(end-start);
+    std::cout << "no  " << duration.count() << std::endl;
     return std::nullopt;
   }
+  auto end = std::chrono::high_resolution_clock::now();
+  auto duration = std::chrono::duration_cast<std::chrono::nanoseconds>(end-start);
+  std::cout << "yes " << duration.count() << std::endl;
 
   // check candidate path is in lanelet
   candidate_path.path = combineReferencePath(prepare_segment, shifted_path.path);
```

It looks like the main branch is broken at the moment, so I also had to modify `autoware_launch` aswell:
```diff
diff --git a/autoware_launch/config/planning/scenario_planning/lane_driving/behavior_planning/behavior_path_planner/behavior_path_planner.param.yaml b/autoware_launch/config/planning/scenario_planning/lane_driving/behavior_planning/behavior_path_planner/behavior_path_planner.param.yaml
index fb1a632..3fd8c6a 100644
--- a/autoware_launch/config/planning/scenario_planning/lane_driving/behavior_planning/behavior_path_planner/behavior_path_planner.param.yaml
+++ b/autoware_launch/config/planning/scenario_planning/lane_driving/behavior_planning/behavior_path_planner/behavior_path_planner.param.yaml
@@ -52,6 +52,13 @@
       priority: 6
       max_module_size: 1
 
+    # tmp
+    lane_change:
+      enable_module: true
+      enable_simultaneous_execution: false
+      priority: 5
+      max_module_size: 1
+
     lane_change_left:
       enable_module: true
       enable_simultaneous_execution: false
```

I compiled in Release mode, then ran the planning simulator on map with a few lane changes:
```
$ ros2 launch autoware_launch planning_simulator.launch.xml map_path:=$MAP_PATH vehicle_model:=sample_vehicle sensor_model:=sample_sensor_kit
```

I set the initial pose, the goal pose, ran a couple simulation with/without lane change then stopped. I then applied your changes on my code base. And tested the same setup again. Here is what I get:

image

Significant performance gain indeed.

But still, I understand where the performance gain comes from.

So I split your changes in 2:
- the "reorder" part:
```c++
  const auto current_lane_poly_2d =
    lanelet::utils::to2D(lanelet::utils::getPolygonFromArcLength(
                           original_lanelets, 0, std::numeric_limits<double>::max()))
      .basicPolygon();
  const auto target_lane_poly_2d =
    lanelet::utils::to2D(lanelet::utils::getPolygonFromArcLength(
                           target_lanelets, 0, std::numeric_limits<double>::max()))
      .basicPolygon();
```
- the "if" part:
```c++
    if (!boost::geometry::covered_by(ll_pt, target_lane_poly_2d)) {
      if (!boost::geometry::covered_by(ll_pt, current_lane_poly_2d)) {
        return false;
      }
    }
```

Then tested again.
This is what I get now:

image

It shows a few interesting things:
1. The first change (the reorder) has no significant impact on the performance.
2. The second change (the if) has a big impact on the performance.
3. But the if change alone has smaller impact than the reorder+if

This could be just the way I have tested, the compiler triggering some optimization, the linker being kind with me etc. I don't say the PR should not be merged, but the performance gain origin is too... mystical. Typically, one day someone will make an innocent change somewhere else and the perf gain will disappear.

For example: I have slightly modified your PR, I have changed the if part into:
```C++
  for (const auto & pt : path.points) {
    const lanelet::BasicPoint2d ll_pt(pt.point.pose.position.x, pt.point.pose.position.y);
    if (!boost::geometry::covered_by(ll_pt, target_lane_poly_2d) && !boost::geometry::covered_by(ll_pt, current_lane_poly_2d)) {
      return false;
    }
  }
```
Instead of a double `if`, I have combined both clause with a `&&`. Since the right part of the `&&` is only evaluated if the left part is `true`, my change is exactly the same than your PR. 

This is what I get:
image

As you can see, your PR is quite faster than mine. Sometimes it is processed 2 times faster!
It is the same though.

TL;DR: I am not convinced that PR brings any permanent performance gain. Today it certainly does, what about tomorrow?

TakanoTaiga:
@VRichardJP 
Thank you for your detailed verification and code review. The information you sent us was very interesting.

TakanoTaiga:
The focus of this change is to investigate the trend of is_in_current and is_in_target states, and found that is_in_current is rarely true. Therefore, we believe that changing the order of the if statements will reduce the number of covered_by function executions.

However, to verify whether this trend applies to all cases, it is necessary to verify using a large number of rosbags instead of a single rosbag. Therefore, we would like to ask for merge after the verification is completed.

VRichardJP:
@TakanoTaiga Oh! I totally missed the test order is changed in the for loop.

Indeed, to test whether the current lanelet contains a point, and if yes whether the target lanelet contains the same point, is not exactly the same than the other way around. From a logical point of view it is the same, but from a performance point of view, it may differ if one of the statement is false more often than the other.

If path points are almost always in the current lanelets, a lot of "meaningless" tests can be avoided by checking the target lane first. This can definitely explain the performance gain.

On a side note, I have observed during my tests that `isPathInLanelets` always returned `true`. In my setup, I print "yes" or "no" depending on `isPathInLanelets` output, and I have only ever observed "yes". Maybe it is just related to the way I tested, but could it be possible that by construction, the path is always found in either the current or target lanelets?

zulfaqar-azmi-t4:
@VRichardJP 
the aim of the function is to avoid lane change path's point from being generated outside of either current or target path. Given the current available conditional statement in the `isPathInLanelets`, this will probably be true in all of the cases. However, we worried that we might missed out on the edge cases as well, therefore the check is added.

PR #3110, opened 2023-03-17
fix(avoidance): not run avoidance module when the shift line is empty

satoshi-ota:
## Description

The avoidance module check adjacent lane to find target object that should be avoid. On the other hand, the ego often don't have to shift even if the target object is not empty because there is enough lateral margin between the ego and that object.

In this PR, I added a new condition to `isExecutionRequested()`, and prevent module execution in scenario where the ego don't have to shift anymore.

NOTE: This modification only works on the new manager for now.


TakaHoribe (approved):
LGTM. 

Note: Implementation between architecture options should be as close as possible. This does not cause big changes.

PR #3125, opened 2023-03-21
feat(avoidance_by_lc): add new module to avoid obstacle by lane change

satoshi-ota:
## Description

:arrow_down: These PRs must be merged before this PR.
https://github.com/tier4/tier4_autoware_msgs/pull/71
https://github.com/autowarefoundation/autoware_launch/pull/261

Add new module `avoidance_by_lc`.
This module execute lane change maneuver when there are obstacles on ego's driving lane.

The obstacle filtering logic and path generation logic are almost the same to the `avoidance` and `lane_change` module. 
(TODO: remove duplicated functions.)

image

## Params

```yaml
/**:
  ros__parameters:
    avoidance_by_lane_change:
      execute_object_num: 1               
      execute_object_longitudinal_margin: 0.0
      execute_only_when_lane_change_finish_before_object: true
```

- `execute_object_num`

This module execute lane change maneuver only when the target number is larger than `execute_object_num`. (If the condition is **NOT** satisfied, the nomal `avoidance` module will be running.)

- `execute_object_longitudinal_margin`

This module execute lane change maneuver only when the distance between the ego and object is larger than `execute_object_longitudinal_margin`. (If the condition is **NOT** satisfied, the nomal `avoidance` module will be running.)

- `execute_only_when_lane_change_finish_before_object`

If it set this param as `true`,  this module will be canceled when the following conditions are satisfied.

1. The lane change end point is behind of the target object.
2. The path change is **NOT** approved.

`execute_only_when_lane_change_finish_before_object: true`

- **Pros**: since the lane change manauver is finished before the object, the path safety is almost guaranteed.
- **Cons**: depending the range of object recognition, it maybe **NOT** possible to execute lane change maneuver.  The wider the detection range is, the longer the grace period to execute lane change maneuver is.

https://user-images.githubusercontent.com/44889564/226545618-dec8233f-9a4c-4fa0-be48-7589eb8f34fa.mp4

---

`execute_only_when_lane_change_finish_before_object: false`

- **Pros**: can execute lane change at any timing.
- **Cons**: maybe output unsafe lane change path.

image

**NOTE: This module also use `avoidance` and `lane_change` module params.**

## Tests performed

Please set following flag to `FALSE`, and confirm to pass all build.

https://github.com/autowarefoundation/autoware.universe/blob/ad5595389a629346476ee06391260ffea53159ef/planning/behavior_path_planner/CMakeLists.txt#L10

Use following PRs' branch.
https://github.com/tier4/tier4_autoware_msgs/pull/71
https://github.com/autowarefoundation/autoware_launch/pull/261

Set `enable_module` as true.

```yaml
...
    avoidance_by_lc:
      enable_module: true
      enable_simultaneous_execution: false
      priority: 3
      max_module_size: 1
```

https://user-images.githubusercontent.com/44889564/226542033-3432b79e-7355-4ab3-8b7a-b169335cf2d9.mp4

### DIFF
--- a/planning/behavior_path_planner/src/scene_module/avoidance_by_lc/module.cpp
+++ b/planning/behavior_path_planner/src/scene_module/avoidance_by_lc/module.cpp
@@ -89,7 +89,7 @@ bool AvoidanceByLCModule::isExecutionRequested() const
   const auto [found_valid_path, found_safe_path] =
     getSafePath(lane_change_lanes, check_distance_, selected_path);
 
-  if (!found_safe_path) {
+  if (!found_valid_path) {
     return false;
   }
 
--- a/planning/behavior_path_planner/src/scene_module/avoidance_by_lc/module.cpp
+++ b/planning/behavior_path_planner/src/scene_module/avoidance_by_lc/module.cpp
@@ -49,38 +49,26 @@ using tier4_autoware_utils::calcLateralDeviation;
 
 AvoidanceByLCModule::AvoidanceByLCModule(
   const std::string & name, rclcpp::Node & node,
-  std::shared_ptr<AvoidanceByLCParameters> parameters,
-  std::shared_ptr<RTCInterface> & rtc_interface_left,
-  std::shared_ptr<RTCInterface> & rtc_interface_right)
-: SceneModuleInterface{name, node},
-  parameters_{parameters},
-  rtc_interface_left_{rtc_interface_left},
-  rtc_interface_right_{rtc_interface_right},
-  uuid_left_{generateUUID()},
-  uuid_right_{generateUUID()}
+  const std::shared_ptr<AvoidanceByLCParameters> parameters,
+  const std::unordered_map<std::string, std::shared_ptr<RTCInterface>> & rtc_interface_ptr_map)
+: SceneModuleInterface{name, node, rtc_interface_ptr_map}, parameters_{parameters}
 {
   steering_factor_interface_ptr_ =
     std::make_unique<SteeringFactorInterface>(&node, "avoidance_by_lane_change");
 }
 
-void AvoidanceByLCModule::onEntry()
+void AvoidanceByLCModule::processOnEntry()
 {
-  RCLCPP_DEBUG(getLogger(), "AVOIDANCE_BY_LC onEntry");
-#ifdef USE_OLD_ARCHITECTURE
-  current_state_ = ModuleStatus::SUCCESS;
-#else
-  current_state_ = ModuleStatus::IDLE;
+#ifndef USE_OLD_ARCHITECTURE
   waitApproval();
 #endif
   current_lane_change_state_ = LaneChangeStates::Normal;
   updateLaneChangeStatus();
 }
 
-void AvoidanceByLCModule::onExit()
+void AvoidanceByLCModule::processOnExit()
 {
   resetParameters();
-  current_state_ = ModuleStatus::SUCCESS;
-  RCLCPP_DEBUG(getLogger(), "AVOIDANCE_BY_LC onExit");
 }
 
 bool AvoidanceByLCModule::isExecutionRequested() const
@@ -731,10 +719,10 @@ void AvoidanceByLCModule::resetPathIfAbort()
     const auto lateral_shift = lane_change_utils::getLateralShift(*abort_path_);
     if (lateral_shift > 0.0) {
       removePreviousRTCStatusRight();
-      uuid_right_ = generateUUID();
+      uuid_map_.at("right") = generateUUID();
     } else if (lateral_shift < 0.0) {
       removePreviousRTCStatusLeft();
-      uuid_left_ = generateUUID();
+      uuid_map_.at("left") = generateUUID();
     }
 #else
     removeRTCStatus();
@@ -1036,9 +1024,15 @@ std::pair<bool, bool> AvoidanceByLCModule::getSafePath(
   return {found_valid_path, found_safe_path};
 }
 
-bool AvoidanceByLCModule::isSafe() const { return status_.is_safe; }
+bool AvoidanceByLCModule::isSafe() const
+{
+  return status_.is_safe;
+}
 
-bool AvoidanceByLCModule::isValidPath() const { return status_.is_valid_path; }
+bool AvoidanceByLCModule::isValidPath() const
+{
+  return status_.is_valid_path;
+}
 
 bool AvoidanceByLCModule::isValidPath(const PathWithLaneId & path) const
 {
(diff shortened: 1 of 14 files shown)


VRichardJP:
I am not familiar with the way behavior modules interact with each other with the new design. 
- Will this module have priority over the normal avoidance and lane change modules? In other words, are the normal avoidance/lane change modules skipped when this one can operate the lane change?
- With this module, would it make sense to disable normal obstacle avoidance if the road has alwways more than 2 driveable lanes? 
- Does this module consider the required distance to change back to the initial lane? For example in this situation, there would not be enough distance left to come back to the initial lane so the normal avoidance should be used:
image

satoshi-ota:
Hi @VRichardJP Thanks for your comments :+1: 
I'll answer your questions.

> Q1. Will this module have priority over the normal avoidance and lane change modules? In other words, are the normal avoidance/lane change modules skipped when this one can operate the lane change?

Yes. I would like to set higher priority to this module than nomal avoidance and lane change module. So, as you said, these modules will be skipped in the following example scenario.

- lane changeable.
- want to execute lane change to right lane in order to arrive the goal.
- vehicle has to avoid object to right side.

-> This module will be executed, and  the vehicle will move to right side lane. As a result, nomal avoidance and normal lane change module executions will be no longer needed.

> Q2. With this module, would it make sense to disable normal obstacle avoidance if the road has alwways more than 2 driveable lanes?

It is not assumed that normal avoidance will not be used at all even if we enable this module in that scenario. I suppose there are some scenarios that are better handled by normal avoidance. For example, 

- single vehicle avoidance -> normal avoidance
- multiple vehicle avoidance (such as following scene.) -> this module

image

> Q3. Does this module consider the required distance to change back to the initial lane? For example in this situation, there would not be enough distance left to come back to the initial lane so the normal avoidance should be used:

This module checks the required distance to back to the original lane at least in following function, but the function doesn't consider the obsrtacle and It does not check to see if there is enough distance left to return to the original lane after completely avoiding the obstacle. I understand this issue should be resolved :+1: 

https://github.com/autowarefoundation/autoware.universe/blob/36bdca4aa41ad910235cc465501b410a01b91f6f/planning/behavior_path_planner/src/util/lane_change/util.cpp#L448-L486

planning/behavior_path_planner/src/scene_module/avoidance_by_lc/module.cpp
```diff
@@ -0,0 +1,1380 @@
...
+    util::getCurrentLanesFromPath(*getPreviousModuleOutput().reference_path, planner_data_);
+#endif
+  const auto lane_change_lanes = getLaneChangeLanes(current_lanes, lane_change_lane_length_);
+
+  LaneChangePath selected_path;
+  const auto [found_valid_path, found_safe_path] =
+    getSafePath(lane_change_lanes, check_distance_, selected_path);
+
+  return found_valid_path;
+}
```
zulfaqar-azmi-t4, line 138: in lane change module, `isExecutionReady()` depends on `found_safe_path`.
May I know why `AvoidanceByLC` checks for `found_valid_path` instead?

satoshi-ota: Maybe that's my mistake. It is okay with `found_safe_path`.

planning/behavior_path_planner/src/scene_module/avoidance_by_lc/module.cpp
```diff
@@ -0,0 +1,1380 @@
...
+  const auto current_lanes =
+    util::getCurrentLanesFromPath(*getPreviousModuleOutput().reference_path, planner_data_);
+#endif
+  const auto lane_change_lanes = getLaneChangeLanes(current_lanes, lane_change_lane_length_);
+
+  LaneChangePath selected_path;
+  const auto [found_valid_path, found_safe_path] =
+    getSafePath(lane_change_lanes, check_distance_, selected_path);
+
+  return found_valid_path;
```
satoshi-ota, line 137:
```suggestion
  return found_safe_path;
```

planning/behavior_path_planner/src/scene_module/avoidance_by_lc/module.cpp
```diff
@@ -0,0 +1,1380 @@
...
+    util::getCurrentLanesFromPath(*getPreviousModuleOutput().reference_path, planner_data_);
+#endif
+  const auto lane_change_lanes = getLaneChangeLanes(current_lanes, lane_change_lane_length_);
+
+  LaneChangePath selected_path;
+  const auto [found_valid_path, found_safe_path] =
+    getSafePath(lane_change_lanes, check_distance_, selected_path);
+
+  if (!found_safe_path) {
+    return false;
```
zulfaqar-azmi-t4, line 93: Because lane change checked for valid path and not safe path during `isExecutionRequested`, we don't concern with the path is safe or not.

How about in Avoidance by Lane change, is it by intention to choose `found_safe_path`?
(i guess it makes sense to avoid only when safe path is found)

zulfaqar-azmi-t4:
@satoshi-ota 
I might be mistaken but the obstacle registration will be performed in both AvoidLC and Avoidance. Is there any consideration in the future to unify the obstacle related process?

zulfaqar-azmi-t4:
## Single AvoidLC
https://user-images.githubusercontent.com/93502286/230019062-f8bfeeef-c6d9-4ecd-a21c-789474086f77.mp4


## Double AvoidLC
https://user-images.githubusercontent.com/93502286/230018731-26811915-7167-45e7-831f-0b0c8dfdf69b.mp4

planning/behavior_path_planner/src/scene_module/avoidance_by_lc/module.cpp
satoshi-ota: oh, it is also my mistake.

planning/behavior_path_planner/src/scene_module/avoidance_by_lc/module.cpp
```diff
@@ -0,0 +1,1380 @@
...
+  const auto current_lanes =
+    util::getCurrentLanesFromPath(*getPreviousModuleOutput().reference_path, planner_data_);
+#endif
+  const auto lane_change_lanes = getLaneChangeLanes(current_lanes, lane_change_lane_length_);
+
+  LaneChangePath selected_path;
+  const auto [found_valid_path, found_safe_path] =
+    getSafePath(lane_change_lanes, check_distance_, selected_path);
+
+  if (!found_safe_path) {
```
satoshi-ota, line 92:
```suggestion
  if (!found_valid_path) {
```

planning/behavior_path_planner/src/scene_module/avoidance_by_lc/module.cpp
satoshi-ota: I fixed in https://github.com/autowarefoundation/autoware.universe/pull/3125/commits/fb573208e1ff9dbe77726f59da5cf82996c89fc7

satoshi-ota:
> @satoshi-ota I might be mistaken but the obstacle registration will be performed in both AvoidLC and Avoidance. Is there any consideration in the future to unify the obstacle related process?

@zulfaqar-azmi-t4 Thanks for comment. You're right.
Unfortunately, there is no concrete future plan to unify these two module. But there are duplicate process to filter avoidance target in this module and normal avoidance module, and I think it is not so good. So, I wanna fix it if I have time :pleading_face:

zulfaqar-azmi-t4 (approved):
LGTM!

rej55 (approved):
LGTM

PR #3139, opened 2023-03-22
fix: multiple markers in markerarray msg visualization

### LINKED ISSUE #3138: Multiple markers in the same MarkerArray message in VirtualWall in RVIZ
### Checklist

### Description

Using the planning simulator and visualizing virtual walls, there are always error due multiple markers with same id and ns in the same MarkerArray message.

### Expected behavior

Status under each VirtualWall to be OK (without errors)

### Actual behavior

Status under VirtualWall always includes error

Screenshot_20230322_144014

Screenshot_20230322_145007

### Steps to reproduce

1. Open planning simulator
2. Plan a scenario with adding a `2D Pose Estimate` and `2D Goal Estimate` , taking into consideration to have any on the behavior planning modules activated (one at lease or multiples) in this scenario. For example, to have `intersection` or `stop_line` or `traffic_light` , ... etc
3. Check VirtualWall status under `planning -> ScenarioPlanning --> LaneDriving --> BehaviorPlanning --> VirutalWall`

This [video](https://youtu.be/O_8vkA155_4) shows example reproduction of the bug.

### Versions

_No response_

#### Symptom Analysis
- The marker array that is publish to the topic `/planning/scenario_planning/lane_driving/behavior_planning/behavior_velocity_planner/virtual_wall/{module_name}` should include 2 entries .. 
  - one for `virtual_wall` that describes its `frame_id`, `pose`, `color`, **`marker_type` which is `ADD` in this case**.. etc
  - other one for `factor_text` that visualized the `module_name` for this `virtual_wall`.

- The problem happens because the marker array includes **4 entries** usually (not 2)
  -  2 mentioned above
  - another 2 with same id and ns but with zero pose, and **more interestingly with marker action `DELETE` and empty `frame_id`**

#### Root Cause Analysis
The flow of code goes as follows : 
- Inside scene module --> `debug.cpp` --> `createVirtualWallMarkerArray()` method, the `createStopVirtualWallMarker` is called which is defined in `common` --> `motion_utils`

 - The markers with zero poses and `DELETE` action are appended to the marker array using a method `createDeletedStopVirtualWallMarker` in `marker_helper.cpp` which uses `createDefaultMarker` in `marker_helper.hpp`
 
 - The `createDeletedStopVirtualWallMarker` is called by `handleVirtualWallMarker` which wants to delete duplicated poses based on the flag `previous_stop_pose_is_in_stop_pose`.

- :point_right:  It looks like this usage of this flag here is problematic or the method that sets this flag `resample_utils::validate_points_duplication` is problematic

- :point_right: Other possible cause, it could be a problem from rviz2 (similar to [this issue in rviz1](https://github.com/ros-visualization/rviz/pull/874)) that markers with `DELETE` action and without `frame_id` are not deleted !!!

### Additional context
Any module other than scene modules under `behavior_velocity_planner` that uses visualization with same approach will have similar issue.

ahmeddesokyebrahim:
## Description
Fixes #3138 

## Related links
In the [issue](https://github.com/autowarefoundation/autoware.universe/issues/3138), you can find steps for reproducing the issue and a demo for that.

More details about the analysis and other possible causes can be found there as well.

## Tests performed
After applying the potential fix, the output of testing for same scenario in the issue in this [video](https://youtu.be/IsVp21Yh3As) or [this video](https://youtu.be/FDrWwI8DAgI) for better zoom in for stop line as example

### DIFF
--- a/common/motion_utils/include/motion_utils/marker/marker_helper.hpp
+++ b/common/motion_utils/include/motion_utils/marker/marker_helper.hpp
@@ -79,7 +79,6 @@ private:
   visualization_msgs::msg::MarkerArray handleVirtualWallMarker(
     const std::vector<Pose> & poses, const std::string & module_name, const rclcpp::Time & now,
     int32_t id, create_wall_function function_create_wall_marker,
-    delete_wall_function function_delete_wall_marker,
     std::vector<geometry_msgs::msg::Pose> & previous_poses, const double longitudinal_offset = 0.0);
 
   std::vector<Pose> previous_stop_poses_;
--- a/common/motion_utils/src/marker/marker_helper.cpp
+++ b/common/motion_utils/src/marker/marker_helper.cpp
@@ -135,31 +135,14 @@ visualization_msgs::msg::MarkerArray createDeletedDeadLineVirtualWallMarker(
 visualization_msgs::msg::MarkerArray VirtualWallMarkerCreator::handleVirtualWallMarker(
   const std::vector<Pose> & poses, const std::string & module_name, const rclcpp::Time & now,
   int32_t id, create_wall_function function_create_wall_marker,
-  delete_wall_function function_delete_wall_marker,
   std::vector<geometry_msgs::msg::Pose> & previous_virtual_wall_poses,
   const double longitudinal_offset)
 {
   size_t id_to_create = id;
-  size_t id_to_delete = id;
   visualization_msgs::msg::MarkerArray wall_marker;
 
   if (poses.size() == 0) {
     return wall_marker;
-  }
-
-  for (const auto & p : previous_virtual_wall_poses) {
-    const bool previous_stop_pose_is_in_stop_pose =
-      std::any_of(poses.begin(), poses.end(), [&](const geometry_msgs::msg::Pose & elem) {
-        std::vector<Pose> poses;
-        poses.push_back(p);
-        poses.push_back(elem);
-        return resample_utils::validate_points_duplication(poses);
-      });
-
-    if (previous_stop_pose_is_in_stop_pose) {
-      appendMarkerArray(function_delete_wall_marker(now, id_to_delete), &wall_marker, now);
-    }
-    id_to_delete++;
   }
 
   for (const auto & p : poses) {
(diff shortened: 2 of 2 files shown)


kyoichi-sugahara:
@ahmeddesokyebrahim 
I appreciate the proposed changes made in this PR. 
However, without a deleter function, there may be situations where virtual walls continue to exist even when they should not. 

Therefore, I believe it is necessary to include an explicit deletion function to ensure proper removal of virtual walls. For this reason, I would like to keep the deleter function in place to address the bug that was pointed out. 

What are your thoughts on this?

ahmeddesokyebrahim:
>I appreciate the proposed changes made in this PR.
However, without a deleter function, there may be situations where virtual walls continue to exist even when they should not.
Therefore, I believe it is necessary to include an explicit deletion function to ensure proper removal of virtual walls. For this reason, I would like to keep the deleter function in place to address the bug that was pointed out.
What are your thoughts on this?

Thanks for your comment Mr @kyoichi-sugahara .

Actually I wonder about the significance of having a deleter function in the time that we add the marker in the creator function we specify a lifetime to it.
Based on the [Marker message definition](https://docs.ros.org/en/noetic/api/visualization_msgs/html/msg/Marker.html), any marker with lifetime (other than zero) will be deleted automatically. And that is already happening in `createDefaultMarker` method. That give me the understanding that we do not need a deleter function.

Let me know your thoughts.

kyoichi-sugahara:
@ahmeddesokyebrahim 
The reason why I think using the delete function is necessary is because when the wall marker is deleted by setting the lifetime, for example, to 0.2 seconds and each module runs at 10Hz, the wall marker will be displayed unnecessarily for 0.1 seconds after it is no longer needed. Although it may seem like a short time when considering real time, when running a heavy simulator at a slower speed, for example, at 1/10 of the normal speed, the marker will be displayed for 1 second despite being unnecessary, which can confuse developers.

For these reasons, I believe that an explicit deletion function is necessary when modules determine that the marker is no longer needed.

Therefore, I would like to proceed by fixing the bug while keeping the delete function for deleting unnecessary markers. What do you think about this approach?

ahmeddesokyebrahim:
Thanks for the explanation Mr. @kyoichi-sugahara . It makes sense for me. 

I just would like to have more elaboration about what you mean by `explicit deletion function`.

xmfcx:
@kyoichi-sugahara could you provide a solution approach to handle this without removing the deleter function?

maxime-clem:
@ahmeddesokyebrahim , your PR is good but removes the "explicit deletion" of the virtual wall markers, i.e., the virtual walls are not deleted by message with action `DELETE` and only disappear when reaching the end of their lifetime.

I think a better approach to fix the issue with multiple markers is to improve the way the `DELETE` markers are generated.
The current issue, as you found, is that we both create `ADD` markers and `DELETE` markers with the same `id` and `ns`.
I made a draft PR with a possible solution: https://github.com/autowarefoundation/autoware.universe/issues/3138

mitsudome-r:
This PR is now moved to https://github.com/autowarefoundation/autoware.universe/pull/3273/files.

PR #3176, opened 2023-03-27
feat(motion_utils): add cropPoints function

takayuki5168:
## Description

add cropPoints and its related functions.

image


takayuki5168:
@purewater0901 Could you review the PR?

VRichardJP:
Sadly I lost my comment and I am a bit lazy to retype everything.

Basically, I wanted to point out that `target_pos` and `target_seg_idx` have overlapping semantics.
There is no particular reason for the `crop*Points()` function to require both. It would be possible to write a very clean version of `crop*Points()` with just `target_seg_idx`. For example, the current behavior could be re-written as:

```C++
double longitudinal_offset = motion_utils::calcLongitudinalOffsetToSegment(points, target_seg_idx, target_pos);
auto cropped = cropPoints(points, target_seg_idx, forward_length + longitudinal_offset, backward_length - longitudinal_offset);
```

The advantage being that the `crop*Points()` only do the cropping, and don't have to care with some strange projection logic.

In particular, when you look at how the `crop*Points()` functions are used currently, both the caller and the `cropPoints()` functions handles "half" the length computation (which makes little sense). For example in `mpt_optimizer.cpp`:
```C++
  size_t ego_seg_idx =
    trajectory_utils::findEgoSegmentIndex(ref_points, p.ego_pose, ego_nearest_param_);
  ref_points = trajectory_utils::cropPoints(
    ref_points, p.ego_pose.position, ego_seg_idx, forward_traj_length + tmp_margin,
    -backward_traj_length - tmp_margin);
```

The function adjusts the length once, then the `cropPoints()` function adjust the length again internally. This could be re-written simply as:
```C++
  size_t ego_seg_idx =
    trajectory_utils::findEgoSegmentIndex(ref_points, p.ego_pose, ego_nearest_param_);
  double longitudinal_offset = motion_utils::calcLongitudinalOffsetToSegment(ref_points, ego_seg_idx, p.ego_pose);
  ref_points = trajectory_utils::cropPoints(
    ref_points, ego_seg_idx, forward_traj_length + longitudinal_offset + tmp_margin,
    -backward_traj_length - longitudinal_offset - tmp_margin);
```

It's one more line on the caller side, but the API of `cropPoints()` would be clean, and could be also used when the "`target_pos`" does not necessarily make sense, or if we mostly care about performance. At the end of the day, even considering `target_pos`=`target_seg_idx` could be an acceptable approximation.

takayuki5168:
@VRichardJP
I prefer "how simple the node's (caller side's) implementation will be" to "separating the utils functions so that each function has its own separated meaning".

A lot of trajectory functions in motion_utils have the arguments, the position and its segment index.
https://github.com/autowarefoundation/autoware.universe/blob/main/common/motion_utils/include/motion_utils/trajectory/trajectory.hpp
The idea is, the algorithm to find the segment index depends on its object type (ego, dynamic object, traffic light object, etc). Giving the information of the algorithm to find the segment index and its parameters to the motion_utils' functions will be very complex. Therefore we decided to calculate the position and its segment index in the caller side and give them to motion_utils's functions as arguments.
https://autowarefoundation.github.io/autoware.universe/main/common/motion_utils/#application-to-various-object

> , and don't have to care with some strange projection logic.

For me, the projection logic does not sound strange. This projection logic is because of the above rule that Autoware's functions should follow.

takayuki5168:
Anyway, this PR follows the existing rule that the position and its segment index are given to the motion_utils's functions.

If you think your idea is better, how about discussing it on the github discussion?

PR #3130, opened 2023-03-21
feat(lidar_centerpoint_tvm): single inference lidar centerpoint node

tohmae:
## Description
I add single inference lidar centerpoint node to lidar_centerpoint_tvm.

## Pre-review checklist for the PR author

The PR author **must** check the checkboxes below when creating the PR.

- [x] I've confirmed the [contribution guidelines].
- [x] The PR follows the [pull request guidelines].

## In-review checklist for the PR reviewers

The PR reviewers **must** check the checkboxes below before approval.

- [ ] The PR follows the [pull request guidelines].

## Post-review checklist for the PR author

The PR author **must** check the checkboxes below before merging.

- [ ] There are no open discussions or they are tracked via tickets.

After all checkboxes are checked, anyone who has write access can merge the PR.

[contribution guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/
[pull request guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/pull-request-guidelines/

PR #3266, opened 2023-04-04
fix(lidar_centerpoint_tvm): wrong output when downsample_factor_> 1

### LINKED ISSUE #2649: Wrong Outputs of CenterPoint TVM
### Checklist

### Description

I've seen sometimes the outputs objects are wrong. It happens somehow randomly. I don't see a way to reproduce it every time. I can confirm that it is not because of the CUDA to C++ implementation because I have verified the output of every individual function. The issue seems to be related to the C++ deployment of TVM somehow. Further investigation is needed. 

### Expected behavior

Bounding boxes should match detected objects. 

### Actual behavior

They are not matched. 

### Steps to reproduce

Run [Rosbag replay tutorial](https://autowarefoundation.github.io/autoware-documentation/main/tutorials/ad-hoc-simulation/rosbag-replay-simulation/)

### Versions

latest master

### Possible causes

https://discuss.tvm.apache.org/t/different-output-for-large-yolo-onnx-model-in-python-correct-and-c-incorrect/11537/20

### Additional context

_No response_

tohmae:
## Description

Fix #2649
I think downsample_factor_ is used in the generateBoxes3D_worker function incorrectly.
In fact, when downsample_factor_=1, the output by lidar_centerpoint_tvm is similar to the output by lidar_centerpoint ,
but when downsample_factor_=2, the output is different from the output by lidar_centerpoint.
I think that when downsample_factor_= 2, we must extract even-numbered rows and columns from 560x560 matrix.

## Tests performed

I compared the output before and after the modification by single_centerpoint_node  in lidar_centerpoint_tvm. (#3130)

- before fix
lidar_centerpoint_tvm

- after fix
lidar_centerpoint_new

- (for reference) generated by single_centerpoint_node  in lidar_centerpoint
lidar_centerpoint

### DIFF
--- a/perception/lidar_centerpoint_tvm/lib/postprocess/generate_detected_boxes.cpp
+++ b/perception/lidar_centerpoint_tvm/lib/postprocess/generate_detected_boxes.cpp
@@ -69,39 +69,40 @@ void generateBoxes3D_worker(
   // heatmap: N = class_size, offset: N = 2, z: N = 1, dim: N = 3, rot: N = 2, vel: N = 2
   for (std::size_t idx = 0; idx < grids_per_thread; idx++) {
     std::size_t grid_idx = thread_idx * grids_per_thread + idx;
-    const auto down_grid_size = config.down_grid_size_y_ * config.down_grid_size_x_;
-    if (grid_idx >= down_grid_size) {
+    std::size_t array_idx = config.downsample_factor_ * grid_idx;
+    const auto grid_size = config.grid_size_y_ * config.grid_size_x_;
+    if (array_idx >= grid_size) {
       return;
     }
 
-    const auto yi = grid_idx / config.down_grid_size_x_;
-    const auto xi = grid_idx % config.down_grid_size_x_;
+    const auto yi = array_idx / config.grid_size_x_;
+    const auto xi = array_idx % config.grid_size_x_;
 
     int32_t label = -1;
     float max_score = -1;
     for (std::size_t ci = 0; ci < config.class_size_; ci++) {
-      float score = sigmoid(out_heatmap[down_grid_size * ci + grid_idx]);
+      float score = sigmoid(out_heatmap[grid_size * ci + array_idx]);
       if (score > max_score) {
         label = ci;
         max_score = score;
       }
     }
 
-    const float offset_x = out_offset[down_grid_size * 0 + grid_idx];
-    const float offset_y = out_offset[down_grid_size * 1 + grid_idx];
+    const float offset_x = out_offset[grid_size * 0 + array_idx];
+    const float offset_y = out_offset[grid_size * 1 + array_idx];
     const float x =
-      config.voxel_size_x_ * config.downsample_factor_ * (xi + offset_x) + config.range_min_x_;
+      config.voxel_size_x_ * (xi + offset_x) + config.range_min_x_;
     const float y =
-      config.voxel_size_y_ * config.downsample_factor_ * (yi + offset_y) + config.range_min_y_;
-    const float z = out_z[grid_idx];
-    const float w = out_dim[down_grid_size * 0 + grid_idx];
-    const float l = out_dim[down_grid_size * 1 + grid_idx];
-    const float h = out_dim[down_grid_size * 2 + grid_idx];
-    const float yaw_sin = out_rot[down_grid_size * 0 + grid_idx];
-    const float yaw_cos = out_rot[down_grid_size * 1 + grid_idx];
+      config.voxel_size_y_ * (yi + offset_y) + config.range_min_y_;
+    const float z = out_z[array_idx];
+    const float w = out_dim[grid_size * 0 + array_idx];
+    const float l = out_dim[grid_size * 1 + array_idx];
+    const float h = out_dim[grid_size * 2 + array_idx];
+    const float yaw_sin = out_rot[grid_size * 0 + array_idx];
+    const float yaw_cos = out_rot[grid_size * 1 + array_idx];
     const float yaw_norm = sqrtf(yaw_sin * yaw_sin + yaw_cos * yaw_cos);
-    const float vel_x = out_vel[down_grid_size * 0 + grid_idx];
-    const float vel_y = out_vel[down_grid_size * 1 + grid_idx];
+    const float vel_x = out_vel[grid_size * 0 + array_idx];
+    const float vel_y = out_vel[grid_size * 1 + array_idx];
 
     boxes3d[grid_idx].label = label;
     boxes3d[grid_idx].score = yaw_norm >= config.yaw_norm_threshold_ ? max_score : 0.f;
--- a/perception/lidar_centerpoint_tvm/lib/postprocess/generate_detected_boxes.cpp
+++ b/perception/lidar_centerpoint_tvm/lib/postprocess/generate_detected_boxes.cpp
@@ -90,10 +90,8 @@ void generateBoxes3D_worker(
 
     const float offset_x = out_offset[grid_size * 0 + array_idx];
     const float offset_y = out_offset[grid_size * 1 + array_idx];
-    const float x =
-      config.voxel_size_x_ * (xi + offset_x) + config.range_min_x_;
-    const float y =
-      config.voxel_size_y_ * (yi + offset_y) + config.range_min_y_;
+    const float x = config.voxel_size_x_ * (xi + offset_x) + config.range_min_x_;
+    const float y = config.voxel_size_y_ * (yi + offset_y) + config.range_min_y_;
     const float z = out_z[array_idx];
     const float w = out_dim[grid_size * 0 + array_idx];
     const float l = out_dim[grid_size * 1 + array_idx];
--- a/perception/lidar_centerpoint_tvm/lib/postprocess/generate_detected_boxes.cpp
+++ b/perception/lidar_centerpoint_tvm/lib/postprocess/generate_detected_boxes.cpp
@@ -69,14 +69,15 @@ void generateBoxes3D_worker(
   // heatmap: N = class_size, offset: N = 2, z: N = 1, dim: N = 3, rot: N = 2, vel: N = 2
   for (std::size_t idx = 0; idx < grids_per_thread; idx++) {
     std::size_t grid_idx = thread_idx * grids_per_thread + idx;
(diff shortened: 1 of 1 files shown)


perception/lidar_centerpoint_tvm/lib/postprocess/generate_detected_boxes.cpp
```diff
@@ -69,39 +69,39 @@ void generateBoxes3D_worker(
...
-    const auto xi = grid_idx % config.down_grid_size_x_;
+    const auto yi = config.downsample_factor_ * (grid_idx / config.down_grid_size_x_);
+    const auto xi = config.downsample_factor_ * (grid_idx % config.down_grid_size_x_);
+    std::size_t array_idx = yi * config.grid_size_x_ + xi;
 
     int32_t label = -1;
     float max_score = -1;
     for (std::size_t ci = 0; ci < config.class_size_; ci++) {
-      float score = sigmoid(out_heatmap[down_grid_size * ci + grid_idx]);
+      float score = sigmoid(out_heatmap[grid_size * ci + array_idx]);
```
angry-crab, line 85: `out_heatmap` is of size `down_grid_size_x_ * down_grid_size_y_ * class_size_`. I think this might end with out of boundary access. 

```
 const auto grid_xy_size = config_.down_grid_size_x_ * config_.down_grid_size_y_;
.....
  head_out_heatmap_d_ = cuda::make_unique<float[]>(grid_xy_size * config_.class_size_);
```

perception/lidar_centerpoint_tvm/lib/postprocess/generate_detected_boxes.cpp
```diff
@@ -69,39 +69,39 @@ void generateBoxes3D_worker(
...
     const auto down_grid_size = config.down_grid_size_y_ * config.down_grid_size_x_;
     if (grid_idx >= down_grid_size) {
       return;
     }
 
-    const auto yi = grid_idx / config.down_grid_size_x_;
-    const auto xi = grid_idx % config.down_grid_size_x_;
+    const auto yi = config.downsample_factor_ * (grid_idx / config.down_grid_size_x_);
+    const auto xi = config.downsample_factor_ * (grid_idx % config.down_grid_size_x_);
+    std::size_t array_idx = yi * config.grid_size_x_ + xi;
```
angry-crab, line 80: Multiplying downsample factor with idx gives discrete values of xi and yi, which does not seem consistent with the original implementation. 

https://github.com/autowarefoundation/autoware.universe/blob/b78b17d6be85611e0686b1c319f3ae0ffe7bd890/perception/lidar_centerpoint/lib/postprocess/postprocess_kernel.cu#L66-L69

angry-crab:
Thanks for the PR. I left some comments please let me know if those make sense.

tohmae:
Thank you for your comment.

https://github.com/autowarefoundation/autoware.universe/blob/ca9cac30795e9bf58732250df487c1db723d5ad0/perception/lidar_centerpoint/launch/lidar_centerpoint.launch.xml#L5-L6

https://github.com/autowarefoundation/autoware.universe/blob/ca9cac30795e9bf58732250df487c1db723d5ad0/perception/lidar_centerpoint/launch/lidar_centerpoint.launch.xml#L21-L24


Checking the source of lidar_centerpoint, the model is different between centerpoint and centerpoint_tiny. Probably the output by the centerpoint model is 560x560 dimensions and the output by the centerpoint_tiny model is 280x280 dimensions.
However, in lidar_centerpoint_tvm, the model is the same for centerpoint and centerpoint_tiny, and the output is the same 560x560 dimensions.

Because in the future, lidar_centerpoint_tvm may also change the model between centerpoint and centerpoint_tiny,
To fix wrong output when downsample_factor > 1 (#2649) , I think a PR that changes downsample_factor in centerpoint_tiny.param.yaml from 2 to only 1 would be better than this PR (#3266).

https://github.com/autowarefoundation/autoware.universe/blob/ca9cac30795e9bf58732250df487c1db723d5ad0/perception/lidar_centerpoint_tvm/config/centerpoint_tiny.param.yaml#L9

tohmae:
If you agree with my opinion, I will close this PR and create a new PR that that just changes downsample_factor in centerpoint_tiny.param.yaml from 2 to 1.

angry-crab:
Thanks for the heads up. 
> If you agree with my opinion above, I will close this PR and create a new PR that that just changes downsample_factor in centerpoint_tiny.param.yaml from 2 to 1.

I don't think we should change it to 1 because it is misleading and could be omitted in the future. Actually, `centerpoint_tiny`
 is not intended to be used for tvm when we were porting the package. But we can make it generic in this PR as well. 

Let me try to compare the output next week and come back with more suggestions.

angry-crab:
Sorry for the late response. According to my test, the current implementation is correct. `downsample_factor_` is handled by onnx and tenssorrt by default. downsampling is not required for generating boxed. 

https://github.com/autowarefoundation/autoware.universe/blob/d73c3ae41b53b76589d78a7a019cd065d031fd9d/perception/lidar_centerpoint/lib/network/network_trt.cpp#L73-L78

As you can see here, the dimension of output is set.

tohmae:
Thank you for taking the time to look into this.

Please let me know what I don't understand. 
The code you showed is for lidar_centerpoint. I also think that the generateBoxes3D_worker in lidar_centerpoint is correct. 
However, I beliave that the generateBoxes3D_worker in lidar_centerpoint_tvm is wrong and that is why issue #2649 is occurring. Has issue #2649 already been resolved ?

Sorry if this is a misguided question.

angry-crab:
Thanks for the comments.

> However, I think that the generateBoxes3D_worker in lidar_centerpoint_tvm is wrong and that is why issue https://github.com/autowarefoundation/autoware.universe/issues/2649 is occurring.

Honestly, I don't think `generateBoxes3D_worker` in `lidar_centerpoint_tvm` is wrong. I tried to compare the outputs of `lidar_centerpoint_tvm` and `lidar_centerpoint`, and the outputs are identical given same inputs. Since the one in `lidar_centerpoint` is correct, how could the one in `lidar_centerpoint_tvm` be wrong given their output are equal. 

Further, `generateBoxes3D_worker` in `lidar_centerpoint` does not do downsampling as well because the output dimension is determined by config files with different onnx as well. 

> Has issue https://github.com/autowarefoundation/autoware.universe/issues/2649 already been resolved ?

Unfortunately, it is not resolved yet. We are still looking into the problem. I thought there might be something wrong with tvm itself.

tohmae:
Thank you for the explanation.

>But the screen shots seem correct. Would you be able to run it a couple more times to see if the detections are good?

Of course. I would like to test other pointclouds with the following 3 patterns. 
case 1. lidar_centerpoint
case 2. lidar_centerpoint_tvm (default)
case 3. lidar_centerpoint_tvm (change downsample_factor from 2 to 1 in centerpoint_tiny.param.yaml)

Would this be OK?

angry-crab:
> case 3. lidar_centerpoint_tvm (change downsample_factor from 2 to 1 in centerpoint_tiny.param.yaml)

Case 3 would not work because the network are differrent. `centerpoint_tiny` onnx file does not exist in `modelzoo`

tohmae:
Sorry for the lack of explanation.

https://github.com/autowarefoundation/autoware.universe/blob/541122f6d214db36b1358c6ba02be2f8d8f7b089/perception/lidar_centerpoint_tvm/launch/lidar_centerpoint_tvm.launch.xml#L8

Currently, the default model_name in the launch file is centerpoint_tiny, so I wrote the above.
As you pointed out, precisely,
case 2. lidar_centerpoint_tvm(downsample_factor=2)
case 3. lidar_centerpoint_tvm(downsample_factor=1)

The test has already been finished and the results will be pasted later.
Thank you in advance for your cooperation.

tohmae:
I tested four patterns.
I am attaching a pcd file and a file detectioned in 3 different ways.
case 1: lidar_centerpoint ->red
case 2. lidar_centerpoint_tvm(downsample_factor=2) ->green
case 3. lidar_centerpoint_tvm(downsample_factor=1) ->blue

- pcd1
pcd1

- detection1 
detection1

- pcd2
pcd2


- detection2
detection2

- pcd3
pcd3


- detection3
detection3

- pcd4
pcd4

- detection4
detection4

To me, case1(lidar_centerpoint) and case3(downsample_factor=1) look close, except for pcd3.
What do you think?

angry-crab:
> I tested four patterns

Thank you so much for testing! 

Could you double check that in case 1 which network model you are using? 
Normally, the default arg is not used because launch files in the upper level would set and pass the value to `lidar_centerpoint`. 

> To me, case1(lidar_centerpoint) and case3(downsample_factor=1) look similar except for pcd3.

I think we were abusing the config file before as `downsample_factor` is supposed to be 1 for this network.

tohmae:
> Could you double check that in case 1 which network model you are using?

In case 1 I started node by [single_inference_lidar_centerpoint.launch.xml](https://github.com/autowarefoundation/autoware.universe/blob/main/perception/lidar_centerpoint/launch/single_inference_lidar_centerpoint.launch.xml).
So I used `centerpoint_tiny` network model in lidar_centerpoint in case 1.
https://github.com/autowarefoundation/autoware.universe/blob/dd0779c04ee4eda4c835d90f823a93bd5578c0b4/perception/lidar_centerpoint/launch/single_inference_lidar_centerpoint.launch.xml#L3

>I just realized that the config files for lidar_centerpoint_tvm are same except for downsample_factor. I think the best way to fix it would be to simply delete centerpoint_tiny config file since it is misleading.

I also think that your fix is better than my fix.
May I close this PR and create a new PR that deletes centerpoint_tiny config file and changes model_name in lidar_centerpoint_tvm.launch.xml from `centerpoint_tiny` to `centerpoint` ?

https://github.com/autowarefoundation/autoware.universe/blob/351af24e02c80a9626a3fb4322755e317de876c5/perception/lidar_centerpoint_tvm/launch/lidar_centerpoint_tvm.launch.xml#L8

angry-crab:
> I also think that your fix is better than my fix. May I close this PR and create a new PR that deletes centerpoint_tiny config file and changes model_name in lidar_centerpoint_tvm.launch.xml from `centerpoint_tiny` to `centerpoint` ?

Sure! Thanks a lot for the help!

PR #3273, opened 2023-04-05
fix(motion_utils): improve mechanism to delete virtual wall markers

maxime-clem:
## Description

Fixes https://github.com/autowarefoundation/autoware.universe/issues/3138

## Tests performed

Not applicable.

### DIFF
--- a/common/motion_utils/src/marker/marker_helper.cpp
+++ b/common/motion_utils/src/marker/marker_helper.cpp
@@ -135,39 +135,27 @@ visualization_msgs::msg::MarkerArray createDeletedDeadLineVirtualWallMarker(
 visualization_msgs::msg::MarkerArray VirtualWallMarkerCreator::handleVirtualWallMarker(
   const std::vector<Pose> & poses, const std::string & module_name, const rclcpp::Time & now,
   int32_t id, create_wall_function function_create_wall_marker,
-  delete_wall_function function_delete_wall_marker,
-  std::vector<geometry_msgs::msg::Pose> & previous_virtual_wall_poses,
+  delete_wall_function function_delete_wall_marker, size_t & previous_virtual_walls_nb,
   const double longitudinal_offset)
 {
-  size_t id_to_create = id;
-  size_t id_to_delete = id;
   visualization_msgs::msg::MarkerArray wall_marker;
+
+  const auto max_id = id + previous_virtual_walls_nb;
 
   if (poses.size() == 0) {
     return wall_marker;
   }
 
-  for (const auto & p : previous_virtual_wall_poses) {
-    const bool previous_stop_pose_is_in_stop_pose =
-      std::any_of(poses.begin(), poses.end(), [&](const geometry_msgs::msg::Pose & elem) {
-        std::vector<Pose> poses;
-        poses.push_back(p);
-        poses.push_back(elem);
-        return resample_utils::validate_points_duplication(poses);
-      });
-
-    if (!previous_stop_pose_is_in_stop_pose) {
-      appendMarkerArray(function_delete_wall_marker(now, id_to_delete), &wall_marker, now);
-    }
-    id_to_delete++;
-  }
-
   for (const auto & p : poses) {
     appendMarkerArray(
-      function_create_wall_marker(p, module_name, now, id_to_create++, longitudinal_offset),
-      &wall_marker);
+      function_create_wall_marker(p, module_name, now, id++, longitudinal_offset), &wall_marker);
   }
-  previous_virtual_wall_poses = poses;
+
+  while (id < max_id) {
+    appendMarkerArray(function_delete_wall_marker(now, id++), &wall_marker, now);
+  }
+
+  previous_virtual_walls_nb = poses.size();
   return wall_marker;
 }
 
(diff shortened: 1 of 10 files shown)


ahmeddesokyebrahim:
Thanks @maxime-clem for this PR and contribution your providing for this issue.

Honestly I am not able to fully understand your solution.
May you please clarify what do you mean by the prefix `_nb`.

I know it is still draft PR, but I have noticed that it is not compiling. I believe if I can test and try your approach out that will help me a lot in understanding how it works.

In all-ways, I do appreciate your contribution.

maxime-clem:
> Honestly I am not able to fully understand your solution. May you please clarify what do you mean by the suffix `_nb`.

I will try to explain my understanding of the problem and why my solution should work.

#### How rviz markers work

- A rviz marker is published with a namespace `ns`  and an `id`. By default, this marker will be displayed forever.
- If a new marker is published with the same `ns` and `id`, it replaces the previous marker (i.e., the previous marker disappears).
- If a marker has a `lifetime`, it will disappear after the given duration.
- A marker can explicitly be deleted by publishing a marker with action `DELETE` and the `ns` and `id` to delete.

#### Virtual wall markers

 When planning module makes ego stop a specific point (e.g., stop before a crosswalk with pedestrians), a rviz marker is published to show a virtual wall in front of where ego will stop. By default, such virtual wall marker has a `lifetime` of `0.5s`. However, when a module no longer make ego stop (e.g., no more pedestrian around the crosswalk), we want the virtual wall marker to disappear instantly. Otherwise, we may see ego drive through the remaining virtual wall and a user may be worried about the behavior of Autoware.
The way to achieve this is to publish a `DELETE` marker to make the virtual wall disappear.

#### Current issue

The issue with the current implementation is that the same `id` may be used for adding a new virtual wall and for removing a previous virtual wall.
```C++
  size_t id_to_create = id;
  size_t id_to_delete = id;
```
I believe this is a mistake and the cause for the issue.

#### Solution of this PR

With how markers work, we only need to publish a `DELETE` marker for an `id` if we do not already publish an `ADD` marker for the same `id`.
With my PR, I store the number (`nb_`) of previously published markers. Then, in the next iteration, if we publish a number of markers `n`, I make sure to publish `DELETE` markers for all ids `n < id < *_nb_`.

The new logic does not work if the "base id" (function parameter `id`) changes between iterations but it seems to never be the case (and I believe it should always be set to `0` anyway). It also does not work if the function is called multiple times in the same iteration (this issue is also present in the current implementation).

I hope it is clear. Let me know if you have any questions.

> I know it is still draft PR, but I have noticed that it is not compiling. I believe if I can test and try your approach out that will help me a lot in understanding how it works.

Sorry about that. I was building with the `-w` option (disable the warning-as-errors) and missed a build error. It should be fixed now.

ahmeddesokyebrahim:
Thanks @maxime-clem  for your contribution and the detailed explanation.

I have carefully read your explanation and checked your implementation as well. I do believe this approach should be solving the current issue.

However, I was testing your PR with planning simulator and faced an issue. AW was not able to plan a path and hence not driving. You see a recorded video for that [here](https://youtu.be/Ym-m7_mwE8Q). Please let me know if you face the same issue, or is it something wrong with my setup.

Thanks again :)

maxime-clem:
@ahmeddesokyebrahim the branch was outdated and some mismatch with the `autoware_launch` probably broke some planning modules.
I updated the branch and it should work fine now.

ahmeddesokyebrahim:
Thanks @maxime-clem for the update.

I have tested it and working as intended :+1:

kyoichi-sugahara:
@maxime-clem @ahmeddesokyebrahim 
Thank you so much for fixing my bug and sorry for making you guys confused.
My intention was if some of previous stop poses are not in new stop poses, delete those previous stop poses.(There is a bug although)
But maxime's proposal is delete all virtual wall marker based on previous number of id and create virtual wall marker based on module output correct?
I think it's good way if "base_id" is always same.

common/motion_utils/src/marker/marker_helper.cpp
```diff
@@ -135,39 +135,27 @@ visualization_msgs::msg::MarkerArray createDeletedDeadLineVirtualWallMarker(
...
+  delete_wall_function function_delete_wall_marker, size_t & previous_virtual_walls_nb,
   const double longitudinal_offset)
 {
-  size_t id_to_create = id;
-  size_t id_to_delete = id;
   visualization_msgs::msg::MarkerArray wall_marker;
 
+  const auto max_id = id + static_cast<int32_t>(previous_virtual_walls_nb);
+
   if (poses.size() == 0) {
```
kyoichi-sugahara, line 145: poses.empty() is better?

kyoichi-sugahara (approved):
LGTM
Thank you so much for fixing my bugs always!

common/motion_utils/src/marker/marker_helper.cpp
kyoichi-sugahara: sorry 
when pose is empty, return should not be written. previuslly published pose should be deleted.
so this `if` part is not necessary

maxime-clem:
@kyoichi-sugahara I removed the `id` from the function parameters. Do you see any issue with that ?

kyoichi-sugahara:
@maxime-clem 
If all module's id starts from 0, i think it's ok.
You said so yesterday?(I don't remember correctly sorry)

kyoichi-sugahara:
Then, let's merge this PR

satoshi-ota (approved):

PR #3385, opened 2023-04-12
perf(control_cmd_gate): reduce delays introduced by the command gate

maxime-clem:
## Description

Before being sent to the vehicle interface, control messages are handled by the `control_cmd_gate`.
The gate select which message to send to the vehicle based on the current mode (autonomous, remote, stopped, ...) and also applies some filtering on the control command to avoid jerky controls.
Unfortunately, this gate introduces several delays when transitioning between a stopped state and a moving state.
This PR reduces this delay in several ways.

### Jerk filtering

When filtering a control command, some jerk limit is used (by default, `5.0m/s³`).
When stopped, a control command with a negative acceleration (by default `-1.0m/s²`).
This means that, when Autoware decides to move (e.g., after an obstacle is gone, or when engaging the autonomous mode), the filtering will limit the jerk assuming a negative acceleration. This leads to the acceleration staying in negative values for several iterations (`200ms` with the default parameters), adding a significant delay when switching between a stopped and moving state.

To solve this, when ego is stopped, the filtering is not applied on the acceleration of the control command (i.e., the negative value), but on the measured acceleration (i.e., `0.0m/s²`).

### Pause interface

The control command gate uses an API to control whether the vehicle is allowed to move or should "pause".
In the gate, if the state is "paused", the control command is set such that ego stops or stays stopped with a negative acceleration.
When the input control command of the gate corresponds to a move forward, the gate will use the API to "request to start". Then, a response is received with the new "paused" state (usually, after the request to start, the state will be unpaused).
These API calls introduce delays due to their update, both on the gate side and the server side. The updates are made at a constant rate (`10hz` by default), introducing some delay (around `200ms` with the default parameters).

This PR changes the logic to do updates as soon as a state changes, minimizing the delays.

## Tests performed

Delays were measured with the `simple_planning_simulator` by plotting the acceleration of the input and output control commands, as well as the measured acceleration and engage topic.
The delay is the time between the engage signal and when ego starts to move (measured acceleration > 0.0).
Keep in mind that this time will includes the control delay (delay on the vehicle side) which is around `200ms`.

### Current Autoware (500ms~600ms).

:warning: This picture shows the case where autoware is "in transition" after the engage, limiting the jerk by a more conservative value (`0.5m/s³` by default). I am not able to reproduce this case with the latest universe and the delay is only around 550ms after engaging.
image

### With this PR (~250ms)
image


control/vehicle_cmd_gate/src/vehicle_cmd_gate.cpp
```diff
@@ -496,9 +500,11 @@ AckermannControlCommand VehicleCmdGate::filterControlCommand(const AckermannCont
   AckermannControlCommand out = in;
   const double dt = getDt();
   const auto mode = current_operation_mode_;
+  const auto current_status_cmd = getActualStatusAsCommand();
+  const auto ego_is_stopped = std::abs(current_status_cmd.longitudinal.speed) < 1e-3;
 
   // Apply transition_filter when transiting from MANUAL to AUTO.
-  if (mode.is_in_transition) {
+  if (!ego_is_stopped && mode.is_in_transition) {
```
TakaHoribe, line 507: @maxime-clem Sorry if I misunderstand, but I guess just adding the condition:

```
if (ego_is_stopped) {
    prev_values.longitudinal = current_status_cmd.longitudinal;
}
```

is enough. Why the `filter_on_transition_` is required when the ego is stopped?

control/vehicle_cmd_gate/src/vehicle_cmd_gate.cpp
```diff
@@ -520,6 +522,13 @@ AckermannControlCommand VehicleCmdGate::filterControlCommand(const AckermannCont
   // filter in manual mode.
   prev_values.longitudinal = out.longitudinal;  // TODO(Horibe): to be removed
 
+  // When ego is stopped, consider the actual vehicle longitudinal state for the next filtering
+  // this is to prevent the jerk limits being applied on the "stop acceleration",
+  // which may be negative and cause delays when restarting the vehicle.
+  if (ego_is_stopped) {
+    prev_values.longitudinal = current_status_cmd.longitudinal;
+  }
```
TakaHoribe, line 530: @maxime-clem I found a concern. Assume that the vehicle will not move with a small target velocity (e.g. 1.0m/s).
In this case, if `prev_values` is set with a current velocity (or acceleration) every time when the vehicle is stopped, the output target velocity will not exceed `v_current + a_lim * dt` (which is usually very small) as long as the vehicle is stopped.

If the vehicle will not move with the target velocity `v_current + a_lim * dt`, then the vehicle will never start moving.

So, I suggest changing the logic like this:

```
bool hasSmallValue(cmd) {
    if (cmd.acceleration < -0.5) {
        true;
    }
    return false;
}

// not only check the stop status but also check if the current acceleration is large to negative direction
if (ego_is_stopped && hasSmallValue(prev_values)) {
    prev_values.longitudinal = current_status_cmd.longitudinal;
}
```

system/default_ad_api/src/motion.cpp
```diff
@@ -138,6 +138,7 @@ void MotionNode::on_is_start_requested(
   const control_interface::IsStartRequested::Message::ConstSharedPtr msg)
 {
   is_start_requested_ = msg->data;
+  update_state();
```
maxime-clem, line 141: @isamu-takagi 
Do you see any issue with this change ?
The goal of this change is to allow the state to be updated as soon as a request to start is received. Otherwise we have to wait for the timer callback (which can take up to `100ms` since it runs at `10hz` by default).

control/vehicle_cmd_gate/src/vehicle_cmd_gate.cpp
maxime-clem: Thank you for finding this issue.
Your solution sounds good. I implemented it by adding a condition:
```
const auto input_cmd_is_stopping = in.longitudinal.acceleration < 0.0;
```

TakaHoribe: Looks good!

system/default_ad_api/src/motion.cpp
isamu-takagi: Thank you. Your point made me realize that other parts also need improvement. Could you please check this commit https://github.com/autowarefoundation/autoware.universe/pull/3385/commits/b8bf15664c77c68914721395dfd4c7b9569efda3? I pushed it to this branch by mistake.

maxime-clem: Thank you. I tested with your commit and it still works well.
Should I revert your commit ?

isamu-takagi: > Should I revert your commit ? (in case you want to create a separate PR).

I think it's fine to merge with this PR.

TakaHoribe (approved):
LGTM!

control/vehicle_cmd_gate/script/delays_checker.py
```diff
@@ -0,0 +1,142 @@
...
+
+from autoware_auto_control_msgs.msg import AckermannControlCommand
+from autoware_auto_vehicle_msgs.msg import Engage
+from geometry_msgs.msg import AccelWithCovarianceStamped
+from nav_msgs.msg import Odometry
+import rclpy
+from rclpy.node import Node
+
+accel_topic = "/localization/acceleration"
+odom_topic = "localization/kinematic_state"
```
yukkysaito, line 25: nitpicking :pray: 
```suggestion
odom_topic = "/localization/kinematic_state"
```

control/vehicle_cmd_gate/script/delays_checker.py
```diff
@@ -0,0 +1,142 @@
+#!/usr/bin/env python3
+
+# Copyright 2023 Tier IV, Inc.
```
yukkysaito, line 3:
```suggestion
# Copyright 2023 TIER IV, Inc.
```

yukkysaito, line 3:
```suggestion
# Copyright 2023 TIER IV, Inc.
```

yukkysaito (approved):
LGTM

PR #3494, opened 2023-04-21
fix(motion_velocity_smoother): insert the point when applying external velocity limit

TomohitoAndo:
## Description

In the current implementation, when applying the external velocity limit, the velocity of the points on the trajectory are being overwritten. However, this implementation has an issue where the inserting position of the external velocity limit is further away than expected and ego vehicle sometimes can't stop at the desired position (see the video below).
In this PR, I have resolved this problem by inserting the point of the external velocity limit.

https://user-images.githubusercontent.com/11865769/233629346-72330fa1-61b0-4bc8-91a6-513f12e440f9.mp4

### Before
image

### After
image

## Tests performed

I confirmed this works with simple planning simulator.

https://user-images.githubusercontent.com/11865769/233630064-8d7fbd16-4cd8-4257-9bb3-7d38d27f1176.mp4

## Effects on system behavior

Not applicable.

### DIFF
--- a/planning/motion_velocity_smoother/src/motion_velocity_smoother_node.cpp
+++ b/planning/motion_velocity_smoother/src/motion_velocity_smoother_node.cpp
@@ -826,16 +826,31 @@ void MotionVelocitySmootherNode::applyExternalVelocityLimit(TrajectoryPoints & t
   trajectory_utils::applyMaximumVelocityLimit(
     0, traj.size(), max_velocity_with_deceleration_, traj);
 
-  const size_t closest_idx = findNearestIndexFromEgo(traj);
+  // insert the point that has external limited velocity
+  const auto & current_pose = current_odometry_ptr_->pose.pose;
+  const size_t closest_seg_idx = motion_utils::findFirstNearestSegmentIndexWithSoftConstraints(
+    traj, current_pose, node_param_.ego_nearest_dist_threshold,
+    node_param_.ego_nearest_yaw_threshold);
+  const auto inserted_index =
+    motion_utils::insertTargetPoint(closest_seg_idx, external_velocity_limit_.dist, traj);
+  if (!inserted_index) {
+    return;
+  }
+  const double original_vel = traj.at(*inserted_index).longitudinal_velocity_mps;
+  traj.at(*inserted_index).longitudinal_velocity_mps =
+    std::min(external_velocity_limit_.velocity, original_vel);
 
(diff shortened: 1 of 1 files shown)


planning/motion_velocity_smoother/src/motion_velocity_smoother_node.cpp
```diff
@@ -826,16 +826,31 @@ void MotionVelocitySmootherNode::applyExternalVelocityLimit(TrajectoryPoints & t
...
+    current_pose, traj.at(closest_seg_idx + 1).pose.position);
+  for (size_t idx = closest_seg_idx + 1; idx < traj.size() - 1; ++idx) {
     if (dist > external_velocity_limit_.dist) {
       trajectory_utils::applyMaximumVelocityLimit(
-        idx + 1, traj.size(), external_velocity_limit_.velocity, traj);
+        idx, traj.size(), external_velocity_limit_.velocity, traj);
       return;
     }
+
+    dist += tier4_autoware_utils::calcDistance2d(traj.at(idx), traj.at(idx + 1));
```
rej55, line 853: It seems it is enough to apply maximum velocity from `*inserted_index` to `traj.size()`.
So, I suggest to apply maximum velocity like following:
```
trajectory_utils::applyMaximumVelocityLimit(
        *inserted_index, traj.size(), external_velocity_limit_.velocity, traj);
```
How do you think?

TomohitoAndo: Thanks! It's simpler and clearer.
I fixed in  https://github.com/autowarefoundation/autoware.universe/pull/3494/commits/87008dc1535283613cfe0a2857133fe6bc571ae0.


I moved this part to the place where the insert index was not found.
However I'm not sure why this is necessary. Is this modification correct?
```
    traj.back().longitudinal_velocity_mps = std::min(
      traj.back().longitudinal_velocity_mps, static_cast<float>(external_velocity_limit_.velocity));
```
https://github.com/autowarefoundation/autoware.universe/pull/3494/files#diff-19a3604d5e77c39fa02ac64cad147bccfa939c955d7ba1743641f4ca16d50cfdR836-R839

rej55: My memory is a little fuzzy, but I seem to recall it was because if a deceleration point was not set, it would not slow down forever.
So, I think the modification is correct.

PR #3603, opened 2023-05-01
fix(intersection): fixed virtual wall marker id duplication

soblin:
## Description

When there are multiple intersection module running and each of them is publishing virtual walls, there ids were duplicate and some walls were not displayed.

## Related links

https://tier4.atlassian.net/browse/RT1-2283

## Tests performed

Psim

### Before this PR

only intersection walls farthest from ego are displayed.

image

### After this PR

image

## Interface changes

Multiple intersection wall marker appears properly when all modules are disapproved.

## Effects on system behavior

Multiple intersection wall marker appears properly when all modules are disapproved.

### DIFF
--- a/planning/behavior_velocity_planner/src/scene_module/intersection/debug.cpp
+++ b/planning/behavior_velocity_planner/src/scene_module/intersection/debug.cpp
@@ -223,7 +223,7 @@ visualization_msgs::msg::MarkerArray IntersectionModule::createVirtualWallMarker
       &wall_marker, now);
   }
 
-  auto id = 0;
+  auto id = planning_utils::bitShift(module_id_);
   for (auto & marker : wall_marker.markers) {
     if (marker.action == visualization_msgs::msg::Marker::ADD) {
       marker.id = id;


shmpwk (dismissed):

planning/behavior_velocity_planner/src/scene_module/intersection/debug.cpp
```diff
@@ -223,7 +223,7 @@ visualization_msgs::msg::MarkerArray IntersectionModule::createVirtualWallMarker
       &wall_marker, now);
   }
 
-  auto id = 0;
+  auto id = planning_utils::bitShift(module_id_);
```
kyoichi-sugahara, line 226: @soblin 
memo for confirming my understanding and TODO below
It fixes the problem in the description but  `VirtualWallMarkerCreator` assume id starts from `0`(this is mistake in my opinion)
So deleting virtualwallmarker feature doesn't work properly with this change I thought.
Am I correct?
And I think I need to fix this [part](https://github.com/autowarefoundation/autoware.universe/blob/f944153e492a6e894fd7b02d818c48290d1f5f95/common/motion_utils/src/marker/marker_helper.cpp/#L135-L157) so that deletating feature can deal with the situation where id doesn't start from 0.

kyoichi-sugahara (approved):
LGTM

planning/behavior_velocity_planner/src/scene_module/intersection/debug.cpp
soblin: @kyoichi-sugahara I think there are 2 solutions:
- All behavior velocity planner module start the id from non-zero.
- add prefix to each behavior_velocity_planner modules like as follows:
  - {ns = "intersection1_stop_factor_text", id=0, name="intersection"}, {ns = "intersection1_stop_factor_text", id=1, name="intersection_occlusion"}
  - {ns = "intersection2_stop_factor_text", id=0, name="intersection"}, {ns = "intersection2_stop_factor_text", id=1, name="intersection_occlusion"}

I prefer the 2nd approach.

kyoichi-sugahara: @soblin 
I agree with this part
>I think all of these modules need to start the id from non-zero value to handle this error.
and about your proposed solution, if namespace can be changed for each intersection, then I think we can solve this problem with keeping that id always starts from 0 i guess?

soblin: I'll merge this for now

PR #3653, opened 2023-05-09
refactor(vehicle_cmd_gate): use lateral command while creating stop cmd

brkay54:
## Description
In the vehicle_cmd_gate package, if the vehicle disengages, it creates a stop command. However, while creating a stop command, it inserts current steering as lateral command and it causes unwanted behavior like in the video:

https://github.com/autowarefoundation/autoware.universe/assets/45468306/6dd86034-de2a-43f5-b972-d860e3f6d568

After disengaging, the vehicle sends current steering so the vehicle moves outside of the road.

## Tests performed

https://github.com/autowarefoundation/autoware.universe/assets/45468306/d3bdf096-2b22-43f6-9a69-6890fae75baf

Not applicable.

## Effects on system behavior

Not applicable.

### DIFF
--- a/control/vehicle_cmd_gate/src/vehicle_cmd_gate.cpp
+++ b/control/vehicle_cmd_gate/src/vehicle_cmd_gate.cpp
@@ -404,7 +404,7 @@ void VehicleCmdGate::publishControlCommands(const Commands & commands)
 
   // Check engage
   if (!is_engaged_) {
-    filtered_commands.control = createStopControlCmd();
+    filtered_commands.control = createStopControlCmd(filtered_commands.control.lateral);
   }
 
   // Check pause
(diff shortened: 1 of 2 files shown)


TakaHoribe:
@brkay54 The stop command can be used in several ways. It can use the target steering value when we want to send the stop signal (as of the function name: `createStopControlCmd`).
However, sending a target steering when the engage is off doesn't make sense.

My proposal to solve the issue reported above is 
1. "STOP" button in RViz should use the stop interface (https://github.com/autowarefoundation/autoware.universe/pull/3790), not engage interface.
2. When the stop interface is called, the target steering should be used at least while the vehicle is moving. (It can use the current steering angle when completely stopped.)

brkay54:
@TakaHoribe Great idea! I will create a new PR after https://github.com/autowarefoundation/autoware.universe/pull/3790 is merged! Thank you for the suggestion!

brkay54:
@TakaHoribe Thank you for your advice, in autoware_state_panel, stop button calls the service to change operation_mode. After operation_mode is changed, operation_mode_transition_manager changes engage to disengage. I want to ask a question about it. What actually means the engage and disengaging for Autoware? Does disengage mean that operation mode is stop? Maybe if the operation mode is stop, calling the moderate stop interface is a better solution. I always think that if autoware disengage, it does not involve controlling the vehicle, but in the current implementation it means stop the vehicle. It is little confusing for me sorry for that. What do you think?

PR #3957, opened 2023-06-13
feat(detected_object_validation): add BOUNDING_BOX interface to lanelet filter

scepter914:
## Description

For now, the lanelet filter handle only POLYGON shape type.
So this PR adapt to BOUNDING_BOX shape type.
This change can be used for radar detection.

## Tests performed

Tested by rosbag.

Screenshot from 2023-06-13 01-01-43

Purple objects is input objects and red objects is filtered objects.

## Effects on system behavior

This PR doesn't effect to POLYGON type, so doesn't effect to system behavior.

### DIFF
--- a/perception/detected_object_validation/src/object_lanelet_filter.cpp
+++ b/perception/detected_object_validation/src/object_lanelet_filter.cpp
@@ -82,6 +82,13 @@ void ObjectLaneletFilterNode::objectCallback(
     return;
   }
 
+  // if shape is BOUNDING_BOX, make foot print
+  for (auto & object : transformed_objects.objects) {
+    if (object.shape.type == object.shape.BOUNDING_BOX) {
+      object.shape.footprint = setFootprint(object);
+    }
+  }
+
   // calculate convex hull
   const auto convex_hull = getConvexHull(transformed_objects);
   // get intersected lanelets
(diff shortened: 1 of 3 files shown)


perception/detected_object_validation/src/object_lanelet_filter.cpp
```diff
@@ -82,6 +82,13 @@ void ObjectLaneletFilterNode::objectCallback(
     return;
   }
 
+  // if shape is BOUNDING_BOX, make foot print
+  for (auto & object : transformed_objects.objects) {
+    if (object.shape.type == object.shape.BOUNDING_BOX) {
+      object.shape.footprint = setFootprint(object);
+    }
+  }
+
```
yukkysaito, line 91: I think it would be better to define it as a separate structure instead of using the DetectedObject footprint all over again.

scepter914: I cannot get your comment properly, but I tried to fix at d0ef539.
If this change doesn't follow your thought, I ask you to present your thoughts in a detail.

yukkysaito: That is exactly what I intended. :+1: 
LGTM

miursh (approved):
LGTM

Issue #2652, opened 2023-01-13
DevOps Dojo: ROS Node Configuration (Phase 4)


kasperornmeck:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I've agreed with the maintainers that I can plan this task.

### Description

DevOps: Ros Node Configuration will address the topic of how ROS nodes are configured. Guidelines, documentation and changes to configuration files will be made, based on best-practices from cloud-native and software defined development methodologies.

Phase 1 - Motivation
- https://github.com/autowarefoundation/autoware.universe/issues/2631

Phase 2 - Definition
- https://github.com/autowarefoundation/autoware.universe/issues/2636

Phase 3 - Execution
- https://github.com/autowarefoundation/autoware.universe/issues/2651

Phase 4 - Integration
- https://github.com/autowarefoundation/autoware.universe/issues/2652

### Purpose

Phase 4 - Integration:
- scale out solution into system
- system wide test and verification
- education and documentation

### Possible approaches

TBD.

### Definition of done

TBD.

kaspermeck-arm:
https://github.com/autowarefoundation/autoware.universe/pull/3992

kaspermeck-arm:
Want to participate?
- Call for volunteers https://github.com/orgs/autowarefoundation/discussions/3636
- Discord https://discord.com/channels/953808765935816715/1131298526501027972

ahuazuipiaoliang:
@kaspermeck-arm I think the Phase 4 is updating the parameter configuration of ROS nodes? Am I correct?

kaspermeck-arm:
@ahuazuipiaoliang - yes, that's correct! Phase 4 (integration) includes the following:
- scale out solution into system
- system wide test and verification
- education and documentation

Are you interested in working with the node config?

kaspermeck-arm:
PR to update coding guidelines to correct schema template:
* https://github.com/autowarefoundation/autoware-documentation/pull/434

kaspermeck-arm:
@mitsudome-r to seek alignment in the AWF with regards to copyright
- configuration files don't have copyright notices
- schema files don't have copyright notices
- should we change copyright to only Autoware Foundation?

mitsudome-r:
For now we would like to stick with the following:
[...]

PR #3992, opened 2023-06-15
refactor(joy_controller): rework parameters

kaspermeck-arm:
## Description

Updated adding a schema according to new ROS node config coding guidelines. Implements https://github.com/autowarefoundation/autoware.universe/issues/2652.

## Tests performed

Built-in schema testing in VS Code. Proper JSON Schema and parameter validation tests into the CI are on being developed.

## Effects on system behavior


control/joy_controller/schema/joy_controller.schema.json
```diff
@@ -0,0 +1,112 @@
...
+            "velocity_gain": {
+              "type": "number",
+              "description": "Ratio to calculate velocity by acceleration",
+              "default": 3.0
+            },
+            "max_forward_velocity": {
+              "type": "number",
+              "description": "Absolute max velocity to go forward",
+              "default": 20.0,
+              "exclusiveMinimum": 0
```
kaspermeck-arm, line 65: Is "exlusiveMinimum" correct here? Can it be 0?

control/joy_controller/schema/joy_controller.schema.json
```diff
@@ -0,0 +1,112 @@
...
+            },
+            "max_backward_velocity": {
+              "type": "number",
+              "description": "Absolute max velocity to go backward",
+              "default": 3.0,
+              "exclusiveMinimum": 0
```
kaspermeck-arm, line 71: Is "exlusiveMinimum" correct here? Can it be 0?

control/joy_controller/schema/joy_controller.schema.json
```diff
@@ -0,0 +1,112 @@
...
+        "steering_angle_velocity": {
+          "type": "number",
+          "description": "Steering angle velocity for operation",
+          "default": 0.1
+        },
+        "accel_sensitivity": {
+          "type": "number",
+          "description": "Sensitivity to calculate acceleration for external API (commanded acceleration is pow(operation, 1 / sensitivity))",
+          "default": 1.0,
+          "exclusive": 0
```
kaspermeck-arm, line 45: To avoid division by 0.

control/joy_controller/schema/joy_controller.schema.json
```diff
@@ -0,0 +1,112 @@
...
+        },
+        "brake_sensitivity": {
+          "type": "number",
+          "description": "Sensitivity to calculate deceleration for external API (commanded acceleration is pow(operation, 1 / sensitivity))",
+          "default": 1.0,
+          "exclusive": 0
```
kaspermeck-arm, line 51: To avoid division by 0.

kaspermeck-arm (commented):
I've added some validators in the schema where I seemed fit. As I'm not familiar with `joy_controller` I've made some assumptions. Would be great to get feedback on the one's I've added and the one's I haven't added.

control/joy_controller/config/joy_controller.param.yaml
```diff
@@ -1,5 +1,6 @@
 /**:
   ros__parameters:
+    joy_type: "P65"
```
mitsudome-r, line 3: It seems like default parameter is set to `DS4` when it is launch from a [launch file](https://github.com/autowarefoundation/autoware.universe/blob/7241c83310e7d806ba1150274204d3a08859937d/control/joy_controller/launch/joy_controller.launch.xml#L2).
Maybe it's better to make the default value "DS4" to be consistent with the default value in launch file.

Also, since we decided to not have default values defined in places other than schema file, we might want to consider removing the default value set in launch file as well.

control/joy_controller/schema/joy_controller.schema.json
```diff
@@ -0,0 +1,112 @@
...
+  "title": "Parameters for Joy Controller node",
+  "type": "object",
+  "definitions": {
+    "joy_controller": {
+      "type": "object",
+      "properties": {
+        "joy_type": {
+          "type": "string",
+          "description": "Joy controller type",
+          "default": "P65",
```
mitsudome-r, line 12: Maybe it's better to set `DS4` to be consistent with original default type.

control/joy_controller/schema/joy_controller.schema.json
mitsudome-r: It looks like all *_sensitivity parameters are standardized to fit within 0.001 to 1.0 in the [code](https://github.com/autowarefoundation/autoware.universe/blob/7241c83310e7d806ba1150274204d3a08859937d/control/joy_controller/src/joy_controller/joy_controller_node.cpp#L142) already, but I think it's more user friendly that we still provide information in schema file as well.

In that case, we might want to add maximum value defined in schema file, too.
```suggestion
          "maximum": 1.0
          "exclusiveMinimum": 0.0
```

control/joy_controller/schema/joy_controller.schema.json
mitsudome-r: It looks like all *_sensitivity parameters are standardized to fit within 0.001 to 1.0 in the [code](https://github.com/autowarefoundation/autoware.universe/blob/7241c83310e7d806ba1150274204d3a08859937d/control/joy_controller/src/joy_controller/joy_controller_node.cpp#L142) already, but I think it's more user friendly that we still provide information in schema file as well if we consider creating README from schema files in the future.

In that case, we might want to add maximum value defined in schema file, too.
```suggestion
          "maximum": 1.0
          "exclusiveMinimum": 0.0
```

control/joy_controller/schema/joy_controller.schema.json
mitsudome-r: Same with max_forward_velocity

```suggestion
              "minimum": 0
```

mitsudome-r (approved):
@kaspermeck-arm 
Thanks for the PR, and sorry for the delay with the review.
It seems the package maintainers are busy with other tasks, I made a review instead.

I also had some feedback regarding the use of JSON Schema in general, which I might want to discuss in Open AD Kit WG.

kaspermeck-arm:
> 

@mitsudome-r - thanks for your review! Yes, adding `additionalProperties` and setting it to false is a good idea. Let's discuss in today's call.

kaspermeck-arm:
@mitsudome-r - with your new changes, can this PR be merged? Thanks!

mitsudome-r:
Hi Kasper, have you pushed your changes to the branch?
You have to push to your remote `kaspermeck-arm:dojo-ros-config-joy_controller` to be reflected to this PR.

control/joy_controller/schema/joy_controller.schema.json
kaspermeck-arm: Changing to "minimum".

control/joy_controller/config/joy_controller.param.yaml
mitsudome-r: I think DS4 would be better since we usually run the node from the launch file.

mitsudome-r:
@kaspermeck-arm Thanks for the update. I think we can merge it once https://github.com/autowarefoundation/autoware.universe/pull/3992#discussion_r1269270494 is resolved.

PR #4149, opened 2023-07-03
fix(traffic_light map_based detector): vibration offset computation

knzo25:
## Description
As mentioned before internally, I think that the computations regarding the offsets produced by the `vibrations` are incorrect. The effect is small when the ROIs are near the optical center but increases as the ROIs get away from it.

## Tests performed
Validated using the # 4 data from 2023/04/17

## Effects on system behavior

Not applicable.

### DIFF
--- a/perception/traffic_light_map_based_detector/src/node.cpp
+++ b/perception/traffic_light_map_based_detector/src/node.cpp
@@ -286,15 +286,25 @@ bool MapBasedDetector::getTrafficLightRoi(
     tf2::Vector3 map2tl = getTrafficLightTopLeft(traffic_light);
     tf2::Vector3 camera2tl = tf_map2camera.inverse() * map2tl;
     // max vibration
-    const double max_vibration_x =
-      std::sin(config.max_vibration_yaw * 0.5) * camera2tl.z() + config.max_vibration_width * 0.5;
-    const double max_vibration_y = std::sin(config.max_vibration_pitch * 0.5) * camera2tl.z() +
-                                   config.max_vibration_height * 0.5;
-    const double max_vibration_z = config.max_vibration_depth * 0.5;
+    double camera2tl_xz_distance = tf2::Vector3(camera2tl.getX(), 0.0, camera2tl.getZ()).length();
+    double camera2tl_yz_distance = tf2::Vector3(0.0, camera2tl.getY(), camera2tl.getZ()).length();
+
+    tf2::Vector3 tlr_z_basis =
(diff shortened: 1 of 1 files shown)


yukkysaito:
Could you please provide slides or something to explain the computation?

knzo25:
Sure thing.
[TIER IV INTERNAL LINK](https://docs.google.com/presentation/d/1TgGQU_nU5TGuvlCdT63SVzbVkKdF9qyh8O3LC4iE0RU/edit?usp=sharing)

knzo25:
cc: @Mingyu1991 @yukkysaito

yukkysaito:
Sorry for late reply.
I'll review today :pray:

yukkysaito:
I don't know how to read the figure, so please tell me. :pray: 
Where is it on the image to be projected?
If the blue line is where the image is projected to, then why not leave it as it is now?
image

knzo25:
@yukkysaito 
It is not "good" to represent the camera projection plane at an arbitrary such as in the blue line (unless it is the same distance for multiple examples) since it can be easy to arrive to misleading conclusions. Additionally, I think is better to understand this problem in the camera coordinate system rather than in the image space (forgetting the projection model for a while)

The current way to compute things it is not wrong per se, it is just not what it says it seeks to do.
A way to see the problem clearly is considering a very wide FOV and the  TLR in the border of the image. In that case:
 - The z component is almost null compared with the x component which is very large (in camera coordinates)
 - With the current way to compute things no matter how large we select the vibration yaw parameter the delta x does not really change. `delta_x = sin(vibration_yaw) * z`. Since `z << x`, no choice of `vibration_yaw` has an relative effect on `x`.

With the current cameras it is not really a problem, but as we move to higher resolutions (larger distances) and wider FOVs, the effect will be less easy to ignore. 
It is always a choice to not adopt this way of computing the deltas since none of our products are really in need of it right now and adding this fix may cause concerns of man-hours (I don't think that is the case, but even so). 
However, I would like at least for the people involved to be aware of the limitations of the current way we compute these deltas so that is we have issues in the future, we know what and how to address it.

kminoda:
@knzo25 What is the status of this PR?

knzo25:
@kminoda 
The relevant people don't think it is relevant so it being left-as-is for now (we are starting to use 120deg lenses so it may become relevant in the future)

kminoda:
@knzo25 Thanks! Let me convert this to a draft but feel free to reopen if necessary.

knzo25:
@YoshiRi @technolojin 
I will be closing this PR in a few weeks.
But please keep in mind that the way the ROIs are computed will not work as good in the corners as in the center of the images for wide cameras

PR #3293, opened 2023-04-06
perf(ring_outlier_filter): a cache friendly impl

VRichardJP:
## Description

This is a "cache friendly" reimplementation of ring outlier filter. See https://github.com/autowarefoundation/autoware.universe/issues/3269 for more details.

The code is based on @sykwer WiP improvements (see https://github.com/autowarefoundation/autoware.universe/pull/3014), which I rewrote to improve memory access (hopefully).

This implementation produce slightly different output than the original implementation:
- output points are in same order than input. In the original implementation, points are reordered and copied by ring.
- Fixes https://github.com/autowarefoundation/autoware.universe/issues/3217
- Fixes https://github.com/autowarefoundation/autoware.universe/issues/3218

## Related links

https://github.com/autowarefoundation/autoware.universe/issues/3269
https://github.com/autowarefoundation/autoware.universe/pull/3014

## Tests performed

Note: these tests are fairly old and refer to older implementation of ring outlier filters. See recent comments for comparisons between the current implementation and this PR.

Simple test on a few random rosbags. Plot data is the output of `/ring_outlier_filter/debug/processing_time_ms`. In blue the current implementation (with hashmap), in orange the proposed implementation (see the more recent comments for more accurate benchmarks):

- On VLP-16 data (~20000 points per message):
image

- On VLP-32 data (~50000 points per message):
image

In both cases, my implementation is roughly 3 times faster than the original one. Note that I am not using the TLSF allocator.

This is of course just a very simple benchmark and I will need to test more. In particular, I want to:
- compare cache performance with old implementation
- compare performance with @sykwer implementation (after his changes have been merged)
- compare performance on vehicle, with all autoware modules running

I don't necessarily expect this implementation to shine when the filter is the only process running on the machine. However I expect it will perform way better when the machine is stressed.

## Pre-review checklist for the PR author

The PR author **must** check the checkboxes below when creating the PR.

- [x] I've confirmed the [contribution guidelines].
- [x] The PR follows the [pull request guidelines].

## In-review checklist for the PR reviewers

The PR reviewers **must** check the checkboxes below before approval.

- [ ] The PR follows the [pull request guidelines].
- [ ] The PR has been properly tested.
- [ ] The PR has been reviewed by the code owners.

## Post-review checklist for the PR author

The PR author **must** check the checkboxes below before merging.

- [ ] There are no open discussions or they are tracked via tickets.
- [ ] The PR is ready for merge.

After all checkboxes are checked, anyone who has write access can merge the PR.

[...]

Issue #3269, opened 2023-04-04
`ring_outlier_filter` blows up my cache (and yours too)


VRichardJP:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I'm convinced that this is not my fault but a bug.

### Description

Current `ring_outlier_filter` implementation is slow. I have not run a precise benchmark, but this filter runs significantly slower than any other filter.
During a recent test drive with our demo vehicle, I observed that when `ring_outlier_filter` is disabled, htop CPU usage is reduced by 20% (!)

I am pretty sure it is a known problem, and @sykwer is working hard on improving the current implementation.
His changes already improve the performance a lot, but I have the feeling his approach does not tackle the true cause of the bad performance, so I think it is worth opening a separate issue.

Let's first look at how the current algorithm is implemented. Roughly it looks like this:
1. Iterate over the input cloud and group point indices by ring
2. For each ring:
2.1. iterate over the ring points, and group points belonging to the same "walk"
3.1. when a walk is closed, copy indexed points to the output cloud if the walk is long enough.

There is nothing wrong with the algorithm in itself. It looks fairly efficient.
It has a big problem though: it is not adapted to the data it is supposed to process.

Let me explain.
By design, LIDAR sensors naturally collect data in "azimut-major order". So unless one of the previous preprocessing node changes that order, the input cloud data of `ring_outlier_filter` always looks like this:
```
p[0] = {ring = 0; azimut = 0}
p[1] = {ring = 1; azimut = 0}
p[2] = {ring = 2; azimut = 0}
p[3] = {ring = 3; azimut = 0}
```

For example here, I have printed the first 100 point indices of ring 0 (from one rosbag of mine):
```
[...]

Issue #3218, opened 2023-03-30
`ring_outlier_filter` ignores ring's "end" point and "start" point are neighbors


VRichardJP:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I'm convinced that this is not my fault but a bug.

### Description

Copied from https://github.com/autowarefoundation/autoware.universe/pull/3014#discussion_r1128802859

In the `ring_outlier_filter` implementation, points that belong to a same ring are stored in a buffer (in azimut order). The filter explores the buffer from start to end, and remove isolated points (i.e. points with no close neighbor). However the current implementation ignores the buffer is circular (the "first" and "last" points of the ring are connected). As a consequence, the filter may delete valid points and create "holes" in the output data.

For example, let's imagine this ring:
```
ring : xxxx..xxxxxxxxxxx..xxxx.xx
index: abcdefghijklmnopqrstuvwxyz
``` 
Where chains of `x` represent points from the same "walk" (close to each other), and `.` the noisy points ment to be filtered (no close neighbor). The index alphabet is just for the explanation. So in the situation here, there are 3 chains:
1. from 'g' to 'q'
2. from 't' to 'w'
3. from 'y' to 'd' (wrapped around)

However in the current implementation, we start at index 0 ('a') and the start/end of the ring are not considered connected. So we end up finding 3 chains, but different:
1. from 'a' to 'd'
2. from 'g' to 'q'
3. from 't' to 'w'

The algorithm would consider the walk 'y' -> 'z', but as it is too short (less than 4 points) the points would be filtered out.

With the default parameters `num_points_threshold_ = 4`, up to 6 points could be missing because of that (3 points at the beginning, 3 points at the end). When using bigger `num_points_threshold_` value, the filter would sometimes produce some big blinspot in the data.

### Expected behavior

In a ring, start and end points are connected, so they may belong to the same walk.

### Actual behavior

end-start connection ignored. Some valid points may be wrongfully deleted.

### Steps to reproduce

N/A

Issue #3217, opened 2023-03-30
`ring_outlier_filter` skips last input point


VRichardJP:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I'm convinced that this is not my fault but a bug.

### Description

Copied from https://github.com/autowarefoundation/autoware.universe/pull/3014#discussion_r1128788929

https://github.com/autowarefoundation/autoware.universe/blob/cb103ec9c8149fffcf2a665ca37b69fe465e0a61/sensing/pointcloud_preprocessor/src/outlier_filter/ring_outlier_filter_nodelet.cpp#L92-L95

I think we are skipping the last point with this logic.
For example, let's imagine a perfect circle ring with N points.

When we first enter the first loop we have :
```
idx = 0
walk_first_idx = 0
walk_last_idx = idx = 0
current_data_idx = indices[0]
next_data_idx= indices[1]
```

The ring is a perfect circle, so all the points are part of the same walk. So we always continue here:

https://github.com/autowarefoundation/autoware.universe/blob/cb103ec9c8149fffcf2a665ca37b69fe465e0a61/sensing/pointcloud_preprocessor/src/outlier_filter/ring_outlier_filter_nodelet.cpp#L111-L116

On the last loop we have:
```
idx = N-2
walk_first_idx = 0
walk_last_idx = idx = N-2
current_data_idx = indices[N-2]
next_data_idx= indices[N-1]
```

So after the loop, we have these lines:

https://github.com/autowarefoundation/autoware.universe/blob/cb103ec9c8149fffcf2a665ca37b69fe465e0a61/sensing/pointcloud_preprocessor/src/outlier_filter/ring_outlier_filter_nodelet.cpp#L133-L141

which basically copy points from `walk_first_idx = 0` to `walk_last_idx = N-2`, thus skipping the last point.

### Expected behavior

All points are processed

### Actual behavior

Last point missing :(

### Steps to reproduce

N/A

VRichardJP:
Problem still exist in main branch:

https://github.com/autowarefoundation/autoware.universe/blob/36cad1ced0f883d1f500c9d9b45cd1f83a4e2c34/sensing/pointcloud_preprocessor/src/outlier_filter/ring_outlier_filter_nodelet.cpp#L100-L103

PR #4185, opened 2023-07-06
perf(ring_outlier_filter): a cache friendly impl (continuation of VRichardJP's work)

mojomex:
## Description

This is the completion of @VRichardJP's PR #3293 for a cache friendly implementation of ring_outlier_filter.
The implementation goes through the point cloud (which is in azimuth-major order) sequentially and keeps track of walks in the LiDAR rings concurrently. See #3269 for a description of the problem solved here.

## Related links

#3293 
#3269 
#3014 

## Tests performed

See the posts below.

### Summary by CodeRabbit

Release Notes:

- New Feature: The `ring_outlier_filter` implementation has undergone changes to its core parameters. The `max_points_num_per_ring` parameter has been removed, and the description for `max_rings_num` is missing.
- Refactor: The `isCluster` function has been removed from the `RingOutlierFilterComponent` class, indicating a change or relocation of the logic for determining clusters.
- Documentation: The package manifest file (`package.xml`) has been updated to include a new dependency on `range-v3`.

> "Code changes dance,
> Parameters take a chance.
> Clusters shift their stance,
> With dependencies enhanced."

### DIFF
--- a/sensing/pointcloud_preprocessor/src/outlier_filter/ring_outlier_filter_nodelet.cpp
+++ b/sensing/pointcloud_preprocessor/src/outlier_filter/ring_outlier_filter_nodelet.cpp
@@ -327,6 +327,7 @@ void RingOutlierFilterComponent::faster_filter(
       output_size += sizeof(PointXYZI);
     }
   }
+  output.data.resize(output_size);
 
   // Note that `input->header.frame_id` is data before converted when `transform_info.need_transform
   // == true`
--- a/sensing/pointcloud_preprocessor/src/outlier_filter/ring_outlier_filter_nodelet.cpp
+++ b/sensing/pointcloud_preprocessor/src/outlier_filter/ring_outlier_filter_nodelet.cpp
@@ -160,11 +160,6 @@ void RingOutlierFilterComponent::faster_filter(
   };
 
   // helper functions
-  double d_alloc_vectors;
-  double d_build_walks;
-  double d_fuse_walks;
-  double d_check_clusters;
-  double d_write_outputs;
 
   // check if walk is a valid cluster
   const float object_length_threshold2 = object_length_threshold_ * object_length_threshold_;
@@ -194,8 +189,6 @@ void RingOutlierFilterComponent::faster_filter(
              azimuth_diff < 100.f;
     };
 
-  stop_watch_ptr_->toc("cyclic_time", true);
-
   // tmp vectors to keep track of walk/ring state while processing points in order (cache efficient)
   std::vector<WalkInfo> walks;            // all walks
   std::vector<RingWalkInfo> rings;        // info for each ring
@@ -209,9 +202,7 @@ void RingOutlierFilterComponent::faster_filter(
   rings.resize(max_rings_num_, RingWalkInfo{nullptr, nullptr});
   // points are initially associated to no walk
   points_walk_id.resize(input->width * input->height, -1UL);
-  walks_cluster_status.reserve(input->width * input->height); // In the worst case, this could grow to the number of input points
-
-  d_alloc_vectors = stop_watch_ptr_->toc("cyclic_time", true);
+  walks_cluster_status.reserve(max_rings_num_ * 2); // In the worst case, this could grow to the number of input points
 
   int invalid_ring_count = 0;
 
@@ -267,8 +258,6 @@ void RingOutlierFilterComponent::faster_filter(
     }
   }
 
-  d_build_walks = stop_watch_ptr_->toc("cyclic_time", true);
-
   // So far, we have processed ring points as if rings were not circular. Of course, the last and
   // first points of a ring could totally be part of the same walk. When such thing happens, we need
   // to merge the two walks
@@ -307,10 +296,6 @@ void RingOutlierFilterComponent::faster_filter(
       walks_cluster_status.at(last_walk.id) = isCluster(last_walk);
     }
   }
-
-  d_fuse_walks = stop_watch_ptr_->toc("cyclic_time", true);
-
-  d_check_clusters = stop_watch_ptr_->toc("cyclic_time", true);
 
   // finally copy points
   output.point_step = sizeof(PointXYZI);
@@ -387,16 +372,6 @@ void RingOutlierFilterComponent::faster_filter(
       get_logger(), "%d points had ring index over max_rings_num (%d) and have been ignored.",
       invalid_ring_count, max_rings_num_);
   }
-
-  d_write_outputs = stop_watch_ptr_->toc("cyclic_time", true);
-
-  //double d_alloc_vectors;
-  //double d_build_walks;
-  //double d_fuse_walks;
-  //double d_check_clusters;
-  //double d_write_outputs;
-  RCLCPP_INFO(get_logger(), "AAA {'av': %f, 'bw': %f, 'fw': %f, 'cc': %f, 'wo': %f, 'nw': %zu, 'in': %u, 'out': %u} BBB", d_alloc_vectors, d_build_walks,
-              d_fuse_walks, d_check_clusters, d_write_outputs, walks.size(), input->width * input->height, output.width);
 
   // add processing time for debug
   if (debug_publisher_) {
--- a/sensing/pointcloud_preprocessor/src/outlier_filter/ring_outlier_filter_nodelet.cpp
+++ b/sensing/pointcloud_preprocessor/src/outlier_filter/ring_outlier_filter_nodelet.cpp
@@ -76,28 +76,28 @@ void RingOutlierFilterComponent::faster_filter(
     RCLCPP_ERROR(get_logger(), "input cloud does not contain 'intensity' data field");
     return;
   }
-  const auto intensity_offset = *maybe_intensity_offset;
+  const auto intensity_offset = input->fields.at(*maybe_intensity_offset).offset;
 
   auto maybe_ring_offset = utils::getRingOffset(*input);
   if (!maybe_ring_offset) {
     RCLCPP_ERROR(get_logger(), "input cloud does not contain 'ring' data field");
     return;
   }
-  const auto ring_offset = *maybe_ring_offset;
+  const auto ring_offset = input->fields.at(*maybe_ring_offset).offset;
 
   auto maybe_azimuth_offset = utils::getAzimuthOffset(*input);
   if (!maybe_azimuth_offset) {
     RCLCPP_ERROR(get_logger(), "input cloud does not contain 'azimuth' data field");
     return;
   }
-  const auto azimuth_offset = *maybe_azimuth_offset;
+  const auto azimuth_offset = input->fields.at(*maybe_azimuth_offset).offset;
 
   auto maybe_distance_offset = utils::getDistanceOffset(*input);
   if (!maybe_distance_offset) {
     RCLCPP_ERROR(get_logger(), "input cloud does not contain 'distance' data field");
     return;
   }
-  const auto distance_offset = *maybe_distance_offset;
+  const auto distance_offset = input->fields.at(*maybe_distance_offset).offset;
 
   // extract intensity from point raw data
   auto getPointIntensity = [=](const unsigned char * raw_p) {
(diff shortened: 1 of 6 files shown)


mojomex:
I have since tested the old and new implementation on VLP16 pointclouds obtained from AWSIM as well as VLS128 pointclouds captured from the actual sensor.

On a laptop with Intel Core i7-12700H, the cache is not the bottleneck and the old and new implementation perform almost the same, the new one performing slightly slower on VLP16 pointclouds (see plots).

image
|Run	|VLS128 - patch	|VLS128 - before	|VLP16 - patch	|VLP16 - before|
|-|-|-|-|-|
|mean	|8.36105	|8.531297	|1.121155	|0.775264|
|std	|1.976964	|2.080717	|0.246097	|0.174693|

It can however be seen that the cache miss rate (here for VLS128 pointclouds) _does_ go down by ~32%.

Screenshot from 2023-07-10 19-09-09

The ring_outlier_node has been run on an isolated core with fixed 4GHz frequency and turbo boost disabled, timing measurements have been done the same as in #3293 through the `debug/processing_time_ms` topic.

As a next step, I will run this on a more limited machine to verify whether the caching behavior makes a significant difference.

mojomex:
It took some time to set up valid performance benchmarks on my new laptop but here they are.

All benchmarks are done on real-world VLS128 data (ca. 60s long), on an isolated core of my machine, with turbo boost disabled and all cores pinned to the highest non-boost frequency (2.3 GHz).
All benchmarks have been repeated 3 times, the combined measurements are seen below.

My machine:
CPU: Intel Core i7-12700H
L1d: 544 KiB (14 instances)
L1i: 704 KiB (14 instances)
L2: 11.5 MiB (8 instances)
L3: 24 MiB (1 instance)
RAM: 32 GB

I have benchmarked the following implementations:
|Label|Implementation|Results|
|-|-|-|
|before|Original implementation|slow|
|before w/ cos|Original implementation, but isCluster is modified to use the law of cosines|equally slow (within measurement tolerance)|
|rich|@VRichardJP's implementation (I fixed the remaining errors)|fast (~25% speedup)|
|rich w/o cos|Richard's implementation, but with the euclidean isCluster from before (needs to store XYZ for each ring's current walk start and end point|a little less fast (~15% speedup)|

This shows that the trigonometry done in isCluster's law of cosines computation is well worth the extra speedup achieved by the emptier cache.

image

Richard's implementation becomes faster where the original becomes slower; to investigate this, I plotted the number of input/remaining output points over time. The original implementation has significant overhead for copying points into the output because it reorders the points in ring-major (instead of the incoming azimuth-major) order.

image

As shown in Richard's pull request, the number of cache misses significantly decreases (even with a big cache), in my case by around 35 %.

image

I will test this tomorrow on VLP16 data to confirm there is no (significant) performance regression, then I'm good to mark this as ready to merge.

VRichardJP:
@mojomex Thank you so much for taking over!

If I understand your second graph, the clouds are more dense and contain fewer isolated points from around 20s in the rosbag. The original implementation performance gets worse because cache miss is the bottleneck, and more points means increased cache miss count. However I don't really understand why my implementation would be much faster. Is it because with this implementation the bottleneck is not the number of points, but rather the number/length of walks?

For example, if we randomly move X% of the input points, so they become noisy points to remove, is there a threshold the original implementation will perform faster? I don't know how noisy lidar data can become under heavy condition (e.g. heavy rain, fog, etc)

mojomex:
Mystery solved, your intuition that the number of walks is the cause of this timing behavior was correct.

The upper plot shows the timing of each of the 5 phases of the function (stacked, in total they equal the whole processing time).
In black, the number of walks generated is overlaid.
I have highlighted 3 sections (startup, many walks, few walks). The 5 phases are shown as box plots for each of those sections.
Dotted black lines show the baseline mean (the last section, few walks).

image

It can be seen that in the "many walks" section, the _build walks_ and _check clusters_ phase are significantly slower than in the last section.

This makes sense when looking at the code:

sensing/pointcloud_preprocessor/src/outlier_filter/ring_outlier_filter_nodelet.cpp
```diff
@@ -61,122 +67,269 @@ void RingOutlierFilterComponent::faster_filter(
...
+    auto & walk = walks[ring.walk_idx];
+    if (isSameWalk(curr_distance, curr_azimuth, ring.prev_distance, ring.prev_azimuth)) {
+      // current point is part of previous walk
+      walk.num_points += 1;
+      walk.last_point_distance = curr_distance;
+      walk.last_point_azimuth = curr_azimuth;
+      point_walk_idx = ring.walk_idx;
+    } else {
+      // previous walk is finished, start a new one
+      walks.push_back(WalkInfo{1, curr_distance, curr_azimuth, curr_distance, curr_azimuth});
```
mojomex, line 243: We reserve `max_rings_num_` walks in the `walks` vector, so it needs to be resized multiple times when appending on the order of 16000 walks, as is the case in the "many walks" section.

sensing/pointcloud_preprocessor/src/outlier_filter/ring_outlier_filter_nodelet.cpp
```diff
@@ -61,122 +67,269 @@ void RingOutlierFilterComponent::faster_filter(
...
 
-      const float & current_distance =
-        *reinterpret_cast<const float *>(&input->data[current_data_idx + distance_offset]);
-      const float & next_distance =
-        *reinterpret_cast<const float *>(&input->data[next_data_idx + distance_offset]);
+  // segregate walks by cluster (for fast lookup)
+  std::vector<bool> walk_is_cluster;
+  walk_is_cluster.reserve(walks.size());
+  for (const auto & walk_info : walks) {
+    walk_is_cluster.push_back(isCluster(walk_info));
```
mojomex, line 283: This section's runtime of course scales linearly with the number of walks.

sensing/pointcloud_preprocessor/src/outlier_filter/ring_outlier_filter_nodelet.cpp
VRichardJP: Indeed, I only allocated the strict minimum (1 walk per ring), but a bigger value might be wiser.

mojomex:
I have made the walks array fixed-size now, storing only the first and previous walk of any given ring.
The `points_walk_idx` array is now the `points_alk_id` array, storing not an index into `walks` but unique walk ID, which is used as an index into the `walks_cluster_status` boolean array, containing the results of `isCluster`. Because walks are immediately discarded when they are finished, `isCluster` is now immediately called when that happens.

The points are then checked against the cluster status of their assigned walk ID, without needing the actual walk struct to exist.
Notably, the `walks_cluster_status` array can also blow up to at most `input->width` elements (as the `walks` array could before). But, since it only stores bools instead of structs with a few values in them, there is a slight speedup (see *purple (new)* vs. *green (proposed by Richard)*).

I have tried reserving a large number of elements for the `walks_cluster_status` array, it does not make a measurable difference to letting it grow.

This is admittedly a small improvement but imho it did not make the code less readable and it also improves memory usage.

image

yukkysaito:
@mojomex Thank you for this PR :+1: 
Is it possible to split the function and abstract it since it is a bit large?

sensing/pointcloud_preprocessor/src/outlier_filter/ring_outlier_filter_nodelet.cpp
```diff
@@ -61,122 +67,286 @@ void RingOutlierFilterComponent::faster_filter(
...
+    const uint16_t ring_idx = getPointRing(raw_p.data());
+    const float curr_azimuth = getPointAzimuth(raw_p.data());
+    const float curr_distance = getPointDistance(raw_p.data());
+
+    if (ring_idx >= max_rings_num_) {
+      // Either the data is corrupted or max_rings_num_ is not set correctly
+      // Note: point_walk_id == -1 so the point will be filtered out
+      invalid_ring_count += 1;
+      continue;
+    }
```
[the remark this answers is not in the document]
mojomex: I have now changed the implementation so we don't have this unsafety anymore:
`WalkInfo` structs are now directly stored in `RingWalkInfo` (because we only need to store 2 walks per ring).
This also resulted in another significant performance gain (see comment below).

mojomex:
@yukkysaito I have simplified the code (and by effect also further improved performance).

I have removed the functions for finding PointCloud2 field offsets and for accessing those fields, that Richard added.
Because the `ring_outlier_filter` docs specify the expected input pointcloud format, it's reasonable to go with the old approach (before Richard's PR). That approach expected all the fields to exist, thus making the safe accessors unnecesary.

I also removed the `walks` vector as we are storing a constant 2 walks per ring now. I am storing 2 `WalkInfo` structs directly in the `RingWalkInfo` struct now. This also results in a performance gain (see the red points in the chart below).

We are now at ~33% faster performance compared to the original implementation.

image

yukkysaito (approved):
LGTM

VRichardJP:
I would advice against blindly assuming the data layout. This has caused this kind of issues in the past: https://github.com/autowarefoundation/autoware.universe/pull/2618

I had opened https://github.com/autowarefoundation/autoware.universe/issues/3318 to deal with that, but at the time ring_outlier_filter was still under rework. 

For example, something like this would be nice:
```c++
{
  bool missing_field = false;
  missing_field |= !autoware_point_types::hasField<autoware_point_types::PointIndex::Ring>(input);
  missing_field |= !autoware_point_types::hasField<autoware_point_types::PointIndex::Azimuth>(input);
  //...

  if(missing_field) {
    // Print err, etc
  }
```

mojomex:
Thanks for the feedback!
I implemented a generic accessor method that looks up the field index and safely reads the offset.
Errors are logged and the function exists if any of the required fields is not there.
The field's data type is also checked against the expected one.

mojomex:
The CI pipeline finally ran through :tada: I think all discussion points have been addressed and all improvements implemented and tested. I think it is ready to merge @yukkysaito

yukkysaito (approved):
LGTM
@miursh May I ask you to review it as well?

yukkysaito:
@mojomex Can I try to Enable review by ChatGPT on a trial basis? It may be a little noisy.

mojomex:
@yukkysaito Sure!

mojomex:
I forgot to add this to the PR description earlier:
This PR fixes #3269, fixes #3218, fixes #3217 and closes #3293

mojomex:
> @mojomex Can I try to Enable review by ChatGPT on a trial basis? It may be a little noisy.

@yukkisaito Seems like it has no valid API key:

```
Warning: Skipped: failed to create summary bot, please check your openai_api_key: Error: Unable to initialize the OpenAI API, both 'OPENAI_API_KEY' environment variable are not available, backtrace: Error: Unable to initialize the OpenAI API, both 'OPENAI_API_KEY' environment variable are not available
```

yukkysaito:
@mojomex 
>Seems like it has no valid API key:

https://github.com/autowarefoundation/autoware.universe/pull/4515

It seemed that the secret key could not be referenced from the forked repository, so it was not working properly. This PR solved it.

I think it is working correctly now.

mojomex:
The CI tasks for this PR seem not to be picked up by the self-hosted runners, they have been queuing for almost 3 weeks.
Is there something wrong with the configuration in my fork or this PR's tags or is it just bad luck?
I would really appreciate if @yukkysaito or @miursh could take a look :bow:

miursh:
@mojomex You can merge this PR after the CI (build-and-test-differential) passes. https://github.com/autowarefoundation/autoware.universe/actions/runs/5945676273/job/16125118660?pr=4185

mojomex:
> @mojomex You can merge this PR after the CI (build-and-test-differential) passes. https://github.com/autowarefoundation/autoware.universe/actions/runs/5945676273/job/16125118660?pr=4185

Thank you for re-adding the label! I don't have write access to this repo, would you mind merging it instead, @miursh?

xmfcx:
This was reverted, to reproduce the problem, @miursh -san shared the data:

- https://github.com/autowarefoundation/autoware.universe/pull/6148#issuecomment-1908006849

Issue #2048, opened 2022-10-11
[map_height_fitter] Improve height estimation logic


VRichardJP:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I've agreed with the maintainers that I can plan this task.

### Description

With the current implementation of [`get_ground_height`](https://github.com/autowarefoundation/autoware.universe/blob/4e080d2c05da5b1a0356c1ec16ae3c77676914f7/localization/pose_initializer/src/map_height_fitter/map_height_fitter_core.cpp#L46) , when the user sets the initial pose using Rviz, if the initial pose is not within 1 meter of any map point, the `map_height_fitter` node will fail to estimate its height. Most likely, the initial localization will be wrong.

That situation typically occurs when the vehicle is located in some garage or building, whose inside has not mapped. Like here:
 
![IMG_20221011_163404](https://user-images.githubusercontent.com/18645627/195047251-551a5319-2b88-4203-8007-e96aba93fbca.jpg)

![IMG_20221011_163505](https://user-images.githubusercontent.com/18645627/195047264-656b73a8-3414-4917-91e0-85c00e81f067.jpg)

Likely, the issue would also occur if the 3D map has no ground (I don't know if it is a thing)

### Purpose

Improve the reliability of the height estimator.

### Possible approaches

Obviously, the hardcoded 1.0 meter radius search makes the implementation not very flexible. 

I think that simply increasing the radius search would have some undesirable side effects.

One simple improvement would be to consider all point within r+1.0 radius, where r is the distance to the closest point. This way we make sure we always find a bunch of points, yet we don't look unecessarily far.

The only downside I see is that it takes slightly more time, since we explore the map twice (once to find the closest point, another time to find min height within radius r+1.0)

Any other suggestion?

### Definition of done

Pose height estimation give a "decent" estimation even when there is no map point very close.

YamatoAndo:
> One simple improvement would be to consider all point within r+1.0 radius, where r is the distance to the closest point. This way we make sure we always find a bunch of points, yet we don't look unecessarily far.

I think it is a good idea!

YamatoAndo:
> The only downside I see is that it takes slightly more time, since we explore the map twice (once to find the closest point, another time to find min height within radius r+1.0)

[...]

PR #4199, opened 2023-07-07
fix(map_height_fitter): return failure service response when ground fitting failed

KYabuuchi:
## Description
If ground fitting fails, return the service immediately as a failure.

Until now, when there was no point cloud map around the specified initial position, the given height (**basically z=0**) was returned as is.  If the correct height is not given, the NDT will fail to estimate. 

## Tests performed

Two tests were performed using logging_simulator and sample rosbag.

### How to test for case 1

```
# 1: Stop rosbag to prevent overwriting with correct gnss.
# 2: Set wrong gnss pose 
 ros2 topic pub /sensing/gnss/pose_with_covariance geometry_msgs/msg/PoseWithCovarianceStamped "{pose: {pose: {position: {x: 89435, y: 42047}}}}"
# 3: Call the initialization service
 ros2 service call /api/localization/initialize autoware_adapi_v1_msgs/srv/InitializeLocalization
```

After the above steps, a error message appears in the console indicating that ground fitting has failed.
Also, the localization status becomes `UNINITIALIZED`.

### How to test for case 2

```
# 1: Perform "2D Pose Estimate" in the area without a point cloud map as shown in the following figure.
```

After the above steps, a warning message appears in the console indicating that ground fitting has failed. 
The localization status does not change.

## Effects on system behavior

If ground fitting fails, the initial position estimation is aborted and a service failure is returned.
(Until now, initialization continued at the given elevation even if ground fitting failed.)


map/map_height_fitter/src/map_height_fitter.cpp
```diff
@@ -186,7 +192,14 @@ std::optional<Point> MapHeightFitter::Impl::fit(const Point & position, const st
     tf2::Transform transform{tf2::Quaternion{}, tf2::Vector3{}};
     tf2::fromMsg(stamped.transform, transform);
     point = transform * point;
-    point.setZ(get_ground_height(point));
+
+    //
```
isamu-takagi, line 196: Probably forgot to remove the comment.

map/map_height_fitter/src/map_height_fitter.cpp
```diff
@@ -186,7 +192,14 @@ std::optional<Point> MapHeightFitter::Impl::fit(const Point & position, const st
...
+    auto point_optional = get_ground_height(point);
```
isamu-takagi, line 197: This value can be `const` and is not a `point`.

isamu-takagi:
For case 2, no error when `enable_partial_load: false`. I think we have to change `get_ground_height`.
Screenshot from 2023-07-10 19-47-07

map/map_height_fitter/src/map_height_fitter.cpp
KYabuuchi: Thank you! I removed it in 92be65d

map/map_height_fitter/src/map_height_fitter.cpp
KYabuuchi: I made it const and modified the name in  [92be65d](https://github.com/autowarefoundation/autoware.universe/pull/4199/commits/92be65d6469a65c9de6546a861b84e4633e89845)

KYabuuchi:
I have made a change so that `get_ground_height()` will fail if there are no points within a 50-meter (`max_search_radius`) range when determining the range of point cloud to consider.  Please see the line 154 to 157 the following link.

https://github.com/autowarefoundation/autoware.universe/blob/0171abbd1cdad18e0cc20dc47db0838a56ba783a/map/map_height_fitter/src/map_height_fitter.cpp#L140-L157

KYabuuchi:
@isamu-takagi With the above changes, I believe the reported issue has been resolved.

> For case 2, no error when `enable_partial_load: false`. I think we have to change `get_ground_height`.

KYabuuchi:
I found the related issue https://github.com/autowarefoundation/autoware.universe/issues/2048
@VRichardJP  If you have any concerns about this PR, please let us know. :+1:

isamu-takagi (approved):

KYabuuchi:
I have noticed that there is a possibility that `YabLoc` and `Eagleye`, which are pose_estimator does not rely on point cloud maps, may not be able to complete initial pose estimation. :cry:

I will temporarily mark this PR as a draft. I will also consider an implementation that works with a general pose_estimator.

PR #4416, opened 2023-07-27
perf(path_smoother): add ReplanChecker to decrease computation cost

maxime-clem:
## Description

Copy the `ReplanChecker` from the `obstacle_avoidance_planner` in order to improve the performance of the `path_smoother`.

Also needs launch PR: https://github.com/autowarefoundation/autoware_launch/pull/482

## Tests performed

Psim.
CPU usage before: 200-300%.
CPU usage after: 30-60%.

## Effects on system behavior

None

### DIFF
--- a/planning/path_smoother/CMakeLists.txt
+++ b/planning/path_smoother/CMakeLists.txt
@@ -7,12 +7,7 @@ autoware_package()
 find_package(Eigen3 REQUIRED)
 
 ament_auto_add_library(path_smoother SHARED
-  # node
-  src/elastic_band_smoother.cpp
-  # algorithms
-  src/elastic_band.cpp
-  # utils
-  src/utils/trajectory_utils.cpp
+  DIRECTORY src
 )
 
 target_include_directories(path_smoother
--- a/planning/path_smoother/config/elastic_band_smoother.param.yaml
+++ b/planning/path_smoother/config/elastic_band_smoother.param.yaml
@@ -35,3 +35,13 @@
     # nearest search
     ego_nearest_dist_threshold: 3.0 # [m]
     ego_nearest_yaw_threshold: 1.046 # [rad] = 60 [deg]
+    # replanning & trimming trajectory param outside algorithm
+    replan:
+      enable: true  # if true, only perform smoothing when the input changes significantly
+      max_path_shape_around_ego_lat_dist: 2.0  # threshold of path shape change around ego [m]
+      max_path_shape_forward_lon_dist: 100.0   # forward point to check lateral distance difference [m]
+      max_path_shape_forward_lat_dist: 0.1     # threshold of path shape change around forward point [m]
+      max_ego_moving_dist: 5.0                 # threshold of ego's moving distance for replan [m]
+      # make max_goal_moving_dist long to keep start point fixed for pull over
+      max_goal_moving_dist: 15.0               # threshold of goal's moving distance for replan [m]
+      max_delta_time_sec: 1.0                  # threshold of delta time for replan [second]
--- a/planning/path_smoother/include/path_smoother/common_structs.hpp
+++ b/planning/path_smoother/include/path_smoother/common_structs.hpp
@@ -30,15 +30,15 @@ struct Bounds;
 
 struct PlannerData
 {
-  // input
-  Header header;
-  std::vector<TrajectoryPoint> traj_points;  // converted from the input path
-  std::vector<geometry_msgs::msg::Point> left_bound;
-  std::vector<geometry_msgs::msg::Point> right_bound;
-
-  // ego
+  std::vector<TrajectoryPoint> traj_points;
   geometry_msgs::msg::Pose ego_pose;
   double ego_vel{};
+
+  PlannerData(
+    std::vector<TrajectoryPoint> traj_points_, geometry_msgs::msg::Pose ego_pose_, double ego_vel_)
+  : traj_points(traj_points_), ego_pose(ego_pose_), ego_vel(ego_vel_)
+  {
+  }
 };
 
 struct TimeKeeper
--- a/planning/path_smoother/include/path_smoother/elastic_band_smoother.hpp
+++ b/planning/path_smoother/include/path_smoother/elastic_band_smoother.hpp
@@ -18,6 +18,7 @@
 #include "motion_utils/motion_utils.hpp"
 #include "path_smoother/common_structs.hpp"
 #include "path_smoother/elastic_band.hpp"
+#include "path_smoother/replan_checker.hpp"
 #include "path_smoother/type_alias.hpp"
 #include "rclcpp/rclcpp.hpp"
 #include "tier4_autoware_utils/tier4_autoware_utils.hpp"
@@ -59,6 +60,7 @@ protected:
 
   // algorithms
   std::shared_ptr<EBPathSmoother> eb_path_smoother_ptr_{nullptr};
+  std::shared_ptr<ReplanChecker> replan_checker_ptr_{nullptr};
 
   // parameters
   CommonParam common_param_{};
--- a/planning/path_smoother/include/path_smoother/replan_checker.hpp
+++ b/planning/path_smoother/include/path_smoother/replan_checker.hpp
@@ -1,0 +1,71 @@
+// Copyright 2023 TIER IV, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef PATH_SMOOTHER__REPLAN_CHECKER_HPP_
+#define PATH_SMOOTHER__REPLAN_CHECKER_HPP_
+
+#include "path_smoother/common_structs.hpp"
+#include "path_smoother/type_alias.hpp"
+
+#include <rclcpp/rclcpp.hpp>
+
+#include <memory>
+#include <vector>
+
+namespace path_smoother
+{
+class ReplanChecker
+{
+public:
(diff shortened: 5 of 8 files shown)


tkimura4:
Before:
ros2 topic echo /planning/scenario_planning/lane_driving/motion_planning/elastic_band_smoother/debug/calculation_time -f --field data
```
        insertFixedPoint:= 0.006 [ms]
        getPaddedTrajectoryPoints:= 0.009 [ms]
        updateConstraint:= 33.534 [ms]
        calcSmoothedTrajectory:= 1.355 [ms]
        convertOptimizedPointsToTrajectory:= 0.021 [ms]
      smoothTrajectory:= 35.182 [ms]
    onPath:= 35.198 [ms]
    applyInputVelocity:= 0.21 [ms]
  extendTrajectory:= 0.114 [ms]
onPath:= 35.563 [ms]
========================================

---

        insertFixedPoint:= 0.005 [ms]
        getPaddedTrajectoryPoints:= 0.004 [ms]
        updateConstraint:= 40.113 [ms]
        calcSmoothedTrajectory:= 0.168 [ms]
        convertOptimizedPointsToTrajectory:= 0.008 [ms]
      smoothTrajectory:= 40.416 [ms]
    onPath:= 40.422 [ms]
    applyInputVelocity:= 0.044 [ms]
  extendTrajectory:= 0.032 [ms]
onPath:= 40.509 [ms]
========================================

---

        insertFixedPoint:= 0.002 [ms]
        getPaddedTrajectoryPoints:= 0.003 [ms]
        updateConstraint:= 27.409 [ms]
        calcSmoothedTrajectory:= 0.108 [ms]
        convertOptimizedPointsToTrajectory:= 0.004 [ms]
      smoothTrajectory:= 27.596 [ms]
    onPath:= 27.599 [ms]
    applyInputVelocity:= 0.028 [ms]
  extendTrajectory:= 0.022 [ms]
onPath:= 27.655 [ms]
========================================

---

        insertFixedPoint:= 0.002 [ms]
        getPaddedTrajectoryPoints:= 0.002 [ms]
        updateConstraint:= 18.86 [ms]
        calcSmoothedTrajectory:= 0.116 [ms]
        convertOptimizedPointsToTrajectory:= 0.005 [ms]
      smoothTrajectory:= 19.052 [ms]
    onPath:= 19.055 [ms]
    applyInputVelocity:= 0.03 [ms]
  extendTrajectory:= 0.023 [ms]
onPath:= 19.113 [ms]
========================================

---

        insertFixedPoint:= 0.002 [ms]
        getPaddedTrajectoryPoints:= 0.002 [ms]
        updateConstraint:= 9.735 [ms]
        calcSmoothedTrajectory:= 0.113 [ms]
        convertOptimizedPointsToTrajectory:= 0.004 [ms]
      smoothTrajectory:= 9.924 [ms]
    onPath:= 9.927 [ms]
    applyInputVelocity:= 0.03 [ms]
  extendTrajectory:= 0.023 [ms]
onPath:= 9.985 [ms]
========================================

---

        insertFixedPoint:= 0.002 [ms]
        getPaddedTrajectoryPoints:= 0.003 [ms]
        updateConstraint:= 15.611 [ms]
        calcSmoothedTrajectory:= 0.163 [ms]
        convertOptimizedPointsToTrajectory:= 0.007 [ms]
      smoothTrajectory:= 15.87 [ms]
    onPath:= 15.876 [ms]
    applyInputVelocity:= 0.046 [ms]
  extendTrajectory:= 0.034 [ms]
onPath:= 15.966 [ms]
========================================
```

After:
```---

    onPath:= 0.003 [ms]
    applyInputVelocity:= 0.193 [ms]
  extendTrajectory:= 0.128 [ms]
onPath:= 0.381 [ms]
========================================

---

    onPath:= 0 [ms]
    applyInputVelocity:= 0.05 [ms]
  extendTrajectory:= 0.039 [ms]
onPath:= 0.109 [ms]
========================================

---

        insertFixedPoint:= 0.003 [ms]
        getPaddedTrajectoryPoints:= 0.002 [ms]
        updateConstraint:= 22.217 [ms]
        calcSmoothedTrajectory:= 0.12 [ms]
        convertOptimizedPointsToTrajectory:= 0.004 [ms]
      smoothTrajectory:= 22.493 [ms]
    onPath:= 22.496 [ms]
    applyInputVelocity:= 0.029 [ms]
  extendTrajectory:= 0.024 [ms]
onPath:= 22.556 [ms]
========================================

---

    onPath:= 0.001 [ms]
    applyInputVelocity:= 0.071 [ms]
  extendTrajectory:= 0.051 [ms]
onPath:= 0.149 [ms]
========================================

---

    onPath:= 0.001 [ms]
    applyInputVelocity:= 0.048 [ms]
  extendTrajectory:= 0.037 [ms]
onPath:= 0.105 [ms]
========================================

---

    onPath:= 0.002 [ms]
    applyInputVelocity:= 0.12 [ms]
  extendTrajectory:= 0.084 [ms]
onPath:= 0.251 [ms]
========================================

```

tkimura4 (approved):
Merging this PR has resulted in the optimization  (smoothTrajectory()) in the path_smoother no longer being executed every frame.
I have also confirmed a significant decrease in CPU usage of path_smoother node.

cc. @takayuki5168

takayuki5168 (approved):

ahmeddesokyebrahim:
Thanks @maxime-clem for this interesting PR.

We - planning & control team in Leo Drive - are curious about the change in this PR as we are facing a specific behavior in a specific scenario while testing autoware which we are trying to understand. And the reason we are curious about this PR is that the behavior before it has changed after.

Before using this PR, we have tested autoware and faced this weird ego vehicle stop as shown in this [video demo](https://www.youtube.com/watch?v=Nywkjx5NsAI) before the stop line (please watch carefully starting from 0:26).

The obvious reason for me (not totally sure), there is a replan (optimization) happening that is why it stopped and replanned the trajectory. The log message during this stop was `Replan since path forward shape changed.`

After using this PR, this weird stop is not longer happening except in very minor trials - giving me the impressions out of the comments above that optimization is no longer happening in each frame. And when it happens, it stops and replan again. 

Knowing that ... 
- This weird stop does not happen if we disable `stop_line` module with exact same scenrio
- This weird stop does not happen if we start the vehicle somehow near to the stop line so that the speed is less that 15 km/h (or with same exact scenario with `speed_bump` module enabled so that it slows down before approaching the stop line)

From these observations and checking the code change, I am curious to understand some questions and would be very grateful to you if you can support us in that ... 

1.  There is a parameter added in `replan` section in `elastic_band_smoother.param.yaml`. The comment associated with this parameter `# if true, only perform smoothing when the input changes significantly`. I am curious to understand how this is used to perform smoothing when input changed significantly in the time that the parameter is only used inside `isReplanRequired()` to return true if it is false. 
2. Is `smoothing` mentioned in the associated comment above equivalent to replanning/optimization in this context ?
3. Are we facing this weird stop with way more less frequency due to replanning/optmization is not happening now in every frame ?
4. What is the relation here between replanning/optimization and the path forward shape change ? 
5. Do you have any ideas in mind  to connect the dots why this happening with `stop_line` module enabled or not happening in lower speeds ? Does that relate somehow to the nature of replanning/optimization happening ?

Thanks in advance for your efforts and support, and please let me know if you would need any further questions/clarifications before answering the questions.

cc: @mehmetdogru - @beyzanurkaya

maxime-clem:
Hello and thank you for your question. I will try to answer as best as I can but I may need to investigate further to provide you with a satisfactory reply. Do not hesitate to ask more questions if things are not clear.

> 1. I am curious to understand how this is used to perform smoothing when input changed significantly in the time that the parameter is only used inside isReplanRequired() to return true if it is false.

As you noted, if `replan.enable` is set to `false`, then `isReplanRequired()` always returns true.
This is correct but maybe the parameter naming is confusing and `replan` should be renamed to `replan_checker`. By default, the `elastic_band_smoother` always smooths the inputs it receives, which is computationally quite expensive. For this reason we use the `replan_checker` to check whether a new smoothing is required.
When `enable` is `true`, we will try to reuse the previous result instead of performing a new smoothing.

> 2. Is smoothing mentioned in the associated comment above equivalent to replanning/optimization in this context ?

It is equivalent to redoing the elastic band optimization. There is no real planning involved as no decisions are made and the input path is just made smoother.

> 3. Are we facing this weird stop with way more less frequency due to replanning/optmization is not happening now in every frame ?

This is a strange issue. If there is a stop line it seems correct if the vehicle stops. However if the behavior is impacted by the `path_smoother` then there could be a bug.

> 4. What is the relation here between replanning/optimization and the path forward shape change ?

If the input path received is very different from the previous one, we want to update the output as well and will redo the elastic band optimization.

> 5. Do you have any ideas in mind to connect the dots why this happening with stop_line module enabled or not happening in lower speeds ? Does that relate somehow to the nature of replanning/optimization happening ?

There could be a bug with the resampling of the velocities such that the stop point (a single point with `0.0` velocity) is somehow "lost" during resampling.

> 6. As replan_checker is now copied to path_smoother and I have seen both isReplanRequired() is being called by path_smoother and obstacle_avoidance_planner , is that intended ? or is it planned in future that replan_checker to be only in path_smoother ?

I think there will be some refactoring in the future, but both the `path_smoother` and the `obstacle_avoidance_planner` use the `replan_checker` logic to determine whether they need to perform a new optimization or not. The goal is the same and the implementation is very similar, but they are both needed.

ahmeddesokyebrahim:
Thanks a lot @maxime-clem for the valuable answers from your side and sorry for my delayed response.
Actually, the interesting observation we have till now for this weird stop, that it does not happen when the first step of cropping trajectory in `smoothTrajectory()` in `elastic_band.cpp` is skipped. I mean by skipping here by depending on the following steps using `traj_points` instead of `cropped_traj_points`. I am not sure till now why cropping is introducing the stop point in this case. 
As well, feeding path to `obstacle_avoidance_planner` directly from `behavior_velocity_planner` instead of `path_smoother` make the problem not to happen. That means looks like the stop point is coming from the path updated by `path_smoother`. This stop point I can not see in `eb_traj` nor in `mpt_traj` but it is visible in `obstacle_avoidance_planner/trajectory`.
If we have the time to investigate more on this, we will update you for sure.
Thanks again for the answers.

maxime-clem:
@ahmeddesokyebrahim Sorry I should have discussed this during the working group today.

It sounds like a significant issue which should be addressed as soon as possible.
Can I ask you to create an issue ? Ideally with some instructions or rosbag to reproduce the problem. This way we can also investigate on our side.

PR #4616, opened 2023-08-13
feat(pid_longitudinal_controller): take into account the current acceleration to calculate delayed pose

brkay54:
## Description

In the current implementation, we are calculating the pose after delay compensation by using only velocity, however, adding current acceleration to the formula is going to improve our estimation.
## Tests performed
Tested planning_simulator and works well. 

Not applicable.

### DIFF
--- a/control/pid_longitudinal_controller/src/longitudinal_controller_utils.cpp
+++ b/control/pid_longitudinal_controller/src/longitudinal_controller_utils.cpp
@@ -129,11 +129,16 @@ double calcElevationAngle(const TrajectoryPoint & p_from, const TrajectoryPoint 
 }
 
 Pose calcPoseAfterTimeDelay(
-  const Pose & current_pose, const double delay_time, const double current_vel)
+  const Pose & current_pose, const double delay_time, const double current_vel,
+  const double current_acc)
 {
+  if (delay_time <= 0.0 || current_vel <= 0.0) {
+    return current_pose;
+  }
   // simple linear prediction
   const double yaw = tf2::getYaw(current_pose.orientation);
-  const double running_distance = delay_time * current_vel;
+  const double running_distance =
+    delay_time * current_vel + 0.5 * current_acc * delay_time * delay_time;
   const double dx = running_distance * std::cos(yaw);
   const double dy = running_distance * std::sin(yaw);
 
--- a/control/pid_longitudinal_controller/src/longitudinal_controller_utils.cpp
+++ b/control/pid_longitudinal_controller/src/longitudinal_controller_utils.cpp
@@ -143,8 +143,9 @@ Pose calcPoseAfterTimeDelay(
     time_to_stop > 0.0 && time_to_stop < delay_time ? time_to_stop : delay_time;
   // simple linear prediction
   const double yaw = tf2::getYaw(current_pose.orientation);
-  const double running_distance =
-    delay_time_calculation * current_vel + 0.5 * current_acc * delay_time * delay_time_calculation;
+  const double running_distance = delay_time_calculation * current_vel + 0.5 * current_acc *
+                                                                           delay_time_calculation *
+                                                                           delay_time_calculation;
   const double dx = running_distance * std::cos(yaw);
   const double dy = running_distance * std::sin(yaw);
 
--- a/control/pid_longitudinal_controller/src/longitudinal_controller_utils.cpp
+++ b/control/pid_longitudinal_controller/src/longitudinal_controller_utils.cpp
@@ -132,13 +132,14 @@ Pose calcPoseAfterTimeDelay(
   const Pose & current_pose, const double delay_time, const double current_vel,
   const double current_acc)
 {
-  if (delay_time <= 0.0 || current_vel <= 0.0) {
+  if (delay_time <= 0.0) {
     return current_pose;
   }
+
   // simple linear prediction
   const double yaw = tf2::getYaw(current_pose.orientation);
   const double running_distance =
-    delay_time * current_vel + 0.5 * current_acc * delay_time * delay_time;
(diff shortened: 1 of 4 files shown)


control/pid_longitudinal_controller/src/longitudinal_controller_utils.cpp
```diff
@@ -129,11 +129,16 @@ double calcElevationAngle(const TrajectoryPoint & p_from, const TrajectoryPoint
 }
 
 Pose calcPoseAfterTimeDelay(
-  const Pose & current_pose, const double delay_time, const double current_vel)
+  const Pose & current_pose, const double delay_time, const double current_vel,
+  const double current_acc)
 {
+  if (delay_time <= 0.0 || current_vel <= 0.0) {
+    return current_pose;
+  }
```
takayuki5168, line 137: I think delay compensation should work when the vehicle drives backward as well.

control/pid_longitudinal_controller/src/longitudinal_controller_utils.cpp
```diff
@@ -129,11 +129,16 @@ double calcElevationAngle(const TrajectoryPoint & p_from, const TrajectoryPoint
...
   // simple linear prediction
   const double yaw = tf2::getYaw(current_pose.orientation);
-  const double running_distance = delay_time * current_vel;
+  const double running_distance =
+    delay_time * current_vel + 0.5 * current_acc * delay_time * delay_time;
```
takayuki5168, line 141: We'd better check if `running_distance` will not exceed 0.
For example, when the current_vel = 1, current_acc = -3, and delay_time = 1, `running_distance` will be negative which is not expected.

What do you think?

control/pid_longitudinal_controller/src/longitudinal_controller_utils.cpp
brkay54: Yes right, running_distance should not be negative so we can use abs(current_vel), what do you think?

control/pid_longitudinal_controller/src/longitudinal_controller_utils.cpp
brkay54: Hi, thanks for advices, I updated the branch can you check it again?

takayuki5168: Thank you.
In the example I mentioned before when the current_vel = 1, current_acc = -3, and delay_time = 1, the `running_distance` will be a certain positive value, but the current implementation makes it zero.
Could you fix it just in case.

brkay54: `  const double running_distance =
    abs(delay_time * current_vel + 0.5 * current_acc * delay_time * delay_time);`
    
With the latest update, it will be 0.5, I couldn't understand your point.

control/pid_longitudinal_controller/src/longitudinal_controller_utils.cpp
brkay54: With the latest commit, this was also changed.

control/pid_longitudinal_controller/src/longitudinal_controller_utils.cpp
takayuki5168: Ah, sorry I misunderstood your implementation.
With `abs`, the running distance will not be negative, but I'm not sure why we can use `abs` here.

What I wanted to say is that if applying the delay_time results in a negative running distance like the case where the current_vel = 1, current_acc = -3, and delay_time = 1, the time to reach zero velocity should be used to calculate running_distance.
The implementation will be as follows. Please check if the calculation is correct.
```cpp
const double running_distance = [&]() {
    const double time_to_zero_vel = current_vel / -current_acc;
    if (0 < time_to_zero_vel && time_to_zero_vel < delay_time) {
      return time_to_zero_vel * current_vel + 0.5 * current_acc * std::pow(time_to_zero_vel, 2);
    }
    return delay_time * current_vel + 0.5 * current_acc * std::pow(delay_time, 2);
}();
```

brkay54: Sorry for my mistake, you are right, why I don't know but I think that running distance always should be positive, it was the wrong idea. I updated the PR, can you check it again?

control/pid_longitudinal_controller/src/longitudinal_controller_utils.cpp
```diff
@@ -129,11 +129,22 @@ double calcElevationAngle(const TrajectoryPoint & p_from, const TrajectoryPoint
...
+  // check time to stop
+  const double time_to_stop = -current_vel / current_acc;
+
+  const double delay_time_calculation =
+    time_to_stop > 0.0 && time_to_stop < delay_time ? time_to_stop : delay_time;
   // simple linear prediction
   const double yaw = tf2::getYaw(current_pose.orientation);
-  const double running_distance = delay_time * current_vel;
+  const double running_distance =
+    delay_time_calculation * current_vel + 0.5 * current_acc * delay_time * delay_time_calculation;
```
takayuki5168, line 147:
```suggestion
    delay_time_calculation * current_vel + 0.5 * current_acc * delay_time_calculation * delay_time_calculation;
```

takayuki5168:
@TakaHoribe Please review the PR.

TakaHoribe (commented):
Thank you, @brkay54. This is a nice improvement.

LGTM. Please merge after @takayuki5168 's comments are addressed with his approval.

takayuki5168:
@brkay54 
FYI: From several weeks ago, we have to add a label of `run-build-and-test-differential` to run CI. I added.
image

PR #4618, opened 2023-08-14
feat(map_based_prediction): plan paths of objects outside the lanelets.

takayuki5168:
## Description

- Added a publisher of calculation time
- Paths are planed even if the object is outside the lanelet if they are close enough.
**without the PR**
image
**with the PR**
image

FYI:
The case where object is in the opposite directional lane
image

## Tests performed

- [x] planning simulator
- [x] scenario simulator
    - https://evaluation.tier4.jp/evaluation/reports/5511d35c-462a-5071-95f4-edb057611521?project_id=prd_jt

## Effects on system behavior

Nothing

### DIFF
--- a/perception/map_based_prediction/src/map_based_prediction_node.cpp
+++ b/perception/map_based_prediction/src/map_based_prediction_node.cpp
@@ -1091,23 +1091,42 @@ LaneletsData MapBasedPredictionNode::getCurrentLanelets(const TrackedObject & ob
     return {};
   }
 
-  LaneletsData closest_lanelets;
+  LaneletsData object_lanelets;
+  std::optional<std::pair<double, lanelet::Lanelet>> closest_lanelet{std::nullopt};
   for (const auto & lanelet : surrounding_lanelets) {
     // Check if the close lanelets meet the necessary condition for start lanelets and
-    // Check if similar lanelet is inside the closest lanelet
+    // Check if similar lanelet is inside the object lanelet
     if (
       !checkCloseLaneletCondition(lanelet, object, search_point) ||
-      isDuplicated(lanelet, closest_lanelets)) {
+      isDuplicated(lanelet, object_lanelets)) {
       continue;
     }
 
-    LaneletData closest_lanelet;
-    closest_lanelet.lanelet = lanelet.second;
-    closest_lanelet.probability = calculateLocalLikelihood(lanelet.second, object);
-    closest_lanelets.push_back(closest_lanelet);
+    // Memorize closest lanelet
+    // NOTE: The object may be outside the lanelet.
+    if (
+      lanelet.first < dist_threshold_for_searching_lanelet_ &&
+      (!closest_lanelet || lanelet.first < closest_lanelet->first)) {
+      closest_lanelet = lanelet;
+    }
+
+    // Check if the obstacle is inside of this lanelet
+    constexpr double epsilon = 1e-3;
+    if (lanelet.first < epsilon) {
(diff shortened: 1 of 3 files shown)


perception/map_based_prediction/src/map_based_prediction_node.cpp
```diff
@@ -1104,11 +1138,6 @@ bool MapBasedPredictionNode::checkCloseLaneletCondition(
     return false;
   }
 
-  // Step2. Check if the obstacle is inside of this lanelet
-  if (!lanelet::geometry::inside(lanelet.second, search_point)) {
-    return false;
-  }
-
```
takayuki5168, line 1111: This part was moved to 
```
constexpr double epsilon = 1e-3;
if (lanelet.first < epsilon) {
```

perception/map_based_prediction/src/map_based_prediction_node.cpp
```diff
@@ -1076,23 +1091,42 @@ LaneletsData MapBasedPredictionNode::getCurrentLanelets(const TrackedObject & ob
...
+    }
   }
 
-  return closest_lanelets;
+  if (!object_lanelets.empty()) {
+    return object_lanelets;
+  }
+  if (closest_lanelet) {
+    return LaneletsData{LaneletData{
+      closest_lanelet->second, calculateLocalLikelihood(closest_lanelet->second, object)}};
```
YoshiRi, line 1127: need to check if `calculateLocalLikelihood` function is valid with obj outside the lanalet

perception/map_based_prediction/src/map_based_prediction_node.cpp
```diff
@@ -1076,23 +1091,42 @@ LaneletsData MapBasedPredictionNode::getCurrentLanelets(const TrackedObject & ob
...
-    closest_lanelet.lanelet = lanelet.second;
-    closest_lanelet.probability = calculateLocalLikelihood(lanelet.second, object);
-    closest_lanelets.push_back(closest_lanelet);
+    // Memorize closest lanelet
+    // NOTE: The object may be outside the lanelet.
+    if (
+      lanelet.first < dist_threshold_for_searching_lanelet_ &&
+      (!closest_lanelet || lanelet.first < closest_lanelet->first)) {
+      closest_lanelet = lanelet;
+    }
```
YoshiRi, line 1111: I think we need validation process to check object yaw and lanelet yaw diff is small enough so that we will not get the unexpected lanelet such as opposite lanelet

YoshiRi (commented):
So, your idea is to allow `getCurrentLanelet` function to return neighbor lanelet if there are no corresponding lanelet with object.

My concern is:
- the program outside the `getCurrentLanelet` function can not distinguish the output is closest_lanelet or object_lanelet
  - For example, this 'current' lanelet is saved to object_history_ and which is used in `checkCloseLaneletCondition` and so on

If possible, it would be safer for later feature development if it is known whether the returned value is a closet_lanelet or an object_lanelet, but since this is for developers, there should be at least no problem if it is described in a comment.
I think we can accept this concern only if this PR pass all of P/C test case.

perception/map_based_prediction/src/map_based_prediction_node.cpp
takayuki5168: We call checkCloseLaneletCondition before the above condition so we do not have to check the angle and even distance (Now I removed the distance check).

YoshiRi (approved):
LGTM

PR #4693, opened 2023-08-22
refactor(gnss_poser): rework parameters

PhoebeWu21:
## Description
Implement the ROS Node configuration layout described in https://github.com/orgs/autowarefoundation/discussions/3371 for the gnss_poser package.

- Remove the default value from the source code in order to ensure all parameter values are passed from the parameter files.
- Create the schema

## Tests performed
Package successfully build and launch locally.
`colcon build --symlink-install --cmake-args -DCMAKE_BUILD_TYPE=Release --packages-up-to gnss_poser`

## Effects on system behavior

- [x] I've confirmed the [contribution guidelines](https://autowarefoundation.github.io/autoware-documentation/main/contributing/).
- [x] The PR follows the [pull request guidelines](https://autowarefoundation.github.io/autoware-documentation/main/contributing/pull-request-guidelines/).

- [x] The PR follows the [pull request guidelines](https://autowarefoundation.github.io/autoware-documentation/main/contributing/pull-request-guidelines/).


kminoda:
@PhoebeWu21
Thank you for your contribution :smile: . The overall idea of this PR looks good to me :+1:

However, if you don't mind, can we hold with this PR and leave this package as it is for a short period of time?

The main reason is that this modification may change how parameters are reflected to `gnss_poser` in various project going on in Autoware Foundation. If we extract some parameters (like `coordinate_system`, which is reflected in [sample_sensor_kit_launch](https://github.com/autowarefoundation/sample_sensor_kit_launch/blob/3cef1e4888a41001e4ebb3afbf63ae823e42e27d/sample_sensor_kit_launch/launch/gnss.launch.xml#L4)) to json file, some parameters like `coordinate_system` will not be reflected any longer, which would change the whole system behavior. It is possible that the similar situation may exists in many Autoware users.

Essentially, this comes from the duplicated parameter maintenance of some parameters like `coordinate_system`, which ideally should be handled with map files. In other words, current gnss_poser parameters are not optimal, but rather in a "temporary" state. That being said, I propose to first do a major refactoring on `gnss_poser` to extract all such parameters (`coordinate_system`, `plane_zone`, `latitude`, etc), and then extract the parameters as well as introducing JSON Schema. This should save us the additional cost mentioned above.

I will handle it by myself, so would you do me a favor of holding this PR for a moment? It shouldn't take that much of time, but may take for several weeks if things gets complicated. But definitely doable.

PhoebeWu21:
> @PhoebeWu21 (cc: @YamatoAndo) Thank you for your contribution 😄 . The overall idea of this PR looks good to me 👍
> 
> However, if you don't mind, can we halt this PR and leave this package as it is for a short period of time?
> 
> The main reason is that this modification may change how parameters are reflected to `gnss_poser` in various project going on in Autoware Foundation. If we extract some parameters (like `coordinate_system`, which is reflected in [sample_sensor_kit_launch](https://github.com/autowarefoundation/sample_sensor_kit_launch/blob/3cef1e4888a41001e4ebb3afbf63ae823e42e27d/sample_sensor_kit_launch/launch/gnss.launch.xml#L4)) to yaml file, `coordinate_system` will not be reflected any longer, which would change the whole system behavior. It is possible that the similar situation may exists in many Autoware projects.
> 
> Essentially, this comes from the duplicated parameter maintenance of some parameters like `coordinate_system`, which ideally should be handled with map files (in [map_projection_loader](https://github.com/autowarefoundation/autoware.universe/blob/main/map/map_projection_loader/README.md)). In other words, current gnss_poser parameters are not optimal, but rather in a "temporary" state. That being said, I propose to first do a major refactoring on `gnss_poser` to remove all such parameters (`coordinate_system`, `plane_zone`, `latitude`, etc), and then extract the parameters as well as introducing JSON Schema as proposed in this PR. This should save us the additional cost of going through some other packages to mitigate the impact on current Autoware projects.
> 
> If you don't mind, I will handle the above major refactoring by myself, so would you do me a favor of holding this PR for a moment? It shouldn't take that much of time, but may take for several weeks if things gets complicated. But definitely doable.

Got it. Okay!

YamatoAndo:
@kminoda Has your refactoring been completed? If it's done, please let us know.

kminoda:
@YamatoAndo Thank you for the reminder :pray: 
@PhoebeWu21 It is now complete, so you may restart working on this PR, thanks

PhoebeWu21:
> @YamatoAndo Thank you for the reminder 🙏 @PhoebeWu21 It is now complete, so you may restart working on this PR, thanks

Okay, thanks!

sensing/gnss_poser/src/gnss_poser_core.cpp
```diff
@@ -1,4 +1,4 @@
-// Copyright 2020 Tier IV, Inc.
+// Copyright 2020-2023 Foxconn, Tier IV, Inc.
```
YamatoAndo, line 1: I apologize, but since there have been no modifications to this source file.
Could you please revert the copyright back to its original state?

PhoebeWu21: Ok. Done.

YamatoAndo: Thank you!

YamatoAndo (approved):

PR #4701, opened 2023-08-22
fix: center point class

taikitanaka3:
## Description

- add build only option which is necessary to run AWS + Docker 
- fix center point class and add build only option
- revert to use center point tiny as default

Note
-  tiny is AWF default model
- the reason we use center point is center point tiny has bug and it was risky to stay still but After fix which @knzo25 made and @shmpwk merged at https://github.com/autowarefoundation/autoware.universe/commit/461f330f27048e9d7b8d48898e43b9758f18eb75
Therefore there is no reason not to use tiny model.

```
ros2 launch lidar_centerpoint lidar_centerpoint.launch.xml model_name:=centerpoint build_only:=true
ros2 launch lidar_centerpoint lidar_centerpoint.launch.xml model_name:=centerpoint_tiny build_only:=true
```

## Tests performed

by AWSIM using center point
Screenshot from 2023-08-22 15-53-50

## Effects on system behavior

Not applicable.

### DIFF
--- a/launch/tier4_perception_launch/launch/perception.launch.xml
+++ b/launch/tier4_perception_launch/launch/perception.launch.xml
@@ -9,7 +9,7 @@
   <arg name="obstacle_segmentation_ground_segmentation_elevation_map_param_path"/>
 
   <!-- centerpoint model configs -->
-  <arg name="centerpoint_model_name" default="centerpoint" description="options: `centerpoint` or `centerpoint_tiny`"/>
+  <arg name="centerpoint_model_name" default="centerpoint_tiny" description="options: `centerpoint` or `centerpoint_tiny`"/>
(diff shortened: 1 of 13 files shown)


yukkysaito:
@taikitanaka3 @yukke42 Should we set it to tiny as default?

taikitanaka3:
@yukkysaito 
- tiny is AWF default model 
- the reason we use center point is center point tiny has bug and it was risky to stay still but After fix which @knzo25 made and @shmpwk merged https://github.com/autowarefoundation/autoware.universe/commit/461f330f27048e9d7b8d48898e43b9758f18eb75
there is no reason not to use tiny model.

so I think it's better to change default to tiny.

shmpwk (approved):
I confirmed that it works properly with AWSIM

PR #4712, opened 2023-08-22
feat(pid_longitudinal_controller): improve delay and slope compensation

brkay54:
## Description

In the current implementation on Autoware, while calculating the output command, we are compensating for the slope and delay in `pid_longitudinal_controller`. 

We are doing the delay compensation by using the estimation of the drive distance of the vehicle after `delay_compensation_time` and we are looking at the future `TrajectoryPoint` and calculating the output command with respect to this future point. 

However, before sending the output command, we are also doing slope compensation with respect to **Current** `TrajectoryPoint`. However, in some scenarios on hilly roads, this way is causing some conflicts because of the time like below.

delayslope drawio

As you can see in this scenario, there is a conflict between slope and delay compensation because they are using different reference times. In this scenario, the node will send FF with respect to the target point however it sums negative slope compensation value because of the current position of the vehicle and it causes undesired behavior. To avoid this behavior, I made some changes to delay and slope compensation in this package to sum compensations at the same time.

Also, after this PR, the following PRs should be merged by order here:

1- https://github.com/autowarefoundation/autoware.universe/pull/5077
2- https://github.com/autowarefoundation/autoware.universe/pull/5079
3- https://github.com/autowarefoundation/autoware.universe/pull/5080
4- https://github.com/autowarefoundation/autoware.universe/pull/5081

## Tests performed

I tested on the real vehicle and also on the planning simulator. It solves our problems in the real environment and there is no behavioral change in the planning simulator.

## Effects on system behavior

The only effect is that we will use the raw pitch info at only starting (while not moving), if the vehicle is moving, it needs the future pitch info, to achieve this, it depends on the pitch values of the trajectory points.

### Summary by CodeRabbit

- New Feature: Improved delay and slope compensation in the `pid_longitudinal_controller` package, resulting in more accurate output commands. The changes include updates to the `getPitchByTraj` function, introduction of the `findTrajectoryPoseAfterDistance` function, and modification of the return type of the `lerpTrajectoryPoint` function. These enhancements may impact the behavior of the code.
- Bug Fix: Addressed delay and slope compensation issues in the `PidLongitudinalController` class of Autoware. The modifications involve using drive distance estimation after a specified delay time for delay compensation and adjusting slope compensation to avoid conflicts. The changes also introduce the `StateAfterDelay` struct and a new `SlopeSource` enum.

### DIFF
--- a/control/pid_longitudinal_controller/src/longitudinal_controller_utils.cpp
+++ b/control/pid_longitudinal_controller/src/longitudinal_controller_utils.cpp
@@ -84,7 +84,7 @@ double getPitchByPose(const Quaternion & quaternion_msg)
 }
 
 double getPitchByTraj(
-  const Trajectory & trajectory, const size_t nearest_idx, const double wheel_base)
+  const Trajectory & trajectory, const size_t start_idx, const double wheel_base)
 {
   // cannot calculate pitch
   if (trajectory.points.size() <= 1) {
@@ -92,17 +92,17 @@ double getPitchByTraj(
   }
 
   const auto [prev_idx, next_idx] = [&]() {
-    for (size_t i = nearest_idx + 1; i < trajectory.points.size(); ++i) {
-      const double dist = tier4_autoware_utils::calcDistance2d(
-        trajectory.points.at(nearest_idx), trajectory.points.at(i));
+    for (size_t i = start_idx + 1; i < trajectory.points.size(); ++i) {
+      const double dist = tier4_autoware_utils::calcDistance3d(
+        trajectory.points.at(start_idx), trajectory.points.at(i));
       if (dist > wheel_base) {
         // calculate pitch from trajectory between rear wheel (nearest) and front center (i)
-        return std::make_pair(nearest_idx, i);
+        return std::make_pair(start_idx, i);
       }
     }
     // NOTE: The ego pose is close to the goal.
     return std::make_pair(
-      std::min(nearest_idx, trajectory.points.size() - 2), trajectory.points.size() - 1);
+      std::min(start_idx, trajectory.points.size() - 2), trajectory.points.size() - 1);
   }();
 
   return tier4_autoware_utils::calcElevationAngle(
@@ -157,6 +157,34 @@ double applyDiffLimitFilter(
   const double max_val = std::fabs(lim_val);
   const double min_val = -max_val;
   return applyDiffLimitFilter(input_val, prev_val, dt, max_val, min_val);
+}
+
+geometry_msgs::msg::Pose findTrajectoryPoseAfterDistance(
+  const size_t src_idx, const double distance,
+  const autoware_auto_planning_msgs::msg::Trajectory & trajectory)
+{
+  double remain_dist = distance;
+  geometry_msgs::msg::Pose p = trajectory.points.back().pose;
+  for (size_t i = src_idx; i < trajectory.points.size() - 1; ++i) {
+    const double dist = tier4_autoware_utils::calcDistance3d(
+      trajectory.points.at(i).pose, trajectory.points.at(i + 1).pose);
+    if (remain_dist < dist) {
+      if (remain_dist <= 0.0) {
+        return trajectory.points.at(i).pose;
+      }
+      double ratio = remain_dist / dist;
+      const auto p0 = trajectory.points.at(i).pose;
+      const auto p1 = trajectory.points.at(i + 1).pose;
+      p = trajectory.points.at(i).pose;
+      p.position.x = interpolation::lerp(p0.position.x, p1.position.x, ratio);
+      p.position.y = interpolation::lerp(p0.position.y, p1.position.y, ratio);
+      p.position.z = interpolation::lerp(p0.position.z, p1.position.z, ratio);
+      p.orientation = interpolation::lerpOrientation(p0.orientation, p1.orientation, ratio);
+      break;
+    }
+    remain_dist -= dist;
+  }
+  return p;
 }
 }  // namespace longitudinal_utils
 }  // namespace autoware::motion::control::pid_longitudinal_controller
--- a/control/pid_longitudinal_controller/src/pid_longitudinal_controller.cpp
+++ b/control/pid_longitudinal_controller/src/pid_longitudinal_controller.cpp
@@ -168,11 +168,26 @@ PidLongitudinalController::PidLongitudinalController(rclcpp::Node & node)
   m_min_jerk = node.declare_parameter<double>("min_jerk");  // [m/s^3]
 
   // parameters for slope compensation
-  m_use_traj_for_pitch = node.declare_parameter<bool>("use_trajectory_for_pitch_calculation");
+  m_adaptive_trajectory_velocity_th =
+    node.declare_parameter<double>("adaptive_trajectory_velocity_th");  // [m/s^2]
   const double lpf_pitch_gain{node.declare_parameter<double>("lpf_pitch_gain")};
   m_lpf_pitch = std::make_shared<LowpassFilter1d>(0.0, lpf_pitch_gain);
   m_max_pitch_rad = node.declare_parameter<double>("max_pitch_rad");  // [rad]
   m_min_pitch_rad = node.declare_parameter<double>("min_pitch_rad");  // [rad]
+
+  // check slope source is proper
+  const std::string slope_source = node.declare_parameter<std::string>(
+    "slope_source");  // raw_pitch, trajectory_pitch or trajectory_adaptive
+  if (slope_source == "raw_pitch") {
+    m_slope_source = SlopeSource::RAW_PITCH;
+  } else if (slope_source == "trajectory_pitch") {
+    m_slope_source = SlopeSource::TRAJECTORY_PITCH;
+  } else if (slope_source == "trajectory_adaptive") {
+    m_slope_source = SlopeSource::TRAJECTORY_ADAPTIVE;
+  } else {
+    RCLCPP_ERROR(logger_, "Slope source is not valid. Using raw_pitch option as default");
+    m_slope_source = SlopeSource::RAW_PITCH;
+  }
 
   // ego nearest index search
   m_ego_nearest_dist_threshold =
@@ -355,6 +370,7 @@ rcl_interfaces::msg::SetParametersResult PidLongitudinalController::paramCallbac
   // slope compensation
   update_param("max_pitch_rad", m_max_pitch_rad);
   update_param("min_pitch_rad", m_min_pitch_rad);
+  update_param("adaptive_trajectory_velocity_th", m_adaptive_trajectory_velocity_th);
 
   rcl_interfaces::msg::SetParametersResult result;
   result.successful = true;
@@ -401,7 +417,7 @@ trajectory_follower::LongitudinalOutput PidLongitudinalController::run(
   updateControlState(control_data);
 
   // calculate control command
-  const Motion ctrl_cmd = calcCtrlCmd(current_pose, control_data);
+  const Motion ctrl_cmd = calcCtrlCmd(control_data);
 
   // publish control command
   const auto cmd_msg = createCtrlCmdMsg(ctrl_cmd, control_data.current_motion.vel);
@@ -428,19 +444,27 @@ PidLongitudinalController::ControlData PidLongitudinalController::getControlData
   // current velocity and acceleration
   control_data.current_motion.vel = m_current_kinematic_state.twist.twist.linear.x;
   control_data.current_motion.acc = m_current_accel.accel.accel.linear.x;
+  control_data.interpolated_traj = m_trajectory;
 
-  // nearest idx
-  const size_t nearest_idx = motion_utils::findFirstNearestIndexWithSoftConstraints(
-    m_trajectory.points, current_pose, m_ego_nearest_dist_threshold, m_ego_nearest_yaw_threshold);
-  const auto & nearest_point = m_trajectory.points.at(nearest_idx).pose;
+  // calculate the interpolated point and segment
+  const auto current_interpolated_pose =
+    calcInterpolatedTrajPointAndSegment(control_data.interpolated_traj, current_pose);
+
+  // Insert the interpolated point
+  control_data.interpolated_traj.points.insert(
+    control_data.interpolated_traj.points.begin() + current_interpolated_pose.second + 1,
+    current_interpolated_pose.first);
+  control_data.nearest_idx = current_interpolated_pose.second + 1;
+  control_data.target_idx = control_data.nearest_idx;
 
   // check if the deviation is worth emergency
   m_diagnostic_data.trans_deviation =
-    tier4_autoware_utils::calcDistance2d(nearest_point, current_pose);
+    tier4_autoware_utils::calcDistance2d(current_interpolated_pose.first, current_pose);
   const bool is_dist_deviation_large =
     m_state_transition_params.emergency_state_traj_trans_dev < m_diagnostic_data.trans_deviation;
   m_diagnostic_data.rot_deviation = std::abs(tier4_autoware_utils::normalizeRadian(
-    tf2::getYaw(nearest_point.orientation) - tf2::getYaw(current_pose.orientation)));
+    tf2::getYaw(current_interpolated_pose.first.pose.orientation) -
+    tf2::getYaw(current_pose.orientation)));
   const bool is_yaw_deviation_large =
     m_state_transition_params.emergency_state_traj_rot_dev < m_diagnostic_data.rot_deviation;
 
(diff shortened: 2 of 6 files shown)


TakaHoribe:
@brkay54 Thank you for this PR. Recent PRs have caused conflicts. Would you fix it? Sorry for your inconvenience.

brkay54:
Hi @TakaHoribe sorry for the conflicts. I updated all the PRs.

control/pid_longitudinal_controller/src/pid_longitudinal_controller.cpp
```diff
@@ -422,19 +421,27 @@ PidLongitudinalController::ControlData PidLongitudinalController::getControlData
...
-    m_trajectory.points, current_pose, m_ego_nearest_dist_threshold, m_ego_nearest_yaw_threshold);
-  const auto & nearest_point = m_trajectory.points.at(nearest_idx).pose;
+  const auto current_interpolated_pose =
+    calcInterpolatedTrajPointAndSegment(control_data.interpolated_traj, current_pose);
+  const size_t nearest_idx = current_interpolated_pose.second + 1;
+  const auto & nearest_point = current_interpolated_pose.first;
+
+  // Insert the interpolated point
+  control_data.interpolated_traj.points.insert(
+    control_data.interpolated_traj.points.begin() + uint(nearest_idx), nearest_point);
```
TakaHoribe, line 434: Do we need casting to `uint`? We can use `size_t` directly, I guess.

control/pid_longitudinal_controller/src/pid_longitudinal_controller.cpp
```diff
@@ -422,19 +421,27 @@ PidLongitudinalController::ControlData PidLongitudinalController::getControlData
...
-  const size_t nearest_idx = motion_utils::findFirstNearestIndexWithSoftConstraints(
-    m_trajectory.points, current_pose, m_ego_nearest_dist_threshold, m_ego_nearest_yaw_threshold);
-  const auto & nearest_point = m_trajectory.points.at(nearest_idx).pose;
+  const auto current_interpolated_pose =
+    calcInterpolatedTrajPointAndSegment(control_data.interpolated_traj, current_pose);
+  const size_t nearest_idx = current_interpolated_pose.second + 1;
+  const auto & nearest_point = current_interpolated_pose.first;
+
+  // Insert the interpolated point
+  control_data.interpolated_traj.points.insert(
```
TakaHoribe, line 433: When you add a new point in the trajectory, please ensure there are no duplicated points. Otherwise, some computation throws an exception.

control/pid_longitudinal_controller/src/pid_longitudinal_controller.cpp
```diff
@@ -443,24 +450,59 @@ PidLongitudinalController::ControlData PidLongitudinalController::getControlData
...
     return control_data;
   }
-  control_data.nearest_idx = nearest_idx;
+
+  // Delay compensation - Calculate the distance we got, predicted velocity and predicted
+  // acceleration after delay
+  control_data.state_after_delay =
+    predictedStateAfterDelay(control_data.current_motion, m_delay_compensation_time);
+
+  // calculate the target motion
```
TakaHoribe, line 459:
```suggestion
  // calculate the target motion for delay compensation
```

control/pid_longitudinal_controller/src/pid_longitudinal_controller.cpp
```diff
@@ -682,6 +713,8 @@ PidLongitudinalController::Motion PidLongitudinalController::calcCtrlCmd(
     raw_ctrl_cmd = calcEmergencyCtrlCmd(control_data.dt);
   }
 
+  raw_ctrl_cmd = keepBrakeBeforeStop(control_data, raw_ctrl_cmd, target_idx);
```
TakaHoribe, line 716: Why this is moved? I think this function is necessary only for DRIVE state.

control/pid_longitudinal_controller/src/pid_longitudinal_controller.cpp
```diff
@@ -783,8 +818,11 @@ double PidLongitudinalController::calcFilteredAcc(
   // store ctrl cmd without slope filter
   storeAccelCmd(acc_max_filtered);
 
+  // if current state is STOPPED, do not apply slope compensation to avoid sudden acceleration
```
TakaHoribe, line 821: I don't know why the slope compensation causes a sudden stop. The sudden acceleration should not depend on the slope compensation, but the stop acceleration parameter, I guess.

control/pid_longitudinal_controller/src/pid_longitudinal_controller.cpp
```diff
@@ -867,83 +905,98 @@ PidLongitudinalController::Motion PidLongitudinalController::keepBrakeBeforeStop
...
+
+  if (m_ctrl_cmd_vec.empty() || m_current_operation_mode.mode != OperationModeState::AUTONOMOUS) {
+    // check time to stop
+    const double time_to_stop = -current_vel / current_acc;
+    const double delay_time_calculation =
+      time_to_stop > 0.0 && time_to_stop < delay_compensation_time ? time_to_stop
+                                                                   : delay_compensation_time;
+    // simple linear prediction
+    pred_vel = current_vel + current_acc * delay_time_calculation;
+    running_distance = abs(
```
TakaHoribe, line 939:
```suggestion
    running_distance = std::abs(
```

control/pid_longitudinal_controller/src/pid_longitudinal_controller.cpp
```diff
@@ -867,83 +905,98 @@ PidLongitudinalController::Motion PidLongitudinalController::keepBrakeBeforeStop
...
+          : std::min(
+              (rclcpp::Time(m_ctrl_cmd_vec.at(i + 1).stamp) -
+               rclcpp::Time(m_ctrl_cmd_vec.at(i).stamp))
+                .seconds(),
+              delay_compensation_time);
+      const double acc = m_ctrl_cmd_vec.at(i).acceleration;
+      // because acc_cmd is positive when vehicle is running backward
+      pred_acc = std::copysignf(1.0, static_cast<float>(pred_vel)) * acc;
+      running_distance +=
+        abs(abs(pred_vel) * time_to_next_acc + 0.5 * acc * time_to_next_acc * time_to_next_acc);
```
TakaHoribe, line 962:
```suggestion
        std::abs(std::abs(pred_vel) * time_to_next_acc + 0.5 * acc * time_to_next_acc * time_to_next_acc);
```

TakaHoribe (commented):
@brkay54 Thank you for this PR. The overall strategy looks god.
Would you check my comments? Thank you!

control/pid_longitudinal_controller/src/pid_longitudinal_controller.cpp
```diff
@@ -443,24 +450,59 @@ PidLongitudinalController::ControlData PidLongitudinalController::getControlData
...
 
   // pitch
   const double raw_pitch = longitudinal_utils::getPitchByPose(current_pose.orientation);
-  const double traj_pitch =
-    longitudinal_utils::getPitchByTraj(m_trajectory, control_data.nearest_idx, m_wheel_base);
-  control_data.slope_angle = m_use_traj_for_pitch ? traj_pitch : m_lpf_pitch->filter(raw_pitch);
+  const double traj_pitch = longitudinal_utils::getPitchByTraj(
+    control_data.interpolated_traj, control_data.target_idx, m_wheel_base);
+
+  // if vehicle is not in driving state, send the raw slope angle
```
TakaHoribe, line 496: Regarding which slope information to use, it should not depend on the logic as a basic policy because it depends on the accuracy of sensors and maps. What is the purpose of this implementation?

control/pid_longitudinal_controller/src/pid_longitudinal_controller.cpp
```diff
@@ -867,83 +905,98 @@ PidLongitudinalController::Motion PidLongitudinalController::keepBrakeBeforeStop
...
-    return pred_vel > 0 ? std::copysign(pred_vel, current_vel) : 0.0;
+    return StateAfterDelay{pred_vel, pred_acc, running_distance};
   }
 
-  double pred_vel = current_vel_abs;
-
-  const auto past_delay_time =
-    clock_->now() - rclcpp::Duration::from_seconds(delay_compensation_time);
   for (std::size_t i = 0; i < m_ctrl_cmd_vec.size(); ++i) {
     if ((clock_->now() - m_ctrl_cmd_vec.at(i).stamp).seconds() < m_delay_compensation_time) {
```
TakaHoribe, line 947: Is it ok to use `m_delay_compensation_time`? Should it be `delay_compensation_time`? There are same implementations in this function.

TakaHoribe (commented):
other comments.

control/pid_longitudinal_controller/src/pid_longitudinal_controller.cpp
```diff
@@ -163,7 +163,6 @@ PidLongitudinalController::PidLongitudinalController(rclcpp::Node & node)
   m_min_jerk = node.declare_parameter<double>("min_jerk");  // [m/s^3]
 
   // parameters for slope compensation
-  m_use_traj_for_pitch = node.declare_parameter<bool>("use_trajectory_for_pitch_calculation");
```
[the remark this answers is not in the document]
brkay54: Because slope compensation won't be done with respect to the user's option. While driving the ego, it is always needed that the slope should be calculated from future position projection to the trajectory of the vehicle.

The pitch information from the sensing only is usable when the vehicle is in the stopped state.

control/pid_longitudinal_controller/src/pid_longitudinal_controller.cpp
brkay54: > Regarding which slope information to use, it should not depend on the logic as a basic policy because it depends on the accuracy of sensors and maps. What is the purpose of this implementation?



Actually, the pitch information from sensing would be more reliable from the trajectory. My purpose is, that if the vehicle is stopping, it can use the pitch information from sensing, with no need for future slope. But if the vehicle is moving (running distance after delay >0.0), it needs pitch information from the future.

While changing the pitch input source, to provide smoother change, I put them into LPF.

Maybe the slope input source should change with respect to running_distance threshold like:

if (running_distance > 0.5?) {
control_data.slope_angle = traj_pitch;
m_lpf_pitch->filter(control_data.slope_angle);
} else {
control_data.slope_angle = raw_pitch;
// send filtered slope angle to smooth transition between raw and traj pitch
control_data.slope_angle = m_lpf_pitch->filter(control_data.slope_angle);
}

Because in the current implementation, it is possible that the vehicle's velocity is < 0.001 but the vehicle may be in the DRIVE state. What do you think?

brkay54:
Hi @TakaHoribe thank you for your review. I updated the branch w.r.t. your feedback. 
Just I didn't update the code snippets related to:
https://github.com/autowarefoundation/autoware.universe/pull/4712#discussion_r1342496103
https://github.com/autowarefoundation/autoware.universe/pull/4712#discussion_r1342528866 
I am waiting for your feedback! Thank you so much for your effort ❤️

brkay54:
Hi @TakaHoribe, sorry to disturb you, but just I want to kindly ping you.

control/pid_longitudinal_controller/src/longitudinal_controller_utils.cpp
```diff
@@ -168,5 +168,39 @@ double applyDiffLimitFilter(
...
+        ratio);
+      p.position.y = interpolation::lerp(
+        trajectory.points.at(i).pose.position.y, trajectory.points.at(i + 1).pose.position.y,
+        ratio);
+      p.position.z = interpolation::lerp(
+        trajectory.points.at(i).pose.position.z, trajectory.points.at(i + 1).pose.position.z,
+        ratio);
+      p.orientation = lerpOrientation(
+        trajectory.points.at(i).pose.orientation, trajectory.points.at(i + 1).pose.orientation,
+        ratio);
```
TakaHoribe, line 198: This would be more simple?

```suggestion
      const auto p0 = trajectory.points.at(i).pose;
      const auto p1 = trajectory.points.at(i).pose;
      p.position.x = interpolation::lerp(p0.position.x, p1.position.x, ratio);
      p.position.y = interpolation::lerp(p0.position.y, p1.position.y, ratio);
      p.position.z = interpolation::lerp(p0.position.z, p1.position.z, ratio);
      p.orientation = lerpOrientation(p0.orientation, p1.orientation, ratio);
```

TakaHoribe:
@brkay54 Sorry for my late reply 

> Actually, the pitch information from sensing would be more reliable from the trajectory. 

This depends on the users. For the pitch source, we still need to keep the sensor-only option for users who can not have an accurate map.

So I would suggest this way:

Provide three options: 

- `vehicle_pitch`: Use the vehicle's pitch value as a slope. The information derived from the vehicle's pitch is often not an accurate representation of the road gradient due to effects such as body vibrations.
- `trajectory` : Use the trajectory slope value as a slope. Calculating the precise road gradient is possible with trajectory data; however, it requires high-quality map information.
- `trajectory_adaptive`: Primarily, the gradient is calculated using trajectory data, but if the gradient cannot be computed at the end of the route or similar situations, the vehicle pitch is used instead.

The default could be `vehicle_pitch` or `trajectory_adaptive`.

What do you think?

brkay54:
@TakaHoribe Thank you for the suggestion. It makes sense to me. I think it is better to make `trajectory_adaptive` as the default option because it can work with delay compensation better. 

For the trajectory_adaptive option, if it is okay, I am planning to make it adaptive with respect to the velocity of the vehicle. It will use the vehicle_pitch in low velocities (because in low velocities, the target trajectory point gonna be very close to the vehicle).

brkay54:
> @TakaHoribe Thank you for the suggestion. It makes sense to me. I think it is better to make `trajectory_adaptive` as the default option because it can work with delay compensation better.
> 
> For the trajectory_adaptive option, if it is okay, I am planning to make it adaptive concerning the velocity of the vehicle. It will use the vehicle_pitch in low velocities (because in low velocities, the target trajectory point gonna be very close to the vehicle).

Hi again @TakaHoribe -san, I wrote like this but after thinking more, IMO, 2 options would be better rather than using 3 options.

Using only trajectory has a problem for the last point, as you said, and also for all cases, we will have vehicle pitch value. I think the `trajectory` option is not necessary because we will need the future slope only for higher velocities, for lower velocities we can use vehicle_pitch (BCS target waypoint will be near ego). If we do not assume that the user may not have vehicle pitch value, the `adaptive_trajectory` option is always the better option than `trajectory`. 

My proposal is that like that:

```
# slope compensation
adaptive_trajectory: true  # if true, trajectory is used as pitch information source for higher velocities and vehicle pitch for lower velocities. if false, vehicle_pitch is used as pitch information.
adaptive_trajectory_treshold: 1.0 # available only adaptive_trajectory is true
```
What do you think?

brkay54:
Hi again @TakaHoribe -san, I updated the PR w.r.t my comment above. Also, I updated the launch [PR](https://github.com/autowarefoundation/autoware_launch/pull/585) and readme [PR](https://github.com/autowarefoundation/autoware.universe/pull/5081). Please let me know if you have any concerns.

TakaHoribe:
@brkay54 Thank you for addressing my comments. 

> because we will need the future slope only for higher velocities, for lower velocities we can use vehicle_pitch

This is only in the ideal case. We need to care about users who can not use an accurate map or rich sensor for pitch estimation. 

Trajectory-only option is necessary for users with poor sensors. 
Pitch-only option is necessary for uses with poor maps.
Adaptive can be available only for users with both ich sensors and rich maps.

brkay54:
Hi @TakaHoribe -san, sorry to be late I couldn't find time to update the PR in the last weeks. 

I updated the slope source option as you said above. You can find the changes in the launch file [here](https://github.com/autowarefoundation/autoware_launch/pull/585). It is ready for review, please let me know if any issue exists.

Also, after this PR, the following PRs should be merged by order here:

1- https://github.com/autowarefoundation/autoware.universe/pull/5077
2- https://github.com/autowarefoundation/autoware.universe/pull/5079
3- https://github.com/autowarefoundation/autoware.universe/pull/5080
4- https://github.com/autowarefoundation/autoware.universe/pull/5081

control/pid_longitudinal_controller/src/pid_longitudinal_controller.cpp
```diff
@@ -428,19 +444,27 @@ PidLongitudinalController::ControlData PidLongitudinalController::getControlData
...
   control_data.current_motion.acc = m_current_accel.accel.accel.linear.x;
+  control_data.interpolated_traj = m_trajectory;
 
   // nearest idx
-  const size_t nearest_idx = motion_utils::findFirstNearestIndexWithSoftConstraints(
-    m_trajectory.points, current_pose, m_ego_nearest_dist_threshold, m_ego_nearest_yaw_threshold);
-  const auto & nearest_point = m_trajectory.points.at(nearest_idx).pose;
+  const auto current_interpolated_pose =
+    calcInterpolatedTrajPointAndSegment(control_data.interpolated_traj, current_pose);
+  const size_t nearest_idx = current_interpolated_pose.second + 1;
```
TakaHoribe, line 452: `nearest_idx` is not always equivalent to `segment_idx + 1`. 
But the `segment_idx + 1` will be `nearest_idx` when the point is inserted in the `interpolated_traj`.

I guess this variable is not used anywhere?

TakaHoribe (commented):
@brkay54 LGTM and almost ready to merge.

Would you address the followings?

- Fix test code. This seems to fail due to the pos/neg definition mismatch. 
- Make a PR to the autoware_launch for new parameters.
- Make sure unrelated parameters are not changed in this PR: parameters attached in the screenshot should not change.

image

control/pid_longitudinal_controller/src/pid_longitudinal_controller.cpp
brkay54: Thank you for the review Horibe-san. Because it will be inserted into interpolated_traj, I set current_interpolated_pose.second + 1` to `nearest_idx` as below:

```
  // Insert the interpolated point
  control_data.interpolated_traj.points.insert(
    control_data.interpolated_traj.points.begin() + nearest_idx, nearest_point);
  control_data.nearest_idx = nearest_idx;
  control_data.target_idx = control_data.nearest_idx; // target_idx will be calculated, we set nearest_idx for now.
  ```
  
  Yes, the variable is not used anywhere, I simplified it. 
  
> `nearest_idx` is not always equivalent to `segment_idx + 1`.

I am not sure I can understand what you mean. I defined it like that because it will be inserted into interpolated_traj (If I understand correctly). I simplified this snipped. Please let me know if there is any problem.

>     * Fix test code. This seems to fail due to the pos/neg definition mismatch.

The test file is updated.

>     * Make a PR to the autoware_launch for new parameters.

I created PR for the parameter update: https://github.com/autowarefoundation/autoware_launch/pull/585
 
>     * Make sure unrelated parameters are not changed in this PR: parameters attached in the screenshot should not change.

Sorry for the changes, I updated.

brkay54:
@TakaHoribe-san, Just to remind: Also, after this PR, the following PRs should be merged by order here:

1- https://github.com/autowarefoundation/autoware.universe/pull/5077
2- https://github.com/autowarefoundation/autoware.universe/pull/5079
3- https://github.com/autowarefoundation/autoware.universe/pull/5080
4- https://github.com/autowarefoundation/autoware.universe/pull/5081

Also, Horibe-san, I made some critical changes to the package. If you don't mind, I would like to maintain this package. Can I add myself as a maintainer?

TakaHoribe (approved):
LGTM!

TakaHoribe:
@brkay54 

> Can I add myself as a maintainer?

I think it is fine. I'll talk with @takayuki5168

PR #4729, opened 2023-08-24
feat(imu_corrector): add gyro_bias_validator

kminoda:
## Description

I would like to add gyro_bias_estimator in imu_corrector in order to **monitor the gyro bias estimation and detect when the bias is not negligible**.

Current Autoware assumes static gyro bias offset, but in reality the offset may change due to various reasons. As a first step, I would like to add this node to detect this change.

Note that the next step would be to estimate the bias online (change dynamically) instead of using the static parameter calibrated beforehand. However, I will add this monitoring node first to mitigate the potential degradation on current system. The output of the node in this PR would not be used anywhere in the system, so it's safe as long as the computation is efficient enough.

## Related links

Must be merged with:
- sample_sensor_kit_launch: https://github.com/autowarefoundation/sample_sensor_kit_launch/pull/70
- aip_launcher: https://github.com/tier4/aip_launcher/pull/168

### Check if the gyro_bias_estimator outputs warnings when gyro bias is not correctly set

Gyro bias is not set for sample_sensor_kit by default, whereas the rosbag data shows a certain offset in gyroscope. 

1. Launch logging_simulator following the Autoware tutorial with sample_sensor_kit and sample_vehicle
2. Check that `gyro_bias_estimator` is correctly launched (using rqt_runtime_monitor)
3. Check `/diagnostics` and verify that `gyro_bias_estimator` outputs "NG"
4. Check `/sensing/imu/gyro_bias` and verify that it outputs some data

image

### Check if the gyro_bias_estimator outputs NO warnings when gyro bias is correctly set

First, reflect the following in `src/param/autoware_individual_params/individual_params/config/default/sample_sensor_kit/imu_corrector.param.yaml`.
```yaml
/**:
  ros__parameters:
    angular_velocity_stddev_xx: 0.00339
    angular_velocity_stddev_yy: 0.00339
    angular_velocity_stddev_zz: 0.00339
    angular_velocity_offset_x: -0.005799
    angular_velocity_offset_y: -0.007148
    angular_velocity_offset_z: -0.001499

```

1. Launch logging_simulator following the Autoware tutorial with sample_sensor_kit and sample_vehicle
2. Check that `gyro_bias_estimator` is correctly launched
3. Check `/diagnostics` and verify that `gyro_bias_estimator` outputs "OK"
4. Check `/sensing/imu/gyro_bias` and verify that it outputs some data

image

### Check if the gtest passes
Checked with 
```bash
colcon test --event-handlers console_cohesion+ --packages-select imu_corrector
```

## Notes for reviewers

Please follow the above test

## Interface changes

Additional interface has been introduced, but not used. 

## Effects on system behavior

None


sensing/imu_corrector/src/gyro_bias_estimator.cpp
```diff
@@ -0,0 +1,109 @@
...
+    // Update diagnostics
+    if (is_bias_small_enough) {
+      stat.add("gyro_bias", "OK");
+      stat.summary(diagnostic_msgs::msg::DiagnosticStatus::OK, "OK");
+    } else {
+      stat.add(
+        "gyro_bias",
+        "Gyro bias may be incorrect. Please calibrate IMU and reflect the result in "
+        "imu_corrector. You may also use the output of gyro_bias_estimator.");
+      stat.summary(diagnostic_msgs::msg::DiagnosticStatus::WARN, "NG");
```
TaikiYamada4, line 101: "NG" is a Japanese English word, so something like "Warning" or "Error" might be more appropriate against native speakers.

sensing/imu_corrector/src/imu_corrector_core.cpp
```diff
@@ -12,7 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include "imu_corrector/imu_corrector_core.hpp"
+#include "imu_corrector_core.hpp"
+
+#include <algorithm>
```
TaikiYamada4, line 17: Why did you add <algorithm> in this code?
I tried building and simulation with line 17 removed, but it seems to work fine.

TaikiYamada4 (commented):
I checked that it works fine on building and logging simulations, but there were just a few things that concerned me, so I commented on them in "Files changed".

sensing/imu_corrector/src/gyro_bias_estimator.cpp
kminoda: Thanks! Fixed in https://github.com/autowarefoundation/autoware.universe/pull/4729/commits/59900c837545ac719f62344fed0c2cec27910042

sensing/imu_corrector/src/imu_corrector_core.cpp
kminoda: Used in std::max, etc
It is true that the compilation of the code without the inclusion in .cpp also works since it's included somewhere else (in this case probably from `imu_corrector_core.hpp`). But then the pre-commit fails:
```cpplint..................................................................Failed
- hook id: cpplint
- exit code: 1

sensing/imu_corrector/src/imu_corrector_core.cpp:24:  Add #include <algorithm> for max  [build/include_what_you_use] [4]
Done processing sensing/imu_corrector/src/imu_corrector_core.cpp
Total errors found: 1
```

(Which often called as "include-what-you-use" practice)

TaikiYamada4 (approved):
LGTM!

PR #4813, opened 2023-08-30
feat(dynamic_avoidance): generate drivable area by ego/object-path-based depending on the case

takayuki5168:
## Description

- If the object's path is a cutting-in path, but the object is very close to the ego which has to be avoided, without the PR, the decision is no avoidance.
    - To deal with this case, if the object path is like above, the decision to avoid is calculated by the ego and object position.
- Fixed the debug marker's z.
- Dealt with the case where the object is crossing the ego's path, but the decision was avoidance without this PR.
image

## Tests performed

planning simulator

## Effects on system behavior

nothing


takayuki5168:
pre-commit.ci run

kosuke55 (approved):

PR #4854, opened 2023-09-01
feat(ndt_scan_matcher): adding state switch mecahism for nearest_voxel_transformation_likelihood for deciding does ndt have problems while matching scans

yvzksgl:
## Description

This PR offers to add state switch mecahism for nearest_voxel_transformation_likelihood.

nearest_voxel_transformation_likelihood is another observable output for deciding does NDT have problems while matching scans. Before this PR it is just in diagnostics, so there is no status switch condition by nearest_voxel_transformation_likelihood, but there is metric for that, if nearest_voxel_transformation_likelihood is smaller than 2.3 then we say ndt's pose estimation is not well. In order to see directly when it's smaller than 2.3 a status switch condition should be added.

## Tests performed

All added options runned through logging simulator and test vehicle. 

## Effects on system behavior

It gives a facility for observing changes in nearest_voxel_transformation_likelihood.

## Pre-review checklist for the PR author

The PR author **must** check the checkboxes below when creating the PR.

- [x] I've confirmed the [contribution guidelines].
- [x] The PR follows the [pull request guidelines].

## In-review checklist for the PR reviewers

The Reviewers **must** check the checkboxes below before approval.

- [ ] The PR follows the [pull request guidelines].

## Post-review checklist for the PR author

The PR author **must** check the checkboxes below before merging.

- [ ] There are no open discussions or they are tracked via tickets.

After all checkboxes are checked, anyone who has write access can merge the PR.

[contribution guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/
[pull request guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/pull-request-guidelines/

Issue #2969, opened 2023-03-01
remove default parameters in planning/control modules


yamazakiTasuku:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I'm convinced that this is not my fault but a bug.

### Description

Currently, default values are set when using declare_parameter. Therefore, when a user sets a parameter, Autoware starts with that default value even if it is not set correctly.

### Expected behavior

When a user sets a parameter, Autoware starts with that default value even if it is not set correctly. This causes Autoware to run with parameters that the user does not intend.

### Actual behavior

There is a possibility of starting with different values than the set parameters.

### Steps to reproduce

Set parameters in the wrong way

### Additional context
Modified module

planning
- [x]  behavior_path_planner
- [x] behavior_velocity_planner
- [x] costmap_generator
- [X] external_velocity_limit_selector
- [x] freespace_planner
- [x] mission_planner
- [X] motion_velocity_smoother
- [x] obstacle_avoidance_planner
- [x] obstacle_cruise_planner
- [X] obstacle_stop_planner
- [x] obstacle_velocity_limiter
- [x] planning_debug_tools
- [x] planning_validator
- [x] route_handler
- [x] rtc_auto_mode_manager
- [x] rtc_interface
- [x] rtc_replayer
- [x] scenario_selector
- [x] static_centerline_optimizer
- [X] surround_obstacle_checker
- 
control
- [X] controller performance analysis
- [x] external_cmd_selector
- [X] joy controller
- [X] lane departure checker
- [x] mpc_lateral_controller
- [x] obstacle_collision_checker
- [x] operation_mode_transition_manager
- [x] pid_longitudinal_controller
- [x]  pure pursuit
- [X] shift decider
- [x] trajectory_follower_base
- [x] trajectory_follower_node
- [x] vehicle_cmd_gate

_No response_

satoshi-ota:
@yamazakiTasuku Thanks for your contribution.
I think it is better to use checkbox to show your progress.

**PROGRESS**

- [ ] obstacle_stop_planner
- [ ] external_velocity_limit_selector
- [ ] motion_velocity_smoother
- [ ] surround_obstacle_checker
- [ ] behavior_velocity_planner
- [ ] ...

PR #4916, opened 2023-09-07
feat(ndt_scan_matcher): adding exe time to diagnostics for checking runtime_monitor

yvzksgl:
## Description

This PR offersto add exe_time to diagnostics.

Exe time is useful for observe does NDT have problems while matching scans. I can say with my observations, if NDT exe_time is exceeded 24ms, it will be most likely started to cannot match scans properly. So while I'm checking runtime_monitor i would like to see NDT exe_time. Formerly this PR is a part of https://github.com/autowarefoundation/autoware.universe/pull/4854.

Related: https://github.com/autowarefoundation/autoware_launch/pull/559

## Tests performed

All added options runned through logging simulator and test vehicle. 

## Effects on system behavior

It gives a facility for observing changes in nearest_voxel_transformation_likelihood and exe_time.

The Reviewers **must** check the checkboxes below before approval.

### DIFF
--- a/localization/ndt_scan_matcher/config/ndt_scan_matcher.param.yaml
+++ b/localization/ndt_scan_matcher/config/ndt_scan_matcher.param.yaml
@@ -87,4 +87,4 @@
     z_margin_for_ground_removal: 0.8
 
     # The execution time which means probably NDT cannot matches scans properly
-    critical_upperbound_exe_time_ms: 24 # [ms]
+    critical_upper_bound_exe_time_ms: 24 # [ms]
--- a/localization/ndt_scan_matcher/config/ndt_scan_matcher.param.yaml
+++ b/localization/ndt_scan_matcher/config/ndt_scan_matcher.param.yaml
@@ -87,4 +87,4 @@
     z_margin_for_ground_removal: 0.8
 
     # The execution time which means probably NDT cannot matches scans properly
-    critical_upper_bound_exe_time_ms: 24 # [ms]
+    critical_upper_bound_exe_time_ms: 100 # [ms]
--- a/localization/ndt_scan_matcher/config/ndt_scan_matcher.param.yaml
+++ b/localization/ndt_scan_matcher/config/ndt_scan_matcher.param.yaml
@@ -85,3 +85,6 @@
 
     # If lidar_point.z - base_link.z <= this threshold , the point will be removed
     z_margin_for_ground_removal: 0.8
+
+    # The execution time which means probably NDT cannot matches scans properly
+    critical_upperbound_exe_time: 24
--- a/localization/ndt_scan_matcher/config/ndt_scan_matcher.param.yaml
+++ b/localization/ndt_scan_matcher/config/ndt_scan_matcher.param.yaml
@@ -87,4 +87,4 @@
     z_margin_for_ground_removal: 0.8
 
     # The execution time which means probably NDT cannot matches scans properly
-    critical_upperbound_exe_time: 24
+    critical_upperbound_exe_time_ms: 24 # [ms]
--- a/localization/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp
+++ b/localization/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp
@@ -101,7 +101,7 @@ NDTScanMatcher::NDTScanMatcher()
   estimate_scores_for_degrounded_scan_(
     declare_parameter("estimate_scores_for_degrounded_scan", false)),
   z_margin_for_ground_removal_(declare_parameter("z_margin_for_ground_removal", 0.8)),
-  critical_upper_bound_exe_time_ms_(24)
+  critical_upper_bound_exe_time_ms_(100)
 {
   (*state_ptr_)["state"] = "Initializing";
   is_activated_ = false;
--- a/localization/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp
+++ b/localization/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp
@@ -141,7 +141,8 @@ NDTScanMatcher::NDTScanMatcher()
     this->declare_parameter<double>("converged_param_nearest_voxel_transformation_likelihood");
 
   lidar_topic_timeout_sec_ = this->declare_parameter<double>("lidar_topic_timeout_sec");
-  critical_upper_bound_exe_time_ms_ = this->declare_parameter("critical_upperbound_exe_time",critical_upper_bound_exe_time_ms_);
+  critical_upper_bound_exe_time_ms_ =
+    this->declare_parameter("critical_upperbound_exe_time", critical_upper_bound_exe_time_ms_);
 
   initial_pose_timeout_sec_ =
     this->declare_parameter("initial_pose_timeout_sec", initial_pose_timeout_sec_);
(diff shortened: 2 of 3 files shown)


localization/ndt_scan_matcher/config/ndt_scan_matcher.param.yaml
```diff
@@ -79,3 +79,6 @@
 
     # If lidar_point.z - base_link.z <= this threshold , the point will be removed
     z_margin_for_ground_removal: 0.8
+
+    # The execution time which means probably NDT cannot matches scans properly
+    critical_upperbound_exe_time: 24
```
kminoda, line 84: Please add a unit
```suggestion
    critical_upperbound_exe_time_ms: 24 # [ms]
```

kminoda (commented):
Thank you for the PR!

localization/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp
```diff
@@ -141,6 +142,9 @@ NDTScanMatcher::NDTScanMatcher()
     "converged_param_nearest_voxel_transformation_likelihood",
     converged_param_nearest_voxel_transformation_likelihood_);
 
+  critical_upperbound_exe_time_ =
+    this->declare_parameter("critical_upperbound_exe_time", critical_upperbound_exe_time_);
```
kminoda, line 146: Please make sure that all the variables are correctly updated through the PR

localization/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp
```diff
@@ -499,6 +509,7 @@ void NDTScanMatcher::callback_sensor_points(
   } else {
     (*state_ptr_)["is_local_optimal_solution_oscillation"] = "0";
   }
+  (*state_ptr_)["exe_time"] = std::to_string(exe_time);
```
kminoda, line 512:
```suggestion
  (*state_ptr_)["execution_time"] = std::to_string(exe_time);
```

localization/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp
```diff
@@ -281,6 +285,12 @@ void NDTScanMatcher::timer_diagnostic()
       diag_status_msg.level = diagnostic_msgs::msg::DiagnosticStatus::ERROR;
       diag_status_msg.message += "skipping_publish_num exceed limit. ";
     }
+    if (
+      state_ptr_->count("exe_time") &&
+      std::stod((*state_ptr_)["exe_time"]) >= critical_upper_bound_exe_time_ms_) {
```
kminoda, line 290:
```suggestion
      state_ptr_->count("execution_time") &&
      std::stod((*state_ptr_)["execution_time"]) >= critical_upper_bound_exe_time_ms_) {
```

kminoda (commented):
LGTM except the following part, thanks :pray:

Motsu-san:
I would like to respond here about [this](https://github.com/autowarefoundation/autoware.universe/pull/4854#discussion_r1315865187)

Your suggestion of adding a condition to generate warnings is very good in itself.
However, I'm aware that the execution time of NDT varies depending on the environment.
It is difficult to approve a warning at 24 ms because there is no evidence or design for Positive/False detecting rate.
For example, there can be an emergency stop system that operates by seeing this warning, and the addition of this condition will simply have the effect of increasing the frequency of activation.
So, how about setting a default value, such as 100 ms, that can be reliably determined this time that the default update cycle cannot be maintained, so that warnings can be issued with matching threshold values in individual environments?
Or has this threshold been discussed and concluded somewhere? ([pull request guidelines](https://autowarefoundation.github.io/autoware-documentation/main/contributing/pull-request-guidelines/))

Motsu-san (changes requested):
I would like to avoid settings that affect the applications this warning is followed by.

localization/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp
```diff
@@ -99,7 +99,8 @@ NDTScanMatcher::NDTScanMatcher()
   regularization_enabled_(declare_parameter("regularization_enabled", false)),
   estimate_scores_for_degrounded_scan_(
     declare_parameter("estimate_scores_for_degrounded_scan", false)),
-  z_margin_for_ground_removal_(declare_parameter("z_margin_for_ground_removal", 0.8))
+  z_margin_for_ground_removal_(declare_parameter("z_margin_for_ground_removal", 0.8)),
+  critical_upper_bound_exe_time_ms_(24)
```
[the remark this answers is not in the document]
kyoichi-sugahara: @Motsu-san 
I apologize for interrupting the conversation.
What is the specific `affect` in this comment?
> I would like to avoid settings that affect the applications this warning is followed by.

Also, defining default value in this case, 100 you proposed is ok but if you don't define in the parameter file in [autoware_launch](https://github.com/autowarefoundation/autoware_launch/blob/main/autoware_launch/config/localization/ndt_scan_matcher.param.yaml), the parameter can't be changed from yaml file.

Share the link about [autoware_launch](https://autowarefoundation.github.io/autoware-documentation/main/contributing/coding-guidelines/ros-nodes/launch-files/#autoware_launch) and [paramter decleration](https://autowarefoundation.github.io/autoware-documentation/main/contributing/coding-guidelines/ros-nodes/parameters/#declare-parameter-function) for your information.

This is also FYI,
In planning/control module, remove all the default value in this [issue](https://github.com/autowarefoundation/autoware.universe/issues/2969) to avoid situation forgetting defining parameter in autoware_launch.

localization/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp
```diff
@@ -281,6 +285,12 @@ void NDTScanMatcher::timer_diagnostic()
       diag_status_msg.level = diagnostic_msgs::msg::DiagnosticStatus::ERROR;
       diag_status_msg.message += "skipping_publish_num exceed limit. ";
     }
+    if (
+      state_ptr_->count("execution_time") &&
+      std::stod((*state_ptr_)["execution_time"]) >= critical_upper_bound_exe_time_ms_) {
+      diag_status_msg.level = diagnostic_msgs::msg::DiagnosticStatus::WARN;
+      diag_status_msg.message += "NDT exe time is too long. ";
```
kminoda, line 292: Would you mind adding the actual execution time as a reference here?

yvzksgl: Sorry I am not sure i understand your request well.

kminoda: What I meant is that the output would be better to be something like
```
NDT exe time is too long (took 39.4 [ms]).
```

localization/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp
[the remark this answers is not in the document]
Motsu-san: @kyoichi-sugahara 
Thank you for your comment.

> What is the specific affect in this comment?

You mean, "You would like to know what application I said above."
Currently, there is a few on Leo Drive, I think. But in the future, more people will use this warning. In that case, they would wonder why the value is 24ms and how they adjust this value to thier purpose.
100ms is needed for NDT's default cycle. so I showed one example easy to avoid side-effects.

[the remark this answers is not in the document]
kyoichi-sugahara: @Motsu-san 
I wandered what is your concern. 
I just couldn't understand and now it's clear thanks!

kminoda:
Please add the autoware_launch PR link in the description, and make sure that both are merged at the same time.

kminoda:
Please fix the following
- Conflict
- semantic pull request (PR titles should not include capital letters)
- pre-commit (fix all the warnings listed)
- DCO (I will make it pass as a maintainer after LGTM, so no need to deal with it)

Motsu-san (approved):
Thank you for your great work! 

yvzksgl:
> Please fix the following
> 
>     * Conflict
> 
>     * semantic pull request (PR titles should not include capital letters)
> 
>     * pre-commit (fix all the warnings listed)
> 
>     * DCO (I will make it pass as a maintainer after LGTM, so no need to deal with it)

I fixed them but github still says there are conflicts. An error should be occured on github or something, i guess. Do you mind if I ask you to take a look? Thank you.

kminoda:
@yvzksgl Sure, but it may take a while since I have other things to do unfortunately.
Those CIs are just general issues and googling or chatgpt-ing may solve the issue.

Motsu-san:
@yvzksgl Please resolve conflicts.

Issue #3137, opened 2023-03-22
Move downloading artifacts outside CMake


esteve:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I've agreed with the maintainers that I can plan this task.

### Description

Some packages download artifacts by default, which is not necessary for compiling them. Additionally, during the process for generating Debian packages, access to the network is restricted, so building these packages fail (see https://github.com/autowarefoundation/autoware/issues/3222#issuecomment-1478085589)

The packages that download ONNX models are:

 - [x] https://github.com/autowarefoundation/autoware.universe/pull/3140
 - [x] https://github.com/autowarefoundation/autoware.universe/pull/3141
 - [x] https://github.com/autowarefoundation/autoware.universe/pull/3142
 - [x] https://github.com/autowarefoundation/autoware.universe/pull/3143
 - [x] https://github.com/autowarefoundation/autoware.universe/pull/3144
 - [x] https://github.com/autowarefoundation/autoware.universe/pull/3145
 - [x] https://github.com/autowarefoundation/autoware.universe/pull/4896
 - [x] https://github.com/autowarefoundation/autoware.universe/pull/4904
 - [x] https://github.com/autowarefoundation/autoware.universe/pull/4905

Other types of artifacts:

 - [x] https://github.com/autowarefoundation/autoware.universe/pull/4110
 - [x] https://github.com/autowarefoundation/autoware/pull/3855
 - [x] https://github.com/autowarefoundation/autoware.universe/pull/4923
 - [x] https://github.com/autowarefoundation/autoware.universe/pull/5367
 - [x] https://github.com/autowarefoundation/autoware.universe/pull/5431

Update ansible scripts to download artifacts files:

 - [x] https://github.com/autowarefoundation/autoware/pull/3375
 - [x] https://github.com/autowarefoundation/autoware/pull/3848
 - [x]  <del>https://github.com/autowarefoundation/autoware/pull/3856</del>
 - [x] https://github.com/autowarefoundation/autoware/pull/3866
 - [x] https://github.com/autowarefoundation/autoware/pull/3916

Update documentation with instructions on how to download ONNX files:

 - [x] https://github.com/autowarefoundation/autoware-documentation/pull/345
 - [x] https://github.com/autowarefoundation/autoware-documentation/pull/456
[...]

PR #4908, opened 2023-09-06
fix(lidar_apollo_segmentation_tvm): add include tier4_autoware_utils and dependency

lexavtanke:
## Description

lidar_apollo_segmentation_tvm package build crushes with command:
```colcon build  --packages-select lidar_apollo_segmentation_tvm --cmake-args -DDOWNLOAD_ARTIFACTS=on```

Output:
```
[0.764s] WARNING:colcon.colcon_core.package_selection:Some selected packages are already built in one or more underlay workspaces:
	'lidar_apollo_segmentation_tvm' is in: /home/alex/autoware/install/lidar_apollo_segmentation_tvm
If a package in a merged underlay workspace is overridden and it installs headers, then all packages in the overlay must sort their include directories by workspace order. Failure to do so may result in build failures or undefined behavior at run time.
If the overridden package is used by another package in any underlay, then the overriding package in the overlay must be API and ABI compatible or undefined behavior at run time may occur.

If you understand the risks and want to override a package anyways, add the following to the command line:
	--allow-overriding lidar_apollo_segmentation_tvm

This may be promoted to an error in a future release of colcon-override-check.
Starting >>> lidar_apollo_segmentation_tvm
--- stderr: lidar_apollo_segmentation_tvm                             
** WARNING ** io features related to pcap will be disabled
/home/alex/autoware/src/universe/autoware.universe/perception/lidar_apollo_segmentation_tvm/src/lidar_apollo_segmentation_tvm.cpp:18:10: fatal error: tier4_autoware_utils/transform/transforms.hpp: No such file or directory
   18 | #include <tier4_autoware_utils/transform/transforms.hpp>
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
gmake[2]: *** [CMakeFiles/lidar_apollo_segmentation_tvm.dir/build.make:76: CMakeFiles/lidar_apollo_segmentation_tvm.dir/src/lidar_apollo_segmentation_tvm.cpp.o] Error 1
gmake[2]: *** Waiting for unfinished jobs....
gmake[1]: *** [CMakeFiles/Makefile2:185: CMakeFiles/lidar_apollo_segmentation_tvm.dir/all] Error 2
gmake: *** [Makefile:146: all] Error 2
---
Failed   <<< lidar_apollo_segmentation_tvm [9.26s, exited with code 2]

Summary: 0 packages finished [9.92s]
[...]

PR #4923, opened 2023-09-07
build(tvm_utility): remove download logic from CMake and update documentation

lexavtanke:
## Description

See https://github.com/autowarefoundation/autoware.universe/issues/3137

According to discussion above remove downloading logic from building stage. For now it will be possible to download model files either by ansible or manually as explained in this [PR](https://github.com/autowarefoundation/autoware-documentation/pull/345)

Also documentation about the package is updated to remove mentions about downloading. Kindly ask for maintainers to read throw the doc to check that it still consistent especially Security consideration part.

## Tests performed

Not applicable.

## Effects on system behavior

Not applicable.


common/tvm_utility/README.md
```diff
@@ -50,35 +50,17 @@ error description.
...
 [Model Zoo](https://github.com/autowarefoundation/modelzoo/) CI pipeline and saved to an S3 bucket.
-This package exports cmake variables and functions for ease of access to those neural networks.
 
 The `get_neural_network` function creates an abstraction for the artifact management.
-The artifacts are saved under the source directory of the package making use of the function; under "data/".
-Priority is given to user-provided files, under "data/user/${MODEL_NAME}/".
-If there are no user-provided files, the function tries to reuse previously-downloaded artifacts.
-If there are no previously-downloaded artifacts, and if the `DOWNLOAD_ARTIFACTS` cmake variable is set, they will be downloaded from the bucket.
-Otherwise, nothing happens.
+Users should provide model files under "data/user/${MODEL_NAME}/". Otherwise, nothing happens and compilation of the package will be skipped.
```
esteve, line 57: If the artifacts are required for compilation, then they should be part of the source tree (via `git-lfs` for example if the files are too big), otherwise it won't be possible to build Debian packages.

perception/lidar_apollo_segmentation_tvm/package.xml
```diff
@@ -21,6 +21,7 @@
   <depend>tf2_eigen</depend>
   <depend>tf2_geometry_msgs</depend>
   <depend>tf2_ros</depend>
+  <depend>tier4_autoware_utils</depend>
```
esteve, line 24: This should be introduced in a separate PR, let me know if you want me to submit it.

lexavtanke: Yes, there is the PR for this changes https://github.com/autowarefoundation/autoware.universe/pull/4908

common/tvm_utility/README.md
lexavtanke: Actually It is a bit confusing for now. Because this packages don't seem to build anyway now because DOWNLOAD_ARTIFACTS flag is disabled by default and this PR kinda show it because last changes was made several months ago and bug was hidden. So here I just follow the same logic. 

And as I understand before we are going to use ansible to provide artifacts. And for packages with tenosorrt support It  is more or less straight forward process as all model conversion from onnx to TRT happens later on the first run not on the build. 

For TVM it is a bit more complicated as it provide models in already compiled form and to use it in the package you will need to provide some model files to build it.  

But if I understand you correctly here. You propose that there should be some default artifacts for all TVM packages which should be part of the source tree and the only way to do it is `git-lfs` because for now `deploy_param.params` file weights 10-20 Mb usually. And If user will want to provide his own models wouldn't it be messy a bit ?

I think there is a kinda similar way to complile the model to TVM before run by the user same as with TRT but I'm not so familiar with TVM.

And what is the idea behind the Debian packages? Should user be able to use them of the shelf with the build in model? I guess in many cases user with have to use model trained on his own data.

ambroise-arm:
I don't understand why it bothers anyone that there is an option to download default-compiled models. As you see the default is to not download the models, and there is already an option for the user to provide their own compiled/tuned models; so to me it looks like it already adheres to the proposed plan in https://github.com/autowarefoundation/autoware.universe/issues/3137.

lexavtanke:
[ambroise-arm](https://github.com/ambroise-arm) The main issue with current approach that it will produce empty Debian package by default as there is no ansible step in their build. And it is not possible to download or provide models in other way. They should be part of the source tree. 

May be you can propose the other way to deal with it? May be it is possible to move model compilation  step further as it done with onnx2trt conversion? So the model files doesn't need for package compilation.  

For now it was offered in Software WG to add header files for default model to the source tree. And the model weights will be provided by the user or downloaded by ansible script. If user want to use his models he will need to rebuild the package or just provide the deploy_param.params file if he just retrain the model. 

Correct me, please, if get wrong how TVM works in general.

common/tvm_utility/README.md
esteve: > For TVM it is a bit more complicated as it provide models in already compiled form and to use it in the package you will need to provide some model files to build it.

As far as I know, we only need the headers for building the packages, not the full models.

> And what is the idea behind the Debian packages? Should user be able to use them of the shelf with the build in model? I guess in many cases user with have to use model trained on his own data.

The Debian packages are a way for users to install Autoware in production environments without having to build it themselves, much like how ROS is distributed. Eventually, Autoware can become part of the ROS distribution, but that will come at a later stage.

esteve:
@ambroise-arm 

> I don't understand why it bothers anyone that there is an option to download default-compiled models. As you see the default is to not download the models, and there is already an option for the user to provide their own compiled/tuned models; so to me it looks like it already adheres to the proposed plan in https://github.com/autowarefoundation/autoware.universe/issues/3137.

Requiring packages to have access to internet during build step can make them vulnerable to supply chain attacks, in addition to not being able to make builds reproducible (e.g. for audits). For example, the Debian and ROS buildfarms restrict access to the internet for packages so they must be built in isolation. The end goal is so that all that logic for downloading artifacts done in CMake is removed and instead use ansible (or instruct the user to download the artifacts manually).

ambroise-arm:
> The main issue with current approach that it will produce empty Debian package by default as there is no ansible step in their build.

I don't understand how removing the option to download models with cmake solves this. The problem is that we lack an ansible step, not that there is a cmake download step as an option. At least from what I see. Is the ansible step mutually exclusive with the cmake download step? I don't know what this ansible step would look like (if you have example code somewhere I am interested), but I suppose it can be used together with the other ways to get the artifact, in the same way that the cmake logic currently prefers user-provided models, and if there are none try to download one.

> May be it is possible to move model compilation step further as it done with onnx2trt conversion? So the model files doesn't need for package compilation.

Yes, but not with the current implementation. I had started to rework the way tvm_utility worked to allow to remove the need for the artifact to be present at compile time, but I didn't see much interest in TVM packages so I parked the work. Maybe I could continue on that, not immediately though.

> Requiring packages to have access to internet during build step can make them vulnerable to supply chain attacks, in addition to not being able to make builds reproducible (e.g. for audits).

It is not a requirement but a convenience option. Currently user-provided artifacts are preferred over downloading one.

> The end goal is so that all that logic for downloading artifacts done in CMake is removed and instead use ansible (or instruct the user to download the artifacts manually).

Ok, if there has been a definitive decision to remove the cmake downloads from all packages for whatever reason I can live with that. But can we agree it isn't justified by any of the points above?

lexavtanke:
I've updated tvm_utils. As example for other tvm based packages. Introduced new constructor for `InferenceEngineTVM` with additional parameter `autoware_data_path`. So package expect to find artifacts for inference (as deploy_graph.json, deploy_lib.so, deploy_param.params) in autoware_data/tvm_utility/ with the same structure as it was in tvm_utility/data folder.

Back compatible with other tvm packages as previous constructor is kept in place

perception/lidar_apollo_segmentation_tvm/package.xml
esteve: I've merged #4908 now, thanks for the patch.

common/tvm_utility/test/yolo_v2_tiny/main.cpp
```diff
@@ -247,7 +247,9 @@ TEST(PipelineExamples, SimplePipeline)
   using PostPT = PostProcessorYoloV2Tiny;
 
   PrePT PreP{config};
-  IET IE{config, "tvm_utility"};
+  std::string home_dir = getenv("HOME");
```
esteve, line 250: If the text fixtures are required to run this, they should be included in this repository, otherwise tests could not be reliably run (e.g. a user may provide different artifacts than what the tests are meant to run with). This is the same scenario as adding the JPEG image, which is also a test fixture.

esteve (changes requested):

common/tvm_utility/test/yolo_v2_tiny/main.cpp
lexavtanke: I'm not sure that it is such a big issue in current situation. Here we are talking about default case. If user will provide different artifacts or put them in the wrong place he will have the error in the test and will be informed that something is not happening as it's expected to be.  

We can introduce some additional env variable and reference it during the test. 
We also can remove test from the package as all tensorrt based packages do not have tests.  

Or git-lfs is the only way for current stage. But as I understand previous discussions It wasn't recognized as appropriate solution.

lexavtanke:
@ambroise-arm Can you please elaborate a bit how tvm_utility should be reworked:

> I had started to rework the way tvm_utility worked to allow to remove the need for the artifact to be present at compile time, but I didn't see much interest in TVM packages so I parked the work.

So we can consider such option as one possible solution.

common/tvm_utility/artifacts/yolo_v2_tiny/test_image_0.jpg
ambroise-arm, line 1: Are we sure that the license of this picture allows us to commit it? There was a reason not to commit it previously.

common/tvm_utility/include/tvm_utility/pipeline.hpp
```diff
@@ -224,6 +224,76 @@ typedef struct
 class InferenceEngineTVM : public InferenceEngine
 {
 public:
+  explicit InferenceEngineTVM(
```
ambroise-arm, line 227: Please don't duplicate the code. What you could do is have `autoware_data_path` as a parameter in the existing function, have it default to `""` and set network_prefix differently depending on whether it is set or not.

ambroise-arm:
@lexavtanke  At a high level `InferenceEngineTVM` needs to have its constructor interface changed so that it requires getting the network name and paths to the files to be dynamically loaded at runtime (the .so, .json and .params) instead of reading it from the artifact header file. And the remaining parameters from that header file need to be extracted from the network at runtime (with TVM's `runtime_mod.GetFunction(...)`). So that there is no dependency on the artifacts at build time.

All the tvm packages in autoware need to be changed to use the new interface. The path to those .so, .json and .params need to become parameters of the nodes making use of tvm_utilty. That probably needs to be integrated with the new way of getting the artifacts you were talking about, in order to make it easy for the user to select a specific backend, and potentially switching between them without too much effort (and also user-provided artifacts).

And then there can be optimizations that can be done to https://github.com/autowarefoundation/modelzoo/ in order to stop generating that header file that won't be used anymore.

common/tvm_utility/artifacts/yolo_v2_tiny/test_image_0.jpg
lexavtanke: Okay, as there is a possible license problem with the fixtures we can provide. I guess we can just remove the test as tensorrt packages don't have tests. And it will solve the current situation https://github.com/autowarefoundation/autoware.universe/pull/4923#discussion_r1331435001 : 
> If the text fixtures are required to run this, they should be included in this repository, otherwise tests could not be reliably run (e.g. a user may provide different artifacts than what the tests are meant to run with). This is the same scenario as adding the JPEG image, which is also a test fixture.

@esteve What do you think? 

And further TVM_utility package can be updated to be able to get model name and path dynamically.

common/tvm_utility/include/tvm_utility/pipeline.hpp
lexavtanke: @ambroise-arm Thank you. Done

lexavtanke:
@ambroise-arm What is the actual purpose of the test in TVM_utility? to be sure that model will be loaded during the inference and inference will happen, correct? 
If so, we can provide the small model (as sum a + b), add it in source tree and use it during the the test. YOLOv2 Tiny will work as example for the user, so he will be able to build and run it if he want.

I guess it is more simple and proper solution for the current situation. As it provide simple test TVM_utility as a tool for other packages and provide example to the user.
What do you think? 

@esteve What are you thoughts about it as well?

ambroise-arm:
> What is the actual purpose of the test in TVM_utility? to be sure that model will be loaded during the inference and inference will happen, correct?

Yes, it doesn't matter what kind of inference, it is just to exercise the TVM runtime and the tvm_utility interface.

> we can provide the small model (as sum a + b), add it in source tree and use it during the the test

Committing the compiled artifacts from TVM to autoware may be a problem as they contain binary files. You may want to double check with maintainers if it is ok.

esteve:
> @esteve What are you thoughts about it as well?

Adding a test model to the tree is fine by me, that's the same as adding a JPEG image as a test fixture, which we already do. My only concern was that some test fixtures were distributed separately from the code, which is problematic, but as long as everything is accessible when cloning this repository, then it's ok.

If the test model is too big, please use `git-lfs` to add it.

common/tvm_utility/artifacts/yolo_v2_tiny/test_image_0.jpg
esteve: Yeah, @ambroise-arm raises a good point about license, but it seems to me that distributing as is, may not be an issue. Otherwise, I'm sure we can use a public domain image as a fixture, or generate it ourselves.

lexavtanke:
Okay, thank you @ambroise-arm and @esteve I will implement it.

xmfcx:
> If so, we can provide the small model (as sum a + b), add it in source tree and use it during the the test. YOLOv2 Tiny will work as example for the user, so he will be able to build and run it if he want.

We shouldn't add binary model files to the git in Autoware Universe. We won't activate Git LFS for the Universe repositories too, if such files need to be a part of the repo, related package(s) should be moved to an external repository and added to the `autoware.repos` file.

So far, [tvm_utility/README.md](https://github.com/autowarefoundation/autoware.universe/blob/ad69c2851b7b84e12c9f0c3b177fb6a9032bf284/common/tvm_utility/README.md#tvm-utility) this repository performs 2 independent tasks:
1) Testing a yolov2 tiny model
2) Providing [tvm_utility-extras.cmake](https://github.com/autowarefoundation/autoware.universe/blob/ad69c2851b7b84e12c9f0c3b177fb6a9032bf284/common/tvm_utility/tvm_utility-extras.cmake) file for other packages to use.

These should be separated into 2 packages.

The package responsible for downloading and testing the packages should be an external repository (Like `tvm_testing`), can have Git LFS or any other downloading method to test it. (Not distributed as a runtime package)

The actual part that is used by other packages can stay in the utility. It is the utility part.

Right now none of the TVM packages are used in anywhere and they are not tested in CI at all.

Maybe separating them like this will allow us to actually test the parts we require.
By separating the heavy test-required parts (requiring to download models) from cheap test parts.

esteve:
@xmfcx

> We shouldn't add binary model files to the git in Autoware Universe. We won't activate Git LFS for the Universe repositories too, if such files need to be a part of the repo, related package(s) should be moved to an external repository and added to the `autoware.repos` file.

I'd say it depends on how big the model files are. If they are tiny (properly tiny), I don't see the problem with adding them to this repository.

However, I'm not opposed to having a separate repository called `tvm_testing` which may contain the models and has Git-LFS enabled.

xmfcx:
@esteve I've checked the file sizes:
```
mfc@mfc-leo:~/Downloads/aaa/yolo_v2_tiny$ tree -s -h
[4.0K]  .
├── [4.0K]  llvm
│   ├── [ 24K]  deploy_graph.json
│   ├── [161K]  deploy_lib.so
│   ├── [ 43M]  deploy_param.params
│   └── [1.6K]  inference_engine_tvm_config.hpp
└── [4.0K]  vulkan
    ├── [ 23K]  deploy_graph.json
    ├── [270K]  deploy_lib.so
    ├── [ 43M]  deploy_param.params
    └── [1.6K]  inference_engine_tvm_config.hpp
```

It's 42~MB for each architecture.

Downloaded and extracted `networks-x86_64.tar.tar` from [here _probably_requires_admin_access_](https://s3.console.aws.amazon.com/s3/object/autoware-modelzoo?region=us-east-2&prefix=networks-x86_64.tar.gz)

xmfcx:
I think we can remove the https://github.com/autowarefoundation/autoware.universe/blob/ad69c2851b7b84e12c9f0c3b177fb6a9032bf284/common/tvm_utility/tvm_utility-extras.cmake file since we will download the models with ansible.

And only other function of this package was to test a tvm pipeline. And we can move this out of universe to do that.
`tvm_utility` shouldn't be a part of debian packages.

And for the dependencies, like:
- https://github.com/leo-drive/autoware.universe/blob/6ef6b548392232a366fa7a4263bb95e8802755a7/perception/lidar_centerpoint_tvm/package.xml#L22
these can be turned into `<test_depend>`

@lexavtanke I can help with creating a new repo for "tvm_testing under awf github.

lexavtanke:
@xmfcx if we want to test tvm_utility ability to load model, we need only  small model with size in the printscreen. 
image
We don't need to have full yolo model in the repo. 
And as I understand the main idea of the tvm_utility package is to provide pipeline for other packages like lidar_centerpoint_tvm and  lidar_apollo_segmentation_tvm.

esteve:
> It's very small then, it's OK to have it in the git tree, considering that we have some large images in the documentation parts of this repository.

It'd good to store it as `git-lfs`, Git repositories are not the best at storing binary data.

> Still, `tvm_utility` doesn't need to have a debian package since it has no other function than being a reference package and having tests in it.

That's irrelevant, even if we don't need a Debian package for `tvm_utiliy`, it's good practice to keep test fixtures in the same place as tests to avoid admin overhead and to make sure tests are easily reproducible. Other projects like for example PCL do that as well, don't see a compelling reason why Autoware should be different.

xmfcx:
> It'd good to store it as `git-lfs`, Git repositories are not the best at storing binary data.

This is stating the obvious but I wish it was a simple decision to just switch this repository to utilize `git-lfs`.

About introducing `git-lfs`, my concerns are stated in [this issue and can be discussed there](https://github.com/autowarefoundation/autoware/issues/3466#issuecomment-1582409848).

An additional thing I've learned about Github LFS billing is,
> In forks, bandwidth and storage usage count against the root of the repository network. 
> Forking and pulling a repository counts against the parent repository's bandwidth limit.
Source: https://docs.github.com/en/billing/managing-billing-for-git-large-file-storage/about-billing-for-git-large-file-storage#about-billing-for-git-large-file-storage 

I've realized this when trying to pull LFS files of https://github.com/lgsvl/simulator and it said the owner didn't pay for the LFS billing.
I had to fork the repo, turn it into an archive, download it as a zip to get access to the LFS files since they didn't pay for the bills.
And forked repo wasn't even able to pull LFS files because forked repo billing is still tied to the original repo.
Only way to keep working with it was to recreate the repository and push it there.

Because of these reasons, I don't see it as a good option for current Universe repository.

> That's irrelevant, even if we don't need a Debian package for `tvm_utiliy`, it's good practice to keep test fixtures in the same place as tests to avoid admin overhead and to make sure tests are easily reproducible. Other projects like for example PCL do that as well, don't see a compelling reason why Autoware should be different.

The reason I wanted to move the package out of the Universe repository was following:
- The package had 2 functions:
   1. Testing the tvm pipeline with a reference implementation.
   2. Providing a cmake download script.
- Download function is being moved to ansible, now it is not needed anymore.
- If testing the pipeline requires a 40MB download, we can't have it in Autoware Universe, it needs to move out.
   - With recent information, I've learned that it's just about 25KB, I am also not too inclined to move it out too.

But with all these information, relevant or not, this package doesn't need a debian package because it is just a test package.

Still, I think it is better to rename this package to `tvm_testing` since that's the only job it does.

ambroise-arm:
For reference, additional previous discussion on using git-lfs (here specifically for TVM models): https://github.com/orgs/autowarefoundation/discussions/2508

> this package doesn't need a debian package

I agree with this part.

> Still, I think it is better to rename this package to `tvm_testing` since that's the only job it does.

It is not there only for testing. This file: https://github.com/autowarefoundation/autoware.universe/blob/main/common/tvm_utility/include/tvm_utility/pipeline.hpp provides a common interface for all Autoware packages making use of TVM to access the models and organize the inference pipeline. Renaming to `tvm_testing` sounds misleading to me.

esteve:
@ambroise-arm 

> >  this package doesn't need a debian package
>
> I agree with this part.

I disagree. If other packages depend on `tvm_utility` and those packages generate Debian packages, it most definitely needs a Debian package.

ambroise-arm:
After removing the artifact download functionality from `tvm_utility`, its only export will be the header file, which is of no use to other packages at runtime. And in my head the Autoware debian packages would only contain compiled binaries and other files needed at runtime. But also I haven't really followed the debian package discussions, so I may be completely off. Disregard my previous comment if it doesn't match reality.

esteve:
@ambroise-arm Debian packages are built one by one in isolation to ensure that dependencies are correct, they are not built all at once from the source tree, therefore if a package depends on anything at build time from another package, that dependency needs to be packaged and exported to downstream packages.

In this case, the Debian package for `tvm_utility` would include the necessary header files that downstream packages depend on so that the code for such dependent packages can be compiled.

If you have any questions about how the Debian generation process works, I'll be happy to answer them.

lexavtanke:
@esteve and @ambroise-arm I've updated the PR. Now there is a small test case. It run a small model with one node abs. It computes element-wise absolute of input tensor. 
Screenshot from 2023-10-04 17-24-13

I kept yolo_v2_tiny in the package as example pipeline. So the user can build and test it if he want. It's build is disableв by default.  

For now artifact ansible role should be updated to download models for lidar_centerpoint_tvm and lidar_apollo_segmentation_tvm. Also documents for manual downloading should be updated.

And final change should be done in TVM packages. tvm_utility::pipeline::InferenceEngineTVM constructor should receive path to the autoware_data folder with models files. 

Сmake function get_neural_network builds dependency for tvm packages and should be renamed. I think it will be better to do in separate PR.

esteve:
@lexavtanke thanks for addressing my feedback, I think the small model you've added looks good.

As for the `yolo_v2_tiny` model, if you still want it to be included and not add the model files, it should be as an example program and not as a test. Test are meant to test the code and if the goal of the the `yolo_v2_tiny` example is to show how to use `tvm_utility`, then it should be moved to an `examples` folder and rewritten as a program instead of a test, a user could pass a path as an argument to the example program, so the model could be stored anywhere. Otherwise it's confusing whether the test is a test or an example.

We can discuss in the future moving the model fixtures to a separate repository if we want to store them with `git-lfs`, but for now it's not possible to enable `git-lfs` for autoware.universe

lexavtanke:
@esteve Thanks for your feedback. I've updated yolo_v2_tiny example. It is separated form the test case completely, and user can build it separately and launch with the launch file if he wants. 

I also separated x86 model and arm model as @ambroise-arm offered in direct messages. So for now tests will works for both systems. 

There is an only issue for now. As there is binary model files spell-check throws errors on it. Do you now work around for it? Is it possible to ignore the whole file? If so how can it be done? Should CI/CD scripts be modified? 

I think that PR is ready for a review again.

lexavtanke:
@esteve Do you know who has right to update [this file](https://raw.githubusercontent.com/tier4/autoware-spell-check-dict/main/.cspell.json)?
 It's used by spell-check and I think .so file extension should be added to it.

lexavtanke:
I have found the repo for spell-check-dict https://github.com/tier4/autoware-spell-check-dict/tree/main and will PR.

common/tvm_utility/CMakeLists.txt
```diff
@@ -22,23 +22,24 @@ autoware_package()
 
 # Library
 
-set(TVM_UTILITY_NODE_LIB_HEADERS
-  "include/${PROJECT_NAME}/pipeline.hpp"
-)
+set(TVM_UTILITY_NODE_LIB_HEADERS "include/${PROJECT_NAME}/pipeline.hpp")
```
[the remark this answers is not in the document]
lexavtanke: Sorry, I guess cmake lint was angry on some white spaces added by vscode and I use auto fix on it. I put it back.

common/tvm_utility/example/yolo_v2_tiny/main.cpp
```diff
@@ -239,24 +243,39 @@ class PostProcessorYoloV2Tiny : public tvm_utility::pipeline::PostProcessor<std:
...
+
+int main(int argc, char * argv[])
+{
+  // init node to use parameters
+  rclcpp::init(argc, argv);
+  auto node = rclcpp::Node::make_shared("tvm_yolo_example");
+  node->declare_parameter("image_filename", IMAGE_FILENAME);
+  node->declare_parameter("label_filename", LABEL_FILENAME);
+  node->declare_parameter("anchor_filename", ANCHOR_FILENAME);
+  node->declare_parameter("data_path", "");
```
ambroise-arm, line 262: Parameter declaration should follow https://autowarefoundation.github.io/autoware-documentation/main/contributing/coding-guidelines/ros-nodes/parameters/

- no default value in the code
- a config file with the default values
- a schema file for the template
- you can then make use of auto-generation of parameter table in the documentation (https://github.com/autowarefoundation/autoware.universe/commit/63033beebc9d45b6185b8ec789f41a7b379ac04a#diff-f5557a723bf6eba2f59cb3ac58f670f0ad7ea7f189fa7c868289ffbc2ab574d9)

lexavtanke: It seems like there is an issue with auto-generation 
image
https://github.com/autowarefoundation/autoware.universe/tree/main/perception/lidar_apollo_segmentation_tvm_nodes

ambroise-arm: That's expected. It gets rendered here: https://autowarefoundation.github.io/autoware.universe/main/perception/lidar_apollo_segmentation_tvm_nodes/#parameters

ambroise-arm: Does `$(find-pkg-share tvm_utility)` in config/yolo_v2_tiny_example.param.yaml work? I have never seen it used in a config file before.

esteve (approved):
@lexavtanke LGTM, thanks for the patience iterating with us :slightly_smiling_face:

common/tvm_utility/example/yolo_v2_tiny/main.cpp
lexavtanke: yes, it works with `allow_substs="true"` in launch file. It takes a while to find it for [ros2](https://docs.ros.org/en/foxy/Releases/Release-Galactic-Geochelone.html).

ambroise-arm: that's great!

ambroise-arm (approved):

esteve:
@angry-crab could you approve this PR when you have a moment? You're one of the codeowners of this package, I approved these changes, but they won't merge unless a codeowner approves it as well. Thanks!

ambroise-arm:
Am I not a codeowner of `tvm_utility`?

esteve:
@ambroise-arm probably, but only @angry-crab had the codeowner logo, I assumed he was the only one. Anyway, @ambroise-arm thanks for approving this PR

image

PR #4942, opened 2023-09-10
fix(mission_planner): routing api uninitialization

TakaHoribe:
## Description

To fix the issue: "the set_route api is rejected even though the autoware_state is `WAIT_FOR_ROUTE`".

Publish `routing_state = UNKNOWN` until the routing system is not initialized.

Without this PR, it gets `autoware_state = WAIT_FOR_ROUTE` and `routing_state = UNSET` even if the mission_planner is not ready. Due to this inconsistency, calling set_route while in the wait_for_route state causes the API to fail.

Before this PR, when the ego pose is not set, the route state is UNSET:
```
ros2 topic echo /planning/mission_planning/route_state
stamp:
  sec: 1694341968
  nanosec: 326244268
state: 0
```

After this PR, when the ego pose is not set, the route state is UNKNOWN:
```
ros2 topic echo /planning/mission_planning/route_state
stamp:
  sec: 1694342360
  nanosec: 163490802
state: 1
---
```

## Related links

[TIERIV Internal ticket](https://tier4.atlassian.net/browse/RT1-3637)

## Tests performed

Run psim

## Notes for reviewers

check 

## Interface changes

None

## Effects on system behavior

route_state behavior will change

### DIFF
--- a/planning/mission_planner/src/mission_planner/mission_planner.cpp
+++ b/planning/mission_planner/src/mission_planner/mission_planner.cpp
@@ -141,6 +141,7 @@ void MissionPlanner::checkInitialization()
   }
 
   // All data is ready. Now API is available.
+  RCLCPP_INFO(get_logger(), "Route API is ready.");
   change_state(RouteState::Message::UNSET);
   data_check_timer_->cancel();  // stop timer callback
 }
--- a/planning/mission_planner/src/mission_planner/mission_planner.cpp
+++ b/planning/mission_planner/src/mission_planner/mission_planner.cpp
@@ -121,7 +121,28 @@ MissionPlanner::MissionPlanner(const rclcpp::NodeOptions & options)
   adaptor.init_srv(srv_clear_mrm_route_, this, &MissionPlanner::on_clear_mrm_route);
   adaptor.init_sub(sub_modified_goal_, this, &MissionPlanner::on_modified_goal);
 
+  // Route state will be published when the node gets ready for route api after initialization,
+  // otherwise the mission planner rejects the request for the API.
+  data_check_timer_ = create_wall_timer(
+    std::chrono::milliseconds(100), std::bind(&MissionPlanner::checkInitialization, this));
+}
+
+void MissionPlanner::checkInitialization()
+{
+  if (!planner_->ready()) {
+    RCLCPP_INFO_THROTTLE(
+      get_logger(), *get_clock(), 5000, "waiting lanelet map... Route API is not ready.");
+    return;
+  }
+  if (!odometry_) {
+    RCLCPP_INFO_THROTTLE(
+      get_logger(), *get_clock(), 5000, "waiting odometry... Route API is not ready.");
+    return;
+  }
+
+  // All data is ready. Now API is available.
   change_state(RouteState::Message::UNSET);
+  data_check_timer_->cancel();  // stop timer callback
 }
 
 void MissionPlanner::on_odometry(const Odometry::ConstSharedPtr msg)
--- a/planning/mission_planner/src/mission_planner/mission_planner.hpp
+++ b/planning/mission_planner/src/mission_planner/mission_planner.hpp
@@ -82,6 +82,9 @@ private:
   void on_map(const HADMapBin::ConstSharedPtr msg);
   void on_reroute_availability(const RerouteAvailability::ConstSharedPtr msg);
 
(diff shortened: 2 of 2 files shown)


TakaHoribe:
@kosuke55 @isamu-takagi This is to fix a specific issue, but I'm not sure this fix is the best way. Would you check this.

isamu-takagi:
UNKNOWN is reserved as an invalid value, so please avoid publishing this as a normal value.

TakaHoribe:
@isamu-takagi 
> UNKNOWN is reserved as an invalid value, so please avoid publishing this as a normal value.

Thank you for your comment. Do you have an idea how to tell the routing system is not initialized yet? Should we add the new `UNINITIALIZED` field in the message?

planning/mission_planner/src/mission_planner/mission_planner.cpp
```diff
@@ -301,18 +329,15 @@ void MissionPlanner::on_set_route(
...
 
-  if (state_.state != RouteState::Message::UNSET) {
-    throw component_interface_utils::ServiceException(
-      ResponseCode::ERROR_ROUTE_EXISTS, "The route is already set.");
-  }
-  if (!planner_->ready()) {
+  if (state_.state == RouteState::Message::UNKNOWN) {  // should be UNINITIALIZED
     throw component_interface_utils::ServiceException(
-      ResponseCode::ERROR_PLANNER_UNREADY, "The planner is not ready.");
+      ResponseCode::ERROR_ROUTE_EXISTS, "The planner is not ready.");
```
isamu-takagi, line 334:
```suggestion
      ResponseCode::ERROR_PLANNER_UNREADY, "The planner is not ready.");
```

planning/mission_planner/src/mission_planner/mission_planner.cpp
```diff
@@ -574,17 +586,13 @@ void MissionPlanner::on_change_route(
 {
   using ResponseCode = autoware_adapi_v1_msgs::srv::SetRoute::Response;
 
-  if (state_.state != RouteState::Message::SET) {
+  if (state_.state == RouteState::Message::UNKNOWN) {  // should be UNINITIALIZED
     throw component_interface_utils::ServiceException(
-      ResponseCode::ERROR_INVALID_STATE, "The route hasn't set yet. Cannot reroute.");
+      ResponseCode::ERROR_ROUTE_EXISTS, "The planner is not ready.");
```
isamu-takagi, line 591:
```suggestion
      ResponseCode::ERROR_PLANNER_UNREADY, "The planner is not ready.");
```

isamu-takagi:
If you want to know whether a route request is possible, I think you should add a flag to indicate this. This is because when considering reroutes and MRM, it is easier to handle them separately from states.

TakaHoribe:
@isamu-takagi 
>If you want to know whether a route request is possible, I think you should add a flag to indicate this. This is because when considering reroutes and MRM, it is easier to handle them separately from states.

I feel this makes it more confusing. The issue we want to fix now is, "the set_route API is rejected even though the autowre_state is WAIT_FOR_ROUTE".

Now the autoware_state is deprecated, but if the role of the `routing_state` is to tell the status of the routing system, it should contain the uninitialized information.


> it is easier to handle them separately from states.
MRM system can see the status through the response of the API, so it does not matter here.

TakaHoribe:
Note:

Even if the `UNINITIALIZED` status is supported, sometimes the set_route might be rejected due to the planning status.

The status of whether the module can receive the route should be managed at a higher level, not only in the mission_planner.

@isamu-takagi will keep investigating the better way. (I'll make an issue to track this.)

TakaHoribe:
>To fix the issue: "the set_route api is rejected even though the autoware_state is WAIT_FOR_ROUTE".

I changed the way to fix the issue. Now it does not publish the route_state until the node is ready for the API.

kosuke55 (approved):

isamu-takagi (approved):
@TakaHoribe 
LGTM. ~However, it need to be fixed so it doesn't wait in the constructor in the future. Because it blocks the initialization of other components.~

I had a misunderstanding. No problem.

TakaHoribe:
@isamu-takagi Thank you for your review. It is already no-wait in the constructor. The time starts running after the constructor is completed, and the route_state msg will be published when all conditions checked in the timer are satisfied.

PR #4946, opened 2023-09-11
feat(ar_tag_based_localizer): add ekf_pose subscriber

SakodaShintaro:
## Description

Add code to subscribe to ekf_pose, and the Pose produced by AR tag-based localizer is compared to Ekf's latest Pose.

The Pose produced by AR tag-based self-location estimation will be published only when the following conditions are met.

* The time difference should be within 5 seconds.
* The position difference must be within 10m

This difference threshold can be changed with a parameter.

## Tests performed

The following have been tested

* The AR tag-based localizer works correctly with `logging_simulator`.
* In the [ar_tag_based_localizer sample AWSIM data](https://github.com/autowarefoundation/autoware.universe/tree/main/localization/ar_tag_based_localizer#rosbag), a False Positive for ID 32 appears at the end of the run, but it is properly ignored even if it is registered in an odd location on the map.
* The default Autoware without `pose_source:=artag` also works correctly (`logging_simulator`).

## Effects on system behavior

AR tag-based localizer nodes used to publish pose with all detected tags, but now only those within a specified range are published.

Other nodes remain unchanged.


KYabuuchi:
Hi! Thanks for creating PR.
I registered marker32 (the marker that is falsely detected in the sample rosbag) on the map and tested this PR. 
Then I received a message as shown in the figure below. , so I think this PR is working as intended.

outlier_rejection_works_well

localization/ar_tag_based_localizer/src/ar_tag_based_localizer_core.cpp
```diff
@@ -268,6 +279,40 @@ void ArTagBasedLocalizer::publish_pose_as_base_link(
   pose_with_covariance_stamped.header.frame_id = "map";
   pose_with_covariance_stamped.pose.pose = tf2::toMsg(map_to_base_link);
 
+  // If latest_ekf_pose_ is older than <ekf_time_tolerance_> seconds compared to current frame, it
+  // will not be published.
+  const double curr_sec = msg.header.stamp.sec + msg.header.stamp.nanosec * 1e-9;
+  const double ekf_sec =
+    latest_ekf_pose_.header.stamp.sec + latest_ekf_pose_.header.stamp.nanosec * 1e-9;
+  if (ekf_sec + ekf_time_tolerance_ < curr_sec) {
```
KYabuuchi, line 287: I guess `double` type may not accurately store this value. Using `rclcpp::Time` and `rclcpp::Duration` might be better. 
```suggestion
  const rclcpp::Duration tolerance{
    static_cast<int32_t>(ekf_time_tolerance_),
    static_cast<uint32_t>(ekf_time_tolerance_ - std::floor(ekf_time_tolerance_))};
  if (rclcpp::Time(latest_ekf_pose_.header.stamp) + tolerance < rclcpp::Time(msg.header.stamp)) {
```

localization/ar_tag_based_localizer/README.md
```diff
@@ -18,10 +18,11 @@ This package includes two nodes.
...
 
-| Name                  | Type                           | Description  |
-| :-------------------- | :----------------------------- | :----------- |
-| `~/input/image`       | `sensor_msgs::msg::Image`      | Camera Image |
-| `~/input/camera_info` | `sensor_msgs::msg::CameraInfo` | Camera Info  |
+| Name                  | Type                                            | Description                     |
+| :-------------------- | :---------------------------------------------- | :------------------------------ |
+| `~/input/image`       | `sensor_msgs::msg::Image`                       | Camera Image                    |
+| `~/input/camera_info` | `sensor_msgs::msg::CameraInfo`                  | Camera Info                     |
+| `~/input/ekf_pose`    | `geometry_msgs::msg::PoseWithCovarianceStamped` | EKF Pose without IMU correction |
```
KYabuuchi, line 25: Could you describe here or elsewhere that how the `ekf_pose` is used in this node?

localization/ar_tag_based_localizer/src/ar_tag_based_localizer_core.cpp
SakodaShintaro: Thank you! I didn't know there was a class for time management. I agree that it is better to use this.
Fixed in https://github.com/autowarefoundation/autoware.universe/pull/4946/commits/3c8fb046b1cec1ff87e74bbfaa3bc0937b5678a0

localization/ar_tag_based_localizer/README.md
KYabuuchi: thank you :+1:

localization/ar_tag_based_localizer/src/ar_tag_based_localizer_core.cpp
KYabuuchi: I'm sorry. I suggested wrong code. :bow: Since the second argument of rclcpp::Duration is nanoseconds, we need to pass a number multiplied by 10^9. 
```cpp
  const rclcpp::Duration tolerance{
    static_cast<int32_t>(ekf_time_tolerance_),
    static_cast<uint32_t>((ekf_time_tolerance_ - std::floor(ekf_time_tolerance_)) * 1e9)};
```

Motsu-san (changes requested):
Thank you for your pull-request!

> The time difference should be within 5 seconds.
> The position difference must be within 10m

Could you let us know how to adjust or design these values?
I think it is important that the above is opened to users for further development.
These seem to be changed by camera sensor resolution and focal specs, odometry accuracy etc.
It would be fine just to add some comments to README if you don't have enough time.

localization/ar_tag_based_localizer/src/ar_tag_based_localizer_core.cpp
SakodaShintaro: Indeed, that is so true. I had missed it.

Fixed in https://github.com/autowarefoundation/autoware.universe/pull/4946/commits/da1f56483b1c972d418512fb87d18cc57f021776

KYabuuchi (approved):
:+1:  LGTM

SakodaShintaro:
@Motsu-san 
Added comment on how to determine the value.
https://github.com/autowarefoundation/autoware.universe/pull/4946/commits/8f1264a5f035a04f4feae96fd887427b8670b4fc

Motsu-san (approved):

PR #4981, opened 2023-09-13
fix(ring_outlier_filter): change a parameter to be less misleading

sykwer:
## Description
The name of a ring_outlier_filter parameter (`max_rings_num`) has been changed to be less misleading (`max_ring_index`), and the source code has been modified accordingly.

In a project using Autoware, `max_rings_num` was set to `128` (i.e. the range of ring index was set to be `0~127`), but actually the maximum value of ring index was `128`.
To prevent this kind of thing, the parameter name should be changed to be less misleading.

[TIER IV internal link](https://tier4.atlassian.net/wiki/spaces/~5ed0b4584824b20c18371c06/pages/2885976248/pilot-auto.xx1+AWSIM+ring+outlier+filter+die)

## Tests performed
Even in projects with a maximum ring index of 128, it worked without error messages.

### DIFF
--- a/sensing/pointcloud_preprocessor/docs/ring-outlier-filter.md
+++ b/sensing/pointcloud_preprocessor/docs/ring-outlier-filter.md
@@ -27,7 +27,7 @@ This implementation inherits `pointcloud_preprocessor::Filter` class, please ref
 | `distance_ratio`          | double  | 1.03          |             |
 | `object_length_threshold` | double  | 0.1           |             |
 | `num_points_threshold`    | int     | 4             |             |
-| `max_rings_num`           | uint_16 | 128           |             |
+| `max_ring_index`          | uint_16 | 128           |             |
 
 ## Assumptions / Known limits
 
--- a/sensing/pointcloud_preprocessor/docs/ring-outlier-filter.md
+++ b/sensing/pointcloud_preprocessor/docs/ring-outlier-filter.md
@@ -22,12 +22,12 @@ This implementation inherits `pointcloud_preprocessor::Filter` class, please ref
 
 ### Core Parameters
 
-| Name                      | Type    | Default Value | Description                |
-| ------------------------- | ------- | ------------- | -------------------------- |
-| `distance_ratio`          | double  | 1.03          |                            |
-| `object_length_threshold` | double  | 0.1           |                            |
-| `num_points_threshold`    | int     | 4             |                            |
-| `max_ring_index`          | uint_16 | 128           |`128` corresponds to vls128.|
+| Name                      | Type    | Default Value | Description                  |
+| ------------------------- | ------- | ------------- | ---------------------------- |
+| `distance_ratio`          | double  | 1.03          |                              |
+| `object_length_threshold` | double  | 0.1           |                              |
+| `num_points_threshold`    | int     | 4             |                              |
+| `max_ring_index`          | uint_16 | 128           | `128` corresponds to vls128. |
 
 ## Assumptions / Known limits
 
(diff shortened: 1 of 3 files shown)


sensing/pointcloud_preprocessor/docs/ring-outlier-filter.md
```diff
@@ -27,7 +27,7 @@ This implementation inherits `pointcloud_preprocessor::Filter` class, please ref
 | `distance_ratio`          | double  | 1.03          |             |
 | `object_length_threshold` | double  | 0.1           |             |
 | `num_points_threshold`    | int     | 4             |             |
-| `max_rings_num`           | uint_16 | 128           |             |
+| `max_ring_index`          | uint_16 | 128           |             |
```
yukkysaito, line 30: @sykwer Could you please use velodyne 128 as an example here and write more details in the description?
Also, is 127 correct for index?
```suggestion
| `max_ring_index`          | uint_16 | 127           |             |
```

sykwer: Sure.

> In a project using Autoware, max_rings_num was set to 128 (i.e. the range of ring index was set to be 0~127), but actually, the maximum value of ring index was 128.
In the case described above, `128` is correct for `max_ring_index`, so I'd like to set the default value to be `128`.

taikitanaka3: it looks like 128 is better because it is vls128 and i think implementation should be corrected according to the parameter.

yukkysaito: >it looks like 128 is better because it is vls128 and i think implementation should be corrected according to the parameter.

Agree with you :+1:

sensing/pointcloud_preprocessor/docs/ring-outlier-filter.md
```diff
@@ -22,12 +22,12 @@ This implementation inherits `pointcloud_preprocessor::Filter` class, please ref
...
-| `distance_ratio`          | double  | 1.03          |             |
-| `object_length_threshold` | double  | 0.1           |             |
-| `num_points_threshold`    | int     | 4             |             |
-| `max_rings_num`           | uint_16 | 128           |             |
+| Name                      | Type    | Default Value | Description                  |
+| ------------------------- | ------- | ------------- | ---------------------------- |
+| `distance_ratio`          | double  | 1.03          |                              |
+| `object_length_threshold` | double  | 0.1           |                              |
+| `num_points_threshold`    | int     | 4             |                              |
+| `max_ring_index`          | uint_16 | 128           | `128` corresponds to vls128. |
```
yukkysaito, line 30: For index, I am not sure if it is 127 or 128 when using vls128. max_rings_num would be easier to understand if it is 128.

yukkysaito: related comment: https://github.com/autowarefoundation/autoware.universe/pull/4981#discussion_r1325316752
cc @taikitanaka3

tkimura4: @yukkysaito 

> In a project using Autoware, max_rings_num was set to 128 (i.e. the range of ring index was set to be 0~127), but actually the maximum value of ring index was 128.
To prevent this kind of thing, the parameter name should be changed to be less misleading.

Based on this explanation, wouldn't it be more appropriate to use 'max_rings_index' rather than 'max_rings_num'?

yukkysaito: @tkimura4 If it weren't for the description, I would think that vls128 would be correct to specify 127 based on the name max_ring_index.

miursh:
We are closing this PR as there are no response for months. Please re-open if the issue still persists.

PR #5160, opened 2023-09-27
feat(time_synchronizer): organize twist input

kminoda:
## Description

Organize twist input of time_synchronizer as well as concatenate_filter

## Tests performed

Not applicable

## Effects on system behavior

The node will subscribe to /sensing/vehicle_velocity_converter/twist_with_covariance from now on. But we see no practical effects on the system behavior given that the node is not used yet in the system as far as we know.

### DIFF
--- a/sensing/pointcloud_preprocessor/include/pointcloud_preprocessor/time_synchronizer/time_synchronizer_nodelet.hpp
+++ b/sensing/pointcloud_preprocessor/include/pointcloud_preprocessor/time_synchronizer/time_synchronizer_nodelet.hpp
@@ -68,7 +68,6 @@
 #include <tier4_autoware_utils/ros/debug_publisher.hpp>
 #include <tier4_autoware_utils/system/stop_watch.hpp>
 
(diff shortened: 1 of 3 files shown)


yukkysaito:
What is the node that uses this synchronizer?

kminoda:
@miursh Would you answer the above question?

miursh:
@yukkysaito This node is currently not used I think. 
Since the current concatenate node has two function ("time synchronization" and "pointcloud concatenation") in it, we (mainly @YoshiRi ) are planning to separate those to another nodes. 
https://github.com/autowarefoundation/autoware.universe/blob/main/sensing/pointcloud_preprocessor/docs/concatenate-data.md#node-separation-options-for-future

yukkysaito:
@kminoda @miursh Thanks :+1: 
Note that the sensing component is a pre-processing step of the localization component, so it is better to avoid using localization output.

kminoda:
@yukkysaito Understood. The reason why the proposed implementation both supports twist (for sensing) and odom (for localization) is that the node can be use in both sensing and perception module.

yukkysaito (approved):
LGTM

Issue #4915, opened 2023-09-07
Vehicle can not move because steering is not controlable when stopped


brkay54:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I'm convinced that this is not my fault but a bug.

### Description

In vehicle_cmd_gate, if the vehicle is in the stopped state it sends the current steering to the low level. However, this causes a bug for the longitudinal controller because of keep_stopped_until_steering_converge parameter. 

Because of this, in planning_simulator, we can not control the steering when velocity is zero in the manual controller.

I think moving keep_stopped_steering_convergence parameter from pid_controller to vehicle_cmd_gate is a good option to fix it.

### Expected behavior

The vehicle is moving when it is in autonomous mode.

### Actual behavior

The vehicle is not moving when it is in autonomous mode.

### Steps to reproduce

1. open planning simulator
2. add manual controller panel
3. try to control steering when velocity is 0

feng123a52:
请问使用禾赛的激光雷达实车在外部传入自己的点云数据，autoware出现了这么个错，The pointcloud layout is not compatible with PointXYZIRCAEDT/PointXYZIRC. Aborting该如何解决

knzo25:
@feng123a52 
Ever since https://github.com/autowarefoundation/autoware.universe/pull/6996
we use a different pointcloud type. You could either modify Hesai drivers or use Nebula (tier4's driver).
Also, please create new issues/discussions and if you want to ease communications, I highly recommend sticking to English

PR #5183, opened 2023-09-29
fix(vehicle_cmd_gate): fix the problems while transition from/to stopped state

brkay54:
## Description

This PR solves the following issues:
* If the user sends stop command to autoware while driving, autoware does not do steering control while stopping: https://github.com/autowarefoundation/autoware.universe/issues/5796
* https://github.com/autowarefoundation/autoware.universe/issues/4915
* https://github.com/autowarefoundation/autoware.universe/issues/6044
## Tests performed
Tested on Psim, working as we want.

### DIFF
--- a/control/vehicle_cmd_gate/src/vehicle_cmd_gate.cpp
+++ b/control/vehicle_cmd_gate/src/vehicle_cmd_gate.cpp
@@ -430,13 +430,17 @@ void VehicleCmdGate::publishControlCommands(const Commands & commands)
 
   // Check engage
   if (!is_engaged_) {
-    filtered_commands.control = createStopControlCmd();
+    filtered_commands.control.longitudinal = createLongitudinalStopControlCmd();
   }
 
   // Check pause. Place this check after all other checks as it needs the final output.
   adapi_pause_->update(filtered_commands.control);
   if (adapi_pause_->is_paused()) {
-    filtered_commands.control = createStopControlCmd();
+    if (is_engaged_) {
+      filtered_commands.control.longitudinal = createLongitudinalStopControlCmd();
+    } else {
+      filtered_commands.control = createStopControlCmd();
+    }
   }
 
   // Check if command filtering option is enable
@@ -595,6 +599,17 @@ AckermannControlCommand VehicleCmdGate::createStopControlCmd() const
   cmd.lateral.steering_tire_rotation_rate = 0.0;
   cmd.longitudinal.speed = 0.0;
   cmd.longitudinal.acceleration = stop_hold_acceleration_;
+
+  return cmd;
+}
+
+LongitudinalCommand VehicleCmdGate::createLongitudinalStopControlCmd() const
+{
+  LongitudinalCommand cmd;
+  const auto t = this->now();
+  cmd.stamp = t;
+  cmd.speed = 0.0;
+  cmd.acceleration = stop_hold_acceleration_;
 
   return cmd;
 }
--- a/control/vehicle_cmd_gate/src/vehicle_cmd_gate.hpp
+++ b/control/vehicle_cmd_gate/src/vehicle_cmd_gate.hpp
@@ -55,6 +55,7 @@ namespace vehicle_cmd_gate
 using autoware_adapi_v1_msgs::msg::MrmState;
 using autoware_adapi_v1_msgs::msg::OperationModeState;
 using autoware_auto_control_msgs::msg::AckermannControlCommand;
+using autoware_auto_control_msgs::msg::LongitudinalCommand;
 using autoware_auto_vehicle_msgs::msg::GearCommand;
 using autoware_auto_vehicle_msgs::msg::HazardLightsCommand;
 using autoware_auto_vehicle_msgs::msg::SteeringReport;
@@ -220,6 +221,7 @@ private:
   // Algorithm
   AckermannControlCommand prev_control_cmd_;
   AckermannControlCommand createStopControlCmd() const;
+  LongitudinalCommand createLongitudinalStopControlCmd() const;
   AckermannControlCommand createEmergencyStopControlCmd() const;
 
   std::shared_ptr<rclcpp::Time> prev_time_;


brkay54:
Hi @TakaHoribe, about the problem mentioned this [issue](https://github.com/autowarefoundation/autoware.universe/issues/4915), I moved the steering converge checking from controllers to vehicle_cmd_gate to solve this issue. If this method is okay, please let me know, I will create another PR to remove steering converge checking from controllers.

brkay54:
Hi @TakaHoribe @tkimura4, sorry to disturb you, but just I want to kindly ping you.

tkimura4:
@kosuke55 

Could you review this PR?

> Hi @TakaHoribe, about the problem mentioned this https://github.com/autowarefoundation/autoware.universe/issues/4915, I moved the steering converge checking from controllers to vehicle_cmd_gate to solve this issue. If this method is okay, please let me know, I will create another PR to remove steering converge checking from controllers.

What do you think about this opinion?

TakaHoribe:
@brkay54 Thank you for your PR.

It is reasonable not to use the stop steering command when the vehicle is trying to stop. However, I can't see the purpose of the move of the steering convergence check. What is the reason why you think the check should be in the gate?

brkay54:
@TakaHoribe, thank you for the reply. 
> However, I can't see the purpose of the move of the steering convergence check. What is the reason why you think the check should be in the gate?
The main cause is https://github.com/autowarefoundation/autoware.universe/issues/4915
Checking steering convergence in the controller creates unnecessary complexity. For this case:

* pid_controller does not allow the vehicle to move because the steering is not converged.
* vehicle_cmd_gate does not allow to control steering because the vehicle stopped.
These two nodes wait for themselves forever.

Because vehicle_cmd_gate checks conditions for start and stopping, I think it makes sense to move steering checking into vehicle_cmd_gate.

TakaHoribe:
@brkay54 Thank you, I understood. It sounds reasonable. Would you reproduce the non-start issue in psim? Or does it happen only in the real vehicle testing?

brkay54:
Yes, I can reproduce in psim.


https://github.com/autowarefoundation/autoware.universe/assets/45468306/d457b2d3-c3cc-40bc-9f47-4781abdca430
Sorry for the screen capture, there is a problem on my computer when recording screen. 

- Vehicle can start at the beginning in auto mode. 
- I set steering about 20 deg with manual controller.
- After I set 20 deg, when I clicked auto, vehicle can not move because of steering converging issue.
- I set steering about 0 def with manual controller.
- Vehicle can move in auto mode.

brkay54:
@TakaHoribe -san, I rebased on the current master and solved the conflicts. And tested again, worked as we wanted.


https://github.com/autowarefoundation/autoware.universe/assets/45468306/19a2484b-9342-430d-a1e1-9957f49bdf7e

Merge order:
1 - https://github.com/autowarefoundation/autoware.universe/pull/5183 and https://github.com/autowarefoundation/autoware_launch/pull/607
2 - https://github.com/autowarefoundation/autoware.universe/pull/5189 and https://github.com/autowarefoundation/autoware_launch/pull/616
3 - https://github.com/autowarefoundation/autoware.universe/pull/5203

Thank you for your effort, if you have any concerns please let me know.

brkay54:
@TakaHoribe -san, sorry to bother you, it is a friendly ping.

rej55 (approved):
LGTM

@TakaHoribe 
Could you approve this PR as a code owner?

control/vehicle_cmd_gate/src/vehicle_cmd_gate.cpp
```diff
@@ -430,13 +433,25 @@ void VehicleCmdGate::publishControlCommands(const Commands & commands)
...
+      filtered_commands.control.longitudinal = createStopTransitionControlCmd();
+    } else {
+      filtered_commands.control = createStopControlCmd();
+    }
+  } else if (enable_keep_steering_until_convergence_) {
+    if (fabs(current_kinematics_.twist.twist.linear.x) < stopped_vel_th) {
+      if (!isSteeringConverged(filtered_commands.control.lateral)) {
+        filtered_commands.control.longitudinal = createStopTransitionControlCmd();
+      }
+    }
```
rej55, line 454: @brkay54 

```
 } else if (enable_keep_steering_until_convergence_) {
    if (fabs(current_kinematics_.twist.twist.linear.x) < stopped_vel_th) {
      if (!isSteeringConverged(filtered_commands.control.lateral)) {
        filtered_commands.control.longitudinal = createStopTransitionControlCmd();
      }
    }
 }
```

Do you have any reasons to implement this block into `vehicle_cmd_gate`?
I think it should be implemented at the end of `trajectory_follower_node` because `vehicle_cmd_gate` should just switch the control commands.

brkay54: Hi thank you for the review. Actually, I explained the reason above [here](https://github.com/autowarefoundation/autoware.universe/pull/5183#issuecomment-1794691616).

ahuazuipiaoliang:
@brkay54 @TakaHoribe @rej55 I checked the code, there are 3 relevant modules, which are trajectory_follower, vehicle_cmd_gate, default_ad_api;
1. If the required steering angle is too large, the steering angle cannot be converged, the the target velocity of longitudinal velocity should be set as 0; 
2. When the target velocity is 0, the vehicle_cmd_gate will output a state of `is_start_requested=false`;
3. If `is_start_requested==false`, the `default_ad_api` will change the state to `PAUSED` and keep calling the service `/control/vehicle_cmd_gate/set_pause` to set `PAUSED` state of `vehicle_cmd_gate`;
4. When the state of `vehicle_cmd_gate` is `PAUSED`, the `vehicle_cmd_gate` will keep sending `stop control cmd` which the steering angle is the `current_steering_angle` of the vehicle.

So the vehicle enters a dead loop and keeps `PAUSED` state.
I suggest create paused_control_cmd for `PAUSED` state in `vehicle_cmd_gate` that the `steering_cmd` is the command of `trajectory_follower`. When the steering angle is converged, the longitudinal controller can output normal control command.

brkay54:
Firstly, we need to avoid sending `createStopControlCmd()`, while the vehicle is stopping (not stopped yet in disengage state). Because it causes sending current_steering while the vehicle is decelerating. You can see the problem [this issue](https://github.com/autowarefoundation/autoware.universe/issues/5796).

Thank you for the suggestion @ahuazuipiaoliang! It makes sense to me, however, it will cause an undesired behavior. If we change both `DISENGAGE` and `PAUSED` states to send control_cmd's steering command, the steering will be controllable all the time. So, the vehicle won't be fully stopped, always it listens to the control_cmd's steering command. cc @rej55 @TakaHoribe

control/vehicle_cmd_gate/src/vehicle_cmd_gate.cpp
brkay54: Sorry to bother you, but could you check the comment [below](https://github.com/autowarefoundation/autoware.universe/pull/5183#issuecomment-1913727693)? @TakaHoribe @rej55

rej55:
@brkay54 cc @TakaHoribe 
Sorry for late reply, I agree for the suggestion from @ahuazuipiaoliang.
IMO, the trajectory_follower should send the control_cmd not to move the steering.
(@TakaHoribe said the same thing when I talked with him)

ahuazuipiaoliang:
@rej55 Thanks, Fujiyama. Actually, I have modified the code in my way and tested on real vehicle for several days. I have not found any issue yet.

brkay54:
@rej55 Thanks for the reply, it makes sense to me. I am updating the PR.

brkay54:
@rej55 @ahuazuipiaoliang Sorry to disturb you guys, thank you for your replies!

To be able to not listen to the trajectory_follower's output when the vehicle fully stopped, I found another way. I changed the structure like this:

```
  // Check engage
  if (!is_engaged_) {
    filtered_commands.control.longitudinal = createLongitudinalStopControlCmd();
  }

  // Check pause. Place this check after all other checks as it needs the final output.
  adapi_pause_->update(filtered_commands.control);
  if (adapi_pause_->is_paused()) {
    if (is_engaged_) {
      filtered_commands.control.longitudinal = createLongitudinalStopControlCmd();
    } else {
      filtered_commands.control = createStopControlCmd();
    }
  }
  ```
  
 In this way, the vehicle will be able to control steering if the vehicle is in PAUSED state. Also, it won't be able to control the steering when it is fully stopped and disengaged. I made some tests to check:
 

https://github.com/autowarefoundation/autoware.universe/assets/45468306/c0e0fb57-1a79-46a7-bdaa-7925b5ff2046

In the video above, I tested whether the vehicle can control the steering when we set a large steering value by using the manual controller or not. When I changed state to `AUTO`, the vehicle did not move until the steering converged as you can see.

In the second part, I set a large steering value again and I changed its state to `STOP`, as you can see, the vehicle did not change its steering. (It changed a little bit because the manual controller sets ENGAGE state to true while running and the steering was changing in the transition between ENGAGE to DISENGAGE)

Also, I checked the behavior of the vehicle while stopping. (It can control the steering or not while moving and disengaged) You can see the result in the video below:
 
https://github.com/autowarefoundation/autoware.universe/assets/45468306/0beac618-ceca-4ac2-8c2a-edec41320e86

As you can see, It still can control steering after I click the STOP button.

Also, I tried this structure too:

```
  // Check engage
  if (!is_engaged_) {
    filtered_commands.control.longitudinal = createLongitudinalStopControlCmd();
  }

  // Check pause. Place this check after all other checks as it needs the final output.
  adapi_pause_->update(filtered_commands.control);
  if (adapi_pause_->is_paused()) {
    filtered_commands.control.longitudinal = createLongitudinalStopControlCmd();
  }
  ```
  
It also solves the problems above, but, when the vehicle fully stopped, trajectory_follower can control the steering as you can see below:

https://github.com/autowarefoundation/autoware.universe/assets/45468306/c22c959c-fb50-4e8c-bc31-5894d6a5c1f6

Therefore I used the first structure. Please share your comments with me! Thank you for your efforts.

rej55 (approved):
LGTM.
Sorry for late review.

brkay54:
@tkimura4 @TakaHoribe We need code-owner approve. Could you check the PR?

TakaHoribe (approved):
LGTM, thank you.
Note: the keep-stop command can send the target steering in order to control steering before moving.

PR #4553, opened 2023-08-08
fix(system_monitor): extend command line to display

ito-san:
## Description

- The process_monitor provides information about processes that are utilizing high CPU and memory resources.
However, the displayed COMMAND field only shows the executable name such as `ros2` or `python3`, and identifying the process solely based on the provided COMMAND name is not easy.
image

- With this change, it's possible to provide a more detailed COMMAND string.

## Related links

[TIER IV INTERNAL LINK TO JIRA](https://tier4.atlassian.net/browse/RT0-28722)

## Tests performed

1. Run Autoware.
2. Run rqt_runtime_monitor.
    ```console 
    ros2 run rqt_runtime_monitor rqt_runtime_monitor
    ```
3. Verify a more detailed COMMAND string is provided.
image

## Notes for reviewers

None

## Interface changes

None

## Effects on system behavior

None

## Pre-review checklist for the PR author

The PR author **must** check the checkboxes below when creating the PR.

- [x] I've confirmed the [contribution guidelines].
- [x] The PR follows the [pull request guidelines].

## In-review checklist for the PR reviewers

The PR reviewers **must** check the checkboxes below before approval.

- [ ] The PR follows the [pull request guidelines].
- [ ] The PR has been properly tested.
- [ ] The PR has been reviewed by the code owners.

## Post-review checklist for the PR author

The PR author **must** check the checkboxes below before merging.

- [ ] There are no open discussions or they are tracked via tickets.
- [ ] The PR is ready for merge.

After all checkboxes are checked, anyone who has write access can merge the PR.

[contribution guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/
[pull request guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/pull-request-guidelines/

PR #5191, opened 2023-10-02
perf(system_monitor): fix program command line reading

Owen-Liuyuxuan:
## Description

Since the enhancement from https://github.com/autowarefoundation/autoware.universe/pull/4553/, it has been reported that reading the full return of the ```top``` command with hundreds of processes could lead to failure in boost::process::child.

In this PR, we step back the ```top```  reading in boost::process::child to the original version in https://github.com/autowarefoundation/autoware.universe/pull/4553/ which means reading only the program names by default. Then for Top-rated processes, we try to read its command-line name in ```/proc/{PID}/cmdline```. If the reading of a process fails (probably the process ends after the top command), we will only show its program name instead.

## Related links

None.

## Tests performed

PSim

image

### DIFF
--- a/system/system_monitor/include/system_monitor/process_monitor/process_monitor.hpp
+++ b/system/system_monitor/include/system_monitor/process_monitor/process_monitor.hpp
@@ -92,8 +92,8 @@ protected:
    * @brief get command line from process id
    * @param [in] pid process id
    * @param [out] command output command line
-   * @return true if success to get command line name 
-  */
+   * @return true if success to get command line name
+   */
   bool getCommandLineFromPiD(const std::string & pid, std::string * command);
 
   /**
--- a/system/system_monitor/include/system_monitor/process_monitor/process_monitor.hpp
+++ b/system/system_monitor/include/system_monitor/process_monitor/process_monitor.hpp
@@ -89,6 +89,14 @@ protected:
   void getHighMemoryProcesses(const std::string & output);
 
   /**
+   * @brief get command line from process id
+   * @param [in] pid process id
+   * @param [out] command output command line
+   * @return true if success to get command line name 
+  */
+  bool getCommandLineFromPiD(const std::string & pid, std::string * command);
+
+  /**
    * @brief get top-rated processes
    * @param [in] tasks list of diagnostics tasks for high load procs
    * @param [in] output top command output
--- a/system/system_monitor/src/process_monitor/process_monitor.cpp
+++ b/system/system_monitor/src/process_monitor/process_monitor.cpp
@@ -419,11 +419,11 @@ bool ProcessMonitor::getCommandLineFromPiD(const std::string & pid, std::string 
   std::string commandLineFilePath = "/proc/" + pid + "/cmdline";
   std::ifstream commandFile(commandLineFilePath);
   if (commandFile.is_open()) {
-      std::getline(commandFile, *command);
-      commandFile.close();
-      return true;
+    std::getline(commandFile, *command);
+    commandFile.close();
+    return true;
   } else {
-      return false;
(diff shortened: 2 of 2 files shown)


asana17 (approved):
LGTM. Thanks for the implementation!

One small thing...
To fulfill the spell check requirements, how about renaming function names and variables to use "command name" instead of "command line", as it is not getting the command line?

Owen-Liuyuxuan:
> LGTM. Thanks for the implementation!
> 
> One small thing... To fulfill the spell check requirements, how about renaming function names and variables to use "command name" instead of "command line", as it is not getting the command line?

I change "commandline" to command_line in variables. I will keep an eye on the CI results later.

asana17:
I think we can avoid file access by buffering the process name previously accessed, but this approach uses additional spaces instead. I will leave it to you if you implement it or not.

Owen-Liuyuxuan:
> I think we can avoid file access by buffering the process name previously accessed, but this approach uses additional spaces instead. I will leave it to you if you implement it or not.

I am not sure.

1. It is said that the ```top``` command essentially reads everything in the ```/proc``` folder (which is a pseudo-file system without real disk i/o), so we are doing what ```top``` is doing, except we are in the main thread and only work on top-rated processes.
   [reference](flag_find_command_line)
2. As for buffering, if it is needed, I was thinking about maintaining an unordered map from PIDs to the command lines. But does it assume that the same PID corresponds to the same command line name? For now, I am not sure how to do this cleanly.

asana17:
> > I think we can avoid file access by buffering the process name previously accessed, but this approach uses additional spaces instead. I will leave it to you if you implement it or not.
> 
> I am not sure.
> 
> 1. It is said that the `top` command essentially reads everything in the `/proc` folder (which is a pseudo-file system without real disk i/o), so we are doing what `top` is doing, except we are in the main thread and only work on top-rated processes.
>    [reference](flag_find_command_line)
> 2. As for buffering, if it is needed, I was thinking about maintaining an unordered map from PIDs to the command lines. But does it assume that the same PID corresponds to the same command line name? For now, I am not sure how to do this cleanly.

Yes, the PID could be duplicated between the dead process and the new process. However, the autoware (ros2) related process will not reboot during the system monitor running and holds the process name I think.

Owen-Liuyuxuan:
> I think we can avoid file access by buffering the process name previously accessed, but this approach uses additional spaces instead. I will leave it to you if you implement it or not.

I believe it is better not to implement this for now. 
1. According to the proc reference and the implementation of ```top```, reading the command line name is rather efficient, especially only on the limited number of processes. If we find that the consumption is still too large for onboard systems, then we could re-evaluate this issue again.
2. PID duplication and the potential expanding memory usage could be two main problems. Extra codes may also be needed to manage these two problems. Therefore, I guess it is better to start that implementation in a new PR if needed :)

asana17:
> > I think we can avoid file access by buffering the process name previously accessed, but this approach uses additional spaces instead. I will leave it to you if you implement it or not.
> 
> I believe it is better not to implement this for now.
> 
> 1. According to the proc reference and the implementation of `top`, reading the command line name is rather efficient, especially only on the limited number of processes. If we find that the consumption is still too large for onboard systems, then we could re-evaluate this issue again.
> 2. PID duplication and the potential expanding memory usage could be two main problems. Extra codes may also be needed to manage these two problems. Therefore, I guess it is better to start that implementation in a new PR if needed :)

I agree with you for re-evaluating it when we found the serious problem related to this implementation, and I think it won't happen. Sorry for confusing you, and thank you for thinking about it!

ito-san (approved):
Thank you very much for the PR! Great help!

PR #5255, opened 2023-10-09
feat(pid_longitudinal_controller): transit to STOPPED even if the sign of the ego velocity goes opposite

takayuki5168:
## Description

Currently, in the pid_longitudinal_controller, the condition to transit to STOPPED is that the absolute values of the velocity and acceleration are small enough.
However, on the steep ascending slope, if the ego tries to stop by the state transiting to STOPPED but the ego moves backward due to the slope, the state will never transit to STOPPED since the absolute value of the velocity is not small enough.

To deal with this case, the condition to transit to STOPPED is changed as follows
- The absolute values of the velocity and acceleration are small enough, which is the original condition.
- **Or the sign of the velocity is opposite to the expected.**
    - e.g. When the shift is forward but the velocity is negative, the condition to transit to STOPPED is met.

## Tests performed

unit test
psim

## Effects on system behavior

can stop on the steep slope.

### DIFF
--- a/control/pid_longitudinal_controller/src/pid_longitudinal_controller.cpp
+++ b/control/pid_longitudinal_controller/src/pid_longitudinal_controller.cpp
@@ -504,6 +504,8 @@ void PidLongitudinalController::updateControlState(const ControlData & control_d
 
   const bool is_stopped = std::abs(current_vel) < p.stopped_state_entry_vel &&
                           std::abs(current_acc) < p.stopped_state_entry_acc;
+  // Case where the ego slips in the opposite direction of the gear due to e.g. a slope is also
+  // considered as a stop
   const bool is_not_running = [&]() {
     if (control_data.shift == Shift::Forward) {
       if (is_stopped || current_vel < 0.0) {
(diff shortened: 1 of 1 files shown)


TakaHoribe (approved):
The condition of "moving" is defined separately in FWD gear and REV gear.
image

control/pid_longitudinal_controller/src/pid_longitudinal_controller.cpp
```diff
@@ -501,9 +501,23 @@ void PidLongitudinalController::updateControlState(const ControlData & control_d
     m_enable_keep_stopped_until_steer_convergence && !lateral_sync_data_.is_steer_converged;
 
   const bool stopping_condition = stop_dist < p.stopping_state_stop_dist;
-  if (
-    std::fabs(current_vel) > p.stopped_state_entry_vel ||
-    std::fabs(current_acc) > p.stopped_state_entry_acc) {
+
+  const bool is_stopped = std::abs(current_vel) < p.stopped_state_entry_vel &&
+                          std::abs(current_acc) < p.stopped_state_entry_acc;
+  const bool is_not_running = [&]() {
```
TakaHoribe, line 507:
```suggestion
  // Case where the ego slips in the opposite direction of the gear due to e.g. a slope is also
  // considered as a stop
  const bool is_not_running = [&]() {
```

PR #5259, opened 2023-10-10
feat(imu_corrector): changed GyroBiasEstimator to use ndt pose instead of twist

SakodaShintaro:
## Description

Before : Use `twist` to determine whether ego-vehicle is stopping, and calculate imu bias by comparing it to 0 while it is stopping.

After: Use `ndt pose` to determine whether ego-vehicle is curving, and calculate imu bias by comparing it to ndt pose while it is not curving.

## Related link
* https://github.com/autowarefoundation/sample_sensor_kit_launch/pull/78
* https://github.com/tier4/aip_launcher/pull/180
* https://github.com/RobotecAI/awsim_sensor_kit_launch/pull/10

## Tests performed

It has been confirmed that it works as expected with sample rosbag.

The red dashed line indicates the range extended from the orange line, which is the calibration value, by gyro_bias_threshold.

If it is within the red dashed line range, the IMU bias is considered normal.

For TIER IV's internal rosbags, the result of the IMU bias variation detection function for data taken on a day not long after calibration is as follows.

For data taken on a day when more time has passed since calibration, the result is as follows.

## Effects on system behavior

Improves performance of IMU bias variation detection.

### DIFF
--- a/sensing/imu_corrector/README.md
+++ b/sensing/imu_corrector/README.md
@@ -58,6 +58,10 @@ Note that the node calculates bias from the gyroscope data by averaging the data
 
 Note that the input pose is assumed to be accurate enough. For example when using NDT, we assume that the NDT is appropriately converged.
 
+Currently, it is possible to use methods other than NDT as a `pose_source` for Autoware, but less accurate methods are not suitable for IMU bias estimation.
+
+In the future, with careful implementation for pose errors, the IMU bias estimated by NDT could potentially be used not only for validation but also for online calibration.
+
 ### Output
 
 | Name                 | Type                                 | Description                   |
--- a/sensing/imu_corrector/README.md
+++ b/sensing/imu_corrector/README.md
@@ -72,4 +72,3 @@ Note that the node calculates bias from the gyroscope data by averaging the data
 | `gyro_bias_threshold`                 | double | threshold of the bias of the gyroscope [rad/s]                                              |
 | `timer_callback_interval_sec`         | double | seconds about the timer callback function [sec]                                             |
 | `straight_motion_ang_vel_upper_limit` | double | upper limit of yaw angular velocity, beyond which motion is not considered straight [rad/s] |
-| `data_num_threshold`                  | int    | number of data used to calculate bias                                                       |
--- a/sensing/imu_corrector/README.md
+++ b/sensing/imu_corrector/README.md
@@ -51,10 +51,10 @@ Note that the node calculates bias from the gyroscope data by averaging the data
 
 ### Input
 
-| Name              | Type                                             | Description      |
-| ----------------- | ------------------------------------------------ | ---------------- |
-| `~/input/imu_raw` | `sensor_msgs::msg::Imu`                          | **raw** imu data |
-| `~/input/pose`    | `geometry_msgs::msg::PoseWithCovarianceStamped`  | ndt pose         |
+| Name              | Type                                            | Description      |
+| ----------------- | ----------------------------------------------- | ---------------- |
+| `~/input/imu_raw` | `sensor_msgs::msg::Imu`                         | **raw** imu data |
+| `~/input/pose`    | `geometry_msgs::msg::PoseWithCovarianceStamped` | ndt pose         |
 
 ### Output
 
--- a/sensing/imu_corrector/README.md
+++ b/sensing/imu_corrector/README.md
@@ -51,10 +51,10 @@ Note that the node calculates bias from the gyroscope data by averaging the data
 
 ### Input
 
-| Name              | Type                                            | Description      |
-| ----------------- | ----------------------------------------------- | ---------------- |
-| `~/input/imu_raw` | `sensor_msgs::msg::Imu`                         | **raw** imu data |
-| `~/input/pose`    | `geometry_msgs::msg::PoseWithCovarianceStamped` | ndt pose         |
+| Name              | Type                                            | Description                                                                                                                               |
+| ----------------- | ----------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------- |
+| `~/input/imu_raw` | `sensor_msgs::msg::Imu`                         | **raw** imu data                                                                                                                          |
+| `~/input/pose`    | `geometry_msgs::msg::PoseWithCovarianceStamped` | Note that the input pose is assumed to be accurate enough. For example when using NDT, we assume that the NDT is appropriately converged. |
 
 ### Output
 
--- a/sensing/imu_corrector/README.md
+++ b/sensing/imu_corrector/README.md
@@ -54,7 +54,7 @@ Note that the node calculates bias from the gyroscope data by averaging the data
 | Name              | Type                                             | Description      |
 | ----------------- | ------------------------------------------------ | ---------------- |
 | `~/input/imu_raw` | `sensor_msgs::msg::Imu`                          | **raw** imu data |
-| `~/input/twist`   | `geometry_msgs::msg::TwistWithCovarianceStamped` | vehicle velocity |
+| `~/input/pose`    | `geometry_msgs::msg::PoseWithCovarianceStamped`  | ndt pose         |
 
 ### Output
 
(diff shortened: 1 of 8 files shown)


kminoda:
Thank you very much! Let me review until tomorrow :pray:

sensing/imu_corrector/src/gyro_bias_estimator.cpp
```diff
@@ -22,59 +29,138 @@ GyroBiasEstimator::GyroBiasEstimator()
...
     "~/input/imu_raw", rclcpp::SensorDataQoS(),
     [this](const Imu::ConstSharedPtr msg) { callback_imu(msg); });
-  twist_sub_ = create_subscription<TwistWithCovarianceStamped>(
-    "~/input/twist", rclcpp::SensorDataQoS(),
-    [this](const TwistWithCovarianceStamped::ConstSharedPtr msg) { callback_twist(msg); });
-
+  pose_sub_ = create_subscription<PoseWithCovarianceStamped>(
+    "~/input/pose", rclcpp::SensorDataQoS(),
+    [this](const PoseWithCovarianceStamped::ConstSharedPtr msg) { callback_pose(msg); });
   gyro_bias_pub_ = create_publisher<Vector3Stamped>("~/output/gyro_bias", rclcpp::SensorDataQoS());
```
kminoda, line 51: Please publish the estimated gyro_bias as a topic as well

kminoda (commented):
Thank you! I left some comments :pray:

sensing/imu_corrector/config/gyro_bias_estimator.param.yaml
```diff
@@ -1,6 +1,6 @@
 /**:
   ros__parameters:
-    gyro_bias_threshold: 0.0015 # [rad/s]
-    velocity_threshold: 0.0 # [m/s]
-    timestamp_threshold: 0.1 # [s]
+    gyro_bias_threshold: 0.0050 # [rad/s]
```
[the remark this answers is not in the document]
SakodaShintaro: This value was chosen so that the OK/NG decision is appropriately split for the two internal data.
However, the results of the survey revealed that the original value should remain the same, so it was reverted.
https://github.com/autowarefoundation/autoware.universe/pull/5259/commits/b012405bbe8d1104798e71641a395af8384a26cc

sensing/imu_corrector/src/gyro_bias_estimator.cpp
SakodaShintaro: Thank you.
It was a mistake to delete. I have restored it.
https://github.com/autowarefoundation/autoware.universe/pull/5259/commits/b762f909210755ae98e9f16a3380f0bff5d76188

kminoda:
Would you create a PR for other sensor kit packages as well? (e.g. aip_launcher)

sensing/imu_corrector/src/gyro_bias_estimation_module.hpp
```diff
@@ -11,33 +11,33 @@
...
-    const size_t data_num_threshold);
-  geometry_msgs::msg::Vector3 get_bias() const;
-  void update_gyro(const double time, const geometry_msgs::msg::Vector3 & gyro);
-  void update_velocity(const double time, const double velocity);
+  explicit GyroBiasEstimationModule(const size_t data_num_threshold);
+  void update_bias(
+    const std::vector<geometry_msgs::msg::PoseStamped> & pose_list,
+    const std::vector<geometry_msgs::msg::Vector3Stamped> & gyro_list);
+  geometry_msgs::msg::Vector3 get_bias_base_link() const;
+  geometry_msgs::msg::Vector3 get_bias_std() const;
```
kminoda, line 35: Would you erase this function as it is currently not used in gyro_bias_estimator?

sensing/imu_corrector/config/gyro_bias_estimator.param.yaml
```diff
@@ -1,6 +1,6 @@
 /**:
   ros__parameters:
     gyro_bias_threshold: 0.0015 # [rad/s]
-    velocity_threshold: 0.0 # [m/s]
-    timestamp_threshold: 0.1 # [s]
+    timer_callback_interval_sec: 0.5 # [sec]
+    straight_motion_ang_vel_upper_limit: 0.015 # [rad/s]
     data_num_threshold: 200 # [num]
```
kminoda, line 6: Did you test with this value? If using 0.5[sec] per data as a frequency, collecting 200 values sounds too large. Would like to know if there is a room for making this value smaller. 

(Of course it's a trade-off between robustness)

SakodaShintaro:
> Would you create a PR for other sensor kit packages as well? (e.g. aip_launcher)

I have also created this pull request.
https://github.com/tier4/aip_launcher/pull/180

sensing/imu_corrector/config/gyro_bias_estimator.param.yaml
SakodaShintaro: Sorry the code is not clear.

The `data_num_threshold` (200) is used as just an upper limit. If even one data item is obtained, a bias is calculated and diagnostics are also output.

I was concerned that if this limit is not defined, the `gyro_bias_deque_` will get more and more.

By removing the `get_bias_std` function, this variable can also be removed, which would reduce confusion.

sensing/imu_corrector/src/gyro_bias_estimation_module.hpp
SakodaShintaro: Erased in
https://github.com/autowarefoundation/autoware.universe/pull/5259/commits/482b3ccb6be56ecc064881fd625f48348c4d718b

sensing/imu_corrector/config/gyro_bias_estimator.param.yaml
kminoda: I see, make sense (I also failed to read the code correctly).

But removing the threshold may not be a good idea, since the bias with only one data item may be unreliable.
IMO it's better to publish the diagnostics when enough data is gathered.

SakodaShintaro: Even the first time, there is some reliability as it corresponds to an average of 0.5 seconds. (The leftmost point on the graph in the pull request description field corresponds to the first point)

Also, it seems more legitimate to use stddev together with reliability rather than specifying a fixed number of times. (Although I deleted stddev...)

kminoda: > Also, it seems more legitimate to use stddev together with reliability rather than specifying a fixed number of times.

True :sweat_smile: 

Understood that the estimate from 0.5 seconds are accurate for most of the cases. Still I am a bit worried about making a false positives when picking unlucky unstable data for the initial data, but should be fine as the first data would likely to be a data when the vehicle is stopped. Thanks for the clarification!

sensing/imu_corrector/README.md
```diff
@@ -51,10 +51,10 @@ Note that the node calculates bias from the gyroscope data by averaging the data
...
 ### Input
 
-| Name              | Type                                             | Description      |
-| ----------------- | ------------------------------------------------ | ---------------- |
-| `~/input/imu_raw` | `sensor_msgs::msg::Imu`                          | **raw** imu data |
-| `~/input/twist`   | `geometry_msgs::msg::TwistWithCovarianceStamped` | vehicle velocity |
+| Name              | Type                                            | Description      |
+| ----------------- | ----------------------------------------------- | ---------------- |
+| `~/input/imu_raw` | `sensor_msgs::msg::Imu`                         | **raw** imu data |
+| `~/input/pose`    | `geometry_msgs::msg::PoseWithCovarianceStamped` | ndt pose         |
```
kminoda, line 57: Would you state here something like as follows as a side note below?
```
Note that the input pose is assumed to be accurate enough. For example when using NDT, we assume that the NDT is appropriately converged.
```

kminoda (commented):
One final comment :pray:

sensing/imu_corrector/README.md
SakodaShintaro: That's a very good. In fact, one needs to be careful when using a pose other than NDT, or when using NDT in an environment where its accuracy will be low.
Currently, it is only output to diagnostics, so there may not be a need to be overly cautious. However, caution is needed when changing the IMU value based on the estimate obtained here for online calibration.

Fixed in https://github.com/autowarefoundation/autoware.universe/pull/5259/commits/f669c20d54eeb471f69edc1e54fabe14ba8a9e16

SakodaShintaro:
I created a pull request to launch `gyro_bias_estimator` in pose mode in AWSIM as well, but I can not seem to add reviewers.

https://github.com/RobotecAI/awsim_sensor_kit_launch/pull/10

sensing/imu_corrector/README.md
kminoda: Thanks! But what I meant was to put it as a side note maybe as a footer. The description `ndt pose` itself was OK.

So something like
```
### Input

| Name              | Type                                            | Description      |
| ----------------- | ----------------------------------------------- | ---------------- |
| `~/input/imu_raw` | `sensor_msgs::msg::Imu`                         | **raw** imu data |
| `~/input/pose`    | `geometry_msgs::msg::PoseWithCovarianceStamped` | ndt pose         |

Note that the input pose is assumed to be accurate enough. For example when using NDT, we assume that the NDT is appropriately converged.

```

kminoda: > However, caution is needed when changing the IMU value based on the estimate obtained here for online calibration.

Correct. That's one of the main concern for using NDT for estimation, which makes some kind of a new closed loop in Autoware. So would like to add some warnings for future developers.

kminoda (approved):
LGTM!

PR #5310, opened 2023-10-14
perf(radius_search_2d_outlier_filter_nodelet): replace PCL KdTree for naive algorithm

veqcc:
Signed-off-by: veqcc [ryuta.kambe@tier4.jp](mailto:ryuta.kambe@tier4.jp)

### Description
First, I want to ask whether this `radius_search_2d_outlier_filter_nodelet` is still used or will be used in the future?
Since now I'm removing PCL from Autoware as much as possible, and I could not find any program which calls the nodelet.
If it is true, then I will discard this PR and create a new one in which `radius_search_2d_outlier_filter_nodelet` is totally deleted.

Otherwise, I will confirm whether this PR really reduces the latency, memory usage and soft page faults.


yukkysaito:
We are still using it for several projects.
If KD-tree is not used, there may be a difference in execution time depending on the number of point clouds in the input. If possible, could you please benchmark this for us?

veqcc:
Thank you for your reply.
I will do it.

yukkysaito:
@veqcc I can't write about the project so I will share it on slack.

miursh:
We are closing this PR as there are no response for months. Please re-open if the issue still persists.

PR #5327, opened 2023-10-17
fix(detected_object_validation): add 3d pointcloud based validator

badai-nguyen:
## Description

- Currently obstacle_pointcloud_based_validator projects pointcloud into xy_plane and check the number of points within object footprint. 
- In some cases like the centerpoint misdetects tree bushes as car, then the high position points of the tree migh be included which cause a wrong validation.
- This PR will add an option to validate obstacle pointcloud base on 3d shape of detected objects.

| frequently misdetection case | tree leaves are also projected and treated as validation points|
| -- | --|
|    | yellow: obstacle points, green points: neighbor, pink: within polygon; green bbox: validated object|
|image |image|

## Related link

[TIER IV INTERNAL LINK](https://tier4.atlassian.net/browse/RT0-29267?focusedCommentId=139292)

## Tests performed

Tested by logging_simulator
- green points: neighbor points 
- pink points: points within object shape
- orange bbox: non-validated centerpoint result
Screenshot from 2023-10-17 10-39-58

CPU usage comparison: there was not significant change in cpu usage of obstacle-pointcloud-based-validator-node 
| 2d validation | 3d validation | 
| --- | ---|
| obstacle-pointcloud-based-validator-node-2d_CPU-Percent| obstacle-pointcloud-based-validator-node-3d_CPU-Percent|

Not applicable.

## Effects on system behavior

Not applicable.

### Summary by CodeRabbit

- New Feature: Added support for 3D shape-based validation of obstacle point clouds in the `ObstaclePointCloudBasedValidator` class.
- New Feature: Introduced a new method `addNeighborPointcloud` to the `Debugger` class, allowing users to add neighbor point clouds.
- New Feature: Added a parameter to the launch file (`obstacle_pointcloud_based_validator.launch.xml`) to choose between 2D and 3D validation methods.
- Documentation: Updated documentation for the `ObstaclePointCloudBasedValidator` class to reflect the changes and provide usage instructions.

### DIFF
--- a/perception/detected_object_validation/include/obstacle_pointcloud_based_validator/debugger.hpp
+++ b/perception/detected_object_validation/include/obstacle_pointcloud_based_validator/debugger.hpp
@@ -71,15 +71,16 @@ public:
   void addNeighborPointcloud(const pcl::PointCloud<pcl::PointXY>::Ptr & input)
   {
     pcl::PointCloud<pcl::PointXYZ>::Ptr input_xyz = toXYZ(input);
-    for (const auto & point : *input_xyz) {
-      neighbor_pointcloud_->push_back(point);
-    }
+    addNeighborPointcloud(input_xyz);
   }
 
   void addNeighborPointcloud(const pcl::PointCloud<pcl::PointXYZ>::Ptr & input)
   {
-    for (const auto & point : *input) {
-      neighbor_pointcloud_->push_back(point);
+    if (!input->empty()) {
+      neighbor_pointcloud_->reserve(neighbor_pointcloud_->size() + input->size());
+      for (const auto & point : *input) {
+        neighbor_pointcloud_->push_back(point);
+      }
     }
   }
   void addPointcloudWithinPolygon(const pcl::PointCloud<pcl::PointXYZ>::Ptr & input)
--- a/perception/detected_object_validation/include/obstacle_pointcloud_based_validator/debugger.hpp
+++ b/perception/detected_object_validation/include/obstacle_pointcloud_based_validator/debugger.hpp
@@ -76,6 +76,12 @@ public:
     }
   }
 
+  void addNeighborPointcloud(const pcl::PointCloud<pcl::PointXYZ>::Ptr & input)
+  {
+    for (const auto & point : *input) {
+      neighbor_pointcloud_->push_back(point);
+    }
+  }
   void addPointcloudWithinPolygon(const pcl::PointCloud<pcl::PointXYZ>::Ptr & input)
   {
     // pcl::PointCloud<pcl::PointXYZ>::Ptr input_xyz = toXYZ(input);
--- a/perception/detected_object_validation/src/obstacle_pointcloud_based_validator.cpp
+++ b/perception/detected_object_validation/src/obstacle_pointcloud_based_validator.cpp
@@ -144,7 +144,7 @@ void ObstaclePointCloudBasedValidator::on3dObjectsAndObstaclePointCloud(
     const auto & object = input_objects->objects.at(i);
     const auto & transformed_object_position =
       transformed_object.kinematics.pose_with_covariance.pose.position;
-    const auto search_radius = getMaxRadius(transformed_object);
+    const auto search_radius = getMaxRadius3D(transformed_object);
     if (!search_radius) {
       output.objects.push_back(object);
       continue;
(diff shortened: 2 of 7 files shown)


yukkysaito:
I'll review this :+1:

perception/detected_object_validation/src/obstacle_pointcloud_based_validator.cpp
```diff
@@ -85,24 +85,109 @@ ObstaclePointCloudBasedValidator::ObstaclePointCloudBasedValidator(
...
+      output.objects.push_back(object);
+    }
+  }
+  objects_pub_->publish(output);
+  if (debugger_) {
+    debugger_->publishRemovedObjects(removed_objects);
+    debugger_->publishNeighborPointcloud(input_obstacle_pointcloud->header);
+    debugger_->publishPointcloudWithinPolygon(input_obstacle_pointcloud->header);
+  }
+}
```
yukkysaito, line 190: Since most of them are the same as callback functions of XY-plane, callback functions may be common.

perception/detected_object_validation/include/obstacle_pointcloud_based_validator/debugger.hpp
```diff
@@ -76,6 +76,12 @@ class Debugger
     }
   }
 
+  void addNeighborPointcloud(const pcl::PointCloud<pcl::PointXYZ>::Ptr & input)
+  {
+    for (const auto & point : *input) {
+      neighbor_pointcloud_->push_back(point);
+    }
+  }
```
yukkysaito, line 84:
```suggestion
  void addNeighborPointcloud(const pcl::PointCloud<pcl::PointXYZ>::Ptr & input)
  {
    for (const auto & point : *input) {
      neighbor_pointcloud_->push_back(point);
    }
  }

```

perception/detected_object_validation/src/obstacle_pointcloud_based_validator.cpp
```diff
@@ -85,24 +85,109 @@ ObstaclePointCloudBasedValidator::ObstaclePointCloudBasedValidator(
...
+
+    if (debugger_) debugger_->addNeighborPointcloud(neighbor_pointcloud);
+    // Filter object that have few pointcloud in them
+    const auto num = getPointCloudNumWithinShape(transformed_object, neighbor_pointcloud);
+    const auto object_distance =
+      std::hypot(transformed_object_position.x, transformed_object_position.y);
+    size_t min_pointcloud_num = std::clamp(
+      static_cast<size_t>(
+        points_num_threshold_param_.min_points_and_distance_ratio.at(object_label_id) /
+          object_distance +
```
yukkysaito, line 173: The formula here is correct in 2D, but incorrect in 3D.
I don't think it will be a big problem with the 2D formula since it is after height crop, but you should leave it in the comments just in case.

perception/detected_object_validation/include/obstacle_pointcloud_based_validator/debugger.hpp
yukkysaito: nitpick :pray:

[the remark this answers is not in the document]
yukkysaito: @badai-nguyen I think the point is a decent one and we might as well reflect it.

perception/detected_object_validation/src/obstacle_pointcloud_based_validator.cpp
badai-nguyen: @yukkysaito thank you for your comment. I changed  at https://github.com/autowarefoundation/autoware.universe/pull/5327/commits/e3cbf6cb524a34ad3b693a8c0eaa1680e3141f72

perception/detected_object_validation/include/obstacle_pointcloud_based_validator/debugger.hpp
badai-nguyen: I fixed at https://github.com/autowarefoundation/autoware.universe/pull/5327/commits/c3317b63547472675d09e7c0d27597d8b7665749

perception/detected_object_validation/src/obstacle_pointcloud_based_validator.cpp
badai-nguyen: @yukkysaito 
IMO:
- Since this is to adapt validation threshold as pointcloud density decreasing as far distance so the formular should be common for 2D and 3D. The min_points_num and max_points_num might be better changing for 2D and 3D. 
- The big difference happens when vehicle is in front of a up/down slope and the LiDAR resolution are largely changed in vertical axis. 
- As you mentioned, since we have height crop so I will leave it as is.

yukkysaito: @badai-nguyen I agree.
Could you please leave a light comment in the code? :pray:

perception/detected_object_validation/src/obstacle_pointcloud_based_validator.cpp
yukkysaito: I'm sorry, I didn't communicate it well.
It was a little different from my image.

Most parts of the callback function are the same for 2D and 3D, so I would be happy if you could abstract and absorb only the parts that make a difference in the interface (using template, function pointer, etc.). If it is difficult, it might be better to have separate callbacks.

badai-nguyen: @yukkysaito I am sorry for late response. Currently, I did not find the efficient way to refactor as your comment yet, so I would like revert to the separate callbacks way first. I will consider to refactor this ASAP in another PR.

yukkysaito (approved):
LGTM

PR #5347, opened 2023-10-18
test(ndt_scan_matcher): add test_ndt_scan_matcher_launch.py

SakodaShintaro:
## Description

Add a test `test_ndt_scan_matcher_launch.py` that only checks the launch of the `ndt_scan_matcher` node.

With this change, the checking of services in the constructor of MapUpdateModule will be changed to checking before the actual service call.

It is somewhat useless to check every time, but I did this because the constructor cannot terminate in test.

## Tests performed

It has been confirmed that the following commands can be executed successfully.

```bash
colcon build --symlink-install --cmake-args -DCMAKE_BUILD_TYPE=Release -DBUILD_TESTING=ON --packages-select ndt_scan_matcher
colcon test --packages-select ndt_scan_matcher
colcon test-result --all --verbose | grep ndt_scan_matcher
```

I also confirmed that double declarations of parameters will cause the test to fail NG.

In addition, it has been confirmed that AWSIM and logging_simulator can run successfully.

## Effects on system behavior

There is no effects on system behavior since it is just an additional test.


localization/ndt_scan_matcher/test/test_ndt_scan_matcher_launch.py
```diff
@@ -0,0 +1,99 @@
...
+        request.data = True
+        future = cli_trigger.call_async(request)
+        rclpy.spin_until_future_complete(self.test_node, future)
+
+        if future.result() is not None:
+            self.test_node.get_logger().info("Result of bool service: %s" % future.result().message)
+        else:
+            self.test_node.get_logger().error(
+                "Exception while calling service: %r" % future.exception()
+            )
```
KYabuuchi, line 92: It would be better to add 
`self.assertTrue(future.result().success, "ndt_scan_matcher is not activated")`  in if section
and 
`raise self.failureException("service trigger failed")` in else section.

SakodaShintaro: I think that's exactly right.
I fixed it and confirmed that the default would be OK and forcing `service_trigger_node` to return false would result in an error.
https://github.com/autowarefoundation/autoware.universe/pull/5347/commits/388c4c583a3958d9c8bc3c6cfb5f7c54e45c7694

KYabuuchi (approved):
LGTM

Issue #5396, opened 2023-10-24
Wrong drivable area generation when expanding for avoidance


mehmetdogru:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I'm convinced that this is not my fault but a bug.

### Description

When expanding the drivable area for avoidance, it is generated wrong. 

https://github.com/autowarefoundation/autoware.universe/assets/48479081/6c176dd9-7057-4805-a953-4037f808a204

### Expected behavior

Drivable area should be generated right after expansion

### Actual behavior

Drivable area generated wrong after expansion

### Steps to reproduce

1. Psim

### Possible causes

Unnecessary skipping monotonic bound points in makeBoundLongitudinallyMonotonic func

PR #5397, opened 2023-10-24
refactor(bpp): make lateral offset smaller for monotonic bounds

mehmetdogru:
## Description

closes https://github.com/autowarefoundation/autoware.universe/issues/5396

This PR solves the specific issue above however logic in the `makeBoundLongitudinallyMonotonic` needs to be fixed as already addressed with a todo comment.

Screenshot from 2023-10-24 12-45-15

## Tests performed

https://github.com/autowarefoundation/autoware.universe/assets/48479081/4692f900-3775-4fc6-94ce-a16836ea70ca

Not applicable.

## Effects on system behavior

Not applicable.

### DIFF
--- a/planning/behavior_path_planner/src/utils/utils.cpp
+++ b/planning/behavior_path_planner/src/utils/utils.cpp
@@ -1935,7 +1935,7 @@ void makeBoundLongitudinallyMonotonic(
 
       const auto theta = normalizeRadian(p1_to_p2 - p2_to_p3);
 
-      return (is_points_left && 0 < theta) || (!is_points_left && theta < 0);
+      return (is_points_left && 0 > theta) || (!is_points_left && theta > 0);
     };
 
   // define a function to remove non monotonic point on bound
@@ -1954,7 +1954,7 @@ void makeBoundLongitudinallyMonotonic(
 
       // NOTE: is_bound_left is used instead of is_points_left since orientation of path point is
       // opposite.
-      const double lat_offset = is_bound_left ? 100.0 : -100.0;
+      const double lat_offset = is_bound_left ? 0.1 : -0.1;
 
       const auto p_bound_1 = getPoint(bound_with_pose.at(bound_idx));
       const auto p_bound_2 = getPoint(bound_with_pose.at(bound_idx + 1));
@@ -1962,7 +1962,7 @@ void makeBoundLongitudinallyMonotonic(
       const auto p_bound_offset =
         calcOffsetPose(getPose(bound_with_pose.at(bound_idx)), 0.0, lat_offset, 0.0);
 
-      if (!is_monotonic(p_bound_1, p_bound_2, p_bound_offset.position, is_points_left)) {
+      if (is_monotonic(p_bound_1, p_bound_2, p_bound_offset.position, is_points_left)) {
         bound_idx++;
         continue;
       }


mehmetdogru:
@takayuki5168 -san I am struggling to understand the reason behind code section here:

https://github.com/autowarefoundation/autoware.universe/blob/db2768626a2e60e859934e0155b42088ef592be5/planning/behavior_path_planner/src/utils/utils.cpp#L1970-L1986

At the moment smaller lateral offset for non-monoton candidate can suppress the issue however non-monotonic bound points stay in the bound with this logic. Since I am not sure of the use case and the reason of the function I am not offering a fix but a workaround. Do you think assigning a smaller value to [lat_offset](https://github.com/autowarefoundation/autoware.universe/blob/db2768626a2e60e859934e0155b42088ef592be5/planning/behavior_path_planner/src/utils/utils.cpp#L1957C20-L1957C30) could cause any problems for the use cases which you created the func to handle? I offered a value for it but please review it and feel free to offer another value or a solution if any

cc: @shmpwk

satoshi-ota:
> I am struggling to understand the reason behind code section here:

@mehmetdogru Have you already seen **[this](https://autowarefoundation.github.io/autoware.universe/main/planning/behavior_path_planner/docs/behavior_path_planner_drivable_area_design/#expansion-with-hatched-road-markings-area)** doc?

At first, let me explain the logic to remove non-monotonic bound.

Step1. The main purpose of this function is to remove points that are not monotonically increasing with respect to the longitudinal direction of the path. So, it appends orientation infomation to each bound point based on the near path point.

https://github.com/autowarefoundation/autoware.universe/blob/64e2ab33dfcfd8a32dc76fe9de0396209a0f5c41/planning/behavior_path_planner/src/utils/utils.cpp#L1885

Step2. Check if the bound is monotonic. **In this step, it uses lateral offset value**. If the angle `P2P1P3` in following fig is larger than 0.0 deg, it is a non-monotonic bound point. 

https://github.com/autowarefoundation/autoware.universe/blob/64e2ab33dfcfd8a32dc76fe9de0396209a0f5c41/planning/behavior_path_planner/src/utils/utils.cpp#L1931-L1939

Step3. If there is a non-monotonic bound, it removes it. In this step, it searches intersect point and append it to bound point.

https://github.com/autowarefoundation/autoware.universe/blob/64e2ab33dfcfd8a32dc76fe9de0396209a0f5c41/planning/behavior_path_planner/src/utils/utils.cpp#L1976-L1985

image

planning/behavior_path_planner/src/utils/utils.cpp
```diff
@@ -1935,7 +1935,7 @@ void makeBoundLongitudinallyMonotonic(
 
       const auto theta = normalizeRadian(p1_to_p2 - p2_to_p3);
 
-      return (is_points_left && 0 < theta) || (!is_points_left && theta < 0);
+      return (is_points_left && 0 > theta) || (!is_points_left && theta > 0);
```
satoshi-ota, line 1938: @mehmetdogru Btw, I might have found a bug. If the **LEFT** bound is monotonic, `theta` will be smaller than zero. On the other hand, if the **RIGHT** bound is monotonic, `theta` will be larger than zero.

Therefore, I guess we should fix it as follow:

```suggestion
      return (is_points_left && 0 < theta) || (!is_points_left && theta > 0);
```

mehmetdogru:
@satoshi-ota -san

Thank you for the explanations and references. I would like to work on this function maybe little bit later and talk about it more on it however currently [these values causes issues](https://github.com/autowarefoundation/autoware.universe/issues/5396). So maybe we can make [lat_offset](https://github.com/autowarefoundation/autoware.universe/blob/db2768626a2e60e859934e0155b42088ef592be5/planning/behavior_path_planner/src/utils/utils.cpp#L1957C20-L1957C30) 10 m for now until we find a fix?

planning/behavior_path_planner/src/utils/utils.cpp
mehmetdogru: Thanks @satoshi-ota -san. I think it is already how you want it to be. What you suggest with the code change is:

If the **LEFT** bound is monotonic, theta will be **larger** than zero. On the other hand, if the **RIGHT** bound is monotonic, theta will be **larger** than zero

satoshi-ota: > If the LEFT bound is monotonic, theta will be larger than zero. On the other hand, if the RIGHT bound is monotonic, theta will be larger than zero

Ah! Sorry, it was a misjudgement :pray:

takayuki5168:
@mehmetdogru
Sorry to be late.
The reason to implement `makeBoundLongitudinallyMonotonic` is just for visualization. Without this implementation, when the map has a complicated hatched road markings area, the drivable area's shape will be very weird.
If it's hard to maintain the function, we can discuss removing the function considering how bag the visualization will be.

mehmetdogru:
@takayuki5168 -san

Thank you for your answer. IMO it is completely ok having a function around created for visualization purposes as long as it does not break any functionality. In this case we have a problem with the functionality ➡️ https://github.com/autowarefoundation/autoware.universe/issues/5396

I am curious about [this variable](https://github.com/autowarefoundation/autoware.universe/blob/db2768626a2e60e859934e0155b42088ef592be5/planning/behavior_path_planner/src/utils/utils.cpp#L1957C20-L1957C30) that why it is 100 m? Is there a use case that requires such big number? 

If not we can assign a reasonable number for now to satisfies both your use cases and sample map drive.

satoshi-ota:
@mehmetdogru Thanks for your proposal.
We set such a large number so that it can find intersecting point in following situation where the vehicle is driving in large drivable area. I think 0.1 ~ 10m is not enough but I can make it smaller value than 100m I guess. So, I'll try to find proper value that doesn't cause the issue at first in sample map and some other maps.

image

satoshi-ota:
Hi @mehmetdogru 
I tried some values for monotonic function. As a result, I confirmed it was ok for most of our use cases to change the value to 20.0m. And this change might be able to fix https://github.com/autowarefoundation/autoware.universe/issues/5396.
Therefore, I think it is ok to change the lateral offset value from 100m to 20.0m right now.

---
FYI: I also tried to solve the underlying cause of the issue. To be honest, I think it is possible to fix the issue without any changing of the variable. I suspected that the is_monotonic function had some bugs, so I re-built the function again in following PR. Then, it seems the issue is fixed. So, I would be appriciate if you could try it when it is convenient for you.

https://github.com/autowarefoundation/autoware.universe/pull/5565

mehmetdogru:
@satoshi-ota 

https://github.com/autowarefoundation/autoware.universe/pull/5565 solves the issue. Thank you for fixing the bug :pray: If it is ok for you lets merge it once you make it open.

satoshi-ota:
@mehmetdogru 
I'm happy to hear that :+1: 
I opened PR. Could you leave some comments on that PR to show that you already confirmed it solves issue https://github.com/autowarefoundation/autoware.universe/issues/5396? (Maybe other maintainer approves it.)

PR #5431, opened 2023-10-27
build(lidar_apollo_segmentation_tvm and lidar_apollo_segmentation_tvm_nodes):  remove download from cmake

lexavtanke:
## Description

According to https://github.com/autowarefoundation/autoware.universe/issues/3137.

This PR updates lidar_apollo_segmentation_tvm and lidar_apollo_segmentation_tvm_nodes packages. Now they use updated version of function get_neural_network which do not download neural networks files. Configuration headers of compiled nets are part of the source tree for now. The rest of the networks files is expected to be in autoware_data folder under lidar_apollo_segmentation directory. 

Tests are removed from lidar_apollo_segmentation_tvm package as they require model files as test fixture which weights 11.9 Mb and can't be part of the source tree. Because we can't use git-lfs for autoware-universe repo. See https://github.com/autowarefoundation/autoware.universe/pull/4923

## Tests performed

Not applicable.

## Effects on system behavior

Not applicable.


perception/lidar_apollo_segmentation_tvm/CMakeLists.txt
```diff
@@ -69,14 +69,6 @@ if(NOT NN_DEPENDENCY STREQUAL "")
     ${tvm_runtime_LIBRARIES}
   )
 
-  if(BUILD_TESTING)
```
ambroise-arm, line 72: I don't think removing the tests is a good change.

I think a better behavior would be to keep what is in this file as is, and instead modify the test so that it checks whether the artifact files are present or not at the very begging; if there are not present print a warning (or something) and pass the test, and if they are present proceed with the test.

ambroise-arm (changes requested):

perception/lidar_apollo_segmentation_tvm/CMakeLists.txt
lexavtanke: Thank you for your feedback. I did some updates to address it. Could you please review it again?

perception/lidar_apollo_segmentation_tvm/test/main.cpp
```diff
@@ -37,9 +39,34 @@ void test_segmentation(bool use_intensity_feature, bool use_constant_feature, bo
...
+
+  fs::path deploy_graph_path = deploy_path / deploy_graph;
+  fs::path deploy_lib_path = deploy_path / deploy_lib;
+  fs::path deploy_param_path = deploy_path / deploy_param;
+
+  if (
+    !fs::exists(deploy_graph_path) || !fs::exists(deploy_lib_path) ||
+    !fs::exists(deploy_param_path)) {
+    printf("Model deploy files not found. Skip test.\n");
+    GTEST_SKIP();
```
ambroise-arm, line 63: nit: would it get logged in a better way with `GTEST_SKIP() << "Model deploy files not found";` instead of printf? (https://google.github.io/googletest/advanced.html#skipping-test-execution)

lexavtanke: I tried yesterday and it doesn't output anything just skip. Same as it doesn't work in the sampling_based_planner 
https://github.com/autowarefoundation/autoware.universe/blob/7789be489dc437d2f2c2b62dba0ddbf27746ff09/planning/sampling_based_planner/sampler_common/test/test_transform.cpp#L49
Screenshot from 2023-11-06 18-00-18

Or Is it only my local issue with gtest?

ambroise-arm: I know `SUCCEED() << "msg";` doesn't print anything, I can see it being the same thing for the skip macro. The printf way looks good enough in the logs (https://github.com/autowarefoundation/autoware.universe/actions/runs/6783547137/job/18438043868?pr=5431#step:7:312):

```
1: [ RUN      ] lidar_apollo_segmentation_tvm.others
1: Model deploy files not found. Skip test.
1: Model deploy files not found. Skip test.
1: Model deploy files not found. Skip test.
1: Model deploy files not found. Skip test.
1: [  SKIPPED ] lidar_apollo_segmentation_tvm.others (0 ms)
```

Another request then: could this skip logic be moved inside of `TEST(lidar_apollo_segmentation_tvm, others)` below so that it only gets executed once instead of 4 times?

lexavtanke: Yes, I updated the code. It check for files only once and provide data_path to test function. 
Output looks like this:
image

ambroise-arm (approved):
LGTM

PR #5467, opened 2023-11-02
feat(ekf_localizer, system_error_monitor): system_error_monitor handles ekf diags

kminoda:
## Description

Although current emergency handling system for localization solely relies on `localization_accuracy`, the criteria may not be accurate enough given that the covariances of NDT, IMU, and wheel odometry are still under development (e.g. the covariance estimation of NDT is still ongoing, and currently we use a static value that is empirically derived). One alternative criteria that we can use (at least temporarily) is to monitor if the valid NDT results have been published recently. In current Autoware implementation, we can monitor this by checking `no_update_count` in `ekf_localizer`. That being said, **I want to make Autoware Universe ready for using the EKF output for emergency handling**. 

Here are the changes that I would like to propose:
- Add EKF diag monitoring in `diagnostic_aggregator`
- Change default parameter of `no_update_count`, since current value (250, which corresponds to 5[sec] with 50Hz EKF) is too large.
- Use the EKF diag for system_error_monitor (only for safe fault by default, see autoware_launch PR)

Note that this PR itself ONLY affects the safe fault and not for single fault or latent fault which are directly connected to the minimum risk maneuver (MRM). The decision whether to really use this for MRM should depend on each reference designs.

Note that we also need to merge https://github.com/autowarefoundation/autoware_launch/pull/674

## Tests performed

Not applicable

## Effects on system behavior

No effect expected on current Autoware, since the diagnostic result of EKF is not used in MRM.

### DIFF
--- a/localization/ekf_localizer/config/ekf_localizer.param.yaml
+++ b/localization/ekf_localizer/config/ekf_localizer.param.yaml
@@ -24,9 +24,9 @@
 
     # for diagnostics
     pose_no_update_count_threshold_warn: 50
-    pose_no_update_count_threshold_error: 250
+    pose_no_update_count_threshold_error: 100
     twist_no_update_count_threshold_warn: 50
-    twist_no_update_count_threshold_error: 250
+    twist_no_update_count_threshold_error: 100
 
     # for velocity measurement limitation (Set 0.0 if you want to ignore)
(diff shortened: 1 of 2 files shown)


YamatoAndo (approved):

kminoda:
@ito-san Would you review this PR?

ito-san (approved):
LGTM

PR #5501, opened 2023-11-07
fix(simple_planning_simulator): set ego pitch to 0 if road slope is not simulated

danielsanchezaran:
### <samp>🤖 Generated by Copilot at 684bffd</samp>

Refactor `simple_planning_simulator_core.cpp` to skip ego pitch calculation when road slope simulation is off and use a constant for gravity. This improves performance and accuracy of the simple planning simulator.

## Tests performed
I found the bug when using the planning simulator on a specific lanelet map that happened to have inclination (I did not know about the inclination). The result was that the controller output would give negative acceleration and the vehicle would not move after planning a path even when road slope simulation is off . The video shows the bug and the image the control output: 

https://github.com/autowarefoundation/autoware.universe/assets/25967964/a46e06de-95f2-478e-bc67-2ee2a73d3a75

image

I saved the specific starting pose of the vehicle and replayed the planning simulator after applying the fix on this PR and the problem was solved: the vehicle could move after planning. The acceleration is no longer negative either.

https://github.com/autowarefoundation/autoware.universe/assets/25967964/55f420fb-371a-4064-b310-3f4d96dfa0e1

## Effects on system behavior

When performing simulations, the simulation calculates the acceleration caused by the vehicle's pitch angle, even if the enable_road_slope_simulation_  parameter is set to false which is an unexpected behavior. With this fix, the acceleration caused by pitch will be zero when enable_road_slope_simulation_  is false. So, in those cases, the acceleration given by the controller will change. 

### DIFF
--- a/simulator/simple_planning_simulator/src/simple_planning_simulator/simple_planning_simulator_core.cpp
+++ b/simulator/simple_planning_simulator/src/simple_planning_simulator/simple_planning_simulator_core.cpp
@@ -326,9 +326,9 @@ void SimplePlanningSimulator::on_timer()
   }
 
   // calculate longitudinal acceleration by slope
-  const double ego_pitch_angle = calculate_ego_pitch();
-  const double acc_by_slope =
-    enable_road_slope_simulation_ ? -9.81 * std::sin(ego_pitch_angle) : 0.0;
+  constexpr double gravity_acceleration = -9.81;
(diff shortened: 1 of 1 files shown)


takayuki5168 (approved):

TakaHoribe (approved):
LGTM

PR #5511, opened 2023-11-07
refactor(landmark_parser): refactored landmark manager

SakodaShintaro:
## Description

Refactored `landmark_parser`.

* ~~Added function `convert_landmark_pose_to_ego_pose`~~
  * ~~https://github.com/autowarefoundation/autoware.universe/pull/5511/files#diff-60f6e4229981714b879a62a22e2d2ee2f2ccae53ee7d68ec67614ba1ddd7b2deR154-R181~~
  * ~~Given the pose of a landmark on the base_link and the ID of the landmark, it returns the Pose of the base_link in the map_frame.~~
* ~~The role was expanded and a class called `LandmarkManager` was created.~~

I have made the following changes instead of the above changes.

* Changed the package name to `landmark_manager`
* Added namespace `landmark_manager`
* Change the return type of `parse_landmark` from map to vector

## Tests performed

It has been confirmed that `ar_tag_based_localizer` works with the same performance as before.

## Effects on system behavior

There is no effect on system behavior.


KYabuuchi (approved):
LGTM :smiley:

SakodaShintaro:
This issue is not new and was not introduced by this pull request; however, Yamato Ando and I have identified a problem. There is a need for better management of multiple tags, especially when multiple tags share the same ID.

```cpp
std::map<std::string, geometry_msgs::msg::Pose> landmarks_;
```
should be changed as follows
```cpp
std::map<std::string, std::vector<geometry_msgs::msg::Pose>> landmarks_;
```

The `convert_landmark_pose_to_ego_pose` function should return the landmark pose that is closest to the EKF Pose when there are multiple landmarks with the same ID. Consequently, the EKF Pose must be passed as an additional input to this function.

For instance, in the case of the LiDAR Intensity Detector, there is currently only one type of board that it can detect.

I think, even with this modification, problems may persist if multiple landmarks with the same ID are detected in a single scan or single image. I believe the current solution is to be careful in operation to prevent this from happening.

For now, I will implement the aforementioned changes.

SakodaShintaro:
Changes have been completed. Please review again.

YamatoAndo (approved):

KYabuuchi (approved):
:+1:

PR #5545, opened 2023-11-09
Updates searchPredictedObject to more accurately detect collision

MrRubyRed:
## Description

When checking whether the collision point is inside the polygon or not, in some instances the searchPredictedObjects fails. This is likely due to the point being on the boundary of the polygon. The change checks for polygon intersection instead of looking if the collision point is inside.

## Tests performed

Tested locally in the scenario_simulator_v2.

## Effects on system behavior

It should stop when using predicted objects in the obstacle stop planner.

Not applicable.

### DIFF
--- a/planning/obstacle_stop_planner/src/node.cpp
+++ b/planning/obstacle_stop_planner/src/node.cpp
@@ -867,14 +867,11 @@ void ObstacleStopPlannerNode::searchPredictedObject(
     bool is_init = false;
     size_t nearest_collision_object_index = 0;
 
-    for (size_t j = 0; j < predicted_object_history_.size(); ++j)
-    {
+    for (size_t j = 0; j < predicted_object_history_.size(); ++j) {
       // check new collision points
-      const auto& obj = predicted_object_history_.at(j).object;
-      if (node_param_.enable_z_axis_obstacle_filtering)
-      {
-        if (!intersectsInZAxis(obj, z_axis_min, z_axis_max))
-        {
+      const auto & obj = predicted_object_history_.at(j).object;
+      if (node_param_.enable_z_axis_obstacle_filtering) {
+        if (!intersectsInZAxis(obj, z_axis_min, z_axis_max)) {
           continue;
         }
       }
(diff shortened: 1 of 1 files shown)


maxime-clem (changes requested):
Thank you for the PR and sorry for the delay in reviewing it.
I will check it in more details ASAP, but for now I have a small code change request.

brkay54:
In the code, both intersect points and the points inside the objects are detected and then pushed into predicted_objects_history_  in the for loop above. You can see the process below of this [line](https://github.com/autowarefoundation/autoware.universe/blob/ebba4f778b79eb6149b9e1b43894d2e855c030ba/planning/obstacle_stop_planner/src/node.cpp#L764). Can you give us more details about failing case?

brkay54:
Hi @MrRubyRed thank you for the effort! Actually, in `obstacle_stop_planner` node, both corner points and intersected points are taken into calculation, however, like you said there is a mistake here. 

At first for loop [here](https://github.com/autowarefoundation/autoware.universe/blob/65bb40985f8c9e488f30d467e13a3cfff9333f28/planning/obstacle_stop_planner/src/node.cpp#L603), it searches collision on trajectory by looking into predicted objects. If there is a collision object, it finds the nearest point of the object (by looking both intersected points and corner points [here](https://github.com/autowarefoundation/autoware.universe/blob/65bb40985f8c9e488f30d467e13a3cfff9333f28/planning/obstacle_stop_planner/src/node.cpp#L805-L830)), and it pushes predicted object, time, and nearest point into predicted_object_history_.

At the second for loop [here](https://github.com/autowarefoundation/autoware.universe/blob/65bb40985f8c9e488f30d467e13a3cfff9333f28/planning/obstacle_stop_planner/src/node.cpp#L854), it searches the nearest collision object in predicted_object_history_ vector. It looks for the pre-calculated nearest point inside the trajectory footprint or not. Like you said, the mistake is here. 

If the nearest point we found in the first loop is an intersected point, bg::intersect might not be true in the second loop, it is a logical mistake, and sorry for that. 
IMO, we need a little structural change. Calculating the nearest collision point in the first loop is not necessary there, it should be in the second loop. The first loop should be closed after the collision object is found. I think moving the nearest point calculations in the first loop to the second loop gonna solve this problem. What do you think?

maxime-clem:
I am closing the PR since there was no update from the author in a very long time. Moreover another PR (https://github.com/autowarefoundation/autoware.universe/pull/5794) was merged and addresses the issue.

Issue #5391, opened 2023-10-24
Autoware won't start on a downward slope


VRichardJP:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I'm convinced that this is not my fault but a bug.

### Description

This is a problem I have observed on site so I wasn't sure what was the cause, but I could easily reproduce it with the planning simulator.

In this example, the vehicle has to cross a short hilly road:

![image](https://github.com/autowarefoundation/autoware.universe/assets/18645627/28a1b00c-28bf-44ae-be35-b494174010f4)

When the vehicle starts from 0km/h on the upward slope, there is no problem. But if it has to start from the downward slope, then it will simply not move. There is no problem if the vehicle has some velocity. 

I understand the longitudinal controller adjusts target acceleration based on the slope, but here the vehicle will simply not start at all. I remember that when this happened on our vehicle, the control command sent by Autoware to the vehicle was stuck at `target velocity = 0.25, target acceleration = -0.4`.

Considering the longitudinal controller is PID based, I would expect the vehicle to reach desired velocity/acceleration even if the slope compensation overshoots. Setting `enable_slope_compensation: false` in the `pid.param.yaml` config file fixes the issue (so it is related to slope compensation logic)

The video below shows the issue, reproduced with the planning simulator:

[downward_slope-2023-10-24_14.28.44.webm](https://github.com/autowarefoundation/autoware.universe/assets/18645627/9d365879-29f2-4420-bfbf-e0308a9c1cf5)

### Expected behavior

Vehicle should be able to start from downward slope.

### Actual behavior

Stuck

### Steps to reproduce

Planning simulator + hilly lanelet

brkay54:
There is no gravity in planning_simulator, this should be the cause of this behavior in the planning simulator. Because while the vehicle in the slope, there is no force in the x direction.

VRichardJP:
@brkay54 I think it is a problem if the slope compensation may stop the vehicle forever. With the PID alone, the vehicle struggles to reach the desired acceleration, but at least it never get stuck.

A few weeks ago I also had the same issue on a different vehicle. There was a stop line on a road going slightly down. Autoware would often stop at the stop line and never restart. I never understood why autoware was sending negative acceleration values while it was supposed to accelerate. But now I think it was caused by the slope compensation.

[...]

PR #5549, opened 2023-11-10
feat(pid_longitudinal_controller): use Integrated error when vehicle is stopped

danielsanchezaran:
## Description

Related to this [Issue](https://github.com/autowarefoundation/autoware.universe/issues/5391)

The goal is to make the PID use the integrative gain to offset possible slope compensation overshoots. The reason this is not currently done is shown on the [documentation](https://github.com/autowarefoundation/autoware.universe/blob/main/control/pid_longitudinal_controller/README.md#pid-control): 

"Also, the integral term is not accumulated when the vehicle is stopped. This is to prevent unintended accumulation of the integral term in cases such as Autoware assumes that the vehicle is engaged, but an external system has locked the vehicle to start."

But I believe that, by tuning the [I effort parameters](https://github.com/autowarefoundation/autoware_launch/blob/c5afec100128e0456aa1b8589187b97b56e0ebe4/autoware_launch/config/control/trajectory_follower/longitudinal_controller.param.yaml#L30):

    max_i_effort: 0.3
    min_i_effort: -0.3
The maximum acceleration caused by the I controller can be tuned to be confortable & safe. So, the velocity limit is not necessary.

## Tests performed

evaluator tests: [TIER4 internal link](https://evaluation.ci.tier4.jp/evaluation/reports/9ae64f05-bc24-5da8-93ab-56531f56d494?project_id=prd_jt) No degradation when compared to today’s baseline [TIER4 internal link](https://evaluation.tier4.jp/evaluation/reports/5f1c1607-9eb6-55e4-942d-87f757294012?project_id=prd_jt)
## Notes for reviewers
Requires launch changes: https://github.com/autowarefoundation/autoware_launch/pull/698

## Effects on system behavior

The use of the integrative gain is no longer limited by the vehicle speed for the PID controller, and the ego can now start on downward slopes.


takayuki5168:
LGTM. Horibe-san will review as well.

TakaHoribe:
@danielsanchezaran Thank you for the PR!

I agree with enabling integration during stops. However, there are several considerations we need to address.

Even in AUTONOMOUS mode, it takes time for a vehicle to start moving. Therefore, the integral term will inevitably accumulate unintentionally during the start. This could have a bad effect on the system, making parameter tuning difficult. When stopped, I think we need a method such as enabling integration only if the vehicle doesn't start moving after a certain period (e.g. 2 seconds).

Additionally, vehicle speed sensors often perform poorly at low speeds, frequently returning zero speed. To avoid unintended integral processing in such vehicles, it remains necessary to have the option to disable integration at low speeds.

danielsanchezaran:
Not sure why the issue was closed, I did not do it.

danielsanchezaran:
With these new changes, the error integration does not kick off until a threshold 
https://github.com/autowarefoundation/autoware.universe/assets/25967964/08fad25b-af3f-4037-a175-3bd4765431c0
There is a parameter to set the threshold limit and a parameter to enable or disable the feature altogether @TakaHoribe @takayuki5168  

Note: the changes to autoware_launch are necessary: https://github.com/autowarefoundation/autoware_launch/pull/698

danielsanchezaran:
> @danielsanchezaran Thank you for the PR!
> 
> I agree with enabling integration during stops. However, there are several considerations we need to address.
> 
> Even in AUTONOMOUS mode, it takes time for a vehicle to start moving. Therefore, the integral term will inevitably accumulate unintentionally during the start. This could have a bad effect on the system, making parameter tuning difficult. When stopped, I think we need a method such as enabling integration only if the vehicle doesn't start moving after a certain period (e.g. 2 seconds).
> 
> Additionally, vehicle speed sensors often perform poorly at low speeds, frequently returning zero speed. To avoid unintended integral processing in such vehicles, it remains necessary to have the option to disable integration at low speeds.

I believe the new changes satisfy these requirements. Could you please take a look again?

takayuki5168:
I leave this PR's review to @TakaHoribe-san.

control/pid_longitudinal_controller/src/pid_longitudinal_controller.cpp
```diff
@@ -959,8 +970,15 @@ double PidLongitudinalController::applyVelocityFeedback(
...
-    (std::abs(current_vel) > m_current_vel_threshold_pid_integrate) && is_under_control;
+
+  const bool vehicle_is_moving = std::abs(current_vel) > m_current_vel_threshold_pid_integrate;
+  const double time_under_control = getTimeUnderControl();
+  const bool vehicle_is_stuck =
+    m_enable_integration_at_low_speed
+      ? !vehicle_is_moving && time_under_control > m_time_threshold_before_pid_integrate
+      : false;
+
+  const bool enable_integration = (vehicle_is_moving || vehicle_is_stuck) && is_under_control;
```
TakaHoribe, line 981: The status of whether a vehicle is stuck should not depend on the flag: `m_enable_integration_at_low_speed`?

```suggestion
  const bool vehicle_is_stuck =
     !vehicle_is_moving && time_under_control > m_time_threshold_before_pid_integrate;

  const bool enable_integration = (vehicle_is_moving || (m_enable_integration_at_low_speed && vehicle_is_stuck)) && is_under_control;
```

TakaHoribe (commented):
@danielsanchezaran LGTM. Two minor comments:

- Would you write a background of this feature on the README? Maybe new users don't know why the time_threshold is necessary. 
- One suggestion below

control/pid_longitudinal_controller/README.md
```diff
@@ -212,6 +217,8 @@ AutonomouStuff Lexus RX 450h for under 40 km/h driving.
 | time_threshold_before_pid_integration | double | How much time without the vehicle moving must past to enable PID error integration. [s]                                                                            | 5.0           |
 | brake_keeping_acc                     | double | If `enable_brake_keeping_before_stop` is true, a certain acceleration is kept during DRIVE state before the ego stops [m/s^2] See [Brake keeping](#brake-keeping). | 0.2           |
 
+Note:
+
```
TakaHoribe, line 221: ?? Should it be removed?

TakaHoribe (approved):
LGTM! (but one odd part in README?)

Issue #3896, opened 2023-06-05
distortion_corrector should update "azimuth" and "distance" field


VRichardJP:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I'm convinced that this is not my fault but a bug.

### Description

Because LIDAR sensors do not shoot laser simultaneously in all directions, raw sensor clouds look distorted when the vehicle turns or drive fast. `pointcloud_preprocessor::DistortionCorrectorComponent` use vehicle twist/imu data to rectify that distortion. Basically, it interpolates cloud points xyz values to make it as if all point laser had been shot simultaneously.

In the implementation `x`, `y` and `z` fields are modified, but not `azimuth` nor `distance`. Downstream nodes thus cannot use these 2 fields reliably.

![distortion_corrector](https://github.com/autowarefoundation/autoware.universe/assets/18645627/39149961-b721-4a1c-86a9-7a8564d3e6f6)

For example, ring outlier filter is used after the distortion corrector filter, and current implementation relies both on azimuth and distance fields values. If the azimuth field is wrong, then the filter output may be incorrect. In ring outlier filter case, distortion effect is not "too bad", because only consecutive points on the same ring are considered (distortion effect over 2 consecutive points is negligible), except for the last and first point of the ring, which would be far away from each other due to distortion.

This leads to another problem: what do "azimuth" and "distance" fields mean? Are these values related to the sensor origin, or the current frame? Currently, these values are never updated, so they represent points in the original sensor frame, even though the cloud frame may be different (typically `base_link`). I don't think it is a problem per se, since these fields most likely only make sense in the sensor frame. For example ring outlier filter uses azimuth to group ring points together, but if the values were expressed in a different frame the logic would not make sense anymore.

So if the distortion corrector needs to update the "azimuth" and "distance" fields, it needs to be done in the original sensor frame, not the current one. This would definitely lead to some convoluted implementation.

Instead, it would be maybe simpler to have the input cloud of distortion corrector (and ring outlier filter) use the sensor frame. For example, in the `sample_sensor_kit`, the `velodyne_node_container` transforms the raw input cloud to `base_link` frame in order to filter points on the vehicle body. The sensor frame could be used as output frame for the distortion filter right after.

### Expected behavior

[...]

PR #5560, opened 2023-11-13
feat(pointcloud_preprocessor): update azimuth and distance in distortion_correction_node

vividf:
## Description
This PR solves issue #3896 
This PR updates the azimuth and distance value based on the undistorted xyz value. The user can decide to use this feature by setting the parameter  `update_azimuth_and_distance` to true.

The distance is calculated with the undistorted xyz value. 
The azimuth is calculated by using an undistorted xy value with the `cv::fastAtan2`. By using `cv::fastAtan2` instead of `atan2`, it can fasten the calculation 6 times faster with negligible error.

**For different frames:**
1. **base_link**: If the frame of the input pointcloud is base_link, even the `update_azimuth_and_distance` is set to true, we won't update the azimuth and distance, as it didn't make sense to update these value when the frame is in base_link.

2. **sensor_frame**: If the frame of the input pointcloud is sensor_frames, we will update the azimuth value and distance value for each point. However different sensors have different coordinate systems after the nebula driver. For example, Velodyne pointcloud defines the x-axis as 0 degrees, but the y-axis as 270 degrees. Hesai pointcloud defines the y-axis as 0 degrees and the x-axis as 90 degrees. Both of them are rotating clockwise. 

To solve this problem, we update the azimuth value based on the Cartesian coordinate system, in which the x-axis is 0 degrees and the y-axis is 90 degrees. For instance, the below result is recorded when the input pointcloud of the distortion correction node is `/sensing/lidar/top/pointcloud_raw_ex `, and the frame_id is velodyne_top.

> before distortion correction (Velodyne): 
> x: -11.3042, y: -3.0857, z: -1.90711, azimuth: 16200, distance: 11.872
> after distortion correction:
> x: -11.2044, y: -3.67424, z: -1.89818, azimuth: 19815.9, distance: 11.9432

Based on this situation, I write documentation in the readme to notify the user that they should be aware of the azimuth of the Lidar.

## Tests performed
Testing rosbag is from the sample rosbag: https://autowarefoundation.github.io/autoware-documentation/main/tutorials/ad-hoc-simulation/rosbag-replay-simulation/

Test with launching `ros2 launch autoware_launch logging_simulator.launch.xml map_path:=$HOME/autoware_map/sample-map-rosbag vehicle_model:=sample_vehicle sensor_model:=sample_sensor_kit` 

Result of the updated azimuth and distance value:
> x: 7.65626, y : -10.5918, z: 1.58103, azimuth: 30585.4, distance: 13.1645

## Notes for reviewers
@drwnz 
The user can only change the parameter via the console with this single PR.
I also set the parameter in the nebula_node_container.launch.py in this PR (https://github.com/autowarefoundation/sample_sensor_kit_launch/pull/81), which can change the parameter more easily.
I also want to do the same thing for the https://github.com/RobotecAI/awsim_sensor_kit_launch/blob/main/common_awsim_sensor_launch/launch/velodyne_node_container.launch.py#L132C1-L140C64

## Interface changes
Added a new parameter `update_azimuth_and_distance` for the user to decide whether they want to update the azimuth and distance value.

## Effects on system behavior
The default value of the `update_azimuth_and_distance` is **False**, which won't cause any change to the system. If the user wants to update the azimuth and distance, then they can set the parameter to **True** via launch file or console `ros2 param set /distortion_corrector_node update_azimuth_and_distance True`.

Once the `update_azimuth_and_distance` is set to True, the time cost will increase by around 13% for the distortion correction node (around 1 ms on my laptop).

**Performance analysis:** 
Below testing runs the rosbag with a total of 81 frames of pointcloud.
1. Without only additional updates (only XYZ)

> test1: 0.00715894
> test2: 0.00704335
> test3: 0.0070083
> avg:  0.00707019666

2. With additional distance and azimuth (cv::fastAtan2) update

> test1:  0.008114
> test2: 0.00785841
> test3: 0.00807316
> avg: 0.00801519,  increasing 13% from no updates

5. With additional distance and azimuth (atan2) update

> test1: 0.0116086
> test2: 0.0122381
> test3: 0.0117111
> avg: 0.0118526,  increasing 63% from no updates


yukkysaito:
@badai-nguyen @miursh Can you review this?

sensing/pointcloud_preprocessor/src/distortion_corrector/distortion_corrector.cpp
```diff
@@ -123,6 +124,8 @@ void DistortionCorrectorComponent::onPointCloud(PointCloud2::UniquePtr points_ms
     return;
   }
 
+  update_azimuth_and_distance_ = get_parameter("update_azimuth_and_distance").as_bool();
```
badai-nguyen, line 127: Is this typo? IMO 'update_azimuth_and_distance_' is already updated in constructor so this can be remove.
```suggestion
```

sensing/pointcloud_preprocessor/src/distortion_corrector/distortion_corrector.cpp
```diff
@@ -39,6 +39,7 @@ DistortionCorrectorComponent::DistortionCorrectorComponent(const rclcpp::NodeOpt
   // Parameter
   time_stamp_field_name_ = declare_parameter("time_stamp_field_name", "time_stamp");
   use_imu_ = declare_parameter("use_imu", true);
+  update_azimuth_and_distance_ = declare_parameter("update_azimuth_and_distance", false);
```
badai-nguyen, line 42: As https://github.com/autowarefoundation/autoware-documentation/blob/main/docs/contributing/coding-guidelines/ros-nodes/parameters.md#parameters, I think it is better to remove default param from declare_parameter
```suggestion
  update_azimuth_and_distance_ = declare_parameter<bool>("update_azimuth_and_distance");
```

badai-nguyen:
@vividf Thank you for your PR. 
Could you also fix spell-check error as it is required?
You also could check those spell errors locally by installing https://github.com/tier4/cspell-dicts

sensing/pointcloud_preprocessor/src/distortion_corrector/distortion_corrector.cpp
```diff
@@ -302,6 +307,10 @@ bool DistortionCorrectorComponent::undistortPointCloud(
     *it_y = static_cast<float>(undistorted_point.getY());
     *it_z = static_cast<float>(undistorted_point.getZ());
 
+    if (update_azimuth_and_distance_) {
+      *it_distance = sqrt(*it_x * *it_x + *it_y * *it_y + *it_z * *it_z);
+      *it_azimuth = cv::fastAtan2(*it_y, *it_x) * 100;
```
badai-nguyen, line 312: @vividf 
As [fastAtan2 function description](https://docs.opencv.org/3.4/db/de0/group__core__utils.html#ga7b356498dd314380a0c386b059852270), the accuracy is 0.3 deg which greater than VLS128 azimuth resolution (~0.2deg) so do you think some beams' order in a ring might be changed?
Anyway, IMO, even it's happened, it changes only 1 beam so it will not affect other filters result, such as Dual_return_filter which uses rertified topic as input. @miursh Do you have any concern about this?

vividf (commented):
.

sensing/pointcloud_preprocessor/src/distortion_corrector/distortion_corrector.cpp
vividf: Sure, I can remove the default value. 
I have one question is that should I create the parameter file for this node? 
Currently, the other node in the pointcloud preprocessor doesn't have the parameter file and also uses declare_parameter with the default value.

vividf: Sure! I will remove the default param! 
Thanks for the advice!

sensing/pointcloud_preprocessor/src/distortion_corrector/distortion_corrector.cpp
drwnz: I think it should be fine for now - if we run into issues we could optionally use a custom look up table implementation or use a method which has higher accuracy (at higher computational cost) but as you mention, I doubt that this will cause any problems for the modules which actually use azimuth.

badai-nguyen: @drwnz 
I agree that it's acceptable to use fastAtan2 for now. 
@vividf Could you add the reference about fastAtan accuracy and a warning about posibility of changing beam order for high azimuth resolution LiDAR in filter README?

sensing/pointcloud_preprocessor/docs/distortion-corrector.md
```diff
@@ -39,9 +39,12 @@ $ ExactPointTime = TimeStamp + TimeOffset $
...
+| ----------------------------- | ------ | ------------- | ----------------------------------------------------------- |
+| `timestamp_field_name`        | string | "time_stamp"  | time stamp field name                                       |
+| `use_imu`                     | bool   | true          | use gyroscope for yaw rate if true, else use vehicle status |
+| `update_azimuth_and_distance` | bool   | false         | update the azimuth and distance based on undistorted xyz    |
 
 ## Assumptions / Known limits
+
+When setting the parameter `update_azimuth_and_distance` to True, the node will calculate the azimuth value based on the undistorted XYZ by using atan2, which will return the azimuth value with the coordinate system that the x-axis is 0 degrees and the y-axis is 90 degrees. Please make sure the frame coordinates are the same as the coordinate system mentioned above. For autoware users, the frame_id of input pointcloud is base_link, which the coordinate system is the same as the coordinate system mentioned above, therefore it won't cause any issues. Also note that by setting the parameter `update_azimuth_and_distance` to True, the time will increase by around 13%.
```
drwnz: Recommended:
When setting the parameter `update_azimuth_and_distance` to `true`, the node will calculate the per-point azimuth and distance values based on the undistorted XYZ using the atan2 function. The azimuth value will have its origin along the x-axis. Please make sure the frame coordinates are the same as the coordinate system mentioned above. For Autoware users, the `frame_id` of the input pointcloud is `base_link`, in which the coordinate system is the same as the coordinate system mentioned above, therefore it will not cause any issues. Also note that by setting the parameter `update_azimuth_and_distance` to `true`, the processing time will increase by approximately 13%.

drwnz (changes requested):
Please see the comments about README. Thanks!

sensing/pointcloud_preprocessor/docs/distortion-corrector.md
vividf: Thanks! I updated the README, could you please help me review it again :).

drwnz:
@vividf can you please update the status of this PR? Thanks!

vividf:
@drwnz 
This PR is blocked by the Velodyne fixes for azimuth value.

PR #5587, opened 2023-11-15
fix(ekf_localizer): fixed bug where calculation of delay_step may be incorrect when timer callback is stuck 

TaikiYamada4:
## Description

In the `measurementUpdatePose` and `measurementUpdateTwist` of ekf_module.cpp, a variable `delay_step`, which is used for extracting previous states in the EKF, have been computed by the following. 
```
const int delay_step = std::roundf(delay_time / dt);
```
Currently, the devisor `dt` is the measured lap time of the timer_callback, but since ekf_localizer is a single thread node, the lap time of the timer callback may be different than expected. Generally, the lap time gets once longer when other callbacks provoke, and then gets shorter to catch up. This results to incorrect extracting of previous states. Thanks to the damping strategy in the EKF, this bug hasn't been exposed as a critical outcome, but yet it should be solved.

To solve this, I implemented a feature that the ekf_localizer will perform prediction forcefully according to its `prediction_frequency` defined in ekf_localizer.param.yaml. The ekf_localizer will measure and accumulate the lap time of the timer callback, and performs prediction when it exceeds the period (calculated from `prediction frequency`)

### DIFF
--- a/localization/ekf_localizer/include/ekf_localizer/ekf_localizer.hpp
+++ b/localization/ekf_localizer/include/ekf_localizer/ekf_localizer.hpp
@@ -163,8 +163,8 @@ private:
   AgedObjectQueue<geometry_msgs::msg::PoseWithCovarianceStamped::SharedPtr> pose_queue_;
   AgedObjectQueue<geometry_msgs::msg::TwistWithCovarianceStamped::SharedPtr> twist_queue_;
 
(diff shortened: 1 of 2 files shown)


TaikiYamada4:
This PR is not needed anymore since a better approach has been merged
https://github.com/autowarefoundation/autoware.universe/pull/5691

PR #5512, opened 2023-11-07
feat(obstacle_cruise_planner)!: change logic to secure safe distance margin as polygons distance

yuki-takagi-66:
## Description

BEFORE:
The current implementation calculate the stop point by the closest collision point.
This cause inappropriate behavior for looped trajectories.
image-20231020-104158

AFTER:
The stop point is calculation relies on the distance between the polygons
image
image
image

The followings are achieved by this PR, 
- Proper handling of collision in extended trajectory is achieved (It is the main motivation of this PR).
- The stop point modification around the goal point is now working properly.
- This PR contains the similar feature to https://github.com/autowarefoundation/autoware.universe/pull/4952, and therefore approaching on curve is now provided as default setting.
- `goal_extension_length` and `goal_extension_interval` may be no longer used and `safe_distance_margin` and `decimate_trajectory_step_length` are used instead.

## Related links
No launch file changes.

## Tests performed
psim test and tier4 internal tests was performed.
Implemented logic works well.
Calculation time seems to increase somewhat but may not be doubled.

https://evaluation.tier4.jp/evaluation/reports/9843ed0c-3a29-54c2-9ded-8522350900ac?project_id=prd_jt

## Notes for reviewers
While no longer used parameters remain, I will remove these parameters after this PR has been agreed upon.

## Interface changes
No interface changes.

## Effects on system behavior
The minor bugs mentioned above have been fixed.
The slow down planning and the cruise planning may not changed.

## Pre-review checklist for the PR author

The PR author **must** check the checkboxes below when creating the PR.

- [x] I've confirmed the [contribution guidelines].
- [x] The PR follows the [pull request guidelines].

## In-review checklist for the PR reviewers

The PR reviewers **must** check the checkboxes below before approval.

- [ ] The PR follows the [pull request guidelines].
- [ ] The PR has been properly tested.
- [ ] The PR has been reviewed by the code owners.

## Post-review checklist for the PR author

The PR author **must** check the checkboxes below before merging.

- [ ] There are no open discussions or they are tracked via tickets.
- [ ] The PR is ready for merge.

After all checkboxes are checked, anyone who has write access can merge the PR.

[contribution guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/
[pull request guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/pull-request-guidelines/

PR #5619, opened 2023-11-17
feat(obstacle_cruise_planner): prior generation of collision distance

yuki-takagi-66:
## Description
BEFORE:
The current implementation calculate the stop point by the closest collision point.
This cause inappropriate behavior for looped trajectories.
image-20231020-104158

AFTER:
Screenshot from 2023-11-17 16-54-44

The followings are achieved by this PR, 
- Proper handling of collision in extended trajectory is achieved (It is the main motivation of this PR).
- The stop point modification around the goal point is now working properly.
- `goal_extension_length` and `goal_extension_interval` may be no longer used and `safe_distance_margin` and `decimate_trajectory_step_length` are used instead.

This PR similar to https://github.com/autowarefoundation/autoware.universe/pull/5512, but this PR maintain more compatibility.
In other words, the function that included in the above PR  _This PR contains the similar feature to https://github.com/autowarefoundation/autoware.universe/pull/4952, and therefore approaching on curve is now provided as default setting._ is obsoleted.

## Tests performed
psim tests and Tier4 internal tests was performed in success.
No degrade directly related to this PR founds.
https://evaluation.tier4.jp/evaluation/reports/52d22759-d1af-5896-b016-3b45e73d4138?project_id=prd_jt

## Notes for reviewers
While no longer used parameters remain, I will remove these parameters after this PR has been agreed upon.

## Interface changes
No interface changes.

## Effects on system behavior
Minor bugs mentioned above have been fixed.
The slow down planning and the cruise planning may not changed.

### DIFF
--- a/planning/obstacle_cruise_planner/include/obstacle_cruise_planner/common_structs.hpp
+++ b/planning/obstacle_cruise_planner/include/obstacle_cruise_planner/common_structs.hpp
@@ -120,7 +120,7 @@ struct StopObstacle : public TargetObstacleInterface
   }
   Shape shape;
   geometry_msgs::msg::Point
-    collision_point;  // TODO: takagi; this member variable still used in
+    collision_point;  // TODO(yuki_takagi): this member variable still used in
                       // calculateMarginFromObstacleOnCurve() and  should be removed as it can be
                       // replaced by ”dist_to_collide_on_decimated_traj”
   double dist_to_collide_on_decimated_traj;
--- a/planning/obstacle_cruise_planner/include/obstacle_cruise_planner/common_structs.hpp
+++ b/planning/obstacle_cruise_planner/include/obstacle_cruise_planner/common_structs.hpp
@@ -110,16 +110,20 @@ struct StopObstacle : public TargetObstacleInterface
     const std::string & arg_uuid, const rclcpp::Time & arg_stamp,
     const geometry_msgs::msg::Pose & arg_pose, const Shape & arg_shape,
     const double arg_lon_velocity, const double arg_lat_velocity,
-    const geometry_msgs::msg::Point arg_collision_point, const double arg_dist_to_collide)
+    const geometry_msgs::msg::Point arg_collision_point,
+    const double arg_dist_to_collide_on_decimated_traj)
   : TargetObstacleInterface(arg_uuid, arg_stamp, arg_pose, arg_lon_velocity, arg_lat_velocity),
     shape(arg_shape),
     collision_point(arg_collision_point),
-    dist_to_collide(arg_dist_to_collide)
+    dist_to_collide_on_decimated_traj(arg_dist_to_collide_on_decimated_traj)
   {
   }
   Shape shape;
-  geometry_msgs::msg::Point collision_point;
-  double dist_to_collide;
+  geometry_msgs::msg::Point
+    collision_point;  // TODO: takagi; this member variable still used in
+                      // calculateMarginFromObstacleOnCurve() and  should be removed as it can be
+                      // replaced by ”dist_to_collide_on_decimated_traj”
+  double dist_to_collide_on_decimated_traj;
 };
 
 struct CruiseObstacle : public TargetObstacleInterface
--- a/planning/obstacle_cruise_planner/include/obstacle_cruise_planner/common_structs.hpp
+++ b/planning/obstacle_cruise_planner/include/obstacle_cruise_planner/common_structs.hpp
@@ -110,14 +110,16 @@ struct StopObstacle : public TargetObstacleInterface
     const std::string & arg_uuid, const rclcpp::Time & arg_stamp,
     const geometry_msgs::msg::Pose & arg_pose, const Shape & arg_shape,
     const double arg_lon_velocity, const double arg_lat_velocity,
-    const geometry_msgs::msg::Point arg_collision_point)
+    const geometry_msgs::msg::Point arg_collision_point, const double arg_dist_to_collide)
   : TargetObstacleInterface(arg_uuid, arg_stamp, arg_pose, arg_lon_velocity, arg_lat_velocity),
     shape(arg_shape),
-    collision_point(arg_collision_point)
+    collision_point(arg_collision_point),
+    dist_to_collide(arg_dist_to_collide)
   {
   }
   Shape shape;
   geometry_msgs::msg::Point collision_point;
+  double dist_to_collide;
 };
 
 struct CruiseObstacle : public TargetObstacleInterface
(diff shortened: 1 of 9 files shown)


planning/obstacle_cruise_planner/include/obstacle_cruise_planner/node.hpp
```diff
@@ -70,6 +70,9 @@ class ObstacleCruisePlannerNode : public rclcpp::Node
   std::optional<geometry_msgs::msg::Point> createCollisionPointForStopObstacle(
     const std::vector<TrajectoryPoint> & traj_points, const std::vector<Polygon2d> & traj_polys,
     const Obstacle & obstacle) const;
+  std::optional<double> calcCollisionDistanceForStopObstacle(
+    const std::vector<TrajectoryPoint> & traj_points, const std::vector<Polygon2d> & traj_polys,
+    const Obstacle & obstacle) const;
```
takayuki5168, line 75:
```suggestion
```
Seems this function declaration is not required.

planning/obstacle_cruise_planner/src/polygon_utils.cpp
```diff
@@ -134,6 +134,25 @@ Polygon2d createOneStepPolygon(
   return hull_polygon;
 }
 
+std::optional<double> calcCollisionDistanceForStopObstacle(
+  const std::vector<TrajectoryPoint> & traj_points, const std::vector<Polygon2d> & traj_polys,
+  const Obstacle & obstacle)
+{
+  auto obstacle_polygon = tier4_autoware_utils::toPolygon2d(obstacle.pose, obstacle.shape);
```
takayuki5168, line 141:
```suggestion
  const auto obstacle_polygon = tier4_autoware_utils::toPolygon2d(obstacle.pose, obstacle.shape);
```

planning/obstacle_cruise_planner/src/polygon_utils.cpp
```diff
@@ -134,6 +134,25 @@ Polygon2d createOneStepPolygon(
...
+  for (size_t col_i = 0; col_i < traj_polys.size(); ++col_i) {
+    if (boost::geometry::intersects(traj_polys.at(col_i), obstacle_polygon)) {
+      try {
+        double prev_index_dist =
```
takayuki5168, line 145:
```suggestion
        const double prev_index_dist =
```

planning/obstacle_cruise_planner/src/polygon_utils.cpp
```diff
@@ -134,6 +134,25 @@ Polygon2d createOneStepPolygon(
...
 }
 
+std::optional<double> calcCollisionDistanceForStopObstacle(
+  const std::vector<TrajectoryPoint> & traj_points, const std::vector<Polygon2d> & traj_polys,
+  const Obstacle & obstacle)
+{
+  auto obstacle_polygon = tier4_autoware_utils::toPolygon2d(obstacle.pose, obstacle.shape);
+  for (size_t col_i = 0; col_i < traj_polys.size(); ++col_i) {
+    if (boost::geometry::intersects(traj_polys.at(col_i), obstacle_polygon)) {
+      try {
```
takayuki5168, line 144: Why do you use try here?
I think we can handle the corner case without using try, which is a better implementation

planning/obstacle_cruise_planner/src/planner_interface.cpp
```diff
@@ -401,12 +373,13 @@ std::vector<TrajectoryPoint> PlannerInterface::generateStopTrajectory(
 
   stop_planning_debug_info_.set(
     StopPlanningDebugInfo::TYPE::STOP_CURRENT_OBSTACLE_DISTANCE,
-    closest_obstacle_dist - abs_ego_offset);  // TODO(murooka)
+    closest_stop_obstacle->dist_to_collide);  // TODO(murooka)
```
takayuki5168, line 376: Is this correct? Seems `dist_to_collide` does not consider the following distances
- the distance from the path start point and the ego position
- the distance from the base link to the ego front position.

planning/obstacle_cruise_planner/include/obstacle_cruise_planner/common_structs.hpp
```diff
@@ -110,14 +110,16 @@ struct StopObstacle : public TargetObstacleInterface
     const std::string & arg_uuid, const rclcpp::Time & arg_stamp,
     const geometry_msgs::msg::Pose & arg_pose, const Shape & arg_shape,
     const double arg_lon_velocity, const double arg_lat_velocity,
-    const geometry_msgs::msg::Point arg_collision_point)
+    const geometry_msgs::msg::Point arg_collision_point, const double arg_dist_to_collide)
   : TargetObstacleInterface(arg_uuid, arg_stamp, arg_pose, arg_lon_velocity, arg_lat_velocity),
     shape(arg_shape),
-    collision_point(arg_collision_point)
+    collision_point(arg_collision_point),
```
takayuki5168, line 116: Is it hard to remove the collision_point here?
If this still exists here, it will be used for some calculation by mistake.

Actually, the nearest search is applied to this collision point, which is not expected.
https://github.com/autowarefoundation/autoware.universe/blob/cb409fa58f516f8499b6970820c39f4de345ac11/planning/obstacle_cruise_planner/src/planner_interface.cpp#L407

takayuki5168: If this variable has to exist for the debug information, we'd better change the variable name to `collision_point_for_debug` for example so that the developer can understand the reason for this variable to exist.

planning/obstacle_cruise_planner/src/planner_interface.cpp
```diff
@@ -269,85 +269,58 @@ std::vector<TrajectoryPoint> PlannerInterface::generateStopTrajectory(
...
-    planner_data.traj_points, 0, closest_stop_obstacle->collision_point);
-
   const auto ego_segment_idx =
     ego_nearest_param_.findSegmentIndex(planner_data.traj_points, planner_data.ego_pose);
-  const auto negative_dist_to_ego = motion_utils::calcSignedArcLength(
-    planner_data.traj_points, planner_data.ego_pose.position, ego_segment_idx, 0);
-  const double dist_to_ego = -negative_dist_to_ego;
+  double dist_to_collide_on_ref_traj =
+    motion_utils::calcSignedArcLength(planner_data.traj_points, 0, ego_segment_idx) +
+    closest_stop_obstacle->dist_to_collide;
```
takayuki5168, line 276: This cannot consider the detailed ego position since `ego_segment_idx` means the resampled index for the ego position.

takayuki5168: Also, the distance from the base link to the ego front position (`abs_ego_offset`) seems not considered with this PR's change.

takayuki5168 (approved):

PR #5651, opened 2023-11-22
AEB: update readme and add some comments

xianglunkai:


xianglunkai:
By the way,there seems to be an issue with the calculation of RSS when vehicles are traveling in opposite directions?

PR #5691, opened 2023-11-27
fix(ekf_localizer): correct the calculation of delay_step in updateMeasurementPose/Twist

TaikiYamada4:
## Description

In the [`measurementUpdatePose()`](https://github.com/autowarefoundation/autoware.universe/blob/017aed4f8888a8cae468ae04e214356e0dd0192f/localization/ekf_localizer/src/ekf_module.cpp#L149) and [`measurementUpdateTwist()`](https://github.com/autowarefoundation/autoware.universe/blob/017aed4f8888a8cae468ae04e214356e0dd0192f/localization/ekf_localizer/src/ekf_module.cpp#L245) of ekf_module.cpp, a variable [`delay_step`](https://github.com/autowarefoundation/autoware.universe/blob/017aed4f8888a8cae468ae04e214356e0dd0192f/localization/ekf_localizer/src/ekf_module.cpp#L173), which is used for extracting previous states in the EKF, have been computed by the following.
```
const int delay_step = std::roundf(delay_time / dt);
```
Currently, the devisor `dt` is the measured lap time of the timer_callback, but since ekf_localizer is a single thread node, the lap time of the timer callback may be different than expected. Generally, the lap time gets once longer when other callbacks provoke, and then gets shorter to catch up. This results to incorrect extracting of previous states. Thanks to the damping strategy in the EKF, this bug hasn't been exposed as a critical outcome, but yet it should be solved.

To solve this, I added a variable to accumulated_delay_times_ to store the time how long the previous states in the EKF are old.
With this, we can search the state index that has the closest delay time to that of the measurement, which is the correct `delay_step` we want.

(Added on 2023/12/04)
I also added a feature to output a WARNING message, if the timer callback period (`ekf_dt_`) gets too long for some reason.

## Related links

To debug these commits, I locally made a debug commit by my self.
https://github.com/TaikiYamada4/autoware.universe/tree/feat/debug_y_ekf_with_accumulated_delay_time

The PR below has the same objective, but since this PR's approach is more simple than the previous one, I'd rather like to merge this PR and the previous PR should be closed if so.
https://github.com/autowarefoundation/autoware.universe/pull/5587

## Tests performed

These changes have been tested through the [rosbag replay simulation in the tutorial](https://autowarefoundation.github.io/autoware-documentation/main/tutorials/ad-hoc-simulation/rosbag-replay-simulation/).
Plus, I have monitored the first element of [variable `y_ekf` in `measurementUpdatePose()`](https://github.com/autowarefoundation/autoware.universe/blob/b5d6c51c26a7ee2597f85dd6fe3c04e75d13414b/localization/ekf_localizer/src/ekf_module.cpp#L201) which suffers from the mentioned bug. This is monitored in the same rosbag replay simulation.
(Extracting values of y_ekf is not included in this PR and done by [my local branch](https://github.com/TaikiYamada4/autoware.universe/tree/feat/debug_y_ekf_with_accumulated_delay_time))

Despite this value should behave in a continuous manner, there are spikes due to the bug mentioned here (Upper figure).
After applying the changes in this PR, they will disappear (Lower figure).

compared_y_ekf

## Interface changes

None

## Effects on system behavior

The predicted observation in the EKF should not look up to wrong previous states.
However, since the smoothing strategy in the EKF have been relieving the effects of the bug, it hardly makes any changes to the outputs in a macro manner.

Maybe this bug fix benefits the system when the timer callback in the ekf_localizer is unstable.


localization/ekf_localizer/src/ekf_localizer.cpp
```diff
@@ -109,9 +109,18 @@ void EKFLocalizer::updatePredictFrequency()
     if (get_clock()->now() < *last_predict_time_) {
       warning_->warn("Detected jump back in time");
     } else {
-      ekf_rate_ = 1.0 / (get_clock()->now() - *last_predict_time_).seconds();
-      DEBUG_INFO(get_logger(), "[EKF] update ekf_rate_ to %f hz", ekf_rate_);
-      ekf_dt_ = 1.0 / std::max(ekf_rate_, 0.1);
+      /* Measure dt */
+      ekf_dt_ = (get_clock()->now() - *last_predict_time_).seconds();
+      ekf_rate_ = 1 / ekf_dt_;
+      DEBUG_INFO(get_logger(), "[EKF] update ekf_dt_ to %f seconds (= %f hz)", ekf_dt_, ekf_rate_);
```
YamatoAndo, line 115: `ekf_rate_` doesn't seem to be used elsewhere anymore, so it may not be necessary to keep it as a member variable.

localization/ekf_localizer/src/ekf_module.cpp
```diff
@@ -38,6 +38,7 @@
 EKFModule::EKFModule(std::shared_ptr<Warning> warning, const HyperParameters params)
 : warning_(std::move(warning)),
   dim_x_(6),  // x, y, yaw, yaw_bias, vx, wz
+  accumulated_delay_times_(params.extend_state_step, 1.0E15),
```
YamatoAndo, line 41: Is there a specific reason for initializing it with a very small value rather than 0?

TaikiYamada4:
I also added a feature to output a WARNING message, if the timer callback period (`ekf_dt_`) gets too long for some reason.

localization/ekf_localizer/src/ekf_module.cpp
YamatoAndo: make sense. OK

YamatoAndo (approved):

Issue #5289, opened 2023-10-12
DevOps Dojo: ROS Node Configuration - GitHub Actions for multiple Schema/Parameter files


kaspermeck-arm:
### Checklist

- [x] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [x] I've searched other issues and no duplicate issues were found.
- [x] I'm convinced that this is not my fault but a bug.

### Description

tbd

### Expected behavior

tbd

### Actual behavior

tbd

### Steps to reproduce

tbd

kaspermeck-arm:
@kaspermeck-arm @ambroise-arm - create a PR and let @mitsudome-r know so he can create a new release so that Autoware.Universe uses the latest GitHub actions

Workflow:
* https://github.com/autowarefoundation/autoware.universe/blob/main/.github/workflows/json-schema-check.yaml

Script:
* https://github.com/autowarefoundation/autoware-github-actions/blob/main/json-schema-check/action.yaml

PR #4902, opened 2023-09-06
refactor(image_projection_based_fusion): add JSON Schema and remove default value spefications

ktro2828:
## Description

- Add ROS node configuration with JSON Schema as discussed in https://github.com/orgs/autowarefoundation/discussions/3371
- Remove default parameter specified in declare_parameter()

## Tests performed

Not applicable.

## Effects on system behavior

Not applicable.

## Pre-review checklist for the PR author

The PR author **must** check the checkboxes below when creating the PR.

- [x] I've confirmed the [contribution guidelines].
- [x] The PR follows the [pull request guidelines].

## In-review checklist for the PR reviewers

The PR reviewers **must** check the checkboxes below before approval.

- [ ] The PR follows the [pull request guidelines].

## Post-review checklist for the PR author

The PR author **must** check the checkboxes below before merging.

- [ ] There are no open discussions or they are tracked via tickets.

After all checkboxes are checked, anyone who has write access can merge the PR.

[contribution guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/
[pull request guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/pull-request-guidelines/

PR #5798, opened 2023-12-06
refactor(json-schema): align schema and config file names

ambroise-arm:
## Description

The current paths of the schema/config files need to be changed slightly to account for a limitation in the current design where a package can't have more than one schema file (otherwise the CI fails). The update to the design is described in https://github.com/autowarefoundation/autoware-github-actions/issues/232#issuecomment-1746971463

The schema files need to follow a NODE_NAME.schema.json naming convention and the configuration files need to follow a NODE_NAME*.param.yaml naming convention. It allows to have one schema file per node in a package.

Rename files that didn't match this pattern.

## Related links

Align the autoware.universe codebase with https://github.com/autowarefoundation/autoware.universe/issues/5289

After this PR is merged, https://github.com/autowarefoundation/autoware-github-actions/pull/263 can be moved out of draft

This will probably unblock https://github.com/autowarefoundation/autoware.universe/pull/4902

## Tests performed

Verified the configurations files renamed were not referred to in other autoware.universe packages.

Only CI tests were run

## Interface changes

Changed the paths of some .schema.json and .param.yaml files.

## Effects on system behavior

None


isamu-takagi:
@ambroise-arm Different nodes may be created with partially matching names (e.g. `my_node` and `my_node_tool`). In this case, `my_node.schema.json` will check both `my_node.param.yaml` and `my_node_tool.param.yaml`.

So, I suggest using a slash that cannot be used in nodes as a separator. In other words, create a directory with the node name like below.
- `schema/my_node.schema.json`
- `schema/my_node_tool.schema.json`
- `config/my_node/setting1.param.yaml`
- `config/my_node/setting2.param.yaml`
- `config/my_node_tool/default.param.yaml`

ambroise-arm:
Ah yes, indeed. I agree that it will not work if nodes have names that partially match. And your suggestion looks good to me. However, it would imply a change touching every single package in autoware.universe. I suppose it would need to be discussed and agreed first. I am not planning to do that in this PR.

I think the change I introduce here is strictly better than the current design, as it would allow to unblock https://github.com/autowarefoundation/autoware.universe/pull/4902 while introducing no regression to the packages that have already implemented a schema file.

isamu-takagi:
@ambroise-arm OK. We will discuss the issue of partially matching names separately.

ambroise-arm:
@miursh @scepter914 Please review when you have time

scepter914 (approved):
LGTM (radar_object_fusion_to_detected_object)

PR #5870, opened 2023-12-14
feat(common): remove independent packages

JianKangEgon:
It's found that the following packages in "universe.common" are not referred by others, so we are wondering if we can remove them during porting.

    glog_component: this package provides the glog (google logging library) feature as a ros2 component library.
    goal_distance_calculator: this node publishes deviation of self-pose from goal pose.
    path_distance_calculator: this node publishes a distance from the closest path point from the self-position to the end point of the path.
    tier4_debug_tools: this package publishers tf2 pose and lateral error.
    time_utils: provided by APEX, transform ROS2 time to standard C+++ time.

**Note**: Confirm the [contribution guidelines](https://autowarefoundation.github.io/autoware-documentation/main/contributing/) before submitting a pull request.

Click the `Preview` tab and select a PR template:

- [Standard change](?expand=1&template=standard-change.md)
- [Small change](?expand=1&template=small-change.md)


isamu-takagi:
These are optional packages and are not included in the default dependencies, but they should be keep available.

isamu-takagi:
However, some packages may not really be used, so we should check it.

JianKangEgon:
> However, some packages may not really be used, so we should check it.

Thanks for the quick reply, glad for further information

h-ohta:
@JianKangEgon 
The path_distance_calculator is used in https://github.com/autowarefoundation/autoware.universe/blob/main/launch/tier4_autoware_api_launch/launch/deprecated_api.launch.xml#L18 
and 
https://github.com/autowarefoundation/autoware.universe/blob/main/launch/tier4_autoware_api_launch/package.xml#L21 .

TakaHoribe:
Thank you for the PR! In my understanding,

- glog_component: used in some packages for debugging
- goal_distance_calculator and path_distance_calculator: this is a debug tool. It may not be used in runtime, but it is used for debugging.
- tier4_debug_tools: this is a debug tool. It does not have a dependency but is used for debugging.
- timer_utils: this is a static library. If this does not depend on anything, maybe we can remove it.

image

JianKangEgon:
Oh thanks for updating, I used the older version of Universe in September. Sure, some useful packages are supposed to be kept.

PR #2339, opened 2022-11-21
feat(ndt_scan_matcher): dynamic map loading

kminoda:
Signed-off-by: kminoda <koji.minoda@tier4.jp>

## Description
I would like to add a dynamic map loading functionality option for `ndt_scan_matcher`.

In order to pass the build CI, we first need to merge
- https://github.com/tier4/ndt_omp/pull/19

Must be merged with 
- https://github.com/autowarefoundation/autoware_launch/pull/151

NOTE:
This PR will replace current `ndt_omp` with `multigrid_ndt_omp`, which is a customized version for dynamic map loading feature. Although most of the interfaces prepared in `ndt_omp` can also be used in `multigrid_ndt_omp`, there are one difference:
- `multigrid_ndt_omp` does not have `neighborhood_search_method` option, since we currently do not implement search method other than kd-tree. 

Thus, do NOTE that **`neighborhood_search_method` parameter in `ndt_scan_matcher` is also removed in this PR**. 

## Related links
- https://github.com/orgs/autowarefoundation/discussions/2696

- https://github.com/orgs/autowarefoundation/discussions/2812

## Tests performed
I tested with a data from Autoware tutorial. The map is divided into 20m grids ([sample-map-rosbag_split.zip](https://github.com/autowarefoundation/autoware.universe/files/10349104/sample-map-rosbag_split.zip)).

- I have confirmed that the Autoware performs the same as the current Autoware when `use_dynamic_map_loading` is set `false`.
- I have confirmed the following when `use_dynamic_map_loading` is set true.
  - localization performance does not change
  - map is dynamically loaded in `ndt_scan_matcher` (see `/localization/pose_estimator/debug/loaded_pointcloud_map`)

In addition, we, TIER IV, have confirmed that the computational resource required for dynamic map loading is low enough for several TIER IV vehicles, based on a several experiments with real hardware.

## Pre-review checklist for the PR author

The PR author **must** check the checkboxes below when creating the PR.

- [x] I've confirmed the [contribution guidelines].
- [x] The PR follows the [pull request guidelines].

## In-review checklist for the PR reviewers

The PR reviewers **must** check the checkboxes below before approval.

- [ ] The PR follows the [pull request guidelines].
- [ ] The PR has been properly tested.
- [ ] The PR has been reviewed by the code owners.

## Post-review checklist for the PR author

The PR author **must** check the checkboxes below before merging.

- [ ] There are no open discussions or they are tracked via tickets.
- [ ] The PR is ready for merge.

[...]

PR #5873, opened 2023-12-14
refactor(ndt_scan_matcher, map_loader): remove map_module

SakodaShintaro:
## Description

This pull request proposes removing map_module. I think this is generally a good change, but I'm open to it being rejected based on discussion.

### Context

Currently, there are two modules related to pcd loading in the `ndt_scan_matcher` package: `map_module` and `map_update_module`. These can be switched by setting `use_dynamic_map_loading` to true/false in the config. (Only one is used at a time).

* map_module : Has a callback for the map point cloud and sets the incoming point cloud to the NDT InputTarget
* map_update_module : Has a client for the pcd_loader service and dynamically loads the map. It also has ekf callback and timer callback.

Functionally, `map_update_module` is upwards compatible, so `map_module` is no longer necessary.

The `map_update_module` was introduced in pull request #2339 about a year ago. It has been working stably since then, so I believe now is a good time to fully migrate to using it instead of `map_module`.

### Changes

Removing `map_module` also requires `enable_differential_load` to always be true in map_loader, so this pull request also includes a fix for that.

If this pull request is acceptable, I will also modify autoware_launch to remove the parameters.

## Related PR
* https://github.com/autowarefoundation/autoware_launch/pull/769

## Tests performed

## Effects on system behavior

map_module will no longer be available. As the function is map_update_module, I don't think there will be any impact.


mitsudome-r:
Does this mean that maps without map_config.yaml won't be supported after the removal of map_loader?

SakodaShintaro:
@mitsudome-r 
Thank you for your comment.
I apologize, I was not aware of the existence of map_config.yaml. After some research, I understand that map_config.yaml is currently not implemented yet and may be used in the future.

I believe that as long as map_config.yaml is properly handled in the map loader, differential_map_loading will run normally. Depending on the implementation at the time, it will be easy to support even without map_config.yaml. 
I don't think the changes in this pull request will have any impact about map_config.yaml.

Motsu-san:
> Does this mean that maps without map_config.yaml won't be supported after the removal of map_module?

@SakodaShintaro Simply you said, "No." didn't you?
@mitsudome-r Could you let us know where you worry? Is any slide needed to show that the changes in this pull request doesn't have any impact about map_config.yaml?

SakodaShintaro:
@Motsu-san 
Yes. This pull request does not affect `map_config.yaml` in any way, I think.

kminoda:
Assuming that @mitsudome-r's concern was about `pointcloud_map_metadata.yaml` and not `map_config.yaml`, the answer is no. Currently, the `differential_map_loading` I/F from `map_loader` will be successfully launched when either of the conditions is met:

1. The pcd is a single file
2. The pcd file consists of multiple files, and an appropriate `pointcloud_map_metadata.yaml` is provided

This does not change even if we disable the ROS 2 topic based I/F between `ndt_scan_matcher` and `map_loader`.

So, to summarize, a map without `pointcloud_map_metadata.yaml` will not be supported only if it consists of multiple PCD files, which is also true even before this PR.

Motsu-san:
@mitsudome-r Does Minoda-san's explanation filled your concerns? Or should we provide an evidence(capture movie?) to prove the following?

> So, to summarize, a map without pointcloud_map_metadata.yaml will not be supported only if it consists of multiple PCD files, which is also true even before this PR.

KYabuuchi:
related discussion: https://github.com/orgs/autowarefoundation/discussions/4062

localization/ndt_scan_matcher/README.md
```diff
@@ -213,20 +213,15 @@ Using the feature, `ndt_scan_matcher` can theoretically handle any large size ma
...
+| `lidar_radius`                        | double | LiDAR radius used for localization (only used for diagnosis) |
 
 ### Enabling the dynamic map loading feature
 
-To use dynamic map loading feature for `ndt_scan_matcher`, you also need to appropriately configure some other settings outside of this node.
-Follow the next two instructions.
-
-1. enable dynamic map loading interface in `pointcloud_map_loader` (by setting `enable_differential_load` to true in the package)
-2. split the PCD files into grids (recommended size: 20[m] x 20[m])
+To use dynamic map loading feature for `ndt_scan_matcher`, you also need to split the PCD files into grids (recommended size: 20[m] x 20[m])
```
KYabuuchi, line 224: "dynamic_loading" is always enabled, so the title of this section seems strange to me.
How about "Notes for dynamic map loading"?

localization/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp
```diff
@@ -410,10 +402,8 @@ void NDTScanMatcher::callback_initial_pose(
         << ". Please check the frame_id in the input topic and ensure it is correct.");
   }
 
-  if (use_dynamic_map_loading_) {
-    std::lock_guard<std::mutex> lock(latest_ekf_position_mtx_);
-    latest_ekf_position_ = initial_pose_msg_ptr->pose.pose.position;
-  }
+  std::lock_guard<std::mutex> lock(latest_ekf_position_mtx_);
+  latest_ekf_position_ = initial_pose_msg_ptr->pose.pose.position;
```
KYabuuchi, line 406: How about adding a scope to emphasize that we are getting a lock?
I think this will make it less likely to be buggy when code is added behind it in the future.
```suggestion
  {
    std::lock_guard<std::mutex> lock(latest_ekf_position_mtx_);
    latest_ekf_position_ = initial_pose_msg_ptr->pose.pose.position;
  }
```

KYabuuchi:
`pointcloud_map` and `input_ekf_odom` are no longer subscribed. So, please remove them in README.
image

image

Actually, input_ekf_odom was erased in [the other PR](https://github.com/autowarefoundation/autoware.universe/pull/5915)

localization/ndt_scan_matcher/config/ndt_scan_matcher.param.yaml
KYabuuchi, line 1: it is necessary to create a PR in the [autoware_launch](https://github.com/autowarefoundation/autoware_launch) repository to remove `use_dynamic_map_loading` and `enable_differential_load` .

localization/ndt_scan_matcher/README.md
SakodaShintaro: Fixed in https://github.com/autowarefoundation/autoware.universe/pull/5873/commits/a74c1ff2b1c5af5578ae94a4a1236c9c9570891c

localization/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp
SakodaShintaro: Fixed in https://github.com/autowarefoundation/autoware.universe/pull/5873/commits/669ce13e6bbc6019d739c86bf5111d076cbaf5a0

localization/ndt_scan_matcher/config/ndt_scan_matcher.param.yaml
SakodaShintaro: created
https://github.com/autowarefoundation/autoware_launch/pull/769

KYabuuchi (approved):
LGTM

mitsudome-r:
@Motsu-san Thanks. I think @kminoda's explanation answers my concern.

PR #5915, opened 2023-12-19
fix(ndt_scan_matcher): fix map_update_module structure

SakodaShintaro:
## Description

In this pull request, the structure of `map_update_module` in `ndt_scan_matcher` has been changed to optimize processing.

### Context
(1) Currently, `map_update_module` continuously outputs ERROR logs due to its inability to subscribe to ekf odometry until `ndt_scan_matcher` estimates the initial pose. Outputting ERROR in this normal sequence is not ideal. To avoid this, a structural change was required because `map_update_module` needs to check the state of `ndt_scan_matcher`.

(2) Upon reviewing the code, I identified another issue: `map_update_module` subscribes to ekf odometry independently of `ndt_scan_matcher`. This redundancy is inefficient. Given that the ekf odometry topic is published at 50Hz, subscribing to it twice could negatively impact CPU usage.

(3) Also, because `map_update_module` creates callback groups in its constructor, it is difficult to understand how many callback groups there were in the entire `ndt_scan_matcher`.

### Changes
* Moved ekf subscriber and timer callbacks to `ndt_scan_matcher_core.cpp`.
* Changed the `map_update_module` to no longer manage the current location and focus only on updating the map.
* Changed the name from main_callback_group to sensor_callback_group.
* Some minor refactoring, such as removing unused arguments.

## Tests performed

The logging_simulator has been verified to run with the same accuracy as before on AWSIM data with GT. It also performs well when dynamic_map_loading is used for divided maps.

## Effects on system behavior

There are no large effects on system behavior. There is a possibility that the CPU load has improved slightly because EKF's double subscription is eliminated.


localization/ndt_scan_matcher/include/ndt_scan_matcher/ndt_scan_matcher_core.hpp
```diff
@@ -199,6 +203,7 @@ class NDTScanMatcher : public rclcpp::Node
 
   std::mutex ndt_ptr_mtx_;
   std::unique_ptr<SmartPoseBuffer> initial_pose_buffer_;
+  std::optional<geometry_msgs::msg::Point> current_position_ = std::nullopt;
```
KYabuuchi, line 206: In the comments, please describe that it is only used when `use_dynamic_map_loading_` is enabled.
Or rename it to something more specific.

localization/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp
```diff
@@ -368,6 +372,28 @@ void NDTScanMatcher::publish_diagnostic()
...
+  if (current_position_ == std::nullopt) {
+    RCLCPP_ERROR_STREAM_THROTTLE(
+      this->get_logger(), *this->get_clock(), 1000,
+      "Cannot find the reference position for map update. Please check if the EKF odometry is "
+      "provided to NDT.");
+    return;
+  }
+  if (!use_dynamic_map_loading_) {
+    return;
+  }
```
KYabuuchi, line 389: It would be better to check `use_dynamic_map_loading_` first.
I don't think it is proper to give the above error message when the variable is false.

localization/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp
```diff
@@ -193,31 +193,36 @@ NDTScanMatcher::NDTScanMatcher()
...
   auto initial_pose_sub_opt = rclcpp::SubscriptionOptions();
   initial_pose_sub_opt.callback_group = initial_pose_callback_group;
-
-  auto main_sub_opt = rclcpp::SubscriptionOptions();
-  main_sub_opt.callback_group = main_callback_group;
-
+  auto sensor_sub_opt = rclcpp::SubscriptionOptions();
+  sensor_sub_opt.callback_group = sensor_callback_group;
+
+  const double map_update_dt = 1.0;
```
KYabuuchi, line 207: You can use constexpr here.

KYabuuchi (commented):
Thanks for your contribution. Please check my following comments. :pray:

localization/ndt_scan_matcher/include/ndt_scan_matcher/ndt_scan_matcher_core.hpp
SakodaShintaro: Fixed in https://github.com/autowarefoundation/autoware.universe/pull/5915/commits/dae9c048805681ea832be4fcc5ba4db495841dad

I also noticed that this variable was causing a data race, so I added std::mutex https://github.com/autowarefoundation/autoware.universe/pull/5915/commits/4313c2e2c3855ef837d1ff4c43692aef0db4083b

localization/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp
SakodaShintaro: It make super sense.
Fixed in https://github.com/autowarefoundation/autoware.universe/pull/5915/commits/fbe915757393a5d0b876a11db64aa453c07fa7d8

localization/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp
SakodaShintaro: Fixed in https://github.com/autowarefoundation/autoware.universe/pull/5915/commits/39ec48782c1c3cce99fbf00b836f96ed48e5b7cc

SakodaShintaro:
Also, the is_activated_ variable is accessed from multiple threads as shown below.
* Read/write by sensor_callback_group thread
* Read in initial_pose_callback_group and timer_callback_group_ threads

I made it atomic. Please let me know if there is any misunderstanding.

KYabuuchi (approved):
LGTM
I think the two additional exclusivity controls make sense. :+1:  (for `is_activated_` and `latest_ekf_position_ `)

PR #5917, opened 2023-12-19
fix(dynamic_avoidance): fix which side to avoid

takayuki5168:
## Description

Changed the logic about which side to avoid
- When the target object velocity is the same direction as the ego
    - Check if the current object position is left or right against the ego's path
- When the target object velocity is NOT the same direction as the ego
    - Check if the object position in the near future is left or right against the ego's path

## Tests performed

psim

## Effects on system behavior

nothing


zjw77889:
When can the dynamic_avoidance  function be used?

takayuki5168:
@zjw77889 
We, TIER IV, will start using the dynamic avoidance feature in two months. Now we are testing and fixing it with a real vehicle.

PR #5951, opened 2023-12-25
fix(ndt_scan_matcher): fixed a lock scope in update_ndt

SakodaShintaro:
## Description

In this pull request, the data race for `ndt_ptr_` has been fixed.

### Context
In the traditional code, within the `update_ndt()` of the `map_update_module`, a copy of the `ndt_ptr_`  was obtained and changes were applied to this copy then it was written back.

However, the class `NormalDistributionsTransform`, which was copied, internally contains a shared_ptr as a member variable. As a result, a simple copy resulted in a shallow copy.

Consequently, when the map update of NDT and the pose estimation by NDT occurred simultaneously, it led to data race issues.

### Changes
Because the shared_ptr is deeply used in the `NormalDistributionsTransform`, implementing a deep copy is difficult.

So the lock scope has been expanded. As a result, it is no longer necessary to make a copy of `ndt_ptr_`, and updates are now applied directly to the original object. Converting the type of the additional points is a heavy process, so this is now pre-calculated outside of the lock scope.

## Tests performed

### Process time
Since the lock scope is larger than previously thought, the time to stop pose estimation by ndt will also be longer.

The time taken to update the map was measured.

Environment : ThinkPad X1 Extreme Gen 4 (11th Gen Intel(R) Core(TM) i7-11800H @ 2.30GHz)

process_time

The first time is adding a new map, so it takes a lot of time to `ndt_ptr_->addTarget`.

From the second time, a differential update is done, and typically, `createVoxelKdtree` is the time-consuming part. If there are many points to add, the `ndt_ptr_->addTarget` also takes time.

In the future, it may be better to recreate `NormalDistributionsTransform` to enable deep copy.

### endurance test
I ran logging_simulator 170 times for about 12 hours and confirmed that ndt_scan_matcher did not crash even once.

## Effects on system behavior

The issue of the node potentially crashing during dynamic map loading is resolved.

However, pose estimation by ndt has a slight delay when the map is differential updated.

### DIFF
--- a/localization/ndt_scan_matcher/src/map_update_module.cpp
+++ b/localization/ndt_scan_matcher/src/map_update_module.cpp
@@ -92,35 +92,36 @@ void MapUpdateModule::update_ndt(
   }
   const auto exe_start_time = std::chrono::system_clock::now();
 
-  NormalDistributionsTransform backup_ndt = *ndt_ptr_;
+  const size_t add_size = maps_to_add.size();
+
+  // Perform heavy processing outside of the lock scope
+  std::vector<pcl::shared_ptr<pcl::PointCloud<PointTarget>>> points_pcl(add_size);
+  for (size_t i = 0; i < add_size; i++) {
+    points_pcl[i] = pcl::make_shared<pcl::PointCloud<PointTarget>>();
+    pcl::fromROSMsg(maps_to_add[i].pointcloud, *points_pcl[i]);
+  }
+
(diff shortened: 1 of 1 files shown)


SakodaShintaro:
I ran logging_simulator 170 times for about 12 hours and confirmed that ndt_scan_matcher did not crash even once.

@YamatoAndo 
Could you review this pull request?

YamatoAndo (approved):

PR #6018, opened 2024-01-05
fix(tier4_localization_launch): add gnss_enabled arg

SakodaShintaro:
## Description

The argument of `gnss_enabled` can now be given from outside of autoware.universe.

Since there may be vehicles that do not have GNSS module, `gnss_enabled` should be able to be set from an external launch.

* https://github.com/autowarefoundation/autoware_launch/pull/771

## Tests performed

* NDT
  * Before : 0.046 [m]
  * After : 0.044 [m]
* artag
  * Before : 0.461 [m]
  * After : 0.461 [m]

## Effects on system behavior

Not applicable.


SakodaShintaro:
As pointed out by @YamatoAndo san, in order to make the handling of arguments in the launch file consistent, it has been changed so that the arg is loaded implicitly.
https://github.com/autowarefoundation/autoware.universe/pull/6018/commits/2e355db8449e9a47382423d4c21c3efee543cb3f

launch/tier4_localization_launch/launch/pose_twist_estimator/pose_twist_estimator.launch.xml
```diff
@@ -2,6 +2,7 @@
 <launch>
   <let name="stop_check_enabled" if="$(eval &quot;'$(var system_run_mode)'=='online'&quot;)" value="true"/>
   <let name="stop_check_enabled" if="$(eval &quot;'$(var system_run_mode)'=='logging_simulation'&quot;)" value="false"/>
+  <arg name="gnss_enabled"/>
```
KYabuuchi, line 5: Hi. Is this buildable now?
In my understanding, if arg is declared here, we need to explicitly pass the arg from the outside launch file (localization.launch.xml).

SakodaShintaro: I thought it was working when I made the change and checked it, but when I checked it now, it didn't work.
I expected that it would work by passing it to implicit, but as I read the page below, that doesn't seem to be the case.
https://design.ros2.org/articles/roslaunch_xml.html

This may not seem like a good idea either, as the arguments are difficult to handle.
For now, the state was changed to draft.

KYabuuchi: ros2 launch is so difficult :cry:

SakodaShintaro:
The issue that `gnss_enabled` cannot be set from `autoware_launch` was resolved by the pull request below, so this pull request is closed.
https://github.com/autowarefoundation/autoware.universe/pull/6144

PR #5846, opened 2023-12-12
feat(pose_estimator_arbiter): add pose_estimator_arbiter to realize multi localizer

KYabuuchi:
## Description

This PR implement a mechanism to launch multiple pose_estimators, and allow for pausing and resuming specific ones according to the situation.

[The package README](https://github.com/KYabuuchi/autoware.universe/tree/feat/multi_localizer/localization/pose_estimator_arbiter) describes the purpose in detail.

## Related links

[[Proposal] Runtime Localization Module Switching for Multi-Pose Estimators #3878](https://github.com/orgs/autowarefoundation/discussions/3878)

## Tests performed

I have confirmed that ndt/yabloc/eagleye/artag is available as before with `logging_simulator.launch.xml` .
Also, I confirmed that by specifying multiple pose_estimators at startup, pose_estimator_arbiter is launched and arbitrates the pose_estimators according to the situation.

### sample data
[sample data (rosbag & map)](https://drive.google.com/file/d/1MxLo1Sw6PdvfkyOYf_9A5dZ9uli1vPvS/view)
The rosbag is simulated data created by [AWSIM](https://tier4.github.io/AWSIM/). The map is an edited version of the original map data published on the AWSIM documentation page to make it suitable for multiple pose_estimators.

### how to launch 

When start a single pose_estimator:
```shell
ros2 launch autoware_launch logging_simulator.launch.xml \
    map_path:=$HOME/MAP vehicle_model:=sample_vehicle sensor_model:=awsim_sensor_kit \
    pose_source:=ndt # or yabloc, eagleye or artag
```

When start multiple pose_estimator:
```bash
ros2 launch autoware_launch logging_simulator.launch.xml \
    map_path:=$HOME/MAP vehicle_model:=sample_vehicle sensor_model:=awsim_sensor_kit \
    pose_source:=ndt_yabloc_eagleye_artag # or ndt_yabloc, eagleye_artag and so on
```

### expected behavior
When start a single pose_estimator, it launches the same nodes as before.

When start multiple pose_estimator, it works as follows.

https://github.com/autowarefoundation/autoware.universe/assets/24854875/62b26fdd-dcf0-4b1c-a1a0-ecd633413196

> [!IMPORTANT]
> We can visualize the debug prints of pose_estimator_arbiter at any time with the following command
> (I recommend enabling logging, because with the default rviz settings, we do not know which pose_estimator is running.)
```
ros2 service call /localization/pose_estimator_arbiter/config_logger \
[...]

PR #6031, opened 2024-01-09
refactor(ar_tag_based_localizer): simplify the launch of artag

SakodaShintaro:
## Description

In the current launch mechanism, input/output topic names are given using remap instead of parameters. This makes launch codes complicated.

Therefore, I changed launch.xml and cpp so that the topic name can be set in the parameter file. As a result, the settings for each environment can be completed in the parameter file, and there is no need to create a launch within `tier4_localization_launch`, reducing the number of layers by one.

## Tests performed

* Before : 0.462 m
* After : 0.460 m

## Effects on system behavior

There are no effects on system behavior.


SakodaShintaro:
This pull request is closed because it conflicts with the implementation of `pose_estimator_arbiter`.

https://github.com/autowarefoundation/autoware.universe/pull/5846

PR #6066, opened 2024-01-12
fix(ekf_localizer): fixed timer in ekf_localizer

SakodaShintaro:
## Description

Changed the handling of timestamps in `ekf_localizer`.

### Context

I found that `this->now()` returns different timestamps even during a single timer callback execution.

Therefore, I think it would be better to obtain the timestamp only once at the beginning of timer callback execution, and use it commonly in one callback. Please let me know if there is any misunderstanding.

## Tests performed

* Before : 0.051 [m]
* After :  0.050 [m]

I plotted the difference between the value of one step and the value of the next step regarding x, y, z and timestamp of Pose output by EKF.

* Before
localization__pose_twist_fusion_filter__pose_differential
* After
localization__pose_twist_fusion_filter__pose_differential

## Effects on system behavior

It seems that the topic timestamps output by EKF will be slightly more stable.

### DIFF
--- a/localization/ekf_localizer/include/ekf_localizer/ekf_localizer.hpp
+++ b/localization/ekf_localizer/include/ekf_localizer/ekf_localizer.hpp
@@ -199,7 +199,7 @@ private:
   /**
    * @brief update predict frequency
    */
-  void updatePredictFrequency();
+  void updatePredictFrequency(const rclcpp::Time & current_time);
 
   /**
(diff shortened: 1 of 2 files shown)


TaikiYamada4 (approved):
Seems to works fine in my environment too.
LGTM!

YamatoAndo (approved):

Issue #5976, opened 2023-12-27
Misc backward driving issues in `motion_velocity_smoother`


VRichardJP:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I've agreed with the maintainers that I can plan this task.

### Description

As I investigate freespace planners, I observe `motion_velocity_smoother` has many issues handling trajectories going backwards. The trajectories are assumed to go forward at many places in the and many calculation are plain wrong when going backward.

I would like to list all issues here.

### Purpose

Find and fix logic when backward trajectories are used.

### Possible approaches

N/A

### Definition of done

N/A

VRichardJP:
`isEngageStatus` does not return `true` when engaging backward (i.e. `target_vel < node_param_.engage_velocity`)

https://github.com/autowarefoundation/autoware.universe/blob/a04f825bc542618fb86a347b45ae90162c1ff17e/planning/motion_velocity_smoother/src/motion_velocity_smoother_node.cpp#L1038-L1043

VRichardJP:
`MotionVelocitySmootherNode::calcExternalVelocityLimit()` assumes forward drive:

https://github.com/autowarefoundation/autoware.universe/blob/a04f825bc542618fb86a347b45ae90162c1ff17e/planning/motion_velocity_smoother/src/motion_velocity_smoother_node.cpp#L362-L363

https://github.com/autowarefoundation/autoware.universe/blob/a04f825bc542618fb86a347b45ae90162c1ff17e/planning/motion_velocity_smoother/src/motion_velocity_smoother_node.cpp#L377-L384

when `v0 < 0.0`, `max_velocity_with_deceleration_` is set to negative value, but it is supposedly absolute velocity.

VRichardJP:
`applyExternalVelocityLimit()` logic looks strange on backward path:

- The function is called by `calcTrajectoryVelocity()` after input velocity has been "flipped" in `onCurrentTrajectory`:

https://github.com/autowarefoundation/autoware.universe/blob/a04f825bc542618fb86a347b45ae90162c1ff17e/planning/motion_velocity_smoother/src/motion_velocity_smoother_node.cpp#L468-L475

[...]

PR #6079, opened 2024-01-15
fix(motion_velocity_smoother): improve backward trajectory handling

VRichardJP:
## Description

Fixes https://github.com/autowarefoundation/autoware.universe/issues/5976

Current support for backward trajectories is fairly limited and sometimes just happen to work "by accident" (e.g. some calculated data is wrong but mostly unused). In particular, I have found many backward driving issues with the `motion_velocity_smoother` node while implementing a new freespace planner. 

Most functions in `motion_velocity_smoother` have been implemented with forward trajectories in mind and don't apply to backward trajectories. The "trick" currently used to handle backward trajectories is to "flip" the trajectory points velocity before processing it, to do as if the vehicle was driving forward and not backward, and then flip the output back. I think the idea is valid, however it is not properly implemented: for example [it causes steering values to be incorrect on backward trajectories](https://github.com/autowarefoundation/autoware.universe/issues/5976#issuecomment-1870862114).

The "proper" way to implement the flipping trick is not just to flip the velocity but to flip everything: trajectory pose orientation, velocity, steering... but also the ego pose and velocity when we need to place ego vehicle on the flipped trajectory.

Note that the way the logic is implemented at the moment (local trajectory but global `is_reverse_` parameter) is error prone and difficult to maintain. It would be worth packing trajectory data and ego state inside a single ("flippable") struct so that sub functions could just assume everything goes forward.

## Related links

https://github.com/autowarefoundation/autoware.universe/issues/5976

## Tests performed

With the freespace planner in the planning simulator:

- curved forward/backward trajectories values for longitudinal_velocity_mps, acceleration_mps2, front_wheel_angle_rad have correct signs.
- same for single point turn trajectories
- external speed limit on above trajectories works fine.
- lanelet based driving still looks ok

## Interface changes

None.

## Effects on system behavior

On all trajectories:
- `motion_velocity_smoother` now ignores trajectories mixing forward/backward driving. AFAIK this is not supposed to be supported and would most likely give strange result. But now it is checked.

On backward trajectory:
- `motion_velocity_smoother` output trajectory points values are now signed properly

Except for these, there should be no other visible behavior change.

### DIFF
--- a/planning/motion_velocity_smoother/src/motion_velocity_smoother_node.cpp
+++ b/planning/motion_velocity_smoother/src/motion_velocity_smoother_node.cpp
@@ -471,17 +471,23 @@ void MotionVelocitySmootherNode::onCurrentTrajectory(const Trajectory::ConstShar
   external_velocity_limit_ptr_ = nullptr;
 
   // Trajectory must be either all forward or all backward.
-  bool is_forward_only = std::all_of(input_points.begin(), input_points.end(),
-    [](const auto& tp) { return tp.longitudinal_velocity_mps >= 0.0; });
-  bool is_backward_only = std::all_of(input_points.begin(), input_points.end(),
-    [](const auto& tp) { return tp.longitudinal_velocity_mps <= 0.0; });
+  bool is_forward_only = std::all_of(input_points.begin(), input_points.end(), [](const auto & tp) {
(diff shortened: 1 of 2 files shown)


xmfcx:
@takayuki5168 -san, could you assign someone from TIER IV for this issue? @VRichardJP is probably busy and this seems like an important feature.

PR #6138, opened 2024-01-22
feat(point cloud preprocessor): add pickup based voxel grid filter

yukkysaito:

### DIFF
--- a/sensing/pointcloud_preprocessor/include/pointcloud_preprocessor/downsample_filter/pickup_based_voxel_grid_downsample_filter.hpp
+++ b/sensing/pointcloud_preprocessor/include/pointcloud_preprocessor/downsample_filter/pickup_based_voxel_grid_downsample_filter.hpp
@@ -1,0 +1,54 @@
+// Copyright 2023 TIER IV, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef POINTCLOUD_PREPROCESSOR__DOWNSAMPLE_FILTER__PICKUP_BASED_VOXEL_GRID_DOWNSAMPLE_FILTER_NODELET_HPP_
+#define POINTCLOUD_PREPROCESSOR__DOWNSAMPLE_FILTER__PICKUP_BASED_VOXEL_GRID_DOWNSAMPLE_FILTER_NODELET_HPP_
+
+#include "pointcloud_preprocessor/filter.hpp"
+
+#include <Eigen/Core>
+
+#include <pcl_conversions/pcl_conversions.h>
+#include <sensor_msgs/msg/point_cloud2.h>
+
+#include <algorithm>
+#include <cmath>
+#include <vector>
+
+namespace pointcloud_preprocessor
+{
+class PickupBasedVoxelGridDownsampleFilterComponent : public pointcloud_preprocessor::Filter
+{
+protected:
+  void filter(
+    const PointCloud2ConstPtr & input, const IndicesPtr & indices, PointCloud2 & output) override;
+
+private:
+  float voxel_size_x_;
+  float voxel_size_y_;
+  float voxel_size_z_;
+
+  /** \brief Parameter service callback result : needed to be hold */
+  OnSetParametersCallbackHandle::SharedPtr set_param_res_;
+
+  /** \brief Parameter service callback */
+  rcl_interfaces::msg::SetParametersResult paramCallback(const std::vector<rclcpp::Parameter> & p);
+
+public:
+  PCL_MAKE_ALIGNED_OPERATOR_NEW
+  explicit PickupBasedVoxelGridDownsampleFilterComponent(const rclcpp::NodeOptions & options);
+};
+}  // namespace pointcloud_preprocessor
+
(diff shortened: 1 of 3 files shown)


sensing/pointcloud_preprocessor/include/pointcloud_preprocessor/downsample_filter/pickup_based_voxel_grid_downsample_filter.hpp
```diff
@@ -0,0 +1,62 @@
...
+  /** \brief Parameter service callback */
+  rcl_interfaces::msg::SetParametersResult paramCallback(const std::vector<rclcpp::Parameter> & p);
+
+public:
+  PCL_MAKE_ALIGNED_OPERATOR_NEW
+  explicit PickupBasedVoxelGridDownsampleFilterComponent(const rclcpp::NodeOptions & options);
+};
+}  // namespace pointcloud_preprocessor
+
+#endif  // POINTCLOUD_PREPROCESSOR__DOWNSAMPLE_FILTER__PICKUP_BASED_VOXEL_GRID_DOWNSAMPLE_FILTER_HPP_
```
mitsudome-r, line 62: NOTE: 
It seems like pre-commit ci is failing because the include guard is very long and exceeds 100 character limitations of cpp-lint.

In order to fix this, you need to add `// NO_LINT` to ignore cpp-lint, but this also fails because clang-format, which runs before cpp-lint, will reformat the comment to split the lines between `// NO_LINT` and the include guard comment.  Therefore, you also have to add `// clang-format off` to turn off clang-format for this line as well.


```suggestion
// clang-format off
#endif  // POINTCLOUD_PREPROCESSOR__DOWNSAMPLE_FILTER__PICKUP_BASED_VOXEL_GRID_DOWNSAMPLE_FILTER_HPP_  // NOLINT
// clang-format on
```

sensing/pointcloud_preprocessor/src/downsample_filter/pickup_based_voxel_grid_downsample_filter.cpp
```diff
@@ -0,0 +1,158 @@
...
+ */
+struct VoxelKeyHash
+{
+  std::size_t operator()(const std::array<int, 3> & k) const
+  {
+    // Primes based on the following paper: 'Investigating the Use of Primes in Hashing for
+    // Volumetric Data'.
+    return (k[0] * 73856093 ^ k[1] * 19349663 ^ k[2] * 83492791);
+    // return ((1 << 16) - 1) & (k[0] * 73856093 ^ k[1] * 19349663 ^ k[2] * 83492791);
+    // return ((1 << 20) - 1) & (k[0] * 73856093 ^ k[1] * 19349663 ^ k[2] * 83492791);
```
KYabuuchi, line 31: Are these two lines necessary? 
If they are necessary, please add a comment to explain the purpose. If not, remove them. :pray:

badai-nguyen (approved):
I compared with[ Approximate_downsample_filter](https://github.com/autowarefoundation/autoware.universe/blob/010523f0d917a3a467de112bf6c82cdf8c57f5b3/sensing/pointcloud_preprocessor/src/downsample_filter/approximate_downsample_filter_nodelet.cpp#L77-L94) and intutively confirmed the output pointcloud number does not changed and processing time is ~ 4times faster.

|     Approximate_downsample_filter | Pickup_based_voxel_grid_downsample_filter |
| -- | -- | 
|||
|Approximation_voxel_grid_2| pickup_based_voxel_grid_downsample_filter_2|

LGTM

drwnz (commented):
This is a nice idea! I have a couple of concerns:
- Please correct me if this is misunderstood, but my understanding is that the filter relies on integer truncation and only updates the hash key for new entries outside the truncated volume. This is potentially problematic, as it will likely introduce a bias in the filter - the chosen representative point is always the first one that appears in the scan within the truncation volume. Although more computationally complex, a median filter would seem like a more rigorous approach and still maintain a real point rather than centroid.
- The name doesn't really describe what this filter does. It seems to be an integer-truncated XOR hashing filter, not a voxel grid filter (and the meaning of pickup is unclear)

sensing/pointcloud_preprocessor/src/downsample_filter/pickup_based_voxel_grid_downsample_filter.cpp
```diff
@@ -0,0 +1,158 @@
...
+namespace pointcloud_preprocessor
+{
+PickupBasedVoxelGridDownsampleFilterComponent::PickupBasedVoxelGridDownsampleFilterComponent(
+  const rclcpp::NodeOptions & options)
+: Filter("PickupBasedVoxelGridDownsampleFilterComponent", options)
+{
+  // Initialization of voxel sizes from parameters
+  voxel_size_x_ = static_cast<float>(declare_parameter("voxel_size_x", 1.0));
+  voxel_size_y_ = static_cast<float>(declare_parameter("voxel_size_y", 1.0));
+  voxel_size_z_ = static_cast<float>(declare_parameter("voxel_size_z", 1.0));
```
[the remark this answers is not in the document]
yukkysaito: Thank you very much. I need to use float for tuning the computation cost, and I need to cast somewhere. :pray:

yukkysaito:
Thanks :+1: 

>Please correct me if this is misunderstood, but my understanding is that the filter relies on integer truncation and only updates the hash key for new entries outside the truncated volume. This is potentially problematic, as it will likely introduce a bias in the filter - the chosen representative point is always the first one that appears in the scan within the truncation volume. Although more computationally complex, a median filter would seem like a more rigorous approach and still maintain a real point rather than centroid.

The part that is changed to an integer value for each xyz is used to indicate the index of the voxel. It is then hashed for speed.

>The name doesn't really describe what this filter does. It seems to be an integer-truncated XOR hashing filter (and the meaning of pickup is unclear)

Do you have any suggestions? :pray:

kminoda:
Memo
Need an approval from maintainer

yukkysaito:
@kminoda Performance tuning was performed. Approximately 3x speedup on large voxels (200,000 voxels) (45msec -> 15msec)

yukkysaito:
BTW, At 200,000 voxels, the existing ones introduced in autoware cause integer overflow and do not work.

PR #6149, opened 2024-01-23
feat(start_planner): keep distance back static objects from start pose

kyoichi-sugahara:
## Description

Following launcher PR should be merged first
https://github.com/autowarefoundation/autoware_launch/pull/815

keep enough distance between back static object and start pose.
This distance is necessary for avoiding stack which is caused by other module's stop point insertion.

NOTE: margin against front objects can be smaller than margin against back objects, because start_planner expect the other module like avoidance to change the trajectory. 

image

## Tests performed

- [x] [PASS TIERIV INTERNAL SCENARIOS](https://evaluation.tier4.jp/evaluation/reports/88b12251-366f-562d-b77d-cb3b165da607?project_id=prd_jt)

## Effects on system behavior

Not applicable.


kyoichi-sugahara:
@kosuke55 
confirmed no degression with evaluator's result.
Could you review the PR?

planning/behavior_path_start_planner_module/src/start_planner_module.cpp
```diff
@@ -964,13 +964,13 @@ std::vector<Pose> StartPlannerModule::searchPullOutStartPoseCandidates(
   const double backward_path_length =
     planner_data_->parameters.backward_path_length + parameters_->max_back_distance;
 
-  const auto stop_objects_in_pull_out_lanes = filterStopObjectsInPullOutLanes(
-    pull_out_lanes, start_pose.position, parameters_->th_moving_object_velocity,
-    backward_path_length, std::numeric_limits<double>::max());
+  const auto back_stop_objects_in_pull_out_lanes = filterStopObjectsInPullOutLanes(
+    pull_out_lanes, start_pose.position, parameters_->th_moving_object_velocity, 0,
+    backward_path_length);
```
kosuke55, line 969: the previous code is mistakes?

planning/behavior_path_start_planner_module/src/start_planner_module.cpp
```diff
@@ -1005,8 +1005,8 @@ std::vector<Pose> StartPlannerModule::searchPullOutStartPoseCandidates(
     }
 
     if (utils::checkCollisionBetweenFootprintAndObjects(
-          local_vehicle_footprint, *backed_pose, stop_objects_in_pull_out_lanes,
-          parameters_->collision_check_margins.back())) {
+          local_vehicle_footprint, *backed_pose, back_stop_objects_in_pull_out_lanes,
+          parameters_->back_objects_collision_check_margin)) {
```
kosuke55, line 1009: use this margin for all directions not only backward?

planning/behavior_path_start_planner_module/src/start_planner_module.cpp
kyoichi-sugahara: yes the order of arguments was opossite...

planning/behavior_path_start_planner_module/src/start_planner_module.cpp
kyoichi-sugahara: sorry I didn't understand this part
> I think it is better to expand margin only backward and keep the other directions(forward and lateral)

But your suggestion is calculate only longitudinal length, correct?
If so, I totally agree!

kosuke55: yes, not forward but only backward longitudinal margin should be `back_objects_collision_check_margin`

kosuke55: these functions may be useful 
https://github.com/tier4/autoware.universe/blob/awf-latest/planning/behavior_path_planner_common/src/utils/utils.cpp#L204-L248

kyoichi-sugahara: I understand and totally agree with your proposal!!
And thank you for sharing useful code.

kosuke55: thanks!
if it is need to check lateral collision, we have two ways. 
thease are pseudo codes
1.
```
utils::checkCollisionBetweenFootprintAndObjects(
          local_vehicle_footprint, *backed_pose, back_stop_objects_in_pull_out_lanes,
          parameters_->back_objects_collision_check_margin))

calcLongitudinalDistanceFromEgoToObject(...)
```

2.
```
backward_expanded_local_vehicle_footprint = expandPolygon(local_vehicle_footprint, ...)
utils::checkCollisionBetweenFootprintAndObjects(
          backward_expanded_local_vehicle_footprint, *backed_pose, back_stop_objects_in_pull_out_lanes,
          parameters_->back_objects_collision_check_margin))

```

kyoichi-sugahara: I thought the change in https://github.com/autowarefoundation/autoware.universe/pull/6149/commits/3d2df33251667536f4e2d4bd9a8887c458f09a8e is expected?

kosuke55: The change skips lateral collision check but is it ok?

kyoichi-sugahara: sorry i missed to check following message
> if it is need to check lateral collision, we have two ways.
thease are pseudo codes

I'm assuming `back_stop_objects_in_pull_out_lanes` are located in same lane correct?
Even in this case, lateral collision should be checked with `back_objects_collision_check_margin`?

planning/behavior_path_start_planner_module/src/util.cpp
```diff
@@ -104,4 +105,35 @@ lanelet::ConstLanelets getPullOutLanes(
...
+    const auto obj_polygon = tier4_autoware_utils::toPolygon2d(obj);
+    for (const auto & obj_outer_point : obj_polygon.outer()) {
+      const auto obj_pose_arc_length = getArcLengthForPoint(lanelets, obj_outer_point);
+      for (const auto & vehicle_corner_point : vehicle_footprint) {
+        const auto vehicle_pose_arc_length = getArcLengthForPoint(lanelets, vehicle_corner_point);
+        const double distance = std::abs(obj_pose_arc_length - vehicle_pose_arc_length);
+        min_distance = std::min(min_distance, distance);
+      }
+    }
+  }
```
kosuke55, line 125: nits: 
this could be heavy.
I think there are better ways to make it faster, such as extracting the closest object, the closest point, etc.

planning/behavior_path_start_planner_module/src/start_planner_module.cpp
kosuke55: memo:
we talked face to face.
use hard margin for all distance, and check arc length for backward

planning/behavior_path_start_planner_module/src/util.cpp
kosuke55: pseudo code
```
backward_objects = extraceBackwardObjects(lanelets, ego_pose, objectse); // extract objects whose arc length is less than ego's one
closest_backward_objects = getClosestObjects(backward_objects); // get object whose arc length is max
closet_object_corner_point = getClosestCornerPoint(closest_backward_objects); // get cornaner whose arc length is max
most_backward_ego_corner_point = getMostBackwardEgoCornerPoint(ego_pose, vehicle_footprint); // get corner whose arc length is min
distance = std::abs(most_backward_ego_corner_point - closet_object_corner_point); // get arc length between two points
```

kyoichi-sugahara: Thanks!
I will consider improving the efficiency

kyoichi-sugahara: In my current implementation, distance `a` is calculated but distance `b` should be calculated, correct?
If so, the method of extracting objects for back distance against back object should be changed. In this sketch, only object [2] should be extracted. 
image

And for caring lateral collision implemented with the following implementetion. objects[0,1,2] should be extracted :thinking:   
```
      (utils::checkCollisionBetweenFootprintAndObjects(
        local_vehicle_footprint, *backed_pose, back_stop_objects_in_pull_out_lanes,
        parameters_->objects_collision_check_margins.back())))
```

kosuke55: > I will consider improving the efficiency
thanks! but it's ok if it takes time to improve because the check target objects may be not so many 

> In my current implementation, distance a is calculated but distance b should be calculated, correct?
no, I think distance a should be calculated,

> And for caring lateral collision implemented with the following implementetion. objects[0,1,2] should be extracted 🤔
I agree

kyoichi-sugahara: I believe that some ingenuity is necessary in extracting objects to ensure `back_distance` in the lane coordinate system, especially in such unique situations. If we were to calculate `a` distance, I'm concerned that it might become impossible for backward driving.
image

red area represents `pull_out_lanes`

kosuke55: yes, right. how about filter by lateral distance?

> backward_objects = extraceBackwardObjects(lanelets, ego_pose, objectse); // extract objects whose arc length is less than ego's one

pseudo
```
extraceBackwardObjects(const lanelet::ConstLanelets & lanelets, const geometry_msgs::msg::Pose & ego_pose, const PredictedObjects & objects, const double lateral_threshold){
  const double ego_arc_coords  = getArcCoordinates(lanelets, ego_pose.position).length;
  for (const auto & object : objects) {
    const double obj_arc_coods = getArcCoordinates(lanelets, object.pose.position);
    if(std::abs(ego_arc_coords.distance -obj_arc_coods.distance) > lateral_threshold){
      continue;
    }
    if (obj_arc_coods.length < ego_arc_coords.length) {
      backward_objects.push_back(object);
    }
  }
  return backward_objects;
}
```

kyoichi-sugahara: how about creating polygon along the `start_pose_candidates_path`?
https://github.com/autowarefoundation/autoware.universe/blob/1121f3b2a5a298ac474835522b97e84d9301dd9a/planning/behavior_path_start_planner_module/src/start_planner_module.cpp#L885

And extracts closest object overlaps with the polygon
image
a: `objects_collision_check_margins.back()`
b: `back_objects_collision_check_margin`

It's heavier but checking lateral distance of object pose is not considering the size of the object?

kosuke55: looks good!
the attention area is  along the `start_pose_candidates_path` not nagative x-coords direction, right?

kosuke55: which pattern did you mean?
image

kosuke55: and, better to check just side to ego?

image

kosuke55: (sorry for my mistake, this pseudo code did not consider footprint)

https://github.com/autowarefoundation/autoware.universe/pull/6149#discussion_r1485610920

```
    const auto obj_arc_coods = getArcCoordinates(lanelets, object.pose.position);
    if(std::abs(ego_arc_coords.distance -obj_arc_coods.distance) > lateral_threshold){
      continue;
    }
```
not just check `object.pose.position` but need to check 4 footprint points by like `std::any`

xmfcx:
@kyoichi-sugahara @kosuke55 could you rebase and work on this PR?

xmfcx:
@kyoichi-sugahara @kosuke55 I will close this issue since there was no response. Feel free to reopen if necessary.

PR #6362, opened 2024-02-08
feat(mission_planner_rviz_plugin): create mission planner tool

isamu-takagi:
## Description

Add rviz plugin to set/clear the mrm route.
Screenshot from 2024-02-08 22-11-18

## Related links

- Related to https://github.com/autowarefoundation/autoware.universe/issues/6319
- Depends on https://github.com/tier4/tier4_autoware_msgs/pull/111
- Depends on https://github.com/autowarefoundation/autoware.universe/pull/6363

## Tests performed

Same as  https://github.com/autowarefoundation/autoware.universe/pull/6363.

https://evaluation.tier4.jp/evaluation/reports/7d8458ad-5132-5f28-b0b2-1c117bc1b7e6?project_id=prd_jt

## Notes for reviewers

None

## Interface changes

None

## Effects on system behavior

None

## Pre-review checklist for the PR author

The PR author **must** check the checkboxes below when creating the PR.

- [x] I've confirmed the [contribution guidelines].
- [x] The PR follows the [pull request guidelines].

## In-review checklist for the PR reviewers

The PR reviewers **must** check the checkboxes below before approval.

- [ ] The PR follows the [pull request guidelines].
- [ ] The PR has been properly tested.
- [ ] The PR has been reviewed by the code owners.

## Post-review checklist for the PR author

The PR author **must** check the checkboxes below before merging.

- [ ] There are no open discussions or they are tracked via tickets.
- [ ] The PR is ready for merge.

After all checkboxes are checked, anyone who has write access can merge the PR.

[contribution guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/
[pull request guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/pull-request-guidelines/

Issue #6319, opened 2024-02-05
Refactor route interface


isamu-takagi:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I've agreed with the maintainers that I can plan this task.

### Description

With the introduction of pull-over MRM, routes specified in the API may be changed for MRM. This may cause applications using the API to confuse pull-over MRM completion with destination arrival, resulting in unintended behavior. Therefore, it is necessary to distinguish between normal routes and MRM routes.

### Purpose

Manage normal routes and MRM routes separately.

### Possible approaches

Add route_selector node to manage the normal and MRM routes switching. The mission_planner can now be treated as a simple route change without distinguishing between normal and MRM routes.

- Normal and MRM routes have their own route and state.
- Add an internal route message/service for the route_selector to specify route ID.
- Add some states for mission_planner and route_selector communication.
- Internally `change_route` is merged into `set_route`.

The diagram below outlines the changes (red: autoware_adapi_v1_msgs, blue: autoware_planning_msgs).

![routing-interface-names](https://github.com/autowarefoundation/autoware.universe/assets/43976882/dea14560-e794-444c-9275-e3ec929b7213)

### Definition of done

- [x] Add internal route message.
- [x] Add route_selector and change mission_planner
- [x] Update API node.

PR #6363, opened 2024-02-08
feat(mission_planner)!: introduce route_selector node

isamu-takagi:
## Description

Introduce the route_selector node to the mission_planner package. The route_selector node manages main/mrm routes and makes requests to the mission_planner node. Therefore, now the mission_planner node only needs to plan the current route.

## Related links

- Related to https://github.com/autowarefoundation/autoware.universe/issues/6319
- Related to https://github.com/autowarefoundation/autoware.universe/pull/6362
- Related to https://github.com/autowarefoundation/autoware.universe/pull/6364
- Depends on https://github.com/tier4/tier4_autoware_msgs/pull/111

## Tests performed

1. Merge the dependent PRs.
2. Launch planning simulator.
3. Add MrmGoalTool and RouteSelectorPanel to rviz.
4. Set topic name from rviz tool property panel.
    - 2D Goal Pose: `/rviz/route_selector/main/goal`
    - Mrm Goal Pose: `/rviz/route_selector/mrm/goal`
5. Check if the normal route can be set (use 2D Goal Pose)
6. Check if the mrm route can be set (use Mrm Goal Pose)
7. Check if it can return to normal route (push clear button of MRM state)
8. Check if the mrm route can be set again (use Mrm Goal Pose)
9. Check if the normal route setting is rejected during MRM (use 2D Goal Pose)

https://evaluation.tier4.jp/evaluation/reports/7d8458ad-5132-5f28-b0b2-1c117bc1b7e6?project_id=prd_jt

## Notes for reviewers

https://github.com/autowarefoundation/autoware.universe/assets/43976882/b841c383-f30b-41fd-9d48-62e7facdeced

## Interface changes

Add:
- /planning/mission_planning/route_selector/main/route
- /planning/mission_planning/route_selector/main/state
- /planning/mission_planning/route_selector/mrm/route
- /planning/mission_planning/route_selector/mrm/state
- /planning/mission_planning/mission_planner/clear_route
- /planning/mission_planning/mission_planner/set_lanelet_route
- /planning/mission_planning/mission_planner/set_waypoint_route
- /planning/mission_planning/route_selector/main/clear_route
- /planning/mission_planning/route_selector/main/set_lanelet_route
- /planning/mission_planning/route_selector/main/set_waypoint_route
- /planning/mission_planning/route_selector/mrm/clear_route
- /planning/mission_planning/route_selector/mrm/set_lanelet_route
- /planning/mission_planning/route_selector/mrm/set_waypoint_route

Type change:
- /planning/mission_planning/state

## Effects on system behavior

The MRM route will no longer be reflected in AD API.


planning/mission_planner/src/mission_planner/route_selector.cpp
```diff
@@ -0,0 +1,218 @@
...
+  if (auto request = std::get_if<RoutePointRequest>(&main_request_)) {
+    // NOTE: Clear the waypoint to avoid returning. Remove this once resuming is supported.
+    (**request).waypoints.clear();
+    res->status = service_utils::sync_call(cli_set_waypoint_route_, *request);
+    return;
+  }
+  if (auto request = std::get_if<RouteRequest>(&main_request_)) {
+    res->status = service_utils::sync_call(cli_set_lanelet_route_, *request);
+    return;
+  }
```
kosuke55, line 179: In the original code, even in the case of RouteRequest, if the reroute fails, it is retried using only the goal.

planning/mission_planner/src/mission_planner/mission_planner.hpp
```diff
@@ -68,86 +70,63 @@ class MissionPlanner : public rclcpp::Node
...
+    const Header & header, const std::vector<Pose> & waypoints, const Pose & goal_pose,
+    const UUID & uuid, const bool allow_goal_modification);
 
-  std::shared_ptr<LaneletRoute> normal_route_{nullptr};
-  std::shared_ptr<LaneletRoute> mrm_route_{nullptr};
+  rclcpp::TimerBase::SharedPtr data_check_timer_;
+  void check_initialization();
+
+  double reroute_time_threshold_;
+  double minimum_reroute_length_;
```
kosuke55, line 128: better to set default value (same for other parts.
because the default value for double is not always 0.0

kosuke55:
```suggestion
  double reroute_time_threshold_{10.0};
  double minimum_reroute_length_{30.0};
```

these are removed default values

planning/mission_planner/src/mission_planner/route_selector.cpp
```diff
@@ -0,0 +1,218 @@
...
+  mrm_operating_ = true;
+  res->status = service_utils::sync_call(cli_set_waypoint_route_, req);
+}
+
+void RouteSelector::on_set_lanelet_route_mrm(
+  SetLaneletRoute::Request::SharedPtr req, SetLaneletRoute::Response::SharedPtr res)
+{
+  mrm_operating_ = true;
+  res->status = service_utils::sync_call(cli_set_lanelet_route_, req);
+}
```
kosuke55, line 195: `mrm_operating_` should be set to `true` only after rerouting success.
Otherwise, it appears that state will also be set if mrm reroute fails. (0:59-1:00 in your demo video)

image

image

planning/mission_planner/src/mission_planner/service_utils.hpp
```diff
@@ -0,0 +1,69 @@
...
+  };
+}
+
+template <class T, class Req>
+ResponseStatus sync_call(T & client, Req req)
+{
+  if (!client->service_is_ready()) {
+    throw ServiceUnready(client->get_service_name());
+  }
+  auto future = client->async_send_request(req);
```
kosuke55, line 63:
```suggestion
  const auto future = client->async_send_request(req);
```

planning/mission_planner/src/mission_planner/mission_planner.cpp
```diff
@@ -174,557 +134,262 @@ void MissionPlanner::on_map(const HADMapBin::ConstSharedPtr msg)
...
 
-// NOTE: The route interface should be mutually exclusive by callback group.
-void MissionPlanner::on_change_route_points(
-  const SetRoutePoints::Service::Request::SharedPtr req,
-  const SetRoutePoints::Service::Response::SharedPtr res)
+LaneletRoute MissionPlanner::create_route(const PoseWithUuidStamped & msg)
 {
-  using ResponseCode = autoware_adapi_v1_msgs::srv::SetRoutePoints::Response;
+  const auto & header = msg.header;
+  const auto & waypoints = std::vector<Pose>();
```
kosuke55, line 353: no need to use & for temporary object?

planning/mission_planner/src/mission_planner/mission_planner.hpp
isamu-takagi: These values are not used because the parameters are always assigned in the constructor.

kosuke55: Yes! It looks OK this time and no need to modify this

We had recently made sure to initialize as well, because death could happen if you forget to pass in parameters.
FYI: https://star4.slack.com/archives/C03QW0GU6P7/p1694699548349669?thread_ts=1694534749.153109&cid=C03QW0GU6P7 (It is bad that the parameter is not set.)

planning/mission_planner/src/mission_planner/route_selector.cpp
isamu-takagi: Thank you. I fixed it to call `set_waypoint_route` with only goal in both cases. https://github.com/autowarefoundation/autoware.universe/pull/6363/commits/09a753374ce85dc786a242fccd385eff0df88bb3

planning/mission_planner/src/mission_planner/service_utils.hpp
isamu-takagi: The `future.get` is not const function. So adding the const will cause a compilation error.

planning/mission_planner/src/mission_planner/mission_planner.hpp
isamu-takagi: The problem is that the parameter is simply not assigned, and setting the initial value here is not the correct solution. I think it is better to avoid introducing unnecessary measures because bugs may be harder to find.

It's a good thing that nodes die and anomalies are easy to detect. I think it's more risky to keep working with unintended parameters.

planning/mission_planner/src/mission_planner/route_selector.cpp
kosuke55: thanks and sorry.

In the original code, 
1. Check if vehicle can return to the original normal root  https://github.com/autowarefoundation/autoware.universe/blob/main/planning/mission_planner/src/mission_planner/mission_planner.cpp#L510-L516
2. If not, check with a new route using only goal. https://github.com/autowarefoundation/autoware.universe/blob/main/planning/mission_planner/src/mission_planner/mission_planner.cpp#L518-L530

kosuke55: same for WaypointRequest,
1. check original normal route
2. check new route with only goal

But the route_selector does not keep the original normal route, so it may be a bit difficult.

planning/mission_planner/src/mission_planner/route_selector.cpp
```diff
@@ -0,0 +1,234 @@
...
+    const auto r = std::make_shared<SetWaypointRoute::Request>();
+    r->header = request->header;
+    r->goal_pose = request->goal_pose;
+    r->uuid = request->uuid;
+    r->allow_modification = request->allow_modification;
+    return r;
+  };
+
+  mrm_operating_ = false;
+  mrm_.change_state(now(), RouteState::UNSET);
```
kosuke55, line 173: clear route mrm sometimes fails, so better to set this after reroute wit normal route success


in the previous code, keep mrm route when reoute fails
https://github.com/autowarefoundation/autoware.universe/blob/main/planning/mission_planner/src/mission_planner/mission_planner.cpp#L525-L529

planning/mission_planner/src/mission_planner/mission_planner.cpp
isamu-takagi: Fixed with https://github.com/autowarefoundation/autoware.universe/pull/6363/commits/4b5cf34f1b170b0ea64af00cd2559d5e27634ffe.

planning/mission_planner/src/mission_planner/route_selector.cpp
isamu-takagi: Fixed so that it first attempts with the planned lanelet segments and then replan with the goal. https://github.com/autowarefoundation/autoware.universe/pull/6363/commits/9fe0af236f554ee9c47b39cbd3af8443854248c3

planning/mission_planner/src/mission_planner/route_selector.cpp
isamu-takagi: Fix to clear mrm mode if the route clear/change is succeeded. https://github.com/autowarefoundation/autoware.universe/pull/6363/commits/4951e1b054dd8a3b29ff2cccf89e02305cff2b22#diff-b8e18a7cd1279c2c60737e4e006028710a1b6f18b055a63d1070c0f3f8a659c1R194-R197.

kosuke55 (approved):
thanks!!! looks very nice!

soblin (approved):

PR #6382, opened 2024-02-12
feat(tier4_simulator_launch): add option to disable all perception related modules

brkay54:
## Description

Part of:
- https://github.com/autowarefoundation/autoware.universe/issues/6547

Required for:
- https://github.com/autowarefoundation/autoware.universe/pull/5954

Reaction_analyzer runs with PSim and it publishes its perception outputs to be able to measure exact reaction times. 

What we want to do is we can be able to disable the perception publishers.

This change does not have any effect on system behavior because its default value is `true`.

## Tests performed
Tested in PSim. Works as we expected. 

## Effects on system behavior
- `launch_simulator_perception_modules` is `true` -> it does not have any effect on the system.
- `launch_simulator_perception_modules` is `false` -> it does not launch perception-related modules.

### DIFF
--- a/launch/tier4_simulator_launch/launch/simulator.launch.xml
+++ b/launch/tier4_simulator_launch/launch/simulator.launch.xml
@@ -6,7 +6,7 @@
   <arg name="occupancy_grid_map_updater"/>
   <arg name="occupancy_grid_map_updater_param_path"/>
   <arg name="pose_initializer_param_path"/>
-
+  <arg name="launch_simulator_perception_modules" default="true"/>
   <arg name="launch_dummy_perception"/>
   <arg name="launch_dummy_vehicle"/>
   <arg name="launch_dummy_localization"/>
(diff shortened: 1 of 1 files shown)


yukkysaito:
Thanks. The change itself is good, but I don't know if the original PR approach is appropriate.
https://github.com/autowarefoundation/autoware.universe/pull/5954#pullrequestreview-1875710722
cc @xmfcx

xmfcx:
@brkay54 could you rebase and resolve the conflicts?

brkay54:
@xmfcx I solved the conflicts, however, currently, I am implementing `message_filters` to `reaction_analyzer`. So I couldn't try it yet, not tested. I am converting it to a draft. I will open it again asap.

launch/tier4_simulator_launch/launch/simulator.launch.xml
```diff
@@ -26,21 +26,6 @@
...
-  </group>
-
-  <!-- Dummy Perception -->
-  <group if="$(var launch_dummy_perception)">
-    <include file="$(find-pkg-share dummy_perception_publisher)/launch/dummy_perception_publisher.launch.xml">
-      <arg name="real" value="$(var perception/enable_detection_failure)"/>
-      <arg name="use_object_recognition" value="$(var perception/enable_object_recognition)"/>
-      <arg name="use_base_link_z" value="$(var perception/use_base_link_z)"/>
-      <arg name="visible_range" value="$(var sensing/visible_range)"/>
-    </include>
```
xmfcx, line 42: Could you bring back this included launch file?

launch/tier4_simulator_launch/launch/simulator.launch.xml
```diff
@@ -26,21 +26,6 @@
...
   <let name="vehicle_model_pkg" value="$(find-pkg-share $(var vehicle_model)_description)"/>
 
-  <group if="$(var scenario_simulation)">
-    <include file="$(find-pkg-share fault_injection)/launch/fault_injection.launch.xml">
-      <arg name="config_file" value="$(var fault_injection_param_path)"/>
-    </include>
-  </group>
-
-  <!-- Dummy Perception -->
-  <group if="$(var launch_dummy_perception)">
```
xmfcx, line 36: Could you bring back condition too? Otherwise this PR will make this launch file act differently than before.

xmfcx:
The way it is right now,
> This change does not have any effect on system behavior because its default value is true.

this is not true because some parameters are removed with this pr.

brkay54:
> The way it is right now,
> 
> > This change does not have any effect on system behavior because its default value is true.
> 
> this is not true because some parameters are removed with this pr.

Sorry, for the missed lines. I updated it now. Tested in PSim with:
`  <arg name="launch_simulator_perception_modules" default="true"/>`
I couldn't see any effect on system behavior. However, when it is `false`, `behavior_path_planner` waits for perception messages and the vehicle does not move. (It is what we expect.)

brkay54:
@xmfcx I tried with both options. When it is true as the default option, there is no effect on system behavior. When it is false, it does not launch perception-related modules except `probabilistic_occupancy_grid_map` (which is needed for `behavior_path_planner`, `behavior_velocity_planner`, and `obstacle_velocity_limiter`) and reaction_analyzer can publish the Obstacles without problem.

xmfcx:
Why is `occupancy_grid_map` not part of this new perception condition?

> `<include file="$(find-pkg-share tier4_perception_launch)/launch/occupancy_grid_map/probabilistic_occupancy_grid_map.launch.xml">`

It is in the `tier4_perception_launch` file.

xmfcx:
Alright, with the last change, this change puts:
- fault_injection
- dummy_perception_publisher
- probabilistic_occupancy_grid_map
- tracking
- prediction
- empty_objects_publisher
- elevation_map_loader
- traffic_light

into a group controlled by "launch_simulator_perception_modules" bool.
It is `true` by default.

xmfcx (approved):
It doesn't change the existing behavior, should be ok to merge.

xmfcx:
@KeisukeShima
@taikitanaka3
@TakaHoribe
@takayuki5168
@tkimura4

Could you review as code owners? Should be ok to merge, simple change.

xmfcx:
@KeisukeShima
@taikitanaka3
@TakaHoribe
@takayuki5168
@tkimura4

Please use this view while reviewing:
- https://github.com/autowarefoundation/autoware.universe/pull/6382/files?diff=split&w=1

launch/tier4_simulator_launch/launch/simulator.launch.xml
```diff
@@ -26,97 +26,100 @@
...
+      <include file="$(find-pkg-share tier4_perception_launch)/launch/occupancy_grid_map/probabilistic_occupancy_grid_map.launch.xml">
+        <arg name="input_obstacle_pointcloud" value="true"/>
+        <arg name="input_obstacle_and_raw_pointcloud" value="false"/>
+        <arg name="input/obstacle_pointcloud" value="/perception/obstacle_segmentation/pointcloud"/>
+        <arg name="output" value="/perception/occupancy_grid_map/map"/>
+        <arg name="occupancy_grid_map_method" value="laserscan_based_occupancy_grid_map"/>
+        <arg name="occupancy_grid_map_param_path" value="$(var laserscan_based_occupancy_grid_map_param_path)"/>
+        <arg name="occupancy_grid_map_updater" value="$(var occupancy_grid_map_updater)"/>
+        <arg name="occupancy_grid_map_updater_param_path" value="$(var occupancy_grid_map_updater_param_path)"/>
+      </include>
```
takayuki5168, line 58: OGM is disabled when launch_simulator_perception_modules is false? Is this the expected behavior?

launch/tier4_simulator_launch/launch/simulator.launch.xml
```diff
@@ -6,7 +6,7 @@
   <arg name="occupancy_grid_map_updater"/>
   <arg name="occupancy_grid_map_updater_param_path"/>
   <arg name="pose_initializer_param_path"/>
-
+  <arg name="launch_simulator_perception_modules" default="true"/>
```
takayuki5168, line 9: To me, this option makes the launch file messy. I wanna look for a better way. If there's nothing, I'm fine with this PR's change.

launch/tier4_simulator_launch/launch/simulator.launch.xml
takayuki5168: > We do not want to disable OGM actually.

So, how will you achieve the above? You will launch OGM by yourself outside the autoware.launch.xml?

xmfcx: Yes, the variable names and conditions should represent what they really do.

OGM is a perception module so it should be disabled with the rest.

@brkay54 will launch it separately from the reaction analyzer node launch file.

If you have a better name for these variables or an alternative solution, we are open for it.

takayuki5168: I also agree to disable all the perception module with launch_simulator_perception_modules.

My question was that the PR's description said

> Therefore, we modify the launch file to disable all perception-related packages except occupancy_grid_map.

but actually, OGM is disabled as well.


Now I understand that you will launch it seprately.

launch/tier4_simulator_launch/launch/simulator.launch.xml
brkay54: Thank you for the review @takayuki5168. I am going to think about a better way coming days. If I find another way, I will create another PR to make launch file tidy.

xmfcx:
This is a very simple change, just adds 3 lines and changes no existing behaviors, can we merge it?

Once a better solution is found, it can easily be reverted.

@takayuki5168 -san

PR #6394, opened 2024-02-13
feat(ndt_scan_matcher): smooth ndt map update

anhnv3991:
## Description

- In ndt_scan_matcher, add a secondary ndt to prefetch map while the vehicle is moving to a new area. The main ndt_ptr can still be used for scan matching without blocking, while the new map is being loaded.
- ndt_omp must be checked out at https://github.com/tier4/ndt_omp/tree/feature/smooth_ndt_map_update
  - https://github.com/tier4/ndt_omp/pull/26

## Tests performed

- rosbag: https://drive.google.com/drive/folders/1VYhi6cHW4OXx5QvGih6aCskrAH-17iJl
- map: https://tier4.github.io/AWSIM/GettingStarted/QuickStartDemo/#launching-autoware
- The map is divided by [MapIV's point cloud divider](https://github.com/MapIV/pointcloud_divider)
- Launch autoware's logging simulator: `ros2 launch autoware_launch logging_simulator.launch.xml \
    map_path:=~/Downloads/nishishinjuku_autoware_map_divided \
    vehicle_model:=sample_vehicle \
    sensor_model:=awsim_sensor_kit`

- Replay the rosbag `ros2 bag play ${ROSBAG_PATH} --rate 1.0 \
  --clock 200`

[Screencast from 07-02-2024 10:56:45.webm](https://github.com/autowarefoundation/autoware.universe/assets/11849905/caaa8d35-b959-4149-8340-dc00f04a8e61)

## Effects on system behavior

None


localization/ndt_scan_matcher/src/map_update_module.cpp
```diff
@@ -94,35 +153,30 @@ void MapUpdateModule::update_ndt(
...
   const auto exe_end_time = std::chrono::system_clock::now();
   const auto duration_micro_sec =
     std::chrono::duration_cast<std::chrono::microseconds>(exe_end_time - exe_start_time).count();
   const auto exe_time = static_cast<double>(duration_micro_sec) / 1000.0;
   RCLCPP_INFO(logger_, "Time duration for creating new ndt_ptr: %lf [ms]", exe_time);
-
-  publish_partial_pcd_map();
 }
 
+int count = 0;
```
SakodaShintaro, line 175: Is this a variable that was used for debugging and is no longer needed?

localization/ndt_scan_matcher/src/map_update_module.cpp
```diff
@@ -94,35 +153,30 @@ void MapUpdateModule::update_ndt(
...
+
 void MapUpdateModule::publish_partial_pcd_map()
 {
   pcl::PointCloud<PointTarget> map_pcl = ndt_ptr_->getVoxelPCD();
-
   sensor_msgs::msg::PointCloud2 map_msg;
```
SakodaShintaro, line 180: It's a very small point, but once you get rid of this line break deletion, the difference may become smaller.

localization/ndt_scan_matcher/include/ndt_scan_matcher/map_update_module.hpp
```diff
@@ -44,36 +45,46 @@ class MapUpdateModule
...
   rclcpp::Logger logger_;
   rclcpp::Clock::SharedPtr clock_;
 
   std::optional<geometry_msgs::msg::Point> last_update_position_ = std::nullopt;
 
   HyperParameters::DynamicMapLoading param_;
+
+  // Indicate if there is a prefetch thread waiting for being collected
+  NdtPtrType secondary_ndt_ptr_;
+  bool rebuild_;
```
SakodaShintaro, line 87: Is it correct to think that this variable called `rebuild_` means "whether a rebuild is required?" In that case, how about naming it `need_rebuild_`?

SakodaShintaro (commented):
Added a quick comment about trivial issues with the code.
A review of the logic will be conducted separately in the future. :+1:

localization/ndt_scan_matcher/src/map_update_module.cpp
anhnv3991: @SakodaShintaro Yes. I used it for checking whether the maps are loaded properly.

localization/ndt_scan_matcher/include/ndt_scan_matcher/map_update_module.hpp
anhnv3991: @SakodaShintaro Yes.
In this PR, I load the map to a secondary ndt ptr, so the scan alignment can be done on the main ndt ptr without locking. When the map loading is done, the main ndt ptr is swapped with the secondary ndt ptr, so the time of blocking is negligible (in my assumption).

However, when the new position is moved to a new position very far from the previous position, aligning scan on the main ndt ptr (while updating the secondary ndt) is not correct. In such cases, the main ndt ptr should be locked and updated _before_ any scan alignment.

This _rebuild__ flag is used to indicate whether the update should be done on the main ndt ptr directly. If it is false, the update is done on the secondary ndt ptr. If it is true, the update is performed directly on the ndt ptr and the pointer is locked until the update is finished.

localization/ndt_scan_matcher/src/map_update_module.cpp
```diff
@@ -28,29 +28,48 @@ MapUpdateModule::MapUpdateModule(
 
   pcd_loader_client_ =
     node->create_client<autoware_map_msgs::srv::GetDifferentialPointCloudMap>("pcd_loader_service");
+
+  secondary_ndt_ptr_.reset(new NormalDistributionsTransform);
+
+  if (ndt_ptr_) {
```
SakodaShintaro, line 34: Is it correct to think that the case where `ndt_ptr_` is false is a rare case? At that time, I felt that it would be good to output an ERROR log.

localization/ndt_scan_matcher/src/map_update_module.cpp
```diff
@@ -28,29 +28,48 @@ MapUpdateModule::MapUpdateModule(
...
+    // If the map does not keep up with the current position,
+    // lock ndt_ptr_ entirely until it is fully rebuilt.
+    need_rebuild_ = true;
   }
+
   return distance > param_.update_distance;
 }
 
-void MapUpdateModule::update_map(const geometry_msgs::msg::Point & position)
+void MapUpdateModule::prefetch_map(const geometry_msgs::msg::Point & position, NdtPtrType & ndt)
```
SakodaShintaro, line 65: This function is named `prefetch_map`, but I feel like the naming is a bit confusing since it also takes `ndt_ptr_` as an argument and updates it.

Just a suggestion, how about bringing all the current `prefetch_map` code into `update_ndt`?

localization/ndt_scan_matcher/src/map_update_module.cpp
```diff
@@ -69,11 +88,51 @@ void MapUpdateModule::update_map(const geometry_msgs::msg::Point & position)
...
+
+  // Memorize the position of the last update
   last_update_position_ = position;
+
+  // Publish the new ndt maps
+  publish_partial_pcd_map();
 }
 
 void MapUpdateModule::update_ndt(
+  NdtPtrType & ndt,
```
SakodaShintaro, line 135: This may just be my stupidity, but I feel that using pointer references makes the code more complex and difficult to understand.
How about making it a function that receives a reference to NormalDistributionsTransform?

localization/ndt_scan_matcher/include/ndt_scan_matcher/map_update_module.hpp
```diff
@@ -44,36 +45,46 @@ class MapUpdateModule
   using PointTarget = pcl::PointXYZ;
   using NormalDistributionsTransform =
     pclomp::MultiGridNormalDistributionsTransform<PointSource, PointTarget>;
+  using NdtPtrType = std::shared_ptr<NormalDistributionsTransform>;
 
 public:
   MapUpdateModule(
     rclcpp::Node * node, std::mutex * ndt_ptr_mutex,
-    std::shared_ptr<NormalDistributionsTransform> ndt_ptr,
+    std::shared_ptr<NormalDistributionsTransform> & ndt_ptr,
```
SakodaShintaro, line 53: Isn't it possible to achieve what you want by simply using a `shared_ptr` instead of a reference to a pointer?
Sorry, but I feel like using pointer references makes the concept a bit more complex and difficult to understand.

SakodaShintaro (commented):
Added some comments. I would appreciate it if you could give me answers.

localization/ndt_scan_matcher/include/ndt_scan_matcher/map_update_module.hpp
```diff
@@ -44,36 +45,46 @@ class MapUpdateModule
...
   rclcpp::Logger logger_;
   rclcpp::Clock::SharedPtr clock_;
 
   std::optional<geometry_msgs::msg::Point> last_update_position_ = std::nullopt;
 
   HyperParameters::DynamicMapLoading param_;
+
+  // Indicate if there is a prefetch thread waiting for being collected
+  NdtPtrType secondary_ndt_ptr_;
+  bool need_rebuild_;
```
SakodaShintaro, line 87: After discussing it with Ando-san, we came to the conclusion that `need_rebuild_` may not be necessary for this pull request. What do you think?

Even if this module needs to rebuild `ndt_ptr_` itself, I thought it is possible to update the `secondary_ndt_ptr_` and copying it to `ndt_ptr_`.
There will be waste copying, but I think it will eliminate conditional branches and make the code simpler.

Please let me know if you have any questions.

localization/ndt_scan_matcher/include/ndt_scan_matcher/map_update_module.hpp
anhnv3991: @SakodaShintaro I don't think it is possible.
If we just use a pointer, when we perform the assignment 
`ndt_ptr_ = secondary_ndt_ptr_`
in the map_update_module.cpp, the value of `ndt_ptr_` in ndt_scan_matcher_core.cpp **_does not_** change. That means the new scans are still aligned to the old `ndt_ptr_`, not the updated one, and that is not correct.

localization/ndt_scan_matcher/include/ndt_scan_matcher/map_update_module.hpp
anhnv3991: @SakodaShintaro There are two cases here:

1. The current position is close to the previously loaded position. In this case, aligning the current scan to the old `ndt_ptr_` still works, because the scan and the previously loaded map overlap to some extent. Hence, we can update the `secondary_ndt_ptr_` while aligning the current scan with the old `ndt_ptr_` just fine.
2. The current position is very far from the previously loaded position. In this case, aligning the current scan to the old ndt_ptr_ produces the incorrect result, because the scan the previously loaded map does not overlap at all. If we does not lock ndt_ptr_ in this case, the system will try to align the incoming scans (until the secondary_ndt_ptr_ is 100% updated) with the old ndt_ptr_, thus giving incorrect results. Therefore, in this case, I think the ndt_ptr_ must be locked and rebuilt before any scan alignment is called.

localization/ndt_scan_matcher/src/map_update_module.cpp
anhnv3991: @SakodaShintaro I added a `RCLCPP_ERROR_STREAM_THROTTLE` to report an error if ndt_ptr is null in the constructor of `MapUpdateModule`.

localization/ndt_scan_matcher/src/map_update_module.cpp
anhnv3991: @SakodaShintaro Done!

localization/ndt_scan_matcher/src/map_update_module.cpp
anhnv3991: @SakodaShintaro Done!

localization/ndt_scan_matcher/include/ndt_scan_matcher/map_update_module.hpp
SakodaShintaro: I understand the idea of not copying the instance, just exchanging pointers.
Great! :+1:

localization/ndt_scan_matcher/include/ndt_scan_matcher/map_update_module.hpp
SakodaShintaro: I also understand the need for `need_rebuild_`.
The code is OK as is.

SakodaShintaro (commented):
I think it's basically OK.

I have only added two comments, so please respond.

localization/ndt_scan_matcher/src/map_update_module.cpp
```diff
@@ -28,29 +28,90 @@ MapUpdateModule::MapUpdateModule(
...
+    // the main ndt_ptr_) overlap, the latency of updating/alignment reduces partly.
+    // If the updating is done the main ndt_ptr_, either the update or the NDT
+    // align will be blocked by the other.
+    update_ndt(position, *secondary_ndt_ptr_);
+
+    ndt_ptr_mutex_->lock();
+    auto input_source = ndt_ptr_->getInputSource();
+    ndt_ptr_ = secondary_ndt_ptr_;
+    ndt_ptr_->setInputSource(input_source);
+    ndt_ptr_mutex_->unlock();
```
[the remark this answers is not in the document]
anhnv3991: @SakodaShintaro You may be right. The NDT destructor could be the reason for that 3~7 msec execution time.

localization/ndt_scan_matcher/src/map_update_module.cpp
```diff
@@ -28,29 +28,90 @@ MapUpdateModule::MapUpdateModule(
 
   pcd_loader_client_ =
     node->create_client<autoware_map_msgs::srv::GetDifferentialPointCloudMap>("pcd_loader_service");
+
+  secondary_ndt_ptr_.reset(new NdtType);
+
+  if (ndt_ptr_) {
+    *secondary_ndt_ptr_ = *ndt_ptr_;
```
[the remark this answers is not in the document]
anhnv3991: @SakodaShintaro I think we should keep it like that. We don't know if the map loader module could be used somewhere else in future, so we should not rely on external code to verify the input parameters/data. The module should be able to verify the input by itself.

localization/ndt_scan_matcher/src/map_update_module.cpp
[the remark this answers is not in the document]
SakodaShintaro: OK. I think it's fast enough.

SakodaShintaro (approved):
Looks Good To Me!

PR #6416, opened 2024-02-15
fix(pose_instability_detector): changed the interval_sec value of pose_instability_detector

SakodaShintaro:
## Description

Currently, `interval_sec` in `pose_instability_detector` is set to 1.0 seconds.

However, the localization_stability timeout in `system_error_monitor`, which aggregates this, is also set to 1.0 seconds.

https://github.com/autowarefoundation/autoware.universe/blob/b2cd2aac9ebd5bb3e8468c10ee2b008a6ef1101b/system/system_error_monitor/config/diagnostic_aggregator/localization.param.yaml#L33-L37

Therefore, if there is even a slight delay, it will become stale.

In thie PR, I have changed `interval_sec` of `pose_instability_detector` to 0.5 seconds so that it does not become stale immediately even if there is a delay.

#### Why not
There are two solutions: increase the `timeout` of `system_monitor` or decrease `interval_sec` of `pose_instability_detector`.
I chose to decrease the `interval_sec` of `pose_instability_detector` in this PR because I was worried that increasing the `system_monitor` timeout would delay noticing timeout errors.

## Tests performed

AWSIM + Autoware works fine without stale in `rqt_robot_monitor`.

```
ros2 run rqt_robot_monitor rqt_robot_monitor
```

Note: Since autoware outputs a large number of topic messages to `/diagnostics`, drops will still occur in `rqt_runtime_monitor` where the queue size is not large enough.

```
```

## Effects on system behavior

ERRORs in `system_error_monitor` due to `pose_instability_detector` stale are less likely to occur.

### DIFF
--- a/localization/pose_instability_detector/config/pose_instability_detector.param.yaml
+++ b/localization/pose_instability_detector/config/pose_instability_detector.param.yaml
@@ -1,6 +1,6 @@
 /**:
   ros__parameters:
-    interval_sec: 1.0  # [sec]
+    interval_sec: 0.5  # [sec]
     threshold_diff_position_x: 1.0  # [m]
     threshold_diff_position_y: 1.0  # [m]
     threshold_diff_position_z: 1.0  # [m]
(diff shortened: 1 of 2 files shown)


TaikiYamada4 (approved):
LGTM!

PR #6475, opened 2024-02-21
fix(joy_controller): joy_type is only set from the param.yaml file

xmfcx:
## Tests performed

Not applicable.

## Effects on system behavior

Not applicable.


xmfcx:
@satoshi-ota @satoshi-ota I've looked for all the references of this `joy_type` parameter and it is not referenced by any other launch file.

Screenshot from 2024-02-21 21-24-02

If someone wants to set this parameter, they should use the param.yaml file as planned.

Please review :bowing_man:

control/joy_controller/config/joy_controller.param.yaml
```diff
@@ -1,6 +1,6 @@
 /**:
   ros__parameters:
-    joy_type: $(var joy_type)
+    joy_type: DS4 # DS4, G29, P65, XBOX
```
kaspermeck-arm, line 3: I think the `# DS4, G29, P65, XBOX` can be removed as these live in the schema:
* https://github.com/autowarefoundation/autoware.universe/blob/main/control/joy_controller/schema/joy_controller.schema.json#L13

kaspermeck-arm: @xmfcx 
I don't think it's necessary but if you want to keep it, I'm OK with it!

kaspermeck-arm (approved):
LGTM!

xmfcx:
@rej55 @satoshi-ota @shmpwk @taikitanaka3 @TakaHoribe @takayuki5168 @tkimura4

Could you review as a code owner? I would like to merge this as soon as possible and make the action required.

rej55:
@xmfcx 
This package should be optionally launched, separate from Autoware, if we would like to control from joy.
The launch command is like following.

```
ros2 launch joy_controller joy_controller.launch.xml joy_type:=G29
```

As you can see, we would like to launch `joy_controller` with `joy_type`argument, so `joy_type` should be able to change dynamically.

xmfcx:
@rej55 you can call it like:

`ros2 launch joy_controller joy_controller.launch.xml config_file:=/path/to/config-file.params.yaml`

And set it in the `.param.yaml` file.

Why is this not enough?

rej55:
@xmfcx 
Thank you for your suggestion.
It seems okay for me.

I will ask the user of this package.
Could you wait?
(It may take a while to reply because it is a holiday... 🙇)

xmfcx:
Another option would be to create a second launch file, called `joy_controller_no_yaml.launch.xml` or `joy_controller_direct_params.launch.xml`.

And this new launch file wouldn't use the param.yaml at all. All params would be set in the launch file.

By default we would use the `joy_controller.launch.xml` and for convenience, you could use the `joy_controller_no_yaml` version.

What do you think?

shmpwk:
I agree launching joy_controller with
> ros2 launch joy_controller joy_controller.launch.xml config_file:=/path/to/config-file.param.yaml

I think it would be better to add instructions on how to launch it to the README for previous users.

rej55:
@xmfcx 
Thank you for waiting, the users also agreed the following suggestion;

> If necessary, we can generate 4 `param.yaml` files for each joystick type, would this be alright?

Could you make these parameter files and add instruction to README?

xmfcx:
@rej55 @shmpwk I've updated the README and the param files and the launch file.

Also added a new simple launch file to make everyone happy.

```
# With default config (ds4)
ros2 launch joy_controller joy_controller.launch.xml

# Default config but select from the existing parameter files
ros2 launch joy_controller joy_controller_param_selection.launch.xml joy_type:=ds4 # or g29, p65, xbox

# Override the param file
ros2 launch joy_controller joy_controller.launch.xml config_file:=/path/to/your/param.yaml
```

The old users can make use of the new `joy_controller_param_selection.launch.xml` launch file to select a config easily from the terminal.

The default launch file makes use of the param files as agreed.

xmfcx:
@rej55 @shmpwk ready for review again.

shmpwk (approved):
LGTM!

rej55 (approved):
LGTM. Thank you.

PR #6545, opened 2024-03-05
feat(start_planner): prevent hindering rear vehicles

danielsanchezaran:
## Description
Current Autoware:
Currently, obstacle collision check with vehicles coming from the rear is performed when the ego vehicle is trying to merge to a lane from the shoulder lane. As follows:

image

The collision check is done until the ego vehicle eventually overlaps with the centerline. However, it is possible that the check is performed when the ego vehicle already impedes the pass of vehicles as shown on the next image: 

image

In the case above, the ego vehicle has not quite overlapped the centerline yet (closest point in purple) but the gap between the ego and the lane boundary, W2, is smaller than the width W1 of the NPC vehicle (black). In this cases, with collision check active, the ego vehicle might stop moving, even though the rear vehicle cannot really pass. The final outcome is that the ego vehicle ends up impeding the previous vehicle from passing and also remains stopped.

This PR: 
With this PR, the ego will not check overlaps with the center line; instead, it will check whether the remaining gap between the ego and the lane is big enough for the rear obstacle to pass. If it is big enough, the ego stops/yields to the coming vehicle, but, if it is smaller, it means the rear vehicle cannot stop and the ego should proceed. There is also an added parameter "extra_width_margin_for_rear_obstacle" which can be used to artificially enlarge or reduce the rear vehicle's perceived width. 

Example:
1) The gap is big enough for the rear vehicle to pass, so the ego stops (start planner's red virtual wall)

image

2) The gap is small, the rear vehicle cannot pass and the ego has "claimed" the lane: The ego vehicle proceeds, no stop point is added (note the ego is not overlapping the centerline yet):

image

3) Same ego pose as 2) but the rear bicycle is small, so it can still fit in the gap left by ego: A stop point is inserted and we let the bicycle pass.

image

## Related links

Related ticket: [TIER IV INTERNAL LINK](https://tier4.atlassian.net/browse/RT1-5260)

## Tests performed
PSIM

Degradation tests -> NO degradation [TIER IV INTERNAL LINK](https://evaluation.ci.tier4.jp/evaluation/reports/a4700631-a45e-5d47-90f6-80ef1599a778?project_id=prd_jt)

## Notes for reviewers

A function to get the backward lanes was moved from lane change module to common so it can be used by the start planner too.

## Effects on system behavior

The ego will not impede upcoming traffic by not committing to a lane entrance it should do.


zulfaqar-azmi-t4 (approved):
LGTM

planning/behavior_path_start_planner_module/src/start_planner_module.cpp
```diff
@@ -287,6 +288,136 @@ bool StartPlannerModule::isCurrentPoseOnMiddleOfTheRoad() const
...
+    return std::fabs(calcLateralOffset(boundary_path, search_pose.position));
+  };
+
+  auto calc_left_lateral_offset = [&](
+                                    const lanelet::ConstLineString2d & boundary_line,
+                                    const geometry_msgs::msg::Pose & search_pose) {
+    return -calc_right_lateral_offset(boundary_line, search_pose);
+  };
+
+  // Get the ego's overhang point closest to the centerline path and the gap between said point an
```
kyoichi-sugahara, line 320:
```suggestion
  // Get the ego's overhang point closest to the centerline path and the gap between said point and
```

planning/behavior_path_start_planner_module/include/behavior_path_start_planner_module/start_planner_module.hpp
```diff
@@ -208,6 +208,8 @@ class StartPlannerModule : public SceneModuleInterface
   bool isModuleRunning() const;
   bool isCurrentPoseOnMiddleOfTheRoad() const;
   bool isOverlapWithCenterLane() const;
```
kyoichi-sugahara, line 210: can you delete this function?

kyoichi-sugahara:
@danielsanchezaran 
If my understanding is correct, the assumption is merging from left lane, correct?
If ego merge from right side, the gap is wrongly calculated?
image

kyoichi-sugahara:
@danielsanchezaran 
It feels like it would be better to determine from which side the merging is happening. Therefore, using `getOriginalStartPose()` and the `target lanes` to judge whether the starting point is on the right or left side of the target lane might be a good idea. 
What do you think?

kyoichi-sugahara:
Can not request a change...
How about changing following description from
```
    if (Can stop within constraints \n && \n no crossing centerline?) then (yes)
```
to
```
    if (Can stop within constraints \n && \n Has sufficient space for rear vehicle to drive?) then (yes)
```

danielsanchezaran:
> @danielsanchezaran If my understanding is correct, the assumption is merging from left lane, correct? If ego merge from right side, the gap is wrongly calculated? image

The current calculates both distances to the left AND right edges of the road and uses the largest one, so, in the case you describe, the gap would be the distance between the left bound of the lane and the ego's overhang point. So it does not matter from which side the merging is happening.

planning/behavior_path_start_planner_module/src/start_planner_module.cpp
```diff
@@ -287,6 +288,136 @@ bool StartPlannerModule::isCurrentPoseOnMiddleOfTheRoad() const
...
+      left_dist_to_lane_border = calc_left_lateral_offset(current_left_bound, point_pose);
+      right_dist_to_lane_border = calc_right_lateral_offset(current_right_bound, point_pose);
+    }
+  }
+  if (smallest_lateral_offset == std::numeric_limits<double>::max()) return false;
+
+  const double gap_between_ego_and_lane_border =
+    (std::abs(left_dist_to_lane_border) > std::abs(right_dist_to_lane_border))
+      ? std::abs(left_dist_to_lane_border)
+      : std::abs(right_dist_to_lane_border);
```
danielsanchezaran, line 367: @kyoichi-sugahara here the largest gap is used

danielsanchezaran: I think there is no priority for left or right merging. I consider both distances and choose the biggest one. What I am concerned about is this situation you showed: 

image

In the case of a very wide road and a thin ego vehicle, it might be possible that another vehicle can pass through the gap, but I feel that, once the ego has crossed the centerline it has "won" the road and other vehicles should not overtake it in this situation, so I believe the centerline overlap check should also be performed. What do you think @kyoichi-sugahara ?

danielsanchezaran:
The code has been updated so the closest ego overhang point to the lane border is used to calculate the gap. The border to calculate the distance is chosen depending on which side the ego is merging the lane from. @kyoichi-sugahara

planning/behavior_path_start_planner_module/include/behavior_path_start_planner_module/start_planner_module.hpp
kyoichi-sugahara: thanks!

kyoichi-sugahara (approved):
Thank you so much for your quick response!
LGTM!

PR #6559, opened 2024-03-07
feat(ndt_scan_matcher): added a function to check maximum distance of sensor points

SakodaShintaro:
## Tests performed

Not applicable.

## Effects on system behavior

Not applicable.

### DIFF
--- a/localization/ndt_scan_matcher/schema/sub/sensor_points.json
+++ b/localization/ndt_scan_matcher/schema/sub/sensor_points.json
@@ -1,0 +1,20 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "Parameters for Ndt Scan Matcher Node",
+  "definitions": {
+    "frame": {
+      "type": "object",
+      "properties": {
+        "required_distance": {
+          "type": "number",
+          "description": "Required distance of input sensor points. If the max distance of input sensor points is lower than this value, the scan matching will not be performed.",
+          "default": "10.0"
+        }
+      },
+      "required": [
+        "required_distance"
+      ],
+      "additionalProperties": false
+    }
+  }
+}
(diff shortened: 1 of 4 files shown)


TaikiYamada4 (commented):
@SakodaShintaro 
Almost perfect to me!
I made a situation that the concatenation doesn't work well by making `timeout_sec` to 0.01 [s] from 0.1 [s], and the implementation in this PR worked well.

Apart for that, could you add the unit [m] to the warning message for clarity?

localization/ndt_scan_matcher/schema/sub/sensor_points.json
TaikiYamada4, line 1: You should also add this to  `ndt_scan_matcher.schema.json`

TaikiYamada4 (changes requested):

SakodaShintaro:
@anhnv3991 
Thank you for your comment :)

Although this is an important point, it seems like it will take some time to fix it, so I would like to try merging this PR as is.

The current autoware.universe is degraded,  which initial position estimation fails with sample-rosbag with a probability of about 50%, so I would like to fix this as soon as possible.

However, since good design is of course important, I will address your points in detail.

(1) Concurrency
You are correct, but in this code, the computational complexity of checking sensor points is O(N). In NDT, N is usually around 1500, so it is not significant. Even in the future, the checking code may not become a heavy process.

(2) Modularity
I agree with you. How about creating a new class in the future, such as SensorPointsValidator and having NDTScanMatcher hold it as a member variable? Just as NDT's align processing is essentially implemented with ndt_omp, it may be desirable to create a dedicated class for checking input point clouds. I think NDTScanMatcher's responsibility is communication in ROS2. In any case, I agree that core logic should be written as little as possible in NDTScanMatcher.

(3) Reusability
I think the validation of input is different for each process. For example, points only from a short-range LiDAR may not be suitable for NDT but could be useful for other processes. In general, the requirements for inputs should be defined for each process.

localization/ndt_scan_matcher/schema/sub/sensor_points.json
SakodaShintaro: Fixed in https://github.com/autowarefoundation/autoware.universe/pull/6559/commits/f0a974ca208208aa435a2288e47b3160de15030b

PR #6611, opened 2024-03-13
fix(multi_object_tracker): mot multi-step prediction is not work as intended

technolojin:
## Description

The extended kalman filter update process was done in multiple-step when the given time-step is too large.
However, the implementation was not work as intended.

* multi-step prediction
```c++
// if dt is too large, shorten dt and repeat prediction
const uint32_t repeat = std::ceil(dt / dt_max_);
const double dt_ = dt / repeat;
for (uint32_t i = 0; i < repeat; ++i) {
  if (!predictStateStep(dt_, tmp_ekf_for_no_update)) {
    return false;
  }
}
```
The motion model prediction is linealization of non-linear motion model. The prediction step assumes the time difference is small enough. By keeping the prediction time step small, a maximum time step was set and the prediction is to be conducted in multiple times.

* Found symptom: the multi-step prediction is not working as intended. 
Until the reason is found, bring back to the one-step prediction.

## FOUND BUG and FIX
A bug was found on `predictStateStep` methods.
* intention: predict the future state of the variable Kalman filter `ekf`
* bug: using the member Kalman filter `ekf_`
* fix: get the state vector from the variable Kalman filter `ekf`

## Tests performed
Tested on recorded data replay recomputation. 

## Effects on system behavior
Not applicable.

### DIFF
--- a/perception/multi_object_tracker/src/tracker/motion_model/motion_model_base.cpp
+++ b/perception/multi_object_tracker/src/tracker/motion_model/motion_model_base.cpp
@@ -50,25 +50,9 @@ bool MotionModel::predictState(const rclcpp::Time & time)
     return true;
   }
 
-  // // if dt is too large, shorten dt and repeat prediction
-  // {
-  //   const uint32_t repeat = std::floor(dt / dt_max_) + 1;
-  //   const double dt_ = dt / repeat;
-  //   for (uint32_t i = 0; i < repeat; ++i) {
-  //     if (!predictStateStep(dt_, ekf_)) {
-  //       return false;
-  //     }
-  //     // add interval to last_update_time_
-  //     last_update_time_ += rclcpp::Duration::from_seconds(dt_);
-  //   }
-  //   std::cout << "MotionModel::predictState predict  dt: " << dt << ", dt_: " << dt_  << ",
-  //   repeat:
-  //   "<< repeat<< std::endl;
-  // }
-
   if (!predictStateStep(dt, ekf_)) return false;
 
-  // update last_update_time_ to the estimation time
+  // update last_update_time_
   last_update_time_ = time;
   return true;
 }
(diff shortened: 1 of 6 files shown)


YoshiRi:
@technolojin 
Successfully built on my laptop and it seems to work, but could you describe more detail so that we can check what was wrong with past PR and how do you think fixed it?

perception/multi_object_tracker/src/tracker/motion_model/motion_model_base.cpp
```diff
@@ -46,16 +46,28 @@ bool MotionModel::predictState(const rclcpp::Time & time)
...
+  //     if (!predictStateStep(dt_, ekf_)) {
+  //       return false;
+  //     }
+  //     // add interval to last_update_time_
+  //     last_update_time_ += rclcpp::Duration::from_seconds(dt_);
+  //   }
+  //   std::cout << "MotionModel::predictState predict  dt: " << dt << ", dt_: " << dt_  << ",
+  //   repeat:
+  //   "<< repeat<< std::endl;
+  // }
```
yukkysaito, line 67: Would you like to leave this comment out?

technolojin: the commented-out code is not necessary for updating SW.
I removed it.

technolojin:
@YoshiRi 
I described definition of `multi-step prediction` and current issue.
The reason of the issue is under investigation. Until then, I would like to roll-back the part to the previous status, and this PR is for that.

technolojin:
The bug solution is found. Please wait for the fix.

technolojin:
the bug is found, the solution is applied, and the explanation is added to the description.

yukkysaito (approved):
LGTM

YoshiRi (approved):
LGTM. 

The delay compensation worked with my rosbag. (blue: detection, yellow: prediction/tracking)
image

Issue #6617, opened 2024-03-14
multi_object_tracker dies when unknown object has only one point in footprint which is 0,0,0


mebasoglu:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I'm convinced that this is not my fault but a bug.

### Description

Sometimes, the input DetectedObjects message has an unknown object that has only one point inside the footprint field which is 0,0,0 point. When this message arrives, the node dies. Here is the terminal output:

```
[multi_object_tracker-63] terminate called after throwing an instance of 'boost::numeric::positive_overflow'
[multi_object_tracker-63]   what():  bad numeric conversion: positive overflow
[multi_object_tracker-63] *** Aborted at 1710407978 (unix time) try "date -d @1710407978" if you are using GNU date ***
[multi_object_tracker-63] PC: @                0x0 (unknown)
[multi_object_tracker-63] *** SIGABRT (@0x3e80001ef73) received by PID 126835 (TID 0x786d4e635f40) from PID 126835; stack trace: ***
[multi_object_tracker-63]     @     0x786d4cd07046 (unknown)
[multi_object_tracker-63]     @     0x786d4e042520 (unknown)
[multi_object_tracker-63]     @     0x786d4e0969fc pthread_kill
[multi_object_tracker-63]     @     0x786d4e042476 raise
[multi_object_tracker-63]     @     0x786d4e0287f3 abort
[multi_object_tracker-63]     @     0x786d4e4a2b9e (unknown)
[multi_object_tracker-63]     @     0x786d4e4ae20c (unknown)
[multi_object_tracker-63]     @     0x786d4e4ae277 std::terminate()
[multi_object_tracker-63]     @     0x786d4e4ae4d8 __cxa_throw
[multi_object_tracker-63]     @     0x786d4cf2f5a5 object_recognition_utils::getIntersectionArea()
[multi_object_tracker-63]     @     0x786d4cf61e91 object_recognition_utils::get2dIoU<>()
[multi_object_tracker-63]     @     0x786d4cf615de DataAssociation::calcScoreMatrix()
[multi_object_tracker-63]     @     0x786d4cee0ba2 MultiObjectTracker::onMeasurement()
[multi_object_tracker-63]     @     0x786d4cef2187 std::_Function_handler<>::_M_invoke()
[...]

PR #6618, opened 2024-03-14
fix(multi_object_tracker): make sure smallest unknown footprint is 0.3

mebasoglu:
## Description

Related issue:
- https://github.com/autowarefoundation/autoware.universe/issues/6617

When unknown object polygon has only one point which is 0,0,0 , the node dies.

## Related links

Fixes: https://github.com/autowarefoundation/autoware.universe/issues/6617

## Tests performed

Confirmed on logging simulator.

## Notes for reviewers

There are two places where this error occurs, on `data_association.cpp` and `multi_object_tracker_core.cpp`. The fix was added those.

### DIFF
--- a/common/object_recognition_utils/include/object_recognition_utils/matching.hpp
+++ b/common/object_recognition_utils/include/object_recognition_utils/matching.hpp
@@ -65,6 +65,12 @@ inline double getUnionArea(const Polygon2d & source_polygon, const Polygon2d & t
 template <class T1, class T2>
 double get2dIoU(const T1 source_object, const T2 target_object, const double min_union_area = 0.01)
 {
+  if (
+    source_object.shape.type == autoware_auto_perception_msgs::msg::Shape::POLYGON &&
+    source_object.shape.footprint.points.size() < 3) {
+    return 0.0;
+  }
+
   const auto source_polygon = tier4_autoware_utils::toPolygon2d(source_object);
   const auto target_polygon = tier4_autoware_utils::toPolygon2d(target_object);
 
--- a/common/object_recognition_utils/include/object_recognition_utils/matching.hpp
+++ b/common/object_recognition_utils/include/object_recognition_utils/matching.hpp
@@ -30,8 +30,18 @@ namespace object_recognition_utils
 {
 using tier4_autoware_utils::Polygon2d;
 
+inline bool validatePolygons(const Polygon2d & source_polygon, const Polygon2d & target_polygon)
+{
+  static constexpr double min_area = 1e-9;
+  if (boost::geometry::area(source_polygon) < min_area) return false;
+  if (boost::geometry::area(target_polygon) < min_area) return false;
+  return true;
+}
+
 inline double getConvexShapeArea(const Polygon2d & source_polygon, const Polygon2d & target_polygon)
 {
+  if (!validatePolygons(source_polygon, target_polygon)) return 0.0;
+
   boost::geometry::model::multi_polygon<Polygon2d> union_polygons;
   boost::geometry::union_(source_polygon, target_polygon, union_polygons);
 
@@ -50,6 +60,8 @@ inline double getSumArea(const std::vector<Polygon2d> & polygons)
 inline double getIntersectionArea(
   const Polygon2d & source_polygon, const Polygon2d & target_polygon)
 {
+  if (!validatePolygons(source_polygon, target_polygon)) return 0.0;
+
   std::vector<Polygon2d> intersection_polygons;
   boost::geometry::intersection(source_polygon, target_polygon, intersection_polygons);
   return getSumArea(intersection_polygons);
@@ -57,6 +69,8 @@ inline double getIntersectionArea(
 
 inline double getUnionArea(const Polygon2d & source_polygon, const Polygon2d & target_polygon)
 {
+  if (!validatePolygons(source_polygon, target_polygon)) return 0.0;
+
   std::vector<Polygon2d> union_polygons;
   boost::geometry::union_(source_polygon, target_polygon, union_polygons);
   return getSumArea(union_polygons);
@@ -65,12 +79,6 @@ inline double getUnionArea(const Polygon2d & source_polygon, const Polygon2d & t
 template <class T1, class T2>
 double get2dIoU(const T1 source_object, const T2 target_object, const double min_union_area = 0.01)
 {
-  if (
-    source_object.shape.type == autoware_auto_perception_msgs::msg::Shape::POLYGON &&
-    source_object.shape.footprint.points.size() < 3) {
-    return 0.0;
-  }
-
   const auto source_polygon = tier4_autoware_utils::toPolygon2d(source_object);
   const auto target_polygon = tier4_autoware_utils::toPolygon2d(target_object);
 
@@ -90,8 +98,11 @@ double get2dGeneralizedIoU(const T1 & source_object, const T2 & target_object)
   const auto target_polygon = tier4_autoware_utils::toPolygon2d(target_object);
 
   const double intersection_area = getIntersectionArea(source_polygon, target_polygon);
+  if (intersection_area == 0.0) return 0.0;
   const double union_area = getUnionArea(source_polygon, target_polygon);
+  if (union_area == 0.0) return 0.0;
   const double convex_shape_area = getConvexShapeArea(source_polygon, target_polygon);
+  if (convex_shape_area == 0.0) return 0.0;
 
   const double iou = union_area < 0.01 ? 0.0 : std::min(1.0, intersection_area / union_area);
   return iou - (convex_shape_area - union_area) / convex_shape_area;


perception/multi_object_tracker/src/multi_object_tracker_core.cpp
```diff
@@ -386,6 +386,15 @@ void MultiObjectTracker::sanitizeTracker(
         continue;
       }
 
+      if (
+        object1.shape.type == autoware_auto_perception_msgs::msg::Shape::POLYGON &&
+        tier4_autoware_utils::getArea(object1.shape) == 0 &&
```
yukkysaito, line 391: It is not very desirable to make equal comparisons between floating-point types and direct numbers. 🙏 
In this case, it is 0, so it is OK, but it may cause unintended bugs.

yukkysaito: reference : https://www.geeksforgeeks.org/problem-in-comparing-floating-point-numbers-and-how-to-compare-them-correctly/

yukkysaito:
Thanks for your PR 👍 
We'll check this

armaganarsln:
/ask rate this PR quality, from 1 to 10. explain why

armaganarsln:
/ask At line 391 of multi_object_tracker_core.cpp, I'm comparing a float against an integer `tier4_autoware_utils::getArea(object1.shape) == 0` I don't think this is a good thing. What should I do instead?

perception/multi_object_tracker/src/multi_object_tracker_core.cpp
xmfcx: @yukkysaito can we use the following directly directly:

https://github.com/autowarefoundation/autoware.universe/blob/52b2fd6a2099f0ff2fc0e9b9c22a9a14163ad849/common/autoware_auto_common/include/autoware_auto_common/helper_functions/float_comparisons.hpp#L106-L122

Or do we have these float comparisons somewhere else?

perception/multi_object_tracker/src/data_association/data_association.cpp
```diff
@@ -207,6 +207,13 @@ Eigen::MatrixXd DataAssociation::calcScoreMatrix(
         if (passed_gate) {
           const double min_iou = min_iou_matrix_(tracker_label, measurement_label);
           const double min_union_iou_area = 1e-2;
+          if (
+            measurement_object.shape.type == autoware_auto_perception_msgs::msg::Shape::POLYGON &&
+            area == 0 && measurement_object.shape.footprint.points.size() == 1) {
```
xmfcx, line 212:
```suggestion
            area == 0 && measurement_object.shape.footprint.points.size() == 1) {
```
Same as here https://github.com/autowarefoundation/autoware.universe/pull/6618/files#r1525002210

perception/multi_object_tracker/src/data_association/data_association.cpp
```diff
@@ -207,6 +207,13 @@ Eigen::MatrixXd DataAssociation::calcScoreMatrix(
...
+            measurement_object.shape.footprint.points.front().x = 0.3;
+            measurement_object.shape.footprint.points.front().y = 0.3;
+            measurement_object.shape.footprint.points.front().z = 0.3;
+          }
```
yukkysaito, line 216: To set it to 0.3 would embed an unintended bug.In fact.
It sometimes results in strange behavior when there are actually objects with smaller values.
I think a better way to handle this issue is to have the getIntersectionArea function return 0 if the polygon does not have 3 or more points in it.

YoshiRi (commented):
I think this issue should be solved with fixing IoU caluculate function, so I do not want to apply this PR.
See https://github.com/autowarefoundation/autoware.universe/issues/6617#issuecomment-2001244853 .

mebasoglu:
Hello everyone, thank you for the feedback. I have pushed a commit for `get2dIoU`. Now, if there aren't enough points for a polygon, it will return `0.0`. It works on logging simulator.

The other approach was checking points size in `getIntersectionArea` function. That will also affect `get2dGeneralizedIoU`, 
`get2dPrecision` and `get2dRecall`.

What do you think about applying the patch to `getIntersectionArea`, or should it be only for `get2dIoU`?

YoshiRi:
@mebasoglu 
Thanks for your work!

As you say, changing  `getIntersectionArea`  may better since most function you mentioned has following escape sequence. I prefer this approach.
```cpp
  const double intersection_area = getIntersectionArea(source_polygon, target_polygon);
  if (intersection_area == 0.0) return 0.0; // you can escape here
```

But it seems `get2dGeneralizedIoU` do not have the escape, so we need to do some work then.

```cpp
  const double intersection_area = getIntersectionArea(source_polygon, target_polygon); // no escape
  const double union_area = getUnionArea(source_polygon, target_polygon); // this part may cause error
  const double convex_shape_area = getConvexShapeArea(source_polygon, target_polygon); // this will cause error
```

It would be helpful if you could change `getUnionArea`, `getConvexShapeArea` and `getIntersectionArea`.

technolojin (changes requested):
I agree to add checkers insures that the two inputs are valid.

The boost documentation of [intersection](https://live.boost.org/doc/libs/1_56_0/libs/geometry/doc/html/geometry/reference/algorithms/intersection.html) mentioned as following.
"Check the [Polygon Concept](https://live.boost.org/doc/libs/1_56_0/libs/geometry/doc/html/geometry/reference/concepts/concept_polygon.html) for the rules that polygon input for this algorithm should fulfill"

Having at least 3 footprint points is a necessary condition, not the sufficient condition. to have a area, the sufficient condition is to have a rank of 2.

I would like to suggest to add checkers of the validity of the polygon, as following.
```
static const double MIN_AREA = 1e-9;
if (boost::geometry::area(source_polygon) < MIN_AREA) return 0.0;
if (boost::geometry::area(target_polygon) < MIN_AREA) return 0.0;
```
In the official [documentation](https://live.boost.org/doc/libs/1_50_0/libs/geometry/doc/html/geometry/reference/algorithms/area/area_2_with_strategy.html) of the area function, 
" 
[Behavior](https://live.boost.org/doc/libs/1_50_0/libs/geometry/doc/html/geometry/reference/algorithms/area/area_2_with_strategy.html#geometry.reference.algorithms.area.area_2_with_strategy.behavior)
pointlike (e.g. point) Returns 0
linear (e.g. linestring) Returns 0
areal (e.g. polygon) Returns the area
"

The checkers may implied in the necessary functions, such as `get2dIoU`, `get2dGeneralizedIoU`, `get2dPrecision`, and `get2dRecall`.

mebasoglu:
Hello @YoshiRi @technolojin , thank you very much for the feedback.

I have created a new function `validatePolygons` . It returns `false` if the polygon areas are very close to zero, `true` otherwise. I added checks with this function on `getConvexShapeArea`, `getIntersectionArea` and `getUnionArea`. They return `0.0` if polygons aren't validated.

The `get2dGeneralizedIoU` function was also updated for the checks.

Tested on logging simulator.

technolojin:
@mebasoglu Thank you for your update. 
Your solution is a simple solution and it is important. However, I would like to give you some feedback.

1) Numerical error
As @mebasoglu checked, comparing a float number against zero (even the type is float) have a risk since the target number (`intersection_area` for example) can have a numerical error (overflow if this number is used for division, and that is what we do here). 

   Here are some related discussions.
   https://softwareengineering.stackexchange.com/questions/309350/avoiding-division-by-zero-using-float-comparison
   https://en.wikipedia.org/wiki/Machine_epsilon

  > Question: at line 391 of multi_object_tracker_core.cpp, i'm comparing a float against an integer `tier4_autoware_utils::getarea(object1.shape) == 0` i don't think this is a good thing. what should i do instead?
  > 
  > Answer: Comparing a float to an integer directly can indeed lead to precision issues due to the way floating point numbers are represented in memory. A better approach would be to use a small threshold value for the comparison.
  > 
  > Instead of `tier4_autoware_utils::getArea(object1.shape) == 0`, you could use something like:
  > 
  > ```c++
  > const float epsilon = 1e-9; // or any other small value that suits your precision needs
  > if (std::abs(tier4_autoware_utils::getArea(object1.shape)) < epsilon) {
  >     // ...
  > }
  > ```
  > 
  > This way, you're checking if the area is "close enough" to zero, taking into account the precision of floating point numbers.
  > **✨ Ask tool usage guide:**


2) Meaning of returning `0.0` on `getConvexShapeArea` and `getUnionArea` functions
As your updated functions, `getConvexShapeArea` and `getUnionArea` returns zero even one polygon have a valid area. As future users, it is confusing result. You implemented to check if the obtained area is zero before calculate necessary ratios, but other future developer may not understand the meaning and can cause catastrophic error (divide-by-zero). 
(Returning zero on `getIntersectionArea` is valid, by its definition.)

3) Duplicated checks
In `get2dGeneralizedIoU`,  the `validatePolygons` runs three times. and it is not necessary. 
The IoU calculations will repeated many many times and this change can make overall object tracking heavy, especially the number of tracking and detected object is large. 
(For example, calculation of matching score matrix have complexity of O(n^2))

I would like to request an efficient solution. Even it becomes a bit complicate, I think it worth.

mebasoglu:
Closing due to https://github.com/autowarefoundation/autoware.universe/pull/6701 was merged.

PR #6630, opened 2024-03-15
fix(out_of_lane): use bigger stop lines to cut predicted paths

maxime-clem:
## Description

Fixes an edge case with the feature to cut predicted paths at red traffic lights.
In some case the predicted path goes slightly around the stop line and it is not correctly cut.
With this PR, the stop line is made longer to avoid this edge case.

- TIER IV INTERNAL LINK: https://tier4.atlassian.net/browse/RT1-5753

## Tests performed

Psim

## Effects on system behavior

Not applicable.

### DIFF
--- a/planning/behavior_velocity_out_of_lane_module/src/filter_predicted_objects.cpp
+++ b/planning/behavior_velocity_out_of_lane_module/src/filter_predicted_objects.cpp
@@ -86,10 +86,11 @@ void cut_predicted_path_beyond_red_lights(
 {
   const auto stop_line = find_next_stop_line(predicted_path, planner_data);
   if (stop_line) {
+    // we use a longer stop line to also cut predicted paths that slightly go around the stop line
     auto longer_stop_line = *stop_line;
     const auto diff = stop_line->back() - stop_line->front();
-    longer_stop_line.front() += diff * -2.0;
-    longer_stop_line.back() += diff * 2.0;
+    longer_stop_line.front() -= diff * 0.5;
+    longer_stop_line.back() += diff * 0.5;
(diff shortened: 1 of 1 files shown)


planning/behavior_velocity_out_of_lane_module/src/filter_predicted_objects.cpp
```diff
@@ -85,7 +85,13 @@ void cut_predicted_path_beyond_red_lights(
   const PlannerData & planner_data, const double object_front_overhang)
 {
   const auto stop_line = find_next_stop_line(predicted_path, planner_data);
-  if (stop_line) cut_predicted_path_beyond_line(predicted_path, *stop_line, object_front_overhang);
+  if (stop_line) {
+    auto longer_stop_line = *stop_line;
+    const auto diff = stop_line->back() - stop_line->front();
+    longer_stop_line.front() += diff * -2.0;
+    longer_stop_line.back() += diff * 2.0;
```
takayuki5168, line 92: Does this mean that the stop line width becomes 5 times the original one? Isn't it too wide?
If you think it's fine, I'm okay.

takayuki5168 (approved):

planning/behavior_velocity_out_of_lane_module/src/filter_predicted_objects.cpp
maxime-clem: You are right I should not need to make it this big but it does not work as expected with smaller values :thinking: I will check what is the issue.

maxime-clem: I fixed the issue and now I only make the stop line twice bigger.

PR #6560, opened 2024-03-07
fix(ndt_scan_matcher): added error handling about update_ndt

SakodaShintaro:
## Description
The current `ndt_scan_matcher` crashes if the initial pose is placed far from the PCD map.

This is because in the case where `update_ndt` fails, `voxel_centroids_ptr_` of ndt remains nullptr and is referenced in subsequent code.

This pull request works around this issue by adding error handling.

## Tests performed

It has been confirmed that the `logging_simulator` runs with the same accuracy as before on AWSIM data with GT.

## Effects on system behavior

Placing the initial position estimate far from the PCD no longer causes a crash.

## Pre-review checklist for the PR author

The PR author **must** check the checkboxes below when creating the PR.

- [x] I've confirmed the [contribution guidelines].
- [x] The PR follows the [pull request guidelines].

## In-review checklist for the PR reviewers

The PR reviewers **must** check the checkboxes below before approval.

- [ ] The PR follows the [pull request guidelines].

## Post-review checklist for the PR author

The PR author **must** check the checkboxes below before merging.

- [ ] There are no open discussions or they are tracked via tickets.

After all checkboxes are checked, anyone who has write access can merge the PR.

[contribution guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/
[pull request guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/pull-request-guidelines/

PR #6464, opened 2024-02-21
fix(ndt_scan_matcher): added coping input_source when rebuilding ndt

SakodaShintaro:
## Description
I found that there was a problem with Driving Log Replayer not working properly, so fix it.

```
[ndt_scan_matcher-39] [WARN 1708474992.863786247] [localization.pose_estimator.ndt_scan_matcher]: No InputSource. Please check the input lidar topic (service_ndt_align() at /home/shintarosakoda/autoware/src/universe/autoware.universe/localization/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp:864)
[pose_initializer_node-41] [INFO 1708474992.864026735] [localization.util.pose_initializer_node]: NDT align server failed. (align_pose() at /home/shintarosakoda/autoware/src/universe/autoware.universe/localization/pose_initializer/src/pose_initializer/ndt_module.cpp:43)
[service_log_checker-4] [ERROR 1708474992.864308161] [system.service_log_checker]: /localization/initialize: status code 4 'NDT align server failed.' (/localization/util/pose_initializer_node) (set_error() at /home/shintarosakoda/autoware/src/universe/autoware.universe/common/component_interface_tools/src/service_log_checker.cpp:83)
[service_log_checker-4] [ERROR 1708474992.865189695] [system.service_log_checker]: /localization/initialize: status code 4 'NDT align server failed.' (/default_ad_api/node/localization) (set_error() at /home/shintarosakoda/autoware/src/universe/autoware.universe/common/component_interface_tools/src/service_log_checker.cpp:83)
[service_log_checker-4] [ERROR 1708474992.865347276] [system.service_log_checker]: /api/localization/initialize: status code 4 'NDT align server failed.' (/default_ad_api/node/localization) (set_error() at /home/shintarosakoda/autoware/src/universe/autoware.universe/common/component_interface_tools/src/service_log_checker.cpp:83)
```

Related PR

* https://github.com/autowarefoundation/autoware.universe/pull/6394

## Tests performed
https://tier4.github.io/driving_log_replayer/quick_start/installation/
Driving Log Replayer works fine
```bash
driving_log_replayer simulation run -p localization --rate 0.5
```

Logging Simulator works fine
```bash
# terminal 1
ros2 launch autoware_launch logging_simulator.launch.xml \
    map_path:=$HOME/autoware_map/sample-map-rosbag \
    vehicle_model:=sample_vehicle \
    sensor_model:=sample_sensor_kit

# terminal 2
[...]

PR #6649, opened 2024-03-18
test(ndt_scan_matcher): added an unit test to ndt_scan_matcher

SakodaShintaro:
## Description

Added two unit test cases to `ndt_scan_matcher`.

These test cases can detect bugs fixed in the following pull requests.

* https://github.com/autowarefoundation/autoware.universe/pull/6464
* https://github.com/autowarefoundation/autoware.universe/pull/6560

#### Limitation
Since `rclcpp::shotdown` must be performed for each test case, it is necessary to prepare a test binary for each test case.

#### The other change
Fixed a bug where input_source was discarded when initial position estimation was performed once outside the map.

https://github.com/autowarefoundation/autoware.universe/pull/6649/files#diff-ec315946878d62892ca32409f28e76703d96bb59470c4865b78bfce34d62be58R80-R82

## Tests performed

## Effects on system behavior

There are no effects on system behavior.


xmfcx:
(I just ran these to see how well PR-agent action performs, you can either use or ignore them. They might be useful though.)

SakodaShintaro:
@xmfcx 
Thank you very much. It looks like it's on trial, so I'll write a review of the review.

## Review
#### (1) Consider using explicit
I've already done this so I don't understand its point.
#### (2) Introduce Fixture
Currently there is only one test case, so it doesn't seem to be necessary, but I will consider it when the number of test cases increases.
#### (3) Avoid using EXIT_FAILURE
Agree
#### (4) Error handling of rcl_params
Error handling is performed when reading, but it may be better to do it when initializing variables as well. I'm a little worried about it becoming redundant.

## Suggestions
#### (1) Avoid using EXIT_FAILURE
I think the suggested modification method would be fine.
#### (2) removing the sleep
I will delete the line instead of commenting it out. The code used during debugging remained.
#### (3) EXPECT_NEAR -> EXPECT_NEAR
I think it would be okay to leave this as is.
#### (4) spin_until_future_complete
I didn't know about this, so I'll try it

Overall, there were some useful comments.

localization/ndt_scan_matcher/test/test_fixture.hpp
```diff
@@ -0,0 +1,89 @@
...
+    for (const auto & param_pair : param_map) {
+      for (const auto & param : param_pair.second) {
+        node_options.parameter_overrides().push_back(param);
+      }
+    }
+    node_ = std::make_shared<NDTScanMatcher>(node_options);
+    rcl_yaml_node_struct_fini(params_st);
+
+    // prepare tf_static "base_link -> sensor_frame"
+    tf_broadcaster_ = std::make_shared<tf2_ros::TransformBroadcaster>(node_);
```
YamatoAndo, line 59:
```suggestion
    tf_broadcaster_ = std::make_shared<tf2_ros::StaticTransformBroadcaster>(node_);
```
https://docs.ros.org/en/humble/Tutorials/Intermediate/Tf2/Writing-A-Tf2-Static-Broadcaster-Cpp.html#

localization/ndt_scan_matcher/test/test_fixture.hpp
```diff
@@ -0,0 +1,89 @@
...
+
+    // prepare stubs
+    pcd_loader_ = std::make_shared<StubPcdLoader>();
+    initialpose_client_ = std::make_shared<StubInitialposeClient>();
+    trigger_node_client_ = std::make_shared<StubTriggerNodeClient>();
+    sensor_pcd_publisher_ = std::make_shared<StubSensorPcdPublisher>();
+  }
+
+  std::shared_ptr<NDTScanMatcher> node_;
+  std::shared_ptr<tf2_ros::TransformBroadcaster> tf_broadcaster_;
```
YamatoAndo, line 82:
```suggestion
  std::shared_ptr<tf2_ros::StaticTransformBroadcaster> tf_broadcaster_;
```

localization/ndt_scan_matcher/test/test_fixture.hpp
SakodaShintaro: Fixed in https://github.com/autowarefoundation/autoware.universe/pull/6649/commits/d538dcfff442328baefa53c6fda61f1c5c925c84

localization/ndt_scan_matcher/test/test_fixture.hpp
SakodaShintaro: Fixed in https://github.com/autowarefoundation/autoware.universe/pull/6649/commits/d538dcfff442328baefa53c6fda61f1c5c925c84

YamatoAndo (approved):

PR #6671, opened 2024-03-22
fix(planning_validator): check max lateral acceleration after resampling trajectory entirely sparsely

kaigohirao:
## Description

Related JIRA Ticket:  https://tier4.atlassian.net/browse/RT0-29352

In the planning validator, among other checks, the max lateral acceleration is calculated and checked whether it is larger than 1G [ms^2] or not from an input of trajectory. The lateral accelerations are calculated from longitudinal velocities and curvatures at every trajectory points. 

The validation function in the planning validator includes two internal trajectory variables, namely `trajectory`, an input from the motion velocity smoother which is densely sampled near the ego and sparsely sampled otherwise with the policy explained [here](https://autowarefoundation.github.io/autoware.universe/main/planning/motion_velocity_smoother/#resample-trajectory,) , and `resampled`, representing the same path as `trajectory` but entirely sparsely resampled with the interval of 1~2m.

This PR puts the lateral acceleration check process after resampling of trajectory for the purpose of numerically safer calculations of curvatures.
## Tests performed
I confirmed the normal operation with the latest version of Autoware in the planning simulation [here](https://autowarefoundation.github.io/autoware-documentation/main/tutorials/ad-hoc-simulation/planning-simulation/).

## Effects on system behavior
Effects on the resulting max lateral accelerations would be small as the [curvature calcuation](https://github.com/autowarefoundation/autoware.universe/blob/e903d11d1fcba308325da3995a912d74bf0dcab2/planning/planning_validator/src/utils.cpp#L93-L154) is originally done after resampling the trajectory with the arc length of about 1m.


TomohitoAndo:
@kaigohirao 
Could you resolve the DCO error to pass CI?
image

PR #6706, opened 2024-03-28
feat(multi_object_tracker): debug timer to work with reduced replay rate recomputation

technolojin:
## Description

* Debug timers
Current timer only works in real-time. 
In a week development environment such as a laptop, usually reduced replay rate and rely on rostime.
This PR modify the timing to use rostime so that the evaluation is working under low speed replay.
Also some new timers were added.

* Refactoring
the debugging class is separated from the main node code.

## Tests performed

A rosbag replay was done in the rate of 0.2, and showed expected cycle time of 100 ms. 

Screenshot from 2024-03-28 19-55-44

## Effects on system behavior

Not applicable.

## Pre-review checklist for the PR author

The PR author **must** check the checkboxes below when creating the PR.

- [ ] I've confirmed the [contribution guidelines].
- [ ] The PR follows the [pull request guidelines].

## In-review checklist for the PR reviewers

The PR reviewers **must** check the checkboxes below before approval.

- [ ] The PR follows the [pull request guidelines].

## Post-review checklist for the PR author

The PR author **must** check the checkboxes below before merging.

- [ ] There are no open discussions or they are tracked via tickets.

After all checkboxes are checked, anyone who has write access can merge the PR.

[contribution guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/
[pull request guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/pull-request-guidelines/

PR #6687, opened 2024-03-25
feat(multi_object_tracker): separate timer usage from enable delay compensation flag

YoshiRi:
## Description

This PR tries to separate timer usage and delay compensation.

- add `timer_based_constant_publish_rate` parameter to utilize timer (default is `false`)
- conventional delay compensation will not use timer to prevent large processing delay

### Problem statement

Previously, `enable_delay_compensation` mode used timer to control publish timing and compensate processing delay.
In this mode, subscription and topic publish is asynchronous. So, there is a problem that it can cause large processing delay depending the phase.

image

In the worst case, we got almost 100ms delay! 
image

However, fixing timer publish phase was difficult since the tracking process often does not perform in constant time.
Keep constant publishing rate while controlling publish timing seems unrealistic, so I want to separate those objective and its implementation. 

## Related links

https://tier4.atlassian.net/browse/RT1-5741?atlOrigin=eyJpIjoiOWMwZWRkN2FjZDM0NDlmMzliZjQwZWFiYmE2OGUwYTQiLCJwIjoiaiJ9

## Pre-review checklist for the PR author

The PR author **must** check the checkboxes below when creating the PR.

- [ ] I've confirmed the [contribution guidelines].
- [ ] The PR follows the [pull request guidelines].

## In-review checklist for the PR reviewers

The PR reviewers **must** check the checkboxes below before approval.

- [ ] The PR follows the [pull request guidelines].
- [ ] The PR has been properly tested.
- [ ] The PR has been reviewed by the code owners.

## Post-review checklist for the PR author

The PR author **must** check the checkboxes below before merging.

- [ ] There are no open discussions or they are tracked via tickets.
- [ ] The PR is ready for merge.

After all checkboxes are checked, anyone who has write access can merge the PR.

[contribution guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/
[pull request guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/pull-request-guidelines/

PR #6710, opened 2024-03-29
feat(multi_object_tracker): reduce publish delay

technolojin:
## Description

This PR suggests a new method to reduce latency about 100 ms.

### Background: 
Current multi_object_tracker publishes the tracked object triggered by a timer. This timer ensures the topic is published in target frequency. The delay is compensated by predict the tracked objects motion.
However, the input (merged objects from multiple detectors) frequency is unstable (see the following plots, name of `input_latency_ms`). The instability of the detection latency is absorbed by the multi_object_tracker and adds latency about one cycle (100 ms of TIER IV configuration). 

### The new method:
there is two triggers to publish the tracked object
1) on measurement : When the publishing frequency is in the target window, publish right after the process is done. The publishing object will be compensated the delay from the measurement to the publishing time.
2) on timer : When the input is late, publish tracked object with delay compensation. 

### Side effect: 
The publishing frequency will not be kept precisely. (see the following plots, name of `cycle_time_ms` )

## Related links

[TIER IV INTERNAL](https://tier4.atlassian.net/browse/RT1-5741)

## Tests performed

Current
MOT_timing_now

Improved
MOT_timing_proposal

## Notes for reviewers

This PR will be merged after 
PR https://github.com/autowarefoundation/autoware.universe/pull/6706
PR https://github.com/autowarefoundation/autoware.universe/pull/6757
PR https://github.com/autowarefoundation/autoware.universe/pull/6775

This PR is an alternative work of PR https://github.com/autowarefoundation/autoware.universe/pull/6687

TODO: improve input (merged object) stability

## Interface changes
N/A

## Effects on system behavior
It may reduces overall latency about one cycle of the perception (100 ms).

### DIFF
--- a/perception/multi_object_tracker/src/multi_object_tracker_core.cpp
+++ b/perception/multi_object_tracker/src/multi_object_tracker_core.cpp
@@ -68,6 +68,7 @@ boost::optional<geometry_msgs::msg::Transform> getTransformAnonymous(
 
 MultiObjectTracker::MultiObjectTracker(const rclcpp::NodeOptions & node_options)
 : rclcpp::Node("multi_object_tracker", node_options),
+  last_published_time_(this->now()),
   tf_buffer_(this->get_clock()),
   tf_listener_(tf_buffer_)
 {
@@ -83,7 +84,7 @@ MultiObjectTracker::MultiObjectTracker(const rclcpp::NodeOptions & node_options)
     create_publisher<autoware_auto_perception_msgs::msg::TrackedObjects>("output", rclcpp::QoS{1});
 
   // Parameters
-  double publish_rate = declare_parameter<double>("publish_rate");
+  double publish_rate = declare_parameter<double>("publish_rate");  // [hz]
   world_frame_id_ = declare_parameter<std::string>("world_frame_id");
   bool enable_delay_compensation{declare_parameter<bool>("enable_delay_compensation")};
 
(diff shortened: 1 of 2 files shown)


YoshiRi:
Now I get your idea.
The key value in your logic is `constexpr double maximum_latency_ratio = 1.11;  // 11% margin`.

This value represents how large this "time keeper" can endure input delay variance.

If this value is small as 1.11, the time keeper is more strict so that it can cause large delay when larger input latency variance comes.
image

If I choose larger value like 1.5, the time keeper would not bother process and then processing time can be stick to raw value. It has drawback that the worst delay becomes larger since time keeper is loose.
image

technolojin:
@YoshiRi 
Your understand is right. If we increase the value `maximum_latency_ratio`, probability of prediction-without-update will reduced.
However, I do not know users can accept the long silent of 150 ms.

One another key tuning point is 
`const double minimum_publish_interval = publisher_period_ * 0.7;  // 70% of the period `
If we allow to publish topics in a short time (lower the `minimum_publish_interval`), The publisher triggered by message will keep up the latest object. 
I adjusted the value to `0.4` and the test result is as following.

Screenshot from 2024-04-05 13-51-33

The cyclic time is occasionally drops near to 40 ms, but the publish interval is kept within 116 ms in my environment.

perception/multi_object_tracker/src/multi_object_tracker_core.cpp
```diff
@@ -225,24 +237,32 @@ std::shared_ptr<Tracker> MultiObjectTracker::createNewTracker(
...
 {
   const rclcpp::Time current_time = this->now();
-  const auto self_transform =
-    getTransformAnonymous(tf_buffer_, world_frame_id_, "base_link", current_time);
-  if (!self_transform) {
-    return;
+  // check the publish period
+  const auto elapsed_time = (current_time - last_published_time_).seconds();
+  // if the elapsed time is over the period, publish objects with prediction
+  constexpr double maximum_latency_ratio = 1.11;  // 11% margin
```
YoshiRi, line 243: [imo]
Since early publish is better for the planning module, I prefer larger value like 1.4 or more.

technolojin: the acceptable publish interval is set to 70% to 111%, as agreed with @yukkysaito

technolojin:
This PR will be reviewed and merged after a bugfix of the MOT https://github.com/autowarefoundation/autoware.universe/pull/6775 is merged.

YoshiRi (approved):
LGTM. Worked within my laptop

PR #6739, opened 2024-04-04
perf(behavior_path_planner): improve getReferencePath to support very long routes

maxime-clem:
## Description

This PR improves the performance of the `getReferencePath` function by modifying the logic, switching from keeping a `route_lanelet_`  to keeping the `current_route_lanelet_`.
This allows computing the reference path quickly even with very long routes.

## Tests performed

Psim and evaluator (TIER IV INTERNAL LINK): https://evaluation.tier4.jp/evaluation/reports/7666b046-dcca-5893-9fba-80c2d580c1a2?project_id=prd_jt

## Effects on system behavior

Not applicable.


satoshi-ota:
@maxime-clem Thanks for your PR.

Mostly looks good. But I found exexpected module behaivor when I tested your branch.

### AUTONOMOUS MODE

Looks good.

https://github.com/autowarefoundation/autoware.universe/assets/44889564/88451e0b-d392-47c5-b67a-aade9af70ccc

### MANUAL DRIVING MODE (I reprodece the situation by moving ego pose by pc mouse.)

When I put ego position on adjacent lane, the avoidance path was vanished and lane change module was activated. 

https://github.com/autowarefoundation/autoware.universe/assets/44889564/40785206-6692-43a7-ba17-7a4c0a1a859b

I guess this caused by following changes. 

image

Basically the behavior_path_planner calls when it's NOT in AUTONOMOUS mode.
https://github.com/autowarefoundation/autoware.universe/blob/a336fa8e8a1053753442ac8d2faf491f4a8f58fb/planning/behavior_path_planner/src/behavior_path_planner_node.cpp#L390-L395

 But, sometimes, we want to check module behavior with manual driving mode. So, the module locked root lanelet if there are approved modules even when it's in manual driving.

Is is possible to leave the locking logic in `resetRootLanelet` function...? Or, do you have other good ideas...?

maxime-clem:
@satoshi-ota thank you for checking the PR and finding the issue with manual driving.
It should be fixed by this new commit: [cfd8cb4](https://github.com/autowarefoundation/autoware.universe/pull/6739/commits/cfd8cb4dbeb7d5aaa1b1d579b5e9b21dfbad01d3)
Here is a video showing the correct behavior while driving with the `ManualController`.

https://github.com/autowarefoundation/autoware.universe/assets/78338830/07471e08-332f-4e3c-b1b4-27c19501ad2b

satoshi-ota (approved):
LGTM :+1:

zulfaqar-azmi-t4 (approved):
LGTM

planning/behavior_path_planner/src/behavior_path_planner_node.cpp
```diff
@@ -372,7 +372,7 @@ void BehaviorPathPlannerNode::run()
   const bool is_first_time = !(planner_data_->route_handler->isHandlerReady());
```
danielsanchezaran, line 372: I suggest using an early return instead of nesting conditions. ie. 

```
if(!route_ptr){
const auto controlled_by_autoware_autonomously =
      planner_data_->operation_mode->mode == OperationModeState::AUTONOMOUS &&
      planner_data_->operation_mode->is_autoware_control_enabled;
    if (!controlled_by_autoware_autonomously && !planner_manager_->hasApprovedModules())
      planner_manager_->resetCurrentRouteLanelet(planner_data_);
return;
}
    planner_data_->route_handler->setRoute(*route_ptr);
    planner_manager_->resetCurrentRouteLanelet(planner_data_);

    // uuid is not changed when rerouting with modified goal,
    // in this case do not need to rest modules.
    const bool has_same_route_id =
      planner_data_->prev_route_id && route_ptr->uuid == planner_data_->prev_route_id;
    // Reset behavior tree when new route is received,
    // so that the each modules do not have to care about the "route jump".
    if (!is_first_time && !has_same_route_id) {
      RCLCPP_INFO(get_logger(), "New uuid route is received. Resetting modules.");
      planner_manager_->reset();
      planner_data_->prev_modified_goal.reset();
    }
```

danielsanchezaran (commented):
I left only one comment, the rest LGTM

planning/behavior_path_planner/src/behavior_path_planner_node.cpp
maxime-clem: Thank you. You are right we want to avoid nested conditions (https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f56-avoid-unnecessary-condition-nesting) but here we cannot do an early return since we want to execute the rest of the `run()` function in both condition branches.

danielsanchezaran: Ah sorry, I got confused, I though it was at the end of the function.

danielsanchezaran (approved):
LGTM

PR #6749, opened 2024-04-05
feat(ndt_scan_matcher): Timeout map update

anhnv3991:
## Description
Add a timeout period in map_update_module. When waiting for pcd load client, after the timeout period expires, if no response from the client, report an error and drop the NDT update.

## Tests performed

Not applicable.

## Effects on system behavior

Not applicable.

### DIFF
--- a/localization/ndt_scan_matcher/src/map_update_module.cpp
+++ b/localization/ndt_scan_matcher/src/map_update_module.cpp
@@ -146,12 +146,26 @@ bool MapUpdateModule::update_ndt(const geometry_msgs::msg::Point & position, Ndt
     request,
     [](rclcpp::Client<autoware_map_msgs::srv::GetDifferentialPointCloudMap>::SharedFuture) {})};
 
+      // Wait for maximum 10 milliseconds
+  std::chrono::milliseconds timeout(10);
+  auto start = std::chrono::system_clock::now();
+
   std::future_status status = result.wait_for(std::chrono::seconds(0));
   while (status != std::future_status::ready) {
     RCLCPP_INFO(logger_, "waiting response");
     if (!rclcpp::ok()) {
       return false;  // No update
     }
+
+    auto cur = std::chrono::system_clock::now();
+
+    // Report an error if wait for too long
+    if (cur - start >= timeout) {
+      RCLCPP_ERROR_STREAM_THROTTLE(
+        logger_, *clock_, 1000, "Waited for incoming PCDs for too long. Abandon NDT update.");
+      return false;
+    }
+
     status = result.wait_for(std::chrono::seconds(1));
   }
 


localization/ndt_scan_matcher/src/map_update_module.cpp
```diff
@@ -146,12 +146,26 @@ bool MapUpdateModule::update_ndt(const geometry_msgs::msg::Point & position, Ndt
...
       return false;  // No update
     }
+
+    auto cur = std::chrono::system_clock::now();
+
+    // Report an error if wait for too long
+    if (cur - start >= timeout) {
+      RCLCPP_ERROR_STREAM_THROTTLE(
+        logger_, *clock_, 1000, "Waited for incoming PCDs for too long. Abandon NDT update.");
+      return false;
```
YamatoAndo, line 166:
```suggestion
```

Please remove this line because if it is returned, it will cause the program to stop running.

localization/ndt_scan_matcher/src/map_update_module.cpp
```diff
@@ -146,12 +146,26 @@ bool MapUpdateModule::update_ndt(const geometry_msgs::msg::Point & position, Ndt
...
     RCLCPP_INFO(logger_, "waiting response");
     if (!rclcpp::ok()) {
       return false;  // No update
     }
+
+    auto cur = std::chrono::system_clock::now();
+
+    // Report an error if wait for too long
+    if (cur - start >= timeout) {
+      RCLCPP_ERROR_STREAM_THROTTLE(
```
YamatoAndo, line 164:
```suggestion
      RCLCPP_WARN_STREAM_THROTTLE(
```

I think this part would  be better to change to `WARN`.

https://autowarefoundation.github.io/autoware-documentation/main/contributing/coding-guidelines/ros-nodes/console-logging/#example

localization/ndt_scan_matcher/src/map_update_module.cpp
```diff
@@ -146,12 +146,25 @@ bool MapUpdateModule::update_ndt(const geometry_msgs::msg::Point & position, Ndt
     request,
     [](rclcpp::Client<autoware_map_msgs::srv::GetDifferentialPointCloudMap>::SharedFuture) {})};
 
+  // Wait for maximum 10 milliseconds
+  std::chrono::milliseconds timeout(10);
```
YamatoAndo, line 150: @anhnv3991 
Is there a reason for the constraint of 10ms?

Example:
・If driving at a speed of 100km/h, you will move y [m] in x [ms]. If you wait longer than this, there is a risk of moving out of the loaded map area.
・This process normally takes less than 1ms. So 10ms seems abnormal.

localization/ndt_scan_matcher/src/map_update_module.cpp
```diff
@@ -146,12 +146,25 @@ bool MapUpdateModule::update_ndt(const geometry_msgs::msg::Point & position, Ndt
...
+  auto start = std::chrono::system_clock::now();
+
   std::future_status status = result.wait_for(std::chrono::seconds(0));
```
YamatoAndo, line 153: @anhnv3991 
What does wait_for(std::chrono::seconds(0)) do?
Does it essentially do nothing by waiting for 0 seconds?
Or does it mean waiting indefinitely until completion?"

localization/ndt_scan_matcher/src/map_update_module.cpp
```diff
@@ -146,12 +146,25 @@ bool MapUpdateModule::update_ndt(const geometry_msgs::msg::Point & position, Ndt
...
+
+    auto cur = std::chrono::system_clock::now();
+
+    // Report an error if wait for too long
+    if (cur - start >= timeout) {
+      RCLCPP_WARN_STREAM_THROTTLE(
+        logger_, *clock_, 1000, "Waited for incoming PCDs for too long...");
+    }
+
     status = result.wait_for(std::chrono::seconds(1));
```
YamatoAndo, line 168: @anhnv3991 Even though the timeout is set to 10 ms, is it okay for this part to take 1 sec?

localization/ndt_scan_matcher/src/map_update_module.cpp
anhnv3991: @YamatoAndo If I understand the async_send_request correctly, wait_for is similar to std future wait_for. That means it waits for a maximum period. So if a response is received before that, it would return immediately with a ready status. If timeout, it would return a timeout status.

localization/ndt_scan_matcher/src/map_update_module.cpp
anhnv3991: @YamatoAndo No idea, it is not my code. Maybe the author wanted to check the status immediately after calling wait_for.

localization/ndt_scan_matcher/src/map_update_module.cpp
anhnv3991: @YamatoAndo So even when no PCD coming at all, the matching still continues? That does not sound right to me. If the map is not available then the matching should drop and a re-initialization should be performed.

localization/ndt_scan_matcher/src/map_update_module.cpp
YamatoAndo: So, how about writing the code like this?
```
while (result.wait_for(std::chrono::milliseconds(10)) != std::future_status::ready) {
  RCLCPP_WARN_STREAM_THROTTLE(
        logger_, *clock_, 1000, "Waited for incoming PCDs for too long...");
}
```

localization/ndt_scan_matcher/src/map_update_module.cpp
YamatoAndo: @anhnv3991 I measured the time it takes for wait_for in my environment, it seems to take about 0 to 300ms depending on the amount of the map being loaded. 
Since there is time required to load the map from the SSD, I think it is impossible to finish within 10ms.

anhnv3991: @YamatoAndo Can you share your data with 300ms loading time? I want to take a look at them.
300ms loading time sounds too much for me. Loading data from SSD should not take that much, unless your PCD are super dense.

YamatoAndo: @YamatoAndo OK. I will send you via TIER IV slack

PR #2913, opened 2023-02-20
perf(distortion_corrector_node): performance tuning

sykwer:
## Description
This PR makes `distortion_corrector_node` faster without changing the logical output (slight changes by approximate calculation of `sin/cos`). The processing of the `distortion_corrector_node` gets speeded up from about `21ms to 13ms` under the following condition.
- Ubuntu22.04 + ROS2 Humble + [Autoware Universe rosbag simulation](https://autowarefoundation.github.io/autoware-documentation/main/tutorials/ad-hoc-simulation/rosbag-replay-simulation/)
- Core Isolated
- Core Frequency Fixed (2.6GHz)
- L3 Cache: 12MB

distortion_corrector_node_performance_improvement_page-0001

This PR contains two kinds of changes.
- Avoid unnecessary instantiation of objects
- Pre-compute sin/cos values

It should be noted that the latter change results in a slight approximation of the logical output.

## Related links
[Background, Implementation Detail, and Performance Analysis](https://tier4.atlassian.net/wiki/spaces/~5ed0b4584824b20c18371c06/pages/2593587602/awf.tutorial+distortion+corrector+node)

## Tests performed
Check if `distortion_corrector_node` publishes the same logical output as before (using [Autoware Universe rosbag simulation](https://autowarefoundation.github.io/autoware-documentation/main/tutorials/ad-hoc-simulation/rosbag-replay-simulation/)).

The output is very slightly different in some parts since `sin/cos` is approximated by precomputation.

## Pre-review checklist for the PR author

The PR author **must** check the checkboxes below when creating the PR.

- [x] I've confirmed the [contribution guidelines].
- [x] The PR follows the [pull request guidelines].

## In-review checklist for the PR reviewers

The PR reviewers **must** check the checkboxes below before approval.

- [ ] The PR follows the [pull request guidelines].
- [ ] The PR has been properly tested.
- [ ] The PR has been reviewed by the code owners.

## Post-review checklist for the PR author

The PR author **must** check the checkboxes below before merging.

- [ ] There are no open discussions or they are tracked via tickets.
- [ ] The PR is ready for merge.

After all checkboxes are checked, anyone who has write access can merge the PR.

[contribution guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/
[pull request guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/pull-request-guidelines/

PR #6758, opened 2024-04-08
fix(distortion_correction_node): remove redundant code

vividf:
## Description
There is a for-loop and a while loop doing the same thing. Basically because when implementing the while loop, contributor forgot to delete it. 

## Tests performed

Not applicable.

## Effects on system behavior

Not applicable.

### DIFF
--- a/sensing/pointcloud_preprocessor/src/distortion_corrector/distortion_corrector.cpp
+++ b/sensing/pointcloud_preprocessor/src/distortion_corrector/distortion_corrector.cpp
@@ -263,12 +263,6 @@ bool DistortionCorrectorComponent::undistortPointCloud(
       // For performance, do not instantiate `rclcpp::Time` inside of the for-loop
       double imu_stamp = rclcpp::Time(imu_it->header.stamp).seconds();
 
-      for (;
-           (imu_it != std::end(angular_velocity_queue_) - 1 &&
-            *it_time_stamp > rclcpp::Time(imu_it->header.stamp).seconds());
-           ++imu_it) {
-      }
-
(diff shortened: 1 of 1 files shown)


yukkysaito:
@vividf Since the code appears to be designed to improve performance, it is better to check the execution time impact.

kminoda:
@yukkysaito My understanding is that this modification does not change any performance, simply increasing the readability of the code by removing "purely" redundant code, given that the following `while` loop does exactly the same thing as the `for` loop.

@vividf Would you confirm if the above argument is correct?

vividf:
@kminoda 
Yes, you are correct :) 

@yukkysaito 
In the PR, [#2913], you can also see that the author removed the for loop and added the while loop for optimization for twist. But for imu, only the while loop was added. 
I also analyze the time difference (green: remove for-loop, red: original)
image

Below is the value of the above graph (100 samples)
Remove for-loop:
min: 5.1, max: 11.98, avg: 6.56

Original
min: 5.29, max: 11.95, avg: 6.88

The value average processing time is lower because the current implementation still uses the for-loop. By removing the for-loop and using the while loop that the author implemented it will decrease the time. 


reference:
#(https://github.com/autowarefoundation/autoware.universe/pull/2913/files#:~:text=it_y%2C%20%2B%2Bit_z%2C%20%2B%2Bit_time_stamp)%20%7B-,for%20(%3B,%2B%2Btwist_it)%20%7B,-while%20(twist_it)

kminoda (approved):
LGTM

badai-nguyen:
@vividf could you resolve conflict?

vividf:
Due to the approval being late, other PR also solved this.

PR #6740, opened 2024-04-04
feat(goal_planner): fix non-thread-safe access in goal_planner

soblin:
## Description

depends https://github.com/autowarefoundation/autoware.universe/pull/6752

As described here https://github.com/autowarefoundation/autoware.universe/issues/6718, current goal_planner module is not thread-safe.

Following variables are not thread safe currently.

1. each shared point member of planner_data_ and their shared pointer members like route_handler->lanelet_map_ptr/routing_graph_ptr
2. getCurrentStatus()
3. getPreviousModuleOutput()
4. occupancy_grid_map_
5. last_previous_module_output_
6. prev_data_
7. goal_planner_data.check_collision
8. parameters_
9. ego_predicted_path_params_
10. objects_filtering_params_
11. safety_check_params_

### surrogate planner_data of GoalPlannerModule

1. PlannerData
2. getCurrentStatus()
3. getPreviousModuleOutput()
4. occupancy_grid_map_
8. parameters_
9. ego_predicted_path_params_
10. objects_filtering_params_
11. safety_check_params_

are saved in GoalPlannerData.

```
  struct GoalPlannerData
  {
    GoalPlannerParameters parameters;
    std::shared_ptr<EgoPredictedPathParams> ego_predicted_path_params;
    std::shared_ptr<ObjectsFilteringParams> objects_filtering_params;
    std::shared_ptr<SafetyCheckParams> safety_check_params;

    PlannerData planner_data;
    ModuleStatus current_status;
    BehaviorModuleOutput previous_module_output;
    std::shared_ptr<OccupancyGridBasedCollisionDetector> occupancy_grid_map;
  };
  std::optional<GoalPlannerData> gp_planner_data_{std::nullopt};
  std::mutex gp_planner_data_mutex_;
```

In each own thread this struct is copied to the thread local variable to acquire ownership of the shared pointer members of planner_data_ on the thread during the lifetime. 

Since the route_handler also contains shared pointer in it, we clone it in the following way:

```
void GoalPlannerModule::GoalPlannerData::update(
  const GoalPlannerParameters & parameters_,
  const std::shared_ptr<EgoPredictedPathParams> & ego_predicted_path_params_,
  const std::shared_ptr<ObjectsFilteringParams> & objects_filtering_params_,
  const std::shared_ptr<SafetyCheckParams> & safety_check_params_,
  const PlannerData & planner_data_, const ModuleStatus & current_status_,
  const BehaviorModuleOutput & previous_module_output_)
{
  parameters = parameters_;
[...]

Issue #6718, opened 2024-03-31
goal_planner (as well as start_planner) module of behavior_path_planner is not thread safe and we have a lot of room for refactoring 


soblin:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I'm convinced that this is not my fault but a bug.

### Description

# 7 non-thread-safe variables

## data race for planner_data_

[...]

PR #6779, opened 2024-04-09
feat(start_planner): fix non-thread-safe access

soblin:
## Description

related work for https://github.com/autowarefoundation/autoware.universe/pull/6740

following variables are copied to thread local variable for thread-safety
- planner_data_
- getCurrentStatus() result
- parameters_
- isStopped() result
- `status_` (PullOutStatus)

## Related links

https://github.com/autowarefoundation/autoware.universe/issues/6718

## Tests performed

https://evaluation.tier4.jp/evaluation/reports/51d9494d-0cb2-5726-bd8e-80583125eb27?project_id=prd_jt

## Interface changes

none

## Effects on system behavior

none

### DIFF
--- a/planning/behavior_path_start_planner_module/include/behavior_path_start_planner_module/start_planner_module.hpp
+++ b/planning/behavior_path_start_planner_module/include/behavior_path_start_planner_module/start_planner_module.hpp
@@ -155,6 +155,24 @@ public:
   bool isFreespacePlanning() const { return status_.planner_type == PlannerType::FREESPACE; }
 
 private:
+  struct StartPlannerData
+  {
+    StartPlannerParameters parameters;
+    PlannerData planner_data;
+    ModuleStatus current_status;
+    PullOutStatus main_thread_pull_out_status;
+    bool is_stopped;
+
+    StartPlannerData clone() const;
+    void update(
+      const StartPlannerParameters & parameters, const PlannerData & planner_data,
+      const ModuleStatus & current_status, const PullOutStatus & pull_out_status,
+      const bool is_stopped);
+  };
+  std::optional<PullOutStatus> freespace_thread_status_{std::nullopt};
+  std::optional<StartPlannerData> sp_planner_data_{std::nullopt};
+  std::mutex sp_planner_data_mutex_;
+
   // Flag class for managing whether a certain callback is running in multi-threading
   class ScopedFlag
   {
@@ -307,7 +325,9 @@ private:
   SafetyCheckParams createSafetyCheckParams() const;
   // freespace planner
   void onFreespacePlannerTimer();
-  bool planFreespacePath();
+  std::optional<PullOutStatus> planFreespacePath(
+    const StartPlannerParameters & parameters,
+    const std::shared_ptr<const PlannerData> & planner_data, const PullOutStatus & pull_out_status);
 
   void setDebugData();
   void logPullOutStatus(rclcpp::Logger::Level log_level = rclcpp::Logger::Level::Info) const;
--- a/planning/behavior_path_start_planner_module/include/behavior_path_start_planner_module/start_planner_module.hpp
+++ b/planning/behavior_path_start_planner_module/include/behavior_path_start_planner_module/start_planner_module.hpp
@@ -170,8 +170,8 @@ private:
       const bool is_stopped);
   };
   std::optional<PullOutStatus> freespace_thread_status_{std::nullopt};
-  std::optional<StartPlannerData> sp_planner_data_{std::nullopt};
-  std::mutex sp_planner_data_mutex_;
+  std::optional<StartPlannerData> start_planner_data_{std::nullopt};
+  std::mutex start_planner_data_mutex_;
 
   // Flag class for managing whether a certain callback is running in multi-threading
   class ScopedFlag
--- a/planning/behavior_path_start_planner_module/include/behavior_path_start_planner_module/start_planner_module.hpp
+++ b/planning/behavior_path_start_planner_module/include/behavior_path_start_planner_module/start_planner_module.hpp
@@ -308,7 +308,6 @@ private:
   bool hasFinishedBackwardDriving() const;
   bool hasCollisionWithDynamicObjects() const;
   bool isStopped();
-  bool isStuck();
   bool hasFinishedCurrentPath();
   void updateSafetyCheckTargetObjectsData(
     const PredictedObjects & filtered_objects, const TargetObjectsOnLane & target_objects_on_lane,
--- a/planning/behavior_path_start_planner_module/src/start_planner_module.cpp
+++ b/planning/behavior_path_start_planner_module/src/start_planner_module.cpp
@@ -1739,11 +1739,11 @@ void StartPlannerModule::StartPlannerData::update(
   /*
     auto route_handler_self = planner_data.route_handler;
     planner_data = planner_data_; // sync planer_data to planner_data_, planner_data.route_handler
-    is once repointed
+    is once re-pointed
 
     if (!planner_data_->is_route_handler_updated && route_handler_self != nullptr) {
       // we do not need to sync planner_data.route_handler with that of planner_data_
-      // repoint to the original again
+      // re-point to the original again
       planner_data.route_handler = route_handler_self;
     } else {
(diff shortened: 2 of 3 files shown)


planning/behavior_path_start_planner_module/src/start_planner_module.cpp
```diff
@@ -172,6 +202,41 @@ void StartPlannerModule::updateObjectsFilteringParams(
...
+  // planWaitingApproval()/plan(), so we can copy latest current_status to sp_planner_data_ here
+
+  // NOTE: onFreespacePlannerTimer copies sp_planner_data to its thread local variable, so we need
+  // to lock sp_planner_data_ here to avoid data race. But the following clone process is
+  // lightweight because most of the member variables of PlannerData/RouteHandler is
+  // shared_ptrs/bool
+  // begin of critical section
+  {
+    std::lock_guard<std::mutex> guard(sp_planner_data_mutex_);
+    if (!sp_planner_data_) {
```
danielsanchezaran, line 217: The if and else are deeply nested into the brackets, which makes the code a little harder to read when the `if (freespace_thread_status_)` line is reached (its an if inside and else , inside a lock_guard set of brackets)
I suggest using a lambda function []() -> StartPlannerData{}
So you can have an early return, something like this:
```
...
 sp_planner_data.value().update(
        *parameters_, *planner_data_, getCurrentStatus(), status_, isStopped());
return sp_planner_data;
}
auto & sp_planner_data = sp_planner_data_.value();
...
```

and you can eliminate the "else"

planning/behavior_path_start_planner_module/src/start_planner_module.cpp
```diff
@@ -93,26 +93,56 @@ void StartPlannerModule::onFreespacePlannerTimer()
...
   }
 
   const bool is_new_costmap =
-    (clock_->now() - planner_data_->costmap->header.stamp).seconds() < 1.0;
+    (clock_->now() - local_planner_data->costmap->header.stamp).seconds() < 1.0;
   if (!is_new_costmap) {
     return;
   }
 
-  if (isStuck()) {
```
danielsanchezaran, line 114: This method is only called here, I suggest either modifying the method to accept is_stopped , pull_out_status.planner_type and pull_out_status.found_pull_out_path

and call
` isStuck( is_stopped , pull_out_status.planner_type ,pull_out_status.found_pull_out_path) `

Alternatively, the isStuck() method can be made into a lambda, inside this same function. 

It can also be deleted altogether, but I feel it is more readable if a method called is_ego_stuck() is called here instead of directly checking for 3 different  conditions on this if()

danielsanchezaran (commented):
Thanks for this great PR, I have some minor comments regarding unused methods and variable names, but the changes themselves look solid, have you ran scenario tests with these changes applied?

planning/behavior_path_start_planner_module/src/start_planner_module.cpp
danielsanchezaran: Also, since the update method is called either way: 

sp_planner_data.update(
        *parameters_, *planner_data_, getCurrentStatus(), 
Do we really need to check if sp_planner_data_ has a value? 
cannot we always do this:
`sp_planner_data_ = StartPlannerData();
 sp_planner_data_.value().update(
        *parameters_, *planner_data_, getCurrentStatus(), status_, isStopped());
`
?
Then you only need to check for `if (freespace_thread_status_)` ?

planning/behavior_path_start_planner_module/include/behavior_path_start_planner_module/start_planner_module.hpp
```diff
@@ -155,6 +155,24 @@ class StartPlannerModule : public SceneModuleInterface
...
+    bool is_stopped;
+
+    StartPlannerData clone() const;
+    void update(
+      const StartPlannerParameters & parameters, const PlannerData & planner_data,
+      const ModuleStatus & current_status, const PullOutStatus & pull_out_status,
+      const bool is_stopped);
+  };
+  std::optional<PullOutStatus> freespace_thread_status_{std::nullopt};
+  std::optional<StartPlannerData> sp_planner_data_{std::nullopt};
```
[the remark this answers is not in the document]
soblin: all right ! I'll replace it with "start_planner_data"

planning/behavior_path_start_planner_module/src/start_planner_module.cpp
```diff
@@ -1665,4 +1732,25 @@ void StartPlannerModule::logPullOutStatus(rclcpp::Logger::Level log_level) const
 
   logFunc("=======================================");
 }
+
+StartPlannerModule::StartPlannerData StartPlannerModule::StartPlannerData::clone() const
```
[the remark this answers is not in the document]
soblin: I want to keep this method for now. As I'm doing in the body "begin/end of critical section", I'm creating a copy of StartPlannerData. Right now I'm making the copied of each member of StartPlannerData because most of the functions takes `const shared_ptr<PlannerData>` as the argument, not `const PlannerData&`, which is the reason why I'm making the copy of `StartPlannerData.planner_data` as the shared pointer.

In the future refactoring maybe I will replace most of the `const shared_ptr<PlannerData>` arguments to `const PlannerData&`, in which case the `clone()` method is useful.

planning/behavior_path_start_planner_module/src/start_planner_module.cpp
soblin: fixed in https://github.com/autowarefoundation/autoware.universe/pull/6779/commits/f12bd11c5b8c4b777bff93ef874d2069c0cc9495

planning/behavior_path_start_planner_module/src/start_planner_module.cpp
soblin: https://github.com/autowarefoundation/autoware.universe/pull/6779/commits/95ed1a51e694db2a8ff153cec78cf66325e22723

I moved the checking of `freespace_thread_status_` out of the `else` scope.

I do not want to refresh `start_planner_data_` value every time due to following reason:
1. PlannerData has the shared pointer to RouteHandler, which contains the shared pointer to `lanelet_map_ptr` and `routing_graph_ptr`. I'm making the deep copy of `start_planner_data.planner_data.route_handler` so StartPlanner can become the owner (== increase reference count) of `lanelet_map_ptr` and `routing_graph_ptr`.
2. But if we use huge lanelet map, this refresh of RouteHandler is heavy because it contains the vector of lanelets. So it is preferable to refresh it only when needed, or when the map/route has changed so to speak.
3. currently PlannerData does not have the flag indicating if `the map/route has changed`, but I will add it in the next PR. With that flag, we can skip the refresh of `StartPlannerData.planner_data.route_handler` if the flag is false.

planning/behavior_path_start_planner_module/src/start_planner_module.cpp
```diff
@@ -93,26 +93,56 @@ void StartPlannerModule::onFreespacePlannerTimer()
...
+      const auto & start_planner_data = start_planner_data_.value();
+      local_planner_data = std::make_shared<PlannerData>(start_planner_data.planner_data);
+      current_status_opt = start_planner_data.current_status;
+      parameters_opt = start_planner_data.parameters;
+      pull_out_status_opt = start_planner_data.main_thread_pull_out_status;
+      is_stopped = start_planner_data.is_stopped;
+    }
+  }
+  // finish copying thread sensitive data
+  if (!local_planner_data || !current_status_opt || !parameters_opt || !pull_out_status_opt) {
```
danielsanchezaran, line 115: Just a small detail, but could you please make the if(!local_planner_data || !current_status_opt || !parameters_opt || !pull_out_status_opt) condensed with a single boolean? 

like const bool is_data_ready = !local_planner_data || !current_status_opt || !parameters_opt || !pull_out_status_opt; 

In the future if  we need to check for more optional variables the if might become too large and CodeScene is complaining about this line.

danielsanchezaran (approved):
other than the multiple boolean comment I made everything else LGTM, I would appreciate it if @kyoichi-sugahara and @kosuke55 can also give their feedback before this PR is merged though. 

Thank you so much for this PR!

kyoichi-sugahara (approved):
LGTM!Thanks

PR #6781, opened 2024-04-09
fix(avoidance): fix invalid road bound distance calculation

satoshi-ota:
## Background

The avoidance module calculates distance to drivable bound by using `tier4_autoware_utils::intersection` with following step.

1. check object offset direction based on current path. (fig: on left)
2. calculate overhang point (=p1). (fig: black point)
3. calculate offset point (=p2) based on overhang points by using `tier4_autoware_utils::calcOffsetPose`.
4. calculate intersect point (=p3) between line p1-p2 and drivable bound.
5. calculate distance (=`D`) p1 to p3.

In most case (fig: nominal case), we can find intersect by following logic. But, in edge case, we can't.

```c++
        const auto p2 =
          calcOffsetPose(p_tmp, 0.0, (isOnRight(object) ? 100.0 : -100.0), 0.0).position;
        const auto opt_intersect =
          tier4_autoware_utils::intersect(p1.second, p2, bound.at(i - 1), bound.at(i));
```

Then, there is following logic as well to calculate `D'`.

```c++
        const auto p2 =
          calcOffsetPose(p_tmp, 0.0, (isOnRight(object) ? -100.0 : 100.0), 0.0).position;
        const auto opt_intersect =
          tier4_autoware_utils::intersect(p1.second, p2, bound.at(i - 1), bound.at(i));
```

IMG_0043

## Description

BUG: The module couldn't calculate distance to drivable bound properly. As a result, the module didn't create avoidance path even when there was enough space to avoid.

Screenshot from 2024-04-09 11-28-17

This issue caused by unexpected intersetion point calculation in following logic. And, I decreased the hard coded value to 1m in order not to get unexpected intersection point.

```c++
        const auto p2 =
          calcOffsetPose(p_tmp, 0.0, (isOnRight(object) ? -100.0 : 100.0), 0.0).position;
        const auto opt_intersect =
          tier4_autoware_utils::intersect(p1.second, p2, bound.at(i - 1), bound.at(i));
```

IMG_0044

## Tests performed

https://github.com/autowarefoundation/autoware.universe/assets/44889564/892447d8-2844-44f4-993c-9af831869bc6

- [x] [PASS TIER IV INTERNAL SCENARIOS](https://evaluation.tier4.jp/evaluation/reports/c2d865f1-f4c3-50e2-a5df-4d6fcf174c8a?project_id=prd_jt)

## Effects on system behavior

Fix bug.

### DIFF
--- a/planning/behavior_path_avoidance_module/src/utils.cpp
+++ b/planning/behavior_path_avoidance_module/src/utils.cpp
@@ -872,8 +872,7 @@ double getRoadShoulderDistance(
       }
 
       {
-        const auto p2 =
-          calcOffsetPose(p_tmp, 0.0, (isOnRight(object) ? -100.0 : 100.0), 0.0).position;
+        const auto p2 = calcOffsetPose(p_tmp, 0.0, (isOnRight(object) ? -1.0 : 1.0), 0.0).position;
         const auto opt_intersect =
           tier4_autoware_utils::intersect(p1.second, p2, bound.at(i - 1), bound.at(i));
 


shmpwk (approved):
LGTM!!

My note
- If there are several intersected points, the closer one is chosen. https://github.com/autowarefoundation/autoware.universe/blob/0f4cb74415654cd3599838d769500885db5ac2f5/planning/behavior_path_avoidance_module/src/utils.cpp#L889-L891

- I understand that the symmetrical lanelet, obstacle position would work as we expected.

PR #6802, opened 2024-04-12
fix(route_handler): planning route with goal lanelet in candidate_lanelets

sebekx:
## Description

There is and issue with route_handler in situation when goal lanelet is in candidate lanelet list. 

`getLaneletSequenceUpTo` function starts from the goal lanelet and iteratively check candidate lanenelet from current lanelet. But when there is goal lanelet in candidate lanelet search is stopped to prevent looping.
However in following situation:
image

We would to set goal on lanelet 4, but function 'getLaneletSequenceUpTo' going backward and for lanelet 3 it has 2 candidate lanelet 4 and 2. 4 is goal lanelet so processing is stopped.

Condition that protects against looping could be rewritten to remove goal lanelet from candidate_lanelets instead braking processing to support such cases.

## Tests performed

Verified that after applying changes i was able to set goal in such case:
Screenshot from 2024-04-10 15-17-45

## Effects on system behavior

Fix issue with generating lanelet segments by mission\_planner in specific case.
Not applicable.


planning/route_handler/src/route_handler.cpp
```diff
@@ -587,12 +587,13 @@ lanelet::ConstLanelets RouteHandler::getLaneletSequenceUpTo(
...
-          [lanelet](auto & prev_llt) { return lanelet.id() == prev_llt.id(); })) {
-      break;
-    }
+
+    // loop check, remove goal lanelet from candidate lanelets
+    candidate_lanelets.erase(
+      std::remove_if(
+        candidate_lanelets.begin(), candidate_lanelets.end(),
+        [lanelet](auto & prev_llt) { return lanelet.id() == prev_llt.id(); }),
+      candidate_lanelets.end());
```
takayuki5168, line 596: Thanks for the fix.
Can you check if the mission planner will not die in some corner cases like putting the goal pose behind the ego in the ego's lanelet.

sebekx: I've verified different cases and didn't observe that mission planner die. Putting the goal behind the ego produce bad trajectory but this is known issue i supposed: 
image

takayuki5168: Thank you. I will approve the change!

takayuki5168 (approved):

mitsudome-r:
@mitsudome-r will take a look and see if this modification is still necessary.

xmfcx:
I've summarized the changes here:

https://gist.github.com/xmfcx/a1a295cf654aceb62f4fa86accebec1b/revisions

- old main (this PR was against this)
- proposal version
- current main (as of 2025-01-27)

This PR added checks and behavior to add consideration for "goal lanelets.

The latest version has its logic updated quite a lot and I see specific consideration for `goal_lanelets_`.

I will assume this was fixed in the latest version and will close this PR.

PR #6843, opened 2024-04-18
perf(pointcloud_preprocessor): prevent excessive log and speed up a bit

KYabuuchi:
## Description

This PR contains two changes:

1. Avoiding calling `RCLCPP_WARN_STREAM_THROTTLE` for a number of points
2. Removing unnecessary `for` loops and `rclcpp::Time` class instantiations

**`RCLCPP_WARN_STREAM_THROTTLE` takes some processing time even if it does not actually output logs**, so calling it for a number of points would take a very long time. :sob:   In practice, printing once per point cloud is sufficient, so I made that change.

The second change provides a slight speedup. In my environment (with about 300,000 points), it resulted in a maximum speedup of about **10 ms**.

---

Also, if the distortion_corrector is delayed for some reason, the following scenario occurs, which is fatal:

1. Processing is significantly delayed for some reason.
2. Twist and IMU are stored in the queue at the correct time, so regardless of which data is paired with the point cloud, the stamp difference is large.
3. When the stamp difference is large, `RCLCPP_WARN_STREAM_THROTTLE` is called for each point cloud (~300,000 times).
4. Even if `RCLCPP_WARN_STREAM_THROTTLE` does not output logs, calling it many times takes a lot of time.
5. If step 4 takes too long, the stamp difference for matching between the vehicle speed and IMU in subsequent iterations also becomes large.
6. Return to step 2 and repeat indefinitely.

When it actually happens, it looks like the image below. This PR resolves this issue.

 

## Tests performed

I ran the logging_simulator on the sample data and measured the processing time. Unfortunately, there was minimal improvement in processing speed with the sample data.
The following graph shows the processing time for two runs before and after the changes. The vertical axis represents processing_time_ms. While there was little effect in the beginning, an improvement can be observed around 80-85 seconds

compare_distortion_corrector

Also, in my environment (with about 300,000 points), it resulted in a maximum speedup of about 10 ms.

## Effects on system behavior

The system behavior does not change.

### DIFF
--- a/sensing/pointcloud_preprocessor/src/distortion_corrector/distortion_corrector.cpp
+++ b/sensing/pointcloud_preprocessor/src/distortion_corrector/distortion_corrector.cpp
@@ -246,38 +246,43 @@ bool DistortionCorrectorComponent::undistortPointCloud(
   bool twist_time_stamp_is_too_late = false;
   bool imu_time_stamp_is_too_late = false;
 
-  while (twist_it != std::end(twist_queue_) - 1 && first_point_time_stamp_sec > twist_stamp) {
-    ++twist_it;
-    twist_stamp = rclcpp::Time(twist_it->header.stamp).seconds();
-  }
-
-  float v{static_cast<float>(twist_it->twist.linear.x)};
-  float w{static_cast<float>(twist_it->twist.angular.z)};
-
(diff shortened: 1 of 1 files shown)


yukkysaito (approved):
LGTM

PR #6871, opened 2024-04-24
fix(ekf_localizer): updated ekf gate_dist params

SakodaShintaro:
## Description
The current `ekf_localizer` has a gate function based on Mahalanobis distance, but the threshold is set to a very large value of 10000.0, so it does not actually work.

The Mahalanobis distance threshold can be set based on a statistical test assuming a chi-square distribution.

We will update the documentation to explain how to determine the significance level, and we will change the default value to a more realistic one.

## Tests performed

I measured the Mahalanobis distance for some data. The blue line is the actual Mahalanobis distance, and the orange dotted line is the threshold.

#### Data with successful localization
| name | image |
|:--------:|:---------:|
| AWSIM Nishishinjuku| diagnostics_ekf_localizer|
| Real data | diagnostics_ekf_localizer |

#### Data with failed localization
| name | image | note |
|:--------:|:--------:|:------:|
| case 1 | diagnostics_ekf_localizer | The threshold value set this time is also large enough, so even if there is a failure, the threshold value may not be reached. |
| case 2 | diagnostics_ekf_localizer | In this failure case, the actual value may exceed the threshold. |

The detection by Mahalanobis distance is also not perfect, so failure cannot always be detected. In particular, if the NDT score is below the score threshold, the NDT pose will not be published, so it cannot be detected by the observation gate in EKF.

## Effects on system behavior

In theory, this makes EKF slightly more robust to outliers. In reality, there will likely be little change.


Motsu-san (approved):

PR #6875, opened 2024-04-24
feat(autonomous_emergency_braking): filter and crop aeb pointcloud

danielsanchezaran:
## Description

This PR introduces an overhaul to the autonomous_emergency_braking (AEB) module. 

1. Added Euclidean clustering to consider only groups of point clouds that are bigger than a certain size (given by cluster tolerance and min amount of point parameters).
2. The Point Cloud from sensing is cropped to only include points close to, or inside the ego predicted path.
3. A Convex hull is created around each cluster of objects. The Vertices of the hull are the points used to calculate if a collision happens or not (the vertices of the hull correspond to the most “external points”, so it really only makes sense to consider said points.
4. Only the closest Object/Vertex is evaluated for collision, (if ego evaluates that the closest point is safe, the other points will be safe too since they have, by definition, larger distances to the ego).

The goal of these changes is to avoid false positive triggering of the AEB by noisy points, reduce computation time by cropping out unrelated points off of the AEB point cloud, and simplify collision detection by checking the closest point

Example of AEB ignoring objects far from path (AEB points show as red points, normal are grey):

https://github.com/autowarefoundation/autoware.universe/assets/25967964/94b2ff62-adae-4e44-86ce-e0c34a56190f

## Tests performed

PSIM tests

LSIM
Replayed these rosbags where AEB was wrongly activated, and there was no longer an activation
[TIER IV INTERNAL LINK](https://tier4.atlassian.net/browse/RT1-5364)
[TIER IV INTERNAL LINK](https://tier4.atlassian.net/browse/RT1-5666)

## Notes for reviewers

Requires changes to Launch: https://github.com/autowarefoundation/autoware_launch/pull/966

## Effects on system behavior

Reduced false positive detection for AEB

### DIFF
--- a/control/autonomous_emergency_braking/include/autonomous_emergency_braking/node.hpp
+++ b/control/autonomous_emergency_braking/include/autonomous_emergency_braking/node.hpp
@@ -148,6 +148,8 @@ public:
     const Path & ego_path, const std::vector<Polygon2d> & ego_polys, const rclcpp::Time & stamp,
     std::vector<ObjectData> & objects);
 
+  void cropPointCloudWithEgoPath(const std::vector<Polygon2d> & ego_polys);
+
   void createClusteredPointCloudObjectData(
     const Path & ego_path, const std::vector<Polygon2d> & ego_polys, const rclcpp::Time & stamp,
     std::vector<ObjectData> & objects);
@@ -161,6 +163,8 @@ public:
   void addCollisionMarker(const ObjectData & data, MarkerArray & debug_markers);
 
   PointCloud2::SharedPtr obstacle_ros_pointcloud_ptr_{nullptr};
+  PointCloud2::SharedPtr cropped_ros_pointcloud_ptr_{nullptr};
+
   VelocityReport::ConstSharedPtr current_velocity_ptr_{nullptr};
   Vector3::SharedPtr angular_velocity_ptr_{nullptr};
   Trajectory::ConstSharedPtr predicted_traj_ptr_{nullptr};
--- a/control/autonomous_emergency_braking/include/autonomous_emergency_braking/node.hpp
+++ b/control/autonomous_emergency_braking/include/autonomous_emergency_braking/node.hpp
@@ -185,7 +185,6 @@ public:
   bool publish_debug_pointcloud_;
   bool use_predicted_trajectory_;
   bool use_imu_path_;
-  bool crop_pointcloud_with_path_footprint_;
   double path_footprint_extra_margin_;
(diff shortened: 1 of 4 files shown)


control/autonomous_emergency_braking/include/autonomous_emergency_braking/node.hpp
```diff
@@ -65,6 +67,33 @@ using visualization_msgs::msg::MarkerArray;
 using Path = std::vector<geometry_msgs::msg::Pose>;
 using Vector3 = geometry_msgs::msg::Vector3;
 
+// For debugging
+class DebugTimer
```
zulfaqar-azmi-t4, line 71: We have tier4_autoware_utils::StopWatch for this

danielsanchezaran:
> Ok, so you are filtering out the obstacle pointclouds that belong to clusters on the (inflated) ego footprint convex hull on the path. And then find the closest point (not the center of the cluster) along ego's path and check safety in the conventional way.
> 
> Is it possible to move `cropeed_pointcloud_ptr` from the member to a local variable / argument ? I prefer having less internal state variables.

Exactly, I think it is fine to move the point cloud to a local variable, I will push a commit with the fix tomorrow!

tkimura4:
Could you update [Readme](https://github.com/autowarefoundation/autoware.universe/blob/main/control/autonomous_emergency_braking/README.md) ?

control/autonomous_emergency_braking/src/node.cpp
```diff
@@ -304,6 +318,8 @@ void AEB::onCheckCollision(DiagnosticStatusWrapper & stat)
 
 bool AEB::checkCollision(MarkerArray & debug_markers)
 {
+  using colorTuple = std::tuple<double, double, double, double>;
```
tkimura4, line 321: FYI

A useful utility function for creating debug markers is available [here](https://github.com/autowarefoundation/autoware.universe/blob/a6c69122541b141cbbf3c37c03151a82009d50fa/control/vehicle_cmd_gate/src/marker_helper.hpp#L61).

danielsanchezaran:
> Could you update [Readme](https://github.com/autowarefoundation/autoware.universe/blob/main/control/autonomous_emergency_braking/README.md) ?

Done!

control/autonomous_emergency_braking/include/autonomous_emergency_braking/node.hpp
danielsanchezaran: Deleted

tkimura4 (approved):

PR #6900, opened 2024-04-26
feat(start_planner): add centerline crossing check

danielsanchezaran:
## Description

Related to this ticket: [TIER IV INTERNAL LINK](https://tier4.atlassian.net/browse/RT1-6090)

on Autoware, the start planner has a check that lets it ignore rear vehicles if the gap between the ego and its merging lane is not big enough to let the rear vehicle pass through, this is done on the isPreventingRearVehicleFromPassingThrough() function. 

However, there are situations where the function isPreventingRearVehicleFromPassingThrough() does not consider objects that are not directly behind the ego vehicle but that have a predicted path that goes through the ego path. In said cases isPreventingRearVehicleFromPassingThrough() will return false because it does not detect an object behind ego (this is for safety) but then the dynamic  obstacle collision check will detect the object an issue a stop wall, even if the ego vehicle has properly merged with its target lane. 

image

Ideally the function isPreventingRearVehicleFromPassingThrough() should categorize vehicles based on intention (predicted path) and not just position, but said behavior requires some though and work time. 

On the mean time I issued this PR to improve the  isPreventingRearVehicleFromPassingThrough() function by checking if any of the ego overhang points has crossed the center of the road. In that case, it does not make much sense to think about dynamic collision with obstacles from behind since the ego has already merged into its target lane. 

With this PR the start planner ignores false dynamic collision cases like in this case:

https://github.com/autowarefoundation/autoware.universe/assets/25967964/c2f79d85-a02f-4bf1-9fce-ffa78f987fb0

## Tests performed

Evaluator tests [TIER IV INTERNAL LINK](https://evaluation.ci.tier4.jp/evaluation/reports/02855474-ede0-557b-90fc-e35f2542ce02?project_id=prd_jt) <- there is degradation but it is not caused by this PR afaik

### DIFF
--- a/planning/behavior_path_start_planner_module/src/start_planner_module.cpp
+++ b/planning/behavior_path_start_planner_module/src/start_planner_module.cpp
@@ -441,7 +441,9 @@ bool StartPlannerModule::isPreventingRearVehicleFromPassingThrough() const
   geometry_msgs::msg::Pose ego_overhang_point_as_pose;
   const auto [gap_between_ego_and_lane_border, corresponding_lateral_gap_with_other_lane_bound] =
     get_gap_between_ego_and_lane_border(ego_overhang_point_as_pose, ego_is_merging_from_the_left);
-  if (!gap_between_ego_and_lane_border.has_value()) return false;
+  if (!gap_between_ego_and_lane_border.has_value()) {
+    return false;
+  }
   if (gap_between_ego_and_lane_border.value() < corresponding_lateral_gap_with_other_lane_bound) {
     // middle of the lane is crossed, no need to check for collisions anymore
     return true;
--- a/planning/behavior_path_start_planner_module/src/start_planner_module.cpp
+++ b/planning/behavior_path_start_planner_module/src/start_planner_module.cpp
@@ -403,7 +403,7 @@ bool StartPlannerModule::isPreventingRearVehicleFromPassingThrough() const
     const auto vehicle_footprint =
       transformVector(local_vehicle_footprint, tier4_autoware_utils::pose2transform(current_pose));
     double smallest_lateral_gap_between_ego_and_border = std::numeric_limits<double>::max();
-    double corresponding_lateral_gap_with_opposite_lane;
+    double corresponding_lateral_gap_with_other_lane_bound;
     for (const auto & point : vehicle_footprint) {
       geometry_msgs::msg::Pose point_pose;
       point_pose.position.x = point.x();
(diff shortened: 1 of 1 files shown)


planning/behavior_path_start_planner_module/src/start_planner_module.cpp
```diff
@@ -416,25 +417,34 @@ bool StartPlannerModule::isPreventingRearVehicleFromPassingThrough() const
       const lanelet::ConstLineString2d current_lane_bound = (ego_is_merging_from_the_left)
                                                               ? closest_lanelet_const.rightBound2d()
                                                               : closest_lanelet_const.leftBound2d();
+      const lanelet::ConstLineString2d opposite_lane_bound =
+        (ego_is_merging_from_the_left) ? closest_lanelet_const.leftBound2d()
+                                       : closest_lanelet_const.rightBound2d();
```
kyoichi-sugahara, line 422:
```suggestion
      const auto [current_lane_bound, opposite_lane_bound] =
        ego_is_merging_from_the_left
          ? std::make_pair(
              closest_lanelet_const.rightBound2d(), closest_lanelet_const.leftBound2d())
          : std::make_pair(
              closest_lanelet_const.leftBound2d(), closest_lanelet_const.rightBound2d());
```

planning/behavior_path_start_planner_module/src/start_planner_module.cpp
```diff
@@ -463,6 +473,7 @@ bool StartPlannerModule::isPreventingRearVehicleFromPassingThrough() const
   // filtering objects based on the current position's lane
   const auto target_objects_on_lane = utils::path_safety_checker::createTargetObjectsOnLane(
     relevant_lanelets.value(), route_handler, filtered_objects, objects_filtering_params_);
+
```
kyoichi-sugahara, line 476:
```suggestion
```

planning/behavior_path_start_planner_module/src/start_planner_module.cpp
```diff
@@ -416,25 +417,34 @@ bool StartPlannerModule::isPreventingRearVehicleFromPassingThrough() const
       const lanelet::ConstLineString2d current_lane_bound = (ego_is_merging_from_the_left)
                                                               ? closest_lanelet_const.rightBound2d()
                                                               : closest_lanelet_const.leftBound2d();
+      const lanelet::ConstLineString2d opposite_lane_bound =
```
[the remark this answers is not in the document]
danielsanchezaran: in this case, the opposite lane bound is the lane to the opposite side of the lane we are checking the gap against.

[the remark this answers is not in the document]
kyoichi-sugahara: I think your way of using opposite is reasonable.

But I am not entirely sure if we have a common understanding, but "opposite lane" refers to the lane traveling in the opposite direction. Therefore, I am concerned that using this term may cause misunderstanding.

If you want to express the lane bound on the opposite side of the current lane bound, how about the following:

- adjacent_lane_bound
- other_side_lane_bound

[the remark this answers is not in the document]
danielsanchezaran: I will use other_side_lane_bound! thanks for the suggestion!

planning/behavior_path_start_planner_module/src/start_planner_module.cpp
```diff
@@ -398,11 +398,12 @@ bool StartPlannerModule::isPreventingRearVehicleFromPassingThrough() const
   auto get_gap_between_ego_and_lane_border =
     [&](
       geometry_msgs::msg::Pose & ego_overhang_point_as_pose,
-      const bool ego_is_merging_from_the_left) -> std::optional<double> {
+      const bool ego_is_merging_from_the_left) -> std::optional<std::pair<double, double>> {
     const auto local_vehicle_footprint = vehicle_info_.createFootprint();
     const auto vehicle_footprint =
       transformVector(local_vehicle_footprint, tier4_autoware_utils::pose2transform(current_pose));
     double smallest_lateral_gap_between_ego_and_border = std::numeric_limits<double>::max();
+    double corresponding_lateral_gap_with_other_lane_bound;
```
soblin, line 406: clang-tidy says this value is not initialized, right ?

danielsanchezaran: Fixed. Thank you!

PR #6343, opened 2024-02-07
fix(multi_object_tracker): tune parameters for pedestrian tracker

technolojin:
## Description

Low performance of initial pedestrian detection was fixed. There was dependency of pedestrian direction, since the initial value of yaw is 0 and this initialization had too much impact. 

1. Tuning parameters for pedestrian tracker (initial covariance matrix, process noise of turn rate), so that the initial tracking can have larger Kalman gain in the beginning.

* initial state covariance
p0_stddev_x         1.0 -> 2.0    // positional standard deviation[m]
p0_stddev_y         1.0 -> 2.0    // positional standard deviation[m]
p0_stddev_yaw     1000           // yaw from the ground, not changed [deg]
p0_stddev_vx        5 -> 120     // velocity standard deviation [km/h]
p0_stddev_wz       10 -> 360   // turn rate standard deviation [deg/s]

* process noise
q_stddev_wz       20 -> 30   // turn rate changes [deg/(s*s)]

2. Refactoring comments and line orders to align trackers

## Tests performed

Tested on the `planning_simulator`.

- Original

https://github.com/autowarefoundation/autoware.universe/assets/42434141/bd18491b-3cc5-4247-be07-02a97388eb3e

- Modified

https://github.com/autowarefoundation/autoware.universe/assets/42434141/41f3f452-84c4-4fae-9c79-6369eca6a0c5

## Effects on system behavior

Not applicable.

## Pre-review checklist for the PR author

The PR author **must** check the checkboxes below when creating the PR.

- [x] I've confirmed the [contribution guidelines].
- [x] The PR follows the [pull request guidelines].

## In-review checklist for the PR reviewers

The PR reviewers **must** check the checkboxes below before approval.

- [ ] The PR follows the [pull request guidelines].

## Post-review checklist for the PR author

The PR author **must** check the checkboxes below before merging.

- [ ] There are no open discussions or they are tracked via tickets.

After all checkboxes are checked, anyone who has write access can merge the PR.

[contribution guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/
[pull request guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/pull-request-guidelines/

PR #6905, opened 2024-04-30
feat(map_based_prediction): incorporate crosswalk user history 

dkoldaev:
## Description

This PR incorporates crosswalk user history to the map based prediction node. The history is useful when detection is unstable and crosswalk user disappear from the detection for some short time. Since current PR, a user that disappeared from the detection for times shorter than the history timeout will get a path based on the latest available position. Additionally, when a crosswalk user is detected in with multiple different IDs it is reasonable to try matching those IDs. This PR introduces simple matching that is based on the position of the user. 

## Related links

[TIER IV INTERNAL LINK](https://star4.slack.com/archives/C03S84LDJGG/p1713766763852949)

## Tests performed

Tested on TIER IV internal data 

https://evaluation.tier4.jp/evaluation/reports/08d0336a-8a00-59ad-984c-850352428143?project_id=prd_jt

## Notes for reviewers

NA

## Interface changes

NA

## Effects on system behavior

The new functionality can be disabled with newly introduced parameters

### DIFF
--- a/perception/map_based_prediction/src/map_based_prediction_node.cpp
+++ b/perception/map_based_prediction/src/map_based_prediction_node.cpp
@@ -612,12 +612,10 @@ std::unordered_set<std::string> removeOldObjectsHistory(
     // Delete old information
     while (!object_data.empty()) {
       const double post_object_time = rclcpp::Time(object_data.front().header.stamp).seconds();
-      if (current_time - post_object_time > buffer_time) {
-        // Delete Old Position
-        object_data.pop_front();
-      } else {
+      if (current_time - post_object_time <= buffer_time) {
         break;
       }
+      object_data.pop_front();
     }
 
     if (object_data.empty()) {


perception/map_based_prediction/src/map_based_prediction_node.cpp
```diff
@@ -926,6 +989,19 @@ void MapBasedPredictionNode::objectsCallback(const TrackedObjects::ConstSharedPt
...
 
+  // get current crosswalk users for later prediction
+  std::unordered_map<std::string, TrackedObject> current_crosswalk_users;
+  for (const auto & object : in_objects->objects) {
+    const auto label_for_prediction =
+      changeLabelForPrediction(object.classification.front().label, object, lanelet_map_ptr_);
+    if (
+      label_for_prediction == ObjectClassification::PEDESTRIAN ||
+      label_for_prediction == ObjectClassification::BICYCLE) {
+      current_crosswalk_users[tier4_autoware_utils::toHexString(object.object_id)] = object;
```
danielsanchezaran, line 1000: I prefer creating a temp variable:
const auto object_id = tier4_autoware_utils::toHexString(object.object_id)
Also, please use the unordered_maps' insert method to insert the object into the map

perception/map_based_prediction/src/map_based_prediction_node.cpp
```diff
@@ -1343,49 +1496,9 @@ void MapBasedPredictionNode::updateObjectData(TrackedObject & object)
   return;
 }
 
-void MapBasedPredictionNode::removeOldObjectsHistory(
-  const double current_time, const TrackedObjects::ConstSharedPtr in_objects)
+void MapBasedPredictionNode::cleanupOldStoppedOnGreenTimes(
```
danielsanchezaran, line 1499: What does this name mean??

perception/map_based_prediction/src/map_based_prediction_node.cpp
```diff
@@ -1123,6 +1218,64 @@ void MapBasedPredictionNode::objectsCallback(const TrackedObjects::ConstSharedPt
...
+  CrosswalkUserData crosswalk_user_data;
+  crosswalk_user_data.header = header;
+  crosswalk_user_data.tracked_object = object;
+  if (crosswalk_users_history_.count(object_id) == 0) {
+    crosswalk_users_history_.emplace(object_id, std::deque<CrosswalkUserData>{crosswalk_user_data});
+  } else {
+    crosswalk_users_history_.at(object_id).push_back(crosswalk_user_data);
+  }
+}
+
```
danielsanchezaran, line 1233:
```suggestion
  const CrosswalkUserData crosswalk_user_data = [&](){
  CrosswalkUserData crosswalk_user_data;
  crosswalk_user_data.header = header;
  crosswalk_user_data.tracked_object = object;
  if (crosswalk_users_history_.count(object_id) == 0) {
    crosswalk_users_history_.emplace(object_id, std::deque<CrosswalkUserData>{crosswalk_user_data});
    return  crosswalk_user_data;
  } 
    crosswalk_users_history_.at(object_id).push_back(crosswalk_user_data);
  return CrosswalkUserData crosswalk_user_data;
}

```

perception/map_based_prediction/src/map_based_prediction_node.cpp
```diff
@@ -1123,6 +1218,64 @@ void MapBasedPredictionNode::objectsCallback(const TrackedObjects::ConstSharedPt
...
+    std::string match_id = known_matches_[object_id];
+    // object in the history is already matched to something (possibly itself)
+    if (crosswalk_users_history_.count(match_id)) {
+      // avoid matching two appeared users to one user in history
+      current_users[match_id] = crosswalk_users_history_[match_id].back().tracked_object;
+      return match_id;
+    } else {
+      RCLCPP_WARN_STREAM(
+        get_logger(), "Crosswalk user was " << object_id << "was matched to " << match_id
+                                            << " but the history has deleted. Rematching");
```
danielsanchezaran, line 1247:
```suggestion
                                            << " but it was deleted from object history. Rematching");
```

danielsanchezaran, line 1247: something like this is cleaner:
```suggestion
const auto match_id_opt = [&]()->std::optional<std::string>{
  if (!known_matches_.count(object_id)) return std::nullopt;
     
    std::string match_id = known_matches_[object_id];
    // object in the history is already matched to something (possibly itself)
    if (crosswalk_users_history_.count(match_id) == 0) {
     RCLCPP_WARN_STREAM(
        get_logger(), "Crosswalk user was " << object_id << "was matched to " << match_id 
        return std::nullopt;}
      // avoid matching two appeared users to one user in history
      current_users[match_id] = crosswalk_users_history_[match_id].back().tracked_object;
      return match_id;
    }
if (  match_id_opt.has_value()) {return  match_id.value()}  
```

perception/map_based_prediction/src/map_based_prediction_node.cpp
```diff
@@ -1123,6 +1218,64 @@ void MapBasedPredictionNode::objectsCallback(const TrackedObjects::ConstSharedPt
...
+    }
+    // TODO(dkoldaev): implement more sophisticated scoring, for now simply dst to last position in
+    // history
+    auto pos = user_history.back().tracked_object.kinematics.pose_with_covariance.pose.position;
+    double score = std::hypot(pos.x - pos0.x, pos.y - pos0.y);
+    if (score < best_score) {
+      best_score = score;
+      match_id = user_id;
+    }
+  }
```
danielsanchezaran, line 1266: Please add const to non changing variables like pos and score. Also, please use more descriptive names, whose position are we talking about?

perception/map_based_prediction/src/map_based_prediction_node.cpp
```diff
@@ -585,6 +585,54 @@ bool hasPotentialToReach(
...
+    const double latest_object_time = rclcpp::Time(object_data.back().header.stamp).seconds();
+
+    // Delete Old Objects
+    if (current_time - latest_object_time > buffer_time) {
+      invalid_object_id.insert(object_id);
+      continue;
+    }
+
+    // Delete old information
+    while (!object_data.empty()) {
```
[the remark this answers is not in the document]
dkoldaev: this is not a new function, but a slightly modified version of `removeOldObjectsHistory` to accommodate crosswalk user history. I agree that it will look better with remove_if, but I think it is not a good idea to make refactoring while adding features.

perception/map_based_prediction/src/map_based_prediction_node.cpp
dkoldaev: do you mean use RAII to avoid nested conditionals? I think something like `known_match_ptr` would be a good name here! thank you

dkoldaev: Hm maybe "but history for the crosswalk user was deleted"...

perception/map_based_prediction/src/map_based_prediction_node.cpp
dkoldaev: Could you elaborate why this change please? does not seem to be equivalent to the original code...

perception/map_based_prediction/src/map_based_prediction_node.cpp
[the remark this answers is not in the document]
danielsanchezaran: @dkoldaev but these are new added lines, so it is not a refactoring of old code? Even so, function logic is not altered so I really suggest you do it on this PR

perception/map_based_prediction/src/map_based_prediction_node.cpp
danielsanchezaran: Sorry, I got confused with the intention of this function, The current implementation is fine, but I recommend inserting an early return on the if condition and getting rid of the else and its brackets, they make the code harder to read.

perception/map_based_prediction/src/map_based_prediction_node.cpp
danielsanchezaran: Well, its an std::optional, not a pointer, so _opt is the correct suffix. It does behave very similarly to a ptr, but the optional is used to specify that you may or may not get a value out of your function. 

The main goal is to avoid nested conditionals as much as possible as you mentioned.

dkoldaev: Oh sorry didnt mean ptr there, but the comment makes sense . I’ll update the code

perception/map_based_prediction/config/map_based_prediction.param.yaml
```diff
@@ -27,6 +27,9 @@
       timeout_set_for_no_intention_to_walk: [1.0, 0.0] # [s] values of map
     # parameter for shoulder lane prediction
     prediction_time_horizon_rate_for_validate_shoulder_lane_length: 0.8
+    use_crosswalk_user_history:
```
YoshiRi, line 30: I do not want these parameter to be `true` in default. 
- These association and object keeping process are handled within some tracking processes. I think post processing in prediction should be a option to prevent confusion.
- FYI: After update in tracker process, [PR6343](https://github.com/autowarefoundation/autoware.universe/pull/6343) for example, pedestrian tracking will much stable than older version. Centerpoint and yolo model update will also improve stability.

perception/map_based_prediction/src/map_based_prediction_node.cpp
dkoldaev: Not the best agree, but it actually cleanups the dictionary with such name. I’ll think of a better name

perception/map_based_prediction/src/map_based_prediction_node.cpp
dkoldaev: any reason to use insert and not `emplace` here? I can see that everywhere else in the code emplace is used

perception/map_based_prediction/config/map_based_prediction.param.yaml
dkoldaev: I agree to switch it off by default, because current stabilization process is not very elaborate. But in general I think map based prediction should not rely on perfect tracking. There should be multiple layers of protection in these safety critical flows such as pedestrian avoidance

perception/map_based_prediction/include/map_based_prediction/map_based_prediction_node.hpp
```diff
@@ -137,6 +144,8 @@ class MapBasedPredictionNode : public rclcpp::Node
   // Object History
   std::unordered_map<std::string, std::deque<ObjectData>> objects_history_;
```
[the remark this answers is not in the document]
dkoldaev: this makes sense to me. I think it is a good idea to make the change in this PR to make things clear.

YoshiRi (approved):
LGTM

perception/map_based_prediction/src/map_based_prediction_node.cpp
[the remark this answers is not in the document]
dkoldaev: @danielsanchezaran I also am not sure how to implement an early return in case of remove if. The predicate is bool, so it will go through all of the iterator list as I understand. The early return is possible, because of while loop and sorted structure of the object data... I would keep it as is for now and address it in a separate PR if you want to improve here

perception/map_based_prediction/src/map_based_prediction_node.cpp
dkoldaev: no problem. I reduced the conditional here by removing else statement

perception/map_based_prediction/src/map_based_prediction_node.cpp
```diff
@@ -585,6 +585,54 @@ bool hasPotentialToReach(
...
+      const double post_object_time = rclcpp::Time(object_data.front().header.stamp).seconds();
+      if (current_time - post_object_time > buffer_time) {
+        // Delete Old Position
+        object_data.pop_front();
+      } else {
+        break;
+      }
+    }
```
danielsanchezaran, line 621: I mean something like this: which is more readable and will not significantly affect performance
```suggestion
    std::remove_if(object_data.begin(),object_data.end(),[&](const auto & obj){
    const double post_object_time = rclcpp::Time(obj.header.stamp).seconds();
    return  (current_time - post_object_time > buffer_time);
    });
```
You could also do this: 
```suggestion
    while (!object_data.empty()) {
      const double post_object_time = rclcpp::Time(object_data.front().header.stamp).seconds();
      if (current_time - post_object_time <= buffer_time) {
        break;
      }
      object_data.pop_front();
    }
```
But the first suggestion is better imo.

perception/map_based_prediction/src/map_based_prediction_node.cpp
danielsanchezaran: emplace is also good @dkoldaev

perception/map_based_prediction/src/map_based_prediction_node.cpp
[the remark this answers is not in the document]
danielsanchezaran: Please see my latest comment on the issue

perception/map_based_prediction/src/map_based_prediction_node.cpp
dkoldaev: Thank you! But as I understand the first suggestion does not have early exit, it will keep iterating through the list getting bool value for all of the objects. I will go with the second solution.

danielsanchezaran: @dkoldaev yes, it does not have but the processing cost is low, I believe the first option is more readable and the performance cost is not significant.

soblin (approved):

PR #6911, opened 2024-05-01
fix(avoidance): unexpected road shoulder distance calculation result

satoshi-ota:
## Description

[TIER IV INTERNAL LINK](https://tier4.atlassian.net/browse/RT1-6050)

Related PR: https://github.com/autowarefoundation/autoware.universe/pull/6781

The module searches most overhanging point by following steps:

1. pick a point from envelope polygon contour (P1).
2. create new point from P1 by using `calcOffsetPose` (P2).
3. search intersection point between P1-P2 and drivable bound.
4. calculate distance between intersection point and P1.
5. _apply step1~4 for all polygon points._
6. sort point by distance calculated in step4.
7. pick the point whose distance is smallest from sorted points as most overhanging point.

```c++
const auto p2 = calcOffsetPose(p_tmp, 0.0, (isOnRight(object) ? -1.0 : 1.0), 0.0).position;
```

Previously, the point (P2) is calculated with constant offset value. **However, I found sometimes it couldn't find intersection point for some polygon points when the distance between P1 and drivable bound was larger than 1.0m.**

In following case, it couldn't find intersection point for bottom left/right points on white polygon. As a result, it got wrong point as the most overhanging point and created infeasible path.

Screenshot from 2024-05-01 13-28-31

In this PR, I fixed the logic for step2 to decide offset value based on polygon width so that it always can find intersection point for all polygon points.

```c++
const auto envelope_polygon_width =
        boost::geometry::area(object.envelope_poly) / object.length;

      {
        const auto p2 =
          calcOffsetPose(p_tmp, 0.0, (isOnRight(object) ? -0.5 : 0.5) * envelope_polygon_width, 0.0)
            .position;
...
```

## Tests performed

Overhang distance is calculated expectedly. And it doesn't create infeasible path.

image

## Effects on system behavior

Nothing.

### DIFF
--- a/planning/behavior_path_avoidance_module/src/utils.cpp
+++ b/planning/behavior_path_avoidance_module/src/utils.cpp
@@ -932,7 +932,8 @@ double getRoadShoulderDistance(
       }
 
       const auto envelope_polygon_width =
-        boost::geometry::area(object.envelope_poly) / object.length;
+        boost::geometry::area(object.envelope_poly) /
+        std::max(object.length, 1e-3);  // prevent division by zero
 
       {
         const auto p2 =
--- a/planning/behavior_path_avoidance_module/src/utils.cpp
+++ b/planning/behavior_path_avoidance_module/src/utils.cpp
@@ -931,8 +931,13 @@ double getRoadShoulderDistance(
         }
       }
 
+      const auto envelope_polygon_width =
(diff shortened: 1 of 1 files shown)


shmpwk (approved):
LGTM

PR #6912, opened 2024-05-01
feat(autonomous_emergency_braking): add obstacle velocity estimation for AEB

danielsanchezaran:
## Description

This PR introduces obstacle velocity estimation to the AEB module. 
Currently, the AEB module uses RSS distance to determine if a collision between ego and the obstacle will collide or not, but the object speed is set to 0 by default. This sometimes causes the ego vehicle to issue a stop when it gets close to an object in front that is moving forward.

This PR uses detected object history and the change of object position between point cloud frames to calculate the object's velocity along the ego path's longitudinal axis. Furthermore the velocity is added to a queue of "object velocities" and the median value is chosen as the current object speed (That is done to account for possible jumps of velocity due to false detection).

This calculation is necessary to correctly compute the RSS distance that AEB uses to determine if a collision will happen, currently AEB just assigns 0 as the object speed. This PR fixes that

Additionally this PR adds debug markers to show the closest object velocity and velocity relative to ego.

## Related links

Launch required changes: https://github.com/autowarefoundation/autoware_launch/pull/978

## Tests performed

PSim

Comparison: 

Before: AEB sends an error (emergency stop) when we suddenly place a car running at 10 m/s in front of ego even though the NPC vehicle is moving faster than the ego and the risk of not being able to slow down on time is not there. You can see the error being sent on the rqt_robot_monitor screen

https://github.com/autowarefoundation/autoware.universe/assets/25967964/74acf839-9a41-4adc-9de0-40408409217b

With this PR: same situation, but the AEB module correctly judges the situation to be safe because it calculates the object speed to be around 10 m/s on the longitudinal axis of the ego path and there is no emergency stop issued:

https://github.com/autowarefoundation/autoware.universe/assets/25967964/3432bcda-2498-4026-aae7-16484cfb034c

## Notes for reviewers

Requires changes to Launch: https://github.com/autowarefoundation/autoware_launch/pull/978

## Effects on system behavior

By estimating object speed, false positive stops will happen less often.

### DIFF
--- a/control/autonomous_emergency_braking/README.md
+++ b/control/autonomous_emergency_braking/README.md
@@ -108,6 +108,33 @@ After Noise filtering, it performs a geometric collision check to determine whet
 
 Finally, the vertex that is closest to the ego vehicle is chosen as the candidate for collision checking: Since rss distance is used to judge if a collision will happen or not, if the closest vertex to the ego is deemed to be safe, the rest of the vertices (and the points in the clusters) will also be safe.
 
+#### Obstacle velocity estimation
+
+Once the position of the closest obstacle/point is determined, the AEB modules uses the history of previously detected objects to estimate the closest object speed using the following equations:
+
+$$
+d_{t} = o_{time stamp} - prev_{time stamp}
+$$
+
+$$
+d_{pos} = norm(o_{pos} - prev_{pos})
+$$
+
+$$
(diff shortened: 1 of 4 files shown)


soblin (approved):
Actually, you do not need lock_guard(mutex_) because the callbacks of this node runs in a single thread.

PR #6957, opened 2024-05-09
fix(componet_state_monitor): remove ndt node alive monitoring

YamatoAndo:
## Description

To monitor the health of the NDT,
it is not appropriate to monitor `/localizaiton/pose_estimator/pose_with_covariance` , and it often leads to false detections. Therefore, I will remove this monitoring.

If NDT health monitoring is still necessary, I will consider alternative methods.

## Tests performed

LSim works.

## Effects on system behavior

N/A

## Interface changes

N/A

### DIFF
--- a/system/component_state_monitor/config/topics.yaml
+++ b/system/component_state_monitor/config/topics.yaml
@@ -50,19 +50,6 @@
     error_rate: 1.0
     timeout: 1.0
 
-- module: localization
-  mode: [online, logging_simulation]
-  type: autonomous
-  args:
-    node_name_suffix: pose_estimator_pose
-    topic: /localization/pose_estimator/pose_with_covariance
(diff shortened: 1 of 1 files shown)


isamu-takagi:
@YamatoAndo Could you please explain in more detail? I expected that NDT scan matching would publish the pose topic periodically, but is there a case where this is not the case?

YamatoAndo:
>  I expected that NDT scan matching would publish the pose topic periodically, but is there a case where this is not the case?

For example, NDT Scan matching does not publish topics when the score is decreasing.

There are several other validations as well, and if it is determined that scan matching is not being performed successfully, it will not publish topics.

isamu-takagi (approved):
@YamatoAndo Thank you. That makes sense. LGTM.

PR #6960, opened 2024-05-09
fix(bvp): don't use numeric limits for start/finish distance

satoshi-ota:
## Description

- Remove redundant process calling `updateRTCStatus`.
- Don't use `std::numeric_limits<double>::lowest()` for start/finish distance in order to avoid ros2 command crashing. https://github.com/ros2/rclpy/issues/848

## Tests performed

- [x] [PASS TIER IV INTERNAL SCENARIOS](https://evaluation.tier4.jp/evaluation/reports/e8201714-b1df-5d2c-85e8-17695c9eb301?project_id=prd_jt)

## Effects on system behavior

Nothing.

### DIFF
--- a/planning/behavior_velocity_blind_spot_module/src/manager.cpp
+++ b/planning/behavior_velocity_blind_spot_module/src/manager.cpp
@@ -72,8 +72,6 @@ void BlindSpotModuleManager::launchNewModules(
       module_id, lane_id, planner_data_, planner_param_, logger_.get_child("blind_spot_module"),
       clock_));
     generateUUID(module_id);
-    updateRTCStatus(
-      getUUID(module_id), true, std::numeric_limits<double>::lowest(), path.header.stamp);
   }
 }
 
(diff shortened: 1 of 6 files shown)


planning/behavior_velocity_blind_spot_module/src/manager.cpp
```diff
@@ -72,8 +72,6 @@ void BlindSpotModuleManager::launchNewModules(
       module_id, lane_id, planner_data_, planner_param_, logger_.get_child("blind_spot_module"),
       clock_));
     generateUUID(module_id);
-    updateRTCStatus(
-      getUUID(module_id), true, std::numeric_limits<double>::lowest(), path.header.stamp);
```
rej55, line 76: If this line is removed, RTC status does not contain this information.
To check the existence of a module with `module_id`, status should contain this information, so I think this information including distance as `-10000.0` is required.

satoshi-ota:
@rej55 Thanks. I think there is no problem since `updateRTCStatus` is always called in `sendRTC` function per every planning cycle.

On the other hand, I found it was able to echo topics that include nan or inf. Then, I close this PR.

PR #7004, opened 2024-05-14
fix(start_planner): issue when ego does not straddle lane bounds and starts from 0 speed

danielsanchezaran:
## Description

When the ego vehicle is starting from a road and not the shoulder lane, it is possible that the blinkers are not properly activated. 
This is caused by 2 things:

1.  The check for lane bound straddling (crossing a lane bound) is failed. This check is not necessary when doing a pull out because we should turn on the blinkers anyways.
2. The shift is small and the ego is stopped. This happens because the ego is starting from complete stop so its speed is close to 0 and if the shift pull out lane is small/has a small shift, the blinkers are ignored, which is not desired in this case. 

To solve the issue I added an override for the first check if the ego is doing a pullout, and for the second issue I slightly modified the variable  `status_.first_engaged_and_driving_forward_time` to only be initialized when the ego vehicle has surpassed a minimum threshold velocity.

## Related links
Ticket: [TIER IV COMPANY INTERNAL LINK](https://tier4.atlassian.net/browse/RT0-31321)

## Tests performed
PSim

https://github.com/autowarefoundation/autoware.universe/assets/25967964/169ba8c2-ca75-4b93-b614-b5f1f4576d5f

### DIFF
--- a/planning/behavior_path_start_planner_module/src/start_planner_module.cpp
+++ b/planning/behavior_path_start_planner_module/src/start_planner_module.cpp
@@ -241,9 +241,12 @@ void StartPlannerModule::updateData()
     DEBUG_PRINT("StartPlannerModule::updateData() received new route, reset status");
   }
 
+  constexpr double moving_velocity_threshold = 0.1;
+  const double & ego_velocity = planner_data_->self_odometry->twist.twist.linear.x;
   if (
     planner_data_->operation_mode->mode == OperationModeState::AUTONOMOUS &&
-    status_.driving_forward && !status_.first_engaged_and_driving_forward_time) {
+    status_.driving_forward && !status_.first_engaged_and_driving_forward_time &&
+    ego_velocity > moving_velocity_threshold) {
     status_.first_engaged_and_driving_forward_time = clock_->now();
   }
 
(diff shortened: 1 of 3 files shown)


planning/behavior_path_start_planner_module/src/start_planner_module.cpp
```diff
@@ -1265,8 +1268,12 @@ TurnSignalInfo StartPlannerModule::calcTurnSignalInfo()
   // In Geometric pull out, the ego stops once and then steers the wheels to the opposite direction.
   // This sometimes causes the getBehaviorTurnSignalInfo method to detect the ego as stopped and
   // close to complete its shift, so it wrongly turns off the blinkers, this override helps avoid
-  // this issue.
+  // this issue. Also, if the ego is not engaged (so it is stopped), the blinkers should still be
+  // activated.
   const bool override_ego_stopped_check = std::invoke([&]() {
+    if (!status_.first_engaged_and_driving_forward_time) {
```
soblin, line 1274: and change this condition to `needToPrepareBlinkerBeforeStratDrivingForward()` ?

danielsanchezaran: @soblin I cannot check the current speed, I need a check that tells me if at some point in the PAST the ego has moved faster than the threshold speed. So this approach cannot work.

soblin (approved):

PR #6902, opened 2024-04-26
feat(image projection based fusion): unrecify 3d point for image projection based fusion

yukkysaito:
## Description

In image_projection_based_fusion, only the internal parameters of the camera were considered, and no distortion correction was applied. As a result, when projecting onto images with significant distortion, misalignments occurred. This Pull Request has been modified to account for distortion.

## Tests performed

In roi_cluster_fusion, operation has been confirmed on the actual machine.
before
image

after
image

## Pre-review checklist for the PR author

The PR author **must** check the checkboxes below when creating the PR.

- [ ] I've confirmed the [contribution guidelines].
- [ ] The PR follows the [pull request guidelines].

## In-review checklist for the PR reviewers

The PR reviewers **must** check the checkboxes below before approval.

- [ ] The PR follows the [pull request guidelines].
- [ ] The PR has been properly tested.
- [ ] The PR has been reviewed by the code owners.

## Post-review checklist for the PR author

The PR author **must** check the checkboxes below before merging.

- [ ] There are no open discussions or they are tracked via tickets.
- [ ] The PR is ready for merge.

After all checkboxes are checked, anyone who has write access can merge the PR.

[contribution guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/
[pull request guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/pull-request-guidelines/

PR #7030, opened 2024-05-15
feat: unrecify image projection based fusion

yukkysaito:
## Description

Fixed this [PR](https://github.com/autowarefoundation/autoware.universe/pull/6902)

In image_projection_based_fusion, only the internal parameters of the camera were considered, and no distortion correction was applied. As a result, when projecting onto images with significant distortion, misalignments occurred. This Pull Request has been modified to account for distortion.

## Tests performed

In roi_cluster_fusion, operation has been confirmed on the actual machine.
before
image

after
image


perception/image_projection_based_fusion/src/utils/utils.cpp
```diff
@@ -13,8 +13,33 @@
...
+
+  cv::Point2d raw_image_point = rectified_image_point;
+
+  try {
+    bool need_unrectify = false;
+    for (const auto & distortion : pinhole_camera_model.cameraInfo().d) {
+      if (distortion != 0.0) {
+        need_unrectify = true;
+      }
+    }
```
kminoda, line 32:
```suggestion
  const bool need_unrectify = std::any_of(pinhole_camera_model.cameraInfo().d.begin(),
    pinhole_camera_model.cameraInfo().d.end(),
    [](double distortion) { return distortion != 0.0; });
```
How about this? It would be easier to read, and also will enable making `need_rectify` a const variable.

perception/image_projection_based_fusion/src/utils/utils.cpp
```diff
@@ -13,8 +13,33 @@
...
 namespace image_projection_based_fusion
 {
+Eigen::Vector2d calcRawImageProjectedPoint(
+  const image_geometry::PinholeCameraModel & pinhole_camera_model, const cv::Point3d & point3d)
+{
+  cv::Point2d rectified_image_point = pinhole_camera_model.project3dToPixel(point3d);
+
+  cv::Point2d raw_image_point = rectified_image_point;
+
+  try {
```
kminoda, line 26: What is this try sentence for?

kminoda (commented):
Thank you for the PR :pray:

perception/image_projection_based_fusion/src/utils/utils.cpp
yukkysaito: The code for determining whether to unrectify in the library of image_geometry seems to change occasionally. If it can't be done, it is thrown, so I want to be able to catch it.

perception/image_projection_based_fusion/src/utils/utils.cpp
yukkysaito: [a86ed7b](https://github.com/autowarefoundation/autoware.universe/pull/7030/commits/a86ed7b1088b6bdd884daa21e519afc6ef0c9c0e)

perception/image_projection_based_fusion/src/utils/utils.cpp
kminoda: Hmm, understood. It is a difficult decision, but IMHO we should avoid catching this error and just logging warning without killing the node, since it will be difficult for developers to be aware of a breaking change or a bug in `image_geometry` library, increasing a risk of a potential degradation of Autoware. What do you think?

yukkysaito: @kminoda So, what exactly do you want to do?

perception/image_projection_based_fusion/src/utils/utils.cpp
```diff
@@ -13,8 +13,30 @@
...
+    const bool need_unrectify = std::any_of(
+      pinhole_camera_model.cameraInfo().d.begin(), pinhole_camera_model.cameraInfo().d.end(),
+      [](double distortion) { return distortion != 0.0; });
+
+    if (need_unrectify) {
+      raw_image_point = pinhole_camera_model.unrectifyPoint(rectified_image_point);
+    }
+  } catch (cv::Exception & e) {
+    RCLCPP_WARN_STREAM(rclcpp::get_logger("image_projection_based_fusion"), e.what());
+  }
```
kminoda, line 36:
```suggestion
  const bool need_unrectify = std::any_of(
    pinhole_camera_model.cameraInfo().d.begin(), pinhole_camera_model.cameraInfo().d.end(),
    [](double distortion) { return distortion != 0.0; });

  if (need_unrectify) {
    raw_image_point = pinhole_camera_model.unrectifyPoint(rectified_image_point);
  }
```
Regarding https://github.com/autowarefoundation/autoware.universe/pull/7030/files#r1604255216, here's my proposal.

perception/image_projection_based_fusion/src/utils/utils.cpp
kminoda: Please see here https://github.com/autowarefoundation/autoware.universe/pull/7030#discussion_r1604267322

yukkysaito: @kminoda This time, when d is 0 in the image_geometry class, it is determined that the image is not calibrated, and thus there is no need to unrectify, resulting in an exception being thrown. This determination of being uncalibrated seems to change frequently in the code. Even if an exception is thrown, there is no need to terminate the process since it only means that unrectifying is not possible. Wouldn't it be better to skip this with a try-catch block?

- https://github.com/ros-perception/vision_opencv/blob/humble/image_geometry/src/pinhole_camera_model.cpp#L157-L160

- https://docs.ros.org/en/diamondback/api/image_geometry/html/c++/pinhole__camera__model_8cpp_source.html#l00328

yukkysaito: I will close this PR as there seems to have been a misunderstanding on my part.

PR #7034, opened 2024-05-16
perf(route_handler): avoid unnecessary copies and heavy loops

maxime-clem:
## Description

This PR improves the performance of a few functions of the `RouteHandler` class.
Behavior should be unchanged.

## Tests performed

Psim.
Evaluator (TIER IV INTERNAL LINK): https://evaluation.tier4.jp/evaluation/reports/b47197c0-55ad-56fb-9e4f-7da9c4b9e806?project_id=prd_jt

## Effects on system behavior

Not applicable.

### DIFF
--- a/planning/route_handler/src/route_handler.cpp
+++ b/planning/route_handler/src/route_handler.cpp
@@ -573,8 +573,9 @@ lanelet::ConstLanelets RouteHandler::getLaneletSequenceUpTo(
 
   lanelet::ConstLanelet current_lanelet = lanelet;
   double length = 0;
+  lanelet::ConstLanelets candidate_lanelets;
   while (rclcpp::ok() && length < min_length) {
-    lanelet::ConstLanelets candidate_lanelets;
+    candidate_lanelets.clear();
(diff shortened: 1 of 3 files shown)


planning/route_handler/src/route_handler.cpp
```diff
@@ -573,44 +573,26 @@ lanelet::ConstLanelets RouteHandler::getLaneletSequenceUpTo(
 
   lanelet::ConstLanelet current_lanelet = lanelet;
```
danielsanchezaran, line 574: not sure how many copies occur normally, but wouldn't it be better to have this as a pointer? so the copy in line 574 is not that expensive?

maxime-clem: I do not see an easy way to use pointers in this case. In the loop we have a temporary `prev_lanelet` that will need to be copied (or moved) anyway.
From my profiling these copies have negligible impact on performance so I think it is okay to ignore them.

zulfaqar-azmi-t4 (approved):

PR #7159, opened 2024-05-29
fix(multi_object_tracker): prevent too large object tracking

technolojin:
## Description
A bug was reported that a huge (length of 6km) object is appeared from the object tracking.

It is happening when a unknown object (shape of `autoware_auto_perception_msgs::msg::Shape::POLYGON`) is assigned to a bicycle tracker.
It was caused by `utils::convertConvexHullToBoundingBox(object, bbox_object)` which treats the footprint points as global (map coordinate) position, while the input is local (the object coordinate).

1. Fix bug on `utils::convertConvexHullToBoundingBox(object, bbox_object)`
2. Add a detected object size checker. Update the object size only if the size is proper

## Tests performed
1. Configure `data_association_matrix.param.yaml` to enable the unknown object can be assigned to bicycle tracker
```yaml
/**:
  ros__parameters:
    can_assign_matrix:
      #UNKNOWN, CAR, TRUCK, BUS,  TRAILER, MOTORBIKE, BICYCLE,PEDESTRIAN <-Measurement
      [1,       0,   0,     0,    0,       0,         0,      0,         #UNKNOWN <-Tracker
       0,       1,   1,     1,    1,       0,         0,      0,         #CAR
       0,       1,   1,     1,    1,       0,         0,      0,         #TRUCK
       0,       1,   1,     1,    1,       0,         0,      0,         #BUS
       0,       1,   1,     1,    1,       0,         0,      0,         #TRAILER
       1,       0,   0,     0,    0,       1,         1,      1,         #MOTORBIKE
       1,       0,   0,     0,    0,       1,         1,      1,         #BICYCLE
       1,       0,   0,     0,    0,       1,         1,      1]         #PEDESTRIAN
```
Replay a file that contains to associate an object that have poligon shape to a tracker. 

https://github.com/autowarefoundation/autoware.universe/assets/42434141/89a3ab5a-1c1f-40fb-bceb-b9d885507fae

Red box: previous tracking object
Green box: fixed tracking object

2. Perception Regular Test
test result: [TIER IV INTERNAL](https://evaluation.tier4.jp/evaluation/reports/5f5ca3c5-8d96-558f-948b-c13b4850e91f?project_id=prd_jt)
261/264 PASS
! This test shows that the reference mode can be used as before, not testing the problem is resolved (since this scene is not in the test)

## Effects on system behavior
Not applicable.

## Interface changes
Not applicable.

### DIFF
--- a/perception/multi_object_tracker/include/multi_object_tracker/utils/utils.hpp
+++ b/perception/multi_object_tracker/include/multi_object_tracker/utils/utils.hpp
@@ -299,12 +299,6 @@ inline void convertConvexHullToBoundingBox(
   const autoware_auto_perception_msgs::msg::DetectedObject & input_object,
   autoware_auto_perception_msgs::msg::DetectedObject & output_object)
 {
-  const Eigen::Vector2d center{
-    input_object.kinematics.pose_with_covariance.pose.position.x,
-    input_object.kinematics.pose_with_covariance.pose.position.y};
-  const auto yaw = tf2::getYaw(input_object.kinematics.pose_with_covariance.pose.orientation);
-  const Eigen::Matrix2d R_inv = Eigen::Rotation2Dd(-yaw).toRotationMatrix();
-
   double max_x = 0;
   double max_y = 0;
   double min_x = 0;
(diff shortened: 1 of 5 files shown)


perception/multi_object_tracker/include/multi_object_tracker/utils/utils.hpp
```diff
@@ -295,40 +295,43 @@ inline void calcAnchorPointOffset(
...
-      input_object.shape.footprint.points.at(i).x, input_object.shape.footprint.points.at(i).y};
-
-    const Eigen::Vector2d local_vertex = R_inv * (vertex - center);
-    max_x = std::max(max_x, local_vertex.x());
-    max_y = std::max(max_y, local_vertex.y());
-    min_x = std::min(min_x, local_vertex.x());
-    min_y = std::min(min_y, local_vertex.y());
-
-    max_z = std::max(max_z, static_cast<double>(input_object.shape.footprint.points.at(i).z));
+    const double foot_x = input_object.shape.footprint.points.at(i).x;
```
YoshiRi, line 314: [imho]
I think footprint is calculated by relative distance from centroid in base_link and not in object coordinate.
So, I guess yaw rotation fix will be necessary before get maximum x,y value.

Currently, it won't make any change because yaw of unknown object may be 0 in all case.

technolojin: Base line: the source of the unknown polygon object is from `shape_estimator`.

1. Object Position: 
   * Coordinate origin: base_link
   * Position: polygon centroid 
   * [evidence](https://github.com/autowarefoundation/autoware.universe/blob/9dcf81ec79c7f465c547196148f8c9c40e6d3411/perception/shape_estimation/lib/model/convex_hull.cpp#L88-L89)
2. Footprint:
   * Coordinate origin: object position (polygon centroid)
   * Positions: convex hull point positions, local
   * [evidence](https://github.com/autowarefoundation/autoware.universe/blob/9dcf81ec79c7f465c547196148f8c9c40e6d3411/perception/shape_estimation/lib/model/convex_hull.cpp#L75-L81)

I also checked the object size, output of this function, and confirmed that the numbers are as expected (about 0.7 m).

Therefore the footprint points are not from the base_link, but from the object origin.

technolojin: If the message defines the footprint is in the object coordinate (as its position is from the object position), the width and length should be on its coordinate, not rotated one.

It doesn't matter whether the `shape_estimator` implementation is done in `yaw = 0` or not.
If the `shape_estimator` do the yaw estimation in the polygon&convex hull function, it should rotate the point cloud when it make the message.

YoshiRi: Understood. Then, your code is fine.

YoshiRi (approved):
LGTM

PR #7100, opened 2024-05-23
feat(perception_online_evaluator): add metric_value not only stat

kosuke55:
## Description

Previously, only stat with `max`, `mean`, and `min` as keys could be handled in diagnostics, but this PR allows `metric_value` with a single value.
This is to allow only one value to be included in metrics such as object counts, since all stat values are the same, which is redundant.

- stat type
```
- level: "\0"
  name: predicted_path_deviation_variance_CAR_5.00
  message: ''
  hardware_id: ''
  values:
  - key: min
    value: '0.039821'
  - key: max
    value: '0.039821'
  - key: mean
    value: '0.039821'
```

- metric_value type
```
- level: "\0"
  name: interval_average_objects_count_UNKNOWN_r100.00_h10.00
  message: ''
  hardware_id: ''
  values:
  - key: metric_value
    value: '4.000000'
```

image

## Tests performed

psim

## Effects on system behavior

Not applicable.

## Interface changes

no interface changes for autoware.

but the diagnostic key used for tools (driving_log_replayer etc) is chaged

## Pre-review checklist for the PR author

The PR author **must** check the checkboxes below when creating the PR.

- [x] I've confirmed the [contribution guidelines].
- [x] The PR follows the [pull request guidelines].

## In-review checklist for the PR reviewers

The PR reviewers **must** check the checkboxes below before approval.

- [ ] The PR follows the [pull request guidelines].

## Post-review checklist for the PR author

The PR author **must** check the checkboxes below before merging.

- [ ] There are no open discussions or they are tracked via tickets.

After all checkboxes are checked, anyone who has write access can merge the PR.

[contribution guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/
[pull request guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/pull-request-guidelines/

PR #7167, opened 2024-05-30
fix(perception_online_evaluator):  add metric_value not only stat (#7100)(#7118) (revert of revert)

kosuke55:
this PR need to wait for driving log replayer

Reverts autowarefoundation/autoware.universe#7118

https://github.com/autowarefoundation/autoware.universe/pull/7100

## Description

Previously, only stat with `max`, `mean`, and `min` as keys could be handled in diagnostics, but this PR allows `metric_value` with a single value.
This is to allow only one value to be included in metrics such as object counts, since all stat values are the same, which is redundant.

- stat type
```
- level: "\0"
  name: predicted_path_deviation_variance_CAR_5.00
  message: ''
  hardware_id: ''
  values:
  - key: min
    value: '0.039821'
  - key: max
    value: '0.039821'
  - key: mean
    value: '0.039821'
```

- metric_value type
```
- level: "\0"
  name: interval_average_objects_count_UNKNOWN_r100.00_h10.00
  message: ''
  hardware_id: ''
  values:
  - key: metric_value
    value: '4.000000'
```

image

## Tests performed

psim

## Effects on system behavior

Not applicable.

## Interface changes

no interface changes for autoware.

but the diagnostic key used for tools (driving_log_replayer etc) is chaged

### DIFF
--- a/evaluator/perception_online_evaluator/include/perception_online_evaluator/perception_online_evaluator_node.hpp
+++ b/evaluator/perception_online_evaluator/include/perception_online_evaluator/perception_online_evaluator_node.hpp
@@ -61,6 +61,8 @@ public:
 
   DiagnosticStatus generateDiagnosticStatus(
     const std::string metric, const Stat<double> & metric_stat) const;
+  DiagnosticStatus generateDiagnosticStatus(
+    const std::string metric, const double metric_value) const;
 
 private:
   // Subscribers and publishers
--- a/evaluator/perception_online_evaluator/include/perception_online_evaluator/perception_online_evaluator_node.hpp
+++ b/evaluator/perception_online_evaluator/include/perception_online_evaluator/perception_online_evaluator_node.hpp
@@ -62,7 +62,7 @@ public:
   DiagnosticStatus generateDiagnosticStatus(
     const std::string metric, const Stat<double> & metric_stat) const;
   DiagnosticStatus generateDiagnosticStatus(
-    const std::string metric, const double metric_value) const;
+    const std::string & metric, const double metric_value) const;
 
 private:
   // Subscribers and publishers
--- a/evaluator/perception_online_evaluator/src/perception_online_evaluator_node.cpp
+++ b/evaluator/perception_online_evaluator/src/perception_online_evaluator_node.cpp
@@ -69,16 +69,25 @@ void PerceptionOnlineEvaluatorNode::publishMetrics()
 
   // calculate metrics
   for (const Metric & metric : parameters_->metrics) {
-    const auto metric_stat_map = metrics_calculator_.calculate(Metric(metric));
-    if (!metric_stat_map.has_value()) {
+    const auto metric_result = metrics_calculator_.calculate(Metric(metric));
+    if (!metric_result.has_value()) {
       continue;
     }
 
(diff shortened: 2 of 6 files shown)


evaluator/perception_online_evaluator/include/perception_online_evaluator/perception_online_evaluator_node.hpp
```diff
@@ -61,6 +61,8 @@ class PerceptionOnlineEvaluatorNode : public rclcpp::Node
 
   DiagnosticStatus generateDiagnosticStatus(
     const std::string metric, const Stat<double> & metric_stat) const;
+  DiagnosticStatus generateDiagnosticStatus(
+    const std::string metric, const double metric_value) const;
```
ktro2828, line 65: [NITS] How about using const reference?  
```suggestion
    const std::string & metric, const double metric_value) const;
```

evaluator/perception_online_evaluator/src/perception_online_evaluator_node.cpp
```diff
@@ -69,16 +69,25 @@ void PerceptionOnlineEvaluatorNode::publishMetrics()
...
-        metrics_msg.status.push_back(generateDiagnosticStatus(metric, stat));
-      }
-    }
+    std::visit(
+      [&metrics_msg, this](auto && arg) {
+        using T = std::decay_t<decltype(arg)>;
+        for (const auto & [metric, value] : arg) {
+          if constexpr (std::is_same_v<T, MetricStatMap>) {
+            if (value.count() > 0) {
+              metrics_msg.status.push_back(generateDiagnosticStatus(metric, value));
```
ktro2828, line 83: [NITS] In terms of performance, `emplace_back` is better than `push_back` to construct instance directly.

ktro2828, line 83: [NITS] In terms of the performance, `emplace_back` is better than `push_back` as it constructs an instance directly.

evaluator/perception_online_evaluator/src/perception_online_evaluator_node.cpp
```diff
@@ -111,6 +120,22 @@ DiagnosticStatus PerceptionOnlineEvaluatorNode::generateDiagnosticStatus(
   return status;
 }
 
+DiagnosticStatus PerceptionOnlineEvaluatorNode::generateDiagnosticStatus(
+  const std::string metric, const double value) const
```
kosuke55, line 124:
```suggestion
  const std::string & metric, const double value) const
```

evaluator/perception_online_evaluator/src/perception_online_evaluator_node.cpp
kosuke55: fixed in d4edf4cf035e2c0c816b4eb0cbe072d27a4533d6

hayato-m126 (approved):
LGTM

I confirmed that the following two types of DianosticsStatus are produced by actually running the code.

This PR code works with https://github.com/tier4/driving_log_replayer/pull/467

kyoichi-sugahara (approved):

PR #7256, opened 2024-06-04
feat(ndt_scan_matcher): direct pcd loader for map update module

anhnv3991:
## Description

- The map_update_module updates the map point cloud via a async call to the pointcloud map loader service, but the latency of such a call is significant. This PR added a map loader to the map update module to enable this module to load PCD files directly (bypass a call to the map loader service).
- This PR also adds the parallel load, which loads PCDs files in parallel, and makes the PCD load overlap with the NDT build. This reduces the total cost of updating NDT (load map + rebuild NDT).

## Tests performed

- Map: Odaiba
- Data: xx1_odaiba rosbag
- Result: The attached image shows the total cost of the function update_ndt in milliseconds, when using the old and the new way of loading data.
map_loader_cmp

## Effects on system behavior

Not applicable.

## Interface changes

None

- [] There are no open discussions or they are tracked via tickets.


anhnv3991:
@YamatoAndo I ran some test and collect some numbers. 
In this attached image, I recorded the total number of points loaded from PCDs in every update_ndt() call.
map_loader_pnum
When comparing this image with the image attached in this PR description, we can see that the patterns of the loading time of the old method and the number of points loaded are quite similar. So I guess when calling the map loader service to load PCDs, the system may trigger some extra memory copies somewhere, thus causes the additional latency.

Anyway, it is just my guess.

PR #7293, opened 2024-06-06
feat(pointcloud_preprocessor): add test for distortion correction node

vividf:
## Description
Add Integration testing for the distortion correction node.

- test empty twist
- test empty imu
- test empty pointcloud
- test normal input without using imu msg
- test normal input using imu msg

## Tests performed
Build
```
colcon build --symlink-install --cmake-args -DCMAKE_BUILD_TYPE=Release --packages-select pointcloud_preprocessor
```

Test
```
colcon test --packages-select pointcloud_preprocessor --event-handlers console_cohesion+
```


knzo25 (changes requested):
I saw you are addressing directly the data of the pointcloud. In python there are still ways to access pointclouds via iterators. If not that approach, ros2_numpy would be an option, but probably not possible in universe. However, **I highly recommend** you re write pointcloud related tests to c++

On another note, please avoid using `don't` and similar abbreviations. They have leaked into other parts where they should not been before

vividf:
@knzo25 

Thanks for the review, 
I agree that writing C++ tests is better for testing, however, this PR is not a Unit test but an Integration test (which tests the single node instead of the function). Regarding the Integration testing on the Autoware contribution ([testing the output of node](https://autowarefoundation.github.io/autoware-documentation/main/contributing/testing-guidelines/integration-testing/#:~:text=Testing%20the%20output,%23)), it suggests us to write in this way. Therefore, I am not sure that I should spend more time migrating it to the C++ version in this case. 

> I saw you are addressing directly the data of the pointcloud. In python there are still ways to access pointclouds via iterators. If not that approach, ros2_numpy would be an option, but probably not possible in universe.

For this comment, could you explain more in detail, Thanks!

knzo25:
```
An integration test is defined as the phase in software testing where individual software modules are combined and tested as a group. Integration tests occur after unit tests, and before validation tests.

The input to an integration test is a set of independent modules that have been unit tested. The set of modules is tested against the defined integration test plan, and the output is a set of properly integrated software modules that is ready for system testing.
```

Having the briefest of looks your test is a unit test, since anyways the distortion correction has essentially one method

vividf:
This PR is set to draft as we want the gtest instead of the component test.

vividf:
The unit test is created by another PR https://github.com/autowarefoundation/autoware.universe/pull/7783.
 Therefore, this pr will be closed.

PR #7312, opened 2024-06-06
ci(cppcheck-all): rename to cppcheck-daily and make it run only daily

kminoda:
## Description
Rename cppcheck-all to cppcheck-monitor to make the name self-contained.
I also would like to make it pass always as it may be a noise for developers.

## Tests performed

Please check the CI results of this PR (it should pass)

## Effects on system behavior

None


.github/workflows/cppcheck-monitor.yaml
```diff
@@ -1,4 +1,4 @@
-name: cppcheck-all
+name: cppcheck-monitor
 
 on:
   pull_request:
```
kminoda, line 4:
```suggestion
```

mitsudome-r (approved):

xmfcx:
Why did you remove
```yaml
  - name: Fail the job if Cppcheck failed
        if: steps.cppcheck.outcome == 'failure'
        run: exit 1
```

Part, I think it's better if it fails on failure.

kminoda:
@xmfcx Now that this CI is going to be a monitor rather than a check, I think it would be better to make it succeed regardless of the cppcheck results. Also it is obvious that cppcheck still outputs bunch of errors.
We can remove this CI eventually when all the cppcheck errors have been mitigated. (This way we may also be able to know when the CI is failing for some other reason, e.g. errors while installing cppcheck)

But if this may be confusing for other developers, I can leave that part. What do you think?

xmfcx:
I still think it should fail on failure. It is a daily job anyways. Most people won't even see it so it is ok.

xmfcx:
> Also it is obvious that cppcheck still outputs bunch of errors.

It is not obvious, it will only be known if you click into the individual workflow run logs.

> We can remove this CI eventually when all the cppcheck errors have been mitigated. (This way we may also be able to know when the CI is failing for some other reason, e.g. errors while installing cppcheck)

This action can stay to run daily forever. Maybe some files will go past cppcheck somehow, this will warn us in general.

> But if this may be confusing for other developers, I can leave that part. What do you think?

It won't be, it is only accessible if someone purposefully goes to https://github.com/autowarefoundation/autoware.universe/actions/workflows/cppcheck-monitor.yaml page. Because it is a scheduled daily job.

And people going to that link wants to know the answer to this question: "since when was this workflow failing?"

If you hide it, then we need to click each workflow run one by one.

.github/workflows/cppcheck-monitor.yaml
```diff
@@ -54,7 +52,3 @@ jobs:
         with:
           name: cppcheck-report
           path: cppcheck-report.xml
```
xmfcx, line 54:
```suggestion
          path: cppcheck-report.xml

      - name: Fail the job if Cppcheck failed
        if: steps.cppcheck.outcome == 'failure'
        run: exit 1
```

.github/workflows/cppcheck-monitor.yaml
```diff
@@ -1,13 +1,12 @@
...
 
 on:
-  pull_request:
   schedule:
     - cron: 0 0 * * *
   workflow_dispatch:
 
 jobs:
-  cppcheck-all:
+  cppcheck-monitor:
```
xmfcx, line 9:
```suggestion
  cppcheck-daily:
```

Same as above.

.github/workflows/cppcheck-monitor.yaml
```diff
@@ -1,13 +1,12 @@
-name: cppcheck-all
+name: cppcheck-monitor
```
xmfcx, line 1:
```suggestion
name: cppcheck-daily
```

This way the scope, schedule and function will be more clear, what do you think?

xmfcx (changes requested):

.github/workflows/cppcheck-monitor.yaml
kminoda: Agreed, thanks for the proposal.

.github/workflows/cppcheck-monitor.yaml
kminoda: Fixed: Fixed: https://github.com/autowarefoundation/autoware.universe/pull/7312/commits/2ed692bee996ebde3c22238b70844cde823891c5

.github/workflows/cppcheck-monitor.yaml
kminoda: > This action can stay to run daily forever. Maybe some files will go past cppcheck somehow, this will warn us in general.

Thanks for the comment. Agreed that it is better to make it fail on failure given that this CI will stay forever.

Fixed: https://github.com/autowarefoundation/autoware.universe/commit/2ed692bee996ebde3c22238b70844cde823891c5

xmfcx (approved):
Thank you! Let's wait for pre-commit.ci to fix the file ending and we can merge ✅

PR #7408, opened 2024-06-10
ci(build-and-test-differential): set checkout fetch-depth to 1

xmfcx:
## Description

Right now this step takes about 2m22s.

This happens for all jobs even before checking if any relevant files are changed.

**Before:** it downloaded the entire history and branches of the repository.
**After:** it only downloads the latest state of the repo (shallow copy).

## Tests performed

This PR CI, now it takes 5s (down from 2m22s).

## Effects on system behavior

Not applicable.

## Pre-review checklist for the PR author

The PR author **must** check the checkboxes below when creating the PR.

- [x] I've confirmed the [contribution guidelines].
- [x] The PR follows the [pull request guidelines].

## In-review checklist for the PR reviewers

The PR reviewers **must** check the checkboxes below before approval.

- [x] The PR follows the [pull request guidelines].

## Post-review checklist for the PR author

The PR author **must** check the checkboxes below before merging.

- [ ] There are no open discussions or they are tracked via tickets.

After all checkboxes are checked, anyone who has write access can merge the PR.

[contribution guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/
[pull request guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/pull-request-guidelines/

PR #7441, opened 2024-06-11
fix: revert "ci(build-and-test-differential): set checkout fetch-depth to 1"

mitsudome-r:
Reverts autowarefoundation/autoware.universe#7408


mitsudome-r:
Since we might merge PRs that might break build of Autoware without this PR, I'm merging it before status checks pass.

PR #7492, opened 2024-06-14
feat(tier4_state_rviz_plugin): change to read topic by polling

Autumn60:
## Description

Based on [the discussion](https://github.com/orgs/autowarefoundation/discussions/4612), change `tier4_state_rviz_plugin` to read topic by polling instead of best_effort callback.

During this change, `VelocitySteeringFactorsPanel` became timer-driven.

## Tests performed

Not applicable.

## Effects on system behavior

Not applicable.


isamu-takagi:
I don't think this change makes much sense. Since takeNewData takes messages from the queue in order, the same number of messages are processed, and there is more wasted processing because of polling.

Autumn60:
@isamu-takagi 
Thanks for your opinion.
As you say, if the frequency at which messages are sent is lower than the loop rate (20 Hz), then all messages will be processed.

So I consider this change as a "performance stabilization" rather than a "performance improvement".
As for "CPU load independent of external topic frequency", this PR makes sense.

isamu-takagi:
Thank you. The PR purpose makes sense. However, since the main purpose of this debug tool is to check data, it is more important to process all data than performance, so I would like to continue using the callback method.

Autumn60:
@isamu-takagi 

Thank you very much for your detailed explanation. :pray: 
I will close the PR here and leave the callback as is.

PR #7494, opened 2024-06-14
feat(tier4_adapi_rviz_plugin): add legacy state panel

isamu-takagi:
## Description

Adding legacy state panel because https://github.com/autowarefoundation/autoware.universe/pull/7036 reduces debug functionality.

## Tests performed

Not applicable.

## Effects on system behavior

Not applicable.


kosuke55 (approved):

xmfcx:
@isamu-takagi I've just realized this was added back.

Why do you think the current one is reducing the debug functionality?

image

Could you provide feedback?

xmfcx:
Which information is missing from the new panel?

armaganarsln:
I also don’t understand why, instead of addressing the issue of what changes are required for better debug functionality, the state panel was added back without consulting.

isamu-takagi:
@xmfcx @armaganarsln 
All I really needed was the operation_mode part, but I copied the old source code because I could use it as is. Sorry for the confusion of having two panels with same functions.

This is redundant in normal use and is not necessary for the `tier4_state_rviz_plugin`, but we needed to display the following to check the correctness of the implementation. I will later extract the relevant part as a panel dedicated to debugging.

- The current mode when is_mode_available is false.

Screenshot from 2024-09-21 09-22-50
Screenshot from 2024-09-21 09-23-12
Screenshot from 2024-09-21 09-24-11

- The destination mode during transition.

Screenshot from 2024-09-21 09-26-29
Screenshot from 2024-09-21 09-26-56

xmfcx:
I see, these cases are not represented correctly in the new version.

(I also understand why you needed the old version now, thanks. I've left a comment in #8933 as well.)

cc. @KhalilSelyan @isamu-takagi 

Could you confirm if the proposed changes would make it better?

## `Stop` button should be selected in these cases

And the remaining buttons should be disabled.

image

image

## `Auto` button should be selected in these cases

And the remaining buttons should be disabled.

image

image

isamu-takagi:
@xmfcx 

> Could you confirm if the proposed changes would make it better?

Yes, it's good that these are visible to the user.
- Stop button should be selected, rest should be disabled
- Stop button should be selected, rest should be enabled
- Auto button should be selected, rest should be disabled
- Auto button should be selected, rest should be enabled


> Should we allow users to press buttons during the transitions?

No, changing modes during a transition is not an expected usage. It is better to allow this operation only from the debug panel.

PR #7500, opened 2024-06-15
fix(vehicle_cmd_gate): put back subscriber rather than using polling subsriber

go-sakayori:
## Description
The previous [PR](https://github.com/autowarefoundation/autoware.universe/pull/7418/files#diff-c7f50a6d7ff06bab2e98e0a30e9384f680d0716e10b8a122f51346ea00eec474L716) changed the subscirbers to polling subscribers which is called in onTimer function. From the perspective of verifiability and latency, polling subscribers are restored to normal subscribers to achieve event-driven approach, except for turn indicators, hazard lights, and gear command. These 3 values are only referred in the onTimer function, and therefore the polling subscriber would be appropriate.

## Tests performed
Could engage on PSim.
Screenshot from 2024-06-15 23-00-45

## Effects on system behavior
Not applicable.


TakaHoribe (approved):
Some callbacks can use the polling subscriber since it just changes the member variables.

However, it is a bit complicated and does not cause critical issues.

PR #7596, opened 2024-06-20
feat(ndt_scan_matcher): estimate the covariance of ndt scan matching in real time instead of using a predefined value

SaltUhey:
## Description

Now, Autoware localizer uses Extended Kalman Filter (EKF) based on estimated pose with covariance and estimated velocity.
Moreover, the covariance of estimated pose is empirically defined in advance and remains constant regardless of location.
However, I think that real-time covariance estimation can make some contribution to addressing the following two LiDAR degeneracy environments.
- Tunnel
- Open spaces such as fields, rice paddies, and wilderness

By properly estimating the covariance at these locations, where ndt_pose are prone to error, the incorrect ndt_pose information is less likely to be integrated into the ekf_pose.
I propose to implement three methods as sub-functions by using [this](https://github.com/tier4/ndt_omp/tree/tier4/main/src/estimate_covariance).

## Related links

- [2D real-time covariance estimation(origin)](https://github.com/autowarefoundation/autoware.universe/tree/main/localization/ndt_scan_matcher#2d-real-time-covariance-estimation)
- [2D real-time covariance estimation(proposal)](https://github.com/SaltUhey/autoware.universe/tree/feat/estimete_ndt_covariance/localization/ndt_scan_matcher#2d-real-time-covariance-estimation)

autoware_launch
* https://github.com/autowarefoundation/autoware_launch/pull/1038

## Tests performed

Tests were conducted in tunnels and open spaces to verify that the covariance can be estimated in real time.

## Notes for reviewers

You need to add the parameter [autoware_launch](https://github.com/autowarefoundation/autoware_launch/blob/main/autoware_launch/config/localization/ndt_scan_matcher/ndt_scan_matcher.param.yaml), which is an external repository as shown in the image below.
image

## Interface changes
There are no changes to the interface other than the addition of sub-functions.

## Effects on system behavior
NDT covariance estimation will not start unless the arguments are changed. In other words, it does not affect the default Autoware behavior.


localization/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp
```diff
@@ -569,7 +572,7 @@ bool NDTScanMatcher::callback_sensor_points_main(
 
   if (is_converged && param_.covariance.covariance_estimation.enable) {
     const auto estimated_covariance =
-      estimate_covariance(ndt_result, initial_pose_matrix, sensor_ros_time);
+      estimate_covariance(ndt_result, initial_pose_matrix, sensor_ros_time, ndt_covariance);
```
SakodaShintaro, line 575: It seems a bit odd that the function `estimate_covariance` takes `ndt_covariance` as an argument.
I think the covariance should be the return value.

localization/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp
```diff
@@ -819,67 +822,91 @@ std::array<double, 36> NDTScanMatcher::estimate_covariance(
...
-
   geometry_msgs::msg::PoseArray multi_ndt_result_msg;
   geometry_msgs::msg::PoseArray multi_initial_pose_msg;
   multi_ndt_result_msg.header.stamp = sensor_ros_time;
-  multi_ndt_result_msg.header.frame_id = param_.frame.map_frame;
   multi_initial_pose_msg.header.stamp = sensor_ros_time;
+  multi_ndt_result_msg.header.frame_id = param_.frame.map_frame;
   multi_initial_pose_msg.header.frame_id = param_.frame.map_frame;
   multi_ndt_result_msg.poses.push_back(matrix4f_to_pose(ndt_result.pose));
   multi_initial_pose_msg.poses.push_back(matrix4f_to_pose(initial_pose_matrix));
```
SakodaShintaro, line 835: Rather than receiving the base covariance as an argument, how about declare it as a local variable on this line?
```
  std::array<double, 36> ndt_covariance = param_.covariance.output_pose_covariance;
```

localization/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp
```diff
@@ -819,67 +822,91 @@ std::array<double, 36> NDTScanMatcher::estimate_covariance(
...
+    ndt_covariance[0 + 6 * 0] = cov_by_mndt(0, 0);
+    ndt_covariance[1 + 6 * 1] = cov_by_mndt(1, 1);
+    ndt_covariance[1 + 6 * 0] = cov_by_mndt(1, 0);
+    ndt_covariance[0 + 6 * 1] = cov_by_mndt(0, 1);
+    const Eigen::Matrix2d cov_by_mndt_rotated =
+      pclomp::rotate_covariance_to_base_link(cov_by_mndt, ndt_result.pose);
+    ndt_dev_x_pub_->publish(
+      make_float32_stamped(sensor_ros_time, std::sqrt(cov_by_mndt_rotated(0, 0))));
+    ndt_dev_y_pub_->publish(
+      make_float32_stamped(sensor_ros_time, std::sqrt(cov_by_mndt_rotated(1, 1))));
```
SakodaShintaro, line 872: Since the processes such as setting the covariance matrix and publishing stddev are common, wouldn't it be better to implement them outside the if statement?

localization/ndt_scan_matcher/include/ndt_scan_matcher/hyper_parameters.hpp
```diff
@@ -29,6 +29,12 @@ enum class ConvergedParamType {
   NEAREST_VOXEL_TRANSFORMATION_LIKELIHOOD = 1
 };
 
+enum class CovarianceEstimationType {
+  LAPLACE_APPROXIMATION = 0,
+  MULTI_NDT = 1,
+  MULTI_NDT_SCORE = 2,
+};
```
SakodaShintaro, line 36: How about adding the traditional "set fixed values" mode here?
In that case, I think the `CovarianceEstimation.enabled` option would be unnecessary.

localization/ndt_scan_matcher/include/ndt_scan_matcher/ndt_scan_matcher_core.hpp
```diff
@@ -152,6 +152,8 @@ class NDTScanMatcher : public rclcpp::Node
     initial_pose_with_covariance_pub_;
   rclcpp::Publisher<geometry_msgs::msg::PoseArray>::SharedPtr multi_ndt_pose_pub_;
   rclcpp::Publisher<geometry_msgs::msg::PoseArray>::SharedPtr multi_initial_pose_pub_;
+  rclcpp::Publisher<tier4_debug_msgs::msg::Float32Stamped>::SharedPtr ndt_dev_x_pub_;
+  rclcpp::Publisher<tier4_debug_msgs::msg::Float32Stamped>::SharedPtr ndt_dev_y_pub_;
```
SakodaShintaro, line 156: These publishers may not be necessary, as these values ​​can be calculated from the output of the covariance matrix. I would like to hear other people's opinions on this.

SakodaShintaro (changes requested):
Thank you very much :+1: 
I have added some comments so please check them.

localization/ndt_scan_matcher/schema/sub/covariance_covariance_estimation.json
```diff
@@ -19,8 +19,16 @@
           "type": "array",
           "description": "Offset arrangement in covariance estimation [m]. initial_pose_offset_model_x & initial_pose_offset_model_y must have the same number of elements.",
           "default": [0.5, -0.5, 0.0, 0.0, 0.0, 0.0]
+        },
+        "covariance_estimation_type": {
+          "type": "float",
```
KYabuuchi, line 24: In json schema we can not use "float". Please replace with "number"
```suggestion
          "type": "number",
```

localization/ndt_scan_matcher/schema/sub/covariance_covariance_estimation.json
```diff
@@ -19,8 +19,16 @@
...
+          "description": "Converged estimation type. 0=LAPLACE_APPROXIMATION, 1=MULTI_NDT, 2=MULTI_NDT_SCORE",
+          "default": 2,
+          "minimum": 0,
+          "maximum": 2
         }
       },
+
       "required": ["enable", "initial_pose_offset_model_x", "initial_pose_offset_model_y"],
```
TaikiYamada4, line 32: Please add `covariance_estimation_type` to this `required` list.

TaikiYamada4 (commented):
Thank you for your PR!

Could you also change the `config/ndt_scan_matcher.param.yaml` as same as the one in autoware.launch?
https://github.com/autowarefoundation/autoware_launch/pull/1038

localization/ndt_scan_matcher/include/ndt_scan_matcher/hyper_parameters.hpp
```diff
@@ -81,7 +87,10 @@ struct HyperParameters
     struct CovarianceEstimation
     {
       bool enable{};
+      CovarianceEstimationType covariance_estimation_type{};
       std::vector<Eigen::Vector2d> initial_pose_offset_model{};
+      std::vector<double> initial_pose_offset_model_x{};
+      std::vector<double> initial_pose_offset_model_y{};
```
KYabuuchi, line 93: `initial_pose_offset_model` and `*_model_x` & `*_model_y` seem to have the same values. Since this is redundant, can't you simplify it?

localization/ndt_scan_matcher/include/ndt_scan_matcher/ndt_scan_matcher_core.hpp
KYabuuchi: I also think these publishers are unnecessary. It might be useful for debugging, but for practical purposes, I think pose_with_covariance is sufficient.

localization/ndt_scan_matcher/README.md
```diff
@@ -244,13 +244,17 @@ This is a function that uses no ground LiDAR scan to estimate the scan matching
 Calculate 2D covariance (xx, xy, yx, yy) in real time using the NDT convergence from multiple initial poses.
 The arrangement of multiple initial poses is efficiently limited by the Hessian matrix of the NDT score function.
 In this implementation, the number of initial positions is fixed to simplify the code.
+To obtain the covariance, MULTI_NDT computes until convergence at each initial position, while MULTI_NDT_SCORE uses the nearest voxel transformation likelihood.
```
KYabuuchi, line 247: This is confusing because it has not yet been explained what MULTI_NDT and MULTI_NDT_SCORE are here. 
To make it easier for the user to understand, please explain first that they are mode names of covariance estimation.

For example,
```Three modes are provided for 2D covariance estimation: LAPLACE_APPROXIMATION, MULTI_NDT, and MULTI_NDT_SCORE. To obtain the covariance,...```

File: localization/ndt_scan_matcher/include/ndt_scan_matcher/ndt_scan_matcher_core.hpp (hunk 5 above)
Comment (KYabuuchi) [reply localization/ndt_scan_matcher/include/ndt_scan_matcher/ndt_scan_matcher_core.hpp:156]: I also think these publishers are unnecessary. 
It might be useful for debugging, but for practical usages, I think pose_with_covariance is sufficient.

File: localization/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp (hunk 1 above)
Comment (SaltUhey) [reply localization/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp:575]: Now that you mention it, I think that's certainly true.
It may be strange to use `ndt_covariance` as an argument to output `estimate_covariance`.
However, within the estimate_covariance(), we need the default covariance (z, roll, pitch, yaw) converted into the base_link coordinate system.

Another option is to have the function return only the covariance of x and y, and not use `ndt_covariance` as an argument. However, there is a concern that the code will become a little longer from line 573 onwards. 
Should I include `ndt_covariance` as an argument, or not?

I also think the easiest way to avoid confusion would be to rename `ndt_covariance` to `default_ndt_covariance`.

File: localization/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp (hunk 2 above)
Comment (SaltUhey) [reply localization/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp:835]: This is one option.
However, it requires using ndt_covariance converted to the base_link coordinate system.
Therefore, the concern is that the code will become longer and calculations will increase.

File: localization/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp (hunk 3 above)
Comment (SaltUhey) [reply localization/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp:872]: I agree with your opinion, thank you for pointing that out.

File: localization/ndt_scan_matcher/include/ndt_scan_matcher/ndt_scan_matcher_core.hpp (hunk 5 above)
Comment (SaltUhey) [reply localization/ndt_scan_matcher/include/ndt_scan_matcher/ndt_scan_matcher_core.hpp:156]: OK! Thanks for pointing that out!

File: localization/ndt_scan_matcher/schema/sub/covariance_covariance_estimation.json (hunk 6 above)
Comment (SaltUhey) [reply localization/ndt_scan_matcher/schema/sub/covariance_covariance_estimation.json:24]: Ok! Thanks for pointing that out!

File: localization/ndt_scan_matcher/schema/sub/covariance_covariance_estimation.json (hunk 7 above)
Comment (SaltUhey) [reply localization/ndt_scan_matcher/schema/sub/covariance_covariance_estimation.json]: OK! Thanks for pointing that out!

File: localization/ndt_scan_matcher/include/ndt_scan_matcher/hyper_parameters.hpp (hunk 4 above)
Comment (SaltUhey) [reply localization/ndt_scan_matcher/include/ndt_scan_matcher/hyper_parameters.hpp:36]: I agree with your opinion. 
I'd like to hear other people's opinions as well.

File: localization/ndt_scan_matcher/README.md (hunk 9 above)
Comment (SaltUhey) [reply localization/ndt_scan_matcher/README.md:247]: OK! Thanks for pointing that out!

File: localization/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp (hunk 1 above)
Comment (SakodaShintaro) [reply localization/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp:573]: I think `estimate_covariance` should return not `std::array<double, 36>` but `Eigen::Matrix2d`. It is good to use `pclomp::adjust_diagonal_covariance` outside of `estimate_covariance`.

File: localization/ndt_scan_matcher/include/ndt_scan_matcher/hyper_parameters.hpp (hunk 4 above)
Comment (SakodaShintaro) [reply localization/ndt_scan_matcher/include/ndt_scan_matcher/hyper_parameters.hpp:36]: @KYabuuchi 
What do you think about this?

File: localization/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp [hunk 10]
```diff
@@ -819,68 +820,67 @@ std::array<double, 36> NDTScanMatcher::estimate_covariance(
...
+      const Eigen::Matrix4f sub_ndt_result = ndt_ptr_->getResult().pose;
+      multi_ndt_result_msg.poses.push_back(matrix4f_to_pose(sub_ndt_result));
+      multi_initial_pose_msg.poses.push_back(matrix4f_to_pose(sub_initial_pose_matrix));
+    }
+    multi_ndt_pose_pub_->publish(multi_ndt_result_msg);
+    multi_initial_pose_pub_->publish(multi_initial_pose_msg);
+  } else if (
+    param_.covariance.covariance_estimation.covariance_estimation_type ==
+    CovarianceEstimationType::MULTI_NDT_SCORE) {
+    const double temperature = 0.1;
```
SakodaShintaro, line 866: Can this variable be parameterized to config?

SakodaShintaro (changes requested):

SakodaShintaro:
Could you also update config/ndt_scan_matcher.param.yaml?
https://github.com/autowarefoundation/autoware.universe/blob/1635cbeb4fbcf931a563a3006a45282330019493/localization/ndt_scan_matcher/config/ndt_scan_matcher.param.yaml

localization/ndt_scan_matcher/include/ndt_scan_matcher/hyper_parameters.hpp
KYabuuchi: I also think it seems fine to add a fixed value mode and remove `enabled` from the parameter. :ok_hand:

SaltUhey: Thank you for your comment.
I will add a fixed value mode and remove `enabled` from the parameters.

localization/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp
SaltUhey: I agree with your opinion.
I change the return value of `estimate_covariance()` to `Eigen::Matrix2d`.

localization/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp
SaltUhey: OK.
I will change the code to handle temperature as a parameter.

SaltUhey: OK!
[Here](https://github.com/SaltUhey/autoware.universe/commit/8249a12ba22f5058f890df6a71e1c8def76e4df4), I added temperature to the parameters.

SakodaShintaro: Thanks!

localization/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp
```diff
@@ -841,47 +839,54 @@ std::array<double, 36> NDTScanMatcher::estimate_covariance(
...
+  } else if (
+    param_.covariance.covariance_estimation.covariance_estimation_type ==
+    CovarianceEstimationType::MULTI_NDT) {
+    const std::vector<Eigen::Matrix4f> poses_to_search = pclomp::propose_poses_to_search(
+      ndt_result, initial_pose_offset_model_x, initial_pose_offset_model_y);
+    const pclomp::ResultOfMultiNdtCovarianceEstimation result_of_multi_ndt_covariance_estimation =
+      estimate_xy_covariance_by_multi_ndt(ndt_result, ndt_ptr_, poses_to_search);
+    for (const auto & sub_initial_pose_matrix : poses_to_search) {
+      auto sub_output_cloud = std::make_shared<pcl::PointCloud<PointSource>>();
+      ndt_ptr_->align(*sub_output_cloud, sub_initial_pose_matrix);
```
SakodaShintaro, line 866: Is this "align" necessary?

The struct `pclomp::ResultOfMultiNdtCovarianceEstimation` already has `std::vector<NdtResult> ndt_results`.

https://github.com/tier4/ndt_omp/blob/ef654bdb1a3fe0bae6f3247b4959e230e63ba429/include/estimate_covariance/estimate_covariance.hpp#L9-L14

SaltUhey: Sure, maybe. I'll fix it.

localization/ndt_scan_matcher/include/ndt_scan_matcher/hyper_parameters.hpp
SaltUhey: `initial_pose_offset_model_x` and `initial_pose_offset_model_y` are necessary to use `pclomp::propose_poses_to_search()`.

On the other hand, `initial_pose_offset_model` is not used as is by `ndt_scan_matcher`.

Therefore, I am  considering deleting `std::vector<Eigen::Vector2d> initial_pose_offset_model` from hyper_parameters.hpp.

KYabuuchi: Your suggestion seems good! :+1:

SakodaShintaro (approved):
Thank you for all the fixes.
I have confirmed that logging_simulator and AWSIM work properly.
Looks Good To Me

PR #7612, opened 2024-06-21
feat(ndt_scan_matcher): warn if the lidar scan has gone out of the range of the loaded map

anhnv3991:
## Description

- Display a warning message when the initial pose of NDT has gone out of the map range. Basically, warn users when the map should be updated but it has not been updated yet.

## Tests performed
- XX1 Odaiba PCD map and rosbag
- In launcher/autoware_launch/config/localization/ndt_scan_matcher/ndt_scan_matcher.param.yaml, lower the dynamic map loading radius to closer to lidar_radius (about 2 meters bigger)
- Launch logging simulator 
- The terminal of the logging simulator should display something like this
`...
[ndt_scan_matcher_node-39] [WARN] [1718934187.199986673] [localization.pose_estimator.ndt_scan_matcher]: Lidar has gone out of the map range
...`

## Effects on system behavior

None

## Interface changes

None


KYabuuchi (commented):
@anhnv3991 
The return value of `should_update_map` indicates "whether the map should be updated", not "whether the vehicle has exited the map boundary". Therefore, it is incorrect to output the WARN message when the function returns true.

The check for whether the vehicle is outside the map boundary is already performed internally within `should_update_map`, and an ERROR diagnostic is issued if a violation occurs. Additionally, since this function is called periodically in the timer callback, I think there is no need to check this in the sensor points callback as well.

The potential issue arises when waiting for a service response from map_loader. This waiting occurs within the timer_callback of MapUpdateModule, during which the conventional `should_update_map` is not called.

So, could you modify the implementation to check that the vehicle does not go out of the map boundary [while waiting for the service response](https://github.com/autowarefoundation/autoware.universe/blob/f3069735bb70d318cc99618c67c3dfda0f35adca/localization/ndt_scan_matcher/src/map_update_module.cpp#L216-L230) ?

anhnv3991:
> The potential issue arises when waiting for a service response from map_loader. This waiting occurs within the timer_callback of MapUpdateModule, during which the conventional `should_update_map` is not called.

@KYabuuchi Why don't we just call a check in the main callback sensor point (of ndt_scan_matcher_core.cpp)? Basically the check by callback timer is performed periodically (e.g. every 200ms if ros2 bag play --clock 200). Hence, within that period (200ms) there is no check, and the vehicle could move outside of the map range if it travels fast enough.

Checking from map_update_module while waiting for service is quite complicated, since map_update_module does not have access to the current position of NDT. Creating a shared pointer for the poses in such cases is unnecessary imo.

KYabuuchi:
@anhnv3991 `--clock 200` does not mean that the timer callback is called every 200ms. It specifies the frequency at which the /clock topic is published from the rosbag, so `--clock 200` means /clock is published at 200Hz. 
And, the MapUpdateModule's timer callback runs at 1Hz. [source code link](https://github.com/autowarefoundation/autoware.universe/blob/main/localization/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp#L107)

From my understanding, NDT runs should_update_map in a separate thread from callback_sensor_points_main because:
* When updating the map, it needs to wait for the map_loader service in a separate thread anyway.
* Checking at a frequency of 1Hz seems to be sufficient. (On the other hand, there is no upper limit to the delay of the map_loader response.)
* Concentrating MapUpdateModule on map management makes the code easier to understand.

---

> Checking from map_update_module while waiting for service is quite complicated, since map_update_module does not have access to the current position of NDT. 

Indeed, what you're saying is correct. I didn't anticipate this difficulty...

anhnv3991:
@KYabuuchi I see. I just checked the should_update_map again. You're right, should_update_map returns true does not mean the lidar has gone out of the map range. The if clause inside that function is what we need for checking.

So how about this? We can define a method of map_update_module to check the map range, and just let the callback_sensor_points call that method on NDT poses.

anhnv3991:
@KYabuuchi Just a question: To change parameters of ndt_scan_matcher, I found 2 files 
```
src/launcher/autoware_launch/config/localization/ndt_scan_matcher/ndt_scan_matcher.param.yaml 
src/universe/autoware.universe/localization/ndt_scan_matcher/config/ndt_scan_matcher.param.yaml
```
which have the same content. Which file do I have to edit to change parameters of ndt_scan_matcher? I thought colcon --symlink-install should link those files but it seems like it does not.

KYabuuchi:
@anhnv3991 When using `ros2 launch autoware_launch`, the parameter files loaded are in `src/launch/autoware_launch/*`.
If you want to experiment with your implementation, you should edit the files in `src/launch/autoware_launch/*`.

These files contain the same content, and you can consider the parameter files in autoware.universe/* as sample parameter files.

localization/ndt_scan_matcher/src/map_update_module.cpp
```diff
@@ -107,6 +107,20 @@ bool MapUpdateModule::should_update_map(
   return distance > param_.update_distance;
 }
 
+bool MapUpdateModule::out_of_map_range(const geometry_msgs::msg::Point & position)
```
KYabuuchi, line 110: `out_of_map` and `should_update_map` have overlapping content. Can you make them common?

localization/ndt_scan_matcher/src/map_update_module.cpp
```diff
@@ -107,6 +107,20 @@ bool MapUpdateModule::should_update_map(
...
+{
+  if (last_update_position_ == std::nullopt) {
+    return true;
+  }
+
+  const double dx = position.x - last_update_position_.value().x;
+  const double dy = position.y - last_update_position_.value().y;
```
KYabuuchi, line 117: The `last_update_position_` and `param_` in MapUpdateModule are accessed from both the callback_sensor_points and the callback_timer, making them susceptible to data races. 

Although it's not a critical issue at the moment, it poses a risk. Please add a lock or comments explaining why data races won't occur, and caution future developers who might work on this code.

localization/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp
```diff
@@ -439,6 +439,18 @@ bool NDTScanMatcher::callback_sensor_points_main(
     add_regularization_pose(sensor_ros_time);
   }
 
+  // Warn if the lidar has gone out of the map range
```
KYabuuchi, line 442: Can you add a reason why it is checked periodically in map_update_module but also here? 
My understanding is that this part is specifically checking for being out of the map during the waiting time for a service reply.

KYabuuchi (commented):
Sorry for the delayed response regarding this. 
Basically, I think this PR will provide the desired functionality. :+1: But, there is a risk of data races between the sensor points callback and the timer callback, so we need to be careful.

localization/ndt_scan_matcher/src/map_update_module.cpp
KYabuuchi: Reading an optional type can sometimes result in a bad optional access, so we need to be cautious. 
Here is an example.
```cpp
#include <iostream>
#include <optional>
#include <thread>
std::optional<int> count = std::nullopt;
void ThreadA()
{
  for (int i = 0; i < 100000; ++i) {
    if (i % 2 == 0)
      count = i;
    else
      count = std::nullopt;
  }
}
void ThreadB()
{
  long sum = 0;
  for (int i = 0; i < 100000; ++i) {
    if (count.has_value())
      sum += count.value(); // This will occur bad optinal access
  }
}
int main()
{
  std::thread th_a(ThreadA);
  std::thread th_b(ThreadB);
  th_a.join();
  th_b.join();
  return 0;
}

```

localization/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp
anhnv3991: @KYabuuchi In the current setting, it is checked once every 1 second. During that 1 second, if the vehicle travels too fast there is a chance that the lidar may get out of the map range (although in the current setting, it is very hard to occur, unless the vehicle speed is about 50 m/s ~ 180 km/h). So this check here is for such cases.

But I think if we reduce the period of the timer callback it would be better (like instead of 1 second we set it to 50 ms). However, doing so would raise another issue: while the map and NDT are being updated (which often takes more than 50ms), there is a chance that there is another should_update_map return true, thus triggers another update_map. Such duplicated operations are not necessary at all.

KYabuuchi: @anhnv3991 Oh, I thought you implemented option 1 because option 2 wouldn't alert us if the vehicle goes out of the map while waiting for the map_loader service (which often takes more than 50ms).

I want to clarify what issues this PR is intended to address. I believe it tackles the following two issues:
1. The lack of checks during the waiting time for the map_loader service in MapUpdateModule.
2. The risk of the vehicle going out of the map within 1 second when dynamic_map_loading parameters are set to hard values (although this is rare with default parameters).

Please let me know if I misunderstood.

Your option 1 addresses both issues, so I think your option 1 is the natural choice. (However, there's also the option to check in the initial_pose_callback instead of the sensor_points_callback.)

SakodaShintaro:
I'm going to check this pull request now, but have all of Yabuuchi-san's comments been resolved yet?

localization/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp
anhnv3991: @KYabuuchi Actually, initially I thought this PR is to address No. 2 (which is pointed out by Ando-san). But No.1 is also a good point of this PR. Shortly, this PR warns us if the vehicle goes out of the map range anytime.

localization/ndt_scan_matcher/src/map_update_module.cpp
SakodaShintaro: @anhnv3991 
I think this is important, so please respond.

localization/ndt_scan_matcher/include/ndt_scan_matcher/map_update_module.hpp
```diff
@@ -88,6 +91,9 @@ class MapUpdateModule
   // Indicate if there is a prefetch thread waiting for being collected
   NdtPtrType secondary_ndt_ptr_;
   bool need_rebuild_;
+  std::mutex last_update_pos_mtx_;
+  // A copy of last update position, used for checking map range
+  std::optional<geometry_msgs::msg::Point> last_update_pos_;
```
SakodaShintaro, line 96: I'm confused about the need to add this variable. Why is not it enough to just use `last_update_position_`?

anhnv3991: @SakodaShintaro I removed the second last_update_pos_. The mutex is now used for last_update_position_.

localization/ndt_scan_matcher/src/map_update_module.cpp
SakodaShintaro: I feel the same way, but there are some minor differences so I am not sure whether the function can be used as common. I am still thinking about it.

SakodaShintaro: I cannot come up with any good ideas to organize the code. :disappointed: 
I think it is okay to merge this as is.

SakodaShintaro (approved):
Looks Good To Me
@KYabuuchi 
Are your all comments already resolved?
I think it would also be good to have your approval before merging.

localization/ndt_scan_matcher/src/map_update_module.cpp
KYabuuchi: Since there is little duplication of code, I also think it is acceptable as is in this time. :ok_man:

localization/ndt_scan_matcher/src/map_update_module.cpp
KYabuuchi: mutex has already been added. So I close this thread.

KYabuuchi (approved):
Sorry for the delayed review. All of my concerns have been addressed.
Looks Good To Me :ok_man:

Issue #7587, opened 2024-06-19
Bugs about euclidean_cluster


beginningfan:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I'm convinced that this is not my fault but a bug.

### Description

1. `min_cluster_size` is not passed to [clustering algorithm](https://github.com/autowarefoundation/autoware.universe/blob/b284599e61d2824049ae76f907ca698abc9bccbe/perception/euclidean_cluster/lib/voxel_grid_based_euclidean_cluster.cpp#L89)
2. When the `max_cluster_size_` is large, clustering takes too much CPU and memory because this [line](https://github.com/autowarefoundation/autoware.universe/blob/b284599e61d2824049ae76f907ca698abc9bccbe/perception/euclidean_cluster/lib/voxel_grid_based_euclidean_cluster.cpp#L109). This bug is related to this [pr](https://github.com/autowarefoundation/autoware.universe/pull/6818)

### Expected behavior

1. The size of the cluster is greater than or equal to the `min_cluster_size`
2. Memory allocation should be based on the actual size of the point cloud

### Actual behavior

1. The input parameter of the cluster algorithm is 1 
2. Memory allocation is based on the maximum number of clusters

### Steps to reproduce

1. print size of clusters before filter in [here](https://github.com/autowarefoundation/autoware.universe/blob/b284599e61d2824049ae76f907ca698abc9bccbe/perception/euclidean_cluster/lib/voxel_grid_based_euclidean_cluster.cpp#L134)
2. use a large number such as 30000 as `max_cluster_size` in [config](https://github.com/autowarefoundation/autoware.universe/blob/b284599e61d2824049ae76f907ca698abc9bccbe/perception/euclidean_cluster/config/voxel_grid_based_euclidean_cluster.param.yaml#L7)

### Versions

OS: ROS2
autoware: latest

PR #7662, opened 2024-06-24
fix(euclidean_cluster): fix euclidean cluster params

beginningfan:
## Description

Fix bug 2 about `max_cluster_size_` related in [here](https://github.com/autowarefoundation/autoware.universe/issues/7587)
The bug about `min_cluster_size` is a mistake and it should not to be changed

## Tests performed

use a large `max_cluster_size_` will not take too much CPU and memory

## Effects on system behavior

Not applicable.


yukkysaito:
@badai-nguyen Can I ask you to review this?

badai-nguyen:
@beginningfan Thank you for your PR and raising issue. 
- Related to `min_cluster_size` I commented here https://github.com/autowarefoundation/autoware.universe/issues/7587#issuecomment-2186540150
- Related to `max_cluster_size`, since the [voxel_grid_filter](https://github.com/autowarefoundation/autoware.universe/blob/main/perception/euclidean_cluster/lib/voxel_grid_based_euclidean_cluster.cpp#L65-L69) was applied to downsized input pointcloud before applying [euclidean_cluster filter](https://github.com/autowarefoundation/autoware.universe/blob/9a456e6a7e41b9e7e16785a939e8f8eaaec67ff0/perception/euclidean_cluster/lib/voxel_grid_based_euclidean_cluster.cpp#L89C1-L93C50) so that the actual output cluster size after reversing the [voxel_grid_filter](https://github.com/autowarefoundation/autoware.universe/blob/9a456e6a7e41b9e7e16785a939e8f8eaaec67ff0/perception/euclidean_cluster/lib/voxel_grid_based_euclidean_cluster.cpp#L117) could be bigger than cluster.indices.size(). It means that applying your proposal might over_filtering out some clusters which has actual `cluster_size` < `max_cluster_size_`.  That is also the reason why `min_cluster_size` for `pcl_euclidean_cluster_filter` is set as 1 and rechecked at later stage.

Please confirm about points and let me known if any question.

beginningfan:
@badai-nguyen thanks for your reply.
Related to `min_cluster_size`, I misunderstood this, so I did not modify min_cluster_size in this PR.
Related to `max_cluster_size`, I understand your approach. I first allocate space according to the downsized cluster size, and then resize it if there is insufficient space when push_back. Because mostly we cannot determine the maximum value of clusters, so we set a larger number to ensure that clusters are not lost. However, using the maximum value to allocate space will cause excessive CPU and memory usage.This method will not cause problems in previous versions.

perception/euclidean_cluster/lib/voxel_grid_based_euclidean_cluster.cpp
```diff
@@ -124,6 +124,10 @@ bool VoxelGridBasedEuclideanCluster::cluster(
         &temporary_clusters.at(map[index]).data[cluster_data_size],
         &pointcloud_msg->data[i * point_step], point_step);
       cluster_data_size += point_step;
+      if (cluster_data_size == temporary_clusters.at(map[index]).data.size()) {
+        temporary_clusters.at(map[index])
+          .data.resize(temporary_clusters.at(map[index]).data.size() * 2);
```
badai-nguyen, line 129: I think decide that the max size of output cluster (`cluster_data_size`) to be doubled of temporary_cluster is not reasonable. (runtime_error might be occurred here for some high dense pointcloud)
```suggestion
          .data.resize(max_cluster_size_ * point_step);
```


Or we can increment `cluster_data_size` and check it repeatedly.

beginningfan: Here is a `for loop` that reallocates twice the space every time the space is full. When the space is full, it will twice the space again.I have tested using a ouster64, and the dense pointcloud is sometimes 50 times larger than the downsampled one. Even so, enough space can be allocated quickly.
If you use `max_cluster_size_`, it might as well just use the original code. If we add 10 points each time, this will result in multiple copies of the vector data.

badai-nguyen: @beginningfan I understood. I have checked and it worked well. Thank you for your improvement.

badai-nguyen (approved):
LGTM

beginningfan:
@mitsudome-r The reviewer of this PR has approved it. Can it be merged now?

PR #6989, opened 2024-05-12
feat(lidar_centerpoint): accelerated preprocessing for centerpoint

knzo25:
## Description
While changing the point type to the new standard, realized that preprocessing was bottleneck in centerpoint.
Apologies to Amadeusz since he did almost the same already in transfusion, and I forgot until writing this PR but I also moved the preprocessing to cuda for gpu acceleration.

During tests in my desktop the detection time went from 16.9ms to 13.0ms, making preprocessing about ~1ms if we forcefully add syncs to measure the time.

## Tests performed
 - Tested using data from the taxi

## Pre-review checklist for the PR author

The PR author **must** check the checkboxes below when creating the PR.

- [x] I've confirmed the [contribution guidelines].
- [x] The PR follows the [pull request guidelines].

## In-review checklist for the PR reviewers

The PR reviewers **must** check the checkboxes below before approval.

- [ ] The PR follows the [pull request guidelines].
- [ ] The PR has been properly tested.
- [ ] The PR has been reviewed by the code owners.

## Post-review checklist for the PR author

The PR author **must** check the checkboxes below before merging.

- [ ] There are no open discussions or they are tracked via tickets.
- [ ] The PR is ready for merge.

After all checkboxes are checked, anyone who has write access can merge the PR.

[contribution guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/
[pull request guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/pull-request-guidelines/

PR #7676, opened 2024-06-25
revert: revert PR #6989

kminoda:
## Description
This reverts commit fffb14946829b5e91171880445237979fc796403.

## Related links
None

## How was this PR tested?
None

## Notes for reviewers
None

## Interface changes

None.

## Effects on system behavior

None.


knzo25:
Instead of reverting this, would it not be less painful at this time to comment out the cuda tests from the cmake?

PR #7667, opened 2024-06-24
chore: update the PR template

xmfcx:
### Old PR pipeline:
- Click on the preview button (Unintuitive)
- Select a PR template
- There are many checkboxes (nobody uses them and things get merged anyways)

Old method:

https://github.com/autowarefoundation/autoware.universe/assets/10751153/d889c718-0891-44a3-bb41-422c547e829a

### New PR pipeline:
- Template comes right away, has the best of both templates in a single simple package.
   - Multi-line comments are color coded, easy to remove.
- Checkboxes are removed and offloaded to a CI action to comment automatically on the PR.
- Verbose/obvious comments are removed, it is very clean now.
- Test new template using [this link](https://github.com/xmfcx/test_temp/compare/main...xmfcx-patch-2?quick_pull=1)
- Renamed the `PULL_REQUEST_TEMPLATE.md` to `pull_request_template.md` following [Github's Creating a pull request template for your repository guidelines](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository)

## Tests performed

- https://github.com/xmfcx/test_temp/compare/main...xmfcx-patch-2?quick_pull=1
- https://github.com/autowarefoundation/autoware.universe/pull/7667#issuecomment-2186550636

## Effects on system behavior

Not applicable.

## Pre-review checklist for the PR author

The PR author **must** check the checkboxes below when creating the PR.

- [x] I've confirmed the [contribution guidelines].
- [x] The PR follows the [pull request guidelines].

## In-review checklist for the PR reviewers

The PR reviewers **must** check the checkboxes below before approval.

- [ ] The PR follows the [pull request guidelines].

## Post-review checklist for the PR author

The PR author **must** check the checkboxes below before merging.

- [ ] There are no open discussions or they are tracked via tickets.

After all checkboxes are checked, anyone who has write access can merge the PR.

[contribution guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/
[pull request guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/pull-request-guidelines/

PR #7680, opened 2024-06-25
ci(comment-on-pr): make it work on forks

xmfcx:
## Description

It failed on this fork pr:
- https://github.com/autowarefoundation/autoware.universe/pull/7678
With error:
- https://github.com/autowarefoundation/autoware.universe/actions/runs/9657331258/job/26636407771?pr=7678#step:3:26
   - `Error: Resource not accessible by integration`

Now changing the trigger to [`pull_request_target`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target) so it can access repo's github token for commenting.

Also changing the general naming to fit the repo's other workflows. Shouldn't affect anything since I've added this new workflow just yesterday.

## Related links

**Follow up from:**

- https://github.com/autowarefoundation/autoware.universe/pull/7667

Related solution:
- https://github.com/marocchino/sticky-pull-request-comment/issues/227

## How was this PR tested?

- Can't be tested within this repo until it is merged (because `pull_request_target` trigger runs on the repo commit version).
- Could be tested on a separate repo but I didn't since it is simple.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/.github/workflows/comment-on-pr.yaml
+++ b/.github/workflows/comment-on-pr.yaml
@@ -1,6 +1,6 @@
-name: Comment on PR
+name: comment-on-pr
 on:
-  pull_request:
+  pull_request_target:
 
 jobs:
   comment-on-pr:


xmfcx:
Alright it works now:
- https://github.com/autowarefoundation/autoware.universe/pull/7681#issuecomment-2188149711

PR #7681, opened 2024-06-25
fix(pose_instability_detector): fix a rare error

SakodaShintaro:
## Description
In `pose_instability_detector`, a bug has been fixed that in rare abnormal cases, it may access to outside the queue size and crash, or it may show abnormal values.

I have added test cases that cause this, and I have confirmed that it fails probabilistically if it is not fixed.
Since accessing outside the range is undefined behavior, it does not necessarily always fail.

By making a modification of pose_instability_detector.cpp, the test will no longer fail even after 100 consecutive attempts.

## Related links

None.

## How was this PR tested?
The following script was used.

```
./test_loop.sh pose_instability_detector
```

```bash
#!/bin/bash
set -eux

TEST_TARGET_PKG=$1
LOOP_NUM=${2:-100}

colcon build --symlink-install \
  --cmake-args \
    -DCMAKE_BUILD_TYPE=Release \
    -DBUILD_TESTING=ON \
  --packages-select $TEST_TARGET_PKG

for ((i=1; i<=LOOP_NUM; i++))
do
  colcon test --packages-select ${TEST_TARGET_PKG} --event-handlers console_cohesion+ --return-code-on-test-failure
  echo "Finish test $i"
done
```

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/localization/pose_instability_detector/test/test.cpp
+++ b/localization/pose_instability_detector/test/test.cpp
@@ -167,16 +167,17 @@ TEST_F(TestPoseInstabilityDetector, does_not_crash_even_if_abnormal_odometry_dat
 {
   // [Condition] There is no twist_msg between the two target odometry_msgs.
   // Normally this doesn't seem to happen.
-  // As far as I can think, this happens when
-  //  (case A) the twist msg stops
-  //  (case B) the odometry msg stops (so the next timer callback will refer to the same odometry
-  //  msg, and the timestamp difference will be calculated as 0)
+  // As far as I can think, this happens when the odometry msg stops (so the next timer callback
+  // will refer to the same odometry msg, and the timestamp difference will be calculated as 0)
+  // This test case shows that an error occurs when two odometry msgs come in close succession and
+  // there is no other odometry msg.
+  // Referring again, this doesn't normally seem to happen in usual operation.
 
   builtin_interfaces::msg::Time timestamp{};
 
   // send the twist message1
   timestamp.sec = 0;
-  timestamp.nanosec = 5e8;  // timestamp is the same as the first odometry message
+  timestamp.nanosec = 0;
   helper_->send_twist_message(timestamp, 0.2, 0.0, 0.0);
 
   // send the first odometry message after the first twist message
@@ -196,12 +197,7 @@ TEST_F(TestPoseInstabilityDetector, does_not_crash_even_if_abnormal_odometry_dat
   timestamp.nanosec = 5e8 + 1e7;
   helper_->send_odometry_message(timestamp, 12.0, 0.0, 0.0);
 
-  // send the twist message2 (move 0.1m in x direction)
-  timestamp.sec = 0;
-  timestamp.nanosec = 7e8 + 5e7;
-  helper_->send_twist_message(timestamp, 0.2, 0.0, 0.0);
-
-  // send the twist message3 (move 0.1m in x direction)
+  // send the twist message2
   timestamp.sec = 1;
   timestamp.nanosec = 0;
   helper_->send_twist_message(timestamp, 0.2, 0.0, 0.0);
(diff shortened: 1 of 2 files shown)


xmfcx:
@SakodaShintaro could you rebase to the latest main?

I want to see if the fix on comment-on-pr workflow fixed the issue.

- https://github.com/autowarefoundation/autoware.universe/pull/7680

localization/pose_instability_detector/test/test.cpp
```diff
@@ -163,6 +163,64 @@ TEST_F(TestPoseInstabilityDetector, output_warn_when_twist_is_too_small)  // NOL
...
 }
 
+TEST_F(TestPoseInstabilityDetector, does_not_crash_even_if_abnormal_odometry_data_comes)  // NOLINT
+{
+  // [Condition] There is no twist_msg between the two target odometry_msgs.
+  // Normally this doesn't seem to happen.
+  // As far as I can think, this happens when
+  //  (case A) the twist msg stops
+  //  (case B) the odometry msg stops (so the next timer callback will refer to the same odometry
+  //  msg, and the timestamp difference will be calculated as 0)
```
TaikiYamada4, line 173: I'm not sure what does this test code tests here.
Is it case A or B?

Either way, this test looks like a normal test of a dead reckoning mismatch.
- Set the velocity 0.2 m/s
- Move the odometry (12 - 10) = 2 meters between 0.09999990 seconds
- The dead reckon shouldn't match and publish a warning.

TaikiYamada4 (commented):
After all, I think the dead reckon process should be skipped when the time of latest and prev is same, and publish a WARN diagnostics. (And then I believe the size of result_deque will be always be > 2.)

But I will mark that as a TODO for me, and I can approve the cpp part.
Can you take a look to my comment in the test part? :pray:

localization/pose_instability_detector/test/test.cpp
SakodaShintaro: Thank you. Thinking about it more deeply, Case A doesn't seem to occur. twist_buffer.size() becomes 1 and is rejected appropriately.
I've revised the test case to make the intent clearer.

https://github.com/autowarefoundation/autoware.universe/pull/7681/commits/d3d011c116483f00898b8a86466dc2c95ebf9dcd

localization/pose_instability_detector/test/test.cpp
```diff
@@ -163,6 +163,59 @@ TEST_F(TestPoseInstabilityDetector, output_warn_when_twist_is_too_small)  // NOL
...
+  helper_->send_odometry_message(timestamp, 12.0, 0.0, 0.0);
+
+  // send the twist message2
+  timestamp.sec = 1;
+  timestamp.nanosec = 0;
+  helper_->send_twist_message(timestamp, 0.2, 0.0, 0.0);
+
+  executor_.spin_some();
+
+  // process the above messages (by timer_callback)
```
TaikiYamada4, line 207: It seems that this test code will not fail in the current version.
I think you have to receive the diagnostics twice (= ensure to run timer callback twice)
This will crash the current version but not the PR version.
```suggestion
  // process the above messages (by timer_callback)
  helper_->received_diagnostic_array_flag = false;
  while (!helper_->received_diagnostic_array_flag) {
    executor_.spin_some();
    std::this_thread::sleep_for(std::chrono::milliseconds(10));
  }
  
  // provoke timer callback again
```

SakodaShintaro: Since the processing for the first Odometry is done before, I thought that diagnostics would only be published once here.
However, if I change it as suggested, an error will definitely occur, so I will make the change.

TaikiYamada4 (approved):
LGTM! :+1:

PR #7686, opened 2024-06-25
fix(lidar_centerpoint): fix constexpr related bugs

veqcc:
## Description

These are fixes based on clang-tidy CRITICAL warnings ( = clang compile error)

For example,
```
constexpr variable 'detection_yaw_sin' must be initialized by a constant expression
```

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


knzo25 (approved):
I would have sworn that stl's math functions were compatible with constexpr :cry: 

LGTM

veqcc:
@knzo25 
Agreed 😮‍💨 
They would be constexpr from C++26

Issue #7668, opened 2024-06-24
build-and-test workflow is failing due to CUDA runtime version


xmfcx:
@knzo25 starting from this PR, The CI for `build-and-test` started failing:
- https://github.com/autowarefoundation/autoware.universe/actions/runs/9590002364/job/26444678737#step:15:22008
```
8: C++ exception with description "cudaErrorInsufficientDriver (35)@/__w/autoware.universe/autoware.universe/perception/lidar_centerpoint/include/lidar_centerpoint/cuda_utils.hpp#L80: CUDA driver version is insufficient for CUDA runtime version" thrown in the test body.
```

For some reason the [has CI passed](https://github.com/autowarefoundation/autoware.universe/actions/runs/9589903760/job/26444432176) for the `build-and-test-differential` cuda yet it fails the `build-and-test` checks.

_Originally posted by @xmfcx in https://github.com/autowarefoundation/autoware.universe/issues/6989#issuecomment-2186485659_

knzo25:
@xmfcx 
Kind of a nasty one as I run the tests a while ago with no errors. Will look into it asap, but this week I am full of deadlines...

knzo25:
@youtalk 

I tried to reproduce the error using a docker and following the ci/cd commands whenever possible.
However, the tests pass:

```
9: -- run_test.py: invoking following command in '/home/kenzolobos/workspace/autoware/build/lidar_centerpoint':
9:  - /home/kenzolobos/workspace/autoware/build/lidar_centerpoint/test_preprocess_kernel
9: [==========] Running 4 tests from 1 test suite.
9: [----------] Global test environment set-up.
9: [----------] 4 tests from PreprocessKernelTest
9: [ RUN      ] PreprocessKernelTest.EmptyVoxelTest
9: [       OK ] PreprocessKernelTest.EmptyVoxelTest (45 ms)
9: [ RUN      ] PreprocessKernelTest.BasicTest
9: [       OK ] PreprocessKernelTest.BasicTest (7 ms)
9: [ RUN      ] PreprocessKernelTest.OutOfRangeTest
9: [       OK ] PreprocessKernelTest.OutOfRangeTest (7 ms)
9: [ RUN      ] PreprocessKernelTest.VoxelOverflowTest
9: [       OK ] PreprocessKernelTest.VoxelOverflowTest (7 ms)
9: [----------] 4 tests from PreprocessKernelTest (66 ms total)
9: 
[...]

PR #7694, opened 2024-06-26
ci: show cuda and nvidia-driver version in build&test ci

kminoda:
## Description

Since build and test of CUDA-related packages may fail, not only due to the package modification but also due to the environment modification in the recent GitHub Actions improvements, showing the CUDA and nvidia-driver version would be helpful for debugging in some cases.

## Related links

None

## How was this PR tested?
None

## Effects on system behavior

No effects on Autoware.


xmfcx:
@kminoda Is this related to:
- https://github.com/autowarefoundation/autoware.universe/issues/7668

PR #7699, opened 2024-06-26
fix(autoware_velocity_smoother): fix duplicateBreak warning

veqcc:
## Description

This is a fix based on cppcheck `duplicateBreak` warning

```
planning/autoware_velocity_smoother/src/node.cpp:1016:3: style: Consecutive return, break, continue, goto or throw statements are unnecessary. [duplicateBreak]
  return AlgorithmType::INVALID;
  ^
```

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/autoware_velocity_smoother/src/node.cpp
+++ b/planning/autoware_velocity_smoother/src/node.cpp
@@ -1013,7 +1013,6 @@ VelocitySmootherNode::AlgorithmType VelocitySmootherNode::getAlgorithmType(
   }
 
   throw std::domain_error("[VelocitySmootherNode] undesired algorithm is selected.");
-  return AlgorithmType::INVALID;
 }
 
 double VelocitySmootherNode::calcTravelDistance() const


TakaHoribe (approved):
I didn't know this code causes an error message, but this fix makes sense to me.

rej55 (approved):
LGTM!

veqcc:
@TakaHoribe 
> I didn't know this code causes an error message

This is because `throw std::domain_error("[VelocitySmootherNode] undesired algorithm is selected.");` will exit the code and the following `return AlgorithmType::INVALID;` will not be executed.
The warning `duplicateBreak` means ""Consecutive return, break, continue, goto or throw statements are unnecessary"" (cited from https://github.com/danmar/cppcheck/blob/1668b0b689885b703039af1eb27c76da77f6445d/lib/checkother.cpp#L893 )

PR #7717, opened 2024-06-27
fix(lane_change): prevent empty path when routing

kosuke55:
## Description

prevent reference path from being empty when rerouting which causes dying.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


zulfaqar-azmi-t4 (approved):
LGTM

PR #6818, opened 2024-04-15
fix(euclidean_cluster): add intensity field

badai-nguyen:
## Description

- To process and output the fields of clusters' pointcloud as the input topic including `intensity` field for `voxel_grid_based_euclidean_cluster` to solve the issue https://github.com/autowarefoundation/autoware.universe/issues/6785

## Related links

- [TIER IV INTERNAL LINK](https://tier4.atlassian.net/browse/RT1-5996?focusedCommentId=166391)

## Tests performed

- Confirmed by logging_simulator that voxel_grid_based_euclidean_cluster
   - Can process and output the topic with intensity field:

image

   - Processing time increase a little bit for a quite complicated scene
image

| Before | After | 
| -- | -- | 
|image| image |

## Pre-review checklist for the PR author

The PR author **must** check the checkboxes below when creating the PR.

- [x] I've confirmed the [contribution guidelines].
- [x] The PR follows the [pull request guidelines].

## In-review checklist for the PR reviewers

The PR reviewers **must** check the checkboxes below before approval.

- [ ] The PR follows the [pull request guidelines].
- [ ] The PR has been properly tested.
- [ ] The PR has been reviewed by the code owners.

## Post-review checklist for the PR author

The PR author **must** check the checkboxes below before merging.

- [ ] There are no open discussions or they are tracked via tickets.
- [ ] The PR is ready for merge.

After all checkboxes are checked, anyone who has write access can merge the PR.

[contribution guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/
[pull request guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/pull-request-guidelines/

PR #7734, opened 2024-06-27
fix(euclidean_cluster): fix max_cluster_size bug

badai-nguyen:
## Description

- To fix the bug of processing the cluster related to `max_cluster_size:
  
   - Due to of mis-matching of  [limitation of data copy](https://github.com/autowarefoundation/autoware.universe/blob/main/perception/euclidean_cluster/lib/voxel_grid_based_euclidean_cluster.cpp#L120) and  [max cluster size checking](https://github.com/autowarefoundation/autoware.universe/tree/main/perception/euclidean_cluster/lib/voxel_grid_based_euclidean_cluster.cpp#L135), the large cluster with number of pointcloud > `max_cluster_size` is still clutered with first `max_cluster_size` copying points. 
   - Checking by small `max_cluster_size = 100` but part of car still clustered.
image
 - Releated PR: https://github.com/autowarefoundation/autoware.universe/pull/6818
- This PR will open for review after https://github.com/autowarefoundation/autoware.universe/pull/7662 is merged.
## How was this PR tested?
 - Confirmed by logging_simulator and confirmed that a large cluster ( > `max_cluster_size`) will not be clustered.
 - Tested with  `max_cluster_size = 100`
image

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


beginningfan:
I think the problem is in [line 120](https://github.com/badai-nguyen/autoware.universe/blob/c708f98b41eb3b24cdb6bdae91b56ed657100909/perception/euclidean_cluster/lib/voxel_grid_based_euclidean_cluster.cpp#L120). When adding points to the cluster, the cluster stops adding when the size of cluster is equal to `max_cluster_size`. Therefore, this incomplete cluster will not be filtered out by [line 134](https://github.com/badai-nguyen/autoware.universe/blob/c708f98b41eb3b24cdb6bdae91b56ed657100909/perception/euclidean_cluster/lib/voxel_grid_based_euclidean_cluster.cpp#L134). According to your modification, the cluster will be filtered out by [line 134](https://github.com/badai-nguyen/autoware.universe/blob/c708f98b41eb3b24cdb6bdae91b56ed657100909/perception/euclidean_cluster/lib/voxel_grid_based_euclidean_cluster.cpp#L134) when it exits when there is one more point than `max_cluster_size`. However, this solution will cause the program to crash due to insufficient space allocated by resize in [line 109](https://github.com/badai-nguyen/autoware.universe/blob/c708f98b41eb3b24cdb6bdae91b56ed657100909/perception/euclidean_cluster/lib/voxel_grid_based_euclidean_cluster.cpp#L109), causing the program to crash. If the space is reallocated using the method I provided in #7662, or add an extra point when allocating space in [line 109](https://github.com/badai-nguyen/autoware.universe/blob/c708f98b41eb3b24cdb6bdae91b56ed657100909/perception/euclidean_cluster/lib/voxel_grid_based_euclidean_cluster.cpp#L109), this problem can be avoided.

badai-nguyen:
@beginningfan Yes, I created this PR by assuming your [PR#7662](https://github.com/autowarefoundation/autoware.universe/pull/7662) is merged. It is reason I still keep it as draft. Could you merge your PR? I'm sorry for insufficient description.

badai-nguyen:
@beginningfan I updated this PR as your PR #7662 was merged. Since you also worked on Euclidean Cluster, could I ask your review on this PR again.

beginningfan (approved):
LGTM

yukkysaito (approved):
LGTM

perception/euclidean_cluster/lib/voxel_grid_based_euclidean_cluster.cpp
```diff
@@ -117,7 +117,7 @@ bool VoxelGridBasedEuclideanCluster::cluster(
       voxel_grid_.getCentroidIndexAt(voxel_grid_.getGridCoordinates(point.x, point.y, point.z));
     if (map.find(index) != map.end()) {
       auto & cluster_data_size = clusters_data_size.at(map[index]);
-      if (cluster_data_size + point_step > std::size_t(max_cluster_size_ * point_step)) {
+      if (cluster_data_size > std::size_t(max_cluster_size_ * point_step)) {
```
YoshiRi, line 120: @badai-nguyen  Do you have any intention to not writing like this?

```suggestion
      if (cluster_data_size >= std::size_t(max_cluster_size_ * point_step)) {
```

YoshiRi: With this change, we also need to fix line139 to not filter out clusters with max_cluster_size

```
static_cast<int>(i_cluster_data_size / point_step) <= max_cluster_size_)) {
```

badai-nguyen: @YoshiRi  if change as [your suggestion](https://github.com/autowarefoundation/autoware.universe/pull/7734#discussion_r1662105558)
we need to change `static_cast<int>(i_cluster_data_size / point_step) <= max_cluster_size_)) {`  
to `static_cast<int>(i_cluster_data_size / point_step) < max_cluster_size_)) {`. 
In the result, the actual biggest cluster can be output will have `max_cluster_size_ -1` points. 
Otherwise, it will cause the same issue as current (as orrange case: also output a parth of cluster-A which actually is bigger than `max_cluster_size_')
 
Untitled Diagram

YoshiRi (approved):
LGTM

PR #7740, opened 2024-06-28
fix(autonomous_emergency_braking): aeb strange mpc polygon

danielsanchezaran:
## Description
Due to the use of resize() the path vector generates a vector full of path points with position x:0 y:0 z:0 . This causes the AEB module to generate a wrong polygon path (because the last path poses are located at the base link). 

This PR fixes the issue by doing reserve() instead, which seems like the original intention of the original commit. 

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

Before: 
cap- 2024-06-28-13-52-30
The polygon is disjointed and an NPC is wrongly detected as an obstacle.

After: 
cap- 2024-06-28-13-50-29
The polygons are correct and no collision is detected!

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/control/autoware_autonomous_emergency_braking/src/node.cpp
+++ b/control/autoware_autonomous_emergency_braking/src/node.cpp
@@ -588,11 +588,11 @@ std::optional<Path> AEB::generateEgoPath(const Trajectory & predicted_traj)
 
   // create path
   Path path;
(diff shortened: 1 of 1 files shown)


soblin (approved):
OMG!

PR #6497, opened 2024-02-27
feat(lidar_centerpoint): remove files those are no longer supported nor used

kminoda:
## Description
Remove unused files that are no longer supported
https://github.com/autowarefoundation/autoware/issues/4165

## Tests performed

Confirmed that build for the package succeeds and Autoware logging_simulator works fine with successful perception output.

## Effects on system behavior
None

Not applicable.

## Pre-review checklist for the PR author

The PR author **must** check the checkboxes below when creating the PR.

- [ ] I've confirmed the [contribution guidelines].
- [ ] The PR follows the [pull request guidelines].

## In-review checklist for the PR reviewers

The PR reviewers **must** check the checkboxes below before approval.

- [ ] The PR follows the [pull request guidelines].

## Post-review checklist for the PR author

The PR author **must** check the checkboxes below before merging.

- [ ] There are no open discussions or they are tracked via tickets.

After all checkboxes are checked, anyone who has write access can merge the PR.

[contribution guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/
[pull request guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/pull-request-guidelines/

PR #7758, opened 2024-06-30
fix(lidar_centerpoint): remove python3-open3d depend

amadeuszsz:
## Description
This is the only `<depend>python3-open3d</depend>` invocation in autoware.universe. We don't need this dependency anymore.

## Related links

**Parent Issue:**

- https://github.com/autowarefoundation/autoware.universe/pull/6497
- https://github.com/orgs/autowarefoundation/discussions/4893

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


knzo25 (approved):
LGTM

it seems the guys who deleted the single inference node forgot to delete this dependency

PR #7761, opened 2024-07-01
fix(detected_object_validation): add virtual destructor to obstacle_pointcloud_based_validator

yhisaki:
## Description

- Because a virtual class requires a virtual destructor, it has been added.
- Some refactorings

## Related links

https://stackoverflow.com/questions/461203/when-to-use-virtual-destructors

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


yukkysaito (approved):
LGTM

PR #7778, opened 2024-07-02
feat(autoware_universe_utils): reduce dependence on Boost.Geometry

mitukou1109:
## Description

This PR adds the following functions to reduce dependence on Boost.Geometry.

|Boost.Geometry <br> (`boost::geometry`)|alternative <br> (`autoware::universe_utils`)|supported types|
|----|----|----|
|`area()`|`area()`|polygon|
|`convex_hull()`|`convex_hull()`|point list|
|`correct()`|`correct()`|polygon|
|`covered_by()`|`covered_by()`|point & polygon|
|`disjoint()`|`disjoint()`|polygon & polygon|
|`distance()`|`distance()`|point & segment (2 edge points) / point & polygon|
|`equals()`|`equals()`|point & point / polygon & polygon|
|`intersects()`|`intersects()`|segment & segment / polygon & polygon|
|`touches()`|`touches()`|point & polygon|
|`within()`|`within()`|point & polygon / polygon & polygon|

Notes:
- It is assumed that polygons are convex and do not have holes.
- Polygons do not have closing points (i.e. the last vertex in the array does not match the first, as any polygon given is considered to be closed).
- `correct()` sorts outer points in clockwise order with respect to the first point AND removes duplicate points.

## Related links

**Parent Issue:**

- https://github.com/autowarefoundation/autoware.universe/issues/8128

## How was this PR tested?

```
colcon test --packages-select autoware_universe_utils
```

Benchmark result with 500 randomly generated polygons with 9 vertices:

|function|Boost|self-impl.|task|
|----|----|----|----|
|`area()`|0.03ms|0.03ms|calculate the area of 500 polygons|
|`convex_hull()`|0.38ms|0.32ms|calculate the convex hull of vertices of 500 polygons|
|`covered_by()`|309.93ms|307.29ms|check if each vertex of 500 polygons is covered by (= inside or on border of) another 500 polygons|
|`disjoint()`|198.91ms|77.51ms|check if 500 polygons are disjoint (runs totally 250,000 times)|
|`intersects()`|199.10ms|58.56ms|check if 500 polygons intersect (runs totally 250,000 times)|
|`touches()`|296.44ms|310.73ms|check if each vertex of 500 polygons touches (= on border of) another 500 polygons|
|`within()`|1058.86ms|55.80ms|check if each of 500 polygons is within 500 polygons|

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


satoshi-ota (approved):

soblin (commented):
Maybe you can change the name to `CvxPolygon` not to confuse the user

common/autoware_universe_utils/include/autoware/universe_utils/geometry/geometry.hpp
```diff
@@ -572,17 +572,70 @@ inline double calcNorm(const geometry_msgs::msg::Vector3 & v)
...
  */
 bool intersects_convex(const Polygon2d & convex_polygon1, const Polygon2d & convex_polygon2);
 
+// Alternatives for Boost.Geometry ----------------------------------------------------------------
+
+// TODO(mitukou1109): remove namespace when migrating to alternatives
+namespace alt
+{
+using Point = tf2::Vector3;
+using CvxPolygon = std::vector<tf2::Vector3>;
```
takayuki5168, line 587: `ConvexPolygon` is easier to understand. A major type of variables should not use the abbreviation.

common/autoware_universe_utils/include/autoware/universe_utils/geometry/geometry.hpp
```diff
@@ -572,17 +572,70 @@ inline double calcNorm(const geometry_msgs::msg::Vector3 & v)
...
+using PointList = std::vector<tf2::Vector3>;
+
+alt::Point fromGeom(const geometry_msgs::msg::Point & point);
```
takayuki5168, line 590: According to the coding guideline, please use the snake_case instead of camelCase.

common/autoware_universe_utils/src/geometry/geometry.cpp
```diff
@@ -361,33 +361,394 @@ bool isTwistCovarianceValid(const geometry_msgs::msg::TwistWithCovariance & twis
...
+  return (seg_end - seg_start).cross(point - seg_start).z() > 0;
+}
+
+std::optional<bool> isClockwise(const alt::CvxPolygon & poly)
+{
+  if (const auto s = area(poly)) {
+    return *s > 0;
+  } else {
+    return std::nullopt;
+  }
```
takayuki5168, line 685: Please use "early return".

takayuki5168 (approved):

common/autoware_universe_utils/test/src/geometry/test_geometry.cpp
```diff
@@ -1980,3 +1983,1053 @@ TEST(geometry, intersectPolygonRand)
...
+    points.push_back({3.7, 1.6});
+    points.push_back({3.4, 2.0});
+    points.push_back({4.1, 3.0});
+    points.push_back({5.3, 2.6});
+    points.push_back({5.4, 1.2});
+    points.push_back({4.9, 0.8});
+    points.push_back({2.9, 0.7});
+    const auto result = convex_hull(points);
+
+    EXPECT_EQ(result.vertices().size(), 7);
```
maxime-clem, line 2034: (nitpick)
With `EXPECT_...` the test will not stop even if the condition is false. In that case this can lead to a segfault in the following lines (e.g., if the size is 0).
In order to prevent tests from crashing, such condition on vector size should be checked with `ASSERT_EQ`.
```suggestion
    ASSERT_EQ(result.vertices().size(), 7);
```

common/autoware_universe_utils/include/autoware/universe_utils/geometry/geometry.hpp
```diff
@@ -583,6 +584,148 @@ std::optional<geometry_msgs::msg::Point> intersect(
  */
 bool intersects_convex(const Polygon2d & convex_polygon1, const Polygon2d & convex_polygon2);
 
+// Alternatives for Boost.Geometry ----------------------------------------------------------------
```
maxime-clem, line 587: (minor) I think the new code could be added to a separate file as the current one is becoming too big.
This can be done in a future PR.

common/autoware_universe_utils/src/geometry/geometry.cpp
```diff
@@ -390,4 +390,425 @@ bool intersects_convex(const Polygon2d & convex_polygon1, const Polygon2d & conv
...
+  const double t = v2.cross(v12) / det;
+  const double s = v1.cross(v12) / det;
+  if (t < 0 || 1 < t || s < 0 || 1 < s) {
+    return false;
+  }
+
+  return true;
+}
+
+bool intersects(const alt::ConvexPolygon2d & poly1, const alt::ConvexPolygon2d & poly2)
```
maxime-clem, line 659: GJK is already implemented (https://github.com/autowarefoundation/autoware.universe/blob/main/common/autoware_universe_utils/src/geometry/gjk_2d.cpp) but it only works for convex polygon.
If your goal is to provide an alternative to the `boost::geometry::intersects` function, then you need to use a more general algorithm that also works for concave planners.

common/autoware_universe_utils/test/src/geometry/test_geometry.cpp
```diff
@@ -1980,3 +1983,1053 @@ TEST(geometry, intersectPolygonRand)
...
+    const Point2d p2 = {2.0, -1.0};
+    const Point2d p3 = {-1.0, -1.0};
+    const Point2d p4 = {-1.0, 1.0};
+    const auto result = distance(p, ConvexPolygon2d({p1, p2, p3, p4}));
+
+    EXPECT_NEAR(result, 0.0, epsilon);
+  }
+}
+
+TEST(geometry, intersects)
```
maxime-clem, line 2271: This test should include non-convex polygons.

common/autoware_universe_utils/test/src/geometry/test_geometry.cpp
```diff
@@ -1980,3 +1983,1053 @@ TEST(geometry, intersectPolygonRand)
...
+      "\tNot disjoint:\n\t\tBoost::geometry = %2.2f ms\n\t\tAlt = %2.2f ms\n",
+      ground_truth_not_disjoint_ns / 1e6, alt_not_disjoint_ns / 1e6);
+    std::printf(
+      "\tTotal:\n\t\tBoost::geometry = %2.2f ms\n\t\tAlt = %2.2f ms\n",
+      (ground_truth_not_disjoint_ns + ground_truth_disjoint_ns) / 1e6,
+      (alt_not_disjoint_ns + alt_disjoint_ns) / 1e6);
+  }
+}
+
+TEST(geometry, intersectsRand)
```
maxime-clem, line 2843: This test (and probably the other random tests as well) should also use random non-convex polygons.
@mraditya01 is working on random polygon generation so maybe you can coordinate with him.

maxime-clem (commented):
I did not check all the code in details yet but I think there are a few issues to fix because the PR can be merged.
- the alternative `intersects` only works for convex polygons (whereas the `boost` implementation also works for non-convex polygons).
- the random test only use convex polygons.
- GJK is already implemented so we should see which implementation to keep.

common/autoware_universe_utils/src/geometry/geometry.cpp
```diff
@@ -390,4 +390,425 @@ bool intersects_convex(const Polygon2d & convex_polygon1, const Polygon2d & conv
...
+      std::min(min_distance, distance(point, vertices[i], vertices[(i + 1) % vertices.size()]));
+  }
+
+  return min_distance;
+}
+
+bool equals(const alt::Point2d & point1, const alt::Point2d & point2)
+{
+  return std::abs(point1.x() - point2.x()) <= std::numeric_limits<double>::epsilon() &&
+         std::abs(point1.y() - point2.y()) <= std::numeric_limits<double>::epsilon();
```
maxime-clem, line 625: From what I read (e.g., [here](https://en.cppreference.com/w/cpp/types/numeric_limits/epsilon) or [here](https://stackoverflow.com/questions/48133572/what-can-stdnumeric-limitsdoubleepsilon-be-used-for)), using `std::numeric_limits<double>::epsilon()` directly is not a good way to check for precision when comparing two numbers.
I think it would make more sense to define an epsilon based on the accuracy we want to reach (e.g., `1e-3` for mm accuracy).

common/autoware_universe_utils/test/src/geometry/test_geometry.cpp
mitukou1109: Fixed: https://github.com/autowarefoundation/autoware.universe/pull/7778/commits/1e825e247538b5a5327f2f547f8ecff8a21524db

common/autoware_universe_utils/src/geometry/geometry.cpp
mitukou1109: I agree, but I already named the class for representing polygons `ConvexPolygon2d` and thus I do not expect the input to be concave (at least for now).

PR #7784, opened 2024-07-02
feat(mrm_handler): operate mrm only when autonomous operation mode

kyoichi-sugahara:
## Description

By switching to the diagnostics graph in the [PR](https://github.com/autowarefoundation/autoware_launch/pull/1043), a problem arose where the state becomes emergency when scenario_simulator_v2 starts up, causing the simulation to fail.

This change fixes the above problem by operating MRM only during autonomous operation mode.

## How was this PR tested?

- [x] run psim
- [x] run simualtion in local environment

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/system/mrm_handler/src/mrm_handler/mrm_handler_core.cpp
+++ b/system/mrm_handler/src/mrm_handler/mrm_handler_core.cpp
@@ -538,7 +538,7 @@ bool MrmHandler::isEmergency() const
          is_operation_mode_availability_timeout;
 }
 
-bool MrmHandler::isAutonomous()
+bool MrmHandler::isControlModeAutonomous()
 {
   using autoware_vehicle_msgs::msg::ControlModeReport;
   auto mode = sub_control_mode_.takeData();
--- a/system/mrm_handler/src/mrm_handler/mrm_handler_core.cpp
+++ b/system/mrm_handler/src/mrm_handler/mrm_handler_core.cpp
@@ -396,13 +396,13 @@ void MrmHandler::updateMrmState()
   }
 
   // Get mode
-  const bool is_auto_mode = isAutonomous();
-  const bool is_operation_mode_auto_mode = isOperationModeAutonomous();
+  const bool is_control_mode_autonomous = isControlModeAutonomous();
+  const bool is_operation_mode_autonomous = isOperationModeAutonomous();
 
   // State Machine
   switch (mrm_state_.state) {
     case MrmState::NORMAL:
-      if (is_auto_mode && is_operation_mode_auto_mode) {
+      if (is_control_mode_autonomous && is_operation_mode_autonomous) {
         transitionTo(MrmState::MRM_OPERATING);
       }
       return;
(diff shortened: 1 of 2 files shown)


system/mrm_handler/src/mrm_handler/mrm_handler_core.cpp
```diff
@@ -397,11 +397,12 @@ void MrmHandler::updateMrmState()
 
   // Get mode
   const bool is_auto_mode = isAutonomous();
```
veqcc, line 399: @kyoichi-sugahara 
Could you rename the followings in terms of readability?
- `is_auto_mode` -> `is_control_mode_autonomous`
- `is_operation_mode_auto_mode` -> `is_operation_mode_autonomous`
- `isAutonomous()` -> `isControlModeAutonomous()

kyoichi-sugahara: @veqcc 
Addressed in https://github.com/autowarefoundation/autoware.universe/pull/7784/commits/18da2cb123f9e0c140803042a1e772088ccd5b9e
Can you approve the PR?

veqcc: Thank you!

veqcc: @kyoichi-sugahara 
You may have to change the function name here:
```
bool MrmHandler::isAutonomous()
{
  using autoware_vehicle_msgs::msg::ControlModeReport;
  auto mode = sub_control_mode_.takeData();
  if (mode == nullptr) return false;
  return mode->mode == ControlModeReport::AUTONOMOUS;
}
```

kyoichi-sugahara: @veqcc 
Sorry forgot to change and addressed in https://github.com/autowarefoundation/autoware.universe/pull/7784/commits/7297667b9f654875a5b5b1178f989ae60f670529

veqcc:
@kyoichi-sugahara 
I have a quiestion on the behavior of `takeData()`.
If it comsums the data in the subscription queue, then the following steps will happen, right?

See the code block first
```
  // Get mode
  const bool is_control_mode_autonomous = isControlModeAutonomous();
  const bool is_operation_mode_autonomous = isOperationModeAutonomous();

  // State Machine
  switch (mrm_state_.state) {
    case MrmState::NORMAL:
      if (is_control_mode_autonomous && is_operation_mode_autonomous) {
        transitionTo(MrmState::MRM_OPERATING);
      }

    case MrmState::MRM_OPERATING:
      ...
      if (isArrivedAtGoal()) {
        transitionTo(MrmState::MRM_SUCCEEDED);
      }
      ...
  }
```

step 1. consume `sub_operation_mode_state_.takeData();` in `isOperationModeAutonomous();`
step 2. the process go into `case MrmState::MRM_OPERATING:` block
step 3. `isArrivedAtGoal()` also tries to consume `sub_operation_mode_state_.takeData();`, but it is already consumed, so always return false


If `takeData` is not consuming, then there seems no problem.

kyoichi-sugahara:
@veqcc 
I understand your concern and `takeData()` is not consuming :+1:

veqcc (approved):
LGTM, thank you!!

isamu-takagi (approved):

Issue #7769, opened 2024-07-01
Add CODECOV_TOKEN to autowarefoundation repositories


xmfcx:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I've agreed with the maintainers that I can plan this task.

### Description

Transitioning from `codecov/codecov-action@v3` to `codecov/codecov-action@v4`, a token is required.

I've added an organization wide `CODECOV_TOKEN` variable that is accessible through `token: ${{ secrets.CODECOV_TOKEN }}`.
This will work for all https://app.codecov.io/github/autowarefoundation repositories.

Forks won't need the token, they are handled automatically:
- https://docs.codecov.com/docs/codecov-uploader#if-youre-using-the-action

### Definition of done

-

xmfcx:
I've updated all tokens for v4 for autowarefoundation organization.

PR #7767, opened 2024-07-01
ci: add codecov token

xmfcx:
## Description

Fix: https://github.com/autowarefoundation/autoware.universe/actions/runs/9740736192/job/26878461354#step:17:57

## Related links

**Parent Issue:**
- https://github.com/autowarefoundation/autoware.universe/issues/7769

- https://github.com/codecov/codecov-action?tab=readme-ov-file#usage

For forks:
- https://docs.codecov.com/docs/codecov-uploader#if-youre-using-the-action
   - > The [Codecov GitHub Action](https://github.com/codecov/codecov-action) is capable of automatically detecting that it is running from a fork and will automatically correctly populate the branch parameter when uploading.
   - Should work out of the box but did not test for forks yet.

## How was this PR tested?

- https://github.com/autowarefoundation/autoware.universe/pull/7767#issuecomment-2199688434

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

PR #7793, opened 2024-07-02
ci(codecov.yaml): make ci succeed regardless of coverage

xmfcx:
## Description

Since CodeCov configuration was fixed with:
- https://github.com/autowarefoundation/autoware.universe/pull/7767

We started having these checks as errors in CI.

Since we are not enforcing this check to pass, I'm reducing the thresholds to allow anything to pass these tests.

And we'll use it as an informative tool for the time being.

https://github.com/autowarefoundation/autoware.universe/commits/main/

image

## Related links

- https://docs.codecov.com/docs/commit-status#project-status

**Partly Related Issue:**
- https://github.com/autowarefoundation/autoware.universe/issues/7769

## How was this PR tested?

Since this works with a GitHub App, I guess we can't test it unless we merge it :D

It's all well documented and change is minimal so it should be alright.

## Notes for reviewers

Here I am pasting the explanation of the values that I've changed.

In summary, I'm making the thresholds as inclusive as possible to make them pass regardless of the coverage change.

### target

https://docs.codecov.com/docs/commit-status#target

> `auto | <number>`
> Choose a minimum coverage ratio that the commit must meet to be considered a `success`.
> 
> `auto` will use the coverage from the base commit (pull request base or parent commit) coverage to compare against. This is the default.
> `<number>` you can specify a target of an exact coverage number such as `75%` or `100%` (`string, int, or float` accepted).

### threshold

https://docs.codecov.com/docs/commit-status#threshold

> `<number>`
> Allow the coverage to drop by `<number>%`, and posting a `success` status.

## Interface changes

None.

## Effects on system behavior

CI will pass

### DIFF
--- a/codecov.yaml
+++ b/codecov.yaml
@@ -1,11 +1,15 @@
+# https://docs.codecov.com/docs/commit-status#project-status
+
 coverage:
   status:
     project:
       default:
-        target: auto
+        target: 0%  # Make CI always succeed
+        threshold: 100%   # Make CI always succeed
     patch:
       default:
-        target: auto
+        target: 0%  # Make CI always succeed
+        threshold: 100%  # Make CI always succeed
 
 comment:
   show_carryforward_flags: true
--- a/codecov.yaml
+++ b/codecov.yaml
@@ -4,12 +4,12 @@ coverage:
   status:
     project:
       default:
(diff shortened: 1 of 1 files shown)


xmfcx:
I'll merge it to test it, revert if it fails.

PR #7803, opened 2024-07-03
ci(codecov.yaml): make it pass regardless of coverage

xmfcx:
## Description

image

I think I should have updated these lines too.

It got better, only 2 are failing now. With these too, all should pass.

## Related links

**Continuation from:**

- https://github.com/autowarefoundation/autoware.universe/pull/7793

## Notes for reviewers

The targets for these were changed [in the previous PR](https://github.com/autowarefoundation/autoware.universe/pull/7793):
image

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/codecov.yaml
+++ b/codecov.yaml
@@ -20,7 +20,9 @@ flag_management:
     statuses:
       - name_prefix: project-
         type: project
-        target: auto
+        target: 0% # Make CI always succeed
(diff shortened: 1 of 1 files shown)


xmfcx:
image

Seems fixed!
cc. @esteve

PR #7804, opened 2024-07-03
refactor(autoware_object_merger): move headers to src and rename package

esteve:
## Description

Move headers to src and rename package

## Related links

**Parent Issue:**

- https://github.com/autowarefoundation/autoware/issues/4569

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

PR #7809, opened 2024-07-03
refactor(autoware_tracking_object_merger): move headers to include/autoware and rename package

esteve:
## Description

move headers to include/autoware and rename package

## Related links

**Parent Issue:**

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


technolojin (changes requested):
1. package renaming
I am thinking to do it after all the other packages are ready.

2. moving headers
the association class will be replaced by shared library. mentioned in https://github.com/autowarefoundation/autoware.universe/pull/7804#pullrequestreview-2157696734

esteve:
> 1. package renaming
>    I am thinking to do it after all the other packages are ready.

Do you have an estimate for when this will be done? Why not do it now for each package? We've done it for other subsystems (planning, localization, common, etc.)

> 2. moving headers
>    the association class will be replaced by shared library. mentioned in [refactor(autoware_object_merger): move headers to src and rename package #7804 (review)](https://github.com/autowarefoundation/autoware.universe/pull/7804#pullrequestreview-2157696734)

It's best to do it now and then if you want to create a third package that provides a shared library, then you can do it, but there's no estimate when the shared library will be available and instead this PR already contains the changes now. You can create the shared library after the changes in this PR (and the other PRs) are merged.

technolojin:
@esteve 
1. please see https://github.com/autowarefoundation/autoware/issues/4569#issuecomment-2212668662
2. Relocating the headers is optional. The other components (planning, localization) it was not done. Why you want to do it in the same time? It was not mentioned in your [ISSUES](https://github.com/autowarefoundation/autoware/issues/4569).

mitsudome-r:
@esteve 
Thanks for creating the PR.
It seems like @technolojin would like to split this PR into two. Could you drop the changes with the change in header location and just rename the package in this PR?

esteve:
@mitsudome-r sure, no problem. However, the headers need to be moved anyway, either to a private location in `src` (what I did in this PR), or to a public one `include/autoware/tracking_object_merger`, so I used the opportunity to move them to a private location. I'll move the headers to the latter to keep them public.

technolojin (changes requested):
Can you fix the launchers?

`$(find-pkg-share tracking_object_merger) -> $(find-pkg-share autoware_tracking_object_merger)`

technolojin (approved):
I fixed the launchers.

PR #7814, opened 2024-07-03
fix(radar_fusion_to_detected_object): fix knownConditionTrueFalse

veqcc:
## Description

This is a cppcheck warning

```
perception/radar_fusion_to_detected_object/src/radar_fusion_to_detected_object.cpp:98:32: style: Condition 'split_objects.size()==1' is always true [knownConditionTrueFalse]
      if (split_objects.size() == 1) {
                               ^
```

What did you mean here?
Maybe my modification `split_object.size() == 1` is incorrect.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/radar_fusion_to_detected_object/src/radar_fusion_to_detected_object.cpp
+++ b/perception/radar_fusion_to_detected_object/src/radar_fusion_to_detected_object.cpp
@@ -95,7 +95,7 @@ RadarFusionToDetectedObject::Output RadarFusionToDetectedObject::update(
     for (auto & split_object : split_objects) {
       // set radars within objects
       std::shared_ptr<std::vector<RadarInput>> radars_within_split_object;
-      if (split_objects.size() == 1) {
+      if (split_object.size() == 1) {
         // If object is not split, radar data within object is same
         radars_within_split_object = radars_within_object;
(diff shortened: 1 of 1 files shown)


technolojin:
I think that is also weird. 

@scepter914 Can you look at this point?

scepter914:
The function of splitting objects is not implemented yet.
https://github.com/autowarefoundation/autoware.universe/blob/f7b9180d29713fe177b158ea19d7abb994c12b65/perception/radar_fusion_to_detected_object/src/radar_fusion_to_detected_object.cpp#L89

So the case that always outputs `'split_objects.size()==1' is always true` works as expected.
It derive from my first design, but I think this package is not used for product and development of this package is low priority for now.

veqcc:
@scepter914 
>So the case that always outputs 'split_objects.size()==1' is always true works as expected.
It derive from my first design, but I think this package is not used for product and development of this package is low priority for now.

OK, then I will add inline suppression for cppcheck.
If you complete your implementation, please remove the inline suppression then :+1:

scepter914 (approved):
LGTM

PR #7789, opened 2024-07-02
refactor(stop_filter): prefix package and namespace with autoware

esteve:
## Description

prefix package and namespace with autoware

## Related links

**Parent Issue:**

- Link https://github.com/autowarefoundation/autoware/issues/4569

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

PR #7818, opened 2024-07-04
revert(stop_filter): revert "prefix package and namespace"

SakodaShintaro:
## Description
This reverts commit fbbc44ceb33c423c00ba3caa826ff77c8958d920.

I don't know the cause, but I'm reverting because the change prevents the stop_filter node from launching and `/localization/kinematic_state` is not output.

## Related links

- https://github.com/autowarefoundation/autoware.universe/pull/7789

## How was this PR tested?

* logging_simulator
* AWSIM +  Autoware

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


SakodaShintaro:
The bug has been fixed in the following pull request, so this pull request will be closed.
https://github.com/autowarefoundation/autoware.universe/pull/7820

PR #7820, opened 2024-07-04
fix(stop_filter): fix plugin name

SakodaShintaro:
## Description
This PR fix the plugin name of `stop_filter`.

If this fix is not applied, the `stop_filter` node will not be launched properly and the `/localization/kinematic_state` topic will not be output.

## Related links

- https://github.com/autowarefoundation/autoware.universe/pull/7789

## How was this PR tested?
* Autoware + AWSIM works
* logging_simulator works

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


KYabuuchi (approved):
I have confirmed that stop_filter is correctly launched with this fix PR :ok_man:

PR #7828, opened 2024-07-04
feat(safety_check): filter safety check targe objects by yaw deviation between pose and lane

satoshi-ota:
## Description

Sometimes, the safety check caused false positive because it couldn't judge which lane the object belonged to properly.

**Example**: the red lanes mean safety check target lane for avoidance maneuver in following movie and a vehicle is trying to merge ego lane. Basically, the avoidance module only check objects which are on shift side adjacent lane, so the ego don't have to revert avoidance path due to the object in this case. But the object becomes safety check target when it's on safety check target lane even though the direction is different from the lane direction.

https://github.com/autowarefoundation/autoware.universe/assets/44889564/2b276be3-a8b3-4356-a4c3-1bb8261c392e

---

After this PR, we're going to be able to set yaw deviation threshold and ignore the object whose pose is not along with the lane. Additionally, the avoidance module set the threshold to `PI/2` by default.As a result, the object won't be a safety check target even though it's on safety target lane (red lane) in following scene.

:warning: But the avoidance module filters objects by yaw deviation only when the object is moving because the head direction is not reliable while object is stopping.

https://github.com/autowarefoundation/autoware.universe/assets/44889564/bfbeb69a-aa1f-4620-8bb2-815069f7b8ff

On the other hand, if the yaw deviation is less than threshold, the safety check module reacts the object in the same way as original.

https://github.com/autowarefoundation/autoware.universe/assets/44889564/5022efd2-1352-44c4-a5a7-ecf9f3383003

## Related links

**Parent Issue:**

- [TIER IV INTERNAL LINK](https://docs.google.com/presentation/d/1SFzKrDvpwxCm1naLeda-oqcpvzIQkKPURID-qRJCQe4/edit#slide=id.g272e7af22cd_4_63)

## How was this PR tested?

- [x]  [PASS TIER IV INTERNAL SCENARIOS](https://evaluation.tier4.jp/evaluation/reports/ea3701dc-6b2b-5024-aa54-35d4e0ae6e1b?project_id=prd_jt)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_path_planner/autoware_behavior_path_avoidance_by_lane_change_module/src/scene.cpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_avoidance_by_lane_change_module/src/scene.cpp
@@ -151,8 +151,8 @@ void AvoidanceByLaneChange::fillAvoidanceTargetObjects(
   const auto [object_within_target_lane, object_outside_target_lane] =
     utils::path_safety_checker::separateObjectsByLanelets(
       *planner_data_->dynamic_object, data.current_lanelets,
-      [](const auto & obj, const auto & lane) {
-        return utils::path_safety_checker::isPolygonOverlapLanelet(obj, lane);
+      [](const auto & obj, const auto & lane, const auto yaw_threshold) {
+        return utils::path_safety_checker::isPolygonOverlapLanelet(obj, lane, yaw_threshold);
       });
 
   // Assume that the maximum allocation for data.other object is the sum of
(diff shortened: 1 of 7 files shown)


soblin (approved):

go-sakayori (approved):
LGTM for the static obstacle avoidance part

kosuke55 (approved):

danielsanchezaran (approved):

zulfaqar-azmi-t4 (approved):

PR #7832, opened 2024-07-04
fix(multi_object_tracker): fix knownConditionTrueFalse

veqcc:
## Description

This is a fix based on cppcheck `knownConditionTrueFalse` warings

```
perception/multi_object_tracker/src/data_association/successive_shortest_path/successive_shortest_path.cpp:70:7: style: Condition 'sparse_cost' is always true [knownConditionTrueFalse]
  if (sparse_cost) {
      ^
perception/multi_object_tracker/src/data_association/successive_shortest_path/successive_shortest_path.cpp:51:22: note: Assignment 'sparse_cost=true', assigned value is 1
  bool sparse_cost = true;
                     ^
perception/multi_object_tracker/src/data_association/successive_shortest_path/successive_shortest_path.cpp:61:24: note: Assuming condition is false
  if (cost.size() == 0 || cost.at(0).size() == 0) {
                       ^
perception/multi_object_tracker/src/data_association/successive_shortest_path/successive_shortest_path.cpp:70:7: note: Condition 'sparse_cost' is always true
  if (sparse_cost) {
      ^
```

@technolojin 
If this is an acceptable change, then I will make the same changes on `object_merger`, `radar_object_tracker` and `tracking_object_merger` packages in this PR.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


technolojin (changes requested):
@veqcc I think this is not the case. 
`sparse_cost` may need to be defined outside the function instead.

veqcc:
@technolojin 
Thank you for your quick response!

> @veqcc I think this is not the case.
sparse_cost may need to be defined outside the function instead.

Understood.
Do you have any idea for avoiding cpocheck warnings, or is there any plan to refactor these packages so that `sparse_cost` is not hard coded any more?

veqcc:
@technolojin @YoshiRi @yukkysaito 
Hi

perception/multi_object_tracker/src/data_association/successive_shortest_path/successive_shortest_path.cpp
```diff
@@ -47,10 +47,6 @@ void SSP::maximizeLinearAssignment(
   const std::vector<std::vector<double>> & cost, std::unordered_map<int, int> * direct_assignment,
   std::unordered_map<int, int> * reverse_assignment)
 {
-  // NOTE: Need to set as default arguments
```
YoshiRi, line 50: With this statement, I think it's better to define sparse_cost in the arguments and give default true value.

perception/multi_object_tracker/src/data_association/successive_shortest_path/successive_shortest_path.cpp
```diff
@@ -47,10 +47,6 @@ void SSP::maximizeLinearAssignment(
   const std::vector<std::vector<double>> & cost, std::unordered_map<int, int> * direct_assignment,
   std::unordered_map<int, int> * reverse_assignment)
```
YoshiRi, line 48:
```suggestion
  std::unordered_map<int, int> * reverse_assignment, const bool sparse_cost = true)
```

With the Notation in the original code, how about this? ( Need .hpp changes )

YoshiRi (changes requested):
How about put the variable into arguments with default value?

perception/multi_object_tracker/src/data_association/successive_shortest_path/successive_shortest_path.cpp
veqcc: Thank you for your advice!! I will accepr your suggestion.

veqcc:
I will close this PR and create a new PR in which I will solve all the warnings around `sparse_cost` in object_merger, radar_object_tracker and tracking_object_merger with the same way here
> With this statement, I think it's better to define sparse_cost in the arguments and give default true value.

PR #7840, opened 2024-07-04
fix(motion_velocity_planner): use the slowdown velocity (instead of 0)

maxime-clem:
## Description

Currently slowdowns in the `motion_velocity_planner` are incorrectly setting a velocity of 0.
This PR fixes the issue by using the correct velocity.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


Owen-Liuyuxuan:
I have also tried out local tests in an internal environment. It worked as expected.

takayuki5168 (approved):

danielsanchezaran (approved):

PR #7853, opened 2024-07-04
feat(universe_utils): add GJK implementation for 2D convex polygon collision check

maxime-clem:
## Description
Implement the GJK algorithm allowing ~4x speed compared to `boost::geometry::intersects` to determine if 2 polygons intersects.
:warning: the polygons must be convex.

## How was this PR tested?

Unit tests.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/common/autoware_universe_utils/include/autoware/universe_utils/geometry/geometry.hpp
+++ b/common/autoware_universe_utils/include/autoware/universe_utils/geometry/geometry.hpp
@@ -1,4 +1,4 @@
-// Copyright 2020 Tier IV, Inc.
+// Copyright 2020-2024 Tier IV, Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/common/autoware_universe_utils/include/autoware/universe_utils/geometry/geometry.hpp
+++ b/common/autoware_universe_utils/include/autoware/universe_utils/geometry/geometry.hpp
@@ -18,8 +18,11 @@
 #include "autoware/universe_utils/geometry/boost_geometry.hpp"
 #include "autoware/universe_utils/math/constants.hpp"
 #include "autoware/universe_utils/math/normalization.hpp"
+#include "autoware/universe_utils/ros/msg_covariance.hpp"
 
+#include <exception>
 #include <string>
+#include <vector>
 
 #define EIGEN_MPL2_ONLY
 #include <Eigen/Core>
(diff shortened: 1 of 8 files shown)


common/autoware_universe_utils/include/autoware/universe_utils/geometry/geometry.hpp
```diff
@@ -577,6 +577,12 @@ std::optional<geometry_msgs::msg::Point> intersect(
   const geometry_msgs::msg::Point & p1, const geometry_msgs::msg::Point & p2,
   const geometry_msgs::msg::Point & p3, const geometry_msgs::msg::Point & p4);
 
+/**
+ * @brief Check if 2 convex polygons intersect using the GJK algorithm
+ * @details much faster than boost::geometry::intersects()
+ */
+bool intersects(const Polygon2d & convex_polygon1, const Polygon2d & convex_polygon2);
```
takayuki5168, line 584: Should we add the `convex` naming to this function so that developers won't use this function for non-convex polygons?

common/autoware_universe_utils/include/autoware/universe_utils/geometry/gjk_2d.hpp
```diff
@@ -0,0 +1,29 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef AUTOWARE__UNIVERSE_UTILS__GEOMETRY__GJK_2D_HPP_
+#define AUTOWARE__UNIVERSE_UTILS__GEOMETRY__GJK_2D_HPP_
+
+#include "autoware/universe_utils/geometry/boost_geometry.hpp"
+
+namespace autoware::universe_utils::gjk
```
takayuki5168, line 20: Do we have to add the `gjk` namespace here?

takayuki5168: Ah, you have `intersects` and `gjk::intersects`. Now I understand the reason.

takayuki5168 (approved):

common/autoware_universe_utils/include/autoware/universe_utils/geometry/geometry.hpp
maxime-clem: Done :+1: [7037fb1](https://github.com/autowarefoundation/autoware.universe/pull/7853/commits/7037fb1d5852d379377ed33b11b822b21b0f683d)

Issue #7106, opened 2024-05-23
Add deep learning based 3D bouding box shape estimation algorithm


kaancolak:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I've agreed with the maintainers that I can plan this task.

### Description

In the current version of Autoware, the L-shape fitting algorithm is used to find 3D bounding boxes for vehicles from object clusters.  However, this algorithm has some limitations for accurately identifying these bounding boxes. If the vehicle's point cloud includes side mirrors, antennas, or only shows the back of the vehicle (forming an I-shape), the algorithm often fails. Additionally, it can struggle in situations where there are only a few points available.

The Shape-fitting algorithm has crucial roles especially when using camera-lidar fusion or using Apollo instance segmentation as an object detector.  

L shape fitting results : 

https://github.com/autowarefoundation/autoware.universe/assets/12658936/fef4f1ae-b364-473e-aaa4-b55e6eb69f30

### Purpose

Add a neural network-based 3D bounding box shape estimation algorithm. This could be an optional feature in the **shape_estimation** package, allowing users to choose it if desired.

### Possible approaches

The point-based method can be used for estimating object bounding boxes. Typically, networks like PointNet/PointNet++ are used for classifying point clouds or performing point-wise instance segmentation, but they can be easily adapted to estimate 3D bounding boxes of objects.

### Definition of done

[ ] - Train and evaluate prototype model 
[ ] - Prepare the TensorRT inference engine and update **shape_estimation** package
[ ] - Add model to the model zoo and prepare documentation  about how to train

kaancolak:
I developed two prototype models using PointNet and PointNet++. Each model is a combination of two sub-models: a regression transform net to estimate the object's center and a PointNet-based network to predict bounding box dimensions and yaw angles. The bounding box estimation part from the Frustum-PointNet paper was used as a reference.

The models take point clouds and labels as inputs.

The PointNet-based model can process around 2,000 objects using pure Python and a 3060 mobile GPU. By using TensorRT quantization and a more powerful GPU, it can process over 4,000 objects without performance issues. It looks applicable to Autoware. 

Although the PointNet++ model is slightly more accurate, it is only half as fast. Pure Python version processing ~900 objects each second. 

[...]

PR #7860, opened 2024-07-05
feat(shape_estimation): add ml shape estimation

kaancolak:
## Description

Pointnet-based vehicle shape estimation module. 

Choosing the estimation method is optional in the config file. 
 

## Related links

ONNX model file: https://drive.google.com/drive/folders/1xj95FShWo3_lL-1wtXMBB7T2OR8oFxB8

**Parent Issue:**
https://github.com/autowarefoundation/autoware.universe/issues/7106

## Notes for reviewers

Test using the given ONNX file, no need to any other change

## Interface changes

None.

## Effects on system behavior

None.


yukkysaito:
@kaancolak This is really amazing! Could you please describe the mechanism and node parameters in the README? If there is a learning method, please also include the user's learning procedure.

kaancolak:
Thank you @yukkysaito  -san, I will prepare documentation about model details and also how to train the model/convert to ONNX.

kaancolak:
@yukkysaito @YoshiRi @miursh 

I updated the readme file that added retraining/deploying guidelines. Please check it out and feel free to share your thoughts.

perception/shape_estimation/README.md
```diff
@@ -10,7 +10,8 @@ This node calculates a refined object shape (bounding box, cylinder, convex hull
 
 - bounding box
 
-  L-shape fitting. See reference below for details.
+  - L-shape fitting: See reference below for details
+  - ML based shape fitting: See ML Based Shape Fitting Impelementation section below for details
```
yukkysaito, line 14: Typo :pray: 
```suggestion
  - ML based shape fitting: See ML Based Shape Fitting Implementation section below for details
```

kaancolak:
@yukkysaito @YoshiRi @miursh 

Could you review or assign a proper reviewer?

YoshiRi:
Hi. @kaancolak
Thanks for your PR.


## Test with sample rosbag

I built and checked your ml based shape estimation with sample rosbag and confirmed it works.

- how to test
  - Let detection_by_tracker_node to published cluster with class label (need slight source code change)
  - use that output to check ml based shape estimation

sample objects after fitting is like following:

| title | figure |
|--|--|
| normal vehicle |Screenshot from 2024-07-19 15-08-46|
|trucks with I-shaped cluster(Not L-shape) | Screenshot from 2024-07-19 15-09-02|
| trailer with L-shape cluster| Screenshot from 2024-07-19 15-09-09|


## Discussion

I have several opinion and question.

- Based on the current CMakeLists setup, it seems that the build process requires a CUDA environment. However, there are situations where we might want to build and run without a CUDA environment. To accommodate this, we should either use #ifdef to conditionally compile or separate the node that uses ML.
- The shape_estimator is also called in other places, such as detection_by_tracker, so we should aim to unify the interface for both ML and non-ML shape_estimator calls as much as possible. It might be better for the non-ML side to accept a list of objects as input instead of individual objects. However, for this PR, it might be acceptable to ignore this issue for now.
- Just a pure question, the ML estimator is designed for vehicle classes, but what kind of output does it produce if the input belongs to other classes?

perception/shape_estimation/CMakeLists.txt
```diff
@@ -10,6 +10,15 @@ find_package(pcl_conversions REQUIRED)
...
 find_package(Eigen3 REQUIRED)
 
+find_package(CUDA)
+find_package(CUDNN)
+find_package(TENSORRT)
+
+if(NOT (${CUDA_FOUND} AND ${CUDNN_FOUND} AND ${TENSORRT_FOUND}))
+  message(WARNING "cuda, cudnn, tensorrt libraries are not found")
+  return()
+endif()
```
YoshiRi, line 20: This package is used in non-cuda environement, so I do not want to skip build.

perception/shape_estimation/config/shape_estimation_ml.param.yaml
```diff
@@ -0,0 +1,8 @@
+/**:
+  ros__parameters:
+    model_params:
+      use_ml_shape_estimator: false
```
YoshiRi, line 4: [imho]
If we separate nodes with or without cuda dependency, this parameter should written in shape_estimation.param.yaml

perception/shape_estimation/include/autoware/shape_estimation/tensorrt_shape_estimator.hpp
```diff
@@ -0,0 +1,92 @@
...
+public:
+  TrtShapeEstimator(
+    const std::string & model_path, const std::string & precision,
+    const tensorrt_common::BatchConfig & batch_config, const size_t max_workspace_size = (1 << 30),
+    const tensorrt_common::BuildConfig build_config =
+      tensorrt_common::BuildConfig("MinMax", -1, false, false, false, 0.0));
+
+  ~TrtShapeEstimator();
+
+  bool inference(const DetectedObjectsWithFeature & input, DetectedObjectsWithFeature & output);
```
YoshiRi, line 58: [just a comment]
I think it's better to share interface between ml and non-ml shape estimation in the future update.

kaancolak:
@YoshiRi -san, thank you for your review. Sorry, I haven't been looked for last week.

Like your comment, I separated the cuda and non-cuda parts of the code that are controlled by cuda and cuda library founding flags.  

> Just a pure question, the ML estimator is designed for vehicle classes, but what kind of output does it produce if the input belongs to other classes?

The model takes one hot vector and point cloud as input. Yet if we feed different objects than vehicle labels,  it will probably return the closest vehicle shape to the fed point cloud. The model could be easily extended for other classes according to the training readme file.

perception/shape_estimation/include/autoware/shape_estimation/tensorrt_shape_estimator.hpp
kaancolak: I completely agree with that. It will make the package better as a design.

YoshiRi (approved):
LGTM and worked with my local environement

perception/autoware_shape_estimation/CMakeLists.txt
```diff
@@ -10,27 +10,44 @@ find_package(pcl_conversions REQUIRED)
...
-  lib/filter/car_filter.cpp
-  lib/filter/bus_filter.cpp
-  lib/filter/truck_filter.cpp
-  lib/filter/trailer_filter.cpp
-  lib/filter/no_filter.cpp
-  lib/filter/utils.cpp
-  lib/corrector/utils.cpp
-  lib/corrector/no_corrector.cpp
+set(${PROJECT_NAME}_SOURCES
+        lib/shape_estimator.cpp
```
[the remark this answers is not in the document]
kaancolak: thank you warning, I fixed the indent

perception/autoware_shape_estimation/lib/tensorrt_shape_estimator.cpp
```diff
@@ -0,0 +1,338 @@
...
+           y_iter(feature_object.feature.cluster, "y"), z_iter(feature_object.feature.cluster, "z");
+           x_iter != x_iter.end(); ++x_iter, ++y_iter, ++z_iter) {
+        input_pc_h_[i * input_chan * input_pc_size + 0 + offset] = -1 * (*y_iter);
+        input_pc_h_[i * input_chan * input_pc_size + 512 + offset] = -1 * (*z_iter);
+        input_pc_h_[i * input_chan * input_pc_size + 1024 + offset] = *x_iter;
+        offset++;
+      }
+
+      int iter_count = static_cast<int>(input_pc_size) / point_size_of_cloud;
+      int remainer_count = static_cast<int>(input_pc_size) % point_size_of_cloud;
```
Shin-kyoto, line 142:
```suggestion
      int remainder_count = static_cast<int>(input_pc_size) % point_size_of_cloud;
```

Spell check failed, so can you change `remainer` to another word? I think `reminder` is better.
https://github.com/autowarefoundation/autoware.universe/actions/runs/10196767728/job/28208175986?pr=7860

YoshiRi:
Rerun CI to check builds. Please wait for a while.

PR #7869, opened 2024-07-05
fix(bluetooth_monitor): apply cppcheck-suppress for cstyleCast

Koichi98:
## Description
This is a fix based on cppcheck warnings.
```
src/universe/autoware.universe/system/bluetooth_monitor/service/l2ping_service.cpp:63:24: style: C-style pointer casting [cstyleCast]
  ret = bind(socket_, (struct sockaddr *)&address, sizeof(address));
```
These nine pieces of cstyleCast code are suppressed because they are actually written in cstyle.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


ito-san (approved):
LGTM

PR #7137, opened 2024-05-27
feat(pointcloud_preprocessor): support for 3d distortion correction algorithm and refactor distortion correction node

vividf:
## Description
This PR solved the issue https://github.com/autowarefoundation/autoware.universe/issues/6657.
This PR provides an option for the user to choose a 3d distortion corrector instead of using the current 2d distortion corrector.
 
Note that by using this, the time will increase around 50 percent. 
This PR also refactored the distortion correction node to make the code cleaner and support different undistortion strategies (ex: hybrid 2d/3d) in the future. 

Changes:
- once the node gets the TF information between imu, baselink, and lidar, it will stop getting the transformation again. 
- add 3d undistorted algorithm
- refactor the code.
- Add parameter file, parameter schema, and launch for this node. 

## Related links
https://github.com/autowarefoundation/autoware.universe/pull/7031
https://github.com/autowarefoundation/autoware.universe/issues/6657

## Tests performed
Run the node
```bash
ros2 launch pointcloud_preprocessor distortion_corrector.launch.xml
```

I tested this PR and ensured the pointcloud output is the same and the processing time has a negligible time difference (the result is different every time).

Before the PR
Input: mirror pointcloud  (58 frames)
distortion corrector: 2d

|   | Minimum | Maximum | Average |
|-----------|---------|---------|---------|
| Time     | 7.15    | 15.29   | 10.09   |

After the PR
Test 1:
|   | Minimum | Maximum | Average |
|-----------|---------|---------|---------|
| Time     | 7.516    | 12.719   | 9.47   |

Test2:
|   | Minimum | Maximum | Average |
|-----------|---------|---------|---------|
| Time     | 7.5    | 14.9   | 10.5   |

---------------------------------
Input: mirror pointcloud  (58 frames)
distortion corrector: 3d

After the PR
|   | Minimum | Maximum | Average |
|-----------|---------|---------|---------|
| Time     | 13.02    | 22.26   | 16.05   |

---------------------------------

Per point calculation time  
- test with 58 frames * 3 times
- 2d algorithm

Before PR

|   | Minimum       | Maximum      | Average      | Std Dev      | 50%         | 99%         |
|---|---------------|--------------|--------------|--------------|-------------|-------------|
| Time | 4.000000e-08 | 3.256100e-05 | 5.722510e-08 | 3.861742e-08 | 5.600000e-08 | 7.000000e-08 |

After PR
|   | Minimum       | Maximum      | Average      | Std Dev      | 50%         | 99%         |
[...]

Issue #6993, opened 2024-05-13
Improve Roll and Pitch estimation in Localization


mitsudome-r:
We can see that there is a wobbling of ego vehicle points as the vehicle drives through speed bumps. We would like to make sure that these are estimated correctly in Localization (perhaps in EKF).

example scene: https://www.youtube.com/watch?v=jd_HBTl_Gtw

meliketanrikulu:
### Progress:
To understand the source of the problem, I performed the following tests one by one:
1. To make sure that EKF is the source of the problem:
  I ran the localization completely with NDT. I gave NDT output as initial pose to NDT and published the map to baselink transform with NDT poses. In this way, I completely removed EKF from the system. I compared these results with the default autoware results.
  You can see the results in the video [here](https://www.youtube.com/watch?v=ckPNYZCzxO0)

  Test Results: As seen in this video, the point cloud shakes up and down in the default autoware (NDT + EKF) when passing through the speed bump. This error does not occur when run with NDT only.

2. Simulation Environment Test: Ensuring that EKF inputs, sensor calibrations or any other vehicle specific problems are not causing the error.
   I created the same scenario in a simulation environment and reproduced the same problem with awsim_labs_sensor_kit and sample_vehicle.
   You can see the results in the video [here](https://www.youtube.com/watch?v=dl00aDIoNnM)
   Test Results: In the simulation environment, I also saw that errors occurred when passing over bumps. From here I came to the conclusion that any vehicle specific problem (calibration, input twist data ) is not the source of the problem.
At the same time, I visualized the changes in pitch angle when it passed through the speed bump, both in the simulation environment and with real life data.I observed similar behavior in both. You can see these results below: 
<div align="center">
<img src="https://github.com/autowarefoundation/autoware.universe/assets/41450930/6345199d-7835-4ab3-9f76-f11cd791a651" alt="description" width="500" height="700"/>
</div>

### Next Steps: 
After making sure that EKF is only the source of the problem, I will try to solve the problem by debugging EKF in the simulation environment.

meliketanrikulu:
I tried to use the Roll and Pitch values ​​from the IMU directly in EKF. 
1. Simulation Test : 
The problem was completely solved with this method in the simulation environment:
Test video is [here](https://www.youtube.com/watch?v=bMwIv_a8RM8)

 NDT and EKF roll and pitch values gave very similar results in simulation environment:
[...]

PR #7875, opened 2024-07-05
fix(ekf_localizer): add imu option for roll pitch and height estimation

meliketanrikulu:
## Description
We can see that there is a wobbling of ego vehicle points as the vehicle drives through speed bumps.I repeated the tests in the simulation environment and obtained similar results. When I tried to optimize parameters in its current  EKF , I saw that I could not achieve good results in its current packet. You can find the tests I have previously done to solve this problem in the [issue](https://github.com/autowarefoundation/autoware.universe/issues/6993) comments.
Currently, 1D Filter is used for roll, pitch and position z(height) estimation in the ekf_localizer package. The angular velocity data that comes with the twist message and the orientation data in the imu data are not used in the calculations. It is calculated only by reference to the NDT poses that come as input.
However, IMU provides us with orientation information at a very high frequency. By using this data , we can localize roll, pitch and position z more accurately.
This PR suggests adding an option for those who want to use imu data instead of 1D Filters
## Related links

Related Issue: https://github.com/autowarefoundation/autoware.universe/issues/6993
- Link

## How was this PR tested?
1. Simulation Test:
 In the tests carried out in the simulation environment, you can see below that using imu directly solves the roll, pitch, and position z in EKF instead of calculating them with the 1D Filter:
sim-COMPARISON-Page-2 drawio
Buna ek olarak aşağıdaki videoları inceleyebilirsiniz:
Default Autoware test video is [here](https://www.youtube.com/watch?v=U3ByvJ8QQXE)
EKF with IMU Integration test video is [here](https://www.youtube.com/watch?v=bMwIv_a8RM8)
2. Real World Test:
Although the same level of performance cannot be achieved in real-world tests due to different factors, there is an improvement. It also seems that the results are further improved by using the [3D Distortion Corrector](https://github.com/autowarefoundation/autoware.universe/pull/7137), which has been newly merged into Autoware.You can find the results below:
COMPARISON_w-ndt_only drawio
Real World Test Video is [here](https://www.youtube.com/watch?v=VXDYWmPCpic&t=16s)
## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/localization/ekf_localizer/config/ekf_localizer.param.yaml
+++ b/localization/ekf_localizer/config/ekf_localizer.param.yaml
@@ -49,3 +49,4 @@
       # for velocity measurement limitation (Set 0.0 if you want to ignore)
       threshold_observable_velocity_mps: 0.0 # [m/s]
       pose_frame_id: "map"
+      use_imu: false
--- a/localization/ekf_localizer/include/ekf_localizer/ekf_localizer.hpp
+++ b/localization/ekf_localizer/include/ekf_localizer/ekf_localizer.hpp
@@ -33,16 +33,19 @@
 #include <geometry_msgs/msg/twist_stamped.hpp>
 #include <geometry_msgs/msg/twist_with_covariance_stamped.hpp>
 #include <nav_msgs/msg/odometry.hpp>
+#include <sensor_msgs/msg/imu.hpp>
 #include <std_srvs/srv/set_bool.hpp>
 #include <tier4_debug_msgs/msg/float64_multi_array_stamped.hpp>
 #include <tier4_debug_msgs/msg/float64_stamped.hpp>
 
 #include <tf2/LinearMath/Quaternion.h>
 #include <tf2/utils.h>
+#include <tf2_ros/buffer.h>
 #include <tf2_ros/transform_broadcaster.h>
 #include <tf2_ros/transform_listener.h>
 
 #include <chrono>
+#include <deque>
 #include <iostream>
 #include <memory>
 #include <queue>
@@ -129,6 +132,8 @@ private:
   //!< @brief measurement twist with covariance subscriber
   rclcpp::Subscription<geometry_msgs::msg::TwistWithCovarianceStamped>::SharedPtr
     sub_twist_with_cov_;
+  //!< @brief imu subscriber
+  rclcpp::Subscription<sensor_msgs::msg::Imu>::SharedPtr sub_imu_;
   //!< @brief time for ekf calculation callback
   rclcpp::TimerBase::SharedPtr timer_control_;
   //!< @brief last predict time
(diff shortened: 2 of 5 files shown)


meliketanrikulu:
I tested branch with hardware sync data and I get better results.Below, you can review the images captured at the moments when the difference between the instantaneous point cloud and the map was maximized while passing through the speed bump. (Normally instant pc and map points are matching but they move away when passing through a speed bump. You can see here the maximum distances (errors)):
 
COMPARISON-Page-1 drawio

Also you can see final results in [this](https://www.youtube.com/watch?v=q7imgbMKI9I) video
According to these results, we can see that by using hardware sync data, the results are very close to NDT, where we get the best results.

Simulation Tests:
In addition, I tested the simulation tests with VLS128 to see the distortions more clearly. You can see the results in [this](https://www.youtube.com/watch?v=q0M2kweqXBI) video.

meliketanrikulu:
Question : 

The [imu_corrector](https://github.com/autowarefoundation/autoware.universe/tree/main/sensing/imu_corrector) package does not fill orientation field of imu topic (/sensing/imu/imu_data) in autoware. It just fill angular_velocity and linear_acceleration parts of [sensor_msgs/msg/Imu](https://docs.ros.org/en/noetic/api/sensor_msgs/html/msg/Imu.html) data. Orientation field is comes with default values. You can look [here](https://github.com/autowarefoundation/autoware.universe/blob/12b31ab091d375ee9a068ec80977b4fc0c4c8353/sensing/imu_corrector/src/imu_corrector_core.cpp#L81) for check that.
I tested this PR with our vehicle's imu sensor output topic . But all of the autoware uses imu_corrector output topic(/sensing/imu/imu_data). I think I need to create PR for adding orientation field to imu_corrector package. 
What is your suggestion in here?

KYabuuchi:
@meliketanrikulu Yes, imu_corrector does not copy orientation from the original IMU data.
The reason for this should be that there is currently no node that uses IMU's orientation.
I think it is good to add the feature to copy orientation to imu_corrector.

SakodaShintaro:
I'm not sure if copying orientation in imu_corrector is a good idea. I think it's fine to have code to copy, but since it's not used in many cases, I feel like it's increasing unnecessary processing.

Another issue is that according to the documentation, inputting the IMU messages directly into the EKF is different from the Recommended Architecture.
https://autowarefoundation.github.io/autoware-documentation/main/design/autoware-architecture/localization/#recommended-architecture

I think one of the following is needed.
(1) Start a discussion to change the recommended architecture
(2) Implement a new EKF package which receives IMU messages
(3) Implement estimators that output Twist and Pose from the IMU

My personal opinion is that I have nothing against the EKF receiving IMU messages, but would it be possible to achieve this improvement using angular velocity instead of orientation?

meliketanrikulu:
> I'm not sure if copying orientation in imu_corrector is a good idea. I think it's fine to have code to copy, but since it's not used in many cases, I feel like it's increasing unnecessary processing.
> 
> Another issue is that according to the documentation, inputting the IMU messages directly into the EKF is different from the Recommended Architecture. https://autowarefoundation.github.io/autoware-documentation/main/design/autoware-architecture/localization/#recommended-architecture
> 
> I think one of the following is needed. (1) Start a discussion to change the recommended architecture (2) Implement a new EKF package which receives IMU messages (3) Implement estimators that output Twist and Pose from the IMU
> 
> My personal opinion is that I have nothing against the EKF receiving IMU messages, but would it be possible to achieve this improvement using angular velocity instead of orientation?

I created discussion --.> https://github.com/orgs/autowarefoundation/discussions/4973 . I would be very happy if you please share your comments. 

What is your suggestion for calculating orientation using the angular velocity of the vehicle? 
I tested it with [dead reckoning](https://en.wikipedia.org/wiki/Dead_reckoning).  However, this method problematically causes its errors to increase and breaks down after a while.

SakodaShintaro:
@meliketanrikulu 
I haven't read the details of the changes in this pull request yet, so I apologize in advance if I say anything wrong.

There are several papers that integrate IMU with EKF, and since they use acceleration and angular velocity rather than orientation, I think it should be possible to do the same here.  

The following paper is one that I recently learned about.

https://arxiv.org/abs/2407.02786

PR #7886, opened 2024-07-08
fix(pointcloud_preprocessor): fix the bug where the geometry message was not saved correctly.

vividf:
## Description
The distortion corrector node didn't save the geometry message correctly which caused the imu to base link transform become identity. 

## Related links

**Parent Issue:**

- Link

## How was this PR tested?
Tested with the sample rosbag and the unit test I am working on. 

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


knzo25 (changes requested):
Not quite related to these changes, but there are methods that should be protected

knzo25:
Also, please write grammatically correct titles

vividf:
@knzo25 
Thanks for the checks, fixed those in https://github.com/autowarefoundation/autoware.universe/pull/7886/commits/e7980ee28e52e6bb344ce03a5c9105a02adbbb40

PR #7890, opened 2024-07-08
feat(lidar_transfusion): update TransFusion-L model

scepter914:
## Description

I update TransFusion-L model.
According to it, I add the parameter of `num_proposals` for ROS parameter interface because new model changes it from before model.

## How was this PR tested?

Tested by a XX1 rosbag that point type of LiDAR doesn't change.

- Visualization

Bounding boxes of solid cube: TransFusion
Bounding boxes with white line: CenterPoint
While CenterPoint tend to larger bounding box than actual size for large vehicle and sometimes lost the large vehicles, TransFusion can detect for large vehicles stably.

https://github.com/autowarefoundation/autoware.universe/assets/16330533/ccbf700e-518b-4295-a7b7-b36e40159998

- 3D visualization

From 3D view, the results of TransFusion can be see as below

https://github.com/autowarefoundation/autoware.universe/assets/16330533/4a048ca3-fff7-49c7-98d7-5b80aa938faf

https://github.com/autowarefoundation/autoware.universe/assets/16330533/bb7bcefb-0e85-4055-ab0d-ffb920963cbc

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


amadeuszsz (approved):
Great work! I validated new model in runtime as well.
Regarding `num_proposals` parameter, I wonder if we should get it directly through TensorRT engine instead of exposing it as a parameter:
 
transfusion_output_shape

I will address it in future PR during `lidar_transfusion` refactor.
Before merge, could you update artifacts? We need to upload new config as well. Here is an example: https://github.com/autowarefoundation/autoware/pull/4835

scepter914:
> I will address it in future PR during lidar_transfusion refactor.

That's nice!
Thank you for maintenance work.

> Before merge, could you update artifacts? We need to upload new config as well. Here is an example: autowarefoundation/autoware#4835

Thank you for comment.
I'll do it.

scepter914:
[The PR that update artifacts](https://github.com/autowarefoundation/autoware/pull/4959) is approved, so I merge this PR.

PR #7896, opened 2024-07-08
feat(autoware_behavior_velocity_planner_common,autoware_behavior_velocity_stop_line_module): revert pr7710 and Check next_lane_id in createTargetPoint on stop line

yhisaki:
## Description

1. Revert [this PR](https://github.com/autowarefoundation/autoware.universe/pull/7710)
2. In the previous implementation, it was checked whether the lane_id of the intersection between the stop line and the path matched the lane_id held by the module. However, for points on the path near the end of a lane, the lane_id might become the next/previous lane_id of the original id due to the resampling of PathWithLaneId. Therefore, if the stop_line is near the end of a lane, the check for matching lane_ids at the intersection might fail. To address this, even if the lane_id at the intersection does not match the intended id of the stop_line_module, the next lane_id and the previous lane_id of the lane_id are allowed.
	The figure below illustrates the issue that occurs in the previous implementation. The red x represents the PathPointWithLaneId before resampling, the green x represents the PathPointWithLaneId after resampling, the blue lines represent the actual lane boundaries, and the purple line represents the stop_line. Ideally, the stop_line should intersect at a point with lane_id=30134, but due to the resampling effect, it is mistakenly determined to intersect at a point with lane_id=30322.

Untitled (1) (1)

## Tests performed

tested on psim

### DIFF
--- a/planning/behavior_velocity_planner/autoware_behavior_velocity_stop_line_module/src/scene.cpp
+++ b/planning/behavior_velocity_planner/autoware_behavior_velocity_stop_line_module/src/scene.cpp
@@ -52,7 +52,7 @@ bool StopLineModule::modifyPathVelocity(PathWithLaneId * path, StopReason * stop
 
   // Calculate stop pose and insert index
   const auto stop_point = arc_lane_utils::createTargetPoint(
-    *path, stop_line, planner_param_.stop_margin,
+    *path, stop_line, lane_id_, planner_param_.stop_margin,
     planner_data_->vehicle_info_.max_longitudinal_offset_m);
 
   // If no collision found, do nothing
--- a/planning/behavior_velocity_planner/autoware_behavior_velocity_stop_line_module/src/scene.cpp
+++ b/planning/behavior_velocity_planner/autoware_behavior_velocity_stop_line_module/src/scene.cpp
@@ -18,12 +18,10 @@
 #include <autoware/behavior_velocity_planner_common/utilization/util.hpp>
 #include <autoware/motion_utils/trajectory/trajectory.hpp>
 
-#include <algorithm>
-#include <vector>
+#include <cstdint>
 
 namespace autoware::behavior_velocity_planner
 {
-namespace bg = boost::geometry;
 
 StopLineModule::StopLineModule(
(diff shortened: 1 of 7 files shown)


takayuki5168 (approved):

takayuki5168 (changes requested):
getNextLaneletWithinRoute should be used instead of getNextLanelets

planning/behavior_velocity_planner/autoware_behavior_velocity_stop_line_module/src/scene.cpp
```diff
@@ -51,8 +49,15 @@ bool StopLineModule::modifyPathVelocity(PathWithLaneId * path, StopReason * stop
     stop_line_[0], stop_line_[1], planner_data_->stop_line_extend_length);
 
   // Calculate stop pose and insert index
+  auto lane = this->planner_data_->route_handler_->getLaneletsFromId(lane_id_);
+  auto next_lanes = this->planner_data_->route_handler_->getNextLanelets(lane);
```
takayuki5168, line 53: Please add a previous lanes as well.

planning/behavior_velocity_planner/autoware_behavior_velocity_stop_line_module/src/scene.cpp
```diff
@@ -51,8 +49,15 @@ bool StopLineModule::modifyPathVelocity(PathWithLaneId * path, StopReason * stop
     stop_line_[0], stop_line_[1], planner_data_->stop_line_extend_length);
 
   // Calculate stop pose and insert index
+  auto lane = this->planner_data_->route_handler_->getLaneletsFromId(lane_id_);
```
takayuki5168, line 52: Please add a comment since this change is not straightforward.

yhisaki:
> getNextLaneletWithinRoute should be used instead of getNextLanelets

It seems that there is a bug in the behavior_velocity_planner where the route is not subscribed, and the getNextLaneletsWithinRoute function cannot be used. In the future, I would like to fix this bug, but temporarily, I addressed it by using getNextLanelets.

takayuki5168 (approved):

PR #7901, opened 2024-07-08
feat(start_planner): add end_pose_curvature_threshold 

kosuke55:
## Description

The curvature threshold which is used for calculating the longitudinal distance of shift. The shift end pose is shifted forward so that its curvature is less than this value. This is to prevent the generated path from having a large curvature when the end pose is on a curve. If a point with a curvature below the threshold is not found, the longitudinal distance is used as the distance to the point with the lowest curvature among the points beyond a certain distance.

## Related links

**Parent Issue:**

- Link

**Private Links:**

- https://tier4.atlassian.net/browse/RT1-7071

## How was this PR tested?

psim

- before

the shift end pose(yellow arrow is moved to straight lane)

image

image

- after

image

image

2024/07/08 https://evaluation.tier4.jp/evaluation/reports/2081abee-9216-5a09-9f1a-40ff53d9bbcf/?project_id=prd_jt

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


kosuke55:
depend https://github.com/autowarefoundation/autoware.universe/pull/7902

planning/behavior_path_planner/autoware_behavior_path_start_planner_module/README.md
```diff
@@ -469,17 +469,18 @@ Pull out distance is calculated by the speed, lateral deviation, and the lateral
...
+| maximum_curvature                              | [1/m]  | double | maximum curvature. The pull out distance is calculated so that the curvature is smaller than this value.                                                                                                                                                                                                                                                                                                                                                                               | 0.07          |
+| end_pose_curvature_threshold                   | [1/m]  | double | The curvature threshold which is used for calculating the longitudinal distance of shift. The shift end pose is shifted forward so that its curvature is less than this value. This is to prevent the generated path from having a large curvature when the end pose is on a curve. If a point with a curvature below the threshold is not found, the longitudinal distance is used as the distance to the point with the lowest curvature among the points beyond a certain distance. | 0.01          |
```
kyoichi-sugahara, line 483:
```suggestion
| end_pose_curvature_threshold                   | [1/m]  | double | The curvature threshold which is used for calculating the longitudinal distance of shift. The shift end pose is shifted forward so that curvature on shift end pose is less than this value. This is to prevent the generated path from having a large curvature when the end pose is on a curve. If a shift end pose with a curvature below the threshold is not found, the longitudinal distance is used as the distance to the point with the lowest curvature among the points beyond a certain distance. | 0.01          |
```
Btw, `longitudinal distance` is different from `shift pull out distance`?

kyoichi-sugahara (approved):
LGTM!

common/autoware_motion_utils/include/autoware/motion_utils/trajectory/trajectory.hpp
```diff
@@ -991,34 +991,50 @@ calcCurvature<std::vector<autoware_planning_msgs::msg::TrajectoryPoint>>(
...
+
+    // The first point has only the next point, so put the distance to that point.
+    // In other words, assign the first segment length at the second point to the
+    // second_segment_length at the first point.
+    if (i == 1) {
+      curvature_and_segment_length_vec.at(0).second.second =
+        autoware::universe_utils::calcDistance2d(p1, p2);
+    }
+
+    // The second_segment_length of the previous point and the first segmentlength of the current
```
satoshi-ota, line 1014:
```suggestion
    // The second_segment_length of the previous point and the first segment length of the current
```

planning/behavior_path_planner/autoware_behavior_path_start_planner_module/README.md
kosuke55: it's same!

kosuke55: I will update the docs

kosuke55: updated

Issue #5594, opened 2023-11-15
Add a new pointcloud filter based on image semantic segmentation


badai-nguyen:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I've agreed with the maintainers that I can plan this task.

### Description

As I described https://github.com/orgs/autowarefoundation/discussions/3974, we (TIER IV) and other teams are also facing difficult to deal with unknown from plants, raindrops/smoke/splash pointcloud or pointcloud of ground segmentation fail. 
 
Current pointcloud filters including dual_return_filter, outlier_filter, ground_segmentation only based on points kinematics information. Compare_map_filter compare pointcloud with map to remove un-movable objects but it frequently fail with growing plants, tree.. etc. 

It should be helpful if image semantic information could be used and fused with pointcloud to remove unnecessary pointclouds for autonomous driving such as pointcloud of plants in background, or  drivable area.

### Purpose

To deploy some kind of image semantic or instance segmentation model and matching with pointcloud to filter pointcloud.

### Possible approaches

1. Train and deploy an image segmentation model:

- By Using a separate model with current detection Yolox model
- Or adding an additional task/header for into current detection yolox model 

2. Publish some 2d mask image of segmentation information or kind of less important image region information for filtering 

3. Create new node to fuse image pixel level with pointcloud and remove pointcloud in less important image region.

### Definition of done

- [x] Train and Deploy an image segmentation model into autoware
- [x] Publish some topics of image segmentation result 
- [x] Matching pointcloud and image segmentation result for filtering

xmfcx:
- Retraining the 2D image segmentation model right now.
- Semantic seg fusion is already merged.
- Will update the pipeline to have this included.
- Test on the real vehicle.
- Eval on old data (e.g. noisy)

xmfcx:
@StepTurtle please check this work too, link it when creating a new issue.

badai-nguyen:
Note: 2D multiple header YOLOX was trained and to be going to add at https://github.com/autowarefoundation/autoware/pull/4012

xmfcx:
@badai-nguyen is this issue completed?

badai-nguyen:
@xmfcx thank you for your remind. 
I close this issue since all neccessary change were merged.

badai-nguyen:
[...]

PR #7905, opened 2024-07-09
fix(tensorrt_yolox): add run length encoding for sematic segmentation mask

badai-nguyen:
## Description
- Since transfering the raw image of sematic segmentation mask between Perception ECU where tensorrt running and Main ECU causes the bandwidth issue and increases tensorrt_yolox node processing_time. 
- Using JPEG compressed method from `image_transport` can reduce bandwidth but pixel value might be changed during decompression that lead the changed of semantic segmentation especially along the boundary of segment.
- Using lossless compression (PNG) of  `image_transport` improve the bandwidth and processing_time but not good as `run_length` method. The reason is that semantic mask of road scene normally has large segment with the same class which is very effective for `run_length encoder` 
- Comparison of bandwidth and processing time between `image_transport` lossless compression (PNG) with the runlength when runing multi-header model on ROSCUBE and transfer result to Main ECU: 

| raw iamge | 'image_transport`'s PNG `png_level: 1` | 'image_transport`'s PNG  `png_level: 9`  | implemented 'run_length' |
| --- | --- |--- | --- |
| Average : 91.5351569282137 ,std: 0.5338352012729306  original_processing_time_data csv | Average: 54.99859166666666 ,std: 1.066178226938265 png_level_1_processing_time csv |Average: 135.7696519721578 ,std: 10.981303755107556  png_level_9_processing_time csv |Average: 49.28519 ,std: 0.7785528951629853 runlength_processing_time_data csv    |

- Releated PR: https://github.com/autowarefoundation/autoware.universe/pull/7909
- Related Issue: https://github.com/autowarefoundation/autoware.universe/issues/5594
## Related links

**Parent Issue:**

- [TIER IV INTERNAL LINK](https://star4.slack.com/archives/C03S84LDJGG/p1720480344949899?thread_ts=1716365558.423419&cid=C03S84LDJGG)
- [TIER IV INTERNAL TICKET LINK](https://tier4.atlassian.net/browse/RT1-7106)

## How was this PR tested?
- Running `ros2 launch autoware_tensorrt_yolox yolox_s_plus_opt.launch.xml` on Embedded Ecu and confirmed that processing time is unchanged when `sematic segmentation mask topic` topic is subscribed from outside of Embedded ECU.

  - Tested reults on Embedded Robotic Controller Powered by NVIDIA® Jetson AGX Xavier ™ Module RQX-58G with preprocess_on_gpu: true:
  - OS version:
```
cat /etc/nv_tegra_release
-> R32 (release), REVISION: 6.1, GCID: 27863751, BOARD: t186ref, EABI: aarch64, DATE: Mon Jul 26 19:36:31 UTC 2021
```

- Confirmed that `publishing_time` is significantly decreased, so that the overal `processing_time` also descreased.

| Before  | After | 
| --- | --- | 
|image|image |
| 1 Raw Semseg Mask Image ~ 24000KB/s image | Bandwidth: 1 Raw Semseg Mask Image ~ 320.72KB/s image|
| https://github.com/user-attachments/assets/26b98e9c-83ff-46d3-95a3-e63246e4b53e  | [Screencast from 2024年07月26日 19時37分19秒.webm](https://github.com/user-attachments/assets/408c1c82-fc41-4a7e-86e5-0a85036326d8) | 

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


perception/autoware_tensorrt_yolox/src/utils.cpp
```diff
@@ -0,0 +1,64 @@
+// Copyright 2024 Tier IV, Inc.
```
yukkysaito, line 1:
```suggestion
// Copyright 2024 TIER IV, Inc.
```

yukkysaito:
@badai-nguyen Thanks for the PR.
Did you come up with the encoder-decoder being implemented this time by yourself?
I think PNG and JPEG (lossless mode) can be used since they are lossless compression formats, but I would like to know the reason if you didn't consider them this time.

badai-nguyen:
> @badai-nguyen Thanks for the PR. Did you come up with the encoder-decoder being implemented this time by yourself? I think PNG and JPEG (lossless mode) can be used since they are lossless compression formats, but I would like to know the reason if you didn't consider them this time.

@yukkysaito 
- The JPEG is not lossless compression so that can cause the change of segmentation mask, especiall along boundary of each segment.
- I compared run_length with PNG from 'image_transport' and found that fatest compress level (level 1) of PNG still increase processing_time, and the 'run_length' shows the best performance. 
Please check the updated PR's description.

perception/autoware_tensorrt_yolox/src/utils.cpp
```diff
@@ -0,0 +1,64 @@
...
+{
+  cv::Mat mask(rows, cols, CV_8UC1, cv::Scalar(0));
+  int idx = 0;
+  int step = sizeof(uint8_t) + sizeof(int);
+  int nb_pixels = 0;
+  for (size_t i = 0; i < rle_data.size(); i += step) {
+    uint8_t value;
+    int length;
+    std::memcpy(&value, &rle_data[i], sizeof(uint8_t));
+    std::memcpy(&length, &rle_data[i + sizeof(uint8_t)], sizeof(int));
```
manato, line 49: I believe this code works with no problem, but let me discuss a semantic (intension of the writer) topic.
In my understanding, the value used in the `[]` operator is indices, which include the size of each element implicitly (i.e., if a case of `std::vector<int> int_vec`, the distance of `&int_vec[i]`  and `&int_vec[i+1]` would be 4), while the `sizeof` operator returns the actual byte size of the specified object. 
To distinguish "index" and "address" (written in bytes), how about the following modification?:
```suggestion
    std::memcpy(&length, &rle_data[i + 1], sizeof(int));    // under the condition that we know rle_data[i] only consume 1 element of the vector
```
```suggestion
    std::memcpy(&length, reinterpret_cast<uint8_t*>(reinterpret_cas<void*>(&rle_data[i]) + sizeof(uint8_t)), sizeof(int));     // straightforward address calculation
```

This point is kind of my preference, so if you feel something is wrong, please let me know!

manato (commented):
@badai-nguyen 
Thank you for your great contribution and sorry for the delay in my review. My major concern was increasing processing time by introducing run-length encoding. But thanks to the evaluation result you shared, it has gone. 
I left two comments for small points. I'd appreciate it if you could consider them before approving this PR.

perception/autoware_tensorrt_yolox/src/utils.cpp
```diff
@@ -0,0 +1,64 @@
...
+  }
+  return compressed_data;
+}
+
+cv::Mat runLengthDecoder(const std::vector<uint8_t> & rle_data, const int rows, const int cols)
+{
+  cv::Mat mask(rows, cols, CV_8UC1, cv::Scalar(0));
+  int idx = 0;
+  int step = sizeof(uint8_t) + sizeof(int);
+  int nb_pixels = 0;
```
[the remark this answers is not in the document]
badai-nguyen: @manato Thank you for your review, you're correct, `nb_pixels` is redundant. I fixed at https://github.com/autowarefoundation/autoware.universe/pull/7905/commits/12307ea62af1a182458bf2dfba812c76796e5d02

perception/autoware_tensorrt_yolox/src/utils.cpp
badai-nguyen: @manato Thank you for your suggestion.

> the value used in the [] operator is indices, 

Here the value in [] is memory index of `int length` in `data`, which is started after `uint8_t value (the semantic class)`, so that [] has 1byte (sizeof(uint8_t)) padding from beggining of each encoded pattern. The result should be the same as your suggestion.

manato: @badai-nguyen 
Thank you for your consideration. 
Yes, I believe the result should be the same regardless of my suggestion being introduced. My concern was the mixture of "index" and "byte-size-oriented operation". Since your latest commit uses `+1` manipulate index, my concern has been gone, anyway.

perception/autoware_tensorrt_yolox/src/tensorrt_yolox_node.cpp
```diff
@@ -177,12 +178,21 @@ void TrtYoloXNode::onImage(const sensor_msgs::msg::Image::ConstSharedPtr msg)
...
         .toImageMsg();
     out_mask_msg->header = msg->header;
+
+    std::vector<std::pair<uint8_t, int>> compressed_data = runLengthEncoder(mask);
+    int step = sizeof(uint8_t) + sizeof(int);
+    out_mask_msg->data.resize(static_cast<int>(compressed_data.size()) * step);
+    for (size_t i = 0; i < compressed_data.size(); ++i) {
+      std::memcpy(&out_mask_msg->data[i * step], &compressed_data.at(i).first, sizeof(uint8_t));
+      std::memcpy(
+        &out_mask_msg->data[i * step + sizeof(uint8_t)], &compressed_data.at(i).second,
```
manato, line 193: same as https://github.com/autowarefoundation/autoware.universe/pull/7905#discussion_r1705378296

badai-nguyen: @manato I fixed at https://github.com/autowarefoundation/autoware.universe/pull/7905/commits/3220095709b0fd653c8bda7c39d5eaa9fe6efa64

manato (approved):
@badai-nguyen 
LGTM. Thank you for handling all of my concerns. Since all my concerns are gone, I approve this PR.

badai-nguyen:
Since seperated the unit test part into https://github.com/autowarefoundation/autoware.universe/pull/8414

PR #7907, opened 2024-07-09
chore(autoware_universe_utils): update document

yhisaki:
## Description

- Changed the example from a script format to a format where nodes are held as objects.
- Recommended the use of ScopedTimeTrack.
- Modified the example to pass the reporter in the constructor and provided an example of using `add_report()` in the comments.
- Added comments indicating the start and end timing of the tree construction, as well as the timing of the reporting.
- Remove String reporter.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


common/autoware_universe_utils/README.md
```diff
@@ -49,56 +49,83 @@ explicit TimeKeeper(Reporters... reporters);
...
 #include <rclcpp/rclcpp.hpp>
 
+#include <std_msgs/msg/string.hpp>
+
+#include <chrono>
 #include <iostream>
 #include <memory>
+#include <thread>
+
+class TimeKeeperExample : public rclcpp::Node
```
takayuki5168, line 78: ExampleNode sounds better since the word `TimeKeeper` is everywhere currently and searchability is not so high.

yhisaki: fixed!

takayuki5168 (approved):
LGTM!

Issue #7916, opened 2024-07-09
Running source/install/setup.bash complains that it cannot find "local_setup.bash" for te


mitsudome-r:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I'm convinced that this is not my fault but a bug.

### Description

Some packages that depends on TensorRT outputs error message as the following when I run `source/install/setup.bash` after building Autoware on machine without CUDA. 

Users can ignore it and run Autoware without an issue, but it would be better if we can make it so that it doesn't output any error for less confusion.

```
not found: "/home/mitsudome-r/autoware/install/tensorrt_common/share/tensorrt_common/local_setup.bash"
not found: "/home/mitsudome-r/autoware/install/tensorrt_classifier/share/tensorrt_classifier/local_setup.bash"
```

### Expected behavior

`source/install/setup.bash` finished without an error.

### Actual behavior

`source/install/setup.bash` outputs an error as described in the Description.

### Steps to reproduce

1. Install Autoware without installing CUDA
2. run `install/setup.bash`

### Versions

- OS: Ubuntu 22.04
- ROS 2: Humble
- Autoware: Main branch on July 9, 2024

### Possible causes

The packages don't call ament_package() nor ament_auto_package() in their CMakeLists.txt files.

Desperadozcx:
I also encountered this problem. Could you please tell me how you solved it?

AnzhiZhang:
Same issue, can ignore it and run Autoware without an issue.

```bash
root@SORA-WS3:/workspace# source install/setup.bash
not found: "/workspace/install/autoware_tensorrt_plugins/share/autoware_tensorrt_plugins/local_setup.bash"
```

However, a different thing is that I have installed it with CUDA.

PR #7917, opened 2024-07-09
fix: make sure ament_package() is called when CUDA is not found

mitsudome-r:
## Description

Resolves https://github.com/autowarefoundation/autoware.universe/issues/7916

## Related links
- https://github.com/autowarefoundation/autoware.universe/issues/7916

## How was this PR tested?
Install Autoware on machines without CUDA installed and run `source install/setup.bash`

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/common/tensorrt_common/CMakeLists.txt
+++ b/common/tensorrt_common/CMakeLists.txt
@@ -13,6 +13,7 @@ find_package(TENSORRT)
 
 if(NOT (CUDAToolkit_FOUND AND CUDNN_FOUND AND TENSORRT_FOUND))
   message(WARNING "cuda, cudnn, tensorrt libraries are not found")
+  ament_package()
   return()
 endif()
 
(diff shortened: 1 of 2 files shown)


common/tensorrt_common/CMakeLists.txt
```diff
@@ -13,6 +13,7 @@ find_package(TENSORRT)
 
 if(NOT (CUDAToolkit_FOUND AND CUDNN_FOUND AND TENSORRT_FOUND))
   message(WARNING "cuda, cudnn, tensorrt libraries are not found")
+  ament_package()
```
mitsudome-r, line 16: I am using ament_package() instead of ament_auto_packages() to match with L82 in the original file.

mitsudome-r:
It looks like some packages (like lidar_apollo_instance_segmentation) uses `if(NOT ${tensorrt_common_FOUND})` and causes failure with this change. I will think about some solutions to fix this issue in more clean way.

PR #7919, opened 2024-07-09
chore: add missing dependency

miursh:
## How was this PR tested?
Confirm that packages depending on perception can be built on my local PC.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


yukkysaito (approved):
LGTM

PR #7932, opened 2024-07-10
fix(autoware_pure_pursuit): fix shadowVariable

kobayu858:
## Description

```
control/autoware_pure_pursuit/src/autoware_pure_pursuit_core/interpolate.cpp:232:12: style: Local variable 'i' shadows outer variable [shadowVariable]
  for (int i = 0; i < static_cast<int>(normalized_idx.size()); ++i) {
           ^
```
## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/control/autoware_pure_pursuit/src/autoware_pure_pursuit_core/interpolate.cpp
+++ b/control/autoware_pure_pursuit/src/autoware_pure_pursuit_core/interpolate.cpp
@@ -229,8 +229,8 @@ bool SplineInterpolate::interpolate(
   generateSpline(base_value);
 
   // interpolate by spline  with normalized index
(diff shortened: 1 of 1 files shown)


veqcc:
@kobayu858 
Could you consider using range-based for loop?
```
  for (const auto & index : normalized_idx) {
    return_value.push_back(getValue(index));
  }
```

kobayu858:
Thanks for the advice. I did change it to a range-based for loop, because I thought that would be more simple.

veqcc (approved):

kobayu858:
@TakaHoribe @takayuki5168 

takayuki5168 (approved):

PR #7933, opened 2024-07-10
fix(perception_online_evaluator): fix shadowVariable

kobayu858:
## Description

```
evaluator/perception_online_evaluator/src/perception_online_evaluator_node.cpp:278:23: style: Local variable 'ns' shadows outer variable [shadowVariable]
    const std::string ns = "debug_marker.";
                      ^
```
## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/evaluator/perception_online_evaluator/src/perception_online_evaluator_node.cpp
+++ b/evaluator/perception_online_evaluator/src/perception_online_evaluator_node.cpp
@@ -247,55 +247,54 @@ rcl_interfaces::msg::SetParametersResult PerceptionOnlineEvaluatorNode::onParame
   updateParam<double>(parameters, "objects_count_window_seconds", p->objects_count_window_seconds);
 
   // update parameters for each object class
-  const auto update_object_param = [&p, &parameters](
-                                     const auto & semantic, const std::string & ns) {
-    auto & config = p->object_parameters.at(semantic);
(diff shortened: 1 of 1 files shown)


evaluator/perception_online_evaluator/src/perception_online_evaluator_node.cpp
```diff
@@ -275,25 +275,25 @@ rcl_interfaces::msg::SetParametersResult PerceptionOnlineEvaluatorNode::onParame
 
   // update debug marker parameters
   {
-    const std::string ns = "debug_marker.";
+    const std::string name_space = "debug_marker.";
```
veqcc, line 278: @kobayu858 
Could you consider making line 250-274 to a code block by surrounding by `{}` ?

Like
```
// update parameters for each object class
{
  ...
}

// update debug marker parameters
{
  ...
}
```

kobayu858: Thanks for the advice.
The warning has been fixed since it was confirmed that the warning was resolved by this method.

veqcc (approved):

kobayu858:
@rej55 @kosuke55 @ktro2828 @kyoichi-sugahara @miursh @YoshiRi 

miursh (approved):

PR #7974, opened 2024-07-11
fix(autoware_behavior_velocity_crosswalk_module): fix shadowVariable

kobayu858:
## Description

```
planning/behavior_velocity_planner/autoware_behavior_velocity_crosswalk_module/src/scene_crosswalk.cpp:385:18: style: Local variable 'base_link2front' shadows outer variable [shadowVariable]
    const double base_link2front = planner_data_->vehicle_info_.max_longitudinal_offset_m;
                 ^
```

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_velocity_planner/autoware_behavior_velocity_crosswalk_module/src/scene_crosswalk.cpp
+++ b/planning/behavior_velocity_planner/autoware_behavior_velocity_crosswalk_module/src/scene_crosswalk.cpp
@@ -382,7 +382,7 @@ std::optional<StopFactor> CrosswalkModule::checkStopForCrosswalkUsers(
   // Check if ego moves forward enough to ignore yield.
   const auto & p = planner_param_;
   if (!path_intersects.empty()) {
-    const double base_link2front = planner_data_->vehicle_info_.max_longitudinal_offset_m;
+    const double inner_base_link2front = planner_data_->vehicle_info_.max_longitudinal_offset_m;
     const double dist_ego2crosswalk =
       calcSignedArcLength(ego_path.points, ego_pos, path_intersects.front());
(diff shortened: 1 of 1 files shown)


veqcc (approved):

planning/behavior_velocity_planner/autoware_behavior_velocity_crosswalk_module/src/scene_crosswalk.cpp
```diff
@@ -382,7 +382,7 @@ std::optional<StopFactor> CrosswalkModule::checkStopForCrosswalkUsers(
   // Check if ego moves forward enough to ignore yield.
   const auto & p = planner_param_;
   if (!path_intersects.empty()) {
-    const double base_link2front = planner_data_->vehicle_info_.max_longitudinal_offset_m;
```
yuki-takagi-66, line 385: Delete this line is sufficient because the outer value is same as this value

yuki-takagi-66 (changes requested):
Thank you for the fix PR. We were foolishly declaring variables with the same name and same value in an inclusive scope, so could you please change it to remove the inner variables.

planning/behavior_velocity_planner/autoware_behavior_velocity_crosswalk_module/src/scene_crosswalk.cpp
kobayu858: I have removed this line as you indicated.

yuki-takagi-66 (approved):
Thank you so much! LGTM

PR #7979, opened 2024-07-11
fix(system_error_monitor): fix shadowVariable

kobayu858:
## Description
```
system/system_error_monitor/src/system_error_monitor_core.cpp:291:18: style: Local variable 'param_key' shadows outer variable [shadowVariable]
    const auto & param_key = split_names.at(1);
```
## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/system/system_error_monitor/src/system_error_monitor_core.cpp
+++ b/system/system_error_monitor/src/system_error_monitor_core.cpp
@@ -288,10 +288,10 @@ void AutowareErrorMonitor::loadRequiredModules(const std::string & key)
     //                     or required_modules.key.module.parameter
     const auto split_names = split(param_name, '.');
     const auto & param_required_modules = split_names.at(0);
-    const auto & param_key = split_names.at(1);
+    const auto & split_param_key = split_names.at(1);
     const auto & param_module = split_names.at(2);
     const auto module_name_with_prefix =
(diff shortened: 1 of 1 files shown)


kobayu858:
I thought about using "param_key" all the time, but I changed the variable name because I thought it would be better not to use it all the time semantically.

veqcc:
> I thought about using "param_key" all the time, but I changed the variable name because I thought it would be better not to use it all the time semantically.

Then, how about changing the first `param_key` from
```
  const auto param_key = std::string("required_modules.") + key;
  const uint64_t depth = 3;
  const auto param_names = this->list_parameters({param_key}, depth).names;
  if (param_names.empty()) {
    throw std::runtime_error(fmt::format("no parameter found: {}", param_key));
  }
```
to
```
  const uint64_t depth = 3;
  const auto param_names = this->list_parameters({std::string(required_modules.) + "key"}, depth).names;
  if (param_names.empty()) {
    throw std::runtime_error(fmt::format("no parameter found: required_modules.{}", key));
  }
```

kobayu858:
I have corrected the above advice as advised above.

veqcc (approved):

ito-san (approved):
LGTM

PR #8013, opened 2024-07-12
fix(autoware_object_velocity_splitter): fix funcArgNamesDifferent

kobayu858:
## Description
This is a fix based on cppcheck funcArgNamesDifferent warnings

```
perception/object_velocity_splitter/src/object_velocity_splitter_node.cpp:68:82: style: inconclusive: Function 'onObjects' argument 1 names different: declaration 'msg' definition 'objects_data_'. [funcArgNamesDifferent]
void ObjectVelocitySplitterNode::onObjects(const DetectedObjects::ConstSharedPtr objects_data_)
                                                                                 ^
```
## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

PR #8014, opened 2024-07-12
fix(autoware_radar_object_clustering): fix funcArgNamesDifferent

kobayu858:
## Description
This is a fix based on cppcheck funcArgNamesDifferent warnings

```
perception/radar_object_clustering/src/radar_object_clustering_node.cpp:94:81: style: inconclusive: Function 'onObjects' argument 1 names different: declaration 'msg' definition 'objects_data_'. [funcArgNamesDifferent]
void RadarObjectClusteringNode::onObjects(const DetectedObjects::ConstSharedPtr objects_data_)
                                                                                ^
```

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_radar_object_clustering/src/radar_object_clustering_node.hpp
+++ b/perception/autoware_radar_object_clustering/src/radar_object_clustering_node.hpp
@@ -52,10 +52,7 @@ private:
   rclcpp::Subscription<DetectedObjects>::SharedPtr sub_objects_{};
 
   // Callback
-  void onObjects(const DetectedObjects::ConstSharedPtr objects_data_);
-
-  // Data Buffer
-  DetectedObjects::ConstSharedPtr objects_data_{};
+  void onObjects(const DetectedObjects::ConstSharedPtr msg);
 
   // Publisher
   rclcpp::Publisher<DetectedObjects>::SharedPtr pub_objects_{};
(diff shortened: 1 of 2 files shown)


veqcc:
Same comment https://github.com/autowarefoundation/autoware.universe/pull/8013#issuecomment-2225300314

kobayu858:
@scepter914 @miursh @technolojin @YoshiRi 

yukkysaito:
It appears that objects_data_ is not used in this change. objects_data_ is likely a typo. It would be better to align the code with msg instead of objects_data_.

perception/autoware_radar_object_clustering/src/radar_object_clustering_node.hpp
```diff
@@ -52,7 +52,7 @@ class RadarObjectClusteringNode : public rclcpp::Node
   rclcpp::Subscription<DetectedObjects>::SharedPtr sub_objects_{};
 
   // Callback
-  void onObjects(const DetectedObjects::ConstSharedPtr msg);
+  void onObjects(const DetectedObjects::ConstSharedPtr objects_data_);
```
YoshiRi, line 55: I think it's better to keep the original declaration and rename the .cpp function instead.

YoshiRi (changes requested):

perception/autoware_radar_object_clustering/src/radar_object_clustering_node.hpp
kobayu858: Thanks for the advice.
I removed objects_data_ and unified it with msg.

This response is the same as [PR#8013](https://github.com/autowarefoundation/autoware.universe/pull/8013).

YoshiRi (approved):
LGTM

PR #8018, opened 2024-07-12
fix(autoware_obstacle_stop_planner): fix funcArgNamesDifferent

kobayu858:
## Description
This is a fix based on cppcheck funcArgNamesDifferent warnings

```
planning/autoware_obstacle_stop_planner/src/adaptive_cruise_control.cpp:203:49: style: inconclusive: Function 'insertAdaptiveCruiseVelocity' argument 7 names different: declaration 'current_velocity_ptr' definition 'current_odometry_ptr'. [funcArgNamesDifferent]
  const nav_msgs::msg::Odometry::ConstSharedPtr current_odometry_ptr, bool * need_to_stop,
                                                ^
```
## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/autoware_obstacle_stop_planner/src/adaptive_cruise_control.hpp
+++ b/planning/autoware_obstacle_stop_planner/src/adaptive_cruise_control.hpp
@@ -53,7 +53,7 @@ public:
     const TrajectoryPoints & trajectory, const int nearest_collision_point_idx,
     const geometry_msgs::msg::Pose self_pose, const pcl::PointXYZ & nearest_collision_point,
     const rclcpp::Time nearest_collision_point_time,
-    const nav_msgs::msg::Odometry::ConstSharedPtr current_velocity_ptr, bool * need_to_stop,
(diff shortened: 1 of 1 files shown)


veqcc (approved):

kobayu858:
@brkay54 @beyzanurkaya @satoshi-ota @shmpwk @taikitanaka3 @tkimura4 

brkay54:
Thanks @kobayu858 -san, do you think we should change the variable name in the same named other function here?
- https://github.com/autowarefoundation/autoware.universe/pull/8018/files#diff-dcf194d2671df058128fd28886cb9184728818de13c9b2e8e043a553a591dda9L52-L58

kobayu858:
@brkay54 

Thanks for the confirmation.

As you said, I thought it would be better to change the names of the arguments in the above section, so I changed them.

The reason is that the argument name in the `.cpp` file is `current_odometry_ptr`.

Strangely enough, cppchecker did not warn me about this, so I overlooked it. Thank you very much.

brkay54 (approved):

PR #8043, opened 2024-07-16
fix(autoware_velocity_smoother): fix constParameterReference

kobayu858:
## Description

```
planning/autoware_velocity_smoother/src/node.cpp:1082:45: style: Parameter 'pt' can be declared as reference to const [constParameterReference]
    points.begin(), points.end(), [](auto & pt) { return pt.longitudinal_velocity_mps < 0; });
                                            ^
```
## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


veqcc (approved):

TakaHoribe (approved):
LGTM

PR #8048, opened 2024-07-16
fix(lane_change): delay lane change cancel

zulfaqar-azmi-t4:
## Description

Whenever there are vehicles parked in the target lane, we want to delay lane change as long as possible. Currently, the feature in implemented in the path generation. However, once the path is approved, delay lane change check is no longer performed.

This PR adds delay lane change after path is approved.
The passParkedVehicle function is also refactored, as the return boolean value do not properly correspond to the function name.

Before PR:

https://github.com/user-attachments/assets/6d24448c-6651-4483-8abc-8d021b062b83

After PR:
Lane change is cancelled, and it will be executed after front-most parked vehicle.

https://github.com/user-attachments/assets/6a01d369-2473-403a-841a-efd5ba2c9961

## Related links

**Parent Issue:**

- [TIER IV internal link](https://tier4.atlassian.net/browse/RT1-6955)

## How was this PR tested?

- PSIM
- [TIER IV internal scenario test](https://evaluation.tier4.jp/evaluation/reports/10098c88-697a-5bb9-aa24-7b408789656c?project_id=prd_jt)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

Might slightly increase computation time for cancel check.

### DIFF
--- a/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/include/autoware/behavior_path_lane_change_module/utils/utils.hpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/include/autoware/behavior_path_lane_change_module/utils/utils.hpp
@@ -178,10 +178,10 @@ bool isParkedObject(
   const ExtendedPredictedObject & object, const double buffer_to_bound,
   const double ratio_threshold);
 
-bool passParkedObject(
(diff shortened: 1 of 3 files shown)


rej55 (approved):
LGTM

PR #8063, opened 2024-07-16
fix(reaction_analyzer): fix constVariableReference

kobayu858:
## Description
This is a fix based on cppcheck constVariableReference warnings

```
tools/reaction_analyzer/src/subscriber.cpp:275:12: style: Variable 'buffer' can be declared as reference to const [constVariableReference]
    auto & buffer = std::get<MessageBuffer>(variant);
           ^

tools/reaction_analyzer/src/subscriber.cpp:366:12: style: Variable 'buffer' can be declared as reference to const [constVariableReference]
    auto & buffer = std::get<MessageBuffer>(variant);
           ^

tools/reaction_analyzer/src/subscriber.cpp:443:12: style: Variable 'buffer' can be declared as reference to const [constVariableReference]
    auto & buffer = std::get<MessageBuffer>(variant);
           ^

tools/reaction_analyzer/src/subscriber.cpp:523:12: style: Variable 'buffer' can be declared as reference to const [constVariableReference]
    auto & buffer = std::get<MessageBuffer>(variant);
           ^
```

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/tools/reaction_analyzer/src/subscriber.cpp
+++ b/tools/reaction_analyzer/src/subscriber.cpp
@@ -272,7 +272,7 @@ void SubscriberBase::on_trajectory(
   if (zero_vel_idx) {
     RCLCPP_INFO(node_->get_logger(), "%s reacted without published time", node_name.c_str());
     // set header time
-    auto & buffer = std::get<MessageBuffer>(variant);
+    const auto & buffer = std::get<MessageBuffer>(variant);
     buffer->header.stamp = msg_ptr->header.stamp;
     buffer->published_stamp = msg_ptr->header.stamp;
     mutex_.lock();
@@ -363,7 +363,7 @@ void SubscriberBase::on_pointcloud(
   if (search_pointcloud_near_pose(pcl_pointcloud, entity_pose_, entity_search_radius_)) {
     RCLCPP_INFO(node_->get_logger(), "%s reacted without published time", node_name.c_str());
     // set header time
-    auto & buffer = std::get<MessageBuffer>(variant);
+    const auto & buffer = std::get<MessageBuffer>(variant);
     buffer->header.stamp = msg_ptr->header.stamp;
     buffer->published_stamp = msg_ptr->header.stamp;
     mutex_.lock();
@@ -440,7 +440,7 @@ void SubscriberBase::on_predicted_objects(
   if (search_predicted_objects_near_pose(*msg_ptr, entity_pose_, entity_search_radius_)) {
     RCLCPP_INFO(node_->get_logger(), "%s reacted without published time", node_name.c_str());
     // set header time
-    auto & buffer = std::get<MessageBuffer>(variant);
+    const auto & buffer = std::get<MessageBuffer>(variant);
     buffer->header.stamp = msg_ptr->header.stamp;
     buffer->published_stamp = msg_ptr->header.stamp;
     mutex_.lock();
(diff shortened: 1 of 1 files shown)


brkay54:
Hi, @kobayu858 -san, we shouldn't make them `const` because we are modifying buffer variables in the same scope.

kobayu858:
@brkay54 

The auto* const form makes the pointer const, so the value can be changed. Therefore, I don't think there is any problem with the above.

However, setting the buffer to a pointer with * caused other build errors.

So I removed the & as a solution.

I think std::get<MessageBuffer>(variant) returns a pointer. Why was it in the form of taking a pointer reference?

brkay54:
> So I removed the & as a solution.
> 
> I think std::get(variant) returns a pointer. Why was it in the form of taking a pointer reference?

Did you try to run the tool? Because I am not sure it returns a pointer. I removed the `&` code snipped below:

```cpp
#include <variant>
#include <iostream>
#include <string>

struct MessageBuffer {
    std::string message;
};

int main() {
    std::variant<int, float, MessageBuffer> variant = MessageBuffer{"Initial message"};

    auto buffer = std::get<MessageBuffer>(variant);
    buffer.message = "Updated message";

    std::cout << "MessageBuffer now contains: " << std::get<MessageBuffer>(variant).message << std::endl;

    return 0;
}
```

For this example, changes in the buffer weren't reflected in the variant.

kobayu858:
> Did you try to run the tool? Because I am not sure it returns a pointer. I removed the & code snipped below:

I mistakenly thought it was a pointer because the buffer access method was->.

Since MessageBuffer is defined as std::optional<PublishedTime>, I realized that I was using -> to access the object that optional holds, even if it is by reference.

Thanks for the advice.

kobayu858:
We are discussing how to respond.

kobayu858:
@brkay54 cc @veqcc 

Corresponded by suppression.

```
$ cppcheck --enable=all --inconclusive --check-level=exhaustive --suppress=*:*/test/* --inline-suppr tools/reaction_analyzer/src/subscriber.cpp 
Checking tools/reaction_analyzer/src/subscriber.cpp ...
tools/reaction_analyzer/src/subscriber.cpp:15:0: information: Include file: "subscriber.hpp" not found. [missingInclude]
#include "subscriber.hpp"
^
tools/reaction_analyzer/src/subscriber.cpp:17:0: information: Include file: <algorithm> not found. Please note: Cppcheck does not need standard library headers to get proper results. [missingIncludeSystem]
#include <algorithm>
^
tools/reaction_analyzer/src/subscriber.cpp:18:0: information: Include file: <memory> not found. Please note: Cppcheck does not need standard library headers to get proper results. [missingIncludeSystem]
#include <memory>
^
tools/reaction_analyzer/src/subscriber.cpp:19:0: information: Include file: <utility> not found. Please note: Cppcheck does not need standard library headers to get proper results. [missingIncludeSystem]
#include <utility>
^
tools/reaction_analyzer/src/subscriber.cpp:224:64: style: Consider using std::find_if algorithm instead of a raw loop. [useStlAlgorithm]
          if (published_time->header.stamp == brake_cmd.stamp) {
                                                               ^
*/test/*:-1:0: information: Unmatched suppression: * [unmatchedSuppression]

^
nofile:0:0: information: Active checkers: 172/592 (use --checkers-report=<filename> to see details) [checkersReport]
```

brkay54 (approved):
Thanks for the effort!

veqcc (approved):

PR #8066, opened 2024-07-16
fix(static_obstacle_avoidance): avoid object behind unavoidance object if unavoidable is not on the path

satoshi-ota:
## Description

Now if front object is NOT avoidable as following fig, this module doesn't generate avoidance path even when behind object is avoidable. I think if the front object blocks ego path, it's proper behavior. But if the ego can go through, this module must generate avoidance path for behind object.

image

After this PR, this module will output avoidance path for behind object if front object is NOT on the path.

Screenshot from 2024-07-17 17-01-30

If front object is on ego path, this module won't generate avoidance path.

image

Additionally, if the front object is RIGHT and behind object is LEFT, this module won't generate avoidance path because the avoidance path for behind object might conflict front object, and vice versa.

image

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

- [x] [PASS TIER IV INTERNAL SCENARIOS](https://evaluation.tier4.jp/evaluation/reports/8e41e948-7474-5688-b3d7-51a5abddf37d?project_id=prd_jt)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


go-sakayori (approved):
LGTM

PR #8067, opened 2024-07-17
fix(mrm_handler): add check for autonomous mode and do some refactoring

danielsanchezaran:
## Description
Currently, if use_emergency_holding is set to true, the ego will enter emergency mode if it is not changed to autonomous mode within "timeout_emergency_recovery" seconds, which is not the expected behavior. 

This PR adds a check for Autonomous mode before increasing the timer, which solves the issue. 

## Related links

**Parent Issue:**

- Link

## How was this PR tested?
PSim

https://github.com/user-attachments/assets/f5441097-dd87-41ba-81b6-93d6faa81668

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/system/mrm_handler/src/mrm_handler/mrm_handler_core.cpp
+++ b/system/mrm_handler/src/mrm_handler/mrm_handler_core.cpp
@@ -95,10 +95,12 @@ void MrmHandler::onOperationModeAvailability(
     return;
   }
 
+  // If no time stamp is available, create one.
   stamp_autonomous_become_unavailable_ = (!stamp_autonomous_become_unavailable_.has_value())
                                            ? this->now()
                                            : stamp_autonomous_become_unavailable_;
 
+  // Check if autonomous mode unavailable time is larger than timeout threshold.
   const auto emergency_duration =
     (this->now() - stamp_autonomous_become_unavailable_.value()).seconds();
   is_emergency_holding_ = (emergency_duration > param_.timeout_emergency_recovery);
--- a/system/mrm_handler/src/mrm_handler/mrm_handler_core.cpp
+++ b/system/mrm_handler/src/mrm_handler/mrm_handler_core.cpp
@@ -95,7 +95,8 @@ void MrmHandler::onOperationModeAvailability(
     return;
   }
 
-  // If no time stamp is available, create one.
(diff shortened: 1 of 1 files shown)


system/mrm_handler/src/mrm_handler/mrm_handler_core.cpp
```diff
@@ -82,28 +82,26 @@ void MrmHandler::onOperationModeAvailability(
...
-        }
-      }
-    }
+  if (msg->autonomous) {
+    stamp_autonomous_become_unavailable_.reset();
+    return;
   }
-  operation_mode_availability_ = msg;
+
+  stamp_autonomous_become_unavailable_ = (!stamp_autonomous_become_unavailable_.has_value())
```
veqcc, line 98: I prefer the following. What do you think?

```
if (stamp_autonomous_become_unavailable_.has_value()) {
  const auto emergency_duration =
    (this->now() - stamp_autonomous_become_unavailable_.value()).seconds();
  is_emergency_holding_ = (emergency_duration > param_.timeout_emergency_recovery);
} else {
  stamp_autonomous_become_unavailable_.emplace(this->now());
}

TetsuKawa:
LGTM If you correct Kambe's points.

system/mrm_handler/src/mrm_handler/mrm_handler_core.cpp
danielsanchezaran: @veqcc I don't think it is worth it to add branching to the code, it decreases readability and increases cyclomatic complexity. Adding the else branch only skips these checks: 

`  const auto emergency_duration =
    (this->now() - stamp_autonomous_become_unavailable_.value()).seconds();
  is_emergency_holding_ = (emergency_duration > param_.timeout_emergency_recovery);`

which are not costly to compute. Also the cost of computing these operations might be lower than having a branching code. I believe Adding comments is ok, but dont agree with using an else branch if it is not necessary.

veqcc: Agreed :+1:

veqcc (approved):

PR #8068, opened 2024-07-17
refactor(freespace_planning_algorithm): refactor and improve astar search

mkquda:
## Description
Current implementation of hybrid a star path finding is very basic and can give undesirable results. Additionally the path finding time is unstable and varies widely depending on the space configuration.

- Current implementation only explores new nodes and does not update open nodes even if found at lower cost
- Current implementation imposes a cost only on direction change, no costs to account for reverse driving, curvature and smoothness
- Current implementation uses Reeds-Shepp distance for heuristic cost estimate, which can be inaccurate in the presence of obstacles

## Changes

- Fix hybrid a star implementation to explore both new and open nodes (when lower cost is found)
- Add cost to account for curvature (cost on steering)
- Add cost to account for smoothness (cost on steering change)
- Separate costs for reverse motion and switching direction
- Refactor node transition table code to use actual vehicle info and kinematic bicycle model to compute transitions
- Compute collision free distance map from goal
- Use maximum of Reeds-Shepp distance and collision free distance as heuristic cost estimate

## Related links

**Parent Issue:**

- Link

## How was this PR tested?
tested on psim.
compared performance with unit tests: (see [here](https://github.com/autowarefoundation/autoware.universe/pull/8091) for complete comparison)

## Notes for reviewers

None.

#### Additions and removals

| Removed| `minimum_turning_radius` | `double` | `9.0` | Max turning radius used for path search |
| Removed| `maximum_turning_radius`| `double` | `9.0` | Min turning radius used for path search  |
| Removed| `turning_radius_size`         | `double` | `1`    | Number of turning steps between min and max turning radius  |
| Added     | `maximum_turning_ratio`   | `double` | `0.7` | Max turn ratio of vehicle max steering used for path search |
| Added     | `turning_steps`                   | `double` | `1`    | Number of turning steps within turning range |
| Added     | `direction_change_weight` | `double` | `1.5` | additional cost factor for switching direction |
| Added     | `expansion_distance`         | `double` | `0.5` | length of expansion for A-star node transitions |
| Added     | `steering_change_weight`  | `double` | `0.5` | A-star cost factor for change in steering angle |

## Effects on system behavior

When using free space planner:

- Generated path should be natural
- Generated path should have minimum number of direction changes
- Generated path should be smooth and feasible for controller to follow
- Path generation time should be consistent and short for different space configurations


maxime-clem (approved):

planning/autoware_freespace_planning_algorithms/src/abstract_algorithm.cpp
```diff
@@ -105,20 +107,15 @@ double PlannerWaypoints::compute_length() const
 void AbstractPlanningAlgorithm::setMap(const nav_msgs::msg::OccupancyGrid & costmap)
 {
   costmap_ = costmap;
-  const auto height = costmap_.info.height;
-  const auto width = costmap_.info.width;
 
+  uint32_t nb_of_cells = costmap_.data.size();
```
kosuke55, line 111:
```suggestion
  const uint32_t nb_of_cells = costmap_.data.size();
```

planning/autoware_freespace_planning_algorithms/src/astar_search.cpp
```diff
@@ -119,66 +72,134 @@ AstarSearch::AstarSearch(
...
 {
-  transition_table_ = createTransitionTable(
-    planner_common_param_.minimum_turning_radius, planner_common_param_.maximum_turning_radius,
-    planner_common_param_.turning_radius_size, planner_common_param_.theta_size,
-    astar_param_.use_back);
+  steering_resolution_ =
+    collision_vehicle_shape_.max_steering / planner_common_param_.turning_steps;
+  heading_resolution_ = 2.0 * M_PI / planner_common_param_.theta_size;
+
+  double avg_steering =
```
kosuke55, line 79:
```suggestion
  const double avg_steering =
```

planning/autoware_freespace_planning_algorithms/src/astar_search.cpp
```diff
@@ -119,66 +72,134 @@ AstarSearch::AstarSearch(
...
   start_pose_ = global2local(costmap_, start_pose);
   goal_pose_ = global2local(costmap_, goal_pose);
 
-  clearNodes();
-  graph_.reserve(100000);
+  setCollisionFreeDistanceMap();
 
   if (!setStartNode()) {
-    return false;
+    throw std::logic_error("Invalid start pose");
```
kosuke55, line 127: what happen if throw, node will die or exception handling?

planning/autoware_freespace_planning_algorithms/src/astar_search.cpp
```diff
@@ -119,66 +72,134 @@ AstarSearch::AstarSearch(
...
 }
 
-void AstarSearch::clearNodes()
+void AstarSearch::resetData()
 {
   // clearing openlist is necessary because otherwise remaining elements of openlist
   // point to deleted node.
   openlist_ = std::priority_queue<AstarNode *, std::vector<AstarNode *>, NodeComparison>();
+  int nb_of_grid_nodes = costmap_.info.width * costmap_.info.height;
+  int total_astar_node_count = nb_of_grid_nodes * planner_common_param_.theta_size;
```
kosuke55, line 147:
```suggestion
  const int nb_of_grid_nodes = costmap_.info.width * costmap_.info.height;
  const int total_astar_node_count = nb_of_grid_nodes * planner_common_param_.theta_size;
```

planning/autoware_freespace_planning_algorithms/src/astar_search.cpp
```diff
@@ -119,66 +72,134 @@ AstarSearch::AstarSearch(
...
+    const auto & index = current.first;
+    for (const auto & offset_x : offsets) {
+      int x = index.x + offset_x;
+      for (const auto & offset_y : offsets) {
+        int y = index.y + offset_y;
+        IndexXY n_index{x, y};
+        double offset = std::abs(offset_x) + std::abs(offset_y);
+        if (isOutOfRange(n_index) || isObs(n_index) || offset < 1) continue;
+        int n_id = indexToId(n_index);
+        double dist = current.second + (sqrt(offset) * costmap_.info.resolution);
```
kosuke55, line 184:
```suggestion
    const int id = indexToId(current.first);
    if (closed[id]) continue;
    const closed[id] = true;

    const auto & index = current.first;
    for (const auto & offset_x : offsets) {
      const int x = index.x + offset_x;
      for (const auto & offset_y : offsets) {
        const int y = index.y + offset_y;
        cosnt IndexXY n_index{x, y};
        const double offset = std::abs(offset_x) + std::abs(offset_y);
        if (isOutOfRange(n_index) || isObs(n_index) || offset < 1) continue;
        const int n_id = indexToId(n_index);
        const double dist = current.second + (sqrt(offset) * costmap_.info.resolution);
```

planning/autoware_freespace_planning_algorithms/include/autoware/freespace_planning_algorithms/kinematic_bicycle_model.hpp
```diff
@@ -0,0 +1,73 @@
...
+  const double yaw, const double base_length, const double steering_angle, const double distance)
+{
+  geometry_msgs::msg::Pose pose;
+  if (abs(steering_angle) < eps) {
+    pose.position.x = distance * cos(yaw);
+    pose.position.y = distance * sin(yaw);
+    return pose;
+  }
+  double R = getTurningRadius(base_length, steering_angle);
+  double beta = distance / R;
```
kosuke55, line 47:
```suggestion
  const double R = getTurningRadius(base_length, steering_angle);
  const double beta = distance / R;
```

planning/autoware_freespace_planning_algorithms/src/astar_search.cpp
kosuke55: I found exception handling

```
  std::string error_msg;
  bool result = false;
  try {
    result = algo_->makePlan(current_pose_in_costmap_frame, goal_pose_in_costmap_frame);
  } catch (const std::exception & e) {
    error_msg = e.what();
  }
```

kosuke55 (approved):

Issue #7864, opened 2024-07-05
[Dense-Urban-ODD] Enhance Cruise Planner's Responsiveness to Other Road Users


brkay54:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I've agreed with the maintainers that I can plan this task.

### Description

Part of:
- #7485

### Dense Urban ODD Characteristics
- **Low-Speed Driving:** Typical speeds around 15 km/h.
- **Complex Traffic Elements:** Includes dense traffic with various dynamic interactions.
- **Dynamic Agents:** Frequent encounters with moving pedestrians and cyclists.
- **Bidirectional Movement:** Vehicles operate in lanes supporting bidirectional flow.
- **Traffic Rule Non-compliance:** Other road users may exhibit behaviors such as:
  - Driving in the wrong direction.
  - Cutting in unexpectedly from off-road areas.
  - Jaywalking pedestrians.
  - Improper overtaking by drivers.
  
In the context of Dense Urban ODD, we need to fine-tune the existing features of the `obstacle_cruise_planner` and potentially develop new features in speed planning to achieve the characteristics mentioned above.

### Improvement in Cruising for Cut-In and Cut-Out Obstacles

- The `obstacle_cruise_planner` package currently uses predicted objects in a limited way for speed planning. We need to consider objects further away from the trajectory by expanding the coverage area.
- The planner should respond more effectively to obstacles. If the predicted paths of other vehicles intersect with our planned path, Autoware should react much earlier to avoid potential collisions.

**Late Reaction to Potential Collisions**

[Video Link](https://github.com/autowarefoundation/autoware.universe/assets/45468306/c7124008-c9b3-437c-820c-432d6081c9b4)

**Late Reaction for Pedestrian**

[Video Link](https://github.com/autowarefoundation/autoware.universe/assets/45468306/fdb76b28-dd86-41bf-ae33-06dbb5ab1d63)

- For objects that briefly enter and exit our path without posing a significant collision risk, the planner should avoid unnecessary reactions. This selective response will prevent abrupt and unnecessary slowdowns, ensuring smoother and more efficient driving.

**Unnecessary Stop for Transient Object**

[Video Link](https://github.com/autowarefoundation/autoware.universe/assets/45468306/12d31bf5-c30b-4d59-95ad-073bc978dd1f)

[...]

PR #8072, opened 2024-07-17
feat(obstacle_cruise_planner): improve stop and cruise behavior for cut-in & out

brkay54:
## Description

- To increase the response capacity of the `obstacle_cruise_planner` within the scope of Dense-Urban ODD, we want it to be sensitive to dynamic objects approaching from outside Ego's path and stop if there is a collision risk, as mentioned in the issue.

- Avoiding unnecessary stopping of Ego in the presence of transient objects that do not pose a risk of collision. To solve these problems and provide a more human-like autonomous driving experience, we needed to improve the stop behavior of the `obstacle_cruise_planner`.

- We want to improve the cruising behavior for the outside obstacle. If another vehicle in the side lanes is approaching Ego's trajectory, cruise planning starts too late. 

**Improvement on filtering:**

In `obstacle_cruise_planner`, all the `PredictedObjects` were filtered if their lateral distance exceeded some small distance limitations. To achieve a behavior determination for PredictedObjects who are away from trajectory but approaching with some velocity, I needed to change the filtering algorithm. 

With the current implementation, if the predicted object is far away but its perpendicular velocity to the trajectory is high and if it is inside the trajectory within the time parameter `max_lateral_time_margin`, it will be considered as a target obstacle and we can use it for the behavior determination. 

road_with_predicted_objects drawio

**Improvement on stop behavior:**

To handle both transient objects and objects that come to the trajectory that may cause a collision, I divided the stop behavior into two: inside and outside.

For the objects that are inside the trajectory (precise_lat_dist > max_lat_margin_for_stop), we are checking the collision point. If there is a collision point, we are looking object's position when the Ego reaches the collision point. If there is no intersection with the future position, we don't set stop, otherwise set stop.

**Improvement on cruising:**

road_with_predicted_objects drawio

To be able to select the vehicles in side lanes who intend to drive through our trajectory, we added a time margin that calculates the time to reach the trajectory.  By using this, we can select the vehicles earlier as cruise obstacle.

## Related links

**Parent Issue:**
- https://github.com/autowarefoundation/autoware.universe/issues/7864

## How was this PR tested?

PR tested by using PSim, however, I am currently working to create some scenarios for a detailed test of this feature. I will upload them into AWF Project asap.

**Transient Obstacle Tests:**

https://github.com/user-attachments/assets/d90577f9-0fab-4d38-954f-1416114dbd3c

**Possible Collision Obstacles:**

https://github.com/user-attachments/assets/b36ca8b0-e039-4113-830f-678e3318c711

**Cruising for Vehicles come from side lane:**

https://github.com/user-attachments/assets/a4cdab86-587f-4d28-81b0-9c44b1fdcf00

## Notes for reviewers
Please test this PR with the parameters below:
- https://github.com/autowarefoundation/autoware_launch/pull/1142
## Interface changes

None.

#### Additions and removals

| Change type   | Parameter Name                                      | Type   | Default Value | Description                                                                      |
|:--------------|:----------------------------------------------------|:-------|:--------------|:---------------------------------------------------------------------------------|
| Modified      | common.min_behavior_stop_margin                     | double | 6.0           | Updated minimum stop margin for behavior-based stopping.                         |
| Modified      | behavior_determination.obstacle_velocity_threshold_from_cruise_to_stop | double | 1.0   | Updated velocity threshold for switching from cruise to stop.                    |
| Modified      | behavior_determination.obstacle_velocity_threshold_from_stop_to_cruise | double | 1.5   | Updated velocity threshold for switching from stop to cruise.                    |
| Modified      | behavior_determination.crossing_obstacle.obstacle_velocity_threshold | double | 1.0   | Updated velocity threshold for determining crossing obstacles for cruise or stop.|
| Modified      | behavior_determination.crossing_obstacle.obstacle_traj_angle_threshold | double | 0.523599 | Updated yaw angle threshold for crossing obstacles.                              |
| Added         | behavior_determination.stop.outside_obstacle.max_lateral_time_margin | double | 4.0   | Maximum time margin for lateral collision checking between obstacles and ego.    |
| Added         | behavior_determination.stop.outside_obstacle.num_of_predicted_paths | int    | 3     | Number of highest confidence predicted paths to check collision between obstacle and ego. |
| Modified      | behavior_determination.stop.crossing_obstacle.collision_time_margin | double | 2.0   | Updated time threshold for collision between obstacle and ego.                   |
| Added         | behavior_determination.cruise.outside_obstacle.max_lateral_time_margin | double | 5.0   | Maximum time margin for lateral collision checking between obstacles and trajectory band with ego's width. |
| Added         | behavior_determination.cruise.outside_obstacle.num_of_predicted_paths | int    | 3     | Number of highest confidence predicted paths to check collision between obstacle and ego. |
| Modified      | behavior_determination.cruise.outside_obstacle.obstacle_velocity_threshold | double | 1.5   | Updated velocity threshold for obstacles outside the trajectory to cruise or stop.|
| Modified      | behavior_determination.cruise.outside_obstacle.ego_obstacle_overlap_time_threshold | double | 0.2 | Updated time threshold to decide cut-in obstacle for cruise or stop.             |
| Modified      | behavior_determination.cruise.outside_obstacle.max_prediction_time_for_collision_check | double | 10.0 | Updated prediction time for checking collision between obstacle and ego.         |
| Added      | behavior_determination.stop.min_velocity_to_reach_collision_point | double | 2.0 |  minimum velocity margin to calculate time to reach collision        |
## Effects on system behavior

None.

### DIFF
--- a/planning/autoware_obstacle_cruise_planner/config/obstacle_cruise_planner.param.yaml
+++ b/planning/autoware_obstacle_cruise_planner/config/obstacle_cruise_planner.param.yaml
@@ -80,6 +80,7 @@
       pointcloud_min_cluster_size: 1
       pointcloud_max_cluster_size: 100000
 
+      max_lateral_time_margin: 3.0 # maximum time margin to filtering obstacles for collision checking
       decimate_trajectory_step_length : 2.0 # longitudinal step length to calculate trajectory polygon for collision checking
 
       prediction_resampling_time_interval: 0.1
--- a/planning/autoware_obstacle_cruise_planner/config/obstacle_cruise_planner.param.yaml
+++ b/planning/autoware_obstacle_cruise_planner/config/obstacle_cruise_planner.param.yaml
@@ -110,6 +110,7 @@
       stop:
         max_lat_margin: 0.3 # lateral margin between the obstacles except for unknown and ego's footprint
         max_lat_margin_against_unknown: 0.3 # lateral margin between the unknown obstacles and ego's footprint
+        min_velocity_to_reach_collision_point: 2.0 # minimum velocity margin to calculate time to reach collision [m/s]
         outside_obstacle:
           max_lateral_time_margin: 4.0 # time threshold of lateral margin between the obstacles and ego's footprint [s]
           num_of_predicted_paths: 3 # number of the highest confidence predicted path to check collision between obstacle and ego
--- a/planning/autoware_obstacle_cruise_planner/config/obstacle_cruise_planner.param.yaml
+++ b/planning/autoware_obstacle_cruise_planner/config/obstacle_cruise_planner.param.yaml
@@ -114,6 +114,8 @@
         outside_obstacle:
           max_lateral_time_margin: 4.0 # time threshold of lateral margin between the obstacles and ego's footprint [s]
           num_of_predicted_paths: 3 # number of the highest confidence predicted path to check collision between obstacle and ego
+          pedestrian_deceleration_rate: 0.5 # planner perceives pedestrians will stop with this rate to avoid unnecessary stops [m/ss]
+          bicycle_deceleration_rate: 0.5 # planner perceives bicycles will stop with this rate to avoid unnecessary stops [m/ss]
         crossing_obstacle:
           collision_time_margin : 4.0 # time threshold of collision between obstacle adn ego for cruise or stop [s]
 
--- a/planning/autoware_obstacle_cruise_planner/config/obstacle_cruise_planner.param.yaml
+++ b/planning/autoware_obstacle_cruise_planner/config/obstacle_cruise_planner.param.yaml
@@ -29,15 +29,26 @@
       suppress_sudden_obstacle_stop: false
 
       stop_obstacle_type:
-        unknown: true
-        car: true
-        truck: true
-        bus: true
-        trailer: true
-        motorcycle: true
-        bicycle: true
-        pedestrian: true
         pointcloud: false
+        inside:
+          unknown: true
+          car: true
+          truck: true
+          bus: true
+          trailer: true
+          motorcycle: true
+          bicycle: true
+          pedestrian: true
+
+        outside:
+          unknown: false
+          car: true
+          truck: true
+          bus: true
+          trailer: true
+          motorcycle: true
+          bicycle: true
+          pedestrian: true
 
       cruise_obstacle_type:
         inside:
@@ -80,7 +91,6 @@
       pointcloud_min_cluster_size: 1
       pointcloud_max_cluster_size: 100000
 
-      max_lateral_time_margin: 3.0 # maximum time margin to filtering obstacles for collision checking
       decimate_trajectory_step_length : 2.0 # longitudinal step length to calculate trajectory polygon for collision checking
 
       prediction_resampling_time_interval: 0.1
@@ -100,6 +110,9 @@
       stop:
         max_lat_margin: 0.3 # lateral margin between the obstacles except for unknown and ego's footprint
         max_lat_margin_against_unknown: 0.3 # lateral margin between the unknown obstacles and ego's footprint
+        outside_obstacle:
+          max_lateral_time_margin: 4.0 # time threshold of lateral margin between the obstacles and ego's footprint [s]
+          num_of_predicted_paths: 3 # number of the highest confidence predicted path to check collision between obstacle and ego
         crossing_obstacle:
           collision_time_margin : 4.0 # time threshold of collision between obstacle adn ego for cruise or stop [s]
 
@@ -109,6 +122,8 @@
           obstacle_velocity_threshold : 3.5 # minimum velocity threshold of obstacles outside the trajectory to cruise or stop [m/s]
           ego_obstacle_overlap_time_threshold : 2.0 #  time threshold to decide cut-in obstacle for cruise or stop [s]
           max_prediction_time_for_collision_check : 20.0 # prediction time to check collision between obstacle and ego
+          max_lateral_time_margin: 5.0 # time threshold of lateral margin between obstacle and trajectory band with ego's width [s]
+          num_of_predicted_paths: 3 # number of the highest confidence predicted path to check collision between obstacle and ego
         yield:
           enable_yield: true
           lat_distance_threshold: 5.0 # lateral margin between obstacle in neighbor lanes and trajectory band with ego's width for yielding
--- a/planning/autoware_obstacle_cruise_planner/include/autoware/obstacle_cruise_planner/common_structs.hpp
+++ b/planning/autoware_obstacle_cruise_planner/include/autoware/obstacle_cruise_planner/common_structs.hpp
@@ -58,10 +58,13 @@ struct Obstacle
   Obstacle(
     const rclcpp::Time & arg_stamp, const PredictedObject & object,
     const geometry_msgs::msg::Pose & arg_pose, const double ego_to_obstacle_distance,
-    const double lat_dist_from_obstacle_to_traj)
+    const double lat_dist_from_obstacle_to_traj, const double tangent_velocity,
+    const double normal_velocity)
   : stamp(arg_stamp),
     ego_to_obstacle_distance(ego_to_obstacle_distance),
     lat_dist_from_obstacle_to_traj(lat_dist_from_obstacle_to_traj),
+    tangent_velocity(tangent_velocity),
+    normal_velocity(normal_velocity),
     pose(arg_pose),
     orientation_reliable(true),
     twist(object.kinematics.initial_twist_with_covariance.twist),
@@ -94,6 +97,8 @@ struct Obstacle
   rclcpp::Time stamp;  // This is not the current stamp, but when the object was observed.
   double ego_to_obstacle_distance;
   double lat_dist_from_obstacle_to_traj;
+  double tangent_velocity;
+  double normal_velocity;
 
   // for PredictedObject
   geometry_msgs::msg::Pose pose;  // interpolated with the current stamp
--- a/planning/autoware_obstacle_cruise_planner/include/autoware/obstacle_cruise_planner/common_structs.hpp
+++ b/planning/autoware_obstacle_cruise_planner/include/autoware/obstacle_cruise_planner/common_structs.hpp
@@ -58,13 +58,13 @@ struct Obstacle
   Obstacle(
     const rclcpp::Time & arg_stamp, const PredictedObject & object,
     const geometry_msgs::msg::Pose & arg_pose, const double ego_to_obstacle_distance,
-    const double lat_dist_from_obstacle_to_traj, const double tangent_velocity,
-    const double normal_velocity)
+    const double lat_dist_from_obstacle_to_traj, const double longitudinal_velocity,
+    const double approach_velocity)
   : stamp(arg_stamp),
     ego_to_obstacle_distance(ego_to_obstacle_distance),
     lat_dist_from_obstacle_to_traj(lat_dist_from_obstacle_to_traj),
-    tangent_velocity(tangent_velocity),
-    normal_velocity(normal_velocity),
+    longitudinal_velocity(longitudinal_velocity),
+    approach_velocity(approach_velocity),
     pose(arg_pose),
     orientation_reliable(true),
     twist(object.kinematics.initial_twist_with_covariance.twist),
@@ -97,8 +97,8 @@ struct Obstacle
   rclcpp::Time stamp;  // This is not the current stamp, but when the object was observed.
   double ego_to_obstacle_distance;
   double lat_dist_from_obstacle_to_traj;
-  double tangent_velocity;
-  double normal_velocity;
+  double longitudinal_velocity;
+  double approach_velocity;
 
   // for PredictedObject
   geometry_msgs::msg::Pose pose;  // interpolated with the current stamp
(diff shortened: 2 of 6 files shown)


planning/autoware_obstacle_cruise_planner/src/node.cpp
```diff
@@ -1454,38 +1506,17 @@ std::optional<StopObstacle> ObstacleCruisePlannerNode::createStopObstacleForPred
...
   if (!isStopObstacle(obstacle.classification.label)) {
     return std::nullopt;
   }
-
   const double max_lat_margin_for_stop =
     (obstacle.classification.label == ObjectClassification::UNKNOWN)
       ? p.max_lat_margin_for_stop_against_unknown
       : p.max_lat_margin_for_stop;
 
+  // Obstacle not inside of trajectory check predicted path of the obstacle
```
yuki-takagi-66, line 1517: May I ask you to refactor `createStopObstacleForPredictedObject()` ?
I would like to reduce duplicate code even if we have to modify the existing code.

yuki-takagi-66 (commented):
Thank you for the PR. I've been waiting for this change.
As I commented out, I feel we have to refactor `createStopObstacleForPredictedObject()`.
Could you check it

brkay54:
Some scenario tests:

**Possible Collision (UC-NTR-003-0001):**
https://github.com/user-attachments/assets/6c87f422-6951-4b24-a0bb-8ad466e62d45

**Pedestrian Jay Walk  (UC-VRU-001-0002):**
https://github.com/user-attachments/assets/c70a39f0-1c20-4b74-9512-020096a737a0

**Transient Object (UC-NTR-003-0002):**
https://github.com/user-attachments/assets/a712aa01-3129-40f3-a689-52bbd4e97a38

**Stop with high negative acceleration:**
https://github.com/user-attachments/assets/30298e50-6e2e-4a12-997f-5c1081a6f550

planning/autoware_obstacle_cruise_planner/src/node.cpp
yuki-takagi-66: Thank you!  I'll response by the next Wednesday. sorry for the late response.

yuki-takagi-66:
@brkay54 
It is a great feature, so many matters need to be considered, and the feature still needs to be refined.
It behaves undesirably in some situations. Could you check the followings?

1. A stop wall is inserted against the object at the lane change destination.
Screenshot from 2024-08-02 11-57-07

2. A stop decision is made even though the ego can pass the collision point before the pedestrian (the pedestrian will stop).
Screenshot from 2024-08-02 12-06-04

3.No stop wall against the in front object
Screenshot from 2024-08-02 13-32-12

brkay54:
@yuki-takagi-66 -san, thank you for your review, and sorry for being late. Actually, I was planning to create two PRs: improvement for stop obstacles and improvement for cruise obstacles. However, without cruise improvements, we made unnecessary stops in some cases (Like the problem 1 you mentioned). 

Therefore I merged both improvements and updated this PR. Could you please take another look at the PR description?

I solved almost all the problems you mentioned and retested scenarios:
 
> 1. A stop wall is inserted against the object at the lane change destination.

After improvements in cruise planning, it was solved.

Video: https://youtu.be/64EdnFO_U-Q

> 2. A stop decision is made even though the ego can pass the collision point before the pedestrian (the pedestrian will stop).

For this problem, I tested all the cases. There are five cases and 4 of them were passed. Only the one case whose Ego velocity 30 kph failed. I investigated the cause and I saw that stop planning for that case made sense to me. I checked the collision time margin:

 time_to_reach_collision_point: 3.00236, time_to_start_cross: 3, time_to_end_cross: 6.2

If the planner does not set a stop, ego and obstacle will crash.

Video: https://youtu.be/aCxZ_bDvbHc

> 3.No stop wall against the in front object

It is solved: https://youtu.be/WxU56BZn6Fs

Thanks for your efforts again. And, could you review the updated PR again? Also, for Autoware Evaluator could you use the parameters I want to merge with this PR?

- https://github.com/autowarefoundation/autoware_launch/pull/1142

yuki-takagi-66:
@brkay54 
Please let me know when your changes are complete.

brkay54:
@yuki-takagi-66 -san, sorry to being late. I just realized a wrong calculation for obstacle velocity projection and I made it safer while labeling transient obstacles. I tested with Dense-Urban ODD scenarios and the scenarios you provided above. All seemed OK to me except "DistR_UC-v2-D-01-00004_004_case01_cmn_general". In this scenario, my implementation sets STOP for obstacles while they were moving to the road. IMO, this behavior is okay because these pedestrians might not stop. The purpose of this improvement is to be reactive to outside obstacles. I think we can ignore this scenario. I commented out the failure condition to show current behavior:

https://github.com/user-attachments/assets/79abf153-634c-43ec-82cc-116819151955

It is ready for the tests currently.

yuki-takagi-66:
@brkay54 
Thank you. I'll response by the next week. Sorry for late.

yuki-takagi-66:
@brkay54 
Thank you. I've tested this PR also by the tier4 scenario tests.
It almost PASSED, except for the situation of pedestrians near crosswalks.

Regarding the subject behavior, I think it depends on the specifics of the situation whether it is reasonable.
In urban areas, it may be appropriate to assume that pedestrians will ignore the red light and enter the street.
In the suburbs, however, this may not be the case.
It may also vary from country to country or region to region.

To solve this problem, for the pedestrians and cyclists, how about limiting the number of seconds of predicted paths to a new parameter value, or adding logic that only sees the path time it takes them to stop, calculated from the specified deceleration rate?
This is based on the assumption that pedestrians can stop in a shorter time (around 1s) than vehicles.

Maybe we can combine them into one parameter with stop_obstacle_type.outside.

We would appreciate your consideration.

brkay54:
@yuki-takagi-66 
Thank you for your feedback and suggestions!

> To solve this problem, for the pedestrians and cyclists, how about limiting the number of seconds of predicted paths to a new parameter value, or adding logic that only sees the path time it takes them to stop, calculated from the specified deceleration rate?

It was a good idea and I implemented it. I tested `DistR_UC-v2-D-01-00004_004_case01_cmn_general` and it passed. I updated this PR and [autoware_launch PR](https://github.com/autowarefoundation/autoware_launch/pull/1142). Please check the changes. 

By the way, I also changed some values to default values in the launch file, I changed them for passing all scenarios you mentioned above. (For instance, min_behavior_stop_margin was 3.0 but I changed it to 6.0 to pass a scenario you provided.) My colleague also works to tune the cruise planner's parameters for following front obstacles so I don't want to conflict with him. 

If you are okay with the current changes, I want to update the readme with another PR because I am also preparing another PR for the behind obstacle I mentioned in the [issue](https://github.com/autowarefoundation/autoware.universe/issues/7864). With that PR, I will also update the readme. 

Thank you again @yuki-takagi-66 -san!

planning/autoware_obstacle_cruise_planner/include/autoware/obstacle_cruise_planner/common_structs.hpp
```diff
@@ -94,6 +97,8 @@ struct Obstacle
   rclcpp::Time stamp;  // This is not the current stamp, but when the object was observed.
   double ego_to_obstacle_distance;
   double lat_dist_from_obstacle_to_traj;
+  double tangent_velocity;
+  double normal_velocity;
```
yuki-takagi-66, line 101: I feel this name has unclear meaning.
This sign of this value depends the location of the obstacle.
So I suggest to change this variable name, for example, `approach velocity`.

planning/autoware_obstacle_cruise_planner/include/autoware/obstacle_cruise_planner/common_structs.hpp
```diff
@@ -94,6 +97,8 @@ struct Obstacle
   rclcpp::Time stamp;  // This is not the current stamp, but when the object was observed.
   double ego_to_obstacle_distance;
   double lat_dist_from_obstacle_to_traj;
+  double tangent_velocity;
```
yuki-takagi-66, line 100: If we will change the name `normal_velocity`, we should rename as well.

planning/autoware_obstacle_cruise_planner/src/node.cpp
```diff
@@ -99,20 +118,48 @@ double calcDiffAngleAgainstTrajectory(
...
+ * It returns the longitudinal and perpendicular components of the obstacle's velocity
+ * with respect to the trajectory. Negative perpendicular velocity indicates that the
+ * obstacle is getting far away from the trajectory.
+ *
+ * @param traj_points The trajectory points.
+ * @param obstacle_pose The current pose of the obstacle.
+ * @param obstacle_twist The twist (velocity) of the obstacle.
+ * @return A pair containing the longitudinal and perpendicular velocity components.
+ */
 std::pair<double, double> projectObstacleVelocityToTrajectory(
```
yuki-takagi-66, line 134: I guess we should change this function name since the significance of the function has changed.

planning/autoware_obstacle_cruise_planner/src/node.cpp
```diff
@@ -1454,71 +1652,81 @@ std::optional<StopObstacle> ObstacleCruisePlannerNode::createStopObstacleForPred
...
-      RCLCPP_INFO_EXPRESSION(
-        get_logger(), enable_debug_info_,
-        "[Stop] Ignore inside obstacle (%s) since it will not collide with the ego.",
-        object_id.c_str());
-      debug_data_ptr_->intentionally_ignored_obstacles.push_back(obstacle);
+    // brkay54: For the pedestrians and bicycles, we need to check the collision point by thinking
+    // they will stop with a predefined deceleration rate to avoid unnecessary stops.
+    double resample_time_horizon = p.prediction_resampling_time_horizon;
+    if (obstacle.classification.label == ObjectClassification::PEDESTRIAN) {
+      resample_time_horizon =
```
yuki-takagi-66, line 1681: For the case to calculate the time to stop, sure tihs `velocty / -acceleration` is correct, however, I think we want to calculate resample_time_horizon as a argument value to `resampleHighestConfidencePredictedPaths()`.
`resampleHighestConfidencePredictedPaths()` assumes the object moves with constant velocity (or twist), but we are assuming the pedestrians are decelerate constantly.
Therefore should we change the formula as `velocty / (2.0 * -acceleration)`?

yuki-takagi-66 (commented):
Overall, LGTM. Almost all scenarios in tier4 has passed!!
I've add some comments and could you check the comments?

planning/autoware_obstacle_cruise_planner/src/node.cpp
```diff
@@ -738,12 +821,23 @@ std::vector<Obstacle> ObstacleCruisePlannerNode::convertToObstacles(
...
   std::vector<Obstacle> target_obstacles;
   for (const auto & predicted_object : objects.objects) {
     const auto & object_id =
       autoware::universe_utils::toHexString(predicted_object.object_id).substr(0, 4);
+
+    // brkay54: When use_prediction is true, we observed wrong orientation for the object in
+    // scenario simulator.
     const auto & current_obstacle_pose =
-      obstacle_cruise_utils::getCurrentObjectPose(predicted_object, obj_stamp, now(), true);
+      obstacle_cruise_utils::getCurrentObjectPose(predicted_object, obj_stamp, now(), false);
```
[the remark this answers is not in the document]
yuki-takagi-66: Can I ask you to create the issue and solve this problem?
I saw this problem also when using the Psim with use_sim_time:=true

If this problem does not blocking any of your work, I feel we have option to ignore this problem.

planning/autoware_obstacle_cruise_planner/config/obstacle_cruise_planner.param.yaml
```diff
@@ -108,6 +125,8 @@
           obstacle_velocity_threshold : 3.5 # minimum velocity threshold of obstacles outside the trajectory to cruise or stop [m/s]
           ego_obstacle_overlap_time_threshold : 2.0 #  time threshold to decide cut-in obstacle for cruise or stop [s]
           max_prediction_time_for_collision_check : 20.0 # prediction time to check collision between obstacle and ego
+          max_lateral_time_margin: 5.0 # time threshold of lateral margin between obstacle and trajectory band with ego's width [s]
+          num_of_predicted_paths: 3 # number of the highest confidence predicted path to check collision between obstacle and ego
```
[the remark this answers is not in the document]
yuki-takagi-66: Thank you. 
I think `num_of_predicted_paths` should be `3`.

I understood that set `num_of_predicted_paths` as `3` is  a significance of this PR and there will no behavior problem.
Due to tier4 internal reasons, we will temporarily adopt `1` in the fork repository, but will try to follow `3` within a few months.

yuki-takagi-66 (approved):
Thank you for the great job.
I have no concerns and all tier4 internal tests have passe.

brkay54:
@yuki-takagi-66 -san thank you so much for your effort and suggestions 🙏🏻 . Also, could you approve the [PR in Autoware Launch](https://github.com/autowarefoundation/autoware_launch/pull/1142)?

yuki-takagi-66:
@brkay54 
>Also, because I made some critical changes, I would like to be a maintainer of this package. If you and colleagues from the TierIV side don't mind, can I add myself as a maintainer?

Thank you for your contribution. TierIV would greatly appreciate it if you could become a maintainer.

PR #8093, opened 2024-07-18
fix(lane_change): filtering object ahead of terminal

zulfaqar-azmi-t4:
## Description

This PR aims to fix object filtering when dealing with objects ahead of terminal point.

Terminal point is where lane change is perform  with the minimum distance either at the terminal lane or near goal.

The end of current lanes will be the last possible lane where ego can perform lane change. We can call this as terminal lane. As shown in the following diagram

lane_change_redesign-Page-7

Any lanes ahead of the of terminal lane is consider ahead terminal.

Similar to if the terminal lane is within goal route section, then anything after goal is consider ahead terminal.

lane_change_redesign-Page-7 (1)

#### Before PR

In some case, object after lane change end pose is still considered as target lane object, even though it is ahead of terminal lane. Note the cyan square indicating target lane objects

https://github.com/user-attachments/assets/be755147-1b4b-4246-b56b-5ac0521f44c1

#### After PR

Object ahead of terminal lane is no longer consider as target lane objects. Note that is is now purple square, indicating other lane object

https://github.com/user-attachments/assets/96d852ce-2b83-4071-91e9-55011888e9b3

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

[TIER IV internal link](https://evaluation.tier4.jp/evaluation/reports/dc4f22d0-db9a-56de-afef-dd91705e1860?project_id=prd_jt)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/README.md
+++ b/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/README.md
@@ -331,8 +331,7 @@ title NormalLaneChange::filterObjects Method Execution Flow
 start
 
 group "Filter Objects by Class" {
-:Iterate through each object in objects list;
-while (has not finished iterating through object list) is (TRUE)
+while (has not finished iterating through predicted object list) is (TRUE)
   if (current object type != param.object_types_to_check?) then (TRUE)
   #LightPink:Remove current object;
 else (FALSE)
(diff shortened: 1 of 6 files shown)


kyoichi-sugahara:
@zulfaqar-azmi-t4 
can you rebase main branch and remove unrelated reviewers? :cry:

rej55 (approved):

PR #8095, opened 2024-07-18
fix(ekf_localizer): add_delay_compensation_for_roll_and_pitch

meliketanrikulu:
## Description
We can see that there is a wobbling of ego vehicle points as the vehicle drives through speed bumps.I repeated the tests in the simulation environment and obtained similar results. When I tried to optimize parameters in its current EKF , I saw that I could not achieve good results in its current packet. You can find the tests I have previously done to solve this problem in the https://github.com/autowarefoundation/autoware.universe/issues/6993 comments

Delay compensation has been added to improve roll and pitch estimation.
## Related links

**Parent Issue:**

Related Issue: https://github.com/autowarefoundation/autoware.universe/issues/6993

## How was this PR tested?
1.) I tested this PR with rosbag file . Added delay compensation for roll and pitch estimation. This PR includes this changes.
2.) I updated 1D Filter params for pitch estimation.
Changed [pitch_filter_proc_dev](https://github.com/autowarefoundation/autoware_launch/blob/244285f5f7999f4390c2f65f36a112c233b61143/autoware_launch/config/localization/ekf_localizer.param.yaml#L34) param as 0.5
3.) Enabled [3D Distortion Corrector](https://github.com/autowarefoundation/sample_sensor_kit_launch/blob/59843a4362684e8eb8b8714f66a01d21337c9758/common_sensor_launch/config/distortion_corrector_node.param.yaml#L5) (optionally)

### Installation
1. Download and unpack a test map files.
  *  You can also download [the map](https://drive.google.com/file/d/1Xx4H6qou_Z764_vXpQTYsh6gio4aOJ8C/view?usp=sharing) manually.
```bash
gdown --id 1Xx4H6qou_Z764_vXpQTYsh6gio4aOJ8C -O ~/autoware_data/  
unzip -d ~/autoware_data/ ~/autoware_data/autoware_map.zip
```
 
2. Download the test rosbag files.
  *  (Test without 3D Distortion Corrector -default-)You can also download [the rosbag file](https://drive.google.com/drive/folders/10V7P2ho4y75lrW5QFasCH5BkkMkgUTNL?usp=sharing) manually.
```bash
gdown --id 17qpFAmRT_iKnll8IGzApIs7qmJN1bgoD -O ~/autoware_data/  
unzip -d ~/autoware_data/ ~/autoware_data/rosbag2_2024_07_23-12_46_03.zip
```

  *  (Test with 3D Distortion Corrector -optional-)You can also download [the rosbag file](https://drive.google.com/drive/folders/1LUy-1BOTiJV4QdU7aJCzID1K0jM8TZ7i?usp=sharing) manually.
```bash
gdown --id 1LjF49QosxdoZDRtRSyy--jR9GMyfq9pZ -O ~/autoware_data/  
unzip -d ~/autoware_data/ ~/autoware_data/rosbag2_2024_07_23-12_36_27.zip
```
### Prepare Autoware to test PR: 
1. Checkout autoware.universe : 
```bash
cd ~/autoware/src/universe/autoware.universe/
git remote add autoware.universe https://github.com/meliketanrikulu/autoware.universe.git
git remote update
git checkout "fix(ekf_localizer)add_delay_compensation_for_roll_and_pitch"

```
2. Checkout autoware_launch : 
```bash
cd ~/autoware/src/launcher/autoware_launch/
git remote add https://github.com/meliketanrikulu/autoware_launch.git
git remote update
git checkout test_pr8095
```
4. Compile required packages : 
```bash
cd ~/autoware
colcon build --symlink-install --packages-select autoware_launch ekf_localizer
```
### Launch Autoware
1. Run Autoware : 
```bash
source ~/autoware/install/setup.bash
ros2 launch autoware_launch test_pr8095.launch.xml map_path:=$HOME/autoware_data/autoware_map
```
2. Run bag file : 

Run this for testing without 3D Distortion Corrector - Default-
```bash
source ~/autoware/install/setup.bash
ros2 bag play ~/autoware_data/rosbag2_2024_07_23-12_46_03/rosbag2_2024_07_23-12_46_03_0.db3
```
Run this for testing with 3D Distortion Corrector. It can be improve the results. -Optional- 

```bash
source ~/autoware/install/setup.bash
ros2 bag play ~/autoware_data/rosbag2_2024_07_23-12_36_27/rosbag2_2024_07_23-12_36_27_0.db3
```
### Test Results:
You can see changes Before/After this PR video is [here](https://www.youtube.com/watch?v=PQeU31DiZRY)
After this test I added 3D Distortion Corrector too . Final Test Video is [here](https://www.youtube.com/watch?v=XTsb5sTRQPQ).

You can see below, the delay compensation effect on EKF output poses with roll and pitch delay compensation : 

roll drawio

pitch drawio

#### Comparison with Previous Works:
I have previously tried to solve the problem by simply changing parameters with the existing autoware, but I could not achieve successful results. You can find the tests I performed at this stage under the [related issue](https://github.com/autowarefoundation/autoware.universe/issues/6993).

[Previous PR](https://github.com/autowarefoundation/autoware.universe/pull/7875) : 
To solve this problem, I previously suggested and tested adding imu to EKF. The addition of IMU also contributed positively to the result. However, since changing Autoware's architecture was not our first choice, I worked on a new method. As a result, I saw that the problem could be solved by making delay compensation for roll and pitch in the ekf_localizer package and I created this PR. I aim to explain in the image below how close the PR I have worked on before is to the result we expected (NDT only localization) and how close this PR is to the result we expected not to happen (NDT only localization).

Below, you can review the images captured at the moments when the difference between the instantaneous point cloud and the map was maximized while passing through the speed bump. (Normally instant pc and map points are matching but they move away when passing through a speed bump. You can see here the maximum distances (errors)):
Note : Ignore the horizontal differences between the two point clouds here.
COMPARISON-Page-1-Page-3 drawio

## Interface changes

None.

## Effects on system behavior

None.


SakodaShintaro:
Thank you for PR :+1: 
I'm glad that the method improved the EKF.

Since this pull request changes EKF, we would like to evaluate new code carefully in our validation data. It takes 6~7 hours.
Please wait a little moment. Evaluation will be done in this tonight or tomorrow.

meliketanrikulu:
> I'm glad that the method improved the EKF.

Thank you very much @SakodaShintaro . I'm waiting for your review .

SakodaShintaro (approved):
Thank you for waiting. We have evaluated this PR with our validation data. There is no significant difference from before. I also reviewed the code. LGTM

YamatoAndo:
@meliketanrikulu Overall, it looks good to me, but if the Twist msg is interrupted, the values of `roll_rate_` and `pitch_rate_` will be retained. Please implement a measure to initialize them to 0 if there are no updates to the Twist msg.

YamatoAndo (changes requested):
@meliketanrikulu Overall, it looks good to me, but if the twist msg is interrupted, the values of roll_rate_ and pitch_rate_ will be retained. Please implement a measure to initialize them to 0 if there are no updates to the twist msg.

localization/ekf_localizer/src/ekf_module.cpp
```diff
@@ -329,6 +341,8 @@ bool EKFModule::measurement_update_twist(
   /* Set measurement matrix */
   Eigen::MatrixXd y(dim_y, 1);
   y << twist.twist.twist.linear.x, twist.twist.twist.angular.z;
+  roll_rate_ = twist.twist.twist.angular.x;
+  pitch_rate_ = twist.twist.twist.angular.y;
```
TaikiYamada4, line 345: I believe that taking the independent factor of the angular velocity vector as roll and pitch velocity is not mathematically correct, because the norm of the angular velocity vector equals to the angular variation around the major axis. In other words, the `roll_rate_` and `pitch_rate_` can be affected by other factors.

In my opinion, calculating the variation using the quaternion library should be better for precise calculation of angular variation. [I think Line 270 to 281 in pose_instability_detector.cpp might be helpful](https://github.com/autowarefoundation/autoware.universe/blob/main/localization/pose_instability_detector/src/pose_instability_detector.cpp#L270-L281). 

Yes, this also calculates the variation of yaw, but since yaw is estimated in a different way we can discard the yaw part after the calculation of angular variation.

TaikiYamada4 (commented):
@meliketanrikulu 
I wrote my comment about the calculation of angular variation below but how do you think?

localization/ekf_localizer/src/ekf_module.cpp
meliketanrikulu: Hello @TaikiYamada4 . Thanks for your review. You are right, it is a more accurate approach to calculate orientation as a whole instead of roll_rate and pitch_rate. I updated the code accordingly and tested it. https://github.com/autowarefoundation/autoware.universe/pull/8095/commits/49935649b5b91a5ccd1161ef5ab85742a1d2a9e4
Could you check again?

meliketanrikulu:
> @meliketanrikulu Overall, it looks good to me, but if the twist msg is interrupted, the values of roll_rate_ and pitch_rate_ will be retained. Please implement a measure to initialize them to 0 if there are no updates to the twist msg.

Hello @YamatoAndo Thanks for your review. I updated the code . Now I am using last_angular_velocity_ to retain all components of angular velocity. 
I initialize the last_angular_velocity_ variable here as (0,0,0) [Here](https://github.com/meliketanrikulu/autoware.universe/blob/03309e236ca9acbdaecdbb626375262e5728b3f1/localization/ekf_localizer/src/ekf_module.cpp#L40). If the twist message is not subscribed to, the values ​​will be accepted as 0, so the result will not change.

localization/ekf_localizer/src/ekf_module.cpp
TaikiYamada4: @meliketanrikulu Thank you for the revision! I also checked it seems to work in my environment too :+1:

TaikiYamada4 (approved):
LGTM!

YamatoAndo:
@meliketanrikulu I am concerned about the case when the twist msg is interrupted during operation.
Please make sure it does not retain the previous values.

TaikiYamada4:
> > @meliketanrikulu I am concerned about the case when the twist msg is interrupted during operation. Please make sure it does not retain the previous values.
> 
> Hello @YamatoAndo . I tested without twist msg. [Here](https://github.com/meliketanrikulu/autoware.universe/blob/03309e236ca9acbdaecdbb626375262e5728b3f1/localization/ekf_localizer/src/ekf_module.cpp#L302) delta_orientation comes as (0,0,0,1) so comes out `curr_orientation = prew_orientation` as a result of this line . It actually works the same way as before the delay compensation.

He means that the `last_angular_velocity_` will be kept forever even when the twist measurement is ceased.
Then, the ekf_localizer will keep calculation with that outdated twist, so its better to reset the `last_angular_velocity_` to zero after the compensation. 

Plus, I feel it is not appropriate to substitute `last_angular_velocity_` in the `measurement_update_twist` function since the process and the function name doesn't match. So how about
- Adding a reset line of `last_angular_velocity_` at the bottom of `compensate_rph_with_delay`
- Move the following line to after [this line](https://github.com/autowarefoundation/autoware.universe/blob/1ddbee81490f8b35ed5e159f011c7dc79fad6091/localization/ekf_localizer/src/ekf_localizer.cpp#L226) 
```
last_angular_velocity_ = tf2::Vector3(
    twist.twist.twist.angular.x, twist.twist.twist.angular.y, twist.twist.twist.angular.z);
```

localization/ekf_localizer/src/ekf_module.cpp
```diff
@@ -282,15 +283,37 @@ bool EKFModule::measurement_update_pose(
...
+  } else {
+    delta_orientation.setValue(0.0, 0.0, 0.0, 1.0);
+  }
+
+  tf2::Quaternion prew_orientation = tf2::Quaternion(
+    pose.pose.pose.orientation.x, pose.pose.pose.orientation.y, pose.pose.pose.orientation.z,
+    pose.pose.pose.orientation.w);
+
+  tf2::Quaternion curr_orientation;
+  curr_orientation = prew_orientation * delta_orientation;
```
TaikiYamada4, line 302: Is this a typo of `prev`?

localization/ekf_localizer/src/ekf_module.cpp
```diff
@@ -282,15 +283,37 @@ bool EKFModule::measurement_update_pose(
...
+  curr_orientation.normalize();
+
   const auto rpy = autoware::universe_utils::getRPY(pose.pose.pose.orientation);
-  const double dz_delay = kalman_filter_.getXelement(IDX::VX) * delay_time * std::sin(-rpy.y);
-  PoseWithCovariance pose_with_z_delay;
-  pose_with_z_delay = pose;
-  pose_with_z_delay.pose.pose.position.z += dz_delay;
-  return pose_with_z_delay;
+  const double delta_z = kalman_filter_.getXelement(IDX::VX) * delay_time * std::sin(-rpy.y);
```
TaikiYamada4, line 306: Isn't this `rpy.y` the one not compensated? Maybe the rpy should be derived from `curr_orientation`.

TaikiYamada4 (commented):
@meliketanrikulu 
Sorry for leaving a comment after approving, but could you check these out too?

localization/ekf_localizer/src/ekf_module.cpp
meliketanrikulu: Sorry for missing this. I fixed with this commit --> https://github.com/autowarefoundation/autoware.universe/pull/8095/commits/33989271a435a9856c46c36fbb6918433a9384cc

localization/ekf_localizer/src/ekf_module.cpp
meliketanrikulu: I didn't see any change in the test results. But theoretically I think it would be more accurate as you said. That's why I updated it. Thank you for your suggestion. @TaikiYamada4 
Related commit --> https://github.com/autowarefoundation/autoware.universe/pull/8095/commits/26f46c4ac9f5422b3adb6e326e281737e4ccd166

meliketanrikulu:
I updated pose_with_delay header stamp with adding delay_time.
Related commit  --> https://github.com/autowarefoundation/autoware.universe/pull/8095/commits/622c0769a2e29fd61b69c1a08d8d5251807fdd1c
Could you check this changes too.

localization/ekf_localizer/src/ekf_module.cpp
```diff
@@ -282,15 +282,39 @@ bool EKFModule::measurement_update_pose(
   return true;
 }
 
-geometry_msgs::msg::PoseWithCovarianceStamped EKFModule::compensate_pose_with_z_delay(
-  const PoseWithCovariance & pose, const double delay_time)
+geometry_msgs::msg::PoseWithCovarianceStamped EKFModule::compensate_rph_with_delay(
+  const PoseWithCovariance & pose, tf2::Vector3 last_angular_velocity_, const double delay_time)
```
YamatoAndo, line 286:
```suggestion
  const PoseWithCovariance & pose, tf2::Vector3 last_angular_velocity, const double delay_time)
```

meliketanrikulu: Updated with --> https://github.com/autowarefoundation/autoware.universe/pull/8095/commits/e33ab613afb6332d5fc633dcb35c4d75f1c34fbc

meliketanrikulu:
@TaikiYamada4 @YamatoAndo Could you review again. Its ready for the review.

YamatoAndo:
@meliketanrikulu 
Thank you for the revision. The source code looks fine.

I have two questions.
In section "How was this PR tested?", it says that "pitch_filter_proc_dev" was set to 0.5.
But is the "pitch_filter_proc_dev" of the Test Results "Without Delay Compensation" remained at 0.01?
Also, is there a reason why "roll_filter_proc_dev" was not changed?
> 2.) I updated 1D Filter params for pitch estimation.
> Changed [pitch_filter_proc_dev](https://github.com/autowarefoundation/autoware_launch/blob/244285f5f7999f4390c2f65f36a112c233b61143/autoware_launch/config/localization/ekf_localizer.param.yaml#L34) param as 0.5

meliketanrikulu:
> @meliketanrikulu 
> Thank you for the revision. The source code looks fine.
> 
> I have two questions.
> In section "How was this PR tested?", it says that "pitch_filter_proc_dev" was set to 0.5.
> But is the "pitch_filter_proc_dev" of the Test Results "Without Delay Compensation" remained at 0.01?
> Also, is there a reason why "roll_filter_proc_dev" was not changed?
> > 2.) I updated 1D Filter params for pitch estimation.
> > Changed [pitch_filter_proc_dev](https://github.com/autowarefoundation/autoware_launch/blob/244285f5f7999f4390c2f65f36a112c233b61143/autoware_launch/config/localization/ekf_localizer.param.yaml#L34) param as 0.5
> Hello @YamatoAndo . Thanks for your review and comments
1.) Both With/Without Delay Compensation tests were performed with a pitch_filter_proc_dev value of 0.5. because I got the best results by adjusting this parameter.
2.) I did not make any changes to the roll_filter_proc_dev parameter because I did not observe that changing it had a positive effect on the results. roll_filter_proc_dev parameter was 0.01 for all tests.

YamatoAndo:
@meliketanrikulu Thank you for the response.
It seems we need to either find an appropriate process noise, change the estimated covariance value of NDT, or[ change the algorithm to use the estimated variance value of NDT as is](https://github.com/autowarefoundation/autoware.universe/blob/3e79b8f67f5a3428c04efb7545c57b6401c83d17/localization/ekf_localizer/src/ekf_localizer.cpp#L456-L459). I would like to consider this separately from this PR.

YamatoAndo (approved):

meliketanrikulu:
> @meliketanrikulu Thank you for the response. It seems we need to either find an appropriate process noise, change the estimated covariance value of NDT, or[ change the algorithm to use the estimated variance value of NDT as is](https://github.com/autowarefoundation/autoware.universe/blob/3e79b8f67f5a3428c04efb7545c57b6401c83d17/localization/ekf_localizer/src/ekf_localizer.cpp#L456-L459). I would like to consider this separately from this PR.

Thanks for the review and support @YamatoAndo . As it is, I have merged this PR and am closing the issue.  Based on your suggestion, I will check the NDT covariance values.

PR #8100, opened 2024-07-19
fix(bytetrack): no need to throw exception when idx is out of range

chihungtzeng:
## Description

Since ColorMapper just returns a unique color to the caller, we simply use modulo to get a valid color index.
In such a way there is no need to check the index validity and throw a runtime exception.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


knzo25:
@chihungtzeng 
I don't think you meant to edit so many files based on the title if your PR. Could you rebase and solve conflicts?

chihungtzeng:
It is a mistake while I merge the files from the main branch. I will resolve it.

maxime-clem:
Please rebase your branch to the latest `main` and only keep your own commit (I think only this one is needed: https://github.com/autowarefoundation/autoware.universe/pull/8100/commits/089dbb3e29c1a2bb1d6c6e6c71224af13cc7fa44).

chihungtzeng:
Github thinks that the src branch adds files belonging to other commits. I should create another PR.

PR #8110, opened 2024-07-19
refactor(ndt_scan_matcher): remove unnecessary code and fix default covariances

SakodaShintaro:
## Description

(1) remove unnecessary code

Covariance estimation is now done in ndt_omp, so it is no longer necessary to calculate covariance rotations in autoware.universe, so remove that code.

(2) fix default covariances

The value 0.0225 was hard-coded, but I think this should be the covariance value specified in the parameters, so I changed it to that.

## How was this PR tested?

I have confirmed that logging_simulator works well.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


KYabuuchi (approved):
LGTM :ok_man:

PR #8120, opened 2024-07-22
fix(autoware_multi_object_tracker): missing parameter schema path fix

technolojin:
## Description
Fix missing transition to new package path

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


YoshiRi (approved):
LGTM

PR #3901, opened 2023-06-05
refactor(mpc_lateral_controller): refactor code, no change behavior.

TakaHoribe:
## Description

Refactor code to improve readability.

## Related links

None

## Tests performed

Tested on Psim and ScenarioSimulator.

For the degradation check, scenario test performed by TIERIV 
- latest main: (1324/1372) https://evaluation.tier4.jp/evaluation/reports/230bdcac-22ff-5f61-957e-14d46175a5c4?project_id=prd_jt
- latest main with this PR: (1322/1372) https://evaluation.tier4.jp/evaluation/reports/d2ed1e5f-adb3-5412-a28e-30f2f7241e1a?project_id=prd_jt

## Interface changes

None

## Effects on system behavior

None

## Pre-review checklist for the PR author

The PR author **must** check the checkboxes below when creating the PR.

- [x] I've confirmed the [contribution guidelines].
- [x] The PR follows the [pull request guidelines].

## In-review checklist for the PR reviewers

The PR reviewers **must** check the checkboxes below before approval.

- [x] The PR follows the [pull request guidelines].
- [x] The PR has been properly tested.
- [x] The PR has been reviewed by the code owners.

## Post-review checklist for the PR author

The PR author **must** check the checkboxes below before merging.

- [x] There are no open discussions or they are tracked via tickets.
- [x] The PR is ready for merge.

After all checkboxes are checked, anyone who has write access can merge the PR.

[contribution guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/
[pull request guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/pull-request-guidelines/

PR #8141, opened 2024-07-22
fix(autoware_mpc_lateral_controller): remove double assignment in MpcLateralController

wakabame:
## Description

In the initialization process of class `MpcLateralController`, there is a section where the vehicle type is set.
However, the idempotent setter `setVehicleModel()` is called twice in succession:
https://github.com/autowarefoundation/autoware.universe/blob/093c20cdecfd63e3db798608a0bbd114ab10905d/control/autoware_mpc_lateral_controller/src/mpc_lateral_controller.cpp#L104-L107
This is clearly unnecessary and has been removed. 

FYI: This issue originated in commit a4d607151dd3212f73998518727d4dc36f62cef6 of https://github.com/autowarefoundation/autoware.universe/pull/3901/

## Related links

**Parent Issue:**

None

## How was this PR tested?

None

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


mitsudome-r:
This was already fixed by another PR.

PR #8164, opened 2024-07-23
fix(autoware_mpc_lateral_controller): add timestamp and frame ID to published trajectory

kyoichi-sugahara:
## Description

In this [PR](https://github.com/autowarefoundation/autoware.universe/pull/5576), predicted path calculated using the Frenet coordinate system were added for debug purpose, but they couldn't be visualized in rviz due to missing header information. 
added the necessary header information with following modification to resolve this issue.

```
    frenet_clipped.header.stamp = m_clock->now();
    frenet_clipped.header.frame_id = "map";
```

## How was this PR tested?
tested in psim and confirmed visualization
Screenshot from 2024-07-23 20-28-54

## Effects on system behavior

None.


TakaHoribe (approved):
Lgtm

PR #8169, opened 2024-07-23
fix(simple_planning_simulator): fix publised acc of actuation simulator

kosuke55:
## Description

fix acceleration of actuation simulator.

Currently, the derivative of acceleration was calculated from the acceleration command only. This can result in inappropriate values (e.g., 0) for the published acceleration topic. Calculate the derivative of acceleration based on the current acceleration state and indication as in other models.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?
- unit test 
- psim

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


TakaHoribe (approved):
The old code does not work as expected for the acceleration topic calculation.

PR #8172, opened 2024-07-24
fix(lidar_apollo_instance_segmentation): fixed preprocessing

knzo25:
During the point cloud type transition, I performed a manual conversion to cast intensities into floats.
However, there was a transformation step that is not compatible with the new type so addressed those issues in this PR

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_lidar_apollo_instance_segmentation/src/detector.cpp
+++ b/perception/autoware_lidar_apollo_instance_segmentation/src/detector.cpp
@@ -17,9 +17,12 @@
 #include "autoware/lidar_apollo_instance_segmentation/feature_map.hpp"
 
 #include <sensor_msgs/point_cloud2_iterator.hpp>
+#include <tf2_sensor_msgs/tf2_sensor_msgs.hpp>
 
 #include <NvInfer.h>
-#include <pcl_conversions/pcl_conversions.h>
+#include <pcl/PointIndices.h>
+#include <pcl/point_cloud.h>
+#include <pcl/point_types.h>
 
 #include <functional>
 #include <memory>
@@ -83,7 +86,6 @@ bool LidarApolloInstanceSegmentation::transformCloud(
   const sensor_msgs::msg::PointCloud2 & input, sensor_msgs::msg::PointCloud2 & transformed_cloud,
   float z_offset)
 {
-  // TODO(mitsudome-r): remove conversion once pcl_ros transform are available.
   // transform pointcloud to target_frame
   if (target_frame_ != input.header.frame_id) {
     try {
@@ -119,7 +121,6 @@ bool LidarApolloInstanceSegmentation::detectDynamicObjects(
 
   // convert from ros to pcl
   pcl::PointCloud<pcl::PointXYZI>::Ptr pcl_pointcloud_raw_ptr(new pcl::PointCloud<pcl::PointXYZI>);
-  // pcl::fromROSMsg(transformed_cloud, *pcl_pointcloud_raw_ptr);
 
   auto & pcl_pointcloud_raw = *pcl_pointcloud_raw_ptr;
   pcl_pointcloud_raw.width = transformed_cloud.width;


yukkysaito (commented):
I think the current implementation was due to the slow execution time of doTransform. It may have been fixed in the latest version, but could you check if there has been any change in the execution time?
https://qiita.com/yakato_jun/items/091f36308b0662110537

knzo25:
@yukkysaito 
The problem is that not all pointcouds can use the `pcl_ros` one (internally they use the data field which is only available in point types with the data field)

yukkysaito:
@knzo25 Understood. Could you measure the execution time to ensure there are no issues?

By the way, can we use this function instead?
https://github.com/autowarefoundation/autoware.universe/blob/main/common/autoware_universe_utils/include/autoware/universe_utils/transform/transforms.hpp#L27-L49

mitsudome-r:
@knzo25 @yukkysaito Do you still need this PR? If not, please close the PR.

perception/autoware_lidar_apollo_instance_segmentation/src/detector.cpp
```diff
@@ -134,7 +125,7 @@ bool LidarApolloInstanceSegmentation::detectDynamicObjects(
   pcl_pointcloud_raw.width = transformed_cloud.width;
   pcl_pointcloud_raw.height = transformed_cloud.height;
   pcl_pointcloud_raw.is_dense = transformed_cloud.is_dense == 1;
-  pcl_pointcloud_raw.resize(transformed_cloud.width * transformed_cloud.height);
+  pcl_pointcloud_raw.reserve(transformed_cloud.width * transformed_cloud.height);
```
xmfcx, line 129: This looks like a serious bug fix.

Let's say you have 5000 points in the cloud.
**With resize, it will create 5000 points at origin.**
Then with emplace_back down below, it will add 5000 more points that are transformed.

This would basically double the point cloud size.

xmfcx:
> By the way, can we use this function instead?
> https://github.com/autowarefoundation/autoware.universe/blob/main/common/autoware_universe_utils/include/autoware/universe_utils/transform/transforms.hpp#L27-L49

@yukkysaito -san, the whole point in this PR is to avoid back-and-forth pcl type conversion, the existing code already uses that.

I will now make a small benchmark to compare both options.

xmfcx:
https://github.com/ros2/geometry2/blob/78ef2e48c31f3a38d67e8198e28cb9596d58c2df/tf2_sensor_msgs/include/tf2_sensor_msgs/tf2_sensor_msgs.hpp#L77-L150

At least, it doesn't look like something that would be slow.

It's as basic as:
```cpp
  Eigen::Transform<float, 3, Eigen::Affine> t = translation * quaternion;
  Eigen::Transform<float, 3, Eigen::Affine> r = Eigen::Translation3f(0, 0, 0) * quaternion;

  sensor_msgs::PointCloud2ConstIterator<float> x_in(p_in, std::string("x"));
  sensor_msgs::PointCloud2ConstIterator<float> y_in(p_in, std::string("y"));
  sensor_msgs::PointCloud2ConstIterator<float> z_in(p_in, std::string("z"));

  sensor_msgs::PointCloud2Iterator<float> x_out(p_out, std::string("x"));
  sensor_msgs::PointCloud2Iterator<float> y_out(p_out, std::string("y"));
  sensor_msgs::PointCloud2Iterator<float> z_out(p_out, std::string("z"));

  for (; x_in != x_in.end(); ++x_in, ++y_in, ++z_in, ++x_out, ++y_out, ++z_out) {
    point = t * Eigen::Vector3f(*x_in, *y_in, *z_in);
    *x_out = point.x();
    *y_out = point.y();
    *z_out = point.z();
  }
```

xmfcx:
@yukkysaito I've benchmarked all 3 options.

### Bench setup

**Inputs:**
- 200.000 points (like vls128) point cloud, random points in `sensor_msgs::msg::PointCloud2` format.
- Random transformation matrix in `geometry_msgs::msg::TransformStamped` format.

**Task:**
- Transform and return in the input format

**Competing methods:**
1. **PCL:** This is the method that is being used in this code right now.
   - Converts from `sensor_msgs` to `pcl`, transform, converts back to `sensor_msgs`.
2. **TF2:** This is the proposed method in this PR.
   - [Directly transform without unnecessary middle steps.](https://github.com/ros2/geometry2/blob/78ef2e48c31f3a38d67e8198e28cb9596d58c2df/tf2_sensor_msgs/include/tf2_sensor_msgs/tf2_sensor_msgs.hpp#L77-L150)
3. **PCL_ROS:** The proposed method in [your link](https://qiita.com/yakato_jun/items/091f36308b0662110537).
   - [Directly transforms the point cloud](https://github.com/ros-perception/perception_pcl/blob/67a5c2ba4c4de3ca21c5cd495812a01ced3fb69a/pcl_ros/src/transforms.cpp#L127-L227) too but does many other things too, ends up being 2 times slower in the benchmark.

### Please check or replicate the benchmark

💎 https://github.com/xmfcx/pointcloud_transform_benchmark_ws

### Results across 50 runs

image

image

xmfcx:
@YoshiRi @yukkysaito @technolojin please review soon, this has been waiting for 1.5 years, I've rebased and updated it.
It also fixes [a critical bug](https://github.com/autowarefoundation/autoware_universe/pull/8172#discussion_r2647173427).

PR #8174, opened 2024-07-24
perf(autoware_pointcloud_preprocessor): lazy & managed TF listeners

amadeuszsz:
## Description
This PR modifies approach for TF listeners within `autoware_pointcloud_preprocessor` filters.

The standard concept considers initialize listener once and look for transformation via `lookupTransform` API whenever is needed.

This PR concept initializes listener for each transformation request. For first call it will use `lookupTransform` API and place transformation to local buffer (class member). Next requests will return transformation which already exists in local buffer - listener is not initialized. Improvement bases on one assumption - transformations are static, therefore modification is not applied for vector map filter. In result, for base sensing pipeline from `sample_sensor_kit_launch`, the number of listeners decreased from 17 to 0. Tested with Intel i7-13700H and 100 % rosbag data rate. Attached graphs show improvement evidence:
* CPU load has decreased.

result_compared

* Average time delay for point cloud processing not affected (negligible improvement).

delays_compared

## Related links

**Parent Issue:**

## How was this PR tested?
* Covered by unit test.
* Sensing pipeline from `sample_sensor_kit_launch` with rosbag ([TIER IV internal link](https://console.data-search.tier4.jp/projects/prd_jt/environments/64d8fdae-7e76-402f-83dd-7136d22e5531/rosbags/cd413e5b-895c-417c-aae4-e908c066503d))

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/static_transform_buffer.hpp
```diff
@@ -0,0 +1,137 @@
...
+    if (
+      pcl::getFieldIndex(cloud_in, "x") == -1 || pcl::getFieldIndex(cloud_in, "y") == -1 ||
+      pcl::getFieldIndex(cloud_in, "z") == -1) {
+      RCLCPP_ERROR(node->get_logger(), "Input pointcloud does not have xyz fields");
+      return false;
+    }
+    Eigen::Matrix4f eigen_transform;
+    if (!get_transform(node, target_frame, cloud_in.header.frame_id, eigen_transform)) {
+      return false;
+    }
```
yukkysaito, line 125: The frame_id coming into the input data may not necessarily differ from the target frame. Therefore, how about including processing that avoids transformation when the input frame and target frame are the same?

amadeuszsz: Please, check `static_tf_buffer_->transform_pointcloud` occurrences within `filter.cpp`. There are conditions which ignore transform call if frames are equal. I didn't change this behavior and I left this as it was. Do you think it will be better to move this logic into StaticTransformBuffer class?

yukkysaito: @amadeuszsz Thank you for your response.
I understand. So this is intended to be used in filter.cpp.

I am just making a note for the future and not making a request.
I think StaticTransformBuffer would be very useful not only for the pointcloud preprocessor but also for Perception/Localization, so it might be a good idea to move it to a common utility in the future.

amadeuszsz:
@yukkysaito 
Sorry, looks like I didn't applied this feature for dependent filter out of `autoware_pointcloud_preprocessor`. Let me switch it for a draft for this time before I finish all filters.

yukkysaito (approved):

amadeuszsz:
@yukkysaito
Only RANSAC ground filter from `autoware_ground_segmentation` uses TF buffer previously defined in parent filter class. The fix I addressed in https://github.com/autowarefoundation/autoware.universe/pull/8174/commits/92bdc998823f4e3970ce14c9a0596a2c40e154f1.
However, I think we can assume static transformation here as well, therefore this feature I addressed in https://github.com/autowarefoundation/autoware.universe/pull/8174/commits/cc4ae04917698526563d0a6d0879e5e0c39bfad2. If it is wrong assumption, we can revert this commit.

yukkysaito:
@amadeuszsz 
If it is to be used in other packages, it would be better to include it in one of the following packages:

- https://github.com/autowarefoundation/autoware.universe/tree/main/common/autoware_universe_utils
- https://github.com/autowarefoundation/autoware_utils/tree/main/autoware_utils

amadeuszsz:
> @amadeuszsz If it is to be used in other packages, it would be better to include it in one of the following packages:
> 
> * https://github.com/autowarefoundation/autoware.universe/tree/main/common/autoware_universe_utils
> * https://github.com/autowarefoundation/autoware_utils/tree/main/autoware_utils

@yukkysaito 
Addressed in https://github.com/autowarefoundation/autoware.universe/pull/8174/commits/96d143c9a627e3bffc399e7d595ae77fc8a39f2f.

yukkysaito (commented):
LGTM for autoware_ground_segmentation package and autoware_universe_utils

yukkysaito:
@YoshiRi cc @badai-nguyen 
Could you please review the sensing-related packages within the team? It might be a good idea to thoroughly review the distortion corrector to ensure there are no changes in its logic.

drwnz:
@yukkysaito @YoshiRi I think @vividf can also double check the logic since he has recently been refactoring and adding features to distortion corrector.

vividf (approved):
LGTM for concatenate, distortion corrector, and filter node!

sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/static_transform_buffer.hpp
amadeuszsz: @yukkysaito 
Sorry, i just read this comment. If we intend to use this implementation out of pointcloud preprocessor, I agree we can skip unnecessary transforms. It will change nothing for pointcloud preprocessor now, but in future we don't need to wrap all transform call around `frame_id` check.
Addressed in https://github.com/autowarefoundation/autoware.universe/pull/8174/commits/d31481c4f32485a6e5a7a62a7eed715fa3df8d1a.

badai-nguyen (approved):
LGTM regarding ground_segmentation and distortion_corrector.

common/autoware_universe_utils/include/autoware/universe_utils/ros/static_transform_buffer.hpp
```diff
@@ -0,0 +1,141 @@
...
+      buffer_[key] = eigen_transform;
+      return true;
+    }
+
+    // Check if transform is needed
+    if (target_frame == source_frame) {
+      eigen_transform = Eigen::Matrix4f::Identity();
+      buffer_[key] = eigen_transform;
+      return true;
+    }
```
TakaHoribe, line 94: Wouldn't it be more efficient to have this code at the beginning of the function?

common/autoware_universe_utils/include/autoware/universe_utils/ros/static_transform_buffer.hpp
```diff
@@ -0,0 +1,141 @@
...
+
+    // Get the transform from the TF tree
+    auto tf_listener = std::make_shared<autoware::universe_utils::TransformListener>(node);
+    auto tf = tf_listener->getTransform(
+      target_frame, source_frame, rclcpp::Time(0), rclcpp::Duration(1000ms));
+    RCLCPP_DEBUG(
+      node->get_logger(), "Trying to enqueue %s -> %s transform to static TFs buffer...",
+      target_frame.c_str(), source_frame.c_str());
+    if (tf == nullptr) {
+      eigen_transform = Eigen::Matrix4f::Identity();
```
soblin, line 104: returning false and setting the return value `eigen_transform = Eigen::Matrix4f::Identity();` at the same time is a bit weird.  If you want to modify existing lvalue Matrix like a huge member variable, then it is OK.

Otherwise is it possible to return optionla<Matrix> instead ?

TakaHoribe (commented):
@amadeuszsz  I think it’s good to have this function in autoware_util.

@amadeuszsz @knzo25 
For my confirmation, does the designer need to decide if the TF is static or dynamic? For example, will it work correctly if we change a static TF dynamically during calibration?

I'm not confident, but for future scalability and error monitoring, it might be better to re-read TF every few seconds and raise an error if the value changes. Please consider this for future extensions. (It does not have to be in this PR though)

common/autoware_universe_utils/include/autoware/universe_utils/ros/static_transform_buffer.hpp
amadeuszsz: Thank you @soblin for your review.
Using optional was my first thought, but from user point of view is more convenient to already have such a transformation in matrix. Less code need on user side and it is still safe since it returns success status.

soblin: All right. I think it is confusing that returning `false` status while causing the mutable side effect `eigen_transform = Eigen::Matrix4f::Identity();` at the same time. Can you note that as a doxygen docstring ?

common/autoware_universe_utils/include/autoware/universe_utils/ros/static_transform_buffer.hpp
amadeuszsz: @TakaHoribe thank you for your review.
We loosing negligible time here on first request, but in next request we will have transformation in our buffer already (first condition). It is about nanoseconds. I would rather to have all history of TF requests in buffer.
* Regarding autoware_utils, we already have `autoware:universe_utils:TransformListener` and this PR brings similar functionalities, thus I put it alongside.
* We discussed dynamic calibration internally and came up with idea of adding service server for such a feature. Request will force TFs update. What do you think about it? We can also investigate custom mechanics for TF listener: default listener under the hood is just two callbacks to /tf and /tf_static. We could check where is transformation and perform suitable operations: for static TFs same as this PR, for dynamic TFs listen as default. It is just idea at it will need deeper investigation for possible drawbacks etc. Applying this for all the Autoware might free some CPU load.

common/autoware_universe_utils/include/autoware/universe_utils/ros/static_transform_buffer.hpp
amadeuszsz: Good idead! Addressed in https://github.com/autowarefoundation/autoware.universe/pull/8174/commits/11c61beebc03a8671aedd16e3e069c0897b3f689.

soblin (approved):
LGTM

common/autoware_universe_utils/include/autoware/universe_utils/ros/static_transform_buffer.hpp
TakaHoribe: Thank you for your reply. We need to investigate the easy way for dynamic calibration, but so far the approach in this PR and your comment are reasonable. 

> Request will force TFs update. What do you think about it?
You mean, when this instance is created, a server automatically starts, and if dynamic calibration is being performed, the executing application will send a request to that service, triggering a force update? I think this is one of a good idea :+1:

TakaHoribe (approved):

common/autoware_universe_utils/include/autoware/universe_utils/ros/static_transform_buffer.hpp
```diff
@@ -0,0 +1,165 @@
...
+   * Listener is destroyed after function call.
+   *
+   * @param node A pointer to the ROS node.
+   * @param target_frame The target frame.
+   * @param source_frame The source frame.
+   * @param eigen_transform The output Eigen transform matrix. Is set to identity if the transform
+   * is not found.
+   * @return True if the transform was successfully retrieved, false otherwise.
+   */
+  bool getTransform(
```
knzo25, line 82: @amadeuszsz 
For the next time, unless we are worrying about heavy algorithms, instead of modifying a reference and returning a boolean, std::optional could be a good alternative

perception/autoware_ground_segmentation/src/ransac_ground_filter/node.hpp
```diff
@@ -80,6 +82,7 @@ class RANSACGroundFilterComponent : public autoware::pointcloud_preprocessor::Fi
   bool debug_ = false;
   bool is_initialized_debug_message_ = false;
   Eigen::Vector3d unit_vec_ = Eigen::Vector3d::UnitZ();
+  std::shared_ptr<autoware::universe_utils::StaticTransformBuffer> static_tf_buffer_{nullptr};
```
knzo25, line 85: unique?

perception/autoware_ground_segmentation/src/ransac_ground_filter/node.hpp
```diff
@@ -16,6 +16,7 @@
 #define RANSAC_GROUND_FILTER__NODE_HPP_
 
 #include "autoware/pointcloud_preprocessor/filter.hpp"
+#include "autoware/universe_utils/ros/static_transform_buffer.hpp"
```
knzo25, line 19: I got confused with headers under the new namespaces

https://autowarefoundation.github.io/autoware-documentation/main/contributing/coding-guidelines/languages/cpp/

Should `universe_utils` use `<>` or `""` ?

sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/concatenate_data/concatenate_pointclouds.hpp
```diff
@@ -139,9 +140,7 @@ class PointCloudConcatenationComponent : public rclcpp::Node
   // XmlRpc::XmlRpcValue input_topics_;
   std::vector<std::string> input_topics_;
 
-  /** \brief TF listener object. */
-  std::shared_ptr<tf2_ros::Buffer> tf2_buffer_;
-  std::shared_ptr<tf2_ros::TransformListener> tf2_listener_;
+  std::shared_ptr<autoware::universe_utils::StaticTransformBuffer> static_tf_buffer_{nullptr};
```
knzo25, line 143: unique?

knzo25: There seem to be more occurrences of this in case you accept the suggestion

sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/distortion_corrector/distortion_corrector.hpp
```diff
@@ -95,11 +97,12 @@ class DistortionCorrector : public DistortionCorrectorBase
     static_cast<T *>(this)->undistortPointImplementation(
       it_x, it_y, it_z, it_twist, it_imu, time_offset, is_twist_valid, is_imu_valid);
   };
+  void convertMatrixToTransform(const Eigen::Matrix4f & matrix, tf2::Transform & transform);
```
knzo25, line 100: Is using a void return type necessary? (in this node there are some optimizations, that get turned off is we return things, but I am not sure about this one)

sensing/autoware_pointcloud_preprocessor/src/filter.cpp
```diff
@@ -57,6 +57,7 @@
 
 #include <pcl/io/io.h>
 
+#include <cstddef>
```
knzo25, line 60: Just a question, but why is this header needed now?
https://cpprefjp.github.io/reference/cstddef.html

common/autoware_universe_utils/include/autoware/universe_utils/ros/static_transform_buffer.hpp
amadeuszsz: Filter class already follow such an approach ([modifying a reference](https://github.com/autowarefoundation/autoware.universe/blob/d4f365da2f2e6403381ce5c64e60d3820428128a/sensing/autoware_pointcloud_preprocessor/src/filter.cpp#L307)), so I didn't find any reason to change this. If there are performance, principle or readability reason we can address this point for whole package during code refactor.

sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/distortion_corrector/distortion_corrector.hpp
amadeuszsz: I think I seen how tf2_eigen do this and across libraries it's 50/50.

common/autoware_universe_utils/include/autoware/universe_utils/ros/static_transform_buffer.hpp
knzo25: Thanks for the clarification !

sensing/autoware_pointcloud_preprocessor/src/filter.cpp
amadeuszsz: Probably IDE equipped with all the boosters did it automatically while doing testing code and I didn't remove it 😢

perception/autoware_ground_segmentation/src/ransac_ground_filter/node.hpp
amadeuszsz: You can check compiler search part via `g++ -v -E -x c++ - < /dev/null`. My output (in docker image) doesn't consider any of Autoware package to be placed in `<>`.

sensing/autoware_pointcloud_preprocessor/src/filter.cpp
amadeuszsz: Addressed in https://github.com/autowarefoundation/autoware.universe/pull/8174/commits/26d125b158c3594d79c646f1611e679497875e0c

sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/concatenate_data/concatenate_pointclouds.hpp
amadeuszsz: https://github.com/autowarefoundation/autoware.universe/pull/8174/commits/c55991a5193aaaa465bb5fd45e087d01136ba8ca

perception/autoware_ground_segmentation/src/ransac_ground_filter/node.hpp
amadeuszsz: https://github.com/autowarefoundation/autoware.universe/pull/8174/commits/c55991a5193aaaa465bb5fd45e087d01136ba8ca

perception/autoware_ground_segmentation/src/ransac_ground_filter/node.hpp
amadeuszsz: Let's follow Autoware guidelines
Addressed in https://github.com/autowarefoundation/autoware.universe/pull/8174/commits/69eb78a52e15dae322ce417b4d85ddbfed9a8b38

mitsudome-r:
This PR should be merged after https://github.com/autowarefoundation/autoware.universe/pull/8337. (Otherwise, the CI won't pass).

Issue #8128, opened 2024-07-22
Reduce dependence on Boost.Geometry


mitukou1109:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I've agreed with the maintainers that I can plan this task.

### Description

Currently, Autoware relies on Boost.Geometry for most geometric calculations. However, Boost's generic implementation is considered to extend the build time.
The goal of this task is to reduce the build time by minimizing (or eliminating) dependency on Boost.Geometry.

### Purpose

This task aims to shorten the build time by reducing dependency on Boost.Geometry. This involves adding utility functions that compile independently of Boost and omitting unnecessary features for Autoware to improve function execution efficiency.

### Possible approaches

The following functions are planned to be replaced:
- [ ] `area()` @mitukou1109 
- [ ] `convex_hull()` @mitukou1109 
- [ ] `covered_by()` @mitukou1109 
- [ ] `disjoint()` @mitukou1109 
- [ ] `distance()` @mitukou1109 
- [ ] `intersects()` @mitukou1109 
- [ ] `within()` @mitukou1109 
- [ ] `difference()` @mraditya01 
- [ ] `union_()` @mraditya01 
- [ ] `envelope()` @mitukou1109 
- [ ] `simplify()` @mitukou1109 
- [ ] `buffer()` @mitukou1109

### Definition of done

- [ ] All replaceable Boost.Geometry functions are migrated to self-implementations.
- [ ] Build and processing time are reduced compared to Boost.Geometry.

mitukou1109:
Here is the current benchmark results with 500 randomly generated polygons with 9 vertices:

|function|Boost|self-impl.|task|
|----|----|----|----|
|`area()`|0.18ms|0.65ms|calculate the area of 500 polygons|
|`convex_hull()`|1.93ms|4.59ms|calculate the convex hull of vertices of 500 polygons|
|`covered_by()`|278.20ms|946.09ms|check if each vertex of 500 polygons is covered by (= inside or on border of) another 500 polygons|
|`disjoint()`|195.58ms|180.81ms|check if 500 polygons are disjoint (runs totally 250,000 times)|
|`intersects()`|195.41ms|145.11ms|check if 500 polygons intersect (runs totally 250,000 times)|
[...]

PR #8186, opened 2024-07-25
feat(autoware_universe_utils): reduce dependence on Boost.Geometry

mitukou1109:
## Description

This PR is part of https://github.com/autowarefoundation/autoware.universe/pull/7778, and adds the following functions to reduce dependence on Boost.Geometry.

|Boost.Geometry <br> (`boost::geometry`)|alternative <br> (`autoware::universe_utils`)|supported types|
|----|----|----|
|`area()`|`area()`|polygon|
|`correct()`|`correct()`|polygon|
|`equals()`|`equals()`|point & point / polygon & polygon|
|`intersects()`|`intersects()`|segment & segment / polygon & polygon|
|`within()`|`within()`|point & polygon / polygon & polygon|

See the original PR for details and notes.

## Related links

**Parent Issue:**

- https://github.com/autowarefoundation/autoware.universe/issues/8128

## How was this PR tested?

```
colcon test --packages-select autoware_universe_utils
```

Benchmark result with 500 randomly generated polygons with 9 vertices:

|function|Boost|self-impl.|task|
|----|----|----|----|
|`area()`|0.04ms|0.04ms|calculate the area of 500 polygons|
|`intersects()`|231.43ms|70.57ms|check if 500 polygons intersect (runs totally 250,000 times)|
|`within()`|1058.18ms|58.60ms|check if each of 500 polygons is within 500 polygons|

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


satoshi-ota:
@mitukou1109 Thanks! But could you fix pre-commit.ci error?

common/autoware_universe_utils/include/autoware/universe_utils/geometry/geometry.hpp
```diff
@@ -583,6 +584,121 @@ std::optional<geometry_msgs::msg::Point> intersect(
...
+
+  double norm() const { return std::hypot(x_, y_); }
+
+  Vector2d vector_triple(const Vector2d & v1, const Vector2d & v2) const
+  {
+    const auto tmp = this->cross(v1);
+    return {-v2.y() * tmp, v2.x() * tmp};
+  }
+
+  const double & x() const { return x_; }
```
takayuki5168, line 611: Is reference (&) required here?

common/autoware_universe_utils/include/autoware/universe_utils/geometry/geometry.hpp
```diff
@@ -583,6 +584,121 @@ std::optional<geometry_msgs::msg::Point> intersect(
...
+
+  double & x() { return x_; }
```
takayuki5168, line 613: Is this reference (&) effective?
If it's not so effective, my opinion is removing it so that the user of this function will not assign a value to `x()` by mistake.

common/autoware_universe_utils/include/autoware/universe_utils/geometry/geometry.hpp
```diff
@@ -583,6 +584,121 @@ std::optional<geometry_msgs::msg::Point> intersect(
...
+  const double & y() const { return y_; }
+
+  double & y() { return y_; }
+
+private:
+  double x_;
+  double y_;
+};
+
+using Point2d = Vector2d;
```
takayuki5168, line 624: Why don't you name Point2d to the class directly?
Regarding Eigen, they use vectors and matrices, so Vector2d makes sense. But in this case, since we use the point, the Point2d class sounds better to me.

common/autoware_universe_utils/include/autoware/universe_utils/geometry/geometry.hpp
takayuki5168: Ah, in that case, we have to create another setter function.

takayuki5168: We have two options.
- option1.
    - x, y are private variables
    - create getter/setter functions
- option2.
    - x, y are public variables

This clas is very simple, so I feel it's fine and better to choose option2. What do you think?

common/autoware_universe_utils/include/autoware/universe_utils/geometry/geometry.hpp
mitukou1109: I thought it would be strange if a class representing a point would have functions of vector operations such as inner product and outer product as its members.

takayuki5168: Makes sense. Thank you for the explanation!
Please write it in this file so that other people understand the reason.

common/autoware_universe_utils/include/autoware/universe_utils/geometry/geometry.hpp
mitukou1109: Isn't it common for accessors to return references? I intend to use `double & x()` to assign values.

takayuki5168: My concern is as follows, but this may be my misunderstanding since `auto x` is not `auto & x` in this case.
```
auto x = vec.x();
x = 2.0; // vec.x() is overwritten.
```

Then, the thing is whether we need to make x and y private or not as we can access x, y, and z of geometry_msgs::msg::Point for example. I don't have a strong opinion here, so let's keep the current implementation. Thank you.

takayuki5168 (approved):

maxime-clem (approved):

common/autoware_universe_utils/include/autoware/universe_utils/geometry/geometry.hpp
```diff
@@ -583,6 +584,125 @@ std::optional<geometry_msgs::msg::Point> intersect(
...
+
+private:
+  double x_;
+  double y_;
+};
+
+// We use Vector2d to represent points, but we do not name the class Point2d directly
+// as it has some vector operation functions.
+using Point2d = Vector2d;
+using PointList = std::vector<Point2d>;
```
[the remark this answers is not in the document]
mitukou1109: How about `Point2dArray`?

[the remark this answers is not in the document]
maxime-clem: But it is not a `std::array`  :sweat_smile:  I think `Points` or `Points2d` is the best way to name this structure but if you don't like it please keep `PointList`.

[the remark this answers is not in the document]
mitukou1109: I changed it to `Points2d`: https://github.com/autowarefoundation/autoware.universe/pull/8186/commits/f216d4b4a9c68f4931a059bba395abe9a87e5650

PR #8190, opened 2024-07-25
fix(autoware_path_optimizer): fix passedByValue

kobayu858:
## Description

```
planning/autoware_path_optimizer/src/utils/trajectory_utils.cpp:139:38: performance: Function parameter 'traj_points' should be passed by const reference. [passedByValue]
  const std::vector<TrajectoryPoint> traj_points, const double interval)
                                     ^

planning/autoware_path_optimizer/src/utils/trajectory_utils.cpp:151:38: performance: Function parameter 'traj_points' should be passed by const reference. [passedByValue]
  const std::vector<TrajectoryPoint> traj_points, const double interval)
                                     ^

planning/autoware_path_optimizer/src/utils/trajectory_utils.cpp:162:37: performance: Function parameter 'ref_points' should be passed by const reference. [passedByValue]
  const std::vector<ReferencePoint> ref_points, const double interval)
                                    ^
```
## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/autoware_path_optimizer/include/autoware/path_optimizer/utils/trajectory_utils.hpp
+++ b/planning/autoware_path_optimizer/include/autoware/path_optimizer/utils/trajectory_utils.hpp
@@ -158,13 +158,13 @@ size_t findEgoSegmentIndex(
 }
 
 std::vector<TrajectoryPoint> resampleTrajectoryPoints(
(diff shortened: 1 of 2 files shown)


kosuke55 (approved):

takayuki5168 (approved):
Thank you!

PR #8192, opened 2024-07-25
fix(autoware_perception_rviz_plugin): fix passedByValue

bathteayo:
## Description

```
common/autoware_perception_rviz_plugin/src/object_detection/object_polygon_detail.cpp:462:21: performance: Function parameter 'label' should be passed by const reference. [passedByValue]
  const std::string label, const std_msgs::msg::ColorRGBA & color_rgba)
                    ^
```

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


veqcc (approved):

veqcc (commented):
.hpp file should also be modified?

kobayu858:
@veqcc 
Indeed, the declaration and definition must match the type.
That is why I fixed it.

However, it is strange because the CI build was passed.

YoshiRi (approved):
LGTM

PR #8193, opened 2024-07-25
feat(path_safety_checker): add rough collision check

kosuke55:
## Description

Calculate the distance between the object and the footprint at orientation when the object is minimum/maximum at the point of ego on the path, respectively. For each of them, return a pair to determine if they are within the margin or not, and make a collision judgment.

Intended use
- If it's safe at min_distance, it's safe
- If max_distance is unsafe, it's unsafe.
- Otherwise, use precise collision check check (e.g. footprint polygon)

image

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

psim

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_path_planner/autoware_behavior_path_planner_common/src/utils/path_safety_checker/safety_check.cpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_planner_common/src/utils/path_safety_checker/safety_check.cpp
@@ -36,6 +36,12 @@ namespace autoware::behavior_path_planner::utils::path_safety_checker
 
 namespace bg = boost::geometry;
 
+using autoware::motion_utils::calcLongitudinalOffsetPoint;
+using autoware::motion_utils::calcLongitudinalOffsetToSegment;
+using autoware::motion_utils::findNearestIndex;
+using autoware::motion_utils::findNearestSegmentIndex;
+using autoware::universe_utils::calcDistance2d;
+
 void appendPointToPolygon(Polygon2d & polygon, const geometry_msgs::msg::Point & geom_point)
 {
   Point2d point;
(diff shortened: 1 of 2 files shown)


planning/behavior_path_planner/autoware_behavior_path_planner_common/src/utils/path_safety_checker/safety_check.cpp
```diff
@@ -722,4 +728,95 @@ void updateCollisionCheckDebugMap(
...
+    double min_length_to_point = std::numeric_limits<double>::max();
+    for (const auto rel_point : shape.footprint.points) {
+      const double length_to_point = std::hypot(rel_point.x, rel_point.y);
+      if (min_length_to_point > length_to_point) {
+        min_length_to_point = length_to_point;
+      }
+    }
+  }
+
+  throw std::logic_error("The shape type is not supported in obstacle_cruise_planner.");
```
soblin, line 747: @kosuke55 Is this function missing return value ? is this function really used ?

kosuke55: oh, thanks!

I check this function is used with `BOUNDING_BOX`
```
  if (shape.type == Shape::BOUNDING_BOX) {
    return std::min(shape.dimensions.x / 2.0, shape.dimensions.y / 2.0);
   }
```

planning/behavior_path_planner/autoware_behavior_path_planner_common/src/utils/path_safety_checker/safety_check.cpp
```diff
@@ -722,4 +728,95 @@ void updateCollisionCheckDebugMap(
...
+    return shape.dimensions.x / 2.0;
+  } else if (shape.type == Shape::POLYGON) {
+    double min_length_to_point = std::numeric_limits<double>::max();
+    for (const auto rel_point : shape.footprint.points) {
+      const double length_to_point = std::hypot(rel_point.x, rel_point.y);
+      if (min_length_to_point > length_to_point) {
+        min_length_to_point = length_to_point;
+      }
+    }
+  }
```
kosuke55, line 745:
```suggestion
    }
    return min_length_to_point;
  }
```

soblin (approved):

PR #7421, opened 2024-06-10
refactor(lidar_centerpoint): rework parameters

tby-udel:
## Description
Implement the ROS Node configuration layout described in https://github.com/orgs/autowarefoundation/discussions/3371 for the `lidar_centerpoint` package.

 * Move the default parameter file to a standardized location. Use the JSON Schema for that.
 * Updated readme file.

## Tests performed
Package built and launched locally.

## Notes for reviewers
 Nothing inside the src code part was changed, only taking the parameters out.

## Interface changes
Parameter values are now required to be passed to the node when launched.

## Effects on system behavior
None

## Pre-review checklist for the PR author
The PR author **must** check the checkboxes below when creating the PR.

 * [x]  I've confirmed the [contribution guidelines](https://autowarefoundation.github.io/autoware-documentation/main/contributing/).
 * [x]  The PR follows the [pull request guidelines](https://autowarefoundation.github.io/autoware-documentation/main/contributing/pull-request-guidelines/).

 ## In-review checklist for the PR reviewers
 The PR reviewers **must** check the checkboxes below before approval.

 * [ ]  The PR follows the [pull request guidelines](https://autowarefoundation.github.io/autoware-documentation/main/contributing/pull-request-guidelines/).
 * [ ]  The PR has been properly tested.
 * [ ]  The PR has been reviewed by the code owners.
 
 ## Post-review checklist for the PR author
 The PR author **must** check the checkboxes below before merging.
 
 * [ ]  There are no open discussions or they are tracked via tickets.
 * [ ]  The PR is ready for merge.
 
 After all checkboxes are checked, anyone who has write access can merge the PR.

PR #8205, opened 2024-07-26
refactor(autoware_lidar_transfusion): split config

amadeuszsz:
## Description

This PR split `autoware_lidar_transfusion` parameter file for two configs:
* Parameters strictly related to deployed model.
* Parameters related to runtime behavior.

The same approach is applied in `autoware_lidar_centerpoint` package.

## Related links

**Parent Issue:**

- https://github.com/autowarefoundation/autoware_launch/pull/1093
- https://github.com/autowarefoundation/autoware/pull/5036

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


amadeuszsz:
@kminoda @knzo25 
`json-schema-check` fails due to same file name prefix [here](https://github.com/autowarefoundation/autoware-github-actions/blob/main/json-schema-check/validate_json_schemas.py#L17).
Unfortunately, we can't remove asterisk from `{base_name}*.param.yaml` in CI because this convention is widely used among Autoware packages to apply schema file on multiple param files. We can:
* Change file name convention from `<MODEL>_ml_package.param.yaml` to `ml_package_<MODEL>.param.yaml`.  We will need to change deployment path and launch files for transfusion, centerpoint and pointpainting.
* Bypass this test by file extension thus don't check this config in CI as it is now in centerpoint.

I would go for change naming convention if I didn't miss any drawback. If you have no objection, I will apply this for centerpoint and pointpainting as well.

kminoda:
@amadeuszsz Thanks for the proposal :pray: 
However, I don't think it is a good idea to rename the parameter file name, as they are already widely used both in Autoware and Pilot.Auto / Web.Auto, and changing those file names would take a bit of time. 

So I would say it's better to consider a workaround.

knzo25:
@amadeuszsz 
In centerpoint we face similar concerns, and as a workaround (https://github.com/autowarefoundation/autoware.universe/pull/7421) we are considering calling the vanilla model `centerpoint_base`, which sounds awful, but is compatible with the schema.
I agree with your name change, but also understand that the change would affect several places, so it would time...

amadeuszsz:
> @amadeuszsz In centerpoint we face similar concerns, and as a workaround (#7421) we are considering calling the vanilla model `centerpoint_base`, which sounds awful, but is compatible with the schema. I agree with your name change, but also understand that the change would affect several places, so it would time...

I'm not a big fan of this. It will require few copies of same schema files and still affects many places due to file renaming, as you said. 

We could restructure design for both packages to:
```
|-config
|----|-ml_package.param.yaml
|----|-ml_package_tiny.param.yaml
|----|-ml_package_sigma.param.yaml
|----|-centerpoint.param.yaml
|----|-centerpoint_tiny.param.yaml
|----|-centerpoint_sigma.param.yaml
|----|-detection_class_remapper.param.yaml
|
|-schema
|----|-ml_package.schema.json
|----|-centerpoint.schema.json
|----|-detection_class_remapper.schema.json
``` 

At the same time we can keep old configs on cloud server to let people pull files if they didn't update repository.

knzo25:
Update: we decided to skip the problematic schemas in the meantime, so this PR can be merged

knzo25 (approved):
LGTM

YoshiRi (approved):
LGTM

PR #8207, opened 2024-07-26
fix(autoware_velocity_smoother): fix passedByValue

kobayu858:
## Description

```
planning/autoware_velocity_smoother/src/smoother/analytical_jerk_constrained_smoother/velocity_planning_utils.cpp:274:89: performance: Function parameter 'times' should be passed by const reference. [passedByValue]
  double v0, double a0, double jerk_acc, double jerk_dec, int type, std::vector<double> times,
                                                                                        ^
```

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


rej55 (approved):
Sorry for late review, LGTM!

PR #8210, opened 2024-07-26
fix(autoware_behavior_velocity_crosswalk_module): fix passedByValue

kobayu858:
## Description

```
planning/behavior_velocity_planner/autoware_behavior_velocity_crosswalk_module/src/occluded_crosswalk.cpp:89:29: performance: Function parameter 'velocity_thresholds' should be passed by const reference. [passedByValue]
  const std::vector<double> velocity_thresholds, const double inflate_size)
                            ^

planning/behavior_velocity_planner/autoware_behavior_velocity_crosswalk_module/src/scene_crosswalk.cpp:148:48: performance: Function parameter 'stop_factor_points' should be passed by const reference. [passedByValue]
  const std::vector<geometry_msgs::msg::Point> stop_factor_points = {})
                                               ^
```
## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_velocity_planner/autoware_behavior_velocity_crosswalk_module/src/occluded_crosswalk.hpp
+++ b/planning/behavior_velocity_planner/autoware_behavior_velocity_crosswalk_module/src/occluded_crosswalk.hpp
@@ -80,7 +80,7 @@ double calculate_detection_range(
 /// @return selected and inflated objects
 std::vector<autoware_perception_msgs::msg::PredictedObject> select_and_inflate_objects(
   const std::vector<autoware_perception_msgs::msg::PredictedObject> & objects,
-  const double velocity_threshold, const bool skip_pedestrians, const double inflate_size);
+  const double & velocity_threshold, const bool skip_pedestrians, const double inflate_size);
 
 /// @brief clear occlusions behind the given objects
 /// @details masks behind the object assuming rays from the center of the grid map
--- a/planning/behavior_velocity_planner/autoware_behavior_velocity_crosswalk_module/src/occluded_crosswalk.hpp
+++ b/planning/behavior_velocity_planner/autoware_behavior_velocity_crosswalk_module/src/occluded_crosswalk.hpp
@@ -80,7 +80,7 @@ double calculate_detection_range(
 /// @return selected and inflated objects
 std::vector<autoware_perception_msgs::msg::PredictedObject> select_and_inflate_objects(
   const std::vector<autoware_perception_msgs::msg::PredictedObject> & objects,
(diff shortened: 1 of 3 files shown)


planning/behavior_velocity_planner/autoware_behavior_velocity_crosswalk_module/src/occluded_crosswalk.hpp
```diff
@@ -80,7 +80,7 @@ double calculate_detection_range(
 /// @return selected and inflated objects
 std::vector<autoware_perception_msgs::msg::PredictedObject> select_and_inflate_objects(
   const std::vector<autoware_perception_msgs::msg::PredictedObject> & objects,
-  const double velocity_threshold, const bool skip_pedestrians, const double inflate_size);
+  const double & velocity_threshold, const bool skip_pedestrians, const double inflate_size);
```
soblin, line 83: In general for literal type like int/double/etc,  passing by value is faster than passing by reference. But it is OK

soblin (approved):

planning/behavior_velocity_planner/autoware_behavior_velocity_crosswalk_module/src/occluded_crosswalk.hpp
kobayu858: Thanks for the advice.

takayuki5168, line 83:
```suggestion
  const double velocity_threshold, const bool skip_pedestrians, const double inflate_size);
```

takayuki5168: I reverted this change.

takayuki5168, line 83:
```suggestion
  const double velocity_threshold, const bool skip_pedestrians, const double inflate_size);
```

takayuki5168: I agree with @soblin 's comment. I created a suggestion.

kobayu858: Thank you very much. I have corrected it.

veqcc (approved):

PR #8218, opened 2024-07-26
fix(lane_change): moving object is filtered in the extended target lanes

zulfaqar-azmi-t4:
## Description

When filtering  lane change target objects, the objects that is close to target lanes' bound is considered as target objects. This is to prevent ego from lane change too near to the object. However, if the object is not moving, we don't really need to consider them as the target object.

### Before PR

https://github.com/user-attachments/assets/b43c0da5-de38-4d80-82ff-67628d6ef968

### After PR

https://github.com/user-attachments/assets/5631fe31-0d5b-4e07-ac45-6a769657401e

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/scene.cpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/scene.cpp
@@ -1201,7 +1201,8 @@ FilteredByLanesObjects NormalLaneChange::filterObjectsByLanelets(
       is_before_terminal()) {
       const auto ahead_of_ego =
         utils::lane_change::is_ahead_of_ego(common_data_ptr_, current_lanes_ref_path, object);
-      if (object.kinematics.initial_twist_with_covariance.twist.linear.x < 1.0) {
+      constexpr double stopped_obj_vel_th = 1.0;
+      if (object.kinematics.initial_twist_with_covariance.twist.linear.x < stopped_obj_vel_th) {
         if (ahead_of_ego) {
           target_lane_leading_objects.push_back(object);
           continue;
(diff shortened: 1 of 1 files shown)


planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/scene.cpp
```diff
@@ -1174,6 +1174,19 @@ FilteredByLanesObjects NormalLaneChange::filterObjectsByLanelets(
       continue;
     }
 
+    if (
+      check_optional_polygon(object, lanes_polygon.expanded_target) && is_lateral_far &&
+      is_before_terminal()) {
+      const auto ahead_of_ego =
+        utils::lane_change::is_ahead_of_ego(common_data_ptr_, current_lanes_ref_path, object);
+      if (object.kinematics.initial_twist_with_covariance.twist.linear.x < 1.0) {
```
maxime-clem, line 1182: (minor) `1.0` can this be a parameter ? Or at least a named `constexpr` (e.g., `stopped_object_velocity_threshold`).

maxime-clem (approved):

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/scene.cpp
zulfaqar-azmi-t4: fixed in https://github.com/autowarefoundation/autoware.universe/pull/8218/commits/60d6a97fad90c12a706406d354ce2b8ae5df1555

zulfaqar-azmi-t4: I didn't make it a param because it make the PR unnecessary long. However, since there are many parts of the code has similar check (for example in the rss selection for parked and non parked), I will make a pr to refactor this.

PR #8231, opened 2024-07-29
fix(autoware_multi_object_tracker): fix passedByValue

kobayu858:
## Description

```
perception/autoware_multi_object_tracker/src/debugger/debug_object.cpp:54:58: performance: Function parameter 'frame_id' should be passed by const reference. [passedByValue]
TrackerObjectDebugger::TrackerObjectDebugger(std::string frame_id)
                                                         ^

perception/autoware_multi_object_tracker/src/debugger/debug_object.cpp:176:46: performance: Function parameter 'object_data_groups' should be passed by const reference. [passedByValue]
  const std::vector<std::vector<ObjectData>> object_data_groups,
                                             ^

perception/autoware_multi_object_tracker/src/debugger/debug_object.hpp:99:48: performance: Function parameter 'object_data_groups' should be passed by const reference. [passedByValue]
    const std::vector<std::vector<ObjectData>> object_data_groups,
                                               ^
```
## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_multi_object_tracker/src/debugger/debug_object.cpp
+++ b/perception/autoware_multi_object_tracker/src/debugger/debug_object.cpp
@@ -51,7 +51,7 @@ int32_t uuidToInt(const boost::uuids::uuid & uuid)
 namespace autoware::multi_object_tracker
 {
 
-TrackerObjectDebugger::TrackerObjectDebugger(std::string frame_id)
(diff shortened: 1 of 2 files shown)


yukkysaito (approved):
LGTM

PR #8232, opened 2024-07-29
fix(autoware_object_merger): fix passedByValue

kobayu858:
## Description

```
perception/autoware_object_merger/src/object_association_merger_node.cpp:39:25: performance: Function parameter 'distance_threshold_map' should be passed by const reference. [passedByValue]
  std::map<int, double> distance_threshold_map,
                        ^

perception/autoware_object_merger/src/object_association_merger_node.cpp:40:31: performance: Function parameter 'generalized_iou_threshold_map' should be passed by const reference. [passedByValue]
  const std::map<int, double> generalized_iou_threshold_map)
                              ^
```
## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


technolojin (approved):
LGTM

PR #8221, opened 2024-07-26
feat(rtc_inteface): add function to check force deactivate

go-sakayori:
## Description
Current rtc interface could not judge whether cooperateStatus was force deactivated or not.
This PR enables to judge force deactivation (ex. deactivation from human operators).

## How was this PR tested?
[TIER IV internal test](https://evaluation.tier4.jp/evaluation/reports/2ab7653e-934b-51d5-b94e-05e2d10d3664?project_id=prd_jt)

## Notes for reviewers
None.

## Interface changes
None.

## Effects on system behavior
None.

PR #8235, opened 2024-07-29
feat(lane_change): force deactivation in prepare phase

go-sakayori:
## Description
Once the cooperateStatus for lane change was deactivated, the module state transits to `WAITING_APPROVAL` from `RUNNING`. This causes a problem when the auto mode is changed to manual when triggering force deactivation: approve from operators would be needed to trigger the lane change since the UUID in the RTC interface would not be changed.
This PR changes the state to `FAILURE` once it is force deactivated in the lane change preparing phase , and solves the above problem by restarting the lane change module; this would publish a new UUID in the RTC interface with the auto mode set to true.

## Related links
[RTC force deactivate](https://github.com/autowarefoundation/autoware.universe/pull/8221)

## How was this PR tested?
Psim + RTC manager
[Before:](https://github.com/user-attachments/assets/c01b9db5-1edc-49c7-a64b-8ef8641a797d) Once force deactivated the lane change does not proceed
[After:](https://github.com/user-attachments/assets/0f84a65e-81d9-4083-9146-fc58afcfc6d1) The lane change proceeds automatically

[TIER IV internal test](https://evaluation.tier4.jp/evaluation/reports/5274f31b-1a78-5c7d-b446-fba8f675ffff?project_id=prd_jt)

## Notes for reviewers
None

## Interface changes
None.

## Effects on system behavior
None.

### DIFF
--- a/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/interface.cpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/interface.cpp
@@ -275,6 +275,19 @@ bool LaneChangeInterface::canTransitFailureState()
       return true;
     }
 
+    const auto force_deactivated = std::any_of(
+      rtc_interface_ptr_map_.begin(), rtc_interface_ptr_map_.end(),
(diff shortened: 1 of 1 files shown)


rej55 (approved):
LGTM

PR #8237, opened 2024-07-29
fix(autoware_traffic_light_fine_detector): fix passedByValue

kobayu858:
## Description

```
perception/traffic_light_fine_detector/src/nodelet.cpp:51:67: performance: Function parameter 'double_vector' should be passed by const reference. [passedByValue]
inline std::vector<float> toFloatVector(const std::vector<double> double_vector)
                                                                  ^
```
## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


veqcc (approved):

miursh (approved):
LGTM

PR #8238, opened 2024-07-29
fix(route_handler): make new route continuous with previous route

rej55:
## Description

Before this change, it plans a new route independently of the previous route.
So, if `mission_planner` receives a modified goal, it may switch a route then it changes a path suddenly.

To fix this problem, I added a change to make a new route continuous with previous route.

## Related links

**Parent Issue:**

- Link

**Private Links:**

- [TIER IV internal link](https://tier4.atlassian.net/browse/RT1-5832)

## How was this PR tested?

- Planning simulator

Before:

https://github.com/user-attachments/assets/7c33ef77-5e70-4473-a003-d1a06b89b224

After:

https://github.com/user-attachments/assets/7727c61a-5cc6-4480-9d9b-e86b132c0120

[TIER IV internal link](https://evaluation.tier4.jp/evaluation/reports/2fa50f32-c70b-5979-b6c3-63726196dc36?project_id=prd_jt)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


kosuke55 (approved):
LGTM!

Issue #8178, opened 2024-07-24
```tensorrt_yolox``` node does not support GPU selection


ismetatabay:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I've agreed with the maintainers that I can plan this task.

### Description

The ```tensorrt_yolox``` node doesn't support GPU selection for using multiple GPUs with multiple cameras.

### Purpose

If the system has multiple cameras and the ECU has multiple GPU devices, the GPU load can be distributed across the GPUs.

### Possible approaches

Update the ```tensorrt_yolox``` node with gpu_id parameter.

### Definition of done

The updated node has:

- [ ] GPU id parameter.
- [ ] Successful GPU selection for different camera inferences.

PR #8245, opened 2024-07-29
feat(autoware_tensorrt_yolox): add GPU - CUDA device option

ismetatabay:
## Description
This PR adds a parameter to specify which GPU will be used for ```tensorrt_yolox``` inference on multiple GPU devices.

## Related links

**Parent Issue:**

- Link: https://github.com/autowarefoundation/autoware.universe/issues/8178

## How was this PR tested?

This PR has been tested with 8 cameras on a 3-GPU computer. Since GPU-0 is used for the lidar centerpoint and other things, the 8 cameras are divided between GPU-1 and GPU-2. The NVIDIA-SMI output can be found in the following image.

tensorrt_yolox_gpu_usage

## Notes for reviewers

None.

## Interface changes

| Added | `gpu_id`   | `int` | `0`         | GPU ID to select CUDA Device |

## Effects on system behavior

None.


kminoda:
@ismetatabay Please fix the CI (json schema check is failing)

perception/autoware_tensorrt_yolox/src/tensorrt_yolox_node.cpp
```diff
@@ -128,6 +133,11 @@ void TrtYoloXNode::onConnect()
 
 void TrtYoloXNode::onImage(const sensor_msgs::msg::Image::ConstSharedPtr msg)
 {
+  if (!setCudaDeviceId(gpu_id_)) {
+    RCLCPP_ERROR(this->get_logger(), "Given GPU doesn't exist or suitable");
```
Owen-Liuyuxuan, line 137: From the viewpoint of english grammar, 
"Given GPU doesn't exist or suitable" 
could be 
"GPU {with ID or not} does not exist or is not suitable."

ismetatabay: @Owen-Liuyuxuan Hello, thank you for your suggestion. It has been updated.

perception/autoware_tensorrt_yolox/src/tensorrt_yolox_node.cpp
```diff
@@ -128,6 +133,11 @@ void TrtYoloXNode::onConnect()
 
 void TrtYoloXNode::onImage(const sensor_msgs::msg::Image::ConstSharedPtr msg)
 {
+  if (!setCudaDeviceId(gpu_id_)) {
+    RCLCPP_ERROR(this->get_logger(), "GPU %d does not exist or is not suitable.", gpu_id_);
```
Shin-kyoto, line 137: nits: How about using `RCLCPP_ERROR_THROTTLE` ?

perception/autoware_tensorrt_yolox/src/tensorrt_yolox_node.cpp
```diff
@@ -128,6 +133,11 @@ void TrtYoloXNode::onConnect()
 
 void TrtYoloXNode::onImage(const sensor_msgs::msg::Image::ConstSharedPtr msg)
 {
+  if (!setCudaDeviceId(gpu_id_)) {
```
manato, line 136: If possible, I'd also appreciate your consideration whether we surely need to call `setCudaDevicdId` on every iteration. Since `TrtYoloX` class maintains addresses of the GPU memory to avoid memory allocation on every iteration, I think it would be nice for the device ID to be fixed through iterations.

perception/autoware_tensorrt_yolox/include/autoware/tensorrt_yolox/tensorrt_yolox_node.hpp
```diff
@@ -94,6 +95,7 @@ class TrtYoloXNode : public rclcpp::Node
   bool is_roi_overlap_segment_;
   bool is_publish_color_mask_;
   float overlap_roi_score_threshold_;
+  int gpu_id_;
```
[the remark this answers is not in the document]
ismetatabay: @Shin-kyoto Thank you for your suggestion. It is an important point and has been updated.

perception/autoware_tensorrt_yolox/src/tensorrt_yolox_node.cpp
ismetatabay: It has been updated, thanks

perception/autoware_tensorrt_yolox/src/tensorrt_yolox_node.cpp
```diff
@@ -90,6 +91,10 @@ TrtYoloXNode::TrtYoloXNode(const rclcpp::NodeOptions & node_options)
   const tensorrt_common::BatchConfig batch_config{1, 1, 1};
   const size_t max_workspace_size = (1 << 30);
 
+  if (!setCudaDeviceId(gpu_id_)) {
```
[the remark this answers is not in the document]
ismetatabay: @manato -san, thank you for your valuable comments. Since I am using the same 3xGPU model for our tests, the created engine file for GPU 0 (default) is valid for the remaining two GPUs. If there is another different model GPU, it will be a problem. Therefore, we can update the [```tensorrt_common```](https://github.com/autowarefoundation/autoware.universe/tree/main/common/tensorrt_common) package to handle different GPUs as well. What do you think about that?

perception/autoware_tensorrt_yolox/src/tensorrt_yolox_node.cpp
ismetatabay: I checked this situation, and if I don't put the ```setCudaDeviceId()``` in the callback, I get the following error since the CUDA device switches ID to 0 (it was checked whit ```cudaGetDevice()```).
```
CUDA error: an illegal memory access was encountered
```

Additionally, I checked the previous implementation in the ```tensorrt_yolo``` package, and it also has a very similar function in the callback:

https://github.com/autowarefoundation/autoware.universe/blob/9f1ae6d9e84efa1b3f04ac76d0105853c1237a3b/perception/tensorrt_yolo/src/nodelet.cpp#L131-L140

perception/autoware_tensorrt_yolox/src/tensorrt_yolox_node.cpp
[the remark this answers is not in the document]
manato: @ismetatabay
Thank you for your consideration! 
Yes, my main concern was such kinds of environments that different model GPUs exist. Regarding building engine on the specific device, it seems to be enough to call `cudaSetDevice()` before an engine builder starts to work, according to [FAQs: Q.How do I use TensorRT on multiple GPUs?](https://docs.nvidia.com/deeplearning/tensorrt/developer-guide/index.html#faq)  (I believe you can confirm this behavior by monitoring which GPU is used using `nvidia-smi` during engine building). If this is the case, I think we don't need to modify `tensorrt_common` because we just need to call `cudaSetDevice` before constructing `tensorrt_common::TrtCommon`.

Sorry for repeatedly asking, but could you consider to move this cuda-related codes into `tensorrt_yolox.cpp` for further encapsulation so that other nodes who use yolox as a module can enjoy your improvement?

perception/autoware_tensorrt_yolox/src/tensorrt_yolox_node.cpp
manato: Thanks for the confirmation.  My concern here was additional overhead (i.e., increasing processing time) by introducing setDevice on every iteration. According to the [document](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__DEVICE.html#group__CUDART__DEVICE_1g159587909ffa0791bbe4b40187a4c6bb), `cudaSetDevice` will take significant time under the limited situation, I guess the practical overhead of this code is not so large.

perception/autoware_tensorrt_yolox/src/tensorrt_yolox_node.cpp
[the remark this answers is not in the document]
ismetatabay: Thank you @manato -san, you are right. I am okay with moving everything to ```tensorrt_yolox.cpp```. With this update, ```traffic_light_fine_detector``` can also select GPU as you mentioned. However, if I move everything to ```TrtYoloX``` class, I still get an illegal memory access error from CUDA unless I call ```cudaSetDevice``` before creating the ```trt_yolox_``` object in the ```TrtYoloXNode``` constructor. Do you have any suggestions on that?

https://github.com/autowarefoundation/autoware.universe/blob/3e79b8f67f5a3428c04efb7545c57b6401c83d17/perception/autoware_tensorrt_yolox/src/tensorrt_yolox_node.cpp#L92-L97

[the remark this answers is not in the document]
manato: Thank you @ismetatabay -san for your thoughtful consideration!

I guess the cause of the issue you observed is [this](https://github.com/autowarefoundation/autoware.universe/blob/3e79b8f67f5a3428c04efb7545c57b6401c83d17/perception/autoware_tensorrt_yolox/include/autoware/tensorrt_yolox/tensorrt_yolox.hpp#L270) line; the line attempts to create `cudaStream` on the default (`0`) device during construction of an instance, and the stream will be used for memory copy before/after inference. Could you please try to move to call `makeCudaStream()` after setting the device in the class constructor and see if it makes a difference?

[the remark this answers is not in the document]
ismetatabay: Hello @manato -san, sorry for the late update. I was on leave and only had a chance to update it today. I have moved the updates to the tensorrt_yolox. Could you please check it? Thank you :raised_hands:

perception/autoware_traffic_light_fine_detector/src/traffic_light_fine_detector_node.cpp
```diff
@@ -65,6 +65,7 @@ TrafficLightFineDetectorNode::TrafficLightFineDetectorNode(const rclcpp::NodeOpt
   std::string model_path = declare_parameter("fine_detector_model_path", "");
   std::string label_path = declare_parameter("fine_detector_label_path", "");
   std::string precision = declare_parameter("fine_detector_precision", "fp32");
+  uint8_t gpu_id = declare_parameter("gpu_id", 0);
```
Shin-kyoto, line 68:
```suggestion
  const uint8_t gpu_id = declare_parameter("gpu_id", 0);
```

perception/autoware_tensorrt_yolox/include/autoware/tensorrt_yolox/tensorrt_yolox.hpp
```diff
@@ -89,15 +89,22 @@ class TrtYoloX
     const std::string & model_path, const std::string & precision, const int num_class = 8,
     const float score_threshold = 0.3, const float nms_threshold = 0.7,
     const tensorrt_common::BuildConfig build_config = tensorrt_common::BuildConfig(),
-    const bool use_gpu_preprocess = false, std::string calibration_image_list_file = std::string(),
-    const double norm_factor = 1.0, [[maybe_unused]] const std::string & cache_dir = "",
+    const bool use_gpu_preprocess = false, const int gpu_id = 0,
```
Shin-kyoto, line 92:
```suggestion
    const bool use_gpu_preprocess = false, const uint8_t gpu_id = 0,
```
Although int is frequently used in Autoware, it is actualy an unsafe type. Please use uint8_t instaed.
cf. https://matt.sh/howto-c#:~:text=Writing%20code-,Types,-If%20you%20find

perception/autoware_tensorrt_yolox/include/autoware/tensorrt_yolox/tensorrt_yolox.hpp
```diff
@@ -280,6 +287,8 @@ class TrtYoloX
 
   // flag whether preprocess are performed on GPU
   bool use_gpu_preprocess_;
+  // GPU id for inference
+  uint8_t gpu_id_;
```
Shin-kyoto, line 291:
```suggestion
  const uint8_t gpu_id_;
```

Please add const and use member initialization list.

Shin-kyoto (changes requested):

perception/autoware_tensorrt_yolox/src/tensorrt_yolox.cpp
```diff
@@ -156,16 +156,25 @@ namespace autoware::tensorrt_yolox
...
-  const bool use_gpu_preprocess, std::string calibration_image_list_path, const double norm_factor,
-  [[maybe_unused]] const std::string & cache_dir, const tensorrt_common::BatchConfig & batch_config,
-  const size_t max_workspace_size, const std::string & color_map_path)
+  const bool use_gpu_preprocess, const int gpu_id, std::string calibration_image_list_path,
+  const double norm_factor, [[maybe_unused]] const std::string & cache_dir,
+  const tensorrt_common::BatchConfig & batch_config, const size_t max_workspace_size,
+  const std::string & color_map_path)
 {
+  gpu_id_ = gpu_id;
+  std::cout << "GPU" << gpu_id_ << "is selected for the inference !" << std::endl;
```
Shin-kyoto, line 165: @ismetatabay 

Please use `RCLCPP_INFO` or `RCLCPP_INFO_THROTTLE`.
cf. https://github.com/autowarefoundation/autoware.universe/blob/3551d649d390ec87daa57695742985c5370ebc93/perception/autoware_tensorrt_yolox/src/tensorrt_yolox_node.cpp#L109

Shin-kyoto (changes requested):

perception/autoware_traffic_light_fine_detector/src/traffic_light_fine_detector_node.cpp
ismetatabay: Done :+1:

perception/autoware_tensorrt_yolox/include/autoware/tensorrt_yolox/tensorrt_yolox.hpp
ismetatabay: Done :+1:

perception/autoware_tensorrt_yolox/include/autoware/tensorrt_yolox/tensorrt_yolox.hpp
ismetatabay: Done :+1:

perception/autoware_tensorrt_yolox/include/autoware/tensorrt_yolox/tensorrt_yolox.hpp
ismetatabay: Done :+1:

perception/autoware_tensorrt_yolox/src/tensorrt_yolox.cpp
ismetatabay: Hello @Shin-kyoto -san, thank you for your valuable suggestions. We've moved our development node to the tensorrt_yolox library (```tensorrt_yolox.cpp```). I am not sure whether to pass rclcpp into ```tensorrt_yolox.cpp``` for logging purposes. What do you think about that?

perception/autoware_tensorrt_yolox/src/tensorrt_yolox_node.cpp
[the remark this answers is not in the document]
manato: Hi @ismetatabay -san, thank you for your continuous updates. The modifications look good to me!

perception/autoware_tensorrt_yolox/src/tensorrt_yolox.cpp
```diff
@@ -156,16 +156,25 @@ namespace autoware::tensorrt_yolox
...
-  const bool use_gpu_preprocess, std::string calibration_image_list_path, const double norm_factor,
-  [[maybe_unused]] const std::string & cache_dir, const tensorrt_common::BatchConfig & batch_config,
-  const size_t max_workspace_size, const std::string & color_map_path)
+  const bool use_gpu_preprocess, const uint8_t gpu_id, std::string calibration_image_list_path,
+  const double norm_factor, [[maybe_unused]] const std::string & cache_dir,
+  const tensorrt_common::BatchConfig & batch_config, const size_t max_workspace_size,
+  const std::string & color_map_path)
+: gpu_id_(gpu_id)
 {
+  std::cout << "GPU " << gpu_id_ << " is selected for the inference!" << std::endl;
```
manato, line 165:
```suggestion
  std::cout << "GPU " << std::to_string(gpu_id_) << " is selected for the inference!" << std::endl;
```
@ismetatabay -san,
This is the final request for modification. Because the current code output is like the following in my environment, I'd appreciate it if you fix it:
```
[autoware_tensorrt_yolox_node_exe-2] GPU  is selected for the inference!
```
Once you handle this, I'll approve this PR!

perception/autoware_tensorrt_yolox/src/tensorrt_yolox.cpp
```diff
@@ -156,16 +156,25 @@ namespace autoware::tensorrt_yolox
...
+  if (!setCudaDeviceId(gpu_id_)) {
+    throw std::runtime_error(
```
Shin-kyoto, line 167: Can you use try-catch and `RCLCPP_ERROR` instead of throw?

cf. https://google.github.io/styleguide/cppguide.html#Exceptions

Shin-kyoto (changes requested):

perception/autoware_tensorrt_yolox/src/tensorrt_yolox.cpp
ismetatabay: Thank you @manato -san, it is done :+1:

perception/autoware_tensorrt_yolox/src/tensorrt_yolox.cpp
ismetatabay: Hello @Shin-kyoto-san :raised_hands: , since tensorrt_yolox.cpp doesn't include rclcpp and is independent of ROS, I believe we don't need to use RCLCPP loggers. What do you think? ([similar comment](https://github.com/autowarefoundation/autoware.universe/pull/8245#discussion_r1718623591))

Shin-kyoto: Hi, @ismetatabay -san !!
Thank you for your reply!

C++ exceptions should only be used when it's acceptable for the process to die if they are not caught. In other words, in situations where Autoware has no other option but to terminate. 
We want to avoid this because there's a possibility that nodes might suddenly die while in use.

@Owen-Liuyuxuan -san, what do you think about it?

ismetatabay: @Shin-kyoto -san & @Owen-Liuyuxuan -san, thank you for your valuable comments. I have implemented a flag and pushed the corresponding commit. I have also removed the exception mechanism. Please check it and let me know if you have any suggestions on that

perception/autoware_tensorrt_yolox/src/tensorrt_yolox.cpp
```diff
@@ -341,6 +348,17 @@ TrtYoloX::~TrtYoloX()
   }
 }
 
+bool TrtYoloX::setCudaDeviceId(const uint8_t gpu_id)
+{
+  cudaError_t cuda_err = cudaSetDevice(gpu_id);
+  if (cuda_err != cudaSuccess) {
+    std::cerr << "Failed to set CUDA device: " << cudaGetErrorString(cuda_err) << std::endl;
```
Shin-kyoto, line 355: @ismetatabay 
Thanks!
Can you replace this `std::cerr` to `RCLCPP_ERROR`, too?

Shin-kyoto (changes requested):

perception/autoware_tensorrt_yolox/src/tensorrt_yolox.cpp
ismetatabay: Thank you as well. Similarly, this function is included in TrtYoloX instead of TrtYoloXNode. Therefore, we cannot directly change it to use ROS logs. We can either remove it since we have already implemented the GPU initialization control and log mechanism in the ROS nodes, or we can keep it. I’m okay with both options. What do you think? @Shin-kyoto -san & @Owen-Liuyuxuan -san

ismetatabay: done :+1:

Shin-kyoto (changes requested):
Thank you for the changes. 
Finally, could you please perform a functionality check with all the changes included and document the verification method and results?

Shin-kyoto (approved):
LGTM
- I confirmed that the content of this PR is enough to merge.
- I did NOT check this PR by running autoware.

manato (approved):
@ismetatabay 
Thank you for your continuous updating to solve my concerns. LGTM!

PR #7095, opened 2024-05-22
feat: perception objects pointcloud better visualization

KhalilSelyan:
## Description
This PR adding new function to autoware_auto_perception_rviz_plugin. Now it has ability to publish point cloud of detected objects. It improve visualization of perception pipeline.
This feature can be enabled/disabled in control panel of the plugin by toggling Publish Objects Pointcloud property.
New point cloud plugin should be added to autoware config to visualize output of the plugin. Like it's shown in this [video](https://youtu.be/cRTaotGZ1x4).

Output topics for point clouds:

/rviz2/output/predicted_objects_pointcloud
/rviz2/output/detected_objects_pointcloud
/rviz2/output/tracked_objects_pointcloud
[Video demo](https://youtu.be/lJhmTXtJ7jY) with predicted objects from autoware demo bagfile.

## Related links
This feature was discussed here https://github.com/autowarefoundation/autoware.universe/issues/2860.

## Tests performed
This feature was tested manually against sample rosbag from Rosbag replay simulation demo. I didn't find any reduces in performance.

## Pre-review checklist for the PR author
- [x] The PR author must check the checkboxes below when creating the PR.

 I've confirmed the [contribution guidelines](https://autowarefoundation.github.io/autoware-documentation/main/contributing/).
- [x] The PR follows the [pull request guidelines](https://autowarefoundation.github.io/autoware-documentation/main/contributing/pull-request-guidelines/).
In-review checklist for the PR reviewers
- [ ] The PR reviewers must check the checkboxes below before approval.

- [ ] The PR follows the [pull request guidelines](https://autowarefoundation.github.io/autoware-documentation/main/contributing/pull-request-guidelines/).
- [ ] The PR has been properly tested.
- [ ] The PR has been reviewed by the code owners.
Post-review checklist for the PR author
The PR author must check the checkboxes below before merging.

- [ ] There are no open discussions or they are tracked via tickets.
- [ ] The PR is ready for merge.
After all checkboxes are checked, anyone who has write access can merge the PR.

PR #8247, opened 2024-07-29
feat: perception objects pointcloud better visualization

KhalilSelyan:
## Description
From the previous [PR](https://github.com/autowarefoundation/autoware.universe/pull/7095), there was a bug related to the implementation of the Pointcloud attached to the objects which caused RViz to crash regularly/randomly, i have not been able to fix the pcl implementation and for that i created a separate PR for just the 2d bottom visualization of the object such as this instead of the bounding boxes

image

image

## Related links

- https://github.com/autowarefoundation/autoware.universe/pull/7095

## Effects on system behavior

None.


xmfcx:
Too old now and there was lack of interest/will to review/merge this. Will abandon it for now.

PR #8256, opened 2024-07-30
fix(autoware_detected_object_validation): fix cppcheck warnings of functionStatic

taisa1:
## Description
```
perception/autoware_detected_object_validation/src/obstacle_pointcloud/debugger.hpp:100:46: performance: inconclusive: Technically the member function 'autoware::detected_object_validation::obstacle_pointcloud::Debugger::toXYZ' can be static (but you may consider moving to unnamed namespace). [functionStatic]
perception/autoware_detected_object_validation/src/obstacle_pointcloud/obstacle_pointcloud_validator.hpp:95:39: performance: inconclusive: Technically the member function 'autoware::detected_object_validation::obstacle_pointcloud::Validator2D::convertToXYZ' can be static (but you may consider moving to unnamed namespace). [functionStatic]
```
## How was this PR tested?

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


miursh (commented):
@taisa1 
~~You just editing header file, don't you need to edit cpp file as well?~~
I guess you don't

miursh (approved):

PR #8258, opened 2024-07-30
fix(traffic_light_occlusion_predictor): fix cppcheck warnings of functionStatic

taisa1:
## Description
```
perception/autoware_traffic_light_occlusion_predictor/src/occlusion_predictor.hpp:78:8: performance: inconclusive: Technically the member function 'autoware::traffic_light::CloudOcclusionPredictor::sampleTrafficLightRoi' can be static (but you may consider moving to unnamed namespace). [functionStatic]
perception/autoware_traffic_light_occlusion_predictor/src/occlusion_predictor.hpp:83:8: performance: inconclusive: Technically the member function 'autoware::traffic_light::CloudOcclusionPredictor::calcRoiVector3D' can be static (but you may consider moving to unnamed namespace). [functionStatic]

```

## How was this PR tested?

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


miursh (commented):
@taisa1 
For my understanding, could you please explain what is the advantages of adding static to the functions?

taisa1:
@miursh 
Adding `static` keyword makes it clear that the function does not modify the state of the class object.
In terms of performance, I think there is a slight improvement (by reducing overhead) or no difference.

miursh:
@taisa1 
Thanks!
If a function does not modify the class object, I thought using the const modifier would be appropriate. 
However, does using static imply that the function neither modifies nor even accesses the class object at all?

taisa1:
@miursh 
Yes, static functions are not bound to any objects, so the static keyword ensures they do not access non-static members (sorry for the insufficient explanation)

miursh (approved):
Thank you for your explanation.
LGTM

PR #8260, opened 2024-07-30
fix(autoware_vehicle_cmd_gate): fix cppcheck warning of functionStatic

taisa1:
## Description
```
control/autoware_vehicle_cmd_gate/src/vehicle_cmd_filter.hpp:90:10: performance: inconclusive: Technically the member function 'autoware::vehicle_cmd_gate::VehicleCmdFilter::limitDiff' can be static (but you may consider moving to unnamed namespace). [functionStatic]
```

Static member functions cannot be `const`, so some `const` are removed.

## Related links

[C++ reference (static)](https://en.cppreference.com/w/cpp/language/static)

## How was this PR tested?

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


TakaHoribe (approved):
LGTM, thank you!

PR #8266, opened 2024-07-30
fix(diagnostic_graph_aggregator): fix cppcheck warning of functionStatic

taisa1:
## Description
```
system/diagnostic_graph_aggregator/src/common/graph/units.hpp:35:8: performance: inconclusive: Technically the member function 'diagnostic_graph_aggregator::UnitLink::initialize_status' can be static (but you may consider moving to unnamed namespace). [functionStatic]
```

## How was this PR tested?

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


isamu-takagi:
This function is prepared for future implementation. Since I plan to update member variables, please leave it as it is.

taisa1:
I understand. To avoid it remaining as a warning, I changed it to suppress a warning with an inline comment. Could you check this? @isamu-takagi

isamu-takagi (approved):
LGTM

Issue #8219, opened 2024-07-26
lidar_centerpoint does not compile with CUDA 12.5


PPeltzerFka:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I'm convinced that this is not my fault but a bug.

### Description

Building Universe fails for me when following the [Source installation](https://autowarefoundation.github.io/autoware-documentation/main/installation/autoware/source-installation/) guide, with running the `./setup-dev-env.sh` on Ubuntu 22, installing the latest Cuda 12.5 (released in May). When trying to build Autoware Universe, the compilation fails for lidar_centerpoint, pointing to missing declaration for the `__syncthreads()` function and more.

### Expected behavior

Following the tutorial, compilation is expected to be successful.

### Actual behavior

The compilation fails for lidar_centerpoint. Unfortunately, I only have a part of the error message right now as I dont have the time capacity to reinstall CUDA 12.5 again. This one was originating from test_preprocess_kernel.cpp :

```
/usr/local/cuda/include/cub/util_ptx.cuh:271:5: error: ‘__syncthreads’ was not declared in this scope
  271 |     __syncthreads();
```

### Steps to reproduce

1. Run ./setup-dev-env.sh
2. Validate you're running CUDA 12.5 (`nvcc --version`)
3. Build autoware universe, or only up to lidar_centerpoint: `colcon build --symlink-install --cmake-args -DCMAKE_BUILD_TYPE=Release --packages-up-to autoware_lidar_centerpoint`

### Versions

- OS: Ubuntu 22
- ROS 2: Humble
- Autoware: Initially tried with 2024.06 tag, but issue appears to be still there in latest Universe main commit

### Possible causes

Google spits out the following threads related to this:

- [Unable to include thrust/host_vector.h and others with CUDA 12.5](https://stackoverflow.com/questions/78784236/unable-to-include-thrust-host-vector-h-and-others-with-cuda-12-5)
- [When upgrade from CUDA12.4 to 12.5 the compilation became broken](https://forums.developer.nvidia.com/t/when-upgrade-from-cuda12-4-to-12-5-the-compilation-became-broken/295814)

There appears to be an API change in the thrust library released in CUDA 12.5, no longer allowing the inclusion of certain (?) headers in the .cpp/.hpp files instead of .cu files (judging from the linked topics).

[...]

PR #8272, opened 2024-07-31
fix(autoware_lidar_centerpoint): place device vector in CUDA device system

amadeuszsz:
## Description

CUDA 12.5 makes device vectors works with CUDA target only by default. Before this PR it works, but it shouldn't. As maintainer says _no Thrust header can be included in a host-only TU_.

## Related links

**Parent Issue:**

- https://github.com/autowarefoundation/autoware.universe/issues/8219

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


knzo25 (approved):
LGTM !

PR #8275, opened 2024-07-31
feat(pose_initializer, ndt_scan_matcher): check initial pose result and publish diag

RyuYamamoto:
## Description

- Check the score of the initial pose estimation result in ndt_scan_matcher and if it is below the threshold, set false to the reliability in PoseWithCovarianceStamped.srv and return it.
- Receive the initial pose estimation result in pose_initializer and if reliability is false, output an ERROR message in the `/diagnostics` topic.

diagnostic_pose_reliability

## Related links
depend https://github.com/tier4/tier4_autoware_msgs/pull/143

**Parent Issue:**

- Link

## How was this PR tested?
- Check that localization works correctly in logging_simulator.
- check that if initial pose estimation is successful,  initial_pose_reliability set to True .
- check that  if initial pose estimation is failed, initial_pose_reliability set to False.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


SakodaShintaro (commented):
I think, since there is no change about yabloc's initializer service, `reliable` will be always false when using yabloc. Is it correct?

If so, in my opinion

(1) Best: judge by yabloc's score (If there is such a thing)
(2) Better: always returns true

RyuYamamoto:
@SakodaShintaro 

I think that's right, for now, I will fix to always returns true.

if implemented reliability judge logic in yabloc in the future, and I want to use it for judge reliability.

localization/yabloc/yabloc_pose_initializer/src/camera/camera_pose_initializer_core.cpp
```diff
@@ -199,6 +199,8 @@ void CameraPoseInitializer::on_service(
   // Estimate orientation
   const double initial_yaw_angle_rad = 2 * std::atan2(orientation.z, orientation.w);
   const auto yaw_angle_rad_opt = estimate_pose(pos_vec3f, initial_yaw_angle_rad, yaw_std_rad);
+  // TODO(localization/mapping team): judge reliability of estimate pose
```
SakodaShintaro, line 202: Thank you for adding TODO comment :+1:

SakodaShintaro (approved):
Looks Good To Me

PR #8279, opened 2024-07-31
fix(diagnostic_graph_aggregator): fix functionConst

kobayu858:
## Description
This is a fix based on cppcheck functionConst warnings.

```
system/diagnostic_graph_aggregator/src/common/graph/graph.hpp:34:8: style: inconclusive: Technically the member function 'diagnostic_graph_aggregator::Graph::update' can be const. [functionConst]
  void update(const rclcpp::Time & stamp);
       ^
```
## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/system/diagnostic_graph_aggregator/src/common/graph/graph.cpp
+++ b/system/diagnostic_graph_aggregator/src/common/graph/graph.cpp
@@ -37,7 +37,7 @@ void Graph::create(const std::string & file, const std::string & id)
   id_ = id;
 }
 
-void Graph::update(const rclcpp::Time & stamp)
+void Graph::update(const rclcpp::Time & stamp) const
 {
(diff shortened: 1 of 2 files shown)


veqcc (approved):

isamu-takagi:
As the name suggests, this function updates the state, so it feels strange to use const. For implementation reasons, the member variables are not updated, but the references, which are the actual internal state, are updated.

kobayu858:
@isamu-takagi 

Thanks for the advice.
I have changed the method to suppress the warning by adding a comment instead of adding const.

isamu-takagi (approved):
@kobayu858 Thank you. LGTM.

PR #8292, opened 2024-07-31
fix(doc, object_merger): fix object merger document path

Owen-Liuyuxuan:
## Description

Fix the path of the object merger documents.

https://autowarefoundation.github.io/autoware.universe/main/perception/autoware_object_merger/

For now the links does not work correctly, we need to fix the path due to package name change.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


technolojin (approved):
LGTM

YoshiRi (approved):
Thanks, LGTM

PR #8306, opened 2024-08-01
fix(docs): fox docs for autoware object range splitter

Owen-Liuyuxuan:
## Description

Fix docs error during adding the `autoware_` prefix. 

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


YoshiRi (approved):
LGTM

PR #8313, opened 2024-08-01
fix(diagnostic_graph_aggregator): fix uninitMemberVar

kobayu858:
## Description
This is a fix based on cppcheck uninitMemberVar warnings.

```
system/diagnostic_graph_aggregator/src/common/graph/config.hpp:51:12: warning: Member variable 'UnitConfig::index' is not initialized in the constructor. [uninitMemberVar]
  explicit UnitConfig(const TreeData & data) : data(data) {}
           ^
```

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/system/diagnostic_graph_aggregator/src/common/graph/config.hpp
+++ b/system/diagnostic_graph_aggregator/src/common/graph/config.hpp
@@ -18,10 +18,10 @@
 #include "data.hpp"
 #include "types.hpp"
 
+#include <limits>
 #include <memory>
 #include <string>
 #include <vector>
-#include <limits>
 
 namespace diagnostic_graph_aggregator
 {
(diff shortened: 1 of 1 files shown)


isamu-takagi:
This variable will be used later by setting a value, so there is no point in initializing it to the default.

kobayu858:
@isamu-takagi 

Thanks for the advice.
The goal this time is to eliminate the cppcheck warning.
Therefore, something needs to be done.

There are three ways to eliminate the warnings. Currently I am using #1.

1. default initialization in member variable declarations
2. initialize in constructor
3. suppress it with a comment

Any opinions on which method is better would be appreciated.

isamu-takagi:
@kobayu858 Could you initialize it with SIZE_MAX to indicate it's uninitialized?

kobayu858:
@isamu-takagi 

Thanks for the advice.
Changed to initialize with the maximum value.

isamu-takagi (approved):
LGTM

PR #8316, opened 2024-08-01
feat(lane_change): use different rss param to deal with parked vehicle

zulfaqar-azmi-t4:
## Description

Available RSS parameters is too conservative for parked vehicle.
In this PR, RSS parameters for parked vehicle is separated to relax the criteria.

## Related links

https://github.com/autowarefoundation/autoware_launch/pull/1104

## How was this PR tested?

1. With PSIM
2. [TIER IV internal link](https://evaluation.tier4.jp/evaluation/reports/b68f2a5a-fea3-5be2-bffc-bc9cf05b3e6f?project_id=prd_jt)

## Notes for reviewers

None.

## Interface changes

None.

#### Additions and removals

| Added | `safety_check.parked.expected_front_deceleration`    | `double` | `-1.0`         | The front object's maximum deceleration when the front vehicle perform sudden braking. |
| Added | `safety_check.parked.expected_rear_deceleration`   | `double` | `-2.0`         | The rear object's maximum deceleration when the rear vehicle perform sudden braking. |
| Added | `safety_check.parked.rear_vehicle_reaction_time`   | `double` | `1.0`         | The reaction time of the rear vehicle driver which starts from the driver noticing the sudden braking of the front vehicle until the driver step on the brake. |
| Added | `safety_check.parked.rear_vehicle_safety_time_margin`   | `double` | `0.8`         | The time buffer for the rear vehicle to come into complete stop when its driver perform sudden braking. |
| Added | `safety_check.parked.lateral_distance_max_threshold`   | `double` | `1.0`         | The lateral distance threshold that is used to determine whether lateral distance between two object is enough and whether lane change is safe.  |
| Added | `safety_check.parked.longitudinal_distance_min_threshold`   | `double` | `3.0`         | The longitudinal distance threshold that is used to determine whether longitudinal distance between two object is enough and whether lane change is safe. |
| Added | `safety_check.parked.longitudinal_velocity_delta_time`   | `double` | `0.8`         | The time multiplier that is used to compute the actual gap between vehicle at each predicted points (not RSS distance)  |

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/scene.cpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/scene.cpp
@@ -2069,9 +2069,14 @@ PathSafetyStatus NormalLaneChange::isLaneChangePathSafe(
     const auto obj_predicted_paths = utils::path_safety_checker::getPredictedPathFromObj(
       obj, lane_change_parameters_->use_all_predicted_path);
     auto is_safe = true;
+    const auto selected_rss_param =
+      (obj.initial_twist.twist.linear.x <=
+       lane_change_parameters_->prepare_segment_ignore_object_velocity_thresh)
+        ? lane_change_parameters_->rss_params_for_parked
+        : rss_params;
     for (const auto & obj_path : obj_predicted_paths) {
       const auto collided_polygons = utils::path_safety_checker::getCollidedPolygons(
-        path, ego_predicted_path, obj, obj_path, common_parameters, rss_params, 1.0,
+        path, ego_predicted_path, obj, obj_path, common_parameters, selected_rss_param, 1.0,
         get_max_velocity_for_safety_check(), collision_check_yaw_diff_threshold,
         current_debug_data.second);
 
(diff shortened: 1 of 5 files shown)


planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/scene.cpp
```diff
@@ -2069,9 +2069,14 @@ PathSafetyStatus NormalLaneChange::isLaneChangePathSafe(
     const auto obj_predicted_paths = utils::path_safety_checker::getPredictedPathFromObj(
       obj, lane_change_parameters_->use_all_predicted_path);
     auto is_safe = true;
+    const auto selected_rss_param =
+      (obj.initial_twist.twist.linear.x <=
+       lane_change_parameters_->prepare_segment_ignore_object_velocity_thresh)
```
maxime-clem, line 2074: (minor) `prepare_segment_ignore_object_velocity_thresh` this parameter should probably be renamed since the object is not ignored (can be done in a future PR).

maxime-clem (approved):
The code looks good :+1:

mkquda (approved):

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/scene.cpp
zulfaqar-azmi-t4: Yeah, I plan to rename it to something else in the later PR.

PR #8325, opened 2024-08-01
feat(pid_longitudinal_controller)!: add acceleration feedback block

yuki-takagi-66:
## Description
This PR provides an acceleration feedback feature for the pid_longitudinal_controller.

The red allows in the following figure are added by this PR.
image

Note that the acceleration feedback gain G_a(s) is set as 0 in the default parameters, https://github.com/autowarefoundation/autoware_launch/pull/1105.
Therefore, no behavior is changed by this PR.

With the current construction of the speed controller, it is difficult to achieve both disturbance tolerance and tracking performance when the response delay of the vehicle is large. In order to better ensure this performance, we have added a block that performs feedback control by looking at acceleration information.
This is because acceleration is closer to the input value to the plant than velocity, and the response is about 90 degrees faster.

## Related links
launch PR: https://github.com/user-attachments/assets/f2258401-329f-489b-8a5f-580adbb294be

## How was this PR tested?
psim and tier4 internal tests

## Notes for reviewers

None.

## Effects on system behavior

None.


takayuki5168 (approved):

yuki-takagi-66:
@Motsu-san 
Just a notification, We will merge this PR, and then, the twist2accel node and stop_fillter node will be included in the feedback control loop.

PR #8340, opened 2024-08-02
refactor(gyro_odometer)!: prefix package and namespace with autoware

a-maumau:
## Description
This PR will add a `autoware_` prefix to `gyro_odometer`.

## Related links

**Parent Issue:**

## How was this PR tested?
Checked it works with autoware sample-rosbag.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


localization/autoware_gyro_odometer/CMakeLists.txt
```diff
@@ -9,6 +9,10 @@ ament_auto_add_library(${PROJECT_NAME} SHARED
   src/diagnostics_module.cpp
 )
 
+include_directories(
+  src
+)
+
```
SakodaShintaro, line 15: Why are these lines necessary?

a-maumau: It seems when changing the directory structure, the compiler can not find the header in this case.

For the original structure case,
```
src/
    gyro_odometer/
        gyro_odometer_core.hpp
```
and the including `#include "gyro_odometer/gyro_odometer_core.hpp"` use the relative path for the header and this make compiler to search the `gyro_odometer_code.hpp` with searching in relative path 
 `gyro_odometer/gyro_odometer_core.hpp`.

But for this PR, I changed the dir structure to `src/gyro_odometer_core.hpp` and the including `#include "gyro_odometer_core.hpp"`. This will result in compiler to search the `gyro_odometer_core.hpp` in same dir, i.e. `test/gyro_odometer_core.hpp`.

ref: https://cmake.org/cmake/help/latest/command/target_include_directories.html

The
```
include_directories(
  src
)
```
itself might not be appropriate which uses `include_directories` , I will make a commit to use `target_include_directories` instead.

a-maumau:
Resolved conflicts.

SakodaShintaro:
@a-maumau 
Sorry, some conflicts occurred again. Please check .github/CODEOWNERS.

a-maumau:
@SakodaShintaro I fixed the conflict.

SakodaShintaro:
The following error occurred:
```
+ rosdep install -y --from-paths src --ignore-src --rosdistro humble
ERROR: the following packages/stacks could not have their rosdep keys resolved
to system dependencies:
tier4_localization_launch: Cannot locate rosdep definition for [gyro_odometer]
```

a-maumau:
I fix the `launch/tier4_localization_launch/package.xml`, so I think the issue is fixed.

SakodaShintaro:
@a-maumau 
Unfortunately, `.github/CODEOWNERS` conflicts again :(

a-maumau:
I resolved the `.github/CODEOWNERS`.

SakodaShintaro (approved):
I have confirmed that logging_simulator works well and the mean error against AWSIM GT is roughly the same as before (7.2cm -> 6.7cm)

- **Directory Name**
  - [x] Add `autoware_` as a prefix to the directory name.

- **packages.xml**
  - [x] Add `autoware_` to the `name` element.

- **CMakeLists.txt**
  - [x] Change the `project` name to `autoware_***`.

- **Launch Files**
  - [x] Add `autoware_` before `find-pkg-share`.
  - [x] Change `node pkg="<pkgname>"` to `autoware_<pkgname>`.

- **Verification Points**
  - [x] Search `find-pkg-share` in autoware
  - [x] Check README file (especially json schema)

PR #8345, opened 2024-08-05
fix(static_obstacle_avoidance): check opposite lane

satoshi-ota:
## Description

I found a bug: this module generates avoidance path for stopped vehicle even if there is opposite lane. In this case, we assume that it's NOT a parked vehicle and ego doesn't have to avoid it.

image

In this PR, I fixed the logic so that this module can check not only same direction lanes but also opposite lanes.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

- [x] Psim

Screenshot from 2024-08-05 10-10-32

- [x] [PASS TIER IV INTERNAL SCENARIOS](https://evaluation.tier4.jp/evaluation/reports/8d5e5dc1-79be-57da-b705-3a5fc7311121?project_id=prd_jt)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_path_planner/autoware_behavior_path_static_obstacle_avoidance_module/src/utils.cpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_static_obstacle_avoidance_module/src/utils.cpp
@@ -693,6 +693,15 @@ bool isNeverAvoidanceTarget(
       }
     }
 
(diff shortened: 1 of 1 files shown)


TomohitoAndo (approved):
LGTM

go-sakayori (approved):

PR #8348, opened 2024-08-05
feat(scenario_selector, freespace_planner): improve freespace planner edge case behavior

mkquda:
## Description

Currently, when operating in parking area, ego can get stuck in following cases:
- Start pose is in parking area, and goal pose is beyond occupancy grid region. Free space planner will fail to find trajectory and ego will be stuck

Screenshot from 2024-08-05 13-00-24

- Ego is following lanelet route to reach parking spot, and there is obstacle in the way. Ego will remain stopped until obstacle is gone.

Screenshot from 2024-08-05 13-01-46

For the first case, we should try to follow lanelet route to get closer to destination instead of waiting for free space planner to find impossible trajectory.

For second case, if destination is close enough we should try to run free space planner, instead of waiting for obstacle to go away.

Additionally, in freespace planner, when an obstacle is detected on the current trajectory, it will immediately stop and trigger replanning of path, in some cases the obstacle detection is momentary and fluctuates due to changes in the occupancy grid configuration. For smoother behavior it is better to introduce some tolerance for obstacle detection, only if obstacle is consistently detected on current trajectory for some time, then stop and replan.

## Changes

Refactor freespace_planner subscribers

Add time threshold for detecting obstacle on trajectory in freespace_planner:
- Add new parameter `th_obstacle_time_sec` to freespace planner params
- When obstacle is found on trajectory, timer is set
- If duration exceeds `th_obstacle_time_sec` then stop and replan
- If no obstacle is found on path timer is reset.

Implement scenario switching feature in scenario_selector: (when is parking area)
- Add new parameter `enable_mode_switching` to enable/disable feature
- If current scenario is PARKING, AND there is no trajectory from freespace_planner for some time, AND ego is in lane, THEN switch to LANE_DRIVING
- If current scenario is LANE_DRIVING, AND is autonomous, AND ego is stopped for sometime, THEN switch to PARKING

## How was this PR tested?

Test on psim:
- Switching to LANE_DRIVING when ego is in lane and freespace_planner fails to find trajectory

https://github.com/user-attachments/assets/78a33d94-18ac-4315-a2a6-39027900543f

- Switching to PARKING, when ego is in parking area and stops for some time while following lanelet route

https://github.com/user-attachments/assets/34d9cff8-22a5-4dc8-a85b-396101ce2c21

## Notes for reviewers

None.

## Interface changes

None.

#### Additions and removals

| Added | `enable_mode_switching`   | `bool` | `True`         | Flag to enable/disable scenario switching feature in parking area|
| Added | `th_obstacle_time_sec`   | `bool` | `True`         | Time threshold for checking obstacle on trajectory|

## Effects on system behavior

- When goal pose is beyond occupancy grid boundaries, and ego is in lane, ego will follow lanelet route to get closer to destination instead of waiting for impossible freespace planner trajectory

- When ego is following lanelet route inside parking area, and stops for obstacle, after some time activate free space planner to reach goal instead of waiting for obstacle to go away

- Freespace navigation is smoother near objects and stop and go motion is reduced


maxime-clem (approved):
the code looks good and I tested the new behavior which looks very nice.

takayuki5168 (approved):

PR #8350, opened 2024-08-05
fix(tensorrt_common): fix unreadVariable

kobayu858:
## Description

```
common/tensorrt_common/src/tensorrt_common.cpp:140:27: style: Variable 'engine_path' is assigned a value that is never used. [unreadVariable]
  std::string engine_path = model_file_path_;
                          ^

common/tensorrt_common/src/tensorrt_common.cpp:196:17: style: Variable 'engine_path' is assigned a value that is never used. [unreadVariable]
    engine_path = cache_engine_path;
                ^
```
## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


kobayu858:
It is actually used here, but only under certain conditions, which is why this warning occurs.
We have suppressed it with a comment because we have determined that it is a false positive.

If you have any comments, please let me know.
```
#if (NV_TENSORRT_MAJOR * 1000) + (NV_TENSORRT_MINOR * 100) + NV_TENSOR_PATCH >= 8200
  // Write profiles for trt-engine-explorer
  // See: https://github.com/NVIDIA/TensorRT/tree/main/tools/experimental/trt-engine-explorer
  std::string j_ext = ".json";
  fs::path json_path{engine_path};
  json_path.replace_extension(j_ext);
  std::string ret = getLayerInformation(nvinfer1::LayerInformationFormat::kJSON);
  std::ofstream os(json_path, std::ofstream::trunc);
  os << ret << std::flush;
#endif
```

manato (approved):
Thank you for fixing to improve the quality. Since I have no concerns about this PR, I approve it.

PR #8357, opened 2024-08-05
refactor(lane_change): check start point directly after getting start point

zulfaqar-azmi-t4:
## Description

Currently checking lane change start point validity is done in the `sample_lat_acc` loop.
This is slightly inefficient, since it can be perform just once after lane change start point is computed.

This PR aims to refactor this by performing the validity check after getting the lane change start point. 

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

PSIM

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/scene.cpp
```diff
@@ -1424,6 +1417,12 @@ bool NormalLaneChange::getLaneChangePaths(
         continue;
       }
 
+      if (!is_valid_start_point(common_data_ptr_, prepare_segment.points.back().point.pose)) {
+        debug_print(
+          "Reject: lane changing start point are not within the preferred lanes or its neighbors");
```
maxime-clem, line 1422: (nitpick)
```suggestion
          "Reject: lane changing start point is not within the preferred lanes or its neighbors");
```

maxime-clem (approved):
Looks good to me :+1:

mkquda (approved):

PR #8361, opened 2024-08-05
fix(autoware_path_optimizer): fix unreadVariable

kobayu858:
## Description

```
planning/autoware_path_optimizer/src/mpt_optimizer.cpp:1437:16: style: Variable 'A_rows_end' is assigned a value that is never used. [unreadVariable]
    A_rows_end += N_u;
               ^
```
## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


kobayu858:
This variable was removed because it was clearly not being used.

planning/autoware_path_optimizer/src/mpt_optimizer.cpp
```diff
@@ -1433,8 +1433,6 @@ MPTOptimizer::ConstraintMatrix MPTOptimizer::calcConstraintMatrix(
       lb(A_rows_end + i) = ref_steer_angle - mpt_param_.max_steer_rad;
       ub(A_rows_end + i) = ref_steer_angle + mpt_param_.max_steer_rad;
     }
-
-    A_rows_end += N_u;
```
takayuki5168, line 1437: This part is creating constraints for the optimization and `A_rows_end` remembers the size of constraints.
When we add a new constraint in the latter part, the removed line is important to know the size of constraints.

Isn't it possible to keep the current implementation even though it is useless currently.

kobayu858: Thanks for the confirmation.

If it is left intentionally, it can be suppressed in the comments and we will change it that way.

takayuki5168: Thank you.

takayuki5168 (approved):

PR #8362, opened 2024-08-05
fix(lane_change): skip generating path if lane changing path is too long

zulfaqar-azmi-t4:
## Description

Rework lane changing length check.

Current method is a little bit confusing, since we're using s_start and s_end.
This method aims to help improve the readability and reuse calculate distance to terminal end function.

Also current method doesn't check if the lane change end point is in between target neighbor, and this potentially caused lane changing path to be longer than expected and may cause unnecessary checks as well.

This PR methods ensures that the lane changing distance remains sufficient and the lane changing path remains in between target neighbor.

check lane changing path length

## Related links

Required PR: https://github.com/autowarefoundation/autoware.universe/pull/8359

**Parent Issue:**

- Link

## How was this PR tested?

using PSIM

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


maxime-clem (approved):
code looks good and a quick test in Psim did not reveal any problem.

PR #8363, opened 2024-08-05
fix(lane_change): skip generating path if longitudinal distance difference is less than threshold

zulfaqar-azmi-t4:
## Description

**Current behavior:**

The number of lane change  samples increases as ego vehicle approaches terminal start, due to the greater number of longitudinal acceleration samples.

The increase in lane change samples also leads to higher computational costs, as each sample generates a candidate path and undergoes various validations.

**Proposed behavior:**

Each lane change sample is computed based on the corresponding longitudinal acceleration sample.

As the ego vehicle approaches the terminal start, the distance differences between each lane change sample become smaller. 

In this case, if the distance difference is small and the previous path is unsafe, it's likely that the new path with a small distance difference will also be unsafe. Therefore, it is feasible to skip generating paths with small distance differences and focus on those with larger differences instead.

Following are the illustrations of the **proposed behavior**

prepare_diff_th

lc_diff_th

**Before PR**

Has more number of candidate path (5 samples) and processing time is between 7~10ms. 

https://github.com/user-attachments/assets/17b3b757-57ef-432d-9668-6c80e0f0f43c

**After PR**

Has less number of candidate path (3 sample) and processing time is between 5~8ms. 

https://github.com/user-attachments/assets/10d88bad-07f4-454d-a7cd-15475303f551

## Related links
Required https://github.com/autowarefoundation/autoware.universe/pull/8362
Required launcher: https://github.com/autowarefoundation/autoware_launch/pull/1108
**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/config/lane_change.param.yaml
+++ b/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/config/lane_change.param.yaml
@@ -26,6 +26,11 @@
       min_longitudinal_acc: -1.0
       max_longitudinal_acc: 1.0
 
+      skip_process:
+        longitudinal_distance_diff_threshold:
+          prepare: 0.5
+          lane_changing: 0.5
+
       # safety check
       safety_check:
         allow_loose_check_for_cancel: true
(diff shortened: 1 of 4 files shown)


maxime-clem (approved):
Code looks good and a quick test in Psim did not reveal any issue.

PR #8365, opened 2024-08-05
fix(autoware_behavior_path_goal_planner_module): fix unreadVariable

kobayu858:
## Description

```
planning/behavior_path_planner/autoware_behavior_path_goal_planner_module/src/manager.cpp:156:26: style: Variable 'ns' is assigned a value that is never used. [unreadVariable]
    const std::string ns = base_ns + "pull_over.parallel_parking.";
                         ^
```
## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


kobayu858:
This variable was removed because it was clearly not being used.

soblin (approved):

PR #8369, opened 2024-08-06
feat(lane_change): ensure LC merging lane stop point is safe

mkquda:
## Description
In case of LC merging lane pattern, the width of the lane at the stop point is not guaranteed to be sufficient for ego vehicle. If LC path is not approved and ego continues up to stop point, ego body can exceed lane boundaries which is unsafe.

Screenshot from 2024-08-05 17-14-23

## Changes
- added function `isMergingLane` in planner common utils to check for merging lane pattern
- added function `getDistanceFromLastFitWidthToEnd` in planner common utils to compute distance from last center line point within width threshold to lane end
- refactor function `getSignedDistanceFromBoundary` in planner common utils
- in case of LC merging lane pattern, subtract computed distance from `distance_to_terminal` to ensure safe stopping point

## How was this PR tested?
tested with psim.

Screenshot from 2024-08-07 10-49-30

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

When ego is approaching end of merging lane and LC path is not approved, ego should stop at safe point where ego body does not exceed lane boundaries


planning/behavior_path_planner/autoware_behavior_path_planner_common/src/utils/utils.cpp
```diff
@@ -580,6 +580,14 @@ bool isEgoWithinOriginalLane(
   return true;  // inside polygon
 }
 
+bool isMergingLane(const lanelet::ConstLanelet & lane)
+{
+  const auto right_boundy_last_point = lane.rightBound2d().back().basicPoint();
+  const auto left_boundy_last_point = lane.rightBound2d().back().basicPoint();
```
zulfaqar-azmi-t4, line 586: this should be `leftBound2d()` rigt?

planning/behavior_path_planner/autoware_behavior_path_planner_common/src/utils/utils.cpp
```diff
@@ -580,6 +580,14 @@ bool isEgoWithinOriginalLane(
   return true;  // inside polygon
 }
 
+bool isMergingLane(const lanelet::ConstLanelet & lane)
+{
```
zulfaqar-azmi-t4, line 584: Probably it's safer to add a guard here to check the size of the lane bounds.

planning/behavior_path_planner/autoware_behavior_path_planner_common/src/utils/utils.cpp
```diff
@@ -816,25 +844,30 @@ PathPointWithLaneId insertStopPoint(const double length, PathWithLaneId & path)
...
-    RCLCPP_ERROR_STREAM(
-      rclcpp::get_logger("behavior_path_planner").get_child("utils"),
-      "closest shoulder lanelet not found.");
+    return getSignedDistanceFromLaneBoundary(closest_lanelet, pose.position, left_side);
   }
 
-  return arc_coordinates.distance;
+  RCLCPP_ERROR_STREAM(
+    rclcpp::get_logger("behavior_path_planner").get_child("utils"),
+    "closest shoulder lanelet not found.");
```
maxime-clem, line 868: is this function limited to shoulder lanelets ? :thinking:

planning/behavior_path_planner/autoware_behavior_path_planner_common/src/utils/utils.cpp
```diff
@@ -648,6 +656,26 @@ double getDistanceToEndOfLane(const Pose & current_pose, const lanelet::ConstLan
...
+double getDistanceFromLastFitWidthToEnd(
+  const lanelet::ConstLanelet & lane, const double width_threshold)
+{
+  const auto center_line = lane.centerline3d().basicLineString();
+  double distance = 0.0;
+  if (center_line.size() <= 1) {
+    return distance;
+  }
+  auto it = center_line.rbegin() + 1;
+  for (; it < center_line.rend(); ++it) {
```
maxime-clem, line 668: Is it possible for the centerline to only contain 2 points ? In which case this code does not work well and interpolation is required to get an accurate distance.

planning/behavior_path_planner/autoware_behavior_path_planner_common/src/utils/utils.cpp
mkquda: I don't think it is, but this is not my addition, it was already like that, so wasn't sure if I should change it ...
Even the input is called `shoulder_lanelets`

planning/behavior_path_planner/autoware_behavior_path_planner_common/src/utils/utils.cpp
mkquda: Modified implementation to check distance between points and interpolate if necessary

planning/behavior_path_planner/autoware_behavior_path_planner_common/src/utils/utils.cpp
```diff
@@ -648,6 +648,56 @@ double getDistanceToEndOfLane(const Pose & current_pose, const lanelet::ConstLan
...
+  constexpr double step_length{1.0};
+  const double seg_length = autoware::universe_utils::calcDistance2d(point, next_point);
+  if (seg_length <= step_length) return distance;
+
+  double yaw = atan2(next_point.y - point.y, next_point.x - point.x);
+  double length = step_length;
+  for (; length < seg_length; length += step_length) {
+    auto mid_point = point;
+    mid_point.x += length * cos(yaw);
+    mid_point.y += length * sin(yaw);
```
maxime-clem, line 693: I don't think you need to calculate the yaw here. You can just interpolate between the two points using `universe_utils::calcInterpolatedPoint` (https://github.com/autowarefoundation/autoware.universe/blob/main/common/autoware_universe_utils/include/autoware/universe_utils/geometry/geometry.hpp#L472).

Alternatively, how about resampling the centerline with a fixed interval before looping over its points ?

mkquda: I don't think its necessary to resample the entire centerline, its more efficient to interpolate the segment where condition applies

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/include/autoware/behavior_path_lane_change_module/utils/calculation.hpp
```diff
@@ -40,6 +40,23 @@ double calc_ego_dist_to_terminal_end(const CommonDataPtr & common_data_ptr);
...
+ *
+ * @param lanelets current ego lanelets
+ * @param src_pose source pose to calculate distance from
+ * @param common_param common parameters used in behavior path planner.
+ * @param margin additional margin for checking lane width
+ * @return distance to last fit width position along the lane
+ */
+double calc_dist_to_last_fit_width(
+  const lanelet::ConstLanelets lanelets, const Pose & src_pose,
+  const BehaviorPathPlannerParameters & common_param, const double margin = 0.1);
```
zulfaqar-azmi-t4, line 59: I would suggest using `bpp_param` instead to avoid confusion with `common_param_ptr`  
```suggestion
  const lanelet::ConstLanelets lanelets, const Pose & src_pose,
  const BehaviorPathPlannerParameters & bpp_param, const double margin = 0.1);
```

planning/behavior_path_planner/autoware_behavior_path_planner_common/src/utils/utils.cpp
```diff
@@ -559,8 +559,8 @@ bool isEgoWithinOriginalLane(
   const lanelet::ConstLanelets & current_lanes, const Pose & current_pose,
   const BehaviorPathPlannerParameters & common_param, const double outer_margin)
 {
-  const auto lane_length = lanelet::utils::getLaneletLength2d(current_lanes);
-  const auto lane_poly = lanelet::utils::getPolygonFromArcLength(current_lanes, 0, lane_length);
+  const auto combined_lane = lanelet::utils::combineLaneletsShape(current_lanes);
```
zulfaqar-azmi-t4, line 562: using `common_data_ptr->lanes_polygon_ptr->current` will allow us to save some time and skip this part. because creating polygon is a bit expensive

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/utils/calculation.cpp
```diff
@@ -42,4 +44,47 @@ double calc_dist_from_pose_to_terminal_end(
...
+double calc_dist_to_last_fit_width(
+  const lanelet::ConstLanelets lanelets, const Pose & src_pose,
+  const BehaviorPathPlannerParameters & common_param, const double margin)
+{
+  if (lanelets.empty()) return 0.0;
+
+  const auto lane_polygon = lanelets.back().polygon2d().basicPolygon();
+  const auto center_line = lanelet::utils::generateFineCenterline(lanelets.back(), 1.0);
+  universe_utils::LineString2d line_string;
+  line_string.reserve(center_line.size() - 1);
```
maxime-clem, line 57: are we sure that `center_line`  is not empty ? If we are not sure please add a check to avoid doing `0U - 1`.

planning/behavior_path_planner/autoware_behavior_path_planner_common/src/utils/utils.cpp
mkquda: @zulfaqar-azmi-t4 this function is in bpp common utils not LC utils

zulfaqar-azmi-t4: Yes, can always make another overloading function, i.e, one with similar arguments, and another takes polygon2d as input.

maxime-clem (approved):

zulfaqar-azmi-t4 (approved):
LGTM. Thank you!

xmfcx:
Once this was merged, we started having test errors in the CI:
https://github.com/autowarefoundation/autoware.universe/actions/runs/10416334778

> 1 package had test failures: autoware_freespace_planning_algorithms

https://github.com/autowarefoundation/autoware.universe/actions/runs/10416334778/job/28848368965#step:16:17822

 @mkquda could you look into it?

xmfcx:
This ci failed last week too:
https://github.com/autowarefoundation/autoware.universe/actions/runs/10314646396

PR #8371, opened 2024-08-06
fix(bluetooth_monitor): fix unreadVariable

kobayu858:
## Description

```
system/bluetooth_monitor/src/bluetooth_monitor.cpp:146:13: style: Variable 'level' is assigned a value that is never used. [unreadVariable]
  int level = DiagStatus::OK;
            ^
```
## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


ito-san (approved):
LGTM

PR #8373, opened 2024-08-06
fix(reaction_analyzer): fix unreadVariable

kobayu858:
## Description

```
tools/reaction_analyzer/src/utils.cpp:225:27: style: Variable 'current_distance' is assigned a value that is never used. [unreadVariable]
  double current_distance = 0.0;
                          ^
```
## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


kobayu858:
@brkay54 

brkay54 (approved):
No need to define `current_distance` before the loop. LGTM. Thanks.

brkay54 (changes requested):
Sorry for the inconvenience, it seems there is another PR which proposed almost same changes: 
- https://github.com/autowarefoundation/autoware.universe/pull/8450

making `const double` is better so IMO we can close this PR.

veqcc:
@brkay54 
Thank you for the reply!! I will colse this PR instead of @kobayu858 ans merge https://github.com/autowarefoundation/autoware.universe/pull/8450

PR #7120, opened 2024-05-24
feat(tier4_autoware_utils): faster sin and cos

veqcc:
## Description

I have implemented a faster sin ans cos calculation.
Calling `sin_and_cos` is faster than calling `sin` and `cos` separately.

It is useful, for example, for `DistortionCorrector`.
I found that around 35% of CPU time in DistortionCorrector is occupied by `sin` and `cos`,
and I have confirmed 10~20% latency improvement with faster `sin_and_cos`. cc @vividf 

## Tests performed

I have written a test code and the test has passed.

## Pre-review checklist for the PR author

The PR author **must** check the checkboxes below when creating the PR.

- [x] I've confirmed the [contribution guidelines].
- [x] The PR follows the [pull request guidelines].

## In-review checklist for the PR reviewers

The PR reviewers **must** check the checkboxes below before approval.

- [ ] The PR follows the [pull request guidelines].

## Post-review checklist for the PR author

The PR author **must** check the checkboxes below before merging.

- [ ] There are no open discussions or they are tracked via tickets.

After all checkboxes are checked, anyone who has write access can merge the PR.

[contribution guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/
[pull request guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/pull-request-guidelines/

PR #8380, opened 2024-08-06
feat(autoware_pointcloud_preprocessor): distortion corrector node update azimuth and distance

### LINKED ISSUE #3896: distortion_corrector should update "azimuth" and "distance" field

vividf:
## Description
This PR solves the issue https://github.com/autowarefoundation/autoware.universe/issues/3896.
Note that this PR is tested with the nebula PR https://github.com/tier4/nebula/pull/180 

- The node updates the per-point azimuth and distance values based on the undistorted XYZ coordinates when the input point cloud is in the sensor frame (not in the `base_link`) and the `update_azimuth_and_distance` parameter is set to `true`. The azimuth values are calculated using the `cv::fastAtan2` function. 
- Please note that updating the azimuth and distance fields increases the execution time by approximately 20%. Additionally, due to the `cv::fastAtan2` algorithm's accuracy of about 0.3 degrees, there is a **possibility of changing beam order for high azimuth resolution LiDAR**.
- LiDARs from different vendors have different azimuth coordinates. Currently, the coordinate systems listed below have been tested, and the node will update the azimuth based on the input coordinate system.
  - `velodyne`: (x: 0 degrees, y: 270 degrees)
  - `hesai`: (x: 90 degrees, y: 0 degrees)
  - `others`: (x: 0 degrees, y: 90 degrees) and (x: 270 degrees, y: 0 degrees)

> before distortion correction (Velodyne): 
> x: -7.27784, y: 1.06172, z: xxxxx, azimuth: 3.28645, distance: 7.356
> after distortion correction:
> x: -7.2696, y: 0.767085, z: xxxxx, azimuth: 3.24671, distance: 7.311

More information in https://github.com/vividf/autoware.universe/blob/feature/distortion_corrector_node_update_azimuth_and_distance/sensing/autoware_pointcloud_preprocessor/docs/distortion-corrector.md.

Additionally, this PR also removes the separate sin and cos by using the sin_and_cos function which can speed up the processing time (from this https://github.com/autowarefoundation/autoware.universe/pull/7120)

## Related links
Previous closed PR: https://github.com/autowarefoundation/autoware.universe/pull/5560

**Parent Issue:**

- Link

## How was this PR tested?
- Tested with Unit test.
```
colcon test --packages-select autoware_pointcloud_preprocessor --event-handlers console_cohesion+
```

- Tested with sample rosbag.
[TIER4_INTERNAL_LINK](https://drive.google.com/drive/folders/16Ip-xR0INuzYZUW2k3aZwYKf3DIeBrgT?usp=drive_link)

1. change the `input/pointcloud` to ` /sensing/lidar/top/pointcloud_raw_ex` in the launch file
2. change the parameter `update_azimuth_and_distance` to `true` in `distortion_corrector_node.param.yaml`
3. `ros2 launch autoware_pointcloud_preprocessor distortion_corrector_node.launch.xml` and `ros2 topic hz /sensing/lidar/top/rectified/pointcloud_ex`
4. play the provided ros bag

Time comparison
Input topic: **/sensing/lidar/top/pointcloud_raw_ex** (number of points is larger than the /sensing/lidar/top/mirror_cropped/pointcloud_ex)
bag: sample rosbag
frame: 121

- Before this PR

|   | Minimum | Maximum | Average |
|-----------|---------|---------|---------|
| Time     | 7.763    | 15.318   | 10.82   |

- After this PR

1. with new sin_and_cos optimization
6. **without** update azimuth and distance

|   | Minimum | Maximum | Average |
|-----------|---------|---------|---------|
| Time     | 7.038    | 14.375   | 10.28   |

1. with new sin_and_cos optimization
2. **with** update azimuth and distance

|   | Minimum | Maximum | Average |
|-----------|---------|---------|---------|
| Time     | 9.078    | 19.907   | 12.33   |

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


sensing/autoware_pointcloud_preprocessor/src/distortion_corrector/distortion_corrector.cpp
```diff
@@ -258,6 +260,25 @@ void DistortionCorrector<T>::undistortPointCloud(
     // Undistort a single point based on the strategy
     undistortPoint(it_x, it_y, it_z, it_twist, it_imu, time_offset, is_twist_valid, is_imu_valid);
 
+    if (!sensor_azimuth_coordinate.empty() && pointcloudTransformNeeded()) {
+      // Input frame should be in the sensor frame.
+      *it_distance = sqrt(*it_x * *it_x + *it_y * *it_y + *it_z * *it_z);
+      float cartesian_coordinate_azimuth = cv::fastAtan2(*it_y, *it_x);
```
yukkysaito, line 266: Here’s the translation:

I'd like to avoid introducing a dependency on OpenCV just for this function. Do you have any good suggestions? If computational cost is the issue, ideally, we could implement a fast atan2 here.
https://github.com/autowarefoundation/autoware.universe/blob/main/common/autoware_universe_utils/include/autoware/universe_utils/math/trigonometry.hpp

sensing/autoware_pointcloud_preprocessor/src/distortion_corrector/distortion_corrector.cpp
```diff
@@ -258,6 +260,25 @@ void DistortionCorrector<T>::undistortPointCloud(
...
+      if (sensor_azimuth_coordinate == "velodyne") {
+        *it_azimuth = (360 - cartesian_coordinate_azimuth) * M_PI / 180;
+      } else if (sensor_azimuth_coordinate == "hesai") {
+        *it_azimuth = (90 - cartesian_coordinate_azimuth) < 0
+                        ? (90 - cartesian_coordinate_azimuth + 360) * M_PI / 180
+                        : (90 - cartesian_coordinate_azimuth) * M_PI / 180;
+      } else {
+        throw std::runtime_error(
+          sensor_azimuth_coordinate + " azimuth coordinate is not supported");
+      }
```
yukkysaito, line 276: Why is this necessary for each vendor? In Autoware, we prioritize designing modules to be adaptive, so we want to avoid vendor-specific processing in the code of this node as much as possible.

vividf: Thanks for your comment, because the definition of azimuth depends on each vendor like the image below. 
image

reference: https://github.com/vividf/autoware.universe/blob/feature/distortion_corrector_node_update_azimuth_and_distance/sensing/autoware_pointcloud_preprocessor/docs/distortion-corrector.md

vividf: Do you have any suggestions for this? Thanks!

sensing/autoware_pointcloud_preprocessor/src/distortion_corrector/distortion_corrector.cpp
vividf: Understand! I will implement the fast atan2 there!

sensing/autoware_pointcloud_preprocessor/src/distortion_corrector/distortion_corrector.cpp
yukkysaito: @vividf I see 👍 Thank you for the quick response. I think it would be better to handle the differences in azimuth definitions within the Nebula decoder so that the preprocessor node doesn't have to account for these differences. What do you think?

vividf: @yukkysaito Thanks for your advice! I talked about this with @drwnz in today's meeting.

sensing/autoware_pointcloud_preprocessor/src/distortion_corrector/distortion_corrector.cpp
vividf: @yukkysaito  I move the fast atan2 function to trigonometry.cpp/hpp and provide unit test for the function.

sensing/autoware_pointcloud_preprocessor/src/distortion_corrector/distortion_corrector.cpp
vividf: @yukkysaito 
In the end, we decided to implement a conversion formula to adapt to different azimuth coordinate systems for this node.
Additionally, there is a big chance that we will also unify the azimuth in Nebula

vividf:
Time analysis is updated in the description!

vividf:
@yukkysaito kindly ping

common/autoware_universe_utils/src/math/trigonometry.cpp
```diff
@@ -72,4 +72,34 @@ std::pair<float, float> sin_and_cos(float radian)
...
+    c2 = c * c;
+    a = 90.f - (((atan2_p7 * c2 + atan2_p5) * c2 + atan2_p3) * c2 + atan2_p1) * c;
+  }
+  if (dx < 0) a = 180.f - a;
+  if (dy < 0) a = 360.f - a;
+
+  a = a * autoware::universe_utils::pi / 180.f;
+  return a;
+}
+
```
yukkysaito, line 104: Is license clear? no need the license?

sensing/autoware_pointcloud_preprocessor/src/distortion_corrector/distortion_corrector.cpp
```diff
@@ -187,15 +188,99 @@ bool DistortionCorrector<T>::isInputValid(sensor_msgs::msg::PointCloud2 & pointc
...
+    // Check if 'b' can be adjusted to 1 or -1
+    if (std::abs(b - 1.0f) <= threshold_b_) {
+      b_ = 1.0f;
+    } else if (std::abs(b + 1.0f) <= threshold_b_) {
+      b_ = -1.0f;
+    } else {
+      RCLCPP_WARN(
+        node_->get_logger(),
+        "Angle between two points exceeds 180 degrees. Iterate to next point ...");
+      std::cout << "b: " << b << "should be close to 1 or -1" << std::endl;
```
yukkysaito, line 243: Can you replace to rclcpp_foo?

yukkysaito:
@vividf @drwnz 
Sorry for the delay in responding. Could you please designate a person from the Sensing team to verify the algorithm's validity, as I haven't had enough time to fully review it?

sensing/autoware_pointcloud_preprocessor/src/distortion_corrector/distortion_corrector.cpp
yukkysaito: @vividf Thank you for your response. Which part of the code are you referring to?

sensing/autoware_pointcloud_preprocessor/src/distortion_corrector/distortion_corrector.cpp
vividf: Thanks for the check! fixed in https://github.com/autowarefoundation/autoware.universe/pull/8380/commits/f509d7d84d5fa92aeb7e6357da3446c803383270

common/autoware_universe_utils/src/math/trigonometry.cpp
vividf: @yukkysaito @drwnz 
Thanks, could you help me check whether this is enough https://github.com/autowarefoundation/autoware.universe/pull/8380/commits/62434faf959d7324e0d826f4e2b44ee84a3faa1d

sensing/autoware_pointcloud_preprocessor/src/distortion_corrector/distortion_corrector.cpp
vividf: @yukkysaito 
This function calculate the conversion formula https://github.com/autowarefoundation/autoware.universe/blob/a3c977b5b7772cb6e83e183e80ec9c9b2894f48e/sensing/autoware_pointcloud_preprocessor/src/distortion_corrector/distortion_corrector.cpp#L192

common/autoware_universe_utils/src/math/trigonometry.cpp
mojomex: Please rename the license such that it is clear that it is for OpenCV. For example, 
* put it in a directory called `third_party_licenses` and name it `opencv-license.md`.
* then, comment where the license can be found in the doc comment of the function (link to original repo and path to the license copy in Autoware).
* finally, add a comment to pointcloud_preprocessor's README.md to explain where third-party licenses can be found (reference the directory created above)

mojomex:
@yukkysaito 
> Could you please designate a person from the Sensing team to verify the algorithm's validity, as I haven't had enough time to fully review it?

I'll take over!

sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/distortion_corrector/distortion_corrector_node.hpp
```diff
@@ -52,6 +52,9 @@ class DistortionCorrectorComponent : public rclcpp::Node
   std::string base_frame_;
   bool use_imu_;
   bool use_3d_distortion_correction_;
+  bool update_azimuth_and_distance_;
+
+  bool can_update_azimuth_and_distance_{false};
```
mojomex, line 57: Does this need to be in `distortion_corrector_node.hpp` as opposed to `distortion_corrector.hpp`? If it was in the latter, it could be combined with the struct for `a_` and `b_` into an `std::optional<AngleConversion>`.

sensing/autoware_pointcloud_preprocessor/src/distortion_corrector/distortion_corrector_node.cpp
```diff
@@ -91,9 +92,27 @@ void DistortionCorrectorComponent::onPointCloud(PointCloud2::UniquePtr pointclou
...
+      RCLCPP_INFO(
+        this->get_logger(),
+        "Success to get the conversion formula between cartesian coordinates and LiDAR azimuth "
+        "coordinates");
+    } else {
+      RCLCPP_ERROR(
+        this->get_logger(),
+        "Failed to get the conversion formula between cartesian coordinates and LiDAR azimuth "
+        "coordinates. Please check the LiDAR azimuth coordinate. This pointcloud will not update "
+        "azimuth and distance");
```
mojomex, line 110: Stopping execution here is imho better because the user configured the node wrong. Trying to continue execution imo is not particularly meaningful.

mojomex (commented):
I haven't quite finished the bigger parts of this PR (like the actual algorithm :sweat_smile:) but I have a few smaller code style related comments, so I'll post them here in the meantime.

sensing/autoware_pointcloud_preprocessor/launch/distortion_corrector_node.launch.xml
```diff
@@ -1,5 +1,5 @@
 <launch>
-  <arg name="input/pointcloud" default="/sensing/lidar/top/mirror_cropped/pointcloud_ex"/>
+  <arg name="input/pointcloud" default="/sensing/lidar/top/pointcloud_raw_ex"/>
```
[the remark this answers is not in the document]
vividf: By using the `/sensing/lidar/top/mirror_cropped/pointcloud_ex`, the azimuth will not update since it is in the `base_link` instead of the `sensor_frame`. (This is mentioned in the Readme)
However, I think I should switch it back to the original one. Thanks!

common/autoware_universe_utils/test/src/math/test_trigonometry.cpp
```diff
@@ -50,3 +50,28 @@ TEST(trigonometry, sin_and_cos)
...
   }
 }
+
+float normalize_angle(double angle)
+{
+  if (angle < 0) {
+    return angle + 2 * autoware::universe_utils::pi;
+  }
+  return angle;
+}
```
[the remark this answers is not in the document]
vividf: ah, because the output of the std::atan2 is restricted to the [-2 * PI; 2 * PI)

sensing/autoware_pointcloud_preprocessor/src/distortion_corrector/distortion_corrector.cpp
```diff
@@ -187,15 +188,98 @@ bool DistortionCorrector<T>::isInputValid(sensor_msgs::msg::PointCloud2 & pointc
...
+    // If the angle exceeds 180 degrees, it may cross the 0-degree axis,
+    // which could disrupt the calculation of the formula.
+    if (
+      abs(*next_it_azimuth - *it_azimuth) >= autoware::universe_utils::pi ||
+      abs(next_cartesian_rad - current_cartesian_rad) >= autoware::universe_utils::pi) {
+      RCLCPP_WARN(
+        node_->get_logger(),
+        "Angle between two points exceeds 180 degrees. Iterate to next point ...");
+      continue;
+    }
```
mojomex, line 229: There should be four explicit cases (both azimuth and cartesian_rad less than PI, bigger/equal to PI, or the first or latter is less and the other is bigger/equal).

Handling these cases explicitly (should only have effects on the sign of `b` and sign/offset of `a`) would get rid of the necessity of looping until the 'easy' case is found.

sensing/autoware_pointcloud_preprocessor/src/distortion_corrector/distortion_corrector.cpp
```diff
@@ -187,15 +188,98 @@ bool DistortionCorrector<T>::isInputValid(sensor_msgs::msg::PointCloud2 & pointc
   return true;
 }
 
+template <class T>
+bool DistortionCorrector<T>::azimuthConversionExists(sensor_msgs::msg::PointCloud2 & pointcloud)
```
mojomex, line 192: This naming is a bit misleading: it suggests that the existence of a conversion is checked and returned, but in reality the conversion is computed **and causes a side effect** (storing values in `a_` and `b_`).

This function would be better if renamed to something like `tryComputeAndSetAzimuthConversion`.
Even better, it would not set `a_` and `b_` directly but return a `std::optional<AngleTransform>` and be called `tryComputeAzimuthConversion`.

mojomex (changes requested):

mojomex (changes requested):
Hi, here's the rest of my review comments.
Functionality-wise, the only things I found is that
* if the first two points have the same azimuth, the algorithm will not work
* for the 180 deg difference handling, the runtime guarantees of the algorithm could be lowered from `O(n)` to `O(1)` in the worst case by implementing the four possible cases by hand, instead of searching for the easy case

Other than that, there are a few smaller style and maintainability comments. Thanks for your work!

common/autoware_universe_utils/test/src/math/test_trigonometry.cpp
```diff
@@ -50,3 +50,28 @@ TEST(trigonometry, sin_and_cos)
...
+
+TEST(trigonometry, opencv_fastAtan2)
+{
+  for (int i = 0; i < 100; ++i) {
+    // Generate random x and y between -10 and 10
+    float x = static_cast<float>(std::rand()) / RAND_MAX * 20.0 - 10.0;
+    float y = static_cast<float>(std::rand()) / RAND_MAX * 20.0 - 10.0;
```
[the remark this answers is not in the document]
vividf: Thanks! added in the https://github.com/autowarefoundation/autoware.universe/pull/8380/commits/f0609f51bde7f99f34237fcf2bc21178020d8d4a

sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/distortion_corrector/distortion_corrector.hpp
```diff
@@ -103,16 +116,21 @@ class DistortionCorrector : public DistortionCorrectorBase
...
     const geometry_msgs::msg::TwistWithCovarianceStamped::ConstSharedPtr twist_msg) override;
 
   void processIMUMessage(
     const std::string & base_frame, const sensor_msgs::msg::Imu::ConstSharedPtr imu_msg) override;
-  void undistortPointCloud(bool use_imu, sensor_msgs::msg::PointCloud2 & pointcloud) override;
+  void undistortPointCloud(
+    bool use_imu, bool update_azimuth_and_distance,
+    sensor_msgs::msg::PointCloud2 & pointcloud) override;
+  bool azimuthConversionExists(sensor_msgs::msg::PointCloud2 & pointcloud) override;
+  std::tuple<float, float> getConversion() override;
```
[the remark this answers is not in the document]
vividf: Thanks! fixed in https://github.com/autowarefoundation/autoware.universe/pull/8380/commits/0a1f4c90eb5ce5b96660abb40e8a70c59f887055

sensing/autoware_pointcloud_preprocessor/src/distortion_corrector/distortion_corrector_node.cpp
vividf: True, add the runtime error in https://github.com/autowarefoundation/autoware.universe/pull/8380/commits/7e1f510b89f2b0425547cef9ff85b228f1f62437. 
Thanks a lot!

sensing/autoware_pointcloud_preprocessor/launch/distortion_corrector_node.launch.xml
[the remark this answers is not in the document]
vividf: fixed in https://github.com/autowarefoundation/autoware.universe/pull/8380/commits/b156fb59cee6536173178bd15941e4f8d7fa0fe6

sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/distortion_corrector/distortion_corrector.hpp
```diff
@@ -79,6 +84,14 @@ class DistortionCorrector : public DistortionCorrectorBase
   std::deque<geometry_msgs::msg::TwistStamped> twist_queue_;
   std::deque<geometry_msgs::msg::Vector3Stamped> angular_velocity_queue_;
 
+  // Equation for the conversion between sensor azimuth coordinates and Cartesian coordinates:
+  // sensor azimuth coordinates = a + b * cartesian coordinates;
+  // a is restricted to be a multiple of 90, and b is restricted to be 1 or -1.
+  float a_{0};
+  float b_{1};
```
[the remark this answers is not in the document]
vividf: thanks! fixed in https://github.com/autowarefoundation/autoware.universe/pull/8380/commits/f11e5295e2523844282672822be2c4bd888a2bc7

sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/distortion_corrector/distortion_corrector.hpp
```diff
@@ -79,6 +84,14 @@ class DistortionCorrector : public DistortionCorrectorBase
...
+  float threshold_a_{0.087f};  // 5 / 180 * M_PI
+  float threshold_b_{0.1f};
```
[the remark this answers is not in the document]
vividf: thanks! fixed in https://github.com/autowarefoundation/autoware.universe/pull/8380/commits/f11e5295e2523844282672822be2c4bd888a2bc7

common/autoware_universe_utils/src/math/trigonometry.cpp
mojomex: See [Apache Software Foundation's licensing how-to](https://infra.apache.org/licensing-howto.html)

vividf: @mojomex Thanks for the advice!
I added them in https://github.com/autowarefoundation/autoware.universe/pull/8380/commits/b9a5c56436665dfbc79a9e6fde6e7e93ee3e97f8. 
However, the pre-commit modifies the layout of the license in https://github.com/autowarefoundation/autoware.universe/pull/8380/commits/b9a5c56436665dfbc79a9e6fde6e7e93ee3e97f8. Not sure about whether this influence :(

common/autoware_universe_utils/test/src/math/test_trigonometry.cpp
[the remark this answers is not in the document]
vividf: To make it more robust, I fixed it in https://github.com/autowarefoundation/autoware.universe/pull/8380/commits/5e55d016a266de4177830d4d2cdc794e5c864f5a. Thanks!

sensing/autoware_pointcloud_preprocessor/src/distortion_corrector/distortion_corrector.cpp
```diff
@@ -187,15 +188,98 @@ bool DistortionCorrector<T>::isInputValid(sensor_msgs::msg::PointCloud2 & pointc
...
+    }
+
+    // Limit the range of a in [0, 360]
+    if (multiple_of_90_degrees < 0) {
+      multiple_of_90_degrees += 4;
+    } else if (multiple_of_90_degrees > 4) {
+      multiple_of_90_degrees -= 4;
+    }
+
+    a_ = multiple_of_90_degrees * (autoware::universe_utils::pi / 2);
```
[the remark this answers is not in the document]
vividf: I changed it to https://github.com/autowarefoundation/autoware.universe/pull/8380/commits/02beb86b0ac1e45b0bc70b3ce76df8714604e69c!

sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/distortion_corrector/distortion_corrector_node.hpp
vividf: fixed in https://github.com/autowarefoundation/autoware.universe/pull/8380/commits/13d981f70a5fa22fd07a8f7dfa302420a1700a6e, https://github.com/autowarefoundation/autoware.universe/pull/8380/commits/8a24a79457c1204a21b0a398d90a4ad5185ea82e, https://github.com/autowarefoundation/autoware.universe/pull/8380/commits/02b12177da2092ae663a3c27b0bccca07a3876c2

sensing/autoware_pointcloud_preprocessor/src/distortion_corrector/distortion_corrector_node.cpp
vividf: Thanks! fixed in https://github.com/autowarefoundation/autoware.universe/pull/8380/commits/1d5ccdb1c8b6717ca20bd527f298b218dcdb3be5

common/autoware_universe_utils/src/math/trigonometry.cpp
```diff
@@ -72,4 +72,42 @@ std::pair<float, float> sin_and_cos(float radian)
...
+    a = (((atan2_p7 * c2 + atan2_p5) * c2 + atan2_p3) * c2 + atan2_p1) * c;
+  } else {
+    c = ax / (ay + atan2_DBL_EPSILON);
+    c2 = c * c;
+    a = 90.f - (((atan2_p7 * c2 + atan2_p5) * c2 + atan2_p3) * c2 + atan2_p1) * c;
+  }
+  if (dx < 0) a = 180.f - a;
+  if (dy < 0) a = 360.f - a;
+
+  a = a * autoware::universe_utils::pi / 180.f;
```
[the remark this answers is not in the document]
vividf: Thanks! fixed in https://github.com/autowarefoundation/autoware.universe/pull/8380/commits/427a0ce2019d10eb346e34f9b0a50224a1415ecd

sensing/autoware_pointcloud_preprocessor/schema/distortion_corrector_node.schema.json
```diff
@@ -21,13 +21,24 @@
           "description": "Use 3d distortion correction algorithm, otherwise, use 2d distortion correction algorithm.",
           "default": "false"
         },
+        "update_azimuth_and_distance": {
+          "type": "boolean",
+          "description": "Update the azimuth and distance value of each points.",
```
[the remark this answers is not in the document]
vividf: regarding this "will the node reject inputs that have azimuth/distance?", the input distortion corrector node has checked that the input pointcloud needs to be in the type of XYZIRCAEDT

sensing/autoware_pointcloud_preprocessor/test/test_distortion_corrector_node.cpp
```diff
@@ -170,7 +171,9 @@ class DistortionCorrectorTest : public ::testing::Test
   }
 
   sensor_msgs::msg::PointCloud2 generatePointCloudMsg(
-    bool generate_points, bool is_lidar_frame, rclcpp::Time stamp)
+    bool generate_points, bool is_lidar_frame, std::string vendor, rclcpp::Time stamp,
```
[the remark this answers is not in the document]
vividf: Thanks! fixed in https://github.com/autowarefoundation/autoware.universe/pull/8380/commits/53f41627c1f8ad1f5d7ff650cef191ecb8f379f1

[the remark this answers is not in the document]
vividf: Unit test file has big modification, probably it is easier to look at the file again :)

sensing/autoware_pointcloud_preprocessor/test/test_distortion_corrector_node.cpp
```diff
@@ -170,7 +171,9 @@ class DistortionCorrectorTest : public ::testing::Test
...
+    bool use_default_pointcloud, std::vector<Eigen::Vector3f> defined_points,
+    std::vector<float> defined_azimuths)
   {
```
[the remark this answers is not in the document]
vividf: Thanks! fixed in https://github.com/autowarefoundation/autoware.universe/pull/8380/commits/a1f076c0429c5cdf4ac16890941c72f9c9a2f96e and  https://github.com/autowarefoundation/autoware.universe/pull/8380/commits/242ab45094d82b2cd6003c8f4d63a2a044bb1250

sensing/autoware_pointcloud_preprocessor/src/distortion_corrector/distortion_corrector.cpp
```diff
@@ -187,15 +188,98 @@ bool DistortionCorrector<T>::isInputValid(sensor_msgs::msg::PointCloud2 & pointc
...
+  auto next_it_azimuth = it_azimuth;
+
+  if (it_x != it_x.end() && it_x + 1 != it_x.end()) {
+    next_it_x = it_x + 1;
+    next_it_y = it_y + 1;
+    next_it_azimuth = it_azimuth + 1;
+  } else {
+    RCLCPP_WARN(
+      node_->get_logger(),
+      "Current point cloud only has a single point. Could not calculate the formula.");
```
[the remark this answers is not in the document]
vividf: fixed in https://github.com/autowarefoundation/autoware.universe/pull/8380/commits/80ad8e21af43bcfef446ae67cfb5f18c32f53595

sensing/autoware_pointcloud_preprocessor/src/distortion_corrector/distortion_corrector.cpp
```diff
@@ -187,15 +188,98 @@ bool DistortionCorrector<T>::isInputValid(sensor_msgs::msg::PointCloud2 & pointc
...
+
+    // If the angle exceeds 180 degrees, it may cross the 0-degree axis,
+    // which could disrupt the calculation of the formula.
+    if (
+      abs(*next_it_azimuth - *it_azimuth) >= autoware::universe_utils::pi ||
+      abs(next_cartesian_rad - current_cartesian_rad) >= autoware::universe_utils::pi) {
+      RCLCPP_WARN(
+        node_->get_logger(),
+        "Angle between two points exceeds 180 degrees. Iterate to next point ...");
+      continue;
```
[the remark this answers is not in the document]
vividf: Thanks! I change them to debug!

sensing/autoware_pointcloud_preprocessor/src/distortion_corrector/distortion_corrector.cpp
```diff
@@ -187,15 +188,98 @@ bool DistortionCorrector<T>::isInputValid(sensor_msgs::msg::PointCloud2 & pointc
...
+    // Check if 'b' can be adjusted to 1 or -1
+    if (std::abs(b - 1.0f) <= threshold_b_) {
+      b_ = 1.0f;
+    } else if (std::abs(b + 1.0f) <= threshold_b_) {
+      b_ = -1.0f;
+    } else {
+      RCLCPP_WARN(
+        node_->get_logger(),
+        "Angle between two points exceeds 180 degrees. Iterate to next point ...");
+      continue;
```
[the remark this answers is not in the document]
vividf: Same :)  Changed to debug and change the description!

sensing/autoware_pointcloud_preprocessor/src/distortion_corrector/distortion_corrector.cpp
vividf: Thanks! fixed in https://github.com/autowarefoundation/autoware.universe/pull/8380/commits/8a24a79457c1204a21b0a398d90a4ad5185ea82e

sensing/autoware_pointcloud_preprocessor/test/test_distortion_corrector_node.cpp
```diff
@@ -180,21 +183,54 @@ class DistortionCorrectorTest : public ::testing::Test
...
+          Eigen::Vector3f(0.0f, -10.0f, 3.0f),  // point 3
+          Eigen::Vector3f(-5.0f, -5.0f, 4.0f),  // point 4
+          Eigen::Vector3f(-10.0f, 0.0f, 5.0f),  // point 5
+          Eigen::Vector3f(-5.0f, 5.0f, -5.0f),  // point 6
+          Eigen::Vector3f(0.0f, 10.0f, -4.0f),  // point 7
+          Eigen::Vector3f(5.0f, 5.0f, -3.0f),   // point 8
+          Eigen::Vector3f(8.0f, 3.0f, -2.0f),   // point 9
+          Eigen::Vector3f(9.0f, 1.0f, -1.0f)    // point 10
+        }};
+
```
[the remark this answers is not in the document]
vividf: Thanks! I add the test case of (0, 0, 0) in https://github.com/autowarefoundation/autoware.universe/pull/8380/commits/a8532934917f9c6b77882c2831cd8efe60aac560

vividf:
@mojomex 
Big thanks for your detailed review!!!
I fixed the all of the comments you left, please let me know if there are other problems :)

common/autoware_universe_utils/src/math/trigonometry.cpp
```diff
@@ -72,4 +72,42 @@ std::pair<float, float> sin_and_cos(float radian)
...
+// 2. output of the function is changed from degrees to radians.
+static const float atan2_p1 =
+  0.9997878412794807f * static_cast<float>(180) / autoware::universe_utils::pi;
+static const float atan2_p3 =
+  -0.3258083974640975f * static_cast<float>(180) / autoware::universe_utils::pi;
+static const float atan2_p5 =
+  0.1555786518463281f * static_cast<float>(180) / autoware::universe_utils::pi;
+static const float atan2_p7 =
+  -0.04432655554792128f * static_cast<float>(180) / autoware::universe_utils::pi;
+static const float atan2_DBL_EPSILON = 2.2204460492503131e-016f;
```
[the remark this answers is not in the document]
vividf: Thanks! fixed in https://github.com/autowarefoundation/autoware.universe/pull/8380/commits/427a0ce2019d10eb346e34f9b0a50224a1415ecd

sensing/autoware_pointcloud_preprocessor/test/test_distortion_corrector_node.cpp
```diff
@@ -180,21 +183,54 @@ class DistortionCorrectorTest : public ::testing::Test
...
+            float cartesian_deg =
+              std::atan2(point.y(), point.x()) * 180 / autoware::universe_utils::pi;
+            if (cartesian_deg < 0) cartesian_deg += 360;
+            float hesai_deg =
+              90 - cartesian_deg < 0 ? 90 - cartesian_deg + 360 : 90 - cartesian_deg;
+            if (hesai_deg == 360) hesai_deg = 0;
+            default_azimuths.push_back(hesai_deg * autoware::universe_utils::pi / 180);
+          } else {  // empty string
+            default_azimuths.push_back(std::atan2(point.y(), point.x()));
+          }
```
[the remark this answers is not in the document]
vividf: Thanks! Added the comments :)

common/autoware_universe_utils/NOTICE
```diff
@@ -0,0 +1,2 @@
+The function 'opencv_fast_atan2' in /autoware/src/universe/autoware.universe/common/autoware_universe_utils/src/math/trigonometry.cpp
+was modified by function 'fastAtan2' in OpenCV project. (https://github.com/opencv/opencv/blob/4.x/modules/core/src/mathfuncs_core.simd.hpp)
```
mojomex, line 2:
```suggestion
is a modified version of 'fastAtan2' in the OpenCV project. (https://github.com/opencv/opencv/blob/4.x/modules/core/src/mathfuncs_core.simd.hpp)
```

sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/distortion_corrector/distortion_corrector.hpp
```diff
@@ -43,17 +43,29 @@
...
+struct AngleConversion
+{
+  // Equation for the conversion between sensor azimuth coordinates and Cartesian coordinates:
+  // sensor azimuth coordinates = offset_rad + sign * cartesian coordinates;
+  // offset_rad is restricted to be a multiple of 90, and sign is restricted to be 1 or -1.
+  float offset_rad{0};
+  float sign{1};
+  float offset_rad_threshold{0.087f};  // (5 / 180) * M_PI
+  float sign_threshold{0.1f};
+};
```
mojomex, line 60: These can be mate `static const` or `static constexpr` so they will still be semantically part of `AngleConversion` but they are not mutable, and they don't show up in the constructor.

common/autoware_universe_utils/src/math/trigonometry.cpp
```diff
@@ -72,6 +72,19 @@ std::pair<float, float> sin_and_cos(float radian)
   }
 }
 
+// This code is modified from part of OpenCV project
+// (https://github.com/opencv/opencv/blob/4.x/modules/core/src/mathfuncs_core.simd.hpp). It is
+// subject to the license terms in the LICENSE file found in the top-level directory of this
+// distribution and at http://opencv.org/license.html.
```
mojomex, line 78: The file is not in the top level directory but in `common/autoware_universe_utils/third_party_licenses/opencv-license.md`

sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/distortion_corrector/distortion_corrector.hpp
```diff
@@ -48,13 +48,24 @@
...
+
 class DistortionCorrectorBase
 {
 public:
   virtual bool pointcloudTransformExists() = 0;
   virtual bool pointcloudTransformNeeded() = 0;
-  virtual std::deque<geometry_msgs::msg::TwistStamped> get_twist_queue() = 0;
-  virtual std::deque<geometry_msgs::msg::Vector3Stamped> get_angular_velocity_queue() = 0;
+  virtual std::deque<geometry_msgs::msg::TwistStamped> getTwistQueue() = 0;
```
mojomex, line 67: As a general comment, the current Autoware coding style mandates `snake_case`, not `camelBack`. Please adhere to that style :pray:
* [.clang-tidy](https://github.com/autowarefoundation/autoware/blob/949a69e49ae8a0fb7d710080e5b613d926e13e21/.clang-tidy#L438-L439)
* [Autoware coding guidelines](https://autowarefoundation.github.io/autoware-documentation/main/contributing/coding-guidelines/languages/cpp/#use-lower-snake-case-for-function-names-required-partially-automated)

sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/distortion_corrector/distortion_corrector_node.hpp
```diff
@@ -54,7 +54,9 @@ class DistortionCorrectorComponent : public rclcpp::Node
   bool use_3d_distortion_correction_;
   bool update_azimuth_and_distance_;
 
-  bool can_update_azimuth_and_distance_{false};
+  std::optional<AngleConversion> angle_conversion_opt_;
+  int angle_conversion_failure_num_{0};
+  int failure_tolerance_{20};
```
mojomex, line 59: Please doc-comment `/// @brief` these two member variables:
why are failures accepted --> in case the first few pointclouds are not containing enough points to calculate formula

Make `failure_tolerance_` `static const` or `static constexpr`

sensing/autoware_pointcloud_preprocessor/src/distortion_corrector/distortion_corrector.cpp
```diff
@@ -189,9 +188,15 @@ bool DistortionCorrector<T>::isInputValid(sensor_msgs::msg::PointCloud2 & pointc
 }
 
 template <class T>
-bool DistortionCorrector<T>::azimuthConversionExists(sensor_msgs::msg::PointCloud2 & pointcloud)
+std::optional<AngleConversion> DistortionCorrector<T>::tryComputeAngleConversion(
+  sensor_msgs::msg::PointCloud2 & pointcloud)
 {
-  if (!isInputValid(pointcloud)) return false;
+  // This function try to compute the angle conversion from Cartesian coordinates to LiDAR azimuth
```
mojomex, line 194:
```suggestion
  // This function tries to compute the angle conversion from Cartesian coordinates to LiDAR azimuth
```
Also, isn't it lidar to cartesian?

sensing/autoware_pointcloud_preprocessor/src/distortion_corrector/distortion_corrector.cpp
```diff
@@ -208,73 +213,81 @@ bool DistortionCorrector<T>::azimuthConversionExists(sensor_msgs::msg::PointClou
   } else {
     RCLCPP_WARN(
```
mojomex, line 214: Use throttle there

sensing/autoware_pointcloud_preprocessor/src/distortion_corrector/distortion_corrector.cpp
```diff
@@ -208,73 +213,81 @@ bool DistortionCorrector<T>::azimuthConversionExists(sensor_msgs::msg::PointClou
...
     // which could disrupt the calculation of the formula.
     if (
-      abs(*next_it_azimuth - *it_azimuth) >= autoware::universe_utils::pi ||
-      abs(next_cartesian_rad - current_cartesian_rad) >= autoware::universe_utils::pi) {
-      RCLCPP_WARN(
-        node_->get_logger(),
-        "Angle between two points exceeds 180 degrees. Iterate to next point ...");
+      abs(*next_it_azimuth - *it_azimuth) == 0 ||
+      abs(next_cartesian_rad - current_cartesian_rad) == 0) {
+      RCLCPP_DEBUG(
```
mojomex, line 230: Use throttle here

sensing/autoware_pointcloud_preprocessor/src/distortion_corrector/distortion_corrector.cpp
```diff
@@ -208,73 +213,81 @@ bool DistortionCorrector<T>::azimuthConversionExists(sensor_msgs::msg::PointClou
...
+    if (std::abs(sign - 1.0f) <= angle_conversion.sign_threshold) {
+      angle_conversion.sign = 1.0f;
+    } else if (std::abs(sign + 1.0f) <= angle_conversion.sign_threshold) {
+      angle_conversion.sign = -1.0f;
     } else {
-      RCLCPP_WARN(
-        node_->get_logger(),
-        "Angle between two points exceeds 180 degrees. Iterate to next point ...");
-      std::cout << "b: " << b << "should be close to 1 or -1" << std::endl;
+      RCLCPP_DEBUG(
```
mojomex, line 252: Use throttle here

sensing/autoware_pointcloud_preprocessor/src/distortion_corrector/distortion_corrector.cpp
```diff
@@ -208,73 +213,81 @@ bool DistortionCorrector<T>::azimuthConversionExists(sensor_msgs::msg::PointClou
...
-    // Check if 'a' can be adjusted to a multiple of π/2
-    int multiple_of_90_degrees = std::round(a / (autoware::universe_utils::pi / 2));
-    if (std::abs(a - multiple_of_90_degrees * (autoware::universe_utils::pi / 2)) > threshold_a_) {
+    float offset_rad = *it_azimuth - sign * current_cartesian_rad;
+    // Check if 'offset_rad' can be adjusted to offset_rad multiple of π/2
+    int multiple_of_90_degrees = std::round(offset_rad / (autoware::universe_utils::pi / 2));
+    if (
+      std::abs(offset_rad - multiple_of_90_degrees * (autoware::universe_utils::pi / 2)) >
+      angle_conversion.offset_rad_threshold) {
+      RCLCPP_DEBUG(
```
mojomex, line 263: Use throttle here

sensing/autoware_pointcloud_preprocessor/src/distortion_corrector/distortion_corrector_node.cpp
```diff
@@ -94,25 +96,29 @@ void DistortionCorrectorComponent::onPointCloud(PointCloud2::UniquePtr pointclou
...
+      RCLCPP_WARN_STREAM_THROTTLE(
+        this->get_logger(), *this->get_clock(), 10000 /* ms */,
+        "Failed to get the angle conversion between Cartesian coordinates and LiDAR azimuth "
+        "coordinates. This pointcloud will not update azimuth and distance");
+      angle_conversion_failure_num_++;
+
+      if (angle_conversion_failure_num_ > failure_tolerance_) {
+        throw std::runtime_error(
+          "Angle conversion failed more than " + std::to_string(failure_tolerance_) +
+          " times. The node has been interrupted. Please check the LiDAR azimuth coordinates.");
```
mojomex, line 116: Because an exception will be thrown, the node stopping does not need to be mentioned separately.
```suggestion
          " times. Please check the LiDAR azimuth coordinates.");
```

sensing/autoware_pointcloud_preprocessor/test/test_distortion_corrector_node.cpp
```diff
@@ -170,10 +171,71 @@ class DistortionCorrectorTest : public ::testing::Test
...
+    empty_pointcloud_msg.header.frame_id = "lidar_top";
+    empty_pointcloud_msg.height = 1;
+    empty_pointcloud_msg.is_dense = true;
+    empty_pointcloud_msg.is_bigendian = false;
+    empty_pointcloud_msg.width = 0;
+    empty_pointcloud_msg.row_step = 0;
+
+    return empty_pointcloud_msg;
+  }
+
```
mojomex, line 235: I think calling `generatePointCloudMsg(true, stamp, {}, {});` will result in the same empty pointcloud, so this function will not be needed. As it is now, the ~10 lines setting the header fields are duplicated.

mojomex (changes requested):
Thanks for addressing the feedback! There's a couple more small things I found, but it's already looking pretty good :raised_hands:

sensing/autoware_pointcloud_preprocessor/test/test_distortion_corrector_node.cpp
vividf: thanks, fixed in https://github.com/autowarefoundation/autoware.universe/pull/8380/commits/320e9610d1dd43eeffc04bdfb3accd17009bd97a

sensing/autoware_pointcloud_preprocessor/src/distortion_corrector/distortion_corrector.cpp
vividf: all of the logging are fixed in https://github.com/autowarefoundation/autoware.universe/pull/8380/commits/03354c85ca8ec274c2360466078429d7d782d2e9

sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/distortion_corrector/distortion_corrector_node.hpp
vividf: Thanks!!! fixed in https://github.com/autowarefoundation/autoware.universe/pull/8380/commits/550a6466f4fe034c7f45ecbaa3f14f20951c3a82

sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/distortion_corrector/distortion_corrector.hpp
vividf: Thanks, I changed all of them to snake case in https://github.com/autowarefoundation/autoware.universe/pull/8380/commits/f6be6374320c33f9a3bea6e6f0f58fd9cc010d44

mojomex (changes requested):
A few small last changes :pray:

Please also address spell check failing on `Testtry` etc. Consider renaming the functions to `Test_try` or `TestTry` etc.

sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/distortion_corrector/distortion_corrector.hpp
```diff
@@ -55,14 +67,18 @@ class DistortionCorrectorBase
...
     const geometry_msgs::msg::TwistWithCovarianceStamped::ConstSharedPtr twist_msg) = 0;
-  virtual void processIMUMessage(
+  virtual void process_imu_message(
     const std::string & base_frame, const sensor_msgs::msg::Imu::ConstSharedPtr imu_msg) = 0;
-  virtual void setPointCloudTransform(
+  virtual void set_pointcloud_transform(
     const std::string & base_frame, const std::string & lidar_frame) = 0;
   virtual void initialize() = 0;
-  virtual void undistortPointCloud(bool use_imu, sensor_msgs::msg::PointCloud2 & pointcloud) = 0;
+  virtual std::optional<AngleConversion> try_compute_angle_conversion(
```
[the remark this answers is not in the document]
vividf: Thanks!!!
I refactored most of the functions from DistortionCorrector to DistortionCorrectorBase in https://github.com/autowarefoundation/autoware.universe/pull/8380/commits/896d843b0f527a5b203f4baacd1e3cc214e57638.

Only the CRTP method is in the DistortionCorrector class now. I felt it is cleaner now :)

mojomex (commented):
The code looks very good now, thanks!!
I tried running it with the sample rosbag from the logging sim tutorial but am constantly getting the following error, along with a pointcloud that jumps or hangs for a frame sometimes:
```
[distortion_corrector_node-1] [1727355131.361596872 WARN] [distortion_corrector_node]: Twist time_stamp is too late. Could not interpolate.
```
Could you tell me what rosbag to use or what to do to set it up correctly?

Here's my commands:
```shell
# All in separate terminals
ros2 launch autoware_pointcloud_preprocessor distortion_corrector_node.launch.xml
ros2 bag play -l ~/autoware_map/sample-rosbag/ -s sqlite3 --remap /sensing/lidar/top/velodyne_packets:=/velodyne_packets
ros2 launch nebula_ros nebula_launch.py sensor_model:=VLS128 launch_hw:=false
ros2 run tf2_ros static_transform_publisher 0 0 0 0 0 0 base_link velodyne
```

As for processing time, I got similar numbers to the ones you stated. My maximum was around 21ms. I guess this module would be perfect to parallelize/gpu-ize in the future.

vividf:
@mojomex 
Could you try with the `ros2 launch autoware_launch logging_simulator.launch.xml map_path:=$HOME/autoware_map/sample-map-rosbag vehicle_model:=sample_vehicle sensor_model:=sample_sensor_kit`. 

From your commands, it seems that you didn't run the vehicle velocity converter and the topic `sensing/vehicle_velocity_converter/twist_with_covariance` is missed. 

I could give you the rosbag that I recorded before that includes that topic on Monday :)

mojomex:
@veqcc CPPcheck [failed](https://github.com/autowarefoundation/autoware.universe/actions/runs/11097207598/job/30828078728?pr=8380) on a file that has not been touched by this PR. How to best handle such a case?

veqcc:
> CPPcheck [failed](https://github.com/autowarefoundation/autoware.universe/actions/runs/11097207598/job/30828078728?pr=8380) on a file that has not been touched by this PR. How to best handle such a case?

Sorry for the inconvenience.
This problem is fixed in https://github.com/autowarefoundation/autoware.universe/pull/8978, so please merge it into your branch!!

mojomex (commented):
It works well for me, thanks! Below are videos to confirm (top: input, bottom: output).
In the first half, I color by `distance`, in the second half by `azimuth`.


https://github.com/user-attachments/assets/6d829d05-aa24-4504-b45d-41458e792f25


Because I'm a noob, I first ran distortion corrector on a pointcloud not in sensor frame, in chich case it failed silently.
Please add a warning/error when a user attempts to do that :bow:

vividf:
@mojomex
Thanks for the check! I added a runtime error to notify the user if they set things wrong :) 
https://github.com/autowarefoundation/autoware.universe/pull/8380/commits/72f5bb264d7ad0f66607f2a75fd4bab42e0009ab

sensing/autoware_pointcloud_preprocessor/src/distortion_corrector/distortion_corrector.cpp
```diff
@@ -368,28 +368,27 @@ void DistortionCorrector<T>::undistort_pointcloud(
...
-        *it_distance = sqrt(*it_x * *it_x + *it_y * *it_y + *it_z * *it_z);
-
-        ++it_azimuth;
-        ++it_distance;
-      } else {
+      if (!pointcloud_transform_needed_) {
         throw std::runtime_error(
           "The pointcloud is not in the sensor frame, thus it will not update azimuth and distance "
           "values. Please change the input pointcloud or set update_azimuth_and_distance_ to "
           "false.");
```
mojomex, line 375:
```suggestion
          "The pointcloud is not in the sensor's frame and thus azimuth and distance cannot be updated. " 
          "Please change the input pointcloud or set update_azimuth_and_distance to false.");
```

mojomex (approved):
LGTM!

mojomex:
@vividf Sorry, applying suggestions through GitHub seems to have wrecked DCO: [it's failing](https://github.com/autowarefoundation/autoware.universe/pull/8380/checks?check_run_id=31012760538).
Could you rebase as suggested in that failing check? :bow:

common/autoware_universe_utils/src/math/trigonometry.cpp
```diff
@@ -72,4 +72,57 @@ std::pair<float, float> sin_and_cos(float radian)
   }
 }
 
+// This code is modified from part of OpenCV project
```
knzo25, line 75: grammar

sensing/autoware_pointcloud_preprocessor/docs/distortion-corrector.md
```diff
@@ -48,3 +48,24 @@ ros2 launch autoware_pointcloud_preprocessor distortion_corrector.launch.xml
 
 - The node requires time synchronization between the topics from lidars, twist, and IMU.
 - If you want to use a 3D distortion corrector without IMU, please check that the linear and angular velocity fields of your twist message are not empty.
+- The node updates the per-point azimuth and distance values based on the undistorted XYZ coordinates when the input point cloud is in the sensor frame (not in the `base_link`) and the `update_azimuth_and_distance` parameter is set to `true`. The azimuth values are calculated using the `cv::fastAtan2` function.
```
knzo25, line 51:
```suggestion
- The node updates the per-point azimuth and distance values based on the undistorted XYZ coordinates when the input point cloud is in the sensor frame (not in the `base_link`) and the `update_azimuth_and_distance` parameter is set to `true`. The azimuth values are calculated using a modified version of OpenCV's `cv::fastAtan2` function.
```

sensing/autoware_pointcloud_preprocessor/docs/distortion-corrector.md
```diff
@@ -48,3 +48,24 @@ ros2 launch autoware_pointcloud_preprocessor distortion_corrector.launch.xml
...
+- Please note that updating the azimuth and distance fields increases the execution time by approximately 20%. Additionally, due to the `cv::fastAtan2` algorithm's accuracy of about 0.3 degrees, there is a **possibility of changing beam order for high azimuth resolution LiDAR**.
```
knzo25, line 52: Can you state clearly if 0.3 is the maximum error?
If possible, put the average and maximum error

knzo25: changing *the* beam order

otherwise it could mean of a "changing beam order", as in a property of the device

sensing/autoware_pointcloud_preprocessor/src/distortion_corrector/distortion_corrector_node.cpp
```diff
@@ -92,10 +93,32 @@ void DistortionCorrectorComponent::onPointCloud(PointCloud2::UniquePtr pointclou
...
+    } else {
+      RCLCPP_WARN_STREAM_THROTTLE(
+        this->get_logger(), *this->get_clock(), 10000 /* ms */,
+        "Failed to get the angle conversion between Cartesian coordinates and LiDAR azimuth "
+        "coordinates. This pointcloud will not update azimuth and distance");
+      angle_conversion_failure_num_++;
+
+      if (angle_conversion_failure_num_ > failure_tolerance) {
+        throw std::runtime_error(
+          "Angle conversion failed more than " + std::to_string(failure_tolerance) +
```
knzo25, line 115: While I get the feeling, I do not think throwing here is good for our projects (imagine if we throw in the middle of an experiment/production).
In my opinion, we could send an error and fallback to the default case while being verbose about this fact.

Also, I do not believe the use of a tolerance is warranted. For this cases, it works or not, right?

sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/distortion_corrector/distortion_corrector_node.hpp
```diff
@@ -52,12 +52,23 @@ class DistortionCorrectorComponent : public rclcpp::Node
...
+  bool update_azimuth_and_distance_;
+
+  std::optional<AngleConversion> angle_conversion_opt_;
+
+  /// @brief Counts the number of pointclouds for which angle conversion failed.
+  int angle_conversion_failure_num_{0};
+
+  /// @brief The maximum number of failed angle conversions due to pointclouds not containing enough
+  /// points before throwing an error.
+  static constexpr int failure_tolerance{20};
```
knzo25, line 64: I left a comment below about this point

sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/distortion_corrector/distortion_corrector.hpp
```diff
@@ -47,74 +47,95 @@
...
-  virtual void setPointCloudTransform(
-    const std::string & base_frame, const std::string & lidar_frame) = 0;
-  virtual void initialize() = 0;
-  virtual void undistortPointCloud(bool use_imu, sensor_msgs::msg::PointCloud2 & pointcloud) = 0;
+  // Equation for the conversion between sensor azimuth coordinates and Cartesian coordinates:
+  // sensor azimuth coordinates = offset_rad + sign * cartesian coordinates;
+  // offset_rad is restricted to be a multiple of 90, and sign is restricted to be 1 or -1.
+  float offset_rad{0};
+  float sign{1};
+  static constexpr float offset_rad_threshold{0.087f};  // (5 / 180) * M_PI
```
knzo25, line 57: out of ignorance, can't you write `(5 / 180) * M_PI` in a constexpr?

sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/distortion_corrector/distortion_corrector.hpp
```diff
@@ -47,74 +47,95 @@
...
-  rclcpp::Node * node_;
   std::unique_ptr<autoware::universe_utils::ManagedTransformBuffer> managed_tf_buffer_{nullptr};
 
   std::deque<geometry_msgs::msg::TwistStamped> twist_queue_;
   std::deque<geometry_msgs::msg::Vector3Stamped> angular_velocity_queue_;
 
-  void getIMUTransformation(const std::string & base_frame, const std::string & imu_frame);
-  void enqueueIMU(const sensor_msgs::msg::Imu::ConstSharedPtr imu_msg);
-  void getTwistAndIMUIterator(
+  rclcpp::Node * node_;
```
knzo25, line 73: can you make this a pointer const?
`rclcpp::Node * const node_;`

sensing/autoware_pointcloud_preprocessor/src/distortion_corrector/distortion_corrector_node.cpp
```diff
@@ -92,10 +93,32 @@ void DistortionCorrectorComponent::onPointCloud(PointCloud2::UniquePtr pointclou
     return;
   }
 
-  distortion_corrector_->setPointCloudTransform(base_frame_, pointcloud_msg->header.frame_id);
-
+  distortion_corrector_->set_pointcloud_transform(base_frame_, pointcloud_msg->header.frame_id);
   distortion_corrector_->initialize();
-  distortion_corrector_->undistortPointCloud(use_imu_, *pointcloud_msg);
+
+  if (update_azimuth_and_distance_ && !angle_conversion_opt_.has_value()) {
```
knzo25, line 99: I kind of feel this block of code belongs inside the `distortion_corrector.cpp`.
For example inside of initialize.
This is just my personal opinion so feel free to ignore it (close this conversation in that case)

sensing/autoware_pointcloud_preprocessor/src/distortion_corrector/distortion_corrector.cpp
```diff
@@ -187,15 +175,139 @@ bool DistortionCorrector<T>::isInputValid(sensor_msgs::msg::PointCloud2 & pointc
...
+  }
+
+  if (is_imu_time_stamp_too_late) {
+    RCLCPP_WARN_STREAM_THROTTLE(
+      node_->get_logger(), *node_->get_clock(), 10000 /* ms */,
+      "IMU time_stamp is too late. Could not interpolate.");
+  }
+}
+
+void DistortionCorrectorBase::convert_matrix_to_transform(
```
knzo25, line 285: This is a common idiom here, but whenever possible (as in it does not increase the processing time), consider returning instead of modifying a reference

sensing/autoware_pointcloud_preprocessor/src/distortion_corrector/distortion_corrector.cpp
```diff
@@ -187,15 +175,139 @@ bool DistortionCorrector<T>::isInputValid(sensor_msgs::msg::PointCloud2 & pointc
...
 template <class T>
-void DistortionCorrector<T>::undistortPointCloud(
-  bool use_imu, sensor_msgs::msg::PointCloud2 & pointcloud)
+void DistortionCorrector<T>::undistort_pointcloud(
+  bool use_imu, std::optional<AngleConversion> angle_conversion_opt,
+  sensor_msgs::msg::PointCloud2 & pointcloud)
 {
-  if (!isInputValid(pointcloud)) return;
+  if (!is_pointcloud_valid(pointcloud)) return;
+  if (twist_queue_.empty()) {
```
knzo25, line 300: It seems you got the twist check outside the isInputValid because you needed a version for the pointcloud.
In that case you could have done them independently while also keeping the isInputValid

knzo25 (changes requested):
Just small change requests

sensing/autoware_pointcloud_preprocessor/src/distortion_corrector/distortion_corrector.cpp
vividf: True, thanks!
Fixed in https://github.com/autowarefoundation/autoware.universe/pull/8380/commits/b1ffa5e40cef1b554e990e2e0bc24d0660f49fa2

sensing/autoware_pointcloud_preprocessor/src/distortion_corrector/distortion_corrector.cpp
```diff
@@ -187,15 +175,139 @@ bool DistortionCorrector<T>::isInputValid(sensor_msgs::msg::PointCloud2 & pointc
...
+      abs(*next_it_azimuth - *it_azimuth) == 0 ||
+      abs(next_cartesian_rad - current_cartesian_rad) == 0) {
+      RCLCPP_DEBUG_STREAM_THROTTLE(
+        node_->get_logger(), *node_->get_clock(), 10000 /* ms */,
+        "Angle between two points is 0 degrees. Iterate to next point ...");
+      continue;
+    }
+
+    // restrict the angle difference between [-180, 180] (degrees)
+    float azimuth_diff = abs(*next_it_azimuth - *it_azimuth) > autoware::universe_utils::pi
```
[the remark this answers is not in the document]
vividf: Thanks for the catch, it's a mistake.
fixed in https://github.com/autowarefoundation/autoware.universe/pull/8380/commits/b1ffa5e40cef1b554e990e2e0bc24d0660f49fa2

sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/distortion_corrector/distortion_corrector.hpp
vividf: @mojomex suggested using reference instead of the pointer in the concatenate PR but I didn't change it here.
Thanks for the catch, I changed it to reference in https://github.com/autowarefoundation/autoware.universe/pull/8380/commits/2594e761149e02d244614b41e8674b5e5864e833.

@knzo25 putting as and const reference will not work because the current design of managed_tf_buffer

sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/distortion_corrector/distortion_corrector.hpp
vividf: Thanks! fixed in https://github.com/autowarefoundation/autoware.universe/pull/8380/commits/bbbc501bfac443c533ecd1057e07f6d8487de8e0

common/autoware_universe_utils/src/math/trigonometry.cpp
vividf: Thanks, fixed in https://github.com/autowarefoundation/autoware.universe/pull/8380/commits/57824efce82fcc9c1116d0d70d87600cff642b07

sensing/autoware_pointcloud_preprocessor/src/distortion_corrector/distortion_corrector_node.cpp
vividf: Understand your consideration, before we think it is better for the user to set the parameter to False after 2 seconds (if it is failed).
Remove the runtime error and set the logging from warning to error in https://github.com/autowarefoundation/autoware.universe/pull/8380/commits/4bc0df1ea17e80b390bca04ba2f66fd0df39afdb

sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/distortion_corrector/distortion_corrector_node.hpp
vividf: removed in https://github.com/autowarefoundation/autoware.universe/pull/8380/commits/4bc0df1ea17e80b390bca04ba2f66fd0df39afdb

sensing/autoware_pointcloud_preprocessor/docs/distortion-corrector.md
vividf: Fixed in https://github.com/autowarefoundation/autoware.universe/pull/8380/commits/96800274df0de221da68fa94607a50af805bfdd6.

From the opencv website (https://docs.opencv.org/3.4/db/de0/group__core__utils.html#ga7b356498dd314380a0c386b059852270) , I couldn't know the average error.

sensing/autoware_pointcloud_preprocessor/src/distortion_corrector/distortion_corrector.cpp
vividf: I think having both `is_input_valid` and `is_pointlcoud_valid` would be a bit confusing.
And in the `is_input_valid` there is only an additional check (3 lines) for checking the twist queue. Also, there is only one place that will call this function.

vividf:
This PR should be merged together with

- https://github.com/tier4/aip_launcher/pull/318
- https://github.com/autowarefoundation/sample_sensor_kit_launch/pull/105/files
- aw_sim and aw_sim_lab (TODO)

vividf:
@knzo25 kindly ping

knzo25 (approved):
LGTM :tada:

vividf:
@TakaHoribe @soblin @takayuki5168 
Seems that this PR need your approval since it include not only pointcloud preprocessor package :) 
Thanks

soblin (approved):

PR #8381, opened 2024-08-06
fix(lane_change): do not cancel when approaching terminal start

zulfaqar-azmi-t4:
## Description

Lane change module will cancel unsafe path unless ego has passed terminal start point.
However, as ego vehicle approaching terminal end, cancelling lane change path might result in difficulties in re-attempting lane change.

This PR aims to fix this, by not cancelling lane change if ego has already near terminal start.

## Related links

None

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/interface.cpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/interface.cpp
@@ -265,6 +265,11 @@ bool LaneChangeInterface::canTransitFailureState()
     return true;
   }
 
+  if (module_type_->is_near_terminal()) {
+    log_debug_throttled("Unsafe, but ego is approaching terminal. Continue lane change");
+    return false;
+  }
+
   if (module_type_->isCancelEnabled() && module_type_->isEgoOnPreparePhase()) {
     if (module_type_->isStoppedAtRedTrafficLight()) {
       log_debug_throttled("Stopping at traffic light while in prepare phase. Cancel lane change");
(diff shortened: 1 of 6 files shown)


planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/utils/calculation.cpp
```diff
@@ -42,4 +43,15 @@ double calc_dist_from_pose_to_terminal_end(
   }
   return utils::getDistanceToEndOfLane(src_pose, lanes);
 }
+
+double calc_backward_buffer(const LCParamPtr & lc_param_ptr)
```
maxime-clem, line 47: please add a comment or rename the function to make it clear it is calculating a "minimum stopping distance".

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/interface.cpp
```diff
@@ -265,6 +265,15 @@ bool LaneChangeInterface::canTransitFailureState()
     return true;
   }
 
+  if (module_type_->is_near_terminal()) {
+    log_debug_throttled("Unsafe, but ego is approaching terminal. Continue lane change");
+
+    if (module_type_->isRequiredStop(post_process_safety_status_.is_trailing_object)) {
+      log_debug_throttled("Module require stopping");
+    }
```
maxime-clem, line 273: This condition is only to add a log message, is this okay ?

maxime-clem (approved):
Code looks good. I only have a small question and a small request.

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/utils/calculation.cpp
zulfaqar-azmi-t4: Renamed in https://github.com/autowarefoundation/autoware.universe/pull/8381/commits/a27ae47ce217cddb0834fad2dc4cac7c51f61033

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/interface.cpp
zulfaqar-azmi-t4: Yes, this is okay.
We will address the stopping method in the redesign, so for now just adding log is fine.

PR #8214, opened 2024-07-26
refactor(lane_change): separate leading and trailing objects

zulfaqar-azmi-t4:
## Description

Separate lane change target objects to front and back in the filtering function  to avoid redoing the front/back check in the isLaneChangePathSafe function

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

PR #8383, opened 2024-08-06
fix(lane_change): unify stuck detection to avoid unnecessary computation

zulfaqar-azmi-t4:
## Description

With the current detection, there will necessary computation when ego vehicle is stuck.
In this case, lane change module first check if there is any safe path. If not safe path, then ego vehicle will attempt to generate lane change path with stuck rss.

In this PR, these steps are unified. Meaning if ego is stuck, in the same loop, ego first check if there is safe path with normal rss, and if no, then then rss param is used.

This helps to reduce computation time of lane change module.

## Related links

https://github.com/autowarefoundation/autoware.universe/pull/8214

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


maxime-clem (approved):
code looks good :+1:

PR #8404, opened 2024-08-08
fix(autoware_lane_departure_checker): not to show error message "trajectory deviation is too large" during manual driving

T-Kimura-MM:
## Description
Update not to show error message "trajectory deviation is too large" during manual driving.
This message is necessary only during autonomous driving.

## Related links

**Parent Issue:**

- related to [link](https://tier4.atlassian.net/wiki/spaces/~6061477cafe465006afac8aa/pages/3159721149/trajectory+deviation+is+too+large)

## How was this PR tested?
It is tested using planning simulator.
You can see error message at rqt_runtime_monitor.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


kyoichi-sugahara:
@T-Kimura-MM 
LGTM for the change and it makes sense to me.
Could you briefly explain how did you test the change?

T-Kimura-MM:
@kyoichi-sugahara 
Here is a way to conduct the test:

1. Launch the planning simulator.
2. Add a manual controller(https://github.com/autowarefoundation/autoware_tools/tree/main/common/tier4_control_rviz_plugin).
3. Use the manual controller to intentionally shift the route.
4. Confirm that no error message appears in this state.
5. Switch to autonomous driving after deviating from the route.
6. Confirm that a message appears when the route is deviated during autonomous driving.

kyoichi-sugahara (approved):
Thank you so much for explanation about how to test! LGTM!

PR #8407, opened 2024-08-08
fix(autoware_multi_object_tracker): output from screen to both

kminoda:
## Description

Set output from screen to both to improve the debugging efficiency of Autoware in general.
With this change, the output from node will not only be displayed in screen but also to the log file.

See https://github.com/ros2/launch/blob/986ea832cec49c18bee4e29ee4df37fe3844fa90/launch/launch/logging/__init__.py#L397-L487

## Related links

[TIER IV INTERNAL THREAD](https://star4.slack.com/archives/C024FRX4UNL/p1705576376841409)

## How was this PR tested?

Not tested

## Notes for reviewers

None

## Interface changes

None.

## Effects on system behavior

Although there will be a slight increase in computational load due to the additional log, I would say that the risk is minimal since the node's output does not occupy terminal at this moment. But just in case, I would like to make this change for several nodes chosen from perception component which tends to have some issues pretty often.


technolojin (approved):
LGTM

yukkysaito:
Please wait for the merge as it is under discussion here.
https://github.com/autowarefoundation/autoware.universe/pull/8411

yukkysaito:
Please wait for the merge as it is under discussion here.
https://github.com/autowarefoundation/autoware.universe/pull/8411

PR #8411, opened 2024-08-08
fix(autoware_traffic_light_map_based_detector): output from screen to both

kminoda:
## Description

Set output from screen to both to improve the debugging efficiency of Autoware in general.
With this change, the output from node will not only be displayed in screen but also to the log file.

See https://github.com/ros2/launch/blob/986ea832cec49c18bee4e29ee4df37fe3844fa90/launch/launch/logging/__init__.py#L397-L487

## Related links

[TIER IV INTERNAL THREAD](https://star4.slack.com/archives/C024FRX4UNL/p1705576376841409)

## How was this PR tested?

Not tested

## Notes for reviewers

None

## Interface changes

None.

## Effects on system behavior

Although there will be a slight increase in computational load due to the additional log, I would say that the risk is minimal since the node's output does not occupy terminal at this moment. But just in case, I would like to make this change for several nodes chosen from perception component which tends to have some issues pretty often.


yukkysaito (commented):
@mitsudome-r @xmfcx 
Do you have any comments on this Pull Request? In the past, the .ros/log files in the system have grown significantly, which caused the disk space on the autonomous driving ECU to become constrained. I'm a bit concerned that if log files are output everywhere, it might lead to unintended system shutdowns.

@kminoda 
Is this change only for this node? Do you plan to apply this to other nodes as well? If so, I think we should establish some rules.

kminoda:
@yukkysaito Thank you for the comment. I am willing to get an feedback from the community :pray: 

As a first step, I would like to merge this PR for several nodes which tend to have a lot of issues. For these nodes, I think this change would make a difference and thus would like to merge. (Also written in the description)

mitsudome-r:
> Do you have any comments on this Pull Request? In the past, the .ros/log files in the system have grown significantly, which caused the disk space on the autonomous driving ECU to become constrained. I'm a bit concerned that if log files are output everywhere, it might lead to unintended system shutdowns.

I think we should consider giving a guidance about how users can setup log rotation if they wish to when we set output to "both" for all the nodes. 
If we are just changing few nodes for the time being, I don't think it is going to be a critical issue.

kminoda:
@mitsudome-r Thank you for the comment. As you've mentioned, I would like to make this type of change to some few nodes where the debugging efficiency is critical even in short term perspective.

@yukkysaito Would you confirm if it is OK to merge several PRs that I've created for now? I will wait for your comment before merging those PRs.

tzhong518 (approved):

youtalk:
If you have another logger outside ROS system (e.g. syslog), I think the `output="both"` option performs duplicate logging.

xmfcx:
@yukkysaito like @mitsudome-r suggested, creating a tutorial to set up https://linux.die.net/man/8/logrotate would be helpful.

As for this PR, we can merge as is.

My suggestion would be to set the log preferences of all nodes from a global variable.

So everything will be logged to "screen or log file or both" when you select it from the main launch file.

Maybe we could also have a simple python node package that monitors the log directory and keeps it at certain size too.

### Now

- Merge the PR.

### Short term

- Allow setting all node log method from a single global variable.

### Long term

- Create a logrotate tutorial
- Or create a log rotation node package

yukkysaito:
>@yukkysaito like @mitsudome-r suggested, creating a tutorial to set up https://linux.die.net/man/8/logrotate would be helpful.
As for this PR, we can merge as is.
My suggestion would be to set the log preferences of all nodes from a global variable.
So everything will be logged to "screen or log file or both" when you select it from the main launch file.
Maybe we could also have a simple python node package that monitors the log directory and keeps it at certain size too.

@mitsudome-r @xmfcx Thanks for your comments. I agree with you.
It would be fine if it's just this PR, but there are many similar PRs coming in. 
- https://github.com/autowarefoundation/autoware.universe/pull/8410
- https://github.com/autowarefoundation/autoware.universe/pull/8409
- https://github.com/autowarefoundation/autoware.universe/pull/8408
- https://github.com/autowarefoundation/autoware.universe/pull/8407

@kminoda If it's not urgent, I might want to consider adopting the proposed short term change (`Allow setting all node log methods from a single global variable`) before proceeding with the merge. What do you think?

xmfcx:
> Allow setting all node log methods from a single global variable

If that's the case, we should close all these PRs and go for this option.
And make the default "screen" to not fill up the disk space.

yukkysaito:
@kminoda cc @xmfcx 
Why was this pull request merged? I understood that it was supposed to be merged after completing this TODO.

>If that's the case, we should close all these PRs and go for this option.
And make the default "screen" to not fill up the disk space.

kminoda:
@yukkysaito Sorry I missed the latest comment :pray: 

Do we have any timeline for incorporating the global variable? Although this PR is not critically important, it is kind of urgent to set `screen=both` for some of the important nodes to reduce the debugging cost. So I still would like to merge these PRs for limited packages (of course not for all the packages :+1: ) unless the global variable feature is incorporated within a short period of time.

yukkysaito:
@kminoda 
If it's not urgent, it's not preferable to change to screen=both for individual packages. If we need to merge such PRs in the future, could you please create a PR for the global variable feature?

PR #8418, opened 2024-08-09
fix(autoware_tracking_object_merger): add merge frame

kaancolak:
## Description

Currently, merging of different track objects is only possible in the "map" frame, it's hardcoded. However, some sensors like radar provide tracked objects as a local frame. I just added basic functionality that merges objects in the target frame. 

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


YoshiRi (approved):
LGTM

PR #8431, opened 2024-08-09
fix(autoware_freespace_planning_algorithms): fix variableScope

Nagi70:
## Description

This is a fix based on cppcheck variableScope warnings

```
planning/autoware_freespace_planning_algorithms/src/reeds_shepp.cpp:679:10: style: The scope of the variable 'phi' can be reduced. [variableScope]
  double phi, v;
         ^
```

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


Nagi70:
Since the double type is primitive, `phi` is initialized in the for statement and the variableScope is fixed.

kosuke55 (approved):

PR #8436, opened 2024-08-10
feat(autoware_mpc_lateral_controller): add predicted trajectory acconts for input delay

kyoichi-sugahara:
## Description

**merging this [PR](https://github.com/autowarefoundation/autoware_launch/pull/1114) first.**

### Current Implementation Issue

The MPC optimization calculation uses an **initial state that accounts for input delay**. However, in the current implementation, when calculating the predicted trajectory from the obtained solution, it uses the **currently observed state** instead of the **state that considers this delay**. This creates an **inconsistency** between the optimization calculation and the trajectory prediction.

### New Visualization Option

To verify the accuracy of following the reference trajectory, I have added an option to **visualize the resampled reference trajectory**.

### Initial Findings

- Initial verification results suggest that the **predicted trajectory considering delay** follows the target path **more accurately**.
- As the **steering angle increases**, the difference between the predicted trajectories **with and without delay consideration** tends to become larger.

> **Note:** Further verification is needed to confirm these initial findings.

image

| Line Color | Description | Coordinate | Topic Name |
|------------|-------------|------------|------------|
| 🔵 Blue (dotted) | Reference trajectory | World | `/control/trajectory_follower/controller_node_exe/debug/resampled_reference_trajectory` |
| ⚪ White | Predicted trajectory | World | `/control/trajectory_follower/controller_node_exe/debug/predicted_trajectory_with_delay` |
| 🔴 Red | Predicted trajectory (with delay consideration) | World | `/control/trajectory_follower/lateral/predicted_trajectory` |

## How was this PR tested?

by visualizing predicted trajectory w/wo delay and reference trajectory

https://github.com/user-attachments/assets/881ea825-1922-4f9d-8057-9537a85e2597

- [x] [PASS TIERVI INTERNAL SCENARIOS](https://evaluation.tier4.jp/evaluation/reports/bb584c09-9066-5bc7-83ab-918cb7e3be93?project_id=prd_jt)
## Notes for reviewers

Ideally, we should interpolate the trajectory from the current state to the state considering input delay, and then generate the predicted trajectory from this delay-considered state. Please note that the trajectory visualized in this PR generates the predicted trajectory directly from the delay-considered state without performing this interpolation.

#### Additions and removals
| Change type | Parameter Name                             | Type    | Default Value | Description                                     |
|:------------|:-------------------------------------------|:--------|:--------------|:------------------------------------------------|
| Added       | `debug_publish_resampled_reference_trajectory` | `bool`  | `false`       | Publish debug resampled reference trajectory    |


control/autoware_mpc_lateral_controller/src/mpc.cpp
```diff
@@ -104,8 +108,25 @@ bool MPC::calculateMPC(
...
+    predicted_trajectory_frenet.header.stamp = m_clock->now();
+    predicted_trajectory_frenet.header.frame_id = "map";
+    m_debug_frenet_predicted_trajectory_pub->publish(predicted_trajectory_frenet);
+
+    // Calculate and publish predicted trajectory in world coordinate with delay
+    auto predicted_trajectory_world_with_delay = calculatePredictedTrajectory(
+      mpc_matrix, x0_delayed, Uex, mpc_resampled_ref_trajectory, prediction_dt, "world");
+    predicted_trajectory_world_with_delay.header.stamp = m_clock->now();
+    predicted_trajectory_world_with_delay.header.frame_id = "map";
+    m_debug_predicted_trajectory_with_delay_pub->publish(predicted_trajectory_world_with_delay);
```
takayuki5168, line 128: It sounds that the predicted_trajectory_frenet does not consider the delay.

control/autoware_mpc_lateral_controller/src/mpc.cpp
```diff
@@ -104,8 +108,25 @@ bool MPC::calculateMPC(
   m_raw_steer_cmd_prev = Uex(0);
 
   /* calculate predicted trajectory */
-  predicted_trajectory =
-    calculatePredictedTrajectory(mpc_matrix, x0, Uex, mpc_resampled_ref_trajectory, prediction_dt);
+  predicted_trajectory = calculatePredictedTrajectory(
+    mpc_matrix, x0, Uex, mpc_resampled_ref_trajectory, prediction_dt, "world");
```
takayuki5168, line 112: Is there any reason not to apply the input delay to the official predicted trajectory?

control/autoware_mpc_lateral_controller/src/mpc.cpp
```diff
@@ -104,8 +108,25 @@ bool MPC::calculateMPC(
...
+
+  // Publish predicted trajectories in different coordinates for debugging purposes
+  if (m_debug_publish_predicted_trajectory) {
+    // Calculate and publish predicted trajectory in Frenet coordinate
+    auto predicted_trajectory_frenet = calculatePredictedTrajectory(
+      mpc_matrix, x0, Uex, mpc_resampled_ref_trajectory, prediction_dt, "frenet");
+    predicted_trajectory_frenet.header.stamp = m_clock->now();
+    predicted_trajectory_frenet.header.frame_id = "map";
+    m_debug_frenet_predicted_trajectory_pub->publish(predicted_trajectory_frenet);
```
takayuki5168, line 121: Sounds like considering the input delay will have more accuracy. Any reason not to consider the input delay here?

control/autoware_mpc_lateral_controller/src/mpc.cpp
kyoichi-sugahara: @takayuki5168 
Based on my understanding, it's better to apply the input delay for the official predicted path. In my additional tests, I've found that when reducing steer_tau, the original predicted path tends to oscillate, whereas the predicted path that considers delay follows the reference path more closely.

However, as an answer to the question, there are no reasons other than conservative ones. The main concern is that the change could affect the validation functionality in the latter stages. Therefore, I am proposing a configuration that allows for visualization for comparison purposes.

Do you think we should adopt a path that considers input_delay as the official predicted path?

takayuki5168: Currently, the control_validator and lane_departure_checker which depend on the predicted path are not used so much, so I think we do not have to be sensitive to that. This is not my strong opinion, but I feel that changing sooner is better.

kyoichi-sugahara: @takayuki5168 
How about adopting the predicted path that considers input delay as the official version, while keeping an option to revert to the original predicted path just in case? I've done some light verification on my end by changing vehicle parameters and trying several rosbags, and it seems fine, but it's better to be prepared.

takayuki5168: I agree.

kyoichi-sugahara: @takayuki5168 
addressed in https://github.com/autowarefoundation/autoware.universe/pull/8436/commits/c10f9fdab5ce5a8f278944cf72f47ca7b2bcdd00

takayuki5168 (approved):

PR #8456, opened 2024-08-13
feat(autoware_universe_utils): add LRU Cache

yhisaki:
## Description

Implement [LRU Cache](https://ja.wikipedia.org/wiki/Least_Recently_Used) in autoware_universe_utils. This will allow us to cache the input and output of the function.

We show the performance of Fibonacci, w and w/o lru_cache in example_lru_cache.cpp.

```
n       With Cache (μs) Without Cache (μs)
-------------------------------------------
30      2               852
31      0               1401
32      0               2212
33      1               3730
34      1               5743
35      1               10082
36      1               14685
37      1               26827
38      1               39542
39      1               70720
40      1               103842
```

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


common/autoware_universe_utils/include/autoware/universe_utils/system/lru_cache.hpp
```diff
@@ -0,0 +1,143 @@
...
+   * @param key The key to retrieve.
+   * @return The value associated with the key.
+   * @throws std::runtime_error if the key is not found.
+   */
+  Value get(const Key & key)
+  {
+    auto it = cache_map_.find(key);
+    if (it == cache_map_.end()) {
+      throw std::runtime_error("Key not found");
+    }
```
maxime-clem, line 99: How about returning an `std::optional` (similar to the [implementation in `boost::compute`](https://www.boost.org/doc/libs/1_65_1/boost/compute/detail/lru_cache.hpp)) ?
Then we do not need to `throw` an exception.

common/autoware_universe_utils/examples/example_lru_cache.cpp
maxime-clem, line 1: I think this example should be transformed into a test. It requires minimal changes and would allow coverage of the added code.

soblin (approved):

maxime-clem (approved):
Very nice ! Just today I wanted to use a cache and had to use `boost::compute`. I will switch to your implementation instead.

common/autoware_universe_utils/examples/example_lru_cache.cpp
yhisaki: That's right. I'll write a test!

common/autoware_universe_utils/include/autoware/universe_utils/system/lru_cache.hpp
yhisaki: Resolved in e136ae8bf9f382d1b1f31ed8f7c4805e5be834af !

common/autoware_universe_utils/examples/example_lru_cache.cpp
yhisaki: Resolved in 6de8284bec34f6f426ab998eb6a6b43e577b9932 !

dkoldaev:
@yhisaki I just saw this PR, it is a pretty useful tool! Nice performance improvement in the map based prediction node (this is how I found this PR)
Just wanted to point your attention to the testing side. You tested the performance improvements and this is great, but what about some edge cases? I think the main would be to make sure that the cache does not overflow. Say, make cache of size 20 and try to put 25 elements. (P.S. I wrote this also because I remember you passion about testing :smile:  )

yhisaki:
Thank you for looking at the PR!
Indeed, the edge case testing is insufficient. I'll prepare a PR to enhance the tests as soon as possible!

PR #8462, opened 2024-08-14
refactor(autoware_map_based_prediction): map based pred time keeper ptr

technolojin:
## Description

1. Implementing the time keeper by pointer.
  a. Can be implemented to sub-module by passing the pointer of the time keeper
  b. enable or disable the measurement
2. Add configuration for on/off debug message processes
  a. If the debugging process is heavy/not-needed, it can be turned off.

This PR do not contain any logic change.

## Related links

- Link

This PR should be merged first. 
https://github.com/autowarefoundation/autoware_launch/pull/1118

## How was this PR tested?

Tested in a local recompute environment. 

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_map_based_prediction/config/map_based_prediction.param.yaml
+++ b/perception/autoware_map_based_prediction/config/map_based_prediction.param.yaml
@@ -50,3 +50,8 @@
       consider_only_routable_neighbours: false
 
     reference_path_resolution: 0.5 #[m]
+
+    # debug parameters
+    publish_processing_time: false
+    publish_processing_time_detail: false
+    publish_debug_markers: false
--- a/perception/autoware_map_based_prediction/src/map_based_prediction_node.cpp
+++ b/perception/autoware_map_based_prediction/src/map_based_prediction_node.cpp
@@ -870,9 +870,9 @@ MapBasedPredictionNode::MapBasedPredictionNode(const rclcpp::NodeOptions & node_
     detailed_processing_time_publisher_ =
       this->create_publisher<autoware::universe_utils::ProcessingTimeDetail>(
         "~/debug/processing_time_detail_ms", 1);
-    time_keeper_ = autoware::universe_utils::TimeKeeper(detailed_processing_time_publisher_);
-    time_keeper_ptr_ = std::make_shared<autoware::universe_utils::TimeKeeper>(time_keeper_);
-    path_generator_->setTimeKeeper(time_keeper_ptr_);
+    auto time_keeper = autoware::universe_utils::TimeKeeper(detailed_processing_time_publisher_);
+    time_keeper_ = std::make_shared<autoware::universe_utils::TimeKeeper>(time_keeper);
+    path_generator_->setTimeKeeper(time_keeper_);
   }
 
   if (use_debug_marker) {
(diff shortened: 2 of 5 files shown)


soblin (changes requested):
Having shared pointer to existing `time_keeper_` is not good because the destructor of this `time_keeper_` will be called twice leading to double free. I suggest removing `time_keeper_` and use `time_keeper_ptr_` for both classes.

technolojin:
> Having shared pointer to existing `time_keeper_` is not good because the destructor of this `time_keeper_` will be called twice leading to double free. I suggest removing `time_keeper_` and use `time_keeper_ptr_` for both classes.

@soblin 
Thank you for your advise!
Resolved in https://github.com/autowarefoundation/autoware.universe/pull/8462/commits/7bdb507d86a45f8c992452feb01a4c24aae673d0

soblin (approved):

perception/autoware_map_based_prediction/src/map_based_prediction_node.cpp
```diff
@@ -831,39 +832,56 @@ MapBasedPredictionNode::MapBasedPredictionNode(const rclcpp::NodeOptions & node_
...
-
-  published_time_publisher_ =
-    std::make_unique<autoware::universe_utils::PublishedTimePublisher>(this);
-  detailed_processing_time_publisher_ =
-    this->create_publisher<autoware::universe_utils::ProcessingTimeDetail>(
-      "~/debug/processing_time_detail_ms", 1);
-  time_keeper_ = autoware::universe_utils::TimeKeeper(detailed_processing_time_publisher_);
 
+  // debug publishers
+  if (use_time_publisher) {
```
YoshiRi, line 858: [just a comment]
Need debugger class to handle all of these things

perception/autoware_map_based_prediction/config/map_based_prediction.param.yaml
```diff
@@ -50,3 +50,8 @@
       consider_only_routable_neighbours: false
 
     reference_path_resolution: 0.5 #[m]
+
+    # debug parameters
+    publish_processing_time: false
+    publish_processing_time_detail: false
+    publish_debug_markers: false
```
YoshiRi, line 57: Need to ask planning team if they need this debug markers as a default.
@kyoichi-sugahara Do you need debug markers as default? (For vehicle lane change intention visualization.)

YoshiRi (approved):
LGTM

perception/autoware_map_based_prediction/config/map_based_prediction.param.yaml
YoshiRi: Oops, all of these are enabled in launcher. So there are no concern about this.

perception/autoware_map_based_prediction/src/map_based_prediction_node.cpp
YoshiRi: I missed select range :(
This is obvious, resolve this

perception/autoware_map_based_prediction/config/map_based_prediction.param.yaml
technolojin: I set those to true in the autoware_launch.

PR #8463, opened 2024-08-14
fix(turn_signal, lane_change, goal_planner): add optional to tackle lane change turn signal and pull over turn signal

Owen-Liuyuxuan:
## Description

General, update the turn signal decider to align the rule with traffic rules.

### Lane Change

For lane change action, the lane change will decide whether there is a left/right lane for the current lanelet or the next lanelet. 

There are cases when the lane change starts at a point inside a lanelet without a left/right lane, but the target lane is beside the "next" lanelet. 

So whether shiftSideLane exists should be decided by the Lane Change module, instead of the turn signal decider. (If there is no shiftedSideLane, Lane Change will not act instead). So the LC module should skip that judgment.

### Goal Planner

When the goal planner is pulling over by the side of a wide lane, it should turn on the turn indicator no matter whether there is a shift side lane or the path will cross the lane boundary. So the goal planner module should skip that judgment.

## Related links

**Parent Issue:**

- Link

**Private Links:**

- [Internal Issue](https://tier4.atlassian.net/browse/RT0-31749)
- - [Internal Issue](https://tier4.atlassian.net/browse/RT0-33029)

## How was this PR tested?

PSim and Internal Evaluator.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


kosuke55 (approved):

kosuke55:
thansk!! LGTM

PR #8464, opened 2024-08-14
feat(behavior_path_planner_common): add calculateRoughDistanceToObjects

kosuke55:
## Description

similar to https://github.com/autowarefoundation/autoware.universe/pull/8193, add safety check function to calculate rough  distance between the ego vehicle and the objects. Calculate the distance when the distance is minimized or maximized when the direction of the ego and the object is changed.

image

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


soblin (approved):
LGTM

PR #8465, opened 2024-08-14
perf(goal_planner): reduce unnecessary recursive lock guard

kosuke55:
## Description

Reduced computational cost just a little bit. Reduced unnecessary recursive locks.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


soblin (approved):
Can you declare set_xxx_no_lock as private ?

kosuke55:
> Can you declare set_xxx_no_lock as private ?
fixed in https://github.com/autowarefoundation/autoware.universe/pull/8465/commits/304915fc59bcd3ee22267c09849d73083d092b13

PR #8467, opened 2024-08-14
fix(autoware_map_based_prediction): use surrounding_crosswalks instead of external_surrounding_crosswalks

yhisaki:
## Description

Fix bug in autoware_map_based_prediction. Details are described in [internal slack](https://star4.slack.com/archives/C03QW0GU6P7/p1723603495769929).

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

Tested on psim.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_map_based_prediction/src/map_based_prediction_node.cpp
+++ b/perception/autoware_map_based_prediction/src/map_based_prediction_node.cpp
@@ -411,8 +411,8 @@ bool withinRoadLanelet(
 
 boost::optional<CrosswalkEdgePoints> isReachableCrosswalkEdgePoints(
   const TrackedObject & object, const lanelet::ConstLanelets & surrounding_lanelets,
-  const lanelet::ConstLanelets & external_surrounding_crosswalks,
-  const CrosswalkEdgePoints & edge_points, const double time_horizon, const double min_object_vel)
+  const lanelet::ConstLanelets & surrounding_crosswalks, const CrosswalkEdgePoints & edge_points,
+  const double time_horizon, const double min_object_vel)
 {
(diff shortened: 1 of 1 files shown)


soblin (approved):

perception/autoware_map_based_prediction/src/map_based_prediction_node.cpp
```diff
@@ -1462,7 +1461,10 @@ PredictedObject MapBasedPredictionNode::getPredictedObjectAsCrosswalkUser(
 
   // try to find the edge points for other surrounding crosswalks and generate path to the crosswalk
   // edge
-  for (const auto & crosswalk : external_surrounding_crosswalks) {
+  for (const auto & crosswalk : surrounding_crosswalks) {
+    if (crossing_crosswalk && crossing_crosswalk.get() == crosswalk) {
```
yuki-takagi-66, line 1465: Does this mean that `crossing_crosswalk` is necessary for computation cost?

yuki-takagi-66 (approved):
LGTM
I mentioned a small concern and would appreciate your confirmation.

perception/autoware_map_based_prediction/src/map_based_prediction_node.cpp
yuki-takagi-66: sorry for my missunderstanding

yhisaki: crosswalk is optional, so I think it is necessary.

YoshiRi (approved):

PR #8482, opened 2024-08-15
fix(autoware_detected_object_validation): fix functionStatic

kobayu858:
## Description
This is a fix based on cppcheck functionStatic warnings

Since this function internally calls another function and updates member variables within it, it was decided that it would be better to call it using an instance. Therefore, I have suppressed the warning with a comment.

```
perception/autoware_detected_object_validation/src/obstacle_pointcloud/debugger.hpp:69:8: performance: inconclusive: Technically the member function 'autoware::detected_object_validation::obstacle_pointcloud::Debugger::addNeighborPointcloud' can be static (but you may consider moving to unnamed namespace). [functionStatic]
  void addNeighborPointcloud(const pcl::PointCloud<pcl::PointXY>::Ptr & input)
       ^
```

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_detected_object_validation/src/obstacle_pointcloud/debugger.hpp
+++ b/perception/autoware_detected_object_validation/src/obstacle_pointcloud/debugger.hpp
@@ -66,6 +66,7 @@ public:
     pointcloud_within_polygon_->clear();
   }
 
+  // cppcheck-suppress functionStatic
   void addNeighborPointcloud(const pcl::PointCloud<pcl::PointXY>::Ptr & input)
   {
     pcl::PointCloud<pcl::PointXYZ>::Ptr input_xyz = toXYZ(input);


Shin-kyoto (approved):
LGTM

- I confirmed that `addNeighborPointcloud(const pcl::PointCloud<pcl::PointXY>::Ptr & input)` calls the member function(`addNeighborPointcloud(const pcl::PointCloud<pcl::PointXYZ>::Ptr & input)`, so `addNeighborPointcloud(const pcl::PointCloud<pcl::PointXY>::Ptr & input)` should not be static function.
- I confirmed that the content of this PR is good to merge.
- I did **NOT** check this PR by runnning autoware.

PR #8484, opened 2024-08-15
fix(autoware_multi_object_tracker): enable trigger publish when delay_compensation is false

technolojin:
## Description

Fix a bug that the object is not published when the option `enable_delay_compensation` is false.

## Related links
Caused by PR https://github.com/tier4/autoware.universe/pull/1436

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


YoshiRi (approved):
it worked! LGTM

PR #8415, opened 2024-08-08
feat(simple_planning_simulator): add VGR sim model

kosuke55:
## Description

As shown in https://github.com/autowarefoundation/autoware.universe/pull/8504, add conversion from steer angle to handle angle by variable gear ratio to raw_vehicle_cmd_converter.
To simulate this, implement a model that works with handle angle. And add atuation status publishing.

image

1) same VGR coef in converter and sim
raw_vehicle_cmd_converter
```
    convert_steer_cmd_method: "vgr"
    vgr_coef_a: 15.713
    vgr_coef_b: 0.053
    vgr_coef_c: 0.042
    convert_actuation_to_steering_status: true
```

sim
```
    accel_time_delay: 0.1
    accel_time_constant: 0.1
    brake_time_delay: 0.1
    brake_time_constant: 0.1
    accel_map_path: $(find-pkg-share simple_planning_simulator)/data/actuation_cmd_map/accel_map.csv
    brake_map_path: $(find-pkg-share simple_planning_simulator)/data/actuation_cmd_map/brake_map.csv

    convert_accel_cmd: true
    convert_brake_cmd: true
    convert_steer_cmd: true

    convert_steer_cmd_method: "vgr"
    vgr_coef_a: 15.713
    vgr_coef_b: 0.053
    vgr_coef_c: 0.042
    enable_pub_steer: false
```

https://github.com/user-attachments/assets/c17a0926-270d-4931-8232-f4a1ba97748d

2) change sim VGR
```
    convert_steer_cmd_method: "vgr"
    vgr_coef_a: 5.0
    vgr_coef_b: 2.0
    vgr_coef_c: 1.0
    convert_actuation_to_steering_status: true
    enable_pub_steer: false
```

https://github.com/user-attachments/assets/cdf3149f-09f4-4763-928e-058c23058fc6

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

- psim 
- add unit test

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

PR #8504, opened 2024-08-16
feat(raw_vehicle_cmd_converter): add steer command conversion with VGR

kosuke55:
## Description

Added conversion from steer angle to handle angle by variable gear ratio.
Both cmd and status are supported. (However, status conversion is not supported for steer_map.)
This is to unify the conversion of control commands, which has been done in vehicle_interface.
Also, status correction will be implemented in the future.

image

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

- psim wtih https://github.com/autowarefoundation/autoware.universe/pull/8415
- unit test
- 
## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


vehicle/autoware_raw_vehicle_cmd_converter/include/autoware_raw_vehicle_cmd_converter/node.hpp
```diff
@@ -102,16 +108,26 @@ class RawVehicleCommandConverterNode : public rclcpp::Node
...
   bool is_debugging_;
-  bool convert_accel_cmd_;  //!< @brief use accel or not
-  bool convert_brake_cmd_;  //!< @brief use brake or not
-  bool convert_steer_cmd_;  //!< @brief use steer or not
+  bool convert_accel_cmd_;                                             //!< @brief use accel or not
+  bool convert_brake_cmd_;                                             //!< @brief use brake or not
+  std::optional<std::string> convert_steer_cmd_method_{std::nullopt};  //!< @brief method to convert
+  double vgr_coef_a_{0.0};
+  double vgr_coef_b_{0.0};
+  double vgr_coef_c_{0.0};
```
TakaHoribe, line 116: Why are these members necessary?

TakaHoribe (commented):
Could you update the README, particularly by adding an image of the steering conversion scheme and clarifying/plotting what kind of value the VGR would be?

vehicle/autoware_raw_vehicle_cmd_converter/include/autoware_raw_vehicle_cmd_converter/node.hpp
kosuke55: not necessary. it was for before refactoring, so removed them.

kosuke55:
@TakaHoribe 

> Could you update the README, particularly by adding an image of the steering conversion scheme and clarifying/plotting what kind of value the VGR would be?

thanks! I updated readme and scripts for plot.

vehicle/autoware_raw_vehicle_cmd_converter/README.md
```diff
@@ -45,3 +68,4 @@ For ease of calibration and adjustments to the lookup table, an auto-calibration
 ## Limitation
 
 The current feed back implementation is only applied to steering control.
+`
```
TakaHoribe, line 71:
```suggestion
```

vehicle/autoware_raw_vehicle_cmd_converter/src/node.cpp
```diff
@@ -86,11 +109,11 @@ RawVehicleCommandConverterNode::RawVehicleCommandConverterNode(
 
 void RawVehicleCommandConverterNode::publishActuationCmd()
 {
-  if (!current_twist_ptr_ || !control_cmd_ptr_ || !current_steer_ptr_) {
+  if (!current_twist_ptr_ || !control_cmd_ptr_ || !current_steer_ptr_ || !actuation_status_ptr_) {
```
TakaHoribe, line 112: Some vehicles do not publish actiation_status. This is not necessary for vehicles that do not require the steering conversion. Please make a specific condition for that.

vehicle/autoware_raw_vehicle_cmd_converter/src/node.cpp
```diff
@@ -215,6 +241,41 @@ void RawVehicleCommandConverterNode::onControlCmd(const Control::ConstSharedPtr
   control_cmd_ptr_ = msg;
   publishActuationCmd();
 }
+
+void RawVehicleCommandConverterNode::onSteering(const Steering::ConstSharedPtr msg)
```
TakaHoribe, line 245: Please mention that this subscriber can not be replaced by `take()` method in a comment.

kosuke55: Originally it was take(), in which case the topic name must be stated in the declaration. In that case, there would be two identical topics. Therefore, subscriber, which does not require a topic name to be specified at the time of declaration, is used.
In any case, add a comment to the code.

image

kosuke55: > In that case, there would be two identical topics.

I mean, this node also need to publish steering topic if converting option is true.

image

vehicle/autoware_raw_vehicle_cmd_converter/src/node.cpp
kosuke55: thanks! fixed in https://github.com/autowarefoundation/autoware.universe/pull/8504/commits/0a660ef011c913acebdc84291856bdea4e366410

vehicle/autoware_raw_vehicle_cmd_converter/src/node.cpp
kosuke55: added comment why not use polling sub

https://github.com/autowarefoundation/autoware.universe/pull/8504/commits/5b667ccf9f6052a37c888062a623273f2eaf4254

TakaHoribe (approved):

PR #8510, opened 2024-08-16
fix(lane_change): modify lane change requested condition

zulfaqar-azmi-t4:
## Description

Currently lane change module with request execution as soon as target lane is found.
Even when target lane is 200 meters away, lane change will transition to `WAITING_APPROVAL` state. Example is shown below (note the planner manager status)

https://github.com/user-attachments/assets/efc5a7cb-22bc-4251-9abc-18ba843b1057

This doesn't make sense anyway, since lane change path will be trigger only when the end of prepare phase is within target lane's neighbor. Therefore in this PR, the execution condition is changed and lane change will only be requested if ego distance to the start of target lane is less than maximum prepare length. Example after PR is shown below (note the planner manager status)

https://github.com/user-attachments/assets/d3778530-5a29-4e47-81a2-88baea44484b

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

PSIM

Evaluator result TBA

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/utils/calculation.cpp
```diff
@@ -101,4 +101,55 @@ double calc_dist_to_last_fit_width(
...
+  const auto left_front_pt = lanelet::utils::conversion::toGeomMsgPt(target_left_bound.front());
+  const auto right_front_pt = lanelet::utils::conversion::toGeomMsgPt(target_right_bound.front());
+
+  const auto ego_position = common_data_ptr->get_ego_pose().position;
+  const auto left_dist =
+    motion_utils::calcSignedArcLength(path.points, ego_position, left_front_pt);
+  const auto right_dist =
+    motion_utils::calcSignedArcLength(path.points, ego_position, right_front_pt);
+
+  return std::min(left_dist, right_dist);
```
mkquda, line 153:
```suggestion
  const auto & route_handler_ptr = common_data_ptr->route_handler_ptr;

  if (!route_handler_ptr || target_lanes.empty() || current_lanes.empty()) {
    return 0.0;
  }

  const auto & target_bound =
    common_data_ptr->direction == autoware::route_handler::Direction::RIGHT ?
    target_lanes.front().leftBound() : target_lanes.front().rightBound();

  if (target_bound.empty()) {
    return 0.0;
  }

  const auto path =
    route_handler_ptr->getCenterLinePath(current_lanes, 0.0, std::numeric_limits<double>::max());

  if (path.points.empty()) {
    return 0.0;
  }

  const auto target_front_pt = lanelet::utils::conversion::toGeomMsgPt(target_bound.front());
  const auto ego_position = common_data_ptr->get_ego_pose().position;

  return motion_utils::calcSignedArcLength(path.points, ego_position, target_front_pt);
```

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/include/autoware/behavior_path_lane_change_module/utils/calculation.hpp
```diff
@@ -75,6 +75,43 @@ double calc_stopping_distance(const LCParamPtr & lc_param_ptr);
...
+ */
+double calc_maximum_prepare_length(const CommonDataPtr & common_data_ptr);
+
+/**
+ * @brief Calculates the distance from the ego vehicle to the start of the target lanes.
+ *
+ * This function computes the shortest distance from the current position of the ego vehicle
+ * to the start of the target lanes by measuring the arc length to the front points of
+ * the left and right boundaries of the target lane. If the target lanes are empty or other
+ * required data is unavailable, the function returns 0.0.
```
mkquda, line 101: should probably return `std::numeric_limits<double>::max()` instead of 0.0, so that LC doesn't activate

mkquda (approved):

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/include/autoware/behavior_path_lane_change_module/utils/calculation.hpp
zulfaqar-azmi-t4: fixed the docstring in https://github.com/autowarefoundation/autoware.universe/pull/8510/commits/b7fd3f9f9e3a9449bcfb726dfef1bab0103528b5

PR #8351, opened 2024-08-05
fix(autoware_vehicle_cmd_gate): fix unreadVariable

kobayu858:
## Description
This is a fix based on cppcheck unreadVariable warnings.

```
control/autoware_vehicle_cmd_gate/src/vehicle_cmd_gate.cpp:488:28: style: Variable 'filtered_commands.gear' is assigned a value that is never used. [unreadVariable]
    filtered_commands.gear = commands.gear;  // tmp
                           ^

control/autoware_vehicle_cmd_gate/src/vehicle_cmd_gate.cpp:501:28: style: Variable 'filtered_commands.gear' is assigned a value that is never used. [unreadVariable]
    filtered_commands.gear = emergency_commands_.gear;  // tmp
                           ^

control/autoware_vehicle_cmd_gate/src/vehicle_cmd_gate.cpp:853:15: style: Variable 'first_msg' is assigned a value that is never used. [unreadVariable]
    first_msg = false;
              ^
```
## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

PR #8518, opened 2024-08-19
refactor(vehicle_cmd_gate): use smaller class for filter command

TakaHoribe:
## Description

Since the `gear` field is not used as mentioned in https://github.com/autowarefoundation/autoware.universe/pull/8351, we no longer have to use the `Commands` struct. I replace it with the `Control` struct.

## Related links

**Parent Issue:**

None

## How was this PR tested?

- build on CI
- run with psim

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


tkimura4 (approved):
LGTM

PR #8300, opened 2024-08-01
feat(autoware_pointcloud_preprocessor): redesign concatenate and time sync node

vividf:
## Description
This PR solved the issue https://github.com/autowarefoundation/autoware.universe/issues/6832.
Previous designs have some issues concatenating the pointcloud correctly, therefore, this PR redesigns the logic of the concatenate node in order to handle the edge cases like pointcloud delay or pointcloud drop. 

Changes
- new algorithm
- diagnostic message
- parameter file, launch file, schema
- unit test: test the logic of the cloud_collector and combine_cloud_handler
- component test: testing edge cases (pointcloud dropped, delayed)

A more detailed description of the algorithm is on the Readme page.https://github.com/vividf/autoware.universe/blob/feature/redesign_concatenate_and_time_sync_node/sensing/autoware_pointcloud_preprocessor/docs/concatenate-data.md

## Related links

**Parent Issue:**

- Link

### Unit test and Component test for concatenate node
```bash
# build autoware_pointcloud_preprocessor
colcon build --symlink-install --cmake-args -DCMAKE_BUILD_TYPE=Release --packages-up-to autoware_pointcloud_preprocessor

# test autoware_pointcloud_preprocessor
colcon test --packages-select autoware_pointcloud_preprocessor --event-handlers console_cohesion+
```

### Tested with sample rosbag
Please change the sample_sensor_kit_launch: https://github.com/autowarefoundation/sample_sensor_kit_launch/pull/108
```
# Terminal 1
ros2 launch autoware_launch logging_simulator.launch.xml map_path:=$HOME/autoware_map/sample-map-rosbag vehicle_model:=sample_vehicle sensor_model:=sample_sensor_kit

# Terminal 2
ros2 bag play sample-rosbag_0.db3
```

The result provided by @mojomex below https://github.com/autowarefoundation/autoware.universe/pull/8300#issuecomment-2443271632

### Tested with vehicle 1
data: [TIER4_INTERNAL_LINK](https://drive.google.com/drive/folders/1Ss4ZJpS9agFTThqzDv-O6wmrvQTngNp2?usp=drive_link)

modify the config file `concatenate_and_time_sync_node.param.yaml`
```bash
/**:
  ros__parameters:
    debug_mode: false
    has_static_tf_only: false
    rosbag_length: 20.0
    maximum_queue_size: 5    
    timeout_sec: 0.2
    is_motion_compensated: true
    publish_synchronized_pointcloud: true
[...]

PR #8525, opened 2024-08-19
chore(autoware_pointcloud_preprocessor): change unnecessary warning message to debug

vividf:
## Description
This PR provides a quick fix for changing unnecessary **warning** messages to **debug** to reduce the console log. 
This PR https://github.com/autowarefoundation/autoware.universe/pull/8300 fixes the issue but will take time to review.

According to  [TIER4_INTERNAL_LINK](https://star4.slack.com/archives/C076E8EBJTG/p1724037248363719?thread_ts=1724036108.812549&cid=C076E8EBJTG)
`[component_container_mt-1] [WARN 1724036039.656393647] [sensing.lidar.concatenate_data] computeTransformToAdjustForOldTimestamp(): "old_stamp is newer than new_stamp," at (/home/autoware/pilot-auto/src/autoware/universe/sensing/pointcloud_preprocessor/src/concatenate_data/concatenate_and_time_sync_nodelet.cpp:L255)`

Warning messages keep showing up due to the algorithm design. 

Current algorithm:
1. Sort the timestamp from large to small in a vector.
2. Pointclouds will find the timestamp smaller than its timestamp (in the vector) to calculate the transformation. But since it is finding the timestamp from large to small, if the pointcloud is not the newest one, it will always encounter the situation that the current pointcloud timestamp is smaller than the largest (or other larger) timestamp in the vector. When this happens the warning will show up.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


yukkysaito (commented):
@vividf Thank you. I believe this should ideally be issued as a warning, so could you explain why it was categorized as debug information instead? The root cause should be addressed as a warning, but by downgrading it to debug, there’s a risk of missing the underlying issue.

That said, if this process has little to no impact on the system’s operation and doesn’t require correction, then treating it as debug information is acceptable.

vividf:
@yukkysaito 
Thanks for the comments. 
From the algorithm described in the description above, all of the pointcloud will show this warning message instead of the pointcloud that has lastest timestamp. 
For instance, if we have 4 pointcloud ready to concatenate, 3 of them satisfy the if condition.  
`  if (old_stamp > new_stamp) {
    RCLCPP_DEBUG_STREAM_THROTTLE(
      get_logger(), *get_clock(), std::chrono::milliseconds(10000).count(),
      "old_stamp is newer than new_stamp,");`

**if** condition is **true**
1st (earliest) pointlcoud:  3 time, 2nd pointcloud: 2 times, 3rd pointcloud: 1 times

Since almost all of the pointcloud will satisfy the if condition and this has no impact on system operation, I don't think a **warning** is a good logging message

YoshiRi (approved):
@vividf I think this PR is temporary because new design for concatenation in https://github.com/autowarefoundation/autoware.universe/pull/8300 will solve this problem, right?
If so, saito-san's concern will not really affect products so much. LGTM.

PR #8526, opened 2024-08-19
fix: cpp17 namespaces

amc-nu:
## Description
Replace C++17-style namespaces with nested ones to allow compilation on Edge-Auto on the Roscube, which has a noncompliant c++17 compiler for the nodes:
- TensorRT YoloX
- ByteTrack

## Related links

**Parent Issue:**
https://github.com/tier4/edge-auto-jetson/pull/57

## How was this PR tested?
It was tested on Edge-Auto platform on both the ADLink AVA3510 and Roscube RQX-58G. https://edge.auto/

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


Shin-kyoto:
Thank for submitting this PR!!

IMO, changing namespace for the specific Hardware (e.g. Roscube) and non standard compilers is not good to keep the autoware codebase healthy.
How do you think about it? @vividf @amadeuszsz @tzhong518

amadeuszsz:
> Thank for submitting this PR!!
> 
> IMO, changing namespace for the specific Hardware (e.g. Roscube) and non standard compilers is not good to keep the autoware codebase healthy. How do you think about it? @vividf @amadeuszsz @tzhong518

Would be nice to keep compatibility with other devices by such a simple syntax change. I don't think we have a lot of nested namespaces within single component, thus this change will not decrease readability.
I can imagine that other Autoware CUDA-based packages find their use case on edge devices. We could populate this change for all packages which utilize CUDA compiler:
```
common/cuda_utils
common/tensorrt_common
perception/autoware_image_projection_based_fusion
perception/autoware_lidar_centerpoint
perception/autoware_lidar_transfusion
perception/autoware_shape_estimation
perception/autoware_tensorrt_classifier
perception/autoware_tensorrt_yolox
perception/autoware_traffic_light_classifier
perception/autoware_traffic_light_fine_detector
```
However, edge devices vendors slowly integrate JetPack 6.0 to their products and in near future this change will be not necessary.
Finally, I have no objection to this PR change. Regarding the other packages, I would like to hear other comments.

manato:
@amc-nu @a-yyg 
I appreciate issuing PR!

@Shin-kyoto @amadeuszsz 
Thank you for discussing. This kind of discussion has significant meaning in keeping OSS quality. As it was pointed out, this modification aims to make yolox package work on relatively old environments such as Jetson-based ECU. IMO, this package is one of the most common packages to be executed on such kinds of edge devices, which sometimes are still in old environments. Since I feel the proposed modification looks understandable and does not decrease code quality, I'd like to approve this proposal as a code owner. I don't intend to prevent this kind of discussion in future PRs, so I'd appreciate it if you guys would feel free to post comments when necessary!

manato (approved):
@amc-nu @a-yyg 
Again, thank you for the PR. LGTM. I confirmed that the modification is valid for X86-based ECU as well as Jetson AGX Xavier-based ECU.

PR #8536, opened 2024-08-20
feat(autoware_perception_rviz_plugin): rviz predicted path mark as triangle

technolojin:
## Description

Replace path marker from circle to triangle. 
It may reduce rendering load and adds information of its direction.

Before

Screenshot from 2024-08-20 18-12-03

After

Screenshot from 2024-08-20 20-00-00

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


YoshiRi (approved):
LGTM

PR #8545, opened 2024-08-21
fix(static_obstacle_avoidance): target object sorting

go-sakayori:
## Description
The targets object was not added to the member variable `registered_objects_`, so the envelope polygon for the target object was updated with a single shot polygon. This PR will fix the above problem by adding the target objects into  `registered_objects_` appropriately.

## Related links
None

## How was this PR tested?
[TIER IV internal test](https://evaluation.tier4.jp/evaluation/reports/3959d9b3-f390-5e97-acf8-35841e62d97d?project_id=prd_jt)

## Notes for reviewers
None.

## Interface changes
None.

## Effects on system behavior
None.


satoshi-ota (approved):
LGTM

PR #8546, opened 2024-08-21
ci(labeler): tag packages that require cuda

yhisaki:
## Description

This PR is to prepare functionality that does not execute `build-and-test-differential (humble, -cuda)` for PRs related to packages that are not related to cuda.

The label `tag:require-cuda-build-and-test` is applied only to `perception/**/*`, `sensing/**/*`, `common/cuda_utils/**/*`, and `common/tensorrt_common/**/*`.

## Related links

The functionality to skip cuda build using this PR is implemented in the following:
https://github.com/autowarefoundation/autoware.universe/pull/8547

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

PR #8547, opened 2024-08-21
feat(build-and-test-differential): skip cuda build when tag:require-cuda-build-and-test is not attached

yhisaki:
## Description

This PR is to prevent the execution of build-and-test-differential (humble, -cuda) for PRs related to packages that are not cuda-related.

For PRs that do not have the `tag:require-cuda-build-and-test` label, all steps within build-and-test-differential (humble, -cuda) will be skipped.

## How was this PR tested?

A similar workflow was tested at https://github.com/yhisaki/ros2-unit-test-sample/actions/runs/10483601306/job/29036576962.
In this workflow, steps are skipped if the label `skip-world` is applied when `matrix.name` is `world`.

`skip-world` corresponds to `tag:require-cuda-build-and-test`, and `matrix.name` corresponds to `matrix.container-suffix`.

## Notes for reviewers

This PR should be merged after https://github.com/autowarefoundation/autoware.universe/pull/8546 is merged.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/.github/workflows/build-and-test-differential.yaml
+++ b/.github/workflows/build-and-test-differential.yaml
@@ -25,6 +25,8 @@ jobs:
   build-and-test-differential:
     needs: make-sure-label-is-present
     if: ${{ needs.make-sure-label-is-present.outputs.result == 'true' }}
+    env:
+      SKIP: ${{ matrix.container-suffix == '-cuda' && !contains(github.event.pull_request.labels.*.name, 'tag:require-cuda-build-and-test') }}
     runs-on: ${{ matrix.runner }}
     container: ${{ matrix.container }}${{ matrix.container-suffix }}
     strategy:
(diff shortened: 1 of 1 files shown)


youtalk (commented):
I think it's not good for maintainability to have `${{ env.SKIP != 'true' }}` appear many times. Can you refactor by making a composite action in `.github/actions` so that the conditional branching can be done in just one place?
https://docs.github.com/en/actions/sharing-automations/creating-actions/creating-a-composite-action

I hope the following code is helpful for you to make it possible.
https://github.com/autowarefoundation/autoware/blob/main/.github/workflows/docker-build-and-push.yaml#L60-L64

yhisaki:
Thank you for review! I will try it!

.github/workflows/build-and-test-differential.yaml
```diff
@@ -25,6 +25,8 @@ jobs:
   build-and-test-differential:
     needs: make-sure-label-is-present
     if: ${{ needs.make-sure-label-is-present.outputs.result == 'true' }}
+    env:
+      SKIP: ${{ matrix.container-suffix == '-cuda' && !contains(github.event.pull_request.labels.*.name, 'tag:require-cuda-build-and-test') }}
```
mitsudome-r, line 29: We had issues around the usage of `github.event.pull_request.labels` in the past that it only gives us the name of the label that was modified. (For example, if you add another tag to the label after `tag:require-cuda-build-and-test`, this variable would be set to false.)

It is better to use the `make-sure-label-is-present` workflow to check if the label is present just like we do for `tag:run-build-and-test-differential` tag.

https://github.com/autowarefoundation/autoware.universe/blob/713d34548742d006b32b4cebc6585b9c1c758370/.github/workflows/build-and-test-differential.yaml#L20

yhisaki:
> I think it's not good for maintainability to have `${{ env.SKIP != 'true' }}` appear many times. Can you refactor by making a composite action in `.github/actions` so that the conditional branching can be done in just one place? https://docs.github.com/en/actions/sharing-automations/creating-actions/creating-a-composite-action
> 
> I hope the following code is helpful for you to make it possible. https://github.com/autowarefoundation/autoware/blob/main/.github/workflows/docker-build-and-push.yaml#L60-L64

@youtalk 

First, we submitted a [PR](https://github.com/autowarefoundation/autoware.universe/pull/8555) that only separates build-and-test-differential into composite actions.

.github/workflows/build-and-test-differential.yaml
yhisaki: I refrect your review in https://github.com/autowarefoundation/autoware.universe/pull/8562 !
Thanks!

PR #8549, opened 2024-08-21
feat(lane_change): fix delay logic that caused timing to be late

zulfaqar-azmi-t4:
## Description

With delay lane change, ego will start changing lane only when it is beside the last parked object.

image

We would like to start changing lanes earlier, so we want to find this point while delaying lane change.

image

#### Before PR

Delay lane change is activated instead of safety check

https://github.com/user-attachments/assets/568f2fa3-864c-45cf-b155-f1fcf08c21e9

#### After PR

Safety check is activated instead of delay lane change.

https://github.com/user-attachments/assets/bb7f7ae4-6cb9-4de9-b032-ba2b631033ad

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

1. PSIM
2. [TIER IV internal link](https://evaluation.tier4.jp/evaluation/reports/d4326584-6e6d-5a51-bc39-f4c3f29d0c8a?project_id=prd_jt)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/utils/utils.cpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/utils/utils.cpp
@@ -14,6 +14,7 @@
 
 #include "autoware/behavior_path_lane_change_module/utils/utils.hpp"
 
+#include "autoware/behavior_path_lane_change_module/utils/calculation.hpp"
 #include "autoware/behavior_path_lane_change_module/utils/data_structs.hpp"
 #include "autoware/behavior_path_lane_change_module/utils/path.hpp"
 #include "autoware/behavior_path_planner_common/marker_utils/utils.hpp"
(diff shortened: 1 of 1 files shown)


planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/utils/utils.cpp
```diff
@@ -989,30 +990,41 @@ bool passed_parked_objects(
...
+        current_lane_path.points, src_point, goal_pose.position);
     }
-  }
+    return motion_utils::calcSignedArcLength(current_lane_path.points, src_point, current_path_end);
+  };
+
+  const auto min_dist_to_end_of_current_lane = std::invoke([&]() {
+    auto min_dist_to_end_of_current_lane = std::numeric_limits<double>::max();
+    for (const auto & point : leading_obj_poly.outer()) {
+      const auto obj_p = autoware::universe_utils::createPoint(point.x(), point.y(), 0.0);
```
maxime-clem, line 1005: (very very minor)
```suggestion
      const auto obj_p = universe_utils::createPoint(point.x(), point.y(), 0.0);
```

maxime-clem (approved):
Tested in Psim and the change looks good.

PR #8551, opened 2024-08-21
feat(static_obstacle_avoidance): update envelope polygon creation method

go-sakayori:
## Description
When the tracking accuracy is low the target object's envelope polygon becomes large. This large envelope polygon remains even when the tracking accuracy becomes high. This tends to happen when the target object is detected from a far distance; once an unnecessary large sized envelope polygon is created the target object could not be avoided as shown in the movie below.
[Before](https://github.com/user-attachments/assets/cf630b2a-504a-49c1-96b0-1e19a38c7ab0)

This PR updates the envelope polygon by considering the pose covariance.  

## Related links
https://github.com/autowarefoundation/autoware_launch/pull/1130

## How was this PR tested?
PSim + ROSbag
[After](https://github.com/user-attachments/assets/f790cb42-ee29-4432-bef5-c3448860a936)

[TIER IV internal test](https://evaluation.tier4.jp/evaluation/reports/a77becac-0c48-5c6c-ab1b-7e7d8da8c0c2?project_id=prd_jt)

## Notes for reviewers
None.

## Interface changes
None.

## Effects on system behavior
None.

### DIFF
--- a/planning/behavior_path_planner/autoware_behavior_path_static_obstacle_avoidance_module/src/utils.cpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_static_obstacle_avoidance_module/src/utils.cpp
@@ -21,6 +21,7 @@
 #include "autoware/behavior_path_static_obstacle_avoidance_module/data_structs.hpp"
 #include "autoware/behavior_path_static_obstacle_avoidance_module/utils.hpp"
 
+#include <Eigen/Dense>
 #include <autoware_lanelet2_extension/utility/message_conversion.hpp>
 
 #include <boost/geometry/algorithms/buffer.hpp>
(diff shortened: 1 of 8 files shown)


planning/behavior_path_planner/autoware_behavior_path_static_obstacle_avoidance_module/src/utils.cpp
```diff
@@ -1558,11 +1559,27 @@ void fillObjectEnvelopePolygon(
...
+    if (error_eclipse_long_radius < object_data.error_eclipse_max) {
+      object_data.error_eclipse_max = error_eclipse_long_radius;
+      object_data.envelope_poly = one_shot_envelope_poly;
+      return;
+    }
+    object_data.envelope_poly = same_id_obj->envelope_poly;
+    return;
+  }
+
+  object_data.error_eclipse_max = object_parameter.th_error_eclipse_long_radius;
```
satoshi-ota, line 1582: It should be like this..? 

```suggestion
  object_data.error_eclipse_max = error_eclipse_long_radius;
```

satoshi-ota (approved):
LGTM

PR #8562, opened 2024-08-21
feat(build-and-test-differential): skip cuda build when tag:require-cuda-build-and-test is not attached

yhisaki:
## Description

This PR is to prevent the execution of build-and-test-differential (humble, -cuda) for PRs related to packages that are not cuda-related.

For PRs that do not have the tag:require-cuda-build-and-test label, all steps within build-and-test-differential (humble, -cuda) will be skipped.

## Related links

This PR reflects the advice of https://github.com/autowarefoundation/autoware.universe/pull/8547. 

## How was this PR tested?

The setting of this action works well in https://github.com/autowarefoundation/autoware.universe/actions/runs/10488733741/job/29051839562?pr=8562.

## Notes for reviewers

This PR should be merged after https://github.com/autowarefoundation/autoware.universe/pull/8555 is merged.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/.github/workflows/build-and-test-differential.yaml
+++ b/.github/workflows/build-and-test-differential.yaml
@@ -61,10 +61,6 @@ jobs:
           ref: ${{ github.event.pull_request.head.sha }}
           fetch-depth: ${{ env.PR_FETCH_DEPTH }}
 
-      - name: Show whether build-and-test-differential is skipped
-        run: echo "${{ !(matrix.container-suffix == '-cuda') || needs.make-sure-require-cuda-label-is-present.outputs.result == 'true' }}"
-        shell: bash
-
       - name: Run build-and-test-differential action
         if: ${{ !(matrix.container-suffix == '-cuda') || needs.make-sure-require-cuda-label-is-present.outputs.result == 'true' }}
         uses: ./.github/actions/build-and-test-differential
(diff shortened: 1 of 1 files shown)


youtalk:
Is this still draft PR?

yhisaki:
@youtalk 
PR is now open! Please review!

.github/workflows/build-and-test-differential.yaml
```diff
@@ -57,6 +62,7 @@ jobs:
           fetch-depth: ${{ env.PR_FETCH_DEPTH }}
 
       - name: Run build-and-test-differential action
+        if: ${{ !(matrix.container-suffix == '-cuda') || needs.make-sure-require-cuda-label-is-present.outputs.result == 'true' }}
```
youtalk, line 65: I think it is AND condition.

```suggestion
        if: ${{ !(matrix.container-suffix == '-cuda') && needs.make-sure-require-cuda-label-is-present.outputs.result == 'true' }}
```

yhisaki: I want to set the value to false only when the suffix is "cuda" and the label is not present. In all other cases, I want to set it to true. Below is the truth table I want to achieve:
| | label is present | label is not present |
| -------- | -------- | -------- |
| suffix == "cuda" | true | false |
| !(suffix == "cuda") | true | true |

I think we can achieve the desired logic using an OR operation. The expression `!(matrix.container-suffix == '-cuda')` ensures that the condition is true when the suffix is not `'cuda'`. The expression `needs.make-sure-require-cuda-label-is-present.outputs.result == 'true'` ensures that the condition is true when the label is present.

youtalk: Oh, I miss the `!` mark. Now I understood.

PR #8563, opened 2024-08-21
fix(goal_planner): fix zero velocity in middle of path

kosuke55:
## Description

When extending a path, if the end of the original target path contains 0 velocity, then 0 velocity remains in the middle of the extended_path. This causes more stuck before the goal.
This issue was observed when the goal planner was activated after the LC was activated, for example.

```
  // NOTE: The previous module may insert a zero velocity at the end of the path, so remove it by
  // setting remove_connected_zero_velocity=true. Inserting a velocity of 0 into the goal is the
  // role of the goal planner, and the intermediate zero velocity after extension is unnecessary.
```

before

image

after

image

## Related links

**Parent Issue:**

- Link

**Private Links:**

- https://tier4.atlassian.net/browse/RT0-31696

## How was this PR tested?

- psim
- 2024/08/21 https://evaluation.tier4.jp/evaluation/reports/fbeded6b-7456-58ba-a6af-a39e9288b3f0/?project_id=prd_jt

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_path_planner/autoware_behavior_path_goal_planner_module/include/autoware/behavior_path_goal_planner_module/util.hpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_goal_planner_module/include/autoware/behavior_path_goal_planner_module/util.hpp
@@ -84,12 +84,31 @@ bool isReferencePath(
 std::optional<PathWithLaneId> cropPath(const PathWithLaneId & path, const Pose & end_pose);
 PathWithLaneId cropForwardPoints(
   const PathWithLaneId & path, const size_t target_seg_idx, const double forward_length);
+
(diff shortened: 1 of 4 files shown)


kosuke55:
@soblin 

I added comment why it is ok to remove zero velocity of previous output path
```
```
  // NOTE: The previous module may insert a zero velocity at the end of the path, so remove it by
  // setting remove_connected_zero_velocity=true. Inserting a velocity of 0 into the goal is the
  // role of the goal planner, and the intermediate zero velocity after extension is unnecessary.
```
```

soblin (approved):

PR #8564, opened 2024-08-22
fix(autoware_pointcloud_preprocessor): blockage diag node add runtime error when the parameter is wrong

vividf:
## Description
[TIER4_INTERNAL_LINK](https://tier4.atlassian.net/browse/RT0-31963)

When the user sets the parameter "vertical_bins" less than the number of LiDAR's channel, it will output a runtime error to let the user know they should set the vertical_bins correctly instead of only getting the `malloc() `error.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


sensing/autoware_pointcloud_preprocessor/src/blockage_diag/blockage_diag_node.cpp
```diff
@@ -190,6 +190,12 @@ void BlockageDiagComponent::filter(
     sky_blockage_range_deg_[1] = angle_range_deg_[1];
   } else {
     for (const auto p : pcl_input->points) {
+      if (p.channel > vertical_bins) {
```
badai-nguyen, line 193: @vividf Thank you for your PR, I have a minor suggestion since p.channel valua range is [0, vertical_bins -1].
```suggestion
      if (p.channel >= vertical_bins) {
```

vividf: @badai-nguyen thanks for the check!

vividf:
@badai-nguyen Could you recheck this, thanks!

sensing/autoware_pointcloud_preprocessor/src/blockage_diag/blockage_diag_node.cpp
```diff
@@ -190,6 +190,12 @@ void BlockageDiagComponent::filter(
     sky_blockage_range_deg_[1] = angle_range_deg_[1];
   } else {
     for (const auto p : pcl_input->points) {
+      if (p.channel >= vertical_bins) {
+        std::ostringstream oss;
```
badai-nguyen, line 194: @vividf 
Q: Is there any reason to select `std::ostringstream` instead of `RCLCPP_ERROR`? I think `RCLCPP_ERROR` is common for autoware and this will keep error log in `ros/log` file?

vividf:
@badai-nguyen Thanks for your suggestion!! I added the logging message.
However, instead of letting the system break by itself because of the memory-allocating problem 
I think it is better to interrupt by ourselves when the parameter setting is wrong. Therefore, I think keeping the runtime error is needed. What do you think?

Thanks!

badai-nguyen:
> @badai-nguyen Thanks for your suggestion!! I added the logging message. However, instead of letting the system break by itself because of the memory-allocating problem I think it is better to interrupt by ourselves when the parameter setting is wrong. Therefore, I think keeping the runtime error is needed. What do you think?
> 
> Thanks!

@vividf Yes, I agree, actually I missed `return;` in my suggestion

badai-nguyen (approved):
LGTM

PR #8569, opened 2024-08-22
feat(autonomous_emergency_braking): enable aeb with only one req path

danielsanchezaran:
## Description
This PR allows the AEB module to work when at least one type of path (IMU or MPC based) is available (and use_imu_path and use_predicted_path parameter are both true). A warning message is printed if one of the path types is missing, but the AEB can still operate an issue emergency stops using the remaining path type. 

https://github.com/user-attachments/assets/a214f2d8-d783-45ec-90be-26c356678080

## Related links

**Parent Issue:**

- Link

## How was this PR tested?
PSim
## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/control/autoware_autonomous_emergency_braking/src/node.cpp
+++ b/control/autoware_autonomous_emergency_braking/src/node.cpp
@@ -488,7 +488,7 @@ bool AEB::checkCollision(MarkerArray & debug_markers)
 
   // step3. make function to check collision with ego path created with sensor data
   const auto has_collision_ego = [&](pcl::PointCloud<pcl::PointXYZ>::Ptr filtered_objects) -> bool {
-    if (!use_imu_path_) return false;
+    if (!use_imu_path_ || !angular_velocity_ptr_) return false;
     const double current_w = angular_velocity_ptr_->z;
     constexpr colorTuple debug_color = {0.0 / 256.0, 148.0 / 256.0, 205.0 / 256.0, 0.999};
     const std::string ns = "ego";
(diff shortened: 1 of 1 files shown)


control/autoware_autonomous_emergency_braking/src/node.cpp
```diff
@@ -348,21 +348,27 @@ bool AEB::fetchLatestData()
...
+  const bool has_predicted_path = std::invoke([&]() {
+    if (!use_predicted_trajectory_) {
+      return false;
+    }
+    predicted_traj_ptr_ = sub_predicted_traj_.takeData();
+    return (!predicted_traj_ptr_) ? missing("control predicted trajectory") : true;
+  });
+
+  if (!has_imu_path && !has_predicted_path) {
+    return missing("any type of path");
```
kyoichi-sugahara, line 371:
```suggestion
    if (use_imu_path_ && use_predicted_trajectory_) {
      RCLCPP_INFO_SKIPFIRST_THROTTLE(
        get_logger(), *get_clock(), 5000,
        "[AEB] At least one path (IMU or predicted trajectory) is required for operation");
    }
    return false;
```
@danielsanchezaran 
Isn't it sufficient to output "At least one path is required" only when both `use_imu_path_` and `use_predicted_trajectory_` are true?

danielsanchezaran: @kyoichi-sugahara this check "(use_imu_path_ && use_predicted_trajectory_)" is not really necessary, the AEB requires at least one path (imu or predicted_path) so the lack of at least 1 path should be enough to trigger the message, in the case that, for example the user sets use_imu_path_ to true and use_predicted_trajectory_ to false and there is NO Imu topic, I believe printing this message is not a problem either. I have modified the code to reflect this

kyoichi-sugahara (approved):

PR #8571, opened 2024-08-22
fix(lane_change): activate turn signal as soon as we have the intention to change lanes

mkquda:
## Description

In the current implementation of LC, the turn signal behaves as follows:
- When approved: activate turn signal when approaching start of lane change
- When waiting for approval: activate turn signal when approaching terminal start

This logic is not intuitive, when it is difficult to find safe LC path due to traffic, turn signal will not be activated until the very end.
We should activate turn signal as soon as we have the intention to do LC, to indicate to other vehicles that we wish to change lanes.

## Changes

LC turn signal behavior will be as follows:

When LC path is approved -> activate immediately
When LC path is awaiting approval:
- If no turn signal command from previous module -> activate immediately
- Else -> activate when approaching terminal start

## How was this PR tested?
- PSIM
- [TIER IV internal Llink](https://evaluation.tier4.jp/evaluation/reports/2f2d4cad-4199-5030-b532-9f24c621e83a?project_id=prd_jt)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

Turn signal will activate as soon as we have the intention to change lanes, to indicate to other vehicles that we wish to change lanes.


zulfaqar-azmi-t4:
Looks good

Before:

Turn signal will be turned off once lane change is cancelled, or while lane changing is not approved.

https://github.com/user-attachments/assets/8066ebf2-ad77-4813-b175-ee00179ecc50



After:

Turn signal remains on.

https://github.com/user-attachments/assets/7a1e301e-2095-4c65-93d7-db82ccd01510

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/scene.cpp
```diff
@@ -319,14 +302,15 @@ BehaviorModuleOutput NormalLaneChange::generateOutput()
     output.path = abort_path_->path;
     insertStopPoint(get_current_lanes(), output.path);
   } else {
-    output.path = getLaneChangePath().path;
+    output.path = status_.lane_change_path.path;
```
zulfaqar-azmi-t4, line 305: Just curious. Why it is changed from `getLaneChangePath()`?

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/include/autoware/behavior_path_lane_change_module/scene.hpp
```diff
@@ -109,14 +109,16 @@ class NormalLaneChange : public LaneChangeBase
 
   bool isStoppedAtRedTrafficLight() const override;
 
-  TurnSignalInfo get_current_turn_signal_info() override;
+  TurnSignalInfo get_current_turn_signal_info() const override;
```
zulfaqar-azmi-t4, line 112:
```suggestion
  TurnSignalInfo get_current_turn_signal_info() const final;
```
It is `override` previously, but since the modules that inherits `NormalLaneChange` doesn't actually change anything, it is better to change to `final`

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/scene.cpp
mkquda: why do we need to call a function if we can directly access the member variable?

zulfaqar-azmi-t4 (approved):

PR #8572, opened 2024-08-22
fix(autoware_lidar_centerpoint): fix unusedFunction

kobayu858:
## Description
It is used in `.cu` files and is suppressed in the comments.

```
perception/autoware_lidar_centerpoint/lib/utils.cpp:22:0: style: The function 'divup' is never used. [unusedFunction]
std::size_t divup(const std::size_t a, const std::size_t b)
^
```
## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_lidar_centerpoint/lib/utils.cpp
+++ b/perception/autoware_lidar_centerpoint/lib/utils.cpp
@@ -19,7 +19,7 @@
 namespace autoware::lidar_centerpoint
 {
 // cspell: ignore divup
-std::size_t divup(const std::size_t a, const std::size_t b)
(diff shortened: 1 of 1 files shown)


kminoda (approved):
LGTM (probably cppcheck is not checking cuda files...?)

veqcc:
@kminoda 
>  (probably cppcheck is not checking cuda files...?)

True,, see https://github.com/autowarefoundation/autoware.universe/blob/257b73c7d0822118e9516fdd923aa059ccde2ddd/.github/workflows/cppcheck-differential.yaml#L56

PR #8576, opened 2024-08-22
refactor(ekf_localizer): rename dev to var in the Simple1DFilter class

YamatoAndo:
## Description
The variable dev_ in Simple1DFilter is being treated as variance, not standard deviation. 
So, I have renamed `dev` to `var` in the Simple1DFilter class

## Related links

**Parent Issue:**

- Link

## How was this PR tested?
Lsim works

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


SakodaShintaro (approved):
I have confirmed that logging_simulator works well and the mean error compared to AWSIM ground truth is almost the same as before (6.5cm -> 6.6cm)
LGTM

PR #8578, opened 2024-08-22
fix(autoware_tracking_object_merger): fix unusedFunction

kobayu858:
## Description

```
perception/autoware_tracking_object_merger/src/utils/utils.cpp:301:0: style: The function 'objectsYawIsReverted' is never used. [unusedFunction]
bool objectsYawIsReverted(const TrackedObject & main_obj, const TrackedObject & sub_obj)
^

perception/autoware_tracking_object_merger/src/utils/utils.cpp:413:0: style: The function 'probabilityMerger' is never used. [unusedFunction]
float probabilityMerger(const float main_prob, const float sub_prob, const MergePolicy policy)
^

perception/autoware_tracking_object_merger/src/utils/utils.cpp:428:0: style: The function 'shapeMerger' is never used. [unusedFunction]
autoware_perception_msgs::msg::Shape shapeMerger(
^
```
## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_tracking_object_merger/include/autoware/tracking_object_merger/utils/utils.hpp
+++ b/perception/autoware_tracking_object_merger/include/autoware/tracking_object_merger/utils/utils.hpp
@@ -80,13 +80,6 @@ autoware_perception_msgs::msg::TrackedObjectKinematics objectKinematicsVXMerger(
 TrackedObject objectClassificationMerger(
   const TrackedObject & main_obj, const TrackedObject & sub_obj, const MergePolicy policy);
 
-// probability merger
-float probabilityMerger(const float main_prob, const float sub_prob, const MergePolicy policy);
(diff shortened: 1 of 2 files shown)


technolojin:
For my point of view, those functions provide additional functionalities for Autoware users. Even it is not used in current Autoware, someone can use those.

@YoshiRi @yukkysaito 
Do you think those unused functions should be removed?

YoshiRi:
@technolojin 
As you say I intended to use those function in the future development.
But this is trivial, so I think we can remove this.
Ref:  https://star4.slack.com/archives/C4P0NSMB5/p1724040335846959

Shin-kyoto (approved):

PR #8580, opened 2024-08-22
fix(diagnostic_graph_aggregator): fix unusedFunction

bathteayo:
## Description

This is a fix based on cppcheck unusedFunction warnings.

```
src/common/graph/graph.hpp:37:0: style: The function 'diags' is never used. [unusedFunction]
  const auto & diags() const { return diags_; }
^
```

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


isamu-takagi (approved):
LGTM

PR #8583, opened 2024-08-22
fix(autoware_traffic_light_fine_detector): fix unusedFunction

kobayu858:
## Description

```
perception/autoware_traffic_light_fine_detector/src/traffic_light_fine_detector_node.cpp:52:0: style: The function 'toFloatVector' is never used. [unusedFunction]
inline std::vector<float> toFloatVector(const std::vector<double> & double_vector)
^
```
## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


Shin-kyoto (approved):
LGTM

- I confirmed that `toFloatVector` is not used in autoware.universe. I checked it by searching with `toFloatVector`.
- I did **NOT** check this PR by running autoware.

PR #8584, opened 2024-08-22
refactor(ekf_localizer): fix tf2 listener

YamatoAndo:
## Description
In the ekf_localizer node,  there is a strange way of writing for tf_listener , so I fixed it.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

Lsim works

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/localization/ekf_localizer/include/ekf_localizer/ekf_localizer.hpp
+++ b/localization/ekf_localizer/include/ekf_localizer/ekf_localizer.hpp
@@ -39,6 +39,7 @@
 
 #include <tf2/LinearMath/Quaternion.h>
 #include <tf2/utils.h>
+#include <tf2_ros/buffer.h>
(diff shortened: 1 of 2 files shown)


SakodaShintaro (approved):
Basically, I think it's fine, but I understand that the behavior has changed.
`get_transform_from_tf` used to wait for tf for up to 5 seconds, it doesn't wait after the change.
I just want to remember that this PR may be a problem in very specific cases.

YamatoAndo:
@SakodaShintaro If you're concerned about it, should we implement it using [canTransform](https://docs.ros.org/en/indigo/api/tf2_ros/html/c++/classtf2__ros_1_1Buffer.html#a2971d78ba006fd6c3dd67c53870317fd)?

PR #8588, opened 2024-08-22
feat(raw_vehicle_cmd_converter): disable actuation to steering

kosuke55:
## Description

same reason to https://github.com/autowarefoundation/autoware_launch/pull/1132

----

disable actuation to steering by defalut.

this is new feature introduced in https://github.com/autowarefoundation/autoware.universe/pull/8504
but many user don't kwnow the details.

In practice, the description config is called, and this launcher config is often not used, but for now it defaults to false.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

this config in universe is just example not affect the behavior

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


TakaHoribe (commented):
LGTM. This is recently introduced, and not used anywhere yet.

TakaHoribe (approved):
LGTM. This is recently introduced, and not used anywhere yet.

Issue #8590, opened 2024-08-22
AEB module incorrectly calculates target obstacle's velocity at first calculation


ismetatabay:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I've agreed with the maintainers that I can plan this task.

### Description

Currently, in the default Autoware setup, point cloud data is used as obstacle information, and the nodes calculate the obstacle's velocity to use in the RSS formula. When obstacle points enter the predicted path, a miscalculation of the target object's velocity results in the RSS distance being calculated too high at the beggining. This situation triggers an emergency when the ego vehicle is overtaken by an NPC due to the excessively high calculated RSS distance.

- Calculated RSS distance for a constant velocity overtaking NPC

RSS Distance    |  Calculated Object Velocity for 5.55 m/s NPC  
:-------------------------:|:-------------------------:
<img src="https://github.com/user-attachments/assets/fce550de-11bf-4646-b68a-c038fc449a86" width="100%"/> |  <img src="https://github.com/user-attachments/assets/44b68dab-3cfd-4ede-aa19-53212a92a951" width="100%"/>

### Purpose

Prevent unwanted emergency braking while driving in the city.

### Possible approaches

Extend the object tracking area to start calculating velocity before the object enters the predicted ego footprint area.

### Definition of done

- Start calculating the speed of obstacles before they become the target obstacle.
- If an object isn’t in the predicted area, don’t trigger an emergency.
- Add a debug marker for the object search area.

PR #8591, opened 2024-08-22
feat(autonomous_emergency_braking): calculate the object's velocity in the search area

ismetatabay:
## Description

This PR enables the calculation of object velocity within the speed calculation area, using the ```speed_calculation_expansion_margin``` parameter.

## Related links

**Parent Issue:**

- Link: https://github.com/autowarefoundation/autoware.universe/issues/8590

#### Before This PR: 

https://github.com/user-attachments/assets/f4de0400-b04d-4932-aa2e-916545a398a5

#### After This PR: 

https://github.com/user-attachments/assets/5b7445cf-e399-465f-8349-a7def3186247

Stopping for an obstacle. (obstacle_cruise_planner (motion_stop_planner) is disabled) 

https://github.com/user-attachments/assets/ed2743d2-11e9-44fe-b872-b58c3d720e04

## Notes for reviewers

None.

## Interface changes

None.

#### Additions and removals

| Added | `speed_calculation_expansion_margin`   | `double` | `0.5`         | Lateral margin used to initiate speed calculation for an object. |

## Effects on system behavior

None.


ismetatabay:
Hi @danielsanchezaran -san,
If you have time, could you look at this this PR and Issue?
- https://github.com/autowarefoundation/autoware.universe/issues/8590

kyoichi-sugahara:
@ismetatabay 
Thank you so much for the PR!
The `expanded_ego_polys` are assigned using the extended footprint only when `use_pointcloud_data_` is true, but when `use_pointcloud_data_` is false, they remain empty. 
Even when empty, they are still used for inside/outside determination in the `within` function. Is this intended?

control/autoware_autonomous_emergency_braking/include/autoware/autonomous_emergency_braking/node.hpp
```diff
@@ -392,7 +394,9 @@ class AEB : public rclcpp::Node
    * @param closest_object Data of the closest object
```
kyoichi-sugahara, line 394: @ismetatabay 
Could you modify doxygen explanation?

kyoichi-sugahara:
@ismetatabay 
Shold I stop reviewing? Asking because you reverted to draft.

kyoichi-sugahara:
@ismetatabay 
Ok thank you for quick replay I will leave some comments,

I can confirm from the description that the results have improved, but do you understand why setting the collision result to false when the object's vertices are outside the extended predicted path helps avoid unnecessary activations?
Also, does this change mean that the speed is being calculated appropriately?

Furthermore, if we were to incorporate this change, I think it would be more natural to implement the process of excluding objects that have vertices not contained within the predicted path in the `createObjectDataUsingPointCloudClusters` function, rather than checking if it's within the predicted path during collision detection.

The reason is that it's preferable to separate the filtering process for potential collision objects from the actual collision detection. 
The current PR seems to be performing the filtering process within the function that executes collision detection, which feels inconsistent with the function name and its processing.
By changing, We can avoid the problem in the [comment](https://github.com/autowarefoundation/autoware.universe/pull/8591#issuecomment-2307041633)

I apologize if I've overlooked any circumstances and my comments are unreasonable :bow:

ismetatabay:
@kyoichi-sugahara-san, thank you for your comments. I plan to refactor my implementation and keep this PR as a draft. My current implementation steps are as follows:

-> Use rough point cloud filtering to keep the velocity information of the closest object in this area.
-> If the closest target object is not located in the predicted path, don't calculate a collision.

As you mentioned, the ```hasCollision``` function might not need to filter obstacles based on whether they are in the predicted trajectory. I will refactor it accordingly. Thank you again.

ismetatabay:
@kyoichi-sugahara -san, my current idea is:
- Add an ```is_target``` attribute to the ObjectData struct (default to true for predicted objects).
- Add all objects located in the expanded footprint area (with ```path_footprint_extra_margin```) to the object list (to calculate velocity).
- Check if the object is located in the usual predicted path, then update the object's ```is_target``` attribute.
- If the tracked object is not a target (is in the expanded area for velocity calculation but does not enter the predicted path), then don't call the ```hasCollision``` function.

What do you think? 

Additionally, I found a mistake in the control.launch.xml launch file and I fixed this topic as well.:
https://github.com/autowarefoundation/autoware.universe/blob/b9b0fa9dbb4e141519dc8de13d37827763b481f2/launch/tier4_control_launch/launch/control.launch.xml#L195

kyoichi-sugahara:
@ismetatabay 
>If the tracked object is not a target (is in the expanded area for velocity calculation but does not enter the predicted path), then don't call the hasCollision function.

So, does this mean that AEB will no longer explicitly react to objects existing outside the normal predicted path?

From referring to the [issue](https://github.com/autowarefoundation/autoware.universe/issues/8590) (this is super useful result :heart:), I understand that the problem is that the estimation results are unstable for the first few steps after speed calculation begins, causing unnecessary activation.
While reducing the detection targets is a solution, I'm concerned that it might make necessary activations impossible.

From my perspective, I feel the following approaches might be valid solutions:
- Introduce a mechanism that doesn't utilize the first few steps when the speed estimation is clearly unstable.
- Prepare separate areas for speed estimation and for detection targets.

What do you think?

ismetatabay:
@kyoichi-sugahara -san, thank you for your comment :heart: . 

```bash
 So, does this mean that AEB will no longer explicitly react to objects existing outside the normal predicted path? 
```
AEB maintains the same mechanism for reacting to obstacles in the predicted path. The default current implementation uses ```ego_polys``` (predicted path + ```expand_width```) for collision detection and ```expanded_ego_polys``` (```predicted path``` + ```expand_width``` + ```path_footprint_extra_margin```) for point cloud cropping.

AEB drawio

My idea is to calculate the object's velocity when it enters the ```expanded_ego_polys``` area. If the object does not enter the ```ego_polys``` area (currently used for object detection in the default implementation), we can set the ```is_target``` value of this object to false. However, this approach may lead to issues, such as when there are two objects in front of the ego vehicle. If the closer object is within the ```expanded_ego_polys``` and the farther one is within the ```ego_polys```, we might not consider the target object (in ```ego_polys```) because it isn't the closest obstacle.

---

```bash
- Introduce a mechanism that doesn't utilize the first few steps when the speed estimation is clearly unstable.
```
- I think this might cause a slight delay in reaction time, but it could be a quick implementation.

---

```bash
- Prepare separate areas for speed estimation and for detection targets.
```
- I agree with this idea, and I plan to work on it. This area could be ```expanded_ego_polys```, but we need to implement multiple object tracking mechanism I think. I will check. :+1:

kyoichi-sugahara:
@ismetatabay 
> AEB maintains the same mechanism for reacting to obstacles in the predicted path. 

Yes you are right, and sorry I noticed my comment was based on misunderstanding

> we might not consider the target object (in ego_polys) because it isn't the closest obstacle.

I understand your concern but, may be the further object can be ignored in my opinion. Only closest object in the predicted path polygon should be considered. 

> I think this might cause a slight delay in reaction time, but it could be a quick implementation.

yes, that's the concern for this idea which I was discussing with @danielsanchezaran today

> I agree with this idea, and I plan to work on it. This area could be expanded_ego_polys, but we need to implement multiple object tracking mechanism I think. I will check. 👍

Thanks! I hope it improves the behavior but still 
I feel that it's likely to occur that when a vehicle enters an `expanded_ego_polys` diagonally, the position of the nearest point doesn't change vertically over time, resulting in a low calculated speed.

ismetatabay:
@kyoichi-sugahara -san, Thank you so much for your valuable comments and suggestions. My main concern regarding the closest object is that if we calculate and only consider the closest object within separate areas, we might miss the actual target object. I will work on addressing this in the following case:

AEB2 drawio

Since NPC1 is the closest object, we will calculate its speed, but it is not our target object. NPC2 is the target object because it is located on our trajectory. I will check on it.

kyoichi-sugahara:
@ismetatabay 
> Thank you so much for your valuable comments and suggestions. My main concern regarding the closest object is that if we calculate and only consider the closest object within separate areas, we might miss the actual target object. I will work on addressing this in the following case:

Ok so I will wait for your change, and let's discuss based on your change.

And I think your change related to debug marker is already ready for review and it's very helpful for developers. So how about spliting these change and merge them in advance?

ismetatabay:
@kyoichi-sugahara -san,

> And I think your change related to debug marker is already ready for review and it's very helpful for developers. So how about spliting these change and merge them in advance?

Do you mean the ```expanded_ego_polys``` visualization?

ismetatabay:
@kyoichi-sugahara -san, @danielsanchezaran -san,

My current implementation works as shown in the following video:

https://github.com/user-attachments/assets/6c716b4f-584c-466b-8a11-a93698e1750f

In this implementation, we are considering consider the closest target object in the AEB prediction. If a target object is not found, we consider the closest one. However, with the current target-switching mechanism implemented in this PR, we may encounter the same issue at the beginning while target object switching, as we only calculate the velocity of the closest object (When the target object is changed we start the calculation of velocity again). To overcome this, we need to implement an object-tracking mechanism or follow Sugahara-san's suggestion below:

> Introduce a mechanism that doesn't utilize the first few steps when the speed estimation is clearly unstable.

ismetatabay:
@kyoichi-sugahara -san, @danielsanchezaran -san,

My current implementation works as shown in the following video:

https://github.com/user-attachments/assets/6c716b4f-584c-466b-8a11-a93698e1750f

In this implementation, we are considering consider the closest target object in the AEB prediction. If a target object is not found, we consider the closest one. However, with the current target-switching mechanism implemented in this PR, we may encounter the same issue at the beginning while target object switching, as we only calculate the velocity of the closest object (When the target object is changed we start the calculation of velocity again). To overcome this, we need to implement an object-tracking mechanism or follow Sugahara-san's suggestion below:

> Introduce a mechanism that doesn't utilize the first few steps when the speed estimation is clearly unstable.

ismetatabay:
@kyoichi-sugahara san, @danielsanchezaran -san If you are not available to meet after Planning and Control WG, we can discuss it here.

kyoichi-sugahara:
@ismetatabay 
can you join with this link?
https://tier4-jp.zoom.us/j/85823903685?pwd=a4LnIu4h2vZjR9Ja1Mjfx8Aqs3XqOY.1

ismetatabay:
@kyoichi-sugahara  -san, sorry I missed the invite. We talked with @danielsanchezaran  and decided to discuss things here. If we need to talk more, we will set up a meeting for next Monday. Is that okay with you as well?

danielsanchezaran:
@ismetatabay first of all, thank you so much for this PR I think the implementation side is good, I might make some code suggestions later. 

Regarding what the PR offers, I am only worried about a thing: 
When I implemented the velocity estimation, I deliberately decided to use the ego path polygons instead of the extended polygons. The reason was that, the velocity that I wanted to measure is the velocity of the closest point inside the ego path, and NOT the velocity of the object itself. It might sound like they are the same thing but that is not the case. Take this image for example: 

AEB-rss-distance-emergency drawio

In the image, between timestep 0 (top image) and timestep 1 (bottom), the position of the closest pointcloud point to the ego vehicle does not change in the global/map reference frame, even though the red car is moving forward with the same speed component Vx as the ego vehicle's  (where x is the ego path's direction). I think this is why there is a discrepancy between the NPC's speed and when the AEB first detects the speed of the closest point.

In current Autoware, the AEB would detect the speed of the red car as ~ 0 m/s and probably issue an emergency stop. Now, this might feel unnecessary in this case because the red car is small and it is not really necessary to stop to prevent a collision. 

The problem is, if you have the exact same situation, but replace the red car by a truck or bus, then it might be necessary to issue an emergency stop, which Universe's AEB will do in this case.

AEB-rss-distance-emergency-2 drawio

However, I am afraid that, with the changes of this PR, the AEB might not issue an emergency stop on the scenario with the blue/cyan truck, because, with this PR, the speed Vx of the closest point is measured as how the closest point position changes according to the expanded polygons (> 0 m/s) instead of against the ego's path (0 m/s). However, I am not sure how much this change would affect the performance of the AEB. From a safety point I think that makes the situation with the truck unsafe. 

On the other hand, I also believe that unnecessary stops are risky or unsafe, and if there is a driver behind the ego, he might not expect the ego to suddenly brake in the case of the red car (which would happen with current Autoware but not with this PR). I think it is not really possible to satisfy every possible scenario, though, and I am also not sure which situation is more dangerous/likely. 

I think it might be ok to merge this PR but the difference between object speed and closest point speed should be well addressed on the README. 

Also, IF this PR is merged, I feel that it is necessary to reduce the default size of the expanded polygons (4 meters) to something like 1 meter or 0.5 meters. I think that would be a good middle ground between what this PR offers and the current AEB. Could you perform some tests with a reduced polygon expansion @ismetatabay ? 

What is your opinion? @ismetatabay @kyoichi-sugahara 

Does my explanation make sense?

danielsanchezaran:
@ismetatabay what I suggest is creating a second set of expanded polygons. One set will be used for cropping the point cloud (as in current Autoware) and the other (smaller) set of expanded polygons will be used for velocity calculation. That way we can keep the path used for cropping at 4m of extra expansion and add a separate expanded path for velocity calculation that has only 0.5m expansion, that way, we don't run into the trouble of overcropping the point cloud by making the expanded polygon path too small and we can always make the AEB the same as current AEB by setting the "speed_calculation_expansion_margin" (or however you want to call the parameter) to 0m

ismetatabay:
Hello @danielsanchezaran -san, I agree with your suggestion. Adding a second expanded polygon for velocity calculation (```speed_calculation_expansion_margin```) provides users with some flexibility according to their use cases. I will update this PR today.

cc: @kyoichi-sugahara -san

ismetatabay:
Hello @danielsanchezaran, I tested the ```speed_calculation_expansion_margin``` parameter with values of 0.5, 1.0, 1.5, and 2.0 meters. With 0.5, 1.0, and 1.5 meters, we encountered a similar situation, but 2.0 meters seems acceptable to me. However, I understand your concern, so if you prefer, we can keep it at 0.0 or 0.5 meters. I will update the documentation so that users can adjust this value according to their operating environment. What do you think?

#### ```speed_calculation_expansion_margin``` is set to 0.5 [m]

https://github.com/user-attachments/assets/d728c37a-35ef-44bf-a1a4-cf01e2b121f9

#### ```speed_calculation_expansion_margin``` is set to 2.0 [m]

https://github.com/user-attachments/assets/0ba50749-6f0d-4dee-a51f-f93acf81881d

ismetatabay:
@danielsanchezaran -san,

I want to discuss something related to the current implementation of this PR. We are checking for collisions along the IMU path first if it is enabled:

https://github.com/autowarefoundation/autoware.universe/blob/b967b6fd70a8591e17f558aa150e80517e9d53de/control/autoware_autonomous_emergency_braking/src/node.cpp#L538-L539

This situation raises an issue with the target object switching mechanism. For example, in the picture below, the target object couldn't be selected because its speed couldn't be calculated (see code snippet below), as it is outside of the IMU predicted path.

https://github.com/autowarefoundation/autoware.universe/blob/b967b6fd70a8591e17f558aa150e80517e9d53de/control/autoware_autonomous_emergency_braking/include/autoware/autonomous_emergency_braking/node.hpp#L264-L267



However, if it enters the IMU path, the switch will be successful, as shown in the image:


The simplest way to solve this issue is to change the order in the collision checking, like this (MPC predicted path is longer than IMU predicted path):

```diff
- const bool has_collision =
-   has_collision_ego(filtered_objects) || has_collision_predicted(filtered_objects);
+ const bool has_collision =
+   has_collision_predicted(filtered_objects) || has_collision_ego(filtered_objects);
```



I want to discuss the best way to solve this issue. What do you think about that?

danielsanchezaran:
> @danielsanchezaran -san,
> 
> I want to discuss something related to the current implementation of this PR. We are checking for collisions along the IMU path first if it is enabled:
> 
> https://github.com/autowarefoundation/autoware.universe/blob/b967b6fd70a8591e17f558aa150e80517e9d53de/control/autoware_autonomous_emergency_braking/src/node.cpp#L538-L539
> 
> This situation raises an issue with the target object switching mechanism. For example, in the picture below, the target object couldn't be selected because its speed couldn't be calculated (see code snippet below), as it is outside of the IMU predicted path.
> 
> https://github.com/autowarefoundation/autoware.universe/blob/b967b6fd70a8591e17f558aa150e80517e9d53de/control/autoware_autonomous_emergency_braking/include/autoware/autonomous_emergency_braking/node.hpp#L264-L267
> 
> 
> 
> However, if it enters the IMU path, the switch will be successful, as shown in the image: 
> 
> The simplest way to solve this issue is to change the order in the collision checking, like this (MPC predicted path is longer than IMU predicted path):
> 
> ```diff
> - const bool has_collision =
> -   has_collision_ego(filtered_objects) || has_collision_predicted(filtered_objects);
> + const bool has_collision =
> +   has_collision_predicted(filtered_objects) || has_collision_ego(filtered_objects);
> ```
> 
> 
> 
> I want to discuss the best way to solve this issue. What do you think about that?

Thanks for your message, I believe that we should not base it on order of execution, there are some applications (like small robots) where we might want the imu path to have the same time horizon as the MPC path. I think the speed calculation should not depend on the path that detected the object first. 

I have updated/refactored the AEB code so pointcloud cropping and clustering is done once, maybe its possible to add the switching there?

control/autoware_autonomous_emergency_braking/src/node.cpp
```diff
@@ -850,7 +895,14 @@ void AEB::getClosestObjectsOnPath(
     obj.distance_to_object = std::abs(dist_ego_to_object);
 
     const Point2d obj_point(p.x, p.y);
-    for (const auto & ego_poly : ego_polys) {
+    // Check if the object is in the predicted path for AEB or not
```
danielsanchezaran, line 898: @ismetatabay I feel if might be worthwhile to consider using the objects lateral distance instead of their polygons to include and not include them in the speed calculation and classifying them as targets or not. 

if the points absolute lat distance is less than vehicle_width / 2 + expand_width_ they are a target and include them in the object list, if not, we check if their lat distance to the ego path is less than ego_width_ /2 + expand_width + speed_calc_marging. So it is not necessary to iterate through the 2 sets of polygons. The added benefit is that we could use the change in lat distance to calculate object lateral speeds in the future, to "predict" if the object will enter or leave the ego path before a collision.

ismetatabay: Hello @danielsanchezaran -san,  agree, I updated the PR accordingly. However, I also made changes to the collision checking steps. I merged objects from both the IMU and MPC, and I also merged path points from the IMU and MPC so that the checking mechanism runs only once. What do you think about that?

danielsanchezaran: Thank you @ismetatabay I would like to test the PR and see if there is degradation due to the change from polygon-based collision detection to arc coordinate based, I will take a look at processing time. Please give me some time to review. 

Also, I am sorry for being too bothersome, but maybe do you think it is possible to separate this PR into 2? one for the velocity calculation and one for the arc-coordinate based collision detection?

control/autoware_autonomous_emergency_braking/src/node.cpp
```diff
@@ -462,15 +468,29 @@ bool AEB::checkCollision(MarkerArray & debug_markers)
...
-      get_objects_on_path(path, points_belonging_to_cluster_hulls, debug_colors, debug_ns);
-    // Get only the closest object and calculate its speed
+  auto check_collision = [&](const Path & path, std::vector<ObjectData> & objects) {
+    time_keeper_->start_track("has_collision");
     const auto closest_object_point = std::invoke([&]() -> std::optional<ObjectData> {
+      // Attempt to find the closest target object
+      const auto closest_target_object_itr =
+        std::min_element(objects.begin(), objects.end(), [](const auto & o1, const auto & o2) {
+          return (o1.is_target && (!o2.is_target || o1.distance_to_object < o2.distance_to_object));
+        });
```
danielsanchezaran, line 478: If o1 and o2 are NOT targets, doesn't that make o2 always the "closest" object regardless of distance? Maybe the logic should include all cases? 

```suggestion
//target objects have priority
if (o1.is_target != o2.is_target) {
    return o1.is_target;
}
return o1.distance_to_object < o2.distance_to_object;       
```

danielsanchezaran (commented):
This PR looks very promising, please let me know your opinion about the comments/reviews I made. 

I tested the PR is simulation and it looks good, however processing time is a concern (I believe the code should be faster with this PR, but benchmarking is necessary)

control/autoware_autonomous_emergency_braking/src/node.cpp
```diff
@@ -529,27 +550,45 @@ bool AEB::checkCollision(MarkerArray & debug_markers)
...
-    const std::string ns = "imu";
-    return check_collision(ego_imu_path, debug_color, ns, points_belonging_to_cluster_hulls);
-  };
+  const auto imu_path_objects = (!use_imu_path_ || !angular_velocity_ptr_)
+                                  ? std::vector<ObjectData>{}
+                                  : get_objects_on_path(
+                                      ego_imu_path, points_belonging_to_cluster_hulls,
+                                      {0.0 / 256.0, 148.0 / 256.0, 205.0 / 256.0, 0.999}, "imu");
+
+  const auto mpc_path_objects = (!use_predicted_trajectory_ || !predicted_traj_ptr_)
```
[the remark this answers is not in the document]
ismetatabay: Done :+1:

control/autoware_autonomous_emergency_braking/src/node.cpp
```diff
@@ -529,27 +550,45 @@ bool AEB::checkCollision(MarkerArray & debug_markers)
...
+
+  // merge object data which comes from the ego (imu) path and predicted path
+  auto merge_objects =
+    [&](const std::vector<ObjectData> & imu_objects, const std::vector<ObjectData> & mpc_objects) {
+      std::vector<ObjectData> merged_objects = imu_objects;
+      merged_objects.insert(merged_objects.end(), mpc_objects.begin(), mpc_objects.end());
+      return merged_objects;
+    };
+
+  auto merge_imu_mpc_objects = merge_objects(imu_path_objects, mpc_path_objects);
```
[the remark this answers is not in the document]
ismetatabay: Done :+1:

control/autoware_autonomous_emergency_braking/test/test.cpp
```diff
@@ -32,6 +32,7 @@
 
 namespace autoware::motion::control::autonomous_emergency_braking::test
 {
+using autoware::universe_utils::Point2d;
```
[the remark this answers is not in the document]
ismetatabay: Removed :+1:

control/autoware_autonomous_emergency_braking/test/test.cpp
```diff
@@ -40,6 +41,7 @@ using geometry_msgs::msg::Pose;
 using geometry_msgs::msg::TransformStamped;
 using geometry_msgs::msg::Vector3;
 using std_msgs::msg::Header;
+namespace bg = boost::geometry;
```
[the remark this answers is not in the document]
ismetatabay: Removed :+1:

control/autoware_autonomous_emergency_braking/src/node.cpp
```diff
@@ -827,7 +871,7 @@ void AEB::getClosestObjectsOnPath(
   const PointCloud::Ptr points_belonging_to_cluster_hulls, std::vector<ObjectData> & objects)
 {
   autoware::universe_utils::ScopedTimeTrack st(__func__, *time_keeper_);
-  // check if the predicted path has valid number of points
+  // check if the predicted path has a valid number of points
   if (ego_path.size() < 2 || ego_polys.empty() || points_belonging_to_cluster_hulls->empty()) {
```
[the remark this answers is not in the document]
ismetatabay: @danielsanchezaran,

Thank you for your valuable reviews and comments. I think it would be beneficial to add a new parameter for the collision detection method (polygons or arc coordinates). In my opinion, it would be better to implement this in a separate PR. After this PR, I can work on that :+1:

control/autoware_autonomous_emergency_braking/src/node.cpp
ismetatabay: I agree, updated :+1:

danielsanchezaran:
@ismetatabay it looks like there might be some trouble with using arc coordinates. I cannot really tell why, but detection seems to happen when it should not. I am using the default params. 

https://github.com/user-attachments/assets/b13d3710-2abd-4287-bf5e-c3010f6d021a

control/autoware_autonomous_emergency_braking/src/node.cpp
```diff
@@ -462,15 +467,33 @@ bool AEB::checkCollision(MarkerArray & debug_markers)
...
+                                                *closest_target_object_itr, path, current_v)
+                                            : std::make_optional<double>(0.0);
+
+        if (closest_object_speed.has_value()) {
+          closest_target_object_itr->velocity = closest_object_speed.value();
+          return std::make_optional<ObjectData>(*closest_target_object_itr);
+        }
+      }
+
+      // If no target object is found, find the closest object overall
```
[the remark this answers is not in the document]
ismetatabay: No worries, let me check it :+1:

ismetatabay:
Hello @danielsanchezaran -san,

I couldn't reproduce the issue you encountered with the planning simulator. If possible, could you share the bag and map files? so I can check the problem

https://github.com/user-attachments/assets/766e0324-d645-4700-9a61-0c629236aa43

ismetatabay:
@danielsanchezaran 
I have concerns about the ground segmentation performance in your video. Some false positives from ground segmentation in the target area might trigger the emergency. Could you check it?

image

image

control/autoware_autonomous_emergency_braking/src/node.cpp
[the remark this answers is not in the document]
ismetatabay: Removed :+1:

danielsanchezaran:
> @danielsanchezaran I have concerns about the ground segmentation performance in your video. Some false positives from ground segmentation in the target area might trigger the emergency. Could you check it?
> 
> image
> 
> image

@ismetatabay thank you for your input. I will try and see if I can reproduce the problem reliably or if it is a tuning problem. The ground segmentation does indeed look suspicious, but I also tried setting the minimum cluster height to 0.1 m to mitigate that, with no luck. Please give me some more time to do some tests.

danielsanchezaran:
@ismetatabay I discovered what was the issue. 

Some of the MPC path points are virtually the same when the ego is starting to move, that causes the calcLatoffset function to misbehave and not caluclate the proper lateral distance. To solve the issue, we can omit mpc path points that are too close to each other:

`--- a/control/autoware_autonomous_emergency_braking/src/node.cpp
+++ b/control/autoware_autonomous_emergency_braking/src/node.cpp
@@ -709,6 +709,9 @@ std::optional<Path> AEB::generateEgoPath(const Trajectory & predicted_traj)
   for (size_t i = 0; i < predicted_traj.points.size(); ++i) {
     geometry_msgs::msg::Pose map_pose;
     tf2::doTransform(predicted_traj.points.at(i).pose, map_pose, transform_stamped.value());
+    if (autoware::universe_utils::calcDistance2d(path.back(), map_pose) < 1e-2) {
+      continue;
+    }
     path.push_back(map_pose);
 `
I would like you to please add this change and also make the minimum distance between points 1e-2 for the IMU path, for consistency. 

Also, please note that calcLateralOffset might return NaN, so please handle that case by adding a 
`    if (std::isnan(lateral_offset)) continue;
`
right after calculating the lat offset on the AEB::getClosestObjectsOnPath function. 

As a note, the calcLateralOffset function checks if 2 points are too close, but the epsilon value used (1e-8 might be too small. I hopefully will make a PR next week to fix it). 

On the mean time, I will approve this PR, I think it is a great addition to AEB and I am grateful for it, but please fix the issues I pointed out before merging!!!!

danielsanchezaran (approved):
Approved, but please fix the issues on my last comment before merging, the rest LGTM

ismetatabay:
@danielsanchezaran Thank you so much for your efforts. Before merging it, I want to test it on a real vehicle as well. I will update you. Have a good weekend!

danielsanchezaran:
> @danielsanchezaran Thank you so much for your efforts. Before merging it, I want to test it on a real vehicle as well. I will update you. Have a good weekend!

Thank you @ismetatabay any idea when you will you be able to do the tests?

control/autoware_autonomous_emergency_braking/src/node.cpp
```diff
@@ -661,13 +692,20 @@ std::optional<Path> AEB::generateEgoPath(const Trajectory & predicted_traj)
...
+    }
+
     path.push_back(map_pose);
 
     if (i * mpc_prediction_time_interval_ > mpc_prediction_time_horizon_) {
       break;
     }
   }
   time_keeper_->end_track("createPath");
+  if (path.empty()) return std::nullopt;
```
ismetatabay, line 708: @danielsanchezaran Hello, I tested it today and things works as expected in lane_driving scenario. However in parking scenario, sometimes the trajectory will be very short and the returned path will be empty. To fix it I pushed this commit: https://github.com/autowarefoundation/autoware.universe/pull/8591/commits/f85fa019f17f225fd1fe2acc21d8f4c9d3aaae49

danielsanchezaran, line 708:
```suggestion
  return (!path.empty()) ? std::make_optional(path) : std::nullopt;
```

danielsanchezaran (commented):
I believe this is more concise

control/autoware_autonomous_emergency_braking/src/node.cpp
danielsanchezaran: @ismetatabay thank you, that is great. So I understand you tested in on the real vehicle right?

PD: I added a small recommendation regarding this line.

ismetatabay: Done :+1:

ismetatabay: @danielsanchezaran Yes, I tested it with ```speed_calculation_expansion_margin``` set to 4.0, and we didn’t stop for any overtaken NPCs during normal lane driving in a campus environment, even though we stopped for a static obstacle at 30 km/h (with -5.0 m/s² emergency deceleration and ```obstacle_cruise_planner``` disabled during the obstacle stop test). Also I pushed your suggestion :+1:  https://github.com/autowarefoundation/autoware.universe/pull/8591/commits/65f869b21d2b38d3ac5d84575a5176cd57e04c05 . If it is okay for you, I think we can merge it.

danielsanchezaran: @ismetatabay thank you, it looks great, any chance you might update the discussion about AEB with the new tests ?https://github.com/orgs/autowarefoundation/discussions/5020#discussioncomment-10312780

Also, please merge the changes!

ismetatabay: @danielsanchezaran Thank you as well. Our research vehicle is undergoing a hardware upgrade, and I will likely update the report next week. :+1:

PR #8599, opened 2024-08-23
fix(ndt_scan_matcher): fix timestamp in service_ndt_align func

YamatoAndo:
## Description

The timestamp of `/initialpose3d` was always set to 0, so I investigated the cause and made the necessary fixes.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?
Lsim works.

and check the timestamp of `/initialpose3d`.

```
ros2 topic echo /initialpose3d --field header
```

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/localization/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp
+++ b/localization/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp
@@ -961,7 +961,8 @@ void NDTScanMatcher::service_ndt_align_main(
   diagnostics_ndt_align_->add_key_value("is_succeed_transform_initial_pose", true);
 
   // transform pose_frame to map_frame
-  const auto initial_pose_msg_in_map_frame = transform(req->pose_with_covariance, transform_s2t);
+  auto initial_pose_msg_in_map_frame = transform(req->pose_with_covariance, transform_s2t);
+  initial_pose_msg_in_map_frame.header.stamp = req->pose_with_covariance.header.stamp;
   map_update_module_->update_map(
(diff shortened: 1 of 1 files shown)


localization/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp
```diff
@@ -961,7 +961,8 @@ void NDTScanMatcher::service_ndt_align_main(
   diagnostics_ndt_align_->add_key_value("is_succeed_transform_initial_pose", true);
 
   // transform pose_frame to map_frame
-  const auto initial_pose_msg_in_map_frame = transform(req->pose_with_covariance, transform_s2t);
+  auto initial_pose_msg_in_map_frame = transform(req->pose_with_covariance, transform_s2t);
+  initial_pose_msg_in_map_frame.header.stamp = req->pose_with_covariance.header.stamp;
```
SakodaShintaro, line 965: [Note]

transform : https://github.com/autowarefoundation/autoware.universe/blob/18b5ee13c446cd8cb2e172214ac49ea6f29bbc84/localization/localization_util/include/localization_util/util_func.hpp#L70-L76

tf2::doTransform


https://github.com/ros2/geometry2/blob/ae8a167e10ee65b4a44a9d7b71cd23d2579215f8/tf2_geometry_msgs/include/tf2_geometry_msgs/tf2_geometry_msgs.hpp#L789-L814

The `header` is copied from `transform`.

SakodaShintaro (approved):

PR #8602, opened 2024-08-23
feat(autoware_map_based_prediction): improve frenet path generation

technolojin:
## Description

1. Add backlash within lanelet when predict paths. (Background: Small bikes are tend to drive side within the lane. However, the predictor yields the paths always to the center of the lane.)

* Before : The target path is to center line position of the target lanelet.
* After : The target path allows offset if the momentum (lateral position + 0.5 * lateral velocity * lateral_duration) is within the backlash width. The backlash width is determined by the path width - the object width. The bias is modeled by a continuous function.

Before

https://github.com/user-attachments/assets/ecfebfdf-47da-4fc3-bb44-e57bc71a9cca

https://github.com/user-attachments/assets/84b1d6b9-be35-4f28-8c25-df0b9eb479f9

After

https://github.com/user-attachments/assets/c4ca2ebd-44a5-429d-8877-2644bc9a8457

https://github.com/user-attachments/assets/573753aa-907c-47bf-bb60-d3abdd757661

2.  Considering proper starting reference path for Frenet path generation

Since the reference path and the starting point of the Frenet path is important, related logic has been changed.

* Before: Search the closest path position and calculate the Frenet path boundary condition. If the angle of the closest path is steep, the generated path could be extremely curved.
* After: The starting path in the reference path will be found considering the object motion. If the path is not suitable to generate proper path, the path prediction will be replaced by a straight path. 

Before

https://github.com/user-attachments/assets/d752db4f-911e-4543-ab4c-a3f20de7beb5

https://github.com/user-attachments/assets/5b6a5600-2262-4456-bb6c-c88ea078bddb

After

https://github.com/user-attachments/assets/3f5d20ed-ecbc-4617-a3a4-ae3c6db936a0

https://github.com/user-attachments/assets/c9fc76ea-1076-4a75-9ed5-036886ee19d0

3. Interpolation and extrapolation of the path generation is customized to support 2

4, Bug fix related to the initial position of the Frenet path generation

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

[TIER IV Cloud](https://evaluation.tier4.jp/evaluation/reports/7a00c6c1-d977-5d10-8990-ead641680a61?project_id=prd_jt)
4127/4130

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_map_based_prediction/include/map_based_prediction/map_based_prediction_node.hpp
+++ b/perception/autoware_map_based_prediction/include/map_based_prediction/map_based_prediction_node.hpp
@@ -127,6 +127,7 @@ struct PredictedRefPath
 {
   float probability;
   double speed_limit;
+  double width;
   PosePath path;
   Maneuver maneuver;
 };
@@ -315,9 +316,11 @@ private:
     const Maneuver & maneuver, std::vector<PredictedRefPath> & reference_paths,
     const double speed_limit = 0.0);
 
-  mutable universe_utils::LRUCache<lanelet::routing::LaneletPaths, std::vector<PosePath>>
+  mutable universe_utils::LRUCache<
+    lanelet::routing::LaneletPaths, std::vector<std::pair<PosePath, double>>>
     lru_cache_of_convert_path_type_{1000};
-  std::vector<PosePath> convertPathType(const lanelet::routing::LaneletPaths & paths) const;
+  std::vector<std::pair<PosePath, double>> convertPathType(
+    const lanelet::routing::LaneletPaths & paths) const;
 
   void updateFuturePossibleLanelets(
     const TrackedObject & object, const lanelet::routing::LaneletPaths & paths);
--- a/perception/autoware_map_based_prediction/include/map_based_prediction/map_based_prediction_node.hpp
+++ b/perception/autoware_map_based_prediction/include/map_based_prediction/map_based_prediction_node.hpp
@@ -292,6 +292,8 @@ private:
     const std::string & object_id);
   std::string tryMatchNewObjectToDisappeared(
     const std::string & object_id, std::unordered_map<std::string, TrackedObject> & current_users);
+  bool searchProperStartingRefPathIndex(
+    const TrackedObject & object, const PosePath & pose_path, size_t & index) const;
   std::vector<PredictedRefPath> getPredictedReferencePath(
     const TrackedObject & object, const LaneletsData & current_lanelets_data,
     const double object_detected_time, const double time_horizon);
--- a/perception/autoware_map_based_prediction/include/map_based_prediction/map_based_prediction_node.hpp
+++ b/perception/autoware_map_based_prediction/include/map_based_prediction/map_based_prediction_node.hpp
@@ -292,8 +292,8 @@ private:
     const std::string & object_id);
   std::string tryMatchNewObjectToDisappeared(
     const std::string & object_id, std::unordered_map<std::string, TrackedObject> & current_users);
-  bool searchProperStartingRefPathIndex(
-    const TrackedObject & object, const PosePath & pose_path, size_t & index) const;
+  std::optional<size_t> searchProperStartingRefPathIndex(
+    const TrackedObject & object, const PosePath & pose_path) const;
   std::vector<PredictedRefPath> getPredictedReferencePath(
     const TrackedObject & object, const LaneletsData & current_lanelets_data,
     const double object_detected_time, const double time_horizon);
--- a/perception/autoware_map_based_prediction/src/map_based_prediction_node.cpp
+++ b/perception/autoware_map_based_prediction/src/map_based_prediction_node.cpp
@@ -2011,29 +2011,46 @@ std::vector<PredictedRefPath> MapBasedPredictionNode::getPredictedReferencePath(
       st_ptr = std::make_unique<ScopedTimeTrack>("searching_refpath_starting_point", *time_keeper_);
 
     auto & pose_path = ref_path.path;
+    if (pose_path.empty()) {
+      continue;
+    }
 
     size_t starting_segment_idx;
     {
-      // starting segment index is the nearest segment index from the object
+      // starting segment index is a segment close enough to the object
       const auto obj_point = object.kinematics.pose_with_covariance.pose.position;
       {
         std::unique_ptr<ScopedTimeTrack> st_ptr;
         if (time_keeper_)
-          st_ptr = std::make_unique<ScopedTimeTrack>("findNearestSegmentIndex", *time_keeper_);
-        starting_segment_idx =
-          autoware::motion_utils::findNearestSegmentIndex(pose_path, obj_point);
+          st_ptr = std::make_unique<ScopedTimeTrack>("find_close_segment_index", *time_keeper_);
+
+        starting_segment_idx = 0;
+        double min_dist_sq = std::numeric_limits<double>::max();
+        constexpr double acceptable_dist_sq = 1.0;  // [m2]
+        for (size_t i = 0; i < pose_path.size(); i++) {
+          const double dx = pose_path.at(i).position.x - obj_point.x;
+          const double dy = pose_path.at(i).position.y - obj_point.y;
+          const double dist_sq = dx * dx + dy * dy;
+          if (dist_sq < min_dist_sq) {
+            min_dist_sq = dist_sq;
+            starting_segment_idx = i;
+          }
+          if (dist_sq < acceptable_dist_sq) {
+            break;
+          }
+        }
       }
-      size_t idx = 0;
 
       // calculate score that object can reach the target path smoothly, and search the
       // starting segment index that have the highest score
+      size_t idx = 0;
       {  // find target segmentation index
         std::unique_ptr<ScopedTimeTrack> st_ptr;
         if (time_keeper_)
           st_ptr = std::make_unique<ScopedTimeTrack>("find_target_seg_index", *time_keeper_);
 
         const double obj_yaw = tf2::getYaw(object.kinematics.pose_with_covariance.pose.orientation);
-        constexpr double search_distance = 30.0;  // [m]
+        constexpr double search_distance = 32.0;  // [m]
         const int search_segment_count = std::floor(search_distance / reference_path_resolution_);
         const uint search_segment_num =
           std::min(search_segment_count, static_cast<int>(pose_path.size() - starting_segment_idx));
@@ -2065,8 +2082,9 @@ std::vector<PredictedRefPath> MapBasedPredictionNode::getPredictedReferencePath(
           constexpr double dist_to_yaw_ratio = 0.0001;  // [rad2/m2]
           double score = delta_yaw * delta_yaw + dist_to_yaw_ratio * distance_sq;
 
+          constexpr double acceptable_score = 1e-3;
           if (score < best_score) {
-            if (score < 1e-3) {
+            if (score < acceptable_score) {
               // if the score is small enough, we can break the loop
               idx = i;
               break;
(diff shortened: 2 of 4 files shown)


perception/autoware_map_based_prediction/src/map_based_prediction_node.cpp
```diff
@@ -1114,7 +1115,7 @@ void MapBasedPredictionNode::objectsCallback(const TrackedObjects::ConstSharedPt
 
         // If predicted reference path is empty, assume this object is out of the lane
         if (ref_paths.empty()) {
-          PredictedPath predicted_path = path_generator_->generatePathForLowSpeedVehicle(
```
technolojin, line 1117: @YoshiRi @soblin 
Is `generatePathForLowSpeedVehicle` right for the empty ref_paths object?
I guess it should be `generatePathForOffLaneVehicle`.

soblin:
I confirmed T4 CI results, no problem
https://evaluation.tier4.jp/evaluation/reports/7a00c6c1-d977-5d10-8990-ead641680a61?project_id=prd_jt

perception/autoware_map_based_prediction/src/path_generator.cpp
```diff
@@ -175,17 +175,35 @@ PredictedPath PathGenerator::generatePathForOffLaneVehicle(
...
+  if (object.kinematics.twist_with_covariance.twist.linear.x < 0.0) {
     return generateStraightPath(object, duration);
   }
 
-  return generatePolynomialPath(object, ref_paths, duration, lateral_duration, speed_limit);
+  // get object width
+  double object_width = 5.0;  // a large number
+  if (
+    object.shape.type == autoware_perception_msgs::msg::Shape::BOUNDING_BOX ||
+    object.shape.type == autoware_perception_msgs::msg::Shape::CYLINDER) {
```
YoshiRi, line 197: [to be fixed]
CYLINDER object won't have `y` value. Please use `x` for CYLINDER.

```idl
      @verbatim (language="comment", text=
        " x: the length of the object (BOUNDING_BOX) or diameter (CYLINDER)"
        " y: the width of the object (BOUNDING_BOX)"
        " z: the overall height of the object")
```

perception/autoware_map_based_prediction/src/path_generator.cpp
```diff
@@ -175,17 +175,35 @@ PredictedPath PathGenerator::generatePathForOffLaneVehicle(
...
+    object_width = object.shape.dimensions.y;
+  }
+  // calculate backlash_width, which may the object can be biased from the reference path
+  constexpr double margin = 0.5;  // margin of 0.5m
```
YoshiRi, line 201: [imho]
I can imagine what this margin means, but it seems to be hard to understand for other people.
Need to update README to explain this backlash based method, I think.

It may okay to be done in another PR.

perception/autoware_map_based_prediction/src/path_generator.cpp
```diff
@@ -210,37 +228,65 @@ PredictedPath PathGenerator::generateStraightPath(
...
-  const auto current_point = getFrenetPoint(object, ref_path, duration, speed_limit);
+  const auto current_point = getFrenetPoint(object, ref_path.at(0), duration, speed_limit);
 
-  // Step1. Set Target Frenet Point
+  // Step 1. Set Target Frenet Point
   // Note that we do not set position s,
   // since we don't know the target longitudinal position
   FrenetPoint terminal_point;
-  terminal_point.s_vel = current_point.s_vel;
+  terminal_point.s_vel = std::hypot(current_point.s_vel, current_point.d_vel);
```
YoshiRi, line 244: You mean that the vehicle should keep its speed norm at the finite state, right? If you agree, just resolve this comment.

perception/autoware_map_based_prediction/src/path_generator.cpp
```diff
@@ -210,37 +228,65 @@ PredictedPath PathGenerator::generateStraightPath(
...
   terminal_point.s_acc = 0.0;
-  terminal_point.d = 0.0;
   terminal_point.d_vel = 0.0;
   terminal_point.d_acc = 0.0;
 
-  // Step2. Generate Predicted Path on a Frenet coordinate
+  // calculate terminal d position, based on backlash width
+  {
+    if (backlash_width < 0.01 /*m*/) {
```
YoshiRi, line 251: [imo]
Need comments to explain the cases. like...

1. when backlash is close to 0 = vehicle is close to the boundary -> converge to center
2. Else, calc coeff and momentum to judge wheather objects come back to the center

perception/autoware_map_based_prediction/src/path_generator.cpp
```diff
@@ -210,37 +228,65 @@ PredictedPath PathGenerator::generateStraightPath(
...
+      terminal_point.d = 0.0;
+    } else {
+      constexpr double return_width = 1.5;  // [m]
+      const double return_coeff =
+        return_width / backlash_width;  // if the object is biased too much, back to the center path
```
YoshiRi, line 256: How this 1.5m decided? What about just use backlash length to switch cases?
Also you'll need to fix `(offset_ratio >= 1 && offset_ratio < return_coeff + 1)` in L265.

If using return_coeff is a better verbalization, I will resolve this comment.

perception/autoware_map_based_prediction/src/path_generator.cpp
technolojin: Yes, it is obvious when the object encountered a curved target lane. The vehicle may keep the speed, not longitudinal speed to some lane direction.

perception/autoware_map_based_prediction/src/path_generator.cpp
technolojin: resolved https://github.com/autowarefoundation/autoware.universe/pull/8602/commits/92851eed0e079ec09e1b415f5e1d39d39f9487bb

perception/autoware_map_based_prediction/src/map_based_prediction_node.cpp
```diff
@@ -2495,6 +2607,14 @@ std::vector<PosePath> MapBasedPredictionNode::convertPathType(
...
         prev_p = current_p;
       }
+
+      // Update minimum width
+      const auto left_bound = lanelet.leftBound2d();
+      const auto right_bound = lanelet.rightBound2d();
+      const double lanelet_width_front = std::hypot(
+        left_bound.front().x() - right_bound.front().x(),
+        left_bound.front().y() - right_bound.front().y());
+      width = std::min(width, lanelet_width_front);
```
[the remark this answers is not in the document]
technolojin: Is there any original attribute of 'minimum lane width'? Please let me know reference of that.

perception/autoware_map_based_prediction/src/map_based_prediction_node.cpp
```diff
@@ -2002,6 +2004,113 @@ std::vector<PredictedRefPath> MapBasedPredictionNode::getPredictedReferencePath(
...
+
+    if (is_position_found) {
+      // Trim the reference path
+      pose_path.erase(pose_path.begin(), pose_path.begin() + starting_segment_idx);
+      ++it;
+    } else {
+      it = all_ref_paths.erase(it);
+    }
+  }
+
```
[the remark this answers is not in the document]
technolojin: Not only this new part, overall code should be refactored. The refactoring shall be done in multiple PRs.

perception/autoware_map_based_prediction/src/path_generator.cpp
technolojin: comment is added https://github.com/autowarefoundation/autoware.universe/pull/8602/commits/5222c9d8b800da62174dad6f1c91397b9dd7e4d6

perception/autoware_map_based_prediction/src/path_generator.cpp
technolojin: comment is added https://github.com/autowarefoundation/autoware.universe/pull/8602/commits/88bd9da91c2cfe53f39eacf3e9da540ec4a9d896

perception/autoware_map_based_prediction/src/path_generator.cpp
technolojin: This part is to return the target offset to zero when the target lane is a side lane, or the object is outside of any lane.

A. 1.5 m is an half width of the general lane width of 3m. It is expensive or impossible to get proper width of switching lane width. Therefore it is an empirical value.

B. `(offset_ratio >= 1 && offset_ratio < return_coeff + 1)` = `(current_momentum_d >= backlash_width && current_momentum_d  <  backlash_width + return_width )`, which is as my intention. 
The bias will be zero if `current_momentum_d` is outside of the backlash width + the return width.

perception/autoware_map_based_prediction/src/map_based_prediction_node.cpp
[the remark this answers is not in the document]
YoshiRi: "width" is an optional attribute and I was not meant to use it. I mean "If there is narrower parts in the lane than the front, would it make some trouble in latter process?". I think you have already prepared escape in such a case.

perception/autoware_map_based_prediction/src/path_generator.cpp
YoshiRi: Thanks!

- In my sense, `(current_momentum_d >= backlash_width && current_momentum_d  <  backlash_width + return_width )` is easier to understand.
- In my understandings, `return_width` originally means the half of neighbor lanelet width. Then can we use current lanelet width instead?(But the function do not have access to current lanelet size :face_with_diagonal_mouth: )

YoshiRi:
Worst case performance in my laptop seems to be OK.

```
⏰ Worst Case Execution Time ⏰
100.00% objectsCallback: total 11.79 [ms], avg. 11.79 [ms], run count: 1
```

perception/autoware_map_based_prediction/src/path_generator.cpp
technolojin: 1. resolved in https://github.com/autowarefoundation/autoware.universe/pull/8602/commits/7b24da2231d9c9647bf781a4dbf2476feffcc366
2. the return width will be calculated as a half of the path width https://github.com/autowarefoundation/autoware.universe/pull/8602/commits/3b85890f7b6bdd9063d0f6d7c747724eedc9a24f 

Test is running https://evaluation.tier4.jp/evaluation/reports/983591ff-2753-5590-a49d-de3163aa7100?project_id=prd_jt

perception/autoware_map_based_prediction/src/map_based_prediction_node.cpp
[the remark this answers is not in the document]
technolojin: this part is separated as a method in https://github.com/autowarefoundation/autoware.universe/pull/8602/commits/63595a3d5b8e8f39a7bbe401c9b7e42a4c0269dd

YoshiRi (approved):
LGTM

perception/autoware_map_based_prediction/include/map_based_prediction/map_based_prediction_node.hpp
```diff
@@ -291,6 +292,8 @@ class MapBasedPredictionNode : public rclcpp::Node
     const std::string & object_id);
   std::string tryMatchNewObjectToDisappeared(
     const std::string & object_id, std::unordered_map<std::string, TrackedObject> & current_users);
+  bool searchProperStartingRefPathIndex(
+    const TrackedObject & object, const PosePath & pose_path, size_t & index) const;
```
soblin, line 296: can you change this function to return `optional<size_t>` as returned index value ?

soblin (approved):

perception/autoware_map_based_prediction/include/map_based_prediction/map_based_prediction_node.hpp
technolojin: resolved https://github.com/autowarefoundation/autoware.universe/pull/8602/commits/f3bfec1d3ae3eb6bd09fc7a10b2d0176d3a4c420

PR #7436, opened 2024-06-11
fix(lane_change): fix state transition in lane change module

rej55:
## Description

Before this change, lane change module sometimes can not transit from `RUNNING` state.
In this change, I fixed the state transition problem in the lane change module.

Furthermore, I fixed the state transition problem that it transits `FAILURE` state if the lane change module finishes aborting maneuver.

## Tests performed

[TIER IV INTERNAL LINK](https://evaluation.tier4.jp/evaluation/reports/4a3e9007-c88b-5ab5-ab81-e8584ca3c683?project_id=prd_jt)

## Interface changes

Not applicable.

### ROS Topic Changes

Not applicable.

### ROS Parameter Changes

Not applicable.

## Effects on system behavior

Not applicable.

## Pre-review checklist for the PR author

The PR author **must** check the checkboxes below when creating the PR.

- [x] I've confirmed the [contribution guidelines].
- [x] The PR follows the [pull request guidelines].

## In-review checklist for the PR reviewers

The PR reviewers **must** check the checkboxes below before approval.

- [ ] The PR follows the [pull request guidelines].
- [ ] The PR has been properly tested.
- [ ] The PR has been reviewed by the code owners.

## Post-review checklist for the PR author

The PR author **must** check the checkboxes below before merging.

- [ ] There are no open discussions or they are tracked via tickets.
- [ ] The PR is ready for merge.

After all checkboxes are checked, anyone who has write access can merge the PR.

[contribution guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/
[pull request guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/pull-request-guidelines/

PR #8604, opened 2024-08-24
fix(lane_change): update rtc status for some failure condition

go-sakayori:
## Description
 In this previous PR https://github.com/autowarefoundation/autoware.universe/pull/7436, some conditions were added so that the state will transit from `RUNNING` to `FAILURE`. The module itself would transit to the appropriate state however, the RTC cooperateStatus would not be updated. In this PR I have resolved this problem by calling the `updateRTCStatus` function by providing the `FAILURE` status.

## Related links
https://github.com/autowarefoundation/autoware.universe/pull/7436

## How was this PR tested?
https://evaluation.tier4.jp/evaluation/reports/1f0f0af5-d1b1-579a-890b-be2aa856e208?project_id=prd_jt

## Notes for reviewers
None.

## Interface changes
None.

## Effects on system behavior
None.


maxime-clem (approved):
the change looks good to me and a quick test in Psim revealed no issue.

PR #8616, opened 2024-08-26
fix(tier4_planning_rviz_plugin): fix unusedFunction

kobayu858:
## Description

```
common/tier4_planning_rviz_plugin/src/tools/jsk_overlay_utils.cpp:65:0: style: The function 'getPixelBuffer' is never used. [unusedFunction]
Ogre::HardwarePixelBufferSharedPtr ScopedPixelBuffer::getPixelBuffer()
^
```

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/common/tier4_planning_rviz_plugin/src/tools/jsk_overlay_utils.cpp
+++ b/common/tier4_planning_rviz_plugin/src/tools/jsk_overlay_utils.cpp
@@ -62,11 +62,6 @@ ScopedPixelBuffer::~ScopedPixelBuffer()
   pixel_buffer_->unlock();
 }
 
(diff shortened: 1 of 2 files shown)


kobayu858:
@takayuki5168 @yukkysaito 

I believe this function can be removed since it is not actually used.

Also, there are similar parts in other packages, but I plan to issue a separate PR after this one, since cppcheck will have to resolve them one by one for its convenience.

kobayu858:
If it is used anywhere, it would be appreciated if you could let us know.

kobayu858:
@takayuki5168 @yukkysaito 

takayuki5168 (approved):

PR #8622, opened 2024-08-26
feat(ekf_localizer): input ekf_dt to simple1dfilter

YamatoAndo:
## Description

Instead of calculating dt within Simple1DFilter, I modified it to pass `ekf_dt_` as an argument.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


YamatoAndo:
@reviewers I'm not confident that this fix is correct. 
So, I would like everyone to consider what the correct approach should be.

SakodaShintaro (commented):
This change makes the "Simple1DFilter" class a ROS-free implementation. I think it's admirable trom view point about readability.

Are you concerned that `ekf_dt_` is not equal to `pose.header.stamp - latest_time_`?

YamatoAndo:
I’m more concerned about potential issues when there are two pose sources, such as NDT and GNSS.
Depending on the output timing of NDT and GNSS poses, the calculation of `pose.header.stamp - latest_time_` could result in a negative value, which seems like an incorrect implementation.
In this case, I'm not confident that `ekf_dt_` is correct.

TaikiYamada4:
@YamatoAndo 
I believe there is nothing wrong with using `ekf_dt_` in the update function. I understand that `dt` in the `update` function determines the "variance of prediction." This variance should be larger as the timer period increases, because the longer the time, the more difficult the prediction becomes. In a traditional Kalman filter, the variance of prediction doesn’t need to be affected by the observation, regardless of how the observation appears.

In terms of multiple pose sources, it still seems natural to me that ekf_dt_ is used in the update function, and it doesn't significantly affect the system, even if multiple pose observations are stacked in the queue. If the timer period becomes longer for some reason, all the pose observations in the queue should have a larger impact.

Perhaps the concern arises from the fact that current 1D filters don't update the variance during delay compensation. If the observations in the queue are too old and the variance doesn’t increase during delay compensation, unexpected results may occur.

SakodaShintaro:
I'll perform the performance tests for all data.

TIER IV INTERNAL LINK
* https://evaluation.tier4.jp/evaluation/reports/72dcc4cf-12c1-5d60-ab9f-aea42cbc3836?project_id=autoware_dev
* https://evaluation.tier4.jp/evaluation/reports/779f0860-8f91-5529-94fa-db21eb4c12c4?project_id=prd_jt

SakodaShintaro (approved):
I have compared the new code with the old code using 30 data points. Since there is some randomness in the logging_simulator, the results may have some deviation, but it can be interpreted that the performance is almost the same as before.

compare_result

LGTM

PR #8625, opened 2024-08-26
feat(control): use output=both in launch instead of output=screen

takayuki5168:
## Description

use `output="both"` in launch instead of `output="screen"` to track the log not only in the terminal but also in the log file.

The command to modify the files:
```bash
sed -i 's/output="screen"/output="both"/g' "$file"
```

## Related links

TIER IV's internal link: https://star4.slack.com/archives/C4P0NSMB5/p1723112515306229

## How was this PR tested?

simple planning simulator worked.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


shmpwk (approved):
LGTM

PR #8648, opened 2024-08-28
feat(lane_change): improve execution condition of lane change module

mkquda:
## Description

Currently lane change module checks each path candidate for validity by checking if it has enough length to regulatory elements (intersection, crosswalk, traffic light). It can be better to not request LC module when we are already too close to a reg. element and wait until we have passed it.

Additionally in recent changes to LC turn signal activation; LC module is intended to enable the turn signal whenever a lane change is requested. However, currently, lane change will be activated whenever a target lane is found. This can cause confusion near reg. elements.

For those reasons. We would like to modify the activation conditions of LC module to consider reg.elements within range.

## Changes
- Add LC utility function to get the distance to the next closest reg. element
- Add function to check if we are too close to reg. element
- Do not request LC module if too close to reg. element
- If LC module already running **AND** still not approved **AND** too close to reg. element **THEN** deactivate LC module

## Related links

None.

## How was this PR tested?

PSIM
- **Before:** Even though there is no LC path around the reg. element, LC module is continuously running and turn signal is active as ego approaches and passes reg. element.

https://github.com/user-attachments/assets/94090303-31c2-47ae-9280-971f6069f514

- **After:** LC module is deactivated and turn signal turns off as ego approaches and passes reg. element. Once ego is beyond reg. element, LC module and turn signal are activated again.

https://github.com/user-attachments/assets/cfbbbd29-8de0-430f-a717-fd5ce72f0e7e

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

Improved LC behavior at regulatory elements (crosswalks, traffic lights, intersection), to remove possibility of confusion for other vehicles.


planning/behavior_path_planner/autoware_behavior_path_lane_change_module/include/autoware/behavior_path_lane_change_module/utils/calculation.hpp
```diff
@@ -113,6 +117,21 @@ double calc_maximum_prepare_length(const CommonDataPtr & common_data_ptr);
 double calc_ego_dist_to_lanes_start(
   const CommonDataPtr & common_data_ptr, const lanelet::ConstLanelets & current_lanes,
   const lanelet::ConstLanelets & target_lanes);
+
+double calc_minimum_lane_change_length(
+  const LaneChangeParameters & lane_change_parameters, const std::vector<double> & shift_intervals);
+
+double calc_minimum_lane_change_length(
+  const std::shared_ptr<RouteHandler> & route_handler, const lanelet::ConstLanelet & lane,
+  const LaneChangeParameters & lane_change_parameters, Direction direction);
```
zulfaqar-azmi-t4, line 126: use `common_data_ptr` and `lanes` instead  for the input parameter.

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/include/autoware/behavior_path_lane_change_module/scene.hpp
```diff
@@ -109,6 +109,8 @@ class NormalLaneChange : public LaneChangeBase
 
   bool isStoppedAtRedTrafficLight() const override;
 
+  bool is_too_close_to_regulatory_element() const override;
```
zulfaqar-azmi-t4, line 112:
```suggestion
  bool is_too_close_to_regulatory_element() const final;
```

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/include/autoware/behavior_path_lane_change_module/utils/calculation.hpp
mkquda: This function is also used in `avoidance_by_lane_change` module, so I didn't want to change its input parameters, so I only moved it from utils to calculation

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/scene.cpp
```diff
@@ -171,7 +170,36 @@ bool NormalLaneChange::isLaneChangeRequired()
...
+bool NormalLaneChange::is_too_close_to_regulatory_element() const
+{
+  const auto & current_lanes = get_current_lanes();
+
+  if (current_lanes.empty()) return false;
+
+  const auto shift_intervals =
+    getRouteHandler()->getLateralIntervalsToPreferredLane(current_lanes.back());
+  double threshold_distance =
+    calculation::calc_minimum_lane_change_length(*lane_change_parameters_, shift_intervals);
```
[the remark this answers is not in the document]
mkquda: changed to use new function

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/README.md
mkquda, line 1: @zulfaqar-azmi-t4 can you check if these changes to README are sufficient

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/scene.cpp
```diff
@@ -1631,41 +1665,6 @@ bool NormalLaneChange::getLaneChangePaths(
...
-          debug_print_lat("Reject: regulate on traffic light!!");
-          continue;
-        }
-
-        if (utils::traffic_light::isStoppedAtRedTrafficLightWithinDistance(
-              get_current_lanes(), candidate_path.value().path, planner_data_,
-              lane_change_info.length.sum())) {
-          debug_print_lat("Ego is stopping near traffic light. Do not allow lane change");
-          continue;
-        }
```
mkquda, line 1668: @zulfaqar-azmi-t4 can you confirm that its okay to remove these checks, knowing that I modified the activation condition [here](https://github.com/autowarefoundation/autoware.universe/pull/8648/files#diff-59cbf60b192ecddbbd9709016c0b4bd8846eda2ddbc91dc28937762391e13ff0R198) to use the maximum possible lane change length.

zulfaqar-azmi-t4 (approved):
LGTM! Dont forget to add evaluator's result.

PR #8657, opened 2024-08-28
feat(autoware_multi_object_tracker): reduce trigger latency

technolojin:
## Description
Reduce latency of the multi object tracker caused by periodic trigger of the process.

First implementation
https://github.com/autowarefoundation/autoware.universe/pull/6710

Revert of the first implementation
https://github.com/autowarefoundation/autoware.universe/pull/8277
https://github.com/autowarefoundation/autoware.universe/pull/8484

### Issue: Latency due to the timer trigger cycle (and its phase)

image
The multi-object-tracker adds 50ms of latency. The latency was increased from 2-5ms to 15-60ms.

image (1)
The latency is changing on each test. 

The reason of additional latency is due to the phase difference between the detected object (after detection and merge) and the multi object tracker timer. The phase delay is determined at the launch time and it is random.

### Solution
Bringing back high frequency checker of publish timing. 

1. The previous implementation checked the latest publish timing when the main input channel arrival and timer. 
2. The interval was set to keep in the range of 70% to 111% of the main frequency (which is 10 Hz in general).

The new implementation has been changed for stability.

1. The publish timing checking is triggered by the timer only.
2. The interval range is set to 85% to 105%.

## Related links

[TIER IV INTERNAL discussion](https://star4.slack.com/archives/C03QW0GU6P7/p1724416259766929?thread_ts=1724377803.853829&cid=C03QW0GU6P7)
Reported increased latency after https://github.com/autowarefoundation/autoware.universe/pull/8277

## How was this PR tested?

[TIER IV Cloud](https://evaluation.tier4.jp/evaluation/reports/925d6489-bf09-5f8e-a151-290733b6b679?project_id=prd_jt)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_multi_object_tracker/src/multi_object_tracker_node.cpp
+++ b/perception/autoware_multi_object_tracker/src/multi_object_tracker_node.cpp
@@ -225,7 +225,7 @@ void MultiObjectTracker::onTimer()
   const rclcpp::Time current_time = this->now();
 
   // ensure minimum interval: room for the next process(prediction)
-  const double minimum_publish_interval = publisher_period_ * minimumPublishIntervalRatio;
+  const double minimum_publish_interval = publisher_period_ * minimum_publish_interval_ratio;
   const auto elapsed_time = (current_time - last_published_time_).seconds();
   if (elapsed_time < minimum_publish_interval) {
     return;
@@ -236,7 +236,7 @@ void MultiObjectTracker::onTimer()
 
   // if there was no update, publish if the elapsed time is longer than the maximum publish latency
   // in this case, it will perform extrapolate/remove old objects
-  const double maximum_publish_interval = publisher_period_ * maximumPublishIntervalRatio;
+  const double maximum_publish_interval = publisher_period_ * maximum_publish_interval_ratio;
   should_publish = should_publish || elapsed_time > maximum_publish_interval;
 
   // Publish with delay compensation to the current time
--- a/perception/autoware_multi_object_tracker/src/multi_object_tracker_node.cpp
+++ b/perception/autoware_multi_object_tracker/src/multi_object_tracker_node.cpp
@@ -231,13 +231,13 @@ void MultiObjectTracker::onTimer()
     return;
   }
 
-  // if there was update, publish
+  // if there was update after publishing, publish new messages
   bool should_publish = last_published_time_ < last_updated_time_;
 
   // if there was no update, publish if the elapsed time is longer than the maximum publish latency
-  // it will perform extrapolate/remove old objects
-  const double maximum_publish_latency = publisher_period_ * 1.05;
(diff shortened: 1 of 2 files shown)


perception/autoware_multi_object_tracker/src/multi_object_tracker_node.cpp
```diff
@@ -221,21 +224,30 @@ void MultiObjectTracker::onTimer()
 {
   const rclcpp::Time current_time = this->now();
 
-  // get objects from the input manager and run process
-  ObjectsList objects_list;
-  const bool is_objects_ready = input_manager_->getObjects(current_time, objects_list);
-  if (is_objects_ready) {
-    onMessage(objects_list);
+  // ensure minimum interval: room for the next process(prediction)
+  const double minimum_publish_interval = publisher_period_ * 0.85;  // 85% of the period
```
YoshiRi, line 228: [imo]
How about declare these magic number in other place?
Like adding following parameter in the node deceleration.

```c++
  static constexpr double minimumPublishIntervalRatio = 0.85;
  static constexpr double maximumPublishIntervalRatio = 1.05;
```

perception/autoware_multi_object_tracker/src/multi_object_tracker_node.cpp
```diff
@@ -221,21 +224,30 @@ void MultiObjectTracker::onTimer()
...
   }
 
-  // Publish with delay compensation
-  checkAndPublish(current_time);
+  // if there was update after publishing, publish new messages
+  bool should_publish = last_published_time_ < last_updated_time_;
+
+  // if there was no update, publish if the elapsed time is longer than the maximum publish latency
+  // in this case, it will perform extrapolate/remove old objects
+  const double maximum_publish_interval = publisher_period_ * 1.05;  // 105% of the period
```
YoshiRi, line 239: Same comment with above.

YoshiRi (approved):
LGTM but would you mind check my comment?

perception/autoware_multi_object_tracker/src/multi_object_tracker_node.cpp
technolojin: resolved https://github.com/autowarefoundation/autoware.universe/pull/8657/commits/ff8b6464e48b21f9a0743ce796e5a2b358c2482a

perception/autoware_multi_object_tracker/src/multi_object_tracker_node.cpp
technolojin: resolved https://github.com/autowarefoundation/autoware.universe/pull/8657/commits/ff8b6464e48b21f9a0743ce796e5a2b358c2482a

PR #8666, opened 2024-08-29
fix(autoware_behavior_velocity_intersection_module): fix unusedFunction

kobayu858:
## Description

```
planning/behavior_velocity_planner/autoware_behavior_velocity_intersection_module/src/debug.cpp:162:0: style: The function 'createPointMarkerArray' is never used. [unusedFunction]
[[maybe_unused]] visualization_msgs::msg::Marker createPointMarkerArray(
^

planning/behavior_velocity_planner/autoware_behavior_velocity_intersection_module/src/scene_intersection.hpp:330:0: style: The function 'getAssociativeIds' is never used. [unusedFunction]
  const std::set<lanelet::Id> & getAssociativeIds() const { return associative_ids_; }
^

planning/behavior_velocity_planner/autoware_behavior_velocity_intersection_module/src/scene_intersection.hpp:332:0: style: The function 'getOcclusionUUID' is never used. [unusedFunction]
  UUID getOcclusionUUID() const { return occlusion_uuid_; }
^

planning/behavior_velocity_planner/autoware_behavior_velocity_intersection_module/src/scene_intersection.hpp:333:0: style: The function 'getOcclusionSafety' is never used. [unusedFunction]
  bool getOcclusionSafety() const { return occlusion_safety_; }
^

planning/behavior_velocity_planner/autoware_behavior_velocity_intersection_module/src/scene_intersection.hpp:334:0: style: The function 'getOcclusionDistance' is never used. [unusedFunction]
  double getOcclusionDistance() const { return occlusion_stop_distance_; }
^

planning/behavior_velocity_planner/autoware_behavior_velocity_intersection_module/src/scene_intersection.hpp:335:0: style: The function 'setOcclusionActivation' is never used. [unusedFunction]
  void setOcclusionActivation(const bool activation) { occlusion_activated_ = activation; }
^

planning/behavior_velocity_planner/autoware_behavior_velocity_intersection_module/src/scene_intersection.hpp:336:0: style: The function 'isOcclusionFirstStopRequired' is never used. [unusedFunction]
  bool isOcclusionFirstStopRequired() const { return occlusion_first_stop_required_; }
^

planning/behavior_velocity_planner/autoware_behavior_velocity_intersection_module/src/scene_intersection.hpp:337:0: style: The function 'getInternalDebugData' is never used. [unusedFunction]
  InternalDebugData & getInternalDebugData() const { return internal_debug_data_; }
^

planning/behavior_velocity_planner/autoware_behavior_velocity_intersection_module/src/util.cpp:324:0: style: The function 'isBeforeTargetIndex' is never used. [unusedFunction]
bool isBeforeTargetIndex(
^

planning/behavior_velocity_planner/autoware_behavior_velocity_intersection_module/src/util.cpp:348:0: style: The function 'hasAssociatedTrafficLight' is never used. [unusedFunction]
bool hasAssociatedTrafficLight(lanelet::ConstLanelet lane)
^
```
## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_velocity_planner/autoware_behavior_velocity_intersection_module/src/debug.cpp
+++ b/planning/behavior_velocity_planner/autoware_behavior_velocity_intersection_module/src/debug.cpp
@@ -159,25 +159,6 @@ visualization_msgs::msg::MarkerArray createLineMarkerArray(
   return msg;
 }
 
-[[maybe_unused]] visualization_msgs::msg::Marker createPointMarkerArray(
-  const geometry_msgs::msg::Point & point, const std::string & ns, const int64_t id, const double r,
-  const double g, const double b)
-{
-  visualization_msgs::msg::Marker marker_point{};
-  marker_point.header.frame_id = "map";
-  marker_point.ns = ns + "_point";
-  marker_point.id = id;
-  marker_point.lifetime = rclcpp::Duration::from_seconds(0.3);
-  marker_point.type = visualization_msgs::msg::Marker::SPHERE;
-  marker_point.action = visualization_msgs::msg::Marker::ADD;
-  marker_point.scale = createMarkerScale(2.0, 2.0, 2.0);
-  marker_point.color = createMarkerColor(r, g, b, 0.999);
-
-  marker_point.pose.position = point;
-
-  return marker_point;
-}
-
 constexpr std::tuple<float, float, float> white()
 {
   constexpr uint64_t code = 0xfdfdfd;
--- a/planning/behavior_velocity_planner/autoware_behavior_velocity_intersection_module/src/scene_intersection.hpp
+++ b/planning/behavior_velocity_planner/autoware_behavior_velocity_intersection_module/src/scene_intersection.hpp
@@ -327,13 +327,20 @@ public:
   visualization_msgs::msg::MarkerArray createDebugMarkerArray() override;
   autoware::motion_utils::VirtualWalls createVirtualWalls() override;
 
+  // cppcheck-suppress unusedFunction
   const std::set<lanelet::Id> & getAssociativeIds() const { return associative_ids_; }
 
+  // cppcheck-suppress unusedFunction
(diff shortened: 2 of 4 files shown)


soblin (changes requested):
cppcheck is clearly WRONG because getOcclusionDistance is actually used in the codebase. See manager.cpp L340 ~ L353 for example.

I looked through other places where "cppcheck-suppress unusedFunction" is used and found I they are detected probably because it is defined in header and used by functions defined in header.

Can you move those annotated functions to source file instead and see if cppcheck reports same warning ?

kobayu858:
I moved the definition from hpp to cpp but could not resolve the warning.

However, we have had similar problems with #8707  and #8709 , which have been resolved.

We are currently investigating other possible causes.

soblin:
@kobayu858 You mean, by moving the definition from header to source, you could eliminate "cppcheck-suprression-xxx" warning from the source code in https://github.com/autowarefoundation/autoware.universe/pull/8707

kobayu858:
@soblin 

That is correct.

However, I have tried the same method for this PR and was unable to resolve the warning.

kobayu858:
Strictly speaking, it is not the definition, but whether the target function is used in cpp.

By moving the definition to cpp, the above condition is satisfied and the warning is resolved.

kobayu858:
@soblin 

The function suppressed in the comment turned out to be a false positive due to a cppcheck parsing error.

Therefore, the comment suppression was removed.

This cppcheck parsing error will be resolved in other PRs as soon as they are ready.
Until then, the unusedFunction warning will be issued.

kobayu858:
The parsing error was that the macro in `PLUGINLIB_EXPORT_CLASS` could not be understood and the specific file could not be parsed.

soblin (approved):

Issue #8530, opened 2024-08-19
Whether there is a bug in the frame_id of the perception module output objects


cyn-liu:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I'm convinced that this is not my fault but a bug.

### Description

I set perception_mode=`lidar` and use `lidar_centerpoint` for object detection, but I find that my detection box does not show up on the lidar target.
I debug and found that the `frame_id` of  the output topic `/perception/object_recognition/detection/centerpoint/objects`  is `velodyne_top`, I think this is unreasonable, it should be `map` or `base_link`, which is configured by param `world_fram_id` in the file `centerpoint.param.yaml`.

**I have correctly configured the extrinsic parameters from `velodyne_top` to `base_link`, and NDT localization  is successful.**

### Expected behavior

The output topic `/perception/object_recognition/objects` of the perception module is drawn on the point cloud object.

### Actual behavior

The output topic `/perception/object_recognition/objects` of the perception module is not drawn on the point cloud object.

`/perception/object_recognition/detection/centerpoint/objects`
![centerpoint](https://github.com/user-attachments/assets/5259ee7d-1b06-481e-be23-548b1234b6f3)

`/perception/object_recognition/detection/clustering/objects`
![clustering](https://github.com/user-attachments/assets/2c505262-69fc-4632-b05d-1e34eb627a5d)

`/perception/object_recognition/detection/temporary_merged_objects`
![merge1](https://github.com/user-attachments/assets/b81b055f-b7a8-446a-81cc-c5bbfdfb76f8)

`/perception/object_recognition/detection/objects_before_filter`
![merge1](https://github.com/user-attachments/assets/a5c86d12-d5e0-41b0-b66b-5332daae0b74)

`/perception/object_recognition/detection/objects`
![detection](https://github.com/user-attachments/assets/b8d29cd7-0147-4d87-ad63-df7496bacd72)

`/perception/object_recognition/objects`
![recognition](https://github.com/user-attachments/assets/a09f3932-003b-4eeb-a5cb-68b2f2a95a31)

### Steps to reproduce

1. modify input/pointcloud topic
2. build Autoware
3. launch Autoware

### Versions

OS: amd
ROS2: humle
Autoware: The latest autoware.universe pulled from the autoware repository

cyn-liu:
[...]

PR #8674, opened 2024-08-29
fix(object_association_merger_node): fix the frame id of output object msg 

vividf:
## Description
This PR solves the issue https://github.com/autowarefoundation/autoware.universe/issues/8530.

If the frame id of `output_msg` is not `base_link`, the merged object will have the wrong coordinate.
This issue happens because all of the objects (0 and 1) are  transformed to base_link, but the frame of `output_msg` is from `input_objects0_msg->header`

Before fixed:
before_fixed

After fixed:
after_fixed

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

- rosbag: sample rosbag
- CenterPoint detection apply on top pointcloud 

```
```

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


cyn-liu:
Thank you very much for your fix, but I think there are some problem with this PR.

I don't think this bug should be fixed in this way. In theory, the frame_id of the two input topic of node  `object_merger` should be `base_link`, so `output_msg.header = input_objects0_msg->header` is not a bug.
 The real problem is that the frame_id of `input/object0` for node `object_merger` is `lidar frame_id` instead of `base_link`, so we should modify the frame_id of  node `lidar_centerpoint` output topic to `base_link`.

What do you think?

vividf:
@cyn-liu 
Thanks for your comment. 
However, I don't think restricting the output of the centerpoint would be a better option. Since centerpoint should be able to run without providing the baselink frame and I think the object frame should also be the same as the pointcloud frame. 

Since the code forces the header of the `output_msg` to be the same as the  `input_objects0_msg`, and the objects are transformed and merged in the base_link (which also means that the input can be in the different frame), so I think restricting the output frame id to base_link here probably is a better option. 

@technolojin @YoshiRi @yukkysaito What do you think, thanks!

technolojin (approved):
I think it was a bug, and will be fixed by this PR.

LGTM

yukkysaito (approved):
LGTM

cyn-liu:
@vividf  Thanks for your fix this bug.

By the way, I want ask a question: What is the mean of parameter `world_frame_id: map`  in package `autoware_lidar_centerpoint`? I  thought  this parameter controlled the output frame_id of node `lidar_centerpoint`, but after verification, it has no relationship with the output frame_id.

vividf:
@cyn-liu 
From the documentation https://autowarefoundation.github.io/autoware.universe/main/perception/autoware_lidar_centerpoint/
image

Issue #8589, opened 2024-08-22
After performance improvements of autoware_pointcloud_preprocessor voxel_based_compare_map_filter doesn't work


sebekx:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I'm convinced that this is not my fault but a bug.

### Description

In latest perf improvements of `autoware_pointcloud_preprocessor` [in this PR](https://github.com/autowarefoundation/autoware.universe/pull/8174) changed the way how tf are retrieved in `filter` class, which is used i.e. in  `voxel_based_compare_map_filter`. New improvements assumes that only static tf's are used, however in  `voxel_based_compare_map_filter` there is transformation from `base_link` to `map`.

### Expected behavior

`voxel_based_compare_map_filter` filter pointcloud based on poincloud map. correctly transform input pointcloud to `map` frame and filter matching points and transform back to `base_link` 

### Actual behavior

`voxel_based_compare_map_filter` after start correctly filter pointcloud but after vehicle moves old tf is used which cause filter to not match input pointcloud points with map points

### Steps to reproduce

1 Start autoware
2. Enable  /perception/object_recognition/detection/pointcloud_map_filtered/pointcloud topic in rviz
3. Observe filtered points after vehicle moves

### Possible causes

performance improvements of autoware_pointcloud_preprocessor [in this PR](https://github.com/autowarefoundation/autoware.universe/pull/8174) which retrieves tf only once which in case of `base_link` <-> `map` transform is incorrect

badai-nguyen:
@amadeuszsz Could you check this issue as it seems related to [your PR](https://github.com/autowarefoundation/autoware.universe/pull/8174)

amadeuszsz:
PR has been merged, closing.

PR #8678, opened 2024-08-29
fix(autoware_pointcloud_preprocessor): static TF listener as Filter option

amadeuszsz:
## Description
Fixes issue for filters which don't use static TFs. Filters can explicitly choose type of TF listener:
* static TFs listener (e.g. if filter will consider TFs between base_link and lidar_top),
* dynamic TFs listener (e.g. if filter will consider TFs between base_link and map).

~This choice could be exposed via config file, but in fact all the filters find their use only in specific scenarios - they will use only static TFs or only dynamic TFs.~
From discussion we moved `has_static_tf_only` as a parameter. Populating this parameter to all launch files and configs will be addressed in separate PRs.

List of filter and their TF listener type.

Dynamic:
* VoxelDistanceBasedCompareMapFilter
* VoxelBasedApproximateCompareMapFilter
* DistanceBasedCompareMapFilter
* VoxelBasedCompareMapFilter
* CompareElevationMapFilter
* VectorMapInsideAreaFilter

Static:
* RANSACGroundFilter
* ScanGroundFilter
* RayGroundFilter
* BlockageDiag
* PointcloudAccumulator
* PassThroughFilterUInt16
* PassThroughFilter
* CropBoxFilter
* PolygonRemover
* VoxelGridDownsampleFilter
* RandomDownsampleFilter
* ApproximateDownsampleFilter
* PickupBasedVoxelGridDownsampleFilterComponent
* VoxelGridOutlierFilter
* RadiusSearch2DOutlierFilter
* DualReturnOutlierFilter
* RingOutlierFilter  

## Related links

**Parent Issue:**

- https://github.com/autowarefoundation/autoware.universe/issues/8589

## How was this PR tested?

Logging simulator with example rosbag.
Screenshot from 2024-08-29 21-01-32

Lower CPU usage is still observed.
result_compared

## Notes for reviewers

One file has to be renamed. For that reason commit with file rename has been isolated - better changes tracking.

## Interface changes

Filter class constructor from `autoware_pointcloud_preprocessor` consider new argument `has_static_tf_only`, by default as false. If set to true, the filter will use custom class for handling TFs presented in https://github.com/autowarefoundation/autoware.universe/pull/8174. Otherwise, the class will use classic approach, thus listener will remain in nodes graph. This feature changes name convention for`StaticTransformBuffer` class, therefore class was renamed to `ManagedTransformBuffer`.

## Effects on system behavior

None.


sensing/autoware_pointcloud_preprocessor/src/filter.cpp
```diff
@@ -65,7 +65,8 @@
 
 //////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 autoware::pointcloud_preprocessor::Filter::Filter(
-  const std::string & filter_name, const rclcpp::NodeOptions & options)
+  const std::string & filter_name, const rclcpp::NodeOptions & options,
+  const bool & has_static_tf_only)
```
badai-nguyen, line 69: @amadeuszsz Thank you for your fixing. 
- Since the [input and output frame used in filter.cpp are declared as parameters](https://github.com/autowarefoundation/autoware.universe/blob/7f36c52fdedbdc91e598a6c567b21e65d0650888/sensing/pointcloud_preprocessor/src/filter.cpp#L71-L72), IMO `has_static_tf_only` should be declared as parameter and assign value denpend on the `input_frame` and `output_frame` values,
- For example, if the parameters of ground_segmentation (https://github.com/autowarefoundation/autoware.universe/blob/0e1761ae4ba000beec9b23b76a744a407c0934d9/launch/tier4_perception_launch/launch/obstacle_segmentation/ground_segmentation/ground_segmentation.launch.py#L268-L269) set to `map` -> `ground_segmentation_node` should be intialized and treated as `has_static_tf_only = false`

badai-nguyen (changes requested):

sensing/autoware_pointcloud_preprocessor/src/filter.cpp
amadeuszsz: @badai-nguyen thank you for review!
If at least one filter can find use with & without static TFs, your concerns are correct and we should expose `has_static_tf_only` as parameter. I will do relevant changes 👍🏻

amadeuszsz: @badai-nguyen 
Adressed in https://github.com/autowarefoundation/autoware.universe/pull/8678/commits/f85f3f2afadeb1cccfaa1b2a16cb1de8df11156c
Currently autoware_pointcloud_preprocessor nodes are widely use among standalone Autoware packages and adding `has_static_tf_only` parameter without default value (false) could cause them to stop working. I suggest to address usage of this performance improvement (static TF buffers) in separate PRs to make this transition safer. By default we will use classic TF listener, however if node inherit from Filter class and have same in & out frame, [we still use improvement version](https://github.com/autowarefoundation/autoware.universe/blob/f85f3f2afadeb1cccfaa1b2a16cb1de8df11156c/sensing/autoware_pointcloud_preprocessor/src/filter.cpp#L119-L125).

common/autoware_universe_utils/include/autoware/universe_utils/ros/managed_transform_buffer.hpp
```diff
@@ -124,42 +189,38 @@ class StaticTransformBuffer
...
     }
-    pcl_ros::transformPointCloud(eigen_transform, cloud_in, cloud_out);
-    cloud_out.header.frame_id = target_frame;
+    pcl_ros::transformAsMatrix(*tf, eigen_transform);
     return true;
   }
 
-private:
   TFMap buffer_;
+  rclcpp::Node * node_;
```
knzo25, line 219: can you const this or use a reference?

common/autoware_universe_utils/include/autoware/universe_utils/ros/managed_transform_buffer.hpp
```diff
@@ -58,12 +58,69 @@ struct PairEqual
   }
 };
 using TFMap = std::unordered_map<Key, Eigen::Matrix4f, std::hash<Key>, PairEqual>;
+constexpr std::array<const char *, 3> warn_frames = {"map", "odom", "world"};
```
knzo25, line 61: nit: this is a personal trauma, but would it be possible to use `std::string_view` for the constexpr?

common/autoware_universe_utils/include/autoware/universe_utils/ros/managed_transform_buffer.hpp
```diff
@@ -58,12 +58,69 @@ struct PairEqual
...
+        return getDynamicTransform(target_frame, source_frame, eigen_transform);
+      };
+    }
+  }
+
+  bool getTransform(
+    const std::string & target_frame, const std::string & source_frame,
+    Eigen::Matrix4f & eigen_transform)
+  {
+    return get_transform_(target_frame, source_frame, eigen_transform);
```
knzo25, line 89: nit: this adds an extra level of indirection. the extra call and stack could be avoided by having the branch here (just ignore this if you do not agree)

common/autoware_universe_utils/include/autoware/universe_utils/ros/managed_transform_buffer.hpp
```diff
@@ -124,42 +189,38 @@ class StaticTransformBuffer
...
-   * Transforms a point cloud from one frame to another.
+  /** @brief Retrieves a transform between two dynamic frames.
    *
-   * @param node A pointer to the ROS node.
-   * @param target_frame The target frame to transform the point cloud to.
-   * @param cloud_in The input point cloud to be transformed.
-   * @param cloud_out The transformed point cloud.
-   * @return True if the transformation is successful, false otherwise.
+   * This function attempts to retrieve a transform between the target frame and the source frame.
+   * If success, the transform matrix is set to the output parameter and the function returns true.
```
knzo25, line 195:
```suggestion
   * If succeds, the transform matrix is set to the output parameter and the function returns true.
```

knzo25, line 195:
```suggestion
   * If successful, the transform matrix is assigned to the output parameter, and the function returns true.
```

common/autoware_universe_utils/include/autoware/universe_utils/ros/managed_transform_buffer.hpp
```diff
@@ -72,17 +129,24 @@ class StaticTransformBuffer
    * Otherwise, transform matrix is set to identity and the function returns false. Transform
    * Listener is destroyed after function call.
    *
-   * @param node A pointer to the ROS node.
    * @param target_frame The target frame.
    * @param source_frame The source frame.
    * @param eigen_transform The output Eigen transform matrix. Is set to identity if the transform
```
knzo25, line 134:
```suggestion
   * @param eigen_transform The output Eigen transform matrix. It is set to the identity if the transform
```

common/autoware_universe_utils/include/autoware/universe_utils/ros/managed_transform_buffer.hpp
```diff
@@ -124,42 +189,38 @@ class StaticTransformBuffer
...
+  /** @brief Retrieves a transform between two dynamic frames.
    *
-   * @param node A pointer to the ROS node.
-   * @param target_frame The target frame to transform the point cloud to.
-   * @param cloud_in The input point cloud to be transformed.
-   * @param cloud_out The transformed point cloud.
-   * @return True if the transformation is successful, false otherwise.
+   * This function attempts to retrieve a transform between the target frame and the source frame.
+   * If success, the transform matrix is set to the output parameter and the function returns true.
+   * Otherwise, transform matrix is set to identity and the function returns false.
```
knzo25, line 196:
```suggestion
   * Otherwise, the transform matrix is set to the identity and the function returns false.
```

common/autoware_universe_utils/include/autoware/universe_utils/ros/managed_transform_buffer.hpp
```diff
@@ -124,42 +189,38 @@ class StaticTransformBuffer
...
+   *
+   * @param target_frame The target frame.
+   * @param source_frame The source frame.
+   * @param eigen_transform The output Eigen transform matrix. Is set to identity if the transform
```
knzo25, line 200:
```suggestion
   * @param eigen_transform The output Eigen transformation matrix. It is set to the identity if the transformation
```

sensing/autoware_pointcloud_preprocessor/src/distortion_corrector/distortion_corrector_node.cpp
```diff
@@ -35,6 +35,8 @@ DistortionCorrectorComponent::DistortionCorrectorComponent(const rclcpp::NodeOpt
   base_frame_ = declare_parameter<std::string>("base_frame");
   use_imu_ = declare_parameter<bool>("use_imu");
   use_3d_distortion_correction_ = declare_parameter<bool>("use_3d_distortion_correction");
+  auto has_static_tf_only =
+    declare_parameter<bool>("has_static_tf_only", false);  // TODO(amadeuszsz): remove default value
```
knzo25, line 39: Only this occurrence has a comment. Is this on purpose or should it be added to all other instances?

amadeuszsz: Other occurrences are placed in nodes which are not refactored yet, therefore default value for parameter is valid.

common/autoware_universe_utils/include/autoware/universe_utils/ros/managed_transform_buffer.hpp
amadeuszsz: Addressed in https://github.com/autowarefoundation/autoware.universe/pull/8678/commits/d905dfe990ddd1186289d5f541a9d7eacafa666f

sensing/autoware_pointcloud_preprocessor/src/distortion_corrector/distortion_corrector_node.cpp
knzo25: Thanks !

sensing/autoware_pointcloud_preprocessor/src/filter.cpp
badai-nguyen: @amadeuszsz I'm sorry for late response.
- I think it is good for now to keep has_static_tf_only with default(false) value. 
- My [concern ](https://github.com/autowarefoundation/autoware.universe/pull/8678#discussion_r1740114981) here is just a posibility of base_link <-> map transformation when parameters changed. 
- NIT: Current default setting, `has_static_tf_only = true` is also acceptable for `scan_ground_segmentation`.

badai-nguyen (approved):

badai-nguyen:
@amadeuszsz 
Q: the cpu% evaluation is for latest version with `has_static_tf_only: = false` as default?

amadeuszsz:
> @amadeuszsz Q: the cpu% evaluation is for latest version with `has_static_tf_only: = false` as default?

@badai-nguyen 
Evaluation was performed with `sample_sensor_kit` lidar pipeline and after this PR modification it will be equivalent if `has_static_tf_only = true`. Result on attached graph show performance comparison with default implementation and improved version (number of listeners 17 -> 0). 

Now, with `has_static_tf_only = false` by default, number of listeners changes 17 -> 9 (input frame ≠ output frame e.g. for `base_link -> imu_link`), therefore improvement will be less significant. We can assume this PR as partial revert.

The next step is to populate `has_static_tf_only` parameter to all config files and launch files for all the autoware.universe packages and Autoware-related repositories. This change will require more time for testing. In addition, since `autoware_pointcloud_preprocessor` filters are during refactor, I would like to check for possible Filter class design changes before adding new parameter.

badai-nguyen:
> > @amadeuszsz Q: the cpu% evaluation is for latest version with `has_static_tf_only: = false` as default?
> 
> @badai-nguyen Evaluation was performed with `sample_sensor_kit` lidar pipeline and after this PR modification it will be equivalent if `has_static_tf_only = true`. Result on attached graph show performance comparison with default implementation and improved version (number of listeners 17 -> 0).
> 
> Now, with `has_static_tf_only = false` by default, number of listeners changes 17 -> 9 (input frame ≠ output frame e.g. for `base_link -> imu_link`), therefore improvement will be less significant. We can assume this PR as partial revert.
> 
> The next step is to populate `has_static_tf_only` parameter to all config files and launch files for all the autoware.universe packages and Autoware-related repositories. This change will require more time for testing. In addition, since `autoware_pointcloud_preprocessor` filters are during refactor, I would like to check for possible Filter class design changes before adding new parameter.

@amadeuszsz thank you for your clarificaiton. It is clear for me know.

soblin (approved):

mitsudome-r:
CI is failing for unknown reason, but since the package does not modify CUDA related packages and the CI is passing for no-cuda workflow, I am merging the PR.

NOTE:
[Failed log
](https://github.com/autowarefoundation/autoware.universe/actions/runs/10736481342/job/29848147519?pr=8678)[Failed CodeBuild in AWS](https://us-east-1.console.aws.amazon.com/codesuite/codebuild/305110990523/projects/autoware-us-east-1/build/autoware-us-east-1%3Aa8eaa44c-090c-412e-b52f-9240d258733f/utilization?region=us-east-1)

PR #8680, opened 2024-08-30
refactor(ekf_localizer): add namespace "autoware::ekf_localizer::"

SakodaShintaro:
## Description

Add namespace `autoware::ekf_localizer::` to `ekf_localizer` package.

## How was this PR tested?

I have confirmed autoware with [autoware_tools](https://github.com/autowarefoundation/autoware_tools) and [localization_evaluation_tools (TIER IV INTERNAL)](https://github.com/tier4/localization_evaluator_tools) can be built.

Also, the mean error in logging_simulator about AWSIM data is almost the same as before. (7.0cm -> 7.3cm)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/localization/ekf_localizer/CMakeLists.txt
+++ b/localization/ekf_localizer/CMakeLists.txt
@@ -25,7 +25,7 @@ ament_auto_add_library(${PROJECT_NAME} SHARED
 )
 
 rclcpp_components_register_node(${PROJECT_NAME}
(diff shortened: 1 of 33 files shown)


TaikiYamada4 (approved):
`colcon build` worked with no errors, and logging simulator in my environment works well too.

LGTM :+1:

PR #8685, opened 2024-08-30
fix(simple_planning_simulator): delete velocity dead band for brake

yuki-takagi-66:
## Description
fix stopping behavior issue
In the previous code, when the vehicle velocity is smaller than dead_band velocity, the vehicle do not stop with enough brake.
Currently, stop processing is done within update().

## How was this PR tested?
scenario sim tell me this problem, and I tested in psim

## Interface changes

None.

<!-- ⬇️🔴

## Effects on system behavior

None.

### DIFF
--- a/simulator/simple_planning_simulator/src/simple_planning_simulator/vehicle_model/sim_model_delay_steer_acc_geared_wo_fall_guard.cpp
+++ b/simulator/simple_planning_simulator/src/simple_planning_simulator/vehicle_model/sim_model_delay_steer_acc_geared_wo_fall_guard.cpp
@@ -166,13 +166,10 @@ Eigen::VectorXd SimModelDelaySteerAccGearedWoFallGuard::calcModel(
         return pedal_acc + input(IDX_U::SLOPE_ACCX);
       }
     } else {
-      constexpr double brake_dead_band = 1e-3;
-      if (vel > brake_dead_band) {
+      if (vel > 0.0) {
         return pedal_acc + input(IDX_U::SLOPE_ACCX);
-      } else if (vel < -brake_dead_band) {
+      } else if (vel < 0.0) {
         return -pedal_acc + input(IDX_U::SLOPE_ACCX);
-      } else if (-pedal_acc >= std::abs(input(IDX_U::SLOPE_ACCX))) {
-        return 0.0;
       } else {
         return input(IDX_U::SLOPE_ACCX);
       }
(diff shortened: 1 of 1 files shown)


simulator/simple_planning_simulator/src/simple_planning_simulator/vehicle_model/sim_model_delay_steer_acc_geared_wo_fall_guard.cpp
```diff
@@ -166,13 +166,10 @@ Eigen::VectorXd SimModelDelaySteerAccGearedWoFallGuard::calcModel(
...
+      if (vel > 0.0) {
         return pedal_acc + input(IDX_U::SLOPE_ACCX);
-      } else if (vel < -brake_dead_band) {
+      } else if (vel < 0.0) {
         return -pedal_acc + input(IDX_U::SLOPE_ACCX);
-      } else if (-pedal_acc >= std::abs(input(IDX_U::SLOPE_ACCX))) {
-        return 0.0;
       } else {
         return input(IDX_U::SLOPE_ACCX);
       }
```
maxime-clem, line 175: If we are stopped (`vel == 0`) and apply the brake (`pedal_acc < 0`) we will return `IDX_U::SLOPE_ACCX` which seems like a mistake :thinking: 
I think the case you removed `-pedal_acc >= std::abs(input(IDX_U::SLOPE_ACCX))` should be kept to return `0` when the brake is stronger than the slope acceleration.

yuki-takagi-66: Exactly! Thank you

yuki-takagi-66: The mentioned behavior is also achieved by the following block in update().
```
if (
    prev_state(IDX::VX) * state_(IDX::VX) <= 0.0 &&
    -state_(IDX::PEDAL_ACCX) >= std::abs(delayed_input(IDX_U::SLOPE_ACCX))) {}
```
However, I think calcModel() should maintain the physical consistency. Therefore I decide to maintain the block 
`if (-pedal_acc >= std::abs(input(IDX_U::SLOPE_ACCX)))`

maxime-clem (approved):

simulator/simple_planning_simulator/src/simple_planning_simulator/vehicle_model/sim_model_delay_steer_acc_geared_wo_fall_guard.cpp
yuki-takagi-66: @maxime-clem 
I've fixed. Thank you so much

Issue #8695, opened 2024-08-30
`common` packages are not functioning as a true "common"


youtalk:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I've agreed with the maintainers that I can plan this task.

### Description

I am working on multi-containerizing Autoware, but I'm having trouble because [`autoware.universe/common`](https://github.com/autowarefoundation/autoware.universe/tree/main/common) is not functioning as a true "common" directory as following. 

1. The packages under `common` depend on packages outside of `common`.
  - For example, `common` depends on `vehicle/autoware_vehicle_info_utils`.  
  - For example, `common` depends on `simulator/dummy_perception_publisher`.  
2. Some packages under `common` are only used in specific component directories.  
  - For example, `cuda_utils` is only used under `perception`. If CUDA dependencies can be removed from `common`, CUDA dependencies can be also removed from containers other than `perception`, leading to a significant reduction in image size.  
  - For example, `common/perception_utils` is only used in packages under `perception`.

While issue 2 is not a problem for now, issue 1 is a concern. When breaking down `autoware-universe` in the second step of the multi-containerization plan mentioned in
https://github.com/orgs/autowarefoundation/discussions/4661#discussioncomment-9995806, the idea is to create a shared `autoware-universe-common` stage to avoid redundant builds of common packages. 

![346866514-690b88ac-ee85-490a-93d8-14f8626c3e2e](https://github.com/user-attachments/assets/74da1ee9-e7f1-4981-a1e6-7faebd30393e)

However, under the current situation described in issue 1, this approach cannot be realized, and all container images will need to rebuild the packages under common repeatedly.

Ref. https://github.com/autowarefoundation/autoware/issues/5077

### Purpose

Ensure that the packages under `common` do not have mutual dependencies with packages outside of `common`.

### Possible approaches

Ensure that the packages under `common` do not have mutual dependencies with packages outside of `common`.

### Definition of done

Exclude dependencies on the following package list from the `common` packages to resolve issue 1.

- [x] autoware_vehicle_info_utils
- [x] dummy_perception_publisher

youtalk:
- [ ] tensorrt_common

PR #8708, opened 2024-09-02
chore(autoware_vehicle_info_utils): move package to common directory

mitsudome-r:
## Description

This moves autoware_vehicle_info_utils into common directory since the package is used by the packages in different modules (planning, control, system, common, etc)

## Related links

**Parent Issue:**

- https://github.com/autowarefoundation/autoware.universe/issues/8695

## Notes for reviewers

This PR does not change anything in the code and shouldn't have any affect on Autoware's behavior.

## Interface changes

None.

## Effects on system behavior

None.


tkimura4:
@mitsudome-r 
I think the following files might also need to be changed.

https://github.com/autowarefoundation/autoware.universe/blob/main/.github/CODEOWNERS

taikitanaka3:
@mitsudome-r 
I think git log was deleted by this PR.
so I think it's better to use git filter repo instead

mitsudome-r:
@tkimura4 Thanks. I think the codeowner should be updated automatically by this workflow: https://github.com/autowarefoundation/autoware.universe/blob/main/.github/workflows/update-codeowners-from-packages.yaml

mitsudome-r:
@taikitanaka3 
> I think git log was deleted by this PR.
> so I think it's better to use git filter repo instead

Thanks for the suggestion. I have never used it before. I can take a look, but is the tool able to retain the git history without making a force-push to the main branch? I rather not do disturbing changes to the commit history of the main branch.

Also FYI, you should still able to track the history by adding `--follow` option. (e.g., `git log --follow common/autoware_vehicle_info_utils/src/vehicle_info_utils.cpp`) You can also check through GitHub UI by clicking "Browse History" in [commit history](https://github.com/mitsudome-r/autoware.universe/commits/mv-vehicle-info-utils/common/autoware_vehicle_info_utils/src/vehicle_info.cpp) page, although this only works with a file and not with a directory.
image

taikitanaka3 (approved):
@mitsudome-r 
lgtm for me

tkimura4 (approved):

youtalk:
@mitsudome-r Could you merge this?

PR #8707, opened 2024-09-02
fix(autoware_behavior_path_planner_common): fix unusedFunction

kobayu858:
## Description
This is a fix based on cppcheck unusedFunction warnings.

I suppressed the cppcheck warning in a previous PR (#8654) with a comment.
However, by moving the part using the target function from hpp to cpp, the warning could be eliminated without comment suppression.
Also, setDecelerationVelocity was removed because it was not used.
```
planning/behavior_path_planner/autoware_behavior_path_planner_common/src/marker_utils/utils.cpp:344:0: style: The function 'createDrivableLanesMarkerArray' is never used. [unusedFunction]
MarkerArray createDrivableLanesMarkerArray(
^

planning/behavior_path_planner/autoware_behavior_path_planner_common/src/utils/utils.cpp:1167:0: style: The function 'setDecelerationVelocity' is never used. [unusedFunction]
PathWithLaneId setDecelerationVelocity(
^

planning/behavior_path_planner/autoware_behavior_path_planner_common/src/utils/utils.cpp:1638:0: style: The function 'convertToSnakeCase' is never used. [unusedFunction]
std::string convertToSnakeCase(const std::string & input_str)
^
```

## Related links

- #8654 

**Parent Issue:**

- #8709 

## How was this PR tested?
`ros2 launch autoware_launch planning_simulator.launch.xml map_path:=$HOME/autoware_map/sample-map-planning vehicle_model:=sample_vehicle sensor_model:=sample_sensor_kit`

Confirmation that it works.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

PR #8659, opened 2024-08-28
fix(autoware_behavior_path_start_planner_module): fix unusedFunction

kobayu858:
## Description
This is a fix based on cppcheck unusedFunction warnings.
Debug.cpp was not used, so the entire file was deleted.

```
planning/behavior_path_planner/autoware_behavior_path_start_planner_module/src/debug.cpp:20:0: style: The function 'updateSafetyCheckDebugData' is never used. [unusedFunction]
void updateSafetyCheckDebugData(
^

planning/behavior_path_planner/autoware_behavior_path_start_planner_module/src/util.cpp:75:0: style: The function 'getBackedPose' is never used. [unusedFunction]
Pose getBackedPose(
^

planning/behavior_path_planner/autoware_behavior_path_start_planner_module/src/util.cpp:106:0: style: The function 'extractCollisionCheckSection' is never used. [unusedFunction]
std::optional<PathWithLaneId> extractCollisionCheckSection(
^
```
## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

PR #8654, opened 2024-08-28
fix(autoware_behavior_path_planner_common): fix unusedFunction

kobayu858:
## Description
This is a fix based on cppcheck unusedFunction warnings.

```
planning/behavior_path_planner/autoware_behavior_path_planner_common/src/marker_utils/utils.cpp:344:0: style: The function 'createDrivableLanesMarkerArray' is never used. [unusedFunction]
MarkerArray createDrivableLanesMarkerArray(
^

planning/behavior_path_planner/autoware_behavior_path_planner_common/src/utils/drivable_area_expansion/static_drivable_area.cpp:625:0: style: The function 'calcCenterOfGeometry' is never used. [unusedFunction]
[[maybe_unused]] geometry_msgs::msg::Point calcCenterOfGeometry(const Polygon2d & obj_poly)
^

planning/behavior_path_planner/autoware_behavior_path_planner_common/src/utils/parking_departure/utils.cpp:107:0: style: The function 'updateSafetyCheckTargetObjectsData' is never used. [unusedFunction]
void updateSafetyCheckTargetObjectsData(
^

planning/behavior_path_planner/autoware_behavior_path_planner_common/src/utils/path_utils.cpp:212:0: style: The function 'getPathTurnSignal' is never used. [unusedFunction]
std::pair<TurnIndicatorsCommand, double> getPathTurnSignal(
^

planning/behavior_path_planner/autoware_behavior_path_planner_common/src/utils/path_utils.cpp:434:0: style: The function 'isCloseToPath' is never used. [unusedFunction]
bool isCloseToPath(const PathWithLaneId & path, const Pose & pose, const double distance_threshold)
^

planning/behavior_path_planner/autoware_behavior_path_planner_common/src/utils/path_utils.cpp:582:0: style: The function 'getFirstStopPoseFromPath' is never used. [unusedFunction]
std::optional<Pose> getFirstStopPoseFromPath(const PathWithLaneId & path)
^

planning/behavior_path_planner/autoware_behavior_path_planner_common/src/utils/utils.cpp:102:0: style: The function 'getDistanceBetweenPredictedPaths' is never used. [unusedFunction]
double getDistanceBetweenPredictedPaths(
^

planning/behavior_path_planner/autoware_behavior_path_planner_common/src/utils/utils.cpp:124:0: style: The function 'getDistanceBetweenPredictedPathAndObject' is never used. [unusedFunction]
double getDistanceBetweenPredictedPathAndObject(
^

[...]

PR #8709, opened 2024-09-02
fix(autoware_behavior_path_start_planner_module): fix unusedFunction

kobayu858:
## Description

I suppressed the cppcheck warning in a previous PR #8654 and #8659 with a comment.
However, by moving the part using the target function from hpp to cpp, the warning could be eliminated without comment suppression.
## Related links

- #8654 
- #8659 

**Parent Issue:**

- #8707 

## How was this PR tested?
`ros2 launch autoware_launch planning_simulator.launch.xml map_path:=$HOME/autoware_map/sample-map-planning vehicle_model:=sample_vehicle sensor_model:=sample_sensor_kit`

Confirmation that it works.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


kobayu858:
`checkCollisionBetweenPathFootprintsAndObjects`

Deletion of comment suppression for this function is done at #8707

soblin (approved):

PR #8710, opened 2024-09-02
fix(mission_planner): improve condition to check if the goal is within the lane

maxime-clem:
## Description

This PR fixes an issue occurring when path lanelets overlap, causing `combine_lanelets_with_shoulder()` to generate a self-intersecting polygon which can result in goal poses being wrongfully rejected.

In this PR, we modify the logic to not use the `combine_lanelets_with_shoulder()` function.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

- [TIER IV interval link](https://evaluation.tier4.jp/evaluation/reports/400707a9-0b3c-591d-8715-7080510143e2?project_id=prd_jt)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


soblin (approved):

soblin:
For my understanding:

C := A - B would be empty if A ⊂ B, because entire A overlaps with B.

PR #8714, opened 2024-09-02
feat(goal_planner): extend pull over lanes inward to extract objects

kosuke55:
## Description

The current algorithm expands the pull over lane only outward and extracts objects in it. In this case, if an object exists without overlapping the narrow shoulder lane, the object is not extracted. For example, since the object does not overlap the shoulder lane at 8509, it is not included in the collision check. As a result, the route is generated ignoring the object, and it gets stuck.

image

image

In this PR, pull over lanes are expanded inward so that objects can be extracted regardless of the width of the lane. The length to be expanded is determined by the goal search area. the extraction plygon is the sky blue one.

image

image

image

## Related links
psim

https://evaluation.tier4.jp/evaluation/reports/41bd57af-9931-5c32-ae12-bbb927ad0dd8?project_id=prd_jt&state=failed
4133/4136

baseline: 4132/4136([2024/09/03](https://evaluation.tier4.jp/evaluation/reports/f91ea6c8-5403-5392-83dd-7adb9961eacc/?project_id=prd_jt))

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_path_planner/autoware_behavior_path_goal_planner_module/src/util.cpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_goal_planner_module/src/util.cpp
@@ -191,7 +191,7 @@ std::optional<Polygon2d> generateObjectExtractionPolygon(
     constexpr double INTERSECTION_CHECK_DISTANCE = 10.0;
     std::vector<Point> modified_bound{};
     size_t i = 0;
-    while (i < bound.size() - 1) {
+    while (i < bound.size() - 2) {
       BoostPoint p1(bound.at(i).x, bound.at(i).y);
       BoostPoint p2(bound.at(i + 1).x, bound.at(i + 1).y);
       LineString p_line{};
--- a/planning/behavior_path_planner/autoware_behavior_path_goal_planner_module/src/util.cpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_goal_planner_module/src/util.cpp
@@ -39,6 +39,7 @@
 namespace autoware::behavior_path_planner::goal_planner_utils
 {
 
+using autoware::universe_utils::calcDistance2d;
 using autoware::universe_utils::calcOffsetPose;
 using autoware::universe_utils::createDefaultMarker;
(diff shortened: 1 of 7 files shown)


planning/behavior_path_planner/autoware_behavior_path_goal_planner_module/src/util.cpp
```diff
@@ -138,6 +139,122 @@ lanelet::ConstLanelets generateBetweenEgoAndExpandedPullOverLanes(
...
+    return Polygon2d{};
+  }
+
+  // set orientation to next point
+  for (size_t i = 0; i < base_boundary_poses.size() - 1; ++i) {
+    const auto & p = base_boundary_poses[i].position;
+    const auto & next_p = base_boundary_poses[i + 1].position;
+    const double yaw = autoware::universe_utils::calcAzimuthAngle(p, next_p);
+    base_boundary_poses[i].orientation = autoware::universe_utils::createQuaternionFromYaw(yaw);
+  }
```
danielsanchezaran, line 171:
```suggestion
for (auto it = base_boundary_poses.begin(); it != std::prev(base_boundary_poses.end()); ++it) {
    const auto &p = it->position;
    const auto &next_p = std::next(it)->position;
    const double yaw = autoware::universe_utils::calcAzimuthAngle(p, next_p);
    it->orientation = autoware::universe_utils::createQuaternionFromYaw(yaw);
}
```
Maybe this is more "modern" c++?

kosuke55: thanks!

soblin (changes requested):

kosuke55:
about: `remove_self_intersection`

1)
Shifting the bound of a lanelet inward can cause self-intersections line gray line.

image


2)
To prevent this, add the self-intersected point to the bound instead. After the modification, the blue line is shown below.

image

soblin (approved):

PR #8719, opened 2024-09-03
feat(surround_obstacle_checker): integrate generate_parameter_library

mitukou1109:
## Description

This PR enables surround_obstacle_checker to handle parameters using [generate_parameter_library](https://github.com/PickNikRobotics/generate_parameter_library).

## How was this PR tested?

Psim

- [x] changes in rqt reconfigure are reflected to the behavior of the node
- [x] errors occur if any parameters are missing in the config file in autoware_launch

[Screencast from 2024年09月03日 11時49分01秒.webm](https://github.com/user-attachments/assets/bd2bf69a-8f0d-4d59-91ed-ea81125b89a7)

Evaluator: https://evaluation.tier4.jp/evaluation/reports/36978351-88c5-59df-967f-03ff39440cae?project_id=prd_jt

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


planning/autoware_surround_obstacle_checker/param/surround_obstacle_checker_node_parameters.yaml
```diff
@@ -0,0 +1,67 @@
+surround_obstacle_checker_node:
```
satoshi-ota, line 1: @mitukou1109 Is it possible to create this yaml file under `config` dir...?

planning/autoware_surround_obstacle_checker/src/node.hpp
```diff
@@ -62,27 +63,7 @@ class SurroundObstacleCheckerNode : public rclcpp::Node
...
-    double pointcloud_surround_check_side_distance;
-    double pointcloud_surround_check_back_distance;
-    double surround_check_hysteresis_distance;
-    double state_clear_time;
-    bool publish_debug_footprints;
-    std::string debug_footprint_label;
-  };
-
 private:
-  rcl_interfaces::msg::SetParametersResult onParam(
```
satoshi-ota, line 83: Where is the alternative function of `onParam`...? Or, the pickNik lib can automatically update the value?

planning/autoware_surround_obstacle_checker/param/surround_obstacle_checker_node_parameters.yaml
```diff
@@ -0,0 +1,67 @@
...
+  pointcloud:
+    enable_check:
+      type: bool
+      default_value: false
```
satoshi-ota, line 5: @mitukou1109 please remove default value.

planning/autoware_surround_obstacle_checker/src/node.hpp
mitukou1109: `ParamListener` adds a callback for each parameter internally and thus we don't have to care about updating its value, but we still need to call `get_params()` to acquire the latest values as done [here](https://github.com/mitukou1109/autoware.universe/blob/ef0b5a4835f0a49ff9cdf84971d3ad820a14dc29/planning/autoware_surround_obstacle_checker/src/node.cpp#L198).

planning/autoware_surround_obstacle_checker/param/surround_obstacle_checker_node_parameters.yaml
mitukou1109: Isn't it confusing if it is in the same directory as the original parameter file? The file is just for code generation.

planning/autoware_surround_obstacle_checker/param/surround_obstacle_checker_node_parameters.yaml
mitukou1109: Fixed: https://github.com/autowarefoundation/autoware.universe/pull/8719/commits/fb362a9a8890a78aa6dcf89c9ff8587d9a05d69d

planning/autoware_surround_obstacle_checker/param/surround_obstacle_checker_node_parameters.yaml
satoshi-ota: :thinking: OK. Let's use param dir for now.

planning/autoware_surround_obstacle_checker/param/surround_obstacle_checker_node_parameters.yaml
```diff
@@ -0,0 +1,57 @@
+surround_obstacle_checker_node:
+  pointcloud:
+    enable_check:
+      type: bool
+      description: Enable to check surrounding obstacle
+  object_types:
+    type: string_array
+    default_value: [unknown, car, truck, bus, trailer, motorcycle, bicycle, pedestrian]
```
satoshi-ota, line 8: @mitukou1109 Is it possible to remove default values...?

And, I think you can add validation like this.
```yaml
    validation:
      subset_of<>: [[unknown, car, truck, bus, trailer, motorcycle, bicycle, pedestrian]]
```

https://github.com/PickNikRobotics/generate_parameter_library?tab=readme-ov-file#built-in-validators

planning/autoware_surround_obstacle_checker/param/surround_obstacle_checker_node_parameters.yaml
```diff
@@ -0,0 +1,57 @@
...
+  __map_object_types:
+    enable_check:
+      type: bool
+      description: Enable to check surrounding obstacle
+  obstacle_types:
+    type: string_array
+    default_value: [pointcloud, unknown, car, truck, bus, trailer, motorcycle, bicycle, pedestrian]
```
satoshi-ota, line 15: Same for above.

planning/autoware_surround_obstacle_checker/param/surround_obstacle_checker_node_parameters.yaml
satoshi-ota: You can add `unique<>` as well.

mitukou1109: `object_types` and `obstacle_types` are for generating parameter maps. Parameters with such names do not exist.

https://github.com/PickNikRobotics/generate_parameter_library?tab=readme-ov-file#mapped-parameters

satoshi-ota: Ah. I see.

satoshi-ota (approved):
LGTM. Good job.

PR #8723, opened 2024-09-03
fix(autoware_universe_utils): fix unusedFunction

kobayu858:
## Description

```
common/autoware_universe_utils/src/system/backtrace.cpp:29:0: style: The function 'print_backtrace' is never used. [unusedFunction]
void print_backtrace()
^
```

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/common/autoware_universe_utils/src/system/backtrace.cpp
+++ b/common/autoware_universe_utils/src/system/backtrace.cpp
@@ -26,6 +26,7 @@
 namespace autoware::universe_utils
 {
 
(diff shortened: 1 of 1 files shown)


kobayu858:
The warning occurs because only hpp template functions are used.

This can be solved by moving it from hpp to cpp, but in this case it is a template function, so it was suppressed with a comment.

takayuki5168 (approved):

PR #8726, opened 2024-09-03
fix(static_obstacle_avoidance): improve turn signal output timing when the ego returns original lane

satoshi-ota:
## Description

Fix bug which caused delay of turning on signal when the ego returns original lane after avoiding. This module outputs turn signal per shift lines but sometimes they include the one whose shift length is small. In original implementation, this module didn't output turn signal for such a shift line. Additionally, since this module only checked front (=nearest) shift line, it may ignore return shift line unexpectedly.

In this PR, I fixed the logic to remove small shift line at first.

```c++
  const auto itr =
    std::remove_if(shift_lines.begin(), shift_lines.end(), [&, this](const auto & s) {
      const auto threshold = planner_data_->parameters.turn_signal_shift_length_threshold;
      return std::abs(s.start_shift_length - s.end_shift_length) < threshold ||
             is_ignore_signal(s.id);
    });
  shift_lines.erase(itr, shift_lines.end());
```

And, I added the logic to check not only nearest shift line but also next one, and outputs turn signal for latter one if its start point is closer than threshold.

```c++
    for (size_t i = 1; i < shift_lines.size(); i++) {
      const auto & s2 = shift_lines.at(i);

      const auto s1_relative_length = s1.start_shift_length - s1.end_shift_length;
      const auto s2_relative_length = s2.start_shift_length - s2.end_shift_length;

      // same side shift
      if (s1_relative_length > 0.0 && s2_relative_length > 0.0) {
        continue;
      }

      // same side shift
      if (s1_relative_length < 0.0 && s2_relative_length < 0.0) {
        continue;
      }

      // different side shift
      const auto & points = path_shifter_.getReferencePath().points;
      const size_t idx = planner_data_->findEgoIndex(points);

      // output turn signal for near shift line.
      if (calcSignedArcLength(points, idx, s1.start_idx) > 0.0) {
        return s1;
      }

      // output turn signal for far shift line.
      if (
        calcSignedArcLength(points, idx, s2.start_idx) <
        getEgoSpeed() * parameters_->max_prepare_time) {
        return s2;
      }
```

## Related links

**Parent Issue:**

- [TIER IV INTERNAL LINK](https://tier4.atlassian.net/browse/ISZ-828)

## How was this PR tested?

- [x] [PASS TIER IV INTERNAL SCENARIOS](https://evaluation.tier4.jp/evaluation/reports/10863e56-4bb0-5b16-84ab-6547d1d42b0d?project_id=prd_jt)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_path_planner/autoware_behavior_path_static_obstacle_avoidance_module/include/autoware/behavior_path_static_obstacle_avoidance_module/debug.hpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_static_obstacle_avoidance_module/include/autoware/behavior_path_static_obstacle_avoidance_module/debug.hpp
@@ -48,7 +48,8 @@ MarkerArray createAmbiguousObjectsMarkerArray(
 MarkerArray createStopTargetObjectMarkerArray(const AvoidancePlanningData & data);
 
 MarkerArray createDebugMarkerArray(
-  const AvoidancePlanningData & data, const PathShifter & shifter, const DebugData & debug,
+  const BehaviorModuleOutput & output, const AvoidancePlanningData & data,
+  const PathShifter & shifter, const DebugData & debug,
   const std::shared_ptr<AvoidanceParameters> & parameters);
 }  // namespace autoware::behavior_path_planner::utils::static_obstacle_avoidance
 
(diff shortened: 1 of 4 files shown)


shmpwk (approved):
LGTM!

PR #8729, opened 2024-09-03
refactor(cuda_utils): move cuda_utils to sensing

YoshiRi:
## Description

According to https://github.com/autowarefoundation/autoware.universe/issues/8695, we want to move cuda_utils from common packages to sensing.

This PR assumes that only perception and sensing are going to use cuda_utils.

## Related links

**Parent Issue:**

- TIER IV INTERNAL DISCUSSION: https://star4.slack.com/archives/C076E8EBJTG/p1725244585043449

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


knzo25 (approved):
LGTM (I will send a PR to add me as a maintainer instead of Nishimatsu-san later)

manato (approved):
LGTM. 
Note: this `cuda_utils` package is currently used in sensing/perception packagesonly; so we move this package into the sensing package to mitigate dependencies in autoware.universe.  Once other packages, such as planning, will start to consider CUDA implementation, we'll move this package into `common` again.

PR #8737, opened 2024-09-03
feat(compare_map_segmentation): publish warning when pcl::voxelgrid failed to filter with large leaf size

YoshiRi:
## Description

pcl::VoxelGrid fails when too large leaf size is needed.（See L243 in [this file](https://pointclouds.org/documentation/voxel__grid_8hpp_source.html)）
This PR enables to publish warning when input and output pointcloud number is the same in the pcl voxelgrid filter.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_compare_map_segmentation/src/voxel_distance_based_compare_map_filter/node.cpp
+++ b/perception/autoware_compare_map_segmentation/src/voxel_distance_based_compare_map_filter/node.cpp
@@ -39,9 +39,18 @@ void VoxelDistanceBasedStaticMapLoader::onMapCallback(
   // voxel
   voxel_map_ptr_.reset(new pcl::PointCloud<pcl::PointXYZ>);
   voxel_grid_.setLeafSize(voxel_leaf_size_, voxel_leaf_size_, voxel_leaf_size_);
+  const bool is_feasible = isFeasibleWithPCLVoxelGrid(map_pcl_ptr, voxel_grid_);
+  if (!is_feasible) {
+    RCLCPP_ERROR(
+      logger_,
+      "Voxel grid filter is not feasible. Check the voxel grid filter parameters and input "
+      "pointcloud.");
+    throw std::runtime_error("Voxel grid filter is not feasible with input pointcloud.");
+  }
   voxel_grid_.setInputCloud(map_pcl_ptr);
   voxel_grid_.setSaveLeafLayout(true);
   voxel_grid_.filter(*voxel_map_ptr_);
+
   // kdtree
   map_ptr_ = map_pcl_ptr;
 
--- a/perception/autoware_compare_map_segmentation/src/voxel_distance_based_compare_map_filter/node.cpp
+++ b/perception/autoware_compare_map_segmentation/src/voxel_distance_based_compare_map_filter/node.cpp
@@ -30,6 +30,8 @@ namespace autoware::compare_map_segmentation
 void VoxelDistanceBasedStaticMapLoader::onMapCallback(
   const sensor_msgs::msg::PointCloud2::ConstSharedPtr map)
 {
+  std::cout << "VoxelDistanceBasedStaticMapLoader::onMapCallback()" << std::endl;
+
   pcl::PointCloud<pcl::PointXYZ> map_pcl;
   pcl::fromROSMsg<pcl::PointXYZ>(*map, map_pcl);
   const auto map_pcl_ptr = pcl::make_shared<pcl::PointCloud<pcl::PointXYZ>>(map_pcl);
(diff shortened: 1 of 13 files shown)


perception/autoware_compare_map_segmentation/src/voxel_distance_based_compare_map_filter/node.cpp
```diff
@@ -42,6 +42,17 @@ void VoxelDistanceBasedStaticMapLoader::onMapCallback(
...
+  // sanity check: check pointcloud size before and after filters
+  const auto voxel_pc_size = voxel_map_ptr_->size();
+  const auto original_pc_size = map_pcl_ptr->size();
+  const bool is_not_filtered = (voxel_pc_size == original_pc_size);
+  if (is_not_filtered) {
+    RCLCPP_WARN(
+      logger_,
+      "Pointcloud has not downsized. If this is not intended, something wrong with PCL voxel grid. "
+      "Check the voxel grid filter parameters.");
+  }
```
yukkysaito, line 54: When an error occurs, isn't an exception (try-catch) supposed to be thrown?

YoshiRi: Hmm, I thought it depends on cases.
I only checking intput/output pointcloud number and it could be the normal operation with sparse input pointcloud and small voxel leaf size.

yukkysaito: The reason is that with the code mentioned above, if the input point cloud is sparse, there is a possibility that the number of points in the point cloud before and after filtering may match.

YoshiRi: Basically the voxel index overflow problem is handled by PCL warning, not the error. So we can not catch it.
https://pointclouds.org/documentation/voxel__grid_8hpp_source.html

If we want to detect this for ourself, we need to check max range of pointclouds.

badai-nguyen: @YoshiRi cc @yukkysaito 
IMO, we cat catch warning from PCL_WARN by using redirectStderrToStream `stderr` to check where `Leaf size is too small for the input dataset...` is output for VoxelGrid Filter or not to RCLCPP_ERROR and stop autoware execution?
I tried [here](https://github.com/autowarefoundation/autoware.universe/compare/main...badai-nguyen:feat/compare_map_filter/catch_overflow_error?expand=1#diff-6fcc7c77ce6950930e74d4badcd339667191e227f50b9eeb68ba4e6a53754fda) and confirmed it work as disired.
```
[component_container_mt-1] [ERROR] [1727941704.709987775] [perception.object_recognition.detection.voxel_based_compare_map_filter]: Leaf size is too small for the input dataset. Integer indices would overflow. PCL_WARN was catch!
[component_container_mt-1] terminate called after throwing an instance of 'std::runtime_error'
[component_container_mt-1]   what():  Voxel grid filter is not feasible with input pointcloud.
[component_container_mt-1] *** Aborted at 1727941704 (unix time) try "date -d @1727941704" if you are using GNU date ***
[component_container_mt-1] PC: @                0x0 (unknown)
[component_container_mt-1] *** SIGABRT (@0x3e80004998d) received by PID 301453 (TID 0x74eb4aa00640) from PID 301453; stack trace: ***
[component_container_mt-1]     @     0x74eb54978046 (unknown)
[component_container_mt-1]     @     0x74eb56642520 (unknown)
[component_container_mt-1]     @     0x74eb566969fc pthread_kill
[component_container_mt-1]     @     0x74eb56642476 raise
[component_container_mt-1]     @     0x74eb566287f3 abort
[component_container_mt-1]     @     0x74eb56aa2b9e (unknown)
[component_container_mt-1]     @     0x74eb56aae20c (unknown)
[component_container_mt-1]     @     0x74eb56aae277 std::terminate()
[component_container_mt-1]     @     0x74eb56aae4d8 __cxa_throw
[component_container_mt-1]     @     0x74eb0c326306 _ZN24VoxelGridStaticMapLoader13onMapCallbackESt10shared_ptrIKN11sensor_msgs3msg12PointCloud2_ISaIvEEEE.cold
[component_container_mt-1]     @     0x74eb0c3a7f18 std::_Function_handler<>::_M_invoke()
[component_container_mt-1]     @     0x74eb31e70c47 _ZNSt8__detail9__variant17__gen_vtable_implINS0_12_Multi_arrayIPFNS0_21__deduce_visit_resultIvEEOZN6rclcpp23AnySubscriptionCallbackIN11sensor_msgs3msg12PointCloud2_ISaIvEEESA_E8dispatchESt10shared_ptrISB_ERKNS5_11MessageInfoEEUlOT_E_RSt7variantIJSt8functionIFvRKSB_EESN_IFvSP_SH_EESN_IFvRKNS5_17SerializedMessageEEESN_IFvSW_SH_EESN_IFvSt10unique_ptrISB_St14default_deleteISB_EEEESN_IFvS14_SH_EESN_IFvS11_ISU_S12_ISU_EEEESN_IFvS1A_SH_EESN_IFvSD_ISO_EEESN_IFvS1F_SH_EESN_IFvSD_ISV_EEESN_IFvS1K_SH_EESN_IFvRKS1F_EESN_IFvS1Q_SH_EESN_IFvRKS1K_EESN_IFvS1W_SH_EESN_IFvSE_EESN_IFvSE_SH_EESN_IFvSD_ISU_EEESN_IFvS25_SH_EEEEEJEEESt16integer_sequenceImJLm8EEEE14__visit_invokeESL_S2B_

```
@YoshiRi if you agree with this way, please take it as suggestion or I can create another PR for this matter?

YoshiRi:
@badai-nguyen 
really!?
that's good news! I will fix tonight.

perception/autoware_compare_map_segmentation/src/voxel_distance_based_compare_map_filter/node.cpp
YoshiRi: @badai-nguyen
That's good news! I will take a look tonight.

P.S.
I saw your code. It's interesting but I don't think it is better idea to capture error stream.
Let us find another method to solve this 🥹

mitsudome-r:
@YoshiRi This has been stale for a long time. Could you check if you still need this PR?

technolojin:
Implemented by https://github.com/autowarefoundation/autoware_universe/pull/10381

Issue #8742, opened 2024-09-03
autoware_pointcloud_preprocessor tests do not link properly


esteve:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I'm convinced that this is not my fault but a bug.

### Description

The templates in autoware_pointcloud_preprocessor are not instantiated so they can't be properly linked with the tests.

### Expected behavior

All binaries in the project linked without issues

### Actual behavior

Linking with stricter flags produces the current error:

```console
2024-08-26T22:06:12.8648001Z [100%] Built target test_utilities
2024-08-26T22:06:14.7299232Z /usr/bin/ld: /tmp/ccvFi0NQ.ltrans0.ltrans.o: in function `DistortionCorrectorTest_TestIsInputValid_Test::TestBody()':
2024-08-26T22:06:14.7305828Z ./obj-x86_64-linux-gnu/./test/test_distortion_corrector_node.cpp:312: undefined reference to `autoware::pointcloud_preprocessor::DistortionCorrector<autoware::pointcloud_preprocessor::DistortionCorrector2D>::isInputValid(sensor_msgs::msg::PointCloud2_<std::allocator<void> >&)'
2024-08-26T22:06:14.7309918Z /usr/bin/ld: ./obj-x86_64-linux-gnu/./test/test_distortion_corrector_node.cpp:320: undefined reference to `autoware::pointcloud_preprocessor::DistortionCorrector<autoware::pointcloud_preprocessor::DistortionCorrector2D>::isInputValid(sensor_msgs::msg::PointCloud2_<std::allocator<void> >&)'
2024-08-26T22:06:14.7315092Z /usr/bin/ld: ./obj-x86_64-linux-gnu/./test/test_distortion_corrector_node.cpp:325: undefined reference to `autoware::pointcloud_preprocessor::DistortionCorrector<autoware::pointcloud_preprocessor::DistortionCorrector2D>::isInputValid(sensor_msgs::msg::PointCloud2_<std::allocator<void> >&)'
[...]

PR #8743, opened 2024-09-03
fix(autoware_pointcloud_preprocessor): instantiate templates so that the symbols exist when linking

esteve:
## Description

This PR instantiates the DistortionCorrector template so that when linking against it the symbols can be found.

## Related links

**Parent Issue:**

- Link https://github.com/autowarefoundation/autoware.universe/issues/8742

## How was this PR tested?

Built Autoware from source in the buildfarm.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


YoshiRi (approved):
LGTM

PR #8750, opened 2024-09-04
feat(lane_change): add checks to ensure the edge of vehicle do not exceed target lane boundary when changing lanes

mkquda:
## Description

Currently, LC path does not guarantee that ego vehicle is within target lane boundaries at all times during LC maneuver. Specifically for larger vehicles (such as bus or trucks), it is more likely that the vehicle frame exceeds the target lane boundary into the next lane during the maneuver, which can result in an unsafe interaction with other vehicles on the road.

This PR adds a safety check for LC candidate path to ensure ego footprint maintains a safe distance from the target lane far boundary for all path points.

If a candidate path does not satisfy the condition it will still be added as a valid path, but will not be considered a safe path.

## Related links

[autoware_launch PR](https://github.com/autowarefoundation/autoware_launch/pull/1157)

## How was this PR tested?

PSIM.

## Notes for reviewers

None.

#### Additions and removals

| Added | `enable_target_lane_bound_check`   | `bool` | `true` | Flag to enable / disable target lane boundary check |

## Effects on system behavior

LC candidates path that result in ego foot print exceeding target lane far boundary will be considered unsafe.

### DIFF
--- a/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/utils/utils.cpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/utils/utils.cpp
@@ -213,7 +213,8 @@ bool isPathInLanelets(
 }
 
 bool pathFootprintExceedsTargetLaneBound(
-  const CommonDataPtr & common_data_ptr, const PathWithLaneId & path, const VehicleInfo & ego_info)
+  const CommonDataPtr & common_data_ptr, const PathWithLaneId & path, const VehicleInfo & ego_info,
+  const double margin)
 {
   if (common_data_ptr->direction == Direction::NONE || path.points.empty()) {
     return false;
@@ -233,7 +234,7 @@ bool pathFootprintExceedsTargetLaneBound(
     const auto dist_to_boundary =
       sign * utils::getSignedDistanceFromLaneBoundary(combined_target_lane, front_vertex, is_left);
 
-    if (dist_to_boundary < 0.0) {
+    if (dist_to_boundary < margin) {
       return true;
     }
   }
--- a/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/utils/utils.cpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/utils/utils.cpp
@@ -225,16 +225,16 @@ bool pathFootprintExceedsTargetLaneBound(
 
   const auto combined_target_lane = lanelet::utils::combineLaneletsShape(target_lanes);
 
-  auto it = path.points.rbegin();
-  for (; it < path.points.rend(); ++it) {
-    const auto & pose = it->point.pose;
-    const auto & front_vertex = getEgoFrontVertex(pose, ego_info, is_left);
+  for (const auto & path_point : path.points) {
+    const auto & pose = path_point.point.pose;
+    const auto front_vertex = getEgoFrontVertex(pose, ego_info, is_left);
 
-    const double sign = is_left ? -1.0 : 1.0;
+    const auto sign = is_left ? -1.0 : 1.0;
     const auto dist_to_boundary =
       sign * utils::getSignedDistanceFromLaneBoundary(combined_target_lane, front_vertex, is_left);
 
     if (dist_to_boundary < margin) {
+      RCLCPP_DEBUG(get_logger(), "Path footprint exceeds target lane boundary");
       return true;
     }
   }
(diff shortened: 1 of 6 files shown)


planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/utils/utils.cpp
```diff
@@ -287,6 +287,36 @@ bool isPathInLanelets(
...
+  const bool is_left = common_data_ptr->direction == Direction::LEFT;
+
+  const auto combined_target_lane = lanelet::utils::combineLaneletsShape(target_lanes);
+
+  auto it = path.points.rbegin();
+  for (; it < path.points.rend(); ++it) {
+    const auto & pose = it->point.pose;
+    const auto & front_vertex = getEgoFrontVertex(pose, ego_info, is_left);
+
+    const double sign = is_left ? -1.0 : 1.0;
```
zulfaqar-azmi-t4, line 308:
```suggestion
    const auto sign = is_left ? -1.0 : 1.0;
```
Suggest using auto most of the time for consistency.

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/utils/utils.cpp
```diff
@@ -287,6 +287,36 @@ bool isPathInLanelets(
...
+
+  const auto & target_lanes = common_data_ptr->lanes_ptr->target;
+  const bool is_left = common_data_ptr->direction == Direction::LEFT;
+
+  const auto combined_target_lane = lanelet::utils::combineLaneletsShape(target_lanes);
+
+  auto it = path.points.rbegin();
+  for (; it < path.points.rend(); ++it) {
+    const auto & pose = it->point.pose;
+    const auto & front_vertex = getEgoFrontVertex(pose, ego_info, is_left);
```
zulfaqar-azmi-t4, line 306:
```suggestion
    const auto front_vertex = getEgoFrontVertex(pose, ego_info, is_left);
```
Reference doesn't have any effects in this case.

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/utils/utils.cpp
```diff
@@ -1123,6 +1153,16 @@ Polygon2d getEgoCurrentFootprint(
...
 }
 
+Point getEgoFrontVertex(
+  const Pose & ego_pose, const autoware::vehicle_info_utils::VehicleInfo & ego_info, bool left)
+{
+  const double lon_offset = ego_info.wheel_base_m + ego_info.front_overhang_m;
+  const double lat_offset = 0.5 * (left ? ego_info.vehicle_width_m : -ego_info.vehicle_width_m);
+  const auto vertex =
+    autoware::universe_utils::calcOffsetPose(ego_pose, lon_offset, lat_offset, 0.0).position;
+  return vertex;
```
zulfaqar-azmi-t4, line 1163: why not directly call calcOffsetPose in the return statement?

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/utils/utils.cpp
```diff
@@ -287,6 +287,36 @@ bool isPathInLanelets(
...
+    return false;
+  }
+
+  const auto & target_lanes = common_data_ptr->lanes_ptr->target;
+  const bool is_left = common_data_ptr->direction == Direction::LEFT;
+
+  const auto combined_target_lane = lanelet::utils::combineLaneletsShape(target_lanes);
+
+  auto it = path.points.rbegin();
+  for (; it < path.points.rend(); ++it) {
```
zulfaqar-azmi-t4, line 304: curious why you didn't add it in the for statement?

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/utils/utils.cpp
```diff
@@ -287,6 +287,36 @@ bool isPathInLanelets(
...
+    const auto & pose = it->point.pose;
+    const auto & front_vertex = getEgoFrontVertex(pose, ego_info, is_left);
+
+    const double sign = is_left ? -1.0 : 1.0;
+    const auto dist_to_boundary =
+      sign * utils::getSignedDistanceFromLaneBoundary(combined_target_lane, front_vertex, is_left);
+
+    if (dist_to_boundary < margin) {
+      return true;
```
zulfaqar-azmi-t4, line 313: can you add RCLCPP_DEBUG here?

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/utils/utils.cpp
mkquda: just thought it might help with readability, but actually I realized we don't need to use reverse iterator, so will change it to use range-based for

zulfaqar-azmi-t4 (approved):

PR #8751, opened 2024-09-04
perf(autoware_map_based_prediction): replace pow

TakaHoribe:
## Description

A performance analysis of the `map_based_prediction` with [VTune](https://www.intel.com/content/www/us/en/developer/tools/oneapi/vtune-profiler.html) was performed. The generated frame graph is below.
image

For example, you see `std::pow` accounts for 50% of the `generateFrenetPath` function's processing time.
image

Replacing `std::pow` with a simple multiplier reduces the computation time of this function from `2.2` ms to `0.6` ms for 1000 calculations. This is the PR.
(The function is called for each combination of "the number of objects" and "the number of predicted paths".)

## Related links

None

## How was this PR tested?

- I visually confirmed that the prediction is working correctly in psim.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


takayuki5168 (approved):

YoshiRi (approved):
LGTM

PR #8754, opened 2024-09-04
fix(static_obstacle_avoidance): ignore objects which has already been decided to avoid

satoshi-ota:
## Description

There is a issue that the avoidance module did wrong avoidable/unavoidable judgement even after it decided to avoid. I found this was caused by the logic to check longitudinal distance, and the module judged it was not avoidable because there was not enough longitudinal distance when the ego got close to avoidance target object.

https://github.com/user-attachments/assets/693c3e24-c33c-4020-9f06-a02e805efa16

In this PR, I fixed the logic to ignore object if the module generates avoidance path for it.

```c++
  const auto is_approved = [this](const auto & object) {
    return (helper_->getShift(object.getPosition()) > 0.0 && isOnRight(object)) ||
           (helper_->getShift(object.getPosition()) < 0.0 && !isOnRight(object));
  };
```

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

https://github.com/user-attachments/assets/17fa8073-6d32-41bd-a06d-4ff22b05b7b4

- [x] [PASS TIER IV INTERNAL SCENARIOS](https://evaluation.tier4.jp/evaluation/reports/10863e56-4bb0-5b16-84ab-6547d1d42b0d?project_id=prd_jt)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


shmpwk (approved):
LGTM

PR #8757, opened 2024-09-04
feat(mission_planner): add option to prevent rerouting in autonomous driving mode

kosuke55:
## Description

add option to prevent rerouting in autonomous driving mode
set true by default.
if false, reroute fails. if true, only safe reroute is allowed.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

psim

set `false`

https://github.com/user-attachments/assets/6655a9d1-d3d8-4b57-8047-2861de81240c

2024/09/04 https://evaluation.tier4.jp/evaluation/reports/00d52220-01c8-5767-ae65-97fd3f71150c/?project_id=prd_jt

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/autoware_mission_planner/README.md
+++ b/planning/autoware_mission_planner/README.md
@@ -19,17 +19,18 @@ It distributes route requests and planning results according to current MRM oper
 
 ### Parameters
 
-| Name                         | Type   | Description                                                                                                      |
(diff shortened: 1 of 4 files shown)


takayuki5168 (approved):

tkimura4 (approved):
LGTM

PR #8761, opened 2024-09-04
ci(build-and-test): reset ccache stats before build

xmfcx:
## Description

Right now ccache stats are not correct because it accumulates over time.

> https://linux.die.net/man/1/ccache
> 
> ```
> -z, --zero-stats
> Zero the cache statistics (but not the configured limits).
> ```

This command will reset it before the build. So we can get to know the real stats after build.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/.github/actions/build-and-test-differential/action.yaml
+++ b/.github/actions/build-and-test-differential/action.yaml
@@ -57,8 +57,10 @@ runs:
         restore-keys: |
           ccache-main-${{ runner.arch }}-${{ inputs.rosdistro }}-
 
-    - name: Show ccache stats before build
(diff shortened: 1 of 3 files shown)


xmfcx:
https://github.com/autowarefoundation/autoware.universe/actions/runs/10702198996/job/29669865121?pr=8761#step:5:10599

image

96% hit rate, it works :D

Thanks @mojomex 

Now I will revert the dummy change and we can merge

xmfcx:
Merging this as is since it's a simple change.

Issue #8460, opened 2024-08-13
`segmentation_pointcloud_fusion` can't work proper with multiple mask


StepTurtle:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I'm convinced that this is not my fault but a bug.

### Description

I want to use `segmentation_pointcloud_fusion` node with 6 cameras and I have a problem. When I have multiple cameras and multiple `/perception/object_recognition/detection/mask?` outputs from `autoware_tensorrt_yolox`, my expectation is to see point cloud which fused from all cameras. However I just see the point cloud fused from a single camera, especially the last camera from [this loop](https://github.com/autowarefoundation/autoware.universe/blob/bbc85252a5ebf6ea9cd901b536f051c5bd57cff2/perception/autoware_image_projection_based_fusion/src/fusion_node.cpp#L217)

In the following loop located in the FusionNode<TargetMsg3D, Obj, Msg2D>::subCallback() function, the output point cloud data is cleared at every step, and the points deleted in the previous step are restored.

```cpp
for (std::size_t roi_i = 0; roi_i < rois_number_; ++roi_i) {

    // ...

    fuseOnSingleImage(
              *input_msg, roi_i, *((cached_roi_msgs_.at(roi_i))[matched_stamp]),
              camera_info_map_.at(roi_i), *output_msg);

    // ...

}

```

As a result of this, I just see that the only points remove related with the last camera. Did I make something wrong, or is it the expected behavior?

fyi. @badai-nguyen, I guess you are the maintainer of this package or node.

### Expected behavior

I expect to see something like that when I run the `segmentation_pointcloud_fusion` node: https://youtu.be/Ho4lkrmtuz4

To obtain this results, I make some basic changes in `fusion_node.cpp` and `segmentation_pointcloud_fusion/node.cpp`.

### Actual behavior

I see something like that when I run the `segmentation_pointcloud_fusion` node: https://youtu.be/GaNBf3vUcOo

### Steps to reproduce

I did not run whole Autoware, I just launch `yolox` and `image_projection_based_fusion`.

There are six cameras and one lidar topic in my bag file.

1. `ros2 launch autoware_tensorrt_yolox multiple_yolox.launch.xml`
[...]

PR #8769, opened 2024-09-05
fix(autoware_image_projection_based_fusion): resolve issue with segmentation pointcloud fusion node failing with multiple mask inputs

vividf:
## Description

This PR addresses [issue #8460](https://github.com/autowarefoundation/autoware.universe/issues/8460). The current design of the segmentation pointcloud fusion node is incompatible with the fusion base class because the `fuseOnSingleImage` function always uses the original input pointcloud in each iteration (camera), instead of the pointcloud that has filtered out points that fall outside the mask before processing. In the end, the current algorithm will lead to the problem that the output pointcloud will only filter out the points with the last camera's mask, instead of all cameras' masks

A simple fix would be to copy the output pointcloud and use it as the input for the next iteration. However, this approach increases processing time by approximately 0.2 ms per iteration (camera). Moreover, it impacts four other nodes in the `autoware_image_projection_based_fusion` package.

To avoid these drawbacks, I opted to modify the algorithm directly within the segmentation pointcloud fusion node. The new solution involves storing the offsets of points that are filtered out in a set. During post-processing, the points in the set are ignored, and the remaining points are copied to the output pointcloud.
## Related links

**Parent Issue:**
- Link: https://github.com/autowarefoundation/autoware.universe/issues/8460

## How was this PR tested?
I test the PR with xx1 bag with using three camera (0, 1, 2)

launch yolox segmentation (modify multiple_yolox.launch.xml by adding output/mask)
```
ros2 launch autoware_tensorrt_yolox multiple_yolox.launch.xml
```
launch fusion node
```
ros2 launch autoware_image_projection_based_fusion segmentation_pointcloud_fusion.launch.xml
```
launch the xx1 tf
```
ros2 launch static_tf_publisher static_transforms_launch.py
```

Before (only filter out the camera 2 information (left) )

Screenshot from 2024-09-05 15-50-08

After
Screenshot from 2024-09-05 15-49-55

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


badai-nguyen (approved):
@vividf I really appreciate your PR.
LGTM

PR #8783, opened 2024-09-06
fix(autoware_route_handler): fix unusedFunction

Nagi70:
## Description

This is a fix based on cppcheck unusedFunction warnings.

```
common/autoware_test_utils/src/autoware_test_utils.cpp:141:0: style: The function 'get_absolute_path_to_route' is never used. [unusedFunction]
std::string get_absolute_path_to_route(
^
```

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/autoware_route_handler/test/test_route_handler.cpp
+++ b/planning/autoware_route_handler/test/test_route_handler.cpp
@@ -81,7 +81,6 @@ TEST_F(TestRouteHandler, getLaneletSequenceWhenOverlappingRoute)
 TEST_F(TestRouteHandler, getClosestRouteLaneletFromLaneletWhenOverlappingRoute)
 {
   set_route_handler("overlap_map.osm");
-  set_test_route("overlap_test_route.yaml");
   ASSERT_TRUE(route_handler_->isHandlerReady());
 
   geometry_msgs::msg::Pose reference_pose;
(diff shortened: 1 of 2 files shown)


takayuki5168:
@mkquda Can you review the PR?

zulfaqar-azmi-t4:
@Nagi70 
`set_test_route` is a necessary function, and removing it will cause the test to fail.

 May I ask your help to clarify the following details:
 ```
The get_absolute_path_to_route function in the autoware_test_utils package was only used for testing in the autoware_route_handler package. Therefore, the set_test_route function has been removed.
```
There is no doubt `get_absolute_path_to_route` is used, and I didn't see the reason to remove the function, and I also didn't understand why `cppcheck` outputs the function in unused🤔

kobayu858:
@zulfaqar-azmi-t4 

First, cppcheck warns that get_absolute_path_to_route is unusedFunciton. There are two reasons for this: get_absolute_path_to_route is not used outside of the test folder, and the test folder is not included in the analysis. This is in accordance with the CI specification.

Also, as a policy to resolve unusedFunciton warnings, functions that are only used within a test function are removed from that entire test. Reference ( #8605 )

In this case, I believe the error is due to the fact that what is deleted extends to the constructor.

kobayu858:
@zulfaqar-azmi-t4 

In this case, since the test function and `get_absolute_path_to_route` are separate packages, we are trying to remove the test function first.

However, if I need to remove even the constructor, as in this case, I believe that I need to remove the test itself.

What are your thoughts on this?

kobayu858:
The goal was to ignore the function warnings associated with test, but the cppcheck command was not properly configured, resulting in the warnings in this PR.

We plan to fix the command in a future PR, so the warnings in this PR will go away.

For these reasons, this PR is closed.

PR #8798, opened 2024-09-06
perf(motion_velocity_smoother): remove some heavy debug logging

maxime-clem:
## Description

Some debug logging of the `analytical_jerk_constrained_smoother` can take a lot of time (I observed up to 20ms when running locally), even when debug logging is disabled.
I have not found a nice solution so this PR simply removes the logging code.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


satoshi-ota (approved):
LGTM

PR #8817, opened 2024-09-10
feat(crosswalk): suppress restart when the ego is close to the next stop point

takayuki5168:
## Description

When the ego stops and is close to the next stop point where the pedestrian is walking in the front crosswalk, it will be dangerous and uncomfortable if the ego restarts a little bit towards the pedestrian.
Therefore, this PR suppresses the restart in this condition.

The video shows
- the crosswalk stop position is on the lanelet map's stop line as usual at first
- after the ego stops, the crosswalk stop position is the ego's front by this PR
- after the ego restarts, the crosswalk stop position is on the lanelet map's stop line as usual.
NOTE: I increased the distance parameter to 5.0m for the demonstration.

https://github.com/user-attachments/assets/7191c2ff-e34a-4e23-ab52-a29b37b1aa23

## How was this PR tested?

scenario test (internal link):
https://evaluation.tier4.jp/evaluation/reports/10b82881-f9bc-5329-b893-90344215df4a?project_id=prd_jt

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_velocity_planner/autoware_behavior_velocity_crosswalk_module/src/scene_crosswalk.cpp
+++ b/planning/behavior_velocity_planner/autoware_behavior_velocity_crosswalk_module/src/scene_crosswalk.cpp
@@ -1325,6 +1325,8 @@ bool CrosswalkModule::checkRestartSuppression(
   const double dist_to_stop =
     calcSignedArcLength(ego_path.points, ego_pos, stop_factor->stop_pose.position);
 
+  // NOTE: min_dist_to_stop_for_restart_suppression is supposed to be the same as
+  //      the pid_longitudinal_controller's drive_state_stop_dist.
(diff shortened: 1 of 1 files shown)


yuki-takagi-66:
@takayuki5168 
I can't imagine why `min_distance_to_stop` is necessary. Could you let me know?

yuki-takagi-66 (approved):
Overall, LGTM to merge

takayuki5168:
@yuki-takagi-66 
There are two reasons
- when the ego stops for the crosswalk exceeding the ideal crosswalk stop position a bit, the planned crosswalk stop position will be moved to the ego's front.
- when the ego stops for the crosswalk, usually where the ego stops is not completely the same as the ideal crosswalk stop position. In this case, the stop position is slightly moved to the ego's front. This happens everytime the ego stops for the crosswalk, which is very weird.

PR #8820, opened 2024-09-10
refactor(autoware_lidar_centerpoint): use std::size_t instead of size_t

kminoda:
## Description
Use `std::size_t` instead of `size_t`

## How was this PR tested?
Build works

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


knzo25 (approved):
LGTM

PR #8826, opened 2024-09-10
feat(remaining_distance_time_calculator): integrate generate_parameter_library

mitukou1109:
## Description

This PR enables remaining_distance_time_calculator to handle parameters using [generate_parameter_library](https://github.com/PickNikRobotics/generate_parameter_library).

## How was this PR tested?

Psim

- [x] errors occur if any parameters are missing in the config file in remaining_distance_time_calculator package

Evaluator: https://evaluation.tier4.jp/evaluation/reports/51089edf-cf82-58b3-b7c9-c571753ad766?project_id=prd_jt

## Notes for reviewers

The node `remaining_distance_time_calculator` does not have dynamic parameters.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/autoware_remaining_distance_time_calculator/CMakeLists.txt
+++ b/planning/autoware_remaining_distance_time_calculator/CMakeLists.txt
@@ -16,7 +16,7 @@ target_link_libraries(${PROJECT_NAME}
   remaining_distance_time_calculator_parameters
 )
 
-target_compile_options(${PROJECT_NAME} PUBLIC
+target_compile_options(${PROJECT_NAME} PRIVATE
   -Wno-error=deprecated-declarations
 )
 
--- a/planning/autoware_remaining_distance_time_calculator/CMakeLists.txt
+++ b/planning/autoware_remaining_distance_time_calculator/CMakeLists.txt
@@ -4,8 +4,16 @@ project(autoware_remaining_distance_time_calculator)
 find_package(autoware_cmake REQUIRED)
 autoware_package()
 
+generate_parameter_library(remaining_distance_time_calculator_parameters
+  param/remaining_distance_time_calculator_parameters.yaml
+)
+
 ament_auto_add_library(${PROJECT_NAME} SHARED
   src/remaining_distance_time_calculator_node.cpp
+)
+
+target_link_libraries(${PROJECT_NAME}
+  remaining_distance_time_calculator_parameters
 )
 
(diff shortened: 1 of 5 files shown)


satoshi-ota (approved):
LGTM :+1:

planning/autoware_remaining_distance_time_calculator/CMakeLists.txt
```diff
@@ -4,10 +4,22 @@ project(autoware_remaining_distance_time_calculator)
...
+
 ament_auto_add_library(${PROJECT_NAME} SHARED
   src/remaining_distance_time_calculator_node.cpp
 )
 
+target_link_libraries(${PROJECT_NAME}
+  remaining_distance_time_calculator_parameters
+)
+
+target_compile_options(${PROJECT_NAME} PUBLIC
```
ahmeddesokyebrahim, line 19: Actually this CMake target compile options rule caught my attention as being `PUBLIC` to suppress deprecated errors.
Using `PUBLIC` in this case is not risky but unnecessary. 
- It is not risky because the error suppression will not propagate to other targets, as other targets cannot "link" to an executable target and here we are using this command with an executable.
- It unnecessary from the sense of best practices, as `PRIVATE` is often the better choice, keeping the compile option strictly for that target. 

That is why I do really recommend to have this rule as 
```
target_compile_options(${PROJECT_NAME} PRIVATE
  -Wno-error=deprecated-declarations
)
```

ahmeddesokyebrahim (changes requested):
Thanks a lot @mitukou1109 -san for your PR. 

It looks great from all sides and ready to be merged. 
I have only one minor comment that you can have a look it.

planning/autoware_remaining_distance_time_calculator/CMakeLists.txt
mitukou1109: Thank you for pointing it out, I fixed it just now https://github.com/autowarefoundation/autoware.universe/pull/8826/commits/e42d4e4e4468a5bfdd3476a1468b599abc672939

ahmeddesokyebrahim (approved):
Thanks again for your efforts
LGTM 👍

mitsudome-r:
@mitukou1109 @satoshi-ota @takayuki5168 The PR is approved. Do you have any concerns with merging this PR?

takayuki5168:
@satoshi-ota 
I don't know the background of this PR, so please handle it.

PR #8840, opened 2024-09-10
fix(tier4_adapi_rviz_plugin): fix unusedFunction

kobayu858:
## Description
Preparation for future CI changes.

```
common/tier4_adapi_rviz_plugin/src/state_panel.cpp:29:0: style: The function 'Bool2String' is never used. [unusedFunction]
inline std::string Bool2String(const bool var)
^
```
## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


isamu-takagi (approved):
LGTM

PR #8845, opened 2024-09-10
fix(tier4_state_rviz_plugin): fix unusedFunction

kobayu858:
## Description
Preparation for future CI changes.

I have tested and confirmed this, but would like to know if it is being used.
```
common/tier4_state_rviz_plugin/src/include/autoware_state_panel.hpp:226:0: style: The function 'updateCustomLabel' is never used. [unusedFunction]
  static void updateCustomLabel(
^
common/tier4_state_rviz_plugin/src/include/autoware_state_panel.hpp:232:0: style: The function 'updateButton' is never used. [unusedFunction]
  static void updateButton(QPushButton * button, QString text, QString style_sheet)
^
```
## Related links

**Parent Issue:**

- Link

## How was this PR tested?
Confirm that the planning simulation works.
Confirm that print debugging is not used.
Confirm that the build passes.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/common/tier4_state_rviz_plugin/src/include/autoware_state_panel.hpp
+++ b/common/tier4_state_rviz_plugin/src/include/autoware_state_panel.hpp
@@ -223,17 +223,6 @@ protected:
     label->setText(text);
     label->setStyleSheet(style_sheet);
   }
-  static void updateCustomLabel(
-    CustomLabel * label, QString text, QColor bg_color, QColor text_color)
-  {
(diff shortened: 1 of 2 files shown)


KhalilSelyan:
I'll have to double check these, will confirm this later tonight or at latest tomorrow

KhalilSelyan (approved):
These are not used anymore can be removed

PR #8849, opened 2024-09-11
refactor(lane_change): add new structs (TransientData) to store commonly used values

zulfaqar-azmi-t4:
## Description

This PR adds `TransientData` to standardize commonly used values in the lane change module, such as the current lanes' buffer and the distance to the terminal start and end. The goal is to ensure consistency when calculating limits and lengths.

## Related links

**Parent Issue:**

- None

## How was this PR tested?

1. Build complete
2. PSIM
3. Scenario Test TBD

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_path_planner/autoware_behavior_path_avoidance_by_lane_change_module/src/scene.cpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_avoidance_by_lane_change_module/src/scene.cpp
@@ -281,7 +281,7 @@ double AvoidanceByLaneChange::calcMinimumLaneChangeLength() const
     return std::numeric_limits<double>::infinity();
   }
 
-  return utils::lane_change::calculation::calc_minimum_lane_change_length(
+  return utils::lane_change::calculation::calc_minimum_lane_change_buffer(
     common_data_ptr_, current_lanes);
 }
 
--- a/planning/behavior_path_planner/autoware_behavior_path_avoidance_by_lane_change_module/src/scene.cpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_avoidance_by_lane_change_module/src/scene.cpp
@@ -282,7 +282,7 @@ double AvoidanceByLaneChange::calcMinimumLaneChangeLength() const
   }
 
   return utils::lane_change::calculation::calc_minimum_lane_change_length(
-    getRouteHandler(), current_lanes.back(), *lane_change_parameters_, direction_);
+    common_data_ptr_, current_lanes);
 }
 
 double AvoidanceByLaneChange::calcLateralOffset() const
--- a/planning/behavior_path_planner/autoware_behavior_path_avoidance_by_lane_change_module/src/scene.cpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_avoidance_by_lane_change_module/src/scene.cpp
@@ -82,7 +82,7 @@ bool AvoidanceByLaneChange::specialRequiredCheck() const
 
   const auto & nearest_object = data.target_objects.front();
   const auto minimum_avoid_length = calcMinAvoidanceLength(nearest_object);
-  const auto minimum_lane_change_length = calcMinimumLaneChangeLength();
+  const auto minimum_lane_change_length = calc_minimum_dist_buffer();
 
   lane_change_debug_.execution_area = createExecutionArea(
     getCommonParam().vehicle_info, getEgoPose(),
@@ -273,12 +273,12 @@ double AvoidanceByLaneChange::calcMinAvoidanceLength(const ObjectData & nearest_
   return avoidance_helper_->getMinAvoidanceDistance(shift_length);
 }
 
-double AvoidanceByLaneChange::calcMinimumLaneChangeLength() const
+double AvoidanceByLaneChange::calc_minimum_dist_buffer() const
 {
-  const auto min_lc_length_and_dist_buffer =
-    utils::lane_change::calculation::calc_min_lc_length_and_dist_buffer(
+  const auto lc_length_and_dist_buffer =
+    utils::lane_change::calculation::calc_lc_length_and_dist_buffer(
       common_data_ptr_, get_current_lanes());
-  return std::get<1>(min_lc_length_and_dist_buffer);
+  return std::get<1>(lc_length_and_dist_buffer).min;
 }
 
 double AvoidanceByLaneChange::calcLateralOffset() const
--- a/planning/behavior_path_planner/autoware_behavior_path_avoidance_by_lane_change_module/src/scene.cpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_avoidance_by_lane_change_module/src/scene.cpp
@@ -275,14 +275,10 @@ double AvoidanceByLaneChange::calcMinAvoidanceLength(const ObjectData & nearest_
 
 double AvoidanceByLaneChange::calcMinimumLaneChangeLength() const
 {
-  const auto current_lanes = get_current_lanes();
-  if (current_lanes.empty()) {
-    RCLCPP_DEBUG(logger_, "no empty lanes");
-    return std::numeric_limits<double>::infinity();
-  }
-
-  return utils::lane_change::calculation::calc_minimum_lane_change_buffer(
-    common_data_ptr_, current_lanes);
+  const auto min_lc_length_and_dist_buffer =
+    utils::lane_change::calculation::calc_min_lc_length_and_dist_buffer(
+      common_data_ptr_, get_current_lanes());
+  return std::get<1>(min_lc_length_and_dist_buffer);
 }
 
 double AvoidanceByLaneChange::calcLateralOffset() const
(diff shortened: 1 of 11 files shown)


planning/behavior_path_planner/autoware_behavior_path_avoidance_by_lane_change_module/src/scene.cpp
```diff
@@ -275,14 +275,10 @@ double AvoidanceByLaneChange::calcMinAvoidanceLength(const ObjectData & nearest_
 
 double AvoidanceByLaneChange::calcMinimumLaneChangeLength() const
```
mkquda, line 276: Maybe better to rename to differentiate between single lc length and entire dist buffer

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/scene.cpp
```diff
@@ -188,17 +239,11 @@ bool NormalLaneChange::is_near_regulatory_element() const
...
-
-  const auto & lc_params = *common_data_ptr_->lc_param_ptr;
   const auto max_prepare_length = calculation::calc_maximum_prepare_length(common_data_ptr_);
-  const auto min_lc_length =
-    calculation::calc_minimum_lane_change_length(lc_params, shift_intervals);
   const auto dist_to_terminal_start =
-    calculation::calc_ego_dist_to_terminal_end(common_data_ptr_) - min_lc_length;
+    calculation::calc_dist_from_pose_to_terminal_end(
+      common_data_ptr_, current_lanes, common_data_ptr_->get_ego_pose()) -
+    common_data_ptr_->transient_data.current_dist_buffer.min;
```
mkquda, line 246: why not directly use `common_data_ptr_->transient_data.dist_from_ego_to_current_terminal_start`

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/scene.cpp
```diff
@@ -1483,7 +1434,9 @@ bool NormalLaneChange::getLaneChangePaths(
       const auto prepare_length = utils::lane_change::calcPhaseLength(
         current_velocity, getCommonParam().max_vel, longitudinal_acc_on_prepare, prepare_duration);
 
-      const auto ego_dist_to_terminal_start = dist_to_end_of_current_lanes - lane_change_buffer;
+      const auto ego_dist_to_terminal_start =
+        common_data_ptr_->transient_data.dist_from_ego_to_current_terminal_end -
+        common_data_ptr_->transient_data.current_dist_buffer.min;
```
mkquda, line 1439: isn't this same as `common_data_ptr_->transient_data.dist_from_ego_to_current_terminal_start`

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/scene.cpp
```diff
@@ -1587,7 +1540,10 @@ bool NormalLaneChange::getLaneChangePaths(
           }
         }
 
-        if (lane_changing_length + prepare_length > dist_to_end_of_current_lanes) {
+        if (
+          lane_changing_length + prepare_length +
+            common_data_ptr_->transient_data.next_lc_buffer.min >
```
mkquda, line 1545: Just curious why did you add the next LC buffer here?

planning/behavior_path_planner/autoware_behavior_path_avoidance_by_lane_change_module/src/scene.cpp
zulfaqar-azmi-t4: fixed in https://github.com/autowarefoundation/autoware.universe/pull/8849/commits/9a0f342290328d85911745afc1e7757a8f8ff7b1

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/scene.cpp
zulfaqar-azmi-t4: I'm making an assumption that we want to do more than 1 lane changes. This is assuming that the end point of current and target lane align with each other. 

Additional note: some additional fix is needed if the end point is not aligned.

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/scene.cpp
zulfaqar-azmi-t4: changed in https://github.com/autowarefoundation/autoware.universe/pull/8849/commits/56c45d8960f7574105c07973019378729f833fab

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/scene.cpp
zulfaqar-azmi-t4: changed in https://github.com/autowarefoundation/autoware.universe/pull/8849/commits/56c45d8960f7574105c07973019378729f833fab

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/scene.cpp
mkquda: but isn't this covered by the check [here](https://github.com/autowarefoundation/autoware.universe/pull/8849/files#diff-59cbf60b192ecddbbd9709016c0b4bd8846eda2ddbc91dc28937762391e13ff0R1548-R1550)

planning/behavior_path_planner/autoware_behavior_path_avoidance_by_lane_change_module/src/scene.cpp
```diff
@@ -273,16 +273,12 @@ double AvoidanceByLaneChange::calcMinAvoidanceLength(const ObjectData & nearest_
...
-    RCLCPP_DEBUG(logger_, "no empty lanes");
-    return std::numeric_limits<double>::infinity();
-  }
-
-  return utils::lane_change::calculation::calc_minimum_lane_change_length(
-    getRouteHandler(), current_lanes.back(), *lane_change_parameters_, direction_);
+  const auto lc_length_and_dist_buffer =
+    utils::lane_change::calculation::calc_lc_length_and_dist_buffer(
+      common_data_ptr_, get_current_lanes());
+  return std::get<1>(lc_length_and_dist_buffer).min;
```
maxime-clem, line 281: (minor) I think you can do something like this which is a bit easier to read.
```suggestion
  const auto [_, dist_buffer] =
    utils::lane_change::calculation::calc_lc_length_and_dist_buffer(
      common_data_ptr_, get_current_lanes());
  return dist_buffer.min;
```

planning/behavior_path_planner/autoware_behavior_path_avoidance_by_lane_change_module/src/scene.cpp
```diff
@@ -273,16 +273,12 @@ double AvoidanceByLaneChange::calcMinAvoidanceLength(const ObjectData & nearest_
   return avoidance_helper_->getMinAvoidanceDistance(shift_length);
 }
 
-double AvoidanceByLaneChange::calcMinimumLaneChangeLength() const
+double AvoidanceByLaneChange::calc_minimum_dist_buffer() const
 {
-  const auto current_lanes = get_current_lanes();
-  if (current_lanes.empty()) {
-    RCLCPP_DEBUG(logger_, "no empty lanes");
-    return std::numeric_limits<double>::infinity();
```
maxime-clem, line 281: If I understand the code correctly, the new return value in case `current_lanes` is empty is `0` compared to `infinity()` previously. Is this okay ?
- the `calc_lc_length_and_dist_buffer()` returns `{}` when `lanes` is empty. 
- `{}` uses the [default constructor of `pair`](https://en.cppreference.com/w/cpp/utility/pair/pair) which uses the default constructor of `Boundary` which sets `min = 0.0`.

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/scene.cpp
```diff
@@ -86,18 +86,82 @@ void NormalLaneChange::update_lanes(const bool is_approved)
...
+  if (
+    !common_data_ptr_ || !common_data_ptr_->is_data_available() ||
+    !common_data_ptr_->is_lanes_available()) {
+    return;
+  }
+
+  auto & transient_data = common_data_ptr_->transient_data;
+  const auto acc_boundary = calcCurrentMinMaxAcceleration();
+  transient_data.acc.min = std::get<0>(acc_boundary);
+  transient_data.acc.max = std::get<1>(acc_boundary);
```
maxime-clem, line 118: I think you can do something like that:
```suggestion
  std::tie(transient_data.acc.min, transient_data.acc.max) = calcCurrentMinMaxAcceleration();
```
or are they the same type ? In which case:
```suggestion
  transient_data = calcCurrentMinMaxAcceleration();
```

planning/behavior_path_planner/autoware_behavior_path_avoidance_by_lane_change_module/src/scene.cpp
maxime-clem: @zulfaqar-azmi-t4 did you check this issue ?

zulfaqar-azmi-t4:
Due to some computational issue, I will set this PR to draft.

zulfaqar-azmi-t4:
close as I will open another PR.

PR #8851, opened 2024-09-11
feat(autoware_mpc_lateral_controller): set steering rate limits and steering limit in predicted trajectory calculation

kyoichi-sugahara:
## Description

There is an issue where the predicted trajectory generated based on the MPC solution would become unstable when the speed decreased. The method of setting the maximum speed to 1 kph can be reproduced by executing the following command.
```
ros2 topic pub /planning/scenario_planning/max_velocity tier4_planning_msgs/msg/VelocityLimit "{stamp: now, max_velocity: 0.2777777910232544, use_constraints: true, constraints: {min_acceleration: -1.0, max_jerk: 1.0, min_jerk: -1.0}, sender: 'api'}" --once
```
visualized topic in the video is `/control/trajectory_follower/lateral/predicted_trajectory`

https://github.com/user-attachments/assets/90177a24-3cca-4b99-af39-27907e5c7795

When examining the values of each trajectory point in function `calculatePredictedTrajectoryInWorldCoordinate` while generating unstable paths, it was found that the steer angle and steer_rate values exceeded the vehicle's constraints. 
By considering these constraints during calculation, it was confirmed that the generation of unstable paths could be prevented as shown in the following video.

https://github.com/user-attachments/assets/fdb1ed63-e1d6-4de1-ada7-5c1edb612c5f

## How was this PR tested?

run planning_simulator and confirmed problem is solved.

- [x] [PASS TIERIV INTERNAL SCENARIOS](https://evaluation.tier4.jp/evaluation/reports/0595e732-1a46-5499-be8e-b56dc210d4cb?project_id=prd_jt)

## Notes for reviewers

I believe it is necessary to improve the following integration method derived from the discretized system of equations:
```
    dstate(0) = velocity * std::cos(yaw);
    dstate(1) = velocity * std::sin(yaw);
    dstate(2) = velocity * std::tan(steer) / m_wheelbase;
    dstate(3) = -(steer - desired_steer) / m_steer_tau;
```

In this case, it is likely that similar issues are occurring in the internal model calculations of the MPC as well. Therefore, I think it is necessary to implement improvements in conjunction with those calculations. I'm planning to investigate and address this in a separate PR.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/control/autoware_mpc_lateral_controller/include/autoware/mpc_lateral_controller/vehicle_model/vehicle_model_interface.hpp
+++ b/control/autoware_mpc_lateral_controller/include/autoware/mpc_lateral_controller/vehicle_model/vehicle_model_interface.hpp
@@ -37,6 +37,8 @@ protected:
   double m_velocity;   //!< @brief vehicle velocity [m/s]
   double m_curvature;  //!< @brief curvature on the linearized point on path
   double m_wheelbase;  //!< @brief wheelbase of the vehicle [m]
+  Eigen::VectorXd
+    m_steer_rate_limits;  //!< @brief steering rate limit along with the trajectory [rad/s]
 
 public:
   /**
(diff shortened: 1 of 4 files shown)


takayuki5168 (approved):

kyoichi-sugahara:
addressed the issue with this [PR](https://github.com/autowarefoundation/autoware.universe/pull/9048)

PR #8852, opened 2024-09-11
fix: change remaning distance when the size of route is 1

shulanbushangshu:
## Description
change  the calculation of  remaning_distance when the size of route is one.
## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


ahmeddesokyebrahim:
Thanks @shulanbushangshu for your PR.
Can you elaborate more what are the problems you faced due to the original implementation and how the new one is fixing them ?

shulanbushangshu:
> Thanks @shulanbushangshu for your PR. Can you elaborate more what are the problems you faced due to the original implementation and how the new one is fixing them ?

When the starting point and end point are in the same path (curvature is not 0), there is an error in using Euclidean distance directly

ahmeddesokyebrahim:
> When the starting point and end point are in the same path (curvature is not 0), there is an error in using Euclidean distance directly

Thanks @shulanbushangshu for your reply.

Unfortunately I am still missing the point of having this change and how it is improving the performance or fixing an existing problem.

Would it be possible that you fill in the section in the PR template of "**How was this PR tested?**" with video demos before this PR & after this PR ? 

I do believe that would help a lot in facilitating the review of this PR.

shulanbushangshu:
there is an error in using Euclidean distance directly
屏幕截图 2024-09-24 100209

mitsudome-r:
I have confirmed that when the vehicle enters the goal lanelet, the PR calculates the remaining distance correctly. 

Originally, it calculated the distance between the ego position and the goal position, but the PR calculates the distance along the lanelet to have more accurate distance until the goal.

mitsudome-r (approved):

PR #8856, opened 2024-09-12
feat(ndt_scan_matcher): visualize matching score each point

SaltUhey:
## Description
This PR visualizes the matching score each point to evaluate it qualitatively as shown in the figure.
The closer the color of the point is to red, the better the score, and the closer it is to blue, the worse the score.

It is important to know how well the map point cloud matches the sensor(LiDAR) point cloud.
Now, Autoware uses  nearest voxel transform probability(NVTL) as a quantitative evaluation of matching score in each frame.

This implementation is expected to facilitate understanding for people.

Screenshot from 2024-09-12 10-51-36
Screenshot from 2024-09-17 13-11-52

## Related links
https://github.com/tier4/ndt_omp/pull/66

## How was this PR tested?
Tests were conducted in some environments in real-time.

## Notes for reviewers
Dependency with ndt_omp.

## Interface changes
Add topic (not shown by default)

#### Additions and removals

| Added | Pub | `/voxel_score_points` | `sensor_msgs::msg::PointCloud2`   | [debug topic] point cloud with scores |

## Effects on system behavior
None.


localization/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp
```diff
@@ -878,6 +895,32 @@ Eigen::Matrix2d NDTScanMatcher::estimate_covariance(
   }
 }
 
+pcl::PointCloud<pcl::PointXYZRGB>::Ptr NDTScanMatcher::visualize_point_score(
+  const pcl::shared_ptr<pcl::PointCloud<PointSource>> & sensor_points_in_map_ptr,
+  const float & lower_nvs, const float & upper_nvs)
+{
+  pcl::PointCloud<pcl::PointXYZI> nvs_points_in_map_ptr_i;
+  nvs_points_in_map_ptr_i =
+    ndt_ptr_->calculateNearestVoxelScoreEachPoint(*sensor_points_in_map_ptr);
```
SakodaShintaro, line 904: I think it's fine to directly assign the value.

localization/ndt_scan_matcher/include/ndt_scan_matcher/ndt_scan_matcher_core.hpp
```diff
@@ -145,6 +145,10 @@ class NDTScanMatcher : public rclcpp::Node
     const pclomp::NdtResult & ndt_result, const Eigen::Matrix4f & initial_pose_matrix,
     const rclcpp::Time & sensor_ros_time);
 
+  pcl::PointCloud<pcl::PointXYZRGB>::Ptr visualize_point_score(
+    const pcl::shared_ptr<pcl::PointCloud<PointSource>> & sensor_points_in_map_ptr,
+    const float & lower_nvs, const float & upper_nvs);
```
SakodaShintaro, line 150: I like that this function returns `pcl::PointCloud<pcl::PointXYZRGB>` instead of `pcl::PointCloud<pcl::PointXYZRGB>::Ptr`.

localization/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp
```diff
@@ -878,6 +895,32 @@ Eigen::Matrix2d NDTScanMatcher::estimate_covariance(
...
+  pcl::PointCloud<pcl::PointXYZRGB>::Ptr nvs_points_in_map_ptr_rgb{
+    new pcl::PointCloud<pcl::PointXYZRGB>};
```
SakodaShintaro, line 906: To avoid multiple memory allocations, I think it would be better to call reserve before the for loop.

localization/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp
```diff
@@ -636,6 +638,21 @@ bool NDTScanMatcher::callback_sensor_points_main(
     *sensor_points_in_baselink_frame, *sensor_points_in_map_ptr, ndt_result.pose);
   publish_point_cloud(sensor_ros_time, param_.frame.map_frame, sensor_points_in_map_ptr);
 
+  // check each of point score
+  const float lower_nvs = 1.0f;
+  const float upper_nvs = 3.5f;
+  if (voxel_score_points_pub_->get_subscription_count() > 0) {
+    pcl::PointCloud<pcl::PointXYZRGB>::Ptr nvs_points_in_map_ptr_rgb{
+      new pcl::PointCloud<pcl::PointXYZRGB>};
```
SakodaShintaro, line 646: Does 'nvs' stand for 'nvtl_score'? If so, it is better to replace it.

SakodaShintaro (commented):
I left some comments. Please either fix them or confirm if they can stay as they are.

SakodaShintaro (approved):
I will address the points I mentioned in a different pull request.
I have confirmed that `logging_simulator` works well the same as before.
@SaltUhey 
Thank you so much!

PR #8858, opened 2024-09-12
fix(autoware_object_merger): default merger priority within enum range

technolojin:
## Description
The default value of the merger priority was out of the enum range.
Fixed the default value to the explained mode `PriorityMode::Confidence = 2`.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


YoshiRi (approved):
LGTM

PR #8862, opened 2024-09-12
fix(autoware_lidar_apollo_instance_segmentation): added existence probability

SamratThapa120:
## Description
The output of apollo gets filtered out in [roi_detected_object_fusion](https://github.com/autowarefoundation/autoware.universe/blob/dabeaa640f25cd7af3ea3333a3fb4533cc64cf5f/perception/autoware_image_projection_based_fusion/src/roi_detected_object_fusion/node.cpp#L60) because the existence_probability is always zero. So, the existence_probability is set as the classification score. 

In addition, the classification score of apollo needed to be normalized using sigmoid function before using it as existence_probability. Ref: https://github.com/ApolloAuto/apollo/blob/2e8ad6fecb323915eeb74efa05cfd1647d6c6138/modules/perception/production/data/perception/lidar/models/cnnseg/velodyne64/deploy.prototxt#L739

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

Tested using several annotated rosbags using [driving_log_replayer](https://github.com/tier4/driving_log_replayer/tree/develop). 

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


yukkysaito (approved):
LGTM

perception/autoware_lidar_apollo_instance_segmentation/src/cluster2d.cpp
```diff
@@ -72,6 +72,11 @@ geometry_msgs::msg::Quaternion getQuaternionFromRPY(const double r, const double
   return tf2::toMsg(q);
 }
 
+double sigmoid(double x)
```
kminoda, line 75:
```suggestion
double sigmoid(const double x)
```

kminoda:
@SamratThapa120 Would you also add the reference to the apollo repository where it mentions about the sigmoid?

SamratThapa120:
@kminoda Thank you. Fixed.

kminoda (approved):

PR #8863, opened 2024-09-12
feat(start_planner): add skip_rear_vehicle_check parameter

kyoichi-sugahara:
## Description

**launcher [PR](https://github.com/autowarefoundation/autoware_launch/pull/1165) should be merged first**

Add the skip_rear_vehicle_check parameter to the start planner module configuration. This parameter allows disabling the rear vehicle check during collision detection. By default, the rear vehicle check is enabled.

## How was this PR tested?

run planning_simulator

default behavior is don't insert stop point from start planner because ego is preventing rear vehicle's progress

```
      skip_rear_vehicle_check: false
```

https://github.com/user-attachments/assets/ced3a866-50a5-4ff4-a4dc-a88bcaa30a3b

by changing paramter to true, don't insert stop point 
```
      skip_rear_vehicle_check: true
```

https://github.com/user-attachments/assets/287daf0d-2ca7-4241-b960-c6f03aeb505a

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


danielsanchezaran (approved):
LGTM

PR #8876, opened 2024-09-13
fix(autoware_behavior_velocity_planner_common): add node clock, fix use sim time

dmoszynski:
## Description
The `behavior_velocity_planner` module has a bug that causes it to work incorrectly when `use_sim_time == True` - it should use the time published on `/clock`, however, it does not do this.
## Related links
--

## How was this PR tested?
I used [tier4 evaluator scenario](https://evaluation.tier4.jp/evaluation/reports/2689eac8-6c54-5595-8e12-681fd1185ca4/tests/c7466edc-cf13-5f76-aa90-dada3a833083/7afa3e07-429e-5739-af3e-2bd350f9c0b7/0d57fbff-8d1a-599c-9e7f-3674ebc7ea4f?project_id=prd_jt) scenario for local testing. The tests included comparison of behavior for the same setup but different values of the `use_sim_time` parameter (either `True` or `False`).

- The result of the scenario is `Passed` for `use_sim_time == False`.

- The result of the scenario is `Failure` for `use_sim_time == True`, as the `Ego`, after stopping before the pedestrian crosswalk (and reaching a linear speed equal to `0.0000`), immediately started moving again by reaching the linear small speed (less than `0.1000`) and stopped again. 

- In the scenario, such behavior is not allowed, so therefore it caused `Failure` and helped detect this issue.

- To get `Passed` I've relaxed the condition that checks the stopped state from `speed<0.0001` to `speed<0.1000`.

After this change, the scenario seemed to run identically for both cases - after Ego stopped before the crosswalk, it gently moved and stopped again (reaching a speed less than `0.1000`, so the conditions were met and got `Passed`). 
- The change in Ego's behavior is related to the `behavior_velocity_crosswalk_module` - which, after the vehicle stops, transitions to `IGNORE` state - this transition does not occur in the case of `use_sim_time == False`.
- Below I've attached the video before the changes:
  - Ego stops when the crosswalk module is in `YIELD` state, 
  - the crosswalk module transitions to `IGNORE` state, 
  - Ego starts again and reaches linear velocity <0.1 (plot),
  - the crosswalk module transitions to `YIELD` state,
  - Ego stops again.

- https://github.com/user-attachments/assets/e623922e-5949-4952-a281-8e907cc5a34e

#### Reason
The reason by which the `behavior_velocity_crosswalk_module` transitions to the `IGNORE` state is a bug in the velocity buffer, when the parameter `use_sim_time == True`.
- To be more precise, the `behavior_velocity_crosswalk_module` behaves as if the velocity buffer is off and the `planner_data_->isVehicleStopped()` call ([here](https://github.com/autowarefoundation/autoware.universe/blob/07a850d04f37fd5422b2204b14069e3018150b12/planning/behavior_velocity_planner/autoware_behavior_velocity_crosswalk_module/src/scene_crosswalk.cpp#L1071-L1072)) returns `True` as immediately as the linear velocity of `Ego` reaches 0.0. 
 - This is because the `TwistStamped` objects in the velocity buffer queue contain timestamps that are consistent with the simulation time (starts at zero) while in the `isVehicleStopped()` method they are compared with the system time (epoch) - [it can be seen here](https://github.com/autowarefoundation/autoware.universe/blob/07a850d04f37fd5422b2204b14069e3018150b12/planning/behavior_velocity_planner/autoware_behavior_velocity_planner_common/include/autoware/behavior_velocity_planner_common/planner_data.hpp#L111-L122).
- This is because the timestamp “now” is acquired using the static method `rclcpp::Clock().now()` instead of using `clock_->now()` - clock that is assigned to the node and takes into account the node parameter `use_sim_time`.  (I realize that in the case of using the static `rclcpp::Clock().now()` method, the time from the topic `/clock` should be used, but this is not happening, even though the publication /clock starts before the start of the module and publishes continuously)

#### Fix
I’ve managed to fix this bug by making sure that module `behavior_velocity_crosswalk_module` uses the clock assigned to the node (takes into account `use_sim_time`) and now timestamps are correctly compared, velocity buffer works as expected. The changes can be seen in this PR. Below I've attached the video after the changes.

https://github.com/user-attachments/assets/b2de4def-e5b2-43a7-91cc-ae5fcb6fee66

## Notes for reviewers

Note: Analysis and tests were performed on the [tier4/v.0.26.1](https://github.com/tier4/autoware.universe/blob/v0.26.1/planning/behavior_velocity_planner_common/include/behavior_velocity_planner_common/planner_data.hpp) version.

## Interface changes

None.

## Effects on system behavior

The module `behavior_velocity_crosswalk_module` uses the clock assigned to the node (takes into account `use_sim_time`).


xmfcx:
@takayuki5168 san could you assign someone to review this? It's a very simple one.

takayuki5168 (approved):
LGTM
Thank you!

takayuki5168:
@rej55 @soblin @shmpwk 
Please approve the PR.

rej55 (approved):

PR #8893, opened 2024-09-17
refactor(universe_utils): remove raw pointers from the triangulation function

mraditya01:
## Description
Refactor the triangulation function to use `std::vector` instead of raw pointer.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?
Random and Unit tested.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/common/autoware_universe_utils/src/geometry/ear_clipping.cpp
+++ b/common/autoware_universe_utils/src/geometry/ear_clipping.cpp
@@ -41,9 +41,9 @@ void perform_triangulation(const Polygon2d & polygon, std::vector<std::size_t> &
 
     std::cout << "Outer ring size: " << len << std::endl;
     auto outer_point = linked_list(outer_ring, true, vertices, points_);
-    std::cout << "Linked list created. Last point index: " << outer_point  << "last_point prev index" << points_[outer_point].prev_index << std::endl;
+    std::cout << "Linked list created. Last point index: " << outer_point  << "last_point prev index" << points_[outer_point].prev_index.value() << std::endl;
 
-    if (outer_point == points_[outer_point].prev_index) {
+    if (outer_point == points_[outer_point].prev_index.value()) {
         std::cerr << "Invalid linked list. Exiting triangulation." << std::endl;
         return;
     }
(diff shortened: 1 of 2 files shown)


common/autoware_universe_utils/src/geometry/ear_clipping.cpp
```diff
@@ -20,311 +20,392 @@
...
+  std::size_t vertices = 0;
   const auto & outer_ring = polygon.outer();
-  len = outer_ring.size();
+  std::size_t len = outer_ring.size();
   points_.reserve(len * 3 / 2);
+
+  if (polygon.outer().empty()) return points_;
+
   indices.reserve(len + outer_ring.size());
+  auto outer_point = linked_list(outer_ring, true, vertices, points_);
```
maxime-clem, line 42: (confusing) `outer_point` is not a point but an index.

common/autoware_universe_utils/src/geometry/ear_clipping.cpp
```diff
@@ -20,311 +20,392 @@
...
+  if (outer_point == points_[outer_point.value()].prev_index.value()) {
```
maxime-clem, line 43: Is it possible that `prev_index` is `nullopt` ? in which case this should be checked.
What about `outer_point` ? If not it should not be a `std::optional`.

common/autoware_universe_utils/src/geometry/ear_clipping.cpp
```diff
@@ -20,311 +20,392 @@
...
+std::size_t construct_point(std::size_t index, const Point2d & point, std::vector<Point> & points)
 {
-  indices.clear();
-  vertices = 0;
-
-  if (polygon.outer().size() == 0) return;
-
-  std::size_t len = 0;
+  points.emplace_back(index, point);
+  return index;
```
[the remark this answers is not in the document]
maxime-clem: I think the `index` is not necessary. If so, please remove it, use `push_back(point)` instead of `emplace_back`, and set the return to either void, or `points.size() - 1`.

common/autoware_universe_utils/src/geometry/ear_clipping.cpp
```diff
@@ -92,271 +247,297 @@ std::optional<std::size_t> linked_list(
...
-    }
+    outer_index = eliminate_hole(q, outer_index, pointsvec);
   }
+
+  return outer_index;
 }
 
 /// @brief check whether ear is valid
-bool is_ear(std::size_t ear_index, const std::vector<Point> & points_vec)
+bool is_ear(const std::size_t ear_index, const std::vector<LinkedPoint> & pointsvec)
```
maxime-clem, line 392: can you rename `pointsvec` to `points` ?

common/autoware_universe_utils/src/geometry/ear_clipping.cpp
```diff
@@ -92,271 +247,297 @@ std::optional<std::size_t> linked_list(
...
+    p_index = pointsvec[p_index].next_index.value();
   }
 
   return true;
 }
 
 /// @brief go through all polygon Points and cure small local self-intersections
 std::size_t cure_local_intersections(
-  std::size_t start_index, std::vector<std::size_t> & indices, std::vector<Point> & points_vec)
+  std::size_t start_index, std::vector<std::size_t> & indices, std::vector<LinkedPoint> & pointsvec)
```
maxime-clem, line 418: can you rename `pointsvec`  to `points` ?

common/autoware_universe_utils/src/geometry/ear_clipping.cpp
```diff
@@ -92,271 +247,297 @@ std::optional<std::size_t> linked_list(
...
   const std::vector<LinearRing2d> & inners, std::size_t outer_index, std::size_t & vertices,
-  std::vector<Point> & points_vec)
+  std::vector<LinkedPoint> & pointsvec)
 {
   const std::size_t len = inners.size();
 
   std::vector<std::size_t> queue;
   for (std::size_t i = 0; i < len; i++) {
-    auto list = linked_list(inners[i], false, vertices, points_vec);
+    auto list = linked_list(inners[i], false, vertices, pointsvec);
```
maxime-clem: (confusing) `list` is not a list

maxime-clem (approved):

soblin (approved):

PR #8933, opened 2024-09-21
feat: operation mode debug panel

isamu-takagi:
## Description

The legacy state panel is used to debug the operation mode.
Since it would be confusing to have a panel with the same function, I extracted that part.
Also, since it will be used as a panel dedicated to debugging, the following changes have been made:
- The mode change button can be used even if `is_mode_*_available` is false.
- The mode change button can be used even if it is the same as the current mode.
- Display raw value when unknown mode

## Related links

- https://github.com/autowarefoundation/autoware.universe/pull/7494#issuecomment-2364792758

## How was this PR tested?

1. Run planning simulation
2. Add OperationModeDebugPanel from rviz menu
3. Change to drive mode and stop mode. 

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


isamu-takagi:
@h-ohta @kosuke55 Could you review this PR?

xmfcx (approved):
image

image

image

It is now clear, it works as it should. I didn't test all functionality but if anyone finds it, we can revert it. Thanks @isamu-takagi -san!

Issue #8940, opened 2024-09-23
Add Direct Initialization Option with GNSS pose


meliketanrikulu:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I've agreed with the maintainers that I can plan this task.

### Description

You need to call service to initialize directly now. [Documentation](https://github.com/autowarefoundation/autoware.universe/tree/main/localization/autoware_pose_initializer#direct-initial-position-set)
 However, if you trust the GNSS pose, you may want to initialize directly with the GNSS pose. For example, we think that the pose coming from GNSS for the simulation environment can be used for initialization directly. We think that the feature of using GNSS poses should be added to activate the direct initialization feature for such cases.

### Purpose

In cases where we trust the GNSS pose (simulation tests etc.), we want to shorten the time spent for pose initlizer and ensure that the direct initilization feature is realized through the subscribed GNSS topic.

### Possible approaches

Adding control of gnss poses when the direct initlizer feature is activated may solve the problem.

### Definition of done

The task is completed when Direct initialization is activated using GNSS poses with parameter changes.

SakodaShintaro:
I believe the desired behavior can be achieved by commenting out the following line:
https://github.com/autowarefoundation/autoware.universe/blob/d25e9a1bbedc176d4e29c70b3ed9586808bdf813/localization/autoware_pose_initializer/src/pose_initializer_core.cpp#L161
How about adding a parameter to the `autoware_pose_initializer` config that allows toggling whether or not to execute the `align_pose`?

Note: Please make sure that `/sensing/gnss/pose_with_covariance`, which `autoware_pose_initializer` subscribes to, outputs the orientation with the appropriate frame ID (map). If the orientation is an invalid value such as (0, 0, 0, 0), a large number of errors will be output by tf.

(cc. @YamatoAndo  Please let me know if there's anything wrong with my understanding.)

SakodaShintaro:
I'm sorry, I missed [this PR](https://github.com/autowarefoundation/autoware.universe/pull/8948).

xmfcx:
Let's discuss here:
[Proposal for Simplifying Pose Initialization in Autoware #5285](https://github.com/orgs/autowarefoundation/discussions/5285)

PR #5285, opened 2023-10-12
feat(map_based_prediction): enable to control lateral path convergence speed by setting control  time horizon

YoshiRi:
## Description

In the current path generation algorithm, lateral displacement is controlled so that the it will arrive center line after the `time_horizon_` = `10sec`. This causes slow lateral motion convergence and lead to inappropriate path outside the lanelet.

I just add one parameter to control this convergence speed. Like figure below.
It simply means to calculate lateral path which converge to centerline by faster time_horizon.

image

Must merge after https://github.com/autowarefoundation/autoware_launch/pull/637.

## Tests performed

Tested with Psim.

See jira ticket [TIER IV INTERNAL LINK](https://tier4.atlassian.net/browse/RT1-2371).

## Notes for reviewers

**Must change `autoware_launch` repo before merging this to main branch**.

## Pre-review checklist for the PR author

The PR author **must** check the checkboxes below when creating the PR.

- [ ] I've confirmed the [contribution guidelines].
- [ ] The PR follows the [pull request guidelines].

## In-review checklist for the PR reviewers

The PR reviewers **must** check the checkboxes below before approval.

- [ ] The PR follows the [pull request guidelines].
- [ ] The PR has been properly tested.
- [ ] The PR has been reviewed by the code owners.

## Post-review checklist for the PR author

The PR author **must** check the checkboxes below before merging.

- [ ] There are no open discussions or they are tracked via tickets.
- [ ] The PR is ready for merge.

After all checkboxes are checked, anyone who has write access can merge the PR.

[contribution guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/
[pull request guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/pull-request-guidelines/

PR #8941, opened 2024-09-23
fix(pose_initializer): add topic subscription to direct initialization

meliketanrikulu:
## Description
You need to call service to initialize directly now. [Documentation](https://github.com/autowarefoundation/autoware.universe/tree/main/localization/autoware_pose_initializer#direct-initial-position-set)
However, if you trust the GNSS pose, you may want to initialize directly with the GNSS pose. For example, we think that the pose coming from GNSS for the simulation environment can be used for initialization directly. We think that the feature of using GNSS poses should be added to activate the direct initialization feature for such cases.

This PR allows you to directly initialize the position from GNSS poses with a parameter change.
## Related links

**Parent Issue:**
https://github.com/autowarefoundation/autoware.universe/issues/8940
- Link

## How was this PR tested?

Test Steps:
### 1. Checkout autoware.universe
```bash
cd ~/autoware/src/universe/autoware.universe/
git remote add autoware.universe https://github.com/meliketanrikulu/autoware.universe.git
git remote update
git checkout fix/add_topic_subscription_to_direct_initialization
```
### 2. Compile updated packages:
```bash
cd ~/autoware
colcon build --symlink-install --packages-select autoware_pose_initializer autoware_default_adapi
```
### 3. Update Parameters
To enable the direct initialization feature using GNSS poses, the parameters must be adjusted accordingly.

> 1. Open the [config/default_adapi.param.yaml](https://github.com/meliketanrikulu/autoware.universe/blob/1f161fa21decb59e9296842113434622953bbd31/system/autoware_default_adapi/config/default_adapi.param.yaml#L12) file.
> 2. Set the `initialization_method` parameter to 1 to activate direct initialization.

### 4. Run autoware
```bash
source ~/autoware/install/setup.bash
```

### 4. Run bag file
```bash
source ~/autoware/install/setup.bash
ros2 bag play ~/autoware_map/sample-rosbag/ -r 0.2 -s sqlite3
```
Instead, you can test PR in any sensor model or simulation environment and observe it with the GNSS pose to see the results.

## Notes for reviewers
What do we expect to see?
If we set initialization method as 1 (DIRECT Initialization method), we expect to see the vehicle initialize at exactly the same point and with the same orientation as the GNSS pose.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/localization/autoware_pose_initializer/README.md
+++ b/localization/autoware_pose_initializer/README.md
@@ -115,6 +115,11 @@ method: 1
 
 The initial position is set directly by the input position without going through localization algorithm.
 
+> [!NOTE]
+> If you want to activate the direct initialization feature with GNSS poses, you must change the parameters. </br>
+> You need to update [config/default_adapi.param.yaml](../../system/autoware_default_adapi/config/default_adapi.param.yaml) </br> 
+> `initialization_method` should be set to 1 for direct initialization. </br>
+
 ### Via ros2 topic pub
 
 ```bash
--- a/localization/autoware_pose_initializer/README.md
+++ b/localization/autoware_pose_initializer/README.md
@@ -117,8 +117,7 @@ The initial position is set directly by the input position without going through
 
 > [!NOTE]
 > If you want to activate the direct initialization feature with GNSS poses, you must change the parameters. </br>
-> You need to update [config/default_adapi.param.yaml](../../system/autoware_default_adapi/config/default_adapi.param.yaml) </br> 
-> `initialization_method` should be set to 1 for direct initialization. </br>
+> You need to update [config/default_adapi.param.yaml](../../system/autoware_default_adapi/config/default_adapi.param.yaml) </br> > `initialization_method` should be set to 1 for direct initialization. </br>
 
 ### Via ros2 topic pub
 
--- a/localization/autoware_pose_initializer/README.md
+++ b/localization/autoware_pose_initializer/README.md
@@ -116,8 +116,9 @@ method: 1
 The initial position is set directly by the input position without going through localization algorithm.
 
 > [!NOTE]
-> If you want to activate the direct initialization feature with GNSS poses, you must change the parameters. </br>
-> You need to update [config/default_adapi.param.yaml](../../system/autoware_default_adapi/config/default_adapi.param.yaml) </br> > `initialization_method` should be set to 1 for direct initialization. </br>
+> To enable the direct initialization feature using GNSS poses, the parameters must be adjusted accordingly.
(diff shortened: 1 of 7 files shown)


YamatoAndo (changes requested):
This makes the ndt_enable and yabloc_enabled parameters meaningless. Please revise the implementation.

meliketanrikulu:
This PR parameter change is based on the addition of DIRECT initialization. However, instead of this, it seems more logical to decide whether the pose will be DIRECT initialized by checking if the covariance of the gnss poses is 0.For this reason I want to close this and open a new PR.
 However, the covariance values ​​of the GNSS poses that come in the autoware_pose_initializer package are automatically changed . This parameter decide the covariance values of GNSS poses --> https://github.com/autowarefoundation/autoware.universe/blob/d25e9a1bbedc176d4e29c70b3ed9586808bdf813/localization/autoware_pose_initializer/config/pose_initializer.param.yaml#L16

 What is the reason for this? Why are default values ​​assigned instead of using the covariance values ​​that come with the poses.
 @isamu-takagi @YamatoAndo

YamatoAndo:
> What is the reason for this? Why are default values ​​assigned instead of using the covariance values ​​that come with the poses.

It's not that we want the covariance of the GNSS, but rather, we want to determine the kind of distribution to use when spreading the particles.

YamatoAndo:
> it seems more logical to decide whether the pose will be DIRECT initialized by checking if the covariance of the gnss poses is 0

I don't think this method is a good approach.

xmfcx:
> > it seems more logical to decide whether the pose will be DIRECT initialized by checking if the covariance of the gnss poses is 0

> I don't think this method is a good approach.

@YamatoAndo -san, Why do you think so?

If all the covariances are 0, or lower than some threshold we define, it means, the pose is very reliable and we can use it for direct initialization.

Overriding the original covariances should be an optional feature, not the default.

The GNSS already provides this information.

xmfcx:
> It's not that we want the covariance of the GNSS, but rather, we want to determine the kind of distribution to use when spreading the particles.

Why shouldn't you distribute it along the covariance ellipse?

YamatoAndo:
> Why shouldn't you distribute it along the covariance ellipse?

In the first place, the position output by the GNSS receiver often does not match the correct position on the point cloud map. While GNSS positioning errors are one factor, it is difficult to perfectly align the point cloud map with GNSS positioning. Despite this, relying on the dispersion output by the GNSS receiver to determine the particle distribution is not a good idea.
Therefore, considering various errors, I believe it would be more robust to distribute particles with a wider spread than the dispersion output by the GNSS receiver.

xmfcx:
Let's discuss here:
[Proposal for Simplifying Pose Initialization in Autoware #5285](https://github.com/orgs/autowarefoundation/discussions/5285)

PR #8943, opened 2024-09-24
fix(autoware_lidar_marker_localization): fix segmentation fault

YamatoAndo:
## Description

To prevent out-of-bounds access errors in the subsequent program when `bin_num` is smaller than `param_.intensity_pattern.size()`, a countermeasure has been implemented.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?
Test with [sample dataset for lidar_marker_localization](https://github.com/autowarefoundation/autoware.universe/tree/main/localization/autoware_landmark_based_localizer/autoware_lidar_marker_localizer#sample-dataset).

Reproduce a situation where bin_num becomes smaller by methods such as narrowing [the range of LiDAR point clouds](https://github.com/autowarefoundation/autoware_launch/blob/main/autoware_launch/config/localization/lidar_marker_localizer/pointcloud_preprocessor/crop_box_filter_measurement_range.param.yaml#L5-L6
).

```
    min_x: -0.1
    max_x: 0.1
```

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


YamatoAndo:
I will test it later and open this PR

PR #8945, opened 2024-09-24
refactor(polar_grid): add autoware prefix to polar_grid

esteve:
## Description

Added `autoware` prefix to `polar_grid`

## Related links

**Parent Issue:**

- https://github.com/autowarefoundation/autoware/issues/5077

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


esteve:
@xmfcx @mitsudome-r @yukkysaito I couldn't find any packages in universe that depend on this package, so I wasn't sure if it'd be best to remove it. But perhaps it has downstream dependencies in other repositories?

xmfcx:
This is used in the default Autoware configuration. Shows the polar grid and the periodic growing circles around the vehicle.

esteve:
@yukkysaito can you please have a look at this PR? Thanks.

xmfcx:
@esteve 

https://github.com/autowarefoundation/autoware_launch/blob/565568d2ea47ad2a0d92d2cb544f27d3b0efc93e/autoware_launch/rviz/autoware.rviz#L233-L246

It is referenced through the plugin class name: `rviz_plugins/PolarGridDisplay`.

So you probably don't need to update it on the RViz config side.

esteve:
@xmfcx thanks for checking, I only saw references to the classes in `autoware_launch`, but not the package or the namespace. So all the changes here are self contained and could be merged, right?

esteve:
@yukkysaito could you have a look at these changes when you have a moment? Thanks.

yukkysaito (approved):

PR #8948, opened 2024-09-24
fix(pose_initializer): switch direct init with gnss cov check

meliketanrikulu:
## Description
Adding switch direct initialization with checking GNSS poses covariance values. This PR checks GNSS covariance values, If your GNSS covariance values are completely Zero, it uses DIRECT initialization method.  
## Related links

**Parent Issue:**
- https://github.com/autowarefoundation/autoware.universe/issues/8940

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/localization/autoware_pose_initializer/src/pose_initializer_core.cpp
+++ b/localization/autoware_pose_initializer/src/pose_initializer_core.cpp
@@ -156,6 +156,13 @@ void PoseInitializer::on_initialize(
 
       auto pose =
         req->pose_with_covariance.empty() ? get_gnss_pose() : req->pose_with_covariance.front();
+      if (is_covariance_zero(pose.pose.covariance)) {
+        RCLCPP_INFO(get_logger(), "Covariance is zero, switching to DIRECT initialization");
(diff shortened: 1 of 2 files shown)


SakodaShintaro:
It’s my first thought.

Instead of switching the behavior by checking whether the covariance of GNSS has only zeros (or small values), how about combining the pose from the GNSS and the pose from the NDT scan matcher, taking into account the covariance of each pose?
If each element of the GNSS covariance matrix is small enough, the combined pose will be almost the same as the GNSS pose.

Theoretically, the combination of two normal distributions can be easily calculated if the two distributions are independent. Of course, since the `align_pose` function of the NDT scan matcher takes the GNSS pose as input, they are not entirely independent.

However, in practice, we can treat them as independent because many particles are evaluated in the `align_pose` function.

xmfcx:
Let's discuss here:
[Proposal for Simplifying Pose Initialization in Autoware #5285](https://github.com/orgs/autowarefoundation/discussions/5285)

PR #8954, opened 2024-09-25
refactor(lane_change): add TransientData to store commonly used lane change-related variables.

zulfaqar-azmi-t4:
## Description

This PR adds TransientData to standardize commonly used values in the lane change module, such as the current lanes' buffer and the distance to the terminal start and end. The goal is to ensure consistency when calculating limits and lengths.

## Related links

- https://github.com/autowarefoundation/autoware.universe/pull/8849 → 8849 has caused some degradation in the Evaluator, so it has been set to draft for the time being. 8954 is a subset of 8849 and has no scenario failures.

## How was this PR tested?

- PSIM
- Evaluator: [TIER IV internal link](https://evaluation.tier4.jp/evaluation/reports/b1ac9431-b48b-5c7d-b6c7-a1ea6f8854e1?project_id=prd_jt)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/include/autoware/behavior_path_lane_change_module/utils/data_structs.hpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/include/autoware/behavior_path_lane_change_module/utils/data_structs.hpp
@@ -27,6 +27,7 @@
 #include <lanelet2_core/primitives/Lanelet.h>
 #include <lanelet2_core/primitives/Polygon.h>
 
+#include <limits>
 #include <memory>
 #include <utility>
 #include <vector>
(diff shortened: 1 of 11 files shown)


planning/behavior_path_planner/autoware_behavior_path_lane_change_module/include/autoware/behavior_path_lane_change_module/utils/data_structs.hpp
```diff
@@ -313,6 +315,25 @@ struct LanesPolygon
...
+};
+
+struct TransientData
+{
+  MinMaxValue acc;
+  MinMaxValue lane_changing_length;
+  MinMaxValue current_dist_buffer;
+  MinMaxValue next_dist_buffer;
+  double dist_from_ego_to_current_terminal_end{std::numeric_limits<double>::min()};
+  double dist_from_ego_to_current_terminal_start{std::numeric_limits<double>::min()};
```
mkquda, line 331: maybe the name can be shorter, maybe `dist_to_terminal_start`, I think unless otherwise specified it is implied that the distance is measured with respect to ego position. can just add a comment in the struct to describe it if someone is confused.

mkquda (approved):
LGTM.

rej55 (approved):

PR #8955, opened 2024-09-25
fix(static_obstacle_avoidance): remove redundant calculation

go-sakayori:
## Description
Inside `fillAvoidanceTargetObjects` function, `filterTargetObjects` and `updateRoadShoulderDistance` are called sequentially.
Both function calls `getRoadShoulderDistance` function, which is relatively time consuming (refer to figure).
The `updateRoadShoulderDistance` function updates the target object's distance to road shoulder (`to_road_shoulder_distance`) and avoidance margin (`avoid_margin`) according to the clip object information.
However, this calculation would be redundant if there is no clip objects.
This PR will resolve the above problem. 

image

## How was this PR tested?
- [x] [TIER IV internal evauator](https://evaluation.tier4.jp/evaluation/reports/7d522c74-6e21-5285-8f7f-a64099582682?project_id=prd_jt)

## Notes for reviewers
None.

## Interface changes
None.

## Effects on system behavior
None.

### DIFF
--- a/planning/behavior_path_planner/autoware_behavior_path_static_obstacle_avoidance_module/src/utils.cpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_static_obstacle_avoidance_module/src/utils.cpp
@@ -1822,6 +1822,9 @@ void updateRoadShoulderDistance(
       clip_objects.push_back(object);
     }
   });
+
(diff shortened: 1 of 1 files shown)


satoshi-ota (approved):
LGTM

PR #8909, opened 2024-09-19
refactor(lane_change): refactor getLaneChangePaths function

mkquda:
## Description

The function `getLaneChangePaths` in LC module is the main part of LC code and is responsible for sampling different candidate paths and verifying the validity and safety of the candidate path. However the function is very big and complex, which makes it difficult to understand and modify.

This PR aims to refactor the `getLaneChangePaths` function to improve code readability and granularity and reduce its complexity.

## Changes
- Add struct `PhaseMetrics` to group relevant measurements (duration, length, velocity, lon. acc, lat. acc) for LC phases
- Add function `calc_prepare_phase_metrics` to compute and return a list of sampled prepare phase metrics
- Add function `calc_shift_phase_metrics` to compute and return a list of sampled lane changing phase metrics
- Remove the necessity of computing target segment, and use reference path from target lane directly instead
- Add function `getCandidatePath` to process LC info and construct the cadidate path
- Add function `checkCandidatePathSafety` to verify LC candidate path

**Flowchart of getLaneChangePaths function after refactoring**
Untitled Diagram

## Related links
None.

## How was this PR tested?
1. Build complete
2. PSIM
3. [Internal TIER IV evaluator](https://evaluation.tier4.jp/evaluation/reports/95c0f44b-6656-5a11-ba20-b642704d4eee?project_id=prd_jt)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

PR #8961, opened 2024-09-26
feat(lane_change): modify lane change target boundary check to consider velocity

mkquda:
## Description
Previously a check was added to ensure the edge of vehicle doesn't exceed target lane boundary during lane change in this [PR](https://github.com/autowarefoundation/autoware.universe/pull/8750)

But in this recent [PR](https://github.com/autowarefoundation/autoware.universe/pull/8909) the check was mistakenly removed.

The purpose of this PR is to add the check again and modify the logic to only apply the boundary check if the lane change velocity is greater than the parameterized minimum lane changing velocity.

## Related links

None.

## How was this PR tested?

- PSIM
- [Internal TIER IV evaluator](https://evaluation.tier4.jp/evaluation/reports/45ea392e-5d35-5fd8-b136-4fe1d4d923c1?project_id=prd_jt)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

During lane change, if the candidate path causes ego footprint to exceed target lane far boundary, and lc velocity is above minimum, the path will be unsafe and will not be approved.


zulfaqar-azmi-t4 (approved):

kosuke55:
@mkquda 
could you please solve cppcheck-differential. maybe it can be solved by just rebase on main

PR #8968, opened 2024-09-26
refactor(static_obstacle_avoidance): move route handler based calculation outside loop

go-sakayori:
## Description
Function `isSatisfiedWithCommonCondition` is called for each object.
However, inside this function there was calculation (specifically route handler based calculation) irrelevant to object information.
In this PR I have replaced some calculation outside the for loop.  

## Related links
None

## How was this PR tested?

- [x] [TIER IV internal link](https://evaluation.tier4.jp/evaluation/reports/5c16651b-e34e-5026-adfc-352b1a202360?project_id=prd_jt)

## Notes for reviewers
None.

## Interface changes
None.

## Effects on system behavior
None.

### DIFF
--- a/planning/behavior_path_planner/autoware_behavior_path_static_obstacle_avoidance_module/src/utils.cpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_static_obstacle_avoidance_module/src/utils.cpp
@@ -751,7 +751,7 @@ bool isObviousAvoidanceTarget(
 
 bool isSatisfiedWithCommonCondition(
   ObjectData & object, const PathWithLaneId & path, const double forward_detection_range,
-  const double to_goal_distance, const Point & ego_pos, const bool & is_allowed_goal_modification,
+  const double to_goal_distance, const Point & ego_pos, const bool is_allowed_goal_modification,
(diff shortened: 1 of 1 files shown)


planning/behavior_path_planner/autoware_behavior_path_static_obstacle_avoidance_module/src/utils.cpp
```diff
@@ -750,8 +750,8 @@ bool isObviousAvoidanceTarget(
 }
 
 bool isSatisfiedWithCommonCondition(
-  ObjectData & object, const AvoidancePlanningData & data, const double forward_detection_range,
-  const std::shared_ptr<const PlannerData> & planner_data,
+  ObjectData & object, const PathWithLaneId & path, const double forward_detection_range,
+  const double to_goal_distance, const Point & ego_pos, const bool & is_allowed_goal_modification,
```
satoshi-ota, line 754: [nits]
```suggestion
  const double to_goal_distance, const Point & ego_pos, const bool is_allowed_goal_modification,
```

satoshi-ota (approved):
LGTM

PR #8811, opened 2024-09-09
feat(autoware_map_based_prediction): improve frenet path generation

technolojin:
## Description
Please see the original PR
https://github.com/autowarefoundation/autoware.universe/pull/8602

* a bug is fixed
The predicted object orientation was wrong.

before

Screenshot from 2024-09-09 16-38-56

after 

Screenshot from 2024-09-09 16-34-44

## Related links

- Link

  - original PR
https://github.com/autowarefoundation/autoware.universe/pull/8602

  - revert PR
https://github.com/autowarefoundation/autoware.universe/pull/8808

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

PR #8973, opened 2024-09-27
fix(autoware_map_based_prediction): adjust lateral duration when object is behind reference path

technolojin:
## Description

The previous update of the `map_based_prediction` had multiple function changes of path generation algorithms.
One of them was `searchProperStartingRefPathIndex`, which finds a proper reference path position to generate natural path.

In case of the `lateral_duration` was too short to reach the reference path point, the path tend to be generated unnaturally.

Planning simulator (PSIM)
condition: simulated car with speed of 1.5 m/s. (which makes the fixed `lateral_duration` short to reach the reference path point.

Before 

Screenshot from 2024-09-27 12-44-17

After

Screenshot from 2024-09-27 12-50-20

## Related links

https://github.com/autowarefoundation/autoware.universe/pull/8602
https://github.com/autowarefoundation/autoware.universe/pull/8811

## How was this PR tested?

Tested in PSIM

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_map_based_prediction/src/path_generator.cpp
+++ b/perception/autoware_map_based_prediction/src/path_generator.cpp
@@ -250,6 +250,15 @@ PredictedPath PathGenerator::generatePolynomialPath(
   terminal_point.d_vel = 0.0;
   terminal_point.d_acc = 0.0;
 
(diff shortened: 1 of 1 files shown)


YoshiRi (approved):
make sense, LGTM

PR #8974, opened 2024-09-27
feat(universe_utils): reduce dependence on Boost.Geometry

mitukou1109:
## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


mitukou1109:
This PR should be merged after #8995. Currently the random tests fail because the concave polygon generator occasionally creates self-intersections.

PR #8976, opened 2024-09-27
fix(autoware_multi_object_tracker): update yaw with range-limited innovation

technolojin:
## Description

The Kalman filter update for yaw should be limited. The innovation should occur to the small side of the angle difference.
The innovation value `fixed_yaw` was calculated but not used.

This PR is to implement the range-limited innovation `fixed_yaw`.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


YoshiRi (approved):
LGTM

PR #8979, opened 2024-09-27
fix(lane_change): fix abort distance enough check

zulfaqar-azmi-t4:
## Description

Before abort path is computed, lane change module check if the remaining distance is sufficient to re-attempt lane change.

Using the following  image as example

image

The total remaining distance should be
```
abort start distance + abort distance + min stopping dist + min dist buffer <= dist to terminal
```

The current computation only

```
abort distance + min dist buffer <= dist to terminal
```
which is incorrect.

This PR aims to fix the incorrect  computation.

## Related links

- [TIER IV internal link](https://star4.slack.com/archives/C04HK9T2E0N/p1727415580962899?thread_ts=1727381189.356109&cid=C04HK9T2E0N)
- [TIER IV Internal link](https://tier4.atlassian.net/browse/RT1-7991)
- [Evaluator](https://evaluation.tier4.jp/evaluation/reports/bf054099-496a-52e7-8291-955797e8b13c?project_id=prd_jt)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


kosuke55 (approved):
LGTM! thanks!!

PR #8995, opened 2024-10-01
fix(autoware_utils): Address self-intersecting polygons in random_concave_generator and handle empty inners() during triangulation

mraditya01:
## Description

This fix two issues in autoware_utils:

- Self-intersecting polygons in random_concave_generator: fix is_valid function to properly validate inserted edges.

- Handling empty inners() during triangulation: ensures that empty inners() are properly handled in the triangulation process.

## How was this PR tested?
- Unit test on self- intersecting polygon and polygon with empty inners().

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/common/autoware_universe_utils/src/geometry/ear_clipping.cpp
+++ b/common/autoware_universe_utils/src/geometry/ear_clipping.cpp
@@ -582,7 +582,12 @@ std::vector<LinkedPoint> perform_triangulation(
   }
 
   if (!polygon.inners().empty()) {
-    outer_point_index = eliminate_holes(polygon.inners(), outer_point_index, vertices, points);
+    const auto & inner_rings = polygon.inners();
+    bool has_non_empty_holes = std::any_of(
+      inner_rings.begin(), inner_rings.end(), [](const auto & ring) { return !ring.empty(); });
+    if (has_non_empty_holes) {
+      outer_point_index = eliminate_holes(inner_rings, outer_point_index, vertices, points);
+    }
   }
 
   ear_clipping_linked(outer_point_index, indices, points);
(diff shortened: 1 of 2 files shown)


common/autoware_universe_utils/src/geometry/ear_clipping.cpp
```diff
@@ -561,6 +564,28 @@ void ear_clipping_linked(
...
+{
+  const auto & points = triangle.outer();
+  double x1 = points[0].x();
+  double y1 = points[0].y();
+  double x2 = points[1].x();
+  double y2 = points[1].y();
+  double x3 = points[2].x();
+  double y3 = points[2].y();
+
+  return std::abs((x1 * (y2 - y3) + x2 * (y3 - y1) + x3 * (y1 - y2)) / 2.0);
```
maxime-clem, line 577: Can you add a source for this formula ?
Or if this function is only needed for the test then just use `boost::geometry::area` in the tests.

common/autoware_universe_utils/src/geometry/ear_clipping.cpp
```diff
@@ -561,6 +564,28 @@ void ear_clipping_linked(
...
+}
+
+double calculate_total_triangle(const std::vector<autoware::universe_utils::Polygon2d> & triangles)
```
maxime-clem, line 580: the name of this function is not clear.
I also think it only needs to be defined in the tests.

mraditya01: Ok, I will move the function to `test` and use `boost::geometry::area` instead

soblin (commented):
I recommend
```
optional<Polygon2d> random_concave_polygon(const size_t vertices, const double max)
```

because Polygon2d()  cannot be a Polygon in any mathematical form. Fundamental issue is that empty constructor for Polygon2d() is defined.

In
```
Polygon2d inward_denting(LinearRing2d & ring)
{
  LinearRing2d convex_ring;
  std::vector<Point2d> q;
  q.reserve(ring.size());
```

if you care about the performance then maybe `q` should be a list instead because `remove_if` for vector causes more overhead than the amount of reduced overhead by calling `reserve` beforehand

soblin (approved):
OK for this change

takayuki5168 (approved):

PR #8996, opened 2024-10-01
fix(autoware_lidar_centerpoint): fix twist covariance orientation

technolojin:
## Description

Fix the twist covariance converter considering the twist covariance matrix is based on the object coordinate.

Before

Screenshot from 2024-10-01 16-29-22

The twist (velocity) estimation of the centerpoint is based on the base_link frame. The covariance is also the same.
In the shown case, a vehicle in front has velocity negative Y direction to the ego vehicle, and the covariance of the velocity is large in Y direction.
However, the twist covariance is directly inserted to Y coordinate of the object message, and it showed large twist covariance in lateral object velocity.

After

Screenshot from 2024-10-01 16-35-32

The twist covariance is translated to the object coordinate, which is by its definition. 
As result, the twist covariance is oriented in Y direction to the ego vehicle, as expected.

NOTE: There is not inference of covariance of XY and YX. The covariance eclipse always oriented in the same direction to the base_link frame. 

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_lidar_centerpoint/include/autoware/lidar_centerpoint/ros_utils.hpp
+++ b/perception/autoware_lidar_centerpoint/include/autoware/lidar_centerpoint/ros_utils.hpp
@@ -35,7 +35,7 @@ uint8_t getSemanticType(const std::string & class_name);
 
 std::array<double, 36> convertPoseCovarianceMatrix(const Box3D & box3d);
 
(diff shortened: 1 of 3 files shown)


knzo25 (approved):
LGTM!

note:

Cov_{object} = R^{T} Cov_{ego} R

PR #9004, opened 2024-10-02
feat(autonomous_emergency_braking): set max imu path length

danielsanchezaran:
## Description
This PR adds a threshold limit to the AEB module's IMU path. 
Reasoning: we want the IMU to represent future positions of the ego vehicle, but, at high speeds, a slight turn of the wheels incurs into a big change on the ego IMU path. By limiting the maximum IMU path length we can still have a good time horizon for the predicted IMU path at low speeds while keeping the IMU path deformation more or less in check. 

Require launch changes: https://github.com/autowarefoundation/autoware_launch/pull/1183
 
## Related links

**Parent Issue:**

- Link

## How was this PR tested?
PSim 

https://github.com/user-attachments/assets/ca8bb17c-02ab-4929-9c46-23752b0d334f

| Change type   |   Topic Name    | Topic Type    | Description        | Default value       |
| change name | min_generated_imu_path_length [m]        | double | minimum distance for a predicted path generated by sensors                                                                                                                                      | 0.5           |
| add | max_generated_imu_path_length [m]        | double | maximum distance for a predicted path generated by sensors| 10.0          |         

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


control/autoware_autonomous_emergency_braking/src/node.cpp
```diff
@@ -648,30 +650,40 @@ Path AEB::generateEgoPath(const double curr_v, const double curr_w)
...
   for (double t = 0.0; t < horizon + epsilon; t += dt) {
     curr_x = curr_x + curr_v * std::cos(curr_yaw) * dt;
     curr_y = curr_y + curr_v * std::sin(curr_yaw) * dt;
     curr_yaw = curr_yaw + curr_w * dt;
     geometry_msgs::msg::Pose current_pose;
     current_pose.position = autoware::universe_utils::createPoint(curr_x, curr_y, 0.0);
     current_pose.orientation = autoware::universe_utils::createQuaternionFromYaw(curr_yaw);
-    if (autoware::universe_utils::calcDistance2d(path.back(), current_pose) < 1e-2) {
+    const double distance_between_points =
+      autoware::universe_utils::calcDistance2d(path.back(), current_pose);
```
tkimura4, line 662: I think this calculation method is fine, but wouldn't using curr_v * dt provide a simpler and more accurate distance calculation?

danielsanchezaran: @tkimura4 you are totally right. I have changed the code to reflect your recommendation.

kyoichi-sugahara:
@danielsanchezaran 
Can you add comment for skipping loop when distance between points are small?

control/autoware_autonomous_emergency_braking/src/node.cpp
danielsanchezaran: @tkimura4 thank you, I have refactored the code to use your suggestion and also delete a lot of repeated code on that function.

danielsanchezaran:
> @danielsanchezaran Can you add comment for skipping loop when distance between points are small?

Done!

kyoichi-sugahara (approved):

PR #9005, opened 2024-10-02
feat(autoware_behavior_path_planner_common): disable feature of turning off blinker at low velocity

kyoichi-sugahara:
## Description

When ego vehicle stops in the lane with turn_direction right or left, blinker turns off because of the following implementation.
```
  const auto requires_turn_signal = [&](const auto & lane_attribute) {
    constexpr double stop_velocity_threshold = 0.1;
    return (
      lane_attribute == "right" || lane_attribute == "left" ||
      (lane_attribute == "straight" && current_vel < stop_velocity_threshold));
  };
```
image

The expected behavior for the scene is the keep turning on when the ego vehicle is on the lane with turn_direction is right or left.

So changed so that keep turning on blinker in this PR.

## Related links

**Parent Issue:**

- [TIERIV INTERNAL TICKET LINK](https://tier4.atlassian.net/browse/RT0-33633)

## How was this PR tested?

- [x] tested with psim
- [ ] passed [TIERIV INTERNAL SCENARIOS](https://evaluation.tier4.jp/evaluation/reports/ff1dbf4c-19e2-5392-b1fe-dbeee2af253a?project_id=prd_jt)

## Notes for reviewers

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_path_planner/autoware_behavior_path_planner_common/src/turn_signal_decider.cpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_planner_common/src/turn_signal_decider.cpp
@@ -138,11 +138,11 @@ std::optional<TurnSignalInfo> TurnSignalDecider::getIntersectionTurnSignalInfo(
   const size_t current_seg_idx, const RouteHandler & route_handler,
   const double nearest_dist_threshold, const double nearest_yaw_threshold)
 {
-  const auto requires_turn_signal = [&](const auto & lane_attribute) {
+  const auto requires_turn_signal = [&](const auto & turn_direction, const bool is_in_turn_lane) {
     constexpr double stop_velocity_threshold = 0.1;
     return (
-      lane_attribute == "right" || lane_attribute == "left" ||
-      (lane_attribute == "straight" && current_vel < stop_velocity_threshold));
+      turn_direction == "right" || turn_direction == "left" ||
+      (turn_direction == "straight" && current_vel < stop_velocity_threshold && !is_in_turn_lane));
   };
   // base search distance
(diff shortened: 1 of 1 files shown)


planning/behavior_path_planner/autoware_behavior_path_planner_common/src/turn_signal_decider.cpp
```diff
@@ -138,11 +138,11 @@ std::optional<TurnSignalInfo> TurnSignalDecider::getIntersectionTurnSignalInfo(
   const size_t current_seg_idx, const RouteHandler & route_handler,
   const double nearest_dist_threshold, const double nearest_yaw_threshold)
 {
-  const auto requires_turn_signal = [&](const auto & lane_attribute) {
+  const auto requires_turn_signal = [&](const auto & turn_direction, const bool is_in_turn_lane) {
```
satoshi-ota, line 141:
```suggestion
  const auto requires_turn_signal = [&current_vel](const auto & turn_direction, const bool is_in_turn_lane) {
```

satoshi-ota (approved):
LGTM

planning/behavior_path_planner/autoware_behavior_path_planner_common/src/turn_signal_decider.cpp
kyoichi-sugahara: applied suggested change in https://github.com/autowarefoundation/autoware.universe/pull/9005/commits/aa288021cd0a9570d7dcd19b70c39c94b06ac4ae
and confirmed it works as expected in psim, thanks!

PR #9016, opened 2024-10-03
feat(build-and-test-differential): modify clang-tidy to run only on specified packages

yhisaki:
## Description

Currently, clang-tidy checks are being executed but are not required and are being ignored. However, clang-tidy outputs important warnings that should not be disregarded.

This PR aims to gradually make clang-tidy checks required on a package-by-package basis. With this PR, the following process related to clang-tidy will be executed on GitHub Actions:

1. [Get modified packages](https://github.com/yhisaki/autoware.universe/blob/97ead9953bb44b9f6a727f1b3630d78f4daef3ab/.github/workflows/build-and-test-differential.yaml#L96): Retrieve the packages that have been modified in the PR.
2. [Get target packages](https://github.com/yhisaki/autoware.universe/blob/97ead9953bb44b9f6a727f1b3630d78f4daef3ab/.github/workflows/build-and-test-differential.yaml#L100): Obtain the intersection of packages listed in [`clang-tidy-required-package.yaml`](https://github.com/yhisaki/autoware.universe/blob/clang-tidy-specific-packages/clang-tidy-required-packages.yaml) and those retrieved in step 1.
3. Run [autoware-clang-tidy](https://github.com/yhisaki/autoware-clang-tidy) on the packages obtained in step 2.

By progressively adding clang-tidy-compliant packages to [`clang-tidy-required-package.yaml`](https://github.com/yhisaki/autoware.universe/blob/clang-tidy-specific-packages/clang-tidy-required-packages.yaml) , we can achieve gradual improvement in code quality. 

This PR is related to [this PR](https://github.com/autowarefoundation/autoware-github-actions/pull/322).

## Related links

- https://github.com/orgs/autowarefoundation/discussions/4876

## How was this PR tested?

Tested on CI.

- [succeeded CI](https://github.com/autowarefoundation/autoware.universe/actions/runs/11230032538/job/31218645764?pr=9016)
- [failed CI](https://github.com/autowarefoundation/autoware.universe/actions/runs/11231089990/job/31219788707?pr=9016)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


clang-tidy-required-packages.yaml
```diff
@@ -0,0 +1,223 @@
+##### common #####
```
SakodaShintaro, line 1: I felt that it might be a little difficult to manage this file when the package name changes or when packages are deleted/added.

map/map_loader/include/map_loader/lanelet2_map_loader_node.hpp
```diff
@@ -28,6 +28,7 @@
 #include <memory>
 #include <string>
 
+// Hi! I'm a comment!
```
SakodaShintaro, line 31: Was this line added just for testing?

clang-tidy-required-packages.yaml
yhisaki: I agree.
However, this file is temporary. In the near future, all packages will pass the clang-tidy check and this file will not be needed. Ideally, we would like to make all packages clang-tidy check and make this check required right now, but that is unrealistic. This file setting is a temporary alternative and will be deleted.

YamatoAndo:
@yhisaki hi, I can't review this PR, so please ask someone else to review it.

clang-tidy-required-packages.yaml
xmfcx, line 1: About this file as a whole, what about removing the commented out packages?

This way the file will be cleaner. Right now it is hard to see the "not commented out" packages.

We can just keep adding new packages as they come through.

.github/workflows/build-and-test-differential.yaml
```diff
@@ -97,20 +97,28 @@ jobs:
...
 
       - name: Run clang-tidy
-        if: ${{ steps.get-changed-files.outputs.changed-files != '' }}
-        uses: autowarefoundation/autoware-github-actions/clang-tidy@v1
+        if: ${{ steps.get-target-packages.outputs.target-packages != '' }}
+        uses: yhisaki/autoware-github-actions/clang-tidy@use-ament_clang_tidy
         with:
           rosdistro: humble
-          target-packages: ${{ steps.get-modified-packages.outputs.modified-packages }}
-          target-files: ${{ steps.get-changed-files.outputs.changed-files }}
```
xmfcx, line 112: Why transition from running it solely on the changed files to applying it across the entire modified packages?

clang-tidy-required-packages.yaml
xmfcx: This way, the prefixing job won't affect it that much.

veqcc: Agreed :+1:

.github/workflows/build-and-test-differential.yaml
veqcc: One reason is that clang-tidy does not work on a `.hpp` file.
In other words, `clang-tidy -p build/ hoge.hpp` will omit a lot of false positive warnings due to the lack of compilation infomation, and thus those .hpp files should be analyzed through .cpp files.

For example, you can compare the following two runs:

```
$ clang-tidy-14 --use-color -header-filter=src/universe/autoware.universe/.* -p ./build/ src/universe/autoware.universe/planning/sampling_based_planner/autoware_sampler_common/include/autoware_sampler_common/constraints/ma
p_constraints.hpp
```

```
$ export FILES=($(find src/universe/autoware.universe/planning/sampling_based_planner/autoware_sampler_common/ -name "*.cpp" -or -name "*.hpp"))
veqcc@dpc2206003:~/work/autoware$ run-clang-tidy -j 16 -p build/ "${FILES[@]}"

Comment (mojomex) [reply .github/workflows/build-and-test-differential.yaml:112]: Filtering the warnings could be done after clang-tidy ran, i.e. run on all files of the modified package, and then only report the errors/warnings that are in the modified files.

Comment (veqcc) [reply .github/workflows/build-and-test-differential.yaml:112]: @mojomex 
Thanks, I underdstand it :+1: 

The problem is that, when `changed-files` contains `.hpp` files, then clang-tidy should be run on `.cpp` files who include those hpp files.
To realize it, we want to analyze all the files in the target package. That is one readon why we want to introduce the `clang-tidy-required-packages.yaml` filtering.

Comment (mojomex) [reply .github/workflows/build-and-test-differential.yaml:112]: @veqcc Ah, sorry, I expressed my idea a bit loosely!

I understand that Clang-Tidy has to run on all files, including (possibly unmodified) `.cpp` files.
My proposal would be to

1. run on all files of the _modified package_ (or, if it needs to run on all of Autoware, that is also fine I guess)
2. Filter the warnings/errors from Clang-Tidy and only report those in the _modified files_

This could be done via the `--line-filter` argument of clang-tidy and passing the _modified files_ to it:
```
  --line-filter=<string>         - List of files with line ranges to filter the
                                   warnings. Can be used together with
                                   -header-filter. The format of the list is a
                                   JSON array of objects:
                                     [
                                       {"name":"file1.cpp","lines":[[1,3],[5,7]]},
                                       {"name":"file2.h"}
                                     ]
```

I'll have to try that when I have time :slightly_smiling_face:

Comment (veqcc) [reply .github/workflows/build-and-test-differential.yaml:112]: @mojomex 
> Filter the warnings/errors from Clang-Tidy and only report those in the modified files

I agree to the idea that it's better to refine errors/warnings to the modified files.

However, I am not sure whether the changes in `.hpp` files never invoke a new error/warning in `.cpp` files who include it.
In case of it, I want to include all the errors/warnings in CI results.

Comment (xmfcx) [reply .github/workflows/build-and-test-differential.yaml:112]: It makes sense, thanks for explaining!

File: clang-tidy-required-packages.yaml (hunk 1 above)
Comment (xmfcx) [reply clang-tidy-required-packages.yaml:1]: Ok thanks, I'll approve once the commented out packages are removed from this file :+1:

Comment (veqcc) [reply clang-tidy-required-packages.yaml:1]: @yhisaki 
Could you update clang-tidy-required-packages.yaml file?

**Comment by veqcc** [comment]
@xmfcx 
I have one question on clang-tidy CI: Is there any other repository which uses `autowarefoundation/autoware-github-actions/clang-tidy`?
If not, what about changing `autowarefoundation/autoware-github-actions/clang-tidy` as follows:
 - delete `target-files` interface and use only `target-packages`
 - use `run-clang-tidy` instead of `clang-tidy`, which is more suitable for a big project

**Comment by xmfcx** [comment]
You can find all the references here:
- https://github.com/search?q=org%3Aautowarefoundation%20autowarefoundation%2Fautoware-github-actions%2Fclang-tidy&type=code

I'm not sure about usages outside our organization.

You are referring to https://github.com/lmapii/run-clang-tidy right?

Could you create a PR to https://github.com/autowarefoundation/autoware-github-actions create a new version for the clang tidy for this.

And it would be a breaking change, we will need to update the major version for the entire actions in the repository since they are not versioned individually.

Once the major version is incremented, any other further references to any of the actions from the repository will need to be incremented if we make a change to them. This is because we refer to them with `@v1` everywhere. Example: https://github.com/autowarefoundation/autoware.universe/blob/d8087f40fdab6d8e289f3374b621aa5228a1e45a/.github/actions/build-and-test-differential/action.yaml#L43

**Comment by xmfcx** [comment]
It might be a good idea to divide https://github.com/autowarefoundation/autoware-github-actions into multiple repositories.

At least, whenever a major change is required for an action, we can put that particular action into its own repository and version it independently.

What do you think @mitsudome-r @HansRobo ?

I think the costs for the new repo are:
- repository settings
  - general settings
  - rulesets
    - status checks will refer to actions below
  - team management
- actions
  - release action?
  - precommit
  - test-composite-actions
  - new pr action
  - these can be automated with a sync action

**Comment by veqcc** [comment]
> You can find all the references here:
https://github.com/search?q=org%3Aautowarefoundation%20autowarefoundation%2Fautoware-github-actions%2Fclang-tidy&type=code

Thank you for your infomation, and I understand it is a breaking change and needs versioning.
Our policy on making clang-tidy a required CI is a mimimum and rapid start, so if it will take some time, we will use the v1 clang-tidy workflow 👍 


> You are referring to https://github.com/lmapii/run-clang-tidy right?
Could you create a PR to https://github.com/autowarefoundation/autoware-github-actions create a new version for the clang tidy for this.

No, perhaps. `run-clang-tidy` is a part of llvm-project:
https://github.com/llvm/llvm-project/blob/main/clang-tools-extra/clang-tidy/tool/run-clang-tidy.py
I will make a PR later!

**Comment by HansRobo** [comment]
> It might be a good idea to divide [autowarefoundation/autoware-github-actions](https://github.com/autowarefoundation/autoware-github-actions) into multiple repositories.
> 
> At least, whenever a major change is required for an action, we can put that particular action into its own repository and version > it independently.

I agree.
It's something we'll need to do eventually as we grow, so if this is the opportunity, we should go ahead with it.

There will be maintenance costs, but I can help with some of that, and most of it can be automated.
However, a lot of the start-up work will have to be done by the AWF maintainers.

By dividing the repository, the simplicity of permission management may be lost, but this can be managed by creating a team of GitHub Actions maintainers on AWF GitHub organization and linking the divided repositories to that team.

(In line with this, `awf/autoware-spell-check-dict` can also be linked to this team for management.)

**Comment by mitsudome-r** [comment]
> It might be a good idea to divide https://github.com/autowarefoundation/autoware-github-actions into multiple repositories.

> At least, whenever a major change is required for an action, we can put that particular action into its own repository and version it independently.

That sounds like a good idea to me.

File: clang-tidy-required-packages.yaml (hunk 1 above)
Comment (veqcc) [reply clang-tidy-required-packages.yaml:1]: @yhisaki 
I have create a new, minimum `.clang-tidy-ci` configuration, so please use it and list up the packages which pass the minimum check.
https://github.com/autowarefoundation/autoware/pull/5465

**Comment by yhisaki** [comment]
While this PR specifies which ROS2 packages to tidy, I want to significantly rewrite it to specify which files should NOT be tidied.

In my plan, I would create a file like `.clang-tidy-ignore` and specify files to ignore in the following format:
```
path/to/ignore1/*
path/to/ignore2/ignored.cpp
```
The reasons for this change are:
1. This approach is common and used in various projects (e.g. [apache/arrow](https://github.com/apache/arrow), [zeek/spicy](https://github.com/zeek/spicy/), [andreasfertig/cppinsights](https://github.com/andreasfertig/cppinsights))
2. This is a more versatile approach.

Let me know if you have any thoughts or opinions.

xmfcx:
@yhisaki it makes sense, let's do it as you've suggested 👍

PR #9019, opened 2024-10-03
feat(autoware_multi_object_tracker): set maximum reverse velocity to bicycle and crtv motion models

technolojin:
## Description

Revert the tracker orientation when the velocity exceed the maximum reverse velocity

The tracking object orientation is determined by the initial object orientation.
If the initial object yaw was reversed (180 deg rotated), and the object moves forward, the object is tracked as reversed direction.
This mismatch can cause yaw instability in case of the bicycle motion model.

- Before

Screenshot from 2024-10-03 18-38-10

The screenshot shows trackedObjects. Triangles of the bounding box corners indicates front of the object. 
The right object heads left upper side but its velocity is right bottom side. This shows the object is reversed motion. 
If the detector has high confidence of its headings, this situation can be trusted. But, the current lidar-based detection has high error rate on its headings.

Considering the current Autoware implementation (the map_based_prediction fixes the object to head its velocity direction anyway), fixing the orientation with a certain threshold is feasible. 

- After

Screenshot from 2024-10-03 18-47-57

If the object is reversed and it reaches a threshold ( -5 kmph by default), the tracked object will be flipped and heads forward to its velocity vector. 

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/tracker/motion_model/ctrv_motion_model.hpp
+++ b/perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/tracker/motion_model/ctrv_motion_model.hpp
@@ -45,13 +45,14 @@ private:
   // motion parameters: process noise and motion limits
   struct MotionParams
   {
-    double q_cov_x = 0.025;     // [m^2/s^2] uncertain position in x, 0.5m/s
-    double q_cov_y = 0.025;     // [m^2/s^2] uncertain position in y, 0.5m/s
-    double q_cov_yaw = 0.1218;  // [rad^2/s^2] uncertain yaw angle, 20deg/s
-    double q_cov_vel = 8.6436;  // [m^2/s^4] uncertain velocity, 0.3G m/s^2
-    double q_cov_wz = 0.5236;   // [rad^2/s^4] uncertain yaw rate, 30deg/s^2
-    double max_vel = 2.78;      // [m/s] maximum velocity
-    double max_wz = 0.5236;     // [rad/s] maximum yaw rate, 30deg/s
+    double q_cov_x = 0.025;          // [m^2/s^2] uncertain position in x, 0.5m/s
+    double q_cov_y = 0.025;          // [m^2/s^2] uncertain position in y, 0.5m/s
+    double q_cov_yaw = 0.1218;       // [rad^2/s^2] uncertain yaw angle, 20deg/s
+    double q_cov_vel = 8.6436;       // [m^2/s^4] uncertain velocity, 0.3G m/s^2
+    double q_cov_wz = 0.5236;        // [rad^2/s^4] uncertain yaw rate, 30deg/s^2
+    double max_vel = 2.78;           // [m/s] maximum velocity
+    double max_wz = 0.5236;          // [rad/s] maximum yaw rate, 30deg/s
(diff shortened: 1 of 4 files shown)


yukkysaito (approved):

perception/autoware_multi_object_tracker/lib/tracker/motion_model/ctrv_motion_model.cpp
```diff
@@ -203,15 +203,26 @@ bool CTRVMotionModel::limitStates()
   Eigen::MatrixXd P_t(DIM, DIM);
   ekf_.getX(X_t);
   ekf_.getP(P_t);
-  X_t(IDX::YAW) = autoware::universe_utils::normalizeRadian(X_t(IDX::YAW));
+
+  // maximum reverse velocity
+  if (X_t(IDX::VEL) < 0 && X_t(IDX::VEL) < motion_params_.max_reverse_vel) {
```
YoshiRi, line 208: Since max_reverse_vel is already negative, do we need to check `X_t(IDX::VEL) < 0` ?

perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/tracker/motion_model/ctrv_motion_model.hpp
```diff
@@ -45,13 +45,14 @@ class CTRVMotionModel : public MotionModel
...
-    double max_vel = 2.78;      // [m/s] maximum velocity
-    double max_wz = 0.5236;     // [rad/s] maximum yaw rate, 30deg/s
+    double q_cov_x = 0.025;          // [m^2/s^2] uncertain position in x, 0.5m/s
+    double q_cov_y = 0.025;          // [m^2/s^2] uncertain position in y, 0.5m/s
+    double q_cov_yaw = 0.1218;       // [rad^2/s^2] uncertain yaw angle, 20deg/s
+    double q_cov_vel = 8.6436;       // [m^2/s^4] uncertain velocity, 0.3G m/s^2
+    double q_cov_wz = 0.5236;        // [rad^2/s^4] uncertain yaw rate, 30deg/s^2
+    double max_vel = 2.78;           // [m/s] maximum velocity
+    double max_wz = 0.5236;          // [rad/s] maximum yaw rate, 30deg/s
+    double max_reverse_vel = -1.38;  // [m/s] maximum reverse velocity, -5km/h
```
YoshiRi, line 55: [suggestion for the future]
This value should be negative. If possible add some comment or add validation  of this struct.

YoshiRi (approved):
LGTM with few minor comments

perception/autoware_multi_object_tracker/lib/tracker/motion_model/ctrv_motion_model.cpp
technolojin: it is fixed to check the parameter is negative and the velocity is smaller than the parameter

perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/tracker/motion_model/ctrv_motion_model.hpp
technolojin: thank you for your comment. I added a comment to the parameter explanation.

PR #9032, opened 2024-10-04
docs(coverage): display component wise coverage matrics

soblin:
## Description

display coverage status
image

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


README.md
```diff
@@ -16,3 +14,37 @@ To dive into the vast world of Autoware and understand how Autoware Universe fit
...
+| Control      | [![codecov](https://img.shields.io/badge/dynamic/json?url=https://codecov.io/api/v2/github/autowarefoundation/repos/autoware.universe/components&label=Control%20Packages&query=$.[1].coverage)](https://app.codecov.io/gh/autowarefoundation/autoware.universe?components%5B0%5D=Control%20Packages)           |
+| Evaluator    | [![codecov](https://img.shields.io/badge/dynamic/json?url=https://codecov.io/api/v2/github/autowarefoundation/repos/autoware.universe/components&label=Evaluator%20Packages&query=$.[2].coverage)](https://app.codecov.io/gh/autowarefoundation/autoware.universe?components%5B0%5D=Evaluator%20Packages)       |
+| Launch       | [![codecov](https://img.shields.io/badge/dynamic/json?url=https://codecov.io/api/v2/github/autowarefoundation/repos/autoware.universe/components&label=Launch%20Packages&query=$.[3].coverage)](https://app.codecov.io/gh/autowarefoundation/autoware.universe?components%5B0%5D=Launch%20Packages)             |
```
mitsudome-r, line 35: It looks like Launch is not showing any percentage maybe because the launch packages only contains XML without any test codes. Do you think it still make sense to include this in the table? I though we could add it when we think that there is a need to show the coverage.
image

soblin: I think launch is worth tracking the coverage because we should check if specified modules are running given the arguments.

mitsudome-r (approved):

PR #9044, opened 2024-10-07
refactor(lane_change): refactor get_lane_change_lanes function

zulfaqar-azmi-t4:
## Description

Refactor get_lane_change_lanes function.
Two utilities functions are added;

1. to_geom_msg_pose which will create pose from lanelet points
2. is_mandatory_lane_change which list the lane change type for mandatory lane change

## Related links

**Parent Issue:**

- Link

Compile autoware and run psim

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/include/autoware/behavior_path_lane_change_module/utils/utils.hpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/include/autoware/behavior_path_lane_change_module/utils/utils.hpp
@@ -51,12 +51,15 @@ using autoware_perception_msgs::msg::PredictedObjects;
 using autoware_perception_msgs::msg::PredictedPath;
 using behavior_path_planner::lane_change::CommonDataPtr;
 using behavior_path_planner::lane_change::LanesPolygon;
+using behavior_path_planner::lane_change::ModuleType;
 using behavior_path_planner::lane_change::PathSafetyStatus;
 using geometry_msgs::msg::Point;
 using geometry_msgs::msg::Pose;
 using geometry_msgs::msg::Twist;
 using path_safety_checker::CollisionCheckDebugMap;
 using tier4_planning_msgs::msg::PathWithLaneId;
+
+bool is_mandatory_lane_change(const ModuleType lc_type);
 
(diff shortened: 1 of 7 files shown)


planning/behavior_path_planner/autoware_behavior_path_lane_change_module/include/autoware/behavior_path_lane_change_module/utils/utils.hpp
```diff
@@ -118,9 +121,9 @@ double getLateralShift(const LaneChangePath & path);
 
 CandidateOutput assignToCandidate(
   const LaneChangePath & lane_change_path, const Point & ego_position);
-std::optional<lanelet::ConstLanelet> getLaneChangeTargetLane(
-  const RouteHandler & route_handler, const lanelet::ConstLanelets & current_lanes,
-  const LaneChangeModuleType type, const Direction & direction);
+
+std::optional<lanelet::ConstLanelet> get_lane_change_target_lane(
+  const CommonDataPtr & common_data_ptr, const lanelet::ConstLanelets & current_lanes);
```
maxime-clem, line 126: Is there any reason for using a pointer here instead of directly passing `CommonData` by reference ?

maxime-clem (approved):

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/include/autoware/behavior_path_lane_change_module/utils/utils.hpp
zulfaqar-azmi-t4: @maxime-clem 
I would like to standardize the input to avoid mixing argument types within the lane change functions. For example, some functions use `CommonData`, while others use `CommonDataPtr`.

Are there any concerns with this implementation?

zulfaqar-azmi-t4: Note: discussion with @maxime-clem 
In this function, there's no reason to use pointer as input argument, since we didn't even check for any null values.

So for standardization, we will make another PR to refactor the usage of CommonDataPtr

PR #9052, opened 2024-10-08
feat(autoware_lidar _ransfusion): fix 3d bounding box orientation

SamratThapa120:
## Description
We found that the 3d bounding boxes were rotated by 90 degrees for transfusion
image

This PR fixes the issue.

### DIFF
--- a/perception/autoware_lidar_transfusion/lib/ros_utils.cpp
+++ b/perception/autoware_lidar_transfusion/lib/ros_utils.cpp
@@ -28,7 +28,6 @@ void box3DToDetectedObject(
   autoware_perception_msgs::msg::DetectedObject & obj)
 {
   obj.existence_probability = box3d.score;
-
   // classification
   autoware_perception_msgs::msg::ObjectClassification classification;
   classification.probability = 1.0f;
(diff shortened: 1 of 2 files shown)


perception/autoware_lidar_transfusion/lib/ros_utils.cpp
```diff
@@ -49,14 +49,14 @@ void box3DToDetectedObject(
 
   // pose and shape
   // mmdet3d yaw format to ros yaw format
-  float yaw = box3d.yaw + autoware::universe_utils::pi / 2;
+  float yaw = box3d.yaw;
   obj.kinematics.pose_with_covariance.pose.position =
```
amadeuszsz, line 53: If we don't need any format conversion, please remove this line and upper comment as well (line 51). We can use `box3d.yaw` directly.

perception/autoware_lidar_transfusion/lib/ros_utils.cpp
```diff
@@ -49,14 +49,14 @@ void box3DToDetectedObject(
...
     autoware::universe_utils::createPoint(box3d.x, box3d.y, box3d.z);
   obj.kinematics.pose_with_covariance.pose.orientation =
     autoware::universe_utils::createQuaternionFromYaw(yaw);
   obj.shape.type = autoware_perception_msgs::msg::Shape::BOUNDING_BOX;
   obj.shape.dimensions =
-    autoware::universe_utils::createTranslation(box3d.length, box3d.width, box3d.height);
+    autoware::universe_utils::createTranslation(box3d.width, box3d.length, box3d.height);
```
amadeuszsz, line 59: Changing order does not match with `createTranslation` funciton call (x, y, z). If we need to do a switch here, it should be done in [kernel](https://github.com/autowarefoundation/autoware.universe/blob/c2438fd2841bce7792be4bc57e88edbe6b34276d/perception/autoware_lidar_transfusion/lib/postprocess/postprocess_kernel.cu#L86-L87). However, the operation in kernel match indices in forward propagation ([TIER IV internal link](https://github.com/tier4/autoware-ml/blob/bf604baf6317d76deb08e78a5327615e8f27ccd0/projects/TransFusion/transfusion/transfusion_head.py#L683-L684)). Could you check training pipeline if there is a mistake?

amadeuszsz (changes requested):
Nice catch! I shared my comments

perception/autoware_lidar_transfusion/lib/ros_utils.cpp
SamratThapa120: FIXED

amadeuszsz (approved):
LGTM. Thanks for fix! 🚀

kminoda:
@SamratThapa120 Would you follow the PR format as others do? Also, if you've run the evalation on Evaluator, please post the link as well.

SamratThapa120:
@SamratThapa120 Thank you for the information. I updated the PR description. I havent yet run the evaluation on Evaluator, I will post the link once it is done.

PR #9057, opened 2024-10-09
fix(autoware_lidar_transfusion): set tensor names by matching with predefined values.

SamratThapa120:
# Description
When making minor architectural changes to onnx checkpoints, we found that the tensorrt IO tensor order changed, which results in the current code failing because the order of tensors is fixed in current code.
https://github.com/autowarefoundation/autoware.universe/blob/c2438fd2841bce7792be4bc57e88edbe6b34276d/perception/autoware_lidar_transfusion/include/autoware/lidar_transfusion/utils.hpp#L39

It seems that the tensor output order of the optimized tensorrt engine can be different from onnx IO order, but the tensor names match the onnx IO names. So, it is better to use the names from trt engine API to find the IO tensor positions. 

We also found out that the current deployed models have faulty output names i.e the names of dir_cls_pred0 and bbox_pred0 are swapped, so they do not work with this PR. Therefore, we had to update the deployed onnx models with this PR. 

Merge with https://github.com/autowarefoundation/autoware/pull/5318

### DIFF
--- a/perception/autoware_lidar_transfusion/lib/network/network_trt.cpp
+++ b/perception/autoware_lidar_transfusion/lib/network/network_trt.cpp
@@ -23,7 +23,7 @@
 namespace autoware::lidar_transfusion
 {
 
-inline NetworkIO nameToNetworkIO(const char* name)
+inline NetworkIO nameToNetworkIO(const char * name)
 {
   static const std::unordered_map<std::string_view, NetworkIO> name_to_enum = {
     {"voxels", NetworkIO::voxels},
@@ -31,8 +31,7 @@ inline NetworkIO nameToNetworkIO(const char* name)
     {"coors", NetworkIO::coors},
     {"cls_score0", NetworkIO::cls_score},
     {"dir_cls_pred0", NetworkIO::dir_pred},
-    {"bbox_pred0", NetworkIO::bbox_pred}
-  };
+    {"bbox_pred0", NetworkIO::bbox_pred}};
 
   auto it = name_to_enum.find(name);
   if (it != name_to_enum.end()) {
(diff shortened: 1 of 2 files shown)


perception/autoware_lidar_transfusion/lib/network/network_trt.cpp
```diff
@@ -23,6 +23,23 @@
...
 
+inline NetworkIO nameToNetworkIO(const char * name)
+{
+  static const std::unordered_map<std::string_view, NetworkIO> name_to_enum = {
+    {"voxels", NetworkIO::voxels},
+    {"num_points", NetworkIO::num_points},
+    {"coors", NetworkIO::coors},
+    {"cls_score0", NetworkIO::cls_score},
+    {"dir_cls_pred0", NetworkIO::dir_pred},
+    {"bbox_pred0", NetworkIO::bbox_pred}};
```
amadeuszsz, line 34: Please, switch `dir_cls_pred0` with `bbox_pred0` in line 33 & 34 and order of [enum](https://github.com/autowarefoundation/autoware.universe/blob/f256c0bed26e8b26b1eab39b8085a5c6c0be3722/perception/autoware_lidar_transfusion/include/autoware/lidar_transfusion/utils.hpp#L39) as well.
This change is only for readability purposes, I want to stick our enum with [original implementation](https://github.com/open-mmlab/mmdeploy/blob/main/mmdeploy/codebase/mmdet3d/models/mvx_two_stage.py#L99).

amadeuszsz (changes requested):
Thanks for PR! Only one minor change requested.

perception/autoware_lidar_transfusion/lib/network/network_trt.cpp
SamratThapa120: agreed. Fixed.

amadeuszsz: Style: there is one unnecessary comma at the end after your change, could you remove it and bring back one enum per line? Maybe that is why pre-commit did this weird style fix.

SamratThapa120: I removed the comma, but pre-commit still uses the wierd style fix

amadeuszsz (approved):
LGTM! Please, merge this PR with model update PR at the same time.

PR #9089, opened 2024-10-16
fix(avoidance): don't insert stop line if the ego can't avoid or return

satoshi-ota:
## Description

Sometimes, the avoidance module inserted stop point even when the ego overran the dead line to avoid/return manuever.
In this PR, I fixed logic not to insert stop point if the ego exceeds stop  line in order to prevent stuck.

Additionally, I fixed not to insert stop point on dead line for return maneuver if it's in following condition:

- there is no new shift line.
- return shift line is NOT feasible.

Screenshot from 2024-10-16 09-35-57

## Related links

**Parent Issue:**

- ticket: [TIER IV INTERNAL LINK](https://tier4.atlassian.net/browse/RT0-33610)

## How was this PR tested?

https://github.com/user-attachments/assets/5130c5e6-9eb8-4812-9d3a-7e6f5df7381c

- [x] [PASS TIER IV INTERNAL SCENARIOS](https://evaluation.tier4.jp/evaluation/reports/deebc6b7-a130-50ab-b2c8-bda253bff92e?project_id=prd_jt)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_path_planner/autoware_behavior_path_static_obstacle_avoidance_module/include/autoware/behavior_path_static_obstacle_avoidance_module/helper.hpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_static_obstacle_avoidance_module/include/autoware/behavior_path_static_obstacle_avoidance_module/helper.hpp
@@ -323,7 +323,7 @@ public:
 
   bool isFeasible(const AvoidLineArray & shift_lines) const
   {
-    const auto JERK_BUFFER = 0.1;  // [m/sss]
+    constexpr double JERK_BUFFER = 0.1;  // [m/sss]
     const auto & values = parameters_->velocity_map;
     const auto idx = getConstraintsMapIndex(0.0, values);  // use minimum avoidance speed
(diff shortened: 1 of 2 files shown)


planning/behavior_path_planner/autoware_behavior_path_static_obstacle_avoidance_module/include/autoware/behavior_path_static_obstacle_avoidance_module/helper.hpp
```diff
@@ -321,6 +321,19 @@ class AvoidanceHelper
     });
   }
 
+  bool isFeasible(const AvoidLineArray & shift_lines) const
+  {
+    const auto JERK_BUFFER = 0.1;  // [m/sss]
```
go-sakayori, line 326: nit: constexpr double might be better
```suggestion
    constexpr double JERK_BUFFER = 0.1;  // [m/sss]
```

go-sakayori (approved):

PR #9091, opened 2024-10-16
refactor(lane_change): refactor longitudinal acceleration sampling

mkquda:
## Description
when sampling lane change paths, the min & max longitudinal accelerations are computed once using the maximum prepare duration setting. The longitudinal acceleration is then sampled between the computed min & max values, and used to generated prepare phase metrics.

However when ego is approaching terminal, the prepare duration will also be sampled between 0 and max prepare duration, in that case the min & max longitudinal acceleration values computed previously can be invalid for prepare duration lower than max prepare duration, and can result in failure to generate LC path.

Therefore, it is more logical to compute min & max longitudinal accelerations and sample it for each prepare duration separately.

## Changes
- Move longitudinal acceleration sampling code to `calculation.cpp`
- Refactor `calc_prepare_phase_metrics` to sample lon acceleration for each prepare duration separately
- Remove unused functions

## Related links

None.

## How was this PR tested?
- PSIM.
- [TIER IV Internal Evaluation](https://evaluation.tier4.jp/evaluation/reports/8671ab7b-8d65-5bac-b777-fc55035e0392?project_id=prd_jt)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/utils/calculation.cpp
```diff
@@ -326,6 +299,92 @@ double calc_phase_length(
...
+  }
+
+  const auto max_dist_buffer = transient_data.current_dist_buffer.max;
+
+  if (max_dist_buffer > transient_data.dist_to_terminal_end) {
+    return calc_acceleration_values(min_accel, max_accel, sampling_num);
+  }
+
+  if (transient_data.is_ego_stuck) {
+    return calc_acceleration_values(min_accel, max_accel, sampling_num);
```
maxime-clem, line 369: These 3 conditions all return the same `calc_acceleration_values(min_accel, max_accel, sampling_num);` (same as the final return of the function l. 385).
Is there any way to simplify the conditions to make the 2 cases more clear.
- case 1: `return {max_accel}`.
- case 2: `return calc_acceleration_values(min_accel, max_accel, sampling_num);`

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/utils/calculation.cpp
```diff
@@ -339,10 +398,32 @@ double calc_lane_changing_acceleration(
...
+                         lc_param_ptr->min_length_for_turn_signal_activation;
+  const auto max_prepare_duration = lc_param_ptr->lane_change_prepare_duration;
+
+  // TODO(Azu) this check seems to cause scenario failures.
+  if (common_data_ptr->transient_data.dist_to_terminal_start >= threshold) {
+    return {max_prepare_duration};
+  }
+
+  std::vector<double> prepare_durations;
+  constexpr double step = 0.5;
```
maxime-clem, line 414: This step is not a parameter ?

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/scene.cpp
```diff
@@ -311,9 +316,7 @@ TurnSignalInfo NormalLaneChange::get_terminal_turn_signal_info() const
 
   const double nearest_dist_threshold = common_param.ego_nearest_dist_threshold;
   const double nearest_yaw_threshold = common_param.ego_nearest_yaw_threshold;
-  const size_t current_nearest_seg_idx =
-    autoware::motion_utils::findFirstNearestSegmentIndexWithSoftConstraints(
-      path.points, current_pose, nearest_dist_threshold, nearest_yaw_threshold);
+  const size_t current_nearest_seg_idx = common_data_ptr_->transient_data.current_path_seg_idx;
```
zulfaqar-azmi-t4, line 319:
```suggestion
  const auto current_nearest_seg_idx = common_data_ptr_->transient_data.current_path_seg_idx;
```

if possible i want to standardize by using auto

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/utils/calculation.cpp
mkquda: originally it wasn't a parameter, so I kept it as it is.
@zulfaqar-azmi-t4 do you think there is a need to make it a parameter?

zulfaqar-azmi-t4 (approved):

PR #9098, opened 2024-10-17
feat(autoware_test_utils): add path with lane id parser

zulfaqar-azmi-t4:
## Description

Some test might require path as input, so this PR adds a parser, so that user can use path for unit testing.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

Build and test, and not unit test failure observed.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


soblin (approved):

mkquda (approved):
LGTM

PR #9101, opened 2024-10-17
fix(mission_planner): return without change_route if new route is empty 

kosuke55:
## Description

Looks like an implementation error.
If we do not return, an emtpy route may be set.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

2024/10/21 https://evaluation.tier4.jp/evaluation/reports/d8006304-00be-5558-9c86-1576794020f1/?project_id=prd_jt

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


isamu-takagi (approved):
Thank you, it needed to abort the process as you say.

PR #9118, opened 2024-10-18
feat(autoware_external_cmd_converter): add ext cmd converter tests

danielsanchezaran:
## Description
Add tests to increase test  coverage of external cmd converter

image

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/vehicle/autoware_external_cmd_converter/src/node.cpp
+++ b/vehicle/autoware_external_cmd_converter/src/node.cpp
@@ -15,6 +15,7 @@
 #include "autoware_external_cmd_converter/node.hpp"
 
 #include <algorithm>
+#include <cmath>
 #include <memory>
 #include <string>
 #include <utility>
@@ -147,6 +148,14 @@ double ExternalCmdConverterNode::calculate_acc(const ExternalControlCommand & cm
 {
   const double desired_throttle = cmd.control.throttle;
   const double desired_brake = cmd.control.brake;
+  if (
+    std::isnan(desired_throttle) || std::isnan(desired_brake) || std::isinf(desired_throttle) ||
+    std::isinf(desired_brake)) {
+    std::cerr << "Input brake or throttle is out of range. returning 0.0 acceleration."
+              << std::endl;
+    return 0.0;
+  }
+
   const double desired_pedal = desired_throttle - desired_brake;
 
   double ref_acceleration = 0.0;
--- a/vehicle/autoware_external_cmd_converter/tests/test_external_cmd_converter.cpp
+++ b/vehicle/autoware_external_cmd_converter/tests/test_external_cmd_converter.cpp
@@ -1,0 +1,169 @@
+// Copyright 2024 TIER IV, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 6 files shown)


vehicle/autoware_external_cmd_converter/tests/test_external_cmd_converter.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+    const ExternalControlCommand & cmd, const double vel)
+  {
+    return external_cmd_converter_->calculate_acc(cmd, vel);
+  }
+
+  std::shared_ptr<ExternalCmdConverterNode> external_cmd_converter_;
+};
+
+TEST_F(TestExternalCmdConverter, testCheckEmergencyStopTimeout)
```
TakaHoribe, line 125: Would you show a brief description of each test case?

vehicle/autoware_external_cmd_converter/tests/test_external_cmd_converter.cpp
```diff
@@ -0,0 +1,169 @@
...
+  EXPECT_DOUBLE_EQ(1.0, test_get_shift_velocity_sign(cmd));
+  cmd.command = GearCommand::LOW;
+  EXPECT_DOUBLE_EQ(1.0, test_get_shift_velocity_sign(cmd));
+  cmd.command = GearCommand::REVERSE;
+  EXPECT_DOUBLE_EQ(-1.0, test_get_shift_velocity_sign(cmd));
+  cmd.command = GearCommand::LOW_2;
+  EXPECT_DOUBLE_EQ(0.0, test_get_shift_velocity_sign(cmd));
+}
+
+TEST_F(TestExternalCmdConverter, testCalculateAcc)
```
TakaHoribe, line 153: Could you make test cases for boundary conditions? For example, a case with `throttle = Inf (large number)`?

danielsanchezaran: @TakaHoribe thanks for your suggestions. I have added descriptions to the tests and functions used to evaluate. I also added combinations of std::numeric_limits<double>::infinity() to the throttle and brake. Could you please check again?

TakaHoribe: You only check if the output is positive/negative for infinite cases. Are you sure that those outputs are not nan nor other invalid numbers?
This PR looks ok if the above concern is solved.

vehicle/autoware_external_cmd_converter/src/node.cpp
```diff
@@ -147,6 +148,14 @@ double ExternalCmdConverterNode::calculate_acc(const ExternalControlCommand & cm
...
   const double desired_throttle = cmd.control.throttle;
   const double desired_brake = cmd.control.brake;
+  if (
+    std::isnan(desired_throttle) || std::isnan(desired_brake) || std::isinf(desired_throttle) ||
+    std::isinf(desired_brake)) {
+    std::cerr << "Input brake or throttle is out of range. returning 0.0 acceleration."
+              << std::endl;
+    return 0.0;
+  }
+
```
danielsanchezaran, line 158:
```suggestion
  if (
    std::isnan(desired_throttle) || std::isnan(desired_brake) || std::isinf(desired_throttle) ||
    std::isinf(desired_brake)) {
    std::cerr << "Input brake or throttle is out of range. returning 0.0 acceleration."
              << std::endl;
    return 0.0;
  }

```
@TakaHoribe I noticed there is no proper handling of Nan or infinite values in the external_cmd_converter so I decided to add this check, but I am wondering if returning 0 acceleration is the right answer in this case? 

I have also made checks to combinations of infinity and nan inputs and checking the output to not be NaN or inf.

TakaHoribe (approved):
LGTM. Thank you for your nice work!

PR #9120, opened 2024-10-18
feat(control_validator): add hold and lpf

yuki-takagi-66:
## Description
This PR add holding function and LPF for the velocity check.
in additon, velocity check is separated to two diags to trigger different MRM behavior.

## Related links
launch PR: https://github.com/autowarefoundation/autoware_launch/pull/1193

**Parent Issue:**

- Link

## How was this PR tested?
tier4 internal tests

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/control/autoware_control_validator/README.md
+++ b/control/autoware_control_validator/README.md
@@ -9,10 +9,10 @@ The `control_validator` is a module that checks the validity of the output of th
 The following features are supported for the validation and can have thresholds set by parameters.
 The listed features below does not always correspond to the latest implementation.
 
-| Description                                                                        | Arguments                                                                              |                Diagnostic equation                | Implemented function name       |
-| ---------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------- | :-----------------------------------------------: | ------------------------------- |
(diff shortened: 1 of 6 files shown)


kyoichi-sugahara:
@yuki-takagi-66 
Please update newly introduced parameter description in README and add parameter in the yaml file in the package with default value.

kyoichi-sugahara:
@yuki-takagi-66 
If there's no particular reason for removing the docstrings in this PR, it would be clearer to add them back in accordance with the changes. Or is it possible that they're being temporarily removed due to potential future changes?

Also, since I'm not familiar with the background of these changes, the processing seems fine to me. However, is it appropriate for me to approve without understanding the context or motivation behind these changes?

yuki-takagi-66:
@kyoichi-sugahara 
Thank you for the review.
I've update README and yaml in universe side.

I think we have no additional information to be added as docstrings from the function name and the arg names.
And we do not operate doxygen docs.
Therefore, I think we should **not** maintain the docstrings for this function, although we have option to maintain.

> Also, since I'm not familiar with the background of these changes, the processing seems fine to me. However, is it appropriate for me to approve without understanding the context or motivation behind these changes?
Maybe it is not a sufficient description, LPF and hold is required to achieve the appropriate behavior along with MRM.
I've add some of descriptions.

kyoichi-sugahara (approved):
LGTM!

PR #9121, opened 2024-10-21
fix(crosswalk): fix occlusion detection range calculation and add debug markers

maxime-clem:
## Description

Fix a mistake in the detection range calculation.
- Before: detection range (`m`) = time for ego to reach the crosswalk (`s`) _divided_ by occluded object velocity (`m/s`).
    - seconds divided by meters per seconds --> second squared per meter :warning:
- After: detection range (`m`) = time for ego to reach the crosswalk (`s`) _multiplied_ by occluded object velocity (`m/s`).
    - seconds multiplied by meters per seconds --> meters. :heavy_check_mark: 

Also do a small refactoring and add some markers to visualize the occlusion detection area
## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


yuki-takagi-66 (approved):
LGTM

PR #9123, opened 2024-10-21
fix(utils): fix envelope polygon update logic

satoshi-ota:
## Description

I found a wrong implementation of envelope polygon update logic. It seemed that object's envelope polygon was updated even when the latest observed data was more acculate than previous one.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

- [x] [PASS TIER IV INTERNAL SCENARIOS](https://evaluation.tier4.jp/evaluation/reports/3fbf20a1-b8c1-5e39-933d-9748e43c6194?project_id=prd_jt)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


go-sakayori (approved):
Thank you very much for fixing my bug!

Issue #9131, opened 2024-10-21
[autoware_point_types] unable to compile under Jazzy: Could not find PCL


Timple:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I'm convinced that this is not my fault but a bug.

### Description

```
$ source /opt/ros/jazzy/setup.bash && colcon build --packages-up-to autoware_point_types
Starting >>> autoware_point_types
--- stderr: autoware_point_types
CMake Error at /opt/ros/jazzy/share/pcl_conversions/cmake/ament_cmake_export_dependencies-extras.cmake:21 (find_package):
  By not providing "FindPCL.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "PCL", but
  CMake did not find one.

  Could not find a package configuration file provided by "PCL" with any of
  the following names:

    PCLConfig.cmake
    pcl-config.cmake

  Add the installation prefix of "PCL" to CMAKE_PREFIX_PATH or set "PCL_DIR"
  to a directory containing one of the above files.  If "PCL" provides a
  separate development package or SDK, be sure it has been installed.
Call Stack (most recent call first):
  /opt/ros/jazzy/share/pcl_conversions/cmake/pcl_conversionsConfig.cmake:41 (include)
  /opt/ros/jazzy/share/pcl_ros/cmake/ament_cmake_export_dependencies-extras.cmake:21 (find_package)
  /opt/ros/jazzy/share/pcl_ros/cmake/pcl_rosConfig.cmake:41 (include)
  /opt/ros/jazzy/share/ament_cmake_auto/cmake/ament_auto_find_build_dependencies.cmake:67 (find_package)
  /opt/ros/jazzy/share/autoware_cmake/cmake/autoware_package.cmake:49 (ament_auto_find_build_dependencies)
  CMakeLists.txt:5 (autoware_package)


---
Failed   <<< autoware_point_types [0.86s, exited with code 1]
```

### Expected behavior

No failed compliation

### Actual behavior

See title and description?

### Steps to reproduce

```Dockerfile
FROM ros:jazzy
SHELL [ "/bin/bash", "-c" ]
RUN apt-get update -qq && apt-get install -qqy python3-colcon-core
RUN mkdir -p /ws/src
WORKDIR /ws
[...]

PR #9132, opened 2024-10-21
Include pcl-dev for <pcl/point_types.h>

Timple:
## Description

`pcl-dev` needs to be installed for this line: 

https://github.com/autowarefoundation/autoware.universe/blob/main/common/autoware_point_types/include/autoware_point_types/types.hpp#L20

## Related links

**Parent Issue:**

- #9131

## How was this PR tested?

With the dockerfile from #9131

## Notes for reviewers

I know currently humble is supported, but this dependency should be explicit anyhow. Instead of coming in luckily via deps :slightly_smiling_face: 

## Interface changes

Added dependency

## Effects on system behavior

None.

### DIFF
--- a/common/autoware_point_types/package.xml
+++ b/common/autoware_point_types/package.xml
@@ -20,6 +20,7 @@
   <depend>ament_cmake_cppcheck</depend>
   <depend>ament_cmake_lint_cmake</depend>
   <depend>ament_cmake_xmllint</depend>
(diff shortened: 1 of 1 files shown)


Timple:
Needs more thought, this seems relevant: https://github.com/autowarefoundation/autoware.universe/issues/9131

Timple:
Not the right place

Issue #9117, opened 2024-10-18
could not set use_intra_process in map.launch.xml


LuZhishan:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I'm convinced that this is not my fault but a bug.

### Description

Default value of `use_intra_process` is `false` [here](https://github.com/autowarefoundation/autoware.universe/blob/ae24913390600d95b3194f7e3ddf43ac823261ce/launch/tier4_map_launch/launch/map.launch.xml#L15),but if set it to `true`, there will be an error

### Expected behavior

load node to `map_container`

### Actual behavior

![image](https://github.com/user-attachments/assets/4c47319c-4dc1-455b-8482-0495b03bce98)

### Steps to reproduce

1. set `use_intra_process` to true
2. launch map.launch.xml

### Versions

- OS : ubuntu22.04
- ROS2 : humble
- AutoWare : latest

LuZhishan:
Could you help me with this problem ? @YamatoAndo @SakodaShintaro @cyn-liu

YamatoAndo:
@LuZhishan 
Sorry, the map component uses [transient_local](https://github.com/autowarefoundation/autoware.universe/blob/main/map/map_loader/src/lanelet2_map_loader/lanelet2_map_loader_node.cpp#L130C75-L130C90), so it may not be possible to set `use_intra_process=true`.

YamatoAndo:
@LuZhishan Yes, I think it should be remove.

PR #9138, opened 2024-10-23
fix(tier4_map_launch): remove parameter use_intra_process

cyn-liu:
## Description
The default value  of parameter `use_intra_process` is `false` in package [tier4_map_launch](https://github.com/autowarefoundation/autoware.universe/blob/ae24913390600d95b3194f7e3ddf43ac823261ce/launch/tier4_map_launch/launch/map.launch.xml#L15), but if set it to true, there will be an error. That's because map component uses [transient_local](https://github.com/autowarefoundation/autoware.universe/blob/main/map/map_loader/src/lanelet2_map_loader/lanelet2_map_loader_node.cpp#L130C75-L130C90), so it may not be possible to set use_intra_process=true. 

To avoid unnecessary errors, delete parameter `use_intra_process`.

## Related links

**Parent Issue:** [issue 9117](https://github.com/autowarefoundation/autoware.universe/issues/9117)

## How was this PR tested?

Build successful.

## Notes for reviewers

None.

#### Additions and removals

| Removed | `use_intra_process`   | `bool` | `false`         | Not used |

## Effects on system behavior

None.

### DIFF
--- a/launch/tier4_map_launch/launch/map.launch.xml
+++ b/launch/tier4_map_launch/launch/map.launch.xml
@@ -11,9 +11,6 @@
   <arg name="map_tf_generator_param_path"/>
   <arg name="map_projection_loader_param_path"/>
 
-  <!-- whether use intra-process -->
-  <arg name="use_intra_process" default="false"/>
-
   <!-- select container type -->
   <arg name="use_multithread" default="false"/>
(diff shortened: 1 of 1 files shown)


YamatoAndo:
@cyn-liu Thank you!

PR #9141, opened 2024-10-23
fix(autonomous_emergency_braking): fix no backward imu path and wrong back distance usage

danielsanchezaran:
## Description
This PR introduces a fix for backwards path generation and collision judgement. It fixes a bug that causes the path length to be negative which makes the AEB NOT generate an IMU path when ego moves in reverse (because ego speed is negative and the length should be always positive). It also makes it so we have to check the ego speed to calculate whatoffset should be used to calculate collisions when moving backwards. These changes only affect backward driving behavior.
## Related links

**Parent Issue:**

- Link

## How was this PR tested?
PSim
## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/control/autoware_autonomous_emergency_braking/src/node.cpp
+++ b/control/autoware_autonomous_emergency_braking/src/node.cpp
@@ -467,7 +467,9 @@ bool AEB::checkCollision(MarkerArray & debug_markers)
       use_pointcloud_data_ && points_belonging_to_cluster_hulls &&
       !points_belonging_to_cluster_hulls->empty()) {
       const auto current_time = obstacle_ros_pointcloud_ptr_->header.stamp;
-      getClosestObjectsOnPath(path, current_time, points_belonging_to_cluster_hulls, objects);
+      const bool ego_moves_forward = current_v > 0.0;
+      getClosestObjectsOnPath(
+        path, current_time, points_belonging_to_cluster_hulls, objects, ego_moves_forward);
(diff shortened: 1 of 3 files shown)


control/autoware_autonomous_emergency_braking/src/node.cpp
```diff
@@ -467,7 +467,9 @@ bool AEB::checkCollision(MarkerArray & debug_markers)
       use_pointcloud_data_ && points_belonging_to_cluster_hulls &&
       !points_belonging_to_cluster_hulls->empty()) {
       const auto current_time = obstacle_ros_pointcloud_ptr_->header.stamp;
-      getClosestObjectsOnPath(path, current_time, points_belonging_to_cluster_hulls, objects);
+      const bool ego_moves_forward = current_v > 0.0;
```
kyoichi-sugahara, line 470: @danielsanchezaran 
How about using `isDrivingForward` util function to judge driving forward or not?
If trajectory is not subscribed, then check velocity directly is ok. The concern is velocity estimation's chattering.

kyoichi-sugahara (approved):
LGTM.
Please check my comment about driving direction judge part.

danielsanchezaran:
> LGTM. Please check my comment about driving direction judge part.

Thank you very much, I added your suggested change and the changes are much simpler and clean!

PR #9143, opened 2024-10-23
feat(autoware_planning_validator): add collision checking for predicted objects

kyoichi-sugahara:
## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


planning/autoware_planning_validator/src/utils.cpp
```diff
@@ -289,6 +293,139 @@ std::pair<double, size_t> calcMaxSteeringRates(
...
+  motion_utils::calculate_time_from_start(filtered_trajectory, current_ego_position);
+
+  // Generate vehicle footprint polygons along the trajectory
+  std::vector<Polygon2d> vehicle_footprints;
+  vehicle_footprints.reserve(filtered_trajectory.size());
+  for (const auto & point : filtered_trajectory) {
+    vehicle_footprints.push_back(createVehicleFootprintPolygon(point.pose, vehicle_info));
+  }
+
+  const double time_tolerance = 0.1;
```
maxime-clem, line 325: should be a parameter

planning/autoware_planning_validator/src/planning_validator.cpp
```diff
@@ -545,8 +553,25 @@ bool PlanningValidator::checkValidForwardTrajectoryLength(const Trajectory & tra
   return forward_length > forward_length_required;
 }
 
+bool PlanningValidator::checkValidTrajectoryCollision(const Trajectory & trajectory)
+{
+  const auto ego_speed = std::abs(current_kinematics_->twist.twist.linear.x);
+  if (ego_speed < 1.0 / 3.6) {
+    return true;  // Ego is almost stopped.
+  }
```
maxime-clem, line 561: Why do we care about the current ego velocity when checking for collisions over the planned trajectory ?

maxime-clem (commented):
If you want to efficiently check for collisions then you should use an rtree to store and do queries on the ego trajectory footprints.

planning/autoware_planning_validator/src/planning_validator.cpp
kyoichi-sugahara: This is because when the ego vehicle is stationary, it is acceptable for its predicted path to intersect with the predicted paths of NPC vehicles. 
For example, there may be cases where an NPC vehicle drives toward the ego vehicle stopped on the road and then avoids it. We want to exclude such cases from being detected as errors.

maxime-clem: I think a maximum time horizon is better.
Even if ego is stopped, if we generate a trajectory that collides with a front NPC at `t=2s`, I think this is an error.
Even if ego is driving, if we generate a trajectory that collides with a NPC at `t=10s`, it may be okay to ignore.
But I do not know all background so feel free to ignore my comments :)

kyoichi-sugahara: @maxime-clem 
Thank you so much for the comment!!!!
For example, let's consider a scenario where ego vehicle needs to stop mid-way during departure from shoulder lane due to an approaching vehicle from behind.
In this case, I want to avoid having ego vehicle's current position footprint overlap with the predicted path of the following vehicle.
Is it possible to determine that the path planning in this scenario is not error with current implementation?

image

planning/autoware_planning_validator/src/utils.cpp
```diff
@@ -289,6 +293,139 @@ std::pair<double, size_t> calcMaxSteeringRates(
...
+        // Skip if timestamps don't match within tolerance
+        if (std::fabs(trajectory_time - predicted_time) > time_tolerance) {
+          continue;
+        }
+
+        // Skip collision check if object is too far away
+        const double distance = autoware::universe_utils::calcDistance2d(
+          trajectory_point.pose.position, selected_predicted_path->path[j].position);
+        if (distance >= 10.0) {
+          continue;
```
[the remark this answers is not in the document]
kyoichi-sugahara: changed to use rtree :+1:  Thanks!

planning/autoware_planning_validator/src/utils.cpp
```diff
@@ -289,6 +293,181 @@ std::pair<double, size_t> calcMaxSteeringRates(
...
+  std::vector<autoware_planning_msgs::msg::TrajectoryPoint> collision_points;
+  for (const auto & [ego_index, obj_index] : collision_index_set) {
+    collision_points.push_back(filtered_trajectory[ego_index]);
+  }
+
+  return collision_points.empty() ? std::nullopt : std::make_optional(collision_points);
+}
+
+Rtree make_ego_footprint_rtree(
+  std::vector<autoware_planning_msgs::msg::TrajectoryPoint> & trajectory,
```
satoshi-ota, line 362: @kyoichi-sugahara Please fix cppcheck error.

satoshi-ota, line 362:
```suggestion
  const std::vector<autoware_planning_msgs::msg::TrajectoryPoint> & trajectory,
```

kyoichi-sugahara:
@satoshi-ota 
After visualizing the footprint of predicted path of ego vehicle and the polygon of obstacles where collision is detected, I will merge them.
Thank you so much for the corrections.

satoshi-ota (approved):

mitsudome-r:
@kyoichi-sugahara This has been stale for a long time. Could you check if you still need this PR?

xmfcx:
@kyoichi-sugahara I'm closing this due to inactivity. Feel free to reopen if needed.

PR #9148, opened 2024-10-24
feat(autoware_shape_estimation): add reference object based corrector

a-maumau:
## Description
This PR introduces a corrector in the `shape_estimation` module based on the reference object (tracked object).
This PR only affects the result of merged object from `detection_by_tracker`.

### method
The original implementation of the corrector in `shape_estimation` which is called by `detection_by_tracker` may result in low IoU bounding boxes in certain cases (see Fig. 1 (a)).
Considering the purpose of correcting an oversegmentation, the corrected bounding box should overlap with the tracked object with a higher IoU.

This PR improves the corrector by choosing a more reasonable direction when adjusting the bounding box.

shape_estimation_algo_change_with_dbt
fig.1: (a) is the current implementation of the [corrector used from `detection_by_tracker`](https://github.com/autowarefoundation/autoware.universe/blob/26f617d45eebcc2fa025528df420f74d52add338/perception/autoware_shape_estimation/lib/corrector/utils.cpp#L335), (b) the method implemented in this PR.

Fig. 1 (b) shows the new method implemented in this PR, which simply uses the center of the reference object to select a more reasonable direction by choosing the farthest point.
This approach mitigates the negative effect caused by a bounding box that faces the opposite direction from the reference object, as shown in Fig. 1 (a).

## Related links

**Parent Issue:**

- Link

## How was this PR tested?
This PR was tested with the data which contains situations like (a).  
Potentially, it might change the behavior in other cases.

## Notes for reviewers

Discussion:
- Justification of the point selection.
  - changes in this PR is intended to be small (avoid high computation cost).
- Edge cases which I might not seeing.

## Interface changes

None.

## Effects on system behavior

None.


yukkysaito:
@kaancolak @YoshiRi @technolojin I haven’t checked the code yet, but I believe the algorithm itself is sound.
I don’t have time for a code review, so I’d like to ask another approver to handle the review itself.

technolojin:
test
node-alive: https://evaluation.tier4.jp/evaluation/reports/b2472102-6ec8-5c94-af9a-bb0c90c0cbbc?project_id=prd_jt

technolojin:
Test (performance) https://evaluation.tier4.jp/evaluation/reports/344d6f50-0e33-5c5d-9c73-4553b73c465e?project_id=prd_jt

technolojin:
@a-maumau 
Can you resolve the cppcheck failure?
image

a-maumau:
@technolojin I have fixed the cppcheck warning in https://github.com/autowarefoundation/autoware.universe/pull/9148/commits/068d7f3a6c6c3c70eb59a1ec58e1603321c5cf40

technolojin (approved):
LGTM!

YoshiRi (approved):
LGTM!!

PR #9152, opened 2024-10-24
Modify Freespace Planner to use Node Time (to reflect use_sim_time in module) Instead of System Time

stevenbrills:
## Description
Modified the autoware_freespace_planner and autoware_freespace_planning_algorithms packages to use the node clock instead of rclcpp detached clock. This allows the module to make use of sim time. Previously during simulation the parking trajectory would have system time in trajectory header messages causing downstream issues like non-clearance of trajectory buffers in motion planning based on elapsed time.

## How was this PR tested?
Tested in sim.

## Effects on system behavior

In the previous version, the use of system time in the free-space planner trajectory header causes the trajectory buffer downstream from clearing and therefore causing the is_steering_converged flag to never reset when simulating causing the system to not change from Parking back to Lane Driving. Now, when use_sim_time is used, the parking trajectory also reflects the correct time and the buffer can be cleared appropriately.

### DIFF
--- a/planning/autoware_freespace_planner/include/autoware/freespace_planner/utils.hpp
+++ b/planning/autoware_freespace_planner/include/autoware/freespace_planner/utils.hpp
@@ -58,13 +58,13 @@ size_t get_next_target_index(
   const size_t current_target_index);
 
 Trajectory get_partial_trajectory(
-  const Trajectory & trajectory, const size_t start_index, const size_t end_index);
+  const Trajectory & trajectory, const size_t start_index, const size_t end_index, const rclcpp::Clock::SharedPtr clock);
 
 Trajectory create_trajectory(
   const PoseStamped & current_pose, const PlannerWaypoints & planner_waypoints,
   const double & velocity);
 
-Trajectory create_stop_trajectory(const PoseStamped & current_pose);
+Trajectory create_stop_trajectory(const PoseStamped & current_pose, const rclcpp::Clock::SharedPtr clock);
 
 Trajectory create_stop_trajectory(const Trajectory & trajectory);
 
--- a/planning/autoware_freespace_planning_algorithms/src/astar_search.cpp
+++ b/planning/autoware_freespace_planning_algorithms/src/astar_search.cpp
@@ -72,7 +72,33 @@ Pose calcRelativePose(const Pose & base_pose, const Pose & pose)
 AstarSearch::AstarSearch(
   const PlannerCommonParam & planner_common_param, const VehicleShape & collision_vehicle_shape,
   const AstarParam & astar_param)
-: AbstractPlanningAlgorithm(planner_common_param, collision_vehicle_shape),
+: AbstractPlanningAlgorithm(planner_common_param, std::make_shared<rclcpp::Clock>(RCL_ROS_TIME), collision_vehicle_shape),
+  astar_param_(astar_param),
+  goal_node_(nullptr),
+  use_reeds_shepp_(true)
+{
+  steering_resolution_ =
+    collision_vehicle_shape_.max_steering / planner_common_param_.turning_steps;
+  heading_resolution_ = 2.0 * M_PI / planner_common_param_.theta_size;
+
+  const double avg_steering =
+    steering_resolution_ + (collision_vehicle_shape_.max_steering - steering_resolution_) / 2.0;
+  avg_turning_radius_ =
+    kinematic_bicycle_model::getTurningRadius(collision_vehicle_shape_.base_length, avg_steering);
+
+  is_backward_search_ = astar_param_.search_method == "backward";
+
+  min_expansion_dist_ = astar_param_.expansion_distance;
+  max_expansion_dist_ = collision_vehicle_shape_.base_length * base_length_max_expansion_factor_;
+
+  near_goal_dist_ =
+    std::max(astar_param.near_goal_distance, planner_common_param.longitudinal_goal_range);
+}
+
+AstarSearch::AstarSearch(
+  const PlannerCommonParam & planner_common_param, const VehicleShape & collision_vehicle_shape,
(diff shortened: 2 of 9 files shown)


planning/autoware_freespace_planner/include/autoware/freespace_planner/utils.hpp
```diff
@@ -58,13 +58,15 @@ size_t get_next_target_index(
   const size_t current_target_index);
 
 Trajectory get_partial_trajectory(
-  const Trajectory & trajectory, const size_t start_index, const size_t end_index);
+  const Trajectory & trajectory, const size_t start_index, const size_t end_index,
+  const rclcpp::Clock::SharedPtr clock);
```
mkquda, line 62: you need to update the [unit test](https://github.com/autowarefoundation/autoware.universe/blob/main/planning/autoware_freespace_planner/test/test_freespace_planner_utils.cpp#L215) as well.

I would suggest setting the timestamp on the calling side after returning the trajectory from this function.
Then you don't need to pass the clock ptr and you don't need to worry about modifying the unit tests.

planning/autoware_freespace_planner/include/autoware/freespace_planner/utils.hpp
```diff
@@ -58,13 +58,15 @@ size_t get_next_target_index(
...
 
 Trajectory create_trajectory(
   const PoseStamped & current_pose, const PlannerWaypoints & planner_waypoints,
   const double & velocity);
 
-Trajectory create_stop_trajectory(const PoseStamped & current_pose);
+Trajectory create_stop_trajectory(
+  const PoseStamped & current_pose, const rclcpp::Clock::SharedPtr clock);
```
mkquda, line 69: you need to update the [unit test](https://github.com/autowarefoundation/autoware.universe/blob/main/planning/autoware_freespace_planner/test/test_freespace_planner_utils.cpp#L259) as well.

I would suggest setting the timestamp on the calling side after returning the trajectory from this function.
Then you don't need to pass the clock ptr and you don't need to worry about modifying the unit tests.

planning/autoware_freespace_planning_algorithms/src/astar_search.cpp
```diff
@@ -72,7 +72,34 @@ Pose calcRelativePose(const Pose & base_pose, const Pose & pose)
...
+  max_expansion_dist_ = collision_vehicle_shape_.base_length * base_length_max_expansion_factor_;
+
+  near_goal_dist_ =
+    std::max(astar_param.near_goal_distance, planner_common_param.longitudinal_goal_range);
+}
+
+AstarSearch::AstarSearch(
+  const PlannerCommonParam & planner_common_param, const VehicleShape & collision_vehicle_shape,
+  const AstarParam & astar_param, const rclcpp::Clock::SharedPtr & clock)
+: AbstractPlanningAlgorithm(planner_common_param, clock, collision_vehicle_shape),
```
mkquda, line 102: this implementation is completely fine, but I'm wondering if it's worth doing it just for [this one line](https://github.com/autowarefoundation/autoware.universe/pull/9152/files#diff-68a0a8dee085da7456b38aba6f7bf632516d1a3499cb5d88506277ba2a9d92c9R434)

I think we can leave the responsibility of setting the time stamp to the parent node in which the freespace planning algorithm is used ...

Just a thought.

planning/autoware_freespace_planner/include/autoware/freespace_planner/utils.hpp
stevenbrills: Hi @mkquda, I've updated the missed unit tests to provide a shared clock ptr. The only reason for keeping it this way is so that if we want to develop other modules that make use of the freespace planner then they can treat it like a black box and not have to worry about doing sanity checks with the output.

planning/autoware_freespace_planner/include/autoware/freespace_planner/utils.hpp
stevenbrills: Hi @mkquda, I've updated the missed unit tests to provide a shared clock ptr. The only reason for keeping it this way is so that if we want to develop other modules that make use of the freespace planner then they can treat it like a black box and not have to worry about doing sanity checks with the output.

planning/autoware_freespace_planning_algorithms/src/astar_search.cpp
stevenbrills: Yes, I agree that would reduce these extra lines in the actual planning algorithm but we wanted to make it such that any other module making use of the freespace planner would be able to use the output as is without having to do extra sanity checks. That's the reason for going this route.

stevenbrills:
Hi @mkquda, is there anything else to be re-evaluated?

mkquda (approved):
LGTM.

danielsanchezaran (approved):

kosuke55 (approved):

kosuke55:
There are other parts of autoware that use `rclcpp::Clock()` besides this PR change, do they all need to be replaced with node clock correctly?

soblin (approved):

stevenbrills:
> There are other parts of autoware that use `rclcpp::Clock()` besides this PR change, do they all need to be replaced with node clock correctly?

@kosuke55 Only if they are being used to populate headers or are directly/indirectly being used by consuming nodes as a reference of when the message was created in the context of the run either in real system time or simulated time. There are places where a detached clock is used for measuring algorithm processing time which does not need to change.

stevenbrills:
@mkquda I forgot to add signoffs to the commit messages, and I am unable to directly rebase and add the signoffs directly to the pull request due to insufficient permissions. Is there a workaround?

stevenbrills:
@maxime-clem @mkquda 
Gentle reminder on this pending merge.

PR #9197, opened 2024-10-30
perf(autoware_pointcloud_preprocessor): introduce managed transform buffer with implicitly defined listener type

amadeuszsz:
## Description

See parent issue.

**Order of PRs merge:**
- https://github.com/autowarefoundation/autoware/pull/5611
- https://github.com/autowarefoundation/autoware_utils/pull/41
- https://github.com/autowarefoundation/autoware_universe/pull/9197 `<---`
- https://github.com/autowarefoundation/autoware_core/pull/360
- https://github.com/autowarefoundation/autoware_launch/pull/1393

## Related links

**Parent Issue:**

- https://github.com/autowarefoundation/autoware/issues/5385

## How was this PR tested?

* See parent issue for results.
* To test locally, you need to pull related PR, https://github.com/autowarefoundation/autoware_utils/pull/41 and https://github.com/autowarefoundation/ManagedTransformBuffer as well.
* This PR applies changes to two packages only. You can also test ManagedTransformBuffer applied for all the autoware.universe:
```
cd ~/autoware
git checkout 0.41.2
vcs import src < autoware.repos
git clone https://github.com/autowarefoundation/ManagedTransformBuffer src/universe/external/managed_transform_buffer
cd src/universe/autoware/autoware.universe
git remote add amadeuszsz https://github.com/amadeuszsz/autoware.universe.git
git pull amadeuszsz feat/managed-tf-buffer-demo-test
git checkout feat/managed-tf-buffer-demo-test
cd ~/autoware
cd src/core/autoware_utils
git remote add amadeuszsz https://github.com/amadeuszsz/autoware_utils.git
git pull amadeuszsz feat/remove-managed-tf-buffer
git checkout feat/remove-managed-tf-buffer
cd ~/autoware
colcon build --symlink-install --cmake-args -DCMAKE_BUILD_TYPE=Release"-DCMAKE_EXPORT_COMPILE_COMMANDS=On
```
```
# press space to resume rosbag
ros2 bag play ~/autoware_data/rosbags/sample-rosbag/ -r 1.0 -p
```
To validate number of listeners during runtime, use `ros2 node list | grep impl -c`.

## Interface changes

- `has_static_tf_only` parameter is not necessary anymore.
- `getTransform` now has form `getTransform<typename T>` (`geometry_msgs::msg::TransformStamped`, `tf2::Transform`, `Eigen::Matrix4f`).
- `getTransform` returns `std::optional`.

## Effects on system behavior

See parent issue.


perception/autoware_ground_segmentation/README.md
```diff
@@ -33,15 +33,14 @@ Detail description of each ground segmentation algorithm is in the following lin
 
 ### Node Parameters
 
-| Name                 | Type   | Default Value | Description                           |
-| -------------------- | ------ | ------------- | ------------------------------------- |
-| `input_frame`        | string | " "           | input frame id                        |
```
knzo25, line 38: Not really for this PR, but
" " does not evaluate as empty, so I think it goes against its apparent objective
https://github.com/autowarefoundation/autoware.universe/blob/b3caa5cea50a6747ff77f96786396b3176840a74/sensing/autoware_pointcloud_preprocessor/src/filter.cpp#L274

@badai-nguyen  what do you think?

badai-nguyen: @knzo25 I'm so sorry, I missed your mention here let have a look.

knzo25:
@amadeuszsz 
Additional nodes have been added since this PR was opened. Can you make them use the managed transform?

knzo25 (changes requested):
@amadeuszsz 
I am sorry, but it seems we will not be able to merge this as is.
With the taxi,we still have a rear lidar configured, even though in 99% of the cases, the cloud will be empty.

The concat waits for the rear cloud, and accepts it even when it is empty. If it does not receive it, it will publish the cloud much later, effectively publishing it 1 frame off, so we need to publish it.

I checked the base autoware and this PR + the awf/managed_transform and could confirm this phenomena.

An easy fix would be to consider a pointcloud transform successful no matter what frames are being transformed when the pointcloud is empty

amadeuszsz:
> @amadeuszsz I am sorry, but it seems we will not be able to merge this as is. With the taxi,we still have a rear lidar configured, even though in 99% of the cases, the cloud will be empty.
> 
> The concat waits for the rear cloud, and accepts it even when it is empty. If it does not receive it, it will publish the cloud much later, effectively publishing it 1 frame off, so we need to publish it.
> 
> I checked the base autoware and this PR + the awf/managed_transform and could confirm this phenomena.
> 
> An easy fix would be to consider a pointcloud transform successful no matter what frames are being transformed when the pointcloud is empty

Thanks for review! Empty cloud is not an error anymore. Also updated latest node.

knzo25:
@amadeuszsz 

 - How do you plan to add this to autoware? 
 - Since you have added all the other universe nodes, can you add the one for the cuda pointcloud preprocessor ?
 - Can the camel case repo name cause issues for the build farm?

knzo25 (changes requested):
@amadeuszsz
Confirmed that it works as expected, including the previous problematic case. However:

 - How do you plan to add this to autoware?
 - Since you have added all the other universe nodes, can you add the one for the cuda pointcloud preprocessor ?
 - Can the camel case repo name cause issues for the build farm?

amadeuszsz:
> @amadeuszsz Confirmed that it works as expected, including the previous problematic case. However:
> 
> * How do you plan to add this to autoware?

PR https://github.com/autowarefoundation/autoware/pull/5611 already awaiting.

> * Since you have added all the other universe nodes, can you add the one for the cuda pointcloud preprocessor ?

This PR targets only `autoware_pointcloud_preprocessor` and other packages can be addressed in separate PRs. Only test branch applied managed TF buffer on all the nodes for testing purposes.

> * Can the camel case repo name cause issues for the build farm?

I don't think so. `snake_case` is required for `autoware_xyz` packages [only](https://autowarefoundation.github.io/autoware-documentation/main/contributing/coding-guidelines/languages/cpp/#use-lower-snake-case-for-function-names-required-partially-automated). However, if you think it would be better to refactor it now, we can do it 🥲

knzo25 (approved):
LGTM !
Can you get approval for the autoware PR before merging this one?
That way we can merge both with minimal time difference

knzo25:
PS: when both PRs are merged remind me to apply the managed transform to the cuda preprocessor. I need to also add the polling subscriber so I can do both in one seating

knzo25:
@badai-nguyen  @soblin 
Can you take a look a this please ? :pray:

amadeuszsz:
I will also address `autoware_core` since those minimal nodes use deprecated `managed_transform_buffer` from `autoware_utils`.

soblin (approved):

badai-nguyen:
@amadeuszsz 
I think a refactor PR to remove `has_static_tf_only` from autoware_launch is needed such as
https://github.com/autowarefoundation/autoware_launch/blob/0b9caea24347f57a062f29b3524caef73a3295ff/sensor_kit/awsim_labs_sensor_kit_launch/awsim_labs_sensor_kit_launch/config/concatenate_and_time_sync_node.param.yaml#L4

amadeuszsz:
> @amadeuszsz I think a refactor PR to remove `has_static_tf_only` from autoware_launch is needed such as https://github.com/autowarefoundation/autoware_launch/blob/0b9caea24347f57a062f29b3524caef73a3295ff/sensor_kit/awsim_labs_sensor_kit_launch/awsim_labs_sensor_kit_launch/config/concatenate_and_time_sync_node.param.yaml#L4

@badai-nguyen 
It's already addressed in https://github.com/autowarefoundation/autoware_launch/pull/1393

knzo25:
@amadeuszsz 
Just to make it faster to get merged.
Can you list all the related PRs and issues ad the beginning of each PR?

badai-nguyen (approved):
LGTM regarding `compare_map_segmentation` and `ground_segmentation`!

sensing/autoware_pointcloud_preprocessor/src/concatenate_data/combine_cloud_handler.cpp
```diff
@@ -219,7 +217,9 @@ ConcatenatedCloudResult CombineCloudHandler::combine_pointclouds(
     convert_to_xyzirc_cloud(cloud, xyzirc_cloud);
 
     auto transformed_cloud_ptr = std::make_shared<sensor_msgs::msg::PointCloud2>();
-    managed_tf_buffer_->transform_pointcloud(output_frame_, *xyzirc_cloud, *transformed_cloud_ptr);
+    managed_tf_buffer_->transformPointcloud(
+      output_frame_, *xyzirc_cloud, *transformed_cloud_ptr, xyzirc_cloud->header.stamp,
+      rclcpp::Duration::from_seconds(1.0), node_.get_logger());
```
badai-nguyen, line 223: @amadeuszsz
Just a question, before change, `managed_tf_buffer_->transform_pointcloud`'s `has_static_only=false` as here https://github.com/autowarefoundation/autoware_launch/blob/d32ac7200f26685d9c00e01d6ca57d69cfc8ec69/sensor_kit/awsim_sensor_kit_launch/awsim_sensor_kit_launch/config/concatenate_and_time_sync_node.param.yaml#L4.
It mean that 
`managed_tf_buffer_->transform_pointcloud` will use latest timestamp for transformed but this you change to pointcloud's timestamp. 
Is this intentional?

amadeuszsz: @badai-nguyen 
Yes. Before I hardcoded time as `now()` without deeper investigation.  In old `autoware_pointcloud_preprocessor` implementation, we mostly used [pcl_ros](https://github.com/autowarefoundation/autoware_universe/blob/9bb228fe5b7fa4c6edb47e4713c73489a02366e1/sensing/autoware_pointcloud_preprocessor/src/concatenate_data/concatenate_and_time_sync_nodelet.cpp#L234), which takes stamp from [msg](https://github.com/ros-perception/perception_pcl/blob/humble/pcl_ros/src/transforms.cpp#L74-L77). However, our implementation got refactor, so now it is difficult to decide which stamp should we use. To reproduce latest behavior, we should put `now()` here, but I assumed it was a mistake forced by previous `managed_transform_buffer` implementation. 
@vividf as you understand inner algorithm well, could you please tell us please which stamp (`now()` or from input cloud msg) should we use?

vividf: @amadeuszsz I think either approach is fine—using now() or the message timestamp—since the transform from LiDAR to base_link is always the same. Personally, I prefer using the message timestamp since it makes more sense as we're applying the transformation to that specific point cloud.

PR #9200, opened 2024-10-30
feat(mission_planner): use allow_while_using_route

kosuke55:
## Description

use allow_while_using_route
https://autowarefoundation.github.io/autoware-documentation/main/design/autoware-interfaces/ad-api/features/routing/#allow_while_using_route

This is because there was a request to change the route only when it is not in use, so that it does not affect the behavior of the vehicle.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

- 2024/10/30 https://evaluation.tier4.jp/evaluation/reports/98f703d9-3601-5070-8c20-2e91c8ea0c88/?project_id=prd_jt
- not tested with this option because converter is not implemented

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/autoware_mission_planner/src/mission_planner/mission_planner.cpp
+++ b/planning/autoware_mission_planner/src/mission_planner/mission_planner.cpp
@@ -47,7 +47,6 @@ MissionPlanner::MissionPlanner(const rclcpp::NodeOptions & options)
   map_frame_ = declare_parameter<std::string>("map_frame");
   reroute_time_threshold_ = declare_parameter<double>("reroute_time_threshold");
   minimum_reroute_length_ = declare_parameter<double>("minimum_reroute_length");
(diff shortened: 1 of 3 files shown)


planning/autoware_mission_planner/src/mission_planner/mission_planner.cpp
```diff
@@ -242,9 +241,9 @@ void MissionPlanner::on_set_lanelet_route(
                               operation_mode_state_->is_autoware_control_enabled
                           : false;
 
-  if (is_reroute && !allow_reroute_in_autonomous_mode_ && is_autonomous_driving) {
+  if (is_reroute && !req->allow_while_using_route) {
```
isamu-takagi, line 244:
```suggestion
  if (is_reroute && is_autonomous_driving && !req->allow_while_using_route) {
```

kosuke55:
cc @isamu-takagi 
close this PR same to
https://github.com/tier4/tier4_autoware_msgs/pull/150

PR #9209, opened 2024-10-31
fix(out_of_lane): correct calculations of the stop pose

maxime-clem:
## Description

Fix bugs where the raw trajectory (which can start behind ego) was used instead of the filtered one (which always starts in front of ego).
Similarly, fix the incorrect use of the `first_ego_trajectory_idx` which is calculated on the raw trajectory but was used with the filtered trajectory.
Finally, fix the time measurements to keep using the last stop pose.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/motion_velocity_planner/autoware_motion_velocity_out_of_lane_module/src/calculate_slowdown_points.cpp
+++ b/planning/motion_velocity_planner/autoware_motion_velocity_out_of_lane_module/src/calculate_slowdown_points.cpp
@@ -130,20 +130,15 @@ std::optional<geometry_msgs::msg::Pose> calculate_slowdown_point(
 
 void calculate_min_stop_and_slowdown_distances(
   out_of_lane::EgoData & ego_data, const PlannerData & planner_data,
-  const std::optional<geometry_msgs::msg::Pose> & previous_slowdown_pose,
-  const double slow_velocity)
+  const std::optional<geometry_msgs::msg::Pose> & previous_slowdown_pose)
 {
   ego_data.min_stop_distance = planner_data.calculate_min_deceleration_distance(0.0).value_or(0.0);
-  ego_data.min_slowdown_distance =
-    planner_data.calculate_min_deceleration_distance(slow_velocity).value_or(0.0);
   if (previous_slowdown_pose) {
     // Ensure we do not remove the previous slowdown point due to the min distance limit
     const auto previous_slowdown_pose_arc_length = motion_utils::calcSignedArcLength(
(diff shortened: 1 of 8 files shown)


planning/motion_velocity_planner/autoware_motion_velocity_out_of_lane_module/src/calculate_slowdown_points.cpp
```diff
@@ -127,4 +127,26 @@ std::optional<geometry_msgs::msg::Pose> calculate_slowdown_point(
...
   return slowdown_point;
 }
+
+void calculate_min_stop_and_slowdown_distances(
+  out_of_lane::EgoData & ego_data, const PlannerData & planner_data,
+  const std::optional<geometry_msgs::msg::Pose> & previous_slowdown_pose,
+  const double slow_velocity)
+{
+  ego_data.min_stop_distance = planner_data.calculate_min_deceleration_distance(0.0).value_or(0.0);
+  ego_data.min_slowdown_distance =
```
mkquda, line 137: It seems that both `min_stop_distance` and `min_slowdown_distance` are only relevant within this function, so I wonder if they need to be part of `EgoData` struct.

Additionally it doesn't seem that `min_slow_distance` has any significance, its just being computed but not used anywhere.

planning/motion_velocity_planner/autoware_motion_velocity_out_of_lane_module/src/types.hpp
```diff
@@ -94,19 +94,24 @@ using OutLaneletRtree = bgi::rtree<LaneletNode, bgi::rstar<16>>;
 /// @brief data related to the ego vehicle
 struct EgoData
 {
-  std::vector<autoware_planning_msgs::msg::TrajectoryPoint> trajectory_points;
+  std::vector<autoware_planning_msgs::msg::TrajectoryPoint>
+    trajectory_points;  // filtered trajectory starting from the 1st point ahead of ego
```
mkquda, line 98: Considering this [line](https://github.com/autowarefoundation/autoware.universe/pull/9209/files#diff-7bd28123bdeeef46ff15191d21a1b2d98e41109269e99fad0158976bb9a372c5R156)
It seems the filtered trajectory starts from the first point of the nearest segment, which is most likely behind ego ... (unless original trajectory starts ahead of ego)

mkquda (approved):

PR #9202, opened 2024-10-30
feat(rtc_interface): add requested field

go-sakayori:
## Description
A new field `requested` was implemented in the `tier4_rtc_msgs`.
This PR modify the functions inside the rtc_interface taking into account of the new msg field.
Details are below:
- `updateCooperateStatus`: take `requested` as argument and set the default value to false (since most module won't use the requested field)
- `isActivated`: consider the state of requested field
- `isForceActivated`: consider the state of requested field

Furthermore, some refactoring and basic unit test were implemented.

## Related links
https://github.com/tier4/tier4_autoware_msgs/pull/148
https://github.com/orgs/autowarefoundation/discussions/4978

## How was this PR tested?
- [x] [TIER IV internal evaluator](https://evaluation.tier4.jp/evaluation/reports/778f2112-fa74-5378-b8c1-469e124d2cbb?project_id=prd_jt)
- [x] Colcon test
image

## Notes for reviewers
None.

## Interface changes
None.

## Effects on system behavior
None.

PR #9211, opened 2024-10-31
fix(rtc_interface): update requested field for every cooperateStatus state

go-sakayori:
## Description
The `requested` field was not updated, when the uuid was updated (it was updated when it was once registered).
This PR fixes the problem.

## Related links
https://github.com/autowarefoundation/autoware.universe/pull/9202

## How was this PR tested?
- [x] [TIER IV internal evaluator](https://evaluation.tier4.jp/evaluation/reports/1ea74b18-8e29-5d35-ab1d-98e2ca2d3bd3?project_id=prd_jt)

## Notes for reviewers
None.

## Interface changes
None.

## Effects on system behavior
None.


satoshi-ota (approved):
LGTM

PR #9224, opened 2024-11-01
fix(autoware_mpc_lateral_controller): fix bugprone-misplaced-widening-cast

kobayu858:
## Description
This is a fix based on clang-tidy `bugprone-misplaced-widening-cast` error.

```
/home/emb4/autoware/autoware/src/universe/autoware.universe/control/autoware_mpc_lateral_controller/src/lowpass_filter.cpp:128:16: error: either cast from 'int' to 'size_t' (aka 'unsigned long') is ineffective, or there is loss of precision before the conversion [bugprone-misplaced-widening-cast,-warnings-as-errors]
      tmp += u[static_cast<size_t>(i + j)];
               ^
```

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/control/autoware_mpc_lateral_controller/src/lowpass_filter.cpp
+++ b/control/autoware_mpc_lateral_controller/src/lowpass_filter.cpp
@@ -125,7 +125,7 @@ bool filt_vector(const int num, std::vector<double> & u)
     }
 
     for (int j = -num_tmp; j <= num_tmp; ++j) {
-      tmp += u[static_cast<size_t>(i + j)];
+      tmp += u[static_cast<size_t>(i) + static_cast<size_t>(j)];
       ++count;
     }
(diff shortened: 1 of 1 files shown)


kobayu858:
Other errors also exist, but they are not corrected at this time because of their low severity.

veqcc:
@kobayu858 
Your change seems to be incorrect. Please check again.

kobayu858:
@veqcc 
I forgot that it is a negative value. Thank you for pointing that out.

kobayu858:
The index is now calculated first to make sure it does not exceed the range before casting.

kobayu858:
@TakaHoribe @takayuki5168 

takayuki5168 (approved):

PR #9227, opened 2024-11-01
fix(autoware_euclidean_cluster): fix bugprone-misplaced-widening-cast

kobayu858:
## Description
This is a fix based on clang-tidy `bugprone-misplaced-widening-cast` error.
Other errors also exist, but they are not corrected at this time because of their low severity.

`max_cluster_size` is always assumed to be positive numbers.
Also, `cluster.point_step` is always a positive number because it is of type uint32.

```
/home/emb4/autoware/autoware/src/universe/autoware.universe/perception/autoware_euclidean_cluster/lib/voxel_grid_based_euclidean_cluster.cpp:120:31: error: either cast from 'int' to 'std::size_t' (aka 'unsigned long') is ineffective, or there is loss of precision before the conversion [bugprone-misplaced-widening-cast,-warnings-as-errors]
      if (cluster_data_size > std::size_t(max_cluster_size_ * point_step)) {
                              ^
```

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


badai-nguyen (approved):
LGTM

PR #9231, opened 2024-11-01
fix(autoware_behavior_path_sampling_planner_module): fix invalid parameter file

mitsudome-r:
## Description
I'm trying to see if I can make tier4_planning_launch able to launch without calling it from autoware_launch as a standalone launch file.

Through the process, I found out that the sample parameter file provided in the autoware_behavior_path_sampling_planner_module package was broken.
This copies the file from autoware_launch file so that people can use it as a default parameter setup.

## Related links

**Parent Issue:**

- https://github.com/autowarefoundation/autoware/issues/5394

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


youtalk (approved):
I've check the [sampling_planner.param.yaml](https://github.com/autowarefoundation/autoware_launch/blob/main/autoware_launch/config/planning/scenario_planning/lane_driving/behavior_planning/behavior_path_planner/sampling_planner/sampling_planner.param.yaml) in awf/autoware_launch. They are exactly the same.

youtalk:
@danielsanchezaran @maxime-clem Please review this.

maxime-clem (approved):

PR #9240, opened 2024-11-05
feat(object_lanelet_filter): add configurable margin for object lanel…

sebekx:
object_lanelet_filter

## Description
Added configurable margin to object lanelet filter which extends area of ​​non filtered objects.

With margin 3m:
image
image

## Related links
Related [autoware launch pr](https://github.com/autowarefoundation/autoware_launch/pull/1210) with configs  
**Parent Issue:**
None.

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_detected_object_validation/config/object_lanelet_filter.param.yaml
+++ b/perception/autoware_detected_object_validation/config/object_lanelet_filter.param.yaml
@@ -19,3 +19,5 @@
         enabled: false
         velocity_yaw_threshold: 0.785398 # [rad] (45 deg)
         object_speed_threshold: 3.0 # [m/s]
+      debug: false
+      lanelet_extra_margin: 0.0
--- a/perception/autoware_detected_object_validation/src/lanelet_filter/debug.cpp
+++ b/perception/autoware_detected_object_validation/src/lanelet_filter/debug.cpp
@@ -1,0 +1,119 @@
+// Copyright 2022 TIER IV, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "lanelet_filter.hpp"
+
+#include <boost/geometry/algorithms/convex_hull.hpp>
+#include <boost/geometry/algorithms/disjoint.hpp>
+#include <boost/geometry/algorithms/intersects.hpp>
+#include <boost/geometry/index/rtree.hpp>
+
+#include <lanelet2_core/geometry/BoundingBox.h>
+#include <lanelet2_core/geometry/Lanelet.h>
+#include <lanelet2_core/geometry/LaneletMap.h>
+#include <lanelet2_core/geometry/Polygon.h>
+
+namespace autoware::detected_object_validation
+{
+namespace lanelet_filter
+{
+
+using visualization_msgs::msg::Marker;
+using visualization_msgs::msg::MarkerArray;
+
+template <typename T>
+visualization_msgs::msg::Marker createPolygonMarker(
+  T polygon, rclcpp::Time stamp, std::string_view ns, size_t marker_id,
+  std_msgs::msg::ColorRGBA color)
+{
+  const constexpr std::string_view FRAME_ID = "map";
+
+  auto create_marker = [&](auto marker_type) {
+    Marker marker;
+    marker.id = marker_id;
+    marker.header.frame_id = FRAME_ID;
+    marker.header.stamp = stamp;
+    marker.ns = ns;
+    marker.type = marker_type;
+    marker.action = Marker::ADD;
+    marker.lifetime = rclcpp::Duration::from_seconds(0.1);
+    return marker;
+  };
+
+  auto marker = create_marker(Marker::LINE_LIST);
+  marker.pose.orientation.w = 1.0;
+  marker.scale.x = 0.1;
+  marker.color.a = color.a;
+  marker.color.r = color.r;
+  marker.color.g = color.g;
+  marker.color.b = color.b;
+
+  std::vector<geometry_msgs::msg::Point> points;
+  points.reserve(polygon.size());
+
+  for (size_t i = 0; i < polygon.size() - 1; ++i) {
+    geometry_msgs::msg::Point point;
+    point.x = polygon[i].x();
+    point.y = polygon[i].y();
+    points.push_back(point);
+
+    point.x = polygon[i + 1].x();
+    point.y = polygon[i + 1].y();
+    points.push_back(point);
+  }
+  geometry_msgs::msg::Point point;
+  point.x = polygon.back().x();
+  point.y = polygon.back().y();
+
+  points.push_back(point);
+  point.x = polygon.front().x();
+  point.y = polygon.front().y();
+
+  points.push_back(point);
+  marker.points = points;
+  return marker;
+}
+
+void ObjectLaneletFilterNode::publishDebugMarkers(
+  rclcpp::Time stamp, const LinearRing2d & hull, const std::vector<BoxAndLanelet> & lanelets)
+{
+  using visualization_msgs::msg::Marker;
+  using visualization_msgs::msg::MarkerArray;
+
+  uint8_t marker_id = 0;
+  Marker delete_marker;
+  constexpr std::string_view lanelet_range = "lanelet_range";
+  constexpr std::string_view roi = "roi";
+
+  MarkerArray marker_array;
+
+  std_msgs::msg::ColorRGBA color;
+  color.a = 0.5f;
+  color.r = 0.3f;
+  color.g = 1.0f;
+  color.b = 0.2f;
+  auto marker = createPolygonMarker(hull, stamp, lanelet_range, ++marker_id, color);
+  marker_array.markers.push_back(std::move(marker));
+  for (auto & box_and_lanelet : lanelets) {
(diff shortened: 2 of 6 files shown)


YoshiRi:
@badai-nguyen @technolojin Could you check this PR? (especially calculation cost)

a-maumau:
@sebekx Hi, thank you for your PR!

I timed the processing time by the [TimeKeeper](https://github.com/autowarefoundation/autoware.universe/tree/main/common/autoware_universe_utils).

| impl. | larger map | debug | lanelet_extra_margin | avg. processing time |
| ---- | ---- | ---- | ---- | ---- |
| current impl. | - | N/A |  N/A | ~6.0 ms |
| this PR | - | false |  0.0 | ~3.8 ms |
| this PR | - | false |  0.1 | ~12.6 ms |
| this PR | - | false |  0.5 | ~11.4ms |
| this PR | - | false |  1.0 | ~10.3 ms |
| this PR | - | false |  3.0 | ~8.5 ms |
| this PR | - | true |  0.0 | ~4.4 ms |
| current impl. | ✓ | N/A |  N/A | ~6.5 ms |
| this PR | ✓ | false |  0.0 | ~4.0 ms |
| this PR | ✓ | false |  0.1 | ~13.6 ms |
| this PR | ✓ | false |  3.0 | ~9.2 ms |
| this PR | - | false |  -1.0 | ~3.9 ms |

"larger map" in the table indicates that the same rosbag with a larger map, so the size of the map is the only difference. 
`debug`, `lanelet_extra_margin` is the parameter in the config file.

The implementation of this PR is faster than the current one when the `lanelet_extra_margin` is set 0.0 (disable).
(It should be coming from the reduction of the method call)

@sebekx since I am not familiar with the `bg::strategy::buffer`, do you know what kind of behavior is expected for `distance_strategy(<negative value>)`? Checking the lanelets with debug=true option in this PR, it seems it does not create the margin for the lanelets.

a-maumau:
Also, it would be nice if you create the update PR for the config file in the [autoware_launch](https://github.com/autowarefoundation/autoware_launch) too!

perception/autoware_detected_object_validation/src/lanelet_filter/lanelet_filter.cpp
```diff
@@ -73,6 +76,37 @@ ObjectLaneletFilterNode::ObjectLaneletFilterNode(const rclcpp::NodeOptions & nod
     std::make_unique<autoware::universe_utils::DebugPublisher>(this, "object_lanelet_filter");
   published_time_publisher_ =
     std::make_unique<autoware::universe_utils::PublishedTimePublisher>(this);
+  viz_pub_ =
+    this->create_publisher<visualization_msgs::msg::MarkerArray>("~/debug/marker", rclcpp::QoS{1});
+}
```
a-maumau, line 81: It might be better to create the publisher when the `filter_settings_.debug == true` since it only be used in `publishDebugMarkers` that is also only get called when `filter_settings_.debug == true`.

perception/autoware_detected_object_validation/src/lanelet_filter/lanelet_filter.cpp
```diff
@@ -59,7 +59,10 @@ ObjectLaneletFilterNode::ObjectLaneletFilterNode(const rclcpp::NodeOptions & nod
     declare_parameter<double>("filter_settings.lanelet_direction_filter.velocity_yaw_threshold");
   filter_settings_.lanelet_direction_filter_object_speed_threshold =
     declare_parameter<double>("filter_settings.lanelet_direction_filter.object_speed_threshold");
-
+  filter_settings_.debug = declare_parameter<bool>("filter_settings.debug");
+  filter_settings_.lanelet_extra_margin =
+    declare_parameter<double>("filter_settings.lanelet_extra_margin");
+  //
```
a-maumau, line 65: remove `//`

a-maumau:
@sebekx also, could you fix the error in the [cppcheck result](https://github.com/autowarefoundation/autoware.universe/actions/runs/11686560104/job/33179930601?pr=9240) ?

sebekx:
> @sebekx since I am not familiar with the `bg::strategy::buffer`, do you know what kind of behavior is expected for `distance_strategy(<negative value>)`? Checking the lanelets with `debug=true` option in this PR, it seems it does not create the margin for the lanelets.

@a-maumau Hi, thanks for review. Normally it will make polygon smaller but in [this](https://github.com/autowarefoundation/autoware.universe/pull/9240/files#diff-222b58890f45f2983bcbdd896de51ef5521a3d81d8efd55842fb6e7749fd589bR260-R262) implementation there is a condition to only expand polygons in whether there is positive value of extra margin.

perception/autoware_detected_object_validation/src/lanelet_filter/lanelet_filter.cpp
sebekx: removed

perception/autoware_detected_object_validation/src/lanelet_filter/lanelet_filter.cpp
sebekx: Added code that creates publisher only whether there is `filter_settings_.debug == true`.

sebekx:
> Also, it would be nice if you create the update PR for the config file in the [autoware_launch](https://github.com/autowarefoundation/autoware_launch) too!

there is [pr](https://github.com/autowarefoundation/autoware_launch/pull/1210)

sebekx:
> @sebekx also, could you fix the error in the [cppcheck result](https://github.com/autowarefoundation/autoware.universe/actions/runs/11686560104/job/33179930601?pr=9240) ?

fixed

perception/autoware_detected_object_validation/src/lanelet_filter/lanelet_filter.cpp
a-maumau: I think it still exist.

a-maumau:
> Hi, thanks for review. Normally it will make polygon smaller but in [this](https://github.com/autowarefoundation/autoware.universe/pull/9240/files#diff-222b58890f45f2983bcbdd896de51ef5521a3d81d8efd55842fb6e7749fd589bR260-R262) implementation there is a condition to only expand polygons whether there is positive value of extra margin.

Ah, I simply forgot that part, thank you.

I first thought shrinking might have some needs in some cases, but it would make some gaps between the lanelet connections, so this is reasonable.

perception/autoware_detected_object_validation/src/lanelet_filter/lanelet_filter.cpp
technolojin, line 65:
```suggestion

```

technolojin:
> > Also, it would be nice if you create the update PR for the config file in the [autoware_launch](https://github.com/autowarefoundation/autoware_launch) too!
> 
> there is [pr](https://github.com/autowarefoundation/autoware_launch/pull/1210)

Can you kindly add this link and description to this PR description?

technolojin:
I am going to test its performance.

technolojin:
I measured processing time in 3 different situation

A. Before 
B. After with `lanelet_extra_margin` of 3.0 
C. After with`lanelet_extra_margin` of zero

Screenshot from 2024-11-25 14-31-33

A: 5ms in low, 10ms of high load
B. 8ms in low, 19ms of high load
C. 3ms in low, 2ms of high load

In conclusion, the performance of the same usage is improved. If the extra margin is added, it causes computational cost (as expected).

technolojin (approved):
LGTM

technolojin:
@YoshiRi can you confirm this PR?

perception/autoware_detected_object_validation/config/object_lanelet_filter.param.yaml
```diff
@@ -19,3 +19,5 @@
         enabled: false
         velocity_yaw_threshold: 0.785398 # [rad] (45 deg)
         object_speed_threshold: 3.0 # [m/s]
+      debug: false
+      lanelet_extra_margin: 0.0
```
Shin-kyoto, line 23: @sebekx 

Could you update README and add the explanation for these params?

Shin-kyoto (changes requested):

YoshiRi (approved):
LGTM

perception/autoware_detected_object_validation/config/object_lanelet_filter.param.yaml
sebekx: updated readme

perception/autoware_detected_object_validation/src/lanelet_filter/debug.cpp
```diff
@@ -0,0 +1,126 @@
...
+  for (size_t i = 0; i < polygon.size() - 1; ++i) {
+    geometry_msgs::msg::Point point;
+    point.x = polygon[i].x();
+    point.y = polygon[i].y();
+    points.push_back(point);
+
+    point.x = polygon[i + 1].x();
+    point.y = polygon[i + 1].y();
+    points.push_back(point);
+  }
```
Shin-kyoto, line 79:
```suggestion
  std::vector<geometry_msgs::msg::Point> points;
  points.reserve(polygon.size() * 2);
  
  for (auto it = polygon.begin(); it != std::prev(polygon.end()); ++it) {
    geometry_msgs::msg::Point point;
    
    point.x = it->x();
    point.y = it->y();
    points.push_back(point);
    
    const auto next = std::next(it);
    point.x = next->x();
    point.y = next->y();
    points.push_back(point);
  }
```
@sebekx 
- reserved size looks like wrong.
- The iterator-based approach (it != std::prev(polygon.end())) provides better safety and container independence compared to index-based access.

Shin-kyoto (changes requested):

perception/autoware_detected_object_validation/src/lanelet_filter/debug.cpp
sebekx: thanks, it's reasonable

Shin-kyoto (approved):
Thanks! LGTM

NOTE: I did **NOT** check this PR by running this code.

technolojin:
@sebekx Thank you for your contribution!

PR #9247, opened 2024-11-06
fix(autonomous_emergency_braking): solve issue with arc length

danielsanchezaran:
## Description
This PR solves an issue that makes the AEB IMU path consider objects outside of the IMU path. The fix checks the arc length of the objects against the path to determine if it should be considered for collision checking or not.

Before: 
Points that are in front of IMU path are being considered for collision checking
image

After:

https://github.com/user-attachments/assets/617fb13b-0cae-41d9-832d-78826fe91c63

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/control/autoware_autonomous_emergency_braking/src/node.cpp
+++ b/control/autoware_autonomous_emergency_braking/src/node.cpp
@@ -900,11 +900,12 @@ void AEB::getClosestObjectsOnPath(
     return autoware::universe_utils::createPoint(p.x, p.y, p.z);
   }();
 
+  const auto path_length = autoware::motion_utils::calcArcLength(ego_path);
   for (const auto & p : *points_belonging_to_cluster_hulls) {
     const auto obj_position = autoware::universe_utils::createPoint(p.x, p.y, p.z);
     const double obj_arc_length =
       autoware::motion_utils::calcSignedArcLength(ego_path, current_p, obj_position);
-    if (std::isnan(obj_arc_length)) continue;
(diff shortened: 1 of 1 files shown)


control/autoware_autonomous_emergency_braking/src/node.cpp
```diff
@@ -900,11 +900,12 @@ void AEB::getClosestObjectsOnPath(
     return autoware::universe_utils::createPoint(p.x, p.y, p.z);
   }();
 
+  const auto path_length = autoware::motion_utils::calcArcLength(ego_path);
   for (const auto & p : *points_belonging_to_cluster_hulls) {
     const auto obj_position = autoware::universe_utils::createPoint(p.x, p.y, p.z);
     const double obj_arc_length =
       autoware::motion_utils::calcSignedArcLength(ego_path, current_p, obj_position);
```
kyoichi-sugahara, line 907: @danielsanchezaran 
Is there a possibility that the starting point of ego_path is located before the ego vehicle's position?
In that case, I think there would be a difference in distance between "from current_p to ego_path endpoint" and "from ego_path start point to endpoint".

danielsanchezaran: The paths are generated from the current ego pose so the first pose is the beggining of the path, however, I forgot about the longitudinal offset between base link and front of ego, it has been fixed! @kyoichi-sugahara

kyoichi-sugahara (approved):
LGTM!

Issue #9145, opened 2024-10-24
Add collision detector 


go-sakayori:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I've agreed with the maintainers that I can plan this task.

### Description

Autoware needs to detect if an object is in contact with the vehicle and trigger an emergency stop, ensuring that the vehicle remains stationary and does not move until the situation is resolved. To implement this using the existing functionality, it will be necessary to send diagnostics (diag) and trigger the Minimum Risk Maneuver (MRM) to achieve the desired behavior. The gap between the current functionality and the required behavior needs to be addressed.

### Purpose

Enhance the system's capability of checking for collisions between the vehicle and objects, triggering diagnostics, and executing the MRM.

### Possible approaches

Below are possible approaches:

- Add a new package `autoware_collision_detector` which detects if there has already been a collision (or about to collide). A prototype has been already developed in TIER IV which works well with the MRM.
- Modify the existing `autoware_obstacle_collision_checker` to detect actual contact between the vehicle and objects. Ensure it triggers diag messages and initiates MRM when contact is detected.

We plan to initially take the first approach and enhance the function of the package. As we enhance the new package, we would like to seek the integration with the existing packages.

### Definition of done
- [x] Add new package `autoware_collision_detector` under control in autoware.universe
- [x] Modify the launch file to ensure the new node launches (default would be off)
- [ ] Adjust the MRM settings to utilize the autoware_collision_detector when necessary

Merging packages would be out of scope for this time. (Remain as future work)

go-sakayori:
As the default setting is off the below task would be out of scope this time.
- Adjust the MRM settings to utilize the autoware_collision_detector when necessary

When the collision_detector is turned on, the above task would be tackled.

PR #9254, opened 2024-11-07
feat(system_diagnostic_monitor): add collision detector to control yaml file

go-sakayori:
## Description
This PR adds the collision_detector node information to the system diagnostic monitor.

## Related links
- https://github.com/autowarefoundation/autoware.universe/issues/9145
- https://github.com/autowarefoundation/autoware_launch/pull/1215

## How was this PR tested?
Could engage in PSim, and checked that the collision_detector diagnostic is disabled with `ros2 run diagnostic_graph_utils dump_node`.

## Notes for reviewers
None.

## Interface changes
None.

## Effects on system behavior
None.


go-sakayori:
As the collision_detector is off by default, there is no need to merge this PR.

PR #9260, opened 2024-11-08
fix(vehicle_cmd_gate): fix processing time measurement

maxime-clem:
## Description

Fix an issue where the processing time was incorrectly measured and published multiple times.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/control/autoware_vehicle_cmd_gate/src/vehicle_cmd_gate.cpp
+++ b/control/autoware_vehicle_cmd_gate/src/vehicle_cmd_gate.cpp
@@ -213,6 +213,8 @@ VehicleCmdGate::VehicleCmdGate(const rclcpp::NodeOptions & node_options)
     std::chrono::duration<double>(update_period));
   timer_ =
     rclcpp::create_timer(this, get_clock(), period_ns, std::bind(&VehicleCmdGate::onTimer, this));
+  timer_pub_status_ = rclcpp::create_timer(
+    this, get_clock(), period_ns, std::bind(&VehicleCmdGate::publishStatus, this));
 
   logger_configure_ = std::make_unique<autoware::universe_utils::LoggerLevelConfigure>(this);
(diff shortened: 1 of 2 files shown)


control/autoware_vehicle_cmd_gate/src/vehicle_cmd_gate.cpp
```diff
@@ -518,6 +516,14 @@ void VehicleCmdGate::onTimer()
     }
     gear_cmd_pub_->publish(gear);
   }
+
+  publishStatus();
```
tkimura4, line 520: I think that the implementation have changed because, when there is an early return due to conditions like !isDataReady() or is_emergency_state_heartbeat_timeout_, not all Status messages are output. Is this not an issue?

maxime-clem: You are right.
Then there are 2 options:
- `publishStatus()` is called in a separate timer callback (current implementation before this PR) but its runtime is not measured (this is probably fine).
- `publishStatus()` is called before every early returns.

tkimura4: > publishStatus() is called in a separate timer callback (current implementation before this PR) but its runtime is not measured (this is probably fine).

The processing within the publishStatus function is simple and lightweight, so I think this should be fine

maxime-clem: Done: a5a4d92b87

tkimura4 (approved):
LGTM

kosuke55:
cc @Kazunori-Nakajima 
please check this PR :pray:

PR #9265, opened 2024-11-08
feat(evaluator): Add Trajectory Evaluator

mraditya01:
## Description

Add trajectory evaluator to evaluate the accuracy of predicted or planned vehicle trajectories by comparing the expected versus actual arrival times at specific trajectory points.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?
Unit and random test.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/evaluator/autoware_trajectory_evaluator/src/trajectory_evaluator_node.cpp
+++ b/evaluator/autoware_trajectory_evaluator/src/trajectory_evaluator_node.cpp
@@ -34,86 +34,118 @@ void TrajectoryEvaluatorNode::onTimer()
     auto current_time = now();
 
     const auto ego_state_ptr = kinematic_state_sub_.takeData();
+    const auto traj_msg = traj_sub_.takeData();
+    onTrajectory(traj_msg, ego_state_ptr);
     onKinematicState(ego_state_ptr);
 
-    const auto traj_msg = traj_sub_.takeData();
-    onTrajectory(traj_msg);
 }
 
 
-void TrajectoryEvaluatorNode::onTrajectory(Trajectory::ConstSharedPtr traj_msg) {
+void TrajectoryEvaluatorNode::onTrajectory(const Trajectory::ConstSharedPtr traj_msg, const nav_msgs::msg::Odometry::ConstSharedPtr odom_msg) {
     if (traj_msg != nullptr) {
-        trajectory_history_.push_back({traj_msg, this->now()});
-        if (trajectory_history_.size() > 100) {
+        TrajectoryWithTimestamp trajectory_with_timestamp;
+        trajectory_with_timestamp.time_stamp = this->now();
+        const auto current_position = odom_msg->pose.pose.position;
+        std::vector<TrajectoryPointWithTime> trajectory_with_time;
+
+        for (const auto& point : traj_msg->points) {
+            TrajectoryPointWithTime point_with_time(point, rclcpp::Duration::from_seconds(0), trajectory_with_timestamp.time_stamp);
+            trajectory_with_timestamp.trajectory_points.push_back(point_with_time);
+        }
+
+        calculate_time_from_start(trajectory_with_timestamp, current_position, 1e-3);
+        trajectory_history_.push_back(trajectory_with_timestamp);
+        if (trajectory_history_.size() > 10) {
             trajectory_history_.erase(trajectory_history_.begin());
         }
     }
 }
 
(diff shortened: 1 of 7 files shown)


evaluator/autoware_trajectory_evaluator/src/trajectory_evaluator_node.cpp
```diff
@@ -0,0 +1,234 @@
...
+  }
+}
+
+void TrajectoryEvaluatorNode::on_kinematic_state(
+  const nav_msgs::msg::Odometry::ConstSharedPtr odom_msg)
+{
+  if (odom_msg != nullptr) {
+    const double velocity = std::sqrt(
+      std::pow(odom_msg->twist.twist.linear.x, 2) + std::pow(odom_msg->twist.twist.linear.y, 2) +
+      std::pow(odom_msg->twist.twist.linear.z, 2));
```
maxime-clem, line 133: it is enough to directly use `odom_msg->twist.twist.linear.x`

evaluator/autoware_trajectory_evaluator/src/trajectory_evaluator_node.cpp
```diff
@@ -0,0 +1,234 @@
...
+      to.time_from_start =
+        rclcpp::Duration::from_seconds(time_delta) + trajectory_points[idx - 1].time_from_start;
+    }
+  }
+}
+
+void TrajectoryEvaluatorNode::on_kinematic_state(
+  const nav_msgs::msg::Odometry::ConstSharedPtr odom_msg)
+{
+  if (odom_msg != nullptr) {
```
maxime-clem, line 130: Instead of having the whole function in the `if` block, you should do an early return.
```suggestion
  if (odom_msg == nullptr) {
    return;
  }
```

evaluator/autoware_trajectory_evaluator/src/trajectory_evaluator_node.cpp
```diff
@@ -0,0 +1,234 @@
...
+        const auto & trajectory_point_pose = trajectory.trajectory_points[i].trajectory_point.pose;
+        double distance = autoware::universe_utils::calcDistance2d(trajectory_point_pose, pose);
+
+        if (distance < min_distance) {
+          min_distance = distance;
+          closest_index = i;
+        }
+      }
+
+      if (min_distance < 1e-6) {
```
maxime-clem, line 155: if ego deviates from the trajectory by more than `1e-6`m then the error is not measured ?

evaluator/autoware_trajectory_evaluator/src/trajectory_evaluator_node.cpp
```diff
@@ -0,0 +1,234 @@
...
+      const geometry_msgs::msg::Pose & pose = odom_msg->pose.pose;
+      for (size_t i = 0; i < trajectory.trajectory_points.size(); ++i) {
+        const auto & trajectory_point_pose = trajectory.trajectory_points[i].trajectory_point.pose;
+        double distance = autoware::universe_utils::calcDistance2d(trajectory_point_pose, pose);
+
+        if (distance < min_distance) {
+          min_distance = distance;
+          closest_index = i;
+        }
+      }
```
maxime-clem, line 153: There is already a function `autoware::motion_utils::findNearestIndex()` to find the nearest index.
BUT you should calculate the expected time more accurately. Please calculate the arc length of the odom pose and interpolate the corresponding expected time.
You can use `autoware::motion_utils::findNearestSegmentIndex()`.

evaluator/autoware_trajectory_evaluator/src/trajectory_evaluator_node.cpp
```diff
@@ -0,0 +1,234 @@
...
+        metrics_msg_.metric_array.push_back(closest_traj_y);
+
+        MetricMsg current_pose_z;
+        current_pose_z.name = "trajectory_metrics/current_pose_z";
+        current_pose_z.value = pose.position.z;
+        metrics_msg_.metric_array.push_back(current_pose_z);
+
+        metrics_pub_->publish(metrics_msg_);
+
+        metrics_pub_->publish(metrics_msg_);
```
maxime-clem, line 218: You are publishing twice, and for each stored trajectory. 
This is too much so please only publish the metrics message once per iteration.

PR #9268, opened 2024-11-08
fix(lane_change): extending lane change path for multiple lane change (RT1-8427)

zulfaqar-azmi-t4:
## Description

Lane change path is not properly extended in multiple lane change case.
This has caused the stop point to not be inserted correctly, as shown in the following two videos.

#### Case 01

https://github.com/user-attachments/assets/d313afb5-47d7-4611-9fb9-56b1c725dbdc

#### Case 02

https://github.com/user-attachments/assets/b5adfee3-75fb-4c64-9839-16634b03f1d7

This PR aims to fix this by removing ego position check and properly extending path is there's no next lane.

#### Case 01 after PR

https://github.com/user-attachments/assets/0afca695-78db-40c4-a269-c713b05f7fc0

#### Case 02 after PR

https://github.com/user-attachments/assets/ae7cf59a-ccd1-4de3-970b-6c9c65bf3f3a

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

1. PSIM
2. Evaluator: TBA

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/scene.cpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/scene.cpp
@@ -595,13 +595,6 @@ std::optional<PathWithLaneId> NormalLaneChange::extendPath()
 {
   universe_utils::ScopedTimeTrack st(__func__, *time_keeper_);
   const auto path = status_.lane_change_path.path;
-  const auto lc_start_point = status_.lane_change_path.info.lane_changing_start.position;
-
-  const auto dist = calcSignedArcLength(path.points, lc_start_point, getEgoPosition());
-
-  if (dist < 0.0) {
-    return std::nullopt;
-  }
 
   auto & target_lanes = common_data_ptr_->lanes_ptr->target;
   const auto & transient_data = common_data_ptr_->transient_data;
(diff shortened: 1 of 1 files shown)


planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/scene.cpp
```diff
@@ -613,22 +606,22 @@ std::optional<PathWithLaneId> NormalLaneChange::extendPath()
     forward_path_length) {
     return std::nullopt;
   }
-
   const auto is_goal_in_target = getRouteHandler()->isInGoalRouteSection(target_lanes.back());
 
+  const auto dist_to_end_of_path =
+    lanelet::utils::getArcCoordinates(target_lanes, path.points.back().point.pose).length;
   if (is_goal_in_target) {
```
mkquda, line 613:
```suggestion
  const auto dist_to_end_of_path =
    lanelet::utils::getArcCoordinates(target_lanes, path.points.back().point.pose).length;

  if (common_data_ptr_->lanes_ptr->target_lane_in_goal_section) {
```

mkquda (approved):
LGTM.

PR #9270, opened 2024-11-08
refactor(goal_planner): remove reference_goal_pose getter/setter

soblin:
## Description

`referance_goal_pose` in GoalSearcher does not need to be saved as a state. So pass then in the  caller side using GoalSearcher.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

- https://evaluation.tier4.jp/evaluation/reports/c515d86c-910d-55e8-86c7-74f2f74c2ce8?project_id=prd_jt
- https://evaluation.tier4.jp/evaluation/reports/e1c8650a-f0d7-57b1-a0f2-83b58e079f62?project_id=prd_jt

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_path_planner/autoware_behavior_path_goal_planner_module/src/goal_planner_module.cpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_goal_planner_module/src/goal_planner_module.cpp
@@ -208,7 +208,6 @@ void GoalPlannerModule::onTimer()
   std::optional<BehaviorModuleOutput> last_previous_module_output_opt{std::nullopt};
   std::shared_ptr<OccupancyGridBasedCollisionDetector> occupancy_grid_map{nullptr};
   std::optional<GoalPlannerParameters> parameters_opt{std::nullopt};
-  std::shared_ptr<GoalSearcherBase> goal_searcher{nullptr};
 
   // begin of critical section
   {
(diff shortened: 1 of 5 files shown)


planning/behavior_path_planner/autoware_behavior_path_goal_planner_module/src/goal_planner_module.cpp
```diff
@@ -437,6 +436,8 @@ void GoalPlannerModule::onFreespaceParkingTimer()
     needPathUpdate(
       local_planner_data->self_odometry->pose.pose, path_update_duration, modified_goal_opt,
       parameters)) {
+    auto goal_searcher = std::make_shared<GoalSearcher>(parameters, vehicle_footprint);
+
```
kosuke55, line 440:
```suggestion
    const auto goal_searcher = std::make_shared<GoalSearcher>(parameters, vehicle_footprint);

```

kosuke55: Since it is now only used within planFreespacePath(), it may be ok to generate it within that function.

kosuke55 (approved):

planning/behavior_path_planner/autoware_behavior_path_goal_planner_module/src/goal_planner_module.cpp
kosuke55: (update() does not depend on search(), and there is no problem if we pass goal candidates. It seems like it would be ok to make update() a static method.)

PR #9241, opened 2024-11-05
refactor(tvm_utility) prefix package and namespace with autoware

esteve:
## Related links

**Parent Issue:**

- https://github.com/autowarefoundation/autoware/issues/5077
- https://github.com/autowarefoundation/autoware/issues/4569

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

PR #9291, opened 2024-11-11
chore(tvm_utility): remove tvm_utility package as it is no longer used

esteve:
## Description

The `tvm_utility` is no longer used.

## Related links

**Parent Issue:**

- https://github.com/autowarefoundation/autoware.universe/pull/9241#issuecomment-2457417234

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


ambroise-arm (approved):
@esteve Alongside this PR, the corresponding artifacts can also be removed from https://github.com/autowarefoundation/autoware/blob/main/ansible/roles/artifacts/tasks/main.yaml#L406-L430

esteve:
@ambroise-arm good catch. I've also submitted https://github.com/autowarefoundation/autoware/pull/5427 to remove those artifacts.

PR #9308, opened 2024-11-13
fix(collision_detector): skip process when odometry is not published

go-sakayori:
## Description
The collision detection published some error information when the vehicle position was not set in PSim.
```
[component_container_mt-39] [ERROR 1731399700.354588656] [control.collision_detector] filterObjects(): Failed to get transform from object frame to base_link:(L185)
```
This was caused due to not finding a connection between 'base_link' and 'map'.
This PR adds a subscriber for ego vehicle odometry information, and skips the process when the odometry information is not published.
As the `collision_detector` package is to be merged into the `surround_obstacle_checker` packages, subscribing the odometry information would not be a big issue, since the `surround_obstacle_checker`  is already subscribing this information.

## Related links
None

## How was this PR tested?
- [x] [TIER IV internal evaluator](https://evaluation.tier4.jp/evaluation/reports/f33be73c-95dc-5339-a46a-f6e3a3581e8f?project_id=prd_jt)
- [x] Confirmed that the error message does not appear before putting the vehicle in any position

## Notes for reviewers
None

#### Additions and removals

| Added| Sub | `~/input/odometry` | `nav_msgs::msg::Odometry`   | Ego vehicle odometry |

## Effects on system behavior
None.

### DIFF
--- a/control/autoware_collision_detector/include/autoware/collision_detector/node.hpp
+++ b/control/autoware_collision_detector/include/autoware/collision_detector/node.hpp
@@ -108,6 +108,8 @@ private:
   rclcpp::TimerBase::SharedPtr timer_;
 
   // publisher and subscriber
+  autoware::universe_utils::InterProcessPollingSubscriber<nav_msgs::msg::Odometry> sub_odometry_{
+    this, "~/input/odometry"};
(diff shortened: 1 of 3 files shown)


kyoichi-sugahara (approved):
Thank you! LGTM!

takayuki5168 (approved):

PR #9313, opened 2024-11-13
fix(behavior_path_planner): change module initial state to WAITING_APPROVAL

go-sakayori:
## Description
Each initial state of scene module in behavior path planner has been set to `RUNNING` by the `setInitState` function: therefore, the state would transfer from `IDLE` to `RUNNING`.
From the perspective of state transition, each module should transfer from `IDLE` to `WAITING_APPROVAL`.
This PR sets the initial state to `WAITING_APPROVAL` inside the `setInitState`.

## Related links
None

## How was this PR tested?
- [x] [TIER IV internal evaluator](https://evaluation.tier4.jp/evaluation/reports/bdff42fc-f7b7-5392-92e5-42fa0e166d4a?project_id=prd_jt)

## Notes for reviewers
None.

## Interface changes
None.

## Effects on system behavior
None.


soblin (approved):
Since LC is already overriding to transit `IDLE -> WAITING_APPROVAL -> RUNNING`, this is OK.

PR #9317, opened 2024-11-13
feat(autoware_test_utils): add parser for LaneletRoute, TrackedObject

soblin:
## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/common/autoware_test_utils/src/mock_data_parser.cpp
+++ b/common/autoware_test_utils/src/mock_data_parser.cpp
@@ -205,6 +205,17 @@ std::vector<LaneletSegment> parse(const YAML::Node & node)
 }
 
 template <>
+LaneletRoute parse(const YAML::Node & node)
+{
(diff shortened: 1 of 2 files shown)


common/autoware_test_utils/src/mock_data_parser.cpp
```diff
@@ -204,6 +204,17 @@ std::vector<LaneletSegment> parse(const YAML::Node & node)
   return segments;
 }
 
+template <>
+LaneletRoute parse(const YAML::Node & node)
```
zulfaqar-azmi-t4, line 208: there's already lanelet route parser in the mock data parser, just that it has different input, so i think you should refactor the code to reuse this function.

zulfaqar-azmi-t4 (approved):
Tested locally. LGTM!

PR #9319, opened 2024-11-14
fix(run_out): output velocity factor

satoshi-ota:
## Description

Set velocity factor when run out module inserts stop point.

Screenshot from 2024-11-14 11-29-49

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

Psim

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


danielsanchezaran (approved):
LGTM

PR #9322, opened 2024-11-14
refactor(lane_change): refactor extended object safety check

mkquda:
## Description

Currently when creating extended objects in lane change module, lane change prepare duration is assumed to be fixed (parameterized value).

In situations where the prepare duration used for candidate path generation does not match the assumed fixed value, such as when sampling multiple prepare durations near terminal, the mismatch can result in incorrect result for collision check causing an unsafe path to be approved.

This PR fixes the issue by refactoring extended object generation code and collision check code to work for candidate paths with varying prepare durations.

## Changes

- refactor LC utility function `transform` to generate predicted object polygons for the entire duration
- refactor LC function `is_collided` to filter the predicted object polygons based on the prepare duration of the candidate path being checked.

## Related links

None.

## How was this PR tested?
- PSIM
- [TIER IV Internal Evaluator](https://evaluation.tier4.jp/evaluation/reports/2bb5e5eb-4c90-5d2f-bc12-f79a264796c8?project_id=prd_jt)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/scene.cpp
```diff
@@ -1961,9 +1965,26 @@ bool NormalLaneChange::is_collided(
...
+    predicted_obj_path.confidence = obj_path.confidence;
+    std::copy_if(
+      obj_path.path.begin(), obj_path.path.end(), std::back_inserter(predicted_obj_path.path),
+      [&](const auto & entry) {
+        if (
+          entry.time < start_time ||
+          (entry.time < prepare_duration && entry.velocity < velocity_threshold)) {
+          return false;
+        }
+        return true;
```
zulfaqar-azmi-t4, line 1983:
```suggestion
        return !(
          entry.time < start_time ||
          (entry.time < prepare_duration && entry.velocity < velocity_threshold));
```

zulfaqar-azmi-t4 (approved):
LGTM

### Before

image

### After

image

PR #9338, opened 2024-11-15
fix(bpp): update collided polygon pose only once

zulfaqar-azmi-t4:
## Description

The collision pose is update at every iteration.
This has caused us to get inaccurate collision information when viewing the debug marker.
The important part is for us to know where the collision first happens.

### Before

image

The position shown is the **last collided pose**.

### After

image

The position shown is the **first collided pose**.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


go-sakayori (approved):
LGTM

PR #9342, opened 2024-11-17
feat(mission_planner): add processing time publisher

Kazunori-Nakajima:
## Description
The processing time of `mission_planner` and `route_selector` ware measured and made to Pub.

Sometimes the processing time is pubbed twice during route generation. 
This is once for on_clear_route() and once for route on_set_lanelet_route().

(TIER IV internal usecase)
This output is then read by the Basic scenario and tested daily for cycle failures.

## How was this PR tested?

- `ros2 topic echo /planning/mission_planning/mission_planner/debug/processing_time_ms`

- `ros2 topic echo /planning/mission_planning/route_selector/debug/processing_time_ms`

https://github.com/user-attachments/assets/821baeb7-799f-47a3-a8ce-cfeb7fe0fdf4

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


kosuke55 (approved):

isamu-takagi (commented):
The on_set_lanelet_route and on_set_waypoint_route functions may abort due to exceptions. In that case, the processing time will not be published, will this be OK?

Kazunori-Nakajima:
I would like to have the processing time output even in the case of an exception.
We have moved the process of pubbing processing_time to handle_exception in service_utils.
This allows us to output processing time even in the case of an exception.

isamu-takagi (approved):
LGTM

Kazunori-Nakajima:
- Codes are confusing and dirty.
- In case of error, it is not necessary to have it.
Because of the above two points, the processing time is not output in case of errors.

PR #9346, opened 2024-11-18
fix(factors_panel): sort by distance to stop/decel or point where it starts moving the steering

satoshi-ota:
## Description

Sort table col by distance.

Screenshot from 2024-11-18 09-53-04

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

Psim

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


h-ohta (approved):
LGTM

PR #9353, opened 2024-11-18
feat(mrm_comfortable_stop_operator): add updateParam for mrm comfortable stop

danielsanchezaran:
## Description
Add updateParam to facilitate comfortable stop param tuning and avoid relaunching autoware to test new params 

Note: I did not add the option to change the Hz of the module since I believe that might cause some troubles and it is best to not change it while Autoware is running

## Related links

**Parent Issue:**

- Link

## How was this PR tested?
PSim

https://github.com/user-attachments/assets/5f8f4841-b9c2-462e-8ef6-7190ab0fea22

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/system/mrm_comfortable_stop_operator/src/mrm_comfortable_stop_operator/mrm_comfortable_stop_operator_core.cpp
+++ b/system/mrm_comfortable_stop_operator/src/mrm_comfortable_stop_operator/mrm_comfortable_stop_operator_core.cpp
@@ -14,6 +14,8 @@
 
 #include "mrm_comfortable_stop_operator/mrm_comfortable_stop_operator_core.hpp"
 
+#include <autoware/universe_utils/ros/update_param.hpp>
+
 namespace mrm_comfortable_stop_operator
 {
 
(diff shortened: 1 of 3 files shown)


system/mrm_comfortable_stop_operator/src/mrm_comfortable_stop_operator/mrm_comfortable_stop_operator_core.cpp
```diff
@@ -65,6 +71,21 @@ void MrmComfortableStopOperator::operateComfortableStop(
   }
 }
 
+rcl_interfaces::msg::SetParametersResult MrmComfortableStopOperator::onParameter(
+  const std::vector<rclcpp::Parameter> & parameters)
+{
+  using autoware::universe_utils::updateParam;
+  updateParam<double>(parameters, "min_acceleration", params_.min_acceleration);
+  params_.min_acceleration = std::abs(params_.min_acceleration);
```
TomohitoAndo, line 79: Why do you use `std::abs` here? :thinking: 
It looks like [the original code](https://github.com/autowarefoundation/autoware.universe/pull/9353/files#diff-8cbb3f9e143a55d21060bfcc6cbd6bc985262024e022991f0ac39fb688187d17R27) doesn't do that.

danielsanchezaran: @TomohitoAndo my bad, I thought the original always changed it to abs value, but it was a mistake when looking at a different module. I have removed that line!

PR #9384, opened 2024-11-19
feat(pid_longitudinal_controller): suppress rclcpp_warning/error

takayuki5168:
## Description

Suppress the too-frequent MPC's warning which fills the terminal.
Instead, set a diag with a detailed reason, and show it on the terminal once as follows.
image

## How was this PR tested?
unit test passed perfectly, and psim worked as expected.
## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


yuki-takagi-66 (approved):
LGTM Thank you

PR #9385, opened 2024-11-19
fix(lane_change): check obj predicted path when filtering

zulfaqar-azmi-t4:
## Description

If there’s object that crosses predicted target lanes but it is not in target lane or target preceding lane, then the object will be treated as other lane object and will be ignored.

This PR aims to fix this.

### Before PR

The following video shows the use case.
If the object is not in cyan lane (current) or blue (target preceding), then it will be considered as others object (purple cube).

https://github.com/user-attachments/assets/40cbb71d-a3eb-48a0-9530-113d8e255a3e

### With PR

Object is now considered as target lane leading object (cyan cube)

https://github.com/user-attachments/assets/f781f984-6249-4c6f-8baa-56300dba6993

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

1. PSIM
2. Internal evaluator: [TIER IV Internal link](https://evaluation.tier4.jp/evaluation/reports/2e106800-248c-5576-8f53-37c0f5a710b0?project_id=prd_jt)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/include/autoware/behavior_path_lane_change_module/utils/utils.hpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/include/autoware/behavior_path_lane_change_module/utils/utils.hpp
@@ -382,5 +382,8 @@ bool filter_target_lane_objects(
   const double dist_ego_to_current_lanes_center, const bool ahead_of_ego,
   const bool before_terminal, TargetLaneLeadingObjects & leading_objects,
   ExtendedPredictedObjects & trailing_objects);
+
+bool has_path_overlapped_target_lanes(
+  const ExtendedPredictedObject & object, const lanelet::BasicPolygon2d & lanes_polygon);
 }  // namespace autoware::behavior_path_planner::utils::lane_change
 #endif  // AUTOWARE__BEHAVIOR_PATH_LANE_CHANGE_MODULE__UTILS__UTILS_HPP_
--- a/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/include/autoware/behavior_path_lane_change_module/utils/utils.hpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/include/autoware/behavior_path_lane_change_module/utils/utils.hpp
@@ -397,7 +397,7 @@ bool filter_target_lane_objects(
  * @return true if any of the object's predicted paths overlap with the lanes polygon, false
  * otherwise.
  */
-bool is_object_path_overlapped_lanes(
+bool object_path_overlaps_lanes(
   const ExtendedPredictedObject & object, const lanelet::BasicPolygon2d & lanes_polygon);
 }  // namespace autoware::behavior_path_planner::utils::lane_change
 #endif  // AUTOWARE__BEHAVIOR_PATH_LANE_CHANGE_MODULE__UTILS__UTILS_HPP_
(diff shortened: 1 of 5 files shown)


planning/behavior_path_planner/autoware_behavior_path_lane_change_module/include/autoware/behavior_path_lane_change_module/utils/utils.hpp
```diff
@@ -368,5 +368,8 @@ bool filter_target_lane_objects(
   const double dist_ego_to_current_lanes_center, const bool ahead_of_ego,
   const bool before_terminal, TargetLaneLeadingObjects & leading_objects,
   ExtendedPredictedObjects & trailing_objects);
+
+bool has_path_overlapped_target_lanes(
```
mkquda, line 372: name suggestion `object_path_overlaps_target_lanes`

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/utils/utils.cpp
```diff
@@ -35,6 +35,8 @@
 #include <autoware_lanelet2_extension/utility/utilities.hpp>
 #include <autoware_vehicle_info_utils/vehicle_info.hpp>
 #include <range/v3/algorithm/any_of.hpp>
+#include <range/v3/range/conversion.hpp>
+#include <range/v3/view/transform.hpp>
```
mkquda, line 39: can't you just `#include <ranges>`

mkquda (approved):
LGTM.

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/utils/utils.cpp
zulfaqar-azmi-t4: ranges library only available from c++20.
non c++20 uses range-v3.
also if we include all, potentially it will increase build time, so it might be better to just include what we use.

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/utils/utils.cpp
```diff
@@ -1206,9 +1201,16 @@ bool filter_target_lane_objects(
...
-      !boost::geometry::disjoint(object.initial_polygon, lanes_polygon.target);
-    if (in_target_lanes) {
+    const auto overlapped_target_lanes = std::invoke([&]() {
+      if (!boost::geometry::disjoint(object.initial_polygon, lanes_polygon.target)) {
+        return true;
+      }
+
+      return !is_stopped && is_vehicle(object.classification) &&
+             object_path_overlaps_lanes(object, lanes_polygon.target);
+    });
```
maxime-clem, line 1211: I think the lambda is unnecessary here.

```suggestion
    const auto overlapped_target_lanes = !boost::geometry::disjoint(object.initial_polygon, lanes_polygon.target) || (!is_stopped && is_vehicle(object.classification) && object_path_overlaps_lanes(object, lanes_polygon.target));
```

maxime-clem (approved):

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/utils/utils.cpp
maxime-clem: If you want to decompose the conditions, then using intermediate boolean variables is probably the best.

PR #9369, opened 2024-11-19
refactor(sensing): move CUDA related packages to `sensing/cuda` directory

youtalk:
## Description

Currently, in the awf/autoware [Dockerfile](https://github.com/autowarefoundation/autoware/blob/main/docker/Dockerfile), sensing/perception packages that are built with CUDA libraries are explicitly listed and built as follows. 

https://github.com/autowarefoundation/autoware/blob/main/docker/Dockerfile#L273-L284
```
RUN --mount=type=cache,target=${CCACHE_DIR} \
  # TODO(youtalk): Move CUDA related packages into a dedicated directory
  --mount=type=bind,from=rosdep-universe-sensing-perception-depend,source=/autoware/src/universe/autoware.universe/perception/autoware_bytetrack,target=/autoware/src/universe/autoware.universe/perception/autoware_bytetrack \
  --mount=type=bind,from=rosdep-universe-sensing-perception-depend,source=/autoware/src/universe/autoware.universe/perception/autoware_lidar_apollo_instance_segmentation,target=/autoware/src/universe/autoware.universe/perception/autoware_lidar_apollo_instance_segmentation \
  --mount=type=bind,from=rosdep-universe-sensing-perception-depend,source=/autoware/src/universe/autoware.universe/perception/autoware_lidar_centerpoint,target=/autoware/src/universe/autoware.universe/perception/autoware_lidar_centerpoint \
  --mount=type=bind,from=rosdep-universe-sensing-perception-depend,source=/autoware/src/universe/autoware.universe/perception/autoware_lidar_transfusion,target=/autoware/src/universe/autoware.universe/perception/autoware_lidar_transfusion \
  --mount=type=bind,from=rosdep-universe-sensing-perception-depend,source=/autoware/src/universe/autoware.universe/perception/autoware_shape_estimation,target=/autoware/src/universe/autoware.universe/perception/autoware_shape_estimation \
  --mount=type=bind,from=rosdep-universe-sensing-perception-depend,source=/autoware/src/universe/autoware.universe/perception/autoware_tensorrt_classifier,target=/autoware/src/universe/autoware.universe/perception/autoware_tensorrt_classifier \
  --mount=type=bind,from=rosdep-universe-sensing-perception-depend,source=/autoware/src/universe/autoware.universe/perception/autoware_tensorrt_common,target=/autoware/src/universe/autoware.universe/perception/autoware_tensorrt_common \
[...]

PR #9394, opened 2024-11-20
refactor(sensing): revert "move CUDA related packages to `sensing/cuda` directory"

youtalk:
Reverts autowarefoundation/autoware.universe#9369

I came up with a way to fix it without modifying the `autoware.universe` repository, so I will revert it.


amadeuszsz (approved):
LGTM

PR #9398, opened 2024-11-20
fix(autoware_obstacle_cruise_planner): fix clang-diagnostic-defaulted-function-deleted

kobayu858:
## Description
This is a fix based on clang-tidy `clang-diagnostic-defaulted-function-deleted` error.

```
/home/emb4/autoware/autoware/src/universe/autoware.universe/planning/autoware_obstacle_cruise_planner/include/autoware/obstacle_cruise_planner/planner_interface.hpp:69:3: error: explicitly defaulted default constructor is implicitly deleted [clang-diagnostic-defaulted-function-deleted]
  PlannerInterface() = default;
  ^
```

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/autoware_obstacle_cruise_planner/include/autoware/obstacle_cruise_planner/planner_interface.hpp
+++ b/planning/autoware_obstacle_cruise_planner/include/autoware/obstacle_cruise_planner/planner_interface.hpp
@@ -66,8 +66,6 @@ public:
       node.declare_parameter<double>("slow_down.moving_object_hysteresis_range");
   }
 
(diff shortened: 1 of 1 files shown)


kobayu858:
There are two ways to fix this: either remove the default constructor or have all member variables have a default constructor.
In this case, the default constructor is not used, so it was removed.
Confirmed that `colcon build` does not produce any errors.

takayuki5168 (approved):

PR #9421, opened 2024-11-21
feat(diagnostic_graph_utils): publish error graph instead of the terminal log

takayuki5168:
## Description

The diag_graph_utils's output to the terminal is a bit too frequent to fill the whole terminal screen.
Instead of the output to the terminal, this PR makes the diag_graph_utils publish the string message of the error diag graph so that the Rviz can visualize it as follows.
image

## How was this PR tested?

psim worked well

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/system/diagnostic_graph_utils/src/node/logging.cpp
+++ b/system/diagnostic_graph_utils/src/node/logging.cpp
@@ -61,7 +61,7 @@ void LoggingNode::on_timer()
   if (root_unit_ && root_unit_->level() != DiagUnit::DiagnosticStatus::OK) {
     dump_text_.str("");
     dump_text_.clear(std::stringstream::goodbit);
-    dump_unit(root_unit_, 0, "    ");
+    dump_unit(root_unit_, 0, "");
 
     if (enable_terminal_log_) {
       RCLCPP_WARN_STREAM(get_logger(), prefix_message << std::endl << dump_text_.str());
--- a/system/diagnostic_graph_utils/src/node/logging.cpp
+++ b/system/diagnostic_graph_utils/src/node/logging.cpp
@@ -57,14 +57,14 @@ void LoggingNode::on_create(DiagGraph::ConstSharedPtr graph)
 
 void LoggingNode::on_timer()
 {
-  static const auto message = "The target mode is not available for the following reasons:";
+  static const auto prefix_message = "The target mode is not available for the following reasons:";
   if (root_unit_ && root_unit_->level() != DiagUnit::DiagnosticStatus::OK) {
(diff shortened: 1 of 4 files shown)


system/diagnostic_graph_utils/src/node/logging.cpp
```diff
@@ -33,8 +33,13 @@ LoggingNode::LoggingNode(const rclcpp::NodeOptions & options) : Node("logging",
   sub_graph_.register_create_callback(std::bind(&LoggingNode::on_create, this, _1));
   sub_graph_.subscribe(*this, 1);
 
+  pub_error_graph_ =
+    create_publisher<tier4_debug_msgs::msg::StringStamped>("~/debug/error_graph", rclcpp::QoS(1));
```
takayuki5168, line 37: @isamu-takagi 
I temporarily used error_graph. Is there any official or better name?

isamu-takagi: Nothing in particular, but since it is a string type, something like `error_graph_text` might be better.

system/diagnostic_graph_utils/src/node/logging.cpp
```diff
@@ -52,12 +57,24 @@ void LoggingNode::on_create(DiagGraph::ConstSharedPtr graph)
...
 void LoggingNode::on_timer()
 {
-  static const auto message = "The target mode is not available for the following reasons:";
+  static const auto prefix_message = "The target mode is not available for the following reasons:";
   if (root_unit_ && root_unit_->level() != DiagUnit::DiagnosticStatus::OK) {
     dump_text_.str("");
     dump_text_.clear(std::stringstream::goodbit);
-    dump_unit(root_unit_, 0, "  ");
-    RCLCPP_WARN_STREAM(get_logger(), message << std::endl << dump_text_.str());
+    dump_unit(root_unit_, 0, "    ");
```
isamu-takagi, line 64: The indentation was redundant so please remove it.
```suggestion
    dump_unit(root_unit_, 0, "");
```

system/diagnostic_graph_utils/src/node/logging.cpp
takayuki5168: Fixed here
https://github.com/autowarefoundation/autoware.universe/pull/9421/commits/618244cd569275cbd17c6e2097a8f6f422db8f0a

takayuki5168:
@isamu-takagi 
Sorry to be late. Please again review the PR.

isamu-takagi (approved):
LGTM

PR #9426, opened 2024-11-21
fix(static_centerline_generator): several bug fixes

takayuki5168:
## Description

Several bug fixes including
- fix package dependency
- use discontinuous curvature calculation instead of spline interpolation which is more robust to the trajectory's interval length
- show steer angle instead of curvature which is easy to understand

## How was this PR tested?

unit test passed.
## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


takayuki5168:
@kosuke55 
Please approve the PR if the changes look good to you. No need to go deep into the changes.

kosuke55 (approved):

PR #9450, opened 2024-11-25
fix(surround_obstacle_checker)!: remove stop reason

satoshi-ota:
## Description

Don't use stop reason. This module uses velocity factor instead of it.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

Please confirm that this module doesn't output stop reason.

```
/planning/scenario_planning/status/stop_reasons
```

- [x] [TIER IV INTERNAL SCENARIOS](https://evaluation.tier4.jp/evaluation/reports/7fd68c7e-5f22-5b47-85c9-e829bdbc1e43?project_id=prd_jt)

## Notes for reviewers

None.

## Interface changes

After this PR, this module only outputs velocity factors.

## Effects on system behavior

None.

### DIFF
--- a/planning/autoware_surround_obstacle_checker/README.md
+++ b/planning/autoware_surround_obstacle_checker/README.md
@@ -34,8 +34,6 @@ else
 
 endif
 
-:Publish stop reason;
-
 stop
 @enduml
 ```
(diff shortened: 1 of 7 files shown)


go-sakayori (approved):
LGTM！

PR #9459, opened 2024-11-25
refactor(evaluators, autoware_universe_utils): rename Stat class to Accumulator and move it to autoware_universe_utils

xtk8532704:
## Description
Here is a class `Stat` used across evaluators for accumulating statistics, which is implemented repeatedly in each evaluator.
This PR do:

- Moved it to `autoware_universe_utils/math` and remove the implements inside evaluators.

- Renamed this class to `accumulator` (because this class is similar to <boost/accumulators/accumulators.hpp>).

Why do this:
This class has been commonly used, and will also be used for other nodes like processing_time_checker for metric statistics, so I do this refactoring.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?
unit test, psim, [evaluator](https://evaluation.tier4.jp/evaluation/reports/e6f0897f-ce9f-5f34-acdc-110de94cc4a0?project_id=prd_jt).

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


kosuke55 (approved):
LGTM,
this class will be used by processing_time_checker too

common/autoware_universe_utils/test/src/math/test_accumulator.cpp
```diff
@@ -0,0 +1,29 @@
...
+{
+  autoware::universe_utils::Accumulator<double> acc;
+  acc.add(1.0);
+  acc.add(2.0);
+  acc.add(3.0);
+
+  EXPECT_DOUBLE_EQ(acc.mean(), 2.0);
+  EXPECT_DOUBLE_EQ(acc.min(), 1.0);
+  EXPECT_DOUBLE_EQ(acc.max(), 3.0);
+}
```
kosuke55, line 29: The functions in universe_utils are used throughout autoware, so they need to be of a littile high quality. Could you add some more tests? For example, negative values.

common/autoware_universe_utils/include/autoware/universe_utils/math/accumulator.hpp
```diff
@@ -65,11 +65,11 @@ class Stat
   unsigned int count() const { return count_; }
 
   template <typename U>
-  friend std::ostream & operator<<(std::ostream & os, const Stat<U> & stat);
+  friend std::ostream & operator<<(std::ostream & os, const Accumulator<U> & accumulator);
 
 private:
   T min_ = std::numeric_limits<T>::max();
-  T max_ = std::numeric_limits<T>::min();
+  T max_ = std::numeric_limits<T>::lowest();
```
xtk8532704, line 72: I found a bug here we should use lowest instead of min, I also found there are many similar mistakes in the whole autoware.universe, I will remind others to fix them.

common/autoware_universe_utils/test/src/math/test_accumulator.cpp
xtk8532704: I added more tests, and by that, I found and fixed a meaningful bug. Thank you!

kosuke55 (approved):

common/autoware_universe_utils/include/autoware/universe_utils/math/accumulator.hpp
```diff
@@ -1,4 +1,4 @@
-// Copyright 2021 Tier IV, Inc.
+// Copyright 2024 Tier IV, Inc.
```
kosuke55, line 1:
```suggestion
// Copyright 2021-2024 Tier IV, Inc.
```
maybe better

kosuke55: or I think that just
```
// Copyright 2021 Tier IV, Inc.
```
is ok

takayuki5168 (approved):

kyoichi-sugahara (approved):

kminoda (approved):

PR #9462, opened 2024-11-25
refactor(autoware_autonomous_emergency_braking): add getObjectOnPathData and getLongitudinalOffset functions

kyoichi-sugahara:
## Description

I moved some functions to utility functions in a util file. No logic changes were made.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

- [x] run psim

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/control/autoware_autonomous_emergency_braking/src/node.cpp
+++ b/control/autoware_autonomous_emergency_braking/src/node.cpp
@@ -856,6 +856,7 @@ void AEB::createObjectDataUsingPredictedObjects(
         obj.position = obj_position;
         obj.velocity = obj_tangent_velocity;
         obj.distance_to_object = std::abs(dist_ego_to_object);
+        obj.is_target = true;
         object_data_vector.push_back(obj);
         collision_points_added = true;
       }
--- a/control/autoware_autonomous_emergency_braking/src/node.cpp
+++ b/control/autoware_autonomous_emergency_braking/src/node.cpp
@@ -934,7 +934,7 @@ void AEB::getClosestObjectsOnPath(
   }
 
   const auto longitudinal_offset_opt = utils::getLongitudinalOffset(
-    ego_path, vehicle_info_.front_overhang_m, vehicle_info_.rear_overhang_m);
+    ego_path, vehicle_info_.max_longitudinal_offset_m, vehicle_info_.rear_overhang_m);
 
   if (!longitudinal_offset_opt.has_value()) return;
   const auto longitudinal_offset = longitudinal_offset_opt.value();
--- a/control/autoware_autonomous_emergency_braking/src/node.cpp
+++ b/control/autoware_autonomous_emergency_braking/src/node.cpp
@@ -815,6 +815,13 @@ void AEB::createObjectDataUsingPredictedObjects(
   const auto transform_stamped_opt =
     utils::getTransform("base_link", predicted_objects_ptr_->header.frame_id, tf_buffer_, logger);
   if (!transform_stamped_opt.has_value()) return;
+
+  const auto longitudinal_offset_opt = utils::getLongitudinalOffset(
+    ego_path, vehicle_info_.max_longitudinal_offset_m, vehicle_info_.rear_overhang_m);
+
+  if (!longitudinal_offset_opt.has_value()) return;
+  const auto longitudinal_offset = longitudinal_offset_opt.value();
+
   // Check which objects collide with the ego footprints
   std::for_each(objects.begin(), objects.end(), [&](const auto & predicted_object) {
     // get objects in base_link frame
@@ -842,17 +849,13 @@ void AEB::createObjectDataUsingPredictedObjects(
 
         // If the object is behind the ego, we need to use the backward long offset. The
         // distance should be a positive number in any case
-        const bool is_object_in_front_of_ego = obj_arc_length > 0.0;
-        const double dist_ego_to_object =
-          (is_object_in_front_of_ego) ? obj_arc_length - vehicle_info_.max_longitudinal_offset_m
-                                      : obj_arc_length + vehicle_info_.min_longitudinal_offset_m;
+        const double dist_ego_to_object = obj_arc_length - longitudinal_offset;
 
         ObjectData obj;
         obj.stamp = stamp;
         obj.position = obj_position;
         obj.velocity = obj_tangent_velocity;
         obj.distance_to_object = std::abs(dist_ego_to_object);
-        obj.is_target = true;
         object_data_vector.push_back(obj);
         collision_points_added = true;
       }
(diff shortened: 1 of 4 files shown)


control/autoware_autonomous_emergency_braking/src/node.cpp
```diff
@@ -842,17 +849,13 @@ void AEB::createObjectDataUsingPredictedObjects(
...
-          (is_object_in_front_of_ego) ? obj_arc_length - vehicle_info_.max_longitudinal_offset_m
-                                      : obj_arc_length + vehicle_info_.min_longitudinal_offset_m;
+        const double dist_ego_to_object = obj_arc_length - longitudinal_offset;
 
         ObjectData obj;
         obj.stamp = stamp;
         obj.position = obj_position;
         obj.velocity = obj_tangent_velocity;
         obj.distance_to_object = std::abs(dist_ego_to_object);
-        obj.is_target = true;
```
danielsanchezaran, line 855: Please dont delete this line. It is not require, but makes the code more explicit.

danielsanchezaran (commented):
Could you please check for degradation @kyoichi-sugahara ?

control/autoware_autonomous_emergency_braking/src/node.cpp
kyoichi-sugahara: @danielsanchezaran 
> Please dont delete this line. It is not require, but makes the code more explicit.

I don't intend to delete this line, This line is deleted in your refactoring branch. 
But ok :+1: and I agree.

kyoichi-sugahara:
@danielsanchezaran 
> Could you please check for degradation
you mean in CI/CD?

control/autoware_autonomous_emergency_braking/src/node.cpp
kyoichi-sugahara: added in https://github.com/autowarefoundation/autoware.universe/pull/9462/commits/8846adaea8fb23731cade10207c7506636509c72

danielsanchezaran:
There seems to be an issue with the end of the predicted path? @kyoichi-sugahara 

https://github.com/user-attachments/assets/c87756fd-9732-45d9-98fe-7c987bfbd5f3

danielsanchezaran:
normal OSS autoware does not seem to have that issue 

https://github.com/user-attachments/assets/3368a26c-7cf0-4207-96f1-442fd5a99f18

kyoichi-sugahara:
@danielsanchezaran 
> There seems to be an issue with the end of the predicted path?

I couldn't understand issue directly sorry :pray: 
Can you specify what's the issue?

danielsanchezaran:
> @danielsanchezaran
> 
> > There seems to be an issue with the end of the predicted path?
> 
> I couldn't understand issue directly sorry 🙏 Can you specify what's the issue?

Did you watch the video? There is no AEB activation when we set a bus over the goal pose

kyoichi-sugahara:
@danielsanchezaran 

>yes, scenario evaluation

I started to run the simulation :+1: 

I did some debugging and it seems all points are being filtered out by this processing:

```
    obj_arc_length > path_length + longitudinal_offset) {
    return {};
  }
```
You can see the attached video.


https://github.com/user-attachments/assets/4f03864f-990a-4fa4-8efd-1bb8d2cbc49a


Regarding the difference between before and after the changes - while it may not be the intended behavior, from the perspective of excluding points that exceed the path length, the behavior after the change seems more correct. What do you think?

By the way, when I commented that "I couldn't understand issue directly sorry" I meant that after watching the video, I couldn't understand why the behavior was problematic. Sorry if that wasn't clear.
I can provide route files that can reproduce the issue in `recrete_p_sim_scenario` along with debug print differences if needed - please let me know if you'd like those.

danielsanchezaran:
> @danielsanchezaran
> 
> > yes, scenario evaluation
> 
> I started to run the simulation 👍
> 
> I did some debugging and it seems all points are being filtered out by this processing:
> 
> ```
>     obj_arc_length > path_length + longitudinal_offset) {
>     return {};
>   }
> ```
> 
> You can see the attached video.
> 
>  2024-11-26_19.30.10.mp4 
> Regarding the difference between before and after the changes - while it may not be the intended behavior, from the perspective of excluding points that exceed the path length, the behavior after the change seems more correct. What do you think?
> 
> By the way, when I commented that "I couldn't understand issue directly sorry" I meant that after watching the video, I couldn't understand why the behavior was problematic. Sorry if that wasn't clear. I can provide route files that can reproduce the issue in `recrete_p_sim_scenario` along with debug print differences if needed - please let me know if you'd like those.

the object arc_length should not be larger than the path length + longitudinal offset, you can visually see the predicted path going inside the parked bus. One of the variables (path length , longitudinal offset, object arc length) is wrong I believe

control/autoware_autonomous_emergency_braking/src/node.cpp
```diff
@@ -928,53 +932,21 @@ void AEB::getClosestObjectsOnPath(
...
-  const bool ego_is_driving_forward = ego_is_driving_forward_opt.value();
-  // select points inside the ego footprint path
-  const auto current_p = [&]() {
-    const auto & first_point_of_path = ego_path.front();
-    const auto & p = first_point_of_path.position;
-    return autoware::universe_utils::createPoint(p.x, p.y, p.z);
-  }();
 
+  const auto longitudinal_offset_opt = utils::getLongitudinalOffset(
+    ego_path, vehicle_info_.front_overhang_m, vehicle_info_.rear_overhang_m);
```
danielsanchezaran, line 937:
```suggestion
  const auto longitudinal_offset_opt = utils::getLongitudinalOffset(
    ego_path, vehicle_info_.max_longitudinal_offset_m, vehicle_info_.rear_overhang_m);
```
Found the issue. Not sure what the front overhang is supposed to represent, but its value is 1 meter which makes no sense. We want the distance from base link to the front of the object

kyoichi-sugahara: Nice!

>Not sure what the front overhang is supposed to represent,

`front_overhang` is the distance between front of the vehicle and front wheel link as attached in the screenshot.

image

danielsanchezaran:
@kyoichi-sugahara I will approve this PR, the changes look good, but please fix the front_overhang and use the max_longitudinal_offset_m before merging!

danielsanchezaran (approved):

control/autoware_autonomous_emergency_braking/src/node.cpp
kyoichi-sugahara: Thank you so much! addressed in https://github.com/autowarefoundation/autoware.universe/pull/9462/commits/5260283861e9c3f1d81432d1b7298561dbad4c20

Issue #7924, opened 2024-07-09
`traffic_light_roi_visualizer` does not handle unknown results well


miursh:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I'm convinced that this is not my fault but a bug.

### Description

When the TLR outputs an unknown result, the visualizer:
- Keeps outputting 'Adjusted position is out of image bounds.' to the terminal
![image](https://github.com/autowarefoundation/autoware.universe/assets/37187849/439cdb95-f6a8-403e-a7c4-b87193ff9ef2)

- The border color of 'unknown' is green
![image](https://github.com/autowarefoundation/autoware.universe/assets/37187849/277f8b44-cac8-40db-9f10-e9f584415ad1)

### Expected behavior

When the TLR outputs an unknown result, the visualizer:
- Should not output any errors to the terminal
- The border color of 'unknown' should be white

### Actual behavior

When the TLR outputs an unknown result, the visualizer:
- Keeps outputting 'Adjusted position is out of image bounds.' to the terminal
- Turns the border color of 'unknown' to green

### Steps to reproduce

1. launch logging_simulator.launch.xml `ros2 launch autoware_launch logging_simulator.launch.xml map_path:=PATH_TP_MAP vehicle_model:=sample_vehicle sensor_model:=sample_sensor_kit vehicle:=false sensing:=true perception:=true planning:=false control:=false map:=true system:=false localization:=false rviz:=false`
2. run rqt_image_view `ros2 run rqt_image_view rqt_image_view ` and see `~/debug/rois` topic

### Versions

OS: Ubuntu22.04
ROS2: humble
Autoware: latest

### Possible causes

The condition in this section is not good.
https://github.com/autowarefoundation/autoware.universe/blob/main/perception/traffic_light_visualization/src/traffic_light_roi_visualizer/shape_draw.cpp#L50-L52

### Possible approaches

- [x] Very short term: Remove std::cerr https://github.com/autowarefoundation/autoware.universe/pull/8294
- [ ] Short term: Fix the condition for drawing unknown
- [ ] Medium to long term: Refactor overall (there are mysterious magic numbers and many variables with essentially the same name, fixed values set in various places within hpp/cpp files and global functions)

vividf:
@miursh 
Miura-san, may I ask who is handling this issue?

[...]

PR #5934, opened 2023-12-22
feat(traffic_light): output undetected traffic light as unknown

tzhong518:
## Description

Previously, only the successfully detected traffic light signals are output. 
In response to the request of the Planning, as long as the rough/roi exists in current frame, the corresponding signal will also exist in the traffic signal message. If it is not detected by the detector due to occlusion etc., the x_offset, y_offset, width, height will be 0 and the class will be unknown.

## Related links

ticket: https://tier4.atlassian.net/browse/RT1-4600

## Tests performed

Screenshot from 2023-12-22 13-55-15

## Pre-review checklist for the PR author

The PR author **must** check the checkboxes below when creating the PR.

- [ ] I've confirmed the [contribution guidelines].
- [ ] The PR follows the [pull request guidelines].

## In-review checklist for the PR reviewers

The PR reviewers **must** check the checkboxes below before approval.

- [ ] The PR follows the [pull request guidelines].
- [ ] The PR has been properly tested.
- [ ] The PR has been reviewed by the code owners.

## Post-review checklist for the PR author

The PR author **must** check the checkboxes below before merging.

- [ ] There are no open discussions or they are tracked via tickets.
- [ ] The PR is ready for merge.

After all checkboxes are checked, anyone who has write access can merge the PR.

[contribution guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/
[pull request guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/pull-request-guidelines/

PR #9467, opened 2024-11-26
fix(traffic_light_roi_visualizer): show unknown results correctly

tzhong518:
## Description

This PR fixes the following issue: https://github.com/autowarefoundation/autoware.universe/issues/7924

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

image
The rough roi of unknown traffic light is shown in white.
The error message resulted by the undetected traffic lights is fixed, so that it will not flood the terminal.

## Notes for reviewers

https://github.com/autowarefoundation/autoware.universe/pull/5934
Previously, it was required to output the signal message as long as the rough roi exists. 
I'm afraid there exists some misunderstanding here, as a result `has_correspond_traffic_signal` and `has_correspond_roi` at [ this part ](https://github.com/autowarefoundation/autoware.universe/blob/main/perception/autoware_traffic_light_visualization/src/traffic_light_roi_visualizer/node.cpp#L208-L225) will always be True.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_traffic_light_visualization/src/traffic_light_roi_visualizer/shape_draw.cpp
+++ b/perception/autoware_traffic_light_visualization/src/traffic_light_roi_visualizer/shape_draw.cpp
@@ -29,6 +29,10 @@ void drawShape(
   cv::Mat & image, const std::vector<ShapeImgParam> & params, int size, const cv::Point & position,
   const cv::Scalar & color, float probability)
 {
+  // skip if the roi position is set as (0,0), which means it is undetected
+  if (position.x == 0 && position.y == 0) {
+    return;
+  }
   // load concatenated shape image
   const auto shape_img = loadShapeImage(params, size);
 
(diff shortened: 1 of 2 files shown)


perception/autoware_traffic_light_visualization/src/traffic_light_roi_visualizer/shape_draw.cpp
```diff
@@ -42,14 +46,13 @@ void drawShape(
...
   const cv::Point rect_position(position.x, position.y - fill_rect_h);
-
   if (
     rect_position.x < 0 || rect_position.y < 0 || rect_position.x + fill_rect_w > image.cols ||
     position.y > image.rows) {
     // TODO(KhalilSelyan): This error message may flood the terminal logs, so commented out
     // temporarily. Need to consider a better way.
 
-    // std::cerr << "Adjusted position is out of image bounds." << std::endl;
+    std::cerr << "Adjusted position is out of image bounds." << std::endl;
```
yukkysaito, line 55: It seems better to use ROS_WARN instead of std::cerr.

MasatoSaeki: @yukkysaito 
This source code does not depend on ROS. So using `cerr` is better than `ROS_WARN` adding ros header, I think. What do you think?

MasatoSaeki (approved):
LGTM

PR #9480, opened 2024-11-26
feat(lane_change): improve delay lane change logic

mkquda:
## Description

Lane change module has a feature to allow delaying of LC when there is parked along the target lane ahead of ego, however the current implementation has some drawbacks.
In the current implementation, we only check the front most static object along the target lane and delay is applied only if:
- The object if considered parked (stopped to the side of the lane)
- Distance from object to terminal is sufficient to do LC

This approach limits the situations under which delay feature can be useful.

To help improve the robustness of delay feature and increase its usefulness, the following method is proposed.
For all static objects ahead of ego along the target lane, we check each object from closest to furthest, if any object passes the following checks, LC will be delay:
- Distance from object to next static object is sufficient to do LC
- Distance from object to terminal is sufficient to do LC
- Object is parked (if parameter to only consider parked objects is enabled).

## Changes
- Implement new function `is_delay_lane_change`
- Add new struct `DelayParameters`
- Refactor LC parameters related to delay feature
- Refactor function `isParkedObject`
- Remove unused functions `passed_parked_objects` & `getLeadingStaticObjectIdx`

## Related links

- [Launcher PR](https://github.com/autowarefoundation/autoware_launch/pull/1256)

## How was this PR tested?
- PSIM
- [TIER IV Internal Evaluator](https://evaluation.tier4.jp/evaluation/reports/85316b75-190c-54c2-9aa4-6fd0072ded7d?project_id=prd_jt)

## Notes for reviewers

None.

#### Additions and removals

| Added | `delay_lane_change.enabled`   | `bool` | `true`         | Enable lane change delay feature |
| Added | `delay_lane_change.check_only_parked_vehicle`   | `bool` | `false`         | Only consider parked vehicles for delay lane change feature |

#### Modifications

| Old     | `object_check_min_road_shoulder_width` | `double` | `0.5`         | Width considered as a road shoulder if the lane does not have a road shoulder |
| New     | `delay_lane_change.min_road_shoulder_width` | `double` | `0.5`         | Width considered as a road shoulder if the lane does not have a road shoulder |
| Old     | `object_shiftable_ratio_threshold` | `double` | `0.6`         | Vehicles around the center line within this distance ratio will be excluded from parking objects |
| New     | `delay_lane_change.th_parked_vehicle_shift_ratio` | `double` | `0.6`         | Vehicles around the center line within this distance ratio will be excluded from parking objects |

## Effects on system behavior

Delay lane change feature is more robust, and can be configured to apply for all static objects or only or parked objects.
Less chance of Ego doing lane change before stopped vehicle and getting stuck.


planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/utils/utils.cpp
```diff
@@ -776,129 +739,65 @@ bool isParkedObject(
...
-  std::optional<size_t> leading_obj_idx = std::nullopt;
-  for (size_t obj_idx = 0; obj_idx < objects.size(); ++obj_idx) {
-    const auto & obj = objects.at(obj_idx);
-    const auto & obj_pose = obj.initial_pose;
+  const auto ego_vel = common_data_ptr->get_ego_speed();
+  const auto min_lon_acc = common_data_ptr->lc_param_ptr->trajectory.min_longitudinal_acc;
+  const auto gap_threshold = std::abs((ego_vel * ego_vel) / (2 * min_lon_acc));
+  auto is_sufficient_gap = [&gap_threshold](
+                             const ExtendedPredictedObject & current_obj,
+                             const ExtendedPredictedObject & next_obj) {
```
zulfaqar-azmi-t4, line 767:
```suggestion
                             const auto & current_obj,
                             const auto & next_obj) {
```

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/utils/utils.cpp
```diff
@@ -776,129 +739,65 @@ bool isParkedObject(
   return clamped_ratio > ratio_threshold;
 }
 
-bool passed_parked_objects(
+bool is_delay_lane_change(
   const CommonDataPtr & common_data_ptr, const LaneChangePath & lane_change_path,
-  const std::vector<ExtendedPredictedObject> & objects, CollisionCheckDebugMap & object_debug)
+  const std::vector<ExtendedPredictedObject> & target_objects,
```
zulfaqar-azmi-t4, line 744:
```suggestion
  const ExtendedPredictedObjects & target_objects,
```

PR #9486, opened 2024-11-27
chore(autoware_lidar_transfusion): added a warning if we are dropping voxels

knzo25:
## Description
If the input pointcloud generates more voxels than the model that handle, we are currently clipping the number of voxels.
If that happens, we should notify users so they can consider increasing the model parameters.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


amadeuszsz (approved):
LGTM

PR #9487, opened 2024-11-27
fix(start_planner): use extended current lanes to fix turn signal issue

danielsanchezaran:
## Description
Current start planner uses CurrentLanes to calculate the path's shift length and decide if the turn signal should be activated and if it should be the left or right turn signal. 

However, this method disregards the previous lanelets, which might cause the shift length's sign to change when the ego enters a new lanelet. 

In the following example, the left turn signal is wrongly activated.
[Screencast from 2024年10月18日 15時24分50秒.webm](https://github.com/user-attachments/assets/99ac2300-49e2-411d-8702-fca980c555d6)

By using the extended CurrentLanes instead, this issue is solved: 

https://github.com/user-attachments/assets/6cbfe315-0b7f-44ea-a047-80b1bdd1aa6f

## Related links
ticket:  ([TIER IV INTERNAL LINK](https://tier4.atlassian.net/browse/RT0-33950))
**Parent Issue:**

- Link

## How was this PR tested?
PSim
## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


kosuke55:
it may be better to keep the lanes for turn signal caluculation?

kosuke55 (approved):

PR #9518, opened 2024-11-28
feat(start_planner): add current lanes as member var

danielsanchezaran:
## Description
Current lanelets are computed around 7-8 times in the code, by moving them as part of start plannerdata, we reduce code size and reduce possible errors. performance might improve very lightly.
## Related links

**Parent Issue:**

- Link

## How was this PR tested?
PSim
## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


kyoichi-sugahara (approved):
code change LGTM! If there is no degradation in the scenario evaluation's result.
Please change to appropriate conventional-commit-types with refactor from feat.

.github/_CODEOWNERS
mitsudome-r, line 1: @danielsanchezaran Could you revert this renaming?

PR #9520, opened 2024-11-28
feat(build-and-test-differential): if unnecessary, completely ignore cuda build

yhisaki:
### Motivation

Previously, we configured PRs without CUDA-related changes to skip the build-and-test step. However, even just pulling the CUDA image or using the self-hosted runner can sometimes cause errors, which may block PRs unrelated to CUDA. This PR prevents pulling the CUDA image and using the self-hosted runner for PRs that are unrelated to CUDA.

### Modified/Added jobs

- prepare-build-and-test-differential
  - Checks if CUDA build is required and sets `cuda_build` output
  - Fails if `tag:build-and-test-differential` label is not present
- build-and-test-differential
  - Non-CUDA build that runs on both CUDA-related and non-CUDA-related PRs
  - Runs on CUDA-related PRs only
- clang-tidy-differential
  - Runs when CUDA build is not required (`cuda_build=false`)
  - Uses `ubuntu-latest` runner and `ghcr.io/autowarefoundation/autoware:universe-devel` container
- clang-tidy-differential-cuda
  - Runs when CUDA build is required (`cuda_build=true`)

> [!IMPORTANT]
> **If this PR is merged, all of the jobs listed above need to be set as required.** [Since skipped jobs are treated as "Success"](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/using-conditions-to-control-job-execution), this should work correctly.

### Behavior details

- If `tag:build-and-test-differential` does not exist ([Github Actions](https://github.com/autowarefoundation/autoware.universe/actions/runs/12115902576?pr=9520)) → Fail `prepare-build-and-test-differential`

   Screenshot from 2024-12-02 18-04-50

- When a package unrelated to cuda is modified ([Github Actions](https://github.com/autowarefoundation/autoware.universe/actions/runs/12115974337?pr=9520))

   Screenshot from 2024-12-02 18-11-56

- When a package related to cuda is changed ([Github Actions](https://github.com/autowarefoundation/autoware.universe/actions/runs/12115974337?pr=9520))

   image

- If both are changed ([Github Actions](https://github.com/autowarefoundation/autoware.universe/actions/runs/12115974337?pr=9520))

   image

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


.github/workflows/build-and-test-differential.yaml
```diff
@@ -27,27 +27,41 @@ jobs:
...
           - rosdistro: humble
             container: ghcr.io/autowarefoundation/autoware:universe-devel
+            container-suffix: ""
             build-depends-repos: build_depends.repos
-          - container-suffix: -cuda
-            runner: codebuild-autoware-us-east-1-${{ github.run_id }}-${{ github.run_attempt }}-ubuntu-7.0-large
-            build-pre-command: taskset --cpu-list 0-5
-          - container-suffix: ""
             runner: ubuntu-latest
             build-pre-command: ""
```
mitsudome-r, line 66: If we are splitting build-and-test-differential and build-and-test-differential-cuda as a job, then we probably don't have to use matrix

xmfcx: I think the separation build-and-test-differential action made its maintenance more complex and the action logs harder to read.

yhisaki: @mitsudome-r 
Matrix has been removed. Thank you!

yhisaki: @xmfcx 

Thank you for your review. I see your point. Could you provide more details on how splitting `build-and-test-differential` makes maintenance more complex or action logs harder to read? I'll consider solutions to these issues!

Alternatively, if you have other ideas for addressing the problems described in the motivation, I'd like to hear them!

xmfcx: Readeability got worse because GitHub dev team is lazy.

**See:** [Better log separation for composite actions #21276](https://github.com/orgs/community/discussions/21276)
And please vote and comment under the post 🙇

The substeps from composite actions are harder to read.

About the maintenance costs, you already have 2 versions of `clang-tidy-differential` for cuda and non-cuda. So what was extracted through separation of the action did not reduce the total lines of code in total by too much.

But added additional complexity due to introduced variables for the composite action.

And probably, handling the new cuda related tag could be done more easily without complicating the logic of the non-cuda.

## This PR

I like what you did with this PR. This was using a lot of codebuild time just to pull images on aws machines. (up to 3 minutes of compute time for each PR and each push) With this PR, if no tag, no codebuild instances will start. So thank you for implementing this.

xmfcx:
I like what you did with this PR. This was using a lot of codebuild time just to pull images on aws machines. (up to 3 minutes of compute time for each PR and each push) With this PR, if no tag, no codebuild instances will start. So thank you for implementing this.

.github/workflows/build-and-test-differential.yaml
```diff
@@ -27,29 +27,58 @@ jobs:
...
+        run: |
+          if ${{ needs.make-sure-require-cuda-label-is-present.outputs.result == 'true' }}; then
+            echo "cuda-build is required"
+            echo "cuda_build=true" >> $GITHUB_OUTPUT
+          else
+            echo "cuda-build is not required"
+            echo "cuda_build=false" >> $GITHUB_OUTPUT
+          fi
+        shell: bash
+      - name: Fail if build is not required
```
xmfcx, line 47:
```suggestion
      - name: Fail if the tag:run-build-and-test-differential is missing
```

This will convey the meaning better. Because, build is always required. But this will fail if the label is missing.

xmfcx (changes requested):

xmfcx (approved):
I need to update the required status checks now.

## Before this PR

image

https://github.com/autowarefoundation/autoware.universe/settings/rules/472262

image

.github/workflows/build-and-test-differential.yaml
yhisaki: Thank you for the detailed explanation!

Regarding readability, it seems we can only hope for improvements from the GitHub dev team in the future.

As for maintenance, the attached materials were very informative for me. Thank you!!!

PR #9525, opened 2024-11-29
fix(autoware_pointcloud_preprocessor): remove unused arg and unavailable param file.

### LINKED ISSUE #9524: Unavailable param file prevents individual preprocessor nodes from starting
### Checklist

### Description

I have setup the use of `crop_box_filter` node from `autoware_pointcloud_preprocessor` package using the individual filter launch file as follows:
```
<include file="$(find-pkg-share autoware_pointcloud_preprocessor)/launch/crop_box_filter_node.launch.xml" >
    <arg name="input_topic_name" value="pointcloud" />
    <arg name="output_topic_name" value="cropbox/filtered" />
    <arg name="cropbox_filter_param_file" value="$(find-pkg-share my_pkg)/param/crop_box_filter.yaml" />
</include>
```

### Expected behavior

The topic `cropbox/filtered` contains cropbox filtered points and I can use it further down the perception stack to do clustering and object detection.

### Actual behavior

The launch file setup above, doesn't start the filter node properly and I see the following warning on launch:
```
[WARNING] [launch_ros.actions.node]: Parameter file path is not a file: /opt/ros/jazzy/share/autoware_pointcloud_preprocessor/config/filter.param.yaml
```
And the topic `cropbox/filtered` is empty. Since I feed this topic to an instance of the euclidean clustering node, this is confirmed by the following message by the euclidean clustering node:
```
[component_container-6] [WARN] [1732860045.511208848] [euclidean_cluster]: Empty sensor points!
```

### Steps to reproduce

- Use the launch file as in the description
- `ros2 launch <path_to_launch_file>`

### Versions

- OS: Ubuntu 24.04
- ROS2: Jazzy
- Autoware: `main`

### Possible causes

The `crop_box_filter_node.launch.xml` sets a parameter `filter_param_file` [here](https://github.com/autowarefoundation/autoware.universe/blob/996336fc51a3df332e7fc0466c56aaee4cc875f7/sensing/autoware_pointcloud_preprocessor/launch/crop_box_filter_node.launch.xml#L6). However, that file [doesn't exist in the config directory](https://github.com/autowarefoundation/autoware.universe/tree/main/sensing/autoware_pointcloud_preprocessor/config) and in the code, I only see parameters from `crop_box_filter_node.param.yaml` being consumed.

The filter node does work correctly when the `filter_param_file` is initialized to point to any file that exists in the config folder. For example, if the above launch file is changed as follows:
```
<include file="$(find-pkg-share autoware_pointcloud_preprocessor)/launch/crop_box_filter_node.launch.xml" >
    <arg name="input_topic_name" value="pointcloud" />
    <arg name="output_topic_name" value="cropbox/filtered" />
    <arg name="cropbox_filter_param_file" value="$(find-pkg-share my_pkg)/config/crop_box_filter.yaml" />
    <arg name="filter_param_file" value="$(find-pkg-share autoware_pointcloud_preprocessor)/config/crop_box_filter_node.param.yaml" />
</include>
```

However, I think the argument `filter_param_file` should just be removed as that seems to be redundant.

### Additional context

_No response_

mbharatheesha:
## Description
This PR removes an unused launch argument from preprocessing filters in the `autoware_pointcloud_preprocessor` package and fixes #9524 

**Parent Issue:**

- Link https://github.com/autowarefoundation/autoware.universe/issues/9524

## How was this PR tested?
Removing the unused launch argument and hence unavailable path to the non-existent param file starts the filters properly. I tested this with the `crop_box_filter_node`.

## Interface changes

None.

#### Removals

| Removed | `filter_param_file` | string | "$(find-pkg-share autoware_pointcloud_preprocessor/config/filter.param.yaml"

## Effects on system behavior

None.


vividf (approved):
Thanks for the fix and your contribution!
It is a bug during refactoring. 

LGTM!

PR #9526, opened 2024-11-29
fix(lane_change): remove overlapping preceding lanes

zulfaqar-azmi-t4:
## Description

When the lanes are short, the preceding lane might overlap with the current lanes, as shown in the video below.

This could potentially affect the safety check results. To avoid this, we should remove the overlapping lane.

#### Before

Preceding lanes (Dark blue lanes) overlap current lanes. 

https://github.com/user-attachments/assets/d9dd7c51-4609-4be3-8d6c-c561d1582aae

#### After

Preceding lanes no longer overlaps current lanes.

https://github.com/user-attachments/assets/24a0bb3d-e4b8-4408-914d-feec7ffbb0c9

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/utils/utils.cpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/utils/utils.cpp
@@ -1287,7 +1287,6 @@ std::vector<lanelet::ConstLanelets> get_preceding_lanes(const CommonDataPtr & co
   return non_overlapping_lanes_vec;
 }
 
-
 bool object_path_overlaps_lanes(
   const ExtendedPredictedObject & object, const lanelet::BasicPolygon2d & lanes_polygon)
 {
--- a/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/utils/utils.cpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/utils/utils.cpp
@@ -34,9 +34,12 @@
 #include <autoware_lanelet2_extension/utility/query.hpp>
 #include <autoware_lanelet2_extension/utility/utilities.hpp>
 #include <autoware_vehicle_info_utils/vehicle_info.hpp>
+#include <range/v3/action/insert.hpp>
 #include <range/v3/algorithm/any_of.hpp>
+#include <range/v3/algorithm/find_if.hpp>
+#include <range/v3/algorithm/for_each.hpp>
 #include <range/v3/range/conversion.hpp>
-#include <range/v3/view/transform.hpp>
+#include <range/v3/view.hpp>
 #include <rclcpp/rclcpp.hpp>
 
 #include <geometry_msgs/msg/detail/pose__struct.hpp>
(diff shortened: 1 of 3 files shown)


maxime-clem (approved):

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/utils/utils.cpp
```diff
@@ -1247,4 +1252,57 @@ bool filter_target_lane_objects(
...
+  const auto & current_lanes = common_data_ptr->lanes_ptr->current;
+  std::unordered_set<lanelet::Id> current_lanes_id;
+  for (const auto & lane : current_lanes) {
+    current_lanes_id.insert(lane.id());
+
+    const auto prev_lanes = route_handler_ptr->getPreviousLanelets(lane);
+    for (const auto & prev_lane : prev_lanes) {
+      current_lanes_id.insert(prev_lane.id());
+    }
+  }
```
mkquda, line 1275: Won't this logic unnecessarily insert duplicated lane ids for each lane in `current_lanes`

Also I can't see why you need to `getPreviousLanelets` for each lane in `current_lanes`, maybe it is sufficient to do it only for first lane?

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/utils/utils.cpp
```diff
@@ -1247,4 +1252,57 @@ bool filter_target_lane_objects(
...
+      if (!is_connected) {
+        not_connected_ids.insert(current_lane.id());
+      }
+    }
+
+    // Step 3: Remove disconnected lanes
+    auto connected_lanes =
+      non_overlapping_lanes | ranges::views::remove_if([&](const lanelet::ConstLanelet & lane) {
+        return not_connected_ids.find(lane.id()) != not_connected_ids.end();
+      });
```
mkquda, line 1302: Can you explain whats the purpose of this?

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/utils/utils.cpp
zulfaqar-azmi-t4: No this logic is necessary. Because there might be lanes that is connected to the current lanes, but not belong to the route.

image

For example these lanes is connected to current lanes.

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/utils/utils.cpp
zulfaqar-azmi-t4: When we removed the overlapping lanes, we might ended up lanes that is not connected. For example

Original: L4, L3, L2, L1. Assume overlapping is L2.
Overlapping removed: L1, L3, L4

Since L1 and L3 is not directly connected, we should remove L1 from the list, which yield
Connected: L4, L3.

zulfaqar-azmi-t4: @mkquda 
fixed in https://github.com/autowarefoundation/autoware.universe/pull/9526/commits/744e6aa6820155abd1b6bdb80ee125915789ba18

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/utils/utils.cpp
```diff
@@ -1247,4 +1252,62 @@ bool filter_target_lane_objects(
...
+  const auto preceding_lanes_list =
+    utils::getPrecedingLanelets(*route_handler_ptr, target_lanes, ego_pose, backward_lane_length);
+
+  const auto & current_lanes = common_data_ptr->lanes_ptr->current;
+  std::unordered_set<lanelet::Id> current_lanes_id;
+  for (const auto & lane : current_lanes) {
+    current_lanes_id.insert(lane.id());
+  }
+  const auto is_overlapping = [&](const lanelet::ConstLanelet & lane) {
+    return current_lanes_id.find(lane.id()) == current_lanes_id.end();
```
mkquda, line 1272:
```suggestion
    return current_lanes_id.find(lane.id()) != current_lanes_id.end();
```

zulfaqar-azmi-t4: Done https://github.com/autowarefoundation/autoware.universe/pull/9526/commits/7a6ca52dc897e5cffa73dde3df5b2eaae02ccf40

mkquda (approved):
LGTM.

Issue #9428, opened 2024-11-22
The Calculation Method of "latency_ms" in Node "multi_object_tracker"  Seems Unreasonable


cyn-liu:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I'm convinced that this is not my fault but a bug.

### Description

`pipeline_latency_ms` represents the time it takes for the entire pipeline from point cloud publishing to the completion of execution at the current node.

We found node `/perception/object_recognition/tracking/multi_object_tracker`'s `debug/pipeline_latency_ms` is large and the data fluctuates greatly.

https://github.com/user-attachments/assets/40e193de-f018-4137-93a4-cf2afb4a1e93

We believe that there is a problem with the calculation method of `/perception/object_recognition/tracking/multi_object_tracker/debug/pipeline_latency_ms` , which cannot actually reflect the pipeline latency from publishing the point cloud  to completing the operation of node `multi_object_tracker` .

### Expected behavior

`/perception/object_recognition/tracking/multi_object_tracker/debug/pipeline_latency_ms` is slightly large than `/perception/object_recognition/detection/object_lanelet_filter/debug/pipeline_latency_ms`.

### Actual behavior

`/perception/object_recognition/tracking/multi_object_tracker/debug/pipeline_latency_ms` is much large than `/perception/object_recognition/detection/object_lanelet_filter/debug/pipeline_latency_ms`.

### Steps to reproduce

1. Launch Autoware `logging_simulator`
2. Play ROS2 bag
3. Use `rqt_plot` to view `/perception/object_recognition/tracking/multi_object_tracker/debug/pipeline_latency_ms`

### Versions

- OS: ubuntu22.04
-  ROS2: humble
-  Autoware: Latest version

### Possible causes

1. Perhaps debug time should not be published in `onTime`.
2. When calculating time, use the timestamp of the oldest data in the input data queue, the amount of data in the queue often changes, resulting in significant fluctuations in the time calculation results.

### Additional context

None

technolojin:
@cyn-liu 
What is your configuration of `enable_delay_compensation`?
If this value is true, it can extrapolate the tracking result (depends on your system situation), and the `pipeline latency` becomes large.

technolojin:
Related PR

https://github.com/autowarefoundation/autoware.universe/pull/6710
[...]

PR #9533, opened 2024-12-02
fix(autoware_multi_object_tracker): measure latency with latest detection update time

technolojin:
## Description
Fixing the pipeline latency measurement point.
The latest updated detection time will be used for the latency measurement.

1. startMeasurementTime uses latest time rather than oldest time
2. `object_oldest_time` is renamed to `object_earliest_time`
3. `onMessage` method is removed since it is not callback function anymore.

## Related links

**Parent Issue:** https://github.com/autowarefoundation/autoware.universe/issues/9428

- Link

## How was this PR tested?

Screenshot from 2024-12-02 13-43-06

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


cyn-liu:
I have tested this PR and the result looks consistent with expectations.
https://github.com/autowarefoundation/autoware.universe/issues/9428#issuecomment-2514049848

cyn-liu (approved):

YoshiRi (approved):
LGTM

PR #9538, opened 2024-12-02
feat(system_monitor): check UDP network errors

iwatake2222:
### Why this  PR is needed

- UDP packet drops make Autoware unstable. Therefore, this is an important indicator for determining whether Autoware is functioning correctly
- There are several causes of UDP packet drops, but the primary reason is drops in the send/receive buffer. This is particularly true in the case of the loopback network

### What this PR changs

- Monitor UDP recv buffer errors and UDP send buffer errors

## Related links

None

## How was this PR tested?

- `total UDP rcv buf errors` and `total UDP snd buf errors` are monitored, and they are retrieved from `/proc/net/snmp`
- When errors per unit time is greater than threshold (>=1), status becomes warning
- The CPU usage of system_monitor doesn't change
  - Without this PR: 13.6% / 1600%
  - With this PR: 12.6% / 1600%
  - (the CPU usage of system_monitor in 16-core ECU)

image

### How to cause UDP buf errors

Run the following scripts

```sender.py
import socket
import time

sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)

# To cause snd buf error, enable the following line and run the following command
# $> sudo tc qdisc add dev lo root netem delay 100ms
# $> sudo tc qdisc del dev lo root    # fix setting
sock.setsockopt(socket.SOL_SOCKET, socket.SO_SNDBUF, 4)

target_ip = "127.0.0.1"
target_port = 12345

while True:
    sock.sendto(b"a" * 1024, (target_ip, target_port))
    time.sleep(0.00001)
```

```receiver.py
import socket

sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
sock.setsockopt(socket.SOL_SOCKET, socket.SO_RCVBUF, 1)
sock.bind(("0.0.0.0", 12345))

while True:
    try:
        data, addr = sock.recvfrom(1024)
    except Exception as e:
        print(f"Error: {e}")
```

## Notes for reviewers

None.

#### Additions and removals

None

#### Modifications

A new member named `net_monitor: UDP Buf Errors` is added into `/diagnostics` topic

#### Additions and removals

system/system_monitor/config/net_monitor.param.yaml

| New     | `udp_buf_errors_check_duration` | `int` | `1`         | UDP buf errors check duration |
| New     | `udp_buf_errors_check_count` | `int` | `1`         | Generates warning when count of UDP buf errors during udp_buf_errors_check_duration reaches a specified value or higher |

#### Modifications

None

## Effects on system behavior

None.


iwatake2222:
@ito-san @TetsuKawa 
Could you review this PR?
Also, please re-add `run-build-and-test-differential` label to run worflows (some workflows are stuck for some reasons...)

system/system_monitor/src/net_monitor/net_monitor.cpp
```diff
@@ -327,6 +338,62 @@ void NetMonitor::check_reassembles_failed(diagnostic_updater::DiagnosticStatusWr
...
+
+  uint64_t total_udp_rcvbuf_errors = 0;
+  if (get_value_from_net_snmp(udp_rcvbuf_errors_index_, total_udp_rcvbuf_errors)) {
+    update_delta_queue(
+      udp_rcvbuf_errors_queue_, last_udp_rcvbuf_errors_, total_udp_rcvbuf_errors,
+      udp_buf_errors_check_duration_);
+    uint64_t unit_udp_rcvbuf_errors =
+      std::accumulate(udp_rcvbuf_errors_queue_.begin(), udp_rcvbuf_errors_queue_.end(), 0);
+    status.add(fmt::format("total UDP rcv buf errors"), total_udp_rcvbuf_errors);
+    status.add(fmt::format("UDP rcv buf errors per unit time"), unit_udp_rcvbuf_errors);
```
ito-san, line 357: I don't think fmt::format is necessary here.

system/system_monitor/src/net_monitor/net_monitor.cpp
```diff
@@ -327,6 +338,62 @@ void NetMonitor::check_reassembles_failed(diagnostic_updater::DiagnosticStatusWr
...
+      whole_level = std::max(whole_level, static_cast<int>(DiagStatus::WARN));
+      error_message += "UDP buf errors";
+    }
+  } else {
+    udp_rcvbuf_errors_queue_.push_back(0);
+    whole_level = std::max(whole_level, static_cast<int>(DiagStatus::ERROR));
+    error_message += "failed to read RcvbufErrors from /proc/net/snmp";
+  }
+
+  uint64_t total_udp_sndbuf_errors = 0;
```
ito-san, line 368: It looks like duplicate code, so please try to make it more reusable.

system/system_monitor/src/net_monitor/net_monitor.cpp
iwatake2222: Fixed in https://github.com/autowarefoundation/autoware.universe/pull/9538/commits/bfbbaac79edc8d02498b363e07098a4e0725ee12

iwatake2222:
@ito-san @TetsuKawa 
Thank you for reviewing it. I updated the code as you suggested.
Could you review it again please?

system/system_monitor/src/net_monitor/net_monitor.cpp
TetsuKawa: @iwatake2222
I'm a bit concerned about the performance impact of reading /proc/net/snmp separately for each of these three items:

- IP fragmentation failed
- UDP send buffer errors
- UDP receive buffer errors

I think it would be better to consolidate these `std::ifstream ifs("/proc/net/snmp");` into a single function. (it is just an idea that creating a class including NetSnmp instances. the design is up to you.).

iwatake2222: @TetsuKawa 
Thank you for your feedback.  
I had exactly the same thought during the implementation.  

However, I concluded that there is no performance concern for the following reasons:
- generally, the overhead of accessing `/proc` is smaller than accessing a normal file
- the actual measurement results showed no noticeable increase in processing load

Of course, there might be a very very slight increase that is not visible at this level, but I judged that the current implementation is the better approach considering the following factors:
- Increased complexity of the implementation
- Higher maintenance costs if the number of monitored metrics increases in the future
- Increased risk of regression (with potential impacts on all `/proc/net/snmp`-related information)  

Therefore, I would like to keep the current design and implementation as it is :pray:

TetsuKawa: I understand. 
As you say, since /proc exists in memory rather than on SSD, it probably doesn't increase the process load much. (This is very educational for me.) 
I agree with your thinking!

TetsuKawa:
LG**TM**
@ito-san 
I leave the rest to your judgment.

iwatake2222:
@ito-san 
Friendly reminder...

ito-san:
> @ito-san Friendly reminder...
Sorry, the codebase was different between this branch and my project, so my Autoware was broken after I merged this branch.
I've updated the branch, so please wait a little longer.

ito-san:
Verified that no errors occur during startup
image

ito-san:
It seems to be working well
image

ito-san (approved):
LGTM

ito-san:
@iwatake2222 
Please fix this.
/__w/autoware.universe/autoware.universe/system/system_monitor/src/net_monitor/net_monitor.cpp:814:26: error: folding type 'unsigned int' into type 'int' might result in loss of precision [bugprone-fold-init-type,-warnings-as-errors]
  value_per_unit_time_ = std::accumulate(queue_.begin(), queue_.end(), 0);
                         ^

iwatake2222:
@ito-san 
Could you add label to run workflow?

iwatake2222:
@ito-san 
Thank you for reviewing it !!
I fixed the code and all required checks are passed (pre-commit-optional fails due to link error, but it's not related to this PR)

Could you merge this PR? I'm asking because `Only those with write access to this repository can merge pull requests.`

PR #9545, opened 2024-12-03
feat(lane_changing): improve computation of lane changing acceleration

mkquda:
## Description

Currently when generating candidate lane change paths, we sample multiple longitudinal acceleration values, however for values below 0.0 (decelerating), it is only applied to the prepare phase, while min acceleration for lane changing phase is limited to 0.0, meaning we assume no deceleration during lane changing phase. And this behavior is not configurable.

Sometimes this assumption can result in not finding a safe candidate path near the terminal, specifically when there is a leading target lane object.

It is better to allow some deceleration during lane changing phase when approaching terminal to increase the chances of finding a safe candidate path.

## Changes

Modified function `calc_lane_changing_acceleration`:
 - IF sampled longitudinal acceleration is negative, AND ego is near terminal, THEN return a ratio of sampled longitudinal acceleration determined by parameter `lane_changing_decel_factor`
 - IF sampled longitudinal acceleration is negative, AND ego is NOT near terminal, THEN return 0.0 (same as before)

## Related links

**Parent Issue:**

- Link

## How was this PR tested?
- PSIM
- [TIER IV Internal Evaluation](https://evaluation.tier4.jp/evaluation/reports/3fe71be5-8d0e-519e-9103-de95af163ad8?project_id=prd_jt)

## Notes for reviewers

None.

## Interface changes

None.

#### Additions and removals

| Added | `lane_changing_decel_factor` | `double` | `0.5` | longitudinal deceleration factor during lane changing phase |

## Effects on system behavior

Increase chances of finding a safe candidate path near terminal when there is a leading target lane object.

### DIFF
--- a/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/README.md
+++ b/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/README.md
@@ -893,7 +893,7 @@ The following parameters are configurable in [lane_change.param.yaml](https://gi
 | `trajectory.lat_acc_sampling_num`            | [-]    | int    | Number of possible lane-changing trajectories that are being influenced by lateral acceleration                        | 3                  |
 | `trajectory.max_longitudinal_acc`            | [m/s2] | double | maximum longitudinal acceleration for lane change                                                                      | 1.0                |
 | `trajectory.min_longitudinal_acc`            | [m/s2] | double | maximum longitudinal deceleration for lane change                                                                      | -1.0               |
-| `trajectory.lane_changing_decel_factor`      | [m/s2] | double | longitudinal deceleration factor during lane changing phase                                                            | 0.5                |
+| `trajectory.lane_changing_decel_factor`      | [-] | double | longitudinal deceleration factor during lane changing phase                                                            | 0.5                |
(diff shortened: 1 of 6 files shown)


planning/behavior_path_planner/autoware_behavior_path_lane_change_module/README.md
```diff
@@ -893,6 +893,7 @@ The following parameters are configurable in [lane_change.param.yaml](https://gi
 | `trajectory.lat_acc_sampling_num`            | [-]    | int    | Number of possible lane-changing trajectories that are being influenced by lateral acceleration                        | 3                  |
 | `trajectory.max_longitudinal_acc`            | [m/s2] | double | maximum longitudinal acceleration for lane change                                                                      | 1.0                |
 | `trajectory.min_longitudinal_acc`            | [m/s2] | double | maximum longitudinal deceleration for lane change                                                                      | -1.0               |
+| `trajectory.lane_changing_decel_factor`      | [m/s2] | double | longitudinal deceleration factor during lane changing phase                                                            | 0.5                |
```
maxime-clem, line 896: The unit is incorrect.

```suggestion
| `trajectory.lane_changing_decel_factor`      | [-] | double | longitudinal deceleration factor during lane changing phase                                                            | 0.5                |
```

zulfaqar-azmi-t4 (approved):

maxime-clem (approved):

PR #9489, opened 2024-11-27
chore(autoware_pyplot): move to visualization folder

esteve:
## Description

`common` should be meant for base packages that multiple other packages depend on.
## Related links

**Parent Issue:**

- https://github.com/orgs/autowarefoundation/discussions/5462#discussioncomment-11384987

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

PR #9546, opened 2024-12-03
docs(autoware_pyplot): add missing readme and move the package to common

soblin:
## Description

add missing README of autoware_pyplot package.

@esteve Hi, is it OK for us to use this package from `common` and `planning` components ? The major motivation for this package is to create the illustrative descriptions of utility functions of `common`/`planning` components as documentation for better development experience. If it is unintended to use `visualization` package from `common` and `planning`, can you move this back to `common` ?

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

 document PR

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


yhisaki (approved):

xmfcx:
This reverts:
- https://github.com/autowarefoundation/autoware.universe/pull/9489

xmfcx:
@soblin can you rebase this to latest main?

xmfcx:
@esteve do you have any objection is on this?

This is a cpp library that provides python bindings to visualize graphs. Will be used and be depended by a lot of packages including the packages from common. That's why it might be a better idea to keep it in common.

esteve:
@xmfcx if it's gonna be used by other packages, sure, no problem. My concern was that it seemed it wasn't being used by any other package other than those in `visualization`, but it's being used by packages that are in other folders, it makes sense to move it back.

I wonder if it might be better to rename `common` to `base`, IMHO it better describes its purpose.

xmfcx:
@soblin You may merge this since we agreed on moving it back to `common` 👍

PR #9547, opened 2024-12-03
refactor(lane_departure_checker): move member functions to util functions

kyoichi-sugahara:
## Description

minor refactoring changes to improve testability and prepare the lane_departure_checker module for unit testing implementation. The changes are minimal and do not affect the core functionality.

- Moved `createVehiclePassingAreas()`, `getCandidateLanelets()` and `createVehiclePassingAreas()`  from `LaneDepartureChecker` class to utility functions
- Added proper documentation for utility functions using Doxygen style comments
- Add proper handling for edge cases in `createVehiclePassingAreas`
  - Handle empty input vector
  - Handle single footprint case
  - Pre-reserve vector capacity for better performance

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

run psim

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

PR #9548, opened 2024-12-03
test(lane_departure_checker): add unit test for createVehiclePassingAreas

kyoichi-sugahara:
## Description
add unit test for `createVehiclePassingAreas`

## How was this PR tested?

confirmed test result was all passed.
```
1: [==========] Running 26 tests from 7 test suites.
1: [----------] Global test environment set-up.
1: [----------] 4 tests from CreateVehiclePassingAreasTest
1: [ RUN      ] CreateVehiclePassingAreasTest.ReturnsEmptyAreaForEmptyInput
1: [       OK ] CreateVehiclePassingAreasTest.ReturnsEmptyAreaForEmptyInput (0 ms)
1: [ RUN      ] CreateVehiclePassingAreasTest.ReturnsSameAreaForSingleFootprint
1: [       OK ] CreateVehiclePassingAreasTest.ReturnsSameAreaForSingleFootprint (0 ms)
1: [ RUN      ] CreateVehiclePassingAreasTest.CreatesValidHullForAdjacentFootprints
1: [       OK ] CreateVehiclePassingAreasTest.CreatesValidHullForAdjacentFootprints (0 ms)
1: [ RUN      ] CreateVehiclePassingAreasTest.HandlesNonAdjacentFootprints
1: [       OK ] CreateVehiclePassingAreasTest.HandlesNonAdjacentFootprints (0 ms)
1: [----------] 4 tests from CreateVehiclePassingAreasTest (0 ms total)
```

## Notes for reviewers

This PR includes refactor [PR](https://github.com/autowarefoundation/autoware.universe/pull/9547) so please check together with the PR. 
## Interface changes

None.

## Effects on system behavior

None.


control/autoware_lane_departure_checker/src/lane_departure_checker_node/utils.cpp
```diff
@@ -163,6 +165,64 @@ std::vector<LinearRing2d> createVehicleFootprints(
...
+    return std::vector<LinearRing2d>();
+  }
+
+  std::vector<LinearRing2d> areas;
+  areas.reserve(vehicle_footprints.size() == 1 ? 1 : vehicle_footprints.size() - 1);
+
+  if (vehicle_footprints.size() == 1) {
+    areas.push_back(vehicle_footprints.front());
+    return areas;
+  }
```
zulfaqar-azmi-t4, line 215:
```suggestion
  if (vehicle_footprints.size() == 1) {
    return {vehicle_footprints.front()};
  }
  
  std::vector<LinearRing2d> areas;
  areas.reserve(vehicle_footprints.size() - 1);
```
Since this is not refactoring PR, so it's fine not to commit this.
But in case you're refactoring, maybe you can do something like this to avoid having to add logic in the reserve.

control/autoware_lane_departure_checker/test/test_create_vehicle_passing_areas.cpp
```diff
@@ -0,0 +1,139 @@
...
+  bool isPointInsideHull(const Point2d & point, const LinearRing2d & hull)
+  {
+    return boost::geometry::within(point, hull) || boost::geometry::covered_by(point, hull);
+  }
+
+  void validateFootprintsInHull(
+    const std::vector<LinearRing2d> & footprints, const LinearRing2d & hull)
+  {
+    for (const auto & footprint : footprints) {
+      for (size_t i = 0; i < footprint.size() - 1; ++i) {
```
zulfaqar-azmi-t4, line 56: Is it supposed to be `i < footprint.size() - 1` or `i < footprint.size()`?
If footprint.size() == 0, then we might have an overflow.

kyoichi-sugahara: @zulfaqar-azmi-t4 
Since `LinearRing2d` represents a closed linear ring, I believe the last point doesn't need to be checked. So I will add a comment to clarify this. 
Regarding the overflow concern, I think when `footprint.size() == 0`, it would lead to undefined behavior. While we expect the test to fail in such cases, this expectation isn't explicitly documented, so I will modify the code and make this clear.

addredded in https://github.com/autowarefoundation/autoware.universe/pull/9548/commits/b476d59e7b9ba6b01914361b67539878ea067fa5

zulfaqar-azmi-t4 (approved):

PR #9554, opened 2024-12-03
feat(autoware_geography_utils)!: move autoware_geography_utils to autoware.core

mitsudome-r:
## Description
We are porting autoware_geography_utils to autoware.core, and this PR removes the package from autoware.universe.

## Related links

**Parent Issue:**
- Issue: https://github.com/autowarefoundation/autoware.core/issues/99
- This has to be merged with the following pull requests: https://github.com/autowarefoundation/autoware.core/pull/100

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


xmfcx (approved):
I've confirmed that the package was cut and pasted only with the following modifications:
- changelog.rst
- package.xml

**Source:**
- https://github.com/autowarefoundation/autoware.universe/pull/9554

**Target:**
- https://github.com/autowarefoundation/autoware.core/pull/100

This should not break any existing configuration as long as we also make sure the autoware main repo points to the correct versions of core and universe.

YamatoAndo (approved):

xmfcx:
I've rebased this to latest main.

Because "include what you use" pr has modified file, it needed manual rebasing.

But in the end it is just deleting these files, so nothing really changed.

mitsudome-r:
@xmfcx Thanks!

Issue #9556, opened 2024-12-03
Fix cpplint errors


xmfcx:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I've agreed with the maintainers that I can plan this task.

### Description

Recently, I wanted to update pre-commit hooks. It seems the cpplint version is also updated. And the new version is better able to detect mistakes.

We have a lot of `include_what_you_use` errors along with some minor errors.

### Errors to fix

**I've made a list of the required changes here:**
🚨 https://gist.github.com/xmfcx/4266f2048fff6a7fa9c11c6798eea456
(I've applied some manual formatting to make it easier to view)

They are from the following update PR:
- https://github.com/autowarefoundation/autoware.universe/pull/8949

To replicate, you can do the following:
- `cd autoware/src/universe/autoware.universe`
- `git pull`
- `git checkout pre-commit-autoupdate`
- `pre-commit autoupdate`
- `pre-commit run -a`

This should run it against all files and give you the list of what to fix.

We should go through these changes and fix them one by one. I couldn't find a way to do this automatically.

cc. @veqcc @yhisaki @mojomex @mitsudome-r

xmfcx:
All the include what you use PRs:
- https://github.com/autowarefoundation/autoware.universe/pull/9564
- https://github.com/autowarefoundation/autoware.universe/pull/9565
- https://github.com/autowarefoundation/autoware.universe/pull/9566
- https://github.com/autowarefoundation/autoware.universe/pull/9567
- https://github.com/autowarefoundation/autoware.universe/pull/9568
- https://github.com/autowarefoundation/autoware.universe/pull/9569
- https://github.com/autowarefoundation/autoware.universe/pull/9570
- https://github.com/autowarefoundation/autoware.universe/pull/9571
- https://github.com/autowarefoundation/autoware.universe/pull/9572
- https://github.com/autowarefoundation/autoware.universe/pull/9573
- https://github.com/autowarefoundation/autoware.universe/pull/9574
- https://github.com/autowarefoundation/autoware.universe/pull/9575
- https://github.com/autowarefoundation/autoware.universe/pull/9576

xmfcx:
With the following PR:
- https://github.com/autowarefoundation/autoware.universe/pull/9557
This task is completed ✅

PR #9563, opened 2024-12-04
feat(bytetrack): remove unreachable code block from lapjv.h

xmfcx:
## Description

This code block is unreachable unless the header is modified by hand and recompiled.

This PR removes that.

### Why?

As part of:
- https://github.com/autowarefoundation/autoware.universe/issues/9556

This weird macro heavy code snippet was confusing the static analysis tool.

It uses `printf` function yet doesn't include `<cstdio>`. Even if I include it, because of the conditional structure, the cpplint wasn't able to recognize it.

So I decided to remove it since it is not used anyways.

## How was this PR tested?

Basic CI.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_bytetrack/lib/include/lapjv.h
+++ b/perception/autoware_bytetrack/lib/include/lapjv.h
@@ -64,40 +64,10 @@
     b = _temp_index;       \
   }
 
-#if 0
-#include <assert.h>
-#define ASSERT(cond) assert(cond)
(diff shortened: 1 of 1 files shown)


xmfcx:
@manato @YoshiRi it's ready for review ✅

manato (approved):
@xmfcx 
LGTM! Thank you for cleaning up the code.

PR #9558, opened 2024-12-03
fix(cpplint): include what you use

xmfcx:
## Description

Fixes most of:
- https://github.com/autowarefoundation/autoware.universe/issues/9556

## Related links

- [Gist of the problem list.](https://gist.github.com/xmfcx/4266f2048fff6a7fa9c11c6798eea456)

## How was this PR tested?

I've generated this PR with the following script and its instructions:

[fix_includes_pre-commit.py](https://gist.github.com/xmfcx/dfd64f9e69fba1a083ad8391f5133a48)

> ```bash
> cd autoware/src/universe/autoware.universe/
> pre-commit update
> wget https://gist.githubusercontent.com/xmfcx/dfd64f9e69fba1a083ad8391f5133a48/raw/18f56fdb1300daf4bf88b47b15b81505cb6e00df/fix_includes_pre-commit.py
> python3 fix_includes_pre-commit.py
> ```
> 
> This should add all the missing includes.
> 
> Then you can run `pre-commit run clang-format -a` a couple of times to fix their orders.
> 
> Finally to verify, run: `pre-commit run -a` and you shouldn't have any "include-what-you-use" errors left. ✅

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

PR #9569, opened 2024-12-04
fix(cpplint): include what you use - perception

xmfcx:
## Description

**Parent Issue:**
- https://github.com/autowarefoundation/autoware.universe/issues/9556

Extracted from:
- https://github.com/autowarefoundation/autoware.universe/pull/9558

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_bytetrack/lib/include/lapjv.h
+++ b/perception/autoware_bytetrack/lib/include/lapjv.h
@@ -66,6 +66,8 @@
 
 #if 0
 #include <assert.h>
+
+#include <cstdio>
 #define ASSERT(cond) assert(cond)
 #define PRINTF(fmt, ...) printf(fmt, ##__VA_ARGS__)
 #define PRINT_COST_ARRAY(a, n)         \
--- a/perception/autoware_bytetrack/lib/include/lapjv.h
+++ b/perception/autoware_bytetrack/lib/include/lapjv.h
@@ -66,8 +66,6 @@
 
 #if 0
 #include <assert.h>
-
-#include <cstdio>
 #define ASSERT(cond) assert(cond)
 #define PRINTF(fmt, ...) printf(fmt, ##__VA_ARGS__)
 #define PRINT_COST_ARRAY(a, n)         \
--- a/perception/autoware_bytetrack/lib/src/byte_tracker.cpp
+++ b/perception/autoware_bytetrack/lib/src/byte_tracker.cpp
@@ -40,6 +40,8 @@
 
 #include <cstddef>
 #include <fstream>
+#include <iostream>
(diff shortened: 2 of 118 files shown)


perception/autoware_bytetrack/lib/include/lapjv.h
```diff
@@ -66,6 +66,8 @@
 
 #if 0
 #include <assert.h>
+
+#include <cstdio>
```
mitsudome-r, line 70: I think we can remove this change because we will be covering it by another PR: https://github.com/autowarefoundation/autoware.universe/pull/9563/files

mitsudome-r: I'm still going to approve it because there will be no harm. (You will just have a conflict in the above PR)

mitsudome-r (approved):

perception/autoware_bytetrack/lib/src/byte_tracker.cpp
mitsudome-r, line 1: @xmfcx Do you have any idea why GitHub is failing to show the correct diffs in this file?

perception/autoware_bytetrack/lib/include/lapjv.h
xmfcx: Good catch!

I'll remove it now.

perception/autoware_bytetrack/lib/src/byte_tracker.cpp
xmfcx: image

It had bad line endings it seems:

**Before (main):**
```console
mfc@whale:~$ file -k /home/mfc/projects/autoware/src/universe/autoware.universe/perception/autoware_bytetrack/lib/src/byte_tracker.cpp
/home/mfc/projects/autoware/src/universe/autoware.universe/perception/autoware_bytetrack/lib/src/byte_tracker.cpp: C source, ASCII text, with CRLF line terminators
```

**After (This PR):**
```console
mfc@whale:~$ file -k /home/mfc/projects/autoware/src/universe/autoware.universe/perception/autoware_bytetrack/lib/src/byte_tracker.cpp
/home/mfc/projects/autoware/src/universe/autoware.universe/perception/autoware_bytetrack/lib/src/byte_tracker.cpp: C source, ASCII text
```

perception/autoware_bytetrack/lib/include/lapjv.h
xmfcx: Reverted with https://github.com/autowarefoundation/autoware.universe/pull/9569/commits/2b170200c51c60969f530dd545174c4709a64c98

PR #9579, opened 2024-12-05
feat(costmap_generator, scenario_selector): improve freespace planning stability

mkquda:
## Description

In parking scenario, ego can sometimes get stuck when navigating close to freespace area boundary. This happens due to the inconsistency of costmap evaluation at the boundaries.
As shown below, Initially freespace generates a valid trajectory, but as ego moves the costmap evaluation fluctuates, and the endpoint of the forward trajectory becomes invalid, and freespace fails to generate a new trajectory.

https://github.com/user-attachments/assets/713a458d-c164-497b-87d5-f1dd5c5452bd

Another issue is, when ego is stuck for sometime in parking scenario, it will attempt to switch to lane driving (feature added in this [previous PR](https://github.com/autowarefoundation/autoware.universe/pull/8348)), However the logic for switching to lane driving doesn't guarantee ego pose is proper enough to follow lanelet route. Which can cause control error due to large deviation as shown below.

https://github.com/user-attachments/assets/0c3ac6bd-b9c1-48aa-a9f2-26365a7496a9

To fix the first issue (unstable costmap evaluation at boundary):
- instead of continuously setting grid center at ego position
- compute the relative position of ego wrt current grid center
- compute the offset as a multiple of grid resolution
- use offset to set the new position of grid center

To fix second issue (incorrect switching to lane driving):
- get closest lane to ego using distance and yaw constraints
- ensure ego is close enough to lane center line

The resulting behavior is shown below.

https://github.com/user-attachments/assets/5debd550-c398-4b98-be48-3c435c2021f2

## Related links

**Private Links:**

- [TIER IV internal link](https://star4.slack.com/archives/C03QW0GU6P7/p1732863756724719)

## How was this PR tested?
- PSIM.
- Scenario simulator

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

Described above.

### DIFF
--- a/planning/autoware_costmap_generator/include/autoware/costmap_generator/costmap_generator.hpp
+++ b/planning/autoware_costmap_generator/include/autoware/costmap_generator/costmap_generator.hpp
@@ -139,6 +139,13 @@ private:
 
   void set_current_pose();
 
+  /// \brief set position of grid center
+  /// \details computes relative position of ego from current grid center,
+  /// if offset is larger than grid resolution, grid center will be updated
+  /// by a multiple of the grid resolution
+  /// \param[in] tf costmap frame to vehicle frame transform
+  void set_grid_center(const geometry_msgs::msg::TransformStamped & tf);
+
   void onTimer();
 
   bool isActive();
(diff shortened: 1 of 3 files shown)


maxime-clem (approved):
The change looks good :+1:

kosuke55 (approved):

PR #9583, opened 2024-12-06
feat(autoware_lidar_centerpoint): add priority map to preprocess kernel and centerpoint_trt

technolojin:
## Description

Current voxel processing is done by order of X, Y sequentially.
IF the number of filled voxel exceed the ML model limit, the detectable range is limited.

The priority map is set by a scalar field as following.
```C++
    // absolute rectangular hyperbola to focus on crosssection
    // multiply sigmoid function to prioritize the front area
    const float score_a = abs((pos_x - 15.0) * pos_y) * (1 / (1 + exp(pos_x * 0.3)) + 0.75);
    // ellipse area focus on (10,0) and (150,0)
    const float score_b = sqrt((pos_x - 150) * (pos_x - 150) + pos_y * pos_y) +
                          sqrt((pos_x - 10) * (pos_x - 10) + pos_y * pos_y) - 140;
    // total score with weight
    // add noise to extend detection area in low-score area
    const float score =
      sqrt(score_a * 0.1 + score_b * 2.5) + normal_distribution(generator) * score_b * 0.01;
```
Figure_2

The following visualization is to show filled voxel distribution. Configurations are as follwing.
* input high-density lidar point cloud
* score_threshold: 0.01
* max_voxel_size: 40000
* point_cloud_range: [-121.6, -76.8, -3.0, 121.6, 76.8, 5.0]
Then lower score object will be shown (due to the lower score_threshold), indirectly visualize where the voxel is filled.

Screenshot from 2024-11-29 18-19-04

## Related links
**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_lidar_centerpoint/include/autoware/lidar_centerpoint/centerpoint_trt.hpp
+++ b/perception/autoware_lidar_centerpoint/include/autoware/lidar_centerpoint/centerpoint_trt.hpp
@@ -67,6 +67,8 @@ public:
 protected:
   void initPtr();
 
+  void initPriorityMap();
+
   virtual bool preprocess(
     const sensor_msgs::msg::PointCloud2 & input_pointcloud_msg, const tf2_ros::Buffer & tf_buffer);
 
@@ -89,6 +91,8 @@ protected:
   std::size_t voxels_size_{0};
   std::size_t coordinates_size_{0};
   std::vector<float> points_;
+  std::vector<unsigned int> priority_map_;
+
   cuda::unique_ptr<float[]> voxels_d_{nullptr};
   cuda::unique_ptr<int[]> coordinates_d_{nullptr};
   cuda::unique_ptr<float[]> num_points_per_voxel_d_{nullptr};
@@ -106,6 +110,7 @@ protected:
   cuda::unique_ptr<float[]> voxels_buffer_d_{nullptr};
   cuda::unique_ptr<unsigned int[]> mask_d_{nullptr};
   cuda::unique_ptr<unsigned int[]> num_voxels_d_{nullptr};
+  cuda::unique_ptr<unsigned int[]> priority_map_d_{nullptr};
   cuda::unique_ptr<unsigned int[]> shuffle_indices_d_{nullptr};
 };
 
--- a/perception/autoware_lidar_centerpoint/include/autoware/lidar_centerpoint/preprocess/preprocess_kernel.hpp
+++ b/perception/autoware_lidar_centerpoint/include/autoware/lidar_centerpoint/preprocess/preprocess_kernel.hpp
@@ -36,9 +36,9 @@ cudaError_t generateVoxels_random_launch(
   float * voxels, cudaStream_t stream);
 
 cudaError_t generateBaseFeatures_launch(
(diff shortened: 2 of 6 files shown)


knzo25:
I am quite opposed to this idea. Here are some of the reasons
 - it introduces random access (there already one in the affected method, but this adds another).
 - Going over the limit should not happen for a specific product/project. In the past, when faced with a similar situation the resolution was to simply increase the size (number of points, in the past). When using centerpoint, we are not limited by memory resources.
 - If this reaaaally needs to be added, I would want it to be enabled automatically when the first overflow of max voxels occurs, or as a parameter.
 - Looking ahead, there are talks of using finer voxels, or dynamic voxelization. In both cases the current voxelization does not scale (can take up to 100ms for some of the cases considered in our internal ml repository). For these cases, I am testing other implementations (e.g., hash), which would immediately make this inapplicable.
 - This is ML talk, but when deploying a ML algorithm in a distribution other than the one used for training, there will be inevitably some sort of degradation in performance (compared to the theoretical case of training on the right distribution). While we can not control all the distribution changes, we should not go out of our way to add more. Yes, adding voxels with priority is better than what happens with a low max number of voxels and without this PR, but remember that the induced distributions of the voxels when projected into the BEV space will change quite a lot, which is prone to cause unexpected phenomena

technolojin:
@knzo25 
Thank you for your comments and opinion.

> it introduces random access 

Yes

> Going over the limit should not happen for a specific product/project. 

I agree. It is for worst cases.

> I would want it to be enabled automatically when the first overflow of max voxels occurs, 

The implementation will be complicated.

> or as a parameter.

May be better option for users.

> Looking ahead, there are talks of using finer voxels, or dynamic voxelization. In both cases the current voxelization does not scale (can take up to 100ms for some of the cases considered in our internal ml repository). For these cases, I am testing other implementations (e.g., hash), which would immediately make this inapplicable.

If better solution exist and can solve the current risk in a short time, I do not think this PR is needed. Are those will come soon?

> distributions of the voxels when projected into the BEV space will change quite a lot, which is prone to cause unexpected phenomena 

It may depends on the model. Is not the voxel inputs are treated as a [set](https://en.wikipedia.org/wiki/Set_(mathematics))? 
Only voxels filled by points will be collected, therefore spatial order is fluctuated anyway.


## Alternative logic proposal
In current logic, the voxels are filled from right-hand side. 
* before
```
  int voxel_idx = floorf((x - min_x_range) / pillar_x_size);
  int voxel_idy = floorf((y - min_y_range) / pillar_y_size);
  unsigned int voxel_index = voxel_idy * grid_x_size + voxel_idx;
```

We can change it to fill from front to rear. 
* after
```
  int voxel_idx = floorf((x - min_x_range) / pillar_x_size);
  int voxel_idy = floorf((y - min_y_range) / pillar_y_size);
  unsigned int voxel_index = (grid_x_size - voxel_idx - 1) * grid_y_size + voxel_idy;
```

- Not a random access
- its only orientation difference. spatial order may similar to the previous implementation

@knzo25 How about this?

technolojin:
Alternative PR https://github.com/autowarefoundation/autoware.universe/pull/9608

technolojin:
This PR will be replaced by https://github.com/autowarefoundation/autoware.universe/pull/9608

Issue #7563, opened 2024-06-18
Misdetects the pedestrian near a car


sgNicola:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I'm convinced that this is not my fault but a bug.

### Description

The perception system fails to detect and track the pedestrian when they are close to the car. It only recognizes the pedestrian once they move away from the vehicle.
![image](https://github.com/autowarefoundation/autoware.universe/assets/115910640/707427b6-0e39-4aee-a937-f74768303fa7)
![walker](https://github.com/autowarefoundation/autoware.universe/assets/115910640/7e8394d0-915a-4778-b32f-dcf13a3fc680)
Here is the video of the rviz2.
https://github.com/autowarefoundation/autoware.universe/assets/115910640/b538ca29-b0c7-4f02-a50a-64bf7100e8c2

### Expected behavior

Detect the pedestrian correctly

### Actual behavior

Miss detects the pedestrian near the car.

### Steps to reproduce

1. cd autoware/src/sensor_kit/external/awsim_sensor_kit_launch/awsim_sensor_kit_launch/launch，modify the awsim_sensor_kit, only one top lidar is used.
![image](https://github.com/autowarefoundation/autoware.universe/assets/115910640/29eb1629-83a9-4ffe-80b9-040e48a65ddb)
2. ros2 launch autoware_launch logging_simulator.launch.xml map_path:=map_path vehicle_model:=sample_vehicle sensor_model:=awsim_sensor_kit
3. replay this bag [https://drive.google.com/file/d/1FFE1XTrY-A8ELE1zi32bvGfe-Hah4u7P/view?usp=sharing](url)

### Versions

-OS: Ubuntu 22.04
-ROS2: Humble
-Autoware: Autoware.universe main

### Possible causes

The point clouds of pedestrians and cars are connected together， and not recognize correctly

sgNicola:
Here is the video of the rviz2 [https://github.com/autowarefoundation/autoware.universe/assets/115910640/b538ca29-b0c7-4f02-a50a-64bf7100e8c2](url)

badai-nguyen:
@sgNicola Thank you for your issue report. 
I think the current DNN model (Centerpoint) is not good enough for pedestrians nearby the vehicle, especially for the large vehicle like bus because of lack of training dataset. 
[...]

PR #1935, opened 2022-09-22
feat(lidar_centerpoint): add IoU-based NMS

yukke42:
Signed-off-by: yukke42 <yusuke.muramatsu@tier4.jp>

## Description

In lidar_centerpoint, center-distance-based non-maximum-suppression (NMS) has been used to remove overlapped objects, but it sometimes fails to do it especially for large objects, s.t. trucks and trailers. I add iou-based NMS to remove overlapped objects without any affects to the small objects.
And then, I change the `circle_nms_dist_threshold` param not to remove overlapped pedestrians that are close to each other.

before:
Screenshot from 2022-09-22 16-18-34

Screenshot from 2022-09-22 18-08-45

after:

Screenshot from 2022-09-22 16-21-50

Screenshot from 2022-09-22 18-08-06

## Related links

- [TIERIV INTERNAL LINK](https://tier4.atlassian.net/browse/T4PB-21175)

## Pre-review checklist for the PR author

The PR author **must** check the checkboxes below when creating the PR.

- [x] I've confirmed the [contribution guidelines].
- [x] The PR follows the [pull request guidelines].

## In-review checklist for the PR reviewers

The PR reviewers **must** check the checkboxes below before approval.

- [ ] The PR follows the [pull request guidelines].
- [ ] The PR has been properly tested.
- [ ] The PR has been reviewed by the code owners.

## Post-review checklist for the PR author

The PR author **must** check the checkboxes below before merging.

- [ ] There are no open discussions or they are tracked via tickets.
- [ ] The PR is ready for merge.

After all checkboxes are checked, anyone who has write access can merge the PR.

[contribution guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/
[pull request guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/pull-request-guidelines/

PR #9595, opened 2024-12-09
fix(lidar_centerpoint): non-maximum suppression target decision logic

technolojin:
## Description

1. fix non-maximum suppression target decision logic
    Before: If both of the object class are in the target label, caluclate IoU. If not, check distance and calculate IoU if the distance is less than a threshold.
    After: If the label pair is not the same and one of them is pedestrian, do not suppress. Then, check distance and calcuclate IoU only if the distance is within a threshold.
 
2. optimize non-maximum suppression search range
   Keep the squared value of the distance threshold.

3. remove NMS type
  Since there is no other NMS option, remove the logic and parameter to select NMS type.

## Related links

NMS PR: https://github.com/autowarefoundation/autoware.universe/pull/1935

Issue: 
 - pedestrian near a car is not detected https://github.com/autowarefoundation/autoware.universe/issues/7563 

## How was this PR tested?
Performance test https://evaluation.tier4.jp/evaluation/reports/16239dd4-5213-5843-ada0-49341293bb6d?project_id=prd_jt
Execution test https://evaluation.tier4.jp/evaluation/reports/26db2be4-4fe5-5fe9-8221-320a88f318d7?project_id=prd_jt

## Notes for reviewers

Remove unused parameter https://github.com/autowarefoundation/autoware_launch/pull/1267 will be merged after this PR.

## Interface changes

Parameter `iou_nms_target_class_names` will be removed.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_image_projection_based_fusion/config/pointpainting.param.yaml
+++ b/perception/autoware_image_projection_based_fusion/config/pointpainting.param.yaml
@@ -9,7 +9,6 @@
     post_process_params:
       # post-process params
       circle_nms_dist_threshold: 0.3
-      iou_nms_target_class_names: ["CAR"]
       iou_nms_search_distance_2d: 10.0
       iou_nms_threshold: 0.1
       score_threshold: 0.35
--- a/perception/autoware_lidar_centerpoint/include/autoware/lidar_centerpoint/postprocess/non_maximum_suppression.hpp
+++ b/perception/autoware_lidar_centerpoint/include/autoware/lidar_centerpoint/postprocess/non_maximum_suppression.hpp
@@ -45,18 +45,6 @@ struct NMSParams
   // double distance_threshold_{};
 };
 
-std::vector<bool> classNamesToBooleanMask(const std::vector<std::string> & class_names)
-{
-  std::vector<bool> mask;
-  constexpr std::size_t num_object_classification = 8;
-  mask.resize(num_object_classification);
-  for (const auto & class_name : class_names) {
-    const auto semantic_type = getSemanticType(class_name);
-    mask.at(semantic_type) = true;
-  }
-
-  return mask;
-}
 
 class NonMaximumSuppression
 {
@@ -66,14 +54,12 @@ public:
   std::vector<DetectedObject> apply(const std::vector<DetectedObject> &);
 
 private:
-  bool isTargetLabel(const std::uint8_t);
 
   bool isTargetPairObject(const DetectedObject &, const DetectedObject &);
 
   Eigen::MatrixXd generateIoUMatrix(const std::vector<DetectedObject> &);
 
   NMSParams params_{};
-  std::vector<bool> target_class_mask_{};
 
   double search_distance_2d_sq_{};
 };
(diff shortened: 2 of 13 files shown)


kminoda:
@technolojin Would you separate the PR for transfusion? I want to make the PR as small as possible (OK for image_projection_based_fusion and centerpoint as those two packages are dependent)

perception/autoware_image_projection_based_fusion/config/pointpainting.param.yaml
```diff
@@ -9,7 +9,6 @@
     post_process_params:
       # post-process params
       circle_nms_dist_threshold: 0.3
-      iou_nms_target_class_names: ["CAR"]
```
kminoda, line 12: Why are we removing this parameter?

(Also, if we change this, don't we need to update "Interface changes" part?)

perception/autoware_lidar_centerpoint/include/autoware/lidar_centerpoint/postprocess/non_maximum_suppression.hpp
```diff
@@ -28,36 +28,12 @@ namespace autoware::lidar_centerpoint
...
 using autoware_perception_msgs::msg::DetectedObject;
 
-// TODO(yukke42): now only support IoU_BEV
-enum class NMS_TYPE {
-  IoU_BEV
-  // IoU_3D
-  // Distance_2D
-  // Distance_3D
-};
-
```
kminoda, line 38: Why do we remove this in this PR? I kind of understand the benefit of this modification, but even so, would be better to do it in another PR.

perception/autoware_image_projection_based_fusion/config/pointpainting.param.yaml
technolojin: > Interface changes
I mentioned this change. Thanks for remind.

The parameter seems to be implemented only for those classes, but the implementation didn't limit the NMS process
As of today, the NMS should be implemented all of the classes. Only exemption is between pedestrian and non-pedestrian.

technolojin:
> @technolojin Would you separate the PR for transfusion? I want to make the PR as small as possible (OK for image_projection_based_fusion and centerpoint as those two packages are dependent)

@kminoda 
Then, the maintainer of the transfusion will decide if the NMS logic to be fixed or not.

perception/autoware_image_projection_based_fusion/config/pointpainting.param.yaml
kminoda: Ah true, now make sense to me, thanks for the explanation 🙏

perception/autoware_lidar_centerpoint/include/autoware/lidar_centerpoint/postprocess/non_maximum_suppression.hpp
technolojin: The logic checks the NMS type everytime, but this logic do not support multiple NMS models.
Even you have plan to implement other mode, this should come at that moment.

technolojin: This PR is to revise the NMS logic and improve the performance. Under this concept, removing unused parameters and logic is also included.

kminoda:
@technolojin I'm OK with it.
(We had some PRs that changes transfusion and centerpoint at the same time, and this made the cherry-picking in some of our projects very difficult. My point here is to avoid such potential conflicts)

perception/autoware_lidar_centerpoint/include/autoware/lidar_centerpoint/postprocess/non_maximum_suppression.hpp
kminoda: OK, understood. (Also the description update LGTM)

kminoda (approved):
LGTM

PR #9596, opened 2024-12-09
feat: remove max rois limit in the image projection based fusion

YoshiRi:
## Description

Currently, nodes in the image_projection_based_fusion package are restricted to subscribe up to 8 roi messages maximum.
Remove this magic number because it does not have any concrete necessity.

## How was this PR tested?

Adding launcher to subscribe more than 8 camera topics.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


badai-nguyen (approved):
LGTM

PR #9612, opened 2024-12-11
fix(autoware_lidar_transfusion): non-maximum suppression target decision logic

technolojin:
## Description
Implement same fix of PR https://github.com/autowarefoundation/autoware.universe/pull/9595

## Related links
1.  fix non-maximum suppression target decision logic
    Before: If both of the object class are in the target label, caluclate IoU. If not, check distance and calculate IoU if the distance is less than a threshold.
    After: If the label pair is not the same and one of them is pedestrian, do not suppress. Then, check distance and calcuclate IoU only if the distance is within a threshold.

2.  optimize non-maximum suppression search range
    Keep the squared value of the distance threshold.

3.  remove NMS type
    Since there is no other NMS option, remove the logic and parameter to select NMS type.

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_lidar_transfusion/config/transfusion.param.yaml
+++ b/perception/autoware_lidar_transfusion/config/transfusion.param.yaml
@@ -10,7 +10,6 @@
     densification_world_frame_id: map
     # post-process params
     circle_nms_dist_threshold: 0.5
-    iou_nms_target_class_names: ["CAR"]
     iou_nms_search_distance_2d: 10.0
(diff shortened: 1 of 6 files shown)


technolojin:
@amadeuszsz 
Can you test and review this PR?
I do not have the model.

amadeuszsz (approved):
LGTM!

PR #9626, opened 2024-12-12
feat(autoware_multi_object_tracker): tracker update with twist x and y

technolojin:
## Description
Since detected object can have twist of x and y, the tracker can update both velocity and slip.
 

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


technolojin:
implemented in https://github.com/autowarefoundation/autoware_universe/pull/11230

PR #9635, opened 2024-12-12
feat(image_projection_based_fusion): add cache for camera projection

a-maumau:
## Description
This PR introduces a caching feature for calculating raw image projections (unrectifyPoint).
This PR must merged with https://github.com/autowarefoundation/autoware_launch/pull/1275

It introduces the CameraProjection class to handle the related processes. The cache is computed based on specific grid centers, and the projected points are represented using these grid centers. These grid centers are determined by `approximation_grid_cell_width` and `approximation_grid_cell_height`.
Pre-cached result will be simply stored in an array and all grid cells on the image will be pre-computed.

For example, when approximation_grid_cell_\* is set to 1.0 pixel, the grid centers will be created as shown in the following image:
image_projection_based_fusion_PR_image

***Note that edge pixels ((w, \*), (\*, h)) will be rounded to fall outside the image bounds***.

#### memory consumption
When using caching, memory consumption is calculated as:

ceil(image_width / grid_cell_width) \* ceil(image_height / grid_cell_height) \* (4 \* 2) \* roi_numbers [bytes]

Here, (4 \* 2) represents the bytes for two float values.
For example, for an image with dimensions 1440 × 1080, a grid size of 1 pixel, and caching enabled for 6 cameras, the memory consumption is:

(1440 / 1.0) \* (1080 / 1.0) \* 4 \* 2 \* 6 / (1024 \* 1024) ≈ 71.2 MiB (theoretical).

Also when using the cache feature, it requires an initialization step (pre-caching). This will make some delay to a node before starting the processing. 

### Other changes
- This PR fixes a bug where the camera parameters were used before being properly initialized in `pointpainting_fusion`.
- take a list input instead of single bool for `point_project_to_unrectified_image` param

## Performance: current impl. vs. this PR
- The following result is done with the same rosbag file, but the starting is determined by the (almost) first topic from the `map_based_prediction` node, so there is no guarantee it is aligned. 
- The processing time is timed with TimeKeeper (subCallback is timed).
- CPU usage and memory usage is collected with [autoware_tools](https://github.com/autowarefoundation/autoware_tools/tree/main/common/autoware_debug_tools)
- The grid cell size is set to a square (width == height), and the camera (roi) number is set to 6.
- `baseline` indicates the current implementation, `unrectified` is the `point_project_to_unrectified_image` and `cache` means `approximate_camera_projection` option in the config.
#### roi_cluster_fusion
- all processing times
all_roi_cluster_fusion_processing_time
 
- current `point_project_to_unrectified_image=false` vs. this PR `point_project_to_unrectified_image=true, cache=true`
roi_cluster_fusion_baseline_vs_cache
This shows the processing time is close to the case of `point_project_to_unrectified_image=false` case when we use the cache feature.

#### pointpainting_fusion
- all processing times
all_pointpainting_processing_time
- current `point_project_to_unrectified_image=false` vs. this PR `point_project_to_unrectified_image=true, cache=true`
pointpainting_baseline_vs_cache

- current `point_project_to_unrectified_image=true` vs. this PR `point_project_to_unrectified_image=true, no cache`
pointpainting_rectified_true_no_cache_compare
This performance difference should be coming from the fix of the bug in FOV filtering (`camera_info` was not properly initialized).

#### roi_cluster_fusion
roi_cluster_fusion_compare_cpu_usage
Plot length differ because of the difference of play rate (0.1x is 3 times longer than 0.3x).
It does not contains the CPU usage of initialization step (which might use 100% of CPU).

#### roi_cluster_fusion
roi_cluster_fusion_compare_memory_usage
Plot length differ because of the difference of play rate (0.1x is 3 times longer than 0.3x).

## Visualization
Following images ware not captured at the same frame.
example_compare_of_grid_cell_size
All images are in the case of `point_project_to_unrectified_image=true`.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?
Checked the rviz output and each fusion node's visualization.
output_is_normal

For checking by yourself, you need some processing like
```
# decompress image
ros2 launch autoware_image_transport_decompressor image_transport_decompressor.launch.xml input_topic_name:=/sensing/camera/camera0/image_rect_color/compressed output_topic_name:=/sensing/camera/camera0/image_rect_color

# relay camera info
ros2 run topic_tools relay /sensing/camera/camera0/camera_info /perception/object_recognition/detection/pointpainting/pointpainting/debug/camera_info
```
and set the `debug_mode: true` in param. For `pointpainting_fusion`, you need to set the `#if 0` to `#if 1` to contain the points in an image.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_image_projection_based_fusion/include/autoware/image_projection_based_fusion/camera_projection.hpp
+++ b/perception/autoware_image_projection_based_fusion/include/autoware/image_projection_based_fusion/camera_projection.hpp
@@ -40,7 +40,7 @@ public:
   explicit CameraProjection(
     const sensor_msgs::msg::CameraInfo & camera_info, const float grid_width,
     const float grid_height, const bool unrectify, const bool use_approximation);
-  CameraProjection() : grid_w_size_(1.0), grid_h_size_(1.0), unrectify_(false) {};
+  CameraProjection() : grid_w_size_(1.0), grid_h_size_(1.0), unrectify_(false) {}
   void initialize();
   std::function<bool(const cv::Point3d &, Eigen::Vector2d &)> calcImageProjectedPoint;
   sensor_msgs::msg::CameraInfo getCameraInfo();
--- a/perception/autoware_image_projection_based_fusion/include/autoware/image_projection_based_fusion/camera_projection.hpp
+++ b/perception/autoware_image_projection_based_fusion/include/autoware/image_projection_based_fusion/camera_projection.hpp
@@ -18,14 +18,12 @@
 #define EIGEN_MPL2_ONLY
 
 #include <Eigen/Core>
-#include <autoware/universe_utils/system/lru_cache.hpp>
 #include <opencv2/core/core.hpp>
 
 #include <sensor_msgs/msg/camera_info.hpp>
 
 #include <image_geometry/pinhole_camera_model.h>
 
-#include <map>
 #include <memory>
 
 namespace autoware::image_projection_based_fusion
--- a/perception/autoware_image_projection_based_fusion/include/autoware/image_projection_based_fusion/camera_projection.hpp
+++ b/perception/autoware_image_projection_based_fusion/include/autoware/image_projection_based_fusion/camera_projection.hpp
@@ -18,12 +18,10 @@
 #define EIGEN_MPL2_ONLY
 
 #include <Eigen/Core>
-
 #include <autoware/universe_utils/system/lru_cache.hpp>
+#include <opencv2/core/core.hpp>
 
 #include <sensor_msgs/msg/camera_info.hpp>
-
-#include <opencv2/core/core.hpp>
 
 #include <image_geometry/pinhole_camera_model.h>
 
@@ -42,12 +40,9 @@ class CameraProjection
 {
 public:
   explicit CameraProjection(
-    const sensor_msgs::msg::CameraInfo & camera_info,
-    const float grid_width, const float grid_height,
-    const bool unrectify,
-    const bool use_approximation
-  );
-  CameraProjection(): grid_w_size_(1.0), grid_h_size_(1.0), unrectify_(false) {};
+    const sensor_msgs::msg::CameraInfo & camera_info, const float grid_width,
+    const float grid_height, const bool unrectify, const bool use_approximation);
+  CameraProjection() : grid_w_size_(1.0), grid_h_size_(1.0), unrectify_(false) {};
   void initialize();
   std::function<bool(const cv::Point3d &, Eigen::Vector2d &)> calcImageProjectedPoint;
   sensor_msgs::msg::CameraInfo getCameraInfo();
@@ -55,9 +50,11 @@ public:
   bool isOutsideVerticalView(const float py, const float pz);
 
 protected:
-  bool calcRectifiedImageProjectedPoint(const cv::Point3d & point3d, Eigen::Vector2d & projected_point);
+  bool calcRectifiedImageProjectedPoint(
+    const cv::Point3d & point3d, Eigen::Vector2d & projected_point);
   bool calcRawImageProjectedPoint(const cv::Point3d & point3d, Eigen::Vector2d & projected_point);
-  bool calcRawImageProjectedPointWithApproximation(const cv::Point3d & point3d, Eigen::Vector2d & projected_point);
+  bool calcRawImageProjectedPointWithApproximation(
+    const cv::Point3d & point3d, Eigen::Vector2d & projected_point);
   void initializeCache();
 
   sensor_msgs::msg::CameraInfo camera_info_;
--- a/perception/autoware_image_projection_based_fusion/include/autoware/image_projection_based_fusion/camera_projection.hpp
+++ b/perception/autoware_image_projection_based_fusion/include/autoware/image_projection_based_fusion/camera_projection.hpp
@@ -68,8 +68,8 @@ protected:
   float half_grid_h_size_;
   float inv_grid_w_size_;
   float inv_grid_h_size_;
-  uint32_t grid_x_num_;
-  uint32_t grid_y_num_;
+  int grid_x_num_;
+  int grid_y_num_;
   float index_grid_out_h_;
   float index_grid_out_w_;
 
--- a/perception/autoware_image_projection_based_fusion/include/autoware/image_projection_based_fusion/camera_projection.hpp
+++ b/perception/autoware_image_projection_based_fusion/include/autoware/image_projection_based_fusion/camera_projection.hpp
@@ -1,0 +1,88 @@
+// Copyright 2024 TIER IV, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef AUTOWARE__IMAGE_PROJECTION_BASED_FUSION__CAMERA_PROJECTION_HPP_
+#define AUTOWARE__IMAGE_PROJECTION_BASED_FUSION__CAMERA_PROJECTION_HPP_
+
+#define EIGEN_MPL2_ONLY
+
+#include <Eigen/Core>
+
+#include <autoware/universe_utils/system/lru_cache.hpp>
+
+#include <sensor_msgs/msg/camera_info.hpp>
+
+#include <opencv2/core/core.hpp>
+
+#include <image_geometry/pinhole_camera_model.h>
+
+#include <map>
+#include <memory>
+
+namespace autoware::image_projection_based_fusion
+{
+struct PixelPos
+{
+  float x;
+  float y;
+};
+
+class CameraProjection
+{
+public:
+  explicit CameraProjection(
+    const sensor_msgs::msg::CameraInfo & camera_info,
+    const float grid_width, const float grid_height,
+    const bool unrectify,
+    const bool use_approximation
+  );
+  CameraProjection(): grid_w_size_(1.0), grid_h_size_(1.0), unrectify_(false) {};
+  void initialize();
+  std::function<bool(const cv::Point3d &, Eigen::Vector2d &)> calcImageProjectedPoint;
+  sensor_msgs::msg::CameraInfo getCameraInfo();
+  bool isOutsideHorizontalView(const float px, const float pz);
+  bool isOutsideVerticalView(const float py, const float pz);
+
+protected:
+  bool calcRectifiedImageProjectedPoint(const cv::Point3d & point3d, Eigen::Vector2d & projected_point);
+  bool calcRawImageProjectedPoint(const cv::Point3d & point3d, Eigen::Vector2d & projected_point);
+  bool calcRawImageProjectedPointWithApproximation(const cv::Point3d & point3d, Eigen::Vector2d & projected_point);
+  void initializeCache();
+
+  sensor_msgs::msg::CameraInfo camera_info_;
+  uint32_t image_h_, image_w_;
+  double tan_h_x_, tan_h_y_;
+
+  uint32_t cache_size_;
+  float grid_w_size_;
+  float grid_h_size_;
+  float half_grid_w_size_;
+  float half_grid_h_size_;
+  float inv_grid_w_size_;
+  float inv_grid_h_size_;
+  uint32_t grid_x_num_;
+  uint32_t grid_y_num_;
+  float index_grid_out_h_;
+  float index_grid_out_w_;
+
+  bool unrectify_;
+  bool use_approximation_;
+
+  std::unique_ptr<PixelPos[]> projection_cache_;
+  image_geometry::PinholeCameraModel camera_model_;
+};
+
+}  // namespace autoware::image_projection_based_fusion
+
+#endif  // AUTOWARE__IMAGE_PROJECTION_BASED_FUSION__CAMERA_PROJECTION_HPP_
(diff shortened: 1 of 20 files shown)


technolojin:
Can you also update the schema file?
https://github.com/autowarefoundation/autoware.universe/blob/main/perception/autoware_image_projection_based_fusion/schema/fusion_common.schema.json

perception/autoware_image_projection_based_fusion/src/camera_projection.cpp
```diff
@@ -0,0 +1,264 @@
...
+  }
+  const float qy =
+    std::round((rectified_image_point.y - half_grid_h_size_) * inv_grid_h_size_) * grid_h_size_ +
+    half_grid_h_size_;
+  if (qy < 0.0 || qy >= image_h_) {
+    return false;
+  }
+
+  const int grid_x = static_cast<int>(std::floor(qx / grid_w_size_));
+  const int grid_y = static_cast<int>(std::floor(qy / grid_h_size_));
```
technolojin, line 219:
```suggestion
  // round to a near grid center
  const float qx =
    std::round((rectified_image_point.x - half_grid_w_size_) * inv_grid_w_size_)  + 0.5f;
  if (qx < 0.0 || qx >= image_w_ /grid_w_size_) {
    return false;
  }
  const float qy =
    std::round((rectified_image_point.y - half_grid_h_size_) * inv_grid_h_size_)  + 0.5f;
  if (qy < 0.0 || qy >= image_h_/grid_h_size_) {
    return false;
  }

  const int grid_x = static_cast<int>(std::floor(qx));
  const int grid_y = static_cast<int>(std::floor(qy));
```

may it can be more efficient
(need to be tested)

a-maumau:
@technolojin I updated the schema file in https://github.com/autowarefoundation/autoware.universe/pull/9635/commits/acd69a8c0cc64463b856f688ccd76c54b64b67fb

perception/autoware_image_projection_based_fusion/src/camera_projection.cpp
```diff
@@ -0,0 +1,264 @@
...
+        std::round((x - half_grid_w_size_) * inv_grid_w_size_) * grid_w_size_ + half_grid_w_size_;
+      const float qy =
+        std::round((y - half_grid_h_size_) * inv_grid_h_size_) * grid_h_size_ + half_grid_h_size_;
+
+      const int grid_x = static_cast<int>(std::floor(qx / grid_w_size_));
+      const int grid_y = static_cast<int>(std::floor(qy / grid_h_size_));
+      const int index = (grid_y)*grid_x_num_ + grid_x;
+
+      // precompute projected point
+      cv::Point2d raw_image_point = camera_model_.unrectifyPoint(cv::Point2d(qx, qy));
```
technolojin, line 149:
```suggestion
  for (int grid_x = 0; y < image_h_/grid_h_size_; y++) {
    for (int grid_x = 0; x < image_w_ /grid_w_size_; x++) {
      const int index = (grid_y)*grid_x_num_ + grid_x;
      const float qx = (grid_x + 0.5f) * grid_w_size_;
      const float qy = (grid_y + 0.5f) * grid_h_size_;
      
      // precompute projected point
      cv::Point2d raw_image_point = camera_model_.unrectifyPoint(cv::Point2d(qx, qy));
```
Better readability and efficiency.
(need to be tested)

technolojin (commented):
I suggest a fix in the cache generation and projection implementation.

Can you test it?

perception/autoware_image_projection_based_fusion/src/camera_projection.cpp
technolojin:
```suggestion
  // round to a near grid center
  const int grid_x = std::floor(rectified_image_point.x * inv_grid_w_size_);
  const int grid_y = std::floor(rectified_image_point.y * inv_grid_h_size_);
  
  if  (grid_x < 0.0 || grid_x>= static_cast<int>(image_w_/grid_w_size_)) return false;
  if  (grid_y < 0.0 || grid_y>= static_cast<int>(image_h_/grid_h_size_)) return false;
```

more simple like this? is it the same?

technolojin:
Let us align naming about grid and grid cells

Grid: A structure consist of repeated geometry. Consists by grid cells.
(Grid) Cell: A geometry element consisting the grid.

By this definition, parameters and variables need to be renamed.

  

- grid_w_size_ -> cell_w_size_ : width size of a grid cell
- (new) grid_width_ : total number of cells in width direction of the grid, calculated by `grid_width_ = image_width / cell_w_size_`

perception/autoware_image_projection_based_fusion/src/camera_projection.cpp
a-maumau: Thank you for the great suggestion!
I confirm the suggestions result in same values and behavior except for the (0, \*), (\*, 0) pixels.
(0, \*), (\*, 0) pixels will be get inside the grid which is better than the original one.

I fix some of the code and added in https://github.com/autowarefoundation/autoware.universe/pull/9635/commits/e8b5b2ffdcfcca886f51c9d58f26f623356a5932

perception/autoware_image_projection_based_fusion/src/camera_projection.cpp
a-maumau: Thank you for the suggestion!
As same as the above suggestion, it is added in https://github.com/autowarefoundation/autoware.universe/pull/9635/commits/e8b5b2ffdcfcca886f51c9d58f26f623356a5932.

a-maumau:
@technolojin I changed the terms in https://github.com/autowarefoundation/autoware.universe/pull/9635/commits/bc1e8ed63118455a65313e171af4360f32f2dc36

perception/autoware_image_projection_based_fusion/src/camera_projection.cpp
```diff
@@ -0,0 +1,250 @@
...
+
+  // round to a near grid cell
+  const int grid_x = static_cast<int>(std::floor(rectified_image_point.x * inv_cell_width_));
+  const int grid_y = static_cast<int>(std::floor(rectified_image_point.y * inv_cell_height_));
+
+  if (grid_x < 0.0 || grid_x >= grid_width_) return false;
+  if (grid_y < 0.0 || grid_y >= grid_height_) return false;
+
+  const uint32_t index = grid_y * grid_width_ + grid_x;
+  projected_point << projection_cache_[index].x, projection_cache_[index].y;
```
technolojin, line 208:
```suggestion
  // round to a near grid cell
  const int grid_idx_x = static_cast<int>(std::floor(rectified_image_point.x * inv_cell_width_));
  const int grid_idx_y = static_cast<int>(std::floor(rectified_image_point.y * inv_cell_height_));

  if (grid_idx_x < 0.0 || grid_idx_x >= grid_width_) return false;
  if (grid_idx_y < 0.0 || grid_idx_y >= grid_height_) return false;

  const uint32_t grid_index = grid_idx_y * grid_width_ + grid_idx_x;
  projected_point << projection_cache_[grid_index].x, projection_cache_[grid_index].y;
```

perception/autoware_image_projection_based_fusion/src/camera_projection.cpp
```diff
@@ -0,0 +1,250 @@
...
+      const float qx = (grid_x + 0.5f) * cell_width_;
+      const float qy = (grid_y + 0.5f) * cell_height_;
+      const uint32_t index = grid_y * grid_width_ + grid_x;
+
+      // precompute projected point
+      cv::Point2d raw_image_point = camera_model_.unrectifyPoint(cv::Point2d(qx, qy));
+      projection_cache_[index] =
+        PixelPos{static_cast<float>(raw_image_point.x), static_cast<float>(raw_image_point.y)};
+    }
+  }
```
technolojin, line 149:
```suggestion
  for (int idx_y = 0; idx_y < grid_height_; idx_y++) {
    for (int idx_x = 0; idx_x < grid_width_; idx_x++) {
      const uint32_t grid_index = idx_y * grid_width_ + idx_x;
      const float qx = (idx_x + 0.5f) * cell_width_;
      const float qy = (idx_y + 0.5f) * cell_height_;
      
      // precompute projected point
      cv::Point2d raw_image_point = camera_model_.unrectifyPoint(cv::Point2d(qx, qy));
      projection_cache_[grid_index] =
        PixelPos{static_cast<float>(raw_image_point.x), static_cast<float>(raw_image_point.y)};
    }
  }
```

perception/autoware_image_projection_based_fusion/src/camera_projection.cpp
```diff
@@ -0,0 +1,250 @@
...
+  image_h_ = camera_info.height;
+
+  // prepare camera model
+  camera_model_.fromCameraInfo(camera_info);
+
+  // cache settings
+  inv_cell_width_ = 1 / cell_width_;
+  inv_cell_height_ = 1 / cell_height_;
+  grid_width_ = static_cast<int>(std::ceil(image_w_ / cell_width_));
+  grid_height_ = static_cast<int>(std::ceil(image_h_ / cell_height_));
```
technolojin, line 44:
```suggestion
  image_width_ = camera_info.width;
  image_height_ = camera_info.height;

  // prepare camera model
  camera_model_.fromCameraInfo(camera_info);

  // cache settings
  inv_cell_width_ = 1 / cell_width_;
  inv_cell_height_ = 1 / cell_height_;
  grid_width_ = static_cast<int>(std::ceil(image_width_ / cell_width_));
  grid_height_ = static_cast<int>(std::ceil(image_height_ / cell_height_));
```

perception/autoware_image_projection_based_fusion/src/camera_projection.cpp
```diff
@@ -0,0 +1,250 @@
...
+  const cv::Point3d ray_top_right = camera_model_.projectPixelTo3dRay(cv::Point(image_w_ - 1, 0));
+  const cv::Point3d ray_bottom_left = camera_model_.projectPixelTo3dRay(cv::Point(0, image_h_ - 1));
+  const cv::Point3d ray_bottom_right =
+    camera_model_.projectPixelTo3dRay(cv::Point(image_w_ - 1, image_h_ - 1));
+  // for checking barrel shape case
+  const cv::Point3d ray_mid_top = camera_model_.projectPixelTo3dRay(cv::Point(ocx, 0));
+  const cv::Point3d ray_mid_left = camera_model_.projectPixelTo3dRay(cv::Point(0, ocy));
+  const cv::Point3d ray_mid_right = camera_model_.projectPixelTo3dRay(cv::Point(image_w_ - 1, ocy));
+  const cv::Point3d ray_mid_bottom =
+    camera_model_.projectPixelTo3dRay(cv::Point(ocx, image_h_ - 1));
```
technolojin, line 63:
```suggestion
  // for checking pincushion shape case
  const cv::Point3d ray_top_left = camera_model_.projectPixelTo3dRay(cv::Point(0, 0));
  const cv::Point3d ray_top_right = camera_model_.projectPixelTo3dRay(cv::Point(image_width_ - 1, 0));
  const cv::Point3d ray_bottom_left = camera_model_.projectPixelTo3dRay(cv::Point(0, image_height_ - 1));
  const cv::Point3d ray_bottom_right =
    camera_model_.projectPixelTo3dRay(cv::Point(image_width_ - 1, image_height_ - 1));
  // for checking barrel shape case
  const cv::Point3d ray_mid_top = camera_model_.projectPixelTo3dRay(cv::Point(ocx, 0));
  const cv::Point3d ray_mid_left = camera_model_.projectPixelTo3dRay(cv::Point(0, ocy));
  const cv::Point3d ray_mid_right = camera_model_.projectPixelTo3dRay(cv::Point(image_width_ - 1, ocy));
  const cv::Point3d ray_mid_bottom =
    camera_model_.projectPixelTo3dRay(cv::Point(ocx, image_height_ - 1));
```

perception/autoware_image_projection_based_fusion/include/autoware/image_projection_based_fusion/camera_projection.hpp
```diff
@@ -0,0 +1,81 @@
...
+protected:
+  bool calcRectifiedImageProjectedPoint(
+    const cv::Point3d & point3d, Eigen::Vector2d & projected_point);
+  bool calcRawImageProjectedPoint(const cv::Point3d & point3d, Eigen::Vector2d & projected_point);
+  bool calcRawImageProjectedPointWithApproximation(
+    const cv::Point3d & point3d, Eigen::Vector2d & projected_point);
+  void initializeCache();
+
+  sensor_msgs::msg::CameraInfo camera_info_;
+  uint32_t image_h_, image_w_;
```
technolojin, line 60:
```suggestion
  uint32_t image_height_, image_width_;
```

technolojin (commented):
Just rename some variables.

* avoid to use too general name such as `index`: more specify like `grid_index`
* name as its role: `grid_x` -> `grid_idx_x`
* align naming style: since `cell_width_` is used, `image_w_` -> `image_width_`

technolojin:
Test the PR in [TIER IV INTERNAL CLOUD TESTER](https://evaluation.tier4.jp/evaluation/reports/47da4af1-5650-5ce9-a53a-ee471f9ea69b?project_id=prd_jt)

technolojin (approved):
LGTM

technolojin:
@badai-nguyen @YoshiRi @yukkysaito 
Dear code-owners, 

Can you review this PR? This PR has been tested.

badai-nguyen (commented):
@a-maumau Thank you for your great idea and PR. I have minor comments.

a-maumau:
@badai-nguyen Thank you for pointing out. I applied your suggestions.

perception/autoware_image_projection_based_fusion/schema/fusion_common.schema.json
```diff
@@ -65,6 +65,21 @@
           "type": "number",
           "description": "Maximum z position [m].",
           "default": 100.0
+        },
```
badai-nguyen, line 68: `point_project_to_unrectified_image` parameter was missing from previous PR, could also add it?
```suggestion
        },
        "point_project_to_unrectified_image": {
          "type": "array",
          "description": "An array of options indicating whether to project point to unrectified image or not.",
          "default": [false, false, false, false, false, false]
        },
```

technolojin: resolved https://github.com/autowarefoundation/autoware.universe/pull/9635/commits/3b0bbd0837696de640b608c17343789ee8fe70e5

badai-nguyen (approved):
LGTM

PR #9637, opened 2024-12-12
fix(autoware_image_projection_based_fusion): fix bug of fov checker in pointpainting

technolojin:
## Description

When the point is projected to the image space, the point is checked if the point is in the filed-of-view of corresponding camera.
It may cut unnecessary calculation of later roi matching.

However, @a-maumau found that the current implementation of loading camera parameter is done at the initialization and do not get the camera info.
```c++
    auto fx = camera_info_map_[roi_i].k.at(0); // fx = 0
    auto x0 = camera_info_map_[roi_i].k.at(2); // x0 = 0
    tan_h_[roi_i] = x0 / fx;                   // tan_h_[0] == NaN
```

```c++
    if (p_z <= 0.0 || p_x > (tan_h_.at(image_id) * p_z) || p_x < (-tan_h_.at(image_id) * p_z)) {
      continue;
    }
    // p_x > (tan_h_.at(image_id) * p_z) == false
    // p_x < (-tan_h_.at(image_id) * p_z) == false

```
Therefore the checker of the FoV was not done correctly.

This PR will not induce any logic difference.

Special thanks to @a-maumau 

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

Visually confirmed that the given roi and the pointcloud class is corresponding.

Screenshot from 2024-12-12 18-19-47

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_image_projection_based_fusion/src/pointpainting_fusion/node.cpp
+++ b/perception/autoware_image_projection_based_fusion/src/pointpainting_fusion/node.cpp
@@ -161,9 +161,7 @@ PointPaintingFusionNode::PointPaintingFusionNode(const rclcpp::NodeOptions & opt
 
   tan_h_.resize(rois_number_);
   for (std::size_t roi_i = 0; roi_i < rois_number_; ++roi_i) {
-    auto fx = camera_info_map_[roi_i].k.at(0);
-    auto x0 = camera_info_map_[roi_i].k.at(2);
-    tan_h_[roi_i] = x0 / fx;
+    tan_h_[roi_i] = 0;
   }
 
   detection_class_remapper_.setParameters(
@@ -272,6 +270,13 @@ void PointPaintingFusionNode::fuseOnSingleImage(
   }
 
   if (!checkCameraInfo(camera_info)) return;
+
+  // update camera fov
+  for (std::size_t roi_i = 0; roi_i < rois_number_; ++roi_i) {
+    auto fx = camera_info_map_[roi_i].k.at(0);
+    auto x0 = camera_info_map_[roi_i].k.at(2);
+    tan_h_[roi_i] = x0 / fx;
+  }
 
   std::vector<sensor_msgs::msg::RegionOfInterest> debug_image_rois;
   std::vector<Eigen::Vector2d> debug_image_points;
--- a/perception/autoware_image_projection_based_fusion/src/pointpainting_fusion/node.cpp
+++ b/perception/autoware_image_projection_based_fusion/src/pointpainting_fusion/node.cpp
@@ -349,11 +349,12 @@ dc   | dc dc dc  dc ||zc|
 
     /// check if the point is in the camera view
     if (p_z <= 0.0) continue;
+    if (p_x < fov_left_.at(image_id) * p_z) continue;
+    if (p_x > fov_right_.at(image_id) * p_z) continue;
+
     // project
     Eigen::Vector2d projected_point = calcRawImageProjectedPoint(
       pinhole_camera_model, cv::Point3d(p_x, p_y, p_z), point_project_to_unrectified_image_);
-    if (projected_point.x() < fov_left_.at(image_id)) continue;
-    if (projected_point.x() > fov_right_.at(image_id)) continue;
 
     // iterate 2d bbox
     for (const auto & feature_object : objects) {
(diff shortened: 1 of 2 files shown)


perception/autoware_image_projection_based_fusion/src/pointpainting_fusion/node.cpp
```diff
@@ -338,12 +347,13 @@ dc   | dc dc dc  dc ||zc|
...
-    if (p_z <= 0.0 || p_x > (tan_h_.at(image_id) * p_z) || p_x < (-tan_h_.at(image_id) * p_z)) {
-      continue;
-    }
+    /// check if the point is in the camera view
+    if (p_z <= 0.0) continue;
     // project
     Eigen::Vector2d projected_point = calcRawImageProjectedPoint(
       pinhole_camera_model, cv::Point3d(p_x, p_y, p_z), point_project_to_unrectified_image_);
+    if (projected_point.x() < fov_left_.at(image_id)) continue;
+    if (projected_point.x() > fov_right_.at(image_id)) continue;
```
a-maumau, line 356: I think this part is wrong since the `projected_point` is projected onto the image space.
After projecting the points onto the image plane, the check should be whether it is within [0, width-1].

So if we don't check before the projection, we don't need the `tan_h_ (fov_left_, fov_right_)` anymore, I guess.

technolojin: https://github.com/autowarefoundation/autoware.universe/blob/932e87fde0614e9342e8889c7961f103e615e4e3/perception/autoware_image_projection_based_fusion/src/pointpainting_fusion/node.cpp#L287-L298

`lidar2cam_affine` is transformation matrix between the frames. the target frame of the camera is to optic center. Therefore, it is not the image space but the camera coordinate system.

It can be checked experimentally by replace `fov_right_.at(image_id)` to zero. if there are points overwritten, then the projected coordinate is not in [0, width-1].

badai-nguyen, line 356: @technolojin Could you change to check fov before `calcRawImageProjectedPoint` function to avoid unnecessary projection calculation for points out of FOV?

technolojin: @badai-nguyen 
We do not know until the lens distortion is corrected. 
In the case the lens warps the view inside, the more points will be inside of FoV. 
In another word, if it comes before the correction, it may miss some points on the edges.

In the case the camera model is the pinhole camera and do not have lens distortion, the previous process `p_x > (tan_h_.at(image_id) * p_z)` will be done in the `calcRawImageProjectedPoint` and it means it doubles the same process anyway.

a-maumau: Sorry, which part are you talking about?
I agree about the `lidar2cam_affine` and the `point_camera`.
https://github.com/autowarefoundation/autoware.universe/blob/932e87fde0614e9342e8889c7961f103e615e4e3/perception/autoware_image_projection_based_fusion/src/pointpainting_fusion/node.cpp#L344

But the `projected_point` is on the image space (pixel coordinate) right?, or am I misunderstanding somethings?
https://github.com/autowarefoundation/autoware.universe/blob/932e87fde0614e9342e8889c7961f103e615e4e3/perception/autoware_image_projection_based_fusion/src/pointpainting_fusion/node.cpp#L353-L354

technolojin: Oh, project3dToPixel adds the optical center... Your understand is correct.

technolojin: resolved by https://github.com/autowarefoundation/autoware.universe/pull/9637/commits/355c183c4b549dac094a59483f0e2f3169dc5149

technolojin: resolved by https://github.com/autowarefoundation/autoware.universe/pull/9637/commits/355c183c4b549dac094a59483f0e2f3169dc5149

technolojin:
the issue will be fixed by related PR https://github.com/autowarefoundation/autoware.universe/pull/9635

PR #9663, opened 2024-12-16
feat(autoware_detected_object_validation): set validate distance in the obstacle pointcloud based validator 

technolojin:
## Description

A set of detected objects is validated by the obstacle pointcloud, which is the result of a rule-based detection `ground segmentation`.
In far ranges, however, it is difficult to separate non-ground obstacle since reflection of ground in far range is very week or absent due to the shallow angle.

A parameter `validate_max_distance_m` is added to not validate object far than this parameter.

## Related links

Launcher PR https://github.com/autowarefoundation/autoware_launch/pull/1277

## How was this PR tested?
[TIER IV INTERNAL](https://evaluation.tier4.jp/evaluation/reports/23555364-8459-5d70-a6ae-66a328a44a02?project_id=prd_jt)

### Experiment condition
* ground segmentation setting : change parameters to make obstacle pointcloud less in far distance (for test purpose)
  - global_slope_max_angle_deg: 30.0
  - local_slope_max_angle_deg: 30.0
  - use_recheck_ground_cluster: false  

* obstacle pointcloud based validator set a short validation distance (for test purpose)
  - validate_max_distance_m: 40.0

### Before
Blue bounding box: detected objects
Yellow bounding box: validated objects
White points: obstacle pointcloud

Screenshot from 2024-12-16 18-43-04

Since the obstacle pointclouds are fail to segment obstacles in far ranges, far objects are filtered out.

### After
Blue bounding box: detected objects
Yellow bounding box: validated objects

Screenshot from 2024-12-16 18-59-16

Even the obstacle pointclouds are absent, objects far than 40m could pass the validator.

## Notes for reviewers

The parameter `validate_max_distance_m` need to be set where the ground segmentation is less accurate. If this parameter is set too close, there is a risk that the detected is object is not validated.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_detected_object_validation/config/obstacle_pointcloud_based_validator.param.yaml
+++ b/perception/autoware_detected_object_validation/config/obstacle_pointcloud_based_validator.param.yaml
@@ -12,5 +12,7 @@
     #UNKNOWN,  CAR,    TRUCK,     BUS,    TRAILER, MOTORBIKE, BICYCLE,      PEDESTRIAN
       [800.0,  800.0,  800.0,    800.0,   800.0,      800.0,    800.0,         800.0]
 
+    validate_max_distance_m: 70.0 # [m]
+
     using_2d_validator: false
     enable_debugger: false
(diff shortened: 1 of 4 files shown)


Shin-kyoto (approved):
LGTM

- I did NOT check this PR by running autoware.

YoshiRi (approved):
LGTM

PR #9671, opened 2024-12-17
fix(autoware_objects_of_interest_marker_interface): fix bugprone-branch-clone

kobayu858:
## Description
This is a fix based on clang-tidy `bugprone-branch-clone` error.

```
/home/emb4/autoware/autoware/src/universe/autoware.universe/planning/autoware_objects_of_interest_marker_interface/src/objects_of_interest_marker_interface.cpp:85:5: error: switch has 2 consecutive identical branches [bugprone-branch-clone,-warnings-as-errors]
    case ColorName::GRAY:
    ^
/home/emb4/autoware/autoware/src/universe/autoware.universe/planning/autoware_objects_of_interest_marker_interface/src/objects_of_interest_marker_interface.cpp:88:38: note: last of these clones ends here
      return coloring::getGray(alpha);
                                     ^
```
## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/autoware_objects_of_interest_marker_interface/src/objects_of_interest_marker_interface.cpp
+++ b/planning/autoware_objects_of_interest_marker_interface/src/objects_of_interest_marker_interface.cpp
@@ -83,7 +83,6 @@ ColorRGBA ObjectsOfInterestMarkerInterface::getColor(
     case ColorName::RED:
       return coloring::getRed(alpha);
     case ColorName::GRAY:
-      return coloring::getGray(alpha);
     default:
       return coloring::getGray(alpha);
   }
--- a/planning/autoware_objects_of_interest_marker_interface/src/objects_of_interest_marker_interface.cpp
+++ b/planning/autoware_objects_of_interest_marker_interface/src/objects_of_interest_marker_interface.cpp
@@ -82,7 +82,8 @@ ColorRGBA ObjectsOfInterestMarkerInterface::getColor(
       return coloring::getAmber(alpha);
     case ColorName::RED:
       return coloring::getRed(alpha);
-    case ColorName::GRAY:
+    case ColorName::GRAY:  // NOLINT
+      return coloring::getGray(alpha);
     default:
       return coloring::getGray(alpha);
   }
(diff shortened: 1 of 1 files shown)


soblin (commented):
@kobayu858 I want to keep this skipped with a ignore lint comment, since the meaning of `GRAY` gets unclear

kobayu858:
@soblin 
Thanks for the confirmation. I have made the change to suppress it in the comments.

kobayu858:
@soblin 

soblin (approved):

planning/autoware_objects_of_interest_marker_interface/src/objects_of_interest_marker_interface.cpp
```diff
@@ -82,7 +82,7 @@ ColorRGBA ObjectsOfInterestMarkerInterface::getColor(
       return coloring::getAmber(alpha);
     case ColorName::RED:
       return coloring::getRed(alpha);
-    case ColorName::GRAY:
+    case ColorName::GRAY:  // NOLINT
```
xmfcx, line 85: Instead of marking it as "NOLINT" you can move this `GRAY` case to the top of the switch chain. And it shouldn't complain.

If you read here: https://clang.llvm.org/extra/clang-tidy/checks/bugprone/branch-clone.html

It explains it in detail.

Specifically this case:

> ```cpp
> switch (ch) {
> case 'a':
> case 'A':
>   return 10;
> case 'b':
> case 'B':
>   return 11;
> default:
>   return 10;
> }
> ```
> Here the check does not warn for the repeated return 10;, which is good if we want to preserve that 'a' is before 'b' and default: is the last branch.

xmfcx (changes requested):

planning/autoware_objects_of_interest_marker_interface/src/objects_of_interest_marker_interface.cpp
kobayu858: Thank you for the advice.
I have confirmed that the error no longer occurs and have fixed it.

xmfcx (approved):

PR #9674, opened 2024-12-17
feat(planning_evaluator): add lateral trajectory displacement metrics

Kazunori-Nakajima:
## Description

The lateral displacement between the current and previous trajectory was added to the metrics.
Capable of detecting large lateral deviations from self-position.

### Procedure
1. Calculate the perpendicular of the previous trajectory and self-position.
2. Calculate the perpendicular of the current trajectory and self-position.
3. Calculate the lateral trajectory displacement from 1 and 2.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

Output of `/planning/planning_evaluator/metrics` when changing Ego position from upper lane to lower lane

image

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


kyoichi-sugahara:
@Kazunori-Nakajima 
Depending on the measurement purpose, if you need to calculate the distance between a trajectory and a pose, using `calcLateralOffset` would seem more appropriate.

kyoichi-sugahara:
@Kazunori-Nakajima 
As you are comparing the trajectory from the previous cycle with the reference_trajectory in the current cycle, could you confirm if this processing is implemented as intended?

kosuke55:
> Depending on the measurement purpose, if you need to calculate the distance between a trajectory and a pose, using calcLateralOffset would seem more appropriate.

calculating lateral distance between **pose** and **pose** but  `calcLateralOffsete` is more approriate right.

evaluator/autoware_planning_evaluator/src/metrics/deviation_metrics.cpp
```diff
@@ -45,6 +45,28 @@ Accumulator<double> calcLateralDeviation(const Trajectory & ref, const Trajector
...
+    return stat;
+  }
+
+  const auto prev_idx =
+    autoware::motion_utils::findNearestIndex(prev_traj.points, ego_pose.position);
+  const auto prev_p = prev_traj.points[prev_idx];
+  const auto ref_idx = autoware::motion_utils::findNearestIndex(ref_traj.points, ego_pose.position);
+  const auto ref_p = ref_traj.points[ref_idx];
+
+  const double nearest_dist = autoware::universe_utils::calcDistance2d(prev_p, ref_p);
```
kosuke55, line 63: better to use `calcLateralDeviation` .

if using `calcDistance2d` and interval of points of two trajectories is diffrent, the distance also become large.

evaluator/autoware_planning_evaluator/config/planning_evaluator.param.yaml
```diff
@@ -13,6 +13,7 @@
       - jerk
       - lateral_deviation
       - yaw_deviation
+      - nearest_pose_deviation
```
kosuke55, line 16:
```suggestion
      - nearest_pose_lateral_deviation
```

evaluator/autoware_planning_evaluator/src/metrics_calculator.cpp
```diff
@@ -47,6 +47,9 @@ std::optional<Accumulator<double>> MetricsCalculator::calculate(
       return metrics::calcLateralDeviation(reference_trajectory_, traj);
     case Metric::yaw_deviation:
       return metrics::calcYawDeviation(reference_trajectory_, traj);
+    case Metric::nearest_pose_deviation:
+      return metrics::calcNearestPoseDeviation(
+        previous_trajectory_, reference_trajectory_, ego_pose_);
```
kosuke55, line 52: previous_trajectory_ and reference_trajectory_ seems not same topic?

kosuke55:
```
      <remap from="~/input/trajectory" to="$(var input/trajectory)"/>   # previous_trajectory_
      <remap from="~/input/reference_trajectory" to="$(var input/reference_trajectory)"/>  # reference_trajectory_
```

kosuke55: should compare same `~/input/trajectory` topic in dirrent cycle

evaluator/autoware_planning_evaluator/src/metrics/deviation_metrics.cpp
kosuke55: `calcLateralOffset` is also ok

evaluator/autoware_planning_evaluator/config/planning_evaluator.param.yaml
Kazunori-Nakajima: After consulting with them, we came up with the variable/function name `lateral_tajectory_displacement`!

evaluator/autoware_planning_evaluator/src/metrics_calculator.cpp
Kazunori-Nakajima: Changed to look at the same topic using `previous_trajectory` and `trajectory`!

evaluator/autoware_planning_evaluator/src/metrics/deviation_metrics.cpp
Kazunori-Nakajima: Yes sir!
update to use `calcLateralOffset`

kyoichi-sugahara (approved):
LGTM! Thanks.
nits: you can improve the doxygen documentation by adding information about the sign convention for lateral displacement 
e.g. positive when the input trajectory is more to the left compared to the previous trajectory, negative when more to the right

kosuke55 (approved):

kosuke55:
oh sorry I merged but

> nits: you can improve the doxygen documentation by adding information about the sign convention for lateral displacement
e.g. positive when the input trajectory is more to the left compared to the previous trajectory, negative when more to the right

this comment is important. better to use abs value

evaluator/autoware_planning_evaluator/src/metrics/deviation_metrics.cpp
```diff
@@ -45,6 +45,24 @@ Accumulator<double> calcLateralDeviation(const Trajectory & ref, const Trajector
...
+
+  if (prev.points.empty() || traj.points.empty()) {
+    return stat;
+  }
+
+  const auto prev_lateral_deviation =
+    autoware::motion_utils::calcLateralOffset(prev.points, ego_pose.position);
+  const auto traj_lateral_deviation =
+    autoware::motion_utils::calcLateralOffset(traj.points, ego_pose.position);
+  const auto lateral_trajectory_displacement = traj_lateral_deviation - prev_lateral_deviation;
```
kosuke55, line 61: - use std::value
- we want calculate lateral distance between nearest point in current and previous

PR #9685, opened 2024-12-17
feat(pid_longitudinal_controller): add virtual wall for dry steering and emergency

takayuki5168:
## Description

This PR added a virtual wall when the pid_longitudinal_controller is in an emergency as follows.

image
image

## How was this PR tested?

psim worked.
## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/control/autoware_pid_longitudinal_controller/src/pid_longitudinal_controller.cpp
+++ b/control/autoware_pid_longitudinal_controller/src/pid_longitudinal_controller.cpp
@@ -563,8 +563,8 @@ PidLongitudinalController::ControlData PidLongitudinalController::getControlData
 
   // distance to stopline
   control_data.stop_dist = longitudinal_utils::calcStopDistance(
-    control_data.interpolated_traj.points.at(control_data.nearest_idx).pose,
-    control_data.interpolated_traj, m_ego_nearest_dist_threshold, m_ego_nearest_yaw_threshold);
+    current_pose, control_data.interpolated_traj, m_ego_nearest_dist_threshold,
+    m_ego_nearest_yaw_threshold);
 
   // pitch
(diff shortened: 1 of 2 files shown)


control/autoware_pid_longitudinal_controller/src/pid_longitudinal_controller.cpp
```diff
@@ -562,8 +563,8 @@ PidLongitudinalController::ControlData PidLongitudinalController::getControlData
 
   // distance to stopline
   control_data.stop_dist = longitudinal_utils::calcStopDistance(
-    control_data.interpolated_traj.points.at(control_data.nearest_idx).pose,
-    control_data.interpolated_traj, m_ego_nearest_dist_threshold, m_ego_nearest_yaw_threshold);
+    current_pose, control_data.interpolated_traj, m_ego_nearest_dist_threshold,
```
takayuki5168, line 566: This change using `current_pose` is necessary to calculate the negative distance when the ego exceeds the stop point so that the pid_longitudinal_controller will be in an emergency.

Is this okay for you considering the following PR where you replaced `current_pose` with `control_data.interpolated_traj.points.at(control_data.nearest_idx).pose`?
https://github.com/autowarefoundation/autoware.universe/pull/4712

takayuki5168: @brkay54

yuki-takagi-66 (approved):

control/autoware_pid_longitudinal_controller/src/pid_longitudinal_controller.cpp
takayuki5168: @brkay54 
I will merge now, but let me know anytime if the change does not look good to you.

PR #9686, opened 2024-12-18
fix: enable to copy all information in pickup based pointcloud downsampler

YoshiRi:
## Description

Previously pickup-based downsampler, which is used in the default downsampler for the perception input, only copies x,y,z attribute and do not copy other one like intensity. This leads to cause some perception issue.

This PR fix this problem by copying all attributes.

## How was this PR tested?

Just build and tested with sample bag file(WIP).

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


badai-nguyen (approved):
LGTM

PR #9709, opened 2024-12-22
ci: introduce build-test-tidy-pr

xmfcx:
## Description

- build-and-test-differential
- clang-tidy-differential

These workflows have quite a history in autoware. They have been modified in many ways to make sure everything builds and is tested correctly. Sometimes they didn't fit into their machines, sometimes they were bound to labels to function.

Lately I had some complaints about the way they were structured:

### `build-and-test-differential` was extracted as a composite action

- It was hard to see the logs.
  - This is a general GitHub issue: [Better log separation for composite actions #21276](https://github.com/orgs/community/discussions/21276)
- It didn't extract enough steps.

To overcome these issues, I've extracted it as a [reusable workflow](https://docs.github.com/en/actions/sharing-automations/reusing-workflows#creating-a-reusable-workflow).

### Checking `cuda` packages required an additional label

This meant, pr-labeler added the label with these rules:
```yaml
"tag:require-cuda-build-and-test":
  - perception/**/*
  - sensing/**/*
```

This required adding the label and re-running the other steps. And it was also possible to add this label by hand and confuse the system.

Also `prepare-build-and-test-differential` job and its downstream logic was too complicated and hard to maintain.

### `clang-tidy-differential` and `clang-tidy-differential-cuda` jobs were in `build-and-test-differential` job ☹️

And `clang-tidy-differential` was duplicated. Also its `cache-key-element:` was not set up correctly, it was missing cache and had to rebuild from scratch.

### There was no clear success condition

We used to have 4 actual jobs:

- build-and-test-differential
- clang-tidy-differential
- clang-tidy-differential-cuda

But also if cuda jobs were skipped due to not having cuda related files changed, we also wanted them count as success.

## Solution

- Remove the `tag:require-cuda-build-and-test` and instead use `tj-actions/changed-files@v45` in a simple way.
- Introduce new workflow `build-test-tidy-pr.yaml` which contains these jobs:
  - `require-label` 🚪 Nothing below won't run without the label, ❌ Will fail if label is missing.
  - `check-if-cuda-job-is-needed` ↪️ Checks if PR modified perception/sensing folders
  - `build-and-test-differential`
  - `build-and-test-differential-cuda`
  - `clang-tidy-differential`
  - `clang-tidy-differential-cuda`

This top level workflow doesn't contain any implementation details. It is for orchestrating the other workflows by reusing those workflows.

The implementation is simple and the logic is easily readable.

- `build-and-test-differential` and `clang-tidy-differential` are made into reusable workflows.

This way we can see all the steps easily and can reuse more steps in them.

I hope you like how it was simplified and cleaned.

### Other changes:

- 🌲 Switched to `ubuntu-24.04` public runners. They provide 10GB more empty space, we can run all on them.
- 🎁 Now uses newly created [require-label](https://github.com/autowarefoundation/autoware-github-actions/pull/331) to fail if label doesn't exists.
- 🎅 Required label is renamed from `tag:run-build-and-test-differential` to `run:build-and-test-differential`

### Links from `main`, before this PR:

- [build-and-test-differential.yaml](https://github.com/autowarefoundation/autoware.universe/blob/b107ac7bc94b3d88e2873ecbc57aa1cde3375eb5/.github/workflows/build-and-test-differential.yaml)
- [build-and-test-differential/action.yaml](https://github.com/autowarefoundation/autoware.universe/blob/main/.github/actions/build-and-test-differential/action.yaml)

### A very old version of this workflow

- [build-and-test-differential.yaml](https://github.com/autowarefoundation/autoware.universe/blob/0e07aa8860475da2181fb4e742f69e64b480b5fa/.github/workflows/build-and-test-differential.yaml) -> Things were simple back then 🙂

## How was this PR tested?

I've ran these tests from scratch sequentially after finishing up my changes, they are all tested on the latest version before review. 🪟

### No relevant files changed to build-and-test

- https://github.com/autowarefoundation/autoware.universe/actions/runs/12458672141 ✅

image

#### 🟢 Good change

- https://github.com/autowarefoundation/autoware.universe/actions/runs/12458524111 ✅

image

#### 🔴 Bad change

- https://github.com/autowarefoundation/autoware.universe/actions/runs/12458573989 ✅

image

#### 🟢 Good change

- https://github.com/autowarefoundation/autoware.universe/actions/runs/12458598376 ✅

image

#### 🔴 Bad change

-  https://github.com/autowarefoundation/autoware.universe/actions/runs/12458466068 ✅

image


xmfcx:
Testing after changes...

## CUDA package changed

### 🟢 Good change

- https://github.com/autowarefoundation/autoware.universe/actions/runs/12479850859 ✅

image

### 🔴 Bad change

- https://github.com/autowarefoundation/autoware.universe/actions/runs/12480040911

mitsudome-r (approved):

xmfcx:
For the record, these are the new required checks with this PR:

image

PR #9733, opened 2024-12-23
feat(autoware_object_merger, autoware_tracking_object_merger): enable anonymized node names to be configurable

technolojin:
## Description
Since the node names are anonymized, the node name changes on every launch.
The random node name introduce complexity on the node diagnostics.

This PR provides the node name variable `node_name` to be configured externally.
If the node name is not set, the name will be anonymized by default.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/launch/tier4_perception_launch/launch/object_recognition/detection/merger/camera_lidar_merger.launch.xml
+++ b/launch/tier4_perception_launch/launch/object_recognition/detection/merger/camera_lidar_merger.launch.xml
@@ -120,6 +120,7 @@
   <!-- 1st merger: camera_lidar_fusion + lidar cluster-->
   <group>
     <include file="$(find-pkg-share autoware_object_merger)/launch/object_association_merger.launch.xml">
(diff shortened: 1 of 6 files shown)


launch/tier4_perception_launch/launch/object_recognition/detection/merger/camera_lidar_merger.launch.xml
```diff
@@ -120,6 +120,7 @@
   <!-- 1st merger: camera_lidar_fusion + lidar cluster-->
   <group>
     <include file="$(find-pkg-share autoware_object_merger)/launch/object_association_merger.launch.xml">
+      <arg name="node_name" value="object_association_merger_alpha"/>
```
YoshiRi, line 123: [imho]
I prefer "1st, 2nd" instead of "alpha, gamma".

YoshiRi (approved):
LGTM

PR #9737, opened 2024-12-23
fix(lane_change): add metrics to valid paths visualization

zulfaqar-azmi-t4:
## Description

Adds path metrics along valid paths' visualization.

https://github.com/user-attachments/assets/61d42394-80dd-4cf4-9c1c-3b77364e185f

Visualization can be turn on and off separately.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/utils/markers.cpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/utils/markers.cpp
@@ -79,12 +79,11 @@ MarkerArray showAllValidLaneChangePath(
     if (lc_path.path.points.empty()) {
       continue;
     }
-    std::ostringstream ss;
-    ss << ns.c_str() << "[" << idx << "]";
+    std::string ns_with_idx = ns + "[" + std::to_string(idx) + "]";
     const auto & color = colors.at(idx);
     const auto & points = lc_path.path.points;
     auto marker = createDefaultMarker(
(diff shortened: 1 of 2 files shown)


maxime-clem (approved):

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/utils/markers.cpp
```diff
@@ -38,34 +38,75 @@ using autoware::universe_utils::createDefaultMarker;
...
     return MarkerArray{};
   }
 
   MarkerArray marker_array;
-  int32_t id{0};
   const auto current_time{rclcpp::Clock{RCL_ROS_TIME}.now()};
 
   const auto colors = colors::colors_list();
-  const auto loop_size = std::min(lanes.size(), colors.size());
+  const auto loop_size = std::min(lane_change_paths.size(), colors.size());
```
mkquda, line 52: so if the number of paths is more than the colors you will leave out the rest of the paths?

mkquda (approved):

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/utils/markers.cpp
zulfaqar-azmi-t4: We have to keep the color of the paths consistent, so at we to use the colors from color list. Probably have to reuse color, but I will do it in other PR.

PR #9746, opened 2024-12-23
feat(behavior_velocity_planner): remove virtual traffic light dependency from behavior_velocity_planner and behavior_velocity_planner_common

takayuki5168:
## Description

Currently, the virtual traffic light's variables in the `behavior_velocity_planner/behavior_velocity_planner_common` are very specific to the `behavior_velocity_virtual_traffic_light_module` and do not need to be shared in other modules.
Therefore, this PR moved the implementation related to the virtual traffic light from the `behavior_velocity_planner/behavior_velocity_planner_common` to the `behavior_velocity_virtual_traffic_light_module`.
## How was this PR tested?

The unit test passed.
The psim worked as expected.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


kosuke55 (approved):
LGTM!

PR #9766, opened 2024-12-24
fix(perception): fix perception docs

vividf:
## Description
There are several pages in perception docs that are broken on the website. 
This PR fixes the issue including the wrong path and schema error.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

PR #9769, opened 2024-12-24
ci(build-test-tidy): introduce success condition jobs

xmfcx:
## Description

I had initially made this change a part of the previous PR:
- https://github.com/autowarefoundation/autoware.universe/pull/9709

But we removed it to allow a simpler structure.

But I think we cannot avoid it due to how we add things to `Status checks that are required` from rulesets menu.

## Problem

Issue is, you add the required "jobs" to this list.

image

You cannot add workflows here, only jobs. And if jobs are within jobs as in reusable workflows, you need to include both names.

For example, for build-and-test-differential job, here is the list of allowed jobs currently available in this repo:
image

### Add `build-and-test-differential` to list

❌ We end up with:
image
image

In a workflow. If we check more closely:
image

It actually needs `build-and-test-differential / build-and-test-differential` to complete.

### Add `build-and-test-differential / build-and-test-differential` to list

image

It is solved ✅

This was to demonstrate how it works. But now I will share the problem part.

#### Success

image

It works if it is not skipped ✅

#### Skipping

❌ We end up with:
image
image

This happened because it was skipped. And sub-workflow did not execute at all.

image

#### Success

❌ We end up with:
image
image

Because the full job name is not in the list.

#### Skipping

image

It works if it is skipped ✅

## Solution

This means, we cannot rely on skipping counted as success with reusable workflows.

So I'm bringing back these jobs that will run at all times and notify you about the success condition.

## Related links

Tested PRs:

- https://github.com/autowarefoundation/autoware.universe/pull/9766 (non-cuda change)
- https://github.com/autowarefoundation/autoware.universe/pull/9768 (cuda change)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/.github/actions/evaluate-job-result/action.yaml
+++ b/.github/actions/evaluate-job-result/action.yaml
@@ -30,7 +30,7 @@ runs:
 
         for RESULT in "${EXPECTED[@]}"; do
           if [[ "$JOB_RESULT" == "$RESULT" ]]; then
-            echo "- ${{ inputs.job_name }}: Success ✅" >> "$GITHUB_STEP_SUMMARY"
+            echo "- **${{ inputs.job_name }}:** "$JOB_RESULT" ✅" >> "$GITHUB_STEP_SUMMARY"
             echo "failed=false" >> "$GITHUB_OUTPUT"
             exit 0
           fi
@@ -38,6 +38,6 @@ runs:
 
         # If no expected result matched
         echo "::error::${{ inputs.job_name }} failed. ❌"
-        "- ${{ inputs.job_name }}: Failed ❌" >> "$GITHUB_STEP_SUMMARY"
+        echo "- **${{ inputs.job_name }}:** failed ❌" >> "$GITHUB_STEP_SUMMARY"
         echo "failed=true" >> "$GITHUB_OUTPUT"
       shell: bash
--- a/.github/actions/evaluate-job-result/action.yaml
+++ b/.github/actions/evaluate-job-result/action.yaml
@@ -16,7 +16,6 @@ inputs:
 outputs:
   failed:
     description: Indicates if the job failed
-    type: boolean
     value: ${{ steps.evaluate.outputs.failed }}
 
 runs:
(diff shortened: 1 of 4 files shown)


xmfcx:
I improve the readability of the results.

xmfcx:
## Tests

### CUDA Package Modification

#### 🟢 Good Change

#### 🔴 Bad Change

- https://github.com/autowarefoundation/autoware.universe/actions/runs/12483758294 ✅

image

image

### Normal Package Modification

mitsudome-r:
@xmfcx 
Thanks for the fix. I was curious so I took a little look at this. 

It seems like another workaround is to just pass the if condition to the reusable workflow as an argument and skip the job inside the reusable workflow. This way, you will have the results from the sub-workflow even if it is skipped, and the required status check won't stop at "expected". (I've created a test PR on my fork here https://github.com/mitsudome-r/autoware.universe/pull/10 if you want to take a look)

This keeps the workflow code simpler, and developers would have similar user experience to check failed conditions as before. However, it is a bit of hacky way to solve it so I'm okay with your solution as well.

Maybe we can wait for developers reaction, and if they feel uncomfortable with the current fix, we can consider using this approach.

image
image
image

xmfcx:
@mitsudome-r In my opinion, counting skipping as a success condition hides intention and confuses even more. It's better to make things explicit than implicit. For example, cuda workflows are marked as required but they can be skipped and it just passes, this, to me is confusing. But if this build-test-pr is marked as required, we explicitly report our success condition to developers.

Right now, if the workflow fails, they can click on the failed workflow and they will be pointed to where to look.

I've checked your solution and yes, it would solve this problem too. And with much less code.
Still, I feel it hides the success condition by allowing skipped items counted as suceeded implicitly.

If you really want to keep it the old way with the small fix, I understand too, we can apply your solution. 👍

xmfcx:
By the way, the reason we have to deal with these complicated success conditions is that we started running CUDA jobs conditionally. Due to this, clang-tidy also branched into 2. It is the most efficient way to use limited compute resources. I think it's unavoidable to have this complicated structure as long as we aim to support cuda and non-CUDA builds in a large codebase that has large parts that explicitly doesn't require CUDA.

mitsudome-r:
@xmfcx Thanks for your comment.
I agree that GitHub's specification of treating skipped workflow as success condition is confusing.

I just created the PR to the upstream just in case if we want to apply the modification: https://github.com/autowarefoundation/autoware.universe/pull/9771 

Maybe we can come back and discuss this in the next Software WG call.

PR #9771, opened 2024-12-24
ci: pass run condition to reusable workflow

mitsudome-r:
## Description

This PR resolves the issue raised in https://github.com/autowarefoundation/autoware.universe/pull/9769 by alternative method.

This PR:
* reverts https://github.com/autowarefoundation/autoware.universe/pull/9769
* creates a new argument `run-condition` to the reusable `build-and-test-diffirential.yaml` workflow

For the details, please refer to the comment [here](https://github.com/autowarefoundation/autoware.universe/pull/9769#issuecomment-2561453062).

## Related links

- https://github.com/autowarefoundation/autoware.universe/pull/9769

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


xmfcx:
Thanks, you need to do the same for the clang tidy one as well.

.github/workflows/build-test-tidy-pr.yaml
```diff
@@ -50,88 +50,30 @@ jobs:
...
-    if: ${{ needs.check-if-cuda-job-is-needed.outputs.cuda_job_is_needed == 'false' }}
     uses: ./.github/workflows/clang-tidy-differential.yaml
     with:
       container: ghcr.io/autowarefoundation/autoware:universe-devel
+      run-condition: ${{ needs.check-if-cuda-job-is-needed.outputs.cuda_job_is_needed == 'false' }}
 
   clang-tidy-differential-cuda:
+    if: ${{ always() }} # always run to provide report for status check
     needs:
       - build-and-test-differential-cuda
```
mitsudome-r, line 74: I had to explicitly set "always()" condition to make sure that "clang-tidy-differential-cuda" job even when build-and-test-differential-cuda is not ran.

xmfcx:
## Tests

### Normal Package Modification

#### 🟢 Good Change

- https://github.com/autowarefoundation/autoware.universe/actions/runs/12546108605 ✅

image

#### 🔴 Bad Change

- https://github.com/autowarefoundation/autoware.universe/actions/runs/12546179494 ✅

image

### CUDA Package Modification

#### 🟢 Good Change

- https://github.com/autowarefoundation/autoware.universe/actions/runs/12546412960 ❌
  - This has skipped both the `clang-tidy-differential` jobs.

image

#### 🔴 Bad Change



### No Package Modification

#### With Label 🟢



#### Without Label 🔴

xmfcx:
## Tests - 02

### Normal Package Modification

#### 🟢 Good Change



#### 🔴 Bad Change



### CUDA Package Modification

#### 🟢 Good Change



#### 🔴 Bad Change



### No Package Modification

#### With Label 🟢



#### Without Label 🔴

xmfcx:
## Tests - 03

### Normal Package Modification

#### 🟢 Good Change

- https://github.com/autowarefoundation/autoware.universe/actions/runs/12547876847

#### 🔴 Bad Change

- 

### CUDA Package Modification

#### 🟢 Good Change

- 

#### 🔴 Bad Change

- 

### No Package Modification

#### With Label 🟢

- 

#### Without Label 🔴

-

xmfcx:
## Tests - 04

<details>
  <summary>🖱️Click here to expand🔛</summary>

### Normal Package Modification

#### 🟢 Good Change

- 


#### 🔴 Bad Change

- 

### CUDA Package Modification

#### 🟢 Good Change


#### 🔴 Bad Change

- 

### No Package Modification

#### With Label 🟢

- 

#### Without Label 🔴

- https://github.com/autowarefoundation/autoware.universe/actions/runs/12548549365 ✅

image

</details>

mitsudome-r:
@xmfcx Thanks, the modification looks fine.

FYI, in Test 3, you said
> Without Label 🔴
> It skipped build-and-test-differential sub-jobs. Needs to be fixed.
> Fixed in: https://github.com/autowarefoundation/autoware.universe/commit/49394beb047101ea2774263bdf6f88002406972f
 you need additional fix 

but I don't think it's a necessary change.
If we skip the sub-jubs, the status check in a PR will tell a developer that it is expecting for the results from the build-and-test-differential jobs, which is fine. The current change will report that the build-and-test-differential & build-and-test-differential have succeeded (skipped).
Either should be okay as long as we set the require-label as status checks.

xmfcx:
> but I don't think it's a necessary change.
If we skip the sub-jubs, the status check in a PR will tell a developer that it is expecting for the results from the build-and-test-differential jobs, which is fine.

We should not send the users mixed signals about skipping and failing. If I didn't make this change, It would show the build and test differential job as expected but it would never come up. Maybe the user would wait for a runner ro pick it up. It's not clear why it is yellow like this:

Screenshot_20241231_044131_Chrome

And would be the only occasion where we would have a job skipped but at the same time we would show it as yellow "expected".

With the fix, it show as skipped because it will be skipped.

> The current change will report that the build-and-test-differential & build-and-test-differential have succeeded (skipped).

The CI treats skipping as success, yes. But this doesn't mean that the job was already skipped.
Showing it as yellow "expected" is more confusing. And require-label job will be the only failing check in this case, making it easier to spot for developers.

mitsudome-r:
okay, let's keep it consistent.

PR #9783, opened 2024-12-25
feat(lane_change): add info text to virtual wall

mkquda:
## Description
add text to lane change stop wall to provide information for reason of stopping.

When there are valid candidate paths but none are safe, stop wall will be placed at terminal start with the following text `lane_change_direction (no safe path)` as shown below.
Screenshot from 2024-12-25 11-02-41

When there are no valid candidate paths , stop wall will be placed at terminal start with the following text `lane_change_direction (no valid path)` as shown below.
Screenshot from 2024-12-25 16-07-37

When there is stopped front object along current lane and there is no blocking object on target lane, stop wall will be placed at a distance from front object with following text `lane_change_direction (maintain distance to front object)` as shown below.
Screenshot from 2024-12-25 11-02-07

## Related links

**Parent Issue:**

- Link

## How was this PR tested?
- PSIM.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


zulfaqar-azmi-t4 (approved):
LGTM

PR #9787, opened 2024-12-25
feat(ekf_localizer): check whether the initialpose has been set

YamatoAndo:
## Description
Since there are cases where the `ekf_localizer` does not function properly without the initialpose being set, I’ve added a check to ensure whether initialpose is configured.

If it is not set:

- The estimation process will not execute, and no estimation results will be output.

- /diagnostics will transition to the WARN state.

Additionally, if the `ekf_localizer` transitions to an inactive state (e.g., due to re-running the `pose_initializer`), the initialpose must be set again.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

- I tested with logging_simulator same as before.
- I changed the topic name for the input initialpose of `ekf_localizer` (`/initialpose3d`) to create a situation where the initialpose is not set, and confirmed that the behavior matches the description above.

| Screenshot from 2024-12-25 18-09-00 |
| :-: |

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


SakodaShintaro (approved):
LGTM

PR #9789, opened 2024-12-25
refactor(autoware_image_projection_based_fusion): organize 2d-detection related members

technolojin:
## Description

Avoid to use index-based member management.
1. A new struct `Det2dStatus` replaces vector of variables related to roi/camera.
2. Index-based accesses were replaced by object-oriented way as much as possible.
3. Templates of input/output types are revised
4. Comments on each process were added

This PR do not contain any logical change.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

[TIER IV INTERNAL 1](https://evaluation.tier4.jp/evaluation/reports/c756425f-246b-5546-a0f3-638c24d66cdc?project_id=prd_jt)
[TIER IV INTERNAL 2](https://evaluation.tier4.jp/evaluation/reports/939590ee-452a-5299-a9cf-bb704d2172bd?project_id=prd_jt)
[TIER IV INTERNAL 3](https://evaluation.tier4.jp/evaluation/reports/5d336e9f-b7c9-5977-863f-3dd091af1a07?project_id=prd_jt)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


kminoda:
@technolojin Are you sure this is "refactor" and not "feat" nor "fix"? At least the logic around mutex is changed, so I think it is risky to name it "refactor".

perception/autoware_image_projection_based_fusion/include/autoware/image_projection_based_fusion/fusion_node.hpp
```diff
@@ -72,81 +92,93 @@ class FusionNode : public rclcpp::Node
...
+    std::lock_guard<std::mutex> lock_det2d_flags(mutex_det2d_flags_);
+    det2d.is_fused = true;
+  }
+  void clearAllDet2dFlags()
+  {
+    std::lock_guard<std::mutex> lock_det2d_flags(mutex_det2d_flags_);
+    for (auto & det2d : det2d_list_) {
+      det2d.is_fused = false;
+    }
+  }
```
kminoda, line 130: We should move this functions to `private` instead of `protected` since it is not intended to be used by the inherited classes. (BTW I think it is applicable to other functions such as `cameraInfoCallback`, `subCallback`, etc... :melting_face: )

perception/autoware_image_projection_based_fusion/include/autoware/image_projection_based_fusion/fusion_node.hpp
```diff
@@ -52,14 +52,34 @@ using autoware_perception_msgs::msg::DetectedObject;
...
+  bool project_to_unrectified_image = false;
+  bool approximate_camera_projection = false;
+  // process flags
+  bool is_fused = false;
+  // timing
+  double input_offset_ms = 0.0;
+  // cache
+  std::map<int64_t, typename Msg2D::ConstSharedPtr> cached_det2d_msgs;
+  std::unique_ptr<std::mutex> mtx_ptr;
+};
```
kminoda, line 80: IMHO it's better to explicitly separate mutable and immutable variables. For instance `id`, `project_to_unrectified_image`, or `approximate_camera_projection` is immutable while `cached_det2d_msgs` or `is_fused` is mutable, right? 

(Below is just a random implementation proposal, but maybe there is a better way to achieve this)
```suggestion
template <class Msg2D>
struct Det2dManager
{
  // camera index
  const std::size_t id;
  // camera projection
  const std::unique_ptr<CameraProjection> camera_projector_ptr;
  const bool project_to_unrectified_image;
  const bool approximate_camera_projection;
  // process flags
  bool is_fused = false;
  // timing
  const double input_offset_ms;
  // cache
  std::map<int64_t, typename Msg2D::ConstSharedPtr> cached_det2d_msgs;
  std::unique_ptr<std::mutex> mtx_ptr;
  
  Det2dManager(std::size_t id_, ...): id(id_), ... {}
};
```

perception/autoware_image_projection_based_fusion/include/autoware/image_projection_based_fusion/fusion_node.hpp
```diff
@@ -52,14 +52,34 @@ using autoware_perception_msgs::msg::DetectedObject;
...
+using ClusterMsgType = tier4_perception_msgs::msg::DetectedObjectsWithFeature;
+using ClusterObjType = tier4_perception_msgs::msg::DetectedObjectWithFeature;
 using tier4_perception_msgs::msg::DetectedObjectWithFeature;
 using PointCloud = pcl::PointCloud<pcl::PointXYZ>;
 using autoware::image_projection_based_fusion::CameraProjection;
 using autoware_perception_msgs::msg::ObjectClassification;
 
-template <class TargetMsg3D, class ObjType, class Msg2D>
+template <class Msg2D>
+struct Det2dManager
```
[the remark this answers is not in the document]
technolojin: I partially got your idea.
I think "status" is more suitable naming. Thank you.

fixed https://github.com/autowarefoundation/autoware.universe/pull/9789/commits/3c73ba99bd4fc0ddfc4e43fea9c8d659cdd1c0e0

perception/autoware_image_projection_based_fusion/include/autoware/image_projection_based_fusion/fusion_node.hpp
```diff
@@ -52,14 +52,34 @@ using autoware_perception_msgs::msg::DetectedObject;
...
+  std::unique_ptr<CameraProjection> camera_projector_ptr;
+  bool project_to_unrectified_image = false;
+  bool approximate_camera_projection = false;
+  // process flags
+  bool is_fused = false;
+  // timing
+  double input_offset_ms = 0.0;
+  // cache
+  std::map<int64_t, typename Msg2D::ConstSharedPtr> cached_det2d_msgs;
+  std::unique_ptr<std::mutex> mtx_ptr;
```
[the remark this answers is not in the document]
technolojin: I removed mutex related changes https://github.com/autowarefoundation/autoware.universe/pull/9789/commits/a01f84fb9ecd0d08f141679dc11573617a55ec8b

mutex related topic may will be treated in different pr

technolojin:
> @technolojin Are you sure this is "refactor" and not "feat" nor "fix"? At least the logic around mutex is changed, so I think it is risky to name it "refactor".

Since the mutex related changes are removed, this PR became a pure refactoring.
https://github.com/autowarefoundation/autoware.universe/pull/9789/commits/a01f84fb9ecd0d08f141679dc11573617a55ec8b

perception/autoware_image_projection_based_fusion/include/autoware/image_projection_based_fusion/fusion_node.hpp
technolojin: reviewed member and method access range

fixed https://github.com/autowarefoundation/autoware.universe/pull/9789/commits/18bf3d062946aa3df91d9e6e86dfe9d7407abc79

perception/autoware_image_projection_based_fusion/include/autoware/image_projection_based_fusion/fusion_node.hpp
technolojin: It is a `struct` and replaces vector of variables.
In terms of readability, the current `struct` approach is appropriate.
I expect developers use it properly.

perception/autoware_image_projection_based_fusion/src/roi_pointcloud_fusion/node.cpp
```diff
@@ -202,15 +193,18 @@ void RoiPointCloudFusionNode::fuseOnSingleImage(
...
-bool RoiPointCloudFusionNode::out_of_scope(__attribute__((unused))
-                                           const DetectedObjectWithFeature & obj)
+void RoiPointCloudFusionNode::publish(const PointCloudMsgType & output_msg)
 {
-  return false;
+  if (point_pub_ptr_->get_subscription_count() < 1) {
+    return;
+  }
+  point_pub_ptr_->publish(output_msg);
 }
```
vividf, line 206: I noticed that this node publishes the required cluster in the postprocess function, while the publish function is used for the unused point cloud. In my opinion, it might be more intuitive and maintainable to consolidate the output logic by publishing the required output directly in the publish function.

For reference, the roi_detected_fusion node handles a similar case by not implementing a separate postprocess function and instead integrating the postprocess logic into its publish function. Adopting a similar approach here could simplify the code and align with that standard.

Additionally, a similar issue arises in the pointpainting node, where the painted point cloud (which is not the required output) is published. While this can be useful for debugging purposes, I would suggest moving the postprocessing logic for the required outputs into the publish function as well. This would ensure clarity in the node's outputs while retaining the option to publish additional data under a debug mode if necessary.

perception/autoware_image_projection_based_fusion/src/roi_cluster_fusion/node.cpp
```diff
@@ -291,6 +270,36 @@ double RoiClusterFusionNode::cal_iou_by_mode(
...
+    if (
+      is_roi_label_known ||
+      feature_object.object.existence_probability >= min_roi_existence_prob_) {
+      known_objects.feature_objects.push_back(feature_object);
+    }
+  }
+  output_cluster_msg.feature_objects = known_objects.feature_objects;
+}
+
+void RoiClusterFusionNode::publish(const ClusterMsgType & output_msg)
```
vividf, line 295:
```suggestion
void RoiClusterFusionNode::publish(const ClusterMsgType & output_cluster_msg)
```

vividf: I think it would be better to have same naming as postprocess

vividf (changes requested):
Thank you for your thoughtful refactoring! The updated code is significantly more readable and well-structured compared to the previous version.
In my opinion, there are likely areas within roi_pointcloud_fusion and pointpainting_fusion that could also be improved further.

perception/autoware_image_projection_based_fusion/src/roi_cluster_fusion/node.cpp
```diff
@@ -291,6 +270,36 @@ double RoiClusterFusionNode::cal_iou_by_mode(
...
+{
+  if (pub_ptr_->get_subscription_count() < 1) {
+    return;
+  }
+  pub_ptr_->publish(output_msg);
```
vividf, line 300:
```suggestion
  pub_ptr_->publish(output_cluster_msg);
```

perception/autoware_image_projection_based_fusion/include/autoware/image_projection_based_fusion/fusion_node.hpp
kminoda: Thanks!

perception/autoware_image_projection_based_fusion/include/autoware/image_projection_based_fusion/fusion_node.hpp
```diff
@@ -71,82 +90,84 @@ class FusionNode : public rclcpp::Node
...
+  typename Msg3D::SharedPtr cached_det3d_msg_ptr_;
   std::mutex mutex_cached_msgs_;
 
-  // output publisher
-  typename rclcpp::Publisher<TargetMsg3D>::SharedPtr pub_ptr_;
+protected:
+  // callback for main subscription
+  void subCallback(const typename Msg3D::ConstSharedPtr input_msg);
+  // callback for roi subscription
+  void roiCallback(const typename Msg2D::ConstSharedPtr input_roi_msg, const std::size_t roi_i);
```
kminoda, line 136: Any reason why you removed "virtual"?

perception/autoware_image_projection_based_fusion/src/pointpainting_fusion/node.cpp
```diff
@@ -156,11 +155,16 @@ PointPaintingFusionNode::PointPaintingFusionNode(const rclcpp::NodeOptions & opt
   const auto min_area_matrix = this->declare_parameter<std::vector<double>>("min_area_matrix");
   const auto max_area_matrix = this->declare_parameter<std::vector<double>>("max_area_matrix");
 
+  // subscriber
   std::function<void(const sensor_msgs::msg::PointCloud2::ConstSharedPtr msg)> sub_callback =
     std::bind(&PointPaintingFusionNode::subCallback, this, std::placeholders::_1);
   sub_ = this->create_subscription<sensor_msgs::msg::PointCloud2>(
     "~/input/pointcloud", rclcpp::SensorDataQoS().keep_last(3), sub_callback);
```
kminoda, line 162:
```suggestion
  std::function<void(const PointCloudMsgType::ConstSharedPtr msg)> sub_callback =
    std::bind(&PointPaintingFusionNode::subCallback, this, std::placeholders::_1);
  sub_ = this->create_subscription<PointCloudMsgType>(
    "~/input/pointcloud", rclcpp::SensorDataQoS().keep_last(3), sub_callback);
```

kminoda (commented):
Would you consider making splitting this PR into several ones? Current changes are bit too many to assess.

My suggestion is:
- PR 1: Simply replacing the type names with "using PointCloudMsgType", "using RoiMsgType"
- PR 2: Introduce `Det2dStatus`

perception/autoware_image_projection_based_fusion/include/autoware/image_projection_based_fusion/fusion_node.hpp
technolojin: Because those functions are defined only in `FusionNode` and no other classes override those.

technolojin:
> Would you consider making splitting this PR into several ones? Current changes are bit too many to assess.
> 
> My suggestion is:
> 
>     * PR 1: Simply replacing the type names with "using PointCloudMsgType", "using RoiMsgType"
> 
>     * PR 2: Introduce `Det2dStatus`

I understand your point. 
Unfortunately, those two changes are overlapped, therefore, it will take a big effort from now.
How much do you think separating PR is important? Is it worth 5h for example?

kminoda:
@technolojin I see, I thought it is worth splitting if it only takes 20-30min, but if it takes 5hr then it's not necessary. Let me review this PR as it is.

perception/autoware_image_projection_based_fusion/include/autoware/image_projection_based_fusion/roi_cluster_fusion/node.hpp
```diff
@@ -24,22 +24,21 @@ namespace autoware::image_projection_based_fusion
...
+  void publish(const ClusterMsgType & output_msg) override;
 
   void fuseOnSingleImage(
-    const DetectedObjectsWithFeature & input_cluster_msg, const std::size_t image_id,
-    const DetectedObjectsWithFeature & input_roi_msg,
-    DetectedObjectsWithFeature & output_cluster_msg) override;
+    const ClusterMsgType & input_cluster_msg, const Det2dStatus<RoiMsgType> & det2d,
+    const RoiMsgType & input_roi_msg, ClusterMsgType & output_cluster_msg) override;
 
+private:
```
kminoda, line 41: Why adding `private` to only the below and not for other functions like `preprocess` or `postprocess` defined above? I think it is safer to assume that this node is not inherited and thus all the functions should be `private`.

(Not sure if we should do this type of improvement in this PR though...)

perception/autoware_image_projection_based_fusion/src/fusion_node.cpp
```diff
@@ -45,8 +46,8 @@ namespace autoware::image_projection_based_fusion
 {
 using autoware::universe_utils::ScopedTimeTrack;
 
-template <class TargetMsg3D, class ObjType, class Msg2D>
-FusionNode<TargetMsg3D, ObjType, Msg2D>::FusionNode(
+template <class Msg3D, class Msg2D, class ExportObj>
+FusionNode<Msg3D, Msg2D, ExportObj>::FusionNode(
   const std::string & node_name, const rclcpp::NodeOptions & options)
 : Node(node_name, options), tf_buffer_(this->get_clock()), tf_listener_(tf_buffer_)
```
kminoda, line 52: (Not related to here)
I thought that we can even make `rois_number_` a local variable instead of a member variable. Would you mind changing this as well?

perception/autoware_image_projection_based_fusion/src/fusion_node.cpp
```diff
@@ -67,98 +68,114 @@ FusionNode<TargetMsg3D, ObjType, Msg2D>::FusionNode(
...
 
+  // 2d detection status initialization
+  det2d_list_.resize(rois_number_);
+  for (std::size_t roi_i = 0; roi_i < rois_number_; ++roi_i) {
+    det2d_list_.at(roi_i).id = roi_i;
+    det2d_list_.at(roi_i).project_to_unrectified_image =
+      point_project_to_unrectified_image.at(roi_i);
+    det2d_list_.at(roi_i).approximate_camera_projection = approx_camera_projection.at(roi_i);
+    det2d_list_.at(roi_i).input_offset_ms = input_offset_ms.at(roi_i);
+  }
```
kminoda, line 156: I was always annoyed with this too lengthy constructor and bunch of variables existing in one single scope, but thanks to your refactoring, now it is clear that a large portion of it is for generating `det2d_list_` object.

That being said, how about creating a dedicated function for generating `det2d_list_`, and isolate all the code that relates to this to that function? This way, we can limit the scope of some variables such as `approx_camera_projection` or `input_offset_ms` and thus increase the readability of this whole code.

e.g. follows (may include some bugs)

```cpp
std::vector<Det2dStatus<Msg2D>> build_det2d_status(rclcpp::Node * node, int rois_number)
{
  // camera offset settings
  std::vector<double> input_offset_ms = node->declare_parameter<std::vector<double>>("input_offset_ms");
  if (!input_offset_ms.empty() && rois_number > input_offset_ms.size()) {
    throw std::runtime_error("The number of offsets does not match the number of topics.");
  }

  // camera projection settings
  std::vector<bool> point_project_to_unrectified_image =
    node->declare_parameter<std::vector<bool>>("point_project_to_unrectified_image");
  if (rois_number > point_project_to_unrectified_image.size()) {
    throw std::runtime_error(
      "The number of point_project_to_unrectified_image does not match the number of rois topics.");
  }
  std::vector<bool> approx_camera_projection =
    node->declare_parameter<std::vector<bool>>("approximate_camera_projection");
  if (rois_number != approx_camera_projection.size()) {
    const std::size_t current_size = approx_camera_projection.size();
    RCLCPP_WARN(
      node->get_logger(),
      "The number of elements in approximate_camera_projection should be the same as in "
      "rois_number. "
      "It has %zu elements.",
      current_size);
    if (current_size < rois_number) {
      approx_camera_projection.resize(rois_number);
      for (std::size_t i = current_size; i < rois_number; i++) {
        approx_camera_projection.at(i) = true;
      }
    }
  }

  // 2d detection status initialization
  det2d_list.resize(rois_number_);
  for (std::size_t roi_i = 0; roi_i < rois_number; ++roi_i) {
    det2d_list.at(roi_i).id = roi_i;
    det2d_list.at(roi_i).project_to_unrectified_image =
      point_project_to_unrectified_image.at(roi_i);
    det2d_list.at(roi_i).approximate_camera_projection = approx_camera_projection.at(roi_i);
    det2d_list.at(roi_i).input_offset_ms = input_offset_ms.at(roi_i);
  }
  return det2d_list;
}
```

perception/autoware_image_projection_based_fusion/include/autoware/image_projection_based_fusion/fusion_node.hpp
```diff
@@ -71,82 +90,84 @@ class FusionNode : public rclcpp::Node
...
+  // parameters for out_of_scope filter
   float filter_scope_min_x_;
   float filter_scope_max_x_;
   float filter_scope_min_y_;
   float filter_scope_max_y_;
   float filter_scope_min_z_;
   float filter_scope_max_z_;
 
+  // output publisher
+  typename rclcpp::Publisher<ExportObj>::SharedPtr pub_ptr_;
```
kminoda, line 166: Why do we need to have `pub_ptr_` in this class when the `publish()` function no longer calls this object?

And if we do this, I think we should also rename the `pub_ptr_` variable in each inherited class, maybe to a more self-contained name.

kminoda (commented):
Sorry for 五月雨式 comments :bow:

perception/autoware_image_projection_based_fusion/src/roi_pointcloud_fusion/node.cpp
technolojin: I realized that too. 
Debug message and the main message are mixed and published in the `postprocess` in some nodes.

technolojin: - pointpainting fusion
  - detected object(main) - published in `postprocess`
  - pointcloud - published in `publish`

- roi cluster fusion
  - cluster(main) - published in `publish`

- roi detected object fusion
  - detected object(main) - published in `publish`

- roi pointcloud fusion
  - cluster(main) - published in `postprocess`
  - pointcloud - published in `publish`

- segment pointcloud fusion
  - pointcloud(main) - published in `publish`

perception/autoware_image_projection_based_fusion/include/autoware/image_projection_based_fusion/fusion_node.hpp
kminoda: Ah true, overlooked that.

Then I think we need to move these two into private as well.

perception/autoware_image_projection_based_fusion/src/roi_pointcloud_fusion/node.cpp
technolojin: postprocess and publish were refactored in https://github.com/autowarefoundation/autoware.universe/pull/9789/commits/3b5f47b3a2da8688bc7e28be8f545e45f9ed9c95 and https://github.com/autowarefoundation/autoware.universe/pull/9789/commits/56f304112cffadc022c869784a85d2aef485492f

perception/autoware_image_projection_based_fusion/src/roi_cluster_fusion/node.cpp
technolojin: please see https://github.com/autowarefoundation/autoware.universe/pull/9789/commits/3b5f47b3a2da8688bc7e28be8f545e45f9ed9c95 and https://github.com/autowarefoundation/autoware.universe/pull/9789/commits/56f304112cffadc022c869784a85d2aef485492f

perception/autoware_image_projection_based_fusion/src/pointpainting_fusion/node.cpp
technolojin: fixed https://github.com/autowarefoundation/autoware.universe/pull/9789/commits/f34b017efd429f81036e1578454a03bc71d5bb4b

perception/autoware_image_projection_based_fusion/include/autoware/image_projection_based_fusion/fusion_node.hpp
technolojin: `subCallback` is called in the inherited class.

https://github.com/autowarefoundation/autoware.universe/blob/e2accc57e9d0b1328641bd88a48e446c74879af3/perception/autoware_image_projection_based_fusion/src/pointpainting_fusion/node.cpp#L159-L162

perception/autoware_image_projection_based_fusion/include/autoware/image_projection_based_fusion/roi_cluster_fusion/node.hpp
technolojin: fixed https://github.com/autowarefoundation/autoware.universe/pull/9789/commits/647f6d68d9c0ae8504106ff607957cdbb079fb5f

perception/autoware_image_projection_based_fusion/include/autoware/image_projection_based_fusion/fusion_node.hpp
technolojin: `publish()` and `pub_ptr_` are used before and after. You may need to consider all variant of the node implementations. 

By refactoring further, it became a common method.
Please see https://github.com/autowarefoundation/autoware.universe/commit/3b5f47b3a2da8688bc7e28be8f545e45f9ed9c95 and https://github.com/autowarefoundation/autoware.universe/commit/56f304112cffadc022c869784a85d2aef485492f

perception/autoware_image_projection_based_fusion/src/fusion_node.cpp
technolojin: nice catch. I will do that

technolojin: fixed in https://github.com/autowarefoundation/autoware.universe/pull/9789/commits/e95a54b1afa3b6e8260fe225941ca8faba35c93e

perception/autoware_image_projection_based_fusion/src/fusion_node.cpp
technolojin: implemented https://github.com/autowarefoundation/autoware.universe/pull/9789/commits/e95a54b1afa3b6e8260fe225941ca8faba35c93e

perception/autoware_image_projection_based_fusion/src/pointpainting_fusion/node.cpp
```diff
@@ -416,21 +418,17 @@ void PointPaintingFusionNode::postprocess(sensor_msgs::msg::PointCloud2 & painte
...
 
   detection_class_remapper_.mapClasses(output_msg);
 
-  if (objects_sub_count > 0) {
-    obj_pub_ptr_->publish(output_msg);
+  // publish debug message: painted pointcloud
+  if (point_pub_ptr_->get_subscription_count() < 1) {
+    point_pub_ptr_->publish(painted_pointcloud_msg);
   }
 }
```
vividf, line 431: It seems that the publish() function for publishing the detected objects is missed

perception/autoware_image_projection_based_fusion/src/pointpainting_fusion/node.cpp
```diff
@@ -416,21 +418,17 @@ void PointPaintingFusionNode::postprocess(sensor_msgs::msg::PointCloud2 & painte
...
+  // prepare output message
   output_msg.objects = iou_bev_nms_.apply(raw_objects);
 
   detection_class_remapper_.mapClasses(output_msg);
 
-  if (objects_sub_count > 0) {
-    obj_pub_ptr_->publish(output_msg);
+  // publish debug message: painted pointcloud
+  if (point_pub_ptr_->get_subscription_count() < 1) {
+    point_pub_ptr_->publish(painted_pointcloud_msg);
```
vividf, line 428: Could you change the variable name from point_pub_ptr_ to painted_pointcloud_pub_ptr_

perception/autoware_image_projection_based_fusion/src/roi_pointcloud_fusion/node.cpp
```diff
@@ -202,14 +167,40 @@ void RoiPointCloudFusionNode::fuseOnSingleImage(
...
+
+  output_fused_objects_.clear();
+
+  // publish debug cluster
+  if (cluster_debug_pub_->get_subscription_count() > 0) {
+    PointCloudMsgType debug_cluster_msg;
+    autoware::euclidean_cluster::convertObjectMsg2SensorMsg(output_msg, debug_cluster_msg);
+    cluster_debug_pub_->publish(debug_cluster_msg);
+  }
+  if (point_pub_ptr_->get_subscription_count() > 0) {
```
vividf, line 191: Is there a specific reason for publishing the point cloud? Since the published point cloud appears to be identical to the input point cloud, it might not add additional value. In my opinion, the point_pub_ptr could potentially be removed to simplify the code and reduce unnecessary overhead.

vividf (changes requested):

perception/autoware_image_projection_based_fusion/src/roi_pointcloud_fusion/node.cpp
technolojin: I do not think it is required.

It may be reviewed in separated PR. This PR is only for refactoring.

perception/autoware_image_projection_based_fusion/src/fusion_node.cpp
```diff
@@ -180,77 +162,127 @@ FusionNode<TargetMsg3D, ObjType, Msg2D>::FusionNode(
...
 
-template <class TargetMsg3D, class Obj, class Msg2D>
-void FusionNode<TargetMsg3D, Obj, Msg2D>::subCallback(
-  const typename TargetMsg3D::ConstSharedPtr input_msg)
+template <class Msg3D, class Msg2D, class ExportObj>
+void FusionNode<Msg3D, Msg2D, ExportObj>::exportProcess()
+{
+  timer_->cancel();
+
+  ExportObj output_msg;
```
vividf, line 240: I am a bit worried that for segmentation_pointcloud_fusion node, this will copy the pointcloud again.

perception/autoware_image_projection_based_fusion/src/roi_pointcloud_fusion/node.cpp
vividf: In the old version, the pointcloud wasn’t being published, and it didn’t seem necessary. 
I’m curious about the reasoning behind adding this functionality in the refactor.

perception/autoware_image_projection_based_fusion/src/fusion_node.cpp
technolojin: I know that the postprocess is coping the pointcloud. 
Can you let me know where the copy is happening again? Any suggestion to avoid copy is welcomed.

perception/autoware_image_projection_based_fusion/src/roi_pointcloud_fusion/node.cpp
technolojin: The original `publish` method of the parent class DO publish the pointcloud.

perception/autoware_image_projection_based_fusion/src/fusion_node.cpp
vividf: Apologies for my misunderstanding! This part looks good to me now!

perception/autoware_image_projection_based_fusion/src/pointpainting_fusion/node.cpp
technolojin: I do not think that change is really needed. 
Just want to be aligned with other node `RoiPointCloudFusionNode`.

Publishing pointcloud is not needed anyway, then should be turn-off by default. It will be done in a separated PR.

vividf: Got it!
Thanks!

perception/autoware_image_projection_based_fusion/src/roi_pointcloud_fusion/node.cpp
technolojin: the pointcloud publishing condition was wrong.
fixed https://github.com/autowarefoundation/autoware.universe/pull/9789/commits/cc51e6b4cb462cb5b689bef6724b35bf56ec7034

vividf: You’re absolutely right! I got confused by the different publisher names. Thanks for pointing that out!
Let's remove that in another PR

vividf (approved):
LGTM for the refactor!

perception/autoware_image_projection_based_fusion/include/autoware/image_projection_based_fusion/fusion_node.hpp
kminoda: Sorry I overlooked it. Thanks :pray:

kminoda (approved):
LGTM

PR #9795, opened 2024-12-26
fix(autoware_lidar_centerpoint): fixed rounding errors that caused illegal memory access

knzo25:
## Description

We encountered rounding errors resulting from floating precision arithmetic in which points nearing the pointcloud ranges would  generate invalid voxel ids. Since it is not a heavy operation, we decided to clip the values which makes it a little more resilient.

```
[component_container_mt-1] ERROR | voxel_idx=480, voxel_idy=189, voxel_index=-291 grid_x_size=480 grid_y_size=480 point.x=76.799995 point.y=-16.133558 point.z=5.022423 min_x_range=-76.800003 min_y_range=-76.800003 min_z_range=-4.000000 max_x_range=76.800003 max_y_range=76.800003 max_z_range=6.000000
```

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

Run the logging simulator 20 times.
Before this PR, and for a particular rosbag, we encountered the issue with a reproduction rate of about 1 in 4.

Evaluator:
[TIER IV INTERNAL LINK](https://evaluation.tier4.jp/evaluation/reports/0cd7fe57-fdb3-5731-a7f7-7666e86393bb?project_id=prd_jt)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

PR #9796, opened 2024-12-26
fix(autoware_lidar_transfusion): fixed rounding errors that caused illegal memory access

knzo25:
## Description

In centerpoint https://github.com/autowarefoundation/autoware.universe/pull/9795
we encountered rounding errors resulting from floating precision arithmetic in which points nearing the pointcloud ranges would  generate invalid voxel ids. Since it is not a heavy operation, we decided to clip the values which makes it a little more resilient.

```
[component_container_mt-1] ERROR | voxel_idx=480, voxel_idy=189, voxel_index=-291 grid_x_size=480 grid_y_size=480 point.x=76.799995 point.y=-16.133558 point.z=5.022423 min_x_range=-76.800003 min_y_range=-76.800003 min_z_range=-4.000000 max_x_range=76.800003 max_y_range=76.800003 max_z_range=6.000000
```

Transfusion uses a different way of generating voxel ids, which makes the problem a bug, but not one that causes cuda errors, so it probably was silent until now :scream: 

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


amadeuszsz (approved):
LGTM

PR #9797, opened 2024-12-26
Replace debug msgs

vish0012:
## Description
changes the tier4_debug_msgs to autoware_internal_debug_msgs
## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


takayuki5168:
@vish0012 
Could you please split the PR into each component (e.g. planning, control, common, etc) so that the PR will not be large?

isamu-takagi (approved):

vish0012:
> @vish0012 Thank you for your contribution! Could you please split the PR into each component (e.g. planning, control, common, etc) so that the PR will not be large?

>@takayuki5168 thank you for your valuable comment and guidance , I will do the needful as per your suggestion. Sorry for the inconvenience.

PR #9813, opened 2024-12-27
feat(pointpainting_fusion): enable cloud display on image

vividf:
## Description
The PointPainting fusion node disables plotting the point cloud on the image because it uses `push_back` within an OpenMP parallel loop.

> #if 0
      // Parallelizing loop don't support push_back
      if (debugger_) {
        debug_image_points.push_back(projected_point);
      }
#endif

This PR introduces a method to store the projected points in local vectors and merge them after the for-loop. These updates make it easier for developers and users to debug.

Before:
Screenshot from 2024-12-27 15-33-27

After
Screenshot from 2024-12-27 13-57-35

## Related links

**Parent Issue:**

- Link

## How was this PR tested?
Test with taxi rosbag. 

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


ktro2828 (approved):
Code looks good to me.

PR #9854, opened 2025-01-08
feat: add build-and-test-packages-above-differential.yaml and packages_above.repos

takayuki5168:
## Description

Since the CI on the autoware.universe's PR does not check the build of above dependencies, the autoware_tools which depends on the autoware.universe is sometimes broken by the autoware.universe's change.
To avoid this, this PR introduced `packages_above.repos` and added a new workflow with `packages_above.repos`.

## How was this PR tested?

With https://github.com/autowarefoundation/autoware-github-actions/pull/333, the test's results are as expected as follows
- test case to fail
    - https://github.com/autowarefoundation/autoware.universe/pull/10251
- test case to pass
    - https://github.com/autowarefoundation/autoware.universe/pull/10253

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


mitsudome-r:
This PR runs colcon build with packages_above.repos as a separate job from build-test-tidy-pr so there will be duplicate work to build all the packages within build_depends.repos and autoware.universe again in this job. Maybe we can consider using cache from build-test-tidy-pr job for more efficiency in the future, but I'm okay to define it as a independent job for easier maintenance.

mitsudome-r:
It might also require modification of build-test-tidy-pr and call this workflow as additional job to trigger the workflow.

mitsudome-r:
1. Reflect recent modification to [build-and-test-differential.yaml](https://github.com/autowarefoundation/autoware.universe/blob/7756eb7ec9c043f921aec1f819710c9046517b33/.github/workflows/build-and-test-differential.yaml#L99)  (such as prepare build_depend.repos) to your workflow.

2. Please update [built-and-test-pr.yaml](https://github.com/autowarefoundation/autoware.universe/blob/main/.github/workflows/build-test-tidy-pr.yaml) to add a new job "build-and-test-differentail-above" with the following content to run your workflow against PRs:

```
    if: ${{ always() }}
    needs:
      - require-label
    uses: ./.github/workflows/build-and-test-differential-above.yaml
    with:
      run-condition: ${{ needs.require-label.outputs.result == 'true' }}
    secrets:
      codecov-token: ${{ secrets.CODECOV_TOKEN }}
```

3. Please run test with https://github.com/autowarefoundation/autoware-github-actions/pull/333.
You can specify the "feat/colcon-build-above-dependency" branch instead of v1 in this line: https://github.com/tier4/autoware.universe/blob/0f795c2b25d2cddcb02eb51c9592f29cb9639556/.github/workflows/build-and-test-packages-above-differential.yaml#L97

takayuki5168:
@mitsudome-r 
I fixed for the comment 1-3. Additionally, since the change of the clang-tidy action is not necessary, I reverted it as follows.
https://github.com/autowarefoundation/autoware-github-actions/pull/333/commits/549dff419b851da89a06de9f758561335d4fe29b

And I tested the actions and the results are as expected. I described them in detail in the PR's description.

Now I'm ready to merge the PRs!

mitsudome-r (approved):
LGTM

mitsudome-r (approved):

PR #9862, opened 2025-01-08
fix(image_projection_based_fusion): remove mutex

technolojin:
## Description
Since [callback group](https://docs.ros.org/en/foxy/How-To-Guides/Using-callback-groups.html) is not set in the `autoware_image_projection_based_fusion`, all the callbacks are in a same group.
Therefore, the mutex is not necessary.

This PR do not contain any logic change.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?
Tested in a local environment.

lidar_detection_model: pointpainting

tested
* pointpainting fusion
* roi cluster fusion
* roi detected-object fusion
* roi pointcloud fusion

not tested
* segment pointcloud fusion

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_image_projection_based_fusion/include/autoware/image_projection_based_fusion/fusion_node.hpp
+++ b/perception/autoware_image_projection_based_fusion/include/autoware/image_projection_based_fusion/fusion_node.hpp
@@ -40,7 +40,6 @@
 
 #include <map>
 #include <memory>
-#include <mutex>
 #include <set>
 #include <string>
 #include <utility>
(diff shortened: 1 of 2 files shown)


kminoda:
Thanks!!!!!!!!
Would you elaborate more on "how this was tested"? (e.g. checked that the pointpainting node outputs some objects in RViz with some random rosbag)

technolojin:
> Thanks!!!!!!!! Would you elaborate more on "how this was tested"? (e.g. checked that the pointpainting node outputs some objects in RViz with some random rosbag)

I added more about the test done.

kminoda (approved):
LGTM
Thank you for your work as always :pray:

PR #9950, opened 2025-01-17
fix(map_based_prediction): fix unintentional accumulation of lanelets

a-maumau:
## Description
In `PredictorVru::setLaneletMap`, it will always accumulate the crosswalk or walkways lanelets when map gets updated.
It may work trouble-free with the current implementation in the current Autoware architecture, where the map is provided only once.
But in multiple map case (loading map like mapA -> mapB -> mapA -> ... for moving between 2 regions), it may potentially cause some extra latency since it will 
 linear scan the lanelets for each targeted objects: https://github.com/autowarefoundation/autoware_lanelet2_extension/blob/c8ff315d21fc2e62329a7272e3d49edfa88fc57e/autoware_lanelet2_extension/lib/query.cpp#L885

## Related links

**Parent Issue:**

- Link

## How was this PR tested?
Checked the logging simulator using the sample ROS bag.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


technolojin (approved):
LGTM

PR #924, opened 2022-05-18
feat(pointcloud_preprocessor): add polygon remover

beginningfan:
Signed-off-by: liangliang fan <beginning.fan@autocore.ai>

## Description

Add polygon remover to filter pointcloud self reflections.

## Related links

resolve issue #703 

## Tests performed
run this command 
```
ros2 launch pointcloud_preprocessor polygon_remover.launch.py
```
and this is result

polygon_remover4

## Notes for reviewers

Code is rewritten based on https://gitlab.com/autowarefoundation/autoware.auto/AutowareAuto/-/tree/master/src/perception/filters/polygon_remover

## Pre-review checklist for the PR author

The PR author **must** check the checkboxes below when creating the PR.

- [x] I've confirmed the [contribution guidelines].
- [x] The PR follows the [pull request guidelines].

## In-review checklist for the PR reviewers

The PR reviewers **must** check the checkboxes below before approval.

- [x] The PR follows the [pull request guidelines].
- [x] The PR has been properly tested.
- [x] The PR has been reviewed by the code owners.

## Post-review checklist for the PR author

The PR author **must** check the checkboxes below before merging.

- [ ] There are no open discussions or they are tracked via tickets.
- [ ] The PR is ready for merge.

After all checkboxes are checked, anyone who has write access can merge the PR.

[contribution guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/
[pull request guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/pull-request-guidelines/

PR #9955, opened 2025-01-17
chore(autoware_pointcloud_preprocessor): add missing dependencies

wentasah:
## Description

GMP is required in CMakeLists.txt. MPFR is required by GMP and is not installed by default with GMP (Debian GMP package has MPFR dependency only as a suggestion: https://packages.debian.org/sid/libgmp-dev).

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

By building the package with the Nix package manager.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/sensing/autoware_pointcloud_preprocessor/package.xml
+++ b/sensing/autoware_pointcloud_preprocessor/package.xml
@@ -35,9 +35,11 @@
   <depend>cv_bridge</depend>
   <depend>diagnostic_updater</depend>
   <depend>image_transport</depend>
+  <depend>libgmp</depend>
   <depend>libopencv-dev</depend>
   <depend>libpcl-all-dev</depend>
   <depend>message_filters</depend>
+  <depend>mpfr</depend>
   <depend>nav_msgs</depend>
   <depend>pcl_conversions</depend>
   <depend>pcl_msgs</depend>


knzo25:
@beginningfan 
It seems you added this dependency (https://github.com/autowarefoundation/autoware.universe/pull/924).
Do you have an opinion about MPFR?
Unless it is actually used, it would just increase the size of the container, etc

wentasah:
I forgot to mention that the build fails without any of these added packages. If MPFR is not available, configure fails with:
```
-- Found GMP: /nix/store/ga9ymzd1mc47z7j5l945j7hld5n4g0a6-gmp-with-cxx-6.3.0/lib/libgmp.so
CMake Error at /nix/store/4k2h9qsps52n1v56zq03yi67dwmxa81i-cmake-3.30.5/share/cmake-3.30/Modules/FindPackageHandleStandardArgs.cmake:233 (message):
  Could NOT find MPFR (missing: MPFR_LIBRARIES MPFR_INCLUDE_DIR)
Call Stack (most recent call first):
  /nix/store/4k2h9qsps52n1v56zq03yi67dwmxa81i-cmake-3.30.5/share/cmake-3.30/Modules/FindPackageHandleStandardArgs.cmake:603 (_FPHSA_FAILURE_MESSAGE)
  /nix/store/94a41wjpnccb9v826cghdpjjj63n7j23-cgal-5.6.2/lib/cmake/CGAL/FindMPFR.cmake:54 (find_package_handle_standard_args)
  /nix/store/94a41wjpnccb9v826cghdpjjj63n7j23-cgal-5.6.2/lib/cmake/CGAL/CGAL_SetupGMP.cmake:25 (find_package)
  /nix/store/94a41wjpnccb9v826cghdpjjj63n7j23-cgal-5.6.2/lib/cmake/CGAL/CGAL_SetupCGALDependencies.cmake:37 (include)
  /nix/store/94a41wjpnccb9v826cghdpjjj63n7j23-cgal-5.6.2/lib/cmake/CGAL/CGALConfig.cmake:168 (include)
  CMakeLists.txt:15 (find_package)
```

beginningfan:
I'm sorry for replying so late. 
The Ubuntu system will automatically install the two packages you mentioned when installing CGAL. How is your CGAL installed? Normally, there should not be a problem of not being able to find GMP
image

wentasah:
Thank you for the reply. You're right that `libcgal-dev` explicitly depends on `gmp` and `mpfr` libraries. I'm actually packaging Autoware for the Nix package manager and there, CGAL doesn't provide these libraries to dependent packages. It might be a bug in Nix CGAL expression (how we call packages in Nix), but I think that if [`CMakeLists.txt` of a package mentions a library](https://github.com/autowarefoundation/autoware.universe/blob/3c84a57939d4f5a7e6ebe6421386989cd8fd3dc2/sensing/autoware_pointcloud_preprocessor/CMakeLists.txt#L202), it is useful if that library is explicitly mentioned in `package.xml`.

So feel free to close this PR, I can carry this change in my Nix packaging. On the other hand, if you're interested in compatibility with non-Ubuntu systems, this PR will unlikely do any harm to Ubuntu users if merged.

PR #9999, opened 2025-01-22
feat(rviz): add new plugin to visualize planning factor

satoshi-ota:
## Description

New rviz plugin to visualize following data which are in planing factor.

- Red wall shows the ego stop position.
- Yellow wall shows the ego slow down position.
- Circle marker shows the safety factors.

## Related links

**Parent Issue:**

None.

## How was this PR tested?

1. launch Psim.
2. Add `PlanningFactorRvizPlugin`.
3. Select topic name.

I confirmed that this plugin could show the marker in Psim.
Screenshot from 2025-01-22 09-38-08
Screenshot from 2025-01-22 09-37-29
image

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


shmpwk (approved):
LGTM

Owen-Liuyuxuan:
@satoshi-ota 
As far as our internal tests go, this plugin is the **only** plugin being used that requires vehicle parameters. This makes it impossible for Rviz to launch without global parameters.

When users launch Rviz with the usual  `rviz2 -d {rviz_config}`, it will only produce an error with `[rviz2]: Could not load display config: Statically typed parameter 'wheel_radius' must be initialized.` which could be very difficult to debug and solve.

Could you further consider updates to avoid the changes in system dependencies?

[internal PR](https://github.com/tier4/autoware_ecu_system_setup/pull/1445)

PR #10062, opened 2025-02-03
feat(diagnostic_graph_aggregator): remove edit feature

isamu-takagi:
## Description

I created edit feature to reuse diagnostic graph settings, but found that dummy_diag_publisher would be a better choice.

## Related links

**Parent Issue:**

- [TIER IV internal link](https://star4.slack.com/archives/C070FQMREVB/p1738567920430969)

## How was this PR tested?

Remove related file and check that there are no links to those files.

## Notes for reviewers

For users who already use it, I will leave the implementation as a deprecated feature for now.

## Interface changes

None.

## Effects on system behavior

None.


sasakisasaki (commented):
Perhaps [this header part](https://github.com/isamu-takagi/autoware.universe/blob/feat/remove-diag-graph-edit/system/autoware_diagnostic_graph_aggregator/src/common/graph/names.hpp#L35-L40) and [config.cpp](https://github.com/isamu-takagi/autoware.universe/blob/feat/remove-diag-graph-edit/system/autoware_diagnostic_graph_aggregator/src/common/graph/config.cpp#L213-L219) also need to be fixed.

@isamu-takagi Could you do double check if this fix is needed or not? I think we can merge this PR if this point gets clear.

isamu-takagi:
@sasakisasaki 
For backwards compatibility, I will keep the implementation.

sasakisasaki (approved):
### Quick Comment

Sounds good.

### Reason for approval
I checked all the changed files and found some points I want to ask. I got answer for my question and I understand all the fixes are assigned as expected.

Now I believe this PR is ready for merge. Thanks!

Issue #10069, opened 2025-02-04
Create a CUDA dependency meta-package


xmfcx:
### Checklist

- [x] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [x] I've searched other issues and no duplicate issues were found.
- [x] I've agreed with the maintainers that I can plan this task.

### Description

**Continuation from:** 
- https://github.com/autowarefoundation/autoware.universe/issues/3135#issuecomment-2634342143

### Purpose

To simplify filtering out packages that depend on CUDA.

### Possible approaches

Create a meta package with a basic CMakeLists that checks for CUDA.

All packages that **directly** depend on CUDA should depend on this meta package.

**Package name:** `autoware_cuda_dependency_meta`

The README of this package should explain why the package exists and should contain possible usages.

**Example:**
- `colcon build --packages-ignore $(colcon list --packages-above autoware_tensorrt_common autoware_cuda_utils --names-only | tr '\n' ' ')`

### Useful graph

`colcon graph --packages-above autoware_cuda_utils autoware_tensorrt_common --dot | dot -Tsvg -Grankdir=LR -Gnodesep=0.2 -Granksep=0.3 -Gconcentrate=true -o graph.svg`

![Image](https://github.com/user-attachments/assets/09aea4a1-25b2-4fd9-9bcf-c3997400f4bd)

### Definition of done

- [ ] Create the meta-package
- [ ] Update the package.xml files of the packages that directly depend on CUDA.

esteve:
The buildfarm builds fine when excluding the `autoware_cuda_dependency_meta` https://github.com/autowarefoundation/autoware-deb-packages/pull/127

esteve:
Closing this now that https://github.com/autowarefoundation/autoware.universe/pull/10073 has been merged and that https://github.com/autowarefoundation/autoware-deb-packages/pull/127 builds fine.

PR #10073, opened 2025-02-05
refactor: add autoware_cuda_dependency_meta

esteve:
## Description

Added meta package that packages that depend on CUDA must depend on.

## Related links

**Parent Issue:**

- https://github.com/autowarefoundation/autoware.universe/issues/10069

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/sensing/livox/autoware_livox_tag_filter/package.xml
+++ b/sensing/livox/autoware_livox_tag_filter/package.xml
@@ -11,7 +11,6 @@
   <buildtool_depend>ament_cmake_auto</buildtool_depend>
   <buildtool_depend>autoware_cmake</buildtool_depend>
 
-  <depend>autoware_cuda_dependency_meta</depend>
   <depend>libpcl-all-dev</depend>
   <depend>pcl_conversions</depend>
   <depend>rclcpp</depend>
--- a/sensing/livox/autoware_livox_tag_filter/package.xml
+++ b/sensing/livox/autoware_livox_tag_filter/package.xml
@@ -11,6 +11,7 @@
   <buildtool_depend>ament_cmake_auto</buildtool_depend>
   <buildtool_depend>autoware_cmake</buildtool_depend>
 
+  <depend>autoware_cuda_dependency_meta</depend>
   <depend>libpcl-all-dev</depend>
   <depend>pcl_conversions</depend>
   <depend>rclcpp</depend>
--- a/system/autoware_diagnostic_graph_utils/package.xml
+++ b/system/autoware_diagnostic_graph_utils/package.xml
@@ -11,7 +11,6 @@
   <buildtool_depend>ament_cmake_auto</buildtool_depend>
   <buildtool_depend>autoware_cmake</buildtool_depend>
 
-  <depend>autoware_cuda_dependency_meta</depend>
   <depend>autoware_internal_debug_msgs</depend>
   <depend>diagnostic_msgs</depend>
   <depend>rclcpp</depend>
--- a/system/autoware_diagnostic_graph_utils/package.xml
+++ b/system/autoware_diagnostic_graph_utils/package.xml
@@ -11,6 +11,7 @@
   <buildtool_depend>ament_cmake_auto</buildtool_depend>
   <buildtool_depend>autoware_cmake</buildtool_depend>
 
+  <depend>autoware_cuda_dependency_meta</depend>
   <depend>autoware_internal_debug_msgs</depend>
   <depend>diagnostic_msgs</depend>
   <depend>rclcpp</depend>
--- a/common/autoware_cuda_dependency_meta/CMakeLists.txt
+++ b/common/autoware_cuda_dependency_meta/CMakeLists.txt
@@ -1,0 +1,6 @@
+cmake_minimum_required(VERSION 3.8)
+project(autoware_cuda_dependency_meta NONE)
+
+find_package(ament_cmake REQUIRED)
+
+ament_package()
--- a/common/autoware_cuda_dependency_meta/README.md
+++ b/common/autoware_cuda_dependency_meta/README.md
@@ -1,0 +1,5 @@
+# autoware_cuda_dependency_meta
+
+## Purpose
+
+This is a virtual package that packages on Autoware.universe that use CUDA must depend on.
--- a/common/autoware_cuda_dependency_meta/package.xml
+++ b/common/autoware_cuda_dependency_meta/package.xml
@@ -1,0 +1,17 @@
+<?xml version="1.0"?>
+<package format="3">
+  <name>autoware_cuda_dependency_meta</name>
+  <version>0.0.1</version>
+  <description>Virtual package that provides an empty cuda stub package</description>
+
+  <author email="esteve.fernandez@tier4.jp">Esteve Fernandez</author>
+  <maintainer email="esteve.fernandez@tier4.jp">Esteve Fernandez</maintainer>
+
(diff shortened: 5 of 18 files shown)


xmfcx:
Hi @esteve thanks for working on this task, is this ready for review? Are there any blockers?

esteve:
@xmfcx I'm testing this in our buildfarm to make sure packages that depend on `autoware_cuda_dependency_meta` can be filtered out easily.

esteve:
This PR is now ready for review, the buildfarm builds fine when excluding the autoware_cuda_dependency_meta https://github.com/autowarefoundation/autoware-deb-packages/pull/127

esteve:
@xmfcx this PR requires approvals from 15 people, would it be possible to merge it with only one or two? I worry it might take too long if we have to wait for 15. Thanks.

amadeuszsz (approved):
LGTM!
I'm not only sure if for new packages we should align package version in `package.xml` with latest tag. Nevertheless, new release will overwrite it.

sensing/livox/autoware_livox_tag_filter/package.xml
```diff
@@ -11,6 +11,7 @@
   <buildtool_depend>ament_cmake_auto</buildtool_depend>
   <buildtool_depend>autoware_cmake</buildtool_depend>
 
+  <depend>autoware_cuda_dependency_meta</depend>
```
knzo25, line 14: why does this package need the dependency? :thinking:

knzo25:
@esteve 
Can you add this packge as well?
https://github.com/autowarefoundation/autoware.universe/blob/main/perception/autoware_probabilistic_occupancy_grid_map/CMakeLists.txt

system/autoware_diagnostic_graph_utils/package.xml
```diff
@@ -11,6 +11,7 @@
   <buildtool_depend>ament_cmake_auto</buildtool_depend>
   <buildtool_depend>autoware_cmake</buildtool_depend>
 
+  <depend>autoware_cuda_dependency_meta</depend>
```
isamu-takagi, line 14: This package is not related to CUDA.

sensing/livox/autoware_livox_tag_filter/package.xml
esteve: Removd in 8f0f65de390118e5de3779f1c1073a5c8b69a31b

system/autoware_diagnostic_graph_utils/package.xml
esteve: Removed in 8f0f65de390118e5de3779f1c1073a5c8b69a31b

esteve:
@knzo25 if the package depends on CUDA in one way or another, yes, it must depend on `autoware_cuda_dependency_meta`, even if some of its components can be built without CUDA. Eventually we want to add `REQUIRED` to `find_package(CUDA)` for those packages that depend on CUDA to make the dependency clear and so that build fails if CUDA is not available.

xmfcx:
```
$ colcon list --packages-above autoware_tensorrt_common autoware_cuda_utils --names-only
autoware_bytetrack
autoware_cuda_utils
autoware_detected_object_validation
autoware_detection_by_tracker
autoware_image_projection_based_fusion
autoware_launch
autoware_lidar_apollo_instance_segmentation
autoware_lidar_centerpoint
autoware_lidar_transfusion
autoware_probabilistic_occupancy_grid_map
autoware_raindrop_cluster_filter
autoware_shape_estimation
autoware_tensorrt_classifier
autoware_tensorrt_common
autoware_tensorrt_yolox
autoware_traffic_light_classifier
autoware_traffic_light_fine_detector
tier4_perception_launch
tier4_simulator_launch
trt_batched_nms
```

Many packages are missing this dependency. Could you update them?

graph

---

> Or rather, a question:
> The [autoware_probabilistic_occupancy_grid_map](https://github.com/autowarefoundation/autoware.universe/blob/main/perception/autoware_probabilistic_occupancy_grid_map/CMakeLists.txt) contains, say, 2 different nodes that share some logic.
> One needs cuda and will not be built if there is no nvcc. The other one does not need cuda so will be built no matter what.
> Should we use the meta dependency in this case?

@knzo25 I'd suggest that we should not allow partial dependency for cuda for packages. A package should either depend on it or not.

And the building behavior set in CMakeLists.txt should reflect that.

If a node out of this package needs to be used in a non-cuda environment, it should be separated into another package.

But until that happens, we will treat the entire package as a cuda package.

knzo25:
@xmfcx 
I agree with that and for new nodes I am trying hard to separate implementations without duplicating code (what will happen in the concat vs cuda concat).
For the OGM we really needed to acelerate it so I cuda-fied it. Then we got complains that some people needed to run one version (the one that is not heavy) in environments without cuda so I modified the cmake to allow for that :cry:

knzo25 (approved):
LGTM

xmfcx (changes requested):
About 6 packages are missing the dependency.

- https://github.com/autowarefoundation/autoware.universe/pull/10073#issuecomment-2699918678

I'm slightly conflicted about `<exec_depend>`ency. I think we can exclude them?

esteve:
@xmfcx 

> Many packages are missing this dependency. Could you update them?

I don't see why those packages should depend on `autoware_cuda_dependency_meta`, the rationale for the package as a I understood it is that if they depend on CUDA directly, then they need to depend on `autoware_cuda_dependency_meta`. Otherwise, if we're just filtering out packages that depend on `autoware_tensorrt_common` and `autoware_cuda_utils`, then I don't see the point of `autoware_cuda_dependency_meta` itself, when we can just filter out via those.

@xmfcx @mitsudome-r @youtalk Could we clarify this? Thanks.

xmfcx:
Right, I got confused when you said:
>  if the package depends on CUDA in one way or another, yes, it must depend on autoware_cuda_dependency_meta, even if some of its components can be built without CUDA.

I thought you've already added some packages that depended on CUDA indirectly. Sorry, for my misunderstanding🙇‍♂️.

I've also stated in the original issue to add only direct dependencies so I'm in line with you 👍

[autoware_bytetrack](https://github.com/autowarefoundation/autoware.universe/blob/fbc7a9d9cb617485d48ebe79f781e5741d0456dd/perception/autoware_bytetrack/CMakeLists.txt#L70) also depends on CUDA to build I guess, at least it is using some CUDA related parameters in its cmakelists file.

xmfcx:
**CMakeLists has tensorrt version variable passed to it, gray area?:**
- autoware_bytetrack

**Indirectly dependent on CUDA packages:**
- autoware_detected_object_validation
- autoware_detection_by_tracker
- autoware_raindrop_cluster_filter
- tier4_perception_launch
- tier4_simulator_launch

The rest are directly dependent and the deps are added within this PR ✅

esteve:
> [autoware_bytetrack](https://github.com/autowarefoundation/autoware.universe/blob/fbc7a9d9cb617485d48ebe79f781e5741d0456dd/perception/autoware_bytetrack/CMakeLists.txt#L70) also depends on CUDA to build I guess, at least it is using some CUDA related parameters in its cmakelists file.

You're right, I missed that one. Basically what I did was to `grep -ri find_package(CUDA` and add `autoware_cuda_dependency_meta` as a dependency in `package.xml`, but forgot to check for TensorRT. I'll add the dependency to `autoware_bytetrack`

xmfcx (approved):

PR #10074, opened 2025-02-05
feat(autoware_pointcloud_preprocessor): reuse collectors to reduce creation of collector and timer

vividf:
## Description
https://github.com/autowarefoundation/autoware.universe/pull/8300 introduces a collector and initializes a timer in each collector with a 100 ms interval, which leads to infinite object creation and deletion. 

This PR reuses collectors that are no longer in use (i.e., those that have completed concatenation) by resetting their timers, effectively converting them into new collectors. This approach helps limit unnecessary collector creation and timer initialization.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?
* Test with unit test. 
* Test with rosbag that provided in https://github.com/autowarefoundation/autoware.universe/pull/8300 
* Test with CARET and proved that it works better than before [TIER4_INTERNAL_LINK](https://star4.slack.com/archives/C07K1PPNPTR/p1738797207345439?thread_ts=1738207255.819239&cid=C07K1PPNPTR)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


YoshiRi (commented):
LGTM in codespace, but I did not check with any rosbags.

sensing/autoware_pointcloud_preprocessor/src/concatenate_data/concatenate_and_time_sync_node.cpp
```diff
@@ -347,13 +362,19 @@ void PointCloudConcatenateDataSynchronizerComponent::manage_collector_list()
...
   }
+
+  if (cloud_collectors_.size() > collectors_threshold) {
+    RCLCPP_WARN_STREAM_THROTTLE(
+      get_logger(), *get_clock(), 1000,
+      "The number of cloud collectors (" << cloud_collectors_.size() << ") exceeds the threshold ("
+                                         << collectors_threshold
+                                         << "), be careful if it keeps increasing.");
+  }
 }
```
mojomex, line 378: I'd like this to be a hard limit. In production, nobody will look at those logs, so the memory footprint could increase forever.

mojomex (changes requested):
Good stuff!
I've left a few comments in places I think could be simplified or where I'd like the more safety guarantees.

sensing/autoware_pointcloud_preprocessor/src/concatenate_data/concatenate_and_time_sync_node.cpp
```diff
@@ -230,20 +230,36 @@ void PointCloudConcatenateDataSynchronizerComponent::cloud_callback(
...
+    std::shared_ptr<CloudCollector> selected_collector = nullptr;
 
-    cloud_collectors_.push_back(new_cloud_collector);
-    cloud_collectors_lock.unlock();
+    for (auto & collector : cloud_collectors_) {
+      if (collector->get_status() == CollectorStatus::Idle) {
+        selected_collector = collector;
+        break;
+      }
+    }
```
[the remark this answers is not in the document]
vividf: Thanks! fixed in https://github.com/autowarefoundation/autoware.universe/pull/10074/commits/c99f1fe1d2f4ce1193ccfe65c62cdb9bcfd02702

sensing/autoware_pointcloud_preprocessor/src/concatenate_data/cloud_collector.cpp
```diff
@@ -67,17 +68,33 @@ bool CloudCollector::process_pointcloud(
...
+    // Check if the map already contains an entry for the same topic. This shouldn't happen if the
+    // parameter 'lidar_timestamp_noise_window' is set correctly.
+    if (topic_to_cloud_map_.find(topic_name) != topic_to_cloud_map_.end()) {
+      RCLCPP_WARN_STREAM_THROTTLE(
+        ros2_parent_node_->get_logger(), *ros2_parent_node_->get_clock(),
+        std::chrono::milliseconds(10000).count(),
+        "Topic '" << topic_name
+                  << "' already exists in the collector. Check the timestamp of the pointcloud.");
+    }
+  }
```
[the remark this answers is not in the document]
vividf: Thanks! 
The finished state is removed in https://github.com/autowarefoundation/autoware.universe/pull/10082

sensing/autoware_pointcloud_preprocessor/src/concatenate_data/concatenate_and_time_sync_node.cpp
vividf: Thanks!
Fixed with resetting collector that has oldest timestamp:
https://github.com/autowarefoundation/autoware.universe/pull/10074/commits/81ec941d6966c0dcefeffa2055c5c5df67fb212c and https://github.com/autowarefoundation/autoware.universe/pull/10074/commits/3375c43297263358f056b245444674bf62e7b356

vividf: fix logging: https://github.com/autowarefoundation/autoware.universe/pull/10074/commits/a8444cb60b93206d9eb3b3ebfe0c27681e63b60f

sensing/autoware_pointcloud_preprocessor/src/concatenate_data/concatenate_and_time_sync_node.cpp
```diff
@@ -368,12 +370,36 @@ void PointCloudConcatenateDataSynchronizerComponent::manage_collector_list()
...
+      // Update the iterator pointing to the collector with the smallest timestamp
+      if (timestamp < min_timestamp) {
+        min_timestamp = timestamp;
+        min_it = it;
+      }
+    }
+
+    // Reset the collector with the oldest timestamp if found
+    if (min_it != cloud_collectors_.end()) {
+      RCLCPP_WARN_STREAM(
```
mojomex, line 406: Probably warn_throttle would be good here as this can run `n_lidars x frequency` times per second in the worst case.

sensing/autoware_pointcloud_preprocessor/src/concatenate_data/concatenate_and_time_sync_node.cpp
mojomex, line 1: As discussed with @drwnz just now, it's probably a goof idea to allocate (on startup) `collectors_threshold` collectors and then not add/remove any during operation.
This makes it easier to reason about the program state.
(would also get rid of L251-L260)

sensing/autoware_pointcloud_preprocessor/src/concatenate_data/concatenate_and_time_sync_node.cpp
mojomex, line 406: Probably warn_throttle would be good here as this can run `n_lidars x frequency` times per second in the worst case.

mojomex (changes requested):
Probably hard-coding number of collectors to 2 or 3 would be okay. I don't see a use case with more collectors at the moment.
Also would allow us to allocate all of them once on startup instead of dynamically.

sensing/autoware_pointcloud_preprocessor/src/concatenate_data/concatenate_and_time_sync_node.cpp
vividf: Thanks, fixed in https://github.com/autowarefoundation/autoware.universe/pull/10074/commits/a8b512f33dea44196061136f37641b598cd6e4f0

sensing/autoware_pointcloud_preprocessor/src/concatenate_data/concatenate_and_time_sync_node.cpp
```diff
@@ -230,20 +231,37 @@ void PointCloudConcatenateDataSynchronizerComponent::cloud_callback(
...
+      [](const auto & collector) { return collector->get_status() == CollectorStatus::Idle; });
+
+    if (it != cloud_collectors_.end()) {
+      selected_collector = *it;
+    }
+
+    if (!selected_collector) {
+      // If no idle collector exists, create a new collector
+      selected_collector = std::make_shared<CloudCollector>(
+        std::dynamic_pointer_cast<PointCloudConcatenateDataSynchronizerComponent>(
```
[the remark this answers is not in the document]
vividf: Thanks!!!
Fixed in https://github.com/autowarefoundation/autoware.universe/pull/10074/commits/85f75e6f599c5bcc7121ee6fef13a1dc8ddbf432

sensing/autoware_pointcloud_preprocessor/src/concatenate_data/concatenate_and_time_sync_node.cpp
vividf: Thanks!!!
Fixed in https://github.com/autowarefoundation/autoware.universe/commit/85f75e6f599c5bcc7121ee6fef13a1dc8ddbf432

vividf:
@mojomex @drwnz  kindly ping

sensing/autoware_pointcloud_preprocessor/src/concatenate_data/concatenate_and_time_sync_node.cpp
```diff
@@ -347,11 +373,50 @@ void PointCloudConcatenateDataSynchronizerComponent::manage_collector_list()
...
+        }
+      }
+    }
+
+    // Reset the collector with the oldest timestamp if found
+    if (min_it != cloud_collectors_.end()) {
+      RCLCPP_WARN_STREAM_THROTTLE(
+        this->get_logger(), *this->get_clock(), 1000,
+        "Reset the oldest collector because the number of processing collectors ("
+          << num_processing_collectors << ") equal to the limit (" << num_of_collectors << ").");
```
drwnz, line 418:
```suggestion
          << num_processing_collectors << ") is equal to the limit of (" << num_of_collectors << ").");
```

drwnz (approved):
Please see the minor grammar comment, otherwise LGTM.

sensing/autoware_pointcloud_preprocessor/src/concatenate_data/concatenate_and_time_sync_node.cpp
vividf: Thanks! fixed in https://github.com/autowarefoundation/autoware.universe/pull/10074/commits/8a677270602b34a1c1444eb429d6622e109d5b90

mojomex (commented):
I've finally gotten around to test it! I used some gen1 taxi data and copied its concat parameters from one of our projects.

Looking good I think!
One question that might as well be me not understanding how to configure concat:
When deliberately messing up the params for one LiDAR, I would (intuitively) expect that there isn't a huge drop in concat output frequency, but rather that a group is dropped only sometimes, or every second time, etc.

I think this might be because we are always killing the oldest group, never giving it a chance to finish. It might also be because my timeouts are so long that almost all waiting collectors will be cleared because the inputs come at 10Hz.

Let me know your thoughts :bow: 

### Naive mode
[small-compressed-naive.webm](https://github.com/user-attachments/assets/7a9c4ad2-a235-4a67-b111-69f964a572c2)


### Advanced mode, correct params
[small-compressed-advanced_correct.webm](https://github.com/user-attachments/assets/3f89bae4-f3ce-41e1-93b0-d2550d8e7f12)

### Advanced mode, incorrect params
[small-compressed-advanced_incorrect.webm](https://github.com/user-attachments/assets/d54fe616-2f00-44c1-8fb8-267d57f42638)

mojomex (approved):
After talking offline, it is clear that a user can always misconfigure concat enough to look like the bottom video - that's the user's fault, but the algorithm itself seems to work fine :+1: 
LGTM

PR #10082, opened 2025-02-07
fix(autoware_pointcloud_preprocessor): fix potential double unlock in concatenate node

vividf:
## Description

This PR resolves a potential double unlock issue in the concatenate node. Previously, under certain conditions, the mutex could be unlocked twice, leading to undefined behavior and potential segmentation faults.

## Related links
Note: This PR needs to rebase after https://github.com/autowarefoundation/autoware.universe/pull/10074 is merged.

**Parent Issue:**

- Link

## How was this PR tested?
* Tested with unit test.
* Tested with sample rosbag

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


mojomex:
As remarked by @knzo25 offline some days ago, and as per the discussion @drwnz @vividf @manato and I just had,. it might be better to remove all locking completely as the callbacks within one node will always be in a mutually exclusive callback group by default, even when using a multithreaded executor.
This makes concat effectively single-threaded and thus no locking is needed. Performance hasn't be a problem even with our more data-heavy architectures.

knzo25:
@mojomex 
My comments though, require checking how threading, callback groups and timers work in ros.
In my experience, some timers (there are several, and some work at a lower level) had weird behaviors (different clocks were used ages ago, during some transitions, probably from ros1)

sensing/autoware_pointcloud_preprocessor/src/concatenate_data/concatenate_and_time_sync_node.cpp
```diff
@@ -242,36 +241,36 @@ void PointCloudConcatenateDataSynchronizerComponent::cloud_callback(
...
 
     if (it != cloud_collectors_.end()) {
       selected_collector = *it;
+    } else {
+      auto oldest_it = find_and_reset_oldest_collector();
+      if (oldest_it != cloud_collectors_.end()) {
+        selected_collector = *oldest_it;
+      } else {
+        // Handle case where no suitable collector is found
+        RCLCPP_WARN(get_logger(), "No available CloudCollector in IDLE state.");
```
mojomex, line 263: This should probably be an error (the invariant of the program, that there is at least 1 idle one after `find_and_reset_oldest_collector` is broken)
```suggestion
        RCLCPP_ERROR(get_logger(), "No available CloudCollector in IDLE state.");
```

mojomex (approved):

mojomex (approved):
LGTM!
There's two small aesthetic suggestions I left but the algorithm itself loogs good and works fine :+1:

sensing/autoware_pointcloud_preprocessor/src/concatenate_data/concatenate_and_time_sync_node.cpp
vividf: addressed in https://github.com/autowarefoundation/autoware.universe/pull/10082/commits/0c5de07353ecceb159c26598f9531dd8939a2351

sensing/autoware_pointcloud_preprocessor/src/concatenate_data/concatenate_and_time_sync_node.cpp
```diff
@@ -229,7 +229,7 @@ void PointCloudConcatenateDataSynchronizerComponent::cloud_callback(
 
   std::shared_ptr<CloudCollector> selected_collector = nullptr;
 
-  // Try to find a matching collector
+  // For each callback, check whether there is a exist collector that matches this cloud
```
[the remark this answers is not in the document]
vividf: addressed in https://github.com/autowarefoundation/autoware.universe/pull/10082/commits/0c5de07353ecceb159c26598f9531dd8939a2351

SakodaShintaro:
Unfortunatelly, merging "main" didn't resolve the failure of build-test-tidy-pr :cry: 

```
 failed to register layer: write /usr/local/cuda-12.4/targets/x86_64-linux/lib/libcublasLt.so.12.4.5.8: no space left on device
  Error: Docker pull failed with exit code 1
```

technolojin:
recommend to update the branch to solve build-test  
https://github.com/autowarefoundation/autoware.universe/pull/10198

PR #10087, opened 2025-02-10
fix(autoware_image_projection_based_fusion): modify incorrect index access in pointcloud filtering for out-of-range points

keita1523:
## Description
Before
In [this line](https://github.com/tier4/autoware.universe/blob/79dc610ee36432aa6ad4522376a7561fec27d908/perception/image_projection_based_fusion/src/pointpainting_fusion/node.cpp#L239), the point cloud filtering process for out-of-range points increments output variables (such as "iter_painted_x") at the same time as "iter_x".
However, in the following process, these output variables are used until "j", which is the size of points within the range.
As a result, data is stored at indices beyond "j", leading to incorrect index access.

After
I modified the timing of the increments for "iter_painted_x" so that points within the range are stored correctly until "j".

## Related links

**Parent Issue:**

- Link

## How was this PR tested?
[TIER IV INTERNAL LINK](https://star4.slack.com/archives/C076ZGRC15X/p1739289211379129?thread_ts=1737939453.841099&cid=C076ZGRC15X)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


technolojin (approved):
LGTM

technolojin:
@YoshiRi @yukkysaito Can you review this PR? it is to fix a bug in the pointpainting.

yukkysaito (approved):
LGTM

PR #9731, opened 2024-12-23
feat(tier4_perception_launch): add option for new TL detector model

badai-nguyen:
## Description
- Add the option for new TL detector which describe at https://github.com/autowarefoundation/autoware.universe/issues/9719

## Related links
- Need to be merged after new packages and autoware_launch update for new TLR are mergerd: 
  - https://github.com/autowarefoundation/autoware.universe/pull/9721
  - https://github.com/autowarefoundation/autoware.universe/pull/9748
  - https://github.com/autowarefoundation/autoware.universe/pull/9828
  - https://github.com/autowarefoundation/autoware/pull/5764
  - And https://github.com/autowarefoundation/autoware_launch/pull/1310 

**Parent Issue:**

- [Link](https://github.com/autowarefoundation/autoware.universe/issues/9719)
- [TIER IV internal link ](https://tier4.atlassian.net/browse/RT1-8911)

## How was this PR tested?
I checked new tlr and existing perception pipeline on Evaluator.
https://evaluation.ci.tier4.jp/evaluation/reports/e545bfca-9777-5031-9e92-818bd694e2df?project_id=prd_jt

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

PR #10089, opened 2025-02-10
fix(autoware_traffic_light_selector): add camera_info into message_filter

MasatoSaeki:
## Description

add camera_info into message_filter.
If we do not add it, have to handle with mutex to be thread safe. And subscribe topics have same timestamp, So It is efficienfy to subscribe together.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

I checked this PR on my local PC and merging below PR.
- https://github.com/autowarefoundation/autoware.universe/pull/9731
- https://github.com/autowarefoundation/autoware_launch/pull/1310

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


YoshiRi (approved):
LGTM

PR #10104, opened 2025-02-12
perf(behavior_velocity_planner): replace getSortedLaneIdsFromPath

Enigamict:
Signed-off-by: Enigamict atsuki.takata@tier4.jp

## Description
cc @veqcc 
In getSortedLaneIdsFromPath(), we originally used std::find to check for duplicates in the lane_id array. However, the linear search with std::find was too expensive, as shown on the FlameGraph, which indicates that getSortedLaneIdsFromPath imposes a significant load.

Screenshot from 2025-02-10 15-05-48

To address this, we changed the duplicate-checking approach to use an unordered_set. Instead of conditionally calling emplace_back(lane_id) only if it is not found in sorted_lane_ids, we now use unordered_set::insert(). This reduces the overall computational cost by eliminating repeated linear searches.

We evaluated the performance impact using percentile metrics (50%, 75%, 90%, 95%, 99%) on BehaviorVelocityPlannerNode::generatePath.

| Percentile | Before (us) | After (us)     |
| ---------- | ----------- | -------------- |
| **50%**    | 9814.00     | 6357.50|
| **75%**    | 13143.25    | 8567.75|
| **90%**    | 15864.80    | 11829.50|
| **95%**    | 17400.50    | 13691.00|
| **99%**    | 20763.05    | 17032.05|

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_velocity_planner/autoware_behavior_velocity_planner_common/src/utilization/util.cpp
+++ b/planning/behavior_velocity_planner/autoware_behavior_velocity_planner_common/src/utilization/util.cpp
@@ -562,16 +562,13 @@ std::set<int64_t> getLaneIdSetOnPath(
 
 std::vector<int64_t> getSortedLaneIdsFromPath(const PathWithLaneId & path)
 {
-  std::vector<int64_t> sorted_lane_ids;
+  std::unordered_set<int64_t> sorted_lane_ids;
   for (const auto & path_points : path.points) {
-    for (const auto lane_id : path_points.lane_ids)
-      if (
-        std::find(sorted_lane_ids.begin(), sorted_lane_ids.end(), lane_id) ==
-        sorted_lane_ids.end()) {
-        sorted_lane_ids.emplace_back(lane_id);
-      }
+    for (const auto lane_id : path_points.lane_ids) {
+      sorted_lane_ids.insert(lane_id);
+    }
   }
-  return sorted_lane_ids;
+  return std::vector<int64_t>(sorted_lane_ids.begin(), sorted_lane_ids.end());
 }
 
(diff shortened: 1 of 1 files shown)


veqcc:
@atsushi421 
Thank you for your contribution!!

Could you add somthing to `How was this PR tested?` section, if you conducted Psim or Lsim?

soblin (changes requested):
As the function name suggests, this function must make the set  of `lane_ids` unique while preserving the order of those elements, which means we need to remember the insertion order on that container.

One approach is use this container: https://boostjp.github.io/tips/multi_index.html#basic-usage (or you can achieve this in your own way of course), thanks

soblin:
@Enigamict For example, if the `lane_ids` are 

```
{ {30, 20}, {20, 10, 15}, {15, 40} }
```
, then the desired output is 

```
{ 30, 20, 10, 15, 40 }
```

Enigamict:
@veqcc @soblin 
Thank you for the review. I now understand the behavior of getSortedLaneIdsFromPath and appreciate the explanation about lane_ids. I realize that the current modification doesn't consider the order, so it needs to be adjusted. I'll make the necessary corrections based on your feedback.

Enigamict:
We apologize for any inconvenience caused. After trying various methods, we have seen no improvement in performance, and the traditional method remains the best at this time.

Thank you very much for your review and feedback. We will now close this pull request.

Enigamict:
@veqcc @soblin
We apologize for any inconvenience caused. After trying various methods, we have seen no improvement in performance, and the traditional method remains the best at this time.

Thank you very much for your review and feedback. We will now close this pull request.

PR #10133, opened 2025-02-17
fix(tier4_vehicle_launch): add missing exec_depend

isamu-takagi:
## Description

Add exec_depend to the package specified by find-pkg-share.

## Related links

None

## How was this PR tested?

Check if rosdep install succeeds.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


yukkysaito (approved):
LGTM

PR #10135, opened 2025-02-17
fix(common): fix package name of find-pkg-share

isamu-takagi:
## Description

~Add exec_depend to the package specified by find-pkg-share.~
The problem was with find-pkg-share, not exec_depend, so I fixed that.

## Related links

None

## How was this PR tested?

Check if rosdep install succeeds.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


taikitanaka3:
@isamu-takagi 
could you fix pr title？

it should be fix launch pkg name etc..
title is not add exec depend.

isamu-takagi:
@taikitanaka3 Thank you. I fixed it.

taikitanaka3 (approved):
lgtm

PR #10146, opened 2025-02-18
ci(build-depends.repos): separate repos for humble and nightly

xmfcx:
## Description

- https://github.com/autowarefoundation/autoware.universe/pull/10148

This PR is for testing purposes.

Once the pr mentioned above is merged, I will cherry pick that and recreate this PR.

## How was this PR tested?

- **check-build-depends:** https://github.com/autowarefoundation/autoware.universe/actions/runs/13384933484/job/37379802371?pr=10146
- **bnt-diff:** https://github.com/autowarefoundation/autoware.universe/actions/runs/13384933539/job/37379921043?pr=10146

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

PR #10148, opened 2025-02-18
ci(build-depends.repos): separate repos for humble and nightly

xmfcx:
## Description

**Related Issue:**
- https://github.com/autowarefoundation/autoware/issues/5792

**Humble counterpart:**
- https://github.com/autowarefoundation/autoware.universe/pull/10146

`build_depends.repos` file is used for pulling the dependencies of this repository while building it, currently used by CI.

With the introduction of [`autoware-nightly.repos`](https://github.com/autowarefoundation/autoware/blob/cad73fb1a431937e674bfeb067a9e1dbab99840f/autoware-nightly.repos)  and pinning of the versions in [`autoware.repos`](https://github.com/autowarefoundation/autoware/blob/cad73fb1a431937e674bfeb067a9e1dbab99840f/autoware.repos), the management of this file became a bit complicated.

This repository has 2 branches: `main` and `humble`. This PR introduces the following configuration for both these:

### PRs against `main` branch & CI that operates on `main`

They should utilize `build_depends_humble.repos` that is overridden by `build_depends_nightly.repos`.

To simplify this, I've added a new action that will overlay one `.repos` file onto another.

Using this new `combine-repos-action` these CI workflows now use the overlaid `.repos` file.

The reason for not having 2 `.build_depends.repos` files with some duplicate information and going with override option is to simplify version maintenance.

### PRs against `humble` branch

These utilize `build_depends_humble.repos` as it is. The condition for this case is only triggered by `*-differential` workflows.

## How was this PR tested?

- **build-and-test-daily:** https://github.com/autowarefoundation/autoware.universe/actions/runs/13384761188
- **check-build-depends:** https://github.com/autowarefoundation/autoware.universe/actions/runs/13384698882/job/37379204369?pr=10148

### DIFF
--- a/.github/actions/combine-repos-action/README.md
+++ b/.github/actions/combine-repos-action/README.md
@@ -1,0 +1,43 @@
+# combine-repos-action
+
+## Description
+
+**Combine Repos Action** is a GitHub Action designed to merge two `.repos` files—a base file and an overlay file—into a single file. The overlay file takes precedence over the base file when duplicate repository entries exist. This is especially useful for projects that need to dynamically combine configuration files for dependency management or CI workflows.
+
+## Usage
+
+Below is an example of how to include it in your workflow:
+
+```yaml
+jobs:
+  combine:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Combine Repos Files
+        uses: ./.github/actions/combine-repos-action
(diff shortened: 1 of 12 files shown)


mitsudome-r (approved):
LGTM. We can merge it once we confirm that https://github.com/autowarefoundation/autoware.universe/actions/runs/13384761188/job/37379350785 passes.

PR #10151, opened 2025-02-18
chore(autoware_map_projection_loader): show map error details

yhisaki:
## Description

Previously, when autoware_map_projection_loader failed to load a map file, it did not show the details of the error. In this PR I have added the codes to show the errors.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/map/autoware_map_projection_loader/src/load_info_from_lanelet2_map.cpp
+++ b/map/autoware_map_projection_loader/src/load_info_from_lanelet2_map.cpp
@@ -21,6 +21,7 @@
 #include <lanelet2_io/Io.h>
 #include <lanelet2_projection/UTM.h>
 
+#include <sstream>
 #include <string>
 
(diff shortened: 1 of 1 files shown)


TaikiYamada4 (approved):
I've checked that the expected error message appear when I overwrote [this line in the test code](https://github.com/autowarefoundation/autoware.universe/blob/69f7964a5872adfd13f2abda5f23e933d4f9c0f5/map/autoware_map_projection_loader/test/test_load_info_from_lanelet2_map.cpp#L98) to apply the test to [a map whose traffic light regulatory element doesn't have a "refers"](https://github.com/tier4/autoware_lanelet2_map_validator/blob/main/autoware_lanelet2_map_validator/test/data/map/traffic_light/traffic_light_regulatory_element_without_refers.osm). :+1:

PR #10175, opened 2025-02-20
fix(autoware_traffic_light_category_merger): add missing dependency to autoware_universe_utils

mitsudome-r:
## Description

This adds missing dependency of autoware_universe_utils in package.xml.

Currently, the dependency is introduced via autoware_test_utils, but autoware_test_utils will no longer depend on autoware_universe_utils after porting to Autoware Core in https://github.com/autowarefoundation/autoware.core/pull/172. Therefore, we should explicitly define dependency to use autoware_universe_utils in this package. (like in [this line](https://github.com/autowarefoundation/autoware.universe/blob/81489202f5f98ef84917de82a7f40337e4f0336c/perception/autoware_traffic_light_category_merger/src/traffic_light_category_merger_node.hpp#L18))

## Related links

- Related Discussion: https://github.com/orgs/autowarefoundation/discussions/5365
- Related PR: https://github.com/autowarefoundation/autoware.core/pull/172

## How was this PR tested?
Build test in this PRs CI.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_traffic_light_category_merger/package.xml
+++ b/perception/autoware_traffic_light_category_merger/package.xml
@@ -13,6 +13,7 @@
   <buildtool_depend>ament_cmake_auto</buildtool_depend>
 
   <depend>autoware_test_utils</depend>
+  <depend>autoware_universe_utils</depend>
   <depend>message_filters</depend>
   <depend>rclcpp</depend>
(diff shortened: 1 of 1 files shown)


MasatoSaeki (approved):
Thank you!

PR #10179, opened 2025-02-20
feat(mission_planner): tolerate goal footprint being inside the previous lanelets of closest lanelet

soblin:
## Description

In addition to this PR https://github.com/autowarefoundation/autoware.universe/pull/8760, if the goal is just on the very beginning of a lanelet, ego's goal footprint becomes outside of the ego_lanes. To avoid this, I added previous lanelet of closest_goal_lanelet as well.

image

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

- https://evaluation.tier4.jp/evaluation/reports/7f55984f-52be-5218-bd44-c119159844aa?project_id=prd_jt

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/autoware_mission_planner_universe/src/lanelet2_plugins/default_planner.cpp
+++ b/planning/autoware_mission_planner_universe/src/lanelet2_plugins/default_planner.cpp
@@ -207,6 +207,13 @@ bool DefaultPlanner::check_goal_footprint_inside_lanes(
     boost::geometry::correct(poly);
     ego_lanes.push_back(poly);
   }
+  // If the goal is on the very beginning of the closest_lanelet_to_goal, baselink ~ rear part of
+  // ego footprint is outside of it. To tolerate it, add previous lanelet
+  for (const auto & ll : route_handler_.getPreviousLanelets(closest_lanelet_to_goal)) {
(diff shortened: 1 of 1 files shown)


planning/autoware_mission_planner_universe/src/lanelet2_plugins/default_planner.cpp
```diff
@@ -207,6 +207,13 @@ bool DefaultPlanner::check_goal_footprint_inside_lanes(
     boost::geometry::correct(poly);
     ego_lanes.push_back(poly);
   }
+  // If the goal is on the very beginning of the closest_lanelet_to_goal, baselink ~ rear part of
+  // ego footprint is outside of it. To tolerate it, add previous lanelet
+  for (const auto & ll : route_handler_.getPreviousLanelets(closest_lanelet_to_goal)) {
+    boost::geometry::convert(ll.polygon2d().basicPolygon(), poly);
+    boost::geometry::correct(poly);
+    ego_lanes.push_back(poly);
+  }
```
kosuke55, line 216: memo for limitation:
If the previous is very short, it may stick out. In such a case, it is necessary to take all previous lanes for footprints.
This is not the case here.

kosuke55 (approved):

PR #10190, opened 2025-02-24
feat(behavior_velocity_planner_common): refactor func name,change package dependency,etc

storrrrrrrrm:
## Related links

**Parent Issue:**

- https://github.com/autowarefoundation/autoware.core/issues/190

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


storrrrrrrrm:
1.do not refactor func name. 
2. create another pr ,mege into autoware.core directly

PR #10197, opened 2025-02-26
build-and-test-differential switch to codebuild test PR

mitsudome-r:
## Description

**Do not merge this PR!!!**

Merge this instead:
- https://github.com/autowarefoundation/autoware.universe/pull/10198

## Related links

- https://github.com/autowarefoundation/autoware.universe/pull/10024

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


.github/workflows/build-and-test-differential.yaml
```diff
@@ -22,6 +22,10 @@ on:
         required: false
         default: ""
         type: string
+      build-pre-command:
+        required: false
+        default: nice -n 19
```
xmfcx, line 27:
```suggestion
        default: build-pre-command: taskset --cpu-list 0-6
```

Let's keep the default functional too.

xmfcx:
https://github.com/autowarefoundation/autoware.universe/actions/runs/13534456218/job/37823440854?pr=10197

It failed for normal reasons (not for disk space) so I think we can merge it.

I'll remove the unrelated changes.

xmfcx:
Closing since the test has accomplished its purpose.

PR #10198, opened 2025-02-26
ci: use Code Build for build-and-test-differentianl-cuda

mitsudome-r:
## Description
Since the size of -cuda image has increased, Autoware build-and-test-differential-cuda started to fail due to disk space issue ([example](https://github.com/autowarefoundation/autoware.universe/actions/runs/13452761443/job/37590161479?pr=10082))

image

Ideally, we should investigate means to reduce the images size, but this PR modifies the workflow to use Code Build which has larger disk space to fix the error for the time being.

## Related links

- https://github.com/autowarefoundation/autoware.universe/actions/runs/13452761443/job/37590161479?pr=10082

## How was this PR tested?

Tested here: https://github.com/autowarefoundation/autoware.universe/actions/runs/13534456218/job/37823440854?pr=10197#logs

I have tested using a example [PR](https://github.com/autowarefoundation/autoware.universe/pull/10024) that uses CUDA. 
The action is failing at the test code that was modified in the example PR, but it at least surpass the "Intialize Container" step so it should be enough to confirm the validity of this PR.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/.github/workflows/build-and-test-differential.yaml
+++ b/.github/workflows/build-and-test-differential.yaml
@@ -22,6 +22,10 @@ on:
         required: false
         default: ""
         type: string
+      build-pre-command:
+        required: false
+        default: nice -n 19
+        type: string
     secrets:
       codecov-token:
         required: true
@@ -113,7 +117,7 @@ jobs:
           target-packages: ${{ steps.get-modified-packages.outputs.modified-packages }}
           build-depends-repos: build_depends.repos
           cache-key-element: ${{ env.BUILD_TYPE_CUDA_STATE }}
-
+          build-pre-command: ${{ inputs.build-pre-command }}
       - name: Show ccache stats after build
         run: du -sh ${CCACHE_DIR} && ccache -s
         shell: bash
--- a/.github/workflows/build-and-test-differential.yaml
+++ b/.github/workflows/build-and-test-differential.yaml
@@ -24,7 +24,7 @@ on:
         type: string
       build-pre-command:
         required: false
-        default: nice -n 19
+        default: ""
         type: string
     secrets:
       codecov-token:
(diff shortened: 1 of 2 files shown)


.github/workflows/build-and-test-differential.yaml
```diff
@@ -22,6 +22,10 @@ on:
         required: false
         default: ""
         type: string
+      build-pre-command:
+        required: false
+        default: nice -n 19
```
xmfcx, line 27:
```suggestion
        default: taskset --cpu-list 0-6
```

xmfcx (changes requested):
let's keep the default functional too

.github/workflows/build-and-test-differential.yaml
xmfcx: On a second look, this is to be used by the non-cuda build, so it's better to keep it empty.
```suggestion
        default: ""
```

mitsudome-r: I just kept it same so that non-cuda build run just like before, but I'm okay with keeping it empty. (I think it will still work fine)
Do you want me to change it?

xmfcx: Still, better to send it empty. I will create a PR to make the default on the action side empty too.

xmfcx: - Created https://github.com/autowarefoundation/autoware-github-actions/pull/336 to fix the default too.

mitsudome-r: fixed in https://github.com/autowarefoundation/autoware.universe/pull/10198/commits/376227b7aec03cba5d2ba156dd56b8bed1574f2e

mitsudome-r:
I'm doing a test with empty build-pre-command here: https://github.com/autowarefoundation/autoware.universe/actions/runs/13536085896/job/37827946062?pr=10197

xmfcx (approved):

PR #10227, opened 2025-03-04
fix(obstacle_cruise_planner): ignore invalid stopping objects

KYabuuchi:
## Description
This PR modifies the conditions for obstacles that should be yielded to.

According to the [README](https://github.com/autowarefoundation/autoware.universe/tree/4649cf69d8ffa876ebd9f4bc80f2f19eda63a6bc/planning/motion_velocity_planner/autoware_motion_velocity_obstacle_cruise_module#yield-for-vehicles-that-might-cut-in-into-the-egos-lane), stopping objects must meet the following three conditions:
1. The object type is designated by `obstacle_filtering.object_type.*` .
2. The lateral offset between the stopping object and the moving object is within `yield.max_lat_dist_between_obstacles`.
3. The lateral offset of the stopping object from the ego trajectory is within `yield.max_lat_dist_between_obstacles` + `yield.max_lat_dist_between_obstacles` .

**However, the current implementation does not check condition ①** . :mechanic:  This PR adds the missing condition ① to the filtering logic.

> [!NOTE]
> :mechanic:   Additionally, I have added a filtering type for stopping objects.

## Related links

**autoware_launch PR**
-  https://github.com/autowarefoundation/autoware_launch/pull/1354

**Private Links:**
- [TIER IV internal link(RT0-35652)](https://tier4.atlassian.net/browse/RT0-35652)

### planning_simulator.launch.xml + manual obstacle placing (rviz plugin)

| before | after|
|--| --|
| :x: Yielded to obstacles behind an UNKNOWN | :heavy_check_mark:  Did not yield to obstacles behind an UNKNOWN |
| image |image |

:heavy_check_mark:  Even after this change, the ego vehicle yielded when the specified type is ahead 

### planning_simulator.launch.xml + perception_reproducer

- I used this rosbag [TIER IV internal link(rosbag)](https://console.data-search.tier4.jp/projects/prd_jt/environments/c73f858a-20d4-4ea8-9a3c-81581584ea0f/rosbags/4e9de40e-e545-48c0-b83d-d88aa7702b9d)

| before | after|
|--| --|
| :x: Yielded to obstacles behind an UNKNOWN | :heavy_check_mark:  Did not yield to obstacles behind an UNKNOWN |
| image|image|

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


takayuki5168 (commented):
Thank you for the PR!
The object type with is_outside_cruise_obstacle is for the object to follow.

In this PR's case, the object type you are going to change is not the object to follow. So can you define this additional object type in the yaml file like the following?
```
      obstacle_filtering:
        object_type:
          inside:
            unknown: true
            car: true
            truck: true
            bus: true
            trailer: true
            motorcycle: true
            bicycle: true
            pedestrian: false

          outside:
            unknown: false
            car: true
            truck: true
            bus: true
            trailer: true
            motorcycle: true
            bicycle: false
            pedestrian: false

          blocking: # <- new: "blocking" is very confusing. Please think about a better name. 
            unknown: false
            car: true
            truck: true
            bus: true
            trailer: true
            motorcycle: true
            bicycle: false
            pedestrian: false
```

KYabuuchi:
@takayuki5168  Hi. I added the `ahead_stopped` parameter based on your feedback.

I couldn't come up with a better parameter name either. So if you have any suggestions for a more appropriate name, please let me know. :pray:

planning/motion_velocity_planner/autoware_motion_velocity_obstacle_cruise_module/config/obstacle_cruise.param.yaml
```diff
@@ -95,6 +95,17 @@
             bicycle: false
             pedestrian: false
 
+          # If an object does not have this label, the ego vehicle does not assume that the yielded obstacle will cut in due to the object.
+          ahead_stopped:
```
takayuki5168, line 99: ahead sounds like the obstalce is ahead of the ego, but on the ego's trajectory.
How about `side_stopped` instead?

KYabuuchi: Thank you! It sound good for me. I will update parameters soon.

KYabuuchi: I updated the parameter in 90f6046

takayuki5168 (approved):

takayuki5168:
Thank you for the fix!

PR #10232, opened 2025-03-05
fix(obstacle_cruise_planner): fix obstacle filtering logic

KYabuuchi:
## Description

This PR fixes two issues related to yielding:  

1. **`find_yield_cruise_obstacles()` being called incorrectly**
2. **Incorrect lateral filtering for yield targets on right side**  

## 1. `find_yield_cruise_obstacles()` being called incorrectly  
`find_yield_cruise_obstacles()` takes `objects` as an argument, but calling it inside a `for(object: objects)` loop is inappropriate. This can cause the following issues:  

- **Multiple unnecessary calls:** `find_yield_cruise_obstacles()` may be called multiple times within a single period, though only the first call is necessary.  
- **Missed yield targets:** If there are no obstacle cruise targets and the loop aborts early, `find_yield_cruise_obstacles()` is never called, even when yield targets exist.  

:mechanic:    To resolve this, `find_yield_cruise_obstacles()` has been moved outside the `for(object: objects)`  loop.  

## 2. Incorrect lateral filtering for yield targets  
Currently, an object is considered a yield target if its relative lateral position is within `yield_lat_distance_threshold`.  

However, because the relative position is evaluated using a **signed** distance (`get_dist_to_traj_lateral()`) , all vehicles to the right of the ego vehicle could be incorrectly considered yield targets.  

:mechanic:   To resolve this, updates the evaluation to use the **absolute** value of the relative position.  

In the original implementation, a vehicle could still be considered a yield target regardless of how far it was to the right of the ego lane, as shown in the diagram below.

| before | after|
| - | -| 
|❌ Yield to a very far obstacle in right side.  |✔️ Does not yield to a very far obstacles in right side |
||image|

## Related links

**Parent Issue:**
- No Parent Issue.

[TIER IV Internal Ticket: RT0-35652](https://tier4.atlassian.net/browse/RT0-35652)

## How was this PR tested?

* planning_simulator.launch.xml + manual obstacle placing (rviz plugin)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


KYabuuchi:
@maxime-clem  @takayuki5168  @yuki-takagi-66  Hi! Please take a look when you have some time. :pray:

takayuki5168 (approved):
Thank you so much!
Did you check if there is no degradation on common scenario tests?

KYabuuchi:
Thanks for the review!
I haven't tested it in the common scenario, so I will evaluate it soon.
I will merge after checking the evaluation results.

KYabuuchi:
Here is Evaluation Result: 
2025-03-14 https://evaluation.tier4.jp/evaluation/reports/ff57f41d-b131-5c51-9c11-4c385c53bafb?project_id=prd_jt

There is no degration, so I am going to merge.

PR #10226, opened 2025-03-04
feat(autoware_pointcloud_preprocessor): add pointcloud_densifier package

kaancolak:
## Description

For more details : https://github.com/autowarefoundation/autoware.universe/issues/10225
## Related links

**Parent Issue:**

- Link : https://github.com/autowarefoundation/autoware.universe/issues/10225

**Private Links:**
- [Tier4 internal link](https://tier4.atlassian.net/browse/ISSJP-2342)

## How was this PR tested?

run -> `ros2 launch autoware_pointcloud_preprocessor pointcloud_densifier.launch.py`

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

Issue #10225, opened 2025-03-04
Add pointcloud_densifier package


kaancolak:
### Checklist

- [x] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [x] I've searched other issues and no duplicate issues were found.
- [x] I've agreed with the maintainers that I can plan this task.

### Description

This issue addresses the implementation of a PointCloud Densifier component for Autoware, which enhances sparse LiDAR point cloud data by leveraging information from previous frames. Long-range LiDAR data is often sparse(after 80-90 meters), limiting the effectiveness of downstream perception algorithms. This component overcomes this limitation by selectively integrating points from previous frames into the current frame, resulting in a denser representation of the environment.

### Purpose

The main purpose is to detect long range stationary vehicles on the road, such as those stopped at an intersection or traffic light.

This package increases the density of long-range points for detects stationary vehicles with rule-based LiDAR algorithms. Additionally, we plan to create a pull request (PR) to add  roi_excluded_downsample_filter for preventing the downsampling of long-range points, with OT128 stationary objects detection accuracy improved to around 150-200 meters.

### Possible approaches

The algorithm works as follows:

1- **ROI Filtering:** First filters the input point cloud to only keep points in a specific region of interest (ROI), typically focused on the distant area in front of the vehicle.

2- **Occupancy Grid Creation:** Creates a basşc 2D occupancy grid from the filtered points to track which areas contain valid points in the current frame.

3- **Previous Frame Integration:** Transforms points from previous frames into the current frame's coordinate system using TF transformations.

4- **Selective Point Addition:** Adds points from previous frames only if they fall into grid cells that are occupied in the current frame. This ensures that only relevant points are added, avoiding ghost points from dynamic objects.

5- **Combined Output:** Returns a combined point cloud that includes both the current frame's points and selected points from previous frames.

The implementation follows Autoware's filtering architecture by inheriting from the `Filter` base class, providing compatibility with the existing preprocessing pipeline and support for point cloud indices.

### Definition of done

- [ ] Create meta package

PR #10233, opened 2025-03-05
feat(autoware_pointcloud_preprocessor): add roi excluded downsample

kaancolak:
## Description

This filter provides selective downsampling that preserves detailed point data in critical regions while reducing overall point cloud density:

1. Points are classified based on their position relative to a defined Region of Interest (ROI)
2. Points inside the ROI are preserved at their original density (no downsampling)
3. Points outside the ROI undergo standard voxel grid downsampling
4. The final output combines both sets of points

The main purpose is to detect long-range stationary vehicles on the road, such as those stopped at an intersection or traffic light. With the point cloud densifier node, points at a distance will be densified and downsampling will not be applied in this densified area thanks to the roi excluded voxel-based downsampling operation.

https://github.com/user-attachments/assets/2898d96e-e418-462c-80d4-9e84a2856a86

**Parent Issue:**

- Relevant issue : https://github.com/autowarefoundation/autoware.universe/issues/10225
  Relevant issue: https://github.com/autowarefoundation/autoware.universe/pull/10226

**Private Links:**
- [Tier4 internal link](https://tier4.atlassian.net/browse/ISSJP-2342)

## Notes for reviewers

run -> `ros2 launch autoware_pointcloud_preprocessor roi_excluded_voxel_grid_downsample_filter.launch.xml`

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/sensing/autoware_pointcloud_preprocessor/config/voxel_grid_downsample_filter_node.param.yaml
+++ b/sensing/autoware_pointcloud_preprocessor/config/voxel_grid_downsample_filter_node.param.yaml
@@ -3,3 +3,8 @@
     voxel_size_x: 0.3
     voxel_size_y: 0.3
     voxel_size_z: 0.1
+    enable_roi_exclusion  : false
+    roi_x_min: 80.0
+    roi_x_max: 220.0
+    roi_y_min: -20.0
+    roi_y_max: 20.0
--- a/sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/downsample_filter/faster_voxel_grid_downsample_filter.hpp
+++ b/sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/downsample_filter/faster_voxel_grid_downsample_filter.hpp
@@ -35,7 +35,8 @@ public:
   FasterVoxelGridDownsampleFilter();
   void set_voxel_size(float voxel_size_x, float voxel_size_y, float voxel_size_z);
   void set_field_offsets(const PointCloud2ConstPtr & input, const rclcpp::Logger & logger);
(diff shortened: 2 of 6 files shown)


kaancolak:
@YoshiRi @knzo25 @drwnz Could you please check this PR or assign an appropriate reviewer?

sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/downsample_filter/pickup_based_voxel_grid_downsample_filter_node.hpp
```diff
@@ -52,7 +52,8 @@ class PickupBasedVoxelGridDownsampleFilterComponent
   OnSetParametersCallbackHandle::SharedPtr set_param_res_;
 
   /** \brief Parameter service callback */
-  rcl_interfaces::msg::SetParametersResult paramCallback(const std::vector<rclcpp::Parameter> & p);
+  rcl_interfaces::msg::SetParametersResult paramCallback(
+    const std::vector<rclcpp::Parameter> & p);  // NOLINT
```
YoshiRi, line 56: Thanks but I do not think this change is needed in this PR.

YoshiRi (commented):
@kaancolak 
Sorry for late reviewing.
Code itself looks okay to me. But do you really need this filter now?
I mean voxel grid downsample in far distance won't do much filtering, so I wonder if just adding roi cropping function in voxel downsampling filter will satisfy your case and that would be more versatile.

kaancolak:
@YoshiRi  -san, I updated the code to enable and disable roi exclusion with a single parameter in the voxel grid downsample filter and kept changes minimum. Could you review it when you have a chance?

PR #10236, opened 2025-03-06
feat(control_validator)!: add overrun check

yuki-takagi-66:
## Description
add overrun stop point check 
[Screencast from 2025年03月04日 17時57分27秒.webm](https://github.com/user-attachments/assets/906ca190-f95b-45b7-a910-f01b8d20f6c7)

This feature is similar to the EMERGENCY status in the pid_longitudinal_controller, but it plays a different role.

## Related links
launch PR: https://github.com/autowarefoundation/autoware_launch/pull/1346

## How was this PR tested?
psim and tier4 scenario tests

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/control/autoware_control_validator/include/autoware/control_validator/control_validator.hpp
+++ b/control/autoware_control_validator/include/autoware/control_validator/control_validator.hpp
@@ -53,6 +53,7 @@ struct ValidationParams
   double rolling_back_velocity;
   double over_velocity_ratio;
   double over_velocity_offset;
+  double overrun_stop_point_dist;
   double nominal_latency_threshold;
 };
 
@@ -93,6 +94,9 @@ public:
     const Trajectory & predicted_trajectory, const Trajectory & reference_trajectory) const;
 
   void calc_velocity_deviation_status(
+    const Trajectory & reference_trajectory, const Odometry & kinematics);
+
+  void calc_stop_point_overrun_status(
     const Trajectory & reference_trajectory, const Odometry & kinematics);
 
 private:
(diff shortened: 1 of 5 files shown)


control/autoware_control_validator/include/autoware/control_validator/control_validator.hpp
```diff
@@ -86,6 +87,9 @@ class ControlValidator : public rclcpp::Node
   void calc_velocity_deviation_status(
     const Trajectory & reference_trajectory, const Odometry & kinematics);
 
+  void calc_stop_point_overrun_status(
```
kyoichi-sugahara, line 90: Please consider adding Doxygen-style comment as following

```suggestion
  /**
   * @brief Calculate whether the vehicle has overrun a stop point in the trajectory.
   * @param reference_trajectory Reference trajectory
   * @param kinematics Current vehicle odometry including position and velocity
   * @details Determines the distance to the nearest stop point and checks if the vehicle has
   *          passed beyond the stop point while still moving.
   */
  void calc_stop_point_overrun_status(
```

kyoichi-sugahara:
@yuki-takagi-66 
Please add an explanation to the function and update the readme at a minimum.
I don't have a strong opinion about the NaN issue, but the implementation could cause problems if `overrun_stop_point_dist` is mistakenly set to a negative value, so it would be LGTM if that point is covered in the explanation.

kyoichi-sugahara (approved):
LGTM!

PR #8927, opened 2024-09-20
refactor(map_loader)!: prefix package and namespace with autoware

a-maumau:
## Description
This PR will add the autoware prefix to `map_loader` package.

## Related links
- autoware-documentation: https://github.com/autowarefoundation/autoware-documentation/pull/617
- autoware-lanelet2-extension: https://github.com/autowarefoundation/autoware_lanelet2_extension/pull/31
- autoware_tools: https://github.com/autowarefoundation/autoware_tools/pull/122

**Parent Issue:**

- https://github.com/autowarefoundation/autoware/issues/4569

## How was this PR tested?
Checked with running the autoware sample-map.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

PR #10247, opened 2025-03-07
fix(autoware_map_loader): exec name renamed in 24652f8

ralwing:
## Description

exec has been renamed in [24652f834f73b13b3e4465](https://github.com/autowarefoundation/autoware.universe/commit/24652f834f73b13b3e4465bfc9ab8335cbe6cdcf#diff-87cd83c29c31953b416406e475004f933bad5ec63e534a5c1ca18e2747d3d534R36)
but wasn't renamed in default launcher.

## Related links

**Parent Issue:**
https://github.com/autowarefoundation/autoware.universe/pull/8927

- Link

https://github.com/autowarefoundation/autoware.universe/commit/24652f834f73b13b3e4465bfc9ab8335cbe6cdcf#diff-87cd83c29c31953b416406e475004f933bad5ec63e534a5c1ca18e2747d3d534R36

## Effects on system behavior

None.


map/autoware_map_loader/launch/lanelet2_map_loader.launch.xml
```diff
@@ -8,7 +8,7 @@
     <param from="$(var lanelet2_map_loader_param_path)" allow_substs="true"/>
   </node>
 
-  <node pkg="autoware_map_loader" exec="lanelet2_map_loader" name="lanelet2_map_loader" output="both">
+  <node pkg="autoware_map_loader" exec="autoware_lanelet2_map_loader" name="autoware_lanelet2_map_loader" output="both">
```
TaikiYamada4, line 11: Could you keep the node name of lanelet2_map_loader?
This is because [the default node name is NOT preferred to have the `autoware_` prefix](https://autowarefoundation.github.io/autoware-documentation/main/contributing/coding-guidelines/ros-nodes/class-design/).
(Yes there are some other nodes that have the `autoware_` prefix but we don't want `autoware_` all over the place when we `ros2 node list`.)

```suggestion
  <node pkg="autoware_map_loader" exec="autoware_lanelet2_map_loader" name="lanelet2_map_loader" output="both">
```

TaikiYamada4 (changes requested):
Thank you for the contribution!
I've confirmed that the independent launch of `lanelet2_map_loader` works well with this commit, but could you check the following comment?

map/autoware_map_loader/launch/lanelet2_map_loader.launch.xml
ralwing: I restored the previous name

TaikiYamada4 (approved):
LGTM :+1: 

PR #10252, opened 2025-03-10
fix(ndt_scan_matcher): fix the covariance calculation

anhnv3991:
## Description
When building a voxel grid for NDT scan matching, the covariance of points in a voxel is computed as follow (the covariances of (x, x), (x, z), (y, z) are similar)

image

which can be shortened to 
image
However, the formula above is not correct. The formula for _the unbiased estimation_ of _the covariance_ of a voxel should be
image
This PR replace the incorrect formula by the correct one. 

## Related links

[The PCL library ver 1.15.0-dev (line 326)](https://pointclouds.org/documentation/voxel__grid__covariance_8hpp_source.html)

## Notes for reviewers

None.

## Interface changes

None.

<!-- ⬇️🔴

#### Additions and removals

#### Modifications

#### Additions and removals

## Effects on system behavior

This PR may changes the behavior of the NDT scan matching.


SakodaShintaro:
I understood this change. It looks good.
Since there is a possibility that the previous form was used for some practical reason, I kicked the performance tests.

- [test1](https://evaluation.tier4.jp/evaluation/reports/315b80fa-4cbd-5ac1-bf82-7b235ac1a2e7?project_id=autoware_dev) (TIERIV INTERNAL)
- [test2](https://evaluation.tier4.jp/evaluation/reports/3e6b6ba5-4419-5ed5-8a29-b1afec00b54b?project_id=prd_jt) (TIERIV INTERNAL)

If there are no regressions, I think it's okay to merge.

SakodaShintaro:
There is some randomness in the testing so there may be some fluctuations, but I don't think it's a problem.

compare_result

localization/autoware_ndt_scan_matcher/src/ndt_omp/multi_voxel_grid_covariance_omp_impl.hpp
```diff
@@ -434,9 +434,7 @@ void pclomp::MultiVoxelGridCovariance<PointT>::computeLeafParams(
   Leaf & leaf) const
 {
   // Single pass covariance calculation
-  leaf.cov_ = (leaf.cov_ - 2 * (pt_sum * leaf.mean_.transpose())) / leaf.nr_points_ +
-              leaf.mean_ * leaf.mean_.transpose();
-  leaf.cov_ *= (leaf.nr_points_ - 1.0) / leaf.nr_points_;
+  leaf.cov_ = (leaf.cov_ - pt_sum * leaf.mean_.transpose()) / (leaf.nr_points_ - 1);
```
SakodaShintaro, line 437: It seems like division by 0 will occur when `leaf.nr_points_ == 1`, but is that not possible?

anhnv3991:
@SakodaShintaro 
> There is some randomness in the testing so there may be some fluctuations, but I don't think it's a problem.
> 
> compare_result

Are there anyway to run this test offline on local PCs? I want to do some optimization for NDT but the localization stuff by pilot-auto.x2 or autoware seems troublesome.

localization/autoware_ndt_scan_matcher/src/ndt_omp/multi_voxel_grid_covariance_omp_impl.hpp
SakodaShintaro: [Note] The value is set in the following line:

https://github.com/autowarefoundation/autoware.universe/blob/eebb4a5eaa46c8b0b5a8ade256ef04bcfae14733/localization/autoware_ndt_scan_matcher/include/autoware/ndt_scan_matcher/ndt_omp/multi_voxel_grid_covariance_omp.h#L227

SakodaShintaro (approved):

PR #10258, opened 2025-03-11
feat(planning_validator): add yaw deviation metric

maxime-clem:
## Description

Add a metric to the `planning_validator` to signal when the yaw difference between the vehicle and the trajectory is higher than a threshold.
Required launch PR to add the threshold parameter: https://github.com/autowarefoundation/autoware_launch/pull/1356

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

Psim

Video for the default threshold of 90 degrees:

https://github.com/user-attachments/assets/1bdcd05e-4772-4d79-80bb-75485c25b860

Reverse trajectory case:

https://github.com/user-attachments/assets/3deeaf8d-d234-4af1-9227-88218ca0ed48

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


kyoichi-sugahara:
@maxime-clem 
can you add readme explanation?

kyoichi-sugahara:
@maxime-clem 
won't there be issues during reverse driving if comparing the ego vehicle's orientation with the planning output?

kyoichi-sugahara (approved):

kosuke55 (approved):
LGTM!

PR #10259, opened 2025-03-11
feat: use mrm in all operation mode state

TetsuKawa:
## Description
This PR changed that  mrm can be used in all operation mode state,

## Related links

**Private Links:**

- [TIER IV internal link](https://star4.slack.com/archives/C024FRX4UNL/p1741689609280229)

## How was this PR tested?

1. prepare for [autoware](https://github.com/autowarefoundation/autoware/tree/c487b9853e1b4dec5262ab12616ce9d22c54e9a9)
2. cherry-pick this PR commit to autoware.universe
3. conlcon build
4. launch psim
5. Set Autonomous and execute `ros2 param set /dummy_diag_publisher "localization: ekf_localizer".status "Error"`
6. Comfirm that transition to MRM

[Screencast from 2025年03月12日 00時32分57秒.webm](https://github.com/user-attachments/assets/4c9cc84d-24dd-478e-a661-a657eeed552a)

7. Set to autoware_launch/system/diagnostics/autoware-main.yaml
```
units:
  - path: /autoware/modes/local
    type: and
    list:
      - { type: link, link: /autoware/vehicle }
      - { type: link, link: /autoware/system }
      - { type: link, link: /autoware/control/local }
      - { type: link, link: /autoware/localization } # <-- add
```

8. launch psim
9. Set Local and execute `ros2 param set /dummy_diag_publisher "localization: ekf_localizer".status "Error"`
10. Comfirm that transition to MRM

[Screencast from 2025年03月12日 00時40分59秒.webm](https://github.com/user-attachments/assets/4e38263f-bc4e-4500-a276-ae83620e5d6d)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

When current operation mode state (all) is not available, Autoware executes MRM.

### DIFF
--- a/system/autoware_mrm_handler/src/mrm_handler/mrm_handler_core.cpp
+++ b/system/autoware_mrm_handler/src/mrm_handler/mrm_handler_core.cpp
@@ -89,26 +89,26 @@ void MrmHandler::onOperationModeAvailability(
   stamp_operation_mode_availability_ = this->now();
   operation_mode_availability_ = msg;
   const bool skip_emergency_holding_check =
-    !param_.use_emergency_holding || is_emergency_holding_ || !isOperationModeAutonomous();
+    !param_.use_emergency_holding || is_emergency_holding_;
 
   if (skip_emergency_holding_check) {
     return;
   }
 
-  if (msg->autonomous) {
-    stamp_autonomous_become_unavailable_.reset();
+  if (isAvailableCurrentOperationMode()) {
+    stamp_current_operation_mode_become_unavailable_.reset();
     return;
   }
 
   // If no timestamp is available, the ego autonomous mode just became unavailable and the current
   // time is recorded.
-  stamp_autonomous_become_unavailable_ = (!stamp_autonomous_become_unavailable_.has_value())
+  stamp_current_operation_mode_become_unavailable_ = (!stamp_current_operation_mode_become_unavailable_.has_value())
                                            ? this->now()
-                                           : stamp_autonomous_become_unavailable_;
+                                           : stamp_current_operation_mode_become_unavailable_;
 
   // Check if autonomous mode unavailable time is larger than timeout threshold.
   const auto emergency_duration =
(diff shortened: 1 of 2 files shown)


system/autoware_mrm_handler/src/mrm_handler/mrm_handler_core.cpp
```diff
@@ -586,6 +577,33 @@ bool MrmHandler::isArrivedAtGoal()
   return state->mode == OperationModeState::STOP;
 }
 
+bool MrmHandler::isAvailableCurrentOperationMode()
```
veqcc, line 580: `isHoge` like function name sounds begin `const`, so updating global variable inside is strange.

TetsuKawa: The reason why` isAvailableCurrentOperationMode() `cannot be marked as const is that it calls `getCurrentOperationMode()`, which internally calls` .take_data() `on a subscriber.
This `.take_data()` simply fetches the latest message from the internal topic buffer and does not modify any member variables.
Therefore, while the function is technically non-const due to` .take_data()`, it does not have side effects in the conventional sense.

TetsuKawa: If the use of isHoge-style function names is considered inappropriate due to the lack of const,
then I suggest creating a separate style-fix PR to rename all of the following functions, since none of them are marked const either:

```
bool isStopped();
bool isEmergency();
bool isControlModeAutonomous();
bool isPullOverStatusAvailable();
bool isComfortableStopStatusAvailable();
bool isEmergencyStopStatusAvailable();
bool isArrivedAtGoal();
bool isAvailableCurrentOperationMode();
```

veqcc: ok

PR #10262, opened 2025-03-12
fix(autoware_behavior_path_planner_common): add explicit test dependency

xmfcx:
## Description

Similar to:
- https://github.com/autowarefoundation/autoware.universe/pull/10261

Required by:
- https://github.com/autowarefoundation/autoware.core/pull/201#issuecomment-2716840238

## How was this PR tested?

Build.


mitsudome-r (approved):

xmfcx:
Didn't add it to `test_${PROJECT_NAME}_turn_signal` because it doesn't use and depend on `autoware_test_utils`.

PR #10269, opened 2025-03-13
fix(obstacle_stop): use max_lat_margin_against_unknown only for predicted object

takayuki5168:
## Description

Currently, `max_lat_margin_against_unknown` is used for the point cloud.
For the users who use not the predicted object but the point cloud for the obstacle stop and slow_down, it's a bit confusing that not `max_lat_margin` but `max_lat_margin_against_unknown` is used.

Therefore, this PR modifies as follows
- use `max_lat_margin` for the point cloud
- rename `max_lat_margin_against_unknown` to `max_lat_margin_against_predicted_object_unknown`.
## How was this PR tested?
planning simulator worked well
## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


PanConChicharron (approved):
LGTM

PR #10281, opened 2025-03-17
fix(autoware_crosswalk_traffic_light_estimator): add process that guard access to empty elements

k-hazama-esol:
## Description
There is a problem that the crosswalk traffic light estimator drops out when a pedestrian signal is received with V2I signal coordination.

This problem is caused by a process in crosswalk_traffic_light_estimator that accesses elements without checking for the presence of signal recognition results.
Accessing an element without checking its value even when it is empty causes a segmentation fault.

This PR solves the problem by adding a process that prevents the result of signal recognition from being referenced in an empty state.

## Related links

**Private Links:**

- [TIER IV internal link](https://tier4.atlassian.net/browse/RT0-34969) 
- [Evaluator](https://evaluation.ci.tier4.jp/evaluation/reports/c913655e-2855-5a2b-ae65-8a983ce3cde3?project_id=prd_jt)

## How was this PR tested?

- PSIM

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_crosswalk_traffic_light_estimator/src/node.cpp
+++ b/perception/autoware_crosswalk_traffic_light_estimator/src/node.cpp
@@ -303,7 +303,11 @@ void CrosswalkTrafficLightEstimatorNode::setCrosswalkTrafficSignal(
         output_traffic_signal_element.color = color;
         output_traffic_signal_element.shape = TrafficSignalElement::CIRCLE;
         output_traffic_signal_element.confidence = 1.0;
-        output.traffic_light_groups[idx].elements[0] = output_traffic_signal_element;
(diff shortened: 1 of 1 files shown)


TomohitoAndo:
@MasatoSaeki @tzhong518 @YoshiRi 
Could you review this PR as codeowner? :pray:

MasatoSaeki:
@k-hazama-esol @TomohitoAndo 
I am sorry for delay, will just review

MasatoSaeki (commented):
LGTM, But I test on Evaluator now.
I will appove if this pass.

MasatoSaeki (approved):
I cheched passed Evaluator. Thank you.

PR #10284, opened 2025-03-17
feat(lidar_centerpoint): add diagnostics for processing time

a-maumau:
## Description

This PR adds diagnostics for processing time.

Consecutive delays will be checked in the timer callback.
`diagnosticsTimerCallback` will check the timestamp of last normally processed one and current timestamp, and if the difference of two timestamp is larger than `max_acceptable_consecutive_delay_ms`, it will determine as an error state.

If the new processing time is below the `max_allowed_processing_time_ms`, it will update the timestamp of last normally processed case.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

Run rqt_runtime_monitor by
```sh
# optional
# ros2 run autoware_diagnostic_graph_utils converter_node

```
https://github.com/user-attachments/assets/4b816d36-ff8d-4777-bb32-395350a1c338

(above video was created in sample-rosbag with very small `max_allowed_processing_time_ms` == 2ms for demonstration)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


perception/autoware_lidar_centerpoint/include/autoware/lidar_centerpoint/node.hpp
```diff
@@ -56,11 +58,19 @@ class LidarCenterPointNode : public rclcpp::Node
   bool has_variance_{false};
   bool has_twist_{false};
 
+  bool is_processing_delayed_{false};
+  double max_allowed_processing_time_ms_;
+  double max_acceptable_consecutive_delay_ms_;
+  // set optional to avoid sending the error diagnostics before node start processing
+  std::optional<rclcpp::Time> last_normal_processing_timestamp_;
```
technolojin, line 65:
```suggestion
  std::optional<rclcpp::Time> last_in_time_processing_timestamp_;
```

perception/autoware_lidar_centerpoint/config/centerpoint_diagnostics.param.yaml
technolojin, line 1: This configuration is for general node parameters. Therefore, it is proper decision to make a separated parameter file.

Suggestion:
`centerpoint_diagnostics.param.yaml` -> `centerpoint_node.param.yaml` or `centerpoint_common.param.yaml`

perception/autoware_lidar_centerpoint/config/centerpoint_diagnostics.param.yaml
```diff
@@ -0,0 +1,10 @@
+/**:
+  ros__parameters:
+    # When the processing time exceeds the max_allowed_processing_time,
+    # a warning will be triggered.
+    max_allowed_processing_time_ms: 200.0 # in milliseconds
+    # If the warning is triggered and if a timestamp of last normal processing is more than
+    # max_acceptable_consecutive_delay_ms, an error diagnostic message will be sent.
+    max_acceptable_consecutive_delay_ms: 1000.0
+    # interval of diagnostic callback in milliseconds
+    validation_callback_interval_ms: 100 # in milliseconds, in integer
```
technolojin, line 10:
```suggestion
    diagnostics:
      max_allowed_processing_time_ms: 200.0 # in milliseconds
      # If the warning is triggered and if a timestamp of last normal processing is more than
      # max_acceptable_consecutive_delay_ms, an error diagnostic message will be sent.
      max_acceptable_consecutive_delay_ms: 1000.0
      # interval of diagnostic callback in milliseconds
      validation_callback_interval_ms: 100 # in milliseconds, in integer
```

technolojin (changes requested):
overall, the implementation looks good.

I suggested some renaming.

a-maumau:
@technolojin I agreed with your suggestions and applied in 934070c to db4d837 .

a-maumau:
I will create a different PR for the launcher related things.

technolojin (approved):
LGTM

knzo25:
@a-maumau 
Can you check why the CI/CD does not pass?

a-maumau:
I am not sure with previous status, but currently it seems it has some issue with docker pull (code: TOOMANYREQUESTS)

technolojin:
clang-tidy was failed
```
private field 'is_processing_delayed_' is not used [clang-diagnostic-unused-private-field][0m
  bool is_processing_delayed_{false};
```

knzo25 (approved):
LGTM

technolojin (changes requested):
I would like to request to implement to two types of diagnostics

1. number of voxel limit (already implemented)
2. processing time diags

when those are implemented, please let me know.

a-maumau:
@technolojin 
I done some refactor in ff8185d to reorganize and merge the diag. to publish in same interface.

Current implementation may have a little bit of downside.
Unlike in map_based_prediction: https://github.com/autowarefoundation/autoware_universe/blob/6fca99a2194c793079468e2fb982d48e6ba38ef4/perception/autoware_map_based_prediction/src/map_based_prediction_node.cpp#L736 , I implemented the diagnosis for consecutive delay in timer callback to ensure some temporal resolution, and this give some unliked output of diagnostics which processing time is set `0.0` since we can not compute in the timer callback.

I wonder this is an issue for later processings like analysis or logging.

sample_diag

a-maumau:
Sorry for my non concrete impelmentation that making many fix after the first review.

I change the code to use `diagnostic_updater` to diagnose the processing time status.
Since node's callback is currently set in mutually exclusive callback, this should make the timer callback suspended till the `pointCloudCallback` finishes, so the "ensure some temporal resolution" things I mentioned does not make much sense.

I made the code more simple.

perception/autoware_lidar_centerpoint/src/node.cpp
```diff
@@ -199,9 +217,73 @@ void LidarCenterPointNode::pointCloudCallback(
...
+      diag_level = diagnostic_msgs::msg::DiagnosticStatus::ERROR;
+    } else {
+      stat.add("is_consecutive_processing_delay_in_range", true);
+    }
+    stat.add("consecutive_processing_delay_ms", delayed_state_duration);
+  } else {
+    stat.add("is_processing_time_ms_in_expected_range", true);
+    stat.add("processing_time_ms", 0.0);
+    stat.add("is_consecutive_processing_delay_in_range", true);
+    stat.add("consecutive_processing_delay_ms", 0.0);
```
technolojin, line 279: May not needed, but acceptable.

a-maumau: I was wondering if it is OK to change fields depending on the state.

technolojin: In the very beginning, the field is empty (due to the diagnostic_updater implementation)

technolojin (approved):
LGTM

perception/autoware_lidar_centerpoint/src/node.cpp
```diff
@@ -199,9 +217,73 @@ void LidarCenterPointNode::pointCloudCallback(
...
+// If the node is consistently delayed, publish an error diagnostic message
+void LidarCenterPointNode::diagnoseProcessingTime(
+  diagnostic_updater::DiagnosticStatusWrapper & stat)
+{
+  const rclcpp::Time timestamp_now = this->get_clock()->now();
+  diagnostic_msgs::msg::DiagnosticStatus::_level_type diag_level =
+    diagnostic_msgs::msg::DiagnosticStatus::OK;
+  std::stringstream message{"OK"};
+
+  // check node has inferenced or not
```
technolojin, line 235:
```suggestion
  // Check if the node has performed inference
```

technolojin:
launcher config https://github.com/autowarefoundation/autoware_launch/pull/1377

PR #10203, opened 2025-02-28
refactor(multi_object_tracker): internal message driven process

technolojin:
## Description
refactor tracker process by internally defined structure messages

- integrated classes 
  - DynamicObject : object related structure. shorten overall codes
    - remove size variables from tracker and integrate to DynamicObject member
  - InputChannel : integrate channel related configuration
  - AssociatorConfig : association configuration is collected as a struct
- simplify argument passses
  - channel size `channel_size` is removed
  - position of the ego vehicle `self_transform` passed to all of the trackers is removed
  - tracker variable `anchor_point` for vehicle tracking is integrated to DynamicObject class
- Shape related process is moved to input manager
  - input channel processes `getNearestCornerOrSurface` for all inputs. it is moved from the vehicle tracker

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

[TIER IV INTERNAL](https://evaluation.tier4.jp/evaluation/reports/59072fe7-5b0c-5e62-8524-c3670e7723c2?project_id=prd_jt)
123/123 passed

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

PR #10286, opened 2025-03-17
fix(autoware_multi_object_tracker): unknown object orientation

technolojin:
## Description

Object shape process for polygon shape (generally for unknown object) should be processed after the object is transformed to world coordinate. 
This PR fixes [the previous PR ](https://github.com/autowarefoundation/autoware.universe/pull/10203)the polygon process before the transform.

BUG

https://github.com/user-attachments/assets/91b5d8c6-59fc-4524-9d11-b3674543f87d

Additionally, fix a bug to fit bounding box to the polygon.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_multi_object_tracker/lib/object_model/shapes.cpp
+++ b/perception/autoware_multi_object_tracker/lib/object_model/shapes.cpp
@@ -26,6 +26,7 @@
 #include <tf2/utils.h>
 
 #include <algorithm>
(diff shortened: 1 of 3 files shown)


YoshiRi (approved):
LGTM

PR #9943, opened 2025-01-16
feat(behavior_velocity_planner)!: remove velocity_factor completely

soblin:
## Description

wrap up this work: https://github.com/autowarefoundation/autoware.universe/pull/9927

- https://github.com/tier4/autoware.universe/pull/1767
- https://github.com/tier4/autoware.universe/pull/1766

## Related links

https://github.com/autowarefoundation/autoware.universe/pull/9927

**Parent Issue:**

- Link

## How was this PR tested?

- https://evaluation.tier4.jp/evaluation/reports/e824150d-92fd-5df5-bd6c-cd9e24411af3?project_id=prd_jt

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

PR #10292, opened 2025-03-18
fix(behavior_velocity_planner): planning factor integration

satoshi-ota:
## Description

All behavior veloicty modules set redundant control point if the module made stop decision.

```patch
diff --git a/planning/behavior_velocity_planner/autoware_behavior_velocity_walkway_module/src/scene_walkway.cpp b/planning/behavior_velocity_planner/autoware_behavior_velocity_walkway_module/src/scene_walkway.cpp
index 4d8651d05c..827655daa2 100644
--- a/planning/behavior_velocity_planner/autoware_behavior_velocity_walkway_module/src/scene_walkway.cpp
+++ b/planning/behavior_velocity_planner/autoware_behavior_velocity_walkway_module/src/scene_walkway.cpp
@@ -123,7 +123,7 @@ bool WalkwayModule::modifyPathVelocity(PathWithLaneId * path)
 
     /* get stop point and stop factor */
     planning_factor_interface_->add(
-      path->points, planner_data_->current_odometry->pose, stop_pose.value(), stop_pose.value(),
+      path->points, planner_data_->current_odometry->pose, stop_pose.value(),
       autoware_internal_planning_msgs::msg::PlanningFactor::STOP,
       autoware_internal_planning_msgs::msg::SafetyFactorArray{}, true /*is_driving_forward*/,
       0.0 /*velocity*/, 0.0 /*shift distance*/, "walkway_stop");
```

## Related links

**Parent Issue:**

- https://github.com/autowarefoundation/autoware.universe/pull/9943

## How was this PR tested?

I confirmed modules output planning factor properly after fixing.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_velocity_planner/autoware_behavior_velocity_crosswalk_module/src/scene_crosswalk.cpp
+++ b/planning/behavior_velocity_planner/autoware_behavior_velocity_crosswalk_module/src/scene_crosswalk.cpp
@@ -881,7 +881,7 @@ void CrosswalkModule::applySlowDown(
   }
   if (slowdown_pose)
     planning_factor_interface_->add(
-      output.points, planner_data_->current_odometry->pose, *slowdown_pose, *slowdown_pose,
+      output.points, planner_data_->current_odometry->pose, *slowdown_pose,
       autoware_internal_planning_msgs::msg::PlanningFactor::SLOW_DOWN,
       autoware_internal_planning_msgs::msg::SafetyFactorArray{}, true /*is_driving_forward*/,
       safety_slow_down_speed, 0.0 /*shift distance*/, "crosswalk_safety_slowdown_for_approaching");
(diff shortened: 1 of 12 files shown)


planning/behavior_velocity_planner/autoware_behavior_velocity_crosswalk_module/src/scene_crosswalk.cpp
```diff
@@ -881,7 +881,7 @@ void CrosswalkModule::applySlowDown(
   }
   if (slowdown_pose)
     planning_factor_interface_->add(
-      output.points, planner_data_->current_odometry->pose, *slowdown_pose, *slowdown_pose,
+      output.points, planner_data_->current_odometry->pose, *slowdown_pose,
```
soblin, line 884: Maybe we have start & end point for slowdown.

soblin (approved):
Okay except for crosswalk

planning/behavior_velocity_planner/autoware_behavior_velocity_crosswalk_module/src/scene_crosswalk.cpp
satoshi-ota: As far as I read the code, crosswalk module inserts one slow down start point, and all trajectory points are overwritten with same slow down velocity. Then, crosswalk has only one control point as well.

Issue #10290, opened 2025-03-18
Compare_map_segmentation degradation at high speed or during turns


badai-nguyen:
### Checklist

- [x] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [x] I've searched other issues and no duplicate issues were found.
- [x] I'm convinced that this is not my fault but a bug.

### Description

- `Compare_map_segmentation`'s result is unstable when ego-vehicle moves at high speed or turns. 
- 
https://github.com/user-attachments/assets/cd194bc2-db09-4187-996c-cbdb144c6ed1

- A large static objects' pointcloud remain even after `compare_map_segmentation`

![Image](https://github.com/user-attachments/assets/61c4e910-8b05-4c5a-a9a6-f137bbe67bf2)

https://github.com/user-attachments/assets/184ae0de-0226-46bd-a72e-c08092304565

  - White point: ground_segmentation result
  - Pink: compare_map_segmentation result
  - Green: downsampled map 

### Expected behavior

- Compare_map_segmentation should remain stable regardless ego-vehicle's speed
- Large static objects will be also removed successfully at high speed.

### Actual behavior

- Compare_map_segmentation is unstable when ego-vehicle moves fast or turn.
- Even large static objects' pointcloud remain at high speed moving.

### Steps to reproduce

- Using `logging_simulator`
- Test with high-speed moving vehicle rosbag 
- Rendering `/perception/object_recognition/detection/pointcloud_map_filtered/pointcloud` topic to check `compare_map_segmentation` result

PR #10299, opened 2025-03-19
fix(voxel_based_compare_map): temporary fix pointcloud transform lookup 

badai-nguyen:
## Description
- This temporary override input_indices_callback to change lookup tf from latest to pointcloud's header stamp.
- To fix this issue: https://github.com/autowarefoundation/autoware.universe/issues/10290
- This could remove when lookup_time could be select when calling `transform_pointcloud` from [autoware_utils](https://github.com/autowarefoundation/autoware_utils/blob/182c967b189f786faf0f01a08dd7cd2da7999f2f/autoware_utils/include/autoware_utils/ros/managed_transform_buffer.hpp#L211-L212) or new [ManagedTransformBuffer](https://github.com/autowarefoundation/ManagedTransformBuffer)

## Related links

- [TIER IV internal link](https://tier4.atlassian.net/browse/RT0-35864)

**Parent Issue:**

- Link: https://github.com/autowarefoundation/autoware.universe/issues/10290

## How was this PR tested?

- By logging_simulator on internal rosbag

| Before | After (a large number pc were filtered)|
| --- | ---|
|Screenshot from 2025-03-18 23-19-44_before| Screenshot from 2025-03-18 23-13-58_after|
|Screenshot from 2025-03-19 11-30-35_before| Screenshot from 2025-03-19 11-30-43_after|
|[Screencast from 2025年03月19日 08時33分35秒_before.webm](https://github.com/user-attachments/assets/18f8d138-decb-4ed1-8a89-44def2663eb9)| [Screencast from 2025年03月19日 08時33分35秒_before.webm](https://github.com/user-attachments/assets/30d7ee6a-ce26-44ed-9d73-cae5ab78e900) |

- By internal evaluator (WIP):

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


mojomex:
@badai-nguyen There seems to be some previously non-existent misalignment in some regions now. It looks like those parts of the pointcloud have an angular offset to the map. 

| Before | After |
|-|-|
| image | image |

The code definitely looks correct now, I wonder if there are other bugs or delays in the inputs causing this, or if this is a visualization issue (delay between pointclouds).

perception/autoware_compare_map_segmentation/src/voxel_based_compare_map_filter/node.cpp
```diff
@@ -62,6 +71,60 @@ VoxelBasedCompareMapFilterComponent::VoxelBasedCompareMapFilterComponent(
...
+      cloud->header.frame_id.c_str(), tf_input_frame_.c_str());
+    // Save the original frame ID
+    // Convert the cloud into the different frame
+    PointCloud2 cloud_transformed;
+
+    try {
+      // Lookup the transform from input frame to "map"
+      geometry_msgs::msg::TransformStamped transform_stamped = tf_buffer_.lookupTransform(
+        tf_input_frame_, tf_input_orig_frame_, rclcpp::Time(cloud->header.stamp),
+        rclcpp::Duration::from_seconds(1.0));
```
mojomex, line 106: This callback is running at `10 Hz` (right?), so this timeout value would delay output up to 10 whole periods.
Depending on other runtime properties of this node, it should probably be much lower (in the order of `10 ms`).

From my understanding, even `0` could be a good choice, guaranteeing that this module never stalls when a transform is late.

perception/autoware_compare_map_segmentation/src/voxel_based_compare_map_filter/node.cpp
```diff
@@ -62,6 +71,60 @@ VoxelBasedCompareMapFilterComponent::VoxelBasedCompareMapFilterComponent(
...
+
+      // Transform the point cloud
+      tf2::doTransform(*cloud, cloud_transformed, transform_stamped);
+      cloud_transformed.header.frame_id = tf_input_frame_;  // Update frame ID to "map"
+      cloud_tf = std::make_shared<PointCloud2>(cloud_transformed);
+    } catch (tf2::TransformException & ex) {
+      RCLCPP_WARN(this->get_logger(), "Could not transform point cloud: %s", ex.what());
+      cloud_tf = cloud;  // Fallback to original data
```
mojomex, line 114: This callback runs at `10 Hz`, so the `_THROTTLE` family of log functions should be used (e.g. log once only every 5s).

mojomex (changes requested):
Thank you for the PR. I have a few timing-related comments :bow:

YoshiRi (approved):
LGTM and it worked within my v4.1.0 environment

YoshiRi:
Pointcloud number decreased and it looks better now. But there seems to be some part is not removed yet.

|Before|After|
|--|--|
|Screenshot from 2025-03-18 14-00-42|image|

perception/autoware_compare_map_segmentation/src/voxel_based_compare_map_filter/node.cpp
badai-nguyen: > This callback is running at `10 Hz` (right?), so this timeout value would delay output up to 10 whole periods. Depending on other runtime properties of this node, it should probably be much lower (in the order of `10 ms`).
> 
> From my understanding, even `0` could be a good choice, guaranteeing that this module never stalls when a transform is late.

@mojomex 
Thanks for your comment, 
> This callback is running at 10 Hz 

yes

However, the input topic of `compare_map_filter` might have a quite large delay (including 100ms of sensing + crop_box_filter + ground_segmentation + voxel_downsample_filter processing time) which caused the gap between latest tf and the tf at the time pointcloud was recorded. In the result, it will cause unmatched of lidar pointcloud and map pointcloud when ego moves fast.

badai-nguyen:
> There seems to be some previously non-existent misalignment in some regions now. It looks like those parts of the pointcloud have an angular offset to the map.

@mojomex 
Thanks for your insight review.
I think transform back output of compare_map_filter from map -> baselink still use the [default convert with latest tf ](https://github.com/autowarefoundation/autoware.universe/blob/ece38099fe94d959e98ce582bc04be49c6c1f12b/sensing/autoware_pointcloud_preprocessor/src/filter.cpp#L321-L363)
Which might cause the misalignment. I will fix it.

perception/autoware_compare_map_segmentation/src/voxel_based_compare_map_filter/node.cpp
mojomex: @badai-nguyen Thank you for your response.

I agree that the delay can be long, so to get accurate transforms, `cloud->header.stamp` has to be used as the lookup timestamp.

However, The base_link -> map transform should be available way earlier than the pointcloud triggering the callback (like you mentioned, the pointcloud has to go through all of preprocessor first - the base_link/map transform does not).

Thus, I would expect the tf to be available already (able to `lookupTransform` instantly) when the pointcloud arrives. Thus, the `timeout` (not `timestamp`) could be set to `0` or a very low millisecond value.

badai-nguyen: @mojomex 
>  the timeout (not timestamp)

Ah, I got your point. I agree with that. Sorry for my misunderstanding.

badai-nguyen:
> @badai-nguyen There seems to be some previously non-existent misalignment in some regions now. It looks like those parts of the pointcloud have an angular offset to the map.
> 
> Before	After
> image	image
> I wonder if it is an issue with this Pr, with other bugs or delays in the inputs, or if this is a visualization issue (delay between pointclouds).

@mojomex 
I fixed this misalignement at 
The result after update as here:
image
video: 
[Screencast from 2025年03月20日 00時17分53秒_after_fix_back_tf.webm](https://github.com/user-attachments/assets/9042fadd-4029-4594-bc44-31f83633fefb)

perception/autoware_compare_map_segmentation/src/voxel_based_compare_map_filter/node.cpp
badai-nguyen: I fixed at https://github.com/autowarefoundation/autoware_universe/pull/10299/commits/106514e67c51bf1578280579fab22b582e0f34dd

perception/autoware_compare_map_segmentation/src/voxel_based_compare_map_filter/node.cpp
badai-nguyen: @mojomex fixed here https://github.com/autowarefoundation/autoware_universe/pull/10299/commits/106514e67c51bf1578280579fab22b582e0f34dd

perception/autoware_compare_map_segmentation/CMakeLists.txt
```diff
@@ -22,11 +22,11 @@ include_directories(
...
 add_library(${PROJECT_NAME} SHARED
-  src/distance_based_compare_map_filter/node.cpp
-  src/voxel_based_approximate_compare_map_filter/node.cpp
+  # src/distance_based_compare_map_filter/node.cpp
+  # src/voxel_based_approximate_compare_map_filter/node.cpp
   src/voxel_based_compare_map_filter/node.cpp
-  src/voxel_distance_based_compare_map_filter/node.cpp
-  src/compare_elevation_map_filter/node.cpp
+  # src/voxel_distance_based_compare_map_filter/node.cpp
+  # src/compare_elevation_map_filter/node.cpp
```
yukkysaito, line 29: This seems like a very temporary workaround. I think it would be better to make it unnecessary to comment out the above before merging. What do you think?

yukkysaito (changes requested):

perception/autoware_compare_map_segmentation/CMakeLists.txt
yukkysaito: Or do you mean that the commented-out feature is intended to be deprecated?

badai-nguyen: @yukkysaito I'm sorry, it was my typo. I fixed https://github.com/autowarefoundation/autoware_universe/pull/10299/commits/89523c90e8b95a7334d1a9cd533c7f9627026dfe
By the way, we also started a discussion about deprecation of unused filters. I will open the disucssion later.

yukkysaito (approved):

badai-nguyen:
> Thank you for the PR. I have a few timing-related comments 🙇

@mojomex 
Do you have future comment for this PR? Since you have requested change, could I ask your review again?

mojomex (approved):
Sorry, missed sending the approval :bow: 
LGTM!

PR #10308, opened 2025-03-19
fix(lane_change): set safety factor properly

satoshi-ota:
## Description

I fixed the logic so that the lane change module can output safety factor properly.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

I confirmed that the lane change module output safety factor whose flag `is_safe` was `false` in following situation.

Screenshot from 2025-03-19 19-56-07

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


mkquda (approved):
LGTM.

zulfaqar-azmi-t4 (approved):

PR #10309, opened 2025-03-19
fix(autoware_tensorrt_common): invalid engine model file

ralwing:
Signed-off-by: Grzegorz Głowacki <gglowacki@autonomous-systems.pl>

## Description
After a CUDA update and the creation of a new `.engine` model, the module can print an error like this:
```
Error code: 1, Description: IRuntime::deserializeCudaEngine: Error Code 1: Serialization (Serialization assertion stdVersionRead == kSERIALIZATION_VERSION failed. Version tag does not match. Note: Current Version: 237, Serialized Engine Version: 236)
```

This error stops the node from loading, and it doesn't resolve itself. This PR fixes this by deleting the incompatible `.engine` file. The file will then be automatically regenerated when the node starts again. This makes software updates easier for production."

## Effects on system behavior

Removes the incompatible engine model.

### DIFF
--- a/perception/autoware_tensorrt_common/src/tensorrt_common.cpp
+++ b/perception/autoware_tensorrt_common/src/tensorrt_common.cpp
@@ -36,6 +36,45 @@ namespace autoware
 namespace tensorrt_common
 {
 
+class TrtErrorRecorder : public nvinfer1::IErrorRecorder
+{
+public:
+  struct Error
+  {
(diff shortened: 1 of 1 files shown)


ralwing:
Thank You @amadeuszsz . Your solution is more convenient 
Replaced with https://github.com/autowarefoundation/autoware_universe/pull/10320

Issue #10311, opened 2025-03-19
Steering Command Converter for CARLA Set Up Incorrectly


stevenbrills:
### Checklist

- [x] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [x] I've searched other issues and no duplicate issues were found.
- [x] I'm convinced that this is not my fault but a bug.

### Description

The config file for the raw_vehicle_command_converter in the current autoware_carla_interface package is set to NOT convert steering commands from Autoware.

![Image](https://github.com/user-attachments/assets/a9e02de4-d9c5-471a-abbc-8934c70349dc)

However, the interface in carla_ros.py uses the steering command as a steering actuation command itself and not a steered tire angle. See the line where out_cmd.steer is being set in the image below. The existing conversion uses the "in_cmd.actuation.steer_cmd" which is the steered wheel angle (since raw vehicle command converter is not converting the tire angle to actuator signals) and multiplies it by the steer ratio and the maximum wheel angle of the vehicle. Then it assigns it to the out_cmd.steer control command which takes in actuator values and not steered tire angles.

![Image](https://github.com/user-attachments/assets/2bb2b01a-da48-49d3-99f8-50987cb7221a)

This apparently works for small tire angles that are used as steering actuation signals directly but breaks down for large angles, particularly during high curvature free-space maneuvers. In any case this is incorrect semantically.

### Expected behavior

The input steering tire angle should be converted to valid steering actuation signals that lie between -1 and 1 and can be consumed by CARLA.

### Actual behavior

The vehicle is unable to track high curvature paths when interfacing with CARLA.

### Steps to reproduce

1. Load a CARLA town with a valid parking lot
2. Launch the stack with CARLA as the sim option and set a goal in a parking lot that requires sharp maneuvers.
3. The is_steering_converged flag should stop the vehicle from continuing mid-way.

### Versions

- OS: Ubuntu 22.04
- ROS2: Humble
- Autoware: main

### Possible causes

Incorrect conversion of the steering tire angle to actuator signals. However, simply setting the conversion to be true is insufficient as CARLA uses a constant high steering rate to simulate the wheel angle movement and leads to oscillations using the current steering map. The out_cmd.steer_cmd should simply be set as the input actuator command times the max_steer_ratio.

maxime-clem:
[...]

PR #10312, opened 2025-03-19
fix(autoware_carla_interface): updated the CARLA interface to convert steer angle to actuator command and to model steering with first order dynamics and updated the "steer_map.csv" file

stevenbrills:
## Description
This PR is related to the issues raised in #10311. It adds a first order dynamics model for steering in CARLA and includes an updated steer_map.csv file that permits the correct conversion of the steering tire angle and steering rate into actuator signals for use by CARLA.

## Related links

**Parent Issue:**
- #10311

## How was this PR tested?
This update was tested in sim with both the default Prius and a Lincoln MKZ. The vehicle is able to better track high curvature paths with this update. Without the updated steering model, a lot of steering chatter was observed due to near instantaneous front tire steering by CARLA which caused the lateral controller to oscillate when using the steered tire angle as feedback. The first order dynamics adds realism to the steering motion.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

Better tracking of high curvature paths for simulations done in CARLA.


maxime-clem (approved):
Thank you for the PR. The lateral error was very noticeable and it is now significantly improved.

| Current | With this PR |
|--------|--------|
|  current |pr  |

PR #10319, opened 2025-03-21
feat(autoware_overlay_rviz_plugin)!: add static blinking option for turn_signal

yhisaki:
## Description

Currently, the display of turn_signal is blinking. However the blinking feature makes difficult to debug planning modules.  This PR removes the blinking feature.

**before** (x3)

https://github.com/user-attachments/assets/f3f774f7-02b2-4e6a-8755-e01e4ab2208f

**after** (x3)

https://github.com/user-attachments/assets/c1093eda-58c0-4a59-8b25-ba7f33b0329c

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/visualization/autoware_overlay_rviz_plugin/autoware_overlay_rviz_plugin/include/turn_signals_display.hpp
+++ b/visualization/autoware_overlay_rviz_plugin/autoware_overlay_rviz_plugin/include/turn_signals_display.hpp
@@ -30,7 +30,8 @@
 #include <OgreMaterial.h>
 #include <OgreTexture.h>
 
-#include <chrono>
+#include <string>
+#include <string_view>
 
 namespace autoware_overlay_rviz_plugin
 {
@@ -44,10 +45,13 @@ public:
     const autoware_vehicle_msgs::msg::TurnIndicatorsReport::ConstSharedPtr & msg);
   void updateHazardLightsData(
     const autoware_vehicle_msgs::msg::HazardLightsReport::ConstSharedPtr & msg);
+  void setBlinkingMode(std::string_view mode);
 
 private:
   QImage arrowImage;
   QColor gray = QColor(79, 79, 79);
+
+  std::string blinking_mode_ = "Static";
 
   int current_turn_signal_;    // Internal variable to store turn signal state
   int current_hazard_lights_;  // Internal variable to store hazard lights state
--- a/visualization/autoware_overlay_rviz_plugin/autoware_overlay_rviz_plugin/include/turn_signals_display.hpp
+++ b/visualization/autoware_overlay_rviz_plugin/autoware_overlay_rviz_plugin/include/turn_signals_display.hpp
@@ -52,10 +52,6 @@ private:
   int current_turn_signal_;    // Internal variable to store turn signal state
   int current_hazard_lights_;  // Internal variable to store hazard lights state
   QImage coloredImage(const QImage & source, const QColor & color);
-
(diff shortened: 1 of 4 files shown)


xmfcx:
@yhisaki could you make it as an option in the rviz panel?

image

Here, add a new option like this:
- Turn Signal Mode
  - Static
  - Blinking

You can achieve this with an `rviz_common::properties::EnumProperty`:
https://github.com/autowarefoundation/autoware_universe/blob/3b269a944e00644718da9bdb94a0497e73b6485a/visualization/autoware_perception_rviz_plugin/include/autoware_perception_rviz_plugin/object_detection/object_polygon_display_base.hpp#L96

xmfcx:
> A compromise might be to let it blink, but have the OFF color be a darker shade of green - so even when the blinker is in its OFF-phase the user can tell that the blinker is active.

This could work too but 2 bad things could come out of it too:
- if too dark, the off color could be hard to see
- if too vibrant, the off color could make it look ugly

So just an option would help debuggers debug and let actual users get experience they would expect from a real vehicle.

visualization/autoware_overlay_rviz_plugin/autoware_overlay_rviz_plugin/include/turn_signals_display.hpp
```diff
@@ -52,10 +52,6 @@ class TurnSignalsDisplay
   int current_turn_signal_;    // Internal variable to store turn signal state
   int current_hazard_lights_;  // Internal variable to store hazard lights state
   QImage coloredImage(const QImage & source, const QColor & color);
-
-  std::chrono::steady_clock::time_point last_toggle_time_;
-  bool blink_on_ = false;
-  const std::chrono::milliseconds blink_interval_{500};  // Blink interval in milliseconds
```
xmfcx, line 58: Also when you make the update, please also make this blink interval `250`.
`500` means it will blink at 1 Hz.
But  it is 2 Hz for most vehicles.
https://www.youtube.com/watch?v=iK2K9e_tcdI

yhisaki:
@xmfcx cc @mojomex 
Thanks for your comments and sorry for the delayed reply.
I added a feature to switch the blinking mode, and also set "Static" mode as default.

[Screencast from 03-27-2025 02:26:06 PM.webm](https://github.com/user-attachments/assets/ec6e2188-940f-48fb-b4d4-88c5e8c8a57b)

visualization/autoware_overlay_rviz_plugin/autoware_overlay_rviz_plugin/src/turn_signals_display.cpp
```diff
@@ -87,28 +91,20 @@ void TurnSignalsDisplay::drawArrows(
...
-    if (leftActive) {
-      scaledLeftArrow = coloredImage(scaledLeftArrow, color);
-    }
-    if (rightActive) {
-      scaledRightArrow = coloredImage(scaledRightArrow, color);
-    }
+  auto now = std::chrono::steady_clock::now().time_since_epoch();
+  auto elapsed = std::chrono::duration_cast<std::chrono::milliseconds>(now).count();
+
+  bool blink_off = blinking_mode_ == "Blinking" && (elapsed / 333) % 2 == 0;
```
xmfcx, line 97: This is much cleaner 👍

xmfcx (approved):
Thanks for implementing this feature!

PR #10326, opened 2025-03-24
feat(control_validator): add acc check

yuki-takagi-66:
## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


kyoichi-sugahara (approved):
LGTM, based on the assumption that
- it has been verified to function correctly using the planning simulator
- no regressions have been detected through regression evaluation using TIERIV INTERNAL SCENARIOS.

PR #10337, opened 2025-03-25
refactor(departure_checker): move lane departure checker class to departure_checker 

zulfaqar-azmi-t4:
### Objective 1
"Lane" from Lane Departure Checker does not provide good semantic. Because
1. check if path departed from lanes (goal and start planner)
2. check if predicted path crosses uncrossable boundary.
3. lane departure checker node itself does not perform lane departure check. The feature is disabled due to false positive.

### Objective 2
Lane departure checker library is used by planning nodes (start and goal planner) and control node (lane departure checker node). But lane departure checker library is placed in the control folder. This caused the messy dependencies. In this PR, the lane departure checker is moved to planning folder to simplify the dependencies. 

### Objective 3
Currently there are two nodes that perform trajectory deviation checks; control validator and lane departure checker node. To prevent multiple modules performing similar checks, and to narrow lane departure checker scope, trajectory deviation check is removed from the lane departure checker.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

Evaluator: [TIER IV internal link](https://evaluation.tier4.jp/evaluation/reports/40173645-a679-52cf-b10e-78217281ac26?project_id=prd_jt)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


planning/autoware_boundary_departure_checker/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
+cmake_minimum_required(VERSION 3.14)
```
takayuki5168, line 1: Should this be in the planning?
I feel that as the lane_departure_checker was in the control component, other components than planning may refer to this class. Therefore, moving this package to the common components sounds better to me.

zulfaqar-azmi-t4: @takayuki5168 
lane departure checker will be removed once boundary departure prevention module is done. So it make sense to place it here.

takayuki5168: @zulfaqar-azmi-t4 
I don't refer to the current node structure. There may be enough possibilities that the control side will check the boundary departure, and in that case, we don't want to move the package's place so often.
The boundary departure feature is not obviously only the planning related feature, and as we have motion_utils, trajectory packages in the common directory, I still think this package should be in the common directory.

zulfaqar-azmi-t4: I see. In that case, I'll move this package to common directory. I will put this PR into draft for now. Once I moved them, I'll reopen the PR.

takayuki5168: Thank you so much.

takayuki5168 (approved):

soblin (approved):

kosuke55 (approved):

kyoichi-sugahara (approved):

mkquda (approved):

PR #10342, opened 2025-03-26
fix(motion_velocity_planner): remove Metric messages

takayuki5168:
## Description

Removed Metric messages since the PlanningFactor message can be used instead.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


maxime-clem (approved):

xtk8532704 (approved):
thank you very much.

PR #10344, opened 2025-03-26
fix(autoware_probabilistic_occupancy_grid_map): fixed transform issues and border conditions

knzo25:
## Description
When enabling the multi lidar version, it was discovered that the node was dying.
It seems to be due to how some kernel parameters are being handled, but instead of relying on implicit copies, I just moved the transform to device memory. In addition, cuda memcheck reported border conditions, so I addressed them as well :pray: 

## Related links

**Parent Issue:**

- [TIERIV INTERNAL LINK](https://star4.slack.com/archives/C076ZGRC15X/p1742782906061879)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


YoshiRi (approved):
It seems okay to me

PR #10346, opened 2025-03-26
fix(autoware_image_projection_based_fusion): unintended new container behavior

knzo25:
## Description

The current launcher was creating a new container with the same name as the `pointcloud_container`.
It effectively make nodes being load twice or in different containers.

Additionally, the new container is not a multithreaded one :sweat_smile:  (thanks to this at least I was able to find the source)

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_image_projection_based_fusion/launch/roi_pointcloud_fusion.launch.xml
+++ b/perception/autoware_image_projection_based_fusion/launch/roi_pointcloud_fusion.launch.xml
@@ -34,7 +34,7 @@
   <arg name="input/image6" default="/image_raw6"/>
   <arg name="input/image7" default="/image_raw7"/>
   <group>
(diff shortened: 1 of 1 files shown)


badai-nguyen (approved):
@knzo25 Sorry, it was my mistake in previous PR
I confirmed that before PR, there was error of duplicated node were created:
```
ros2 node list | grep roi_pointcloud
WARNING: Be aware that are nodes in the graph that share an exact name, this can have unintended side effects.
/perception/object_recognition/detection/clustering/roi_cluster/roi_pointcloud_fusion
/perception/object_recognition/detection/clustering/roi_cluster/roi_pointcloud_fusion
```
And this PR fixed this bug. 

Thanks you for your fixing. :bow:

PR #10347, opened 2025-03-26
feat(autoware_multi_object_tracker): update overlapped tracker removing process in multi obj tracker 

TadaKazuto:
## Description
Update sanitization(prune) process to fix the issue of temp lost for a laying adult found in LSim perception test

ソート更新：
Known objects(NO-UNKNOWN objs)かどうかを第一条件、meas timeを第二条件としてソートすることで、tracking objの削除が発生する場合は、”必ずインデックスの小さい方が大きい方を消す”ようにして、forループ以後の処理を単純化。
更に第三条件として、meas 最終updateからの経過時間の小さいという条件を加えることで、area threshの制限等でassociation処理で関連付けできなかったobjに対して、無駄に100ms周期で次々とspawnされるという副作用を抑制する
(
Added sorting concept:
By sorting with known objects (NO-UNKNOWN objs) as the first condition and mean time as the second condition, when a tracking obj is deleted, the smaller index will always delete the larger one, so that simplifying the processing after the for loop.
Furthermore, by adding a third condition that the time elapsed since the last update of meas must be short, the side effect of unnecessary spawning one after another at 100ms intervals for obj that could not be associated by the association process due to limitations of area threshold and etc., is suppressed.
)

## Related links

**Parent Issue:**

- Link

https://github.com/tier4/autoware_universe/pull/1915

## How was this PR tested?
LSim Regression result shows some improvements.
Total Pass Number: 207->223 (Total 740)

## Notes for reviewers
In addition to the simplification objective explained in the "Description", the sorting process works to limit unnecessary object removal in some cases.
image

 

The condition check to see elapsed time before erasing fix the issue of object lost when a known object(as a CAR, a PED and etc.) is detected as a unknown object in the next frame.
image

 

By adding a third condition that the time elapsed since the last update of meas must be short, the side effect of unnecessary spawning one after another at 100ms intervals for obj that could not be associated by the association process due to limitations of area threshold and etc., is suppressed.
image

 

## Interface changes

None.

## Effects on system behavior

None.


technolojin:
[TIER IV INTERNAL TEST](https://evaluation.tier4.jp/evaluation/reports/7c7e8aac-a316-5e3d-a234-73bd27468806?project_id=prd_jt)

technolojin:
The sorting algorithm can cause more calculation cost then current algorithm. 

Can you proof that the additional cost is acceptable? 
Example: conduct a test with many objects and check the processing time is not growing significantly.

PR #10353, opened 2025-03-27
refactor(autoware_traffic_light_map_based_detector): split utils and add test

MasatoSaeki:
## Description
Split file to perform the test and add unit test.

related PR : https://github.com/autowarefoundation/autoware_launch/pull/1448

## Related links

**Parent Issue:**

- Link

## How was this PR tested?
Evaluator
https://evaluation.ci.tier4.jp/evaluation/reports/f4120879-eda1-5408-8a97-e71964ba2019?project_id=prd_jt

Unit Test
```
colcon build --merge-install --symlink-install --packages-up-to autoware_traffic_light_map_based_detector --cmake-args -DCMAKE_BUILD_TYPE=Debug
```
```
colcon test --merge-install --packages-select autoware_traffic_light_map_based_detector --event-handlers console_cohesion+
```

## Notes for reviewers

In the `getVibrationMargin**`,  it is correct if you want to calculate correctly.

```
const double x = std::cos(margin_yaw * 0.5)* position.x() + std::sin(margin_yaw * 0.5) * position.z() + margin_width * 0.5;
const double y = std::cos(margin_pitch * 0.5)* position.y() - std::sin(margin_pitch * 0.5) * position.z() + margin_height * 0.5;
```

## Interface changes

None.

## Effects on system behavior

None.


perception/autoware_traffic_light_map_based_detector/src/traffic_light_map_based_detector_node.cpp
```diff
@@ -303,47 +221,41 @@ bool MapBasedDetector::getTrafficLightRoi(
...
+    tf2::Vector3 tl_camera_optical = tf_map2camera.inverse() * tl_map;
     // max vibration
-    const double max_vibration_x =
-      std::sin(config.max_vibration_yaw * 0.5) * camera2tl.z() + config.max_vibration_width * 0.5;
-    const double max_vibration_y = std::sin(config.max_vibration_pitch * 0.5) * camera2tl.z() +
-                                   config.max_vibration_height * 0.5;
-    const double max_vibration_z = config.max_vibration_depth * 0.5;
+    tf2::Vector3 point3d = utils::getVibrationMarginTopLeft(
+      tl_camera_optical, config.max_vibration_pitch, config.max_vibration_yaw,
+      config.max_vibration_height, config.max_vibration_width, config.max_vibration_depth);
```
technolojin, line 229: 1. `getVibrationMarginTopLeft` and `getVibrationMarginBottomRight` are almost same. integrate it into one `getVibrationMargin`
2. hand-over all the config margins are complicate. give a variable as `Config` type. to do so, may need to separate the type definition

```suggestion
    tf2::Vector3 margin = utils::getVibrationMarginTopLeft(tl_camera_optical.z, config);
    tf2::Vector3 point3d = tl_camera_optical - margin;
```

perception/autoware_traffic_light_map_based_detector/src/traffic_light_map_based_detector_utils.cpp
technolojin, line 1: It is named `utils`, but the contents are specific functions for this node.


node: interface and process definition
`process`: all the method/functions to process

`utils` or `common`: common functions to be shared over nodes/packages

technolojin (commented):
I have some minor improvement points, and those are not very mandatory.

perception/autoware_traffic_light_map_based_detector/src/traffic_light_map_based_detector_node.cpp
MasatoSaeki: Regarding 1
Fixed in https://github.com/autowarefoundation/autoware_universe/pull/10353/commits/ecc788bc953dae001a36d6a5c8b02a24c2345f28

Regarding 2
I agree with you. But I leave it at that.
`Config` is defined in private member of `TrafficLightMapBasedDetectorNode`. So I would not like to pass as argument if use function like `utils`. And this `Config` is not sophisticated so I have the plann which I will reduce variable in it.

perception/autoware_traffic_light_map_based_detector/src/traffic_light_map_based_detector_utils.cpp
MasatoSaeki: Thank you for the advice and agree with you.
But I split them because I would like to add unit test. What should I do in order to do it?(So should I name file `traffic_light_map_based_detector_process`?)

MasatoSaeki: Fixed in https://github.com/autowarefoundation/autoware_universe/pull/10353/commits/796808a5e52205f2393b986583eb0272438176b0

technolojin (approved):
LGTM

PR #10363, opened 2025-03-27
refactor(control validaor): refactor control_validator

yuki-takagi-66:
## Description
Initially, control_validator was implemented as a single-function validator.
However, it has since grown significantly to incorporate five distinct detection capabilities, each with a high independence.
This historical evolution has led to the current implementation of autoware_control_validator becoming somewhat complex and causing some warning issues.
For example, I have identified the following issues:

- The node member variable LowpassFilter1d `vehicle_vel_` is used in both the velocity_validator and overrun_validator, even though access to this variable is sensitive.
- In the `ControlValidator::validate()` function, both the local variable `reference_trajectory` and the global `current_reference_trajectory` are accessed, despite them being references to the same underlying data.

This refactor PR fix the above issues by the following approach.
- reduce member variables of the node class 
- separate validation logic into individual classes
- align the validate functions

## How was this PR tested?

- tier4 scenario test with connecting MRM https://evaluation.tier4.jp/evaluation/reports/9b942722-25b2-5d48-8efa-02f3566b480f?project_id=prd_jt
- DLR tests https://tier4.atlassian.net/wiki/spaces/PCPT/pages/3458434691/DLR+control-validator
- engage aviable in psim

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/control/autoware_control_validator/include/autoware/control_validator/control_validator.hpp
+++ b/control/autoware_control_validator/include/autoware/control_validator/control_validator.hpp
@@ -39,8 +39,6 @@
 #include <tuple>
 #include <utility>
 
-class AccelerationValidatorTest;
-
 namespace autoware::control_validator
 {
 using autoware_control_msgs::msg::Control;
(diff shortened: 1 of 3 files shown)


control/autoware_control_validator/include/autoware/control_validator/control_validator.hpp
```diff
@@ -69,28 +95,76 @@ struct ValidationParams
...
-      autoware_utils::get_or_declare_parameter<double>(node, "thresholds.acc_error_offset");
-    e_scale = autoware_utils::get_or_declare_parameter<double>(node, "thresholds.acc_error_scale");
-    const double acc_lpf_gain =
-      autoware_utils::get_or_declare_parameter<double>(node, "acc_lpf_gain");
-    desired_acc_lpf.setGain(acc_lpf_gain);
-    measured_acc_lpf.setGain(acc_lpf_gain);
-  };
+  : e_offset{autoware_utils::get_or_declare_parameter<double>(node, "thresholds.acc_error_offset")},
+    e_scale{autoware_utils::get_or_declare_parameter<double>(node, "thresholds.acc_error_scale")},
+    desired_acc_lpf{autoware_utils::get_or_declare_parameter<double>(node, "acc_lpf_gain")},
```
kyoichi-sugahara, line 102: NITS:
From a consistency perspective, initializing acc_lpf_gain with const qualification and then using this variable to initialize the LowpassFilter1d constructor is one of the approach.

control/autoware_control_validator/src/control_validator.cpp
```diff
@@ -56,8 +73,64 @@ bool AccelerationValidator::is_in_error_range() const
...
+  const double v_vel = vehicle_vel_lpf.filter(kinematics.twist.twist.linear.x);
+
+  // NOTE: the same velocity threshold as autoware::motion_utils::searchZeroVelocity
+  res.has_overrun_stop_point = res.dist_to_stop < -overrun_stop_point_dist_th &&
+                               res.nearest_trajectory_vel < 1e-3 && v_vel > 1e-3;
+}
+
 ControlValidator::ControlValidator(const rclcpp::NodeOptions & options)
-: Node("control_validator", options), validation_params_(), vehicle_info_()
+: Node("control_validator", options), vehicle_info_()
```
kyoichi-sugahara, line 133:
```suggestion
: Node("control_validator", options), vehicle_info_(),
  latency_validator{*this},
  trajectory_validator{*this},
  // other validator
```

NITS:
By initializing Validators declared as members of the ControlValidator class in the constructor's initialization list, objects are directly constructed, which provides performance benefits.

kyoichi-sugahara (approved):
LGTM! Thank you for good refactoring! 
I left some comments as NITS and you can ignore them.

control/autoware_control_validator/include/autoware/control_validator/control_validator.hpp
yuki-takagi-66: We can't take this approach because we have to change its stored value

control/autoware_control_validator/src/control_validator.cpp
yuki-takagi-66: Thank you.

PR #10375, opened 2025-04-01
fix(pred_path_checker,dyn_obstacle_stop): use output of removeOverlaps

maxime-clem:
## Description

There are currently build issues that can occur because the output of `removeOverlapPoints` is marked `[[no_discard]]` but is discarded.
```
error: ignoring return value of 'T autoware::motion_utils::removeOverlapPoints(const T&, size_t) [with T = std::vector<autoware_planning_msgs::msg::TrajectoryPoint_<std::allocator<void> > >; size_t = long unsigned int]', declared with attribute 'nodiscard' [-Werror=unused-result]
  126 |   autoware::motion_utils::removeOverlapPoints(ego_data.trajectory);
      |                   
```
This PR fixes the issue in 2 packages.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?
Psim
## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/control/autoware_predicted_path_checker/src/predicted_path_checker_node/predicted_path_checker_node.cpp
+++ b/control/autoware_predicted_path_checker/src/predicted_path_checker_node/predicted_path_checker_node.cpp
@@ -485,7 +485,7 @@ Trajectory PredictedPathCheckerNode::cutTrajectory(
     cut.points.push_back(point);
     total_length += points_distance;
   }
-  autoware::motion_utils::removeOverlapPoints(cut.points);
+  cut.points = autoware::motion_utils::removeOverlapPoints(cut.points);
 
   return cut;
 }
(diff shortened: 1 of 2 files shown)


soblin (approved):

kyoichi-sugahara (approved):
build fail is fixed Thanks!!!

sasakisasaki (approved):
I verified the build errors are resolved by applying the fix provided by this PR. Thank you so much! I used following command to check if the build-failed packages ~~are~~ become build-able.

```
colcon build --base-paths . --symlink-install --cmake-args -DCMAKE_BUILD_TYPE=Release --packages-up-to autoware_motion_velocity_dynamic_obstacle_stop_module autoware_motion_velocity_dynamic_obstacle_stop_module
```

PR #10389, opened 2025-04-03
feat(repos): fix agnocast version to 2.0.1

sykwer:
## Description
https://github.com/autowarefoundation/autoware/pull/5977
https://github.com/tier4/agnocast/tree/2.0.1

Some publicly exposed environment variable names have changed in a backward-incompatible way, so I made a major update. Currently, agnocast's public API is not used in Autoware, so I've confirmed that the major update has no actual impact. That said, would it be possible to upgrade?

## Related links

**Parent Issue:**

- Link

## Notes for reviewers
We've discovered that due to the specifications of PPA, releasing a new version of agnocast-kmod or agnocast-heaphook causes the previous version to become unavailable, with no workaround. As a result, we'll be changing our release policy. Apologies for the lack of prior investigation into the PPA behavior.

We will provide separate packages for each major version—for example: agnocast-kmod-v1, agnocast-kmod-v2, agnocast-heaphook-v1, agnocast-heaphook-v2, and so on.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/build_depends_humble.repos
+++ b/build_depends_humble.repos
@@ -62,4 +62,4 @@ repositories:
   middleware/external/agnocast:
     type: git
     url: https://github.com/tier4/agnocast.git
-    version: main # TODO(TIER IV): to be updated to v1.1.0
+    version: 2.0.1


sykwer:
> We've discovered that due to the specifications of PPA, releasing a new version of agnocast-kmod or agnocast-heaphook causes the previous version to become unavailable, with no workaround. As a result, we'll be changing our release policy. Apologies for the lack of prior investigation into the PPA behavior.
We will provide separate packages for each major version—for example: agnocast-kmod-v1, agnocast-kmod-v2, agnocast-heaphook-v1, agnocast-heaphook-v2, and so on.

We'll re-open this PR after this is solved.

PR #10397, opened 2025-04-04
feat(pointpainting_fusion): add diagnostics for processing time of pointpainting

a-maumau:
## Description
This PR adds diagnostics for pointpainting's processing time.

Consecutive delays will be also checked.
If the difference of two timestamp (current and last in-time processed timestamp) is larger than `max_acceptable_consecutive_delay_ms`, it will determine as an error state.

If the new processing time is below the `max_allowed_processing_time_ms`, it will update the timestamp of last in-time processed case.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

Run rqt_runtime_monitor by
```sh
```

Screenshot from 2025-04-04 13-56-44

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_image_projection_based_fusion/launch/pointpainting_fusion.launch.xml
+++ b/perception/autoware_image_projection_based_fusion/launch/pointpainting_fusion.launch.xml
@@ -25,6 +25,7 @@
   <arg name="ml_package_param_path" default="$(var model_path)/$(var model_name)_ml_package.param.yaml"/>
   <arg name="class_remapper_param_path" default="$(var model_path)/detection_class_remapper.param.yaml"/>
   <arg name="sync_param_path" default="$(find-pkg-share autoware_image_projection_based_fusion)/config/fusion_common.param.yaml"/>
+  <arg name="common_param_path" default="$(find-pkg-share autoware_image_projection_based_fusion)/config/pointpainting_common.param.yaml"/>
   <arg name="build_only" default="false" description="shutdown node after TensorRT engine file is built"/>
 
   <arg name="use_pointcloud_container" default="false" description="use pointcloud_container"/>
@@ -51,6 +52,7 @@
         <param from="$(var model_param_path)" allow_substs="true"/>
         <param from="$(var ml_package_param_path)" allow_substs="true"/>
         <param from="$(var sync_param_path)"/>
+        <param from="$(var common_param_path)"/>
         <param from="$(var class_remapper_param_path)"/>
         <param name="rois_number" value="$(var input/rois_number)"/>
 
--- a/perception/autoware_image_projection_based_fusion/config/pointpainting_common.param.yaml
+++ b/perception/autoware_image_projection_based_fusion/config/pointpainting_common.param.yaml
@@ -1,11 +1,11 @@
 /**:
   ros__parameters:
-    # When the processing time exceeds the max_allowed_processing_time,
-    # a warning will be triggered.
     diagnostics:
-      max_allowed_processing_time_ms: 700.0 # in milliseconds
+      # When the processing time exceeds the max_allowed_processing_time,
+      # a warning will be triggered.
+      max_allowed_processing_time_ms: 200.0 # in milliseconds
       # If the warning is triggered and if a timestamp of last normal processing is more than
       # max_acceptable_consecutive_delay_ms, an error diagnostic message will be sent.
-      max_acceptable_consecutive_delay_ms: 3000.0
+      max_acceptable_consecutive_delay_ms: 1000.0
       # interval of diagnostic callback in milliseconds
       pointpainting_validation_callback_interval_ms: 100.0 # in milliseconds
--- a/perception/autoware_image_projection_based_fusion/config/pointpainting_common.param.yaml
+++ b/perception/autoware_image_projection_based_fusion/config/pointpainting_common.param.yaml
@@ -1,0 +1,11 @@
+/**:
+  ros__parameters:
+    # When the processing time exceeds the max_allowed_processing_time,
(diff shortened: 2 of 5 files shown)


perception/autoware_image_projection_based_fusion/launch/pointpainting_fusion.launch.xml
```diff
@@ -25,6 +25,7 @@
   <arg name="ml_package_param_path" default="$(var model_path)/$(var model_name)_ml_package.param.yaml"/>
   <arg name="class_remapper_param_path" default="$(var model_path)/detection_class_remapper.param.yaml"/>
   <arg name="sync_param_path" default="$(find-pkg-share autoware_image_projection_based_fusion)/config/fusion_common.param.yaml"/>
+  <arg name="common_param_path" default="$(find-pkg-share autoware_image_projection_based_fusion)/config/pointpainting_common.param.yaml"/>
```
badai-nguyen, line 28: @a-maumau Thanks for your PR.
cc @technolojin

```suggestion
  <arg name="common_param_path" default="$(find-pkg-share autoware_image_projection_based_fusion)/config/pointpainting_diagnostic.param.yaml"/>
```
- If you dont have a parameters located in this param files, I prefer `pointpainting_diagnostic.param.yaml` 

- BTW, Currently, there are already 3 param files related to pointpainting, I think merging diagnostic param to one of them could be better? such as pointpainting.param.yaml. How do you think?

technolojin: @badai-nguyen 

existing 3 parameter files are related to the detection model.
This configuration is for node implementation.

To combine parameter files, the diagnostics parameter files is named pointpainting_common.

If too many parameter files are issue, we may need to integrate the model related parameter files to 1 or 2.

a-maumau: @badai-nguyen thank you for your suggestion!
For my answer for this is like https://github.com/autowarefoundation/autoware_universe/pull/10398#issuecomment-2781824805
TL;DR: I agree with your points, and also agree with @technolojin 's points.
So we should discuss this considering future development/features and file management.

badai-nguyen: > So we should discuss this considering future development/features and file management.

@a-maumau @technolojin Then I think it good for now. However, IMO, we should have a further discussion about parameter file management later. Especially,  including the consolidation of other diagnostics parameters files from other package into the same directory to make it easier to manage the MRM trigger conditions.

technolojin (changes requested):
The evaluating processing time contains all the message synchronization latency.
`stop_watch_ptr_->toc("processing_time", true)`

1. It could be a very fluctuating value due to input message timings.
2. It is not actual `processing time` of the model
3. The process of image projection is done my multiple callbacks. It is hard to expect total amount of process.

Therefore, I would like to suggest to measure independent processing times, such as

- `fuse_on_single_image` loop
- `export_process` or `postprocess`

perception/autoware_image_projection_based_fusion/launch/pointpainting_fusion.launch.xml
technolojin: I agree. From my perspective, this topic just triggered a common issue of the lidar detector parameter management.

For MRM trigger condition, it is discussed in the failure detection WG (incl. Horibe-san). 

- Each component/node/package shall have its own diagnostics parameter (not MRM config)
- MRM condition is configured in the system side
 
Therefore, node configuration will not be in a same place

a-maumau:
@technolojin 
> The evaluating processing time contains all the message synchronization latency.
stop_watch_ptr_->toc("processing_time", true)

Sorry, I overlooked my typo.
The part of `stop_watch_ptr_->toc("processing_time")` was not intended, it should be `stop_watch_ptr_->toc("pointpainting_processing_time")`, which I fixed it in [c7cdf57](https://github.com/autowarefoundation/autoware_universe/pull/10397/commits/c7cdf57163c5e6e4f940b12856bec78eed9c3f6a).

So, it is now only checking the `postprocess` function, and corresponding with this change, I fixed the parameter in [b7b3257](https://github.com/autowarefoundation/autoware_universe/pull/10397/commits/b7b3257b37e3a13663e7a791925f17f923272825).

technolojin (approved):
LGTM

badai-nguyen:
@a-maumau cc @technolojin 
I wondering why we don't use the same processing_time here for diagnostics? which is recorded in rosbag and could be used for later debuging purpose?
https://github.com/autowarefoundation/autoware_universe/blob/e9d43e45ca477ec70e58421be59cebf867f11f5e/perception/autoware_image_projection_based_fusion/src/fusion_node.cpp#L333-L334

technolojin:
> @a-maumau cc @technolojin I wondering why we don't use the same processing_time here for diagnostics? which is recorded in rosbag and could be used for later debuging purpose?
> 
> https://github.com/autowarefoundation/autoware_universe/blob/e9d43e45ca477ec70e58421be59cebf867f11f5e/perception/autoware_image_projection_based_fusion/src/fusion_node.cpp#L333-L334

The implementation of the `processing_time` includes fuse_on_single_image, but it is not predictable.
* not fixed number of ROIs
* not fixed trigger timing

It is impossible to set a proper fixed processing time threshold, if we use the `processing_time`.

Therefore, in the pointpainting, implementing diagnostics only for the ml model inference part.

badai-nguyen (approved):
LGTM

PR #10398, opened 2025-04-04
feat(lidar_transfusion): add diagnostics for processing time

a-maumau:
## Description
This PR adds processing time diagnostics.

Consecutive delays will be also checked.
If the difference of two timestamp (current and last in-time processed timestamp) is larger than `max_acceptable_consecutive_delay_ms`, it will determine as an error state.

If the new processing time is below the `max_allowed_processing_time_ms`, it will update the timestamp of last in-time processed case.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

Run rqt_runtime_monitor by
```sh
```

Screenshot from 2025-04-04 15-36-39

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_lidar_transfusion/config/transfusion_common.param.yaml
+++ b/perception/autoware_lidar_transfusion/config/transfusion_common.param.yaml
@@ -1,0 +1,11 @@
+/**:
+  ros__parameters:
+    # When the processing time exceeds the max_allowed_processing_time,
+    # a warning will be triggered.
+    diagnostics:
+      max_allowed_processing_time_ms: 200.0 # in milliseconds
+      # If the warning is triggered and if a timestamp of last normal processing is more than
+      # max_acceptable_consecutive_delay_ms, an error diagnostic message will be sent.
+      max_acceptable_consecutive_delay_ms: 1000.0
+      # interval of diagnostic callback in milliseconds
+      validation_callback_interval_ms: 100.0 # in milliseconds
--- a/perception/autoware_lidar_transfusion/config/transfusion_common.param.yaml
+++ b/perception/autoware_lidar_transfusion/config/transfusion_common.param.yaml
@@ -1,11 +1,6 @@
 /**:
   ros__parameters:
-    # When the processing time exceeds the max_allowed_processing_time,
-    # a warning will be triggered.
     diagnostics:
-      max_allowed_processing_time_ms: 200.0 # in milliseconds
-      # If the warning is triggered and if a timestamp of last normal processing is more than
-      # max_acceptable_consecutive_delay_ms, an error diagnostic message will be sent.
+      max_allowed_processing_time_ms: 200.0
       max_acceptable_consecutive_delay_ms: 1000.0
-      # interval of diagnostic callback in milliseconds
-      validation_callback_interval_ms: 100.0 # in milliseconds
+      validation_callback_interval_ms: 100.0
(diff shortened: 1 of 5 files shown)


perception/autoware_lidar_transfusion/config/transfusion_common.param.yaml
```diff
@@ -0,0 +1,11 @@
+/**:
```
amadeuszsz, line 1: Could you please remove all the comments? You already provided parameters description in schema file.

amadeuszsz (changes requested):
Thanks for your PR! Overall looks good, but just a simple question. Do we need to have another `param.yaml` file for storing these parameters? Couldn't we just reuse `transfusion.param.yaml` file? We can have separate file for these parameters if we plan to share this config among all related nodes (transfusion and centerpoint). In that case, file could be named `lidar_detector_diag.param.yaml`. Then we could have one parameter file for both packages, stored in `autoware_launch`, also called `lidar_detector_diag.param.yaml`. I see you already use specific name for [centerpoint](https://github.com/autowarefoundation/autoware_universe/pull/10396/files), so I guess it wasn't your intention. What do you think about this idea?

a-maumau:
@amadeuszsz thank you for your suggestions.

> Do we need to have another param.yaml file for storing these parameters? Couldn't we just reuse transfusion.param.yaml file? We can have separate file for these parameters if we plan to share this config among all related nodes (transfusion and centerpoint). In that case, file could be named lidar_detector_diag.param.yaml. Then we could have one parameter file for both packages, stored in autoware_launch, also called lidar_detector_diag.param.yaml

Actually No, we can merge the parameter into the model parameter file. Honestly, at least in the current state, it might be better to do so since the parameter is only related to diagnostics for the LiDAR model itself.
The reason I originally separated the file is that, if diagnostics become more diverse in the future, it might be easier to manage them in a dedicated file, separate from the model parameters.

Also, I didn't make the file shared across different LiDAR detection models because I wanted to allow flexibility in setting warnings/errors based on the specific expectations of each model.

I understand that this diagnosis feature is intended to ensure the safety of the AD component, and that the parameters required for safe driving should ideally be consistent—so keeping everything in a single file could be simpler and more intuitive.

Does that make sense?
I'm happy to make the change based on your suggestion, as I agree with your concern.

perception/autoware_lidar_transfusion/config/transfusion_common.param.yaml
a-maumau: I removed it in [0733f27](https://github.com/autowarefoundation/autoware_universe/pull/10398/commits/0733f27c765fa25b1066e6df3fa1115f057d53fd).

amadeuszsz:
> I wanted to allow flexibility in setting warnings/errors based on the specific expectations of each model.
> I'm happy to make the change based on your suggestion, as I agree with your concern.

Thanks for your clarification! I prefer simplification, but if we plan to define various constraints across models, we should keep separate files. I'll leave this decision up to you 👍🏻

amadeuszsz (approved):
LGTM!

a-maumau:
@amadeuszsz Since the `lidar_centerpoint` package is already done in this manner, for now, I will keep this manner here too if other reviewers don't mind with this. If the file management get troublesome for some process, I will make it in a shared file or merge in to the existing parameter file.

PR #10400, opened 2025-04-04
fix: missing dependency on tf2_sensor_msgs

Timple:
## Description

Used here: https://github.com/autowarefoundation/autoware_universe/blob/main/sensing/autoware_pointcloud_preprocessor/src/pointcloud_densifier/pointcloud_densifier_node.cpp#L20

## Related links

**Parent Issue:**

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


knzo25 (approved):
LGTM. Thanks for the catch :pray:

PR #10403, opened 2025-04-07
fix(autoware_path_optimizer): incorrect application of input velocity due to badly mapping output trajectory to input trajectory

PanConChicharron:
## Description
When applying the input velocity to the output of MPT, the initial pose of the output of MPT is used to trim the input trajectory. In certain cases when  there is a sharp U-turn, depending on the width of the adjacent, opposing lane into which Ego is turning into, the initial pose of the output trajectory may be mapped to a point in the opposing lane - leading to incorrect input velocity application.

This PR adds a yaw-thresholded closest-distance point search step to ensure that this doesn't occur and also trims the output of MPT to start from the current ego pose so that the initial pose of the MPT output will always map to a point within the current lane.
## Related links

**Parent Issue:**
https://tier4.atlassian.net/browse/RT0-35764

## How was this PR tested?
Evaluator: https://evaluation.ci.tier4.jp/evaluation/reports/tables/new?catalog_id=095ae8b3-a23b-4a08-8674-10a681449693&filter=&job_id=4ce421ba-005b-5ca9-ae29-1f9a908a8733&project_id=prd_jt&table_config=date&target_ids=83c8bbb7-fdb5-53c8-ba72-4979a27e45b6

Failing tests are due to start-up delay - not caused due to changes
## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/autoware_path_optimizer/src/node.cpp
+++ b/planning/autoware_path_optimizer/src/node.cpp
@@ -419,21 +419,20 @@ void PathOptimizer::applyInputVelocity(
 {
   autoware_utils::ScopedTimeTrack st(__func__, *time_keeper_);
 
-  const size_t ego_seg_idx =
-    trajectory_utils::findEgoSegmentIndex(input_traj_points, ego_pose, ego_nearest_param_);
-
   // trim to ego-pose
   const size_t ego_seg_idx_output_traj =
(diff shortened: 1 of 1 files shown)


takayuki5168 (approved):

mitsudome-r (approved):

youtalk (approved):
After merging https://github.com/autowarefoundation/autoware_core/pull/355, LGTM

PR #10422, opened 2025-04-09
feat(control_validator): add over run estimation feature

yuki-takagi-66:
## Description
To guarantee DO NOT overrun the stop points or the goal point, this PR adds a new validation fetature.
The node publish error diag for the cases which the ego will overrun the stop points even if the ego decelerate by assumed deceleration value.
https://github.com/autowarefoundation/autoware_universe/pull/10422/files#diff-492e13c1cee80cc9825bff55dc294565ae2382b5e117a5f7f0cc03512280b08fR126-R127

## Related links
https://github.com/autowarefoundation/autoware_launch/pull/1394

**Parent Issue:**

- Link

## How was this PR tested?
With connnecting MRM,
- engage available in psim 
- tier4 internal DLR tests
- tier4 internal scenario tests

## Notes for reviewers

None.

## Interface changes

None.

<

## Effects on system behavior

None.

### DIFF
--- a/control/autoware_control_validator/src/control_validator.cpp
+++ b/control/autoware_control_validator/src/control_validator.cpp
@@ -123,10 +123,18 @@ void OverrunValidator::validate(
       .longitudinal_velocity_mps;
 
   const double v_vel = vehicle_vel_lpf.filter(kinematics.twist.twist.linear.x);
+  res.pred_dist_to_stop =
+    res.dist_to_stop - v_vel * assumed_delay_time - v_vel * v_vel / (2.0 * assumed_limit_acc);
 
   // NOTE: the same velocity threshold as autoware::motion_utils::searchZeroVelocity
-  res.has_overrun_stop_point = res.dist_to_stop < -overrun_stop_point_dist_th &&
-                               res.nearest_trajectory_vel < 1e-3 && v_vel > 1e-3;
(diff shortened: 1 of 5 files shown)


kyoichi-sugahara (approved):
LGTM!
Please add Supported features in readme before merge

yuki-takagi-66:
@maxime-clem 
Could you please review the detection formula of this PR?
Considering whether this formula is appropriate for the purpose of the PR.

control/autoware_control_validator/src/control_validator.cpp
```diff
@@ -227,6 +235,12 @@ void ControlValidator::setup_diag()
       stat, !validation_status_.has_overrun_stop_point,
       "The vehicle has overrun the front stop point on the trajectory.");
   });
+  d.add(ns + "will_overrun_stop_point", [&](auto & stat) {
+    set_status(
+      stat, !validation_status_.has_overrun_stop_point,
```
maxime-clem, line 240:
```suggestion
      stat, !validation_status_.will_overrun_stop_point,
```

maxime-clem (changes requested):
The calculations look good but are a bit hard to understand so some comments should be added.
I also think there is one mistake when adding the status for `will_overrun_stop_point`.

maxime-clem (approved):

PR #10440, opened 2025-04-11
feat(planning_validator): add lateral jerk validation feature

kyoichi-sugahara:
## Description

**Parameter change [PR](https://github.com/autowarefoundation/autoware_launch/pull/1406) shuold be merged first**

- In a previous [PR](https://github.com/autowarefoundation/autoware_universe/pull/10385), I projected acceleration laterally and considered it in the calculation of lateral acceleration. However, when the embedded acceleration and velocity in the path points are directed toward the next path point, this projection was unnecessary, so I reverted the change in this PR.
- Added calculation of lateral jerk and validation based on those values. 

### Lateral Acceleration
Invalid if the expected lateral acceleration/deceleration is too large. Lateral acceleration is calculated using:

$$a_{lat} = v_{lon}^2 \cdot \kappa$$

Where:
- $v_{lon}$ is longitudinal velocity
- $\kappa$ is curvature

Note: Since the acceleration embedded in path points is perpendicular to the derived lateral acceleration, projections are not considered. The velocity and acceleration assigned to each point are directed toward the next path point.

### Lateral Jerk
Invalid if the rate of change of lateral acceleration is too large. Lateral jerk is calculated using:

$$j_{lat} = v_{lon}^3 \cdot \frac{d\kappa}{ds} + 3 \cdot v_{lon}^2 \cdot a_{lon} \cdot \kappa$$

Where:
- $v_{lon}$ is longitudinal velocity
- $\kappa$ is curvature
- $a_{lon}$ is longitudinal acceleration
- $\frac{d\kappa}{ds}$ is the rate of curvature change with respect to distance

Implementation note: In this implementation, the curvature change ($\frac{d\kappa}{ds}$) is not considered, simplifying the calculation to only the second term. The lateral jerk represents how quickly the lateral acceleration changes, which affects ride comfort and vehicle stability.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

- [x] run psim and confirmed jerk value is calculated without unexpected high or low value
- [x] passed newly created unit test
- [x] [PASSED TIERIV INTERNAL SCENARIOS](https://evaluation.tier4.jp/evaluation/reports/2eb1f5fc-413e-5d44-98a4-1003d166b733?project_id=prd_jt)

## Notes for reviewers

I have set several assumptions when calculating lateral jerk and lateral acceleration.
I intend to explain these in the README, this PR's description, and within the code. I would like to review these concepts and check if there are any points that need to be addressed.

## Interface changes

None.

## Effects on system behavior

None.


planning/autoware_planning_validator/src/utils.cpp
```diff
@@ -229,13 +221,61 @@ std::pair<double, size_t> calcMaxLateralAcceleration(const Trajectory & trajecto
...
+  // j_lat = v_lon^3 * (dk/ds) + 3 * v_lon^2 * a_lon * k
+  // However, the dk/ds term is omitted here because the curvature calculation
+  // is currently unstable, making it difficult to derive an accurate rate of
+  // curvature change. Therefore, we only use the second term for a stable estimation.
+  for (size_t i = 0; i < trajectory.points.size(); ++i) {
+    const double v_lon = trajectory.points.at(i).longitudinal_velocity_mps;
+    const double a_lon = trajectory.points.at(i).acceleration_mps2;
+
+    // Calculate lateral jerk using simplified formula: jerk_lat = 3 * v_lon^2 * a_lon * curvature
+    lateral_jerk_vector.at(i) = 3.0 * v_lon * v_lon * a_lon * curvature_vector.at(i);
```
maxime-clem, line 256: please add a TODO here to indicate that the full equation will be used when accurate curvatures become available.

maxime-clem (approved):

planning/autoware_planning_validator/src/utils.cpp
kyoichi-sugahara: left TODO comment in https://github.com/autowarefoundation/autoware_universe/pull/10440/commits/177b506b8844d2aa12d8232ad1d087f9cdccb13a

kyoichi-sugahara: left TODO comment in https://github.com/autowarefoundation/autoware_universe/pull/10440/commits/177b506b8844d2aa12d8232ad1d087f9cdccb13a

planning/autoware_planning_validator/src/planning_validator.cpp
```diff
@@ -462,6 +465,22 @@ bool PlanningValidator::checkValidLateralAcceleration(const Trajectory & traject
...
+{
+  if (!params_.validation_params.lateral_jerk.enable) {
+    return true;
+  }
+
+  const auto [max_lateral_jerk, i] = calc_max_lateral_jerk(trajectory);
+  validation_status_.max_lateral_jerk = max_lateral_jerk;
+  if (max_lateral_jerk > params_.validation_params.acceleration.lateral_th) {
+    debug_pose_publisher_->pushPoseMarker(trajectory.points.at(i), "lateral_jerk");
+    is_critical_error_ |= params_.validation_params.acceleration.is_critical;
```
mkquda, line 478:
```suggestion
    is_critical_error_ |= params_.validation_params.lateral_jerk.is_critical;
```

mkquda (approved):

Owen-Liuyuxuan:
@kyoichi-sugahara 

Given acceleration: 

$$a_{lat} = v_{lon}^2 \cdot \kappa$$

Jerk as the acceleration derivatives w.r.t. time: 

$$j_{lat} = v_{lon}^3 \cdot \frac{d\kappa}{ds} + 2 \cdot v_{lon} \cdot a_{lon} \cdot \kappa$$

It does not look like $3 v^2$ in your original formula.

From this formula, the code should be `2 * v_lon * a_lon * curvature`.

PR #10441, opened 2025-04-11
feat(autoware_bevdet): implementation of bevdet using tensorrt

RahulSundarMCW:
## Description

**Parent Issue:**
- https://github.com/autowarefoundation/autoware/issues/4635

This PR is the continuation of:
- https://github.com/autowarefoundation/autoware_universe/pull/7956

Integrating BEVDet into Autoware for 3D object detection based on multi-view images.

**NEWS:** Refactoring the integration method of bevdet-tensorrt-cpp to Autoware:

We wrap the original [bevdet-tensorrt-cpp](https://github.com/LCH1238/bevdet-tensorrt-cpp/tree/one) code as a [vendor package](https://github.com/autowarefoundation/bevdet_vendor) and maintain it outside Autoware with different CIs settings and code standards. Then we added through `autoware.repos` and just call the libraries in `autoware_tensorrt_bevdet` node.

**NOTE:** This PR depends on:

- https://github.com/autowarefoundation/autoware/pull/5456
- https://github.com/autowarefoundation/bevdet_vendor/pull/1

## How was this PR tested?

This PR has been tested on local environment.

**env1:**
```
GeForce RTX 2080 Ti 
CUDA 12.4
TensorRT 10.8.0
```

## Notes for reviewers
Tested using [nuscenes dataset](https://github.com/Owen-Liuyuxuan/ros2_dataset_bridge/blob/main/docs/nusc_readme.md).
[onnx model file](https://drive.google.com/file/d/1eMGJfdCVlDPBphBTjMcnIh3wdW7Q7WZB/view?usp=sharing)
Instructions for AWSIM Testing will be added here soon

Reviewers can test the functionality of this PR with respect to [nuscenes dataset](https://github.com/Owen-Liuyuxuan/ros2_dataset_bridge/blob/main/docs/nusc_readme.md) using this [test branch](https://github.com/autowarefoundation/autoware_universe/compare/main...RahulSundarMCW:autoware_universe:testing_nuscenes) and follow the instructions given [here](https://github.com/RahulSundarMCW/autoware_universe/blob/fb1fe4b9975971a4a92c7f6db5ef5a85bdd06c55/perception/autoware_tensorrt_bevdet/README.md).

## Interface changes

Add a new `perception_mode`, If set `perception_mode = camera`, the detector will launch `tensorrt_bevdet` node.

### Additions and removals

| Change type | Topic Type | Topic Name                | Message Type                                             | Description                          |
|-------------|------------|---------------------------|----------------------------------------------------------|--------------------------------------|
| Added       | Sub        | ~/input/topic_img_fl      | sensor_msgs::msg::Image                                  | input front_left camera image        |
| Added       | Sub        | ~/input/topic_img_f       | sensor_msgs::msg::Image                                  | input front camera image             |
| Added       | Sub        | ~/input/topic_img_fr      | sensor_msgs::msg::Image                                  | input front_right camera image       |
| Added       | Sub        | ~/input/topic_img_bl      | sensor_msgs::msg::Image                                  | input back_left camera image         |
| Added       | Sub        | ~/input/topic_img_b       | sensor_msgs::msg::Image                                  | input back camera image              |
| Added       | Sub        | ~/input/topic_img_br      | sensor_msgs::msg::Image                                  | input back_right camera image        |
| Added       | Pub        | ~/output/boxes            | autoware_perception_msgs::msg::DetectedObjects           | detected objects                     |
|Added       | Pub         | ~/output_bboxes      | visualization_msgs::msg::MarkerArray                  | detected objects for nuScenes visualization |

### Additions and removals

| Change type | Parameter Name      | Type   | Default Value | Description                    |
|-------------|---------------------|--------|----------------|--------------------------------|
| Added       | bev_detection_model | string | bevdet         | Set BEV detection model type   |

### Modifications

| Version | Parameter Name  | Type   | Default Value | Description                              |
|---------|------------------|--------|----------------|------------------------------------------|
| Old     | perception_mode  | string | lidar          | Set perception mode                      |
| New     | perception_mode  | string | lidar          | Add a new perception mode of `camera`    |

## TODO

- [ ] Testing with AWSIM Labs simulation
- [ ] Validating fps with existing Autoware telemetry package
- [ ] Adding Int8 quantization support
- [ ] Exploring Integration to Autoware Launch rep ([#Issue 5912](https://github.com/autowarefoundation/autoware/issues/5912))

## Effects on system behavior

None.

Looping in  @Selventhiran-Rengaraj-MCW , @ramaseshanms-mcw.


perception/autoware_tensorrt_bevdet/src/bevdet_node.cpp
```diff
@@ -0,0 +1,235 @@
...
+  }
+  Eigen::Matrix3f intrinsics;
+  getCameraIntrinsics(msg, intrinsics);
+  cams_intrin_[idx] = intrinsics;
+
+  Eigen::Quaternion<float> rot;
+  Eigen::Translation3f translation;
+  getTransform(
+    tf_buffer_->lookupTransform("base_link", msg->header.frame_id, rclcpp::Time(0)), rot,
+    translation);
```
a-maumau: we should add try-catch here since it might fail.

perception/autoware_tensorrt_bevdet/src/bevdet_node.cpp
```diff
@@ -0,0 +1,235 @@
...
+  // Generate engine file name based on precision
+  std::string engine_file_base =
+    this->declare_parameter<std::string>("engine_path", "bevdet_one_lt_d");
+  engine_file_ = engine_file_base + (precision_ == "fp16" ? "_fp16.engine" : "_fp32.engine");
+
+  imgs_name_ = this->declare_parameter<std::vector<std::string>>("data_params.cams");
+  class_names_ =
+    this->declare_parameter<std::vector<std::string>>("post_process_params.class_names");
+
+  RCLCPP_INFO_STREAM(this->get_logger(), "Successful load config!");
```
a-maumau: Grammatically, it should be "Successfully loaded *".
Same thing for other same parts.

perception/autoware_tensorrt_bevdet/src/bevdet_node.cpp
```diff
@@ -0,0 +1,235 @@
...
+  imgs.emplace_back(img_br);
+
+  imageTransport(imgs, imgs_dev_, img_w_, img_h_);
+
+  // uchar *imgs_dev
+  sampleData_.imgs_dev = imgs_dev_;
+
+  std::vector<Box> ego_boxes;
+  ego_boxes.clear();
+  float time = 0.f;
```
a-maumau: I think this is for a processing time or something, so it should have more explicit name than a `time`.

perception/autoware_tensorrt_bevdet/src/bevdet_node.cpp
```diff
@@ -0,0 +1,235 @@
...
+  onnx_file_ = this->declare_parameter<std::string>("onnx_path", "bevdet_one_lt_d.onnx");
+
+  // Generate engine file name based on precision
+  std::string engine_file_base =
+    this->declare_parameter<std::string>("engine_path", "bevdet_one_lt_d");
+  engine_file_ = engine_file_base + (precision_ == "fp16" ? "_fp16.engine" : "_fp32.engine");
+
+  imgs_name_ = this->declare_parameter<std::vector<std::string>>("data_params.cams");
+  class_names_ =
+    this->declare_parameter<std::vector<std::string>>("post_process_params.class_names");
```
a-maumau: I prefer to have these in the constructor, since loading the ROS 2 config parameters isn't directly related to model initialization.

perception/autoware_tensorrt_bevdet/src/bevdet_node.cpp
```diff
@@ -0,0 +1,235 @@
...
+  cams2ego_rot_ = std::vector<Eigen::Quaternion<float>>(img_N_);
+  cams2ego_trans_ = std::vector<Eigen::Translation3f>(img_N_);
+
+  tf_buffer_ = std::make_unique<tf2_ros::Buffer>(this->get_clock());
+  tf_listener_ = std::make_shared<tf2_ros::TransformListener>(*tf_buffer_);
+
+  startCameraInfoSubscription();
+
+  // Wait for camera info and tf transform initialization
+  timer_ = this->create_wall_timer(
```
a-maumau: I also prefer this one to be more specific. not just "timer_".

perception/autoware_tensorrt_bevdet/include/autoware/tensorrt_bevdet/bevdet_node.hpp
```diff
@@ -0,0 +1,197 @@
...
+{
+class TRTBEVDetNode : public rclcpp::Node
+{
+  /**
+   * @class TRTBEVDetNode
+   * @brief This class represents a node for performing object detection using TensorRT on BEV
+   * (Bird's Eye View) images.
+   */
+private:
+  size_t img_N_;    ///< Number of images
```
a-maumau: I think it should be `img_n_` as same as `img_w_` or `img_h_`.

perception/autoware_tensorrt_bevdet/include/autoware/tensorrt_bevdet/bevdet_node.hpp
```diff
@@ -0,0 +1,197 @@
...
+
+  std::string onnx_file_;    ///< Path to the ONNX file
+  std::string engine_file_;  ///< Path to the TensorRT engine file
+
+  std::string precision_;  // Store precision setting
+
+  std::vector<std::string> imgs_name_;    ///< Names of the images
+  std::vector<std::string> class_names_;  ///< Names of the object classes
+
+  camsData sampleData_;             ///< Sample data for camera parameters
```
a-maumau: It is not a sample data, so it should be change to more better name.

perception/autoware_tensorrt_bevdet/include/autoware/tensorrt_bevdet/bevdet_node.hpp
```diff
@@ -0,0 +1,197 @@
...
+    sub_b_img_;  ///< Subscriber for back camera image
+  message_filters::Subscriber<sensor_msgs::msg::Image>
+    sub_bl_img_;  ///< Subscriber for back-left camera image
+  message_filters::Subscriber<sensor_msgs::msg::Image>
+    sub_br_img_;  ///< Subscriber for back-right camera image
+
+  typedef message_filters::sync_policies::ApproximateTime<
+    sensor_msgs::msg::Image, sensor_msgs::msg::Image, sensor_msgs::msg::Image,
+    sensor_msgs::msg::Image, sensor_msgs::msg::Image, sensor_msgs::msg::Image>
+    MySyncPolicy;
```
a-maumau: It may be obvious, but I prefer to use a name that describes what it does, rather than something like `MySyncPolicy`.

perception/autoware_tensorrt_bevdet/src/bevdet_node.cpp
```diff
@@ -0,0 +1,235 @@
...
+  imgs.emplace_back(img_b);
+  imgs.emplace_back(img_br);
+
+  imageTransport(imgs, imgs_dev_, img_w_, img_h_);
+
+  // uchar *imgs_dev
+  sampleData_.imgs_dev = imgs_dev_;
+
+  std::vector<Box> ego_boxes;
+  ego_boxes.clear();
```
a-maumau: It should be like `detected_boxes` or something, and I think we don't need the `clear()` here.

perception/autoware_tensorrt_bevdet/src/bevdet_node.cpp
```diff
@@ -0,0 +1,235 @@
...
+  const sensor_msgs::msg::Image::ConstSharedPtr & msg_br_img)
+{
+  cv::Mat img_fl, img_f, img_fr, img_bl, img_b, img_br;
+  std::vector<cv::Mat> imgs;
+  img_fl = cv_bridge::toCvShare(msg_fl_img, "bgr8")->image;
+  img_f = cv_bridge::toCvShare(msg_f_img, "bgr8")->image;
+  img_fr = cv_bridge::toCvShare(msg_fr_img, "bgr8")->image;
+  img_bl = cv_bridge::toCvShare(msg_bl_img, "bgr8")->image;
+  img_b = cv_bridge::toCvShare(msg_b_img, "bgr8")->image;
+  img_br = cv_bridge::toCvShare(msg_br_img, "bgr8")->image;
```
a-maumau: Are these ensured to be safe for in-place operation?

In imageTransport, there's a line: `cv::cvtColor(imgs[i], imgs[i], cv::COLOR_BGR2RGB);`
I **think** OpenCV’s cvtColor handles this safely by creating a new `cv::Mat` internally.
Do you know if this is safe to do? Or could you check whether it's okay to overwrite the input like this?"

Or, we can just create a copy of the images.

perception/autoware_tensorrt_bevdet/src/ros_utils.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+void box3DToDetectedObjects(
+  const std::vector<Box> & boxes, autoware_perception_msgs::msg::DetectedObjects & bevdet_objects,
+  const std::vector<std::string> & class_names, float score_thre, const bool has_twist = true)
+{
+  for (auto b : boxes) {
+    if (b.score < score_thre) continue;
+    autoware_perception_msgs::msg::DetectedObject obj;
+
+    Eigen::Vector3f center(b.x, b.y, b.z + b.h / 2.);
```
a-maumau: I think we don't need center. It is only used for
```
p.x = center.x();
p.y = center.y();
p.z = center.z();
```
we can replace the initialization with the values for center.
we can add a comment indicating that this is the center.

perception/autoware_tensorrt_bevdet/src/ros_utils.cpp
```diff
@@ -0,0 +1,140 @@
...
+void getCameraIntrinsics(
+  const sensor_msgs::msg::CameraInfo::SharedPtr msg, Eigen::Matrix3f & intrinsics)
+{
+  intrinsics << msg->k[0], msg->k[1], msg->k[2], msg->k[3], msg->k[4], msg->k[5], msg->k[6],
+    msg->k[7], msg->k[8];
+}
+
+void imageTransport(std::vector<cv::Mat> imgs, uchar * out_imgs, size_t width, size_t height)
+{
+  uchar * temp = new uchar[width * height * 3];
```
a-maumau: Since `temp` is not used, we can remove `temp` and the later `delete[] temp;`.

launch/tier4_perception_launch/launch/object_recognition/detection/detector/camera_bev_detector.launch.xml
```diff
@@ -0,0 +1,49 @@
...
+      <arg name="input/img_f" value="$(var input/camera1/image)"/>
+      <arg name="input/img_f/camera_info" value="$(var input/camera1/info)"/>
+      <arg name="input/img_fr" value="$(var input/camera2/image)"/>
+      <arg name="input/img_fr/camera_info" value="$(var input/camera2/info)"/>
+      <arg name="input/img_bl" value="$(var input/camera3/image)"/>
+      <arg name="input/img_bl/camera_info" value="$(var input/camera3/info)"/>
+      <arg name="input/img_b" value="$(var input/camera4/image)"/>
+      <arg name="input/img_b/camera_info" value="$(var input/camera4/info)"/>
+      <arg name="input/img_br" value="$(var input/camera5/image)"/>
+      <arg name="input/img_br/camera_info" value="$(var input/camera5/info)"/>
```
a-maumau: It may be obvious, but I prefer to use a name, like `input/img_front_left`, for these config-like things.

perception/autoware_tensorrt_bevdet/launch/tensorrt_bevdet.launch.xml
```diff
@@ -0,0 +1,53 @@
...
+  <arg name="input/img_b" default="/nuscenes/CAM_BACK/image"/>
+  <arg name="input/img_br" default="/nuscenes/CAM_BACK_RIGHT/image"/>
+  <arg name="output/boxes" default="/nuscenes/bboxes"/>
+
+  <arg name="input/img_fl/camera_info" default="/nuscenes/CAM_FRONT_LEFT/camera_info"/>
+  <arg name="input/img_f/camera_info" default="/nuscenes/CAM_FRONT/camera_info"/>
+  <arg name="input/img_fr/camera_info" default="/nuscenes/CAM_FRONT_RIGHT/camera_info"/>
+  <arg name="input/img_bl/camera_info" default="/nuscenes/CAM_BACK_LEFT/camera_info"/>
+  <arg name="input/img_b/camera_info" default="/nuscenes/CAM_BACK/camera_info"/>
+  <arg name="input/img_br/camera_info" default="/nuscenes/CAM_BACK_RIGHT/camera_info"/>
```
a-maumau: Default value should be renamed.
In the `launch/tier4_perception_launch/launch/object_recognition/detection/detection.launch.xml` of this PR, it is using `input/camera0/image` and so on.

perception/autoware_tensorrt_bevdet/src/bevdet_node.cpp
```diff
@@ -0,0 +1,235 @@
+// Copyright 2024 AutoCore, Inc.
```
a-maumau: We can update this to 2025. Also all other files that has the license too.

perception/autoware_tensorrt_bevdet/src/bevdet_node.cpp
```diff
@@ -0,0 +1,235 @@
...
+
+  sync_->registerCallback(std::bind(&TRTBEVDetNode::callback, this, _1, _2, _3, _4, _5, _6));
+}
+
+void TRTBEVDetNode::startCameraInfoSubscription()
+{
+  // cams: ["CAM_FRONT_LEFT", "CAM_FRONT", "CAM_FRONT_RIGHT", "CAM_BACK_LEFT", "CAM_BACK",
+  // "CAM_BACK_RIGHT"]
+  sub_fl_caminfo_ = this->create_subscription<sensor_msgs::msg::CameraInfo>(
+    "~/input/topic_img_fl/camera_info", rclcpp::QoS{1},
```
a-maumau: Could you change the QoS to something like "rclcpp::SensorDataQoS{}.keep_last(1)"?
I can't received the camera info within the autoware (it was possible with your ros2_dataset_bridge).

a-maumau (commented):
@RahulSundarMCW Hi, thank you for the contribution! I'm testing your code now.

I have a few points to mention.
Also, adding a schema file (like in the other repositories) and including it in the README would be a nice!

By the way, I need to change https://github.com/Owen-Liuyuxuan/ros2_dataset_bridge/blob/67c0286b1bd7a69247f40e43739208d3c967423b/ros2_dataset_bridge/utils/ros_util.py#L92 to `encoding="bgr8"` to run the code.

a-maumau:
@RahulSundarMCW

Also, did you test with other data inputs, particularly with different image sizes?
When I tested with other data with Autoware, I encountered an error like: `GPUassert: out of memory.`
I believe I have enough GPU memory (I can run the code with nuScenes data using your provided setup).
My image size is 1920×1080 for 6 cameras, but since there’s a resize step in BEVDet, so it shouldn’t matter. (Is this correct?)

I also received the following TensorRT message: `Error Code 3: API Usage Error (Parameter check failed at: runtime/api/executionContext.cpp::setInputShape::2295, condition: satisfyProfile. Runtime dimension does not satisfy any optimization profile.)`
I suspect this may be due to a shape mismatch causing the error (maybe related to OOM?).

If you have any insights into this, I’d appreciate your input!

perception/autoware_tensorrt_bevdet/src/bevdet_node.cpp
RahulSundarMCW: Thanks for pointing this out, directly modifying the image returned by cv_bridge::toCvShare(...)->image is unsafe because it references the ROS message's internal memory. To ensure safe in-place operations like cv::cvtColor(...), I’ve updated the code to call .clone() on each image, which creates a deep copy. This guarantees we're working with a private buffer and avoids any risk of modifying shared ROS message memory.

perception/autoware_tensorrt_bevdet/launch/tensorrt_bevdet.launch.xml
RahulSundarMCW: Sure, since we have primarily testing with nuScenes dataset, we use default value to launch node separately subscribing ros2_dataset_bridge topics.

RahulSundarMCW:
> @RahulSundarMCW
> 
> Also, did you test with other data inputs, particularly with different image sizes? When I tested with other data with Autoware, I encountered an error like: `GPUassert: out of memory.` I believe I have enough GPU memory (I can run the code with nuScenes data using your provided setup). My image size is 1920×1080 for 6 cameras, but since there’s a resize step in BEVDet, so it shouldn’t matter. (Is this correct?)
> 
> I also received the following TensorRT message: `Error Code 3: API Usage Error (Parameter check failed at: runtime/api/executionContext.cpp::setInputShape::2295, condition: satisfyProfile. Runtime dimension does not satisfy any optimization profile.)` I suspect this may be due to a shape mismatch causing the error (maybe related to OOM?).
> 
> If you have any insights into this, I’d appreciate your input!

@a-maumau 
Trying to reproduce this error, will get back shortly.

RahulSundarMCW:
> > @RahulSundarMCW
> > Also, did you test with other data inputs, particularly with different image sizes? When I tested with other data with Autoware, I encountered an error like: `GPUassert: out of memory.` I believe I have enough GPU memory (I can run the code with nuScenes data using your provided setup). My image size is 1920×1080 for 6 cameras, but since there’s a resize step in BEVDet, so it shouldn’t matter. (Is this correct?)
> > I also received the following TensorRT message: `Error Code 3: API Usage Error (Parameter check failed at: runtime/api/executionContext.cpp::setInputShape::2295, condition: satisfyProfile. Runtime dimension does not satisfy any optimization profile.)` I suspect this may be due to a shape mismatch causing the error (maybe related to OOM?).
> > If you have any insights into this, I’d appreciate your input!
> 
> @a-maumau Trying to reproduce this error, will get back shortly.

@a-maumau 
Thanks for pointing, we found and added the missing resize step and verified the implementation with the following input resolutions:

- 960×540
- 1280×720
- 1600×900
- 1920×1080
- 2560×1440

perception/autoware_tensorrt_bevdet/README.md
```diff
@@ -0,0 +1,155 @@
...
+- [bevdet_vendor](https://github.com/autowarefoundation/bevdet_vendor) package are copied from the [original codes](https://github.com/LCH1238/bevdet-tensorrt-cpp/tree/one) (The TensorRT, C++ implementation by LCH1238) and modified.
+- This package is ported version toward Autoware from [bevdet_vendor](https://github.com/autowarefoundation/bevdet_vendor).
+
+## Inputs / Outputs
+
+### Inputs
+
+| Name                               | Type                           | Description                         |
+| ---------------------------------- | ------------------------------ | ----------------------------------- |
+| `~/input/topic_img_fl`             | `sensor_msgs::msg::Image`      | input front_left camera image       |
```
a-maumau: We should change this topic name as same as the launch file: `~/input/img_front_left`.
As same for other ones.

perception/autoware_tensorrt_bevdet/include/autoware/tensorrt_bevdet/bevdet_node.hpp
```diff
@@ -0,0 +1,204 @@
...
+#include "autoware/tensorrt_bevdet/marker_utils.hpp"
+#include "autoware/tensorrt_bevdet/ros_utils.hpp"
+#include "tf2_ros/buffer.h"
+#include "tf2_ros/transform_listener.h"
+
+#include <Eigen/Core>
+#include <Eigen/Geometry>
+
+#include <sensor_msgs/msg/image.hpp>
+
```
a-maumau, line 29: could you add `#include <autoware_perception_msgs/msg/detected_objects.hpp>` here too?
it may included from `ros_utils.hpp`, it would be better to explicitly include it.

perception/autoware_tensorrt_bevdet/launch/tensorrt_bevdet.launch.xml
```diff
@@ -0,0 +1,58 @@
...
+
+  <arg name="data_path" default="$(env HOME)/autoware_data"/>
+  <arg name="model_name" default="bevdet_one_lt_d"/>
+  <arg name="model_path" default="$(var data_path)/tensorrt_bevdet"/>
+  <arg name="model_config" default="$(find-pkg-share autoware_tensorrt_bevdet)/config/bevdet_r50_4dlongterm_depth.yaml"/>
+  <arg name="param_file" default="$(find-pkg-share autoware_tensorrt_bevdet)/config/bevdet.param.yaml"/>
+
+  <arg name="precision" default="fp16" description="Precision mode (fp16 or fp32)"/>
+  <!-- supports visualization with nuScenes when enabled -->
+  <arg name="debug" default="false" description="Enable debug mode"/>
```
a-maumau: I think it should be fine to set the debug flag in launch file, but other packages do this in the config file, so I prefer to move the `debug` into the `bevdet.param.yaml`

perception/autoware_tensorrt_bevdet/package.xml
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0"?>
+<package format="3">
+  <name>autoware_tensorrt_bevdet</name>
+  <version>0.0.1</version>
```
a-maumau: I am not sure about this, but recently Autoware is doing versioning for the whole packages, we should set this to 0.46.0 (?).

perception/autoware_tensorrt_bevdet/src/bevdet_node.cpp
```diff
@@ -0,0 +1,261 @@
...
+
+  CHECK_CUDA(cudaMalloc(
+    reinterpret_cast<void **>(&imgs_dev_), img_n_ * 3 * img_w_ * img_h_ * sizeof(uchar)));
+
+  pub_boxes_ = this->create_publisher<autoware_perception_msgs::msg::DetectedObjects>(
+    "~/output/boxes", rclcpp::SensorDataQoS{}.keep_last(1));
+  if (debug_mode_) {
+    pub_markers_ = this->create_publisher<visualization_msgs::msg::MarkerArray>(
+      "~/output_bboxes", rclcpp::QoS{1});
+  }
```
a-maumau: Could you change the `pub_boxes_`'s QoS to `rclcpp::QoS{1}`?
And this code also not directly related to the `initModel`, it should move to other places like in the constructor.

perception/autoware_tensorrt_bevdet/src/bevdet_node.cpp
```diff
@@ -0,0 +1,261 @@
...
+  if (caminfo_received_[idx])
+    return;  // already received;  not expected to modify because of we init the model only once
+
+  if (!initialized_) {  // load image width and height from model config YAML
+    YAML::Node config = YAML::LoadFile(model_config_);
+    auto src_size = config["data_config"]["src_size"];
+    img_h_ = src_size[0].as<size_t>();  // height
+    img_w_ = src_size[1].as<size_t>();  // width
+    initialized_ = true;
+  }
```
a-maumau: I think this part only runs once and the initialization does not require the camera info. So we can just move this into constructor.

perception/autoware_tensorrt_bevdet/src/bevdet_node.cpp
```diff
@@ -0,0 +1,261 @@
...
+  const sensor_msgs::msg::Image::ConstSharedPtr & msg_br_img)
+{
+  std::vector<cv::Mat> imgs;
+  auto clone_and_resize = [&](const sensor_msgs::msg::Image::ConstSharedPtr & msg) -> cv::Mat {
+    cv::Mat img = cv_bridge::toCvShare(msg, "bgr8")->image.clone();
+    if (img.size() != cv::Size(img_w_, img_h_)) {
+      cv::resize(img, img, cv::Size(img_w_, img_h_));  // img_w_ = 1600, img_h_ = 900
+    }
+    return img;
+  };
```
a-maumau: If there is no specific reason to do this, then we can make this expression into a method.

perception/autoware_tensorrt_bevdet/src/ros_utils.cpp
```diff
@@ -0,0 +1,137 @@
...
+  } else if (class_name == "pedestrian") {
+    return Label::PEDESTRIAN;
+  } else {
+    return Label::UNKNOWN;
+  }
+}
+
+void box3DToDetectedObjects(
+  const std::vector<Box> & boxes, autoware_perception_msgs::msg::DetectedObjects & bevdet_objects,
+  const std::vector<std::string> & class_names, float score_thre, const bool has_twist = true)
```
a-maumau: we can make `float score_thre` to `const float & score_thre`.

perception/autoware_tensorrt_bevdet/src/ros_utils.cpp
```diff
@@ -0,0 +1,137 @@
...
+{
+  for (auto b : boxes) {
```
a-maumau: we can make `auto b : boxes` to `const auto & b : boxes`.

a-maumau (commented):
Thank you for fixing the points I have pointed out. I have few more points to mention.

a-maumau:
@RahulSundarMCW
> Thanks for pointing, we found and added the missing resize step and verified the implementation with the following input resolutions:
960×540
1280×720
1600×900
1920×1080
2560×1440

Thank you so much for checking the inputs!

It seems to have resolved the issue I encountered.  
However, the error "Runtime dimension does not satisfy any optimization profile." still occurs, even with the fixed image.

I need to update the profile parameters in  https://github.com/autowarefoundation/bevdet_vendor/blob/2dd62dcd62bb455e393c48e5ab79d81ec3f3489a/src/bevdet.cpp#L636-L648.
So, I think we need to make those runtime-dynamic parameters configurable via parameter file for general use.
(This is beyond this PR, though)

a-maumau:
@RahulSundarMCW 
By the way, what is the debug (`pub_markers_`) used for?
I check the output, but it seems as same as the normal output with different color and polygon?

perception/autoware_tensorrt_bevdet/package.xml
RahulSundarMCW: Yes, I have looked few recent merged packages (autoware_bevfusion, ..) where they have been using this 0.46.0 and changed accordingly in bevdet package.

RahulSundarMCW:
> @RahulSundarMCW By the way, what is the debug (`pub_markers_`) used for? I check the output, but it seems as same as the normal output with different color and polygon?

Since this model was trained and evaluated on the nuScenes dataset, @xmfcx has requested that visualization support be added specifically for nuScenes data when running in debug mode. The **pub_marker** publisher uses the visualization_msgs::msg::MarkerArray message type, whereas **pub_boxes** uses the autoware_perception_msgs::msg::DetectedObjects message type.

RahulSundarMCW:
> @RahulSundarMCW
> 
> > Thanks for pointing, we found and added the missing resize step and verified the implementation with the following input resolutions:
> > 960×540
> > 1280×720
> > 1600×900
> > 1920×1080
> > 2560×1440
> 
> Thank you so much for checking the inputs!
> 
> It seems to have resolved the issue I encountered. However, the error "Runtime dimension does not satisfy any optimization profile." still occurs, even with the fixed image.
> 
> I need to update the profile parameters in https://github.com/autowarefoundation/bevdet_vendor/blob/2dd62dcd62bb455e393c48e5ab79d81ec3f3489a/src/bevdet.cpp#L636-L648. So, I think we need to make those runtime-dynamic parameters configurable via parameter file for general use. (This is beyond this PR, though)

Thanks @a-maumau  for pointing out the root cause.
Since it's beyond the scope of this PR, may I suggest we proceed with merging this PR if everything looks good as it stands?
@xmfcx @mitsudome-r

a-maumau:
I checked with TIER4 internal data, and confirmed the package runs.
Other than the CI failing, from my perspective, LGTM.

@RahulSundarMCW could you fix the failing CIs?

RahulSundarMCW:
> I checked with TIER4 internal data, and confirmed the package runs. Other than the CI failing, from my perspective, LGTM.
> 
> @RahulSundarMCW could you fix the failing CIs?

Hi @a-maumau 
Fixed all possible failing CIs from our end,

1. **build-and-test-differential-cuda** : This failure is expected due to missing API compatibility. The build depends on updated support for TensorRT 10.x, which is not yet available in [bevdet_vendor](https://github.com/autowarefoundation/bevdet_vendor/tree/bevdet_vendor-ros2). A separate [PR](https://github.com/autowarefoundation/bevdet_vendor/pull/1) is currently pending that will introduce the required compatibility.
2. **CodeScene Cloud Delta Analysis** : Since our callback explicitly requires all 6 camera inputs, the threshold of 4 is insufficient. Could we consider suppressing it?

@xmfcx @mitsudome-r

a-maumau:
@RahulSundarMCW thank you for fixing the CI fails.
Could you also fix the spell check?
I think some of the words like `bevdet` need to be ignored and others might need to be renamed.
We also can update the spell check dictionary for a word like "bevdet": https://github.com/autowarefoundation/autoware-spell-check-dict

For `CodeScene Cloud Delta Analysis`, I agree with you.
It is not marked as "Required", so it can be ignored for now (many PRs actually ignore this).

For
> build-and-test-differential-cuda : This failure is expected due to missing API compatibility. The build depends on updated support for TensorRT 10.x, which is not yet available in [bevdet_vendor](https://github.com/autowarefoundation/bevdet_vendor/tree/bevdet_vendor-ros2). A separate https://github.com/autowarefoundation/bevdet_vendor/pull/1 is currently pending that will introduce the required compatibility.

, if there is no cyclic dependency (which I believe there isn't), we should follow the intended order: merge the PR in `bevdet_vendor` first, then this PR.
I know the PR for `bevdet_vendor` has been inactive for a few months, but this approach is better and should be followed.

RahulSundarMCW:
> @RahulSundarMCW thank you for fixing the CI fails. Could you also fix the spell check? I think some of the words like `bevdet` need to be ignored and others might need to be renamed. We also can update the spell check dictionary for a word like "bevdet": https://github.com/autowarefoundation/autoware-spell-check-dict
> 
> For `CodeScene Cloud Delta Analysis`, I agree with you. It is not marked as "Required", so it can be ignored for now (many PRs actually ignore this).
> 
> For
> 
> > build-and-test-differential-cuda : This failure is expected due to missing API compatibility. The build depends on updated support for TensorRT 10.x, which is not yet available in [bevdet_vendor](https://github.com/autowarefoundation/bevdet_vendor/tree/bevdet_vendor-ros2). A separate [autowarefoundation/bevdet_vendor#1](https://github.com/autowarefoundation/bevdet_vendor/pull/1) is currently pending that will introduce the required compatibility.
> 
> , if there is no cyclic dependency (which I believe there isn't), we should follow the intended order: merge the PR in `bevdet_vendor` first, then this PR. I know the PR for `bevdet_vendor` has been inactive for a few months, but this approach is better and should be followed.

Thanks @a-maumau 

Hi @xmfcx @mitsudome-r 
Could you please help on this by reviewing and merging the [feat/api_support_TRT_10.x.x](https://github.com/autowarefoundation/bevdet_vendor/pull/1) PR we have raised to **bevdet_vendor** which will help in merging this PR.

xmfcx:
- Fixed the spell check with https://github.com/autowarefoundation/autoware-spell-check-dict/pull/41
- https://github.com/autowarefoundation/bevdet_vendor/pull/1 is merged ✅
- Trying to merge the following but there is some issue with the docker build CI (not related to the PR):
  - https://github.com/autowarefoundation/autoware/pull/5456
  - https://github.com/autowarefoundation/autoware/pull/6138

RahulSundarMCW:
Hi @xmfcx  @mitsudome-r @a-maumau 
Please let us know if there is any support required from our end to merge this PR.

xmfcx:
## Updates

- https://github.com/autowarefoundation/autoware_universe/pull/10441/commits/6723ece4db13d36cd5837cb5c9c65d148e2f6190
  - Renamed `/nuscenes/bboxes` topic in launch to `/debug/markers_boxes` to avoid the clash between the dataset topics.
- https://github.com/autowarefoundation/autoware_universe/pull/10441/commits/6c19f35a52978918e9c1871c9c1c3d806ec2ad0a
  - Fixed the rviz markerarray issue, there were 2 markers with id:0
- https://github.com/autowarefoundation/autoware_universe/pull/10441/commits/563bc0628e7d6b7bac516f95ebf1a0e1a1765a5e
  - Renamed parameter `debug` to `debug_mode`
  - Removed the defaults for the parameters from the source code.
    - https://autowarefoundation.github.io/autoware-documentation/main/contributing/coding-guidelines/ros-nodes/parameters/#workflow see here for the rationale
  - `precision` param reference is removed from the launch file, the param file should be used instead.
- https://github.com/autowarefoundation/autoware_universe/pull/10441/commits/1f83b3acd1a4d5eca74ff8a2370f100b496cf206
  - Download instructions are updated to point to ansible
- https://github.com/autowarefoundation/autoware_universe/pull/10441/commits/04e1fee415ea78379f0dbe1e50574a65ab6ea5b0
  - `CAM_NUM` renamed to `camera_count` for consistency
- https://github.com/autowarefoundation/autoware_universe/pull/10441/commits/283401dffe02dfe693e7a25a73076d0fbc2e1344
  - Spellcheck fixes are finalized.

perception/autoware_tensorrt_bevdet/config/bevdet_r50_4dlongterm_depth.yaml
```diff
@@ -0,0 +1,65 @@
...
+  Ncams: 6
+  cams: [CAM_FRONT_LEFT, CAM_FRONT, CAM_FRONT_RIGHT, CAM_BACK_LEFT, CAM_BACK, CAM_BACK_RIGHT]
+  resize_radio: 0.44
+  crop: [140, 0]
+  flip: true
+  input_size: [256, 704]
+  resize: [-0.06, 0.11]
+  resize_test: 0.0
+  rot: [-5.4, 5.4]
+  src_size: [900, 1600]
```
xmfcx, line 22: Apart from `data_config.src_size` none of the other fields in this config file are used anywhere, is this expected? Is this file auto-generated?

xmfcx (approved):
There are many things to be fixed to conform to https://autowarefoundation.github.io/autoware-documentation/main/contributing/coding-guidelines/ such as snake_case for function names.
But for the first functional prototype it is alright.

**My Test Video:** https://www.youtube.com/watch?v=4_ZgcpOuySQ

perception/autoware_tensorrt_bevdet/config/bevdet_r50_4dlongterm_depth.yaml
mitsudome-r: We have decided to merge the PR first and merge this src_size into bevdet.param.yaml as another PR by @RahulSundarMCW

RahulSundarMCW: @xmfcx @mitsudome-r 
We are using the config file in two places, one is for `data_config.src_size` and we are also passing it as an argument to **bevdet_vendor** where the fields of this config file are being used.
So there won't require another PR.

PR #10455, opened 2025-04-15
feat(diagnostic_graph_aggregator): diagnostic graph test interface

isamu-takagi:
## Description

Add test interface to validate your own diagnostic graph settings.

## Related links

**Private Links:**

- [TIER IV internal link](https://tier4.atlassian.net/browse/RT0-36084)

## How was this PR tested?

See autoware_diagnostic_graph_test_examples package.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


examples/autoware_diagnostic_graph_test_examples/test/example.cpp
```diff
@@ -0,0 +1,84 @@
...
+    diags.status.push_back(status);
+  }
+
+  const auto result = graph.test(diags);
+  std::unordered_map<std::string, DiagnosticLevel> output;
+  for (const auto & status : result) {
+    output[status.path] = status.level;
+  }
+  return output;
+}
```
sasakisasaki: Perhaps it might be better to wrap with namespace `autoware::diagnostic_graph_aggregator`

isamu-takagi: For what reason? This code is not referenced by other packages.

sasakisasaki: You are right, I was misunderstanding.

sasakisasaki (approved):
I agree with merging this PR

PR #10456, opened 2025-04-15
feat(multi object tracker): tracker overlap threshold

technolojin:
## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


perception/autoware_multi_object_tracker/src/multi_object_tracker_node.cpp
```diff
@@ -188,19 +188,34 @@ MultiObjectTracker::MultiObjectTracker(const rclcpp::NodeOptions & node_options)
 
     AssociatorConfig associator_config;
     {
-      const auto tmp = this->declare_parameter<std::vector<int64_t>>("can_assign_matrix");
-      const std::vector<int> can_assign_matrix(tmp.begin(), tmp.end());
-      associator_config.can_assign_matrix = can_assign_matrix;
+      auto initializeMatrixInt = [](const std::vector<int64_t> & vector) {
+        const int label_num = static_cast<int>(std::sqrt(vector.size()));
+        std::vector<int> converted_vector(vector.begin(), vector.end());
```
YoshiRi, line 193: [ minor comment]
`int_64` to `int` conversion may cause warning. IMO we do not need 64 bit to express this field. Anyway, it can be fixed within another PR.

perception/autoware_multi_object_tracker/src/multi_object_tracker_node.cpp
```diff
@@ -188,19 +188,34 @@ MultiObjectTracker::MultiObjectTracker(const rclcpp::NodeOptions & node_options)
...
+        Eigen::Map<Eigen::MatrixXi> matrix_tmp(converted_vector.data(), label_num, label_num);
+        // transpose to make it row-major
+        return matrix_tmp.transpose();
+      };
+      auto initializeMatrixDouble = [](const std::vector<double> & vector) {
+        const int label_num = static_cast<int>(std::sqrt(vector.size()));
+        Eigen::Map<Eigen::MatrixXd> matrix_tmp(
+          const_cast<double *>(vector.data()), label_num, label_num);
```
YoshiRi, line 201: How about just return const and not using const_cast? Because input is const and we do not intend to modify input in this conversion.

```suggestion
        Eigen::Map<const Eigen::MatrixXd> matrix_map( 
            vector.data(), label_num, label_num
        );
```

YoshiRi (changes requested):

perception/autoware_multi_object_tracker/src/multi_object_tracker_node.cpp
technolojin: In the ROS2 parameter, integer array is only for `long int`

https://docs.ros2.org/foxy/api/rclcpp/classrclcpp_1_1Parameter.html#af7575c8b1c49c8c4756a66d4cad3347f

perception/autoware_multi_object_tracker/src/multi_object_tracker_node.cpp
technolojin: fixed https://github.com/autowarefoundation/autoware_universe/pull/10456/commits/aa85148a584392de201db453957bd36ee39eb2ea

PR #10494, opened 2025-04-17
fix(autoware_gyro_odometer): fix deprecated autoware_utils header

a-maumau:
## Description
This PR fixes the include headers of `autoware_utils` to follow the current package style (`autoware_utils_*`) for `gyro_odometer` package.

## Related links

**Parent Issue:**

- https://github.com/autowarefoundation/autoware_universe/issues/10474
  - https://github.com/autowarefoundation/autoware_universe/issues/10482

## How was this PR tested?
Checked the behavior with the sample rosbag and confirmed it works as expected.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/localization/autoware_gyro_odometer/package.xml
+++ b/localization/autoware_gyro_odometer/package.xml
@@ -18,7 +18,10 @@
   <buildtool_depend>autoware_cmake</buildtool_depend>
 
   <depend>autoware_localization_util</depend>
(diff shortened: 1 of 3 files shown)


RyuYamamoto (approved):
Confirmed that it works with logging_simulator.

LGTM

PR #10498, opened 2025-04-17
fix(localization_error_monitor): fix deprecated autoware_utils header

a-maumau:
## Description
This PR fixes the include headers of `autoware_utils` to follow the current package style (`autoware_utils_*`) for `localization_error_monitor` package.

## Related links

**Parent Issue:**

- https://github.com/autowarefoundation/autoware_universe/issues/10474
  - https://github.com/autowarefoundation/autoware_universe/issues/10486

## How was this PR tested?
Checked the behavior with the sample rosbag and confirmed it works as expected.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


RyuYamamoto (approved):
Confirmed that it works with logging_simulator.

LGTM

PR #10499, opened 2025-04-17
fix(autoware_image_projection_based_fusion): add missing params `common_param_path` when not use container

MasatoSaeki:
## Description

This PR adds missing parameter `common_param_path` to use pointpainting when not use container. Caused by https://github.com/autowarefoundation/autoware_universe/pull/10397

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

https://evaluation.ci.tier4.jp/evaluation/reports/5968877a-f3ae-52ea-bce3-aaeeac420201?project_id=prd_jt

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


ktro2828 (approved):
LGTM

PR #10502, opened 2025-04-18
ci: fix conditions of prepare build_depends.repos file (main branch)

rej55:
## Description

Current implementation of build testing workflow has hard coding conditions.
To be robust with branch name changing, I fixed the conditions in workflow files.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

https://github.com/autowarefoundation/autoware_universe/actions/runs/14529930679/job/40767929231?pr=10502

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/.github/workflows/build-and-test-differential-arm64.yaml
+++ b/.github/workflows/build-and-test-differential-arm64.yaml
@@ -68,7 +68,7 @@ jobs:
         shell: bash
 
       - name: Prepare build_depends.repos file (main branch)
-        if: ${{ github.event.pull_request.base.ref == 'main' }}
+        if: ${{ github.event.pull_request.base.ref != 'humble' }}
         uses: ./.github/actions/combine-repos-action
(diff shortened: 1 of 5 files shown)


mitsudome-r (approved):

xmfcx:
@rej55 github.event.pull_request.base.ref refers to the target branch of the PR. Could you provide an example where you needed a PR against neither main or humble?

rej55:
@xmfcx First, we found the problem in our forked repository https://github.com/tier4/autoware_universe/actions/runs/14509323423/job/40704472423#step:15:68 
(We use some maintained branches except for `main` in our repository)

This is just our problem, so we should handle with this problem in our repository only, however, I think this change may be useful for users who use forked repository.

xmfcx:
Alright, no problems 👌

PR #10506, opened 2025-04-20
fix(autoware_system_monitor): quick fix for autoware_system_monitor

liuXinGangChina:
fix sompile issue

## Description
compile using colcon with latest autoware main branch code, i got an error like below

Screenshot from 2025-04-20 15-10-33

it seems that the author want to set value '\0' to the last element of the vector to use it as a terminator
i look through the code make sure there is null pointor protection code serveral lines before, so i modify the last element access method by a tiny fix

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

1. using AWF official docker
2. clone the latest main branch code
3. compile the entier autoware with " colcon build --cmake-args -DCMAKE_BUILD_TYPE=Release --symlink-install"
4. no error occours any more

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/system/autoware_system_monitor/src/process_monitor/process_monitor.cpp
+++ b/system/autoware_system_monitor/src/process_monitor/process_monitor.cpp
@@ -567,7 +567,7 @@ bool ProcessMonitor::getCommandLineFromPid(const std::string & pid, std::string 
   // Leave the last 0x00 (end-of-C-string) intact.
   std::replace(buffer.begin(), (buffer.end() - 1), '\0', ' ');
   // Make sure the last character is a null terminator.
-  *(buffer.end() - 1) = '\0';
+  buffer[buffer.size() - 1] = '\0';
   try {
     std::string cmdline = std::string(buffer.begin(), (buffer.end() - 1));
     // Remove trailing spaces
--- a/system/autoware_system_monitor/src/process_monitor/process_monitor.cpp
+++ b/system/autoware_system_monitor/src/process_monitor/process_monitor.cpp
@@ -567,6 +567,10 @@ bool ProcessMonitor::getCommandLineFromPid(const std::string & pid, std::string 
   // Leave the last 0x00 (end-of-C-string) intact.
   std::replace(buffer.begin(), (buffer.end() - 1), '\0', ' ');
   // Make sure the last character is a null terminator.
+
+  // Although buffer.back() is safer and more idiomatic,
+  // using it here causes a compile error with -Werror=stringop-overflow.
+  // Therefore, we safely access the last element by index.
   buffer[buffer.size() - 1] = '\0';
   try {
     std::string cmdline = std::string(buffer.begin(), (buffer.end() - 1));


sasakisasaki:
Thank you for fixing the issue. Yes, I certainly observed the same issue. Now I'm testing on my local machine by `colcon build`

system/autoware_system_monitor/src/process_monitor/process_monitor.cpp
```diff
@@ -567,7 +567,7 @@ bool ProcessMonitor::getCommandLineFromPid(const std::string & pid, std::string
   // Leave the last 0x00 (end-of-C-string) intact.
   std::replace(buffer.begin(), (buffer.end() - 1), '\0', ' ');
   // Make sure the last character is a null terminator.
-  *(buffer.end() - 1) = '\0';
+  buffer[buffer.size() - 1] = '\0';
```
sasakisasaki, line 570:
```suggestion

  // Although buffer.back() is safer and more idiomatic,
  // using it here causes a compile error with -Werror=stringop-overflow.
  // Therefore, we safely access the last element by index.
  buffer[buffer.size() - 1] = '\0';
```
Perhaps

sasakisasaki (changes requested):
### Reason of this Proposal

Perhaps including a comment would be helpful for understandability :+1: 

I understand using the `buffer.size()` is safer, but it seems the compile option `-Werror=stringop-overflow` which is enabled on the CI cannot determine if the `buffer` is empty or not even though the sanity check is performed at line 563-565:
```
  if (buffer.size() == 0) {  // cmdline is empty if it is a kernel process
    return false;
  }
```

Also, since we can't use #pragma (like the one below) in our CI environment, I agree with your proposed fix. **Let's include a comment explaining the reason to improve understandability. :+1:**
```
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wstringop-overflow"
// buffer is checked for emptiness above, safe to call back()
buffer.back() = '\0';
#pragma GCC diagnostic pop
```

system/autoware_system_monitor/src/process_monitor/process_monitor.cpp
liuXinGangChina: Hi sasaki san @sasakisasaki 

Thanks for your reviewing

I have modified the pr according to your comment

Best regards

心刚

sasakisasaki (approved):
### Reason for Approval

From my side, the proposal looks good. I've tried a build it on my local machine and verified the build succeeded.

TetsuKawa (approved):
LGTM!
I have confirmed that it builds successfully on my local machine as well

nishikawa-masaki:
@liuXinGangChina 
Thank you very much for your quick fix. I feel sorry for any inconvenience caused by the change I made.
I see no problem with your change.
I approve this PR.

nishikawa-masaki (approved):

PR #10508, opened 2025-04-21
fix(control_validator): fix less trajectory point check

yuki-takagi-66:
## Description
This PR addresses the following issue:
Currently, the control_validator cancels the validation process if the point size of predicted_trajectory is less than two. This causes longitudinal validations to be canceled when the ego vehicle overruns the goal pose, even though such situations are critical.

In addition, to reduce maintenance costs, `calcSignedArcLength()` in the `dist_to_stop` calculation is now used correctly.

## How was this PR tested?
psim:
error diags are published when the ego overrun the goal.

tier4 scenario tests https://evaluation.tier4.jp/evaluation/reports/18d7c46d-2b9d-56fb-8c73-ab3ea47f996d?project_id=prd_jt

tier4 DLR tests https://evaluation.tier4.jp/evaluation/reports/75697820-226d-564f-81ad-34bc94c764ca?project_id=prd_jt

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/control/autoware_control_validator/src/control_validator.cpp
+++ b/control/autoware_control_validator/src/control_validator.cpp
@@ -102,21 +102,11 @@ void OverrunValidator::validate(
   ControlValidatorStatus & res, const Trajectory & reference_trajectory,
   const Odometry & kinematics)
 {
-  res.dist_to_stop = [](const Trajectory & traj, const geometry_msgs::msg::Pose & pose) {
-    const auto stop_idx_opt = autoware::motion_utils::searchZeroVelocityIndex(traj.points);
-
-    const size_t end_idx = stop_idx_opt ? *stop_idx_opt : traj.points.size() - 1;
-    const size_t seg_idx =
-      autoware::motion_utils::findFirstNearestSegmentIndexWithSoftConstraints(traj.points, pose);
-    const double signed_length_on_traj = autoware::motion_utils::calcSignedArcLength(
-      traj.points, pose.position, seg_idx, traj.points.at(end_idx).pose.position,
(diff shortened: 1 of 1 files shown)


control/autoware_control_validator/src/control_validator.cpp
```diff
@@ -102,21 +102,11 @@ void OverrunValidator::validate(
...
-    if (std::isnan(signed_length_on_traj)) {
-      return 0.0;
-    }
-    return signed_length_on_traj;
-  }(reference_trajectory, kinematics.pose.pose);
+  const auto stop_idx_opt =
+    autoware::motion_utils::searchZeroVelocityIndex(reference_trajectory.points);
+  const size_t end_idx = stop_idx_opt ? *stop_idx_opt : reference_trajectory.points.size() - 1;
+  res.dist_to_stop = autoware::motion_utils::calcSignedArcLength(
+    reference_trajectory.points, kinematics.pose.pose.position, end_idx);
```
kyoichi-sugahara, line 109: @yuki-takagi-66 
It appears that the code could cause an out-of-bounds access if the reference_trajectory size is 0. 
Is the assumption that this is acceptable because size checking is done in the `on_control_cmd` function?

To prevent issues like this time: 
where a validate function might unexpectedly not be called when adding new features

We should modify each validate function to include necessary checks internally? This would make the functions more robust.
If addressing this in the current PR is challenging due to time constraints, please add a TODO note describing the approach for future fixes.

kyoichi-sugahara (approved):

PR #10523, opened 2025-04-21
perf(cuda_sync_blocking)

primenumber:
## Description

Depending on the type of GPU, busy waiting may be used when synchronizing with a CUDA device (cudaStreamSynchronize, etc.). (cf. [CUDA Runtime API :: CUDA Toolkit Documentation](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__DEVICE.html#group__CUDART__DEVICE_1g69e73c7dda3fc05306ae7c811a690fac))
In that case, a CPU core will be occupied while waiting for the device.
In this pull request, we use the CUDA Runtime API to configure it to block the CPU thread while waiting for the device.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

Please let me know if you know of any other nodes that use CUDA.

## Interface changes

None.

## Effects on system behavior

This change will slightly increase the latency when synchronizing with the device.
Since `cudaSetDeviceFlags` affects the entire process, it may affect the behavior of other nodes in the same component.

### DIFF
--- a/perception/autoware_bevfusion/src/bevfusion_node.cpp
+++ b/perception/autoware_bevfusion/src/bevfusion_node.cpp
@@ -28,6 +28,10 @@ namespace autoware::bevfusion
 BEVFusionNode::BEVFusionNode(const rclcpp::NodeOptions & options)
 : Node("bevfusion", options), tf_buffer_(this->get_clock())
 {
+  // Set CUDA device flags
+  // note: Device flags are process-wide
+  cudaSetDeviceFlags(cudaDeviceScheduleBlockingSync);
+
   auto descriptor = rcl_interfaces::msg::ParameterDescriptor{}.set__read_only(true);
 
   // Modality
--- a/perception/autoware_image_projection_based_fusion/src/pointpainting_fusion/node.cpp
+++ b/perception/autoware_image_projection_based_fusion/src/pointpainting_fusion/node.cpp
@@ -95,6 +95,10 @@ inline bool isUnknown(int label2d)
 PointPaintingFusionNode::PointPaintingFusionNode(const rclcpp::NodeOptions & options)
 : FusionNode<PointCloudMsgType, RoiMsgType, DetectedObjects>("pointpainting_fusion", options)
 {
+  // Set CUDA device flags
+  // note: Device flags are process-wide
+  cudaSetDeviceFlags(cudaDeviceScheduleBlockingSync);
+
   omp_num_threads_ = this->declare_parameter<int>("omp_params.num_threads");
   const float score_threshold =
     static_cast<float>(this->declare_parameter<double>("post_process_params.score_threshold"));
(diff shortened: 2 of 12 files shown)


amadeuszsz:
@primenumber 

Thanks for your PR!

Is there a reason why we don't have to keep default, "auto" behavior? As I understood, in default configuration, CUDA set `cudaDeviceScheduleBlockingSync` for low-powered devices. Isn't sufficient for us? 

If latency is negligible, let's say less than few % change of total processing time, personally I think we can merge it without deeper investigation (latency plot comparison would be very welcome). Additionaly, we don't need to apply it to each separate node if I'm not mistaken. Since all TensorRT-based packages uses `autoware_tensorrt_common`, can we just add `cudaSetDeviceFlags(cudaDeviceScheduleBlockingSync)` [here](http://github.com/autowarefoundation/autoware_universe/blob/f0cb1e5113357dd55401e094ef9cc0a0c1f1c44d/perception/autoware_tensorrt_common/src/tensorrt_common.cpp#L70)?

primenumber:
@amadeuszsz Thank you for your comment!


> Is there a reason why we don't have to keep default, "auto" behavior? As I understood, in default configuration, CUDA set `cudaDeviceScheduleBlockingSync` for low-powered devices. Isn't sufficient for us?

It also helps to make better use of CPU time, so I think it's worth enabling on all devices.

> Since all TensorRT-based packages uses autoware_tensorrt_common, can we just add cudaSetDeviceFlags(cudaDeviceScheduleBlockingSync) [here](https://github.com/autowarefoundation/autoware_universe/blob/f0cb1e5113357dd55401e094ef9cc0a0c1f1c44d/perception/autoware_tensorrt_common/src/tensorrt_common.cpp#L70)?

That's very cool! I'm going to use it except for `cuda_pointcloud_preprocessor_node`, which doesn't use TensorRT.

## Latency comparison

These are the results of measurements taken with Nsight Systems when running [Rosbag replay simulation](https://autowarefoundation.github.io/autoware-documentation/main/tutorials/ad-hoc-simulation/rosbag-replay-simulation/) on my machine (i7-13700H + RTX 4060 Laptop).

Before 

Screenshot from 2025-04-22 11-34-01

After

Screenshot from 2025-04-22 11-34-08

We can see that `sem_wait` is called during `cudaStreamSynchronize` to release CPU resources.
The average latency of `cudaStreamSynchronize` remains almost unchanged, while the median appears to have increased by about 3us. I think this value is small enough.

amadeuszsz:
@primenumber 
Thanks for explanation and detailed summary!

> That's very cool! I'm going to use it except for `cuda_pointcloud_preprocessor_node`, which doesn't use TensorRT.

Great! If you undo your changes and apply flag in `autoware_tensorrt_common`, I can approve this PR.

I confirm it shouldn't break older devices with e.g. JetPack 5.1.2 (CUDA 11.4) - API is [same](https://docs.nvidia.com/cuda/archive/11.4.0/cuda-runtime-api/group__CUDART__DEVICE.html#group__CUDART__DEVICE_1g69e73c7dda3fc05306ae7c811a690fac) and this flag is already implemented.

PR #10529, opened 2025-04-22
fix(tensorrt_plugins): remove unused function logInfo

veqcc:
## Description

Based on cppcheck `unusedFunction` warning
```
perception/autoware_tensorrt_plugins/src/plugin_utils.cpp:33:0: style: The function 'logInfo' is never used. [unusedFunction]
void logInfo(char const * msg)
^
```

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


knzo25 (approved):
LGTM (will reintroduce if necessary tho)

sasakisasaki:
@mitsudome-r (CC: @veqcc ) As the CI issue in this workflow is causing due to cache related, it might be difficult to resolve via PR. Thus I'd like to apply by-pass merge for this PR. Could you kindly provide your idea? :pray: Thank you in advance!

mitsudome-r:
I would like to investigate a little more on why it's failing in the CI.

amadeuszsz:
> I would like to investigate a little more on why it's failing in the CI.

@mitsudome-r 

If I'm not mistaken:
1. We added spconv to Autoware https://github.com/autowarefoundation/autoware/pull/6032
2. Deployment of `universe-devel-cuda` keeps failing due to out of memory https://github.com/autowarefoundation/autoware/actions/runs/15059681722/job/42336990430 (already rerun, let's wait...)
3. Package build silently fails because of `spconv is NOT available` in [build-and-test-differential-cuda](https://github.com/autowarefoundation/autoware_universe/actions/runs/15013940500/job/42187658034?pr=10529#step:15:648).
4. `FileNotFoundError: [Errno 2] No such file or directory: 'build/compile_commands.json'` occurs because we didn't build any package.

We can do one of the following:
1. Wait for `universe-devel-cuda` successful deployment.
2. Reconsider if silent failing is correct approach.
3. Adjust this package CMakeLists. For now, if CUDA is available and spconv not, we let silent fail. We could:
* If CUDA not available - silent fail (build succeed).
* If CUDA available:
  * If sub-dependency (spconv) not available - build fails and `build-and-test-differential-cuda` workflow not succeed.
  * If sub-dependency (spconv) available - build succeed.

amadeuszsz:
`universe-devel-cuda` was deployed successfully, we can merge it now.

PR #10531, opened 2025-04-22
fix(image_projection_based_fusion): fix redundantAssignment warning

veqcc:
## Description

Based on cppcheck `redundantAssignment` warning
```
perception/autoware_image_projection_based_fusion/src/camera_projection.cpp:83:39: style: Variable 'y_bottom' is assigned an expression that holds the same value. [redundantAssignment]
  if (ray_bottom_left.y < y_bottom.y) y_bottom = ray_bottom_left;
                                      ^
perception/autoware_image_projection_based_fusion/src/camera_projection.cpp:71:24: note: y_bottom is assigned 'ray_bottom_left' here.
  cv::Point3d y_bottom = ray_bottom_left;
                       ^
perception/autoware_image_projection_based_fusion/src/camera_projection.cpp:83:39: note: Variable 'y_bottom' is assigned an expression that holds the same value.
  if (ray_bottom_left.y < y_bottom.y) y_bottom = ray_bottom_left;
                                      ^
```

I think the initial value of `y_bottom` is wrong.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_image_projection_based_fusion/src/camera_projection.cpp
+++ b/perception/autoware_image_projection_based_fusion/src/camera_projection.cpp
@@ -68,7 +68,7 @@ CameraProjection::CameraProjection(
   cv::Point3d x_left = ray_top_left;
   cv::Point3d x_right = ray_top_right;
   cv::Point3d y_top = ray_top_left;
-  cv::Point3d y_bottom = ray_bottom_left;
+  cv::Point3d y_bottom = ray_bottom_right;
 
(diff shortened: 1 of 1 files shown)


ktro2828 (approved):
@veqcc Thank you for catching! LGTM.
(It seems this bug hasn't caused any problems because x value of `y_bottom` is not used in the lines below)

PR #10538, opened 2025-04-23
feat(autoware_probabilistic_occupancy_grid_map): integrated the blackboard into the ogm

knzo25:
## Description
As part of https://github.com/autowarefoundation/autoware_universe/issues/9722
The OGM was accelerated via CUDA.
However, the first implementation did not leverage the blackboard, so now that the blackboard and most of the other nodes had it integrated, I am also integrating this to the OGM

## Related links

**Parent Issue:**

- Link

## How was this PR tested?
 - Tested locally. As expected, nothing changes
## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


technolojin:
@YoshiRi How we proceed it?

PR #10379, opened 2025-04-01
fix(process_monitor): get process statistics directly from /proc files to avoid process spawning of Linux commands

nishikawa-masaki:
## PR Type
- Bug Fix

## Related links

**Parent Issue:**

- None

**Private Links:**

- [TIER IV internal link](https://tier4.atlassian.net/browse/RT1-9423)

## Description
- Current version of process_monitor node in autoware_system_monitor is implemented with Linux commands (top, sort, sed, pipe, echo).
- Boost::process::child is used to execute the Linux commands from both Timer thread and Publishing thread.
- Execution of Linux commands with boost::process::child causes the following problems.
  - High CPU workload caused by process_monitor itself.
    - Process spawning by boost::process::child consumes processing time of the calling threads.
    - Process spawning blocks context switching in the kernel space and interferes with other running processes.
- Linux commands have some problems.
  - They are designed to be versatile. It means they are not so efficient.
  - "top" command sleeps during execution to calculate CPU usage (%CPU)

Process_monitor needs to be implemented to read process statistics directly from files under /proc pseudo file system to stop using boost::process::child and Linux commands.

Unit testing of process monitor should be improved so that maintenance and porting in the future becomes easier.

## How was this PR tested?
- Unit testing with real /proc and dummy /proc data (for extreme cases)
- Integration testing by monitoring /diagnostics topic from autoware_system_monitor

## Notes for reviewers

- The implementation has changed somewhat drastically to solve the performance issue.
  - The current implementation employs Linux commands (top, echo, sed, sort, pipe) to collect process information.
  - This PR removes use of those Linux commands by reading process information directly from /proc pseudo file-system.
    - To avoid use of boost::process::child()
    - To avoid sleeping in "top -bn1 -w 128" command line.
  - Unit test cases for missing "top", "echo", "sed", "sort", "pipe" are removed.
  - ProcessMonitor::onTimer() is activated every 1 second and executes the following tasks.
    - Scan /proc/[pid] directories to collect process information.
      - Data fields are stored in their raw format (not as strings).
      - Data fields are converted to strings when they are published.
    - Store the process information in new_map.
    - Calculate CPU usage (%CPU) of each process by comparing data in new_map and prev_map.
    - Update "top 5 lists" of CPU Usage and Memory Usage.
- It is desirable if you can run both current implementation and this PR version and compare the results and performance.
- Published strings are expected to be same as the current implementation (emulating the behavior of "top" command).
- A minor bug of the current implementation is fixed during this change.
[...]

PR #10545, opened 2025-04-23
Fix/cpu monitor collect info on timer thread

nishikawa-masaki:
## Description
**This PR is still in Draft: state.
Descriptions will be added later.**

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/system/autoware_system_monitor/src/cpu_monitor/intel_cpu_monitor.cpp
+++ b/system/autoware_system_monitor/src/cpu_monitor/intel_cpu_monitor.cpp
@@ -173,7 +173,7 @@ void CPUMonitor::getTemperatureFileNames()
       continue;
     }
 
-    cpu_temperature_info temperature;
+    CpuTemperatureInfo temperature;
     temperature.path_ = temperature_input;
     temperature.label_ = path.filename().generic_string();
 
@@ -190,7 +190,7 @@ void CPUMonitor::getTemperatureFileNames()
     temperatures_.push_back(temperature);
   }
 
-  std::sort(temperatures_.begin(), temperatures_.end(), [](const cpu_temperature_info & c1, const cpu_temperature_info & c2) {
+  std::sort(temperatures_.begin(), temperatures_.end(), [](const CpuTemperatureInfo & c1, const CpuTemperatureInfo & c2) {
     std::smatch match;
     const std::regex filter(".*temp(\\d+)_input");
     int n1 = 0;
--- a/system/autoware_system_monitor/src/cpu_monitor/intel_cpu_monitor.cpp
+++ b/system/autoware_system_monitor/src/cpu_monitor/intel_cpu_monitor.cpp
@@ -1,4 +1,4 @@
-// Copyright 2020 Autoware Foundation
+// Copyright 2020,2025 Autoware Foundation
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -41,11 +41,11 @@ CPUMonitor::CPUMonitor(const rclcpp::NodeOptions & options) : CPUMonitorBase("cp
 {
   msr_reader_port_ = declare_parameter<int>("msr_reader_port", 7634);
 
-  this->getTempNames();
-  this->getFreqNames();
+  this->getTemperatureFileNames();
+  this->getFrequencyFileNames();
 }
 
-void CPUMonitor::checkThrottling(diagnostic_updater::DiagnosticStatusWrapper & stat)
+void CPUMonitor::checkThermalThrottling(diagnostic_updater::DiagnosticStatusWrapper & stat)
 {
   // Remember start time to measure elapsed time
   const auto t_start = SystemMonitorUtility::startMeasurement();
@@ -140,18 +140,18 @@ void CPUMonitor::checkThrottling(diagnostic_updater::DiagnosticStatusWrapper & s
       level = DiagStatus::ERROR;
     }
 
-    stat.add(fmt::format("CPU {}: Pkg Thermal Status", index), thermal_dict_.at(level));
+    stat.add(fmt::format("CPU {}: Pkg Thermal Status", index), thermal_dictionary_.at(level));
 
     whole_level = std::max(whole_level, level);
   }
 
-  stat.summary(whole_level, thermal_dict_.at(whole_level));
+  stat.summary(whole_level, thermal_dictionary_.at(whole_level));
 
   // Measure elapsed time since start time and report
   SystemMonitorUtility::stopMeasurement(t_start, stat);
 }
 
-void CPUMonitor::getTempNames()
+void CPUMonitor::getTemperatureFileNames()
 {
   const fs::path root("/sys/devices/platform/coretemp.0");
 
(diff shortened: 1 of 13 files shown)


sasakisasaki:
I'll start testing on my local environment :+1:

nishikawa-masaki:
@sasakisasaki 
Thank you for your reviewing.
To monitor the output from cpu_monitor (and other autoware_system_component nodes), you can use the following command lines.

Terminal1:
> $ ros2 launch autoware_system_monitor system_monitor.launch.xml

Terminal2:
> $ ros2 topic echo -f /diagnostics

As cpu_monitor reports about many CPU cores, you need to add "-f" to catch all pieces of information.
You may want to redirect the output to a file for investigation.

In addition to the "/diagnostics" topic, cpu_monitor also publishes another topic "/cpu_monitor/cpu_usage".
The topic seems to be a kind of "private topic" and it can't be monitored with "ros2 topic echo -f /cpu_monitor/cpu_usage".
> $ ros2 topic list
> /cpu_monitor/cpu_usage
> /diagnostics
> /parameter_events
> /rosout
> $ ros2 topic echo -f /cpu_monitor/cpu_usage
> The message type 'tier4_external_api_msgs/msg/CpuUsage' is invalid

I'll ask other developers about it and share the result here.

Unit tests for autoware_system_monitor (including cpu_monitor) has been disabled since employment of ROS2 according to a comment left in CMakeLists.txt.
I enabled only the unit tests for process_monitor in PR [#10379](https://github.com/autowarefoundation/autoware_universe/pull/10379)  when I worked on process_monitor.
I'll fix the unit tests for cpu_monitor and enable them in a coming PR (step 2 for issue [#10554](https://github.com/autowarefoundation/autoware_universe/issues/10554) ).

sasakisasaki:
@nishikawa-masaki Thank you so much for providing the detailed procedure! I have checked on my machine.

### What I observed (I have a question in this section)
- I could see the published topic by the following procedure.
> Terminal1:
$ ros2 launch autoware_system_monitor system_monitor.launch.xml
Terminal2:
$ ros2 topic echo -f /diagnostics

- But I observed the following error when `ros2 launch autoware_system_monitor system_monitor.launch.xml`. I'm not sure if this is my environment specific problem (sorry if I'm missing needed configuration and so on). Could you investigate why this error happens? I provided the error message and the lines showed error as follows. Thank you so much in advance!

#### Observed Error
```
[component_container_mt-1] [ERROR 1746177461.707465686] [net_monitor]: Failed to connect socket. No such file or directory (connect_service() at /home/junyasasaki/autoware/src/universe/autoware_universe/system/autoware_system_monitor/src/net_monitor/net_monitor.cpp:658)
```

#### Around `net_monitor.cpp:658`
```
648 bool NetMonitor::connect_service()
649 {
650   local::stream_protocol::endpoint endpoint(socket_path_);
651   socket_ = std::make_unique<local::stream_protocol::socket>(io_service_);
652 
653   // Connect socket
654   boost::system::error_code error_code;
655   socket_->connect(endpoint, error_code);
656 
657   if (error_code) {
658     RCLCPP_ERROR_ONCE(get_logger(), "Failed to connect socket. %s", error_code.message().c_str());
659     return false;
660   }
661 
662   return true;
663 }
```

nishikawa-masaki:
@sasakisasaki 
Thank you very much for your reviewing and testing.
I apologize for my delayed response. I was having some problems with re-enabling unit tests for "cpu_monitor", which had been disabled since employment of ROS2.
I changed this PR to "Draft" state because I need to add some changes to the source codes so that the unit tests work correctly.
I'll let you know when this PR becomes ready for re-reviewing.

As for the problem with the "net_monitor" node you reported, you can solve it with the following steps, though it is out of the scope of this PR.
* You may need to install "nethogs" command package as "traffic_reader" application invokes the command.
* Run "traffic_reader" application before executing "ros2 launch autoware_system_monitor system_monitor.launch.xml".
  * Change working directory to Autoware's workspace.
  * Execute "sudo ./build/autoware_system_monitor/traffic_reader" at the workspace directory.
  * Make it sure that a "traffic_reader" process is running (spawned by the command) and a socket named "/temp/traffic_monitor" exists.

sasakisasaki:
@nishikawa-masaki Thank you so much for providing the detailed procedure in running the node. Also I understood the status of this PR :+1: . I'm going to do tests again after this PR gets ready for review. Good luck!

nishikawa-masaki:
I re-opened this PR for reviewing after enabling unit tests, which had been disabled.

For reviewers' convenience, the following are the changes made in this PR.
* CPU information is collected and published in different threads.
  * The version before this PR did both tasks in the publishing context thread.
    * This was the reason why the "cpu_monitor" node was not able to publish topics every second (1 Hz) as expected.
  * Newly introduced onTimer() callback context thread collects information with checkXXX().
  * The publishing context thread publishes the information with updateXXX().
  * "mutex_snapshot_ " is used for guarding data structure between those two threads.
* Enabled unit tests.
  * "mutex_context_" is used for guarding variables between the unit test thread and the timer callback thread.
  * However, currently, unit tests disable the timer and call onTimer() by themselves.
    * If the timer was running, it would interfere with unit tests.
    * This "disabling" makes the mutual exclusion with "mutex_context_" meaningless, but the mutex is kept for future change.
 * "Lazy initialization" with getTemperatureFileNames() and getFrequencyFileNames() is introduced.
   * They are executed on the first call of onTimer().
   * The implementation before this PR seemed to have tried to employ polymorphism, but it didn't work.
 * Removed obscure abbreviations.
   * Temp -> Temperature
   * Throttling -> ThermalThrottling
   * Freq -> Frequency
   * Avg -> Average
   * Dict -> Dictionary
   * "warn" is left intentionally, because the ROS2 topics "usage_warn" and "usage_warn_count" can't be changed for keeping compatibility.
 * The ROS2 topics are explicitly declared as "read only".
 * Updated copyright years.

nishikawa-masaki:
Warnings from CodeCov and CodeScene are ignored in this PR because the warned lines of code will be removed or modified in the next PR.

The "cpu_monitor" node should be built and tested on different CPU platforms.
Currently, it is built and tested only on Intel platform.
Building and testing on other platforms are being conducted in parallel with this PR.

system/autoware_system_monitor/CMakeLists.txt
```diff
@@ -31,6 +31,14 @@ endif()
...
 ## Declare a C++ executable
 
+## Warning/Error about temperature used to be implemented,
+## but they were removed in favor of warning/error about thermal throttling.
+## add_definitions(-DENABLE_TEMPERATURE_DIAGNOSTICS)
+
+## Warning/Error about CPU load average used to be implemented,
+## but they were removed to avoid false alarms.
+## add_definitions(-DENABLE_LOAD_AVERAGE_DIAGNOSTICS)
+
```
sasakisasaki: If these lines might be added in the future development, the comment needs to be improved (for example, "TODO: We are going to implement XXX ... thus, this option might be used in the near future"). Otherwise please consider removing these comments.

system/autoware_system_monitor/src/cpu_monitor/intel_cpu_monitor.cpp
```diff
@@ -166,43 +226,46 @@ void CPUMonitor::getTempNames()
...
+    if (!std::regex_match(temperature_input.c_str(), match, std::regex(".*temp(\\d+)_input"))) {
       continue;
     }
 
-    cpu_temp_info temp;
-    temp.path_ = temp_input;
-    temp.label_ = path.filename().generic_string();
+    CpuTemperatureInfo temperature;
+    temperature.path_ = temperature_input;
+    temperature.label_ = path.filename().generic_string();
```
sasakisasaki, line 238: Thank you for using non-abbreviated word here, this is much easy-to-understand :+1:

sasakisasaki (approved):
### Reason for Approval

I have tested as follows. Good job!

- Terminal 1:
```
(Download attached `tested-repos.txt`)
$ mkdir src
$ vcs import src < tested-repos.txt
$ rosdep install -y --from-paths ./src --ignore-src --rosdistro $ROS_DISTRO

(For `traffic_reader`)
$ sudo apt install -y nethogs
$ sudo ./build/autoware_system_monitor/traffic_reader
$ ps augx | grep traffic_reader
(Verified it is running)

$ source install/setup.bash
$ ros2 launch autoware_system_monitor system_monitor.launch.xml
```

- Terminal 2:
```
$ source install/setup.bash
$ ros2 topic echo -f /diagnostics
(Verified the topic that of CPU monitor is shown)
```

- Attachment:
[tested-repos.txt](https://github.com/user-attachments/files/20653166/tested-repos.txt)

sasakisasaki:
@nishikawa-masaki I believe we can merge if there is no any additional changes :+1:

system/autoware_system_monitor/CMakeLists.txt
nishikawa-masaki: As there is no plan to re-enable these 2 diagnostics, I removed the comment lines.

PR #10550, opened 2025-04-24
feat(default_adapi): add mrm request api

isamu-takagi:
## Description

**This PR depends on https://github.com/autowarefoundation/autoware_adapi_msgs/pull/91**
The MRM request API implementation publishes `/diagnostics`. MRM is triggered by this diagnostics.

## Related links

**Parent Issue:**

- [MRM request API](https://autowarefoundation.github.io/autoware-documentation/main/design/autoware-interfaces/ad-api/features/fail-safe/#mrm-request)
- https://github.com/autowarefoundation/autoware/issues/3096

## How was this PR tested?

1. Launch planning simulation.
2. Change mode to autonomous.
3. Execute this command and check that the MRM is operating.
    ```
    ros2 service call /api/fail_safe/mrm_request/send autoware_adapi_v1_msgs/srv/SendMrmRequest "{request: {strategy: 2}}"
    ````
4. Execute this command and check that the MRM is canceled.
    ```
    ros2 service call /api/fail_safe/mrm_request/send autoware_adapi_v1_msgs/srv/SendMrmRequest "{request: {strategy: 1}}"
    ````

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/system/autoware_default_adapi/src/mrm_request.cpp
+++ b/system/autoware_default_adapi/src/mrm_request.cpp
@@ -1,0 +1,82 @@
+// Copyright 2025 The Autoware Contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "mrm_request.hpp"
+
+namespace autoware::default_adapi
+{
+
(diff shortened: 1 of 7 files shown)


system/autoware_default_adapi/src/mrm_request.cpp
```diff
@@ -0,0 +1,82 @@
...
+}
+
+void MrmRequestNode::publish_mrm_request_list()
+{
+  MrmRequestList::Message msg;
+  msg.stamp = now();
+  for (const auto & [sender, request] : mrm_requests_) {
+    msg.requests.push_back(request);
+  }
+  pub_mrm_request_list_->publish(msg);
```
nishikawa-masaki, line 54: Do you need to care the order of registering requests?
Std::unordered_map doesn't care the order of entries.

If multiple threads access an instance of MrmRequestNode class, mutual exclusion is necessary.

I don't know the exact requirement for the class. These are just reminders.

mitsudome-r (approved):

PR #10551, opened 2025-04-24
fix(autoware_bevfusion): build error when using ninja-build tool

zulfaqar-azmi-t4:
## Description

Fix build failure when using ninja.
The error was caused by using `$(...)` instead of `${...}` for a CMake variable, which breaks Ninja builds.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

Build  success.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


knzo25 (approved):
LGTM

PR #10555, opened 2025-04-24
feat(cuda_pointcloud_preprocessor): update filtering parameter and process

ktro2828:
## Description

This PR relates to https://github.com/autowarefoundation/autoware_universe/pull/10537, and contains the following changes:
- Modify default value of `object_length_threshold = 0.05`
- Remove process checking the number of points inside cluster and `num_points_threshold`.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

Uploading without_cuda.mp4…

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_pointcloud_preprocessor/outlier_kernels.cu
+++ b/sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_pointcloud_preprocessor/outlier_kernels.cu
@@ -20,8 +20,7 @@ namespace autoware::cuda_pointcloud_preprocessor
 
 __global__ void ringOutlierFilterKernel(
   const InputPointType * points, std::uint32_t * output_mask, int num_rings,
-  int max_points_per_ring, float distance_ratio, float object_length_threshold_squared,
-  int num_points_threshold)
+  int max_points_per_ring, float distance_ratio, float object_length_threshold_squared)
 {
   int idx = blockIdx.x * blockDim.x + threadIdx.x;
   int j = idx / max_points_per_ring;
@@ -31,10 +30,10 @@ __global__ void ringOutlierFilterKernel(
     return;
   }
 
-  int min_i = max(i - num_points_threshold, 0);
-  int max_i = min(i + num_points_threshold, max_points_per_ring);
+  const int window_size = 5;
+  int min_i = max(i - window_size, 0);
+  int max_i = min(i + window_size, max_points_per_ring);
 
-  int walk_size = 1;
   int left_idx = min_i;
   int right_idx = min_i + 1;
 
@@ -51,12 +50,10 @@ __global__ void ringOutlierFilterKernel(
         min(left_point.distance, right_point.distance) * distance_ratio &&
       azimuth_diff < 1.f * M_PI / 180.f) {
       // Determined to be included in the same walk
-      walk_size++;
       right_idx++;
     } else if (k >= i) {
       break;
     } else {
-      walk_size = 1;
       left_idx = k + 1;
       right_idx = k + 2;  // this is safe since we break if k >= i
     }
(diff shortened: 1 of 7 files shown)


mojomex:
Hi @ktro2828, what's the status for this PR?

There are currently some frictions in our products / AIP launcher because the CPU version has been updated, but the CUDA changes have not been tested yet. If I can help with testing or anything, please feel free to reach out!

mojomex:
Whoops, I accidentally clicked rebase on the wrong PR's tab :bow: :bow:

mojomex:
@ktro2828 It looks like the `num_points_threshold` must be removed from the default config file:

Failing [json-schema-check](https://github.com/autowarefoundation/autoware_universe/actions/runs/15394418875/job/43311419986?pr=10555):

```
🔍 Validating: sensing/autoware_cuda_pointcloud_preprocessor/config/cuda_pointcloud_preprocessor.param.yaml 🆚 sensing/autoware_cuda_pointcloud_preprocessor/schema/cuda_pointcloud_preprocessor.schema.json ❌ Failed
    Schema validation errors were encountered.
      sensing/autoware_cuda_pointcloud_preprocessor/config/cuda_pointcloud_preprocessor.param.yaml::$./**.ros__parameters: Additional properties are not allowed ('num_points_threshold' was unexpected)
```

ktro2828:
@mojomex Thank you for your comment!
I visualized pointcloud output both of with and without cuda, confirmed there is no big difference between them.

https://github.com/user-attachments/assets/4ac24c76-1e7d-41b0-a4b8-98de07f381be 

https://github.com/user-attachments/assets/3d75293c-795b-4561-8fa2-f87566687344

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_pointcloud_preprocessor/outlier_kernels.cu
```diff
@@ -31,10 +30,10 @@ __global__ void ringOutlierFilterKernel(
     return;
   }
 
-  int min_i = max(i - num_points_threshold, 0);
-  int max_i = min(i + num_points_threshold, max_points_per_ring);
+  const int window_size = 5;
```
mojomex, line 33: `window_size` limits the maximum length of a checked walk to `5 + 1 + 5 = 11` points. While this did not cause any issues during review, this implicitly puts a limit on how big the object length computed between left/right point can get, possibly causing walks to be lost because the 11 points of space were still smaller than `object_length_threshold`.

For now, I think this is an acceptable workaround to bound computation time per point.

@amadeuszsz Opinions?

mojomex (commented):
Thank you! Confirmed working on my end too, with 2nd gen bus ODD data.
I was also not able to find any differences.

I left one comment about the introduced artificial limit on walk length, but I think this is fine for now.

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_pointcloud_preprocessor/outlier_kernels.cu
amadeuszsz: @mojomex 
For me, we can keep `5` hardcoded for now. However, if possible, I would prefer to see `window_size` and `azimuth_diff_threshold` (for radians) or `azimuth_diff_threshold_deg` as this kernel arguments for better readability purposes.

mojomex: @amadeuszsz Fully agree. For now however, let's prioritize this getting merged so people don't need to refactor all their launchers to use CUDA.

mojomex (approved):
LGTM!

mojomex:
@amadeuszsz I'm not a CUDA preprocessor codeowner, you'll have to review as well :disguised_face:

PR #10558, opened 2025-04-25
fix(autoware_probabilistic_occupancy_grid_map): skipped first iteration

knzo25:
## Description

The first iteration was printing errors due to the delta poses between the null pose and the first pose being to big.
However, for the first iteration, the update routine is not needed (the map is empty), so I just skipped it

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

Tested that the error disappeared with this PR

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


YoshiRi (approved):
make sense, LGTM

PR #10566, opened 2025-04-28
feat: accelerate voxel filter

manato:
## Description
This PR adds a CUDA-accelerated voxel grid downsample implementation.

## Related links

**Parent Issue:**

- [TIER IV internal Link](https://tier4.atlassian.net/browse/RT1-9189)

## How was this PR tested?
```bash
colcon build --symlink-install --continue-on-error --cmake-args -DCMAKE_BUILD_TYPE=Release --packages-up-to autoware_cuda_pointcloud_preprocessor 

ros2 launch autoware_cuda_pointcloud_preprocessor cuda_voxel_grid_downsample_filter.launch.xml
```
After this launch, a node named `/cuda_voxel_grid_downsample_filtrer` subscribes `/sensing/lidar/top/pointcloud_raw` as input and publishes `/sensing/lidar/top/downsampled` as output

The following figure shows execution time comparison with `autoware::pointcloud_preprocessor::FasterVoxelGridDownsampleFilter` (denoted as `before`) and the implementation in this PR (denoted as `after`). 
voxel_grid_filter_comparison

## Notes for reviewers

I qualitatively confirmed that the output of this PR implementation is almost identical to the ones from `autoware::pointcloud_preprocessor::FasterVoxelGridDownsampleFilter`. However, I also observed that the number of output points is sometimes different (1 point difference), which I concluded comes from numerical computation difference

## Interface changes

None.

## Effects on system behavior

None.


manato:
@knzo25 
Thank you very much for pointing out regarding the output pointcloud format. I added a support to output `XYZIRC` format described [here](https://autowarefoundation.github.io/autoware-documentation/main/design/autoware-architecture/sensing/data-types/point-cloud/), so it would be really appreciated if you could review this PR again.

knzo25:
@manato 
It seems that the CI/CD is not passing. Can you check that?

sensing/autoware_cuda_pointcloud_preprocessor/include/autoware/cuda_downsample_filter/cuda_voxel_grid_downsample_filter.hpp
```diff
@@ -0,0 +1,137 @@
...
+    size_t output_offsets[6];
+    ThreeDim<float> voxel_size;
+    ThreeDim<int> min_coord;
+    ThreeDim<int> max_coord;
+  };
+
+  // This filter only returns XYZIRC regardless of the input point type
+#pragma pack(push, 1)
+  struct OutputPointType
+  {
```
knzo25, line 79: You should be able to use https://github.com/autowarefoundation/autoware_universe/blob/d82b071a2aca41c9a2be5be964d16b0594351ee5/sensing/autoware_cuda_pointcloud_preprocessor/include/autoware/cuda_pointcloud_preprocessor/point_types.hpp#L23
If you are worried about packing (the structure was aligned though) you can add the macro there

sensing/autoware_cuda_pointcloud_preprocessor/include/autoware/cuda_downsample_filter/thrust_custom_allocator.hpp
```diff
@@ -0,0 +1,66 @@
...
+  using Base = thrust::device_malloc_allocator<uint8_t>;
+  using pointer = typename Base::pointer;
+  using size_type = typename Base::size_type;
+
+  explicit ThrustCustomAllocator(cudaStream_t stream) : stream_(stream) {}
+
+  pointer allocate(size_type num)
+  {
+    uint8_t * buffer(nullptr);
+    auto result = cudaMallocAsync(&buffer, num, stream_);
```
knzo25, line 38: What about https://github.com/autowarefoundation/autoware_universe/blob/main/sensing/autoware_cuda_utils/include/autoware/cuda_utils/cuda_check_error.hpp
?

sensing/autoware_cuda_pointcloud_preprocessor/include/autoware/cuda_downsample_filter/thrust_custom_allocator.hpp
```diff
@@ -0,0 +1,66 @@
...
+      s << ::cudaGetErrorName(result) << " : " << ::cudaGetErrorString(result);
+      throw std::runtime_error{s.str()};
+    }
+
+    cudaMemsetAsync(buffer, 0, num, stream_);
+
+    return pointer(thrust::device_pointer_cast(buffer));
+  }
+
+  void deallocate(pointer ptr, size_t)
```
knzo25, line 50: can the size_t variable be eliminated?

sensing/autoware_cuda_pointcloud_preprocessor/include/autoware/cuda_downsample_filter/thrust_custom_allocator.hpp
```diff
@@ -0,0 +1,66 @@
...
+  {
+    auto result = cudaFreeAsync(thrust::raw_pointer_cast(ptr), stream_);
+    if (result != ::cudaSuccess) {
```
knzo25, line 53: What about
What about https://github.com/autowarefoundation/autoware_universe/blob/main/sensing/autoware_cuda_utils/include/autoware/cuda_utils/cuda_check_error.hpp
?

sensing/autoware_cuda_pointcloud_preprocessor/CMakeLists.txt
```diff
@@ -78,6 +80,38 @@ rclcpp_components_register_node(cuda_pointcloud_preprocessor
...
+  ${cuda_blackboard_INCLUDE_DIRS}
+  ${sensor_msgs_INCLUDE_DIRS}
+)
+
+# Targets
+ament_auto_add_library(cuda_voxel_grid_downsample_filter SHARED
+  src/cuda_downsample_filter/cuda_voxel_grid_downsample_filter_node.cpp
+)
+
+ament_target_dependencies (cuda_voxel_grid_downsample_filter # XXX: DEBUG
```
knzo25, line 99: What is the meaning of
`XXX: DEBUG`?

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_downsample_filter/cuda_voxel_grid_downsample_filter_node.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+void CudaVoxelGridDownsampleFilterNode::cudaPointcloudCallback(
+  const cuda_blackboard::CudaPointCloud2::ConstSharedPtr msg)
+{
+  // The following only checks compatibility with xyzi
+  // (i.e., just check the first four elements of the point field are x, y, z, and intensity
+  // and don't care the rest of the fields)
+  if (!cuda_pointcloud_preprocessor::is_data_layout_compatible_with_point_xyzi(msg->fields)) {
+    RCLCPP_ERROR(
+      this->get_logger(), "Input pointcloud data layout is not compatible with PointXYZI");
```
knzo25, line 52: if it is not compatible there should be an early return

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_downsample_filter/cuda_voxel_grid_downsample_filter.cu
```diff
@@ -0,0 +1,471 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "autoware/cuda_downsample_filter/cuda_voxel_grid_downsample_filter.hpp"
+#include "autoware/cuda_downsample_filter/thrust_custom_allocator.hpp"
+
+#include <sensor_msgs/msg/detail/point_field__struct.hpp>
```
knzo25, line 18: Out of curiosity, why is the standard header not used here?

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_downsample_filter/cuda_voxel_grid_downsample_filter.cu
```diff
@@ -0,0 +1,471 @@
...
+
+  // Calculate unique voxel indices that all points in input_points belong to
+  // Create lookup table to show correspondence between spatial voxel grid position
+  // and memory index
+  auto num_valid_voxel = searchValidVoxel(
+    input_points, voxel_index_buffer_dev, point_index_buffer_dev, return_type_field_dev,
+    channel_field_dev);
+
+  // Allocate region actually result filtered pointcloud requires and fill them
+  auto filtered_output = std::make_unique<cuda_blackboard::CudaPointCloud2>();
```
knzo25, line 262: nit: allocation can be performed after publishing by allocating space of the larger pointcloud processed so far (the amount of cuda memory is negligible I think)

sensing/autoware_cuda_pointcloud_preprocessor/include/autoware/cuda_downsample_filter/cuda_voxel_grid_downsample_filter.hpp
```diff
@@ -0,0 +1,137 @@
...
+#include <cuda_blackboard/cuda_pointcloud2.hpp>
+#include <cuda_blackboard/cuda_unique_ptr.hpp>
+
+#include <cuda_runtime.h>
+
+#include <cstdint>
+#include <iostream>
+#include <memory>
+
+#define CUDA_VGDF_ERROR_CHECK(e) (CUDA_BLACKBOARD_CHECK_CUDA_ERROR(e))
```
knzo25, line 29: nit: in autoware, should not we use the cuda_utils's version?

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_downsample_filter/cuda_voxel_grid_downsample_filter.cu
```diff
@@ -0,0 +1,471 @@
...
+{
+template <typename T>
+__device__ T * getElementPointer(
+  uint8_t * data, const size_t point_index, const size_t point_step, const size_t offset)
+{
+  return reinterpret_cast<T *>(data + point_index * point_step + offset);
+}
+
+template <typename T>
+__device__ const T getElementValue(
```
knzo25, line 46: I am not sure how it works in cuda, but in c++ this looks like aliasing. Can you not assume that the input point type is the one defined in this package? That way you can reduce a lot of clutter IMO

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_downsample_filter/cuda_voxel_grid_downsample_filter.cu
```diff
@@ -0,0 +1,471 @@
...
+
+  // calculate voxel index that each input point belong to
+  calculateVoxelIndexKernel<<<grid_dim, block_dim, 0, stream_>>>(
+    points->data.get(), num_voxels, voxel_index_buffer_dev, point_index_buffer_dev,
+    return_type_field_dev, channel_field_dev);
+
+  // get number of valid voxels
+  // Because thrust::unique_count returns the number of "consective" equal values,
+  // sorting is required beforehand to get all identical values
+  thrust::sort_by_key(
```
knzo25, line 413: Would there be any advantange of using  CUB here?
Sorting seems like the heaviest thing around and with CUB there is no attempt of allocation (which you have covered anyways) and you can save some bits since you know the ranges of the voxel coords (CUB allows to sort a subset of bits when using radix sort)

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_downsample_filter/cuda_voxel_grid_downsample_filter.cu
```diff
@@ -0,0 +1,471 @@
...
+#include <thrust/extrema.h>
+#include <thrust/scan.h>
+#include <thrust/sort.h>
+#include <thrust/unique.h>
+
+#include <cmath>
+#include <cstdint>
+#include <memory>
+#include <stdexcept>
+#include <type_traits>
```
knzo25, line 32: is this used?

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_downsample_filter/cuda_voxel_grid_downsample_filter.cu
```diff
@@ -0,0 +1,471 @@
...
+  CUDA_VGDF_ERROR_CHECK(cudaStreamCreate(&stream_));
+
+  // create memory pool to make repeated allocation efficient
+  {
+    int current_device_id = 0;
+    CUDA_VGDF_ERROR_CHECK(cudaGetDevice(&current_device_id));
+    CUDA_VGDF_ERROR_CHECK(cudaDeviceGetDefaultMemPool(&mem_pool_, current_device_id));
+
+    // Configure memory pool reusing allocation
+    // Following CUDA sample, set high release threshold so that allocated memory region
```
knzo25, line 194:
```suggestion
    // Configure the memory pool reusing allocation
    // Following the CUDA documentation, we set a high release threshold so that the allocated memory region
```

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_downsample_filter/cuda_voxel_grid_downsample_filter.cu
```diff
@@ -0,0 +1,471 @@
...
+  voxel_info_.output_offsets[5] =
+    voxel_info_.output_offsets[4] + sizeof(OutputPointType::return_type);
+
+  CUDA_VGDF_ERROR_CHECK(cudaStreamCreate(&stream_));
+
+  // create memory pool to make repeated allocation efficient
+  {
+    int current_device_id = 0;
+    CUDA_VGDF_ERROR_CHECK(cudaGetDevice(&current_device_id));
+    CUDA_VGDF_ERROR_CHECK(cudaDeviceGetDefaultMemPool(&mem_pool_, current_device_id));
```
knzo25, line 191: I think the way memory pools are being used makes them shared among nodes in the same process, which would lead to undefined behavior if nodes configure them differently. Maybe a singleton class in cuda_utils would be nice

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_downsample_filter/cuda_voxel_grid_downsample_filter.cu
```diff
@@ -0,0 +1,471 @@
...
+
+    // Configure memory pool reusing allocation
+    // Following CUDA sample, set high release threshold so that allocated memory region
+    // will be reused
+    uint64_t pool_release_threshold = ULONG_MAX;
```
knzo25, line 196: this effectively makes infinite caching.
I wonder if this is ok, or if we limit it, would it have a negative impact on tensorrt nodes, etc (tensorrt should not allocate memory each iteration right?)

knzo25:
@manato 
Thanks for the implementation. I learned a lot.
Left some comments, so please check them when you have time.
In addition, the documentation and the schema seems to be missing

knzo25:
@amadeuszsz @drwnz 
Could you notify Hirabayashi-san? I think we does not get universe notifications

sensing/autoware_cuda_pointcloud_preprocessor/CMakeLists.txt
```diff
@@ -78,6 +80,38 @@ rclcpp_components_register_node(cuda_pointcloud_preprocessor
...
+
+target_link_libraries(cuda_voxel_grid_downsample_filter
+  ${CUDA_LIBRARIES}
+  cuda_voxel_grid_downsample_filter_lib
+)
+
+rclcpp_components_register_node(cuda_voxel_grid_downsample_filter
+  PLUGIN "autoware::cuda_downsample_filter::CudaVoxelGridDownsampleFilterNode"
+  EXECUTABLE cuda_voxel_grid_downsample_filter_node
+)
```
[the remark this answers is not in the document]
manato: I misunderstood the packages' structure. I rearranged the structure in 37f7a0c14

sensing/autoware_cuda_pointcloud_preprocessor/include/autoware/cuda_downsample_filter/cuda_voxel_grid_downsample_filter.hpp
manato: Thanks for pointing this out. I modified the code to use the struct you suggested in 76a4f76ee

sensing/autoware_cuda_pointcloud_preprocessor/include/autoware/cuda_downsample_filter/thrust_custom_allocator.hpp
manato: Thanks. In 3ccc84d61, I changed the code to use macro defined in autoware_cuda_utils for CUDA related error check

sensing/autoware_cuda_pointcloud_preprocessor/include/autoware/cuda_downsample_filter/thrust_custom_allocator.hpp
manato: same as https://github.com/autowarefoundation/autoware_universe/pull/10566#discussion_r2095511517

sensing/autoware_cuda_pointcloud_preprocessor/CMakeLists.txt
manato: Sorry, I left this code carelessly. I deleted here in the latest code.

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_downsample_filter/cuda_voxel_grid_downsample_filter_node.cpp
manato: I modified the code so that the filter can handle different types of data for the intensity field of the input pointcloud in 0c949b35d. 
While `pointcloud_preprocessor::utils::is_data_layout_compatible_with_point_xyzi` confirms that the intensity field has float values, my filter still may work with other type such as uint8_t. For this reason, I changed here from `RCLCPP_ERROR` to `RCLCPP_WARN` without an early return.

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_downsample_filter/cuda_voxel_grid_downsample_filter.cu
manato: This line was accidentally inserted by my editor. I creaned up the headre includes by 6b7383607

sensing/autoware_cuda_pointcloud_preprocessor/include/autoware/cuda_downsample_filter/cuda_voxel_grid_downsample_filter.hpp
manato: same as https://github.com/autowarefoundation/autoware_universe/pull/10566#discussion_r2095511517. Thanks for your detailed review.

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_downsample_filter/cuda_voxel_grid_downsample_filter.cu
manato: same as https://github.com/autowarefoundation/autoware_universe/pull/10566#discussion_r2095529301. Thanks

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_downsample_filter/cuda_voxel_grid_downsample_filter.cu
manato: Yes, this is just aliasing. Since this kind of raw pointer arithmetic easily causes bugs, I chose to use aliasing with template typename. Besides, as handled in 0c949b35d, the intensity field may have a different data type, so I chose this design to keep availability as much as possible, even though the result becomes clutter :P

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_downsample_filter/cuda_voxel_grid_downsample_filter.cu
manato: I applied CUB sorting instead of thrust in 0e31c08f9. Thrust's sorting function uses CUB inside, but direct CUB API calls seem to reduce the number of kernels to be called. Thanks for the suggestion.

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_downsample_filter/cuda_voxel_grid_downsample_filter.cu
manato: Thanks for pointing this out. As you suggested, using the device-level default memory pool operation may lead to undefined behavior if it is modified from multiple places. It was revealed that we could have private memory pools, so I implemented it instead of sharing the one big pool among multiple nodes via a singleton class (b590259).  If you still have further concerns, I'd appreciate letting me know

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_downsample_filter/cuda_voxel_grid_downsample_filter.cu
manato: Thanks for your insight.
I monitored GPU memory usage over multiple subscribe call executions and found that  GPU memory usage is almost stable. Although the size you mentioned is just a maximum size for the pool and not the actual size of the allocation, I set a variable maximum size that can be passed via ROS parameter so that users can tune the amount and avoid infinite caching (6a3ff55)

manato:
@knzo25 
I really appreciate your detailed/insightful review and help. And truly sorry for the delay in responding.
I guess I handled the items you pointed out, so I'd appreciate your check when it is convenient for you.

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_downsample_filter/cuda_voxel_grid_downsample_filter.cu
knzo25: @manato 
What about what I just mentioned about limiting the bits that are processed?
In the worst case (for a 0.3m grid) it uses less than half the iterations.
For a 3.0m grid it is even better

knzo25 (changes requested):
@manato 
Thank you for addressing most of the comments. However, I believe there are still 2 unresolved comments

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_downsample_filter/cuda_voxel_grid_downsample_filter.cu
manato: @knzo25 
Thanks for pointing this out. I missed your old comment. I added a modification to set the most significant bit for radix sort in 0b9a1bb.

manato:
@knzo25 
Thank you for your review, and sorry for taking your time again and again. 

> There is one comment that I would consider optional

I guess you refer to  https://github.com/autowarefoundation/autoware_universe/pull/10566#discussion_r2092425200. But I couldn't catch the exact meaning. So, would you mind confirming my question in the responses of that comment?

sensing/autoware_cuda_pointcloud_preprocessor/README.md
```diff
@@ -14,6 +14,7 @@ A detailed description of each filter's algorithm is available in the following
 | ----------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------- | -------------------------------------------- |
 | cuda_pointcloud_preprocessor        | Implements the cropping, distortion correction, and outlier filtering (ring-based) of the `autoware_pointcloud_preprocessor`'s CPU versions. | [link](docs/cuda-pointcloud-preprocessor.md) |
 | cuda_concatenate_and_time_sync_node | Implements pointcloud concatenation an synchronization following `autoware_pointcloud_preprocessor`'s CPU implementation.                    | [link](docs/cuda-concatenate-data.md)        |
+| cuda_voxel_grid_downsample_filter   | Implements voxel downsample filtering of the `autoware_pointcloud_preprocessor`'s CPU version                                                | [link](docs/cuda-voxel-grid-downsample.md)   |
```
Owen-Liuyuxuan, line 17: Sorry, why does the readme state that` cuda_*_preprocessor -- implements ** CPU version`, while the package  looks like a GPU implementation?

PR #10572, opened 2025-04-30
fix(autoware_compare_map_segmentation): handle empty output point cloud 

technolojin:
## Description

When the filtered pointcloud is empty, we expect to publish the empty message.

Fix a check of the output data, and split in the case the output message is a normal empty message.

## Related links

**Parent Issue:**

- Link

**Private Links:**

- [TIER IV internal link](https://star4.slack.com/archives/C03S84LDJGG/p1745981452114449)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


badai-nguyen (approved):
Thank you for your fixing. 
LGTM

technolojin:
I would like to merge after the effect is confirmed.

PR #10573, opened 2025-04-30
feat(start_planner): add virtual stop detail

kyoichi-sugahara:
## Description

There are several factors that `start_planner` can't generate path, such as when the vehicle deviates from its lane or when safety cannot be guaranteed with stationary objects. Since it is difficult to definitively determine these factors, I added details in the [PR](https://github.com/autowarefoundation/autoware_universe/pull/10029).

However, from a user perspective, there are many complex elements such as planner priorities, so I added the route generation failure factors to the detail field as a judgment criterion that can handle most cases without considering edge cases.

I expect that detailed judgments will still be made based on the explanation in this [PR](https://github.com/autowarefoundation/autoware_universe/pull/10029).

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

- [x] [PASS TIERIV INTERNAL SCENARIOS](https://evaluation.tier4.jp/evaluation/reports/297fc4af-8e36-512d-ae3a-cf87d137ead6?project_id=prd_jt)
- [x] run psim and expected detail information is visualized in rvis

### Lane Departure Case
image

### Insufficient Clearance Case
image

### Stop Against Dynamic Object Case
image

All test cases confirmed that the expected detailed information is properly displayed in RVIZ as intended.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_path_planner/autoware_behavior_path_start_planner_module/src/start_planner_module.cpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_start_planner_module/src/start_planner_module.cpp
@@ -683,21 +683,21 @@ bool StartPlannerModule::isExecutionReady() const
     // Future work should provide appropriate messages for each case.
 
     if (!parameters_->enable_back && !is_inside_lanelets) {
-      stop_reason = "ego footprint is out of lanelets";
+      stop_reason = "ego outside lanes";
     } else if (is_inside_lanelets) {
-      stop_reason = "insufficient clearance against static objects";
+      stop_reason = "unsafe against static objects";
     } else {
-      stop_reason = "failed to generate valid pull out path";
+      stop_reason = "failed to generate valid path";
     }
 
   } else if (isWaitingApproval()) {
     // Check for moving objects around
     if (!noMovingObjectsAround()) {
       is_safe = false;
-      stop_reason = "moving objects around: before approval";
+      stop_reason = "unsafe against moving objects around";
     } else if (requiresDynamicObjectsCollisionDetection() && hasCollisionWithDynamicObjects()) {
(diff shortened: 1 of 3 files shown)


planning/behavior_path_planner/autoware_behavior_path_start_planner_module/src/start_planner_module.cpp
```diff
@@ -335,6 +336,57 @@ bool StartPlannerModule::hasCollisionWithDynamicObjects() const
...
+    vehicle_info_.createFootprint(), autoware_utils::pose2transform(current_pose));
+
+  lanelet::BasicPolygon2d footprint_polygon;
+  for (const auto & point : vehicle_footprint) {
+    footprint_polygon.push_back({point.x(), point.y()});
+  }
+
+  // Find lanelets that intersect with the vehicle footprint
+  const auto & lanelets_distance_pair = lanelet::geometry::findWithin2d(
+    planner_data_->route_handler->getLaneletMapPtr()->laneletLayer, footprint_polygon, 0.0);
```
kosuke55, line 352: Do we need to check all the lanes, not just the lane departure check lane?

kosuke55: Ah I see lane_departure_checker_ also check all lanelets
https://github.com/tier4/autoware_universe/blob/c509904eb2b0fb38f4645860508dd79073331af9/planning/behavior_path_planner/autoware_behavior_path_start_planner_module/src/shift_pull_out.cpp#L108-L110

kosuke55 (approved):

PR #10614, opened 2025-05-12
feat(autoware_probabilistic_occupancy_grid_map): added target architectures for the ogm

knzo25:
## Description
As per the title

## Related links

**Parent Issue:**

- https://github.com/autowarefoundation/autoware_universe/issues/10610

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


YoshiRi (approved):
I roughly understood what it wants to do reading your issue. But how should I check this?
I do LGTM here so that my check will not become a bottleneck.

knzo25:
Very old architectures (Volta and prior ones; e.g., 1XXX series) will no longer be able to use this node (if someone really wants to execute for those targets, we can add it). For now all 2XXX 3XXX 4XXX and even 5XXX (these ones with ptx) will work with no issues, so as long as the node compiles and runs there would be no issues

PR #10624, opened 2025-05-13
feat(tier4_planning_rviz_plugin): add glog in the path visualization

takayuki5168:
## Description

When the rviz is directly not via autoware.launch.xml, since glog is not included in the tier4_planning_rviz_plugin, when the rviz process dies, there is no information of glog for the debugging.
This PR introduces the glog in the path visualization of the tier4_planning_rviz_plugin package.
## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


takayuki5168:
@kosuke55 
Please review this PR.

kosuke55 (approved):

PR #10631, opened 2025-05-14
fix(traffic_light_utils): remove condition in `setSignalUnknown`

MasatoSaeki:
## Description

This PR gets to delete if condition and wrong comment.
- `// the default value is -1, which means to not set confidence` is wrong because default value is 0 in  numeric type ros msg (and that value does not set -1 in autoware_traffic_light_classifier).
  - https://design.ros2.org/articles/interface_definition.html
- if condition `if (confidence > 0)` does not need because of avobe description. And `setSignalUnknown` is only used in autoware_traffic_light_occlusion_predictor. But in this package, it is used as if confidence 0.0 can be set. (actually it is not set as 0.0 now.)

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

I have checked this PR works(as traffic light recognition) on my local pc.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


ktro2828 (approved):
LGTM

PR #10632, opened 2025-05-14
feat(autoware_tensorrt_classifier): split cuda and cpp

MasatoSaeki:
## Description

This PR get to split cuda and cpp.

As mentioned [here](https://github.com/autowarefoundation/autoware_universe/blob/main/perception/autoware_tensorrt_yolox/CMakeLists.txt#L97-L101), `.cu` can be compiled with add_library, but it can make bugs. Therefore, should use `cuda_add_library` for `.cu`.

Below error occured with CMake so far. (when implement test code in `autoware_traffic_light_classifier`)
```
/usr/bin/ld: /home/masatosaeki/t4/pilot-auto/test_classifier/install/autoware_tensorrt_classifier/lib/libautoware_tensorrt_classifier.so: undefined reference to `resize_bilinear_letterbox_nhwc_to_nchw32_batch_gpu(float*, unsigned char*, int, int, int, int, int, int, int, float, CUstream_st*)'
collect2: error: ld returned 1 exit status
gmake[2]: *** [CMakeFiles/test_utils.dir/build.make:900: test_utils] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:270: CMakeFiles/test_utils.dir/all] Error 2

```

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

https://evaluation.ci.tier4.jp/evaluation/reports/18c9aea3-d30c-5c2b-b538-af771c11dc66?project_id=prd_jt

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


ktro2828 (approved):
LGTM

PR #10633, opened 2025-05-14
refactor(autoware_traffic_light_classifier): split utils and add test

MasatoSaeki:
## Description

Split file to perform the test and add unit test.
Need to merge https://github.com/autowarefoundation/autoware_universe/pull/10632 first.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

Evaluator
https://evaluation.ci.tier4.jp/evaluation/reports/70c2868e-bd6b-5093-b36b-e236b5107943?project_id=prd_jt

Unit Test
```
colcon build --merge-install --symlink-install --packages-up-to autoware_traffic_light_classifier --cmake-args -DCMAKE_BUILD_TYPE=Debug
```
```
colcon test --merge-install --packages-select autoware_traffic_light_classifier --event-handlers console_cohesion+
```

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


technolojin (approved):
LGTM

Personally, I don't prefer 'utils' because the meaning is not specific, but it is way better than current all-in-one coding.

MasatoSaeki:
@technolojin
Since other TLR code uses utils, we'll leave it as is for now, but once refactoring is complete, we'll change it to process or similar.

PR #10634, opened 2025-05-14
ci(build-and-test): ensure sequential run and remove core limitation

xmfcx:
## Description

For some reason, these workflows are getting canceled when a new one arrives. This PR attempts to ensure they all run without cancelling the old ones.

Will follow https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/control-the-concurrency-of-workflows-and-jobs#example-using-concurrency-and-the-default-behavior and use `group: ${{ github.workflow }}-${{ github.ref }}`.

Also removed the core limitation since this is running on 8c16t self hosted machines.

image

image

> Canceling since a higher priority waiting request for build-and-test exists

## How was this PR tested?

build-and-test-daily works under the same machine without the core limitations just fine and much faster.

image

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -6,8 +6,8 @@ on:
       - main
 
 concurrency:
-  # Ensures sequential execution of this workflow
-  group: ${{ github.workflow }}
+  # Ensure sequential execution of this workflow
+  group: ${{ github.workflow }}-${{ github.ref_name }}
   cancel-in-progress: false
 
 env:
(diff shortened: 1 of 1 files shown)


mitsudome-r (approved):

xmfcx:
It seems the sequential issue is on the github side, their limitations:

- [Using concurrency in workflows shouldn't cancel previously pending workflows #41518](https://github.com/orgs/community/discussions/41518)
- https://taurit.pl/github-canceling-since-a-higher-priority-waiting-request-exists/

At least this PR removed the core count limitations. And the new concurrency group name is better.

PR #10638, opened 2025-05-15
feat(collision_detector): time buffer and ignore behind rear axle

maxime-clem:
## Description

Required launch PR: https://github.com/autowarefoundation/autoware_launch/pull/1443

## Related links

**Parent Issue:**

- Link

**Private Links:**

- [TIER IV internal link](https://tier4.atlassian.net/browse/L4P-173)

## How was this PR tested?

Psim

I visualize the new debug markers.
- We can see the ego footprint used for collision detection and with the new option the polygon does not extend behind the rear axle.
- When the diagnostic of the `collision_detector` is in the `ERROR` state, the footprint turns red. Otherwise it is green.

In the following video, we can see that once the ego footprint overlaps the object footprint, the footprint becomes red (`ERROR` diagnostic) and becomes extended by the hysteresis. After the obstacle disappear, it takes some time for the diagnostic to go back to the normal state (footprint goes back to the normal size and to the green color).

https://github.com/user-attachments/assets/5d93a27d-8c93-4786-9032-7a7ea21f7af3

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/control/autoware_collision_detector/src/node.cpp
+++ b/control/autoware_collision_detector/src/node.cpp
@@ -362,6 +362,9 @@ void CollisionDetectorNode::checkCollision(diagnostic_updater::DiagnosticStatusW
   const auto is_collision_found =
     !nearest_obstacle ? false : nearest_obstacle->first < node_param_.collision_distance;
 
+  // When a collision is detected, update timestamps to track collision duration
+  // - start_of_consecutive_collision_stamp_: marks when a continuous collision began
+  // - most_recent_collision_stamp_: records the latest collision detection time
   if (is_collision_found) {
     if (!start_of_consecutive_collision_stamp_.has_value()) {
       start_of_consecutive_collision_stamp_ = this->now();
@@ -371,6 +374,13 @@ void CollisionDetectorNode::checkCollision(diagnostic_updater::DiagnosticStatusW
     start_of_consecutive_collision_stamp_.reset();
   }
 
+  // Define condition to determine error state based on diagnostic mode
+  // 1. When already in error state (is_error_diag_ == true):
+  //    - Stay in error if time since last collision is less than off_buffer time
+  //    - This creates hysteresis to prevent rapid switching between states
+  // 2. When in normal state (is_error_diag_ == false):
+  //    - Enter error if collision has been continuous for longer than on_buffer time
+  //    - This prevents triggering on brief/momentary collisions
   const auto condition_to_trigger_error = [&]() {
     if (is_error_diag_) {
       return (this->now() - *most_recent_collision_stamp_).seconds() < node_param_.time_buffer.off;
--- a/control/autoware_collision_detector/src/node.cpp
+++ b/control/autoware_collision_detector/src/node.cpp
@@ -109,12 +109,13 @@ autoware_utils_geometry::Polygon2d createObjPolygonForCylinder(
 }
 
 autoware_utils_geometry::Polygon2d createSelfPolygon(
-  const VehicleInfo & vehicle_info, const bool ignore_behind_rear_axle)
+  const VehicleInfo & vehicle_info, const double extra_offset, const bool ignore_behind_rear_axle)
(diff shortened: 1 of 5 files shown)


kyoichi-sugahara:
@maxime-clem 
Thank you for the change! 
Can you briefly explain what you tested with psim?

kyoichi-sugahara:
@maxime-clem 
Please update parameter file on autoware_universe side as well

control/autoware_collision_detector/src/node.cpp
```diff
@@ -349,45 +354,66 @@ void CollisionDetectorNode::checkCollision(diagnostic_updater::DiagnosticStatusW
...
+  const auto nearest_obstacle = getNearestObstacle(ego_polygon);
 
-  const auto is_obstacle_found =
-    !nearest_obstacle ? false : nearest_obstacle.get().first < node_param_.collision_distance;
+  const auto is_collision_found =
+    !nearest_obstacle ? false : nearest_obstacle->first < node_param_.collision_distance;
 
-  if (is_obstacle_found) {
-    last_obstacle_found_stamp_ = this->now();
+  if (is_collision_found) {
```
kyoichi-sugahara, line 365:
```suggestion
  // When a collision is detected, update timestamps to track collision duration
  // - start_of_consecutive_collision_stamp_: Marks when a continuous collision began
  // - most_recent_collision_stamp_: Records the latest collision detection time
  if (is_collision_found) {
```

control/autoware_collision_detector/src/node.cpp
```diff
@@ -349,45 +354,66 @@ void CollisionDetectorNode::checkCollision(diagnostic_updater::DiagnosticStatusW
...
+      start_of_consecutive_collision_stamp_ = this->now();
+    }
+    most_recent_collision_stamp_ = this->now();
+  } else {
+    start_of_consecutive_collision_stamp_.reset();
   }
 
-  const auto is_obstacle_found_recently =
-    (this->now() - last_obstacle_found_stamp_).seconds() < 1.0;
+  const auto condition_to_trigger_error = [&]() {
```
kyoichi-sugahara, line 374:
```suggestion
  // Define condition to determine error state based on diagnostic mode
  // 1. When already in error state (is_error_diag_ == true):
  //    - Stay in error if time since last collision is less than off_buffer time
  //    - This creates hysteresis to prevent rapid switching between states
  // 2. When in normal state (is_error_diag_ == false):
  //    - Enter error if collision has been continuous for longer than on_buffer time
  //    - This prevents triggering on brief/momentary collisions
  const auto condition_to_trigger_error = [&]() {
```

kyoichi-sugahara (approved):
Thanks you for refactoring and adding debug infotmation.
Approved but please consider adding comment for easier understanding

maxime-clem:
@kyoichi-sugahara 
> Can you briefly explain what you tested with psim?

I visualize the new debug markers.
- We can see the ego footprint used for collision detection and with the new option the polygon does not extend behind the rear axle.
- When the diagnostic of the `collision_detector` is in the `ERROR` state, the footprint turns red. Otherwise it is green.

In the following video, we can see that once the ego footprint overlaps the object footprint, the footprint becomes red (`ERROR` diagnostic) and becomes extended by the hysteresis. After the obstacle disappear, it takes some time for the diagnostic to go back to the normal state (footprint goes back to the normal size and to the green color).

https://github.com/user-attachments/assets/5d93a27d-8c93-4786-9032-7a7ea21f7af3

control/autoware_collision_detector/src/node.cpp
maxime-clem: 20dce759b

control/autoware_collision_detector/src/node.cpp
maxime-clem: 20dce759b

PR #10642, opened 2025-05-15
ci(build-and-test*): simplify with reusable workflows

xmfcx:
## Description

We had a lot of code repetition in the following:
- **`build-and-test.yaml`**
   - runs per merged commit to main, cuda only
   - pushes and updates the ccache used by everything else
- **`build-and-test-daily.yaml`**
   - runs daily with and without cuda
   - only used the ccache
- **`build-and-test-daily-arm64.yaml`**
   - runs daily with and without cuda
   - its ccache config was broken, now removed explicitly
   - it uses arm codebuild with 7/8 cores

And they had matrices in them split by cuda, it made it harder to modularize them further.

This PR simplifies them by introducing a reusable workflow called `build-and-test-reusable`.

Now, the repetition is minimized and it is much easier to maintain these workflows.

## How was this PR tested?

- `build-and-test-daily.yaml` will be tested by calling with "workflow_dispatch"
   - https://github.com/autowarefoundation/autoware_universe/actions/runs/15068845283 ✅
- But `build-and-test.yaml` will need to be tested once it is merged.
   - They were so similar that they can also be inspected visually.

image

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


xmfcx:
image

https://github.com/autowarefoundation/autoware_universe/actions/runs/15068845283

Only arm64's are failing but it's a test issue, also failed before. The workflows are being executed as expected as before ✅

mitsudome-r (approved):

PR #10648, opened 2025-05-16
refactor(autoware_multi_object_tracker): update Mahalanobis distance threshold for data association

technolojin:
## Description
Changed the Mahalanobis distance threshold from 3.035 to a new critical value of 3.717, corresponding to a 99.99% confidence level for improved accuracy in object tracking.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


ogrex (commented):
As mentioned in https://github.com/autowarefoundation/autoware_universe/issues/10622#issuecomment-2899671927 it's possible to tune the Q and R parameters to achieve a value of 3.4—as concluded in the previous analysis. If we maintain the original Q and R settings to avoid affecting the tracking behavior itself, the value of 3.717 is reasonable and aligns with experimental results.

ogrex (approved):

PR #10649, opened 2025-05-16
feat(autoware_detected_object_validation): add lanelet object elevation filter

a-maumau:
## Description
This PR adds 3D filtering to the lanelet filter that will filter objects under the lanelets.
In this PR implementation, it will only filter the unknown objects if the unknown objects are a filter target. For other object class, this filter will not apply.

This PR must merge with https://github.com/autowarefoundation/autoware_launch/pull/1446 .

# Main Goal
The filter introduced in this PR aims to handle ghost point artifacts that appear in rainy scenarios (see Fig. 1).

From the LiDAR’s (vehicle's) perspective, some points may appear below the road (lanelet). However, the actual LiDAR rays bounce off the wet road surface and reflect off nearby objects, creating ghost points under the road surface.

These ghost points will get detected as unknown objects and may harm the later stage like planning modules.

This PR introduces a new filter that uses lanelet geometry to identify and remove such ghost points by checking whether the points are located below the lanelet (road surface).

pr_aim
Fig.1

# Implementation
The filtering process follows these steps:

1. Generate query points for each unknown object. Query point will be like a centroid (and, centroid+height, centroid-height) of the object.
2. Create triangle meshes from relevant lanelets (each query point is matched to one candidate lanelet).
3. Check whether the query points (centroid+height, centroid-height) lies in certain distance range from the closest mesh polygon of the closest lanelet.

This check is performed by calculating the dot product between the polygon's normal vector and the query points. If one of the query point is in the range of  [`min_elevation_threshold`, `max_elevation_threshold`], object will be remained.
See Fig. 2 and Fig. 3 for example visualization of the algorithm.

This implementation is assuming the unknown object's position is the centroid of the cluster (convex hull). See the shape estimation's implementation: https://github.com/autowarefoundation/autoware_universe/blob/595b62ff13753f621a6ec37627f1ebe01cfe57f1/perception/autoware_shape_estimation/lib/model/convex_hull.cpp#L93-L95

algorithm_overview
Fig. 2
For only using a centroid of a cluster (in the actual implementation, it will be the object position), we will compute the distance with the plane that fits with the triangle points since projection point may go out the bounds of a triangle mesh.
In this figure, the yellow polygon is selected as the closest.

sample_top_bottom_querying
Fig. 3
Red squares are the query points. Visualization code is not included in this PR.
For querying the nearest polygon of lanelet, the object's center (position) will be used. The filter decision is based on the top and bottom query points.

# Computational Cost
Following result was tested in a benchmark that uses an rosbag adding many unknown objects with relatively large footprints (100 unknown objects, each with 64 points – see Fig. 5).

Comparison of processing times
cmp_processing_time
Fig. 4

mean processing time of Fig. 4 scene
``` text
Without object elevation filter: 3.84 ms
With    object elevation filter: 4.21 ms
```

This filter will not introduce a large computational overhead in typical scenarios.

This benchmark may not reflect exact performance in a real-world environment.
This has not been tested in environments where lanelets contain large numbers of points in each LineString.

lanelet_3d_filter_benchmark_env_ss
Fig. 5 Adding some artificial unknown objects to make the scene complex and stress the filter.

# Known Limitations
1. Lanelet Accuracy
The lanelets must accurately represent the actual road surface. In Fig. 3 (from the Autoware sample map data), lanelets are defined at a fixed height, causing many query points (red squares) to appear below them.

2. Lanelet Shape
Lanelets should not contain extremely steep segments. The algorithm computes a perpendicular projection to find the distance from the query point to each triangle polygon. This process assumes the surface is convex-like.
In Fig. 6(b), although the blue segment seems correct, the algorithm may choose the purple segment, leading to incorrect filtering.

limitation_steep_angle
Fig. 6
The problem is in the (b) case. The object that must not be filtered will be filtered.

3. Layered Lanelets
The algorithm selects a candidate lanelet by comparing the z-coordinate of the first point of each lanelet’s left bound. On long slopes with large height differences, this may produce incorrect results (see Fig. 7). In this case, it should select the lanelet 1, but it would be the lanelet 2.
It is possible to solve this case by only using lanelets that is connected to the lanelets that vehicle is currently riding on.

limitation_layered_lanelets
Fig. 7

5. Input Cluster Shape (Specification)
If ghost points are merged into a cluster of non-ghost points, the filter will not be able to remove them (see Fig. 8). This is an expected behavior based on the filter's design.

limitation_cluster
Fig. 8

6. Might Make Unable to Detect Potholes
This filter will filter the underground objects, so if there is an actual object (or like a cluster generated from a pothole), it will not be able to detect.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


technolojin:
test on high vertical curve road

https://github.com/user-attachments/assets/3a930ab0-dde4-4ff5-9a4a-f6aeeb2d8a50

Grid: baselink surface
Road: lanelet map
Blue box: detected&tracked objects
White convex: Unknown objects

objects over curved lanelet did not filtered, as expected.

technolojin (changes requested):
I have a few points to be considered.

1. existing height filter
Existing the height filter is a similar functionality.
How do you consider this lanelet filter?
    a. lanelet filter replaces height filter
    b. user may choose one of the filter
    c. user can apply filters interdependently
If you choose one of them, make a clear description of choice and clarify the config file in aligned format.

2. Lanelet filter for bounding box object and convex hull object
The current logic imply the filter only to convex hull object.
May some use case, applying the filter to bounding object also needed.

3. Height threshold relative to lanelet
Current filter only considers the surface of the lanelet. Is possible to set minimum and maximum threshold to the projected point to the reference(query) point?

5. Control flow of  `isObjectOverlapLanelets`
Readability of `isObjectOverlapLanelets` function is bad.
Please make it clear.

technolojin:
Control flow of `isObjectOverlapLanelets`


```c++
    bool is_point_in_polygon = false;
    ...
        if (isInPolygon(point_transformed.x, point_transformed.y, candidate.second.polygon, 0.0)) {
          if (filter_settings_.filter_object_under_lanelet) {
            is_point_in_polygon = true;
            break;
          } else {
            return true;
          }
        }
      }

      if (is_point_in_polygon) break;
    }

    // filter the object under the lanelet
    if (filter_settings_.filter_object_under_lanelet && is_point_in_polygon) {
      ...
      return isCentroidAboveLanelet(candidates, centroid, object.shape.dimensions.z * 0.5);
    }
    return is_point_in_polygon;
```

* usage of `is_point_in_polygon` is not intuitive.
* `filter_settings_.filter_object_under_lanelet` switches filter characteristics, so it is better to be higher layer of control, not scattered

a-maumau:
@technolojin I refactor and change the filter algorithm to check both top and bottom in https://github.com/autowarefoundation/autoware_universe/pull/10649/commits/840b27a4ebf2e72185001659ab4e8ce04ba43774 (sorry for merging the change in a single commit).

technolojin (approved):
LGTM

a-maumau:
@technolojin The problem seems to be the following line:
```
constexpr double cos_threshold = std::cos(M_PI / 3.0);
```

According to https://en.cppreference.com/w/cpp/numeric/math/cos , `constexpr` for std::cos is supported in c++26.
(Some compiler may adapt it by their own.)
So I made the value hard coded in https://github.com/autowarefoundation/autoware_universe/pull/10649/commits/6072e22708bfae37d0f9ca0c94e56099a7e75512 .

I hope this will solve the CI issue.

PR #10650, opened 2025-05-17
fix(build-and-test*): use local mixin for colcon build and test

xmfcx:
## Description

- Resolves https://github.com/autowarefoundation/autoware-github-actions/issues/344

## How was this PR tested?

- https://github.com/autowarefoundation/autoware_universe/actions/runs/15087742105/job/42412454193#step:15:385

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


xmfcx:
- https://github.com/autowarefoundation/autoware-github-actions/pull/345 is merged while making the option true by default, this PR is not needed anymore.

PR #10655, opened 2025-05-19
fix(roi_pointcloud_fusion): add remap output option

badai-nguyen:
## Description
- This add parameters into roi_pointcloud_fusion to allow override output of node into unknown when the node fusion pointcloud with all classes of yolox model.
- Allow the node avoiding early skipping when pointcloud numbers over max threshold, and  fusing all pointcloud inside ROIs before refine the closest pointcloud for output clusters. Improve the cases as below figure 
image

- Add `max_object_size` of object to avoid fusing pointcloud at similar distance but in separate clusters. Issue illustrated as below 
image

- Merge after https://github.com/autowarefoundation/autoware_launch/pull/1450

## Related links

**Parent Issue:**

- TIER IV internal link: https://tier4.atlassian.net/browse/RT0-37158

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


Owen-Liuyuxuan:
@badai-nguyen 
Could you also help adding the parameters in the launcher?

badai-nguyen:
@Owen-Liuyuxuan I think the paremeters were added by this https://github.com/autowarefoundation/autoware_launch/pull/1450/files ?

Owen-Liuyuxuan:
@badai-nguyen 
Thank you. I didn't realize they have different names.

PR #10660, opened 2025-05-20
fix(gpu_monitor): correct gpu memory calculation

takam5f2:
## Description
GPU monitor has a function to report how much GPU memory is used. However, I found that it refers the wrong data source for memory consumption. I'm writing the PR

In the current implementation, [`nvmlUtilization_t`](https://docs.nvidia.com/deploy/nvml-api/structnvmlUtilization__t.html#structnvmlUtilization__t
)-based result is referred as GPU memory usage. `nvmlDeviceGetUtilizationRates` method is called in the [`nvml_gpu_monitor.cpp`](https://github.com/autowarefoundation/autoware_universe/blob/1de1753fca9a19e14364545411d8851b256a6714/system/autoware_system_monitor/src/gpu_monitor/nvml_gpu_monitor.cpp#L144C11-L144C40), and [`itr->utilization.memory`]( https://github.com/autowarefoundation/autoware_universe/blob/1de1753fca9a19e14364545411d8851b256a6714/system/autoware_system_monitor/src/gpu_monitor/nvml_gpu_monitor.cpp#L289) is used as the GPU memory usage.

According to  [`nvmlUtilization_t`](https://docs.nvidia.com/deploy/nvml-api/structnvmlUtilization__t.html#structnvmlUtilization__t
), `itr->utilization.memory` means that `Percent of time over the past sample period during which global (device) memory was being read or written.` It does NOT mean FOOTPRINT of GPU memory.

For the expected behavior, it should be replaced by the result from nvmlDeviceGetMemoryInfo method. This PR would like to fixed it.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


nishikawa-masaki:
The PR title should be modified to "fix(gpu_monitor): fix gpu memory calculation #10660", I think.

system/autoware_system_monitor/src/gpu_monitor/nvml_gpu_monitor.cpp
```diff
@@ -295,7 +303,7 @@ void GPUMonitor::checkMemoryUsage(diagnostic_updater::DiagnosticStatusWrapper &
 
     stat.add(fmt::format("GPU {}: status", index), load_dict_.at(level));
     stat.add(fmt::format("GPU {}: name", index), itr->name);
-    stat.addf(fmt::format("GPU {}: usage", index), "%d.0%%", itr->utilization.memory);
+    stat.addf(fmt::format("GPU {}: usage", index), "%.1f%%", (usage * 100.0));
```
nishikawa-masaki, line 306: Though I believe the modified line produces the same format of string before and after the fix, please make it sure that there is no change, because even a small difference of data format may surprise the subscribers.

For your information, unit testing for gpu_monitor (and other nodes in autoware_system_monitor) has been disabled since the change of infra-structure from ROS to ROS2.
Enabling unit testing again requires considerable amount of effort.
Therefore, I believe you don't have to do it.

nishikawa-masaki: I consulted with ChatGPT about the code.
(I'm using **ChatGPT Teams** plan so that I can keep secrets. Please DO NOT use _free_ version of ChatGPT for work.)

It pointed that "%d.0%%" is ambiguous for both human readers and compilers and the code may result in an error depending on the environment.
* It seems to be expected to produce "GPU 1: usage 10.0%" when "itr->utilization.memory" is 10.
  * This seems to be the intent of the original author.
  * For this output format, ChatGPT recommends "stat.add(fmt::format("GPU {}: usage: {}.0%", index, itr->utilization.memory));" because the intent of the code is clearer.
  * Do you get this result with the original code?
 * It also can be recognized as you are expecting "decimal number with no sub-fraction" as you do with "%f.0" for float.
   * "%d" doesn't allow this kind of precision-specifier and it is invalid.
   * However, some compilers may handle this format string with error or undefined-behavior. So, it should be avoided.
 
ChatGPT prefers your fix in [872357b](https://github.com/autowarefoundation/autoware_universe/commit/872357b3b6a9556a7d4506306fd928c58e98ca27) because:
* The intent of the code is clearer for human readers.
* It is safe for compilers.
* It gives "sub-fraction" precision.
* As "usage" is between 0.0 and 1.0, you don't have to worry about exponential expression.

"stat.add(fmt::format("GPU {}: usage: {:.1f}%", index, (usage * 100.0f));" is also valid and is clearer.

Your fix in [e220f60](https://github.com/autowarefoundation/autoware_universe/commit/e220f609d6c48b33210af631bf3a606847d416bd) makes me feel a bit uneasy because "unsigned int" is passed to "%d".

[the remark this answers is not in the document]
nishikawa-masaki: Please see the next discussion on the same line.

nishikawa-masaki:
> The PR title should be modified to "fix(gpu_monitor): fix gpu memory calculation https://github.com/autowarefoundation/autoware_universe/pull/10660", I think.

Sorry if my comment confused you.
My main point was this PR is a fix (of a bug), not a change for chore.
I think PR name should include "fix" so that everyone can understand that a problem was fixed by the PR.

You seem to have named the branch as "chore/fix-gpu-memory-calculation".
I don't think you need to rename it as changing the name of a "already pushed" branch is troublesome.

system/autoware_system_monitor/src/gpu_monitor/nvml_gpu_monitor.cpp
nishikawa-masaki: Thank you for conducting a test.
It looks good to me.

nishikawa-masaki (approved):
Now, it looks good to me.

PR #10665, opened 2025-05-20
feat(diagnostic_graph_utils): show the error graph on the terminal only when it changes

takayuki5168:
## Description

Currently, when the MRM is triggered, the error graph (the cause of the diag triggering MRM) is visualized on the rviz.
However, there is no error message on the terminal, and it's hard to know the cause when we cannot access the rviz for the low-level debugging.

This PR changed the diagnostic_graph_utils package to show the error graph on the terminal only when it changes to suppress the frequent log.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


isamu-takagi:
@takayuki5168 There is no log showing when the level is OK, so we can't know when the error is resolved. Please add logs with level OK if necessary for analysis.

PR #10668, opened 2025-05-21
fix: fix set previous control command to zero when we transit to driving

Owen-Liuyuxuan:
## Description

In PID controller, set previous control command to zero when we transit from stopped to driving.

This is copied from the code of transiting from stopping to driving, and tested in Planning Simulation.

This would slightly increase the launch speed.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


shmpwk (approved):
LGTM

PR #10677, opened 2025-05-21
ci(build-and-test*): use self hosted fleet

xmfcx:
## Description

Now that we have all these strong runners:
image

image

Let's start using them!

This PR:
- Updates the string input of the runner to JSON string format.
   - Reason: https://github.com/orgs/community/discussions/11692#discussioncomment-3541856
   - TLDR: apparently, this is how you pass string arrays as a string
- Passes `"['self-hosted', 'Linux', 'X64']"` instead of codebuild for build-and-test-differential-cuda and clang-tidy-differential-cuda jobs.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/.github/workflows/build-and-test-daily.yaml
+++ b/.github/workflows/build-and-test-daily.yaml
@@ -33,7 +33,7 @@ jobs:
   build-and-test-daily-arm64:
     uses: ./.github/workflows/build-and-test-reusable.yaml
     with:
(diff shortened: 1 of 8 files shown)


xmfcx:
It was successful!

image

Now reverting the dummy change.

mitsudome-r (approved):
LGTM!

xmfcx:
Let's goo

PR #10681, opened 2025-05-22
fix(mission checking if goal footprint is in ego lanes

mitukou1109:
## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


mitukou1109:
I added some tests for `utils::get_arc_length_on_path()`, but it seems the changes should be made in #449. I will move them later.

sasakisasaki:
@mitukou1109 Thank you so much!! Honestly at the beginning of review, it was hard for me to understand why the provided fix is needed. Now I understand as follows.

- If the goal position is located on the **_duplicated_** lane, assuming no fix by this PR, the difference would be remaining even the goal position is inside of the lane because the duplicated part will remain as non-empty.

sasakisasaki (approved):
As I understand as above :+1:

kosuke55:
@mitukou1109 
Is it correct that when using
boost::geometry::difference(goal_footprint, ego_lanes, difference);
the overlapped parts in ego_lanes might be subtracted multiple times, potentially causing the difference to remain non-empty

kosuke55 (approved):

mitukou1109:
> @mitukou1109 Is it correct that when using
> 
> ```
> boost::geometry::difference(goal_footprint, ego_lanes, difference);
> ```
> 
> the overlapped parts in ego_lanes might be subtracted multiple times, potentially causing the difference to remain non-empty?

Are you referring to [this line](https://github.com/mitukou1109/autoware_universe/blob/654c3b777ea4d43df7637b0c5ceb08a8886beccd/planning/autoware_mission_planner_universe/src/lanelet2_plugins/default_planner.cpp#L220) or the original implementation [here](https://github.com/autowarefoundation/autoware_universe/blob/b354190d55832b28a623055d381a0924e6569836/planning/autoware_mission_planner_universe/src/lanelet2_plugins/default_planner.cpp#L218)?

mitukou1109:
> @mitukou1109 Is it correct that when using
> 
> ```
> boost::geometry::difference(goal_footprint, ego_lanes, difference);
> ```
> 
> the overlapped parts in ego_lanes might be subtracted multiple times, potentially causing the difference to remain non-empty?

As `boost::geometry::difference` returns `MultiPolygon2d`, I guess the difference is calculated for each element of `ego_lanes`, and the set of non-empty differences is returned. Therefore, even if one lane completely contains the footprint, the result will not be empty if there are other lanes where the footprint extrudes.

kosuke55:
https://evaluation.ci.tier4.jp/evaluation/reports/472c9a0b-8868-5610-b0a4-4ba00345bec2?project_id=prd_jt 3952/3952

PR #10690, opened 2025-05-23
fix(cmake): update spconv availability messages to use STATUS and WAR…

yukkysaito:
## Description
Since the following output was shown despite not being an error, I have made a correction.
```
--- stderr: autoware_tensorrt_plugins
spconv is available!
---
Finished <<< autoware_tensorrt_plugins [1.26s]

```

## Related links

**Parent Issue:**

None

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


amadeuszsz (approved):
LGTM

PR #10694, opened 2025-05-26
fix(lane_change, behavior_path_planner): fix failure to do lane change

mkquda:
## Description

When a new instance of lane_change module is launched, the module status is set to IDLE, since this status is not handled explicitly in the lane change implementation, it can sometimes result in undefined behavior.

This PR fixes the issue by setting the status of lane_change module to WAITING_APPROVAL when the module is constructed.

This PR also fixes a bug in the planner manager: the `deleted_modules` was not being set, which results in new instances of a module being created before the old instance and its output is cleared

## Related links

**Parent Issue:**

- Link

## How was this PR tested?
- PSIM
- TIER IV Internal Evaluator (Pending)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


soblin:
related PR (previous version)
https://github.com/autowarefoundation/autoware_universe/pull/8117

planning/behavior_path_planner/autoware_behavior_path_planner/src/planner_manager.cpp
```diff
@@ -850,6 +851,7 @@ SlotOutput SubPlannerManager::runApprovedModules(
       results.erase(m->name());
       debug_info_.scene_status.emplace_back(
         m, SceneModuleUpdateInfo::Action::DELETE, "From Approved");
+      deleted_modules.push_back(m);
```
soblin, line 854:
```suggestion
      // NOTE(soblin): m is copied, so it is okay to call deleteExpiredModules
      deleted_modules.push_back(m);
```

soblin (approved):

PR #10697, opened 2025-05-27
feat(mission_planner): print route state when set_route fails

kosuke55:
## Description

print route state when set_route fails

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

psim

check follwoing in terminal
```
  const auto message = fmt::format("state_.state: {}", routeStateToString(state_.state));
  RCLCPP_INFO(get_logger(), message.c_str());
```

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/autoware_mission_planner_universe/src/mission_planner/mission_planner.cpp
+++ b/planning/autoware_mission_planner_universe/src/mission_planner/mission_planner.cpp
@@ -41,7 +41,7 @@ namespace
   case RouteState::state:       \
     return #state;
 
-std::string routeStateToString(const uint8_t state)
+std::string route_state_to_string(const uint8_t state)
 {
   switch (state) {
     ROUTE_STATE_CASE(UNKNOWN)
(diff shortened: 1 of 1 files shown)


planning/autoware_mission_planner_universe/src/mission_planner/mission_planner.cpp
```diff
@@ -34,6 +35,31 @@
...
+    ROUTE_STATE_CASE(ARRIVED)
+    ROUTE_STATE_CASE(ABORTED)
+    ROUTE_STATE_CASE(INTERRUPTED)
+    default:
+      return "UNKNOWN(" + std::to_string(static_cast<int>(state)) + ")";
+  }
+}
+
+#undef ROUTE_STATE_CASE
+}  // namespace
```
kosuke55, line 62: FYI:
we can't not use enum_magic bevause RouteState is not enum but uint8_t

kosuke55:
@isamu-takagi @takayuki5168 
could you review again?

PR #10719, opened 2025-05-27
fix(control_evaluator): add check route handler readiness

namiota:
## Description

`RouteHandler` has field  `lanelet_map_ptr_` which is `nullptr` when object is created using default constructor. This field is set only when `setMap(..)` method call. Using `RouteHandler` without checking if handler is ready can cause segmentation fault error. 

This fix adds check before function call that uses `RouteHandler`. 

## Related links

**Parent Issue:**

- None

## How was this PR tested?

- run planning and control modules
- set goal
- restart control launch.

Some of tests with this sequence result in segmentation fault error:
```
ERROR] [control_evaluator-3]: process has died [pid 22050, exit code -11 (...)].
```

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/evaluator/autoware_control_evaluator/src/control_evaluator_node.cpp
+++ b/evaluator/autoware_control_evaluator/src/control_evaluator_node.cpp
@@ -435,7 +435,7 @@ void ControlEvaluatorNode::onTimer()
   }
 
   // add boundary distance metrics
-  if (odom && behavior_path) {
+  if (odom && behavior_path && route_handler_.isHandlerReady()) {
     const Pose ego_pose = odom->pose.pose;
     AddBoundaryDistanceMetricMsg(*behavior_path, ego_pose);
   }


danielsanchezaran:
@namiota DCO is failing https://github.com/autowarefoundation/autoware_universe/pull/10719/checks?check_run_id=42959009312 Can you follow the instructions given in the link to fix the issue? basically, you need to sign your commits to pass the check. 

@xtk8532704  shouldn't the ` AddStopDeviationMetricMsg(*odom);` line also have a check for odometry? Since the AddxxxMetricMsg methods are all void, cannot you move the pointer checks inside the methods themselves? Alternatively, you could check if the odom msg exists and if not, you can skip a whole lot of if's that depend on odom evaluating to true.  CodeScene evaluation shows the code complexity is getting higher https://github.com/autowarefoundation/autoware_universe/pull/10719/checks?check_run_id=42959008959

PR #10722, opened 2025-05-28
fix(autoware_multi_object_tracker): bug fix of anchor point

technolojin:
## Description

~~1. anchor point position was wrong between side point and edge point~~
2. invalid anchor point of (0,0) need to consider the size by absolute value
3. when the association was done with opposite object direction, the anchor point also need to be turned

----
PS
1 was my misunderstand. However, readability should be improved.

## Related links

previous PRs

https://github.com/autowarefoundation/autoware_universe/pull/2238
- original PR
- the bug (1,3) has been created

https://github.com/autowarefoundation/autoware_universe/pull/7170
- implementing size update
- the bug (1,3) persists

https://github.com/autowarefoundation/autoware_universe/pull/10203
- refactoring
- the bug (1,3) persists
- the bug (2) has been created

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


perception/autoware_multi_object_tracker/lib/object_model/shapes.cpp
```diff
@@ -183,16 +183,16 @@ void getNearestCornerOrSurface(
   if (xl > length / 2.0) {
     anchor_x = length / 2.0;
   } else if (xl > -length / 2.0) {
-    anchor_x = 0;
-  } else {
     anchor_x = -length / 2.0;
```
ogrex, line 186: I still don't fully understand this part. The original version seems more intuitive to me. Could you explain the fix(For ex. why anchor should be -length / 2.0 instead of 0 here )?

perception/autoware_multi_object_tracker/lib/tracker/model/vehicle_tracker.cpp
```diff
@@ -260,8 +260,29 @@ bool VehicleTracker::measure(
...
+    const double updating_yaw = tf2::getYaw(updating_object.pose.orientation);
+    double yaw_diff = updating_yaw - this_yaw;
+    while (yaw_diff > M_PI) yaw_diff -= 2 * M_PI;
+    while (yaw_diff < -M_PI) yaw_diff += 2 * M_PI;
+    if (std::abs(yaw_diff) > M_PI_2) {
+      tf2::Quaternion q;
+      q.setRPY(0, 0, updating_yaw + M_PI);
+      updating_object.pose.orientation = tf2::toMsg(q);
+      updating_object.anchor_point.x = -updating_object.anchor_point.x;
+      updating_object.anchor_point.y = -updating_object.anchor_point.y;
```
YoshiRi, line 275: Currently we can keep it. If we want to use this in 3D rotation, we may need something like:

```
tf2::Quaternion q_orig;
tf2::fromMsg(updating_object.pose.orientation, q_orig);

tf2::Quaternion q_flip;
q_flip.setRotation(tf2::Vector3(0, 0, 1), M_PI);  // π rad = 180°

tf2::Quaternion q_new = q_orig * q_flip;  // locally rotate 180deg
q_new.normalize();
```

IMO this option should not activated for computation resource issue.

YoshiRi (commented):
Can we share some sample bug case?

perception/autoware_multi_object_tracker/lib/object_model/shapes.cpp
technolojin: fixed https://github.com/autowarefoundation/autoware_universe/pull/10722/commits/f969c1df3820685c12503a6e35b731ef8ace5ad9

ogrex (approved):
LGTM

PR #10723, opened 2025-05-28
fix(cuda_pointcloud_preprocessor): reset data when receiving zero siz…

keita1523:
## Description
When a zero-sized point cloud was received, the previous frame’s point cloud was mistakenly published. I added a reset to handle this case.

## Related links
[TIER IV INTERNAL URL](https://tier4.atlassian.net/browse/RT0-37470?focusedCommentId=224997)

## How was this PR tested?
In the above URL.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


mojomex (commented):
@keita1523 Thank you for the fix!
~It seems DCO is failing because your commit was not signed off with` git commit -s`, please refer to the [Autoware docs](https://autowarefoundation.github.io/autoware-documentation/main/contributing/pull-request-guidelines/commit-guidelines/#sign-off-your-commits-required-automated) for more information. [Here](https://github.com/autowarefoundation/autoware_universe/pull/10723/checks?check_run_id=43009205138) are the instructions to fix the issue.~

As for the fix itself, `data.reset()` will set `data = nullptr` (it is a special CUDA pointcloud type, and `data` is not a `std::vector` but a pointer to CUDA memory).

I think the safest fix would be to revert the changed line, and to remove the `if (num_output_points > 0)` here, leaving its contents to be always executed, even with 0 points:

```c++
  if (num_output_points > 0) {
    extractPointsLaunch(
      device_transformed_points, device_ring_outlier_mask, device_indices, num_organized_points_,
      reinterpret_cast<OutputPointType *>(output_pointcloud_ptr_->data.get()), threads_per_block_,
      blocks_per_grid, stream_);
  }
```
:arrow_right: 
```c++
    extractPointsLaunch(
      device_transformed_points, device_ring_outlier_mask, device_indices, num_organized_points_,
      reinterpret_cast<OutputPointType *>(output_pointcloud_ptr_->data.get()), threads_per_block_,
      blocks_per_grid, stream_);
```

keita1523:
Thank you for your response.

```
-  if (num_output_points > 0) {
-    extractPointsLaunch(
-      device_transformed_points, device_ring_outlier_mask, device_indices, num_organized_points_,
-      reinterpret_cast<OutputPointType *>(output_pointcloud_ptr_->data.get()), threads_per_block_,
-      blocks_per_grid, stream_);
-  }
+  extractPointsLaunch(
+    device_transformed_points, device_ring_outlier_mask, device_indices, num_organized_points_,
+    reinterpret_cast<OutputPointType *>(output_pointcloud_ptr_->data.get()), threads_per_block_,
+    blocks_per_grid, stream_);
+
 ```
I tried the fix, but it didn’t work as expected. Unnecessary point clouds are still being published.
I think the reason might be that when a zero-sized point cloud is received, even if extractPointsLaunch is enabled, the `for` loop ([here](https://github.com/autowarefoundation/autoware_universe/blob/7403d0d8f52ba4056920d353a1d9f20f7e0b25f1/sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_pointcloud_preprocessor/common_kernels.cu#L83)) doesn’t run at all, so the output variable isn’t updated.

**Comment by mojomex** [comment]
Okay, now I'm properly confused :sweat_smile::

```c++
if (num_raw_points_ == 0) {
    output_pointcloud_ptr_->data.reset();
    return;
  }
```
This definitely results in a `nullptr`. It just never causes a problem because `num_raw_points_` is `0` and thus the `extractPointsKernel` does not ever actually access that pointer.

However, this fix should have no effect at all because:

In this function

https://github.com/autowarefoundation/autoware_universe/blob/85a868339285e573b5f1ab684d0c3f503b27d52a/sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor_node.cpp#L236-L237

the output of `cuda_pointcloud_preprocessor.cu` is moved from, so all contents of the object are destroyed (actually, primitive fields such as width etc. probably remain but are reset in `preprocess` anyway). The pointer to CUDA memory is definitely invalidated (see [CUDA blackboard](https://github.com/autowarefoundation/cuda_blackboard/blob/1837689df2891f6223f07c178c21aed252566ede/src/cuda_pointcloud2.cpp#L42-L43)):
```c++
    data = std::move(other.data);
    other.data = nullptr;
```

https://github.com/autowarefoundation/autoware_universe/blob/85a868339285e573b5f1ab684d0c3f503b27d52a/sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor_node.cpp#L295-L301

Then, after publishing, the whole `output_pointcloud_ptr_` is replaced by a new one, leaving no chance to leave any old points for the next iteration:

https://github.com/autowarefoundation/autoware_universe/blob/85a868339285e573b5f1ab684d0c3f503b27d52a/sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor_node.cpp#L316-L316

I currently have no time to compile and test with pilot-auto.x1 so I cannot really reproduce and debug the issue myself. Could you share a video or graph of the pointcloud sizes going into and coming out of cuda preprocessor to prove that it is at fault? :bow:

manato:
@keita1523
Thank you for issuing this PR & discussing the problem. As we discussed offline, I added some modifications onto your branch, including diff contents I shared before to eliminate the root cause of the problem.
Would you mind testing the latest code of this branch using your data to see if the problem is solved?

keita1523:
@manato 
I followed the instructions and tested the changes.
[TIER IV INTERNAL URL](https://tier4.atlassian.net/browse/RT0-37470?focusedCommentId=225852)
As a result, the unnecessary point cloud is removed, and the original issue with "Unknown" no longer occurs.
Thank you.

I reopen this PR.

amadeuszsz (approved):

PR #10730, opened 2025-05-29
feat(tier4_state_rviz_plugin): add control state visualization panel

kyoichi-sugahara:
## Description

The Operation Mode currently displayed in the state panel references information from `/api/operation_mode/state`, but this topic is only published when a state transition occurs in the `OperationModeTransitionManager`.
 
Therefore, when performing rosbag-based analysis, depending on when the rosbag recording started, this topic may not be included, making it difficult to understand the current operation mode.

Since there is a need to check the vehicle's current control state during rosbag analysis, as a workaround using non-API information, I would like to display `/vehicle/status/control_mode` in the state panel to assist with analysis.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

switch control mode and confirmed mode is properly changed.
[Screencast from 2025年05月29日 14時42分52秒.webm](https://github.com/user-attachments/assets/515e1c4b-9482-46eb-b9a8-53d10d574bb5)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/visualization/tier4_state_rviz_plugin/README.md
+++ b/visualization/tier4_state_rviz_plugin/README.md
@@ -17,6 +17,7 @@ This plugin also can engage from the panel.
 | `/api/motion/state`                      | `autoware_adapi_v1_msgs::msg::MotionState`                     | The topic represents the state of motion                      |
 | `/api/autoware/get/emergency`            | `tier4_external_api_msgs::msg::Emergency`                      | The topic represents the state of external emergency          |
 | `/vehicle/status/gear_status`            | `autoware_vehicle_msgs::msg::GearReport`                       | The topic represents the state of gear                        |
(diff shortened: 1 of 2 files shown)


kyoichi-sugahara:
@h-ohta @isamu-takagi 
Could I ask either of you two to review this PR?

isamu-takagi:
@kyoichi-sugahara AutowareStatePanel is currently used as the default Autoware RViz UI, so please avoid adding dependencies other than the API. If necessary, create a separate plugin.

kyoichi-sugahara:
@isamu-takagi 
Thank you for the review!
Is showing the additional plugin by default an option?

kyoichi-sugahara:
close this PR to avoid no API information for default RViz plugin

PR #10743, opened 2025-06-02
feat: add adaptive covariance threshold for tracker lifecycle management

linick-kang:
## Description

This feature makes the position covariance threshold dynamically adapt based on the target's distance from the ego vehicle and its roughly estimated size in the bird's-eye view (BEV).
Components of adaptive covariance threshold
- Base threshold: Accounts for sensor noise and detection model uncertainty.
- BEV size influence: Reflects errors in the target's shape or size estimation.
- Distance influence: Compensates for increased noise and reduced information at longer distances.
If distance to ego vehicle is unavailable, fallback to a preset value (same as current configuration)

BEV_size_and_distance_influence_binned

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

- Qualitative evaluation with Lsim

https://github.com/user-attachments/assets/f2599f2e-0694-451f-8f90-0add7f5f09bc

- Quantitative evaluation

[T4_evaluator_result](https://evaluation.tier4.jp/evaluation/reports/43f3ae6f-b771-5d3d-af52-a9a64277a1d4?project_id=prd_jt)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


perception/autoware_multi_object_tracker/lib/tracker/model/tracker_base.cpp
```diff
@@ -285,6 +285,44 @@ void Tracker::getPositionCovarianceEigenSq(
...
+  const double distance = getDistanceToEgo();
+  if (distance <= 0.0) return fallback_threshold;
+
+  constexpr double MAX_BEV_AREA = 20.0;  // [m^2]
+  const double clamped_bev_area = std::min(getBEVArea(), MAX_BEV_AREA);
+
+  // bev_area_influence ranges from 0 to 1 (normalized quadratic curve)
+  // distance_influence ranges from 0.16 to 3 for distance 0 to 120 m (offset sigmoid curve)
+  const double bev_area_influence = 0.03 * std::pow(clamped_bev_area, 2) / 12.0;
+  const double distance_influence = 6.0 / (1 + std::exp(-0.03 * (distance - 120.0)));
```
ogrex: we can make these magic numbers named variables for better understanding

perception/autoware_multi_object_tracker/lib/tracker/model/tracker_base.cpp
```diff
@@ -285,6 +285,44 @@ void Tracker::getPositionCovarianceEigenSq(
...
+    const auto & p = object_.pose.position;
+    const auto & e = ego_pose_->position;
+    return std::hypot(p.x - e.x, p.y - e.y);
+  }
+}
+
+double Tracker::computeAdaptiveThreshold(double base_threshold, double fallback_threshold) const
+{
+  const double distance = getDistanceToEgo();
+  if (distance <= 0.0) return fallback_threshold;
```
ogrex: Just to confirm—although a distance of 0 might not occur, this return value will jump at the boundary of 0. Is this intended behavior?

perception/autoware_multi_object_tracker/lib/tracker/model/tracker_base.cpp
```diff
@@ -285,6 +285,44 @@ void Tracker::getPositionCovarianceEigenSq(
...
+  const auto & dims = object_.shape.dimensions;
+  return dims.x * dims.y;
+}
+
+double Tracker::getDistanceToEgo() const
+{
+  constexpr double INVALID_DISTANCE = -1.0;
+  if (!ego_pose_) {
+    return INVALID_DISTANCE;
+  } else {
```
ogrex: "else" unnecessary

perception/autoware_multi_object_tracker/src/processor/processor.cpp
```diff
@@ -45,13 +45,15 @@ TrackerProcessor::TrackerProcessor(
...
-void TrackerProcessor::predict(const rclcpp::Time & time)
+void TrackerProcessor::predict(
+  const rclcpp::Time & time, const std::optional<geometry_msgs::msg::Pose> & ego_pose)
 {
   std::unique_ptr<ScopedTimeTrack> st_ptr;
   if (time_keeper_) st_ptr = std::make_unique<ScopedTimeTrack>(__func__, *time_keeper_);
 
   for (auto itr = list_tracker_.begin(); itr != list_tracker_.end(); ++itr) {
     (*itr)->predict(time);
+    (*itr)->setEgoPose(ego_pose);
```
ogrex: Actually, all trackers share the same ego_pose, rather than each tracker having its own. It seems more natural for the ego_pose to be owned by the TrackerProcessor and updated just once.
Just my personal opinion though — feel free to handle it however you prefer.

ogrex (changes requested):

perception/autoware_multi_object_tracker/lib/tracker/model/tracker_base.cpp
linick-kang: done.

perception/autoware_multi_object_tracker/lib/tracker/model/tracker_base.cpp
linick-kang: done. I excluded equal condition.

perception/autoware_multi_object_tracker/lib/tracker/model/tracker_base.cpp
linick-kang: done.

perception/autoware_multi_object_tracker/src/processor/processor.cpp
linick-kang: Thank you for good suggestion. It's much better to store just one ego_pose instead of copy it to each tracker.

perception/autoware_multi_object_tracker/lib/tracker/model/tracker_base.cpp
```diff
@@ -285,7 +285,50 @@ void Tracker::getPositionCovarianceEigenSq(
...
+
+double Tracker::getDistanceToEgo(const std::optional<geometry_msgs::msg::Pose> & ego_pose) const
+{
+  constexpr double INVALID_DISTANCE = -1.0;
+  if (!ego_pose) {
+    return INVALID_DISTANCE;
+  }
+  const auto & p = object_.pose.position;
+  const auto & e = ego_pose->position;
+  return std::hypot(p.x - e.x, p.y - e.y);
```
technolojin: return squared value for calculation efficiency
```suggestion
  double dx = p.x - e.x;
  double dy = p.y - e.y;
  return dx * dx + dy * dy
```

perception/autoware_multi_object_tracker/lib/tracker/model/tracker_base.cpp
```diff
@@ -285,7 +285,50 @@ void Tracker::getPositionCovarianceEigenSq(
...
+}
+
+double Tracker::computeAdaptiveThreshold(
+  double base_threshold, double fallback_threshold,
+  const std::optional<geometry_msgs::msg::Pose> & ego_pose) const
+{
+  const double distance = getDistanceToEgo(ego_pose);
```
technolojin:
```suggestion
  const double distance_sq = getDistanceToEgo(ego_pose);
```

perception/autoware_multi_object_tracker/lib/tracker/model/tracker_base.cpp
```diff
@@ -285,7 +285,50 @@ void Tracker::getPositionCovarianceEigenSq(
...
+  constexpr double DISTANCE_GAIN = 6.0;         // Max value added by distance influence
+  constexpr double DISTANCE_SLOPE = 0.03;       // Slope of sigmoid
+  constexpr double DISTANCE_OFFSET = 120.0;     // Inflection point of sigmoid
+
+  const double clamped_bev_area = std::min(getBEVArea(), MAX_BEV_AREA);
+
+  // bev_area_influence ranges from 0 to 1 (normalized quadratic curve)
+  // distance_influence ranges from 0.16 to 3 for distance 0 to 120 m (offset sigmoid curve)
+  const double bev_area_influence =
+    BEV_AREA_GAIN * std::pow(clamped_bev_area, 2) / BEV_AREA_NORMALIZER;
```
technolojin:
```suggestion
   constexpr double BEV_AREA_NORMALIZER_INV = 1 / BEV_AREA_NORMALIZER;
    BEV_AREA_GAIN * clamped_bev_area * clamped_bev_area * BEV_AREA_NORMALIZER_INV;
```

perception/autoware_multi_object_tracker/lib/tracker/model/tracker_base.cpp
```diff
@@ -285,7 +285,50 @@ void Tracker::getPositionCovarianceEigenSq(
...
+  const double distance_influence =
+    DISTANCE_GAIN / (1.0 + std::exp(-DISTANCE_SLOPE * (distance - DISTANCE_OFFSET)));
```
technolojin: avoid to use division, exponential

even if the accuracy is lower, this threshold should be fine

perception/autoware_multi_object_tracker/lib/tracker/model/tracker_base.cpp
```diff
@@ -285,7 +285,50 @@ void Tracker::getPositionCovarianceEigenSq(
...
+  constexpr double INVALID_DISTANCE = -1.0;
+  if (!ego_pose) {
+    return INVALID_DISTANCE;
+  }
+  const auto & p = object_.pose.position;
+  const auto & e = ego_pose->position;
+  return std::hypot(p.x - e.x, p.y - e.y);
+}
+
+double Tracker::computeAdaptiveThreshold(
```
technolojin, line 307: I think it is better to have calculated thresholds per distance, and return them.

In my opinion, it is fine even the returning threshold is not continuous by distance.

perception/autoware_multi_object_tracker/lib/tracker/model/tracker_base.cpp
linick-kang: I keep current one for reason explained above.

perception/autoware_multi_object_tracker/lib/tracker/model/tracker_base.cpp
linick-kang: Thanks for this nice advice. I now use cache to store calculated adaptive threshold components with binned distance/bev_area for faster process.

perception/autoware_multi_object_tracker/lib/tracker/model/tracker_base.cpp
linick-kang: Thanks for the suggestion. I currently keep the square root for distance calculation for easier cache search. 
(squared value may need more comparison process)

perception/autoware_multi_object_tracker/lib/tracker/model/tracker_base.cpp
linick-kang: Thank you for the suggestion. I replaced divide and power process with alternatives in calculation.

perception/autoware_multi_object_tracker/lib/tracker/model/tracker_base.cpp
linick-kang: done.

ogrex:
You can install pre-commit locally to run this check on your machine

technolojin (approved):

PR #10745, opened 2025-06-03
feat(planning_validator): subscribe additional topic for collision detection

satoshi-ota:
## Description

This PR enhances the planning_validator module by subscribing to an additional topic for improved collision detection. By incorporating more data sources, the system can perform more accurate validation of planned trajectories, contributing to safer autonomous driving behavior.

```c++
  autoware_utils::InterProcessPollingSubscriber<PointCloud2> sub_pointcloud_{
    this, "~/input/pointcloud", autoware_utils::single_depth_sensor_qos()};
  autoware_utils::InterProcessPollingSubscriber<
    LaneletRoute, autoware_utils::polling_policy::Newest>
    sub_route_{this, "~/input/route", rclcpp::QoS{1}.transient_local()};
  autoware_utils::InterProcessPollingSubscriber<
    LaneletMapBin, autoware_utils::polling_policy::Newest>
    sub_lanelet_map_bin_{this, "~/input/lanelet_map_bin", rclcpp::QoS{1}.transient_local()};
```

## Related links

**Parent Issue:**

none

## How was this PR tested?

- [x] Psim
- [x] [PASS TIER IV INTERNAL SCENARIOS](https://evaluation.tier4.jp/evaluation/reports/5c4fd567-6913-545a-8582-ac7abcd33c07?project_id=prd_jt)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/planning_validator/autoware_planning_validator/src/node.cpp
+++ b/planning/planning_validator/autoware_planning_validator/src/node.cpp
@@ -111,6 +111,8 @@ void PlanningValidatorNode::setData()
   data->current_acceleration = sub_acceleration_.take_data();
   data->obstacle_pointcloud = sub_pointcloud_.take_data();
   data->set_current_trajectory(sub_trajectory_.take_data());
+  data->set_route(sub_route_.take_data());
+  data->set_map(sub_lanelet_map_bin_.take_data());
 }
 
 void PlanningValidatorNode::onTimer()
--- a/planning/planning_validator/autoware_planning_validator/src/node.cpp
+++ b/planning/planning_validator/autoware_planning_validator/src/node.cpp
@@ -109,6 +109,7 @@ void PlanningValidatorNode::setData()
   auto & data = context_->data;
   data->current_kinematics = sub_kinematics_.take_data();
   data->current_acceleration = sub_acceleration_.take_data();
+  data->obstacle_pointcloud = sub_pointcloud_.take_data();
   data->set_current_trajectory(sub_trajectory_.take_data());
 }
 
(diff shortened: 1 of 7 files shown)


planning/planning_validator/autoware_planning_validator/src/node.cpp
```diff
@@ -109,7 +109,10 @@ void PlanningValidatorNode::setData()
   auto & data = context_->data;
   data->current_kinematics = sub_kinematics_.take_data();
   data->current_acceleration = sub_acceleration_.take_data();
+  data->obstacle_pointcloud = sub_pointcloud_.take_data();
   data->set_current_trajectory(sub_trajectory_.take_data());
+  data->set_route(sub_route_.take_data());
+  data->set_map(sub_lanelet_map_bin_.take_data());
```
mkquda, line 115: I wonder if we need to do something like [this](https://github.com/autowarefoundation/autoware_universe/pull/10750/files#diff-5d4d6c3df9de823d46867bfb645eea730ebd487634f385a943541656f6eac75bR142-R167), this is how route and map are updated in `behavior_path_planner` node.

mkquda (approved):

planning/planning_validator/autoware_planning_validator/src/node.cpp
satoshi-ota: @mkquda Thanks.
> I wonder if we need to do something like [this](https://github.com/autowarefoundation/autoware_universe/pull/10750/files#diff-5d4d6c3df9de823d46867bfb645eea730ebd487634f385a943541656f6eac75bR142-R167), this is how route and map are updated in behavior_path_planner node.

Previously, we used callback functions with subscribers, and this implementation was necessary to ensure thread safety in multi-threaded processing. However, with the introduction of the polling subscriber, callback functions for subscribers are no longer needed, so I understand that this implementation is no longer necessary.

PR #10746, opened 2025-06-03
fix(run_out): guard against decreasing ego trajectory times

maxime-clem:
## Description

Fix crashes that can occur when the calculate times along the ego trajectory are decreasing.

## Related links

- [TIER IV internal link](https://tier4.atlassian.net/browse/RT0-37423)

## How was this PR tested?

Psim

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


kyoichi-sugahara (approved):
Thank you for the quick fix! and LGTM!

PR #10748, opened 2025-06-03
fix(cuda_pointcloud_preprocessor): ensure ordered twist/imu queues

mojomex:
## Description

Although the twist/IMU handling logic in the CUDA preprocessor should be equivalent to the CPU version, there are often occurrences of the twist and IMU queues becoming unordered.

This PR ensures upon insertion of new messages that they are inserted in-order, even if they arrive out-of-order.

As to _why_ they arrive out-of-order: I have no idea. I have confirmed the insertion algorithms are logically identical to the CPU version, and the same rosbag has no issues on the CPU version. Even the same polling subscribers are used.
If anyone has a clue, please let me know :disguised_face: 

## How was this PR tested?

Tested with Pilot Auto X2 launch setup and [TIER IV Internal](https://console.mob.tier4.jp/projects/x2_dev/rosbag?viz=stream&agg=count&from_ts=1746340146&to_ts=1748932146&live=true&query=log_file_name%3A%28cfa23601-97c4-4d47-a37e-edfc7e080d8c_2025-05-29-17-00-12_p0900_11%29&file_id=e4c77db5-5700-4442-9317-9be5e85a77ca) rosbag, see below.
The time jumps observed in both videos are the result of my pitiful attempt at video compression and are not happening in reality.

<table>
<tr>
<th>Before</th><th>After</th>
</tr>
<tr>
<td>

https://github.com/user-attachments/assets/748a39cf-86e5-4213-a29b-1ac683b04fb7

</td>
<td>

https://github.com/user-attachments/assets/5b465538-a60b-4296-b953-195c3761286e

</td>
</tr>
</table>

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor_node.cpp
+++ b/sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor_node.cpp
@@ -204,10 +204,12 @@ void CudaPointcloudPreprocessorNode::imuCallback(
   const sensor_msgs::msg::Imu::ConstSharedPtr imu_msg)
 {
   while (!angular_velocity_queue_.empty()) {
+    /* *INDENT-OFF* */
     // for rosbag replay
-    bool backwards_time_jump_detected =
-      (rclcpp::Time(angular_velocity_queue_.front().header.stamp) >
-       rclcpp::Time(imu_msg->header.stamp));
+    bool backwards_time_jump_detected = rclcpp::Time(angular_velocity_queue_.front().header.stamp) >
+                                        rclcpp::Time(imu_msg->header.stamp);
+    /* *INDENT-ON* */
+
     bool is_queue_longer_than_1s =
       rclcpp::Time(angular_velocity_queue_.front().header.stamp) <
       rclcpp::Time(imu_msg->header.stamp) - rclcpp::Duration::from_seconds(1.0);


sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor_node.cpp
```diff
@@ -193,25 +191,18 @@ void CudaPointcloudPreprocessorNode::twistCallback(
       break;
     }
   }
+
+  auto it = std::lower_bound(
+    twist_queue_.begin(), twist_queue_.end(), twist_msg_ptr->header.stamp,
+    [](const auto & twist, const auto & stamp) {
+      return rclcpp::Time(twist.header.stamp) < stamp;
```
amadeuszsz, line 198: nit:
Do we need `rclcpp::Time` casting here and in `imuCallback` down below?

mojomex: Yes, the rosidl stuff has no comparators defined. Casting one to rclcpp time implicitly casts the other as well.
If trying to remove the cast, this happens:
```
error: no match for ‘operator<’ (operand types are ‘const _stamp_type’ {aka ‘const builtin_interfaces::msg::Time_<std::allocator<void> >’} and ‘const builtin_interfaces::msg::Time_<std::allocator<void> >’)
  198 |       return twist.header.stamp < stamp;
```

mojomex:
@mitsudome-r @youtalk @xmfcx Hi, the CUDA build is [failing](https://github.com/autowarefoundation/autoware_universe/actions/runs/15458796550/job/43519969726#step:9:30) on `Get modified packages` for no obvious reason, would one of you mind taking a look? I already tried re-running multiple times. Thank you :bow:

xmfcx:
@mojomex just rebasing fixed it 👍

PR #10750, opened 2025-06-03
feat(planning_validator): implement redundant collision prevention feature when ego makes a turn at intersection

mkquda:
## Description

When ego turns left/right at intersections, if object prediction/tracking is not available for any reason, there is a chance of collision with oncoming/merging vehicles.

This PR adds a layer of redundancy to check for possible collisions when turning left/right at intersections.
A new plugin module is added to planning_validator node to verify trajectory has no collision, the logic uses obstacle pointcloud so it is robust to prediction errors.

Example 1:

https://github.com/user-attachments/assets/a0e99cf1-bdf7-4d39-a8d7-34e6321a008e

Example 2:

https://github.com/user-attachments/assets/31c090e1-8a60-45c2-ba4f-8306fe3b008d

Example 3:

https://github.com/user-attachments/assets/c6383dcb-ebf4-452a-b134-c513744a80c7

## Related links

- [Launcher PR](https://github.com/autowarefoundation/autoware_launch/pull/1485)
- [Internal TIER IV Link](https://tier4.atlassian.net/wiki/spaces/CT/pages/3627878954/RTL4)

## How was this PR tested?
- PSIM.
- [Internal TIER IV Evaluator](https://evaluation.tier4.jp/evaluation/reports/8e2d4b67-41d2-58f3-a014-c26d74a57d69?project_id=prd_jt)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

redundant collision prevention featurte at intersections.


planning/planning_validator/autoware_planning_validator_collision_checker/package.xml
```diff
@@ -0,0 +1,35 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+  <name>autoware_planning_validator_collision_checker</name>
```
satoshi-ota: Given that this feature is intended for detecting collisions at intersections, would it make sense to rename it to `autoware_planning_validator_intersection_collision_checker`?

planning/planning_validator/autoware_planning_validator_collision_checker/src/collision_checker.cpp
```diff
@@ -0,0 +1,523 @@
+// Copyright 2025 Tier IV, Inc.
```
satoshi-ota:
```suggestion
// Copyright 2025 TIER IV, Inc.
```

planning/planning_validator/autoware_planning_validator_collision_checker/src/collision_checker.cpp
```diff
@@ -0,0 +1,523 @@
...
+    ec.extract(cluster_idx);
+    return cluster_idx;
+  });
+
+  for (const auto & indices : cluster_indices) {
+    PointCloud::Ptr cluster(new PointCloud);
+    bool cluster_above_height_threshold{false};
+    for (const auto & index : indices.indices) {
+      const auto & p = (*input)[index];
+      cluster_above_height_threshold |= (p.z > params_.pointcloud.clustering.min_height);
```
satoshi-ota: This point cloud has already been transformed and is in the map coordinate frame, right? In that case, instead of comparing the z value directly to a parameter, shouldn't we compare the difference between the z value and the ego vehicle's `base_link` z coordinate against the parameter value?

planning/planning_validator/autoware_planning_validator_collision_checker/include/autoware/planning_validator_collision_checker/collision_checker.hpp
```diff
@@ -0,0 +1,93 @@
...
+
+#include <memory>
+#include <string>
+
+namespace autoware::planning_validator
+{
+using sensor_msgs::msg::PointCloud2;
+using PointCloud = pcl::PointCloud<pcl::PointXYZ>;
+
+class CollisionChecker : public PluginInterface
```
satoshi-ota: How about naming it `IntersectionCollisionChecker`?

planning/planning_validator/autoware_planning_validator_collision_checker/src/collision_checker.cpp
```diff
@@ -0,0 +1,523 @@
...
+        object.velocity = 0.0;        // too high acceleration, reset velocity
+        object.track_duration = 0.0;  // reset track duration
+      } else {
+        object.velocity = autoware::signal_processing::lowpassFilter(
+          raw_velocity, object.velocity, 0.5);  // apply low-pass filter to velocity
+      }
+      object.last_update_time = new_data.last_update_time;
+      object.pose = new_data.pose;
+      object.distance_to_overlap = new_data.distance_to_overlap;
+      object.ttc = object.distance_to_overlap / object.velocity;
```
satoshi-ota: Given that point cloud processing can introduce a delay of several hundred milliseconds, it might be advisable to implement delay compensation. Without such compensation, the TTC could be overestimated. However, it's also acceptable to handle this in a separate PR after reviewing the results of the evaluation experiment.

planning/planning_validator/autoware_planning_validator_collision_checker/src/collision_checker.cpp
```diff
@@ -0,0 +1,523 @@
...
+  clock_ = node.get_clock();
+  logger_ = node.get_logger();
+  context_ = context;
+
+  setup_parameters(node);
+
+  setup_diag();
+}
+
+void CollisionChecker::setup_parameters(rclcpp::Node & node)
```
satoshi-ota: How about trying out [generate_parameter_library](https://github.com/PickNikRobotics/generate_parameter_library)? It eliminates the need to define parameter structs manually, and also auto-generates code for tasks like updating all parameters. It also supports parameter validation. That said, it's totally fine to handle this in a separate PR.

satoshi-ota (commented):
It looks generally good to me. However, I’ve added a few comments regarding some points I’d like to clarify.

planning/planning_validator/autoware_planning_validator_collision_checker/package.xml
mkquda: Yes you are right, will rename it

planning/planning_validator/autoware_planning_validator_collision_checker/include/autoware/planning_validator_collision_checker/collision_checker.hpp
mkquda: will rename it

planning/planning_validator/autoware_planning_validator_collision_checker/src/collision_checker.cpp
mkquda: I see, okay I will make a separate PR to introduce delay compensation

planning/planning_validator/autoware_planning_validator_collision_checker/src/collision_checker.cpp
mkquda: will do in a separate PR

planning/planning_validator/autoware_planning_validator_collision_checker/src/collision_checker.cpp
mkquda: @satoshi-ota Yes we should, you are right!
(Btw this implementation is based on `rear_obstacle_checker` so need to fix over there too)

satoshi-ota: Actually, I also noticed the same mistake last week. :laughing:

planning/planning_validator/autoware_planning_validator_collision_checker/src/utils.cpp
```diff
@@ -0,0 +1,237 @@
...
+      rclcpp::Duration(ego_traj.front_traj[overlap_index->first].time_from_start).seconds();
+    overlap_time.second =
+      rclcpp::Duration(ego_traj.back_traj[overlap_index->second].time_from_start).seconds();
+    if (overlap_time.first > time_horizon) continue;
+    lanelets.target_lanelets.emplace_back(
+      ll.id(), extend_lanelet(ll, overlap_point), overlap_point, overlap_time);
+  }
+}
+
+void set_left_turn_target_lanelets(
```
maxime-clem: I did not try to deeply understand the logic of the functions, but why are the `set_right_turn_target_lanelets` and `set_left_turn_target_lanelets` so different ? I would expect the only change to be the check on the lanelet turn direction.

planning/planning_validator/autoware_planning_validator_collision_checker/src/collision_checker.cpp
```diff
@@ -0,0 +1,523 @@
...
+#include <lanelet2_core/geometry/Point.h>
+#include <lanelet2_core/geometry/Polygon.h>
+#include <lanelet2_core/primitives/BoundingBox.h>
+#include <lanelet2_routing/RoutingGraph.h>
+
+#ifdef ROS_DISTRO_GALACTIC
+#include <tf2_eigen/tf2_eigen.h>
+#else
+#include <tf2_eigen/tf2_eigen.hpp>
+#endif
```
maxime-clem: This is no longer needed since we dropped support for `GALACTIC`.

planning/planning_validator/autoware_planning_validator_collision_checker/src/collision_checker.cpp
```diff
@@ -0,0 +1,523 @@
...
+    filter.filter(*filtered_pointcloud);
+  }
+
+  if (filtered_pointcloud->empty()) return;
+
+  {
+    geometry_msgs::msg::TransformStamped transform_stamped;
+    try {
+      transform_stamped = context_->tf_buffer.lookupTransform(
+        "map", input->header.frame_id, input->header.stamp, rclcpp::Duration::from_seconds(0.1));
```
maxime-clem: There can be some performance issue when looking up the transform. Here at most you can have a 100ms delay. This can be avoided with replacing `input->header.stamp` with `tf2::TimePointZero` but this can lead to some error in the transform.

planning/planning_validator/autoware_planning_validator_collision_checker/src/collision_checker.cpp
```diff
@@ -0,0 +1,523 @@
...
+    const auto pcd_object = get_pcd_object(time_stamp, filtered_point_cloud, target_lanelet);
+    if (!pcd_object.has_value()) continue;
+
+    auto update_object = [&](PCDObject & object, const PCDObject & new_data) {
+      const auto dt = (new_data.last_update_time - object.last_update_time).seconds();
+      const auto dl = object.distance_to_overlap - new_data.distance_to_overlap;
+      if (dt < 1e-6) return;  // too small time difference, skip update
+
+      object.track_duration += dt;
+      const auto max_accel = 30.0;
```
maxime-clem: Should this be a parameter ?

planning/planning_validator/autoware_planning_validator_collision_checker/src/collision_checker.cpp
```diff
@@ -0,0 +1,523 @@
...
+      } else {
+        context_->debug_pose_publisher->pushPointMarker(
+          existing_object.pose.position, "collision_checker_pcd_objects", 1);
+      }
+    }
+  }
+
+  auto itr = history_.begin();
+  while (itr != history_.end()) {
+    if ((clock_->now() - itr->second.last_update_time).seconds() > 1.0) {
```
maxime-clem:
```suggestion
    constexpr auto max_history_time = 1.0;
    if ((clock_->now() - itr->second.last_update_time).seconds() > max_history_time) {
```

maxime-clem (commented):
Documentation is missing (docstrings and README) but otherwise the code looks good overall.

planning/planning_validator/autoware_planning_validator_collision_checker/src/collision_checker.cpp
```diff
@@ -0,0 +1,523 @@
...
+  if (!context_->data->obstacle_pointcloud) {
+    return skip_validation("point cloud data is not available, skipping collision check.");
+  }
+
+  if (!context_->data->route_handler->isHandlerReady()) {
+    return skip_validation("route handler is not ready, skipping collision check.");
+  }
+
+  PointCloud::Ptr filtered_pointcloud(new PointCloud);
+  filter_pointcloud(context_->data->obstacle_pointcloud, filtered_pointcloud);
```
satoshi-ota: Since this point cloud processing is heavy, I think it should be executed only when absolutely necessary. Wouldn't it be better to run it right before the `collision_check`, for example?

planning/planning_validator/autoware_planning_validator_collision_checker/src/collision_checker.cpp
```diff
@@ -0,0 +1,523 @@
...
+
+  if (lanelets.trajectory_lanelets.empty()) {
+    return skip_validation("failed to get trajectory lanelets, skipping collision check.");
+  }
+
+  if (lanelets.target_lanelets.empty()) {
+    return skip_validation("failed to get target lanelets, skipping collision check.");
+  }
+
+  context_->validation_status->is_valid_collision_check = check_collision(
```
satoshi-ota: Is the feature to skip collision detection when an upper-level module has already inserted a stop point before the conflicting lane planned to be handled in a separate PR?

satoshi-ota: Ah, sorry. I just found a function that performs that processing. -> `get_ego_trajectory`

planning/planning_validator/autoware_planning_validator_collision_checker/src/collision_checker.cpp
```diff
@@ -0,0 +1,523 @@
...
+  params_.pointcloud.clustering.max_size =
+    get_or_declare_parameter<int>(node, "collision_checker.pointcloud.clustering.max_size");
+}
+
+void CollisionChecker::setup_diag()
+{
+  context_->add_diag(
+    "intersection_validation_collision_check",
+    context_->validation_status->is_valid_collision_check, "risk of collision at intersection turn",
+    params_.is_critical);
```
satoshi-ota: Is it set up to output to the syslog when a collision is detected? Doing so would make it much easier to later understand the situation in which the collision was detected using the `MobilityObserver`, which would greatly simplify the analysis.

planning/planning_validator/autoware_planning_validator_collision_checker/src/utils.cpp
mkquda: @maxime-clem for right turns we want to check lanelets that the ego trajectory cross/overlap with, for left turns we only want to check lanelets that merge in the same lane we are turning into, so the logic to compute target lanelets is different

planning/planning_validator/autoware_planning_validator_collision_checker/src/collision_checker.cpp
mkquda: @maxime-clem 

> but this can lead to some error in the transform.

That's not okay is it? :sweat_smile:

planning/planning_validator/autoware_planning_validator_collision_checker/src/collision_checker.cpp
mkquda: I think its okay not to be a parameter, I don't expect it needs to be tuned

planning/planning_validator/autoware_planning_validator_collision_checker/src/collision_checker.cpp
mkquda: @satoshi-ota planning_validator will output to syslog [here](https://github.com/autowarefoundation/autoware_universe/pull/10750/files#diff-5d4d6c3df9de823d46867bfb645eea730ebd487634f385a943541656f6eac75bR282), do you think it is sufficient?

planning/planning_validator/autoware_planning_validator_collision_checker/src/collision_checker.cpp
mkquda: makes sense

mkquda:
> Documentation is missing (docstrings and README) but otherwise the code looks good overall.

@maxime-clem due to time limit, I will add them in a later PR

planning/planning_validator/autoware_planning_validator_collision_checker/src/collision_checker.cpp
maxime-clem: It's not okay but we do not really have good solutions for this as far as I know.
There is this old ticket where I wrote my understanding of the issue: https://tier4.atlassian.net/browse/RT0-33566

planning/planning_validator/autoware_planning_validator_collision_checker/src/utils.cpp
maxime-clem: Does this mean we assume right-hand traffic ? I think this should be adjustable with a parameter to allow the use of the module with left-hand traffic.

maxime-clem: (this is fine to add the parameter in a future PR)

planning/planning_validator/autoware_planning_validator_collision_checker/src/collision_checker.cpp
maxime-clem: This can be further investigated later.

planning/planning_validator/autoware_planning_validator_collision_checker/src/collision_checker.cpp
satoshi-ota: I think it's sufficient :+1:

planning/planning_validator/autoware_planning_validator_collision_checker/src/collision_checker.cpp
mkquda: resolved in https://github.com/autowarefoundation/autoware_universe/pull/10750/commits/6a66d1ad3c51bddd8d57b190c44e21900bf70316

planning/planning_validator/autoware_planning_validator_collision_checker/src/collision_checker.cpp
mkquda: resolved in https://github.com/autowarefoundation/autoware_universe/pull/10750/commits/6a66d1ad3c51bddd8d57b190c44e21900bf70316

maxime-clem (approved):
Tested some intersection scenarios and could observe the validation status switch to `is_valid_collision_check: false`.

planning/planning_validator/autoware_planning_validator_intersection_collision_checker/src/intersection_collision_checker.cpp
```diff
@@ -0,0 +1,532 @@
...
+  set_lanelets_debug_marker(lanelets);
+
+  if (turn_direction == Direction::NONE) return;
+
+  if (lanelets.trajectory_lanelets.empty()) {
+    return skip_validation("failed to get trajectory lanelets, skipping collision check.");
+  }
+
+  if (lanelets.target_lanelets.empty()) {
+    return skip_validation("failed to get target lanelets, skipping collision check.");
```
satoshi-ota, line 160: It seems that this warning is also being displayed even when the distance to the intersection is long — is this the intended behavior?
Since the warning keeps appearing continuously in the terminal, it might be better to change it to a DEBUG message or something similar.

It's fine to make the modification in a separate PR.

mkquda: once the trajectory reaches a turn direction lane, the logic is executed, even if the distance is still long.
I will change it to DEBUG

satoshi-ota (approved):

PR #10751, opened 2025-06-04
feat(simulator_launch): add an option to disable to launch occupancy grid map

Owen-Liuyuxuan:
## Description

The occupancy launch will conflict with autoware's perception module when we simultaneously launch autoware's perception module along with the simulator.

It would be easier for an integrator to work around the system if we could select whether the occupancy grid map will be launched or not.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

Additional launch argument of `perception/enable_occupancy_grid_map` with `default=True` to preserve the original  behavior.

## Effects on system behavior

None.


paulsohn:
@Owen-Liuyuxuan cc @KeisukeShima @taikitanaka3 @TakaHoribe @takayuki5168 @tkimura4

As we decided to migrate tier4 launchers into autoware_launch repository,
a separate PR for tier4 launchers must be created unless this PR is merged before 2026/02/10 07:00 UTC (16:00 JST)

* https://github.com/orgs/autowarefoundation/discussions/6780
* https://github.com/autowarefoundation/autoware/issues/6774

PR #10754, opened 2025-06-04
feat(run_out): option to preserve parts of ignored predicted paths

maxime-clem:
## Description

When ignoring objects or predicted paths, this PR adds options to set some distance and duration that is still used for collision detection. This allows to keep some small parts of the predicted paths to guarantee safety.

Launch PR: https://github.com/autowarefoundation/autoware_launch/pull/1476

## Related links

- [TIER IV internal link](https://tier4.atlassian.net/browse/RT1-9581)

## How was this PR tested?

[TIER IV Evaluator](https://evaluation.tier4.jp/evaluation/reports/3ac53984-eb09-5dcc-b6ff-7b4b00722122?project_id=prd_jt).

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/motion_velocity_planner/autoware_motion_velocity_run_out_module/README.md
+++ b/planning/motion_velocity_planner/autoware_motion_velocity_run_out_module/README.md
@@ -67,6 +67,9 @@ First, we only keep predicted paths that have a confidence value above the `conf
 If, `confidence_filtering.only_use_highest` is set to `true` then for each object only the predicted paths that have the higher confidence value are kept.
 Next, the remaining predicted paths are cut according to the segments prepared in the previous step.
 
+To guarantee that parts of the predicted paths are never ignored,
+parameters `preserved_duration` and `preserved_distance` can be used to set a minimum duration and/or distance that cannot be cut or ignored.
+
 The following figures shows an example where crosswalks are used to ignore pedestrians and to cut their predicted paths.
 
 | debug markers (`objects_footprints`)                        | objects of interest                                     |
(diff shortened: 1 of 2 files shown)


planning/motion_velocity_planner/autoware_motion_velocity_run_out_module/src/objects_filtering.cpp
```diff
@@ -172,43 +180,96 @@ void calculate_predicted_path_footprints(
...
+  auto t = 0.0;
+  auto dist = 0.0;
+  const auto & ls = footprint.predicted_path_footprint.corner_linestrings[front_left];
+  for (; cut_index + 1 < ls.size(); ++cut_index) {
+    const auto & segment = universe_utils::Segment2d(ls[cut_index], ls[cut_index + 1]);
+    t += footprint.time_step;
+    dist += static_cast<double>(boost::geometry::length(segment));
+    if ((!cut_time || t > cut_time) && (!cut_distance || dist > cut_distance)) {
+      break;
+    }
```
mkquda: this condition is a bit confusing ...
just to clarify the meaning:
- if both thresholds are provided -> both values need to satisfy **(is this intended?)**
- if either threshold is provided -> only provided threshold need to be satisfied

maxime-clem: Yes your understanding is correct.
I reworked the implementation and now this logic is inside the `get_preserved_cut_index()`  function, which is only used to the get the index corresponding to the new parameters (`preserved_time` and `preserved_distance`).
Since the goal of these parameters is to guarantee _at least_ some duration and/or distance, we want both values to be satisfied when they are set.

PR #10760, opened 2025-06-05
chore(run_out): add alqudah.mohammad@tier4.jp as codeowner

maxime-clem:
## Description

Add @mkquda as codeowner of the `run_out` module for the `motion_velocity_planner`.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


isamu-takagi (changes requested):
@maxime-clem CODEOWNERS file is auto-updated based on the maintainer in package.xml, so this changes will be overwritten. Could you update package.xml in run_out module instead?

maxime-clem:
@isamu-takagi Thank you I forgot about this system. I reopened the PR: https://github.com/autowarefoundation/autoware_universe/pull/10762

PR #10770, opened 2025-06-06
feat(tier4_simulator_launch): add use_pointcloud_container argument

mitsudome-r:
## Description
**This PR must be merged after https://github.com/autowarefoundation/autoware_launch/pull/1479**

We are currently trying to run simulator.launch.xml as standalone launch in order to run Autoware in multi-container setup in this issue. However, probabilistic_occupancy_grid_map.launch.xml fails to launch properly because it cannot find pointcloud_container, which is not created in multi-container settings. (Check [my comment](https://github.com/autowarefoundation/autoware/pull/6184#issuecomment-2948129756) for the details)

This PR adds `use_pointcloud_container` to tier4_simulator launch so that users can configure whether to use pointcloud_container from launch argument to resolve the issue. 

It is set to true by default to keep the behavior when we launch this file from planning_simulator.launch.xml as before. 

## Related links

**Parent Issue:**
- https://github.com/orgs/autowarefoundation/projects/6?pane=issue&itemId=110566193&issue=autowarefoundation%7Cautoware%7C6140

Related PRs:
- https://github.com/autowarefoundation/autoware/pull/6184
- https://github.com/autowarefoundation/autoware_launch/pull/1479

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


youtalk (approved):
LGTM. Thank you!

PR #10771, opened 2025-06-06
feat(autoware_multi_object_tracker): unknown as static object

technolojin:
## Description

Provide user to choose whether estimate the unknown object motion estimation or not.

1. let user select if estimate motion (velocity) of unknown object or not.
2. adjust object shape (footprint) to be aligned to the input detected object
3. distinguish getTrackedObject for internal use or to publish

## Related links

[INTERNAL DISCUSSION](https://star4.slack.com/archives/C076ZGRC15X/p1748838403258199)

## How was this PR tested?

* before
try to estimate its shape, extrapolate by time. therefore, the cluster and the tracked object shape are not matched

https://github.com/user-attachments/assets/503cd104-2bba-4acb-a9c8-4b7a57ae4915

https://github.com/user-attachments/assets/d857f812-e0d2-48f3-beee-acaeccda1489

* after (with velocity estimation)
```
    enable_unknown_object_velocity_estimation: true
    enable_unknown_object_motion_output: false
```

it tracks position and velocity, but the shape is originated by the latest measurement. no shape extrapolation by time.
the cluster and the tracked object shape are matched all the time. 
user can select whether estimated velocity to be exported or not, by setting `enable_unknown_object_motion_output`

https://github.com/user-attachments/assets/caa0afb5-60c2-4a5e-a20d-c51c10068512

https://github.com/user-attachments/assets/b8ce9b24-61e8-47bd-a8fd-0c384b32ffa6

* after (static motion model)
```
    enable_unknown_object_velocity_estimation: false
    enable_unknown_object_motion_output: false
```
it dose not estimate velocity. 
the cluster and the tracked object shape are matched all the time

https://github.com/user-attachments/assets/3ba42ff0-feb2-4c52-bc7d-f8de1a3b6d18

https://github.com/user-attachments/assets/a6dce074-9448-4e19-b7f5-248ada4fb78e

## Notes for reviewers
need to be merged with https://github.com/autowarefoundation/autoware_launch/pull/1481

Default values of new parameters are
```
    enable_unknown_object_velocity_estimation: true
    enable_unknown_object_motion_output: false
``` 
It is recommended to enable velocity estimation for better moving object traceability, but the estimated velocity is not reliable since the velocity is mixture of shape changes and its actual motion.

## Interface changes

None.

## Effects on system behavior

None.


perception/autoware_multi_object_tracker/config/multi_object_tracker_node.param.yaml
```diff
@@ -15,6 +15,7 @@
     ego_frame_id: base_link
     enable_delay_compensation: false
     consider_odometry_uncertainty: false
+    enable_unknown_object_velocity_estimation: false
```
ogrex: default is true in launcher

perception/autoware_multi_object_tracker/lib/tracker/motion_model/static_motion_model.cpp
```diff
@@ -0,0 +1,196 @@
...
+#include <autoware_utils/math/unit_conversion.hpp>
+#include <autoware_utils/ros/msg_covariance.hpp>
+
+#include <tf2/utils.h>
+
+namespace autoware::multi_object_tracker
+{
+using autoware_utils::xyzrpy_covariance_index::XYZRPY_COV_IDX;
+
+StaticMotionModel::StaticMotionModel() : logger_(rclcpp::get_logger("StaticMotionModel"))
```
ogrex, line 33: The logger_ variable seems unused 
While it's understandable that this follows the pattern of other implementations, we could consider adding some meaningful logging to justify its presence. Alternatively, it might make more sense to move it into the parent class MotionModel.
This is a minor point, it's OK to keep it as is

ogrex (commented):
Tested locally and works as expected

perception/autoware_multi_object_tracker/config/multi_object_tracker_node.param.yaml
technolojin: the launcher side is turned to false by default
https://github.com/autowarefoundation/autoware_launch/pull/1481/commits/85aabd7e298ab2c5523a112bea4ef15dfae430cc

perception/autoware_multi_object_tracker/lib/tracker/motion_model/static_motion_model.cpp
technolojin: let me fix it in future PR

ogrex (approved):
Tested locally and LGTM

PR #10773, opened 2025-06-06
fix(autoware_adapi_specs): add autoware_adapi_v1_specs as dependency

mitsudome-r:
## Description
I have noticed that autoware_adapi_specs uses messages from autoware_adapi_v1_msgs and autoware_adapi_version_msgs, but package.xml does not state it as dependency.

## Related links

**Parent Issue:**

- None

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/common/autoware_adapi_specs/package.xml
+++ b/common/autoware_adapi_specs/package.xml
@@ -11,8 +11,8 @@
   <buildtool_depend>ament_cmake_auto</buildtool_depend>
   <buildtool_depend>autoware_cmake</buildtool_depend>
 
-  <depend>autoware_adapi_v1_msgs</depend>
-  <depend>autoware_adapi_version_msgs</depend>
+  <build_export_depend>autoware_adapi_v1_msgs</build_export_depend>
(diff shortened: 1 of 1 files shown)


common/autoware_adapi_specs/package.xml
```diff
@@ -11,6 +11,9 @@
   <buildtool_depend>ament_cmake_auto</buildtool_depend>
   <buildtool_depend>autoware_cmake</buildtool_depend>
 
+  <depend>autoware_adapi_v1_msgs</depend>
+  <depend>autoware_adapi_version_msgs</depend>
```
isamu-takagi: `build_export_depend` is more suitable because this package itself is not built.

```suggestion
  <build_export_depend>autoware_adapi_v1_msgs</build_export_depend>
  <build_export_depend>autoware_adapi_version_msgs</build_export_depend>
```

isamu-takagi (approved):
Thank you. LGTM

xmfcx:
Needs to be reopened in autoware_core since the package is moved.

PR #10784, opened 2025-06-09
feat(goal_planner): check pull over lane crossing validity when triggering candidate generation thread

soblin:
## Description

If lane_change is running when goal_planner's thread is generating path candidates, lane_change can suddenly swerve the path. If ego is still driving on the preferred_lanelets before swerving, and goal_planner still happened to be IDLE, syncWithThreads() is not triggered, thus the backgroud thread keeps using old reference path, which is aligned on the old preferred lanelets. This is problematic when pull over planners process "previous_module_path" w.r.t reference path.

To deal with this issue:
1. added the flag `lane_change_status_changed_` to goal_planner to signal background thread
2. each planner use `get_reference_lanelets_for_pullover` instead of `getPullOverLanes`
 
image

## Related links

- https://tier4.atlassian.net/browse/RT0-35453

## How was this PR tested?

- https://evaluation.tier4.jp/evaluation/reports/7fdf2fc0-e7b0-51ba-b36a-03955607323d?project_id=prd_jt&state=failed

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_path_planner/autoware_behavior_path_goal_planner_module/include/autoware/behavior_path_goal_planner_module/util.hpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_goal_planner_module/include/autoware/behavior_path_goal_planner_module/util.hpp
@@ -252,6 +252,23 @@ bool hasDeviatedFromPath(
  */
 bool has_stopline_except_terminal(const PathWithLaneId & path);
 
+/**
+ * @brief find the lanelet that has changed "laterally" from previous lanelet on the routing graph
+ * @return the lanelet that changed "laterally" if the path is lane changing, otherwise nullopt
+ */
+std::optional<lanelet::ConstLanelet> find_lane_change_completed_lanelet(
+  const PathWithLaneId & path, const lanelet::LaneletMapConstPtr lanelet_map,
+  const lanelet::routing::RoutingGraphConstPtr routing_graph);
+
+/**
+ * @brief generate lanelets with which pull over path is aligned
+ * @note if lane changing path is detected, this returns lanelets aligned with later part of the
+ * lane changing path
+ */
+lanelet::ConstLanelets get_reference_lanelets_for_pullover(
+  const PathWithLaneId & path, const std::shared_ptr<const PlannerData> & planner_data,
+  const double backward_length, const double forward_length);
+
 }  // namespace autoware::behavior_path_planner::goal_planner_utils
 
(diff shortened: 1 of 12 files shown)


planning/behavior_path_planner/autoware_behavior_path_goal_planner_module/include/autoware/behavior_path_goal_planner_module/util.hpp
```diff
@@ -252,6 +252,23 @@ bool hasDeviatedFromPath(
  */
 bool has_stopline_except_terminal(const PathWithLaneId & path);
 
+/**
+ * @brief find the lanelet that has changed "laterally" from previous lanelet on the routing graph
+ * @return the lanelet that changed "laterally" if the path is lane changing, otherwise nullopt
+ */
+std::optional<lanelet::ConstLanelet> find_lane_change_completed_lanelet(
+  const PathWithLaneId & path, const lanelet::LaneletMapConstPtr lanelet_map,
```
kosuke55, line 260: question: 

>  A    1  |  3  |  5 | 7
B    2  |  4  |  6 | 8
For example changing lanes from 1→8
1 → 3, 4 → 5,6 → 8

this function returns `4` as `lane_change_completed_lanelet` ? I imagined `8` from the function name.

planning/behavior_path_planner/autoware_behavior_path_goal_planner_module/src/util.cpp
```diff
@@ -1076,4 +1076,51 @@ bool has_stopline_except_terminal(const PathWithLaneId & path)
...
+  const auto & lanelet_map = planner_data->route_handler->getLaneletMapPtr();
+  const auto lane_change_complete_lane =
+    find_lane_change_completed_lanelet(path, lanelet_map, routing_graph);
+  if (!lane_change_complete_lane) {
+    return utils::getExtendedCurrentLanesFromPath(
+      path, planner_data, backward_length, forward_length,
+      /*forward_only_in_route*/ false);
+  }
+  return planner_data->route_handler->getLaneletSequence(
+    *lane_change_complete_lane, backward_length, forward_length);
```
kosuke55: this can't get lanes after goal?

soblin: fixed in b385a4ab5fa26cc7d2e7c2d6fc79990001856024

PR #10790, opened 2025-06-10
fix(multi_object_tracker): correct area calculation for cylinder shape in getArea function

technolojin:
## Description

Previous PR https://github.com/autowarefoundation/autoware_universe/pull/10744 contains area calculation cache.
The function getArea had a bug on area calculation.

- before
```c++
      return shape.dimensions.x * shape.dimensions.x * M_PI;
```
- after
```c++
      return shape.dimensions.x * shape.dimensions.x * M_PI * 0.25;
```

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


ogrex (approved):
LGTM

PR #10799, opened 2025-06-12
fix(planning_validator): fix conflict between trajectory shift and distance deviation checks

mkquda:
## Description

When `trajectory_checker` detects a "critical" error, then the appropriate behavior should be executed as specified by planning validator configs.
Currently if `trajectory_checker` detects another "noncritical" error simultaneously, the expected behavior might be overriden by the diagnostics published for the "noncritical" error.

This PR addresses the issue by using one common `is_critical_error_` flag for the trajectory diagnostics update instead of using a separate flag for each error type. Doing so will ensure that the handling method of critical error is prioritized in case multiple errors are detected at the same time

## Related links

None.

## How was this PR tested?
- PSIM.
- [Internal TIER IV Evaluator](https://evaluation.tier4.jp/evaluation/reports/626a9dcf-9f23-564f-8ec6-fee5ee192e21?project_id=prd_jt)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/planning_validator/autoware_planning_validator/README.md
+++ b/planning/planning_validator/autoware_planning_validator/README.md
@@ -179,3 +179,7 @@ The following parameters can be set for the `autoware_planning_validator`:
 | `validity_checks.trajectory_shift.forward_shift_th`  | double | max valid Longitudinal distance between two consecutive trajectories (measured at nearest point to ego) [m] | 1.0           |
 | `validity_checks.trajectory_shift.backward_shift_th` | double | min valid longitudinal distance between two consecutive trajectories (measured at nearest point to ego) [m] | 0.1           |
 | `validity_checks.trajectory_shift.is_critical`       | bool   | if true, will use handling type specified for critical checks                                               | true          |
+
+!!! warning
+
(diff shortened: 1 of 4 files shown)


kyoichi-sugahara (approved):

planning/planning_validator/autoware_planning_validator/README.md
```diff
@@ -179,3 +179,7 @@ The following parameters can be set for the `autoware_planning_validator`:
 | `validity_checks.trajectory_shift.forward_shift_th`  | double | max valid Longitudinal distance between two consecutive trajectories (measured at nearest point to ego) [m] | 1.0           |
 | `validity_checks.trajectory_shift.backward_shift_th` | double | min valid longitudinal distance between two consecutive trajectories (measured at nearest point to ego) [m] | 0.1           |
 | `validity_checks.trajectory_shift.is_critical`       | bool   | if true, will use handling type specified for critical checks                                               | true          |
+
+!!! warning
+
+    When a validity check with `is_critical = true` is triggered, AND `handling_type.critical` is set to apply soft stop on last valid trajectory, THEN any other validity check triggered simultanuously will publish a WARN level diagnostic instead of ERROR level. This is to avoid contradictory behavior for different types of checks.
```
mkquda:
```suggestion
    When a validity check with `is_critical = true` is triggered, AND `handling_type.critical` is set to apply soft stop on last valid trajectory, THEN any other validity check triggered simultaneously will publish a WARN level diagnostic instead of ERROR level. This is to avoid contradictory behavior for different types of checks.
```

PR #10800, opened 2025-06-12
feat(planning_validator): implement a collision detection feature for rearward objects of the vehicle

satoshi-ota:
## Description

This update introduces a new plugin called rear_collision_checker, which is responsible for detecting potential rear-end collisions based on predicted object motion and the ego vehicle's planned trajectory. The plugin is specifically designed to enhance safety when the vehicle is about to change lanes or merge into an adjacent lane, where rear-approaching vehicles may pose a threat. Key features include:

LIDAR-based obstacle detection: Directly analyzes raw point cloud data to identify rear-approaching objects not yet recognized by perception.

Nearest-point velocity estimation: Calculates relative velocity of closest rear obstacle points for more accurate collision prediction.

RSS-based risk metrics: Computes collision risk using estimated velocity and spatial-temporal proximity.

Dynamic object filtering: Ignores static or irrelevant objects based on motion and direction filtering.

These changes contribute to safer and more robust trajectory validation, especially in lane-change or merging scenarios. The goal is to mitigate risks in situations where object recognition is lost or when the planning module makes incorrect decisions.

This scenario represents an obstacle avoidance case where a vehicle is approaching from behind in the adjacent lane. The obstacle avoidance module detects the rear vehicle as well and, when there is a potential risk of collision, it avoids performing an evasive maneuver. Instead, it waits for the vehicle in the adjacent lane to pass before executing the avoidance. In this case, the module does not activate.

https://github.com/user-attachments/assets/62e0163c-4638-415f-a8d4-0a273c2b6ba9

On the other hand, if the avoidance module fails to operate correctly and performs a lane change in such a situation, this module detects the potential collision as shown below. (In this case, we deliberately disabled the safety check functionality of the avoidance module to intentionally simulate a situation where the planning module fails to make the correct decision.)

https://github.com/user-attachments/assets/9b99228e-11d9-476e-924e-21b34bb9c9d3

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

- [x] [TIER IV INTERNAL SCENARIOS (disconnect MRM)](https://evaluation.tier4.jp/evaluation/reports/2b7d94a6-8b8e-5158-9ad8-4f2629b8eb42?project_id=prd_jt)
- [x] [TIER IV INTERNAL SCENARIOS (connect MRM)](https://evaluation.tier4.jp/evaluation/reports/f1712149-6b08-53ff-9244-be975634230e?project_id=prd_jt)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/planning_validator/autoware_planning_validator_rear_collision_checker/param/rear_collision_checker_node_parameters.yaml
+++ b/planning/planning_validator/autoware_planning_validator_rear_collision_checker/param/rear_collision_checker_node_parameters.yaml
@@ -65,6 +65,10 @@ rear_collision_checker_node:
           type: double
           validation:
             gt_eq<>: [0.0]
+        max_acceleration:
+          type: double
+          validation:
+            gt_eq<>: [0.0]
       latency:
         type: double
         validation:
--- a/planning/planning_validator/autoware_planning_validator_rear_collision_checker/param/rear_collision_checker_node_parameters.yaml
+++ b/planning/planning_validator/autoware_planning_validator_rear_collision_checker/param/rear_collision_checker_node_parameters.yaml
@@ -105,16 +105,18 @@ rear_collision_checker_node:
         type: double
 
     ego:
-      footprint_buffer:
-        type: double
       reaction_time:
         type: double
         validation:
           gt<>: [0.0]
       min_velocity:
         type: double
+        validation:
+          gt<>: [0.0]
       max_velocity:
         type: double
+        validation:
+          gt<>: [0.0]
       max_acceleration:
         type: double
         validation:
@@ -127,6 +129,14 @@ rear_collision_checker_node:
           gt<>: [0.0]
       max_negative_jerk:
         type: double
+      nominal_deceleration:
+        type: double
+      nominal_positive_jerk:
+        type: double
+        validation:
+          gt<>: [0.0]
+      nominal_negative_jerk:
+        type: double
 
     blind_spot:
       check:
@@ -134,13 +144,6 @@ rear_collision_checker_node:
           type: bool
         right:
           type: bool
-      active_distance:
-        start:
-          type: double
-          validation:
-            gt<>: [0.0]
-        end:
-          type: double
       offset:
         inner:
           type: double
--- a/planning/planning_validator/autoware_planning_validator_rear_collision_checker/param/rear_collision_checker_node_parameters.yaml
+++ b/planning/planning_validator/autoware_planning_validator_rear_collision_checker/param/rear_collision_checker_node_parameters.yaml
@@ -92,7 +92,7 @@ rear_collision_checker_node:
       max_deceleration:
         type: double
 
-    vru:
+    vulnerable_road_user:
       reaction_time:
         type: double
         validation:
--- a/planning/planning_validator/autoware_planning_validator_rear_collision_checker/param/rear_collision_checker_node_parameters.yaml
+++ b/planning/planning_validator/autoware_planning_validator_rear_collision_checker/param/rear_collision_checker_node_parameters.yaml
@@ -15,11 +15,7 @@ rear_collision_checker_node:
           type: double
           validation:
             gt_eq<>: [0.0]
-        forward:
-          type: double
-          validation:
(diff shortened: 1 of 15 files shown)


planning/planning_validator/autoware_planning_validator_rear_collision_checker/param/rear_collision_checker_node_parameters.yaml
```diff
@@ -0,0 +1,158 @@
...
+        validation:
+          gt<>: [0.0]
+      max_velocity:
+        type: double
+        validation:
+          gt<>: [0.0]
+      max_deceleration:
+        type: double
+
+    vru:
```
mkquda: What is `vru`? is it a standard acronym :sweat_smile: 
maybe better to use full term?

planning/planning_validator/autoware_planning_validator_rear_collision_checker/src/rear_collision_checker.cpp
```diff
@@ -0,0 +1,888 @@
...
+      cluster->push_back(point);
+    }
+    if (!cluster_surpasses_threshold_height) continue;
+    // Make a 2d convex hull for the objects
+    pcl::ConvexHull<pcl::PointXYZ> hull;
+    hull.setDimension(2);
+    hull.setInputCloud(cluster);
+    std::vector<pcl::Vertices> polygons;
+    PointCloud::Ptr surface_hull(new PointCloud);
+    hull.reconstruct(*surface_hull, polygons);
```
mkquda: you don't use `std::vector<pcl::Vertices> polygons` anywhere so you can just call `hull.reconstruct(*surface_hull);`

planning/planning_validator/autoware_planning_validator_rear_collision_checker/src/rear_collision_checker.cpp
```diff
@@ -0,0 +1,888 @@
...
+    debug.cluster_points = obstacle_pointcloud;
+  }
+
+  constexpr double forward = 100.0;
+  constexpr double backward = 100.0;
+  const auto current_lanes = utils::get_current_lanes(context_, forward, backward);
+  const auto combine_lanelet = lanelet::utils::combineLaneletsShape(current_lanes);
+
+  if (current_lanes.empty()) {
+    debug.text = "failed to identify the current driving lane.";
```
mkquda, line 745: can return here?

mkquda (approved):
LGTM.

planning/planning_validator/autoware_planning_validator_rear_collision_checker/param/rear_collision_checker_node_parameters.yaml
satoshi-ota: it's Vulnerable Road User. But as you saild it might be unclear to some, so I'll revise it.

planning/planning_validator/autoware_planning_validator_rear_collision_checker/src/rear_collision_checker.cpp
satoshi-ota: fixed in https://github.com/autowarefoundation/autoware_universe/pull/10800/commits/91760ac54d47fbbeffdbb416957f5ff3f2926706

planning/planning_validator/autoware_planning_validator_rear_collision_checker/param/rear_collision_checker_node_parameters.yaml
satoshi-ota: fixed in https://github.com/autowarefoundation/autoware_universe/pull/10800/commits/acdf836707e1551296ac28f3fc2e6504bb3c2372

planning/planning_validator/autoware_planning_validator_rear_collision_checker/src/rear_collision_checker.cpp
satoshi-ota: fixed in https://github.com/autowarefoundation/autoware_universe/pull/10800/commits/9e6250222156462efb99df28f49a7944fa436756

PR #10801, opened 2025-06-12
fix(planning_validator): fix the lateral distance calculation

maxime-clem:
## Description

Distance deviation was incorrectly calculated as Euclidian distance from the nearest trajectory _point_ instead of lateral distance from nearest segment.
This PR fixes that.

Example impact:
| Before | After |
| --- | --- |
| image | image |

## Related links

**Private Links:**

- [TIER IV internal link](https://tier4.atlassian.net/browse/RT1-10026)
## How was this PR tested?

Psim
Perception reproducer

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/planning_validator/autoware_planning_validator/README.md
+++ b/planning/planning_validator/autoware_planning_validator/README.md
@@ -151,14 +151,14 @@ The following parameters can be set for the `autoware_planning_validator`:
 
 #### Deviation Check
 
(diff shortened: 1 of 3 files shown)


kyoichi-sugahara (approved):
Thanks!!! LGTM!

PR #10818, opened 2025-06-16
feat(planning_validator): add condition to check the yaw deviation

maxime-clem:
## Description

In general, yaw deviation is not caused by planning but by bad control performance.
The only times when planning should care about the yaw deviation is when changing the path, such as in lane change or avoidance situations.
Therefore, this PR adds a condition to the `planning_validator` for checking the yaw deviation.
The trajectory yaw nearest to the ego pose is calculated both on the current and on the previous trajectories. If this difference is above some threshold (new parameter), then we consider that the path changed and the yaw deviation (difference between the ego yaw and the current trajectory yaw) is checked.

## Related links

Launch PR: https://github.com/autowarefoundation/autoware_launch/pull/1494

**Private Links:**

- [TIER IV internal link](https://tier4.atlassian.net/browse/RT1-10025)
- 
## How was this PR tested?

Psim
Rosbags

## Notes for reviewers

None.

## Interface changes

None.

#### Additions and removals

| Added | `nearest_yaw_trajectory_shift_required_for_checking`   | `double` | `0.1`         | [rad] yaw deviation is only checked if the nearest yaw jumped by more than this value between the previous and current trajectory |

## Effects on system behavior

Yaw deviation is only checked when the path changes drastically.


kyoichi-sugahara:
@maxime-clem 
> The only times when planning should care about the yaw deviation is when changing the path, such as in lane change or avoidance situations.

can you include above description in readme so that other people can understand the feature's background?

kyoichi-sugahara (approved):
LGTM! Thanks.

PR #10822, opened 2025-06-16
fix(pid): fix a bug that acceleration feedback does not go in the correct direction when reverse

Owen-Liuyuxuan:
## Description

When the vehicle is reversing, the PID controller does not correctly handle the acceleration feedback.

`control_data.current_motion.acc = m_current_accel.accel.accel.linear.x;` directly comes from localization topics. So it will be minus, when we are accelerating in reverse mode.

`raw_ctrl_cmd.acc` comes from `applyVelocityFeedback`, where we have moderated the driving direction.  It will be plus when we are accelerating in reverse mode.

The computation of `acc_err` should revert the current acceleration when we are reversing.

## Related links

**Parent Issue:**

- Link

**Private Links:**

- [CompanyName internal link](https://tier4.atlassian.net/browse/RT0-37856)

## How was this PR tested?

Tested in  PSim.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/control/autoware_pid_longitudinal_controller/src/pid_longitudinal_controller.cpp
+++ b/control/autoware_pid_longitudinal_controller/src/pid_longitudinal_controller.cpp
@@ -864,8 +864,8 @@ PidLongitudinalController::Motion PidLongitudinalController::calcCtrlCmd(
 
     // calc acc feedback
     const double vel_sign = (control_data.shift == Shift::Forward)
-                            ? 1.0
-                            : (control_data.shift == Shift::Reverse ? -1.0 : 0.0);
+                              ? 1.0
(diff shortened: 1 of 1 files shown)


yuki-takagi-66 (approved):
Thank you so much. LGTM

yuki-takagi-66:
NOTE: vel_sign is not correct name because the gear direction and vehicle velocity is not one-to-one relation, for example on slope. But It is compatible name to the other similar code.

Owen-Liuyuxuan:
@yuki-takagi-66 
Many thanks for the review. I would love to fix it with a more correct name and improve the coding quality.
But I am considering cherry-picking this compilable fix into private development first.

PR #10831, opened 2025-06-17
fix(roi_cluster_fusion): fix target frame's timestamp

vividf:
## Description
Big thanks to @amadeuszsz  for finding this issue.

From https://github.com/autowarefoundation/autoware_universe/pull/9635 until now, we have only collected the camera info once. 

```
// Create the CameraProjection only when the camera info arrives for the first time.
  // This assume the camera info does not change while the node is running
  auto & det2d_status = det2d_status_list_.at(rois_id);
  if (det2d_status.camera_projector_ptr == nullptr && check_camera_info(*input_camera_info_msg)) {
    det2d_status.camera_projector_ptr = std::make_unique<CameraProjection>(
      *input_camera_info_msg, approx_grid_cell_w_size_, approx_grid_cell_h_size_,
      det2d_status.project_to_unrectified_image, det2d_status.approximate_camera_projection);
    det2d_status.camera_projector_ptr->initialize();

    std::unique_lock<std::mutex> fusion_collectors_lock(fusion_collectors_mutex_);
    for (auto & collector : fusion_collectors_) {
      collector->add_camera_projection(rois_id, det2d_status.camera_projector_ptr);
    }
  }
```
However, in [roi_cluster_fusion](https://github.com/autowarefoundation/autoware_universe/blob/26ed00ba7b72fd10b3115f51d9d16c0196cd0548/perception/autoware_image_projection_based_fusion/src/roi_cluster_fusion/node.cpp#L91) node it utilize the camera info timestamp for find the tf between `xxxx/camera_optical_link` (from camera_info) and `map` (cluster). Which will always change when the timestamp is changed

Also, in all of the other fusion-related nodes, we used the roi header timestamp to get the tf. For example, [roi_detected_object_fusion](https://github.com/autowarefoundation/autoware_universe/blob/26ed00ba7b72fd10b3115f51d9d16c0196cd0548/perception/autoware_image_projection_based_fusion/src/roi_detected_object_fusion/node.cpp#L97).

Thus, this PR also changes the camera info timestamp to roi timestamp (which is original same) but lead to different result for getting tf.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


technolojin (approved):
LGTM

badai-nguyen (approved):
@vividf 
As default, the clusters' `frame_id` is `base_link` which is fixed so that we did not notice the difference between using camera_info and roi_msg until now. 
Thanks for your raising and fixing.
LGTM

Issue #10830, opened 2025-06-17
autoware_system_monitor/process_monitor outputs huge integer for negative process priority


nishikawa-masaki:
### Checklist

- [x] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [x] I've searched other issues and no duplicate issues were found.
- [x] I'm convinced that this is not my fault but a bug.

### Description

After merging of #10379 , /diagnostics topic published by the "autoware_system_monitor/process_monitor" node sometimes contain huge integer values for "PR" (priority) keys in process data.
The versions before #10379 employed Linux command "top" for retrieving process data .
The "top" command uses "rt", which means "real-time", for special system processes with negative priority values smaller than -99.

As the values for "PR" is not used for further processing, this bug won't cause any serious problem.

Example:
> - level: "\0"
>   name: 'process_monitor: High-load Proc[10]'
>   message: OK
>   hardware_id: dpc2500001
>   values:
>   - key: COMMAND
>     value: migration/1
>   - key: '%CPU'
>     value: '1.0'
>   - key: '%MEM'
>     value: '0.0'
>   - key: PID
>     value: '160'
>   - key: USER
>     value: root
>   - key: PR
>     value: '**18446744073709551516**'
>   - key: NI
>     value: '0'

### Expected behavior

The "autoware_system_monitor/process_monitor" node should handle negative priority values accordingly and use special value "rt" for processes with negative priority smaller than -99 and positive priority larger than 999.
These conditions are derived from the source code of the "top" command.

Example:
> - level: "\0"
>   name: 'process_monitor: High-load Proc[10]'
>   message: OK
>   hardware_id: dpc2500001
>   values:
>   - key: COMMAND
>     value: migration/1
>   - key: '%CPU'
>     value: '1.0'
>   - key: '%MEM'
>     value: '0.0'
>   - key: PID
>     value: '160'
>   - key: USER
>     value: root
>   - key: PR
>     value: '**rt**'
>   - key: NI
>     value: '0'

### Actual behavior

See "Description".

### Steps to reproduce

1. Modify the "num_of_procs" parameter in install/autoware_system_monitor/share/autoware_system_monitor/config/process_monitor.param.yaml 
 so that more processes than the default (5) are reported. 40 is large enough.
2. On terminal 1, execute the following command.
> $  ros2 launch autoware_system_monitor system_monitor.launch.xml
[...]

PR #10832, opened 2025-06-17
fix: process_monitor should support negative priority

nishikawa-masaki:
## Description

This PR fixes the problem reported in #10830 .
When processes reported by autoware_system_monitor/process_monitor has negative priority values, the negative values are printed accordingly.

For priorities smaller than -99 or larger than 999, the special string "rt", which means "real-time", is used instead of numeric values.

## Related links

**Parent Issue:**

- #10830 

**Private Links:**

- [TIER IV internal link](https://tier4.atlassian.net/browse/RT1-10416)

## How was this PR tested?

- Executed the reproduction procedure described in #10830 and verified that negative priorities are output as expected.
- Added a unit-test case for the issue #10830 to verify the fix.
  - Before PR: Failed.
  - After PR: Succeeded.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/system/autoware_system_monitor/src/process_monitor/process_monitor.cpp
+++ b/system/autoware_system_monitor/src/process_monitor/process_monitor.cpp
@@ -591,7 +591,8 @@ void ProcessMonitor::fillTaskInfo(
   info.userName = convertUidToUserName(raw_p->status_info.real_uid);
   // For backward compatibility with the old implementation with Linux "top" command,
   // real-time processes need exceptional handling.
-  // Linux "top" command shows priority less than -99 and more than 999 as "rt", which means "real-time".
+  // Linux "top" command shows priority less than -99 and more than 999 as "rt", which means
+  // "real-time".
   if ((raw_p->stat_info.priority < -99) || (raw_p->stat_info.priority > 999)) {
     info.priority = "rt";
   } else {


system/autoware_system_monitor/include/system_monitor/process_monitor/process_information.hpp
```diff
@@ -42,9 +42,9 @@ struct StatInfo
   uint64_t c_maj_flt;
   uint64_t utime_tick;
   uint64_t stime_tick;
-  uint64_t c_utime_tick;
-  uint64_t c_stime_tick;
-  uint64_t priority;
+  int64_t c_utime_tick;
+  int64_t c_stime_tick;
+  int64_t priority;
```
nishikawa-masaki, line 47: Self comment:
c_utime_tick and c_stime_tick should be int64_t according to "man 5 proc", too. (Commented as "%ld")
These 2 members are not used after being read from /proc/[pid]/stat.

nishikawa-masaki (commented):
Just added a "self comment" for the reviewers.

system/autoware_system_monitor/test/src/process_monitor/test_process_monitor.cpp
```diff
@@ -232,15 +232,6 @@ class ProcessMonitorTestSuite : public ::testing::Test
...
   }
-
-  void modifyPath()
-  {
-    // Modify PATH temporarily
-    auto env = boost::this_process::environment();
-    std::string new_path = env["PATH"].to_string();
-    new_path.insert(0, fmt::format("{}:", exe_dir_));
-    env["PATH"] = new_path;
-  }
```
nishikawa-masaki, line 243: Self comment:
This function should have been removed in #10379 .
It had been used for testing the "process_monitor" node implemented with Linux commands like "top", "sort", etc.

After PR #10379 , Linux commands are not used anymore and modifyPath() is unnecessary.

system/autoware_system_monitor/test/src/process_monitor/test_process_monitor.cpp
```diff
@@ -269,6 +260,8 @@ DummyProcFile dummy_proc_files[] = {
   {"dummy_proc_unreadable.tar.bz2"},          // Test data with unreadable /proc file system.
   {"dummy_proc_not_mounted.tar.bz2"},         // Test data with /proc not mounted.
   {"dummy_proc_no_uptime.tar.bz2"},           // Test data with no /proc/uptime.
+  {"dummy_proc_negative_priority.tar.bz2"},   // Test data with negative priority and real-time
+                                              // priority.
```
nishikawa-masaki, line 264: Self comment:

Added data for testing the following cases:
* priority = -99
* priority = -100 (should be output as "rt")
* priority = 999
* priority = 1000 (should be output as "rt")

sasakisasaki:
@nishikawa-masaki Thank you for fixing an issue! I'm going to do tests on my local machine. Could you provide commands for verification? :pray:

nishikawa-masaki:
@sasakisasaki 
Thank you for your reviewing.
Please execute tests as the followings:
* To observe the reported problem more frequently, increase the number of processes reported by "process_monitor".
  * Edit the configuration file "install/autoware_system_monitor/share/autoware_system_monitor/config/process_monitor.param.yaml" and change the value of the parameter "num_of_procs" from 5 (default) to 40 (or more).
  * On Intel CPU platforms, the reported problem can't be observed with the default "num_of_procs".
* Open 2 terminals.
* On terminal 1, launch autoware_system_monitor.
> $ ros2 launch autoware_system_monitor system_monitor.launch.xml
* On terminal 2, monitor the "/diagnostics" topic and store the log in to a file.
> $  ros2 topic echo -f /diagnostics > /tmp/log
* Terminate ros2 on the two terminals with Ctrl-C.
* Investigate the log file.
  * Search for processes named as "migration/(X=number)". They have "rt" priority (-100).
  * Search for processes named as "idle_inject/(X=number)". They have negative priority (-51).
  * As autoware_system_monitor expects that helper applications "msr_reader", "hdd_reader", "traffic_reader" are running, other nodes ("cpu_monitor", "hdd_monitor", "ntp_monitor") report errors, which can be ignored in this test for "process_monitor".
* As processes on Linux change their states dynamically, you may need to monitor the "/diagnostics" topic for a few minutes.
  * The "/diagnostics" topic is published every second.

sasakisasaki:
Sorry for being late, now doing review :pray:

sasakisasaki:
@nishikawa-masaki Thank you so much for providing the detailed test process. I could see the following log. But I could not see a log related to "idle_inject/(X=number)". Could you kindly tell me what is the possible missing procedure? :pray:. I waited at least 3 minutes for obtaining the log. I'm not attaching the log file as it is large. But I can send you if necessary.

- "rt" priority in "migration/(X=number)"
```
- level: "\0"
  name: 'process_monitor: High-load Proc[24]'
  message: OK
  hardware_id: dpc2403002
  values:
  - key: COMMAND
    value: migration/16
  - key: '%CPU'
    value: '1.0'
  - key: '%MEM'
    value: '0.0'
  - key: PID
    value: '112'
  - key: USER
    value: root
  - key: PR
    value: rt
  - key: NI
    value: '0'
  - key: VIRT
    value: '0'
  - key: RES
    value: '0'
  - key: SHR
    value: '0'
  - key: S
    value: S
  - key: TIME+
    value: '0:14.10'
```

nishikawa-masaki:
@sasakisasaki 
Thank you very much for your reviewing and testing.
> But I could not see a log related to "idle_inject/(X=number)". 

It may be because your testing environment (PC, CPU, OS version) is different from mine.
I conducted the test on my desktop PC.

Please run the Linux "top" command on your environment and search for processes with negative priority.
If you can find the same (or similar) process name in the log file and its "PR" value is negative, it is OK.

sasakisasaki:
@nishikawa-masaki Thank you for your feedback, I found a process whose PR is negative as follows. If this is expected behavior, then I agree with merging this PR.

```
- level: "\0"
  name: 'process_monitor: High-load Proc[31]'
  message: OK
  hardware_id: dpc2403002
  values:
  - key: COMMAND
    value: irq/205-nvidia
  - key: '%CPU'
    value: '4.0'
  - key: '%MEM'
    value: '0.0'
  - key: PID
    value: '1236'
  - key: USER
    value: root
  - key: PR
    value: '-51'
  - key: NI
    value: '0'
  - key: VIRT
    value: '0'
  - key: RES
    value: '0'
  - key: SHR
    value: '0'
  - key: S
    value: S
  - key: TIME+
    value: '4:38.79'
```

system/autoware_system_monitor/src/process_monitor/process_monitor.cpp
```diff
@@ -589,7 +589,12 @@ void ProcessMonitor::fillTaskInfo(
   ProcessInfo info;
   info.processId = std::to_string(raw_p->stat_info.pid);
   info.userName = convertUidToUserName(raw_p->status_info.real_uid);
-  info.priority = std::to_string(raw_p->stat_info.priority);
+  // "top" command shows priority less than -99 and more than 999 as "rt", which means real-time.
+  if ((raw_p->stat_info.priority < -99) || (raw_p->stat_info.priority > 999)) {
```
sasakisasaki, line 596: Sorry, I have a question just before approval. Is there any reason why you chose the value -99 and 999?

sasakisasaki: (I want to know if such the values are popular in the context of system)

nishikawa-masaki: The values are taken from the source code of Linux command "top" so that the new implementation produces the same result with the old implementation with the "top" command.
[top.c L3122-L3128](https://github.com/mmalecki/procps/blob/fe4c4a7314f32907b9f558ad0d8b8d0ff1cc76be/top.c#L3122-L3128)

I added a few comment lines for future readers.

sasakisasaki: Thank you! I understand :+1:

sasakisasaki (approved):
I agree with merging this PR :+1:

PR #10836, opened 2025-06-18
feat(static_obstacle_avoidance): add infomation to PlanningFactor topic

Kotakku:
## Description
This PR adds the missing information to PlanningFactor in static_obstacle_avoidance_module.
I also added a unit test so that colcon test can run tests for the PlanningFactor interface.

## How was this PR tested?
- [x] colcon test
- [x] [PASS TIER IV INTERNAL SCENARIOS](https://evaluation.tier4.jp/evaluation/reports/51e50a84-e0c2-5886-8901-521189729415?project_id=prd_jt)
Several Evaluator test cases failed, but I determined that they were random failures because similar failures had occurred in past test results.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


planning/behavior_path_planner/autoware_behavior_path_static_obstacle_avoidance_module/src/scene.cpp
```diff
@@ -1274,11 +1274,25 @@ CandidateOutput StaticObstacleAvoidanceModule::planCandidate() const
     return output.lateral_shift > 0.0 ? PlanningFactor::SHIFT_LEFT : PlanningFactor::SHIFT_RIGHT;
   });
 
+  const std::string planning_factor_detail = std::invoke([&output]() {
+    return output.lateral_shift > 0.0 ? "candidate left shift" : "candidate right shift";
+  });
```
yhisaki:
```suggestion
  const std::string planning_factor_detail
    = output.lateral_shift > 0.0
      ? "candidate left shift"
      : "candidate right shift";
```

Kotakku: Thanks for the suggestions!
I'll proceed with the changes as you suggested. Also, I'll remove the word ```"candidate"``` from ```planning_factor_detail``` because it's not used in any other module.

Kotakku: I've made the corrections. 
Also, it's a minor fix, but I corrected the test namespace as it was wrong.

yhisaki (approved):

PR #10847, opened 2025-06-18
feat(autoware_pointcloud_preprocessor): add concat status

vividf:
## Description
As discussed in the internal channel, we would like to avoid having the fusion node subscribe to the /diagnostics topic.
To address this, this PR introduces a new topic, /concatenate_status, which provides the necessary pointcloud concatenation status for downstream modules.
In addition, this change helps reduce latency, as the fusion node no longer needs to filter through hundreds of unrelated diagnostic messages.

Note that this PR needs to be merged together with https://github.com/autowarefoundation/autoware_utils/pull/75

## Related links

**Parent Issue:**

- Link

## How was this PR tested?
Internal rosbag

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


vividf:
I will close this PR because we have decided to use autoware_sensing_msg to pass the status.

vividf:
I will close this PR because we have decided to use autoware_sensing_msg to pass the status.

PR #10864, opened 2025-06-20
feat(multi_object_tracker): improve shape unstable object tracking

linick-kang:
## Description

- Implement 3d giou for multi_object_tracker association
- Replace angle gate and iou gate with mixed-iou gate for different trackers
  - for pedestrian, use 1d-iou for computational efficiency
  - for non-pedestrian without trustworthy extension info, use 2d-giou for association
  - for non-pedestrian with trustworthy extension info, use 3d-giou for association
- Implement logic to check significant shape change (as a signal for unstable detection) between tracker and measurement for vehicle trackers, meanwhile implement corresponding conditioned update logic for such detection to mitigate the influence of noisy shape on velocity estimation:
  - partially update vehicle's front/rear anchor point based on well-matched edge.  
  - weakly update vehicle position tending to predicted states when edge matching is not confident
  - when a new stable shape is obtained, adapt the new shape and adjust position accordingly to avoid introduction of additional velocity

Requires the launch PR [https://github.com/autowarefoundation/autoware_launch/pull/1511](https://github.com/autowarefoundation/autoware_launch/pull/1511)

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

- Qualitative
  - similar tracking performance for normal objects
  - improved tracking performance for vehicle with unstable shape size, meanwhile reduce/relax association gating parameters
      - [Parking-vehs](https://console.data-search.tier4.jp/projects/x2_dev/environments/190ce3fb-2480-4710-a8db-b976814f90cb/rosbags/a5487adb-4074-40cf-8277-3942d81d2cb9)
      - https://github.com/user-attachments/assets/cbd02a38-0eb0-4446-8e5a-17e5bea178b7

      - [Cloud-peds](https://console.data-search.tier4.jp/projects/x2_dev/environments/190ce3fb-2480-4710-a8db-b976814f90cb/rosbags/3d23a7a7-074b-4ba4-b745-3254ba31a678)
      - https://github.com/user-attachments/assets/d7b947f6-b723-43b3-a577-9a8280fb2090

      - [Large-intersection](https://console.data-search.tier4.jp/projects/x2_dev/environments/190ce3fb-2480-4710-a8db-b976814f90cb/rosbags/0d92e8e9-f69f-4048-9fb8-56fce1c423e0)
      - https://github.com/user-attachments/assets/11f5c3f2-343a-4a1c-8a94-0c1ff9d0f1ff

      - [RT0-37024](https://console.data-search.tier4.jp/projects/x2_dev/environments/b4448f17-3a93-4e5b-b332-ace03e6b690f/rosbags/68406969-5498-4769-8191-93353d84c5de)
      - https://github.com/user-attachments/assets/b1126dfe-4174-4c79-a3d5-1d95f1b8c7bb

      - [RT0-35962](https://console.data-search.tier4.jp/projects/x2_dev/environments/190ce3fb-2480-4710-a8db-b976814f90cb/rosbags/53dacb2a-13ad-4a6d-ad38-25cac1cda561)
      - https://github.com/user-attachments/assets/c8384fa0-9c62-448f-8ec0-6c88d6efc93f

      - [RT0-37297](https://console.data-search.tier4.jp/projects/x2_dev/environments/190ce3fb-2480-4710-a8db-b976814f90cb/rosbags/b1b24e39-becf-4bd6-a238-61fab17d737e)
      - https://github.com/user-attachments/assets/b6f3332d-d5b9-48a2-b5bc-26c29c12d2b9

      - [RT0-37768](https://console.data-search.tier4.jp/projects/x2_dev/environments/459879d9-959c-4e3b-b5d3-b55c36d4a12e/rosbags/68618b98-e348-4367-8711-a8a1eb67c547)
      - https://github.com/user-attachments/assets/bc9b4456-1715-41f9-8ba1-63bc1fdb8e0d

      - [RT0-35335](https://console.data-search.tier4.jp/projects/Mu1UupL2/environments/6916ef2d-8ad6-4b66-b436-f8dcf6a7506e/rosbags/998b4353-bd0b-4545-b07a-e1016b49e91d)
      - https://github.com/user-attachments/assets/7fb38aac-a6cb-4f7d-838a-02fc417069a7

- Quantitative
  - Execution time comparison via the [performance unit test](https://github.com/autowarefoundation/autoware_universe/pull/10974)
    - Pedestrian/Unknown processing time: no significant change  
    - Vehicle processing time: +24%
  
| Car  | Pedestrian | Unknown |
|---|---|---|
| image | image | image |

 <details> <summary>Unittest result (Origin)</summary>
  
=== Performance (ms) vs Car Count (1-100) ===
Car            ,TotalTime   ,PredictTime ,AssociateTime,UpdateTime  ,PruneTime   ,SpawnTime   
1              ,0.012       ,0.000       ,0.011         ,0.001       ,0.000       ,0.000       
6              ,0.076       ,0.002       ,0.063         ,0.005       ,0.006       ,0.001       
11             ,0.105       ,0.002       ,0.086         ,0.006       ,0.009       ,0.002       
16             ,0.149       ,0.003       ,0.123         ,0.008       ,0.013       ,0.003       
21             ,0.201       ,0.004       ,0.162         ,0.011       ,0.020       ,0.004       
26             ,0.277       ,0.006       ,0.221         ,0.016       ,0.028       ,0.006       
31             ,0.299       ,0.006       ,0.250         ,0.017       ,0.023       ,0.005       
36             ,0.343       ,0.007       ,0.285         ,0.019       ,0.027       ,0.005       
41             ,0.395       ,0.007       ,0.322         ,0.022       ,0.039       ,0.006       
46             ,0.447       ,0.008       ,0.368         ,0.024       ,0.041       ,0.007       
51             ,0.478       ,0.008       ,0.394         ,0.025       ,0.044       ,0.008       
56             ,0.540       ,0.010       ,0.441         ,0.029       ,0.052       ,0.009       
61             ,0.607       ,0.011       ,0.485         ,0.034       ,0.067       ,0.010       
66             ,0.630       ,0.011       ,0.520         ,0.034       ,0.056       ,0.010       
71             ,0.656       ,0.012       ,0.546         ,0.036       ,0.051       ,0.011       
76             ,0.742       ,0.014       ,0.602         ,0.039       ,0.073       ,0.015       
81             ,0.780       ,0.014       ,0.640         ,0.040       ,0.072       ,0.014       
86             ,0.832       ,0.015       ,0.684         ,0.044       ,0.075       ,0.014       
91             ,0.861       ,0.016       ,0.713         ,0.046       ,0.073       ,0.013       
96             ,0.923       ,0.017       ,0.764         ,0.049       ,0.078       ,0.014       
[       OK ] MultiObjectTrackerTest.PerformanceVsCarCount (2793 ms)
[ RUN      ] MultiObjectTrackerTest.PerformanceVsPedestrianCount

=== Performance (ms) vs Pedestrian Count (1-100) ===
Pedestrian     ,TotalTime   ,PredictTime ,AssociateTime,UpdateTime  ,PruneTime   ,SpawnTime   
1              ,0.008       ,0.000       ,0.005         ,0.000       ,0.001       ,0.002       
6              ,0.041       ,0.002       ,0.021         ,0.002       ,0.007       ,0.009       
11             ,0.079       ,0.003       ,0.040         ,0.005       ,0.013       ,0.018       
16             ,0.106       ,0.004       ,0.054         ,0.006       ,0.018       ,0.024       
21             ,0.154       ,0.006       ,0.076         ,0.010       ,0.027       ,0.036       
26             ,0.173       ,0.007       ,0.084         ,0.010       ,0.032       ,0.041       
31             ,0.213       ,0.008       ,0.102         ,0.012       ,0.039       ,0.052       
36             ,0.250       ,0.010       ,0.121         ,0.015       ,0.046       ,0.057       
41             ,0.287       ,0.011       ,0.140         ,0.016       ,0.052       ,0.068       
46             ,0.315       ,0.012       ,0.156         ,0.018       ,0.057       ,0.072       
51             ,0.359       ,0.014       ,0.182         ,0.020       ,0.062       ,0.081       
56             ,0.399       ,0.016       ,0.201         ,0.023       ,0.069       ,0.091       
61             ,0.449       ,0.018       ,0.226         ,0.026       ,0.077       ,0.103       
66             ,0.478       ,0.018       ,0.251         ,0.027       ,0.075       ,0.107       
71             ,0.522       ,0.021       ,0.276         ,0.030       ,0.079       ,0.116       
76             ,0.578       ,0.024       ,0.309         ,0.034       ,0.089       ,0.122       
81             ,0.606       ,0.025       ,0.323         ,0.034       ,0.097       ,0.127       
86             ,0.638       ,0.026       ,0.342         ,0.037       ,0.098       ,0.135       
91             ,0.663       ,0.027       ,0.354         ,0.038       ,0.102       ,0.144       
96             ,0.761       ,0.033       ,0.408         ,0.046       ,0.119       ,0.156       
[       OK ] MultiObjectTrackerTest.PerformanceVsPedestrianCount (2020 ms)
[ RUN      ] MultiObjectTrackerTest.PerformanceVsUnknownObjectCount

=== Performance (ms) vs Unknown Count (1-100) ===
Unknown        ,TotalTime   ,PredictTime ,AssociateTime,UpdateTime  ,PruneTime   ,SpawnTime   
1              ,0.015       ,0.000       ,0.015         ,0.000       ,0.000       ,0.000       
6              ,0.100       ,0.000       ,0.086         ,0.001       ,0.011       ,0.002       
11             ,0.186       ,0.000       ,0.163         ,0.002       ,0.018       ,0.003       
16             ,0.260       ,0.001       ,0.228         ,0.003       ,0.025       ,0.005       
21             ,0.324       ,0.001       ,0.282         ,0.003       ,0.033       ,0.006       
26             ,0.399       ,0.001       ,0.343         ,0.004       ,0.043       ,0.007       
31             ,0.461       ,0.001       ,0.400         ,0.004       ,0.048       ,0.008       
36             ,0.534       ,0.001       ,0.460         ,0.005       ,0.060       ,0.009       
41             ,0.614       ,0.001       ,0.533         ,0.006       ,0.065       ,0.009       
46             ,0.671       ,0.001       ,0.590         ,0.007       ,0.064       ,0.010       
51             ,0.757       ,0.001       ,0.660         ,0.008       ,0.078       ,0.010       
56             ,0.822       ,0.001       ,0.721         ,0.008       ,0.080       ,0.012       
61             ,0.934       ,0.002       ,0.821         ,0.009       ,0.092       ,0.011       
66             ,0.976       ,0.002       ,0.850         ,0.009       ,0.103       ,0.013       
71             ,1.050       ,0.002       ,0.916         ,0.010       ,0.108       ,0.014       
76             ,1.125       ,0.002       ,0.978         ,0.012       ,0.120       ,0.014       
81             ,1.207       ,0.002       ,1.045         ,0.012       ,0.134       ,0.014       
86             ,1.347       ,0.002       ,1.172         ,0.013       ,0.143       ,0.017       
91             ,1.369       ,0.002       ,1.192         ,0.013       ,0.143       ,0.019       
96             ,1.452       ,0.003       ,1.279         ,0.015       ,0.137       ,0.019       
[       OK ] MultiObjectTrackerTest.PerformanceVsUnknownObjectCount (3911 ms)
[----------] 3 tests from MultiObjectTrackerTest (8724 ms total)

</details>
<details> <summary>Unittest result (Develop)</summary>
  
=== Performance (ms) vs Car Count (1-100) ===
Car            ,TotalTime   ,PredictTime ,AssociateTime,UpdateTime  ,PruneTime   ,SpawnTime   
1              ,0.030       ,0.001       ,0.027         ,0.002       ,0.001       ,0.001       
6              ,0.101       ,0.002       ,0.086         ,0.005       ,0.006       ,0.001       
11             ,0.140       ,0.002       ,0.120         ,0.007       ,0.010       ,0.002       
16             ,0.220       ,0.003       ,0.188         ,0.010       ,0.015       ,0.003       
21             ,0.265       ,0.004       ,0.224         ,0.011       ,0.021       ,0.004       
26             ,0.362       ,0.005       ,0.308         ,0.016       ,0.027       ,0.005       
31             ,0.378       ,0.005       ,0.329         ,0.017       ,0.023       ,0.004       
36             ,0.436       ,0.006       ,0.379         ,0.019       ,0.026       ,0.005       
41             ,0.495       ,0.007       ,0.422         ,0.023       ,0.038       ,0.006       
46             ,0.558       ,0.008       ,0.480         ,0.024       ,0.040       ,0.006       
51             ,0.620       ,0.009       ,0.533         ,0.026       ,0.045       ,0.007       
56             ,0.707       ,0.010       ,0.601         ,0.030       ,0.058       ,0.008       
61             ,0.740       ,0.010       ,0.631         ,0.031       ,0.059       ,0.009       
66             ,0.819       ,0.012       ,0.707         ,0.035       ,0.055       ,0.010       
71             ,0.861       ,0.013       ,0.748         ,0.038       ,0.052       ,0.011       
76             ,0.936       ,0.014       ,0.799         ,0.040       ,0.069       ,0.015       
81             ,0.997       ,0.014       ,0.853         ,0.042       ,0.074       ,0.014       
86             ,1.027       ,0.014       ,0.883         ,0.043       ,0.073       ,0.013       
91             ,1.079       ,0.015       ,0.932         ,0.047       ,0.074       ,0.012       
96             ,1.186       ,0.018       ,1.025         ,0.051       ,0.078       ,0.014       
[       OK ] MultiObjectTrackerTest.PerformanceVsCarCount (3452 ms)
[ RUN      ] MultiObjectTrackerTest.PerformanceVsPedestrianCount

=== Performance (ms) vs Pedestrian Count (1-100) ===
Pedestrian     ,TotalTime   ,PredictTime ,AssociateTime,UpdateTime  ,PruneTime   ,SpawnTime   
1              ,0.008       ,0.000       ,0.006         ,0.000       ,0.001       ,0.002       
6              ,0.042       ,0.002       ,0.021         ,0.002       ,0.007       ,0.010       
11             ,0.070       ,0.003       ,0.035         ,0.004       ,0.012       ,0.017       
16             ,0.106       ,0.004       ,0.053         ,0.006       ,0.019       ,0.024       
21             ,0.139       ,0.005       ,0.067         ,0.008       ,0.024       ,0.034       
26             ,0.177       ,0.007       ,0.085         ,0.011       ,0.033       ,0.040       
31             ,0.214       ,0.008       ,0.101         ,0.012       ,0.039       ,0.054       
36             ,0.255       ,0.010       ,0.125         ,0.015       ,0.046       ,0.059       
41             ,0.291       ,0.012       ,0.143         ,0.017       ,0.050       ,0.070       
46             ,0.317       ,0.012       ,0.156         ,0.018       ,0.057       ,0.074       
51             ,0.362       ,0.014       ,0.183         ,0.021       ,0.062       ,0.082       
56             ,0.404       ,0.016       ,0.203         ,0.023       ,0.069       ,0.093       
61             ,0.442       ,0.017       ,0.224         ,0.025       ,0.072       ,0.104       
66             ,0.475       ,0.019       ,0.249         ,0.027       ,0.075       ,0.106       
71             ,0.510       ,0.020       ,0.264         ,0.030       ,0.078       ,0.117       
76             ,0.548       ,0.022       ,0.288         ,0.032       ,0.085       ,0.122       
81             ,0.629       ,0.026       ,0.330         ,0.037       ,0.100       ,0.137       
86             ,0.640       ,0.026       ,0.339         ,0.037       ,0.100       ,0.139       
91             ,0.665       ,0.027       ,0.350         ,0.039       ,0.102       ,0.148       
96             ,0.730       ,0.030       ,0.384         ,0.043       ,0.115       ,0.158       
[       OK ] MultiObjectTrackerTest.PerformanceVsPedestrianCount (2000 ms)
[ RUN      ] MultiObjectTrackerTest.PerformanceVsUnknownObjectCount

=== Performance (ms) vs Unknown Count (1-100) ===
Unknown        ,TotalTime   ,PredictTime ,AssociateTime,UpdateTime  ,PruneTime   ,SpawnTime   
1              ,0.016       ,0.000       ,0.015         ,0.000       ,0.000       ,0.000       
6              ,0.101       ,0.000       ,0.087         ,0.001       ,0.011       ,0.002       
11             ,0.182       ,0.000       ,0.159         ,0.002       ,0.017       ,0.004       
16             ,0.271       ,0.001       ,0.237         ,0.003       ,0.026       ,0.005       
21             ,0.332       ,0.001       ,0.289         ,0.003       ,0.033       ,0.006       
26             ,0.395       ,0.001       ,0.340         ,0.004       ,0.043       ,0.007       
31             ,0.467       ,0.001       ,0.406         ,0.005       ,0.048       ,0.008       
36             ,0.542       ,0.001       ,0.466         ,0.006       ,0.061       ,0.009       
41             ,0.621       ,0.001       ,0.540         ,0.006       ,0.065       ,0.009       
46             ,0.771       ,0.002       ,0.676         ,0.009       ,0.075       ,0.011       
51             ,0.770       ,0.001       ,0.671         ,0.008       ,0.080       ,0.010       
56             ,0.832       ,0.001       ,0.729         ,0.009       ,0.082       ,0.012       
61             ,0.927       ,0.002       ,0.815         ,0.010       ,0.090       ,0.011       
66             ,1.004       ,0.002       ,0.884         ,0.011       ,0.095       ,0.013       
71             ,1.034       ,0.002       ,0.918         ,0.011       ,0.091       ,0.014       
76             ,1.211       ,0.002       ,1.071         ,0.013       ,0.111       ,0.014       
81             ,1.216       ,0.002       ,1.071         ,0.013       ,0.117       ,0.014       
86             ,1.282       ,0.002       ,1.124         ,0.013       ,0.127       ,0.016       
91             ,1.386       ,0.002       ,1.219         ,0.015       ,0.131       ,0.019       
96             ,1.429       ,0.002       ,1.262         ,0.015       ,0.132       ,0.019       
[       OK ] MultiObjectTrackerTest.PerformanceVsUnknownObjectCount (3959 ms)
[----------] 3 tests from MultiObjectTrackerTest (9411 ms total)

</details>

- Evaluator
  https://evaluation.tier4.jp/evaluation/reports/df91c38b-2ebf-5b49-b148-8a1e7d7293ef?project_id=prd_jt

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_multi_object_tracker/lib/tracker/model/vehicle_tracker.cpp
+++ b/perception/autoware_multi_object_tracker/lib/tracker/model/vehicle_tracker.cpp
@@ -361,8 +361,7 @@ UpdateStrategy VehicleTracker::determineUpdateStrategy(
   const EdgePositions meas_edges = calculateEdgeCenters(measurement);
 
   // 2. Calculate alignment distances between measurement and prediction edges
-  const EdgeAlignment alignment =
-    findAlignedEdges(meas_edges, prediction);
+  const EdgeAlignment alignment = findAlignedEdges(meas_edges, prediction);
 
   // 3. Check if any edge is well-aligned (within threshold ratio of vehicle length)
   const double predicted_length = prediction.shape.dimensions.x;
@@ -377,8 +376,8 @@ UpdateStrategy VehicleTracker::determineUpdateStrategy(
 
   // 5. Determine aligned edge and calculate anchor point
   strategy.type = (alignment.aligned_pred_edge == Edge::FRONT)
-  ? UpdateStrategyType::FRONT_WHEEL_UPDATE
-  : UpdateStrategyType::REAR_WHEEL_UPDATE;
+                    ? UpdateStrategyType::FRONT_WHEEL_UPDATE
+                    : UpdateStrategyType::REAR_WHEEL_UPDATE;
   strategy.anchor_point = calculateAnchorPoint(alignment, measurement);
 
   return strategy;
@@ -424,13 +423,17 @@ VehicleTracker::EdgeAlignment VehicleTracker::findAlignedEdges(
   const double pred_rear_axis = pred_center_axis - predicted_half_length;
 
   // Define all four edge alignment candidates
-  struct Candidate { double distance; Edge pred_edge; Edge meas_edge; };
-  const std::array<Candidate, 4> candidates = {{
-    {std::abs(meas_front_axis - pred_front_axis), Edge::FRONT, Edge::FRONT},
-    {std::abs(meas_rear_axis - pred_front_axis), Edge::FRONT, Edge::REAR},
-    {std::abs(meas_front_axis - pred_rear_axis), Edge::REAR, Edge::FRONT},
-    {std::abs(meas_rear_axis - pred_rear_axis), Edge::REAR, Edge::REAR}
-  }};
+  struct Candidate
+  {
+    double distance;
+    Edge pred_edge;
+    Edge meas_edge;
+  };
+  const std::array<Candidate, 4> candidates = {
+    {{std::abs(meas_front_axis - pred_front_axis), Edge::FRONT, Edge::FRONT},
+     {std::abs(meas_rear_axis - pred_front_axis), Edge::FRONT, Edge::REAR},
+     {std::abs(meas_front_axis - pred_rear_axis), Edge::REAR, Edge::FRONT},
+     {std::abs(meas_rear_axis - pred_rear_axis), Edge::REAR, Edge::REAR}}};
 
   // Find the best aligned edge pair
   const auto best = std::min_element(


perception/autoware_multi_object_tracker/lib/object_model/shapes.cpp
```diff
@@ -287,6 +304,47 @@ void calcAnchorPointOffset(
...
+  if (height_overlap <= 0.0) return INVALID_SCORE;
+
+  const double total_height = std::max(z_max_src, z_max_tgt) - std::min(z_min_src, z_min_tgt);
+
+  const double intersection_volume = intersection_area * height_overlap;
+  const double union_volume = union_area * total_height;
+  const double convex_volume = convex_area * total_height;
+
+  const double iou = std::min(1.0, intersection_volume / union_volume);
+  return iou - (convex_volume - union_volume) / convex_volume;
```
ogrex: This should be equivalent and simpler?(not verified)

```
const double iou = intersection_area * height_overlap / (union_area * total_height);
return std::min(0, iou- 1.0  ) + union_area / convex_area;
```

perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/object_model/types.hpp
```diff
@@ -44,6 +44,7 @@ namespace types
 // constants
 constexpr size_t max_channel_size = 12;
 constexpr float default_existence_probability = 0.75;
+constexpr double diameter_coefficient = 1.1283791670955126;  // = sqrt(4/pi);
```
ogrex: since constexpr value is evaluated at compile time, this runtime cost should still be zero
`std::sqrt(4.0 / M_PI);
`

perception/autoware_multi_object_tracker/lib/association/association.cpp
```diff
@@ -245,55 +233,72 @@ Eigen::MatrixXd DataAssociation::calcScoreMatrix(
...
-  }
 
-  const double ratio_sq = dist_sq / max_dist_sq;
-  const double score = 1.0 - std::sqrt(ratio_sq);
-  if (score < score_threshold_) return 0.0;
+  // dist gate
+  if (dist_sq > max_dist_sq) return INVALID_SCORE;
+
+  constexpr uint8_t VehicleLabelMin = 1;  // CAR
+  constexpr uint8_t VehicleLabelMax = 4;  // TRAILER
```
ogrex: there should be a constant Label::PEDESTRIAN can be used

perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/tracker/model/tracker_base.hpp
```diff
@@ -51,6 +53,19 @@ class Tracker
...
+  static constexpr double SHAPE_VARIATION_THRESHOLD = 0.2;
+  static constexpr size_t WEAK_UPDATE_MAX_COUNT = 10;
+  static constexpr size_t STABLE_STREAK_THRESHOLD = 4;
+  static constexpr size_t UNSTABLE_STREAK_THRESHOLD = 2;
+
+  size_t weak_update_count_{0};
+  size_t shape_stable_streak_{0};
+  size_t shape_unstable_streak_{0};
+  bool ema_shape_initialized_{false};
+  Eigen::Vector3d ema_shape_;
```
technolojin: EMA stands for Exponential Moving Average

please add a comment that the word will be used by the abbreviation.

perception/autoware_multi_object_tracker/lib/association/association.cpp
```diff
@@ -231,11 +212,18 @@ Eigen::MatrixXd DataAssociation::calcScoreMatrix(
...
       // The actual distance check was already done in the R-tree query
       double score = calculateScore(
         tracked_object, tracker_label, measurement_object, measurement_label,
-        tracker_inverse_covariances[tracker_idx]);
+        tracker_inverse_covariances[tracker_idx], significant_shape_change);
       score_matrix(tracker_idx, measurement_idx) = score;
+
+      if (significant_shape_change) {
+        significant_shape_change_set_.insert(
+          (static_cast<uint64_t>(tracker_idx) << 32) | measurement_idx);
```
technolojin: Add a comment about the usage of the `significant_shape_change_set_`

_hash the indeces_

perception/autoware_multi_object_tracker/lib/association/association.cpp
technolojin: Why CamelCase?

perception/autoware_multi_object_tracker/lib/association/association.cpp
```diff
@@ -245,55 +233,72 @@ Eigen::MatrixXd DataAssociation::calcScoreMatrix(
...
+  // dist gate
+  if (dist_sq > max_dist_sq) return INVALID_SCORE;
+
+  constexpr uint8_t VehicleLabelMin = 1;  // CAR
+  constexpr uint8_t VehicleLabelMax = 4;  // TRAILER
+
+  // gates for non-vehicle objects
+  const double area_meas = measurement_object.area;
+  const bool is_vehicle_tracker =
+    tracker_label >= VehicleLabelMin && tracker_label <= VehicleLabelMax;
```
technolojin: make it more explicit

```suggestion
  const bool is_vehicle_tracker = tracker_label == Label::CAR || label == Label::BUS || label == Label::TRUCK || label == Label::TRAILER;
```

perception/autoware_multi_object_tracker/lib/tracker/model/tracker_base.cpp
```diff
@@ -163,15 +164,100 @@ bool Tracker::updateWithMeasurement(
...
+        object_.shape.dimensions.y = ema_shape_[1];
+        object_.shape.dimensions.z = ema_shape_[2];
+      }
+
+      // measure to update Kalman filter internal states
+      measure(object, measurement_time, channel_info);
+
+      resetShapeUpdateCount();
+    }
+  }
```
technolojin, line 212: 1. Please make it separated function(s) for readability. I suggest to make a class for EMA.
2. in case of polygon shape, an equivalent diameter is used. is it only for EMA?
3. the weak update is done whatever the object shape type is. Do you think it is still valid for polygon (mostly Unknown) objects?

perception/autoware_multi_object_tracker/lib/tracker/model/tracker_base.cpp
```diff
@@ -163,15 +164,100 @@ bool Tracker::updateWithMeasurement(
...
+    pred.pose.position.x = pred.pose.position.x * (1 - w_pose) + object.pose.position.x * w_pose;
+    pred.pose.position.y = pred.pose.position.y * (1 - w_pose) + object.pose.position.y * w_pose;
+
+    // Blend yaw orientation
+    double yaw_pred = tf2::getYaw(pred.pose.orientation);
+    double yaw_meas = tf2::getYaw(object.pose.orientation);
+    double yaw_fused = yaw_pred * (1 - w_pose) + yaw_meas * w_pose;
+    tf2::Quaternion q;
+    q.setRPY(0, 0, yaw_fused);
+    pred.pose.orientation = tf2::toMsg(q);
```
technolojin: Yaw is not always trusted or available.
Please see the other trackers.

technolojin:
By a similar PR https://github.com/autowarefoundation/autoware_launch/pull/1564, there are conflicts. 
You can overwrite the change introduced, but also need to be co-exist.

perception/autoware_multi_object_tracker/lib/object_model/shapes.cpp
linick-kang: I think they are not equivalent when two boxes do not have same heights. and giou value ranges from -1 to 1.

perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/object_model/types.hpp
linick-kang: `std::sqrt(4.0 / M_PI)` isn't a constant expression under c++17, so I hard-coded the value here.

perception/autoware_multi_object_tracker/lib/association/association.cpp
linick-kang: fixed. aee5b42aa03aa85007b9da99723def6f3adefc85

perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/tracker/model/tracker_base.hpp
linick-kang: fixed. aee5b42aa03aa85007b9da99723def6f3adefc85

perception/autoware_multi_object_tracker/lib/association/association.cpp
linick-kang: fixed. aee5b42aa03aa85007b9da99723def6f3adefc85

perception/autoware_multi_object_tracker/lib/association/association.cpp
linick-kang: fixed. aee5b42aa03aa85007b9da99723def6f3adefc85

perception/autoware_multi_object_tracker/lib/association/association.cpp
linick-kang: fixed. aee5b42aa03aa85007b9da99723def6f3adefc85

perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/object_model/types.hpp
ogrex: I see

perception/autoware_multi_object_tracker/lib/tracker/model/tracker_base.cpp
```diff
@@ -163,15 +164,100 @@ bool Tracker::updateWithMeasurement(
...
+
+    // Update blended pose
+    object_.pose = pred.pose;
```
technolojin: May need to care about using the `object_` member.

The object position is in double-maintained by this member and the Kalman filter state vector.

When required, the object position is obtained 
https://github.com/autowarefoundation/autoware_universe/blob/0b66f575f92f1a136d4a331b6b9fafc18a32831e/perception/autoware_multi_object_tracker/lib/tracker/model/vehicle_tracker.cpp#L312
and the blended pose is forgotten

perception/autoware_multi_object_tracker/lib/tracker/model/tracker_base.cpp
linick-kang: Thank you for the valuable comment and suggestions. 
1. fixed eb08a22538ea559a257498b50ea07aa27068f64d
2. it is mainly for fast association of polygon shape unknown objects
3. I think weak update concept is applicable if area (shape) changes significantly, but EMA only for bounding boxes, I change implementation accordingly in eb08a22538ea559a257498b50ea07aa27068f64d

perception/autoware_multi_object_tracker/lib/tracker/model/tracker_base.cpp
linick-kang: Thank you for pointing out this issue.
I correct it by making yaw blend only when it is available. eb08a22538ea559a257498b50ea07aa27068f64d

perception/autoware_multi_object_tracker/lib/tracker/model/tracker_base.cpp
linick-kang: Thank you for this kind notice.
In current implementation, I think `object` member will only be accessed once at the same time?

technolojin: Let be explain its background.

1. update `object_` memeber is to update labels, headers, shape types... not tracked by the motion model
2. if you want to replace the motion model position, https://github.com/autowarefoundation/autoware_universe/blob/98c1055c164f828f487944a36668b8d15c7e0ead/perception/autoware_multi_object_tracker/lib/tracker/motion_model/bicycle_motion_model.cpp#L292-L307 is for you
3. Think twice if you update the position before update the tracker. measureWithPose is the way to update the Kalman filter, and your implementation will do the pose update twice.

perception/autoware_multi_object_tracker/lib/object_model/shapes.cpp
```diff
@@ -343,6 +348,47 @@ void calcAnchorPointOffset(
...
+
+  const auto [z_min_src, z_max_src] = getObjectZRange(source_object);
+  const auto [z_min_tgt, z_max_tgt] = getObjectZRange(target_object);
+
+  const double height_overlap =
+    std::max(0.0, std::min(z_max_src, z_max_tgt) - std::max(z_min_src, z_min_tgt));
+
+  if (height_overlap <= 0.0) return INVALID_SCORE;
+
+  const double total_height = std::max(z_max_src, z_max_tgt) - std::min(z_min_src, z_min_tgt);
```
technolojin, line 280: The GIoU also checks object heights.

However, some detectors are not accurate for the height (cluster type of detectors)
example: 
a. `use_low_height_cropbox` crops the obstacle pointcloud. 
b. `margin_max_z` and `detection_range_z_max` ground segmentation limits the cluster pointcloud height
The GIoU will become low, and the association will going to fail.

**suggestion**
1. ensure all the detectors that height detection reliable
2. lower the impact of height mismatch. 
    a. separate 2D GIoU and Z GIoU?
    b. z difference only for gating?
3. let user choose which channel enables 3d GIoU

perception/autoware_multi_object_tracker/lib/tracker/model/tracker_base.cpp
```diff
@@ -192,6 +242,35 @@ bool Tracker::updateWithoutMeasurement(const rclcpp::Time & timestamp)
...
 }
 
+bool Tracker::createPseudoMeasurement(
+  const types::DynamicObject & meas, types::DynamicObject & pred)
+{
+  // Apply linear fall‑off weight on dist square
+  const double dx = meas.pose.position.x - pred.pose.position.x;
+  const double dy = meas.pose.position.y - pred.pose.position.y;
+  const double dist2 = dx * dx + dy * dy;
+  constexpr double d_max_square_inv = 1 / 2.0;
```
technolojin: w_pose will saturate when dist2 becomes 2.0 (distance of 1.414 m) 
```suggestion
  constexpr double d_max_square_inv = 1 / 2.0; # saturate when distance overs 1.414m
```

perception/autoware_multi_object_tracker/lib/tracker/model/tracker_base.cpp
```diff
@@ -192,6 +242,35 @@ bool Tracker::updateWithoutMeasurement(const rclcpp::Time & timestamp)
...
+  constexpr double min_w = 0.05;
+  const double w_pose = std::clamp(1.0 - dist2 * d_max_square_inv, min_w, 1.0);
+
+  // Blend position
+  pred.pose.position.x = pred.pose.position.x * (1 - w_pose) + meas.pose.position.x * w_pose;
+  pred.pose.position.y = pred.pose.position.y * (1 - w_pose) + meas.pose.position.y * w_pose;
+
+  // Blend orientation by yaw availability
+  bool is_yaw_available =
+    meas.kinematics.orientation_availability != types::OrientationAvailability::UNAVAILABLE;
```
technolojin: also consider `OrientationAvailability::SIGN_UNKNOWN`


**What is SIGN_UNKNOWN?**
Some detectors (lidar detectors) are poor to distinguish vehicle front or rear, but knows its orientations

**What will happen?**
tracker : 0 deg
detection : 180 deg
blended : 90 deg when weight is 0.5

_sudden rotation of tracker_

**How to fix?**
crop the yaw difference in range of -90 deg to 90 deg

perception/autoware_multi_object_tracker/lib/tracker/model/tracker_base.cpp
```diff
@@ -163,11 +164,60 @@ bool Tracker::updateWithMeasurement(
...
+        if (ema_shape_.getBBoxValue(ema_shape)) {
+          object_.shape.dimensions.x = ema_shape[0];
+          object_.shape.dimensions.y = ema_shape[1];
+          object_.shape.dimensions.z = ema_shape[2];
+          object_.shape.type = autoware_perception_msgs::msg::Shape::BOUNDING_BOX;
+        }
+      }
+
+      // measure to update Kalman filter internal states
+      measure(object, measurement_time, channel_info);
```
technolojin: whatever `object_` is updated, the kalman filter is running with the input `object`.

* If your intention is to modify the input
make an object `types::DynamicObject in_object` and modify `in_object` by createPseudoMeasurement and EMA

* If your intention is to modify the tracker
implement createPseudoMeasurement and EMA in `measureWithPose` and `measureWithShape  of all the trackers

perception/autoware_multi_object_tracker/lib/tracker/model/tracker_base.cpp
```diff
@@ -163,11 +164,60 @@ bool Tracker::updateWithMeasurement(
...
 
-  // Update object status
-  getTrackedObject(measurement_time, object_);
+    // Weak update based on prediction
+    ++weak_update_count_;
+    types::DynamicObject pred;
+    getTrackedObject(measurement_time, pred);
+
+    // Create pseudo measurement
+    createPseudoMeasurement(object, pred);
```
technolojin: By the motion model improvement https://github.com/autowarefoundation/autoware_universe/pull/11230, partial update https://github.com/technolojin/autoware.universe/blob/9a85c170e9f0b0133a1617e0cf28cda17c53b957/perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/tracker/motion_model/bicycle_motion_model.hpp#L111-L115 is implemented.

I would like to suggest the EMA update to these `updateStatePoseRear` or `updateStatePoseFront` update

perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/tracker/model/vehicle_tracker.hpp
```diff
@@ -27,6 +27,15 @@
...
 
+// Vehicle update strategy for partial updates
+enum class UpdateStrategy { FRONT_WHEEL, REAR_WHEEL, BODY };
+
+struct WheelInfo
+{
+  UpdateStrategy strategy;
+  geometry_msgs::msg::Point wheel_position;  // Only used for FRONT_WHEEL and REAR_WHEEL
+};
+
```
technolojin: Terminology `WheelInfo` is ambiguous.
By it's context, I suggest it as `UpdateStrategy`, and set the wheel position as `anchor point`.

perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/tracker/model/vehicle_tracker.hpp
```diff
@@ -51,9 +60,19 @@ class VehicleTracker : public Tracker
...
+    const autoware_perception_msgs::msg::Shape & tracker_shape,
+    const rclcpp::Time & measurement_time, const types::InputChannel & channel_info) override;
+
   bool getTrackedObject(
     const rclcpp::Time & time, types::DynamicObject & object,
     const bool to_publish = false) const override;
+
+  void setObjectShape(const autoware_perception_msgs::msg::Shape & shape) override;
+
+  WheelInfo estimateUpdateWheel(
```
technolojin: Terminology `estimateUpdateWheel` is leaned to its algorithm. 
I prefer to rename it `determineUpdateStrategy` or similar.

perception/autoware_multi_object_tracker/lib/association/association.cpp
```diff
@@ -248,11 +229,18 @@ Eigen::MatrixXd DataAssociation::calcScoreMatrix(
...
+      bool significant_shape_change = false;
+      double score = calculateScore(
+        tracked_object, tracker_label, measurement_object, measurement_label,
+        tracker_inverse_covariances[tracker_idx], significant_shape_change);
       score_matrix(tracker_idx, measurement_idx) = score;
+
+      if (significant_shape_change) {
+        // hash the tracker and measurement index pair
+        significant_shape_change_set_.insert(
+          (static_cast<uint64_t>(tracker_idx) << 32) | measurement_idx);
```
technolojin: do not implement hash function distributed.

1. make a class manages the tracker-measure hash for `significant_shape_change_set_`
2. make a method to update the hash, input of tracker_idx and measurement_idx, instead of this.
3. make a checker, instead of `hasSignificantShapeChange`

perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/tracker/model/vehicle_tracker.hpp
```diff
@@ -51,9 +60,19 @@ class VehicleTracker : public Tracker
   bool measureWithPose(
     const types::DynamicObject & object, const types::InputChannel & channel_info);
 
+  bool conditionedUpdate(
```
technolojin, line 67: may `conditionedUpdate` be eliminated from this PR. at this implementation, conditionedUpdate will never triggerd.

perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/tracker/model/tracker_base.hpp
```diff
@@ -64,6 +67,16 @@ class Tracker
...
   std::vector<autoware_perception_msgs::msg::ObjectClassification> classification_;
 
+  // conditioned update configs
+  // EMA/ema below are abbreviation for exponential moving average
+  static constexpr double EMA_ALPHA_WEAK = 0.05;
+  static constexpr double EMA_ALPHA_STRONG = 0.2;
+  static constexpr double SHAPE_VARIATION_THRESHOLD = 0.1;
+  static constexpr size_t STABLE_STREAK_THRESHOLD = 4;
+
+  ExponentialMovingAverageShape ema_shape_{
```
technolojin: rename ema_shape_ more meaningful.
From its context, it is a shape tracker/filter monitoring object extension.

perception/autoware_multi_object_tracker/lib/tracker/model/tracker_base.cpp
```diff
@@ -163,12 +170,45 @@ bool Tracker::updateWithMeasurement(
...
+      const auto tracker_shape = object_.shape;
+
+      // Get predicted object
+      types::DynamicObject predicted_object;
+      getTrackedObject(measurement_time, predicted_object);
+
+      // Perform conditioned update
+      conditionedUpdate(object, predicted_object, tracker_shape, measurement_time, channel_info);
+    }
+  }
```
technolojin: Clarify three streams.

1. normal update: update position and extension(shape)
2. extension update: if a big size change is consist, update the tracker filter
3. partial update: the detection is partially visible. go to further check
    a. front update: the detection matches to the front of the tracker
    b. rear update: the detection matches to the rear of the tracker
    c. weak update: the detection does not matches well. minimum update

perception/autoware_multi_object_tracker/lib/tracker/model/vehicle_tracker.cpp
```diff
@@ -301,4 +301,153 @@ bool VehicleTracker::getTrackedObject(
...
+    } else {
+      // Rear wheel: move inward from rear edge
+      wheel_info.wheel_position.x = meas_rear_x + edge_to_wheel_offset * meas_cos_yaw;
+      wheel_info.wheel_position.y = meas_rear_y + edge_to_wheel_offset * meas_sin_yaw;
+    }
+    wheel_info.wheel_position.z = measurement.pose.position.z;
+  }
+
+  return wheel_info;
+}
```
technolojin: can you try refactoring for readability?

perception/autoware_multi_object_tracker/lib/tracker/motion_model/bicycle_motion_model.cpp
```diff
@@ -332,6 +332,53 @@ bool BicycleMotionModel::updateStatePoseFront(
...
+  const double current_yaw = getYawState();
+  const double cos_yaw = std::cos(current_yaw);
+  const double sin_yaw = std::sin(current_yaw);
+
+  // Calculate new wheel positions directly from current rear wheel position
+  // Rear wheel moves by the difference in lr distances
+  const double current_length = getLength();
+  const double lr_current = current_length * motion_params_.lr_ratio;
+  const double lr_new = new_length * motion_params_.lr_ratio;
+  const double lr_delta = lr_new - lr_current;
```
technolojin:
```suggestion
  const double current_length = getLength();
  const double lr_delta = (new_length - current_length) * motion_params_.lr_ratio;
```

perception/autoware_multi_object_tracker/lib/tracker/motion_model/bicycle_motion_model.cpp
```diff
@@ -332,6 +332,53 @@ bool BicycleMotionModel::updateStatePoseFront(
...
+
+  const double new_x1 = X_t(IDX::X1) - lr_delta * cos_yaw;
+  const double new_y1 = X_t(IDX::Y1) - lr_delta * sin_yaw;
+
+  // Front wheel position based on new wheelbase from new rear wheel position
+  const double new_wheelbase = new_length * (motion_params_.lf_ratio + motion_params_.lr_ratio);
+  const double new_x2 = new_x1 + new_wheelbase * cos_yaw;
+  const double new_y2 = new_y1 + new_wheelbase * sin_yaw;
```
technolojin: updateStateLength extends/contracts the length, while the center position kept.
Do you think it is always true? don't you think about anchor point when object size is changing?

For example, when the given tracked object is measured partially, matched with its rear side, the size update should be done to the front part is moving, not the rear position. With this algorithm, the given object will move the rear side.

perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/tracker/model/tracker_base.hpp
```diff
@@ -88,11 +101,11 @@ class Tracker
   // object update
   bool updateWithMeasurement(
     const types::DynamicObject & object, const rclcpp::Time & measurement_time,
-    const types::InputChannel & channel_info);
+    const types::InputChannel & channel_info, bool significant_shape_change = false);
```
technolojin:
```suggestion
    const types::InputChannel & channel_info, bool had_significant_shape_change = false);
```

technolojin (changes requested):
Can you update evidence of the improvement?
Please update the videos with updated version.

perception/autoware_multi_object_tracker/lib/tracker/model/tracker_base.cpp
linick-kang: Done.

perception/autoware_multi_object_tracker/lib/object_model/shapes.cpp
linick-kang: Done.

perception/autoware_multi_object_tracker/lib/tracker/model/tracker_base.cpp
linick-kang: Done.

perception/autoware_multi_object_tracker/lib/tracker/model/tracker_base.cpp
linick-kang: Done.

perception/autoware_multi_object_tracker/lib/association/association.cpp
linick-kang: Done. 7e86dd76c70c7ee511b47ebf29ce486f03c2558f

perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/tracker/model/vehicle_tracker.hpp
linick-kang: Done. 7e86dd76c70c7ee511b47ebf29ce486f03c2558f

perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/tracker/model/vehicle_tracker.hpp
linick-kang: Done. 7e86dd76c70c7ee511b47ebf29ce486f03c2558f

perception/autoware_multi_object_tracker/lib/association/association.cpp
linick-kang: Done. 09bd03d4fa56bf8ac3cd8b8f3428d2d2b456a77c

perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/tracker/model/vehicle_tracker.hpp
linick-kang: I commented out not tested conditionedUpdate for non-vehicle in tracker_base.cpp. 7e86dd76c70c7ee511b47ebf29ce486f03c2558f

perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/tracker/model/tracker_base.hpp
linick-kang: Thank you for the suggestion. Renamed it to unstable_shape_filter now. 7e86dd76c70c7ee511b47ebf29ce486f03c2558f

perception/autoware_multi_object_tracker/lib/tracker/model/tracker_base.cpp
linick-kang: Thank you for the advice. Add comment in 09bd03d4fa56bf8ac3cd8b8f3428d2d2b456a77c

perception/autoware_multi_object_tracker/lib/tracker/model/vehicle_tracker.cpp
linick-kang: Sorry for the long unstructured code. I tried to break the function into parts and added comments for readibility in 7e86dd76c70c7ee511b47ebf29ce486f03c2558f.

perception/autoware_multi_object_tracker/lib/tracker/motion_model/bicycle_motion_model.cpp
linick-kang: Done as suggested.

perception/autoware_multi_object_tracker/lib/tracker/motion_model/bicycle_motion_model.cpp
linick-kang: Nice point! I previously considered about it too. For simplicity (and avoiding choosing wrong anchor) I stick to center position.
But you're right, it could cause unwanted velocity in following update. So I decided to choose the anchor point based on last conditioned update strategy as in  df3d1c769b196f84ff4c4c0ef9698c12a8ed6441.
(I need to test more data to evaluate this implementation, please kindly give you opinion on this added feature or we may also separate it for another PR)

perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/tracker/model/tracker_base.hpp
linick-kang: Thank you for the suggestion.
I changed it to `has_***` to align with the naming convention.

perception/autoware_multi_object_tracker/lib/tracker/model/tracker_base.cpp
```diff
@@ -192,6 +240,71 @@ bool Tracker::updateWithoutMeasurement(const rclcpp::Time & timestamp)
...
+      yaw_diff = std::fmod(yaw_diff + M_PI, 2 * M_PI) - M_PI;
+      if (yaw_diff > M_PI_2) {
+        yaw_diff -= M_PI;
+      } else if (yaw_diff < -M_PI_2) {
+        yaw_diff += M_PI;
+      }
+      yaw_meas = yaw_pred + yaw_diff;
+    }
+
+    double yaw_fused = yaw_pred * (1 - w_pose) + yaw_meas * w_pose;
```
technolojin: Even for the orientation is available, normalization of yaw difference to [-pi, pi] necessary

```suggestion
    double yaw_diff = yaw_meas - yaw_pred;
    // Normalize yaw_diff to [-π, π] using fmod
    yaw_diff = std::fmod(yaw_diff + M_PI, 2 * M_PI) - M_PI;

    // Handle SIGN_UNKNOWN: limit yaw difference to [-90°, 90°]
    if (meas.kinematics.orientation_availability == types::OrientationAvailability::SIGN_UNKNOWN) {
      if (yaw_diff > M_PI_2) {
        yaw_diff -= M_PI;
      } else if (yaw_diff < -M_PI_2) {
        yaw_diff += M_PI;
      }
    }

    double yaw_fused = yaw_pred + yaw_diff * w_pose;
```

perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/association/index_pair_checker.hpp
```diff
@@ -0,0 +1,73 @@
...
+// limitations under the License.
+
+#ifndef AUTOWARE__MULTI_OBJECT_TRACKER__TRACKER__UTIL__INDEX_PAIR_CHECKER_HPP_
+#define AUTOWARE__MULTI_OBJECT_TRACKER__TRACKER__UTIL__INDEX_PAIR_CHECKER_HPP_
+
+#include <cstddef>
+#include <cstdint>
+#include <unordered_set>
+
+namespace autoware::multi_object_tracker
```
technolojin, line 22: this code is for association. move this file under association

perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/association/adaptive_threshold_cache.hpp
```diff
@@ -36,8 +36,8 @@ class AdaptiveThresholdCache
   static constexpr double MAX_DISTANCE = 150.0;  // meter
```
technolojin, line 36: this code is for association. move this file under association

perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/tracker/shape_model/unstable_shape_filter.hpp
```diff
@@ -0,0 +1,61 @@
...
+#ifndef AUTOWARE__MULTI_OBJECT_TRACKER__TRACKER__UTIL__UNSTABLE_SHAPE_FILTER_HPP_
+#define AUTOWARE__MULTI_OBJECT_TRACKER__TRACKER__UTIL__UNSTABLE_SHAPE_FILTER_HPP_
+
+#include "autoware/multi_object_tracker/object_model/types.hpp"
+
+#include <Eigen/Core>
+
+#include <autoware_perception_msgs/msg/shape.hpp>
+
+namespace autoware::multi_object_tracker
```
technolojin, line 24: this code is for tracker, specifically for shape (extension) tracking. we may can set the folder
`tracker/shape_model`

perception/autoware_multi_object_tracker/lib/tracker/model/vehicle_tracker.cpp
```diff
@@ -301,4 +307,191 @@ bool VehicleTracker::getTrackedObject(
...
+  const EdgeAlignmentDistances alignment_distances =
+    calculateAlignmentDistances(meas_edges, prediction);
+
+  // 3. Check if any edge is well-aligned (within threshold ratio of vehicle length)
+  const double predicted_length = prediction.shape.dimensions.x;
+  const double min_alignment_distance = std::min(
+    alignment_distances.front_alignment_distance, alignment_distances.rear_alignment_distance);
+  constexpr double alignment_ratio_threshold = 0.09;  // 9% of length tolerance
+  const bool is_edge_aligned =
+    (min_alignment_distance / predicted_length) < alignment_ratio_threshold;
```
technolojin: this parameter is very important 

let's move it to hpp
```suggestion
  const bool is_edge_aligned =
    (min_alignment_distance / predicted_length) < ALIGNMENT_RATIO_THRESHOLD;
```

perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/tracker/model/vehicle_tracker.hpp
```diff
@@ -51,9 +64,41 @@ class VehicleTracker : public Tracker
...
+    const rclcpp::Time & measurement_time, const types::InputChannel & channel_info) override;
+
   bool getTrackedObject(
     const rclcpp::Time & time, types::DynamicObject & object,
     const bool to_publish = false) const override;
+
+  void setObjectShape(const autoware_perception_msgs::msg::Shape & shape) override;
+
+  UpdateStrategy determineUpdateStrategy(
+    const types::DynamicObject & measurement, const types::DynamicObject & prediction) const;
```
technolojin, line 80:
```suggestion
  const double ALIGNMENT_RATIO_THRESHOLD = 0.09;
  UpdateStrategy determineUpdateStrategy(
    const types::DynamicObject & measurement, const types::DynamicObject & prediction) const;
```

perception/autoware_multi_object_tracker/lib/association/association.cpp
```diff
@@ -262,62 +248,80 @@ Eigen::MatrixXd DataAssociation::calcScoreMatrix(
...
+  } else if (use_3d_iou) {
+    iou_score = shapes::get3dGeneralizedIoU(measurement_object, tracked_object);
+  } else {
+    iou_score = shapes::get2dGeneralizedIoU(measurement_object, tracked_object);
+  }
+  if (iou_score < min_iou) return INVALID_SCORE;
+
+  // check if shape changes too much for vehicle labels
+  constexpr double CheckGiouThreshold = 0.7;
+  constexpr double AreaRatioThreshold = 1.3;
```
technolojin: these parameters are also important. bring it to hpp

perception/autoware_multi_object_tracker/lib/tracker/model/tracker_base.cpp
linick-kang: Thank you for the kind suggestion. I correct it as suggested.

perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/association/index_pair_checker.hpp
linick-kang: Done. 8d80b1c6389811fec8bad2e719a47b4964916040

perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/association/adaptive_threshold_cache.hpp
linick-kang: Done. 8d80b1c6389811fec8bad2e719a47b4964916040

perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/tracker/shape_model/unstable_shape_filter.hpp
linick-kang: Done. 8d80b1c6389811fec8bad2e719a47b4964916040

perception/autoware_multi_object_tracker/lib/tracker/model/vehicle_tracker.cpp
linick-kang: Done. 8d80b1c6389811fec8bad2e719a47b4964916040

perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/tracker/model/vehicle_tracker.hpp
linick-kang: Done. 8d80b1c6389811fec8bad2e719a47b4964916040

perception/autoware_multi_object_tracker/lib/association/association.cpp
linick-kang: Done. 8d80b1c6389811fec8bad2e719a47b4964916040

perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/association/adaptive_threshold_cache.hpp
technolojin: can you remove util folder??? util means nothing here.

technolojin (changes requested):

perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/association/adaptive_threshold_cache.hpp
linick-kang: Done. ba16cc336a56ecdd0830a2ad43ae12ada477f76f

technolojin:
[RT0-37297](https://console.data-search.tier4.jp/projects/x2_dev/environments/190ce3fb-2480-4710-a8db-b976814f90cb/rosbags/b1b24e39-becf-4bd6-a238-61fab17d737e) degradation is observed.

At 50 second of the video, tracked result of truck in front of ego has moved forward. It can cause mis-detection or/and unwanted departure. @linick-kang can you investigate that this phenomena related to this PR? or it is occasional random tracking result?  
image

technolojin:
@linick-kang 
please check https://github.com/autowarefoundation/autoware_launch/pull/1511#issuecomment-3623875391

linick-kang:
> [RT0-37297](https://console.data-search.tier4.jp/projects/x2_dev/environments/190ce3fb-2480-4710-a8db-b976814f90cb/rosbags/b1b24e39-becf-4bd6-a238-61fab17d737e) degradation is observed.
> 
> At 50 second of the video, tracked result of truck in front of ego has moved forward. It can cause mis-detection or/and unwanted departure. @linick-kang can you investigate that this phenomena related to this PR? or it is occasional random tracking result? image

@technolojin 
I did some investigations on this data. It seems that conditioned update (front/rear wheel) is likely the cause or a partial cause for this degradation. And some polygon detections resulted from detection_by_tracker or camera_lidar_fusion is also likely related to the moving forward (always observed when tracker moves). I'm still looking into this issue and trying to fix it.
image

> @linick-kang
> please check [autowarefoundation/autoware_launch#1511 (comment)](https://github.com/autowarefoundation/autoware_launch/pull/1511#issuecomment-3623875391)

Thank you for this notice. I updated the config in autoware_universe to align with autoware_launch one.(882c0481c89b9f705be8241b9db7bac5b851f228)

linick-kang:
> [RT0-37297](https://console.data-search.tier4.jp/projects/x2_dev/environments/190ce3fb-2480-4710-a8db-b976814f90cb/rosbags/b1b24e39-becf-4bd6-a238-61fab17d737e) degradation is observed.
> At 50 second of the video, tracked result of truck in front of ego has moved forward. It can cause mis-detection or/and unwanted departure. @linick-kang can you investigate that this phenomena related to this PR? or it is occasional random tracking result? image

@technolojin I managed to solve the degradation in last several commits (1547428dfc70201c640ae3727bbbbca79a528d6c)

technolojin (approved):
LGTM

Issue #10554, opened 2025-04-24
The "cpu_monitor" node takes more than 1 second for publishing while it is expected to publish /diagnostics topic with 1 second intervals.


nishikawa-masaki:
### Checklist

- [x] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [x] I've searched other issues and no duplicate issues were found.
- [x] I'm convinced that this is not my fault but a bug.

### Description

The "cpu_monitor" node of the "autoware_system_monitor" component spends more than 1 second (typically 1.4sec) for publishing "/diagnostics" topic while it is expected to publish the topic with 1 second intervals.
This causes the following problems:
* Extra CPU load is consumed by "autoware_system_monitor" itself.
* The safe margin against error detection is decreased.
  * If "/diagnostics" topic stops being published for more than 3 seconds, it is recognized as an error.

### Expected behavior

The "cpu_monitor" node should publish "/diagnostics" topic with 1 second intervals.

### Actual behavior

The "cpu_monitor" node publishes "/diagnostics" topic with 1.4 second intervals or more.

### Steps to reproduce

1. Run "ros2 launch autoware_system_monitor system_monitor.launch.xml" on a terminal.
2. Run "ros2 topic echo -f /diagnostics > /tmp/log" on another terminal.
3. Inspect timestamp intervals of topics published by "cpu_monitor". The intervals are expected to be about 1 second, but actual intervals are considerably longer than 1 second.
4. Observe that the "execution time" values of 'cpu_monitor: CPU Usage' are more than 1000 ms (1 second), too.

### Possible causes

This is caused by the following reasons:
* The "cpu_monitor" node uses "mpstat -P ALL 1 1 -o JSON" command to get CPU usage.
  * The command sleeps 1 second in it to collect data with fair amount of interval for calculating CPU usage.
    * The interval can't be shorter than 1 second. ("mpstat" takes only integer arguments)
* The "cpu_monitor" node uses boost::process::child() to execute "mpstat" command.
  * Spawning of child processes takes time and introduces extra CPU load.
    * It also interferes with context switching in Linux kernel.
* The "cpu_monitor" node collects statistics information in the publishing thread context.
  * Execution of "mpstat" for CPU load takes 1 second.
  * Other statistics are read from /proc pseudo-filesystem, but it also takes some time.
  * A similar problem with "process_monitor" has been reported and fixed.
    * #939

### Additional context

[...]

PR #10873, opened 2025-06-23
fix(cpu monitor): read /proc/stat directly instead of executing mpstat

nishikawa-masaki:
## Description
As reported on issue #10554 , the "cpu_monitor" node executes the command line "mpstat -P ALL 1 1 -o JSON" to collect load information of CPU cores.
This implementation has 2 problems:
* "mpstat 1 1" sleeps for 1 second in it.
* Spawning of "mpstat" process with boost::process::child() introduces non-negligible amount of CPU load.
  * "cpu_monitor" itself need to wait for spawning of "mpstat" process every second.
  * Other processes may suffer from delay of task switching in the kernel when a new process is added to the task queue and the kernel context is locked.

This PR makes the following changes:
* "Cpu_monitor" does not execute "mpstat" command any more.
* Instead, "cpu_monitor" reads CPU load information directly from "/proc/stat".
* Unit tests about "mpstat" are removed because "mpstat" is not used.

Both of this PR and another preceding PR #10545 are needed to fix the issue #10554.

## Related links

**Parent Issue:**

- [The "cpu_monitor" node takes more than 1 second for publishing while it is expected to publish /diagnostics topic with 1 second intervals. #10554](https://github.com/autowarefoundation/autoware_universe/issues/10554)

**Private Links:**

- [TIER IV internal link](https://tier4.atlassian.net/browse/RT1-9621)

## How was this PR tested?
- All PR CI tests passed.
- The source code can be built for different CPU platforms.
- Testing on other CPU platforms than Intel is being conducted in parallel to this PR.
  - Other CPU platforms also employs Linux as their OS and they have "/proc/stat" file.
  - The pseudo files for measuring CPU temperature and monitoring thermal throttling status are different among platforms.
- The number of entries and their values in "/diagnostics" topic published by the "cpu_monitor" node are same as before.
- All of unit tests passed.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### Improvement on system behavior
With this PR, "cpu_monitor" can publish CPU diagnostics every second as expected, with much lighter CPU load.

### DIFF
--- a/system/autoware_system_monitor/src/cpu_monitor/cpu_monitor_base.cpp
+++ b/system/autoware_system_monitor/src/cpu_monitor/cpu_monitor_base.cpp
@@ -20,12 +20,10 @@
 #include "system_monitor/cpu_monitor/cpu_monitor_base.hpp"
 
 #include "system_monitor/cpu_monitor/cpu_information.hpp"
+#include "system_monitor/cpu_monitor/cpu_usage_statistics.hpp"
 #include "system_monitor/system_monitor_utility.hpp"
 
 #include <boost/filesystem.hpp>
-#include <boost/process.hpp>
-#include <boost/property_tree/json_parser.hpp>
-#include <boost/property_tree/ptree.hpp>
 #include <boost/thread.hpp>
 
 #include <fmt/format.h>
@@ -38,9 +36,7 @@
 #include <utility>
 #include <vector>
 
-namespace bp = boost::process;
 namespace fs = boost::filesystem;
-namespace pt = boost::property_tree;
 
 CPUMonitorBase::CPUMonitorBase(const std::string & node_name, const rclcpp::NodeOptions & options)
 : Node(node_name, options),
@@ -49,7 +45,6 @@ CPUMonitorBase::CPUMonitorBase(const std::string & node_name, const rclcpp::Node
   num_cores_(0),
   temperatures_(),
   frequencies_(),
-  mpstat_exists_(false),
   usage_warn_(declare_parameter<float>(
     "usage_warn", 0.96,
     rcl_interfaces::msg::ParameterDescriptor().set__read_only(true).set__description(
@@ -90,10 +85,6 @@ CPUMonitorBase::CPUMonitorBase(const std::string & node_name, const rclcpp::Node
   num_cores_ = boost::thread::hardware_concurrency();
   usage_warn_check_count_.resize(num_cores_ + 2);   // 2 = all + dummy
   usage_error_check_count_.resize(num_cores_ + 2);  // 2 = all + dummy
-
-  // Check if command exists
-  fs::path p = bp::search_path("mpstat");
-  mpstat_exists_ = (p.empty()) ? false : true;
 
   updater_.setHardwareID(hostname_);
   // Update diagnostic data collected by the timer callback.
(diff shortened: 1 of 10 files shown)


system/autoware_system_monitor/test/src/cpu_monitor/test_intel_cpu_monitor.cpp
```diff
@@ -160,26 +155,15 @@ class CPUMonitorTestSuite : public ::testing::Test
     // Get directory of executable
     const fs::path exe_path(argv_[0]);
     exe_dir_ = exe_path.parent_path().generic_string();
```
nishikawa-masaki: Self comment:
The member "exe_dir_" is unnecessary because testing with dummy "mpstat" has been removed.
The local variable "exe_path" is also unnecessary.

system/autoware_system_monitor/src/cpu_monitor/cpu_usage_statistics.cpp
```diff
@@ -0,0 +1,165 @@
...
+    return;
+  }
+
+  // Process each CPU's statistics
+  for (const auto & core_info : current_statistics_) {
+    const std::string & cpu_name = core_info.name;
+    const CpuStatistics & stats = core_info;
+
+    // Skip if we don't have previous stats for this CPU
+    auto prev_iter =
```
nishikawa-masaki: Self comment:
"stats" and "prev_iter" are abbreviations.
"statistics" and "previous_statistics" may be better.

system/autoware_system_monitor/src/cpu_monitor/cpu_usage_statistics.cpp
```diff
@@ -0,0 +1,165 @@
...
+      std::find_if(previous_statistics_.begin(), previous_statistics_.end(),
+        [cpu_name](const CpuStatistics & statistics) { return statistics.name == cpu_name; });
+    if (prev_iter == previous_statistics_.end()) {
+      continue;
+    }
+
+    // Calculate deltas
+    uint64_t user_delta = stats.user - prev_iter->user;
```
nishikawa-masaki: Self comment:
It would help readers if a comment like "The results never become negative." exists.

system/autoware_system_monitor/src/cpu_monitor/cpu_monitor_base.cpp
```diff
@@ -229,167 +220,52 @@ void CPUMonitorBase::checkUsage()
...
+    cpu_usage.nice = usage.nice_percent;
+    cpu_usage.sys = usage.system_percent;
+    cpu_usage.iowait = usage.iowait_percent;
+    cpu_usage.idle = usage.idle_percent;
+    cpu_usage.total = usage.total_usage_percent;
+    // Some of the members of CpuUsageStatistics::CoreUsageInfo are not used.
+
+    // CpuUsageToLevel() converts usage to warning and error levels.
+    // It also counts occurrence of warning and error in addition to conversion.
+    float total_usage = usage.total_usage_percent * 1e-2;
```
sasakisasaki:
```suggestion
    const float total_usage = usage.total_usage_percent * 1e-2;
```
Consider applying this change if `total_usage` is not reassigned within `CpuUsageToLevel`.

system/autoware_system_monitor/src/cpu_monitor/cpu_usage_statistics.cpp
```diff
@@ -0,0 +1,170 @@
...
+    uint64_t total_delta = user_delta + nice_delta + system_delta + iowait_delta + irq_delta +
+                           softirq_delta + steal_delta;
+    uint64_t total_all_delta = total_delta + idle_delta;
+
+    // Skip if total time delta is zero
+    if (total_all_delta == 0) {
+      continue;
+    }
+
+    float total_all_delta_float = static_cast<float>(total_all_delta);
```
sasakisasaki:
```suggestion
    const float total_all_delta_float = static_cast<float>(total_all_delta);
```

system/autoware_system_monitor/src/cpu_monitor/cpu_usage_statistics.cpp
```diff
@@ -0,0 +1,170 @@
...
+    uint64_t softirq_delta    = current_data.softirq    - previous_data.softirq;
+    uint64_t steal_delta      = current_data.steal      - previous_data.steal;
+    uint64_t guest_delta      = current_data.guest      - previous_data.guest;
+    uint64_t guest_nice_delta = current_data.guest_nice - previous_data.guest_nice;
+    // clang-format on
+
+    // Calculate total time delta
+    uint64_t total_delta = user_delta + nice_delta + system_delta + iowait_delta + irq_delta +
+                           softirq_delta + steal_delta;
+    uint64_t total_all_delta = total_delta + idle_delta;
```
sasakisasaki:
```suggestion
    // Calculate total time delta
    uint64_t total_all_delta = user_delta + nice_delta + system_delta + idle_delta + iowait_delta
                               + irq_delta + softirq_delta + steal_delta;
```
If this is not my misunderstanding, it seems `total_delta` is not used in other places. So we would be able to calculate `total_all_delta` directly.

sasakisasaki (commented):
Perhaps this is my misunderstanding and out of this PR's scope.

Looking [this method (CPUMonitorBase::getFrequencyFileNames())](https://github.com/autowarefoundation/autoware_universe/pull/10873/files#diff-03335b148d89a2466da8daebdda1961db9e265fdf67612159db0fc1b4bb0befaR545-R577), even the `frequencies_` (assumed to be protected) is used, there is no guard by `std::lock_guard<std::mutex> lock_context(mutex_context_);`.

Could you kindly check if this is issue or not? :pray:

sasakisasaki:
@nishikawa-masaki Thank you very much for preparing super cleaned-up PR! I believe this is very nice refactoring work as the attached picture shows. After understanding some minor points from me, I will be agreeing merging this PR. Good job! :+1: 

### Attachment
Screenshot from 2025-06-25 18-57-40

nishikawa-masaki:
> Perhaps this is my misunderstanding and out of this PR's scope.
> 
> Looking [this method (CPUMonitorBase::getFrequencyFileNames())](https://github.com/autowarefoundation/autoware_universe/pull/10873/files#diff-03335b148d89a2466da8daebdda1961db9e265fdf67612159db0fc1b4bb0befaR545-R577), even the `frequencies_` (assumed to be protected) is used, there is no guard by `std::lock_guard<std::mutex> lock_context(mutex_context_);`.
> 
> Could you kindly check if this is issue or not? 🙏

There is a comment at the top of CPUMonitorBase::getFrequencyFileNames():
> // This function is called from a locked context in the timer callback.

This comment is not a perfect solution, but better than using std::recursive_mutex, which is less efficient.
Normal mutex can't be used as it would cause dead-locking.

system/autoware_system_monitor/src/cpu_monitor/cpu_usage_statistics.cpp
nishikawa-masaki: I remember that "total_delta" and "total_all_delta" was distinguished because "idle_delta" is more special than others, but your comment makes sense.
I applied 3 changes to CpuUsageStatistics::update_cpu_statistics().
* Removed "total_delta without idle_delta" and renamed "total_all_delta" to "total_delta" because "totall_all" sounds redundant.
* Add "const" to local variables including "total_delta" and "total_delta_float".
* Moved a half of the function lines to CpuUsageStatistics::update_current_cpu_statistics().
  *  CodeSense complains that the original function is too complex.

system/autoware_system_monitor/src/cpu_monitor/cpu_usage_statistics.cpp
nishikawa-masaki: I fixed it with other changes. Please double-check.

sasakisasaki (approved):
Thank you, I agree with merging this PR.

PR #10875, opened 2025-06-24
feat(lane_change): add infomation to PlanningFactor topic

Kotakku:
## Description
This PR adds the missing information to PlanningFactor in lane_change_module.
I also added a unit test so that colcon test can run tests for the PlanningFactor interface.

## How was this PR tested?
- [x] I checked the PlanningFactor topic using psim and `ros2 topic echo` command.
- [x] colcon test
- [x] [PASS TIER IV INTERNAL SCENARIOS](https://evaluation.tier4.jp/evaluation/reports/936ef239-3eb1-5ce2-a3dd-1ff83d6381e7?project_id=prd_jt)
 
## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


planning/behavior_path_planner/autoware_behavior_path_lane_change_module/test/test_planning_factor.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+Odometry loadOdometryInYaml(const std::string & yaml_file = "vehicle_odometry_data.yaml")
+{
+  return loadMessageInYaml<Odometry>(yaml_file, {"pose"});
+}
+
+PredictedObjects loadPathObjectsInYaml(const std::string & yaml_file = "dynamic_objects_data.yaml")
+{
+  return loadMessageInYaml<PredictedObjects>(yaml_file, {"objects"});
+}
```
zulfaqar-azmi-t4, line 86: there's already parser available in [test_utils](https://github.com/autowarefoundation/autoware_core/blob/main/testing/autoware_test_utils/src/autoware_test_utils.cpp), so just curious, what the motivation for writing a separate parser?

Kotakku: I'm already using the `T parse(const YAML::Node &)` function in `test_utils`. The new function was defined to create a `YAML::Node`. In retrospect, specializing the `T parse(const std::string & filename)` template would have been better, but the amount of code would likely not differ significantly.

zulfaqar-azmi-t4 (approved):

PR #10876, opened 2025-06-24
feat(control_validator): 2 thresholds for the yaw deviation (warn/error)

maxime-clem:
## Description

Add a 2nd threshold for the `yaw_deviation` metric allowing publishing of a warning instead of an error.

Requires the launch PR: https://github.com/autowarefoundation/autoware_launch/pull/1508

## Related links

**Private Links:**

- [TIER IV internal link](https://tier4.atlassian.net/browse/RT1-10098)

## How was this PR tested?

Psim

Example with default parameters (threshold of `0.5` for `WARN` and `1.0` for `ERROR`).

https://github.com/user-attachments/assets/6050ad79-92a3-4219-b454-3b93eef4d6c7

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


kyoichi-sugahara (approved):
Thank you! LGTM!
Not only for yaw_deviation, it's better to have error threshuold option for other check, so can you add TODO for having error threshold option?

PR #10879, opened 2025-06-25
feat(autoware_ground_segmentation): change point follow logic (#294)

kaancolak:
## Description

Issue in Non-Grid Mode Ground Classification (**POINT_FOLLOW** Logic)

Only assign **POINT_FOLLOW** if the previous point was classified as GROUND and the current point is very close to it

The Bug: How Misclassifications Happened
A ray of points hits the side of a car / barrier (**NON_GROUND**) and then the road surface (**GROUND**).

The first few points are correctly identified as **NON_GROUND**.

The algorithm then processes the first point that is actually on the road. This point is very close to the last point on the car's tire / barrier.

The Flaw: The original code would see that this new road point was close to the previous non-ground point and assign it the temporary **POINT_FOLLOW** label.

The code would then resolve this temporary label. It had a rule: "If the previous point was **NON_GROUND** and the current point is **POINT_FOLLOW**, then the current point is also **NON_GROUND**."

The Result: The first actual ground point was incorrectly classified as non-ground. This error would then cascade. The algorithm, now believing the ground was higher than it was, would proceed to misclassify the rest of the actual road points in that ray.

Before - After Video : 
Please check especially **00:51**

https://github.com/user-attachments/assets/e6a84ac9-f436-4570-88ae-f21e7c0f153f

## Related links

- [Tier4 internal link](https://tier4.atlassian.net/browse/ISSJP-2594?focusedCommentId=229275)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_ground_segmentation/src/scan_ground_filter/node.cpp
+++ b/perception/autoware_ground_segmentation/src/scan_ground_filter/node.cpp
@@ -248,30 +248,39 @@ void ScanGroundFilterComponent::classifyPointCloud(
 
       float radius_distance_from_gnd = pd.radius - prev_gnd_radius;
       float height_from_gnd = point_curr.z - prev_gnd_point.z;
-      float height_from_obj = point_curr.z - non_ground_cluster.getAverageHeight();
+      float height_from_obj = 0.0f;
+      if (non_ground_cluster.point_num > 0) {
+        height_from_obj = point_curr.z - non_ground_cluster.getAverageHeight();
+      }
       bool calculate_slope = true;
       bool is_point_close_to_prev =
         (points_distance <
          (pd.radius * radial_divider_angle_rad_ + split_points_distance_tolerance_));
 
-      float global_slope_ratio = point_curr.z / pd.radius;
+      if (is_point_close_to_prev) {
+        if (ground_cluster.point_num > 0) {
+          height_from_gnd = point_curr.z - ground_cluster.getAverageHeight();
+          radius_distance_from_gnd = pd.radius - ground_cluster.getAverageRadius();
+        }
+      }
+
+      float global_slope_ratio = pd.radius > 0.0f ? point_curr.z / pd.radius : 0.0f;
       // check points which is far enough from previous point
       if (global_slope_ratio > global_slope_max_ratio_) {
         point_label_curr = PointLabel::NON_GROUND;
         calculate_slope = false;
       } else if (
-        (point_label_prev == PointLabel::NON_GROUND) &&
+        (point_label_prev == PointLabel::NON_GROUND) && (non_ground_cluster.point_num > 0) &&
         (std::abs(height_from_obj) >= split_height_distance_)) {
         calculate_slope = true;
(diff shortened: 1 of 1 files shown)


technolojin:
@badai-nguyen Can you review this change?

perception/autoware_ground_segmentation/src/scan_ground_filter/node.cpp
```diff
@@ -248,30 +248,39 @@ void ScanGroundFilterComponent::classifyPointCloud(
...
       if (global_slope_ratio > global_slope_max_ratio_) {
         point_label_curr = PointLabel::NON_GROUND;
         calculate_slope = false;
       } else if (
-        (point_label_prev == PointLabel::NON_GROUND) &&
+        (point_label_prev == PointLabel::NON_GROUND) && (non_ground_cluster.point_num > 0) &&
         (std::abs(height_from_obj) >= split_height_distance_)) {
         calculate_slope = true;
-      } else if (is_point_close_to_prev && std::abs(height_from_gnd) < split_height_distance_) {
+      } else if (
```
badai-nguyen, line 276: @kaancolak Thank you for your updating. 
Actually the non-grid classifier' logic were not maintained much recently. But I think the FOLLOW_POINT for NON-GROUND point is neccessary for the case that object has a inclined surface like the below figure. It would be better if you confirm the removing of FOLLOW-POINT for Non-Ground point doesn't cause the degradation if you have such data.
image

BTW, could you check if the misclassification is happened with `elevation_grid_mode` mode?

kaancolak: @badai-nguyen thank you so much for your review!

 If we have an inclined surface like the image illustrated, after this PR, the algorithm calculates the local slope angle using the previous ground cluster ([calculate_slope](https://github.com/autowarefoundation/autoware_universe/blob/main/perception/autoware_ground_segmentation/src/scan_ground_filter/node.cpp#L252)) as default. The surfaces of vehicles can be given as an example of this situation.
 
 

https://github.com/user-attachments/assets/ea1b6135-e21c-4462-a4ff-2a07f5d9660d

Screenshot from 2025-06-27 11-44-59


`elevation_grid_mode` uses a more advanced algorithm and can handle turning cases. However, in the current environment, the area (as shown in the images below) causes `elevation_grid_mode` to fail. Ray based algorithm better handle this case.


ground_remover drawio

badai-nguyen: @kaancolak  Thank you for pointing out and the confirmation. LGTM for this change. 

- By the way, FYI, we have also faced a the same issue with the case before a slope as you mentioned when the density of pointcloud increased.
- [ this PR change ](https://github.com/autowarefoundation/autoware_universe/pull/10686) with [this parameter update](https://github.com/autowarefoundation/autoware_launch/pull/1464) was worked well for us. 
- 
image

badai-nguyen (approved):
LGTM

PR #10882, opened 2025-06-26
feat(out_of_lane): validate predicted paths on lanelets

maxime-clem:
## Description

Add an option to the `out_of_lane` module to validate predicted paths before considering an out of lane collision.

## Related links

Launch PR: https://github.com/autowarefoundation/autoware_launch/pull/1516

**Private Links:**

- [TIER IV internal link](https://tier4.atlassian.net/browse/RT0-37348)

## How was this PR tested?

Psim and rosbag.

## Notes for reviewers

None.

#### Additions and removals

| Added |  `validate_predicted_paths_on_lanelets`  | bool   | true | [-] if true, an out of lane collision is only considered if the predicted path fully follows a sequence of lanelets that include the out of lane lanelet |

## Effects on system behavior

Less false positives with the `out_of_lane`.


mkquda (approved):
LGTM :+1:

PR #10883, opened 2025-06-26
fix(autoware_detected_object_validation): templatize lanelet filter

a-maumau:
## Description
This PR templatizes `lanelet filter` in the `autoware_detected_object_validation` package to support both `DetectedObjects` and `TrackedObjects` messages.

The node name of lanelet filter for `DetectedObjects` remains, so we don't need to modify existing files. 

## Related links

**Parent Issue:**

- Link

## How was this PR tested?
Tested with the test in this PR and Autoware's sample rosbag.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


technolojin (approved):
LGTM

PR #10890, opened 2025-06-27
fix(static_obstacle_avoidance): fix coding error

yhisaki:
## Description

**Avoidance of iterator invalidation**
When you call `push_back` on a `std::vector`, it may reallocate its internal storage, which invalidates all existing iterators and references into that vector. By switching to index‐based access (`lanes[i]`) inside a loop that re-evaluates `lanes.size()` each iteration, you obtain each element directly from the vector’s current storage. This ensures you never hold onto a stale iterator or reference, so your code remains correct even if the vector grows and moves its data.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


go-sakayori (approved):
Very nice! LGTM

PR #10893, opened 2025-06-27
feat(diagnostic_graph_aggregator): add initializing flag

isamu-takagi:
## Description

Add initializing flag that prevents error of diagnostic unit with latch settings during initialization.

## Related links

**Private Links:**

- [TIER IV internal link](https://tier4.atlassian.net/wiki/spaces/AIP/pages/3790963545/2025-06-27)

## How was this PR tested?

Add latch settings to diagnostic graph.
Launch aggregator.
Set initializing flag to true and check latch level.

```
ros2 service call /<node-name>/set_initializing std_srvs/srv/SetBool "data: true"
```

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/system/autoware_diagnostic_graph_aggregator/src/common/graph/levels.cpp
+++ b/system/autoware_diagnostic_graph_aggregator/src/common/graph/levels.cpp
@@ -32,6 +32,14 @@ LatchLevel::LatchLevel(ConfigYaml yaml)
   warn_stamp_ = std::nullopt;
   error_stamp_ = std::nullopt;
   input_level_ = DiagnosticStatus::STALE;
+
+  initializing_ = false;
+}
+
+void LatchLevel::set_initializing(bool initializing)
+{
+  initializing_ = initializing;
+  if (initializing_) reset();
 }
 
 void LatchLevel::reset()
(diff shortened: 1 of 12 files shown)


system/autoware_diagnostic_graph_aggregator/src/common/graph/levels.cpp
```diff
@@ -47,7 +55,7 @@ void LatchLevel::update(const rclcpp::Time & stamp, DiagnosticLevel level)
 {
   input_level_ = level;
 
-  if (latch_enabled_) {
+  if (latch_enabled_ && !initializing_) {
```
sasakisasaki, line 58: Ah, I understand why the initialization flag is needed. We need to update the status only when initialization is done.

sasakisasaki (approved):
Thank you so much. Code is organized cleanly and very readable. I also understand what does this PR do. I agree with merging this PR :+1: 

Perhaps, doing test using not only `true` and also `false` is better as follows.
```
ros2 service call /<node-name>/set_initializing std_srvs/srv/SetBool "data: false"
```

isamu-takagi:
Done false case test.

PR #10895, opened 2025-06-27
feat(tier4_state_rviz_plugin): use mrm description api

isamu-takagi:
## Description

Use MRM description API to display user defined MRM name.
https://autowarefoundation.github.io/autoware-documentation/main/design/autoware-interfaces/ad-api/features/fail-safe/#mrm-description

## Related links

**Private Links:**

- [TIER IV internal link](https://tier4.atlassian.net/browse/RT0-36871)

## How was this PR tested?

Add user defined MRM to [config file](https://github.com/autowarefoundation/autoware_universe/blob/main/system/autoware_default_adapi_universe/config/default_adapi.param.yaml#L36-L44) like following.

```diff
       comfortable_stop:
         behavior: 3
         description: "Comfortable stop"
+      example_mrm:
+        behavior: 9
+        description: "Example MRM"
```

Launch adapi node and rviz.

```bash
ros2 launch autoware_default_adapi_universe test_default_adapi.launch.xml 
rviz2 -d src/launcher/autoware_launch/autoware_launch/rviz/autoware.rviz 
```

Publish mrm state topic and check MRM Behavior text on rviz plugin.

```bash
ros2 topic pub -1 /api/fail_safe/mrm_state autoware_adapi_v1_msgs/msg/MrmState "behavior: 1"
```
- 1: inactive
- 2: emergency stop
- 3: comfortable stop
- 4: pull over
- 9: example mrm
- other: unknown

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


visualization/tier4_state_rviz_plugin/src/autoware_state_panel.cpp
```diff
@@ -139,6 +148,26 @@ void AutowareStatePanel::onInitialize()
...
+      if (!client_list_mrm_->service_is_ready()) return;
+      timer_list_mrm_->cancel();
+      client_list_mrm_->async_send_request(
+        std::make_shared<ListMrmDescription::Request>(),
+        [this](rclcpp::Client<ListMrmDescription>::SharedFuture future) {
+          const auto color = autoware::state_rviz_plugin::colors::default_colors.danger;
+          const auto response = future.get();
+          if (response->status.success) {
+            for (const auto & mrm : response->descriptions) {
+              mrm_behaviors_[mrm.behavior] = {mrm.name, color};
```
sasakisasaki: Perhaps this is just my misunderstanding. The color is always `autoware::state_rviz_plugin::colors::default_colors.danger` for each `mrm.behavior` and `mrm.name`. Could you kindly check if this is expected behavior?

isamu-takagi: There is no information about the color in the API, so the plugin can only set the same color.

visualization/tier4_state_rviz_plugin/src/autoware_state_panel.cpp
```diff
@@ -139,6 +148,26 @@ void AutowareStatePanel::onInitialize()
     "/api/fail_safe/mrm_state", rclcpp::QoS{1}.transient_local(),
     std::bind(&AutowareStatePanel::onMRMState, this, _1));
 
+  client_list_mrm_ =
+    raw_node_->create_client<ListMrmDescription>("/api/fail_safe/list_mrm_description");
+
+  timer_list_mrm_ = rclcpp::create_timer(
+    raw_node_, raw_node_->get_clock(), std::chrono::milliseconds(1000), [this]() {
+      if (!client_list_mrm_->service_is_ready()) return;
+      timer_list_mrm_->cancel();
```
mitsudome-r: Should we cancel the timer here? I think it would be better to do it once we get a valid response in case service request fails for some reasons.

visualization/tier4_state_rviz_plugin/src/autoware_state_panel.cpp
sasakisasaki: I see, thank you for checking.

One more question. According to [these lines](https://github.com/autowarefoundation/autoware_universe/pull/10895/files#diff-ef9fdb8ca0ab4122d38f713ed0e7cfc3e0ac2588d0c045690c0f68178a50a19dR87-R91), there looks four colors. Are following my guesses looks correct?

- In the initialization step, the `mrm_behaviors_` has four colors
- But once this for-loop is triggered, all the colors are eventually overwritten by `autoware::state_rviz_plugin::colors::default_colors.danger` and will not be changed unless re-launch

isamu-takagi: Yes. That's right, but this is the intended behavior. Until now, the type of MRM was predefined, so the plugin decided the MRM and color combination. However, this cannot be supported with MRM description API that dynamically provides MRM information.

visualization/tier4_state_rviz_plugin/src/autoware_state_panel.cpp
isamu-takagi: Fixed it  https://github.com/autowarefoundation/autoware_universe/pull/10895/commits/8e5311d06a80bbc6286858be2c0c500385ff8cae

PR #10897, opened 2025-06-30
feat(goal_planner): add infomation to the PlanningFactor topic 

Kotakku:
## Description
This PR adds the missing information to ```PlanningFactor``` within goal_planner_module.
I also included a unit test so that ```colcon test``` can run tests for the ```PlanningFactor```.

## How was this PR tested?

- [x] I checked the PlanningFactor topic using psim and ```ros2 topic echo /planning/planning_factors/goal_planner``` command.
image
- [x]  pass colcon test
- [x] [PASS TIER IV INTERNAL SCENARIOS](https://evaluation.tier4.jp/evaluation/reports/3f3f19dc-e8f8-5004-821b-442b6a7e74c8?project_id=prd_jt)
Some test cases failed. but I determined that it's not due to this change because similar failures have occurred in other reports as well.

## Notes for reviewers
- The ```safety_factor```'s ```object_id``` currently only includes objects determined to be dangerous. It may not include objects that were deemed not dangerous but still affected the plan. Those objects will be added in the future.
- The ```safety_factor```'s ```predicted_path``` will be added in the future.

## Interface changes

None.

## Effects on system behavior

None.


planning/behavior_path_planner/autoware_behavior_path_goal_planner_module/src/goal_planner_module.cpp
```diff
@@ -1426,8 +1426,24 @@ void GoalPlannerModule::updateSteeringFactor(
     return PlanningFactor::NONE;
   });
 
+  const auto & pull_over_path = context_data.pull_over_path_opt.value();
```
kosuke55, line 1432: better to check has_value()

```suggestion
  if (context_data.pull_over_path_opt.has_value()) {
    return;
  }
  const auto & pull_over_path = context_data.pull_over_path_opt.value();
```
?

kosuke55: ah, need `!`

PR #10898, opened 2025-06-30
feat: enhance bytracker classification management

Owen-Liuyuxuan:
## Description

Proposal for enhancing the label assignment of bytetrack.

### Current problem

We found that for each tracked object, the byte track adheres to the original label created during track initialization, regardless of how the follow-up detection may claim otherwise.

This can be confirmed from the code that the label of a tracker is only intialized but never updated.

### The history of implementation

- ByteTrack is based on the official repos  https://github.com/FoundationVision/ByteTrack
- The original paper/code does not handle "labels". It takes detected objects with IDs and keeps track of the ID; the label is not managed in the original python/C++ codes.
- To adapt the bytetrack to Autoware, we initialized a very simple classification management method that only initializes the label at first, and does not update again.

### The enhanced implementation

- If the tracker is reactivated, it will follow the newly detected label.
- If the tracker is constantly updated, we will take the max score among _the decayed score_ and _the newly detected score_ if the newly detected label is also the current label. When the detected label is different from the current label, it will
    - first decay the current score exponentially.
    - then if the new detection score is higher than the decayed score, it will switch to the new label. Otherwise, it will just keep the current label.

#### Parameter understanding

By setting the parameter `classification_decay_constant` closer to 1, it will tend to keep the current label.

- If we set the parameter `classification_decay_constant` to `1.0`, the mechanism essentially becomes "keeping the label following the frame with the highest detection scores".
- If we set the parameter to `0.0`, the mechanism becomes "always adopt the latest label".

## Related links

**Private Links:**

- [TIER IV internal link](https://star4.slack.com/archives/CEV8XMJBV/p1749799032592199?thread_ts=1749632065.978159&cid=CEV8XMJBV)

## Notes for reviewers

None.

#### Additions and removals

| Added | `classification_decay_constant`   | `double` | `0.95`         |  Exponential decay constant for tracked classification scores |

## Effects on system behavior

We should improve tracking results when the YOLOX classification is not stable (especially starting from the edge of the image or far away.)

### DIFF
--- a/perception/autoware_bytetrack/lib/src/strack.cpp
+++ b/perception/autoware_bytetrack/lib/src/strack.cpp
@@ -44,6 +44,7 @@
 
 #include <yaml-cpp/yaml.h>
 
+#include <algorithm>
 #include <iostream>
 #include <string>
 #include <vector>
@@ -168,13 +169,14 @@ void STrack::update(STrack & new_track, int frame_id, double classification_deca
   this->is_activated = true;
 
   // Classification update algorithm
-  // if new track label is the same, then the label will not change
+  // if new track label is the same, only update the score with decayed score or new score.
   // if new track label is different, we will:
   //       1. decay the existing score
   //       2. choose new/old label based on decayed score and new score
 
   if (this->label == new_track.label) {
-    this->score = new_track.score;  // same with existing methods
+    this->score =
+      std::max(classification_decay_constant * this->score, static_cast<double>(new_track.score));
   } else {
     this->score = this->score * classification_decay_constant;
     if (this->score < new_track.score) {


perception/autoware_bytetrack/lib/src/strack.cpp
```diff
@@ -164,7 +167,18 @@ void STrack::update(STrack & new_track, int frame_id)
...
+  // if new track label is different, we will:
+  //       1. decay the existing score
+  //       2. choose new/old label based on decayed score and new score
+
+  if (this->label == new_track.label) {
+    this->score = new_track.score;  // same with existing methods
+  } else {
+    this->score = this->score * classification_decay_constant;
+    this->label = (this->score > new_track.score) ? this->label : new_track.label;
+  }
```
YoshiRi, line 186: [comment]
Why not updating score with the new_track.score?

perception/autoware_bytetrack/lib/src/strack.cpp
```diff
@@ -164,7 +167,18 @@ void STrack::update(STrack & new_track, int frame_id)
...
 
-  this->score = new_track.score;
+  // Classification update algorithm
+  // if new track label is the same, then the label will not change
+  // if new track label is different, we will:
+  //       1. decay the existing score
+  //       2. choose new/old label based on decayed score and new score
+
+  if (this->label == new_track.label) {
+    this->score = new_track.score;  // same with existing methods
```
YoshiRi: [Just a comment]
When newer class label score is lower than current one, it will be decayed. It sounds weired. :(

perception/autoware_bytetrack/lib/src/strack.cpp
Owen-Liuyuxuan: @YoshiRi 
That is a bug sorry for that. 

My original thought:
If the label changed, we should directly use the new score; if the label not changed, we should keep the decayed score

I need to fix this.

perception/autoware_bytetrack/lib/src/strack.cpp
Owen-Liuyuxuan: @YoshiRi 
When the label is the same, I think it is reasonable to keep the maximum of the current score and the tracked score (instead of just keeping the new score as the original implementation)

Owen-Liuyuxuan:
@YoshiRi 

> Roughly agree with your idea. Personally, I prefer handle each class probability as array and update with low pass filter style and output class label with highest probability. Although this style consumes much more memory than proposed.
> 
> p c l a s s = ( 1 − a ) ∗ p c l a s s − n o w + a ∗ p m e a s u r e m e n t

One sad fact is that I remember Yolox only outputs the classification score with the final prediction label, instead of a list of the full prediction scores. 
The theoretically best method would be modeling the classification distribution into a motion model (slowly falls to uniform) and an observation model (includes new observations).  But YOLOX probably only gives the single best score I remember, so I doubt this is practical.

perception/autoware_bytetrack/lib/src/strack.cpp
YoshiRi: Good. Following is just my hubmle comment and I left it to you.

- I do not like a strong brief effects forever.
- So, I think deprecation should calculated even label is the same with the mesurement.

```c++
// decay should applied first 
this->score = std::max(classification_decay_constant * this->score, new_track.score);
```

Owen-Liuyuxuan: @YoshiRi 
Thank you! I think your idea is excellent and aligns perfectly with my thoughts. 

As of the latest commit (5cf2e89), I’ve fixed the previously discovered bug and updated it to incorporate this formula.

perception/autoware_bytetrack/lib/src/strack.cpp
```diff
@@ -164,7 +168,22 @@ void STrack::update(STrack & new_track, int frame_id)
   this->state = TrackState::Tracked;
   this->is_activated = true;
 
-  this->score = new_track.score;
+  // Classification update algorithm
+  // if new track label is the same, only update the score with decayed score or new score.
+  // if new track label is different, we will:
+  //       1. decay the existing score
+  //       2. choose new/old label based on decayed score and new score
+
```
YoshiRi, line 176: [Just a comment]
You can decay current score first and simplify latter description. This is optional.

```cpp
this->score = this->score * classification_decay_constant;
```

YoshiRi (approved):
LGTM with optional comment

PR #10899, opened 2025-06-30
feat(intersection_collision_checker) reduce false positive occurrence

mkquda:
## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


planning/planning_validator/autoware_planning_validator_rear_collision_checker/src/rear_collision_checker.cpp
```diff
@@ -790,16 +790,15 @@ bool RearCollisionChecker::is_safe(DebugData & debug)
...
   {
     if (is_safe(pointcloud_objects, debug)) {
-      last_safe_time_ = now;
       if ((now - last_unsafe_time_).seconds() > p.common.off_time_buffer) {
+        last_safe_time_ = now;
         return true;
       }
+    } else if ((now - last_safe_time_).seconds() < p.common.on_time_buffer) {
+      RCLCPP_WARN(logger_, "[RCC] Momentary collision risk detected.");
+      return true;
```
satoshi-ota, line 799: @mkquda Looks good. thank you.

planning/planning_validator/autoware_planning_validator_intersection_collision_checker/src/intersection_collision_checker.cpp
```diff
@@ -370,48 +386,59 @@ std::optional<PCDObject> IntersectionCollisionChecker::get_pcd_object(
...
-    const auto arc_length_to_overlap = overlap_arc_coord.length - arc_coord.length;
+  for (const auto & obj : clustered_objects) {
+    if (obj.empty()) continue;
+    double arc_length_sum = 0.0;
+    auto local_min_arc_length = std::numeric_limits<double>::max();
+    auto closest_point = obj.front();
+    for (const auto & point : obj) {
+      const auto center_pose = lanelet::utils::getClosestCenterPose(combine_lanelet, point);
+      const auto arc_coord =
+        lanelet::utils::getArcCoordinates(target_lanelet.lanelets, center_pose);
```
satoshi-ota: @mkquda Although this change was not made in the current update, what is the reason for calculating the closest pose on the centerline here? I assume that getClosestCenterPose does not project onto the centerline itself, so depending on the sampling interval of the centerline, the difference between center_pose and point may not be negligible.

Would there be any issues if we passed point directly to getArcCoordinates?

planning/planning_validator/autoware_planning_validator_intersection_collision_checker/src/intersection_collision_checker.cpp
```diff
@@ -370,48 +386,59 @@ std::optional<PCDObject> IntersectionCollisionChecker::get_pcd_object(
...
+    min_arc_length = local_min_arc_length;
+
+    PCDObject pcd_obj;
+    pcd_obj.last_update_time = time_stamp;
+    pcd_obj.pose.position = closest_point;
+    pcd_obj.overlap_lanelet_id = target_lanelet.id;
+    pcd_obj.track_duration = 0.0;
+    pcd_obj.distance_to_overlap = local_min_arc_length;
+    pcd_obj.delay_compensated_distance_to_overlap = local_min_arc_length;
+    pcd_obj.avg_distance_to_overlap = arc_length_sum / obj.size();
```
satoshi-ota: [memo] calculate average position here.

satoshi-ota (commented):
@mkquda The changes are generally fine. There was just one point I was curious about, so I’d appreciate it if you could clarify. Also, I wasn't sure about the motivation for storing the lanelet to be focused on, so I’d appreciate a brief explanation.

planning/planning_validator/autoware_planning_validator_intersection_collision_checker/src/intersection_collision_checker.cpp
satoshi-ota: By the way, are these state variables easily visible during debugging?

mkquda: I'm not currently visualizing these states in the debug markers, but if you think its necessary I can add them in a separate PR

mkquda:
> @mkquda The changes are generally fine. There was just one point I was curious about, so I’d appreciate it if you could clarify. Also, I wasn't sure about the motivation for storing the lanelet to be focused on, so I’d appreciate a brief explanation.

@satoshi-ota when we compute the target lanelets we also compute the overlap point, if we recompute it every cycle, the overlap point position fluctuates slightly which can introduce noise when measuring object's distance to the overlap point.

planning/planning_validator/autoware_planning_validator_intersection_collision_checker/src/intersection_collision_checker.cpp
mkquda: @satoshi-ota I think the reason was that the function "getArcCoordinates" takes a pose as input, so I thought the orientation information should be set correctly, so instead of passing a pose with uninitialized orientation I just get the nearest center pose and use it.

Ofcourse if you have a better suggestion, please let me know :bowing_man:

satoshi-ota: You're right. When tracing through the getArcCoordinates process, it does indeed use orientation information in the step where it finds the nearest lanelet. If `target_lanelet.lanelets` contains overlapping lanelet pairs, it seems to return the one with the smallest yaw difference. 

Was it originally intended for `target_lanelet.lanelets` to include overlapping lanelets?

satoshi-ota: If overlapping lanelets are not supposed to be included, then I think it's fine to just insert an arbitrary orientation. However, if relying on such implicit behavior is undesirable, one possible approach could be to compute the orientation using `getClosestCenterPose`, combine it with the original point for the position to construct a Pose, and then pass that to `getArcCoordinates`.

planning/planning_validator/autoware_planning_validator_intersection_collision_checker/src/intersection_collision_checker.cpp
satoshi-ota: Yes, I think it would be helpful for debugging. Please consider adding them in a separate PR. Thanks!

planning/planning_validator/autoware_planning_validator_intersection_collision_checker/src/intersection_collision_checker.cpp
mkquda: @satoshi-ota in the current implemetation `target_lanelet.lanelets` should not include overlapping lanelets, so for the sake of reducing unnecessary computation, I will remove the `getClosestCenterPose` step and use arbitrary orientation

planning/planning_validator/autoware_planning_validator_intersection_collision_checker/src/intersection_collision_checker.cpp
mkquda: @satoshi-ota made a [ticket](https://tier4.atlassian.net/browse/RT1-10512)

planning/planning_validator/autoware_planning_validator_intersection_collision_checker/src/intersection_collision_checker.cpp
mkquda: resolved in [3f6e0bc](https://github.com/autowarefoundation/autoware_universe/pull/10899/commits/3f6e0bcedff2630207300145fcb9f176184b3715)

satoshi-ota (approved):
LGTM

kyoichi-sugahara (approved):

PR #10907, opened 2025-07-01
fix(autoware_map_based_prediction): orientation availability update on trackers

technolojin:
## Description

Indirect solution of https://github.com/autowarefoundation/autoware_universe/pull/10906

The tracker initialized in spawn time and the orientation availability is set, but never was updated by input.

Implementing the orientation availability update logic.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


perception/autoware_multi_object_tracker/lib/tracker/model/tracker_base.cpp
```diff
@@ -151,6 +151,18 @@ bool Tracker::updateWithMeasurement(
...
+  // Update orientation availability
+  if (object.kinematics.orientation_availability == types::OrientationAvailability::AVAILABLE) {
+    // if the incoming object is AVAILABLE, set the orientation availability to AVAILABLE
+    object_.kinematics.orientation_availability = types::OrientationAvailability::AVAILABLE;
+  } else if (
+    object.kinematics.orientation_availability == types::OrientationAvailability::SIGN_UNKNOWN &&
+    object_.kinematics.orientation_availability == types::OrientationAvailability::UNAVAILABLE) {
+    // if the incoming object is SIGN_UNKNOWN and the tracker is UNAVAILABLE, set the orientation
+    // availability to SIGN_UNKNOWN
+    object_.kinematics.orientation_availability = types::OrientationAvailability::SIGN_UNKNOWN;
```
YoshiRi, line 163: This mean we never output orientation availability with unknown state, right?

technolojin: there is no meaning to overwrite the state from the others to unknown. 
since the tracker get the state available or sign_unknown once, then the tracker knows its orientation state.

PR #10912, opened 2025-07-03
fix(planning_validator_trajectory_checker): set is_critical_error flag to false at start of validation

mkquda:
## Description

The flag `is_critical_error_` is not reset to false at the start of validation. If a "critical" anomaly is detected at any point during operation, and the flag is set to true, it will remain true forever, causing unexpected behavior when another "noncritical" anomaly is detected later on.

This PR fixes the issue by properly setting the flag to false at the start of validation.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


satoshi-ota (approved):
@mkquda Due to the changes, is_critical_error_ is initially initialized as false, but for evaluation items where is_critical is set to true, is_critical_error_ will become true if an anomaly is detected. So, I understand that the functionality for gently decelerating when a sudden path shift is detected will continue to work as before.

Just to be sure, please run the Evaluator.

PR #10914, opened 2025-07-03
fix(boundary_departure): exception and log handling

zulfaqar-azmi-t4:
## Description

Exception might be thrown while boundary departure is running.
The reproducibility is low, which makes it difficult to apply direct fix. 
To prevent it's causing autoware to crash, we are improving error handling in this PR.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

1. Experiment
2. [TIER IV internal link](https://evaluation.tier4.jp/evaluation/reports/8fa7598f-0131-5a74-901d-d0891a4010aa?project_id=prd_jt)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/motion_velocity_planner/autoware_motion_velocity_boundary_departure_prevention_module/src/boundary_departure_prevention_module.cpp
+++ b/planning/motion_velocity_planner/autoware_motion_velocity_boundary_departure_prevention_module/src/boundary_departure_prevention_module.cpp
@@ -224,11 +224,14 @@ VelocityPlanningResult BoundaryDeparturePreventionModule::plan(
 
   take_data();
 
-  if (!is_data_ready()) {
+  if (const auto invalid_data_opt = is_data_invalid(raw_trajectory_points)) {
+    RCLCPP_WARN_SKIPFIRST_THROTTLE(
+      logger_, *clock_ptr_, throttle_duration_ms, "%s", invalid_data_opt->c_str());
     return {};
   }
 
-  if (is_data_timeout(planner_data->current_odometry)) {
+  if (const auto is_timeout_opt = is_data_timeout(planner_data->current_odometry)) {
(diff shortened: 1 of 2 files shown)


planning/motion_velocity_planner/autoware_motion_velocity_boundary_departure_prevention_module/src/boundary_departure_prevention_module.cpp
```diff
@@ -245,68 +248,66 @@ VelocityPlanningResult BoundaryDeparturePreventionModule::plan(
...
+  try {
+    auto result_opt = plan_slow_down_intervals(raw_trajectory_points, planner_data);
 
-  processing_time_publisher_->publish(std::invoke([&]() {
-    autoware_internal_debug_msgs::msg::Float64Stamped msg;
-    msg.stamp = clock_ptr_->now();
-    msg.data = stopwatch_ms.toc();
-    return msg;
-  }));
+    processing_time_publisher_->publish(std::invoke([&]() {
```
maxime-clem, line 254: (question) what is the advantage of creating the message in a lambda here ?

maxime-clem (approved):
Beautiful code improvement :star_struck:

planning/motion_velocity_planner/autoware_motion_velocity_boundary_departure_prevention_module/src/boundary_departure_prevention_module.cpp
zulfaqar-azmi-t4: There’s really no advantage. Instead of calling from outside and passing the result to publish, I think it’s cleaner and more self-contained to just use a lambda directly.

mkquda (approved):

PR #10917, opened 2025-07-03
feat(control_command_gate): support vehicle_cmd_gate interface

isamu-takagi:
## Description

Add vehicle_cmd_gate interface for comatibility.

## Related links

https://github.com/autowarefoundation/autoware_universe/pull/10169

## How was this PR tested?

1. Launch planning simulation.
1. Check that the vehicle stops when Set Emergency button on RViz is pushed.
1. Check that the vehicle stops when following command is executed.
```bash
ros2 service call /control/vehicle_cmd_gate/set_stop tier4_control_msgs/srv/SetStop "stop: true"
```

## Notes for reviewers

The moderate_stop_interface.cpp and hpp are same as vehicle_cmd_gate.

## Interface changes

None.

## Effects on system behavior

None.


TetsuKawa (approved):
LGTM！

PR #10918, opened 2025-07-03
feat(autoware_default_adapi_universe)!: remove interface ported to Autoware Core

mitsudome-r:
## Description
This removes the AD API implementation ported to Autoware Core from Autoware Universe to avoid maintaining duplicate codes.

## Related links

- https://github.com/autowarefoundation/autoware_core/pull/543

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


isamu-takagi (approved):
LGTM. I checked that the planning simulation works.

PR #10919, opened 2025-07-03
feat(detection_area): set stop point behind actual stop line when ego exceeds max acceleration

yhisaki:
### Overview

Add support for adjusting the stop point by considering the vehicle’s maximum allowable acceleration in the Detection Area module.

### Changes

* **Configuration** (`detection_area.param.yaml`)

  * Added two new parameters under the planner namespace:

    ```yaml
    use_max_acceleration: true    # Enable stop‐point adjustment based on max acceleration  
    max_acceleration: 1.0         # Maximum acceleration in m/s²  
    ```
* **Manager** (`manager.cpp`)

  * Retrieve `use_max_acceleration` and `max_acceleration` from ROS parameters.
* **Scene**

  * **Header** (`scene.hpp`):

    * Introduce a new member `forward_offset_to_stop_line_` to store the computed forward offset.
  * **Implementation** (`scene.cpp`):

    * When `use_max_acceleration` is true and the state transitions to STOP:

      1. Compute the feasible stopping distance using `current_velocity² / (2 · max_acceleration)`.
      2. Subtract the original stop distance to determine how far ahead of the stop line to place the stop point.
      3. Find the corresponding segment on the path and update the stop pose and segment index.
* **Utilities** (`utils.hpp` / `utils.cpp`)

  * Added function

    ```cpp
    double feasible_stop_distance_by_max_acceleration(
      double current_velocity,
      double max_acceleration
    );
    ```

    which returns `current_velocity * current_velocity / (2.0 * max_acceleration)` for use in stop‐point calculation.

### How It Works

When `use_max_acceleration` is enabled, the module computes the distance required to safely decelerate from the current speed without exceeding the specified `max_acceleration`. It then shifts the stop point forward by that distance (clamped to zero if smaller than the original stop line offset), ensuring that the controller never demands deceleration beyond the allowed limit.

### Configuration Example

```yaml
  use_max_acceleration: true      # Enable acceleration‐based stop‐point shift
  max_acceleration: 1.0           # [m/s²] maximum deceleration capability
```

### Impact & Notes

* If `use_max_acceleration` is set to false, the module retains its original behavior (stop exactly at the stop line minus vehicle offset).
* Setting `max_acceleration` too low may shift the stop point excessively far ahead—tune it according to your vehicle’s capabilities.
* No downstream API changes: all new behavior is gated by the added parameters.

## Related links

- https://github.com/autowarefoundation/autoware_launch/pull/1535

## How was this PR tested?

- https://evaluation.tier4.jp/evaluation/reports/308590e7-cb35-5f54-9f35-34d176532c4d?project_id=prd_jt

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_velocity_planner/autoware_behavior_velocity_detection_area_module/README.md
+++ b/planning/behavior_velocity_planner/autoware_behavior_velocity_detection_area_module/README.md
@@ -19,6 +19,8 @@ This module is activated when there is a detection area on the target lane.
 | `state_clear_time`                  | double | [s] when the vehicle is stopping for certain time without incoming obstacle, move to STOPPED state                                                       |
 | `stop_margin`                       | double | [m] a margin that the vehicle tries to stop before stop_line                                                                                             |
 | `dead_line_margin`                  | double | [m] ignore threshold that vehicle behind is collide with ego vehicle or not                                                                              |
+| `use_max_acceleration`              | bool   | [-] whether to consider feasible stop distance based on maximum acceleration when inserting stop point                                                   |
+| `max_acceleration`                  | double | [m/s^2] maximum acceleration used to calculate feasible stop distance when `use_max_acceleration` is true                                                |
 | `hold_stop_margin_distance`         | double | [m] parameter for restart prevention (See Algorithm section)                                                                                             |
 | `distance_to_judge_over_stop_line`  | double | [m] parameter for judging that the stop line has been crossed                                                                                            |
 | `suppress_pass_judge_when_stopping` | bool   | [m] parameter for suppressing pass judge when stopping                                                                                                   |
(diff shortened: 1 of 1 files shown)


kyoichi-sugahara:
@yhisaki 
Please update README

yhisaki:
@kyoichi-sugahara 
Done!

planning/behavior_velocity_planner/autoware_behavior_velocity_detection_area_module/README.md
```diff
@@ -114,3 +116,9 @@ This module has parameter `hold_stop_margin_distance` in order to prevent from t
   ![example](docs/keep_stopping.svg){width=1000}
   <figcaption>inside the hold_stop_margin_distance</figcaption>
 </figure>
+
+#### Feasible stop distance
+
+If `use_max_acceleration` is _true_, the module ensures the vehicle can stop within the physical limit set by `max_acceleration`.
```
kyoichi-sugahara, line 122: @yhisaki 
The stopping points are inserted by considering the braking distance calculated from max_acceleration, and the module does not ensure that the vehicle can actually stop.

PR #10920, opened 2025-07-03
fix(lane_change): add curvature threshold for frenet planner

zulfaqar-azmi-t4:
## Description
Add curvature threshold for frenet planner

## Related links

https://github.com/autowarefoundation/autoware_launch/pull/1536

**Parent Issue:**

- Link

## How was this PR tested?

[TIER IV internal link](https://evaluation.tier4.jp/evaluation/reports/2e20f912-8b56-5922-8cf6-608d89d48c37?project_id=prd_jt)

## Notes for reviewers

None.

## Interface changes

None.

#### Additions and removals

| Added | `th_average_curvature`   | `double` | `0.020`         | Remove path with average curvature above the threshold. Path only removed if there is more than 1 candidate, and the first path is always kept. |

## Effects on system behavior

None.


planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/utils/path.cpp
```diff
@@ -614,6 +614,17 @@ std::vector<lane_change::TrajectoryGroup> generate_frenet_candidates(
            calc_average_curvature(p2.lane_changing.curvatures);
   });
 
+  if (trajectory_groups.size() > 1) {
+    auto subrange = ranges::subrange(trajectory_groups.begin() + 1, trajectory_groups.end());
+
+    auto remove_start = ranges::remove_if(subrange, [&](const auto & traj) {
+      return calc_average_curvature(traj.lane_changing.curvatures) >
```
maxime-clem: *Average* curvature is calculated, but the threshold is for the *max* curvature.
To me checking the max is better, but it may cause issues if the curvature is very noisy.

zulfaqar-azmi-t4: I rename it to th_average_curvature [here](https://github.com/autowarefoundation/autoware_universe/pull/10920/commits/722cd6c709f9e850cdac89ff77f10ed1cad21fc3)

As you said, max curvature is very noise and it also a little difficult to parameterizes.
with average curvature, we can ignore any zero curvature points, and will make the parameterization process simpler.

mkquda (approved):
LGTM

PR #10925, opened 2025-07-04
feat(autoware_object_sorter): add object_sorter package

a-maumau:
## Description
This PR adds a new package, `autoware_object_sorter`.
`autoware_object_sorter` performs the filtering processes of `radar_range_splitter` and `radar_velocity_splitter` in a single node.
The package handles both `DetectedObjects` and `TrackedObjects` messages.

The node only publishes objects that remain after the filtering process (i.e., objects with velocity >= `velocity_threshold` and range >= `range_threshold`).

## Related links

**Parent Issue:**

- Link

## How was this PR tested?
Tested with a rosbag that contains a radar data and confirmed filtering is working.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


technolojin:
tested with pipeline integration https://github.com/autowarefoundation/autoware_universe/pull/10916 (WIP)


https://github.com/user-attachments/assets/b693b50e-ff7b-4b0f-8d88-f928961ee132

technolojin (changes requested):
1. clarify the filters
   a. the velocity threshold is to drop late object
   b. the distance threshold is to drop near object from base_link

2. configuration
    a. make the target frame `base_link`, since not all the user think the distance shall be measured from `base_link`

3. when tf is failed : make the default operation to skip the distance filter, not return empty objects

technolojin:
Additional request: add yourself to the codeowners

a-maumau:
@technolojin I’ve updated the code according to your suggestions, reflected in the commits between 89c6037 to 
456aa52

technolojin (approved):
LGTM

It may replaces autoware_object_range_splitter and autoware_object_velocity_splitter, but should be done with alignment.

PR #10928, opened 2025-07-07
feat(intersection_collision_checker): improve target lane selection and velocity estimation to reduce false positives

mkquda:
## Description

This PR applies the following changes to `intersection_collision_checker` module to reduce false positive occurrence.
- Target lane selection:
  - Use turn lanelets centerline string (instead of trajectory linesting) to create a bounding box to get the set initial candidate lanelets
  - This will prevent unnecessarily checking lanelets that are not around the turn area before ego arrives at turn
- PCD Object velocity estimation:
  - Apply a reset acceleration threshold (if measured acceleration exceeds threshold then reset tracking info)
  - Apply a max velocity threshold to avoid outlier data affecting subsequent measurements

## Related links

- [Launcher PR](https://github.com/autowarefoundation/autoware_launch/pull/1538)

## How was this PR tested?
- PSIM
- LSIM
- [Internal TIER IV Evaluator](https://evaluation.tier4.jp/evaluation/reports/56437ec0-2df2-566c-89e4-bb207e6b9d61?project_id=prd_jt)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

Reduce false positive occurrence for `intersection_collision_checker`

### DIFF
--- a/planning/planning_validator/autoware_planning_validator_intersection_collision_checker/include/autoware/planning_validator_intersection_collision_checker/types.hpp
+++ b/planning/planning_validator/autoware_planning_validator_intersection_collision_checker/include/autoware/planning_validator_intersection_collision_checker/types.hpp
@@ -78,7 +78,7 @@ struct TargetLanelet
 
 struct EgoLanelets
 {
-  lanelet::ConstLanelet first_turn_lanelet;
+  lanelet::ConstLanelets turn_lanelets;
   lanelet::ConstLanelets trajectory_lanelets;
   lanelet::ConstLanelets connected_lanelets;
 };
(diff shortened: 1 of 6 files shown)


planning/planning_validator/autoware_planning_validator_intersection_collision_checker/include/autoware/planning_validator_intersection_collision_checker/types.hpp
```diff
@@ -78,7 +78,7 @@ struct TargetLanelet
 
 struct EgoLanelets
 {
-  lanelet::ConstLanelet first_turn_lanelet;
+  lanelet::ConstLanelets turn_lanelets;
```
satoshi-ota, line 81: @mkquda Could you share the motivation behind representing the lanes that the ego vehicle drives on (used for collision detection) as a vector?

mkquda: @satoshi-ota we need this information later on for target lane selection:
- To get initial set of candidate lanelets for detection space
- To exclude any lanelets the ego drives along from detection space

PR #10930, opened 2025-07-07
fix(autoware_map_based_prediction): take most probable class for prediction

technolojin:
## Description
This is a bugfix found during the radar tracked-object pipeline implementation.

In the previous, the first classification was taken. 
```c++
const auto & label_ = transformed_object.classification.front().label;
```
this is implicit implementation of taking highest probable classification. It works only if the input classification vector is sorted  high to low

- The logic is fixed to check all the probability and take the highest one. 
- Renamed `changeLabelForPrediction` to `changeVRULabelForPrediction` for readability

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


perception/autoware_map_based_prediction/src/map_based_prediction_node.cpp
```diff
@@ -680,9 +680,10 @@ void MapBasedPredictionNode::objectsCallback(const TrackedObjects::ConstSharedPt
       transformed_object.kinematics.pose_with_covariance.pose = pose_in_map.pose;
     }
 
-    // get tracking label and update it for the prediction
-    const auto & label_ = transformed_object.classification.front().label;
-    const auto label = utils::changeLabelForPrediction(label_, object, lanelet_map_ptr_);
+    // get the maximum probability label from the classification array
+    const auto & label_ = utils::getMaxProbabilityLabel(transformed_object.classification);
```
YoshiRi: Do you have any reason to not using `autoware::object_recognition_utils::getHighestProbLabel(transformed_object.classification);` ?

Do you have any meaning to avoid add dependency to object_recognition_utils?
`#include <autoware/object_recognition_utils/object_recognition_utils.hpp>
`

technolojin: fixed https://github.com/autowarefoundation/autoware_universe/pull/10930/commits/690d39b888f4f143fad35b352935e477292ab353

YoshiRi (approved):
LGTM

PR #10933, opened 2025-07-07
feat(behavior_path_planner_common): ignore the turn signals when using the zebra zone

yhisaki:
## Description

When avoidance is performed using only the zebra zone, it is not necessary to output a turn signal. Previously, if there were no lanelets on either side of the current lanelet, it was assumed that a zebra zone was being used, and the turn signal was not output. However, this logic caused the turn signal not to be output even when there were neither adjacent lanelets nor zebra zones. In this PR, the logic has been updated to explicitly check for the presence of zebra zones on either side, and the turn signal is now suppressed only when avoidance is performed solely using zebra zones.

### before

[turn_signal_before.webm](https://github.com/user-attachments/assets/601c5672-3a81-45ef-9b0f-501256597496)

### after

[Screencast from 07-07-2025 05:13:16 PM.webm](https://github.com/user-attachments/assets/e7f989f2-4058-4352-a7ac-53e95391af4d)

## Related links

- https://github.com/autowarefoundation/autoware_core/pull/565
- https://star4.slack.com/archives/C03QW0GU6P7/p1751018455659809

## How was this PR tested?

- https://evaluation.tier4.jp/evaluation/reports/8ea91a15-8580-51da-ae50-9d08eb492d02?project_id=prd_jt

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_path_planner/autoware_behavior_path_planner_common/include/autoware/behavior_path_planner_common/turn_signal_decider.hpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_planner_common/include/autoware/behavior_path_planner_common/turn_signal_decider.hpp
@@ -179,7 +179,7 @@ private:
     return start_shift_length < threshold && end_shift_length < threshold;
   };
 
-  inline bool shiftSideOnlyHatchedRoadMarking(
+  inline bool isAdjacentToHatchedRoadMarking(
     const double start_shift_length, const double end_shift_length,
     const bool left_hatched_road_marking_only, const bool right_hatched_road_marking_only,
     const double threshold) const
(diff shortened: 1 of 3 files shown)


planning/behavior_path_planner/autoware_behavior_path_planner_common/include/autoware/behavior_path_planner_common/turn_signal_decider.hpp
```diff
@@ -179,57 +179,58 @@ class TurnSignalDecider
     return start_shift_length < threshold && end_shift_length < threshold;
   };
 
-  inline bool existShiftSideLane(
-    const double start_shift_length, const double end_shift_length, const bool no_left_lanes,
-    const bool no_right_lanes, const double threshold) const
+  inline bool shiftSideOnlyHatchedRoadMarking(
```
satoshi-ota: I think `isAdjacentToHatchedRoadMarking` might be a more appropriate name, as it clearly indicates that the function returns a boolean and describes its purpose more intuitively.

satoshi-ota (approved):

PR #10934, opened 2025-07-07
fix(simple object merger): handle when tf could not get properly

technolojin:
## Description

When tf is not obtained properly, the transform becomes null pointer.

https://github.com/autowarefoundation/autoware_universe/blob/4d0d0ec40afb827eb87d0a3ca23e4760ced6bbc4/common/autoware_universe_utils/include/autoware/universe_utils/ros/transform_listener.hpp#L60-L75

This PR is to handle this exemption properly.

- if the target frame is the same as the input frame, skip the transform process
- if the obtained transform is nullptr, skip the process

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_simple_object_merger/src/simple_detected_object_merger_node.cpp
+++ b/perception/autoware_simple_object_merger/src/simple_detected_object_merger_node.cpp
@@ -38,18 +38,23 @@ void SimpleDetectedObjectMergerNode::approximateMerger(
   const DetectedObjects::ConstSharedPtr & object_msg0,
   const DetectedObjects::ConstSharedPtr & object_msg1)
 {
-  transform_ = transform_listener_->get_transform(
+  auto transform0 = transform_listener_->get_transform(
     node_param_.new_frame_id, object_msg0->header.frame_id, object_msg0->header.stamp,
     rclcpp::Duration::from_seconds(0.01));
+  if (!transform0) {
+    return;
+  }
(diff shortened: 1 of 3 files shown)


perception/autoware_simple_object_merger/src/simple_detected_object_merger_node.cpp
```diff
@@ -38,18 +38,31 @@ void SimpleDetectedObjectMergerNode::approximateMerger(
...
-    getTransformedObjects(object_msg0, node_param_.new_frame_id, transform_);
-
-  // input1
-  transform_ = transform_listener_->get_transform(
-    node_param_.new_frame_id, object_msg1->header.frame_id, object_msg1->header.stamp,
-    rclcpp::Duration::from_seconds(0.01));
-  DetectedObjects::SharedPtr transformed_objects1 =
-    getTransformedObjects(object_msg1, node_param_.new_frame_id, transform_);
+  DetectedObjects::SharedPtr transformed_objects0;
+  if (node_param_.new_frame_id == object_msg0->header.frame_id) {
```
YoshiRi, line 42: memo: tf2 core has early return but it checks tf tree, so this statement is valuable.

YoshiRi (approved):
LGTM

Issue #10936, opened 2025-07-07
autoware_system_monitor (cpu/gpu_monitor)  cannot report CPU, GPU temperatures on NVIDIA Jetson GLX Orin


nishikawa-masaki:
### Checklist

- [x] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [x] I've searched other issues and no duplicate issues were found.
- [x] I'm convinced that this is not my fault but a bug.

### Description

NVIDIA Jetson AGX Orin is a member of NVIDIA's Tegra family.

On NVIDIA Jetson AGX Orin, the "/diagnostics" topic from the "autoware_system_monitor" contains the following lines for "CPU Temperature" and "GPU Temperature", failing to get the actual temperature data about CPU and GPU.

**Additional Info:**
NVIDIA Jetson AGX Xavier is also a member of NVIDIA's Tegra family.
The current implementation had been tested on NVIDIA Jetson AGX Xavier, but Xavier supports only JetPack 5.x with Ubuntu Linux 20.04.
Xavier doesn't support JetPack 6.x with Ubuntu Linux 22.04, which is necessary for ROS2 Humble (or later) required by current implementation of Autoware.
Briefly, the current implementation of Autoware can't work on NVIDIA Jetson AGX Xavier any more.

### Expected behavior

The following lines should be found in the '/diagnostics' topic.

> status:
> - level: "\0"
>   name: 'cpu_monitor: CPU Temperature'
>   message: OK
>  (measured temperature data follow)

> status:
> - level: "\0"
>   name: 'gpu_monitor: GPU Temperature'
>   message: OK
>  (measured temperature data follow)

### Actual behavior

The following lines are found in the '/diagnostics' topic.
It means that the "autoware_system_monitor (cpu_monitor and gpu_monitor) cannot locate the files for reading temperature data."

> status:
> - level: "\x02"
>   name: 'cpu_monitor: CPU Temperature'
>   message: temperature files not found

> status:
> - level: "\x02"
>   name: 'gpu_monitor: GPU Temperature'
>   message: temperature files not found

### Steps to reproduce

1. On "terminal1", execute the following line.

> $ ros2 launch autoware_system_monitor system_monitor.launch.xml

2. On "terminal2", execute the following line.

> $ ros2 topic echo -f /diagnostics > /tmp/log

3. Investigate the "/tmp/log" file.

### Versions

[...]

PR #10937, opened 2025-07-07
fix(autoware_system_monitor): allow variations of arm64 thermal zone type name

nishikawa-masaki:
## Related links

**Parent Issue:**

- #10936

**Private Links:**

- [TIER IV internal link](https://tier4.atlassian.net/browse/RT1-10584)

## How was this PR tested?

- All of CI workflows have succeeded.
- Correct CPU, GPU temperature (Thermal Zone 0, 1) data are reported as a part of the "/diagnostics" topic on NVIDIA Jetson AGX Orin.

## Notes for reviewers

The names of thermal zone types for CPU and GPU on NVIDIA Jetson AGX Orin are different from those on Jetson AGX Xavier.

**Jetson AGX Xavier: (older model)**
CPU: "CPU-therm"
GPU: "GPU-therm"

**Jetson AGX Orin:**
CPU: "cpu-thermal"
GPU: "gpu-thermal"

**The names of thermal zone types can be read from the following pseudo-files on both platforms.**
CPU : /sys/class/thermal/thermal_zone0/type
GPU : /sys/class/thermal/thermal_zone1/type

To keep backward compatibility and allow possible future variations, type name comparison is executed in the following way.
1. Transform the expected strings to lowercase. ("CPU-therm" -> "cpu-therm", "GPU-therm" -> "gpu-therm")
2. Transform the strings read from "type" files to lowercase. ("cpu-thermal" , "gpu-thermal")
3. Check if the latter contains the former from the first letter.

According to my searching results, the strings for thermal zone types are determined by the drivers for the thermal sensors.
Therefore, it would be safer if we allow possible variations like "cpu-therm" or "CPU-THERM".

Just replacing the string constants "CPU-therm" and "GPU-therm" with "cpu-thermal" and "gpu-thermal" does not seem to be a good idea.

## Interface changes

None.

## Effects on system behavior

None.


ito-san (approved):
LGTM

PR #10946, opened 2025-07-09
fix(simulator.launch): replace object recognition tracking with multi-object tracker integration

technolojin:
## Description

Simplify object tracker in simulation.

In the simulation situation, only the detected object `/perception/object_recognition/objects` is simulated.
Therefore, other tracker merger (such as radar pipeline) is not working anyway.

This PR simplify the object tracker by directly calling the multi-object-tracker, so that unused requirement of configurations are not needed.

This PR will not introduce any logical change. It may reduce computational cost for the simulation.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

tested by launching the planning-simulator

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/launch/tier4_simulator_launch/launch/simulator.launch.xml
+++ b/launch/tier4_simulator_launch/launch/simulator.launch.xml
@@ -93,20 +93,14 @@
         <!-- tracking module -->
         <group>
           <push-ros-namespace namespace="tracking"/>
-          <include file="$(find-pkg-share tier4_perception_launch)/launch/object_recognition/tracking/tracking.launch.xml">
-            <arg
-              name="object_recognition_tracking_radar_object_tracker_data_association_matrix_param_path"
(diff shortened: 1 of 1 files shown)


rej55:
@technolojin 
Thank you for fix!
Could you write following items in description?

* Why it didn't run planning simulator with radar tracker?
* What did this PR change?
  * In my understanding, it runs multi object tracker only instead of all tracking modules (including radar tracker)

technolojin:
> @technolojin Thank you for fix! Could you write following items in description?
> 
>     * Why it didn't run planning simulator with radar tracker?
> 
>     * What did this PR change?
>       
>       * In my understanding, it runs multi object tracker only instead of all tracking modules (including radar tracker)

I added intention of the PR to the description.

rej55 (approved):
I checked it can run planning simulator.
LGTM.

tkimura4 (approved):
LGTM

PR #10948, opened 2025-07-09
fix(behavior_path_static_obstacle_avoidance_module): fix isWithinLanes

yhisaki:
## Description

This pull request modifies the logic of the isWithinLanes function to improve its robustness on curved paths.

Previously, the check used the vehicle's full footprint, including its front corners. On sharp curves, this could result in a part of the footprint protruding beyond the lane boundary, causing isWithinLanes to return false even when the vehicle is reasonably within the drivable area.

To address this issue, the check has been updated to use only a line segment connecting the left and right sides of the vehicle's base_link (i.e., a line orthogonal to the vehicle's longitudinal axis), instead of the full footprint. This provides a more conservative and stable representation of the vehicle's lane occupancy, especially on curved trajectories.

## Related links

This PR solves the following issue.

- https://github.com/autowarefoundation/autoware_universe/issues/10285

## How was this PR tested?

- https://evaluation.tier4.jp/evaluation/reports/efbd2b07-e787-522f-94bb-7a3c8557326a?project_id=prd_jt
- https://evaluation.tier4.jp/evaluation/reports/5abbfb79-92b0-54af-bd20-fc2f491929b5?project_id=prd_jt

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_path_planner/autoware_behavior_path_static_obstacle_avoidance_module/src/utils.cpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_static_obstacle_avoidance_module/src/utils.cpp
@@ -23,6 +23,7 @@
 
 #include <Eigen/Dense>
 #include <autoware_lanelet2_extension/utility/message_conversion.hpp>
+#include <autoware_utils_geometry/boost_geometry.hpp>
 #include <autoware_utils_uuid/uuid_helper.hpp>
 
 #include <boost/geometry/algorithms/buffer.hpp>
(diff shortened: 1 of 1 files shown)


planning/behavior_path_planner/autoware_behavior_path_static_obstacle_avoidance_module/src/utils.cpp
```diff
@@ -1213,9 +1214,14 @@ bool isWithinLanes(
 {
   const auto & rh = planner_data->route_handler;
   const auto & ego_pose = planner_data->self_odometry->pose.pose;
+  double vehicle_width = planner_data->parameters.vehicle_info.vehicle_width_m;
```
go-sakayori:
```suggestion
  const double vehicle_width = planner_data->parameters.vehicle_info.vehicle_width_m;
```

go-sakayori (approved):

PR #10949, opened 2025-07-09
feat(out_of_lane): cut trajectory footprints behind the current ego front

maxime-clem:
## Description

In order to avoid detecting out of lane collisions that are unavoidable, this PR cuts the trajectory footprints that are behind the current ego front.

| Before | After |
| --- | --- |
|<video src="https://github.com/user-attachments/assets/4d160355-9e48-4bd3-bf4c-d682dd148377"/> | <video src="https://github.com/user-attachments/assets/3c60bfda-efca-4021-a85c-72bcbf251784"/> |

## Related links

**Private Links:**

- [TIER IV JIRA](https://tier4.atlassian.net/browse/RT1-10493)

## How was this PR tested?

Psim, rosbag, and evaluator
- [TIER IV Evaluator](https://evaluation.tier4.jp/evaluation/reports/d17a5846-24dc-57ef-a8ee-8fc197924ca3?project_id=prd_jt) : no degradation.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

Remove some possible false positives with the `out_of_lane` module.

### DIFF
--- a/planning/motion_velocity_planner/autoware_motion_velocity_out_of_lane_module/src/footprint.cpp
+++ b/planning/motion_velocity_planner/autoware_motion_velocity_out_of_lane_module/src/footprint.cpp
@@ -28,7 +28,6 @@
 
 #include <lanelet2_core/geometry/Polygon.h>
 
-#include <algorithm>
 #include <vector>
 
 namespace autoware::motion_velocity_planner::out_of_lane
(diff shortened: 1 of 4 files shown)


planning/motion_velocity_planner/autoware_motion_velocity_out_of_lane_module/src/footprint.cpp
```diff
@@ -43,32 +48,48 @@ autoware_utils::Polygon2d make_base_footprint(const PlannerParam & p, const bool
...
+      continue;
+    }
+    if (i + 1 < ego_data.trajectory_points.size()) {
+      arc_length +=
+        autoware_utils::calc_distance2d(trajectory_point, ego_data.trajectory_points[i + 1]);
+    }
+    lanelet::BasicPolygons2d cut_result;
+    boost::geometry::difference(trajectory_footprint, cut_polygon, cut_result);
+    trajectory_footprints.push_back(
+      cut_result.empty() ? lanelet::BasicPolygon2d() : cut_result.front());
```
mkquda, line 92: why push back an empty polygon? just skip if empty?

maxime-clem: In next steps there is an assumption that the indexes of the trajectory points matches with the indexes of the footprints (i.e., `trajectory_footprints[i]` corresponds to `trajectory.points[i]`), therefore we use an empty polygon.

PR #10954, opened 2025-07-10
feat(tier4_autoware_api_launch): add arg for replacing default adapi …

TetsuKawa:
## Description
Enable to change adapi param path form autoware_launch.
This PR is necessary for redundant configuration.

## Related links

Mrm description is needed to change for every production.
https://github.com/autowarefoundation/autoware_universe/pull/10838

## How was this PR tested?

- Confirmed build succeeded
- Real minibus testing

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/launch/tier4_autoware_api_launch/launch/autoware_api.launch.xml
+++ b/launch/tier4_autoware_api_launch/launch/autoware_api.launch.xml
@@ -9,6 +9,9 @@
   <arg name="rosbridge_respawn" default="true"/>
   <arg name="rosbridge_max_message_size" default="10000000"/>
 
+  <!-- Parameter files -->
+  <arg name="default_adapi_param_path" default="$(find-pkg-share autoware_default_adapi)/config/default_adapi.param.yaml"/>
+
   <!-- Deprecated API -->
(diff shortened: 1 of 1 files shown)


launch/tier4_autoware_api_launch/launch/autoware_api.launch.xml
```diff
@@ -9,14 +9,19 @@
   <arg name="rosbridge_respawn" default="true"/>
   <arg name="rosbridge_max_message_size" default="10000000"/>
 
+  <!-- Parameter files -->
+  <arg name="default_adapi_param_path" default="$(find-pkg-share autoware_default_adapi)/config/default_adapi.param.yaml"/>
```
isamu-takagi:
```suggestion
  <arg name="default_adapi_param_path" default="$(find-pkg-share autoware_default_adapi_universe)/config/default_adapi.param.yaml"/>
```

TetsuKawa: It was still using the old package name. Thank you for pointing that out.

PR #10960, opened 2025-07-10
fix(command_mode_switcher_plugins): fix velocity limit message

isamu-takagi:
## Description

Fix velocity limit message type of comfortable stop plugin.

## Related links

None

## How was this PR tested?

Check that the message type is unique using the following command.
```
ros2 topic info /planning/scenario_planning/clear_velocity_limit
ros2 topic info /planning/scenario_planning/max_velocity_candidates 
```

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


TetsuKawa (approved):
LGTM！

PR #10968, opened 2025-07-11
feat(autoware_camera_streampetr): implementation of StreamPETR using tensorrt

SamratThapa120:
## Description

This PR introduces StreamPETR to autoware using TensorRT. This is the first camera-only 3d object detection model in Autoware. 
For more information about the node, [check the readme](https://github.com/SamratThapa120/autoware.universe/tree/feat/streampetr_node/perception/autoware_camera_streampetr#readme)

image

### ~~Example on R-VIZ (pointcloud was not used for inference, only visualization)~~

The following video was taken using six 2880x1860 images as input, so there is high drop rate and unstable predictions.
https://github.com/user-attachments/assets/5512026a-8611-4369-8aee-0c92ff4273a4

### Example on R-VIZ (pointcloud was not used for inference, only visualization)
The following video was taken using six 1440x930 images as input, so there is no drop of images which leads to smoother predictions compared to  2880x1860 resolution. Note that 1440x930 is not the resolution of the input to the model (which is 640x480), it is the resolution of the image in the input image topic which affects the data transfer to the node. The input resolution to the model remains unchanged.

https://github.com/user-attachments/assets/4681ea8c-2765-4fd7-9c38-83d5ba92e698

## Related links

[StreamPETR AWML training code](https://github.com/tier4/AWML/tree/main/projects/StreamPETR)
[StreamPETR Official](https://github.com/exiawsh/StreamPETR/tree/main)
[NVIDIDA DL4AGX TensorRT](https://github.com/NVIDIA/DL4AGX/tree/master/AV-Solutions/streampetr-trt)

## How was this PR tested?

Tested using several internal rosbags for both taxi and bus.
[Example TIER IV internal link for taxi](https://console.data-search.tier4.jp/projects/prd_jt/t4datasets/d7b73bd6-5b73-4511-8c8a-2dfe58ffcc86)
[Example TIER IV internal link for bus](https://console.data-search.tier4.jp/projects/x2_dev/t4datasets/f3056f18-b7f6-4469-8d93-d01a283d9a53)

## Notes for reviewers
You can download the onnx files for testing from the following links and place them under 
`/home/autoware/autoware_data/camera_streampetr`
1.  [simplify_extract_img_feat.onnx](https://download.autoware-ml-model-zoo.tier4.jp/autoware-ml/models/streampetr/streampetr-vov99/t4base/v1.7/simplify_extract_img_feat.onnx)
2. [simplify_position_embedding.onnx](https://download.autoware-ml-model-zoo.tier4.jp/autoware-ml/models/streampetr/streampetr-vov99/t4base/v1.7/simplify_position_embedding.onnx)
3. [simplify_pts_head_memory.onnx](https://download.autoware-ml-model-zoo.tier4.jp/autoware-ml/models/streampetr/streampetr-vov99/t4base/v1.7/simplify_pts_head_memory.onnx)

Build the package, and launch using the following command

```bash
ros2 launch autoware_camera_streampetr tensorrt_stream_petr.launch.xml
```

## Interface changes

None.

## Effects on system behavior

None.

## Acknowledgement

Special thanks to @kminoda @scepter914  for contributions to the development of this package.


lzy910:
where are internal rosbags for both taxi and bus？

SamratThapa120:
@lzy910 I'm sorry at the moment,  the rosbags are only accessible to TIER IV employees due to data privacy concerns.

yukkysaito:
@SamratThapa120 Have you completed the verification using the [tutorial](https://autowarefoundation.github.io/autoware-documentation/main/tutorials/ad-hoc-simulation/rosbag-replay-simulation/)?
If possible, could you show a video of RViz from the tutorial to this PR?

yukkysaito:
@SamratThapa120 If possible, could you show an RViz video to this PR?
It's okay if the camera data is not shown.

SamratThapa120:
@yukkysaito I have added a video of predictions visualized over pointcloud in RVIZ.

yukkysaito:
@SamratThapa120 Thank you
Why is the bounding box flickering?

lzy910:
> @lzy910 I'm sorry at the moment, the rosbags are only accessible to TIER IV employees due to data privacy concerns.

Does the point cloud data of our internal rosbags have a ring field, and how can I add this field to the rosbags without the ring field?

perception/autoware_camera_streampetr/src/node.cpp
```diff
@@ -0,0 +1,404 @@
...
+  const Image::ConstSharedPtr & input_camera_image_msg, const int camera_id)
+{
+  if (!prestep()) return;
+
+  data_store_->update_camera_image(camera_id, input_camera_image_msg);
+
+  if (camera_id == anchor_camera_id_) step();
+}
+
+void StreamPetrNode::camera_image_callback(
```
KSeangTan: Let's use `template` for input_camera_image_msg and combine with the above function?

SamratThapa120:
@yukkysaito The problem is not due to the StreamPETR node itself, but due to camera synchronization problems.
Due to the large size of images in the bus rosbag, there was significant drop in image topics when I play the rosbag in my PC, therefore the images synchronization fails, and the predictions are flickering. So, I have re-generated a new visualization (check PR description) by downsampling the image size in the topics to half the original resolution, which fixes the drop issue -> enables synchronization and gives smoother predictions.

perception/autoware_camera_streampetr/src/node.cpp
```diff
@@ -0,0 +1,404 @@
...
+        [this, roi_i](const CompressedImage::ConstSharedPtr msg) {
+          this->camera_image_callback(msg, static_cast<int>(roi_i));
+        });
+    } else {
+      camera_image_subs_.at(roi_i) = this->create_subscription<Image>(
+        "~/input/camera" + std::to_string(roi_i) + "/image", rclcpp::SensorDataQoS{}.keep_last(1),
+        [this, roi_i](const Image::ConstSharedPtr msg) {
+          this->camera_image_callback(msg, static_cast<int>(roi_i));
+        });
+    }
```
ktro2828: It would be nice to use `image_transport::create_subscription(...)` instead.
It allows us to handle both raw or compressed images, and it decompresses image under the hood.

Then, you don't have to implement `decompress_image` and declare the overload of the callback for `CompressedImage`.

Here is the [example](https://github.com/ktro2828/mmros/blob/d7be33960cf73d3859f2e4b337e6ca471c238600/mmros/src/node/detection2d_node.cpp#L95-L98).

perception/autoware_camera_streampetr/include/autoware/camera_streampetr/cuda_utils.hpp
```diff
@@ -0,0 +1,370 @@
...
+  T * ptr = reinterpret_cast<T *>(workspace);
+  workspace = reinterpret_cast<void *>(reinterpret_cast<uintptr_t>(workspace) + size);
+  return ptr;
+}
+
+template <typename T>
+void clear_async(T * ptr, std::size_t num_elem, cudaStream_t stream)
+{
+  CHECK_CUDA_ERROR(::cudaMemsetAsync(ptr, 0, sizeof(T) * num_elem, stream));
+}
```
ktro2828: It looks some of these functions are duplicated with the functions in `autoware_cuda_utils`. Can you replace to them?

perception/autoware_camera_streampetr/lib/postprocess/non_maximum_suppression.cpp
```diff
@@ -0,0 +1,96 @@
...
+    autoware::object_recognition_utils::getHighestProbLabel(object1.classification);
+  const auto label2 =
+    autoware::object_recognition_utils::getHighestProbLabel(object2.classification);
+
+  // if labels are not the same, and one of them is pedestrian, do not suppress
+  if (label1 != label2 && (label1 == Label::PEDESTRIAN || label2 == Label::PEDESTRIAN)) {
+    return false;
+  }
+
+  const auto sqr_dist_2d = autoware_utils::calc_squared_distance2d(
```
ktro2828: In my local environment, I got the following build error:

```shell
Starting >>> autoware_camera_streampetr
--- stderr: autoware_camera_streampetr                                 
/home/ktro2828/workspace/autoware.universe/perception/autoware_camera_streampetr/lib/postprocess/non_maximum_suppression.cpp: In member function ‘bool autoware::camera_streampetr::NonMaximumSuppression::isTargetPairObject(const DetectedObject&, const DetectedObject&)’:
/home/ktro2828/workspace/autoware.universe/perception/autoware_camera_streampetr/lib/postprocess/non_maximum_suppression.cpp:49:28: error: ‘autoware_utils’ has not been declared
   49 |   const auto sqr_dist_2d = autoware_utils::calc_squared_distance2d(
      |                            ^~~~~~~~~~~~~~
gmake[2]: *** [CMakeFiles/autoware_camera_streampetr_lib.dir/build.make:76: CMakeFiles/autoware_camera_streampetr_lib.dir/lib/postprocess/non_maximum_suppression.cpp.o] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:143: CMakeFiles/autoware_camera_streampetr_lib.dir/all] Error 2
---
Failed   <<< autoware_camera_streampetr [10.7s, exited with code 2]
```

Looks you need to use `autoware_utils_geometry::calc_square_distance2d(...)` instead. I confirmed it occurred no build error.

perception/autoware_camera_streampetr/src/node.cpp
```diff
@@ -0,0 +1,404 @@
...
+      }
+    }
+
+    Eigen::Matrix4f T_lidar2img = K_4x4 * T_lidar2cam;
+    Eigen::Matrix4f T_img2lidar = T_lidar2img.inverse();
+
+    // std::cout << "Final res vector:\n";
+    for (int row = 0; row < 4; ++row) {
+      for (int col = 0; col < 4; ++col) {
+        // std::cout << T_img2lidar(row, col) << " ";
```
KSeangTan: Remove?

perception/autoware_camera_streampetr/src/node.cpp
```diff
@@ -0,0 +1,404 @@
...
+        res.push_back(T_img2lidar(row, col));
+      }
+      // std::cout <<  "\n";
```
KSeangTan: Remove

perception/autoware_camera_streampetr/launch/tensorrt_stream_petr.launch.xml
```diff
@@ -0,0 +1,65 @@
+<launch>
+  <arg name="build_only" default="false" description="shutdown node after TensorRT engine file is built"/>
+  <arg name="debug_mode" default="true" description="debug mode"/>
+  <arg name="param_path" default="$(find-pkg-share autoware_camera_streampetr)/config/tensorrt_stream_petr.param.yaml"/>
+  <arg name="model_path" default="/home/autoware/autoware_data/camera_streampetr"/>
```
YoshiRi:
```suggestion
  <arg name="model_path" default="$(env HOME)/autoware_data/camera_streampetr"/>
```

perception/autoware_camera_streampetr/launch/tensorrt_stream_petr.launch.xml
```diff
@@ -0,0 +1,65 @@
...
+  <arg name="rviz" default="true" description="launch rviz"/>
```
YoshiRi: Use may not want to launch rviz as a default.

perception/autoware_camera_streampetr/launch/tensorrt_stream_petr.launch.xml
```diff
@@ -0,0 +1,65 @@
...
+    <node
+      pkg="rviz2"
+      exec="rviz2"
+      name="rviz2"
+      output="screen"
+      if="$(var rviz)"
+      args="-d $(var rviz_config) -s $(find-pkg-share autoware_launch)/rviz/image/autoware.png"
+      respawn="$(var rviz_respawn)"
+    />
+  </group>
```
YoshiRi: I think it's better to separate rviz settings from this launcher. You can create debug launcher so that user can sorely launch nodes they needs.

perception/autoware_camera_streampetr/config/tensorrt_stream_petr.param.yaml
```diff
@@ -0,0 +1,19 @@
+/**:
+  ros__parameters:
+    # see https://github.com/open-mmlab/mmdetection3d/blob/fe25f7a51d36e3702f961e198894580d83c4387b/mmdet3d/evaluation/metrics/nuscenes_metric.py#L386-L393
+    rois_number: 6
+    model_params:
+      class_names: ["CAR", "TRUCK", "BUS", "BICYCLE", "PEDESTRIAN"]
+      fp16_mode: true
+      use_temporal: true
+      detection_range: [-61.2, -61.2, -10.0, 61.2, 61.2, 10.0]
```
YoshiRi: I think we need more explanation about this parameter.
- Is this array matches with each camera output? or xyz min and xyz max?
- Is these parameter are defined with base_link? or other frame?

perception/autoware_camera_streampetr/config/tensorrt_stream_petr.param.yaml
```diff
@@ -0,0 +1,19 @@
...
+      input_image_height: 480
+      input_image_width: 640
+      num_proposals: 5400
+      max_camera_time_diff: 0.15                   # Maximum permitted time difference between camera images
```
YoshiRi: Just for my understandings.
Are there any parameters already included in onnx file description?
I mean if some of the parameters are highly depends on onnx file, those parameter should be provided with model file.

YoshiRi (commented):
Now trying to build and running.
First review only focused on launch and config file.

perception/autoware_camera_streampetr/src/node.cpp
```diff
@@ -0,0 +1,366 @@
...
+  const std::string engine_backbone_path =
+    initEngine(backbone_path, fp16_mode, false, get_logger());
+  const std::string engine_head_path = initEngine(head_path, fp16_mode, true, get_logger());
+  const std::string engine_position_embedding_path =
+    initEngine(position_embedding_path, fp16_mode, false, get_logger());
+
+  if (build_only) {
+    RCLCPP_INFO(get_logger(), "TensorRT engine files built successfully. Shutting Down...");
+    rclcpp::shutdown();
+  }
```
YoshiRi, line 81: Need to add return statements to prevent running initialization after rclcpp::shutdown. Otherwise you'll get error.

```suggestion
    return;
  }
```

YoshiRi (changes requested):

perception/autoware_camera_streampetr/include/autoware/camera_streampetr/cuda_utils.hpp
```diff
@@ -0,0 +1,304 @@
...
+
+  ~Tensor()
+  {
+    if (ptr) {
+      cudaFree(ptr);
+      ptr = nullptr;
+    }
+  }
+
+  void mov(std::shared_ptr<Tensor> other, cudaStream_t stream)
```
KSeangTan: Is it copy or move?

perception/autoware_camera_streampetr/include/autoware/camera_streampetr/cuda_utils.hpp
```diff
@@ -0,0 +1,304 @@
...
+    if (data.size() != static_cast<size_t>(volume)) {
+      std::cerr << "Data size mismatch! Expected " << volume << " elements." << std::endl;
+      return;
+    }
+
+    size_t dsize = volume * getElementSize(dtype);
+    cudaMemcpy(ptr, data.data(), dsize, cudaMemcpyHostToDevice);
+  }
+
+  std::vector<float> cpu() const
```
KSeangTan: by look at the name, we dont know what's the purpose of this function
maybe rename to `copy_tensor_to_host_buffer`?

perception/autoware_camera_streampetr/include/autoware/camera_streampetr/cuda_utils.hpp
```diff
@@ -0,0 +1,304 @@
...
+  {
+    std::vector<float> buffer(volume);
+    cudaMemcpy(buffer.data(), ptr, volume * sizeof(float), cudaMemcpyDeviceToHost);
+    return buffer;
+  }
+
+  std::vector<char> load_ref(std::string fname)
+  {
+    size_t bsize = volume * sizeof(float);
```
KSeangTan: `std::size_t`

perception/autoware_camera_streampetr/include/autoware/camera_streampetr/cuda_utils.hpp
```diff
@@ -0,0 +1,304 @@
...
+  /**
+   * @brief Save tensor data as a numpy array file (.npy). Useful for debugging.
+   * @param filepath Path where to save the numpy file
+   * @return true if successful, false otherwise
+   */
+  bool save_as_numpy(const std::string & filepath) const
+  {
+    // Check if file already exists
+    if (std::filesystem::exists(filepath)) {
+      std::cerr << "File already exists: " << filepath << std::endl;
```
KSeangTan: It's not an error if file already exists? `std::cerr` doesn't look like an appropriate choice, what about `cout` or `printf` instead?

perception/autoware_camera_streampetr/include/autoware/camera_streampetr/cuda_utils.hpp
```diff
@@ -0,0 +1,304 @@
...
+        return false;
+      }
+    }
+
+    // Copy data from GPU to CPU
+    std::vector<float> cpu_data = cpu();
+
+    // Open file for writing
+    std::ofstream f(filepath, std::ios::out | std::ios::binary);
+    if (!f) {
```
KSeangTan: Avoid one-letter variable

perception/autoware_camera_streampetr/include/autoware/camera_streampetr/network/build_trt.hpp
```diff
@@ -0,0 +1,156 @@
...
+    throw std::runtime_error("Failed to parse ONNX file");
+  }
+
+  // Create config
+  auto config = std::unique_ptr<nvinfer1::IBuilderConfig>(builder->createBuilderConfig());
+  if (!config) {
+    throw std::runtime_error("Failed to create builder config");
+  }
+
+  // Set workspace size (8 GiB)
```
KSeangTan: is `8 GiB` recommended size? I believe it's pretty huge

perception/autoware_camera_streampetr/include/autoware/camera_streampetr/network/memory.cuh
```diff
@@ -0,0 +1,39 @@
...
+{
+  int mem_len = 1280;
+  int pre_len = 1024;
+  void * mem_buf;    // 1024
+  float * pre_buf;   // 1, 1024, 1
+  float * post_buf;  // 1, 1280, 1
+
+  cudaStream_t mem_stream;
+
+  Memory() { cudaMalloc(&mem_buf, sizeof(float) * mem_len); }
```
KSeangTan: Use `cudaMallocAsync`?

perception/autoware_camera_streampetr/include/autoware/camera_streampetr/network/network.hpp
```diff
@@ -0,0 +1,227 @@
...
+      context->enqueueV3(stream);
+    }
+  }
+
+  ~SubNetwork() {}
+
+  void EnableCudaGraph(cudaStream_t stream)
+  {
+    // run first time to avoid allocation
+    this->Enqueue(stream);
```
KSeangTan: If `stream` is called by many places in this class, maybe it's better to have a class member variable`stream_`

perception/autoware_camera_streampetr/launch/tensorrt_stream_petr.launch.xml
SamratThapa120: I removed the rviz for now completely, and leave it upto the user to launch their own rviz.

perception/autoware_camera_streampetr/config/tensorrt_stream_petr.param.yaml
SamratThapa120: > Is this array matches with each camera output? or xyz min and xyz max?

This array defines the maximum range [-x,-y,-z,x,y,z]  upto which the model should make predictions.The model is not trained to predict objects outside this range. So, anything predicted outside this range is wrong, and this parameter is used to filter out objects that are outside this range. It is used more like a rule-based filter, after model makes predictions.

> Is these parameter are defined with base_link? or other frame?

base link. 


I will update the config file to add explanations.

perception/autoware_camera_streampetr/config/tensorrt_stream_petr.param.yaml
SamratThapa120: The parameter highly dependent on onnx files are `rois_number`, `input_image_height` , `input_image_width`, `num_proposals`, `class_names`, `use_temporal`. So, I will move the `rois_number` under model_params too. The `max_camera_time_diff` can be adjusted as necessary, so I will move it to launcher.

perception/autoware_camera_streampetr/src/node.cpp
SamratThapa120: THanks, FIXED

perception/autoware_camera_streampetr/src/node.cpp
SamratThapa120: Based on ktro2828's comment, I moved to using image_transport, so there is no need for two separate functions.

perception/autoware_camera_streampetr/include/autoware/camera_streampetr/network/build_trt.hpp
SamratThapa120: Not really, I just used a big value initially to maximize optimization. I was able to test that it doesn't affect performance with 4GB, so I will set it to that if its okay.

SamratThapa120:
@ktro2828 @YoshiRi @KSeangTan 
Thank you for your comments. I have addressed them, could you please re-review when you have time.

perception/autoware_camera_streampetr/include/autoware/camera_streampetr/network/build_trt.hpp
KSeangTan: Let's make it a variable instead?

SamratThapa120: I made the default value 4GB, and made it possible to override that value using `model_params.workspace_size` parameter during runtime

perception/autoware_camera_streampetr/src/node.cpp
```diff
@@ -0,0 +1,382 @@
...
+        }
+      }
+    }
+    geometry_msgs::msg::TransformStamped transform_stamped;
+    try {
+      transform_stamped =
+        tf_buffer_.lookupTransform(camera_links[i], "base_link", tf2::TimePointZero);
+    } catch (const tf2::TransformException & ex) {
+      throw std::runtime_error(
+        "Could not transform from base_link to " + camera_links[i] + ": " + std::string(ex.what()));
```
ktro2828: [Q] Should we throw exception here even though catching `tf2::TransformException`?

perception/autoware_camera_streampetr/src/node.cpp
```diff
@@ -0,0 +1,382 @@
...
+    }
+  }
+
+  return res;
+}
+
+std::pair<std::vector<float>, std::vector<float>> StreamPetrNode::get_ego_pose_vector() const
+{
+  if (!latest_kinematic_state_ || !initial_kinematic_state_) {
+    throw std::runtime_error("Kinematic states have not been received.");
```
ktro2828: [Q] Should we throw exception here?

perception/autoware_camera_streampetr/src/node.cpp
```diff
@@ -0,0 +1,382 @@
...
+  }
+}
+
+std::vector<float> StreamPetrNode::get_camera_extrinsics_vector(
+  const std::vector<std::string> & camera_links)
+{
+  std::vector<float> intrinsics_all = data_store_->get_camera_info_vector();
+
+  std::vector<float> res;
+  res.reserve(camera_links.size() * 16);
```
ktro2828: [NITS] `16` is the magic number, so we should avoid it and it would be clear to declare as follows:

```cpp
constexpr size_t num_row = 4;
constexpr size_t num_col = 4;
res.reserve(camera_links.size() * num_row * num_col);

// ...
size_t offset = i * num_row * num_col;
for (size_t row = 0; row < num_row; ++row) {
  for (size_t col = 0; col < num_col; ++col) {
    // ...
  }
}
```

ktro2828 (commented):
@SamratThapa120 Thank you for updates! I finished final review, so please take a look.

perception/autoware_camera_streampetr/src/node.cpp
SamratThapa120: Fixed in  2e2dfde

perception/autoware_camera_streampetr/src/node.cpp
SamratThapa120: I agree. I removed the throw statement, and just show an error message, and skip inference in this case
Fixed in  2e2dfde

perception/autoware_camera_streampetr/src/node.cpp
SamratThapa120: I agree. I removed the throw statement, and just show an error message, and skip inference in this case
Fixed in https://github.com/autowarefoundation/autoware_universe/commit/2e2dfde358c581440dfcff3e56b4eb39022c6f3b

SamratThapa120:
@KSeangTan Could you re-review when you have time.

perception/autoware_camera_streampetr/include/autoware/camera_streampetr/network/network.hpp
```diff
@@ -0,0 +1,268 @@
...
+{
+using cuda::Tensor;
+using nvinfer1::DataType;
+using nvinfer1::Dims;
+using nvinfer1::ICudaEngine;
+using nvinfer1::IExecutionContext;
+using nvinfer1::ILogger;
+using nvinfer1::IRuntime;
+
+class SubNetwork
```
KSeangTan, line 76: Not a must, maybe we can modify this part to use `autoware_tensorrt_common` in another PR

perception/autoware_camera_streampetr/include/autoware/camera_streampetr/network/memory.cuh
```diff
@@ -0,0 +1,59 @@
...
+ * limitations under the License.
+ */
+
+#include <NvInferRuntime.h>
+#include <cuda_fp16.h>
+#include <cuda_runtime_api.h>
+
+struct Memory
+{
+  int mem_len;
```
KSeangTan: Would be better if you can explicitly tell what are these in comment

perception/autoware_camera_streampetr/include/autoware/camera_streampetr/network/preprocess.hpp
```diff
@@ -0,0 +1,32 @@
...
+// limitations under the License.
+
+#ifndef AUTOWARE__CAMERA_STREAMPETR__NETWORK__PREPROCESS_HPP_
+#define AUTOWARE__CAMERA_STREAMPETR__NETWORK__PREPROCESS_HPP_
+
+#include <cstdint>
+
+namespace autoware::camera_streampetr
+{
+cudaError_t resizeAndExtractRoi_launch(
```
KSeangTan, line 22: Would you reminding me is this the same function also used in other image pipelines? If so, maybe it's better to move it to a standalone package in another PR.

perception/autoware_camera_streampetr/launch/tensorrt_stream_petr.launch.xml
```diff
@@ -0,0 +1,54 @@
...
+
+    <!-- Image of ~/input/camerai will be the ith image in the model input -->
+    <remap from="~/input/camera0/camera_info" to="/sensing/camera/camera0/camera_info"/>
+    <remap from="~/input/camera1/camera_info" to="/sensing/camera/camera1/camera_info"/>
+    <remap from="~/input/camera2/camera_info" to="/sensing/camera/camera2/camera_info"/>
+    <remap from="~/input/camera3/camera_info" to="/sensing/camera/camera3/camera_info"/>
+    <remap from="~/input/camera4/camera_info" to="/sensing/camera/camera4/camera_info"/>
+    <remap from="~/input/camera5/camera_info" to="/sensing/camera/camera5/camera_info"/>
+
+    <!--  Uncomment if you want to use uncompressed images, and set is_compressed_image to false -->
```
KSeangTan: Possible to make a condition checking for using this?

KSeangTan (approved):
LGTM, left some minor comments

KSeangTan (approved):

perception/autoware_camera_streampetr/launch/tensorrt_stream_petr.launch.xml
SamratThapa120: fixed in [a0a6001](https://github.com/autowarefoundation/autoware_universe/pull/10968/commits/a0a6001e038c9bfee7d75569bf77fe07115ff617)

perception/autoware_camera_streampetr/include/autoware/camera_streampetr/network/memory.cuh
SamratThapa120: fixed in [a0a6001](https://github.com/autowarefoundation/autoware_universe/pull/10968/commits/a0a6001e038c9bfee7d75569bf77fe07115ff617)

perception/autoware_camera_streampetr/include/autoware/camera_streampetr/network/network.hpp
SamratThapa120: fixed in [a0a6001](https://github.com/autowarefoundation/autoware_universe/pull/10968/commits/a0a6001e038c9bfee7d75569bf77fe07115ff617)

perception/autoware_camera_streampetr/include/autoware/camera_streampetr/network/preprocess.hpp
SamratThapa120: This implementation is slightly different from BEVFusion in that, it is meant to save the output in contiguous memory locations for all input cameras. I think it is better to update BEVFusion to use a similar logic, and we can unify the code to a stanalone package in the future.

SamratThapa120:
**NEWPR: https://github.com/autowarefoundation/autoware_universe/pull/11139
This PR was abandoned because I mistakenly signed older commits and force pushed them while taking care of the DCO Check.**

PR #10970, opened 2025-07-11
feat(start_planner_module): register start_planner module in bus stop

mehmetdogru:
## Description

This PR is created to run `start_planner` in case ego is in a bus stop

## Related links

**Parent Issue:**

- https://github.com/autowarefoundation/autoware_universe/issues/10963

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_path_planner/autoware_behavior_path_start_planner_module/src/shift_pull_out.cpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_start_planner_module/src/shift_pull_out.cpp
@@ -289,10 +289,10 @@ std::vector<PullOutPath> ShiftPullOut::calcPullOutPaths(
 
   bool has_non_shifted_path = false;
 
-  // if shift length is too short, add non sifted path
-  constexpr double MINIMUM_SHIFT_LENGTH = 0.01;
+  // if shift length is too short, add non shifted path
   const double shift_length = arc_position_start.distance;
-  const bool is_smaller_than_minimum = std::abs(shift_length) < MINIMUM_SHIFT_LENGTH;
+  const bool is_smaller_than_minimum =
+    std::abs(shift_length) < parameters_.th_distance_to_middle_of_the_road;
 
   if (is_smaller_than_minimum) {
     candidate_paths.push_back(non_shifted_path);
--- a/planning/behavior_path_planner/autoware_behavior_path_start_planner_module/src/start_planner_module.cpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_start_planner_module/src/start_planner_module.cpp
@@ -394,13 +394,14 @@ bool StartPlannerModule::isExecutionRequested() const
   }
 
   // Return false and do not request execution if any of the following conditions are true:
-  // - The start pose is on the centerline or on "waypoints" (custom centerline)
-  // - The vehicle has already arrived at the start position planner.
+  // - The vehicle is on the centerline or on "waypoints" (custom centerline)
+  //    [ignore this check if the vehicle is on a bus stop]
+  // - The vehicle is far enough from the original start position.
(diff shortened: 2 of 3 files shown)


kyoichi-sugahara:
@mehmetdogru 
Can you update README to refletct updated start condtion?
https://github.com/autowarefoundation/autoware_universe/blob/9dc67e8696eb7db0f521295a8b49e5e76f755aae/planning/behavior_path_planner/autoware_behavior_path_start_planner_module/README.md#startend-conditions

planning/behavior_path_planner/autoware_behavior_path_start_planner_module/src/shift_pull_out.cpp
```diff
@@ -289,10 +289,10 @@ std::vector<PullOutPath> ShiftPullOut::calcPullOutPaths(
 
   bool has_non_shifted_path = false;
 
-  // if shift length is too short, add non sifted path
-  constexpr double MINIMUM_SHIFT_LENGTH = 0.01;
+  // if shift length is too short, add non shifted path
   const double shift_length = arc_position_start.distance;
-  const bool is_smaller_than_minimum = std::abs(shift_length) < MINIMUM_SHIFT_LENGTH;
+  const bool is_smaller_than_minimum =
```
kyoichi-sugahara: `th_distance_to_middle_of_the_road` and `MINIMUM_SHIFT_LENGTH` are parameters with different values. Is there an intention to replace it with th_distance_to_middle_of_the_road? 
While I think constant definitions are not preferable, I believe replacing it with parameters of different meanings would be inappropriate.

planning/behavior_path_planner/autoware_behavior_path_start_planner_module/src/start_planner_module.cpp
```diff
@@ -1126,9 +1137,17 @@ bool StartPlannerModule::findPullOutPath(
   PlannerDebugData debug_data{
     planner->getPlannerType(), backwards_distance, collision_check_margin, {}};
 
-  const auto pull_out_path =
-    planner->plan(start_pose_candidate, goal_pose, planner_data_, debug_data);
-  debug_data_vector.push_back(debug_data);
+  const auto pull_out_path = std::invoke([&]() -> std::optional<PullOutPath> {
+    if (isCurrentPoseOnEgoCenterline()) {
```
kyoichi-sugahara: @mehmetdogru 
When `isCurrentPoseOnEgoCenterline` returns true here, no planner path is generated and a centerline-following path is generated instead, correct?

In that case, no planner is selected, `collision_check_margin` is not considered, and irrelevant values being stored in `debug_data`.

To avoid this, I feel it would be appropriate to check `isCurrentPoseOnEgoCenterline` before calling this function and generate the `pull_out_path`.

planning/behavior_path_planner/autoware_behavior_path_start_planner_module/src/shift_pull_out.cpp
mehmetdogru: Thank you @kyoichi-sugahara 

Initially I though both values presented the same thing: to check whether ego is on centerline or not. However, since after the changes no planner will be called if `isCurrentPoseOnEgoCenterline() == true` I will actually revert this change with the latest commit :+1:

planning/behavior_path_planner/autoware_behavior_path_start_planner_module/src/start_planner_module.cpp
mehmetdogru: Thank you @kyoichi-sugahara for this suggestion. 

It actually make so much sense :smile: Please let me know if it is alright for you after the latest commit.

PR #10980, opened 2025-07-14
fix(lane_change): prevent cancel if ego is near terminal end

zulfaqar-azmi-t4:
## Description

When ego is near terminal and lane change is cancelled, there is a high possibility that the module fails to generate path, and ego ended  up fails to change lanes, as shown in the following video

https://github.com/user-attachments/assets/b38c34ca-4f90-41ef-a703-a57c43ba6a27

To solve the above issue, we increase the priority of ego's position check, and prevent canceling even if unsafe. 

⚠️  note that there is a feature to insert stop point if there is an incoming vehicle from the rear.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


maxime-clem (approved):

zulfaqar-azmi-t4:
pre-commit.ci run

PR #10981, opened 2025-07-14
fix(multi_object_tracker): correct index usage of association matrix

technolojin:
## Description

Bug fix of data association matrix > max_dist_matrix usage for tracker pruning.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


ogrex (commented):
There is similar logic in DataAssociation::updateMaxSearchDistances. Should it be fixed together as well?

technolojin:
> There is similar logic in DataAssociation::updateMaxSearchDistances. Should it be fixed together as well?

that part is as intended.

ogrex (approved):
LGTM

PR #10983, opened 2025-07-14
feat(autoware_motion_velocity_obstacle_cruise_module): add obstacle cruise filter time-varying range

XiaoyuWang0601:
## Description
So far, the obstacle_cruise_module evaluates every obstacle within the lateral range in front of the ego car's trajectory. However, the maximum sensing range is approximately 120 meters. The detection of obstacles is very noisy at that distance, which can lead to incorrect behavior sometimes. And sometimes the RSS distance is only several dozen meters, much shorter than 120m. Therefore, it's not necessary to evaluate obstacles that are so far away, such as those above 100m.

So, a varying obstacle filter range is introduced here to filter the unnecessary obstacles based on ego_vel, ego_acc, ego_jerk and obstacle_vel.
I used this table to pick out a min/max value (65m/120m) for safe filter range, and varying inside this range. The max_jerk is chosen as 0.15 according to the table.
image

PSim test:
When the obstacle_cruise virtual wall first appeared, we can tell the difference of the obstacle filter range according to different obstacle_vel. This confirms the obstacle filter range is time-varying now.
image
## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


planning/motion_velocity_planner/autoware_motion_velocity_obstacle_cruise_module/src/obstacle_cruise_module.cpp
```diff
@@ -225,7 +226,27 @@ std::vector<CruiseObstacle> ObstacleCruiseModule::filter_cruise_obstacle_for_pre
       continue;
     }
 
-    // 1.2. Check if the rough lateral distance is smaller than the threshold.
+    // 1.2. Check if the obstacle is out of a filtering distance determined by ego velocity,
+    // obstacle velocity, ego acceleration and ego_jerk.
+    // TODO(xiaoyu-wang): Introduce current ego acceleration to this function.
+    //                    So far, the hard code below is enough.
+    const double obstacle_filter_ego_current_acc = 0.0;
```
takayuki5168, line 233: You can use `/localization/acceleration` topic.

planning/motion_velocity_planner/autoware_motion_velocity_obstacle_cruise_module/src/obstacle_cruise_module.cpp
```diff
@@ -225,7 +226,27 @@ std::vector<CruiseObstacle> ObstacleCruiseModule::filter_cruise_obstacle_for_pre
...
+    const double obstacle_filter_ego_planning_dec = 0.0;
```
takayuki5168, line 234: Is there any reason not to use the jerk and accel/decel pramaters in the common.param.yaml?

planning/motion_velocity_planner/autoware_motion_velocity_obstacle_cruise_module/src/obstacle_cruise_module.cpp
```diff
@@ -225,7 +226,27 @@ std::vector<CruiseObstacle> ObstacleCruiseModule::filter_cruise_obstacle_for_pre
...
+    auto obstacle_filter_dist = autoware::motion_utils::calcDecelDistWithJerkAndAccConstraints(
+      current_vel.linear.x, object->get_lon_vel_relative_to_traj(traj_points),
+      obstacle_filter_ego_current_acc, obstacle_filter_ego_planning_dec,
+      obstacle_filtering_param_.obstacle_filter_ego_jerk,
+      -obstacle_filtering_param_.obstacle_filter_ego_jerk);
+    if (obstacle_filter_dist > obstacle_filtering_param_.obstacle_filter_dist_max_range) {
+      obstacle_filter_dist = obstacle_filtering_param_.obstacle_filter_dist_max_range;
+    } else if (obstacle_filter_dist < obstacle_filtering_param_.obstacle_filter_dist_min_range) {
```
takayuki5168, line 242: If possible, we wanna remove parameters that are a bit hard to understand and tune.
Is this min parameter really necessary?

PR #10987, opened 2025-07-15
fix(autoware_cuda_pointcloud_preprocessor): ensure type and API safety

mojomex:
## Description

This PR implements multiple type, memory and API safety measures for `autoware_cuda_pointcloud_preprocessor`.

While invalid input pointclouds caused an error to be logged, callback execution continued on the invalid pointcloud, leading to undefined behavior. This PR makes the clalback return early upon a received invalid input pointcloud.

The CUDA API returns an error/success for most operations, and it is required to check for errors to ensure defined behavior. The CUDA pointcloud preprocessor had no such checks implemented, so this PR adds them.

Lints regarding integer casting and pointer arithmetic have largely been addressed in this PR, except for the CUDA kernels themselves.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


sensing/autoware_cuda_pointcloud_preprocessor/include/autoware/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor.hpp
```diff
@@ -76,16 +75,16 @@ class CudaPointcloudPreprocessor
 
   int num_rings_{};
   int max_points_per_ring_{};
-  std::size_t num_raw_points_{};
-  std::size_t num_organized_points_{};
+  int num_raw_points_{};
```
manato: Since this member holds the number of points, I feel using `size_t` is a reasonable choice because the value never goes negative.  Besides, assigned data ( 
[`sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor.cu:l320`](https://github.com/autowarefoundation/autoware_universe/blob/675e107392e3d3c79d924632d4e85d19d22e8255/sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor.cu#L320)) is `input_pointcloud_msg.width * input_pointcloud_msg.height`(resulting in [`uint32`](https://docs.ros2.org/foxy/api/sensor_msgs/msg/PointCloud2.html)), which also looks `size_t` is suitable to me.'

Do you have any idea why clang-tidy complains that this should be int?

sensing/autoware_cuda_pointcloud_preprocessor/include/autoware/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor.hpp
```diff
@@ -103,17 +102,17 @@ class CudaPointcloudPreprocessor
...
+  thrust::device_vector<OutputPointType> device_output_points_;
+  thrust::device_vector<std::uint32_t> device_crop_mask_;
+  thrust::device_vector<std::uint8_t> device_nan_mask_;
+  thrust::device_vector<std::uint8_t> device_mismatch_mask_;
+  thrust::device_vector<std::uint32_t> device_ring_outlier_mask_;
+  thrust::device_vector<std::uint32_t> device_indices_;
+  thrust::device_vector<TwistStruct2D> device_twist_2d_structs_;
+  thrust::device_vector<TwistStruct3D> device_twist_3d_structs_;
+  thrust::device_vector<CropBoxParameters> host_crop_box_structs_;
+  thrust::device_vector<CropBoxParameters> device_crop_box_structs_;
```
manato, line 115: As far as I searched, using `{}` on thrust::device_vector is valid to initialize it as an empty vector. Could you teach me your motivation for removing `{}` here?

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor.cu
```diff
@@ -37,68 +40,44 @@ namespace autoware::cuda_pointcloud_preprocessor
...
-  channel_field.datatype = sensor_msgs::msg::PointField::UINT16;
-  channel_field.count = 1;
-
-  static_assert(sizeof(OutputPointType) == 16, "OutputPointType size is not 16 bytes");
-  static_assert(offsetof(OutputPointType, x) == 0);
-  static_assert(offsetof(OutputPointType, y) == 4);
-  static_assert(offsetof(OutputPointType, z) == 8);
-  static_assert(offsetof(OutputPointType, intensity) == 12);
-  static_assert(offsetof(OutputPointType, return_type) == 13);
-  static_assert(offsetof(OutputPointType, channel) == 14);
```
manato, line 78: These `static_assert` seem to be dropped. Is it intended?

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor.cu
```diff
@@ -482,21 +464,21 @@ std::unique_ptr<cuda_blackboard::CudaPointCloud2> CudaPointcloudPreprocessor::pr
...
+  CHECK_CUDA_ERROR(cudaMemcpyAsync(
+    &num_output_points, device_indices + num_organized_points_ - 1, sizeof(int),
+    cudaMemcpyDeviceToHost, stream_));
 
-  cudaStreamSynchronize(stream_);
+  CHECK_CUDA_ERROR(cudaStreamSynchronize(stream_));
 
   // Get information and extract points after filters
-  std::uint32_t num_crop_box_passed_points =
+  long num_crop_box_passed_points =
```
manato: The actual return type of `thrust::count` should depend on the iterator that is passed as an argument. Since `device_crop_mask_` is defined as `thrust::device_vector<std::uint32_t>`, and `long` is a signed value,  I feel using `std::uint32_t` is more intuitively proper. Do you have the intention to use `long`?

manato (commented):
@mojomex 
Thank you for your contribution to improving the code quality. 
Regarding some changes you suggested, I left some comments for my further understanding of your intention. I would appreciate your consideration of them.

sensing/autoware_cuda_pointcloud_preprocessor/include/autoware/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor.hpp
mojomex: I agree, and I will change it back. Some CUDA APIs (such as `SortKeys` used in `querySortWorkspace`) use `int` for array lengths / item counts :nauseated_face: so there was always a warning somewhere. As other places (correctly) use `size_t`, even changing these fields to `int` did not solve all linter errors.

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor.cu
mojomex: The function returns a `typename thrust::iterator_traits<normal_iterator<device_ptr<unsigned int>>>::difference_type (aka long)` (there is no way this would ever be negative, so I wonder why it's a `difference_type` and not a `size_t`...). ~I'll change it to `ssize_t` as that expresses the usage as a size, without raising linter errors for signed/unsigned conversions.~ Turns out, changing to `size_t` doesn't upset Clang-Tidy 🤷‍♂️

sensing/autoware_cuda_pointcloud_preprocessor/include/autoware/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor.hpp
mojomex: This is purely a cosmetic change -- Clang Tidy flags the `{}` as [readability-redundant-member-init](https://clang.llvm.org/extra/clang-tidy/checks/readability/redundant-member-init.html), so half of this file was littered with warnings because of it. Is it fine to leave this change as part of this PR?

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor.cu
mojomex: Addressed in 
5193d98

sensing/autoware_cuda_pointcloud_preprocessor/include/autoware/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor.hpp
mojomex: Reverted to `size_t` in 2af311f

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor.cu
mojomex: I found the placement a bit random - it should be done in `point_types.hpp` instead, but indeed I haven't added it there. Will do!

mojomex: Added in 
5a97d81

sensing/autoware_cuda_pointcloud_preprocessor/include/autoware/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor.hpp
manato: Thanks for explanation. I understand the background!

sensing/autoware_cuda_pointcloud_preprocessor/include/autoware/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor.hpp
manato: Thank you for the detailed description. (I never knew such Clang Tidy function for the first time.) 
IMO, it is fine to leave this change as it is!

manato (approved):
@mojomex 
Thank you for dealing with my comments.  Now it LGTM!

sensing/autoware_cuda_pointcloud_preprocessor/include/autoware/cuda_pointcloud_preprocessor/memory.hpp
```diff
@@ -26,7 +26,7 @@ namespace autoware::cuda_pointcloud_preprocessor
 
 // TODO(knzo25): this checks already exists within the pointcloud_preprocessor. After this PR is
 // merged I will move it somewhere all classes that require checks can access it
-bool is_data_layout_compatible_with_point_xyzircaedt(
+[[nodiscard]] inline bool is_data_layout_compatible_with_point_xyzircaedt(
```
amadeuszsz, line 29: nit:
Using https://github.com/autowarefoundation/autoware_core/blob/bb0ef397fb8eeb6d5b8ab409837b6a954e111d57/common/autoware_point_types/include/autoware/point_types/memory.hpp will cause build error for CUDA library due to indirect PCL inclusion, right? If so, we might later consider slight update over there to make `memory.hpp` useful for CUDA code too.

mojomex: To keep this PR concise I'll address this in another PR.

PR #10989, opened 2025-07-15
fix(behavior_path_planner): resample the path before generating the drivable area

maxime-clem:
## Description

When smoothly connecting the goal to the rest of the path, the `goal_planner` removes some path points prior to the goal. When the path is turning just before the goal, this leads to the curve of the path being lost and can cause issues in the drivable area expansion feature (see https://github.com/autowarefoundation/autoware_universe/issues/9844).

This PR fixes the issue by resampling the path before generating the drivable area, instead of after.

## Related links

**Parent Issue:**

- https://github.com/autowarefoundation/autoware_universe/issues/9844

## How was this PR tested?

Psim
Evaluator [TIER IV INTERNAL LINK](https://evaluation.tier4.jp/evaluation/reports/39c542db-2ecf-5ded-9d84-7508f3d20d20?project_id=prd_jt)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_path_planner/autoware_behavior_path_goal_planner_module/src/default_fixed_goal_planner.cpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_goal_planner_module/src/default_fixed_goal_planner.cpp
@@ -14,7 +14,6 @@
 
 #include "autoware/behavior_path_goal_planner_module/default_fixed_goal_planner.hpp"
 
-#include "autoware/behavior_path_planner_common/utils/path_utils.hpp"
 #include "autoware/behavior_path_planner_common/utils/utils.hpp"
 
 #include <autoware_lanelet2_extension/utility/query.hpp>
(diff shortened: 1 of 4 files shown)


kosuke55 (approved):
Need to resample before drivable area generation.
Only one resample is better but seems good as temporal

- smooth goal connection
- resample :new: !!
- dynamic drivable expansion
 
  - [https://github.com/tier4/autoware_universe/blob/2c9554e17cd945b8deda36c7d3400361b7[…]_planner/autoware_behavior_path_planner/src/planner_manager.cpp](https://github.com/tier4/autoware_universe/blob/2c9554e17cd945b8deda36c7d3400361b76fe63b/planning/behavior_path_planner/autoware_behavior_path_planner/src/planner_manager.cpp#L155)
- resample
  - [https://github.com/tier4/autoware_universe/blob/awf-latest/planning/behavior_path_[…]toware_behavior_path_planner/src/behavior_path_planner_node.cpp](https://github.com/tier4/autoware_universe/blob/awf-latest/planning/behavior_path_planner/autoware_behavior_path_planner/src/behavior_path_planner_node.cpp#L724-L726)

PR #10992, opened 2025-07-15
feat(rear_collision_checker): improve collision detection logic

satoshi-ota:
## Description

This PR includes two functional improvements.

First, a new parameter named `lookahead_time` was introduced to allow more intuitive configuration of the collision detection start timing. Previously, it was difficult to adjust how far ahead collisions should be predicted, which sometimes resulted in delayed transitions to MRM just before a potential collision, as shown in the video below.

```yaml
      adjacent_lane:
        lookahead_time: 4.0                         # [s]
```

| BEFORE | AFTER |
| --- | --- |
|  <video src="https://github.com/user-attachments/assets/8779f5e4-ad20-4266-932a-460b055e8673"/> |  <video src="https://github.com/user-attachments/assets/171a5b6e-b88c-439d-a9d8-e950af121cd9"/> |

Second, the collision detection logic was extended to include objects located in front of the ego vehicle. Previously, such objects were excluded from collision checks, which led to situations like the one shown in the video below, where AEB did not activate. To address this, RCC was enhanced to predict collisions with forward objects as well. Whether or not to predict collisions with forward objects can be configured using the following parameter.

```yaml
      adjacent_lane:
        ...
        check:
          front: true
```

| BEFORE | AFTER |
| --- | --- |
|  <video src="https://github.com/user-attachments/assets/a40bb0c8-83e1-4815-96e3-b6c40f3f8fef"/> |  <video src="https://github.com/user-attachments/assets/fdbe81b8-5f7f-4a9f-b114-ece420265f58"/> |

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/planning_validator/autoware_planning_validator_rear_collision_checker/src/utils.cpp
+++ b/planning/planning_validator/autoware_planning_validator_rear_collision_checker/src/utils.cpp
@@ -134,6 +134,34 @@ std::pair<lanelet::BasicPoint2d, double> get_smallest_enclosing_circle(
   return std::make_pair(center, radius_squared);
 }
 
+auto calc_lookahead_line(
+  const std::shared_ptr<PlanningValidatorContext> & context, const double lookahead_time)
+  -> std::optional<std::pair<autoware_utils::LineString3d, double>>
+{
+  const auto & points = context->data->current_trajectory->points;
+  const auto & ego_pose = context->data->current_kinematics->pose.pose;
+  const auto & current_velocity = context->data->current_kinematics->twist.twist.linear.x;
+  const auto & vehicle_info = context->vehicle_info;
+
+  const auto lookahead_distance = current_velocity * lookahead_time;
+  const auto p_future = autoware::motion_utils::calcLongitudinalOffsetPose(
+    points, ego_pose.position, vehicle_info.max_longitudinal_offset_m + lookahead_distance);
+
+  if (!p_future.has_value()) {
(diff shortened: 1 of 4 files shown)


mkquda (approved):
LGTM.

planning/planning_validator/autoware_planning_validator_rear_collision_checker/src/utils.cpp
```diff
@@ -211,9 +239,8 @@ auto check_shift_behavior(
   }
 
   const auto constraints = parameters.common.ego;
-  const auto reachable_point = calc_predicted_stop_line(
-    context, constraints.nominal_deceleration, constraints.nominal_positive_jerk,
```
maxime-clem, line 215: `constraints.nominal_positive_jerk`, `nominal_negative_jerk`, and `nominal_deceleration` are no longer used. Should they be removed completely ?

maxime-clem (approved):
Looks good but some parameters are no longer used and should probably be removed.

planning/planning_validator/autoware_planning_validator_rear_collision_checker/src/utils.cpp
satoshi-ota: I guess there may be some cases where I want to use. But if I find they are no longer necessary, I'll remove them in other PR.

PR #10997, opened 2025-07-16
feat(autoware_cuda_utils): add thrust utils

mojomex:
## Description

Add functions that facilitate the use of Thrust on specific CUDA streams.
This PR is a prerequisite for the thrust-on-stream performance optimization for `autoware_cuda_pointcloud_preprocessor`.

This PR adds the following:
* a `autoware/cuda_utils/thrust_utils.hpp` header
  * a `thrust_on_stream(stream)` function to get a Thrust execution policy that executes on the stream
  * a `thrust_stream` namespace containing convenient wrappers around `thrust::fill`, `thrust::count` etc. that take a `stream` argument
* unit tests for the above functionality

### Usage

```c++
thrust::fill(thrust_on_stream(my_stream), decive_vec.begin(), device_vec.end(), 42.0);
thrust_stream::fill(decive_vec.begin(), device_vec.end(), 42.0, my_stream);
thrust_stream::fill(device_vec, 42.0, my_stream);

// The same APIs exist for these functions as well:
thrust_stream::fill_n(device_vec, 10, 42.0, my_stream);
size_t result = thrust_stream::count(device_vec, 42.0, my_stream);
```

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

Unit tests pass.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/sensing/autoware_cuda_utils/test/test_thrust_utils.cu
+++ b/sensing/autoware_cuda_utils/test/test_thrust_utils.cu
@@ -73,14 +73,15 @@ TEST_F(ThrustUtilsTest, CountFunctionality)
 {
   auto stream = autoware::cuda_utils::makeCudaStream();
 
-  thrust::device_vector<int> vec(6);
-  thrust::fill(vec.begin(), vec.end(), 1);
-  thrust::fill(vec.begin() + 2, vec.end() - 1, 0);  // Set last 3 elements to 0
+  thrust::device_vector<int> vec(20);
+  thrust::fill(vec.begin(), vec.end(), 0);
+  thrust::fill(vec.begin(), vec.begin() + 3, 1);  // Three '1's at the beginning
+  thrust::fill(vec.end() - 3, vec.end(), 1);      // Three '1's at the end
 
   long count = autoware::cuda_utils::thrust_stream::count(vec, 1, *stream);
 
   // Synchronize to ensure the operation completed
   cudaStreamSynchronize(*stream);
 
-  EXPECT_EQ(count, 3);  // There are three '1's in the vector
+  EXPECT_EQ(count, 6);  // There are six '1's in the vector
 }
--- a/sensing/autoware_cuda_utils/test/test_thrust_utils.cu
+++ b/sensing/autoware_cuda_utils/test/test_thrust_utils.cu
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include "autoware/cuda_utils/cuda_gtest_utils.hpp"
 #include "autoware/cuda_utils/stream_unique_ptr.hpp"
 
 #include <autoware/cuda_utils/thrust_utils.hpp>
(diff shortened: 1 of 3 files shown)


mojomex:
@amadeuszsz @manato I've heard some stories some time ago that CUDA unit tests in our CI didn't work out well, but sinec we have `build-and-test-differential-cuda` this should not be a problem anymore, right? It seems the driver version is too old for the CUDA runtime version we're using :eyes:

amadeuszsz:
> @amadeuszsz @manato I've heard some stories some time ago that CUDA unit tests in our CI didn't work out well, but sinec we have `build-and-test-differential-cuda` this should not be a problem anymore, right? It seems the driver version is too old for the CUDA runtime version we're using 👀

@mojomex 
We don't have GPU runner. You can build, but runtime is not available 😢

mojomex:
@amadeuszsz Ah, I see. Is there a standard way to skip the unit tests in CI? I feel like removing them is not a very sustainable solution, given that our CUDA code is critical for functionality/safety of the software.

amadeuszsz:
> @amadeuszsz Ah, I see. Is there a standard way to skip the unit tests in CI? I feel like removing them is not a very sustainable solution, given that our CUDA code is critical for functionality/safety of the software.

@mojomex 
I can imagine there is a way how to turn off unit tests for particular packages, but it may increase CI configuration complexity. I would rather to comment it out (as we did for CenterPoint and TransFusion) and wait for GPU runner, we really need it.

mojomex:
@amadeuszsz I moved the non-thrust unit tests to PR #10999 for clarity. Let's continue the discussion regarding [your above comment](https://github.com/autowarefoundation/autoware_universe/pull/10997#issuecomment-3077766754) there.

mojomex:
@amadeuszsz @manato The prerequisite PR #10999 has been merged and this PR rebased, it is now ready for review :raised_hands:

sensing/autoware_cuda_utils/test/test_thrust_utils.cu
```diff
@@ -0,0 +1,86 @@
...
+  thrust::fill(vec.begin(), vec.end(), 1);
+  thrust::fill(vec.begin() + 2, vec.end() - 1, 0);  // Set last 3 elements to 0
+
+  long count = autoware::cuda_utils::thrust_stream::count(vec, 1, *stream);
+
+  // Synchronize to ensure the operation completed
+  cudaStreamSynchronize(*stream);
+
+  EXPECT_EQ(count, 3);  // There are three '1's in the vector
+}
```
amadeuszsz: `// Set last 3 elements to 0` is not true I think. It uses half-open range convention, as you can see in first fill. Even though result is correct (because vector is `[1, 1, 0, 0, 0, 1]`), maybe better to make it clear and invert logic - we count ones, not zeros. And also let's check begin and end of vector as these regions are bug prone.
```suggestion
TEST_F(ThrustUtilsTest, CountFunctionality)
{
  auto stream = autoware::cuda_utils::makeCudaStream();

  thrust::device_vector<int> vec(20);
  thrust::fill(vec.begin(), vec.end(), 0);
  thrust::fill(vec.begin(), vec.begin() + 3, 1);  // Three '1's at the beginning
  thrust::fill(vec.end() - 3, vec.end(), 1);      // Three '1's at the end

  long count = autoware::cuda_utils::thrust_stream::count(vec, 1, *stream);

  // Synchronize to ensure the operation completed
  cudaStreamSynchronize(*stream);

  EXPECT_EQ(count, 6);  // There are six '1's in the vector
}
```

amadeuszsz: I checked implementation again and we don't need to test thrust library 😆 Just please make this vector initialization clear 🙏🏻

PR #10998, opened 2025-07-16
perf(autoware_cuda_pointcloud_preprocessor): execute thrust operations on the node's own CUDA stream

mojomex:
## Description

This PR replaces all `thrust::` algorithm calls with their `thrust_stream::` equivalents that were introduced in #10997.
Thrust executes algorithms operating on device vectors on the process's default stream by default, and `thrust_stream` ensures execution on the stream the user provided.

Since Autoware CUDA Pointcloud Preprocessor is using CUDA Blackboard which requires all GPU data pub/sub to be done in one process, multiple CUDA-based nodes are commonly launched into the same node container (= process).
Thus, nodes triggering a sync on the default stream might cause delays and inefficiencies for other nodes in the container.

This PR reduces the `39.2%` default stream utilization to just `0.3%` -- a 130x improvement :innocent: 

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

Ran on software version [TIER IV INTERNAL](https://github.com/tier4/pilot-auto.x2/tree/sensing/beta/v4.3), on data [TIER IV INTERNAL](https://console.mob.tier4.jp/projects/x2_dev/rosbag?viz=stream&agg=count&from_ts=1737521882&to_ts=1753160282&live=true&query=log_file_name%3A%28b364d6c9-a553-42b1-842b-6afd55d3b268_2025-04-23-15-37-43%29&file_id=4e333ff5-4cae-4916-82d2-63ea9714dbbe).

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

Expected performance increase in LiDAR pipeline, including Centerpoint. Less contention on default stream.

### DIFF
--- a/sensing/autoware_cuda_pointcloud_preprocessor/include/autoware/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor.hpp
+++ b/sensing/autoware_cuda_pointcloud_preprocessor/include/autoware/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor.hpp
@@ -20,6 +20,7 @@
 
 #include <Eigen/Core>
 #include <Eigen/Geometry>
+#include <autoware/cuda_utils/thrust_utils.hpp>
 #include <cuda_blackboard/cuda_pointcloud2.hpp>
 
 #include <geometry_msgs/msg/transform_stamped.hpp>
--- a/sensing/autoware_cuda_pointcloud_preprocessor/include/autoware/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor.hpp
+++ b/sensing/autoware_cuda_pointcloud_preprocessor/include/autoware/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor.hpp
@@ -67,6 +67,8 @@ public:
     const std::uint32_t first_point_rel_stamp_nsec);
 
 private:
+  static cudaStream_t initialize_stream();
+
   void organizePointcloud();
 
   CropBoxParameters self_crop_box_parameters_{};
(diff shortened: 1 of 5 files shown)


mojomex:
# Evaluation

Default stream usage is reduced from ~39% to ~0.3% by setting the right execution policy. :tada: 

The remaining usage stems from `thrust::fill` calls in `device_vector::resize()` due to the execution policy defaulting to one that uses the default stream. I was not successful in coaxing the allocator->execution policy deduction into yielding a `thrust::cuda_cub::par.on(stream)` instead.
Accordingly, using the mempool allocator does not have a measurable effect on default stream usage.

| Commit | Stream Usage |
| :--- | --- |
| 2161d822d8 (base) | 2161d822d8 |
| 20d2c1466a (+ thrust_stream) | 20d2c1466a |
| ~04f2f97541 (+ mempool alloc)~ | 04f2f97541 |
| cd362b4632 (- mempool alloc) | cd362b4632 |

04f2f97541 was not resulting in any performance gains, so it was reverted and the pre-existing non-functional mempool allocator was removed in cd362b4632.

mojomex:
Next steps:

`device_vector.resize` still runs on the default stream. This is because the execution poicy Thrust uses internally is derived from the device vector's `iterator_system`, which is derived through a large chain of templated code, and ultimately resolves to `thurst::device` for `device_vector`. I tried hacking my way around this by defining my own iterator system and get Thrust to use it, but this turned out to be either not possible or extremely hacky.

Thus, the next step is to implement either
1. a CUDA preprocessor that never calls `resize` during runtime (needs parameters for expected input sizes :slightly_frowning_face:)
2. a subclass of `thrust::device_vector` or `thrust::base_vector` that uses the correct execution policy for `resize`

Currently, the `TwistSruct2D` vector is resized on almost every iteration as the queue length fluctuates ever-so-slightly (e.g. 29, 31, 32, 30, ...). This could be optimized as well, e.g. by keeping older-than-necessary data up to a threshold.

mojomex:
@amadeuszsz @manato This is ready for review, please have a look once you find the time.
 I think it's gone under between all the PRs and rebases :bow:

manato (approved):
@mojomex 
Sorry for the delay in my reviewing. My concern was unexpected asynchronous behavior caused by vanishing implicit synchronization by introducing non-default stream operations. I checked this point, and there seemed to be no problem. So LGTM!

PR #10999, opened 2025-07-16
test(cuda_utils): add unit tests, CUDA test helpers that skip if no CUDA device is available

mojomex:
## Description

After coming to the sad realization in #10997 that our CI currently has no GPUs and thus cannot execute CUDA unit tests, I implemented a minimal set of helpers to run CUDA tests if a GPU is available, while skipping if there is not.

The helpers are in `autoware/cuda_utils/cuda_gtest_utils.hpp` and consist of:
* the `bool is_cuda_runtime_available()` function
* the `SKIP_TEST_IF_CUDA_UNAVAILABLE()` for use with individual test cases
* the `class CudaTest : public ::testing::Test` class to use as the base class for a suite of CUDA-related tests

I also moved the non-thrust unit tests from #10997 to here since they are unrelated to that PR, and serve as validation for the gtest utils implemented here.

Unit tests do not use `ament_add_gtest` directly, as CUDA code would be compiled with a C++ compiler and thus fail compilation.
Instead, the executable is build by hand using `cuda_add_executable`, and the tests are registered by `ament_add_gtest_test`.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

Unit tests pass on my local machine, and are correctly skipped in [CI](https://github.com/autowarefoundation/autoware_universe/actions/runs/16317329159/job/46086319301?pr=10999#logs):

```
 [==========] 13 tests from 4 test suites ran. (0 ms total)
1: [  PASSED  ] 2 tests.
1: [  SKIPPED ] 11 tests, listed below:
1: [  SKIPPED ] CudaUniquePtrTest.MakeUniqueDeviceMemory
1: [  SKIPPED ] CudaUniquePtrTest.MakeUniqueDeviceArray
1: [  SKIPPED ] CudaUniquePtrTest.MakeUniqueHostMemory
1: [  SKIPPED ] CudaUniquePtrTest.MakeUniqueHostArray
1: [  SKIPPED ] CudaUniquePtrTest.DeleterFunctionality
1: [  SKIPPED ] CudaUtilsTest.ClearAsyncFunction
1: [  SKIPPED ] StreamUniquePtrTest.MakeCudaStreamDefault
1: [  SKIPPED ] StreamUniquePtrTest.MakeCudaStreamWithFlags
1: [  SKIPPED ] StreamUniquePtrTest.StreamDeleterFunctionality
1: [  SKIPPED ] StreamUniquePtrTest.StreamReset
1: [  SKIPPED ] StreamUniquePtrTest.StreamMove
```

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/sensing/autoware_cuda_utils/include/autoware/cuda_utils/cuda_gtest_utils.hpp
+++ b/sensing/autoware_cuda_utils/include/autoware/cuda_utils/cuda_gtest_utils.hpp
@@ -1,0 +1,67 @@
+// Copyright 2022 Tier IV, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// This code is licensed under CC0 1.0 Universal (Public Domain).
+// You can use this without any limitation.
+// https://creativecommons.org/publicdomain/zero/1.0/deed.en
+// borrowed from https://proc-cpuinfo.fixstars.com/2019/02/cuda_smart_pointer/
+
+#ifndef AUTOWARE__CUDA_UTILS__CUDA_GTEST_UTILS_HPP_
+#define AUTOWARE__CUDA_UTILS__CUDA_GTEST_UTILS_HPP_
+
+#include <cuda_runtime.h>
+#include <gtest/gtest.h>
+
+#define SKIP_TEST_IF_CUDA_UNAVAILABLE()                              \
+  if (!autoware::cuda_utils::is_cuda_runtime_available()) {          \
+    GTEST_SKIP() << "CUDA runtime is not available. Skipping test."; \
+  }
+
+namespace autoware::cuda_utils
+{
+
+/**
+ * @brief Check if CUDA is available at runtime.
+ * @return true if CUDA is available, false otherwise.
+ */
+inline bool is_cuda_runtime_available()
+{
+  int device_count = 0;
+  cudaError_t error = cudaGetDeviceCount(&device_count);
+  return (error == cudaSuccess && device_count > 0);
+}
+
+/**
+ * @brief Base class for CUDA-related tests that runs only if CUDA is available.
+ *
+ * Usage: Inherit from this class and use the child class as a test fixture.
+ * Example:
+ *
+ * ```c++
+ * class MyTestSuite : public autoware::cuda_utils::CudaTest {};
+ *
+ * TEST_F(MyTestSuite, MyTest) {
+ *   // Your test code here
+ * }
+ * ```
+ */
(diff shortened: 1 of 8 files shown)


mojomex:
@amadeuszsz I've moved the non-thrust unit tests from #10997 to this PR.

In response to [this comment](https://github.com/autowarefoundation/autoware_universe/pull/10997#issuecomment-3077766754):
> @mojomex I can imagine there is a way how to turn off unit tests for particular packages, but it may increase CI configuration complexity. I would rather to comment it out (as we did for CenterPoint and TransFusion) and wait for GPU runner, we really need it.

I've also added a minimal helper to skip CUDA-related tests in unsupported environments. I think this is the cleanest solution, as it will work automatically on GPU runners, and on local machines. It will also explicitly skip the tests, such that it is visible in CI logs that not all tests ran. This might be worth adopting in CenterPoint and TransFusion too, as commented-out tests that are never run could very easily become outdated.

Let me know what you think.

sensing/autoware_cuda_utils/include/autoware/cuda_utils/cuda_gtest_utils.hpp
```diff
@@ -0,0 +1,67 @@
...
+{
+
+/**
+ * @brief Check if CUDA is available at runtime.
+ * @return true if CUDA is available, false otherwise.
+ */
+inline bool is_cuda_runtime_available()
+{
+  int device_count = 0;
+  cudaError_t error = cudaGetDeviceCount(&device_count);
```
amadeuszsz, line 41: nit:
[cudaGetDeviceCount](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__DEVICE.html#group__CUDART__DEVICE_1g18808e54893cfcaafefeab31a73cc55f) is not thread safe. Colcon (whether CI runner with GPU or local machine) run tests for multiple packages -> First thread calls `cudaGetDeviceCount()` and triggers CUDA runtime initialization -> Second thread calls `cudaGetDeviceCount()` simultaneously and due to race condition might get initialization error `cudaErrorInitializationError`.

I marked it as nit because it will fail silently - test will be skipped.

amadeuszsz (approved):
LGTM! Just minor comment, not sure if we need to address it.

This PR makes unit tests for CUDA way easier to implement, thank you!

> This might be worth adopting in CenterPoint and TransFusion too, as commented-out tests that are never run could very easily become outdated.

For sure, we will soon address it for remaining packages!

sensing/autoware_cuda_utils/include/autoware/cuda_utils/cuda_gtest_utils.hpp
mojomex: I read the documentation you shared, but I am not finding info on thread safety.

I let copilot make a script that hammers `cudaGetDeviceCount` with 200 threads for 5 seconds and collects any non-successes. In multiple runs, it did not result in any errors, which leads me to believe CUDA is ensuring thread safety internally. Didn't find any docs explicitly stating that, though.

amadeuszsz: > I read the documentation you shared, but I am not finding info on thread safety.

I probably overinterpreted the lazy loading behavior and wrongly assumed that if there is no mention of thread safety, it is probably not safe. I also couldn't reproduce the error with the generated unit test. Sorry for the trouble! :pepe-shy:

PR #11001, opened 2025-07-16
feat(crosswalk): improve robustness to prevent stop decision from being canceled by transient noise

satoshi-ota:
## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


satoshi-ota:
Alternative
https://github.com/autowarefoundation/autoware_universe/pull/11027
https://github.com/autowarefoundation/autoware_universe/pull/11028

PR #11020, opened 2025-07-18
feat(autoware_system_monitor): publish cpu_temperature

iwatake2222:
## Description

- This PR introduces a feature to publish CPU temperature data 

## Related links

**Private Links:**

## How was this PR tested?

I confirmed `/system/system_monitor/cpu_monitor/cpu_temperature` topic shows the measured data

Screenshot from 2025-07-18 19-33-16

## Notes for reviewers

- ~This feature currently supports only Intel CPUs and Raspberry Pi.~ On platforms other than Intel CPUs and Raspberry Pi, thermal throttling detection is not implemented, and THERMAL_THROTTLING_OFF will be reported.
- ~I tried adding const to `getThermalThrottlingStatus()`, but it caused a compilation error due to the use of std::mutex, so I left it non-const.~

#### Additions and removals

| Added | Pub | `/system/system_monitor/cpu_monitor/cpu_temperature` | `tier4_external_api_msgs/msg/CpuTemperature`   | [CPU temperature](https://github.com/iwatake2222/tier4_autoware_msgs/blob/feat_external_api_system_monitor/tier4_external_api_msgs/msg/CpuTemperature.msg) |

## Effects on system behavior

None.

### DIFF
--- a/system/autoware_system_monitor/include/system_monitor/cpu_monitor/cpu_monitor_base.hpp
+++ b/system/autoware_system_monitor/include/system_monitor/cpu_monitor/cpu_monitor_base.hpp
@@ -28,6 +28,7 @@
 
 #include <tier4_external_api_msgs/msg/cpu_status.hpp>
 #include <tier4_external_api_msgs/msg/cpu_usage.hpp>
+#include <tier4_external_api_msgs/msg/cpu_temperature.hpp>
 
 #include <atomic>
 #include <climits>
@@ -148,6 +149,13 @@ protected:
     diagnostic_updater::DiagnosticStatusWrapper & stat);  // NOLINT(runtime/references)
 
   /**
+   * @brief Get CPU thermal throttling status
+   * @return DiagStatus::OK or WARN or ERROR
+   * @note ERROR indicates that thermal throttling is active or status retrieval failed
+   */
+  virtual int getThermalThrottlingStatus();
+
+  /**
    * @brief timer callback to collect cpu statistics
    */
   void onTimer();
@@ -157,6 +165,11 @@ protected:
    * @param [in] usage CPU usage
    */
   virtual void publishCpuUsage(tier4_external_api_msgs::msg::CpuUsage usage);
+
+  /**
+   * @brief publish CPU temperature as an independent topic
+   */
+  void publishCpuTemperature();
 
   // Updater won't be changed after initialization. No need to protect it with mutex.
   diagnostic_updater::Updater updater_;  //!< @brief Updater class which advertises to /diagnostics
@@ -213,6 +226,7 @@ protected:
 
   // Publisher
   rclcpp::Publisher<tier4_external_api_msgs::msg::CpuUsage>::SharedPtr pub_cpu_usage_;
+  rclcpp::Publisher<tier4_external_api_msgs::msg::CpuTemperature>::SharedPtr pub_cpu_temperature_;
 
   rclcpp::TimerBase::SharedPtr timer_;  //!< @brief timer to collect cpu statistics
   rclcpp::CallbackGroup::SharedPtr timer_callback_group_;  //!< @brief Callback Group
--- a/system/autoware_system_monitor/include/system_monitor/cpu_monitor/cpu_monitor_base.hpp
+++ b/system/autoware_system_monitor/include/system_monitor/cpu_monitor/cpu_monitor_base.hpp
@@ -153,7 +153,7 @@ protected:
    * @return DiagStatus::OK or WARN or ERROR
    * @note ERROR indicates that thermal throttling is active or status retrieval failed
    */
-  virtual int getThermalThrottlingStatus();
+  virtual int getThermalThrottlingStatus() const;
 
   /**
    * @brief timer callback to collect cpu statistics
@@ -231,11 +231,11 @@ protected:
   rclcpp::TimerBase::SharedPtr timer_;  //!< @brief timer to collect cpu statistics
   rclcpp::CallbackGroup::SharedPtr timer_callback_group_;  //!< @brief Callback Group
 
-  std::mutex mutex_snapshot_;         //!< @brief mutex for protecting snapshot
-  TemperatureData temperature_data_;  //!< @brief snapshot of CPU temperature
-  UsageData usage_data_;              //!< @brief snapshot of CPU usage
-  LoadData load_data_;                //!< @brief snapshot of CPU load average
-  FrequencyData frequency_data_;      //!< @brief snapshot of CPU frequency
+  mutable std::mutex mutex_snapshot_;  //!< @brief mutex for protecting snapshot
+  TemperatureData temperature_data_;   //!< @brief snapshot of CPU temperature
+  UsageData usage_data_;               //!< @brief snapshot of CPU usage
+  LoadData load_data_;                 //!< @brief snapshot of CPU load average
+  FrequencyData frequency_data_;       //!< @brief snapshot of CPU frequency
 
 private:
   /**
--- a/system/autoware_system_monitor/include/system_monitor/cpu_monitor/cpu_monitor_base.hpp
+++ b/system/autoware_system_monitor/include/system_monitor/cpu_monitor/cpu_monitor_base.hpp
@@ -27,8 +27,8 @@
 #include <rclcpp/rclcpp.hpp>
 
 #include <tier4_external_api_msgs/msg/cpu_status.hpp>
-#include <tier4_external_api_msgs/msg/cpu_usage.hpp>
 #include <tier4_external_api_msgs/msg/cpu_temperature.hpp>
+#include <tier4_external_api_msgs/msg/cpu_usage.hpp>
 
 #include <atomic>
 #include <climits>
--- a/system/autoware_system_monitor/src/cpu_monitor/cpu_monitor_base.cpp
+++ b/system/autoware_system_monitor/src/cpu_monitor/cpu_monitor_base.cpp
@@ -110,6 +110,9 @@ CPUMonitorBase::CPUMonitorBase(const std::string & node_name, const rclcpp::Node
   pub_cpu_usage_ =
     this->create_publisher<tier4_external_api_msgs::msg::CpuUsage>("~/cpu_usage", durable_qos);
 
+  pub_cpu_temperature_ =
+    this->create_publisher<tier4_external_api_msgs::msg::CpuTemperature>("~/cpu_temperature", durable_qos);
+
   using namespace std::literals::chrono_literals;
   // Start timer for collecting cpu statistics
   timer_callback_group_ = this->create_callback_group(rclcpp::CallbackGroupType::MutuallyExclusive);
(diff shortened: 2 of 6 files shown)


iwatake2222:
@nishikawa-masaki 
Could you review this PR?

system/autoware_system_monitor/include/system_monitor/cpu_monitor/cpu_monitor_base.hpp
```diff
@@ -147,6 +148,13 @@ class CPUMonitorBase : public rclcpp::Node
   virtual void updateThermalThrottlingImpl(
     diagnostic_updater::DiagnosticStatusWrapper & stat);  // NOLINT(runtime/references)
 
+  /**
+   * @brief Get CPU thermal throttling status
+   * @return DiagStatus::OK or WARN or ERROR
+   * @note ERROR indicates that thermal throttling is active or status retrieval failed
+   */
+  virtual int getThermalThrottlingStatus();
```
nishikawa-masaki: To make getThermalThrottlingStatus()  a "const" member method,
please use the following declaration.
```
mutable std::mutex mutex_snapshot_;         //!< @brief mutex for protecting snapshot
```
This is conformant to the original purpose of introducing the "mutable" keyword to C++ standards.
[Related article (in Japanese)](https://jp-seemore.com/iot/25650/)

system/autoware_system_monitor/src/cpu_monitor/cpu_monitor_base.cpp
```diff
@@ -585,11 +595,32 @@ void CPUMonitorBase::publishCpuUsage(tier4_external_api_msgs::msg::CpuUsage usag
...
+
 void CPUMonitorBase::onTimer()
 {
   checkTemperature();
   checkUsage();
   checkLoad();
   checkFrequency();
   checkThermalThrottling();
+
+  publishCpuTemperature();
```
nishikawa-masaki, line 635: What is the reason why CPUMonitorBase::publishCpuTemperature() is called in CPUMonitorBase::onTimer() in the timer-callback context,
while CPUMonitorBase::publishCpuUsage() is called in CPUMonitorBase::updateUsage() in the publishing context.

I just want to know the reason.
CPUMonitorBase::publishCpuUsage() is called in CPUMonitorBase::updateUsage() in the publishing context.

nishikawa-masaki:
> Notes for reviewers
> This feature currently supports only Intel CPUs and Raspberry Pi.

Autoware_system_monitor is not tested on the latest Raspberry Pi (yet), though it had been tested when it was implemented around 2002.
Changes of operating system, ROS and hardware may have affected its operation.

As for NVIDIA's Tegra family, CPU temperature is measured on Jetson AGX Orin platform (verified by tests), but thermal throttling status is not observed (not implemented yet).

As for generic ARM64 architecture, neither CPU temperature nor thermal throttling is observed.

system/autoware_system_monitor/src/cpu_monitor/cpu_monitor_base.cpp
```diff
@@ -585,11 +595,32 @@ void CPUMonitorBase::publishCpuUsage(tier4_external_api_msgs::msg::CpuUsage usag
...
+  cpu_temperature.stamp = this->now();
+  cpu_temperature.hostname = hostname_;
+  {
+    std::lock_guard<std::mutex> lock_snapshot(mutex_snapshot_);
+    if (!temperature_data_.core_data.empty()) {
+      cpu_temperature.temperature = temperature_data_.core_data[0].temperature;
+    }
+  }
+  int thermal_throttling = getThermalThrottlingStatus();
+  cpu_temperature.thermal_throttling = thermal_throttling == DiagStatus::OK
```
nishikawa-masaki: I recommend putting the condition "thermal_throttling == DiagStatus::OK" into a pair of parentheses for helping readers.

system/autoware_system_monitor/src/cpu_monitor/cpu_monitor_base.cpp
```diff
@@ -585,11 +595,32 @@ void CPUMonitorBase::publishCpuUsage(tier4_external_api_msgs::msg::CpuUsage usag
...
+void CPUMonitorBase::publishCpuTemperature()
+{
+  using tier4_external_api_msgs::msg::CpuTemperature;
+  CpuTemperature cpu_temperature;
+  cpu_temperature.stamp = this->now();
+  cpu_temperature.hostname = hostname_;
+  {
+    std::lock_guard<std::mutex> lock_snapshot(mutex_snapshot_);
+    if (!temperature_data_.core_data.empty()) {
+      cpu_temperature.temperature = temperature_data_.core_data[0].temperature;
```
nishikawa-masaki, line 617: On Intel/AMD platforms, temperature of each CPU core is available.
Is reporting only the first core intentional (to make things simpler among different platforms)?
I just wanted to make it sure.

system/autoware_system_monitor/src/cpu_monitor/cpu_monitor_base.cpp
```diff
@@ -585,11 +595,32 @@ void CPUMonitorBase::publishCpuUsage(tier4_external_api_msgs::msg::CpuUsage usag
   pub_cpu_usage_->publish(usage);
 }
 
+void CPUMonitorBase::publishCpuTemperature()
```
nishikawa-masaki, line 608: If this feature is available only on Intel/AMD x86_64 architecture and Raspberry Pi, publishing CpuTemperature every second on ARM and NVIDIA Tegra platforms are meaningless and can be harmful.

system/autoware_system_monitor/src/cpu_monitor/cpu_monitor_base.cpp
nishikawa-masaki, line 617: If temperature_data_.core_data.empty() is true, no value is assigned to cpu_temperature.temperature.
Is it OK?

nishikawa-masaki (changes requested):
Sorry for late reviewing.
Please think about the commented issues.

system/autoware_system_monitor/src/cpu_monitor/cpu_monitor_base.cpp
iwatake2222: @nishikawa-masaki 
In other `ooo_monitor`s that have a timer_callback,  the publish function is called from within the timer callback. Therefore, I followed that approach and called `publishCpuTemperature()` from onTimer(). Since the processing is not directly related to the diagnostics "update", calling it from `onTimer()` felt more natural to me. Also, cpu_temperature holds both the temperature and thermal throttling information. For that reason, I feel that calling it from either of `updateTemperature()` or `updateThermalThrottling()` functions is not appropriate.

I believe it would be better to call `publishCpuUsage` from `onTimer()` as well, but we should handle the change  as a separate PR for refactoring.

system/autoware_system_monitor/src/cpu_monitor/cpu_monitor_base.cpp
iwatake2222: @nishikawa-masaki 
Yes, reporting only the first core's temperature is  mainly for simplicity and consistency across different platforms.
In most practical use cases, the first core’s temperature (often labeled as "Package id 0" or similar) is representative enough of the overall CPU thermal status.
That said, if needed, we can consider extending it to report per-core temperatures in the future. In that case, we need to modify the message definition, though.

system/autoware_system_monitor/include/system_monitor/cpu_monitor/cpu_monitor_base.hpp
iwatake2222: @nishikawa-masaki 
Thanks a lot! Good to know that technique.
Fixed in 7ae7e40602b508793b618e16827ff16424f01892

system/autoware_system_monitor/src/cpu_monitor/cpu_monitor_base.cpp
iwatake2222: @nishikawa-masaki 
Fixed in 7ae7e40602b508793b618e16827ff16424f01892

system/autoware_system_monitor/src/cpu_monitor/cpu_monitor_base.cpp
iwatake2222: @nishikawa-masaki 
(I made a separate comment earlier, but I’ve made the necessary corrections.
Also, I encountered some errors in the CI, so I performed a rebase and force-pushed the branch.
As a result, some unnecessary messages or notifications may have been triggered — I apologize for the inconvenience.)


I have updated the description in this PR.

Old:
This feature currently supports only Intel CPUs and Raspberry Pi.

New:
On platforms other than Intel CPUs and Raspberry Pi, thermal throttling detection is not implemented, and `THERMAL_THROTTLING_OFF` will be reported.

Regarding temperature monitoring, it can technically be retrieved on Tegra and other ARM-based CPUs as well if running on a Linux system.
On the other hand, thermal throttling detection is only supported on Intel CPUs and Raspberry Pi. However, since the reported result will always be `THERMAL_THROTTLING_OFF` on unsupported platforms, I believe this will not cause any negative side effects.

system/autoware_system_monitor/src/cpu_monitor/cpu_monitor_base.cpp
iwatake2222: @nishikawa-masaki 
Temperature monitoring is supported on all platforms including Tegra and other ARM CPUs, so the value is unlikely to be empty. Additionally,  `cpu_temperature.temperature` is always set to 0 by the constructor generated by ROS2 message.
I probably should have had a field to explicitly indicate data validity. I’ll consider adding that as a future improvement.

system/autoware_system_monitor/src/cpu_monitor/cpu_monitor_base.cpp
iwatake2222: @nishikawa-masaki 
Fixed in 2fc28fb6ba3ec86a3f49d605deb27545c7028115

system/autoware_system_monitor/include/system_monitor/cpu_monitor/cpu_monitor_base.hpp
iwatake2222: @nishikawa-masaki 
Thanks a lot! Good to know that technique.
Fixed in 2fc28fb6ba3ec86a3f49d605deb27545c7028115

iwatake2222:
@nishikawa-masaki 
Could you please re-review this PR and check my comments?
(I encountered some errors in the CI, so I performed a rebase and force-pushed the branch.)

system/autoware_system_monitor/src/cpu_monitor/cpu_monitor_base.cpp
iwatake2222: > Regarding temperature monitoring, it can technically be retrieved on Tegra and other ARM-based CPUs as well if running on a Linux system.

I noticed your comment "As for generic ARM64 architecture, neither CPU temperature nor thermal throttling is observed." , but temperature monitoring for ARM CPUs has already been implemented, so the lack of temperature readings is an existing issue from before this PR. Therefore, I’d like to treat this issue separately from the current PR.

nishikawa-masaki: > I noticed your comment "As for generic ARM64 architecture, neither CPU temperature nor thermal throttling is observed." , but temperature monitoring for ARM CPUs has already been implemented, so the lack of temperature readings is an existing issue from before this PR. Therefore, I’d like to treat this issue separately from the current PR.

Yes, temperature monitoring for ARM64 CPUs has been implemented, but does not work (at least in my testing).
It may require additional package for operation.
According to one of the original implementors, arm_cpu_monitor.cpp was implemented as a reference code for generic ARM64 architectures which are not NVIDIA Tegra or Raspberry Pi. It has never been tested.
Therefore, it is left unmaintained. We will need to work on it in near future.

You can resolve this thread when you read this explanation.

nishikawa-masaki (approved):
It looks good to me after your last fix.

PR #11021, opened 2025-07-18
feat(autoware_system_monitor): publish memory_status

iwatake2222:
## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/system/autoware_system_monitor/src/mem_monitor/mem_monitor.cpp
+++ b/system/autoware_system_monitor/src/mem_monitor/mem_monitor.cpp
@@ -44,6 +44,12 @@ MemMonitor::MemMonitor(const rclcpp::NodeOptions & options)
   if (!bp::search_path("edac-util").empty()) {
     updater_.add("Memory ECC", this, &MemMonitor::checkEcc);
   }
+
+  // Publisher
+  rclcpp::QoS durable_qos{1};
+  durable_qos.transient_local();
+  pub_memory_status_ =
+    this->create_publisher<tier4_external_api_msgs::msg::MemoryStatus>("~/memory_status", durable_qos);
 }
 
 void MemMonitor::update()
@@ -91,7 +97,7 @@ void MemMonitor::checkUsage(diagnostic_updater::DiagnosticStatusWrapper & stat)
   std::string line;
   int index = 0;
   std::vector<std::string> list;
-  float usage;
+  float usage = 0.0f;
   size_t mem_total = 0;
   size_t mem_shared = 0;
   size_t mem_available = 0;
@@ -157,6 +163,8 @@ void MemMonitor::checkUsage(diagnostic_updater::DiagnosticStatusWrapper & stat)
   }
 
   stat.summary(level, usage_dict_.at(level));
+
+  publishMemoryStatus(usage);
 
   // Measure elapsed time since start time and report
   SystemMonitorUtility::stopMeasurement(t_start, stat);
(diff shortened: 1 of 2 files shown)


iwatake2222:
@nishikawa-masaki 
Could you review this PR?

system/autoware_system_monitor/src/mem_monitor/mem_monitor.cpp
```diff
@@ -158,6 +164,8 @@ void MemMonitor::checkUsage(diagnostic_updater::DiagnosticStatusWrapper & stat)
 
   stat.summary(level, usage_dict_.at(level));
 
+  publishMemoryStatus(usage);
```
nishikawa-masaki, line 167: Please make it sure that adding "publishMemoryStatus()" here does not add extra execution time to "checkUsage()".
(I believe it doesn't, but I want to make it sure.)
Line 170 SystemMonitorUtility::stopMeasurement() measures the execution time and the time is reported in "/diagnostics" topic.

nishikawa-masaki (approved):
Other than the execution time issue I commented, it looks good to me.

system/autoware_system_monitor/src/mem_monitor/mem_monitor.cpp
```diff
@@ -44,6 +44,12 @@ MemMonitor::MemMonitor(const rclcpp::NodeOptions & options)
   if (!bp::search_path("edac-util").empty()) {
     updater_.add("Memory ECC", this, &MemMonitor::checkEcc);
   }
+
+  // Publisher
+  rclcpp::QoS durable_qos{1};
+  durable_qos.transient_local();
+  pub_memory_status_ = this->create_publisher<tier4_external_api_msgs::msg::MemoryStatus>(
+    "~/memory_status", durable_qos);
```
nishikawa-masaki, line 52: According to ROS2 documents, create_publisher() throws an exception on errors.
That means we need to catch the exception and handle it accordingly, but I've never seen such handling in Autoware.
I need to investigate about it.

nishikawa-masaki: I had Cursor search for exception handling with create_publisher(), but there is no such code block.
In this particular call of create_publisher(), I believe there is no need to worry about exceptions because the parameters are static and valid.
If the parameters were dynamic (and may be invalid), we would need to handle exceptions.

system/autoware_system_monitor/src/mem_monitor/mem_monitor.cpp
iwatake2222: @nishikawa-masaki 

I compared the execution time before and after the changes in this PR. The average execution time before the change (original) was 20.1 ms, while after the change (This PR) it was 18.7 ms — in fact, it was slightly faster after the change. As shown in the graph, the main cause of variation is the randomly occurring spikes, which are likely due to running the Autoware on a regular PC. Excluding those spikes, there doesn't appear to be a significant difference between before and after the change.

image

system/autoware_system_monitor/src/mem_monitor/mem_monitor.cpp
iwatake2222: Thank you for the comment and investigation. I'm relieved to hear that it doesn't have an impact in this case. It seems that we need to be more careful when dealing with dynamic parameters — I learned something new from this.

PR #11022, opened 2025-07-18
feat(autoware_system_monitor): publish gpu_status

iwatake2222:
## Description

- This PR introduces a feature to publish GPU status

## Related links

**Private Links:**

## How was this PR tested?

I confirmed `/system/system_monitor/gpu_monitor/gpu_status` topic shows the measured data

Screenshot from 2025-07-18 19-34-27

## Notes for reviewers

This feature currently supports only NVIDIA GPUs which support NVML library.

#### Additions and removals

| Added | Pub | `/system/system_monitor/gpu_monitor/gpu_status` | `tier4_external_api_msgs/msg/GpuStatus`   | [GPU status](https://github.com/iwatake2222/tier4_autoware_msgs/blob/feat_external_api_system_monitor/tier4_external_api_msgs/msg/GpuStatus.msg) |

## Effects on system behavior

None.


iwatake2222:
@nishikawa-masaki 
Could you review this PR?

system/autoware_system_monitor/include/system_monitor/gpu_monitor/gpu_monitor_base.hpp
```diff
@@ -96,8 +109,29 @@ class GPUMonitorBase : public rclcpp::Node
   virtual void checkFrequency(
     diagnostic_updater::DiagnosticStatusWrapper & stat);  // NOLINT(runtime/references)
 
+  /**
+   * @brief get GPU status
+   * @return GPU status list
+   */
+  virtual std::vector<GpuStatus> getGPUStatus();
```
nishikawa-masaki: Is it possible to declare that getGPUStatus() is a const method?
Generally, "get" metchods are expected to be const methods.

system/autoware_system_monitor/include/system_monitor/gpu_monitor/gpu_monitor_base.hpp
```diff
@@ -44,6 +48,15 @@ class GPUMonitorBase : public rclcpp::Node
 protected:
   using DiagStatus = diagnostic_msgs::msg::DiagnosticStatus;
 
+  struct GpuStatus
+  {
+    std::string name = "";
+    float usage = 0.0f;
+    unsigned int clock = 0;
+    unsigned int temperature = 0;
```
nishikawa-masaki: "Unsigned" should be used only for bit patterns.
"Not expected to be negative" can't be the reason to use "unsigned".
At least, temperature can be negative, theoretically.
Autoware coding guideline is based on ROS2's guideline.
ROS2 guideline is based on Google's.
[Google C++ Style Guide : Integer Types](https://google.github.io/styleguide/cppguide.html#Integer_Types)

nishikawa-masaki (commented):
Though I'm still reviewing this PR, let me submit my comments at the end of the day.

nishikawa-masaki:
I'd like to ask 2 questions before continuing review of this PR.

Q1.
This PR is only for NVIDIA GPUs with NVML library support and it doesn't support Tegra family GPUs.
For example, Connect Tech (CTI) ECU Anvil is based on NVIDIA Jetson AGX Orin, which belongs to Tegra family.
Is it OK to leave Anvil without this change?

Q2.
This PR employs newly defined onTimer() timer callback for collecting GPU information and publishing "/system/system_monitor/gpu_monitor/gpu_status" topic.

This design has 2 problems. What is your view about them?
* NVML libraries are called twice (by the original publishing thread and the timer thread).
  * According to NVIDIA NVML documents, most of NVML functions (except nvmlInit() and nvmlShutdown()) are thread-safe.
  * There won't be so much performance penalty as long as they are called a few times per second, according to the document, but that doesn't justify calling the same functions twice.
* The design doesn't match the current implementation of process_monitor and cpu_monitor. They were modified to improve their performance with the following implementation.
  * Timer-thread callback "onTimer()" corrects information. Some information requires considerable amount of time to get them.
  * The publishing thread publishes the information collected by "onTimer()". Publishing itself is light.
  * This means that two threads share a group of data and protect the data with a mutex.

By the way, I found that NvmlGpuMonitor and GpuMonitorBase don't have explicit destructors.
As far as I understand, they should have explicit destructors and join the publishing thread and the timer thread in them, then call shut_down() to execute nvmlShutdown(), but they don't.
I'm afraid that nvmlShutdown() is not called at all.
This isn't your fault, but somebody need to work on it.

iwatake2222:
@nishikawa-masaki 

> Q1.
This PR is only for NVIDIA GPUs with NVML library support and it doesn't support Tegra family GPUs.
For example, Connect Tech (CTI) ECU Anvil is based on NVIDIA Jetson AGX Orin, which belongs to Tegra family.
Is it OK to leave Anvil without this change?

For this PR, I would like to limit the scope to NVIDIA GPUs that support the NVML library only.
As for Tegra-based devices, I plan to address them separately when I have access to the appropriate hardware and time for implementation and testing. ([TIER IV INTERNAL LINK](https://tier4.atlassian.net/browse/RT0-38673))


> Q2. This PR employs newly defined onTimer() timer callback for collecting GPU information and publishing "/system/system_monitor/gpu_monitor/gpu_status" topic.
> This design has 2 problems. What is your view about them?
> NVML libraries are called twice (by the original publishing thread and the timer thread).

Since these are get-type APIs, I considered it acceptable to call them multiple times without significant issues.
I also considered caching the values obtained inside the `checkXXX` functions, but was concerned about potential discrepancies in the timing of data acquisition—for example, time gap between temperature and clock frequency readings. (That said, in reality, the `checkXXX` functions are called in quick succession by the Updater, so timing differences are unlikely to cause practical issues.)

This issue likely stems from the fact that, as mentioned later, `gpu_monitor` performs both data collection and updates at the timing of each `checkXXX` function called by updater_.


> The design doesn't match the current implementation of process_monitor and cpu_monitor. They were modified to improve their performance with the following implementation. 

The monitor implementations can be categorized as follows.
Here, "API publish" refers to the functionality newly introduced in this PR.

- `cpu_monitor` [PR #10545](https://github.com/autowarefoundation/autoware_universe/pull/10545)
    - timer: Data collection and API publish
    - updater_: Update diagnostics using pre-collected data

- `net_monitor`
    - timer: Data collection and API publish
    - updater_: Update diagnostics using pre-collected data

- `hdd_monitor`
    - timer: Data collection and API publish
    - updater_:
        - Except usage: Update diagnostics using pre-collected data
        - For usage: Data collection and Update diagnostics   <<<

- `gpu_monitor`
    - timer (new): API publish
    - updater_: Data collection and Update diagnostics   <<<

- `mem_monitor`
    - timer: Not used
    - updater_: Data collection and Update diagnostics. API publish    <<<

Note that not only `gpu_monitor` but also `mem_monitor` and parts of `hdd_monitor` differ from other monitors in structure.
While this discrepancy may not have a significant impact on performance, standardizing the structure across all monitors may be beneficial as a refactoring step for better maintainability. In addition, using updater_ alone makes it difficult to perform custom processing on the full set of collected data. One such example is the API publish feature added in this PR. For this reason, I added a separate timer in `gpu_monitor`. As for `mem_monitor`, I did not introduce a separate timer since, at this point, there is no data that needs to be processed in an integrated manner.

I think aligning the structure of parts of `hdd_monitor`, as well as `gpu_monitor` and `mem_monitor`, with that of `cpu_monitor` makes sense, but I would prefer to treat it as a low-priority refactoring task in a separate PR.

> By the way, I found that NvmlGpuMonitor and GpuMonitorBase don't have explicit destructors.

~If the changes are minor, I plan to handle them in this PR.~
Fixed in 9c6fe28da67c9eb4f5e084fc927623bbd4cffe48
I confirmed locally by adding logs that both `GPUMonitorBase::shut_down` and `GPUMonitor::shut_down()` are called.

system/autoware_system_monitor/include/system_monitor/gpu_monitor/gpu_monitor_base.hpp
iwatake2222: Fixed in a07ca63120de2a34fdc4ebfbbd8bf8c5ddf44c5e

system/autoware_system_monitor/include/system_monitor/gpu_monitor/gpu_monitor_base.hpp
iwatake2222: Fixed in a07ca63120de2a34fdc4ebfbbd8bf8c5ddf44c5e

system/autoware_system_monitor/src/gpu_monitor/nvml_gpu_monitor.cpp
```diff
@@ -447,6 +452,66 @@ void GPUMonitor::checkFrequency(diagnostic_updater::DiagnosticStatusWrapper & st
...
+          thermal_throttling = DiagStatus::ERROR;
+          break;
+      }
+    }
+
+    GpuStatus gpu_status;
+    gpu_status.name = itr->name;
+    gpu_status.usage = utilization.gpu;
+    gpu_status.clock = clock;
+    gpu_status.temperature = temp;
```
nishikawa-masaki: Please make type conversion explicitly.
`gpu_status.usage = static_cast<float>(utilization.gpu);  // casting unsigned int to float`

The following two lines should show casting explicitly.
These casting are caused by contradiction beween NVIDIA API and Google's coding style.
```
    gpu_status.clock = static_cast<int>(clock);  // casting unsigned int to int
    gpu_status.temperature = static_cast<int>(temp);  // casting unsigned int to int
```

iwatake2222: @nishikawa-masaki 
Thank you for the suggestion. I fixed it in 7b83805d47c7bbfce8475acda467100f0137ac86
Could you review it again?

system/autoware_system_monitor/src/gpu_monitor/nvml_gpu_monitor.cpp
```diff
@@ -78,6 +78,11 @@ GPUMonitor::GPUMonitor(const rclcpp::NodeOptions & options) : GPUMonitorBase("gp
   }
 }
 
+GPUMonitor::~GPUMonitor()
+{
+  shut_down();
```
nishikawa-masaki, line 83: > Fixed in https://github.com/autowarefoundation/autoware_universe/commit/9c6fe28da67c9eb4f5e084fc927623bbd4cffe48
I confirmed locally by adding logs that both GPUMonitorBase::shut_down and GPUMonitor::shut_down() are called.

Thank you for the fix.
The shutdown sequence is as the followings and it looks OK to me.
1. Autoware is requested to terminate.
2. The executor stops spinning and the callbacks (including timer-callbacks and updater-callbacks) won't be called after that. So, nvmlXXX() API won't be called anymore.
3. Node objects start being destructed.
4. GPUMonitor::~GPUMonitor() is called (nvml_gpu_monitor.cpp version).
5. GPUMonitor::shut_down() is called and nvmlShutdown() is called.
6. GPUMonitorBase::~GPUMonitorBase() is called.
7. GPUMonitorBase::shut_down() is called and it does nothing.

I still have the following concerns.
* GPUMonitorBase::shut_down() is declared as a public method. I think we need to declare it as a protected method so that it can't be called from other classes.
  * "shut_down" iss not inherited from rcpcpp::Node class.
  * The comment for GPUMonitorBase::shut_down() seems to tell that it is intended for internal use, but I can't be sure enough.
    * "Terminate the node, log final statements."
* There is no guarantee that GPUMonitor (nvml_gpu_monitor.cpp) is the only user of NVML library.
  * I have no good idea to guarantee that.
  * For now, we don't need to work on this issue because there is no real harm.
    * This can be a problem if somebody uses NVML library somewhere else or if there are multiple instances of GPUMonitor (nvml_gpu_monitor.cpp) class.

nishikawa-masaki (approved):
It looks good to me.
Please declare "shut_down()" as a protected method and there is no problem in building Autoware.

system/autoware_system_monitor/src/gpu_monitor/nvml_gpu_monitor.cpp
iwatake2222: I made `shut_down()` method protected in https://github.com/autowarefoundation/autoware_universe/pull/11022/commits/2dcca7b5a96fa346f53fc5b8e8da3041031e2509

As you mentioned, the shutdown sequence is the following (I just locally added the log to confirm).

```
^C[WARNING] [launch]: user interrupted with ctrl-c (SIGINT)
[component_container_mt-1] [INFO 1754546789.420718420] [rclcpp]: signal_handler(signum=2)
[component_container_mt-1] [ERROR 1754546789.443904197] [gpu_monitor]: GPUMonitor::~GPUMonitor()
[component_container_mt-1] [ERROR 1754546789.443922952] [gpu_monitor]: GPUMonitor::shut_down()
[component_container_mt-1] [ERROR 1754546789.455159581] [gpu_monitor]: GPUMonitorBase::~GPUMonitorBase()
[component_container_mt-1] [ERROR 1754546789.455172430] [gpu_monitor]: GPUMonitorBase::shut_down()
[INFO] [component_container_mt-1]: process has finished cleanly [pid 584664]
```

PR #11023, opened 2025-07-18
feat(autoware_system_monitor): publish hdd_status

iwatake2222:
## Description

- This PR introduces a feature to publish HDD status

## Related links

**Private Links:**

## How was this PR tested?

I confirmed `/system/system_monitor/hdd_monitor/hdd_status` topic shows the measured data

Screenshot from 2025-07-18 19-35-00

## Notes for reviewers

None.

#### Additions and removals

| Added | Pub | `/system/system_monitor/hdd_monitor/hdd_status` | `tier4_external_api_msgs/msg/HddStatus`   | [HDD status](https://github.com/iwatake2222/tier4_autoware_msgs/blob/feat_external_api_system_monitor/tier4_external_api_msgs/msg/HddStatus.msg) |

## Effects on system behavior

None.

### DIFF
--- a/system/autoware_system_monitor/include/system_monitor/hdd_monitor/hdd_monitor.hpp
+++ b/system/autoware_system_monitor/include/system_monitor/hdd_monitor/hdd_monitor.hpp
@@ -24,6 +24,10 @@
 
 #include <diagnostic_updater/diagnostic_updater.hpp>
 
+#include <tier4_external_api_msgs/msg/hdd_device_status.hpp>
+#include <tier4_external_api_msgs/msg/hdd_partition_status.hpp>
+#include <tier4_external_api_msgs/msg/hdd_status.hpp>
+
 #include <climits>
 #include <map>
 #include <string>
@@ -89,7 +93,7 @@ struct SysfsDevStat
 };
 
 /**
- * @brief statistics of HDD
+ * @brief statistics of HDD for each physical disk
  */
 struct HddStat
 {
@@ -102,6 +106,23 @@ struct HddStat
   SysfsDevStat last_sysfs_dev_stat_;  //!< @brief last statistics of sysfs device
 
   HddStat() : read_data_rate_MBs_(0.0), write_data_rate_MBs_(0.0), read_iops_(0.0), write_iops_(0.0)
+  {
+  }
+};
+
+/**
+ * @brief usage of HDD for each partition
+ */
+struct HddUsage
+{
+  int size;                //!< @brief Total size of the filesystem in megabytes (MB)
+  int used;                //!< @brief Used space in the filesystem in megabytes (MB)
+  int avail;               //!< @brief Available (free) space in the filesystem in megabytes (MB)
+  int capacity;            //!< @brief Percentage of used space in the filesystem (0–100)
+  std::string filesystem;  //!< @brief Name of the filesystem device (e.g., "/dev/nvme0n1p2")
+  std::string mounted_on;  //!< @brief Mount point of the filesystem (e.g., "/")
+
+  HddUsage() : size(0), used(0), avail(0), capacity(0)
   {
   }
 };
@@ -321,6 +342,11 @@ protected:
    */
   int unmountDevice(std::string & device);
 
+  /**
+   * @brief publish HDD status
+   */
+  void publishHddStatus();
+
   diagnostic_updater::Updater updater_;  //!< @brief Updater class which advertises to /diagnostics
   rclcpp::TimerBase::SharedPtr timer_;   //!< @brief timer to get HDD information from HddReader
 
@@ -337,6 +363,10 @@ protected:
   diagnostic_updater::DiagnosticStatusWrapper connect_diag_;
   HddInfoList hdd_info_list_;               //!< @brief list of HDD information
   rclcpp::Time last_hdd_stat_update_time_;  //!< @brief last HDD statistics update time
+
+  std::vector<HddUsage> hdd_usages_;  //!< @brief list of HDD usage
+
+  rclcpp::Publisher<tier4_external_api_msgs::msg::HddStatus>::SharedPtr pub_hdd_status_;  //!< @brief publisher
 
   /**
    * @brief HDD SMART status messages
--- a/system/autoware_system_monitor/include/system_monitor/hdd_monitor/hdd_monitor.hpp
+++ b/system/autoware_system_monitor/include/system_monitor/hdd_monitor/hdd_monitor.hpp
@@ -29,6 +29,7 @@
 #include <tier4_external_api_msgs/msg/hdd_status.hpp>
 
 #include <climits>
+#include <cstdint>
 #include <map>
 #include <string>
 #include <vector>
--- a/system/autoware_system_monitor/include/system_monitor/hdd_monitor/hdd_monitor.hpp
+++ b/system/autoware_system_monitor/include/system_monitor/hdd_monitor/hdd_monitor.hpp
@@ -122,9 +122,7 @@ struct HddUsage
   std::string filesystem;  //!< @brief Name of the filesystem device (e.g., "/dev/nvme0n1p2")
   std::string mounted_on;  //!< @brief Mount point of the filesystem (e.g., "/")
 
-  HddUsage() : size(0), used(0), avail(0), capacity(0)
-  {
-  }
+  HddUsage() : size(0), used(0), avail(0), capacity(0) {}
 };
 
 /**
@@ -366,7 +364,8 @@ protected:
 
   std::vector<HddUsage> hdd_usages_;  //!< @brief list of HDD usage
 
-  rclcpp::Publisher<tier4_external_api_msgs::msg::HddStatus>::SharedPtr pub_hdd_status_;  //!< @brief publisher
+  rclcpp::Publisher<tier4_external_api_msgs::msg::HddStatus>::SharedPtr
+    pub_hdd_status_;  //!< @brief publisher
 
   /**
    * @brief HDD SMART status messages
--- a/system/autoware_system_monitor/include/system_monitor/hdd_monitor/hdd_monitor.hpp
+++ b/system/autoware_system_monitor/include/system_monitor/hdd_monitor/hdd_monitor.hpp
@@ -113,16 +113,16 @@ struct HddStat
 /**
  * @brief usage of HDD for each partition
  */
-struct HddUsage
+struct HddPartitionStatus
 {
-  int size;                //!< @brief Total size of the filesystem in megabytes (MB)
-  int used;                //!< @brief Used space in the filesystem in megabytes (MB)
-  int avail;               //!< @brief Available (free) space in the filesystem in megabytes (MB)
-  int capacity;            //!< @brief Percentage of used space in the filesystem (0–100)
+  int32_t size;            //!< @brief Total size of the filesystem in megabytes (MB)
+  int32_t used;            //!< @brief Used space in the filesystem in megabytes (MB)
+  int32_t avail;           //!< @brief Available (free) space in the filesystem in megabytes (MB)
+  int32_t capacity;        //!< @brief Percentage of used space in the filesystem (0–100)
   std::string filesystem;  //!< @brief Name of the filesystem device (e.g., "/dev/nvme0n1p2")
   std::string mounted_on;  //!< @brief Mount point of the filesystem (e.g., "/")
 
-  HddUsage() : size(0), used(0), avail(0), capacity(0) {}
+  HddPartitionStatus() : size(0), used(0), avail(0), capacity(0) {}
 };
 
 /**
@@ -362,7 +362,7 @@ protected:
   HddInfoList hdd_info_list_;               //!< @brief list of HDD information
   rclcpp::Time last_hdd_stat_update_time_;  //!< @brief last HDD statistics update time
 
-  std::vector<HddUsage> hdd_usages_;  //!< @brief list of HDD usage
+  std::vector<HddPartitionStatus> hdd_partition_statuses_;  //!< @brief list of partition status
 
   rclcpp::Publisher<tier4_external_api_msgs::msg::HddStatus>::SharedPtr
     pub_hdd_status_;  //!< @brief publisher
--- a/system/autoware_system_monitor/src/hdd_monitor/hdd_monitor.cpp
+++ b/system/autoware_system_monitor/src/hdd_monitor/hdd_monitor.cpp
@@ -73,6 +73,11 @@ HddMonitor::HddMonitor(const rclcpp::NodeOptions & options)
   // start HDD transfer measurement
   startHddTransferMeasurement();
 
+  // Publisher
+  rclcpp::QoS durable_qos{1};
+  durable_qos.transient_local();
+  pub_hdd_status_ = this->create_publisher<tier4_external_api_msgs::msg::HddStatus>("~/hdd_status", durable_qos);
+
   timer_ = rclcpp::create_timer(this, get_clock(), 1s, std::bind(&HddMonitor::onTimer, this));
 }
 
@@ -243,6 +248,8 @@ void HddMonitor::checkUsage(diagnostic_updater::DiagnosticStatusWrapper & stat)
     return;
   }
 
+  hdd_usages_.clear();
+
   int hdd_index = 0;
   int whole_level = DiagStatus::OK;
   std::string error_str = "";
@@ -298,7 +305,10 @@ void HddMonitor::checkUsage(diagnostic_updater::DiagnosticStatusWrapper & stat)
     std::string line;
     int index = 0;
     std::vector<std::string> list;
+    int size;
+    int used;
     int avail;
+    int capacity;
 
     while (std::getline(is_out, line) && !line.empty()) {
       // Skip header
@@ -310,11 +320,21 @@ void HddMonitor::checkUsage(diagnostic_updater::DiagnosticStatusWrapper & stat)
       boost::split(list, line, boost::is_space(), boost::token_compress_on);
 
       try {
-        avail = std::stoi(list[3].c_str());
+        size = std::stoi(list[1]);
+        used = std::stoi(list[2]);
+        avail = std::stoi(list[3]);
+        std::string capacity_str = list[4];
+        if (!capacity_str.empty() && capacity_str.back() == '%') {
+          capacity_str.pop_back();
+        }
+        capacity = std::stoi(capacity_str);
       } catch (std::exception & e) {
+        size = -1;
+        used = -1;
         avail = -1;
+        capacity = -1;
         error_str = e.what();
-        stat.add(fmt::format("HDD {}: status", hdd_index), "avail string error");
+        stat.add(fmt::format("HDD {}: status", hdd_index), "string error");
       }
 
       if (avail <= itr->second.free_error_) {
@@ -339,6 +359,15 @@ void HddMonitor::checkUsage(diagnostic_updater::DiagnosticStatusWrapper & stat)
         }
       }
       stat.add(fmt::format("HDD {}: mounted on", hdd_index), mounted_.c_str());
+
+      HddUsage hdd_usage;
+      hdd_usage.size = size;
+      hdd_usage.used = used;
+      hdd_usage.avail = avail;
+      hdd_usage.capacity = capacity;
+      hdd_usage.filesystem = list[0];
+      hdd_usage.mounted_on = mounted_;
+      hdd_usages_.push_back(hdd_usage);
 
       whole_level = std::max(whole_level, level);
       ++index;
@@ -583,6 +612,8 @@ void HddMonitor::onTimer()
   updateHddConnections();
   updateHddInfoList();
   updateHddStatistics();
+
+  publishHddStatus();
 }
 
 void HddMonitor::updateHddInfoList()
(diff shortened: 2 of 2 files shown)


iwatake2222:
@nishikawa-masaki 
Could you review this PR?

system/autoware_system_monitor/include/system_monitor/hdd_monitor/hdd_monitor.hpp
```diff
@@ -106,6 +110,21 @@ struct HddStat
...
 
+/**
+ * @brief usage of HDD for each partition
+ */
+struct HddUsage
+{
+  int size;                //!< @brief Total size of the filesystem in megabytes (MB)
+  int used;                //!< @brief Used space in the filesystem in megabytes (MB)
+  int avail;               //!< @brief Available (free) space in the filesystem in megabytes (MB)
+  int capacity;            //!< @brief Percentage of used space in the filesystem (0–100)
```
nishikawa-masaki: **This is just an "afterthought" comment. Please ignore it.**

---------------------------------
I found that [feat(tier4_external_api_msgs): add system monitor #187](https://github.com/tier4/tier4_autoware_msgs/pull/187) had already merged and commenting on naming of "already fixed" structure members was meaningless, just after I finished writing the following comments.
I'm leaving the comment just for future reference, but using "int32_t" instead of "int" is still preferable because "tier4_external_api_msgs" uses "int32".

---------------------------------
Though there are discussions among engineers, I prefer "Application Hungarian"-like naming, especially when units and scales matter.
(Please note that there are 2 types of Hungarian Notation. "System Hungarian" is meaningless and harmful in C++, but "Application Hungarian" can prevent possible errors. See [Wikipedia(En)](https://en.wikipedia.org/wiki/Hungarian_notation) and [Wikipedia(Jp)](https://ja.wikipedia.org/wiki/%E3%83%8F%E3%83%B3%E3%82%AC%E3%83%AA%E3%82%A2%E3%83%B3%E8%A8%98%E6%B3%95).

My suggestions are:
```
  int32_t total_mb;              //!< @brief Total size of the filesystem in megabytes (MB)
  int32_t used_mb;               //!< @brief Used space in the filesystem in megabytes (MB)
  int32_t available_mb;          //!< @brief Available (free) space in the filesystem in megabytes (MB)
  int32_t usage_rate;            //!< @brief Percentage of used space in the filesystem (0–100)
```
If more clarity is desirable, "MB" (1MB = 1000 * 1000 byte) and "MiB" (1MiB = 1024 * 1024 byte) need to be distinguished because the difference can produce 5% error.
([Wikipedia(En)](https://en.wikipedia.org/wiki/Byte#Multiple-byte_units) , [Wikipedia(Jp)](https://ja.wikipedia.org/wiki/%E3%83%A1%E3%83%93%E3%83%90%E3%82%A4%E3%83%88))

"avail" is replaced by "available_mb" to avoid unnecessary abbreviation.

system/autoware_system_monitor/include/system_monitor/hdd_monitor/hdd_monitor.hpp
```diff
@@ -106,6 +110,21 @@ struct HddStat
   }
 };
 
+/**
+ * @brief usage of HDD for each partition
+ */
+struct HddUsage
```
nishikawa-masaki: This structure is tightly coupled with "tier4_external_api_msgs/msg/HddPartitionStatus.msg".
How about "HddPartitionStatus"?

system/autoware_system_monitor/src/hdd_monitor/hdd_monitor.cpp
```diff
@@ -340,6 +361,15 @@ void HddMonitor::checkUsage(diagnostic_updater::DiagnosticStatusWrapper & stat)
...
       stat.add(fmt::format("HDD {}: mounted on", hdd_index), mounted_.c_str());
 
+      HddUsage hdd_usage;
+      hdd_usage.size = size;
+      hdd_usage.used = used;
+      hdd_usage.avail = avail;
+      hdd_usage.capacity = capacity;
+      hdd_usage.filesystem = list[0];
+      hdd_usage.mounted_on = mounted_;
+      hdd_usages_.push_back(hdd_usage);
```
nishikawa-masaki: Please consider using emplace_back().

system/autoware_system_monitor/src/hdd_monitor/hdd_monitor.cpp
```diff
@@ -310,11 +321,21 @@ void HddMonitor::checkUsage(diagnostic_updater::DiagnosticStatusWrapper & stat)
...
       try {
-        avail = std::stoi(list[3].c_str());
+        size = std::stoi(list[1]);
+        used = std::stoi(list[2]);
+        avail = std::stoi(list[3]);
+        std::string capacity_str = list[4];
+        if (!capacity_str.empty() && capacity_str.back() == '%') {
+          capacity_str.pop_back();
+        }
+        capacity = std::stoi(capacity_str);
```
nishikawa-masaki: I believe the contents of "list" should be accessed with std::vector::at(), not with std::vector::[].
The reference manual says the following:
* list.at(x) throws an exception if x >= list.size().
* If x >= list.size(), the behavior of list[x] is undefined. It is not guaranteed to throw an exception (and it may crash).

In addition to that, the following code block (from line 355 to line 361 in the current version) should be moved to this "try" block. "[5]" should be replaced with "at(5)".
```
      std::string mounted_ = list[5];
      if (list.size() > 6) {
        std::string::size_type pos = line.find("% /");
        if (pos != std::string::npos) {
          mounted_ = line.substr(pos + 2);  // 2 is "% " length
        }
      }
```

system/autoware_system_monitor/src/hdd_monitor/hdd_monitor.cpp
```diff
@@ -340,6 +361,15 @@ void HddMonitor::checkUsage(diagnostic_updater::DiagnosticStatusWrapper & stat)
       }
       stat.add(fmt::format("HDD {}: mounted on", hdd_index), mounted_.c_str());
 
+      HddUsage hdd_usage;
+      hdd_usage.size = size;
+      hdd_usage.used = used;
+      hdd_usage.avail = avail;
+      hdd_usage.capacity = capacity;
+      hdd_usage.filesystem = list[0];
```
nishikawa-masaki: Please declare "std::string filesystem" and assign the value of list.at(0) in the preceding "try" block.
There is no guarantee that list[0] exists.

The same caution applies to the following original [code block](https://github.com/autowarefoundation/autoware_universe/pull/11023/files#:~:text=stat.add(fmt%3A%3Aformat(%22HDD%20%7B%7D%3A%20filesystem,HDD%20%7B%7D%3A%20use%22%2C%20hdd_index)%2C%20list%5B4%5D.c_str())%3B) (from line 350 to line 354 in the current version).
```
      stat.add(fmt::format("HDD {}: filesystem", hdd_index), list[0].c_str());
      stat.add(fmt::format("HDD {}: size", hdd_index), (list[1] + " MiB").c_str());
      stat.add(fmt::format("HDD {}: used", hdd_index), (list[2] + " MiB").c_str());
      stat.add(fmt::format("HDD {}: avail", hdd_index), (list[3] + " MiB").c_str());
      stat.add(fmt::format("HDD {}: use", hdd_index), list[4].c_str());

```

system/autoware_system_monitor/src/hdd_monitor/hdd_monitor.cpp
```diff
@@ -925,5 +957,44 @@ int HddMonitor::unmountDevice(std::string & device)
...
+
+  for (const auto & hdd_usage : hdd_usages_) {
+    tier4_external_api_msgs::msg::HddPartitionStatus partition_status;
+    partition_status.size = hdd_usage.size;
+    partition_status.used = hdd_usage.used;
+    partition_status.avail = hdd_usage.avail;
+    partition_status.capacity = hdd_usage.capacity;
+    partition_status.filesystem = hdd_usage.filesystem;
+    partition_status.mounted_on = hdd_usage.mounted_on;
+    hdd_status.partitions.push_back(partition_status);
```
nishikawa-masaki: Please consider using emplace_back().

system/autoware_system_monitor/src/hdd_monitor/hdd_monitor.cpp
```diff
@@ -925,5 +957,44 @@ int HddMonitor::unmountDevice(std::string & device)
...
+    float read_iops = hdd_stats_[itr->first].read_iops_;
+    float write_iops = hdd_stats_[itr->first].write_iops_;
+
+    tier4_external_api_msgs::msg::HddDeviceStatus device_status;
+    device_status.name = itr->second.disk_device_;
+    device_status.read_data_rate = read_data_rate;
+    device_status.write_data_rate = write_data_rate;
+    device_status.read_iops = read_iops;
+    device_status.write_iops = write_iops;
+    hdd_status.devices.push_back(device_status);
```
nishikawa-masaki: Please consider using emplace_back().

nishikawa-masaki (changes requested):
I found some dangerous code blocks, I mean the use of std::vector::[],  in the original source file before this PR, but they won't cause any trouble  under normal conditions.
Though it is not your fault, I believe it should be fixed because some changes in this PR share the same problem.

The rest of the comments are suggestions. You can ignore them if you have any reason.
For example, the performance improvement with replacing push_back() with emplace_back() is small while the change may confuse readers who are not familiar with emplace_back().

system/autoware_system_monitor/include/system_monitor/hdd_monitor/hdd_monitor.hpp
iwatake2222: I also thought it would be better to use `int32_t`. However, since other parts of the same code used `int`, I aligned with that. That said, upon reviewing the code again, I noticed there was no consistency, with some places using `uint64_t` and others. Therefore, I decided to standardize it by using `int32_t`.

Fixed in https://github.com/autowarefoundation/autoware_universe/pull/11023/commits/0163caee6733584756bb171bd3e86ec99234f129

system/autoware_system_monitor/include/system_monitor/hdd_monitor/hdd_monitor.hpp
iwatake2222: Fixed in https://github.com/autowarefoundation/autoware_universe/commit/0163caee6733584756bb171bd3e86ec99234f129

system/autoware_system_monitor/src/hdd_monitor/hdd_monitor.cpp
iwatake2222: Fixed in https://github.com/autowarefoundation/autoware_universe/commit/0163caee6733584756bb171bd3e86ec99234f129

system/autoware_system_monitor/src/hdd_monitor/hdd_monitor.cpp
iwatake2222: Great Suggestion! Thank you!!
Fixed in https://github.com/autowarefoundation/autoware_universe/commit/0163caee6733584756bb171bd3e86ec99234f129

system/autoware_system_monitor/src/hdd_monitor/hdd_monitor.cpp
iwatake2222: Fixed in https://github.com/autowarefoundation/autoware_universe/commit/0163caee6733584756bb171bd3e86ec99234f129

system/autoware_system_monitor/src/hdd_monitor/hdd_monitor.cpp
iwatake2222: Fixed in https://github.com/autowarefoundation/autoware_universe/commit/0163caee6733584756bb171bd3e86ec99234f129

system/autoware_system_monitor/src/hdd_monitor/hdd_monitor.cpp
iwatake2222: Fixed in https://github.com/autowarefoundation/autoware_universe/commit/0163caee6733584756bb171bd3e86ec99234f129

iwatake2222:
@nishikawa-masaki 
Thank you for your careful review. I fixed the code. Could you please review it again?

system/autoware_system_monitor/src/hdd_monitor/hdd_monitor.cpp
nishikawa-masaki: I'm afraid that a part of my intention in the previous comment didn't reach you.
Thank you for moving list member access to the preceding "try" block and using "at()".
I meant the changes will enable writing the following lines without this new "try" block.

        stat.add(fmt::format("HDD {}: status", hdd_index), usage_dict_.at(level));
        stat.add(fmt::format("HDD {}: filesystem", hdd_index), filesystem.c_str());
        stat.add(fmt::format("HDD {}: size", hdd_index), fmt::format("{} MiB", size));
        stat.add(fmt::format("HDD {}: used", hdd_index), fmt::format("{} MiB", used));
        stat.add(fmt::format("HDD {}: avail", hdd_index), fmt::format("{} MiB", avail));
        stat.add(fmt::format("HDD {}: use", hdd_index), fmt::format("{}", capacity));
        stat.add(fmt::format("HDD {}: mounted on", hdd_index), mounted.c_str());

I wrote the following comment a few minutes ago, but I need to investigate the behavior on an exception more carefully.
> I found that the original code set "-1" to "size" and "used" on exception and still published their raw values.
> (but an undefined behavior would have happen with "[]" under such condition)
> Therefore, strictly speaking, the code above behaves differently from the original code, but I believe it is acceptable.
> ("Undefined behavior" v.s. "size, used, avail, use are set to -1")

system/autoware_system_monitor/include/system_monitor/hdd_monitor/hdd_monitor.hpp
```diff
@@ -106,6 +110,21 @@ struct HddStat
...
 
+/**
+ * @brief usage of HDD for each partition
+ */
+struct HddPartitionStatus
+{
+  int32_t size;            //!< @brief Total size of the filesystem in megabytes (MB)
+  int32_t used;            //!< @brief Used space in the filesystem in megabytes (MB)
+  int32_t avail;           //!< @brief Available (free) space in the filesystem in megabytes (MB)
+  int32_t capacity;        //!< @brief Percentage of used space in the filesystem (0–100)
```
nishikawa-masaki, line 122: Thank you for using int32_t.
Please add "\<cstdint\>" in the included file list.

system/autoware_system_monitor/src/hdd_monitor/hdd_monitor.cpp
```diff
@@ -310,11 +323,29 @@ void HddMonitor::checkUsage(diagnostic_updater::DiagnosticStatusWrapper & stat)
...
+          }
+        }
       } catch (std::exception & e) {
+        size = -1;
+        used = -1;
         avail = -1;
+        capacity = -1;
         error_str = e.what();
-        stat.add(fmt::format("HDD {}: status", hdd_index), "avail string error");
+        stat.add(fmt::format("HDD {}: status", hdd_index), "string error");
```
nishikawa-masaki: Just set an error flag here and handle the exceptional case later.
Please see one of the following comments.

system/autoware_system_monitor/src/hdd_monitor/hdd_monitor.cpp
```diff
@@ -325,20 +356,25 @@ void HddMonitor::checkUsage(diagnostic_updater::DiagnosticStatusWrapper & stat)
...
       }
-      stat.add(fmt::format("HDD {}: mounted on", hdd_index), mounted_.c_str());
+
+      auto & hdd_partition_status = hdd_partition_statuses_.emplace_back();
+      hdd_partition_status.size = size;
+      hdd_partition_status.used = used;
+      hdd_partition_status.avail = avail;
+      hdd_partition_status.capacity = capacity;
+      hdd_partition_status.filesystem = filesystem;
+      hdd_partition_status.mounted_on = mounted;
```
nishikawa-masaki: The flag set in the "catch" block on line 342 should be used here to publish a pair of key "HDD X: status", value "string error".
If any exception is caught, the values of size, used, avail, capacity, filesystem, mounted are unreliable.
I think those unreliable data should not be included in hdd_partition_status.

system/autoware_system_monitor/src/hdd_monitor/hdd_monitor.cpp
iwatake2222: Fixed in https://github.com/autowarefoundation/autoware_universe/pull/11023/commits/04197d356c5b174d2cff02cf9db038d80374a363
Thank you for your suggestion.

> Therefore, strictly speaking, the code above behaves differently from the very original code, but I believe it is acceptable.
("Undefined behavior" v.s. "size, used, avail, use are set to -1")

I think setting -1 is much better than "undefined behavior". ~, also I don't think there's much benefit in reporting raw values.~ 
In case error occurs in `stoi` , reporting raw values may be useful. However, I believe that's not something we need to be concerned about during regular operation.

iwatake2222: @nishikawa-masaki 
Could you please re-check this PR?

system/autoware_system_monitor/src/hdd_monitor/hdd_monitor.cpp
```diff
@@ -925,5 +959,42 @@ int HddMonitor::unmountDevice(std::string & device)
...
+    auto & device = hdd_status.devices.emplace_back();
+    device.name = itr->second.disk_device_;
+    device.read_data_rate = read_data_rate;
+    device.write_data_rate = write_data_rate;
+    device.read_iops = read_iops;
+    device.write_iops = write_iops;
+  }
+  pub_hdd_status_->publish(hdd_status);
+}
+
```
nishikawa-masaki, line 1002: PublishHddStatus() is executed in timer-callback context "onTimer()" and accesses the "hdd_partition_statuses_" vector member.
On the other hand, checkUsage() is executed in the publishing context and it modifies the "hdd_partition_statuses_" vector member.
There can be a racing condition and it may cause a crashing bug.
(Example : Reading from a vector while it is being cleared.)
We need to protect the data structures which are shared between threads, using mutex-es.
Also, we may need to consider about the role of threads.

The followings are my opinions:
* Publishing from the timer callback thread is not good (I think).
* Data should be updated in the timer callback thread.
  * Because some data requires considerable time to collect.
 * The publishing thread should only execute publishing.
 
 This means that we need to something similar to https://github.com/autowarefoundation/autoware_universe/pull/10545 **before** this PR.
 
 If it is "too much to do", the followings is another way to implement safe publishing for the time being.
 * Define another vector "hdd_partition_statuses_snapshot_".
 * The contents of "hdd_partion_statuses_" is copied to "hdd_partition_statuses_snapshot_" in a "protected by a mutex" block.
 * PublishHddStatus() reads data from "hdd_partition_statuses_snapshot_" in a "protected by a mutex" block.
   * PublishHddStatus() should not access "hdd_partition_statuses_".

nishikawa-masaki: Another option is calling PublishHddStatus() in the publishing thread's context.
What is the reason why PublishHddStatus() is being called in onTimer()?

nishikawa-masaki: I also found none of the functions called in onTimer() is thread-safe.
```
void HddMonitor::onTimer()
{
  updateHddConnections();
  updateHddInfoList();
  updateHddStatistics();

  publishHddStatus();
}
```
I wonder why they don't cause any fatal problem with current implementation. It looks dangerous.
It may be because the design of ROS2 Humble executors are not real multi-thread.
This is just a guess and I need to ask somebody or consult with ROS2 documents.

system/autoware_system_monitor/src/hdd_monitor/hdd_monitor.cpp
iwatake2222: @nishikawa-masaki 
In the original code, only "avail" was wrapped in a try-catch just because it was the only field being converted to a number. So I don’t think "avail" was treated as special for any deeper reason.

In the current implementation, added columns are parsed in a similar way. If a column is missing or the value cannot be converted to a number, the error is caught, and a "string error" message is left behind. So I think the current approach is consistent and sufficient.

system/autoware_system_monitor/src/hdd_monitor/hdd_monitor.cpp
iwatake2222: @nishikawa-masaki 
In ROS 2, unless otherwise specified, all callbacks (timers, subscriptions, etc.) are assigned to the same default callback group. Even when using a MultiThreadedExecutor, callbacks within the same callback group are executed sequentially, not in parallel.

In `hdd_monitor`, both `onTimer()` (which calls `PublishHddStatus()`) and `checkUsage()` (called via `updater_`) belong to the same callback group. Therefore, they cannot run concurrently, and access to the shared `hdd_partition_statuses_` vector is effectively serialized. For this reason, explicit locking is not necessary here.

On the other hand, in `cpu_monitor`, the timer callback is explicitly assigned to a different callback group (`timer_callback_group_ = this->create_callback_group(rclcpp::CallbackGroupType::MutuallyExclusive);`). This likely allows time-consuming operations to run in parallel with other callbacks, and thus, mutex protection is needed there.

So, the difference in synchronization handling between `hdd_monitor` and `cpu_monitor` is due to the use (or not) of separate callback groups.

I thought about adding mutexes, but given the current design, I think it's unnecessary. Both `onTimer()` and `checkUsage()` run in the same callback group, so they won’t be executed at the same time—even with a MultiThreadedExecutor. If we later introduce parallelism by using multiple callback groups, we can add synchronization at that point. For now, I prefer to keep the code simple.

iwatake2222: reference: https://docs.ros.org/en/foxy/Concepts/About-Executors.html
reference: https://docs.ros.org/en/foxy/How-To-Guides/Using-callback-groups.html
> Mutually Exclusive Callback Group prevents its callbacks from being executed in parallel - essentially making it as if the callbacks in the group were executed by a SingleThreadedExecutor.

system/autoware_system_monitor/include/system_monitor/hdd_monitor/hdd_monitor.hpp
iwatake2222: Fixed in 08a9ecb6b751409d940dc631ad3b28a982e44178

system/autoware_system_monitor/src/hdd_monitor/hdd_monitor.cpp
iwatake2222: Fixed in 08a9ecb6b751409d940dc631ad3b28a982e44178

system/autoware_system_monitor/src/hdd_monitor/hdd_monitor.cpp
iwatake2222: Fixed in 08a9ecb6b751409d940dc631ad3b28a982e44178

system/autoware_system_monitor/src/hdd_monitor/hdd_monitor.cpp
nishikawa-masaki: Thank you for the detailed explanation. You are right and I'm sorry for making erroneous comments.
Please ignore the preceding comments.

nishikawa-masaki (approved):
Now, it looks good to me.
Thank you for your feedback and fixes.

PR #11024, opened 2025-07-18
feat(autoware_system_monitor): publish network_status

iwatake2222:
## Description

- This PR introduces a feature to publish network status

## Related links

**Private Links:**

## How was this PR tested?

I confirmed `/system/system_monitor/net_monitor/network_status` topic shows the measured data

Screenshot from 2025-07-18 19-35-44

## Notes for reviewers

None.

#### Additions and removals

| Added | Pub | `/system/system_monitor/net_monitor/network_status` | `tier4_external_api_msgs/msg/NetworkStatus`   | [Network status](https://github.com/iwatake2222/tier4_autoware_msgs/blob/feat_external_api_system_monitor/tier4_external_api_msgs/msg/NetworkStatus.msg) |

## Effects on system behavior

None.

### DIFF
--- a/system/autoware_system_monitor/src/net_monitor/net_monitor.cpp
+++ b/system/autoware_system_monitor/src/net_monitor/net_monitor.cpp
@@ -69,6 +69,11 @@ NetMonitor::NetMonitor(const rclcpp::NodeOptions & options)
   updater_.add("IP Packet Reassembles Failed", this, &NetMonitor::check_reassembles_failed);
   updater_.add("UDP Buf Errors", this, &NetMonitor::check_udp_buf_errors);
 
+  // Publisher
+  rclcpp::QoS durable_qos{1};
+  durable_qos.transient_local();
+  pub_network_status_ = this->create_publisher<tier4_external_api_msgs::msg::NetworkStatus>("~/network_status", durable_qos);
+
   nl80211_.init();
 
   // Run I/O service processing loop
@@ -374,6 +379,7 @@ void NetMonitor::on_timer()
 {
   // Update list of network information
   update_network_list();
+  publishNetworkStatus();
 }
 
 void NetMonitor::shutdown_nl80211()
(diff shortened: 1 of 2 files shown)


iwatake2222:
@nishikawa-masaki 
Could you review this PR?

~Also, I noticed some cppchecks fails due to existing code. I'll fix them later.~ -> Done

nishikawa-masaki (approved):
If you use emplace_back() instead of push_back(), the performance would be a bit better.
Other than that, it looks good to me.

system/autoware_system_monitor/src/net_monitor/net_monitor.cpp
```diff
@@ -736,6 +743,45 @@ void NetMonitor::close_connection()
...
+
+    tier4_external_api_msgs::msg::NetworkInterfaceStatus interface_status;
+    interface_status.name = network.interface_name;
+    interface_status.rx_traffic = network.rx_traffic;
+    interface_status.tx_traffic = network.tx_traffic;
+    interface_status.capacity = network.speed;
+    interface_status.rx_errors = network.rx_errors;
+    interface_status.tx_errors = network.tx_errors;
+    interface_status.collisions = network.collisions;
+    network_status.interfaces.push_back(interface_status);
```
nishikawa-masaki: If there is no special reason, replacing push_back() with emplace_back() would produce a bit faster executable.
push_back() よりも emplace_back() の方が少し速いはずなので、特に問題がなければ emplace_back() を使うのが良いと思います。

iwatake2222: @nishikawa-masaki 
Thanks for the suggestion!
I updated the code to use emplace_back() properly by constructing the object directly and setting the fields afterward, instead of just replacing push_back, which wouldn’t have improved performance.
Fixed in https://github.com/autowarefoundation/autoware_universe/pull/11024/commits/6b6c47b64383f73545533df24b72bb2de5228c81

PR #11028, opened 2025-07-20
feat(map_based_prediction): prevent predicted path from chattering under noisy pose and velocity estimation

satoshi-ota:
## Description

This PR improves the robustness of crossing intention estimation against noise in velocity and pose estimation.
Previously, small fluctuations in perceived object velocity or orientation could cause the estimated intention to switch frequently, leading to unstable behavior near crosswalks.

To address this, a time-based filtering mechanism was introduced. The intention is now only judged as true or false if the corresponding condition has persisted for a specified duration.

`crossing_intention_duration` defines how long the system must continuously detect a crossing intention before accepting it.

`no_crossing_intention_duration` defines how long the system must continuously detect the absence of intention before clearing it.

As a result, the intention estimation becomes more stable and less sensitive to temporary perception noise, improving overall system reliability.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

| BEFORE | AFTER |
| --- | --- |
|  <video src="https://github.com/user-attachments/assets/c8d3f868-2ee3-462c-a36c-e7911bd36404"/> |  <video src="https://github.com/user-attachments/assets/7dcb7950-d2e5-42fa-bf45-db80e343c721"/> |

```
webauto ci scenario run --project-id prd_jt --scenario-id 8f516365-80ed-40e0-9d64-7e90d456002f --scenario-version-id 38 --scenario-parameters '__tier4_modifier_deceleration_parameter=-3,__tier4_modifier_ego_speed=11.1111,__tier4_modifier_front_overhang=1,__tier4_modifier_max_blinker_change_count=5,__tier4_modifier_max_deceleration=10,__tier4_modifier_max_goal_lateral_deviation_abs=1000,__tier4_modifier_max_goal_yaw_deviation_abs=5,__tier4_modifier_max_lateral_deviation_abs=15,__tier4_modifier_max_lateral_trajectory_displacement_local=15,__tier4_modifier_max_processing_time=1000,__tier4_modifier_max_resampled_relative_angle=5,__tier4_modifier_max_stand_still=100,__tier4_modifier_max_steer_change_count=20,__tier4_modifier_max_steer_rate_fast=15,__tier4_modifier_max_steer_rate_slow=13,__tier4_modifier_max_stop_deviation_abs=50,__tier4_modifier_min_npc_distance_euclidian=0,__tier4_modifier_min_npc_longitudinal_distance_euclidian=0,__tier4_modifier_min_npc_longitudinal_distance_lateral=0,__tier4_modifier_pedestrian_speed=2.0833,__tier4_modifier_start_pedestrian_distance=-6,__tier4_modifier_wheel_base=2.79' --simulation-name planning_sim_v2
```

- [x] [[PR check (OTA)][Lexus][CommonScenario_Basic] planning test](https://evaluation.tier4.jp/evaluation/reports/94bdce39-48ce-56e2-be03-38acaeca98af?project_id=prd_jt)
- [x] [[PR check (OTA)][Lexus][ControlDLRCommon_Basic] DLR test](https://evaluation.tier4.jp/evaluation/reports/382cb653-4acc-59d6-83dc-cdf0e9bd772b?project_id=prd_jt)
- [x] [PR check (OTA)][Lexus][FuncVerificationScenario_Basic] planning test

## Notes for reviewers

None.

## Interface changes

| Version | Parameter Name               | Type   | Default Value | Description                                                                                     |
|:--------|:-----------------------------|:-------|:---------------|:------------------------------------------------------------------------------------------------|
| New     | `crossing_intention_duration`    | double | `1.0`          | Minimum duration for which the crossing intention must continuously persist to be considered true  |
| New     | `no_crossing_intention_duration` | double | `1.0`          | Minimum duration for which the absence of crossing intention must continuously persist to be considered false |

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_map_based_prediction/README.md
+++ b/perception/autoware_map_based_prediction/README.md
@@ -191,6 +191,13 @@ This module treats **Pedestrians** and **Bicycles** as objects using the crosswa
 
 If there are a reachable crosswalk entry points within the `prediction_time_horizon` and the objects satisfies above condition, this module outputs additional predicted path to cross the opposite side via the crosswalk entry point.
 
+To prevent the predicted crossing path from chattering due to noise in the estimated pose or velocity of a pedestrian who is judged to have crossing intention, the module holds the crossing intention state for a certain duration based on the following parameters:
+
+| Parameter                        | Unit | Type   | Description                                                                                      |
+| -------------------------------- | ---- | ------ | ------------------------------------------------------------------------------------------------ |
+| `crossing_intention_duration`    | [s]  | double | Minimum duration that crossing intention must continuously persist to be judged as true          |
+| `no_crossing_intention_duration` | [s]  | double | Minimum duration that lack of crossing intention must continuously persist to be judged as false |
+
 This module takes into account the corresponding traffic light information.
 When RED signal is indicated, we assume the target object will not walk across.
 In addition, if the target object is stopping (not moving) against GREEN signal, we assume the target object will not walk across either.
@@ -243,8 +250,6 @@ If the target object is inside the road or crosswalk, this module outputs one or
 | `object_buffer_time_length`                                      | [s]   | double | Time span of object history to store the information                                                                                  |
 | `history_time_length`                                            | [s]   | double | Time span of object information used for prediction                                                                                   |
 | `prediction_time_horizon_rate_for_validate_shoulder_lane_length` | [-]   | double | prediction path will disabled when the estimated path length exceeds lanelet length. This parameter control the estimated path length |
-| `crossing_intention_duration`                                    | [s]   | double | Minimum duration that crossing intention must continuously persist to be judged as true                                               |
-| `no_crossing_intention_duration`                                 | [s]   | double | Minimum duration that lack of crossing intention must continuously persist to be judged as false                                      |
 
 ## Assumptions / Known limits
 
--- a/perception/autoware_map_based_prediction/README.md
+++ b/perception/autoware_map_based_prediction/README.md
@@ -198,6 +198,10 @@ To prevent the predicted crossing path from chattering due to noise in the estim
 | `crossing_intention_duration`    | [s]  | double | Minimum duration that crossing intention must continuously persist to be judged as true          |
 | `no_crossing_intention_duration` | [s]  | double | Minimum duration that lack of crossing intention must continuously persist to be judged as false |
 
+!!! note
+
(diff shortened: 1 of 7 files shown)


maxime-clem (approved):
Code looks good and from the video the improvement is clear.

It does add a small delay, but the first detection is used directly so it might not be a big issue, although there might be in the following case:
- initial detection of the object: high noise and the initial prediction is `no_crossing_intention`.
- next iteration: better detection but since previous decision was `no_crossing_intention`, we cannot switch to the `crossing_intention` for the given `crossing_intention_duration`.

It would be nice to add a small paragraph about this feature in the README and mention possible risks with delays.

satoshi-ota:
> Code looks good and from the video the improvement is clear.
> 
> It does add a small delay, but the first detection is used directly so it might not be a big issue, although there might be in the following case:
> 
> * initial detection of the object: high noise and the initial prediction is `no_crossing_intention`.
> * next iteration: better detection but since previous decision was `no_crossing_intention`, we cannot switch to the `crossing_intention` for the given `crossing_intention_duration`.
> 
> It would be nice to add a small paragraph about this feature in the README and mention possible risks with delays.

@maxime-clem Thank you for quick review. I'll write small explanation tomorrow.

kyoichi-sugahara (approved):

technolojin:
Here is an example scenario. 
After `no_crossing_intention_duration` ego vehicle will accelerate to cross the crosswalk. If the pedestrian also start to cross, will the ego vehicle stop? How the prediction switched to `cross` state when the pedestrian start to move?

satoshi-ota:
> Here is an example scenario. After `no_crossing_intention_duration` ego vehicle will accelerate to cross the crosswalk. If the pedestrian also start to cross, will the ego vehicle stop? How the prediction switched to `cross` state when the pedestrian start to move?

Thank you for your question.

As described in the README, the `crossing_intention_duration` parameter defines the minimum duration that a crossing intention must persist for the system to output a predicted path that crosses the crosswalk. Therefore, it is not recommended to set this parameter to a large value, as it may delay the system's response to actual pedestrian crossing behavior.

In the situation you mentioned, the system will determine that the pedestrian intends to cross only after the duration specified by `crossing_intention_duration` has elapsed. However, this does not affect the linear predicted path that is based on the object's estimated pose and velocity. So in case of emergency, the system can still perform a stop based on that predicted path.

PR #11043, opened 2025-07-23
feat(mission_planner): manual lane selection

PanConChicharron:
## Description
Allows the user to force a lane change using an rviz UI.

## Related links

**Parent Issue:**

- [Link](https://tier4.atlassian.net/browse/RT1-10751)

## How was this PR tested?
Example use-cases:

[left-lane change.webm](https://github.com/user-attachments/assets/fad6ede7-122d-452d-b1de-4f2e23bd61dd)

Rejecting requests when the current lane does not have a shiftable lane:
[current_lane_reject.webm](https://github.com/user-attachments/assets/a56d5498-862f-4030-8e47-4c529d9bebe8)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/autoware_planning_msgs/CMakeLists.txt
+++ b/autoware_planning_msgs/CMakeLists.txt
@@ -15,6 +15,7 @@ rosidl_generate_interfaces(${PROJECT_NAME}
   "msg/PathPoint.msg"
   "msg/RouteState.msg"
   "srv/ClearRoute.srv"
+  "srv/SetLaneChangeOverride.srv"
   "srv/SetLaneletRoute.srv"
   "srv/SetWaypointRoute.srv"
   DEPENDENCIES
--- a/autoware_planning_msgs/srv/SetLaneChangeOverride.srv
+++ b/autoware_planning_msgs/srv/SetLaneChangeOverride.srv
@@ -1,0 +1,3 @@
+uint8 lane_change_direction
+---
+autoware_common_msgs/ResponseStatus status


planning/autoware_mission_planner_universe/src/mission_planner/mission_planner.hpp
```diff
@@ -63,9 +64,16 @@ using geometry_msgs::msg::PoseStamped;
 using nav_msgs::msg::Odometry;
 using std_msgs::msg::Header;
 using tier4_planning_msgs::msg::RerouteAvailability;
+using tier4_planning_msgs::srv::SetLaneChangeOverride;
```
takayuki5168: Instead of SetLaneChangeOverride,
- SetPreferredLane
- ChangePreferredLane
sounds better to me.

In that case, 
- lane_change_direction
    - -> preferred_lane would be better
- enum class of DIRECTION
    - -> The following would be better
```enum class PREFERRED_LANE {
  MANUAL_LEFT,
  MANUAL_RIGHT,
  AUTO,
};
```

planning/autoware_mission_planner_universe/src/mission_planner/mission_planner.cpp
```diff
@@ -263,6 +269,145 @@ void MissionPlanner::on_clear_route(
...
+  RCLCPP_INFO_STREAM(
+    get_logger(), "Changing lane "
+                    << (override_direction == DIRECTION::LEFT    ? "left"
+                        : override_direction == DIRECTION::RIGHT ? "right"
+                                                                 : "auto"));
+  RCLCPP_INFO_STREAM(get_logger(), "Current route segments: " << route.segments.size());
+
+  auto final_iter = route.segments.end();
+
+  if (override_direction != DIRECTION::AUTO) {
```
takayuki5168: Please create a new member function for this if-else block.

planning/autoware_mission_planner_universe/src/mission_planner/mission_planner.cpp
```diff
@@ -263,6 +269,145 @@ void MissionPlanner::on_clear_route(
...
+
+  change_state(is_reroute ? RouteState::REROUTING : RouteState::ROUTING);
+  auto route = *current_route_;
+
+  boost::uuids::random_generator gen;
+  boost::uuids::uuid uuid = gen();
+  std::copy(uuid.begin(), uuid.end(), route.uuid.uuid.begin());
+
+  // modify the preferred lanelet to change the lane
+  DIRECTION override_direction = req->lane_change_direction == 0   ? DIRECTION::LEFT
```
kosuke55:
```suggestion
  const DIRECTION override_direction = req->lane_change_direction == 0   ? DIRECTION::LEFT
```

planning/autoware_mission_planner_universe/src/mission_planner/mission_planner.cpp
```diff
@@ -263,6 +269,145 @@ void MissionPlanner::on_clear_route(
...
+                                                                   : DIRECTION::AUTO;
+
+  RCLCPP_INFO_STREAM(
+    get_logger(), "Changing lane "
+                    << (override_direction == DIRECTION::LEFT    ? "left"
+                        : override_direction == DIRECTION::RIGHT ? "right"
+                                                                 : "auto"));
+  RCLCPP_INFO_STREAM(get_logger(), "Current route segments: " << route.segments.size());
+
+  auto final_iter = route.segments.end();
```
kosuke55:
```suggestion
  const auto final_iter = route.segments.end();
```

planning/autoware_mission_planner_universe/src/mission_planner/mission_planner.cpp
```diff
@@ -263,6 +269,145 @@ void MissionPlanner::on_clear_route(
...
+      if (override_direction == DIRECTION::LEFT && index > 0) {
+        // shift to the primitive on the left
+        segment.preferred_primitive = segment.primitives.at(index - 1);
+        RCLCPP_INFO_STREAM(
+          get_logger(), "Shifted left to primitive ID: " << segment.preferred_primitive.id);
+      } else if (override_direction == DIRECTION::RIGHT && index + 1 < segment.primitives.size()) {
+        // shift to the primitive on the right
+        segment.preferred_primitive = segment.primitives.at(index + 1);
+        RCLCPP_INFO_STREAM(
+          get_logger(), "Shifted right to primitive ID: " << segment.preferred_primitive.id);
```
kosuke55: `segment.primitives.at`

need valid access check?

planning/behavior_path_planner/autoware_behavior_path_goal_planner_module/src/goal_planner_module.cpp
```diff
@@ -690,6 +690,9 @@ std::pair<LaneParkingResponse, FreespaceParkingResponse> GoalPlannerModule::sync
 
 void GoalPlannerModule::updateData()
 {
+  if (!utils::isAllowedGoalModification(planner_data_->route_handler)) {
+    return;
+  }
```
kosuke55: Regarding the goal planner, it'll be LGTM if the tests pass.

planning/autoware_mission_planner_universe/src/mission_planner/manual_lane_change_handler.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+  if (!original_route_) {
+    // Save the original route if not already saved
+    original_route_ = *current_route_;
+  }
+
+  auto final_iter = std::prev(route.segments.end(), 2);
+  auto start_iter = final_iter;
+
+  // Find the segment that contains the start_lanelet_id
```
takayuki5168: ego_lanelet_id would be easier to understand.

planning/autoware_mission_planner_universe/src/mission_planner/manual_lane_change_handler.cpp
```diff
@@ -0,0 +1,140 @@
...
+    auto start_iter_primitive = std::find_if(
+      iter->primitives.begin(), iter->primitives.end(),
+      [&start_lanelet_id](const LaneletPrimitive & p) { return p.id == start_lanelet_id; });
+    if (start_iter_primitive != iter->primitives.end()) {
+      start_iter = iter;
+      break;
+    }
+  }
+
+  bool route_updated = false;
```
takayuki5168: Please add a comment like `The lane selection is cancled due to the XXX conditions`.
The additional and detailed reasons.

PanConChicharron: Done.

planning/autoware_mission_planner_universe/src/mission_planner/manual_lane_change_handler.cpp
PanConChicharron: Done.

planning/autoware_mission_planner_universe/src/mission_planner/mission_planner.hpp
PanConChicharron: Done

planning/autoware_mission_planner_universe/src/mission_planner/mission_planner.cpp
PanConChicharron: Done.

planning/autoware_mission_planner_universe/src/mission_planner/mission_planner.cpp
PanConChicharron: Done.

planning/autoware_mission_planner_universe/src/mission_planner/mission_planner.cpp
PanConChicharron: `index + 1 < segment.primitives.size()`

This handles that, no?

planning/autoware_mission_planner_universe/src/mission_planner/mission_planner.cpp
PanConChicharron: Moved to handler

planning/behavior_path_planner/autoware_behavior_path_goal_planner_module/src/goal_planner_module.cpp
PanConChicharron: These were no longer needed

isamu-takagi:
I'll leave the detailed route planning algorithm to you, so I'd like to mainly confirm the interface specifications.

The current interface appears to specify left and right relative to the reference lane, but which of the following is the reference lane?
- Preferred lanes without manual overrides (This means it cannot shift more than one lanes)
- Preferred lanes with manual overrides (This means it can shift more than one lanes by multiple service calls)
- Current lane (It is intuitive for the operator if the vehicle is not on preferred lanes. However, we need to define what happens when the vehicle moves to the preferred lane)

And what happens if the route is changed while the vehicle is driving?
- The manual override state is maintained and the left/right shift is recalculated for the new preferred lane
- Manual override state automatically returns to auto.

Is the override range sufficient? Currently, it is auto and left/right (shift amount is plus or minus 1)
- Is a shift amount of 2 or more required?
- Is a zero shift necessary? (Stay in the current lane even if the preferred lane moves)
- How about specifying the lanes directly and absolutely instead of relative values?

What is the manual override result if the preferred lanes shift mid-route?
- Shifts for each preferred lane.
- Fixed to the shifted lane based on the current preferred lane.

PR #11046, opened 2025-07-23
fix(autoware_behavior_path_static_obstacle_avoidance_module): prevent chattering in avoidance decision

yhisaki:
## Description
This PR addresses the issue of chattering in avoidance decisions for static obstacles. Previously, the avoidance decision (whether an object is avoidable with desired shift length) could fluctuate between planning cycles for the same object, leading to unstable behavior. The fix maintains the initial avoidance decision for each object to ensure consistent avoidance behavior.

### Changes
- Add a new member variable `avoidable_by_desired_shift_length_objects_` to `ShiftLineGenerator` class to maintain avoidance decisions for objects
- Modify `ShiftLineGenerator::generateAvoidOutline` to store and preserve the avoidance decision once an object is determined to be avoidable with desired shift length
- Update the method signature of `ShiftLineGenerator::generateAvoidOutline` to be non-const to allow storing avoidance decisions

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


planning/behavior_path_planner/autoware_behavior_path_static_obstacle_avoidance_module/include/autoware/behavior_path_static_obstacle_avoidance_module/shift_line_generator.hpp
```diff
@@ -235,6 +237,8 @@ class ShiftLineGenerator
 
   std::optional<ShiftLine> last_{std::nullopt};
 
+  std::set<std::string> avoidable_by_desired_shift_length_objects_;
```
satoshi-ota: Would it be possible to place this variable inside `AvoidancePlanningData`?

planning/behavior_path_planner/autoware_behavior_path_static_obstacle_avoidance_module/src/shift_line_generator.cpp
```diff
@@ -116,7 +118,7 @@ AvoidLineArray ShiftLineGenerator::generate(
 }
 
 AvoidOutlines ShiftLineGenerator::generateAvoidOutline(
-  AvoidancePlanningData & data, [[maybe_unused]] DebugData & debug) const
+  AvoidancePlanningData & data, [[maybe_unused]] DebugData & debug)
```
satoshi-ota: I’d like to keep it `const` as much as possible.

PR #11055, opened 2025-07-24
fix(autoware cuda pointcloud preprocessor): trim IMU/twist queues correctly

mojomex:
## Description

The twist and IMU queues were trimmed too aggressively, removing all messages before the first point's timestamp. As the time between first point and first twist/IMU message can potentially be multiple milliseconds, some of the points in the cloud might not be distortion-corrected correctly (ironic, I know!)

See this video (showing concat, and a perception pointcloud, both exhibiting a flickering artifact in the center back area)

[0-artifacts-concat-vs-perception.webm](https://github.com/user-attachments/assets/a424128a-3a3a-4702-9cbd-da1f4284a1b9)

The fix is to trim the queues such that the last message before the first point is preserved, ensuring points are covered from the very beginning.

There are no more flickering artifacts (ignore the slight packet loss though, it is unrelated :bow:)

[7-no_artifacts-perception-fix.webm](https://github.com/user-attachments/assets/93cfe0c2-6851-4631-ac2f-6e6c7e0e0782)

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

More robust pointcloud pipeline, more stable perception results.


drwnz (approved):
LGTM, nice find!

PR #11059, opened 2025-07-25
fix(lane_change): properly handle force activation when there is no valid LC path

mkquda:
## Description

If force activation request is received when there is no valid lc path, LC module state will be left hanging in RUNNING state and cannot properly transition to FAILURE.

This PR fixes this issue by confirming there is a valid path when force activation request is received. If there is no valid path, the request is ignored and a warning msg is displayed.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?
- PSIM

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/interface.cpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/interface.cpp
@@ -238,7 +238,7 @@ bool LaneChangeInterface::canTransitSuccessState()
 
 bool LaneChangeInterface::canTransitFailureState()
 {
-  const auto force_activated = [&]() {
+  const auto force_activated = std::invoke([&]() {
     const bool is_force_activated = std::any_of(
       rtc_interface_ptr_map_.begin(), rtc_interface_ptr_map_.end(),
       [&](const auto & rtc) { return rtc.second->isForceActivated(uuid_map_.at(rtc.first)); });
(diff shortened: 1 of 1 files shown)


kyoichi-sugahara (approved):
I think the changes are reasonable, so approve assuming you've confirmed the behavioral improvements by reproducing the situation.

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/interface.cpp
```diff
@@ -238,11 +238,20 @@ bool LaneChangeInterface::canTransitSuccessState()
...
+    if (is_force_activated && !module_type_->isValidPath()) {
+      RCLCPP_WARN_THROTTLE(
+        getLogger(), *clock_, 1000, "Force activated, but no valid path. Ignore force activation.");
+      return false;
+    }
+    return is_force_activated;
+  };
 
-  if (force_activated) {
+  if (force_activated()) {
```
zulfaqar-azmi-t4: (nits) since `force_activated()` is directly called after the lambda, it's nicer to directly invoke the lambda in-place.

mkquda: resolved in [da0fa4f](https://github.com/autowarefoundation/autoware_universe/pull/11059/commits/da0fa4ff372a70c6a7f33591dc3bef4f391b569a)

PR #11060, opened 2025-07-25
feat(autoware_agnocast_wrapper): support runtime fallback to ROS 2 behavior

bdm-k:
## Description

This PR implements tier4/agnocast#658.

## Notes for reviewers

This PR depends on a feature of Agnocast that will be introduced in tier4/agnocast#670. So, this PR is intended to be merged only after a new version of Agnocast including that feature is released.

That said, I would appreciate any feedback in advance, as this is my first contribution to Autoware Universe and there might be things I'm missing. Thank you for your kind support.

## Interface changes

- `AUTOWARE_POLLING_SUBSCRIBER(MessageT)` is replaced by `AUTOWARE_POLLING_SUBSCRIBER_PTR(MessageT)`. While this is technically a breaking change, `AUTOWARE_POLLING_SUBSCRIBER` is unused currently.
- `AUTOWARE_CREATE_POLLING_SUBSCRIBER(message_type, topic, qos)` is added to help create a polling subscriber.

## Effects on system behavior

When the system is built with Agnocast enabled but the environment variable `ENABLE_AGNOCAST` is not set to 1 at runtime, it falls back to ROS 2 behavior instead of using Agnocast.


sykwer:
draft for now (will be made ready after Agnocast 2.1.2 released)

common/autoware_agnocast_wrapper/include/autoware/agnocast_wrapper/autoware_agnocast_wrapper.hpp
```diff
@@ -16,27 +16,469 @@
...
+
+  virtual agnocast::ipc_shared_ptr<MessageT> move_agnocast_ptr() && noexcept = 0;
+  virtual std::shared_ptr<MessageT> move_ros2_ptr() && noexcept = 0;
+};
+
+template <typename MessageT, OwnershipType Ownership>
+class agnocast_message : public message_interface<MessageT, Ownership>
+{
+  using ros2_ptr_t = std::conditional_t<
+    Ownership == OwnershipType::Unique, std::unique_ptr<MessageT>, std::shared_ptr<MessageT>>;
```
Koichi98, line 104: This declaration of ros2_ptr_t looks redundant but can't be unified since it uses type parameters from the class it belong to. Am I right? Is there a good way to avoid this duplication?

common/autoware_agnocast_wrapper/include/autoware/agnocast_wrapper/autoware_agnocast_wrapper.hpp
```diff
@@ -16,27 +16,469 @@
...
+{
+  static const int sv = get_ENABLE_AGNOCAST();
+  return sv == 1;
+}
+
+template <typename MessageT>
+class Subscription
+{
+  // We are not storing the subscription objects for now, but this may chage once
+  // agnnocast::Subscription supports some functionality.
```
Koichi98: better to fix typo : change, agnocast.

common/autoware_agnocast_wrapper/include/autoware/agnocast_wrapper/autoware_agnocast_wrapper.hpp
```diff
@@ -16,27 +16,469 @@
...
+  autoware::agnocast_wrapper::message_ptr<    \
+    MessageT, autoware::agnocast_wrapper::OwnershipType::Shared>
+#define AUTOWARE_SUBSCRIPTION_PTR(MessageT) \
+  typename autoware::agnocast_wrapper::Subscription<MessageT>::SharedPtr
+#define AUTOWARE_PUBLISHER_PTR(MessageT) \
+  typename autoware::agnocast_wrapper::Publisher<MessageT>::SharedPtr
+#define AUTOWARE_POLLING_SUBSCRIBER_PTR(MessageT) \
+  typename autoware::agnocast_wrapper::PollingSubscriber<MessageT>::SharedPtr
 
 #define AUTOWARE_CREATE_SUBSCRIPTION(message_type, topic, qos, callback, options) \
```
Koichi98, line 43: @sykwer 
It seems that we need to specify 'options' to create subscription but Agnocast interface doesn't necessarily need them. Why isn't a non-options version of create_subscription prepared?

common/autoware_agnocast_wrapper/include/autoware/agnocast_wrapper/autoware_agnocast_wrapper.hpp
bdm-k: That's right. I don't think there is a good way.

Koichi98:
@bdm-k 
Would it be possible to add me as a collaborator? I'd like to make a few corrections.

common/autoware_agnocast_wrapper/include/autoware/agnocast_wrapper/autoware_agnocast_wrapper.hpp
Koichi98: It is probably just that it is not used when applying agnocast to the specific node we chose for testing. We may better add it in the later PR.

bdm-k:
@Koichi98 
I just sent the invite

Koichi98:
@bdm-k 
Thank you! Could you also check the permission? I need a higher level of permissions than Write access.

sykwer:
Agnocast v2.1.2 introduced.
https://github.com/autowarefoundation/autoware/pull/6412
https://github.com/autowarefoundation/autoware/pull/6421
https://github.com/autowarefoundation/autoware_universe/pull/11231

Koichi98:
@bdm-k 
The DCO check is failing on your pull request. It looks like several of your commits are missing the required Signed-off-by line. Could you solve this?

sykwer (approved):

PR #11075, opened 2025-07-29
fix(autoware_multi_object_tracker): set default existence probability for untrusted channels

technolojin:
## Description

the channel configuration `trust_existence_probability` was only implemented on the tracker initialization, but not in the update.
This PR implements default existence probability when the option is false.

## Related links

**Parent Issue:**

[TIER IV INTERNAL](https://star4.slack.com/archives/C076ZGRC15X/p1753424015987739)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


technolojin:
merged to https://github.com/autowarefoundation/autoware_universe/pull/11109

PR #5904, opened 2023-12-19
feat(behavior_path_planner): add a method to notify the planner manager of a successful lane change (update root lanelet)

danielsanchezaran:
## Description

Create a method for the scene modules to notify the planner manager that a lane change has occurred. By default the method returns false, but its overridden for the lane_change scene_module. This change helps simplify the root lanelet change procedure and makes it so you don't need to modify the planner manager to request a lane change if you add a new behavior path module.

## Tests performed

PSim -> perform lane changes bu setting goal in different lanelet and verified that the root lanelet updates after the LC module is successful. 

## Pre-review checklist for the PR author

The PR author **must** check the checkboxes below when creating the PR.

- [x] I've confirmed the [contribution guidelines].
- [x] The PR follows the [pull request guidelines].

## In-review checklist for the PR reviewers

The PR reviewers **must** check the checkboxes below before approval.

- [ ] The PR follows the [pull request guidelines].
- [ ] The PR has been properly tested.
- [ ] The PR has been reviewed by the code owners.

## Post-review checklist for the PR author

The PR author **must** check the checkboxes below before merging.

- [ ] There are no open discussions or they are tracked via tickets.
- [ ] The PR is ready for merge.

After all checkboxes are checked, anyone who has write access can merge the PR.

[contribution guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/
[pull request guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/pull-request-guidelines/

PR #11076, opened 2025-07-29
feat(autoware_bevformer): implementation of bevformer using tensorrt

naveen-mcw:
## Description

This PR introduces BEVFormer as a new 3D object detection method in Autoware, leveraging multi-view camera images and temporal fusion for improved performance.

**News:** This node is based on and adapted for Autoware from [Multicoreware's](https://multicorewareinc.com/) BEVFormer ROS 2 C++ repository.

## How was this PR tested?

- **Environment:**
  - GPU: GeForce RTX 2080 Ti
  - CUDA: 12.4
  - TensorRT: 10.8.0

## Note for reviewers
  - Tested using [nuScenes dataset](https://www.nuscenes.org/nuscenes#)
  - ONNX model [bevformer_small.onnx](https://multicorewareinc1-my.sharepoint.com/:u:/g/personal/naveen_sathiyaseelan_multicorewareinc_com/ERQSpS-BoAZGh4R4zNZhITcB58aqDW_tu9aKHLpit6aLAg?e=IZ5nZN)

Reviewers can test the functionality of this PR by refering [here](https://github.com/naveen-mcw/autoware.universe/blob/feat/add_tensorrt_bevformer/perception/autoware_tensorrt_bevformer/README.md).

#### Add a new `perception_mode`
- If `perception_mode = camera_lidar`, the detector will launch the `tensorrt_bevformer` node.

### Topic changes

| Change Type | Direction | Topic Name                | Message Type                                 | Description                                |
|-------------|-----------|--------------------------|----------------------------------------------|--------------------------------------------|
| Added       | Sub       | ~/input/topic_img_fl     | sensor_msgs::msg::Image                      | input front_left camera image              |
| Added       | Sub       | ~/input/topic_img_f      | sensor_msgs::msg::Image                      | input front camera image                   |
| Added       | Sub       | ~/input/topic_img_fr     | sensor_msgs::msg::Image                      | input front_right camera image             |
| Added       | Sub       | ~/input/topic_img_bl     | sensor_msgs::msg::Image                      | input back_left camera image               |
| Added       | Sub       | ~/input/topic_img_b      | sensor_msgs::msg::Image                      | input back camera image                    |
| Added       | Sub       | ~/input/topic_img_br     | sensor_msgs::msg::Image                      | input back_right camera image              |
| Added       | Sub       | ~/input/can_bus          | autoware_localization_msgs::msg::KinematicState        | CAN bus data for ego-motion                |
| Added       | Pub       | ~/output/boxes           | autoware_perception_msgs::msg::DetectedObjects| detected objects                           |
| Added       | Pub       | ~/output_bboxes          | visualization_msgs::msg::MarkerArray         | detected objects for NuScenes visualization|

## ROS Parameter Changes

| Change Type   | Parameter Name        | Type    | Default Value | Description                               |
|---------------|----------------------|---------|---------------|-------------------------------------------|
| Added         | bev_detection_model   | string  | bevformer     | Set BEV detection model type              |
| Modified      | perception_mode       | string  | lidar  | Add a new perception mode of `camera_lidar`     |

## Effects on System Behavior

None.

Looping in @Selventhiran-Rengaraj-MCW, @ramaseshanms-mcw.


xmfcx:
https://github.com/user-attachments/assets/f4dd1ec7-5c82-4be7-98e6-7fb452617992

I've got the following error when I ran the node and played the dataset:

```
[autoware_tensorrt_bevformer_node-1] [INFO 1754770551.536393563] [tensorrt_bevformer]: Image processing pipeline is now active. (checkInitialization() at /home/mfc/projects/autoware/src/universe/autoware_universe/perception/autoware_tensorrt_bevformer/src/bevformer_node.cpp:260)
[autoware_tensorrt_bevformer_node-1] /home/mfc/projects/autoware/install/autoware_tensorrt_bevformer/lib/autoware_tensorrt_bevformer/autoware_tensorrt_bevformer_node: symbol lookup error: /home/mfc/projects/autoware/install/autoware_tensorrt_bevformer/lib/libautoware_tensorrt_bevformer.so: undefined symbol: _ZNK5Eigen10MatrixBaseINS_6MatrixIdLi3ELi3ELi0ELi3ELi3EEEE7inverseEv
[ERROR] [autoware_tensorrt_bevformer_node-1]: process has died [pid 3624941, exit code 127, cmd '/home/mfc/projects/autoware/install/autoware_tensorrt_bevformer/lib/autoware_tensorrt_bevformer/autoware_tensorrt_bevformer_node --ros-args -r __node:=tensorrt_bevformer --params-file /tmp/launch_params_7_wu_beg --params-file /tmp/launch_params_ekg5knom --params-file /tmp/launch_params_db142xi4 --params-file /tmp/launch_params_q8tn_02d --params-file /tmp/launch_params_tllvycc2 --params-file /tmp/launch_params_c42yocaq --params-file /tmp/launch_params_o3ww1mhj --params-file /tmp/launch_params_ibgx4acb -r ~/input/topic_img_fl:=/nuscenes/CAM_FRONT_LEFT/image -r ~/input/topic_img_f:=/nuscenes/CAM_FRONT/image -r ~/input/topic_img_fr:=/nuscenes/CAM_FRONT_RIGHT/image -r ~/input/topic_img_bl:=/nuscenes/CAM_BACK_LEFT/image -r ~/input/topic_img_b:=/nuscenes/CAM_BACK/image -r ~/input/topic_img_br:=/nuscenes/CAM_BACK_RIGHT/image -r ~/input/can_bus:=/nuscenes/can_bus -r ~/input/scene_token:=/nuscenes/scene_tokens -r ~/input/topic_img_fl/camera_info:=/nuscenes/CAM_FRONT_LEFT/camera_info -r ~/input/topic_img_f/camera_info:=/nuscenes/CAM_FRONT/camera_info -r ~/input/topic_img_fr/camera_info:=/nuscenes/CAM_FRONT_RIGHT/camera_info -r ~/input/topic_img_bl/camera_info:=/nuscenes/CAM_BACK_LEFT/camera_info -r ~/input/topic_img_b/camera_info:=/nuscenes/CAM_BACK/camera_info -r ~/input/topic_img_br/camera_info:=/nuscenes/CAM_BACK_RIGHT/camera_info -r ~/output/boxes:=/tensorrt_bevformer/output/boxes -r ~/output_bboxes:=/nuscenes/bboxes'].
```

**Short:**
`symbol lookup error: /home/mfc/projects/autoware/install/autoware_tensorrt_bevformer/lib/libautoware_tensorrt_bevformer.so: undefined symbol: _ZNK5Eigen10MatrixBaseINS_6MatrixIdLi3ELi3ELi0ELi3ELi3EEEE7inverseEv`

Here are the full logs: https://gist.github.com/xmfcx/fc00d47e34547a60ca87bd8ce33a3103
Also shared my `ldd` results under: https://gist.github.com/xmfcx/fc00d47e34547a60ca87bd8ce33a3103?permalink_comment_id=5713959#gistcomment-5713959

xmfcx:
## about canbus

I've checked where this came from, where it gets published from, what it contains, where it is used.

### origin

In the context of this PR, it is expected that naveen-mcw's fork of ros2_dataset_bridge creates and publishes it [here](https://github.com/naveen-mcw/ros2_dataset_bridge/blob/7dcefa203353d3311d8c4c163e480c2eebb48b21/ros2_dataset_bridge/nuscenes_visualize_node.py#L139-L178).

And that's origin is [here from the fundamentalvision/BEVFormer repo](https://github.com/fundamentalvision/BEVFormer/blob/66b65f3a1f58caf0507cb2a971b9c0e7f842376c/tools/data_converter/nuscenes_converter.py#L154-L176).

### content

It's a float array of size 18.

It wasn't clear what it contained, I've added some log statements and here are the contents:
Here’s a compact summary and table for your `_get_can_bus_info()` return value:

| Index Range | Source Field Group | Description                           | Example Values (first log)          |
| ----------- | ------------------ | ------------------------------------- | ----------------------------------- |
| 0–2         | `pos`              | Position (X, Y, Z) in meters          | \[398.1693, 1141.3784, 0.0]         |
| 3–6         | `orientation`      | Quaternion (w, x, y, z)               | \[0.6116346, 0.0, 0.0, -0.7911404]  |
| 7–9         | `accel`            | Linear acceleration (X, Y, Z) in m/s² | \[-0.578977, 0.1205777, 9.6584166]  |
| 10–12       | `rotation_rate`    | Angular velocity (X, Y, Z) in rad/s   | \[-0.0415433, 0.0093976, 0.0574695] |
| 13–15       | `vel`              | Linear velocity (X, Y, Z) in m/s      | \[6.0879527, 0.0, 0.0]              |
| 16–17       | padding            | Reserved / unused                     | \[0.0, 0.0]                         |

### where is it used?

https://github.com/naveen-mcw/autoware.universe/blob/5165de5d9b6bba77375d862067ac9a646f82fd48/perception/autoware_tensorrt_bevformer/src/bevformer_inference_engine.cpp#L759C12-L780

> ```cpp
> void * d_can_bus = cuda_mem.allocate(sizeof(float) * can_bus.size());
> // ...
> CHECK_CUDA(cudaMemcpyAsync(
>   d_can_bus, can_bus.data(), sizeof(float) * can_bus.size(), cudaMemcpyHostToDevice, stream));
> ```

I am guessing the model is trained to have this in that format, I didn't look into that part more deeply.

### how is the node making use of this?

1. bevformer_node has a callback that takes in images, can_bus_msg and scene token.
2. first, the pos and orientation are re-read from the tf [here](https://github.com/naveen-mcw/autoware.universe/blob/5165de5d9b6bba77375d862067ac9a646f82fd48/perception/autoware_tensorrt_bevformer/src/bevformer_node.cpp#L499-L517)
3. then these are used to overwrite the first 7 values of the received  canbus [here](https://github.com/naveen-mcw/autoware.universe/blob/5165de5d9b6bba77375d862067ac9a646f82fd48/perception/autoware_tensorrt_bevformer/src/bevformer_node.cpp#L539-L548)
4. then here in the [bevformer_data_manager.cpp/processCanBus function](https://github.com/naveen-mcw/autoware.universe/blob/5165de5d9b6bba77375d862067ac9a646f82fd48/perception/autoware_tensorrt_bevformer/src/bevformer_data_manager.cpp#L80-L129)
   - overwriting of /tf **pos** and **orientation** happens
   - yaw angle is recalculated from the tf
   - yaw angle is assigned to `can_bus[16]` as radian and `can_bus[17]` as degrees.

### verdict

This new `canbus` message is unnecessary and cannot be accepted into Autoware. This is specific to the way this node was trained and its contents can be given to this node in proper formats.

The pos and orientation are already being derived from `/tf`. 
Yet all of the contents can be just transmitted with a [autoware_localization_msgs/msg/KinematicState.msg](https://github.com/autowarefoundation/autoware_msgs/blob/main/autoware_localization_msgs/msg/KinematicState.msg).

| KinematicState Field Path               | Value from canbus array |
| --------------------------------------- | ----------------- |
| `pose_with_covariance.pose.position`    | `pos`             |
| `pose_with_covariance.pose.orientation` | `orientation`     |
| `twist_with_covariance.twist.linear`    | `vel`             |
| `twist_with_covariance.twist.angular`   | `rotation_rate`   |
| `accel_with_covariance.accel.linear`    | `accel`           |

- You should publish this from the ros2_dataset_bridge instead of creating this new canbus message.
- Then the bevformer node can subscribe to this, take in the necessary parts.
- If model was trained to take in a `float[18]` array, package this in the node and do the inference.

## about scene_token

So far, only use case I could find to `scene_token` is that it is used to determine if the nuscenes scene has changed. So that it can discard things. This kind of behavior would only be useful for doing inference for the dataset. Not useful for normal vehicle usage.
So instead of this, maybe publish a [std_msgs/Bool message](https://github.com/ros2/common_interfaces/blob/rolling/std_msgs/msg/Bool.msg) to transmit that the scene has changed through a topic like "/reset_bevformer_history" ? This way we don't need a new message for this.

## some questions

- Why are you overwriting the pos and orientation from tf?
- Is the scene_token used for another purpose?

xmfcx:
Here I also created a simple flowchart on what to change.

**Before**

```mermaid
flowchart TB
    A1[ros2_dataset_bridge] -->|"autoware_internal_perception_msgs/msg/CanBusData.msg"| B1
    subgraph B1[bevformer_node]
        BN1[Receive message] --> BN2[Modify data]
        BN2 --> BN3[BEVFormerInferenceEngine::runInference]
    end
```

**After**

```mermaid
flowchart TB
    A2[ros2_dataset_bridge] -->|"autoware_localization_msgs/msg/KinematicState.msg"| B2
    subgraph B2[bevformer_node]
        BN4[Convert to float18 array] --> BN5[Modify data]
        BN5 --> BN6[BEVFormerInferenceEngine::runInference]
    end
```

naveen-mcw:
> 2025-08-09.23-15-44.mp4 
> I've got the following error when I ran the node and played the dataset:
> 
> ```
> [autoware_tensorrt_bevformer_node-1] [INFO 1754770551.536393563] [tensorrt_bevformer]: Image processing pipeline is now active. (checkInitialization() at /home/mfc/projects/autoware/src/universe/autoware_universe/perception/autoware_tensorrt_bevformer/src/bevformer_node.cpp:260)
> [autoware_tensorrt_bevformer_node-1] /home/mfc/projects/autoware/install/autoware_tensorrt_bevformer/lib/autoware_tensorrt_bevformer/autoware_tensorrt_bevformer_node: symbol lookup error: /home/mfc/projects/autoware/install/autoware_tensorrt_bevformer/lib/libautoware_tensorrt_bevformer.so: undefined symbol: _ZNK5Eigen10MatrixBaseINS_6MatrixIdLi3ELi3ELi0ELi3ELi3EEEE7inverseEv
> [ERROR] [autoware_tensorrt_bevformer_node-1]: process has died [pid 3624941, exit code 127, cmd '/home/mfc/projects/autoware/install/autoware_tensorrt_bevformer/lib/autoware_tensorrt_bevformer/autoware_tensorrt_bevformer_node --ros-args -r __node:=tensorrt_bevformer --params-file /tmp/launch_params_7_wu_beg --params-file /tmp/launch_params_ekg5knom --params-file /tmp/launch_params_db142xi4 --params-file /tmp/launch_params_q8tn_02d --params-file /tmp/launch_params_tllvycc2 --params-file /tmp/launch_params_c42yocaq --params-file /tmp/launch_params_o3ww1mhj --params-file /tmp/launch_params_ibgx4acb -r ~/input/topic_img_fl:=/nuscenes/CAM_FRONT_LEFT/image -r ~/input/topic_img_f:=/nuscenes/CAM_FRONT/image -r ~/input/topic_img_fr:=/nuscenes/CAM_FRONT_RIGHT/image -r ~/input/topic_img_bl:=/nuscenes/CAM_BACK_LEFT/image -r ~/input/topic_img_b:=/nuscenes/CAM_BACK/image -r ~/input/topic_img_br:=/nuscenes/CAM_BACK_RIGHT/image -r ~/input/can_bus:=/nuscenes/can_bus -r ~/input/scene_token:=/nuscenes/scene_tokens -r ~/input/topic_img_fl/camera_info:=/nuscenes/CAM_FRONT_LEFT/camera_info -r ~/input/topic_img_f/camera_info:=/nuscenes/CAM_FRONT/camera_info -r ~/input/topic_img_fr/camera_info:=/nuscenes/CAM_FRONT_RIGHT/camera_info -r ~/input/topic_img_bl/camera_info:=/nuscenes/CAM_BACK_LEFT/camera_info -r ~/input/topic_img_b/camera_info:=/nuscenes/CAM_BACK/camera_info -r ~/input/topic_img_br/camera_info:=/nuscenes/CAM_BACK_RIGHT/camera_info -r ~/output/boxes:=/tensorrt_bevformer/output/boxes -r ~/output_bboxes:=/nuscenes/bboxes'].
> ```
> 
> **Short:** `symbol lookup error: /home/mfc/projects/autoware/install/autoware_tensorrt_bevformer/lib/libautoware_tensorrt_bevformer.so: undefined symbol: _ZNK5Eigen10MatrixBaseINS_6MatrixIdLi3ELi3ELi0ELi3ELi3EEEE7inverseEv`
> 
> Here are the full logs: https://gist.github.com/xmfcx/fc00d47e34547a60ca87bd8ce33a3103 Also shared my `ldd` results under: https://gist.github.com/xmfcx/fc00d47e34547a60ca87bd8ce33a3103?permalink_comment_id=5713959#gistcomment-5713959

Hi @xmfcx ,

Thanks for the detailed report.

Issue: Missing Eigen3 library linking caused undefined symbol errors at runtime.

Fix: Added Eigen3::Eigen to target_link_libraries for both main and plugin libraries.

Verified: The undefined symbol is now properly resolved:

$ nm -D /home/naveen/autoware_bevdet/install/autoware_tensorrt_bevformer/lib/libautoware_tensorrt_bevformer.so | grep inverse
00000000009bccb8 W *ZNK5Eigen10MatrixBaseINS*6MatrixIdLi3ELi3ELi0ELi3ELi3EEEE7inverseEv

Could you please test the launch to confirm the crash is resolved?

naveen-mcw:
> ## about canbus
> I've checked where this came from, where it gets published from, what it contains, where it is used.
> 
> ### origin
> In the context of this PR, it is expected that naveen-mcw's fork of ros2_dataset_bridge creates and publishes it [here](https://github.com/naveen-mcw/ros2_dataset_bridge/blob/7dcefa203353d3311d8c4c163e480c2eebb48b21/ros2_dataset_bridge/nuscenes_visualize_node.py#L139-L178).
> 
> And that's origin is [here from the fundamentalvision/BEVFormer repo](https://github.com/fundamentalvision/BEVFormer/blob/66b65f3a1f58caf0507cb2a971b9c0e7f842376c/tools/data_converter/nuscenes_converter.py#L154-L176).
> 
> ### content
> It's a float array of size 18.
> 
> It wasn't clear what it contained, I've added some log statements and here are the contents: Here’s a compact summary and table for your `_get_can_bus_info()` return value:
> 
> Index Range	Source Field Group	Description	Example Values (first log)
> 0–2	`pos`	Position (X, Y, Z) in meters	[398.1693, 1141.3784, 0.0]
> 3–6	`orientation`	Quaternion (w, x, y, z)	[0.6116346, 0.0, 0.0, -0.7911404]
> 7–9	`accel`	Linear acceleration (X, Y, Z) in m/s²	[-0.578977, 0.1205777, 9.6584166]
> 10–12	`rotation_rate`	Angular velocity (X, Y, Z) in rad/s	[-0.0415433, 0.0093976, 0.0574695]
> 13–15	`vel`	Linear velocity (X, Y, Z) in m/s	[6.0879527, 0.0, 0.0]
> 16–17	padding	Reserved / unused	[0.0, 0.0]
> ### where is it used?
> https://github.com/naveen-mcw/autoware.universe/blob/5165de5d9b6bba77375d862067ac9a646f82fd48/perception/autoware_tensorrt_bevformer/src/bevformer_inference_engine.cpp#L759C12-L780
> 
> > ```c++
> > void * d_can_bus = cuda_mem.allocate(sizeof(float) * can_bus.size());
> > // ...
> > CHECK_CUDA(cudaMemcpyAsync(
> >   d_can_bus, can_bus.data(), sizeof(float) * can_bus.size(), cudaMemcpyHostToDevice, stream));
> > ```
> 
> I am guessing the model is trained to have this in that format, I didn't look into that part more deeply.
> 
> ### how is the node making use of this?
> 1. bevformer_node has a callback that takes in images, can_bus_msg and scene token.
> 2. first, the pos and orientation are re-read from the tf [here](https://github.com/naveen-mcw/autoware.universe/blob/5165de5d9b6bba77375d862067ac9a646f82fd48/perception/autoware_tensorrt_bevformer/src/bevformer_node.cpp#L499-L517)
> 3. then these are used to overwrite the first 7 values of the received  canbus [here](https://github.com/naveen-mcw/autoware.universe/blob/5165de5d9b6bba77375d862067ac9a646f82fd48/perception/autoware_tensorrt_bevformer/src/bevformer_node.cpp#L539-L548)
> 4. then here in the [bevformer_data_manager.cpp/processCanBus function](https://github.com/naveen-mcw/autoware.universe/blob/5165de5d9b6bba77375d862067ac9a646f82fd48/perception/autoware_tensorrt_bevformer/src/bevformer_data_manager.cpp#L80-L129)
>    
>    * overwriting of /tf **pos** and **orientation** happens
>    * yaw angle is recalculated from the tf
>    * yaw angle is assigned to `can_bus[16]` as radian and `can_bus[17]` as degrees.
> 
> ### verdict
> This new `canbus` message is unnecessary and cannot be accepted into Autoware. This is specific to the way this node was trained and its contents can be given to this node in proper formats.
> 
> The pos and orientation are already being derived from `/tf`. Yet all of the contents can be just transmitted with a [autoware_localization_msgs/msg/KinematicState.msg](https://github.com/autowarefoundation/autoware_msgs/blob/main/autoware_localization_msgs/msg/KinematicState.msg).
> 
> KinematicState Field Path	Value from canbus array
> `pose_with_covariance.pose.position`	`pos`
> `pose_with_covariance.pose.orientation`	`orientation`
> `twist_with_covariance.twist.linear`	`vel`
> `twist_with_covariance.twist.angular`	`rotation_rate`
> `accel_with_covariance.accel.linear`	`accel`
> * You should publish this from the ros2_dataset_bridge instead of creating this new canbus message.
> * Then the bevformer node can subscribe to this, take in the necessary parts.
> * If model was trained to take in a `float[18]` array, package this in the node and do the inference.
> 
> ## about scene_token
> So far, only use case I could find to `scene_token` is that it is used to determine if the nuscenes scene has changed. So that it can discard things. This kind of behavior would only be useful for doing inference for the dataset. Not useful for normal vehicle usage. So instead of this, maybe publish a [std_msgs/Bool message](https://github.com/ros2/common_interfaces/blob/rolling/std_msgs/msg/Bool.msg) to transmit that the scene has changed through a topic like "/reset_bevformer_history" ? This way we don't need a new message for this.
> 
> ## some questions
> * Why are you overwriting the pos and orientation from tf?
> * Is the scene_token used for another purpose?


Hi @xmfcx,

Thank you for your detailed analysis and suggestions regarding the CAN bus implementation.

## Implementation Update

Regarding CAN bus, we have followed your suggested approach and implemented `autoware_localization_msgs::msg::KinematicState` in both `ros2_dataset_bridge` and `autoware_tensorrt_bevformer`. The implementation is working well now. Here is a [sample log](https://github.com/user-attachments/files/21711985/echo_canbus.txt) published from CAN bus using the `autoware_localization_msgs::msg::KinematicState` message type.

The recent commit from our side uses the `autoware_localization_msgs::msg::KinematicState` message type for CAN bus instead of creating a new message type. Kindly review from your side and let us know if any changes are required.

## Answers to Your Questions

### 1. Why are you overwriting the pos and orientation from tf?

We are following the **official BEVFormer implementation** approach from the fundamentalvision repository.

The BEVFormer model was trained using this specific method where:

- **Position and orientation** come from tf (more accurate and synchronized with camera frames)
- **Motion data** (velocity, acceleration, rotation rates) come from CAN bus (authentic vehicle dynamics)

As stated in the official BEVFormer code comment: *"the nuScenes provides the rotation and translation of each sample, which is more accurate than we obtained from can bus."*

**Why this works better:**
- tf data is perfectly aligned with camera timestamps
- CAN bus data has timing issues that require interpolation
- This hybrid approach gives the best accuracy for the model

This is the standard implementation used by the BEVFormer model.

**Reference**: [[BEVFormer Official Repository - CAN Bus Documentation](https://github.com/fundamentalvision/BEVFormer/blob/master/docs/can_bus.ipynb)](https://github.com/fundamentalvision/BEVFormer/blob/master/docs/can_bus.ipynb)

### 2. Is the scene_token used for another purpose?

The primary purpose of the scene token is to determine whether to use the previous BEV features or not and control the CAN bus relative pose encoding.

**How it works:**

1. If the scene token changes, a flag called `use_prev_bev` is set accordingly
2. This flag is passed to the model to decide whether to use BEV embedding(BEV features of a sample) from the previous frame
3. CAN bus position and yaw angle modifications: Based on the `use_prev_bev` flag, the CAN bus data is processed differently:

   **When scene changes (`use_prev_bev = false`):**
   - Position values `can_bus[:3]` are set to `[0, 0, 0]` (reference frame)
   - Yaw angle `can_bus[-1]` is set to `0` (reference angle)

   **Subsequent frames (`use_prev_bev = true`):**
   - Position values `can_bus[:3]` contain relative deltas (`current_pos - previous_pos`)
   - Yaw angle `can_bus[-1]` contains relative rotation (`current_angle - previous_angle`)

## Questions for your consideration

1. Would you prefer us to implement the boolean message type for `/reset_bevformer_history` as you suggested?
2. Or should we simplify this for real-world usage by using a simple `is_first_frame` variable that defaults `use_prev_bev` to `false` for the first frame and `true` for all others?

---

We appreciate your thorough review and guidance on implementing proper Autoware message standards.

xmfcx:
> Would you prefer us to implement the boolean message type for /reset_bevformer_history as you suggested? Or should we simplify this for real-world usage by using a simple `is_first_frame` variable that defaults `use_prev_bev` to `false` for the first frame and `true` for all others?

I believe that `/reset_bevformer_history` clearly conveys the action, and since it's `false` by default, it seems like a good approach.

For the linking error, I will test as soon as possible thanks :+1:

Dhinesh-MCW:
> > Would you prefer us to implement the boolean message type for /reset_bevformer_history as you suggested? Or should we simplify this for real-world usage by using a simple `is_first_frame` variable that defaults `use_prev_bev` to `false` for the first frame and `true` for all others?
> 
> I believe that `/reset_bevformer_history` clearly conveys the action, and since it's `false` by default, it seems like a good approach.
> 
> For the linking error, I will test as soon as possible thanks 👍

Hi @xmfcx ,
As per your suggestion, we have added the new topic `/input/reset_bevformer_history`, which gives `false` by default. If any scene changes occur, it will return `true` for that instance alone.
Since we have followed this approach, the supporting [PR](https://github.com/autowarefoundation/autoware_internal_msgs/pull/78) raised for the scene info and can bus will be closed.

Thank you for your insights, and we look forward to your further review.

naveen-mcw:
Hi @xmfcx,

As discussed, the current approach works as follows:

- For the first frame, the use_prev_bev flag is set to false, ensuring the model does not use previous BEV features. From the second frame onwards, the flag is set to true, allowing the model to continue using previous BEV features.
- The CAN bus position encoding behaves accordingly, as described in the official BEVFormer [documentation](https://github.com/fundamentalvision/BEVFormer/blob/master/docs/can_bus.ipynb) on CAN bus.

From our observations, there are no negative impacts on predictions or performance during inference — everything appears to be functioning as expected.

Kindly review and let us know if any further changes are required.

naveen-mcw:
Hi Everyone,
While fixing DCO issues in my PR, I mistakenly force-pushed during a rebase, affecting some commits. It’s now fixed with all 34 feature commits rebased on a clean base. Please ignore earlier notifications.
Thanks.

xmfcx:
- [remove redundant calls that are already performed by autoware_package()](https://github.com/autowarefoundation/autoware_universe/pull/11076/commits/5eb00ab77fc29eee6fb7a2ead73c34a230d3e2cd)
  - Here I removed things that are already done by autoware_package() here: https://github.com/autowarefoundation/autoware_cmake/blob/6829fde0b7f8ef6b0155f1902022b7f6d89b068b/autoware_cmake/cmake/autoware_package.cmake#L18-L19 and by ament_auto_find_build_dependencies() which includes packages from package.xml dependencies.
- [use eigen3_cmake_module and utilize target_include_directories](https://github.com/autowarefoundation/autoware_universe/pull/11076/commits/af1a0d2bec817350ed414b4f9aefc3f062ae5f95)
  - Then here, I removed the old style generalized `include_directories` and replace it with modern targeted `target_include_directories` calls for both targets. Also removed `Eigen3::Eigen` from `target_link_libraries` because these will be handled by `eigen3_cmake_module` package.
  - Also added `target_link_options(${PROJECT_NAME} PRIVATE -Wl,--no-undefined)` to let build break if linking error occurs. With this, we don't need to run the package to find out if a linking error will happen.
- [explicitly add #include <Eigen/LU> for inverse](https://github.com/autowarefoundation/autoware_universe/pull/11076/commits/c6941dd2f622c6178c4ca37ad40689d1c4627a02)
  - Here is the actual fix. The `MatrixBase::inverse()` operation resides in `#include <Eigen/LU>` but it was not included anywhere. `#include <Eigen/Dense>` includes most of these but you only included `<Eigen/Core>` and `<Eigen/Geometry>` in most `.hpp` files.
  - I recommend checking https://eigen.tuxfamily.org/dox-devel/group__QuickRefPage.html "Modules and Header files".
  - Because the header was missing, the linking had problems.

With these fixes, now it compiles and works 👍

xmfcx:
Could you make sure [clang-tidy-differential](https://github.com/autowarefoundation/autoware_universe/actions/runs/17080638817/job/48434476008?pr=11076) workflow passes?

Similarly spell-check too.

image

https://github.com/autowarefoundation/autoware-spell-check-dict/actions/workflows/add-word.yaml

- You can add it by triggering this action.
- This action will create a PR [here](https://github.com/autowarefoundation/autoware-spell-check-dict/pulls) for that word.
- In the PR description explain what the word stands for. 
- Then you can approve it and it will be in the dictionary.

You can ignore the DCO for now.

xmfcx:
[fixes.yaml](https://gist.github.com/xmfcx/2a81efcc22806b39e0ca3c00490ca98e) here are the fixes proposed by the clang-tidy-differential workfow.

Some can be fixed by you but many come from local cuda installation.

Example:
```yaml
- BuildDirectory: /__w/autoware_universe/autoware_universe/build/autoware_tensorrt_bevformer
  DiagnosticMessage:
    FileOffset: 169243
    FilePath: /usr/local/cuda/include/cuda_fp16.h
    Message: must specify at least one argument for '...' parameter of variadic macro
    Replacements: []
```

@veqcc -san could you tell us how to ignore cuda folders from this workflow?

xmfcx:
## Test video with default FP16

https://github.com/user-attachments/assets/fe57e699-75a1-4a45-8aae-9de196a1d721

## Latency

```
[autoware_tensorrt_bevformer_node-1] [INFO 1755636397.679645972] [tensorrt_bevformer]: -------------------------------------------------- (callback() at /home/mfc/projects/autoware/src/universe/autoware_universe/perception/autoware_tensorrt_bevformer/src/bevformer_node.cpp:674)
[autoware_tensorrt_bevformer_node-1] [INFO 1755636397.987416486] [tensorrt_bevformer]: Pre-processing : 238 ms (callback() at /home/mfc/projects/autoware/src/universe/autoware_universe/perception/autoware_tensorrt_bevformer/src/bevformer_node.cpp:599)
[autoware_tensorrt_bevformer_node-1] [INFO 1755636398.068334785] [tensorrt_bevformer]: TRT Engine : 60.763 ms (runInference() at /home/mfc/projects/autoware/src/universe/autoware_universe/perception/autoware_tensorrt_bevformer/src/bevformer_inference_engine.cpp:812)
[autoware_tensorrt_bevformer_node-1] [INFO 1755636398.081565723] [tensorrt_bevformer]: Inference : 94 ms (callback() at /home/mfc/projects/autoware/src/universe/autoware_universe/perception/autoware_tensorrt_bevformer/src/bevformer_node.cpp:627)
[autoware_tensorrt_bevformer_node-1] [INFO 1755636398.093444878] [tensorrt_bevformer]: Post-processing : 11 ms (callback() at /home/mfc/projects/autoware/src/universe/autoware_universe/perception/autoware_tensorrt_bevformer/src/bevformer_node.cpp:665)
[autoware_tensorrt_bevformer_node-1] [INFO 1755636398.094219414] [tensorrt_bevformer]: Detections: 188 (callback() at /home/mfc/projects/autoware/src/universe/autoware_universe/perception/autoware_tensorrt_bevformer/src/bevformer_node.cpp:673)
```

Random Frame:

| Stage            | Latency (ms) |
|------------------|--------------|
| Pre-processing   | 238          |
| TRT Engine       | 60       |
| Inference        | 94           |
| Post-processing  | 11           |
| **Total**        | **403**  |

Single frame taking approximately **403ms** on an RTX3090 so it is fairly slow. It's running at about **2.5 Hz**.
Is this expected? Or is my configuration wrong?

xmfcx:
## Dataset playback at 1Hz

https://github.com/user-attachments/assets/b8365505-3907-4184-98b3-f3bc232c0a67

Also even if I play the data at half the speed of inference, the inference results don't match with the ground truth. I think something is wrong with sync or transformations.

Dhinesh-MCW:
Hi @xmfcx,

As per your comments, we have resolved the following issues:

- Removed the redundant dependency declarations from CMakeLists.txt.
- Created a PR ([Link](https://github.com/autowarefoundation/autoware-spell-check-dict/pull/46)) to address the spell check, kindly approve it to pass the spell check.
- Fixed the clang-related issues. (The remaining issues are related to your local CUDA setup.)

Regarding the synchronization issue, we have verified on our side that the output results are properly synchronized. We also implemented a few fixes, tested them, and confirmed that the results are now synchronizing correctly.

I would suggest you to review this from your side as well.

naveen-mcw:
> ## Test video with default FP16
>  2025-08-19.23-46-05.mp4 
> ## Latency
> ```
> [autoware_tensorrt_bevformer_node-1] [INFO 1755636397.679645972] [tensorrt_bevformer]: -------------------------------------------------- (callback() at /home/mfc/projects/autoware/src/universe/autoware_universe/perception/autoware_tensorrt_bevformer/src/bevformer_node.cpp:674)
> [autoware_tensorrt_bevformer_node-1] [INFO 1755636397.987416486] [tensorrt_bevformer]: Pre-processing : 238 ms (callback() at /home/mfc/projects/autoware/src/universe/autoware_universe/perception/autoware_tensorrt_bevformer/src/bevformer_node.cpp:599)
> [autoware_tensorrt_bevformer_node-1] [INFO 1755636398.068334785] [tensorrt_bevformer]: TRT Engine : 60.763 ms (runInference() at /home/mfc/projects/autoware/src/universe/autoware_universe/perception/autoware_tensorrt_bevformer/src/bevformer_inference_engine.cpp:812)
> [autoware_tensorrt_bevformer_node-1] [INFO 1755636398.081565723] [tensorrt_bevformer]: Inference : 94 ms (callback() at /home/mfc/projects/autoware/src/universe/autoware_universe/perception/autoware_tensorrt_bevformer/src/bevformer_node.cpp:627)
> [autoware_tensorrt_bevformer_node-1] [INFO 1755636398.093444878] [tensorrt_bevformer]: Post-processing : 11 ms (callback() at /home/mfc/projects/autoware/src/universe/autoware_universe/perception/autoware_tensorrt_bevformer/src/bevformer_node.cpp:665)
> [autoware_tensorrt_bevformer_node-1] [INFO 1755636398.094219414] [tensorrt_bevformer]: Detections: 188 (callback() at /home/mfc/projects/autoware/src/universe/autoware_universe/perception/autoware_tensorrt_bevformer/src/bevformer_node.cpp:673)
> ```
> 
> Random Frame:
> 
> Stage	Latency (ms)
> Pre-processing	238
> TRT Engine	60
> Inference	94
> Post-processing	11
> **Total**	**403**
> Single frame taking approximately **403ms** on an RTX3090 so it is fairly slow. It's running at about **2.5 Hz**. Is this expected? Or is my configuration wrong?


Hi @xmfcx,

This is expected, but kindly ensure that you are not running any other processes or jobs in the background, since the observed latency on our side is as follows:

## Log Output

```
[autoware_tensorrt_bevformer_node-1] [INFO] [1755679736.819579299] [tensorrt_bevformer]: --------------------------------------------------
[autoware_tensorrt_bevformer_node-1] [INFO] [1755679737.368426911] [tensorrt_bevformer]: Pre-processing : 163 ms
[autoware_tensorrt_bevformer_node-1] [INFO] [1755679737.488555234] [tensorrt_bevformer]: TRT Engine : 91.130 ms
[autoware_tensorrt_bevformer_node-1] [INFO] [1755679737.497347436] [tensorrt_bevformer]: Inference : 128 ms
[autoware_tensorrt_bevformer_node-1] [INFO] [1755679737.501785624] [tensorrt_bevformer]: Post-processing : 4 ms
[autoware_tensorrt_bevformer_node-1] [INFO] [1755679737.503259713] [tensorrt_bevformer]: Detections: 170
[autoware_tensorrt_bevformer_node-1] [INFO] [1755679737.503295432] [tensorrt_bevformer]: --------------------------------------------------
[autoware_tensorrt_bevformer_node-1] [INFO] [1755679738.077615339] [tensorrt_bevformer]: Pre-processing : 180 ms
[autoware_tensorrt_bevformer_node-1] [INFO] [1755679738.198378253] [tensorrt_bevformer]: TRT Engine : 91.086 ms
[autoware_tensorrt_bevformer_node-1] [INFO] [1755679738.207108628] [tensorrt_bevformer]: Inference : 129 ms
[autoware_tensorrt_bevformer_node-1] [INFO] [1755679738.211580066] [tensorrt_bevformer]: Post-processing : 4 ms
[autoware_tensorrt_bevformer_node-1] [INFO] [1755679738.213104674] [tensorrt_bevformer]: Detections: 174
[autoware_tensorrt_bevformer_node-1] [INFO] [1755679738.213120564] [tensorrt_bevformer]: --------------------------------------------------
[autoware_tensorrt_bevformer_node-1] [INFO] [1755679738.766672272] [tensorrt_bevformer]: Pre-processing : 169 ms
[autoware_tensorrt_bevformer_node-1] [INFO] [1755679738.890490454] [tensorrt_bevformer]: TRT Engine : 91.181 ms
[autoware_tensorrt_bevformer_node-1] [INFO] [1755679738.899217768] [tensorrt_bevformer]: Inference : 132 ms
[autoware_tensorrt_bevformer_node-1] [INFO] [1755679738.903631180] [tensorrt_bevformer]: Post-processing : 4 ms
[autoware_tensorrt_bevformer_node-1] [INFO] [1755679738.905090394] [tensorrt_bevformer]: Detections: 168
[autoware_tensorrt_bevformer_node-1] [INFO] [1755679738.905104818] [tensorrt_bevformer]: --------------------------------------------------
[autoware_tensorrt_bevformer_node-1] [INFO] [1755679739.536422866] [tensorrt_bevformer]: Pre-processing : 162 ms
[autoware_tensorrt_bevformer_node-1] [INFO] [1755679739.657106048] [tensorrt_bevformer]: TRT Engine : 91.051 ms
[autoware_tensorrt_bevformer_node-1] [INFO] [1755679739.665697550] [tensorrt_bevformer]: Inference : 129 ms
[autoware_tensorrt_bevformer_node-1] [INFO] [1755679739.670034740] [tensorrt_bevformer]: Post-processing : 4 ms
[autoware_tensorrt_bevformer_node-1] [INFO] [1755679739.671468686] [tensorrt_bevformer]: Detections: 162
[autoware_tensorrt_bevformer_node-1] [INFO] [1755679739.671484509] [tensorrt_bevformer]: --------------------------------------------------
```

## Performance Summary

| Stage           | Latency   |
|-----------------|-----------|
| Pre-processing  | 165 ms    |
| TRT Engine      | 91.283 ms |
| Inference       | 134 ms    |
| Post-processing | 4 ms      |
| Detections      | 163       |

**Note:** The `Inference` time already includes the TRT Engine time as well as the additional overhead for handling temporal data.

So, the **total end-to-end latency is ~303 ms** on our RTX 2080 Ti machine.

As you can see, the preprocessing and post-processing times are lesser than yours for a random frame.

The model is taken from **DerryHub**, kindly take a look at our BEVFormer [discussion page](https://github.com/orgs/autowarefoundation/discussions/5904#discussioncomment-13918364) for further details.

launch/tier4_perception_launch/launch/object_recognition/detection/detection.launch.xml
```diff
@@ -6,6 +6,7 @@
   <!-- Pipeline junctions -->
   <arg name="mode">
     <choice value="camera"/>
+    <choice value="camera_lidar"/>
     <choice value="camera_lidar_radar_fusion"/>
     <choice value="camera_lidar_fusion"/>
```
mitsudome-r: We already have camera_lidar_fusion, do we need to set another choice here as camera_lidar?
Also, have you already tested the full perception pipeline using AWSIM or any other dataset?

perception/autoware_tensorrt_bevformer/package.xml
```diff
@@ -0,0 +1,42 @@
...
+  <maintainer email="naveen.sathiyaseelan@multicorewareinc.com">Naveen Sathiyaseelan</maintainer>
+  <maintainer email="dhinesh.panneerselvam@multicorewareinc.com">Dhinesh Panneerselvam</maintainer>
+  <maintainer email="rahul.sundar@multicorewareinc.com">Rahul Gandhi Sundar</maintainer>
+  <license>Apache License 2.0</license>
+
+  <buildtool_depend>ament_cmake_auto</buildtool_depend>
+  <buildtool_depend>autoware_cmake</buildtool_depend>
+  <buildtool_depend>eigen3_cmake_module</buildtool_depend>
+
+  <depend>ament_index_cpp</depend>
```
mitsudome-r, line 19: Could you add dependency to autoware_cuda_dependency_meta?
We use it to control packages that depend on CUDA.

```suggestion
  <depend>autoware_cuda_dependency_meta</depend>
  <depend>autoware-index_cpp</depend>
```

launch/tier4_perception_launch/launch/object_recognition/detection/detection.launch.xml
technolojin: I could suggest to make a separated mode.
for example, `camera_lidar_bev`

Dhinesh-MCW: I have renamed the mode to camera_lidar_bev from camer_lidar as per your suggestions.
Thank You

perception/autoware_tensorrt_bevformer/package.xml
Dhinesh-MCW: I have added autoware_cuda_dependency_meta to package.xml.

perception/autoware_tensorrt_bevformer/TensorRT/.gitignore
```diff
@@ -0,0 +1 @@
+.idea
```
mitsudome-r: I believe this is for your local environment. Please remove this file if this ins't applicable to OSS users.

launch/tier4_perception_launch/launch/perception.launch.xml
```diff
@@ -118,6 +119,7 @@
   <arg name="image_raw8" default="/sensing/camera/camera8/image_rect_color"/>
   <arg name="camera_info8" default="/sensing/camera/camera8/camera_info"/>
   <arg name="detection_rois8" default="/perception/object_recognition/detection/rois8"/>
+  <arg name="can_bus" default="/sensing/can_bus"/>
```
mitsudome-r: I believe we no longer use /sensing/can_bus anymore so we can remove this line.

mitsudome-r: I don't think we have tested any tests with perception.launch.
Maybe we could focus on merging the autoware_tensorrt_bevformer package as a standalone package for now, then update the perception.launch.xml in a separate PR. It makes us easier to review the PR.

perception/autoware_tensorrt_bevformer/include/autoware/tensorrt_bevformer/bevformer_node.hpp
```diff
@@ -0,0 +1,211 @@
...
+  rclcpp::Subscription<sensor_msgs::msg::CameraInfo>::SharedPtr sub_bl_caminfo_;
+  rclcpp::Subscription<sensor_msgs::msg::CameraInfo>::SharedPtr sub_br_caminfo_;
+
+  // Initialization flags
+  std::vector<bool>
+    caminfo_received_;  // Flag indicating if camera info has been received for each camera
+  bool camera_info_received_flag_ = false;
+
+  // can_bus subscription
+  message_filters::Subscriber<autoware_localization_msgs::msg::KinematicState> sub_can_bus_;
```
mitsudome-r: I suggest to update the subscriber name to match with what it is subscribing. 

```suggestion
  message_filters::Subscriber<autoware_localization_msgs::msg::KinematicState> sub_kinematic_state;

```

mitsudome-r (commented):
I think we are almost ready to merge this, but I made few suggestions.

Also, it would be nice if you could consider moving the header files under includes files to be moved under source folder unless you want them to be used by other packages.
Include files under include directories are meant for exported include files.

Ref: https://docs.ros.org/en/rolling/How-To-Guides/Ament-CMake-Documentation.html
image

perception/autoware_tensorrt_bevformer/TensorRT/.gitignore
naveen-mcw: Got it, thanks for pointing that out. We'll remove it.

launch/tier4_perception_launch/launch/perception.launch.xml
naveen-mcw: Sure, that makes sense.
We'll remove the line from the perception launch file and proceed with bevformer as a standalone package for now.

perception/autoware_tensorrt_bevformer/include/autoware/tensorrt_bevformer/bevformer_node.hpp
naveen-mcw: Thanks for pointing that out. We'll update the subscriber name to match the subscribed topic (sub_kinematic_state).

xmfcx:
Could you fix the cppcheck CI?

https://github.com/autowarefoundation/autoware_universe/actions/runs/17463938744/job/49617150237?pr=11076#step:13:7

naveen-mcw:
Hi @xmfcx,

We have solved the CPP CI Check. The checks for build-test-tidy-pr are failing due to IP Connection time out, please check it from your side. FYI, we moved the headers to src folder as per @mitsudome-r suggestion.

amadeuszsz:
@xmfcx @naveen-mcw 

For TensorRT plugins, we use [dedicated package](https://github.com/autowarefoundation/autoware_universe/tree/main/perception/autoware_tensorrt_plugins). Do you think it is possible to refactor and move plugins to this package? There might be some ops/plugins already implemented. This way we can avoid boilerplate code. If so, please follow naming convention.

naveen-mcw:
Hi @mitsudome-r @xmfcx,

Could you please share your thoughts.

xmfcx:
## Replicate clang-tidy checks on local machine

```console
sudo apt-get install fd-find

cd autoware/

sed -i '/- -Wno-c11-extensions/a\  - -I/usr/include/c++/11\n  - -I/usr/include/x86_64-linux-gnu/c++/11' .clang-tidy-ci

run-clang-tidy -p build/   -config="$(cat .clang-tidy-ci)"   -export-fixes clang-tidy-fixes.yaml   -j $(nproc)   $(fdfind -e cpp -e hpp --full-path "/autoware_tensorrt_bevformer/")   > clang-tidy-report.log
```

- reference: https://github.com/autowarefoundation/autoware/pull/5543#issuecomment-2533325979 cc. @veqcc 

## CUDA system header issue

Not sure why it worked but it got fixed when I did this:

### Before

```cmake
  target_include_directories(${TENSORRT_OPS_TARGET}
    PRIVATE
      ${CMAKE_CURRENT_SOURCE_DIR}/TensorRT/common
    SYSTEM PRIVATE
      ${CUDA_INCLUDE_DIRS}
      ${OpenCV_INCLUDE_DIRS}
  )
```

### After

```cmake
  target_include_directories(${TENSORRT_OPS_TARGET}
    PRIVATE
      ${CMAKE_CURRENT_SOURCE_DIR}/TensorRT/common
  )
  # Separate SYSTEM include to avoid warnings in TensorRT headers with clang-tidy
  target_include_directories(${TENSORRT_OPS_TARGET}
    SYSTEM PRIVATE
      ${CUDA_INCLUDE_DIRS}
      ${OpenCV_INCLUDE_DIRS}
  )
```

I copied this pattern from here:
https://github.com/autowarefoundation/autoware_universe/blob/c542d8950d6976526d97426d41ea4ac10e8ba580/perception/autoware_lidar_centerpoint/CMakeLists.txt#L125-L130

But not sure if they did it for the same reason 🤷‍♂️

## Using `autoware_tensorrt_plugins` package

This PR has been open for too long, let's merge it if it works and create an issue to make it use the `autoware_tensorrt_plugins` as the next task. cc. @amadeuszsz @naveen-mcw

xmfcx:
With my last commit, here are the only errors left:

<details>
  <summary>🖱️Click here to expand🔛</summary>

```yaml
Diagnostics:
- BuildDirectory: ''
  DiagnosticMessage:
    FileOffset: 0
    FilePath: ''
    Message: unknown warning option '-Wno-maybe-uninitialized'; did you mean '-Wno-uninitialized'?
    Replacements: []
  DiagnosticName: clang-diagnostic-unknown-warning-option
  Level: Error
- BuildDirectory: /__w/autoware_universe/autoware_universe/build/autoware_tensorrt_bevformer
  DiagnosticMessage:
    FileOffset: 2973
    FilePath: /__w/autoware_universe/autoware_universe/perception/autoware_tensorrt_bevformer/TensorRT/plugin/multi_scale_deformable_attn/multiScaleDeformableAttnPlugin.cpp
    Message: function previously declared with an implicit exception specification
      redeclared with an explicit exception specification
    Replacements: []
  DiagnosticName: clang-diagnostic-implicit-exception-spec-mismatch
  Level: Error
  Notes:
  - FileOffset: 1931
    FilePath: /__w/autoware_universe/autoware_universe/perception/autoware_tensorrt_bevformer/TensorRT/plugin/multi_scale_deformable_attn/multiScaleDeformableAttnPlugin.h
    Message: previous declaration is here
    Replacements: []
- BuildDirectory: ''
  DiagnosticMessage:
    FileOffset: 0
    FilePath: ''
    Message: unknown warning option '-Wno-maybe-uninitialized'; did you mean '-Wno-uninitialized'?
    Replacements: []
  DiagnosticName: clang-diagnostic-unknown-warning-option
  Level: Error
- BuildDirectory: /__w/autoware_universe/autoware_universe/build/autoware_tensorrt_bevformer
  DiagnosticMessage:
    FileOffset: 3415
    FilePath: /__w/autoware_universe/autoware_universe/perception/autoware_tensorrt_bevformer/TensorRT/plugin/modulated_deformable_conv2d/modulatedDeformableConv2dPlugin.cpp
    Message: function previously declared with an implicit exception specification
      redeclared with an explicit exception specification
    Replacements: []
  DiagnosticName: clang-diagnostic-implicit-exception-spec-mismatch
  Level: Error
  Notes:
  - FileOffset: 2062
    FilePath: /__w/autoware_universe/autoware_universe/perception/autoware_tensorrt_bevformer/TensorRT/plugin/modulated_deformable_conv2d/modulatedDeformableConv2dPlugin.h
    Message: previous declaration is here
    Replacements: []
- BuildDirectory: ''
  DiagnosticMessage:
    FileOffset: 0
    FilePath: ''
    Message: unknown warning option '-Wno-maybe-uninitialized'; did you mean '-Wno-uninitialized'?
    Replacements: []
  DiagnosticName: clang-diagnostic-unknown-warning-option
  Level: Error
- BuildDirectory: ''
  DiagnosticMessage:
    FileOffset: 0
    FilePath: ''
    Message: unknown warning option '-Wno-maybe-uninitialized'; did you mean '-Wno-uninitialized'?
    Replacements: []
  DiagnosticName: clang-diagnostic-unknown-warning-option
  Level: Error
MainSourceFile: ''
```
</details>

xmfcx:
Could you remove the suppression of at least `-Werror=maybe-uninitialized` and update the code to initialize them correctly?

naveen-mcw:
Hi @xmfcx,

Thank you for pointing out the issue and for the changes you made to resolve the CUDA system header problem.

We have addressed the -Werror=maybe-uninitialized issue along with other warnings in the plugins. As suggested, we also verified the changes locally using Clang-Tidy checks. Please have a look.

perception/autoware_tensorrt_bevformer/launch/bevformer.launch.xml
```diff
@@ -0,0 +1,70 @@
...
+  <arg name="input/img_front_left/camera_info" default="/nuscenes/CAM_FRONT_LEFT/camera_info"/>
+  <arg name="input/img_front/camera_info" default="/nuscenes/CAM_FRONT/camera_info"/>
+  <arg name="input/img_front_right/camera_info" default="/nuscenes/CAM_FRONT_RIGHT/camera_info"/>
+  <arg name="input/img_back_left/camera_info" default="/nuscenes/CAM_BACK_LEFT/camera_info"/>
+  <arg name="input/img_back/camera_info" default="/nuscenes/CAM_BACK/camera_info"/>
+  <arg name="input/img_back_right/camera_info" default="/nuscenes/CAM_BACK_RIGHT/camera_info"/>
+
+  <!-- Engine and ONNX paths -->
+  <arg name="onnx_file" default="$(env HOME)/autoware_data/tensorrt_bevformer/bevformer_small.onnx"/>
+  <arg name="engine_file" default="$(env HOME)/autoware_data/tensorrt_bevformer/bevformer_small.engine"/>
```
technolojin: please use `data_path` variable

launch/tier4_perception_launch/launch/object_recognition/detection/detection.launch.xml
```diff
@@ -6,6 +6,7 @@
   <!-- Pipeline junctions -->
   <arg name="mode">
     <choice value="camera"/>
+    <choice value="camera_lidar_bev"/>
```
technolojin: The bevformer detector does not use lidar.
Therefore, it should be in the camera mode.

launch/tier4_perception_launch/launch/object_recognition/detection/detection.launch.xml
```diff
@@ -318,6 +319,30 @@
       <arg name="output/objects" value="$(var output/objects)"/>
     </include>
   </group>
+  <group scoped="false" if="$(eval '&quot;$(var mode)&quot;==&quot;camera_lidar_bev&quot;')">
+    <!-- Camera Lidar-based bev detectors-->
+    <include file="$(find-pkg-share tier4_perception_launch)/launch/object_recognition/detection/detector/cam_lidar_bev_detector.launch.xml">
```
technolojin: This detector is the second multi-camera detector.

Please see how the lidar_dnn_detector launcher handles variable lidar-only models, and add this detector as option of camera-only dnn detector.

launch/tier4_perception_launch/launch/object_recognition/detection/detection.launch.xml
```diff
@@ -318,6 +319,30 @@
...
+      <arg name="input/camera3/info" value="$(var input/camera3/info)"/>
+      <arg name="input/camera4/image" value="$(var input/camera4/image)"/>
+      <arg name="input/camera4/info" value="$(var input/camera4/info)"/>
+      <arg name="input/camera5/image" value="$(var input/camera5/image)"/>
+      <arg name="input/camera5/info" value="$(var input/camera5/info)"/>
+      <arg name="input/camera6/image" value="$(var input/camera6/image)"/>
+      <arg name="input/camera6/info" value="$(var input/camera6/info)"/>
+      <arg name="input/camera7/image" value="$(var input/camera7/image)"/>
+      <arg name="input/camera7/info" value="$(var input/camera7/info)"/>
+      <arg name="input/can_bus" value="$(var input/can_bus)"/>
```
technolojin: `input/can_bus` is not defined

launch/tier4_perception_launch/launch/object_recognition/detection/detector/cam_lidar_bev_detector.launch.xml
```diff
@@ -0,0 +1,49 @@
...
+  <arg name="input/camera4/image"/>
+  <arg name="input/camera4/info"/>
+  <arg name="input/camera5/image"/>
+  <arg name="input/camera5/info"/>
+  <arg name="input/camera6/image"/>
+  <arg name="input/camera6/info"/>
+  <arg name="input/camera7/image"/>
+  <arg name="input/camera7/info"/>
+  <arg name="input/can_bus"/>
+  <arg name="input/scene_token"/>
```
technolojin: https://github.com/autowarefoundation/autoware_universe/pull/11076#issuecomment-3172211526

as result, is still `input/scene_token` required? I don't think it shall not be provided in real vehicle operations.

launch/tier4_perception_launch/launch/object_recognition/detection/detection.launch.xml
technolojin: I would like to suggest to add odometry/kinematic state converter or something in `cam_lidar_bev_detector.launch.xml` for integration, so that the bevformer could be used in autoware systems, not only in nuScenes dataset replay.

Dhinesh-MCW:
Hi @technolojin,

Thank you for your review comments, The launch file is functioning as intended. We’ve included a screenshot showing BEVFormer successfully launched within Autoware using the command: 
`ros2 launch autoware_launch e2e_simulator.launch.xml map_path:=$HOME/autoware_map/sample-map-planning perception_mode:=camera_lidar_bev launch_perception:=true`

autoware_launch_bevformer

We understand that it’s not working on your end. Could you kindly share the exact command you have used to launch it so we can help debug the issue?

**Please note:** As discussed during our Software Working Group meetings this PR primarily focuses on the independent integration of BEVFormer and the overall integration we will address that in a future PR. 
[#11076(comment)](https://github.com/autowarefoundation/autoware_universe/pull/11076#discussion_r2316313785)

xmfcx:
With https://github.com/autowarefoundation/autoware_universe/pull/11076/commits/78084e2a3299f88e858af092e3b4728960cc1ca6 I've removed the modifications in the launch folder. @technolojin for now this is a standalone package.

launch/tier4_perception_launch/launch/object_recognition/detection/detector/cam_lidar_bev_detector.launch.xml
xmfcx: these changes are removed from this pr.

launch/tier4_perception_launch/launch/object_recognition/detection/detection.launch.xml
xmfcx: these changes are removed from this pr.

launch/tier4_perception_launch/launch/object_recognition/detection/detection.launch.xml
xmfcx: these changes are removed from this pr.

launch/tier4_perception_launch/launch/object_recognition/detection/detection.launch.xml
xmfcx: these changes are removed from this pr.

xmfcx (changes requested):
https://github.com/user-attachments/assets/278e6fd0-acea-47e1-824f-1176231e53a1

- The output is `2.26 Hz` on an RTX 3090 with FP16 is that expected?
- The `~/output/boxes` topic has wrong orientation, should be fixed

Dhinesh-MCW:
> 2025-09-17.12-26-37.mp4 
> * The output is `2.26 Hz` on an RTX 3090 with FP16 is that expected?
> * The `~/output/boxes` topic has wrong orientation, should be fixed


Hi @xmfcx ,

[#11076 comment](https://github.com/autowarefoundation/autoware_universe/pull/11076#issuecomment-3205714763)
As previously discussed, this behavior is expected given the model’s performances (Discussed at: [link](https://github.com/orgs/autowarefoundation/discussions/5904#discussioncomment-13918364)). However, based on your visualization, it appears that the detections are not synchronized. On our RTX 2080 Ti system, the results are synchronized as expected. We recommend verifying this on your end.

I have also fixed the orientation in the output topic and implemented some fixes from our side.

xmfcx:
> However, based on your visualization, it appears that the detections are not synchronized. On our RTX 2080 Ti system, the results are synchronized as expected. We recommend verifying this on your end.

Could you share a video of this latest version running correctly as I recorded? I'm just following the instructions on the readme file which are:

```bash
# Terminal 1
ros2 launch ros2_dataset_bridge nuscenes_launch.xml

# Terminal 2
ros2 launch autoware_tensorrt_bevformer bevformer.launch.xml debug_mode:=true
```

And these are the changes in my `nuscenes_launch.xml`:

```xml
    <arg name="NUSCENES_DIR" default="/home/mfc/datasets/nuscenes"/>
    <arg name="NUSCENES_VER" default="v1.0-mini"/>
    <arg name="NUSCENES_CAN_BUS_DIR" default="/home/mfc/datasets/nuscenes/can_bus"/>
    <arg name="UPDATE_FREQUENCY" default="10.0"/>
```

https://github.com/user-attachments/assets/1f09f544-89cb-4fd7-8bed-ca5c6a041232

Yet I'm still having the same choppiness issue. This is expected since the output is `2.26 Hz` which means the latency is `1000/2.26 = 442ms` and the lifetime of the markers is `100ms`

> I have also fixed the orientation in the output topic and implemented some fixes from our side.

I've checked the added commits but I couldn't find any which addressed the wrong orientation transformation of this `~/output_boxes` message. And it is still about 90 degrees off.

Dhinesh-MCW:
Hi @xmfcx,
 
You are following all the instructions in the README correctly. But considering the latency of BEVFormer this is expected, the preprocessing and inference stages naturally take some time for the processes. In the recent commit, we have also observed a slight synchronization issue, whereas the earlier tested versions were fully synchronized. I have attached the inference video below for reference:
 
https://github.com/user-attachments/assets/93463458-aee8-431b-81c5-9cc01e1b501a

We will debug this issue soon and will get back to you.
 
Regarding the orientation issue, I have made some changes in ros_utils.cpp. This should resolve the problem, but if you are still facing any issues, kindly provide detailed information so that we can investigate further.

xmfcx:
https://github.com/autowarefoundation/autoware_universe/pull/11076/commits/38d37e83252943c6e267ff3b7cf1a1908b41a426 with this change I've eliminated the discrepancy between `autoware_perception_msgs::msg::DetectedObjects` and the `visualization_msgs::msg::MarkerArray` object.


https://github.com/user-attachments/assets/72502a12-319a-4221-82b8-eb0274ec74cb

There are still these issues:
1. They don't align well (even for still objects when ego car is stationary)
2. I don't know why we apply these magic number transformations to the inference results
3. It's slow
4. The boxes come from behind

Also I don't know why it is slow in general. According to the original implementation at [DerryHub/BEVFormer_tensorrt/README.md](https://github.com/DerryHub/BEVFormer_tensorrt/blob/303d3140c14016047c07f9db73312af364f0dd7c/README.md#bevformer-tensorrt-with-custom-plugins-support-nv_half-nv_half2-and-int8) I should be able to get 15.8 to 18.6 FPS on an RTX3090 as they tested on this hardware.

## [FPS values tested on RTX 3090](https://github.com/DerryHub/BEVFormer_tensorrt/blob/303d3140c14016047c07f9db73312af364f0dd7c/README.md#bevformer-tensorrt-with-custom-plugins-support-nv_half-nv_half2-and-int8)

> **FP16 Plugins with nv_half**
> 
> |      Model      |   Data   | Batch Size | Float/Int |     Quantization Method     |          NDS/mAP          |  **FPS**/Improve  |  Size (MB)  | Memory (MB)  |  **Device**  |
> | :-------------: | :------: | :--------: | :-------: | :-------------------------: | :-----------------------: | :-----------: | :---------: | :----------: | :------: |
> | BEVFormer small | NuScenes |     1      |   FP16    |              -              | NDS: 0.479<br/>mAP: 0.37  | **15.8** (x2.40)  | 127 (x0.52) | 1729 (x0.37) | **RTX 3090** |
> 
> **FP16 Plugins with nv_half2**
> 
> |      Model      |   Data   | Batch Size | Float/Int |     Quantization Method     |          NDS/mAP          |  **FPS**/Improve  |  Size (MB)  | Memory (MB)  |  **Device**  |
> | :-------------: | :------: | :--------: | :-------: | :-------------------------: | :-----------------------: | :-----------: | :---------: | :----------: | :------: |
> | BEVFormer small | NuScenes |     1      |   FP16    |              -              | NDS: 0.479<br/>mAP: 0.371 | **18.6** (x2.82)  | 124 (x0.51) | 1725 (x0.37) | **RTX 3090** |

For this PR, I'd like you to check if there is anything else you'd like to change on [my latest commit](https://github.com/autowarefoundation/autoware_universe/pull/11076/commits/38d37e83252943c6e267ff3b7cf1a1908b41a426) and if you are satisfied, we can merge this and focus on improvements in future PRs.

@naveen-mcw @Dhinesh-MCW @mitsudome-r

Dhinesh-MCW:
Hi @xmfcx,

[c273fce](https://github.com/autowarefoundation/autoware_universe/pull/11076/commits/c273fce7c092203b2db67c68691f630d2b363f2c)
With this commit, we have significantly improved the alignment issue on our side. Previously, we were handling the marker array with different frame IDs, which led to the misalignments. We have now updated it to ensure proper matching. 

Currently, We are experimenting with offsets to yaw for further improvement in the alignment.

Regarding the issue you mentioned in your [comment](https://github.com/autowarefoundation/autoware_universe/pull/11076#issuecomment-3325404285)
In Derryhub’s BEVFormer TensorRT implementation, FPS was calculated incorrectly. They only included inference time, excluding pre and post processings, whereas in our implementation we have calculated FPS for the entire pipeline. For reference, Kindly see [Discussion #5904](https://github.com/orgs/autowarefoundation/discussions/5904#discussioncomment-13918364)

Could you confirm if the recent fix has improved the alignment on your end? Now it looks good from our side.

xmfcx (approved):
https://github.com/user-attachments/assets/f8479feb-0bbb-40f7-91d0-9de1b42b05a0

It looks like this, there is misalignment (even when ego and target cars are stationary) and it is very slow even on and RTX3090 and boxes come from behind. But these can be addressed in future PRs. Thanks for all the efforts, will merge now.

Dhinesh-MCW:
Hi @xmfcx,
We were able to get the correct output when we set it up from scratch. Since you are following all the instructions correctly, I recommend rebuilding everything from scratch and then check the visualizations again.

mitsudome-r:
@Dhinesh-MCW 
I believe the alignment issue is already solved by MCW and confirmed by Fatih as we can see in the attached video. https://github.com/autowarefoundation/autoware_universe/pull/11076#pullrequestreview-3274770745.

We still have issues with preprocessing and inference speed of the module so we can do optimization as separate PRs.

PR #11079, opened 2025-07-29
feat(blind_spot): new re-designed blind_spot module

soblin:
## Description

(the document follows this PR)

### left turn

https://github.com/user-attachments/assets/386431f4-76c9-4ed4-b71e-94fd23abe0dd

### right turn

https://github.com/user-attachments/assets/90f460b0-6872-4502-8832-94c64fcec06c

### how it works

first, `virtual_blind_lane_boundary_after_turning` (purple line in below image) is calculated by extending the boundary of `last_blind_spot_lanelet_before_turning`

image

`virtual_ego_straight_path_after_turning` is computed from the normal direction of `last_blind_spot_lanelet_before_turning`'s end, starting from the intersection of ( ego's left side ) and ( the end line of  `last_blind_spot_lanelet_before_turning` ).
The area enclosed by the 4 purple lines in below image forms the `atttetion_area`.

image

Then `ego_passage_interval` is obtained by computing `ego_future_profile` as illustrated by dotted footprints in below image, and compute the element that intersects with `virtual_ego_straight_path_after_turning` and `virtual_blind_lane_boundary_after_turning`.

image

`path_lanelet` is an artificial lanelet comprising of ego vehicle's path on the intersection. The time interval when the object in the `attention_area` is in the `path_lanelet` is obtained as `object_passage_interval`.

image

## Related links

https://tier4.atlassian.net/browse/RT0-38170

## How was this PR tested?

- [Evaluator Basic](https://evaluation.tier4.jp/evaluation/reports/1aa14328-37ba-56a0-8b29-83e92338bb82?page_size=50&project_id=prd_jt)
- [Evaluator FuncVerification](https://evaluation.tier4.jp/evaluation/reports/68e67ec6-8798-5ee3-bcb6-33b7b456d71c?project_id=prd_jt)
- [Evaluator DLR](https://evaluation.tier4.jp/evaluation/reports/f51b4b6b-26b2-52cd-ad8a-410c9fe67ef4?project_id=prd_jt)

### processing time

for ~3 objects, processing time is about 12ms.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/config/blind_spot.param.yaml
+++ b/planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/config/blind_spot.param.yaml
@@ -1,14 +1,20 @@
 /**:
   ros__parameters:
     blind_spot:
-      use_pass_judge_line: true
-      stop_line_margin: 1.0 # [m]
-      backward_detection_length: 100.0 # [m]
-      ignore_width_from_center_line: 0.7 # [m]
-      adjacent_extend_width: 1.5 # [m]
-      opposite_adjacent_extend_width: 1.5 # [m]
-      max_future_movement_time: 10.0 # [second]
-      ttc_min: -5.0 # [s]
-      ttc_max: 5.0 # [s]
-      ttc_ego_minimal_velocity: 5.0 # [m/s]
+      backward_attention_length: 100.0 # [m]
+      ttc_start_margin: 2.0 # [s]
+      ttc_end_margin: 1.0 # [s]
+      minimum_default_velocity: 1.388 # [m/s]
+      collision_judge_debounce: 0.5 # [s]
+      critical_stopline_margin: 1.0 # [m]
+      brake:
+        critical:
+          deceleration: -6.0 # [m/ss]
+          jerk: -5.0 # [m/sss]
+        semi_critical:
+          deceleration: -4.0 # [m/ss]
+          jerk: -5.0 # [m/sss]
+      brake_for_ttc:
+        critical_threshold_ub: 3.0 # [s]
+        semi_critical_threshold_lb: 5.0 # [s]
       enable_rtc: false # If set to true, the scene modules require approval from the rtc (request to cooperate) function. If set to false, the modules can be executed without requiring rtc approval
--- a/planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/include/autoware/behavior_velocity_blind_spot_module/scene.hpp
+++ b/planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/include/autoware/behavior_velocity_blind_spot_module/scene.hpp
@@ -1,4 +1,4 @@
-// Copyright 2020 Tier IV, Inc.
+// Copyright 2025 Tier IV, Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -16,9 +16,11 @@
 #define AUTOWARE__BEHAVIOR_VELOCITY_BLIND_SPOT_MODULE__SCENE_HPP_
 
 #include <autoware/behavior_velocity_blind_spot_module/parameter.hpp>
+#include <autoware/behavior_velocity_blind_spot_module/time_to_collision.hpp>
 #include <autoware/behavior_velocity_blind_spot_module/util.hpp>
 #include <autoware/behavior_velocity_planner_common/utilization/state_machine.hpp>
 #include <autoware/behavior_velocity_rtc_interface/scene_module_interface_with_rtc.hpp>
+#include <autoware/lanelet2_utils/intersection.hpp>
 #include <rclcpp/rclcpp.hpp>
 
 #include <autoware_perception_msgs/msg/predicted_objects.hpp>
@@ -50,7 +52,6 @@ struct OverPassJudge
 struct Unsafe
 {
   const size_t stop_line_idx;
-  const std::optional<autoware_perception_msgs::msg::PredictedObject> collision_obstacle;
 };
 
 struct Safe
@@ -60,14 +61,23 @@ struct Safe
 
 using BlindSpotDecision = std::variant<InternalError, OverPassJudge, Unsafe, Safe>;
 
+std::string format_blind_spot_decision(const BlindSpotDecision & decision, const lanelet::Id id);
+
+using TurnDirection = autoware::experimental::lanelet2_utils::TurnDirection;
+
 class BlindSpotModule : public SceneModuleInterfaceWithRTC
 {
 public:
   struct DebugData
   {
     std::optional<geometry_msgs::msg::Pose> virtual_wall_pose{std::nullopt};
-    std::optional<lanelet::CompoundPolygon3d> detection_area;
-    autoware_perception_msgs::msg::PredictedObjects conflicting_targets;
+    std::optional<lanelet::CompoundPolygon3d> attention_area;
+    std::optional<lanelet::CompoundPolygon3d> path_polygon;
+    std::optional<lanelet::ConstLineString3d> virtual_blind_lane_boundary_after_turning;
+    std::optional<lanelet::ConstLineString3d> virtual_ego_straight_path_after_turning;
+    std::optional<std::pair<double, double>> ego_passage_interval;
+    std::optional<double> critical_time;
+    std::optional<std::vector<UnsafeObject>> unsafe_objects;
   };
 
 public:
@@ -77,7 +87,8 @@ public:
     const rclcpp::Logger logger, const rclcpp::Clock::SharedPtr clock,
     const std::shared_ptr<autoware_utils::TimeKeeper> time_keeper,
     const std::shared_ptr<planning_factor_interface::PlanningFactorInterface>
-      planning_factor_interface);
+      planning_factor_interface,
+    const rclcpp::Publisher<std_msgs::msg::String>::SharedPtr decision_state_pub);
 
   /**
    * @brief plan go-stop velocity at traffic crossing with collision check between reference path
@@ -89,12 +100,16 @@ public:
   std::vector<autoware::motion_utils::VirtualWall> createVirtualWalls() override;
 
 private:
-  // (semi) const variables
+  // const variables
   const int64_t lane_id_;
   const PlannerParam planner_param_;
   const TurnDirection turn_direction_;
-  std::optional<lanelet::ConstLanelet> sibling_straight_lanelet_{std::nullopt};
-  std::optional<lanelet::ConstLanelets> blind_spot_lanelets_{std::nullopt};
+
+  // (semi) const variables
+  std::optional<lanelet::ConstLanelet> road_lanelets_before_turning_merged_{std::nullopt};
+  std::optional<lanelet::ConstLanelets> blind_side_lanelets_before_turning_{std::nullopt};
+  std::optional<lanelet::ConstLineString3d> virtual_blind_lane_boundary_after_turning_{
+    std::nullopt};
 
   // state variables
   bool is_over_pass_judge_line_{false};
@@ -117,40 +132,19 @@ private:
     const Decision & decision, autoware_internal_planning_msgs::msg::PathWithLaneId * path);
 
   /**
-   * @brief Generate a stop line and insert it into the path.
-   * A stop line is at an intersection point of straight path with vehicle path
-   * @param detection_areas used to generate stop line
-   * @param path            ego-car lane
-   * @param stop_line_idx   generated stop line index
-   * @param pass_judge_line_idx  generated pass judge line index
-   * @return false when generation failed
+   * @brief obtain object with ttc information which is considered dangerous
+   * @return return unsafe objects, in order of collision time (front element is nearest)
    */
-  std::optional<std::pair<size_t, size_t>> generateStopLine(
-    const InterpolatedPathInfo & interpolated_path_info,
-    autoware_internal_planning_msgs::msg::PathWithLaneId * path) const;
-
-  std::optional<OverPassJudge> isOverPassJudge(
-    const autoware_internal_planning_msgs::msg::PathWithLaneId & input_path,
-    const geometry_msgs::msg::Pose & stop_point_pose) const;
-
-  double computeTimeToPassStopLine(
-    const lanelet::ConstLanelets & blind_spot_lanelets,
-    const geometry_msgs::msg::Pose & stop_line_pose) const;
+  std::vector<UnsafeObject> collect_unsafe_objects(
+    const std::vector<autoware_perception_msgs::msg::PredictedObject> & attention_objects,
+    const lanelet::ConstLanelet & ego_path_lanelet,
+    const std::pair<double, double> & ego_passage_time_interval) const;
 
   /**
-   * @brief Check obstacle is in blind spot areas.
-   * Condition1: Object's position is in broad blind spot area.
-   * Condition2: Object's predicted position is in narrow blind spot area.
-   * If both conditions are met, return true
-   * @param path path information associated with lane id
-   * @param objects_ptr dynamic objects
-   * @param closest_idx closest path point index from ego car in path points
-   * @return true when an object is detected in blind spot
+   * @brief filter objects whose position is inside the attention_area and whose type is target type
    */
-  std::optional<autoware_perception_msgs::msg::PredictedObject> isCollisionDetected(
-    const lanelet::ConstLanelets & blind_spot_lanelets,
-    const geometry_msgs::msg::Pose & stop_line_pose, const lanelet::CompoundPolygon3d & area,
-    const double ego_time_to_reach_stop_line);
+  std::vector<autoware_perception_msgs::msg::PredictedObject> filter_attention_objects(
+    const lanelet::BasicPolygon2d & attention_area) const;
 
   /**
    * @brief Check if object is belong to targeted classes
@@ -160,19 +154,19 @@ private:
   bool isTargetObjectType(const autoware_perception_msgs::msg::PredictedObject & object) const;
 
   /**
-   * @brief Modify objects predicted path. remove path point if the time exceeds timer_thr.
-   * @param objects_ptr target objects
-   * @param time_thr    time threshold to cut path
+   * @brief compute the deceleration and jerk for collision stop from `ttc`
+   * if ttc < critical_threshold_ub, use critical profile
+   * if ttc > semi_critical_lb, use semi_critical profile
+   * otherwise, interpolated between the two
    */
-  autoware_perception_msgs::msg::PredictedObject cutPredictPathWithDuration(
-    const std_msgs::msg::Header & header,
-    const autoware_perception_msgs::msg::PredictedObject & object_original,
-    const double time_thr) const;
+  std::pair<double, double> compute_decel_and_jerk_from_ttc(const double ttc) const;
 
   StateMachine state_machine_;  //! for state
 
   // Debug
   mutable DebugData debug_data_;
+
+  rclcpp::Publisher<std_msgs::msg::String>::SharedPtr decision_state_pub_;
 };
 }  // namespace autoware::behavior_velocity_planner
 
--- a/planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/include/autoware/behavior_velocity_blind_spot_module/time_to_collision.hpp
+++ b/planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/include/autoware/behavior_velocity_blind_spot_module/time_to_collision.hpp
@@ -1,0 +1,101 @@
+// Copyright 2025 Tier IV, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef AUTOWARE__BEHAVIOR_VELOCITY_BLIND_SPOT_MODULE__TIME_TO_COLLISION_HPP_
+#define AUTOWARE__BEHAVIOR_VELOCITY_BLIND_SPOT_MODULE__TIME_TO_COLLISION_HPP_
+
+#include "autoware_utils_geometry/boost_geometry.hpp"
+
+#include <autoware/behavior_velocity_planner_common/planner_data.hpp>
+
+#include <autoware_internal_planning_msgs/msg/path_with_lane_id.hpp>
+#include <autoware_internal_planning_msgs/msg/safety_factor.hpp>
+#include <autoware_perception_msgs/msg/predicted_object.hpp>
+#include <geometry_msgs/msg/pose.hpp>
+
+#include <lanelet2_core/Forward.h>
+
+#include <memory>
+#include <optional>
+#include <tuple>
+#include <utility>
+#include <vector>
+
+namespace autoware::behavior_velocity_planner
+{
+
+struct FuturePosition
+{
+  const autoware_internal_planning_msgs::msg::PathPointWithLaneId position;
+  const double duration;
+};
+
+/**
+ * @brief calculate ego vehicle's future position & duration from current position
+ */
+std::vector<FuturePosition> calculate_future_profile(
+  const autoware_internal_planning_msgs::msg::PathWithLaneId & path,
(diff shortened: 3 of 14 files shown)


planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/config/blind_spot.param.yaml
```diff
@@ -1,14 +1,18 @@
...
-      stop_line_margin: 1.0 # [m]
-      backward_detection_length: 100.0 # [m]
-      ignore_width_from_center_line: 0.7 # [m]
-      adjacent_extend_width: 1.5 # [m]
-      opposite_adjacent_extend_width: 1.5 # [m]
-      max_future_movement_time: 10.0 # [second]
-      ttc_min: -5.0 # [s]
-      ttc_max: 5.0 # [s]
-      ttc_ego_minimal_velocity: 5.0 # [m/s]
+      backward_attention_length: 100.0 # [m]
```
satoshi-ota, line 4: Wouldn't it be better from a tuning perspective to calculate this parameter internally using the maximum speed of the considered object, the maximum velocity of the ego vehicle, and the prediction horizon for TTC as parameters?

planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/src/scene.cpp
```diff
@@ -63,84 +90,190 @@ void BlindSpotModule::initializeRTCStatus()
...
+  debug_data_.unsafe_objects.emplace(unsafe_objects);
+  const bool is_safe_now = unsafe_objects.empty();
+
+  // provide hold time for STOP -> GO
   state_machine_.setStateWithMarginTime(
-    collision_obstacle.has_value() ? StateMachine::State::STOP : StateMachine::State::GO,
-    logger_.get_child("state_machine"), *clock_);
+    !is_safe_now ? StateMachine::State::STOP : StateMachine::State::GO,
+    logger_.get_child("collision state_machine"), *clock_);
+  const bool is_safe = (state_machine_.getState() == StateMachine::State::GO);
```
satoshi-ota, line 209: Can momentary perception loss be handled using this state machine...?

planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/src/util.cpp
```diff
@@ -243,122 +348,334 @@ std::vector<lanelet::Id> find_lane_ids_upto(
...
+      insert_pose, path, ego_nearest_dist_threshold, ego_nearest_yaw_threshold);
+    if (!insert_index_opt) {
+      return std::nullopt;
+    }
+    default_stopline = insert_index_opt.value();
+  }
+  const auto critical_stopline_pose =
+    interpolated_path_info.path.points.at(critical_stopline_ip.value()).point.pose;
+  const auto critical_stopline_index_opt = helper::insert_point_index(
+    critical_stopline_pose, path, ego_nearest_dist_threshold, ego_nearest_yaw_threshold);
```
satoshi-ota: I think it would be better to allow the margin before stopping at the conflict area to be configurable via a parameter, rather than stopping immediately in front of it.

planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/src/scene.cpp
```diff
@@ -63,84 +90,190 @@ void BlindSpotModule::initializeRTCStatus()
...
+      const double acceleration = planner_data_->current_acceleration->accel.accel.linear.x;
+      const double braking_dist = planning_utils::calcJudgeLineDistWithJerkLimit(
+        velocity, acceleration, deceleration, jerk_for_deceleration, delay_response_time);
+      return autoware::motion_utils::calcSignedArcLength(path->points, closest_idx, stop_line_idx) >
+             braking_dist;
+    };
+
+  if (is_safe) {
+    // NOTE: over_pass_judge judgement is only allowed when Safe is ensured
+    if (!can_smoothly_stop_at(
```
satoshi-ota, line 247: Why is it necessary to check whether stopping is possible even when the situation is judged to be safe?

satoshi-ota: I understood the reason after reading the above comment.

planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/src/scene.cpp
```diff
@@ -63,84 +90,190 @@ void BlindSpotModule::initializeRTCStatus()
...
-  return Safe{stop_line_idx};
+  const auto & most_unsafe_object = unsafe_objects.front();
+  const auto most_nearest_collision_time = most_unsafe_object.critical_time;
+  debug_data_.critical_time = most_nearest_collision_time;
+
+  const auto [deceleration, jerk] = compute_decel_and_jerk_from_ttc(most_nearest_collision_time);
+
+  autoware_internal_planning_msgs::msg::SafetyFactorArray safety_factor;
+  safety_factor.factors.push_back(most_unsafe_object.to_safety_factor());
+  if (default_stopline && can_smoothly_stop_at(default_stopline.value(), deceleration, jerk)) {
```
satoshi-ota, line 268: Based on experience, there's a concern that if the system checks whether stopping is possible in every cycle, it may fail to stop properly when approaching a stop point if the vehicle's control performance for the commanded velocity is poor. Has this been taken into account?

planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/include/autoware/behavior_velocity_blind_spot_module/time_to_collision.hpp
```diff
@@ -0,0 +1,99 @@
+// Copyright 2025 Tier IV, Inc.
```
satoshi-ota, line 1:
```suggestion
// Copyright 2025 TIER IV, Inc.
```

planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/src/time_to_collision.cpp
```diff
@@ -0,0 +1,239 @@
+// Copyright 2025 Tier IV, Inc.
```
satoshi-ota, line 1:
```suggestion
// Copyright 2025 TIER IV, Inc.
```

planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/src/util.cpp
```diff
@@ -70,155 +81,249 @@ static std::optional<std::pair<size_t, size_t>> findLaneIdInterval(
...
+
+/**
+ * @brief extend the last part of given `line` by some length
+ */
+lanelet::LineString3d extend_linestring_end(const lanelet::ConstLineString3d & line)
+{
+  const auto size = line.size();
+  const auto & p1 = line[size - 2];
+  const auto & p2 = line[size - 1];
+  static constexpr double extend_length = 20.0;
```
maxime-clem: (minor) it would be more flexible if `extend_length` was a function argument with default value `20.0`.

planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/src/util.cpp
```diff
@@ -70,155 +81,249 @@ static std::optional<std::pair<size_t, size_t>> findLaneIdInterval(
...
+  const double dx = p1.x() - p0.x();
+  const double dy = p1.y() - p0.y();
+  const double d = std::hypot(dx, dy);
+  return {-dy / d * length, dx / d * length, 0.0};
+}
+
+/**
+ * @brief extend the last part of given `line` by some length
+ */
+lanelet::LineString3d extend_linestring_end(const lanelet::ConstLineString3d & line)
```
maxime-clem: (minor) "extend" sounds like points will be added to the original linestring.
How about `segment_beyond_linestring_end` ?

planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/config/blind_spot.param.yaml
soblin: I think all of them can be computed by the user beforehand and so attention_length can be determined too. Like our D&T team computes the required attention_length for each ODD using there own method

planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/include/autoware/behavior_velocity_blind_spot_module/scene.hpp
```diff
@@ -117,40 +132,19 @@ class BlindSpotModule : public SceneModuleInterfaceWithRTC
...
   /**
-   * @brief Generate a stop line and insert it into the path.
-   * A stop line is at an intersection point of straight path with vehicle path
-   * @param detection_areas used to generate stop line
-   * @param path            ego-car lane
-   * @param stop_line_idx   generated stop line index
-   * @param pass_judge_line_idx  generated pass judge line index
-   * @return false when generation failed
+   * @brief obtain object with ttc information which is considered dangerous
+   * @return return unsafe objects, in order of collision time (front element is most nearest)
```
maxime-clem:
```suggestion
   * @return return unsafe objects, in order of collision time (front element is nearest)
```

planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/src/scene.cpp
```diff
@@ -63,84 +90,190 @@ void BlindSpotModule::initializeRTCStatus()
...
+      lane_ids_upto_intersection, lane_id_);
+    if (road_and_blind_lanelets_opt) {
+      const auto & [road_lanelets, blind_side_lanelets_before_turning] =
+        road_and_blind_lanelets_opt.value();
+      road_lanelets_before_turning_merged_ = lanelet::utils::combineLaneletsShape(road_lanelets);
+      blind_side_lanelets_before_turning_ = blind_side_lanelets_before_turning;
+    }
+  }
+  if (!blind_side_lanelets_before_turning_ || !road_lanelets_before_turning_merged_) {
+    return InternalError{"there are no lanelets before turning"};
```
maxime-clem, line 123: If I understand correctly this code is only going to be run once but the `path` is needed so it cannot be calculated in the constructor.
Question: can change in the `path` result in different *_lanelets_before_turning ?

planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/include/autoware/behavior_velocity_blind_spot_module/util.hpp
```diff
@@ -45,68 +44,100 @@ struct InterpolatedPathInfo
...
- * @brief generate a new lanelet object from the `turn_direction` side neighboring lanelet of the
- * input `lanelet` by the width of `adjacent_extend_width`
- * @param new lanelet object
+ * @brief return the extend outer boundary of `leftmost_lanelet`
  */
-lanelet::ConstLanelet generateExtendedAdjacentLanelet(
-  const lanelet::ConstLanelet lanelet, const TurnDirection direction,
-  const double adjacent_extend_width);
+lanelet::ConstLineString3d generate_virtual_blind_side_boundary_after_turning(
+  const lanelet::ConstLanelet & leftmost_lanelet,
```
maxime-clem: For right turns this would be the rightmost lanelet right ? Then I suggest naming the variable `outermost_lanelet`.
```suggestion
  const lanelet::ConstLanelet & outermost_lanelet,
```

planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/src/util.cpp
```diff
@@ -243,122 +348,334 @@ std::vector<lanelet::Id> find_lane_ids_upto(
...
+  const lanelet::ConstLineString3d entry_line{
+    lanelet::InvalId, lanelet::Points3d{
+                        remove_const(intersection_lanelet.leftBound().front()),
+                        remove_const(intersection_lanelet.rightBound().front())}};
+  const auto intersection_point_opt =
+    find_intersection_point(path_linestring, lanelet::utils::to2D(entry_line).basicLineString());
+  if (!intersection_point_opt) {
+    return std::nullopt;
+  }
+  const auto intersection_point = intersection_point_opt.value();
```
maxime-clem:
```suggestion
  const auto & intersection_point = intersection_point_opt.value();
```

planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/src/util.cpp
```diff
@@ -243,122 +348,334 @@ std::vector<lanelet::Id> find_lane_ids_upto(
...
+  lanelet::Points3d attention_area_right_boundary;
+
+  auto & far_side_boundary = (turn_direction == TurnDirection::Left)
+                               ? attention_area_left_boundary
+                               : attention_area_right_boundary;
+  auto & near_side_boundary = (turn_direction == TurnDirection::Left)
+                                ? attention_area_right_boundary
+                                : attention_area_left_boundary;
+
+  // left bound
```
maxime-clem:
```suggestion
  // far side bound
```

planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/src/util.cpp
```diff
@@ -243,122 +348,334 @@ std::vector<lanelet::Id> find_lane_ids_upto(
...
+    far_side_boundary.push_back(remove_const(point));
+  }
+  for (const auto & point : virtual_blind_side_boundary_after_turning) {
+    far_side_boundary.push_back(remove_const(point));
+  }
+  if (far_side_boundary.size() < 2) {
+    return std::nullopt;
+  }
+
+  // right bound
```
maxime-clem:
```suggestion
  // near side bound
```

planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/src/time_to_collision.cpp
```diff
@@ -0,0 +1,239 @@
...
+  factor.object_id = object.object_id;
+  factor.predicted_path = predicted_path;
+  factor.ttc_begin = std::get<0>(object_passage_interval);
+  factor.ttc_end = std::get<1>(object_passage_interval);
+  factor.points.push_back(object.kinematics.initial_pose_with_covariance.pose.position);
+  factor.is_safe = false;
+  return factor;
+}
+
+std::optional<double> is_critical_for_ego(
```
maxime-clem: the name of the function can be improved. Maybe `get_unsafe_time` ?

planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/src/time_to_collision.cpp
```diff
@@ -0,0 +1,239 @@
...
+  const std::pair<double, double> & ego_passage_interval,
+  const std::pair<double, double> & object_passage_interval, const double ttc_start_margin,
+  const double ttc_end_margin)
+{
+  const auto & [ego_entry, ego_exit] = ego_passage_interval;
+  const auto & [object_entry, object_exit] = object_passage_interval;
+  // case1: ego will be still in conflict_area, when the object enters conflict_area
+  if (object_entry - ttc_start_margin < ego_exit) {
```
maxime-clem, line 247: This assumes `ego_entry < object_entry` (ego entered before the object) which can be wrong and cause false positives.

https://github.com/user-attachments/assets/e1044802-605a-44ca-bfa7-f5f37ca02ad1

planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/src/time_to_collision.cpp
```diff
@@ -0,0 +1,239 @@
...
+  }
+  return calculate_future_profile_impl(
+    smoothed_reference_path, current_pose, time_to_restart, nearest_dist_threshold,
+    nearest_yaw_threshold);
+}
+
+std::optional<std::pair<double, double>> compute_time_interval_for_passing_line(
+  const std::vector<FuturePosition> & future_positions,
+  const autoware_utils_geometry::LinearRing2d & footprint, const lanelet::ConstLineString3d & line1,
+  const lanelet::ConstLineString3d & line2)
```
maxime-clem, line 121: I think some additional filtering can be performed using the `ego_passage_interval` in this function.
If I understand correctly, we only care about the position of the object during `ego_passage_interval` (+/- the TTC margins). So I think the calculations can be skipped when the time is too far from the `ego_passage_interval`.

This should improve performance and may also fix the current bug with the false positives.

planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/src/time_to_collision.cpp
```diff
@@ -0,0 +1,239 @@
...
+{
+  // search forward
+  std::optional<double> entry_time{};
+  for (const auto & [path_point, time] : future_positions) {
+    const auto & base_pose = path_point.point.pose;
+    const auto path_point_footprint =
+      autoware_utils::transform_vector(footprint, autoware_utils::pose2transform(base_pose));
+    if (boost::geometry::intersects(
+          path_point_footprint, lanelet::utils::to2D(line1).basicLineString())) {
```
maxime-clem, line 130: (perf) from a quick profiling, it seems the 2 `boost::geometry::intersects` account for ~60% of the runtime of the module.
I think putting all the pairs `(path_point_footprint, time)` in a Rtree and querying all the intersections with the lines would be much faster.

This is low priority and should be kept for a future PR.

maxime-clem (changes requested):
Overall the code looks good and I only have minor comments.
My main concern regarding the logic is if the `blind_side_lanelets_before_turning` are not straight but curved. Then I think the virtual lines and the whole calculation may become wrong.
Additionally I think I found a bug with the conditions to detect a collision that needs to be fixed.

planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/config/blind_spot.param.yaml
satoshi-ota: Since it's a niche topic, it's fine to address it later, but do you think there would be less room for tuning if we implemented the method/model used by the D&T team and allowed users to specify only intuitive values such as vehicle speed?

planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/src/scene.cpp
soblin: > If I understand correctly this code is only going to be run once but the path is needed so it cannot be calculated in the constructor.

Excatly. This snippet is to cache the result.

planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/src/scene.cpp
satoshi-ota: It might be enough.

planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/src/scene.cpp
maxime-clem: It would be easier to understand if this code was put in a separate function (e.g., `cache_lanelets_preceding_the_turn()`).
But this can be done in a future refactoring PR.

planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/src/util.cpp
satoshi-ota: Since the system is designed to use the maximum possible braking force to stop, I understand that the margin parameter is unnecessary.

planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/include/autoware/behavior_velocity_blind_spot_module/time_to_collision.hpp
soblin: fixed in f3d3f9802c8a7424d5775c7e7155f65633b8dc83, let me close

planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/src/scene.cpp
soblin: closed in internal discussion

planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/src/time_to_collision.cpp
soblin: fixed in f3d3f9802c8a7424d5775c7e7155f65633b8dc83, let me close

planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/src/util.cpp
soblin: addressed in f3d3f9802c8a7424d5775c7e7155f65633b8dc83,

planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/src/util.cpp
soblin: addressed in f3d3f9802c8a7424d5775c7e7155f65633b8dc83

planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/include/autoware/behavior_velocity_blind_spot_module/scene.hpp
soblin: addressed in f3d3f9802c8a7424d5775c7e7155f65633b8dc83

planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/include/autoware/behavior_velocity_blind_spot_module/util.hpp
soblin: addressed in f3d3f9802c8a7424d5775c7e7155f65633b8dc83

planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/src/util.cpp
soblin: addressed in f3d3f9802c8a7424d5775c7e7155f65633b8dc83

planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/src/util.cpp
soblin: addressed in f3d3f9802c8a7424d5775c7e7155f65633b8dc83

planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/src/util.cpp
soblin: addressed in f3d3f9802c8a7424d5775c7e7155f65633b8dc83

planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/src/time_to_collision.cpp
soblin: addressed in f3d3f9802c8a7424d5775c7e7155f65633b8dc83

planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/src/time_to_collision.cpp
soblin: I will deal with it in future PR !

planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/src/time_to_collision.cpp
soblin: in f3d3f9802c8a7424d5775c7e7155f65633b8dc83, if `object_exit < ego_entry` this FP is ignored properly

planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/config/blind_spot.param.yaml
soblin: I think that will be nice if we prepare some script to calculate the attention_area_length from given ODD condition

yhisaki (approved):

PR #11086, opened 2025-07-30
fix(goal_planner): fix goal pose footprint check within area

kosuke55:
## Description

The current implementation incorrectly rejects valid goal poses that are close to a boundary.

The root cause is a mismatch in the pose used for footprint generation. The final goal pose is oriented so the vehicle front is parallel to the boundary. However, the footprint used for the validity check is generated from the original, unaligned pose. This leads to a faulty in/out-of-area check.

This PR resolves the issue by generating the footprint from the aligned pose, ensuring the validity check is performed on the correct configuration.

before

image

image

after 

image

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


kosuke55:
move to https://github.com/autowarefoundation/autoware_universe/pull/11224

PR #11095, opened 2025-07-31
feat(pose instability detector): change diag level

YamatoAndo:
## Description

To link with MRM, change diag level from `WARN` to `ERROR`.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

Just checking if it can be built

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


KYabuuchi (approved):
Looks Good To Me!

PR #11100, opened 2025-08-01
feat(autoware_traffic_light_arbiter): handle multiple external sources

dkoldaev:
## Description

Currently the traffic light arbiter assumes that there is only one external source. However, it might not be the case, so in this PR we allow for multiple external sources. To do so, latest message is not overridden, but merged into existing message. Also, old messages are cleaned up using timeout.  

Known limitation: there is no algorithm of deciding in a conflict situation - the latest data will always be the priority. However, we do not expect two external sources provide information about the same traffic light.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

Unit tests are added to test multiple sources

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


perception/autoware_traffic_light_arbiter/include/autoware/traffic_light_arbiter/traffic_light_arbiter.hpp
```diff
@@ -40,6 +41,12 @@ class TrafficLightArbiter : public rclcpp::Node
   using TrafficSignalArray = autoware_perception_msgs::msg::TrafficLightGroupArray;
   using TrafficSignal = autoware_perception_msgs::msg::TrafficLightGroup;
 
+  struct ExternalTrafficLightInfo
```
shmpwk: How about using TrafficSignalArray msg which has traffic siganal and timestamp?　https://github.com/autowarefoundation/autoware_msgs/blob/f49c35c7a7b0110ad7bf5d91f67967833fd1c0d9/autoware_perception_msgs/msg/TrafficSignalArray.msg

shmpwk (commented):
LGTM for the logic. The unit tests greatly cover the feature. I have one small comment.

perception/autoware_traffic_light_arbiter/include/autoware/traffic_light_arbiter/traffic_light_arbiter.hpp
dkoldaev: @shmpwk thank you, good point. I will make the update. Note that it is TrafficLightGroupArray, not TrafficSignalArray, because `  using TrafficSignalArray = autoware_perception_msgs::msg::TrafficLightGroupArray;
`

shmpwk (approved):
LGTM

shmpwk:
@MasatoSaeki Could you briefly check this PR as a code owner?

shmpwk:
@YoshiRi @technolojin @MasatoSaeki 
Could you help to review it as a code owner?

MasatoSaeki:
@dkoldaev 
Thank you for your PR and sorry for delay.
I am running on Evaluator, so if pass I will approve this PR. (https://evaluation.ci.tier4.jp/evaluation/reports/e3aee5ce-6027-5d00-8bb9-110804b04f6a?project_id=prd_jt)

I have one question.
This PR get to be able to handle with multiple source V2I msg, but actually it's an extension to the function of handling (arbitrate) multiple traffic lights, right? (Or does each external msg only have information about one traffic light?)

dkoldaev:
> I have one question.
> This PR get to be able to handle with multiple source V2I msg, but actually it's an extension to the function of handling (arbitrate) multiple traffic lights, right? (Or does each external msg only have information about one traffic light?)

@MasatoSaeki thank you for your review! It will handle the case where each message contains more than one traffic light. It will also handle the case where two sources provide information for the same traffic light, but in a naive way - the later message will be chosen. As the external traffic light message does not have any information about the source we can't create a better arbitration algorithm. The main goal of this PR is to not override the data once we receive a new message, but merge it.

PR #11103, opened 2025-08-01
fix(control_evaluator): publish dist to uncrossable boundary

zulfaqar-azmi-t4:
## Description

Fix control evaluator's distance to uncrossable boundary, so that it constantly publishing values. 

The new behaviors are
if boundary is nearby = publishes distance to boundary
if no boundary found = publishes contant `search_distance` value.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


xtk8532704 (approved):
So it now publishes `search_distance` if nearby_uncrossable_lines_opt is null.

LGTM

PR #11104, opened 2025-08-01
fix(tier4_perception_launch): add yolox 2d detection launch with bytetrack

badai-nguyen:
## Description

- Add `bytetrack` option into launch to easily reproduce the same result compare with real product's ROIs

## Related links

**Parent Issue:**

- [TIER IV internal link](https://star4.slack.com/archives/C03S84LDJGG/p1754633292038409?thread_ts=1754040174.204459&cid=C03S84LDJGG)

## How was this PR tested?

- Run `ros2 launch tier4_perception_launch camera_2d_detector.launch.xml enable_visualizer:=true image_number:=3` and confirm the bytetrack node is loaded together with `tensorrt_yolox`

```
ros2 node list | grep -e byte -e yolox
/bytetrack0
/bytetrack1
/bytetrack2
/bytetrack_visualizer0
/bytetrack_visualizer1
/bytetrack_visualizer2
/tensorrt_yolox0
/tensorrt_yolox1
/tensorrt_yolox2
```

- Run `ros2 launch autoware_launch logging_simulator.launch.xml map_path:=$HOME/sample_autoware/sample-map vehicle_model:=sample_vehicle sensor_model:=sample_sensor_kit control:=false planning:=false perception:=true perception_mode:=camera_lidar_fusion localization:=false use_traffic_light_recognition:=false` with option: `"enable_2d_detection" default="true"` and example `"image_number" value="2"`
  - Confirm that 2 pair of yolox and bytetrack are executed with autoware
``` 
ros2 node list | grep -e byte -e yolox
/perception/object_recognition/detection/bytetrack0
/perception/object_recognition/detection/bytetrack1
/perception/object_recognition/detection/tensorrt_yolox0
/perception/object_recognition/detection/tensorrt_yolox1
```

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


ktro2828:
@badai-nguyen Do you have any special reasons why you updated only specific launch files?

badai-nguyen:
> @badai-nguyen Do you have some special reasons why you updated only specific launch files?

@ktro2828 cc @YoshiRi As [an internal discussion](https://star4.slack.com/archives/C03S84LDJGG/p1751517950158029?thread_ts=1751355812.275449&cid=C03S84LDJGG), some time we want to run Autoware with Yolox on Evaluator. However currently, there is no option to enable bytrack to get a similar result as real vehicle which use bytetrack as default.

PR #11109, opened 2025-08-04
Fix/multi object tracker/fix existence probability

technolojin:
## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


ogrex (approved):
No abnormal issues observed.
LGTM as a quick fix.

Just a note for future consideration: we might consider limiting this fix to nearby areas only? unless there’s evidence that the same issue affects far objects as well.

technolojin:
> No abnormal issues observed. LGTM as a quick fix.
> 
> Just a note for future consideration: we might consider limiting this fix to nearby areas only? unless there’s evidence that the same issue affects far objects as well.

This kind of tracker lost might happen under certain condition
1. roi cluster fusion gives very stable position and classification
2. then, the object position and size becomes very stable
3. as of tracking result, the position covariance becomes very low
4. at the association gating, mahalanobis distance becomes very large, too sensitive on small disturbance of object position

therefore, this situation may not happen for far distance objects

PR #11111, opened 2025-08-04
fix(multi object tracker): fix classification filter implementation

technolojin:
## Description

1. the class vector was not updated (parent class member could not be accessed from child method) Therefore, the initial class was always used.
2. classification probability update to use Bayes theory
3. introduce channel priority: lower the index, higher the priority
4. when trackers are merged, update existence probabilities, classification, and shape  
5. object cache is improved, by additionally checking update counts
6. adjust max_area_matrix parameters for large pedestrian (pedestrian with baby car, wheelchair, walking with bicycle)

## Related links

Issues
- [TIER IV INTERNAL](https://tier4.atlassian.net/browse/RT0-38684)
- [TIER IV INTERNAL](https://tier4.atlassian.net/browse/RT0-38654)

launcher PR https://github.com/autowarefoundation/autoware_launch/pull/1589

## How was this PR tested?
multi channel mode is enabled

* before - initialized as motorcycle

https://github.com/user-attachments/assets/016e90cf-2127-46ff-b604-f6e93767d337

* before - initialized as pedestrian

https://github.com/user-attachments/assets/32792c5e-b541-4452-a22d-f653a161b813

* after

https://github.com/user-attachments/assets/774dfa6a-ec0a-4133-9e34-cdfe61c33a42

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


perception/autoware_multi_object_tracker/lib/tracker/model/tracker_base.cpp
```diff
@@ -193,63 +196,87 @@ void Tracker::updateClassification(
...
-    for (auto & old_class : classification) {
-      if (new_class.label == old_class.label) {
-        old_class.probability += new_class.probability * gain;
-        found = true;
-        break;
-      }
+  constexpr float true_positive_rate = 0.8f;
+  constexpr float true_negative_rate = 0.4f;
+  constexpr float false_positive_rate = 0.2f;
+  constexpr float false_negative_rate = 0.2f;
```
ogrex: If it’s meant literally, shouldn’t false_positive_rate be 1 - true_negative_rate?
Currently, false_negative_rate equals 1 - true_positive_rate, but perhaps we should compute it directly for clarity.

perception/autoware_multi_object_tracker/lib/tracker/model/tracker_base.cpp
```diff
@@ -193,63 +196,87 @@ void Tracker::updateClassification(
...
-      [remove_threshold](const auto & a_class) { return a_class.probability < remove_threshold; }),
-    classification.end());
+  // Normalization
+  {
+    float sum = 0.0;
+    for (const auto & a_class : classification_) {
+      sum += a_class.probability;
+    }
+    // Normalize only if the total probability is greater than 1.0
+    if (sum > 1.0) {
```
ogrex, line 255: I’m not sure I fully understand, but why is normalization applied only when the total probability exceeds 1.0?

perception/autoware_multi_object_tracker/lib/tracker/model/tracker_base.cpp
ogrex: Should this be made configurable?

perception/autoware_multi_object_tracker/lib/tracker/model/tracker_base.cpp
technolojin: the rest is probability of UNKNOWN

perception/autoware_multi_object_tracker/lib/tracker/model/tracker_base.cpp
technolojin: Mathematically you are right. 
Not only that, likelihood of it->probability * true_positive_rate also need to be considered.

Let me revise the overall algorithm.

technolojin: @ogrex 
Refactored https://github.com/autowarefoundation/autoware_universe/pull/11111/commits/9b6f144194205833f26e662a31639f6eefc47099 https://github.com/autowarefoundation/autoware_universe/pull/11111/commits/4bcdaf8ddb0e55c27f9fee01ae629af93f18342b

Actually, the current algorithm does not violate the mathematical condition.
It differentiates probability by the input has classification or not.

technolojin: @ogrex 

> Should this be made configurable?

At this moment, I don't consider these parameters to be configured.
If we want to make the parameter configurable, it is better to align classification vectors of detectors.

technolojin: YoloX: 100% in a class, nothing else
Euclidean cluster: 100% unknown, nothing else
Centerpoint: 0-100% per class, no unknown -> 100% in a class, nothing else

ogrex (approved):
LGTM

PR #11115, opened 2025-08-04
feat(autoware_cuda_dependency_meta): pass -fno-lto compile flag to downstream packages of autoware_cuda_dependency_meta

esteve:
## Description

This PR adds a `CONFIG_EXTRAS` file so that downstream dependencies disable LTO. The Debian packaging tools (`sbuild`, `fakeroot`, etc.) enable LTO by default for all packages, but that optimization is not supported in CUDA and causes link errors.

See https://forums.developer.nvidia.com/t/link-time-optimization-with-cuda-on-linux-flto/55530

## Related links

https://forums.developer.nvidia.com/t/link-time-optimization-with-cuda-on-linux-flto/55530

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/common/autoware_cuda_dependency_meta/CMakeLists.txt
+++ b/common/autoware_cuda_dependency_meta/CMakeLists.txt
@@ -3,4 +3,4 @@ project(autoware_cuda_dependency_meta NONE)
 
 find_package(ament_cmake REQUIRED)
 
-ament_package()
+ament_package(CONFIG_EXTRAS autoware_cuda_dependency_meta-extras.cmake)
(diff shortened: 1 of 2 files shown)


esteve:
This change is needed so that CUDA-dependent packages can be built in the Debian buildfarm.

esteve:
@mitsudome-r @xmfcx This change is needed so that CUDA-dependent packages can be built in the Debian buildfarm.

xmfcx:
Added to the dict:
- https://github.com/autowarefoundation/autoware-spell-check-dict/pull/82
- https://github.com/autowarefoundation/autoware-spell-check-dict/pull/83

xmfcx (approved):

PR #11120, opened 2025-08-05
fix(run_out): adjust unfeasible stop points to become feasible

maxime-clem:
## Description

Previously, when a stop point was found to be unfeasible (i.e., breaking the deceleration limit), a diagnostic would be published but the stop point would still be used.
This PR modifies the stop points to become feasible by using the minimum stopping distance plus some assumed control delay.

## Related links

**Private Links:**

- [TIER IV JIRA](https://tier4.atlassian.net/browse/RT0-38018?focusedCommentId=249310&sourceType=mention)

## How was this PR tested?

Psim + Reproducer
Evaluator: https://evaluation.tier4.jp/evaluation/reports/87e015ec-b472-59c5-ba07-42e7e2e96561?project_id=x2_dev

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

`run_out` is less likely to publish unfeasible stop points.

### DIFF
--- a/planning/motion_velocity_planner/autoware_motion_velocity_run_out_module/src/slowdown.cpp
+++ b/planning/motion_velocity_planner/autoware_motion_velocity_run_out_module/src/slowdown.cpp
@@ -78,7 +78,6 @@ std::optional<geometry_msgs::msg::Point> calculate_stop_position(
   const auto & check_unfeasible_stop = [&](const auto stop_distance) -> bool {
     const auto stop_decel = (current_velocity * current_velocity) / (2 * stop_distance);
     if (stop_decel > params.stop_deceleration_limit) {
-      unfeasible_stop_deceleration = stop_decel;
       if (stop_decel > unfeasible_stop_deceleration.value_or(0.0)) {
         unfeasible_stop_deceleration = stop_decel;
       }


planning/motion_velocity_planner/autoware_motion_velocity_run_out_module/src/slowdown.cpp
```diff
@@ -75,13 +75,16 @@ std::optional<geometry_msgs::msg::Point> calculate_stop_position(
...
-    if (
-      stop_decel > params.stop_deceleration_limit &&
-      stop_decel > unfeasible_stop_deceleration.value_or(0.0)) {
+    if (stop_decel > params.stop_deceleration_limit) {
       unfeasible_stop_deceleration = stop_decel;
+      if (stop_decel > unfeasible_stop_deceleration.value_or(0.0)) {
+        unfeasible_stop_deceleration = stop_decel;
+      }
+      return true;
     }
```
zulfaqar-azmi-t4: In this part, it looks to be like the nested `if` is redundant?

maxime-clem: The nested if is to only update the `unfeasible_stop_deceleration` when it is higher than the currently stored value. This allow getting the maximum unfeasible acceleration in the end.

Previously, I only cared about updating this variable, but now I want to return `true` even if the value is not the maximum, which is why I split the if condition.

The error is that I did not remove the previous assignment and I now removed it: fafc7b26ff

zulfaqar-azmi-t4 (approved):
LGTM

PR #11121, opened 2025-08-05
feat(tier4_autoware_api_launch): add api mode

isamu-takagi:
## Description

Add api_mode launch option for API migration.
- `api_mode:=0` : Launch all APIs (default)
- `api_mode:=1` : APIs up to EOL 2025/09 will not be launched.
- `api_mode:=2` : APIs up to EOL 2025/12 will not be launched.

## Related links

**Parent Issue:**

- https://github.com/autowarefoundation/autoware/issues/3096

## How was this PR tested?

Checkout this PR: https://github.com/tier4/tier4_ad_api_adaptor/pull/168
For EOL, see https://github.com/tier4/tier4_ad_api_adaptor/blob/tier4/universe/README.md

- Check that the same APIs are launched as before with option `launch_deprecated_api:=true api_mode:=0`
- Check that the APIs EOL 2025/09 are not launched `launch_deprecated_api:=true api_mode:=1`
- Check that the APIs EOL 2025/12 are not launched `launch_deprecated_api:=true api_mode:=2`

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/launch/tier4_autoware_api_launch/launch/deprecated_api.launch.xml
+++ b/launch/tier4_autoware_api_launch/launch/deprecated_api.launch.xml
@@ -1,4 +1,6 @@
 <launch>
+  <arg name="api_mode" default="0"/>
+
   <!-- awapi -->
   <group>
     <include file="$(find-pkg-share awapi_awiv_adapter)/launch/awapi_awiv_adapter.launch.xml"/>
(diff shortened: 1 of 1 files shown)


launch/tier4_autoware_api_launch/launch/deprecated_api.launch.xml
```diff
@@ -1,4 +1,6 @@
 <launch>
+  <arg name="api_mode" default="0"/>
```
mitsudome-r: Could you add description for this argument? (LIke this [line](https://github.com/autowarefoundation/autoware_launch/blob/a75b0536f244762821ad5077299834b75656b85d/autoware_launch/launch/components/tier4_perception_component.launch.xml#L49) in autoware_launch)

sasakisasaki (commented):
Just a quick comment/question, it seems only the case `api_mode < 2` is handled. Is it okay if we do not handle the case when `api_mode == 2`, or `api_mode == 2`'s case is already handled?

launch/tier4_autoware_api_launch/launch/deprecated_api.launch.xml
isamu-takagi: Added https://github.com/autowarefoundation/autoware_universe/pull/11121/commits/6b657497e3fc840b0c4e01f95143a32ffabd7d47

isamu-takagi:
> Just a quick comment/question, it seems only the case api_mode < 2 is handled. Is it okay if we do not handle the case when api_mode == 2, or api_mode == 2's case is already handled?

@sasakisasaki It should be OK. If it is 2 or more, the include action will be disabled and the entire deprecated API will not be invoked. The cases of 0 or 1 will be left to processing on the adapter repository side.

PR #10996, opened 2025-07-16
Feat/polar voxel filter

drwnz:
## Description

[Draft PR for polar voxel filter, for testing purposes]

PR #11122, opened 2025-08-05
feat(autoware_cuda_pointcloud_preprocessor): cuda/polar voxel filter

manato:
## Description
This PR provides a GPU-accelerated implementation of the polar voxel outlier filter, which was introduced #10996 ("CPU implementation", hereafter).

## Related links

**Parent Issue:**

- https://github.com/autowarefoundation/autoware_universe/pull/10996

## How was this PR tested?
Using
```bash
ros2 launch autoware_cuda_pointcloud_preprocessor cuda_polar_voxel_outlier_filter.launch.xml \
  input/pointcloud:=<input_cloud>
```
with a sample rosbag

## Notes for reviewers

- The outputs of this filter are almost similar to the ones from the CPU implementation, but not perfectly identical due to differences in floating-point arithmetic between CPUs and GPUs
- The node parameters are common with the CPU implementation

## Interface changes

None.

## Effects on system behavior

None.


sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor_node.cpp
```diff
@@ -149,6 +150,7 @@ CudaPointcloudPreprocessorNode::CudaPointcloudPreprocessorNode(
 
   cuda_pointcloud_preprocessor_ = std::make_unique<CudaPointcloudPreprocessor>();
   cuda_pointcloud_preprocessor_->setRingOutlierFilterParameters(ring_outlier_filter_parameters);
+  cuda_pointcloud_preprocessor_->setRingOutlierFilterActive(enable_ring_outlier_filter);
```
mojomex, line 153: (Not for this PR) I tried to push for this in the original CUDA preprocessor PR review: the components should be separated into nodes, like with the CPU preprocessor. With CUDA blackboard, I don't think there should be any performance loss in doing so.

sensing/autoware_cuda_utils/include/autoware/cuda_utils/cuda_memory_pool.hpp
```diff
@@ -0,0 +1,63 @@
...
+ * small blocks of memory quickly and efficiently.
+ *
+ * \param max_mem_pool_size_in_byte The maximum size of the memory pool in bytes. WHen more than the
+ * this value of memory are held by the memory pool, the allocator will try to release memory.
+ * Defaults to 1GB.
+ * \param device_id The ID of the CUDA device to create the memory pool on. Defaults to 0.
+ * \return A CUDA memory pool handler.
+ */
+cudaMemPool_t create_memory_pool(
+  const size_t & max_mem_pool_size_in_byte = 1e9, const int device_id = 0)
```
mojomex: Both of these should probably not have defaults. Especially the 1GB default for mempool size seems detrimental. The user of this function should think about reasonable limits for the mempool.

sensing/autoware_cuda_pointcloud_preprocessor/docs/cuda-polar-voxel-outlier-filter.md
```diff
@@ -0,0 +1,54 @@
+# cuda_polar_voxel_outlier_filter
+
+## Purpose
+
+This node is a CUDA accelerated version of the `PolarVoxelOutlierFilter` available in [autoware_cuda_pointcloud_preprocessor](../../autoware_pointcloud_preprocessor).
+
+## Inner-workings / Algorithms
+
+This node re implements of the function of `autoware::pointcloud_preprocessor::PolarVoxelOutlierFilterComponent`, which filters outlier based on polar coordinate space instead of Cartesian coordinate space.
```
mojomex:
```suggestion
This node is an alternative implementation to `autoware::pointcloud_preprocessor::PolarVoxelOutlierFilterComponent`, which filters outliers based on voxels in polar coordinate space instead of Cartesian coordinate space.
```

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_outlier_filter/cuda_polar_voxel_outlier_filter_node.cpp
```diff
@@ -0,0 +1,525 @@
...
+
+  // Create visibility publisher
+  visibility_pub_ = create_publisher<autoware_internal_debug_msgs::msg::Float32Stamped>(
+    "~/debug/visibility", rclcpp::SensorDataQoS());
+
+  // Create ratio publisher
+  ratio_pub_ = create_publisher<autoware_internal_debug_msgs::msg::Float32Stamped>(
+    "~/debug/filter_ratio", rclcpp::SensorDataQoS());
+
+  sub_ =
```
mojomex: For clarity, `pointcloud_sub_` might be better.

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_outlier_filter/cuda_polar_voxel_outlier_filter_node.cpp
```diff
@@ -0,0 +1,525 @@
...
+      "PointXYZ format has not been supported by "
+      "autoware_cuda_pointcloud_preprocessor::cuda_polar_voxel_outlier_filter yet.");
+  }
+
+  filtered_cloud = std::move(filter_return.filtered_cloud);
+  noise_cloud = std::move(filter_return.noise_cloud);
+  filter_ratio_ = filter_return.filter_ratio;
+  visibility_ = filter_return.visibility;
+
+  if (filtered_cloud.get() == nullptr) {
```
mojomex: `.get()` is not needed, and ClangTidy complains about it.

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_outlier_filter/cuda_polar_voxel_outlier_filter_node.cpp
```diff
@@ -0,0 +1,525 @@
...
+    {"visibility_warn_threshold", {validate_normalized, [this](const rclcpp::Parameter & p) {
+                                     filter_params_.visibility_warn_threshold = p.as_double();
+                                   }}}};
+
+  for (const auto & param : params) {
+    auto it = param_ops.find(param.get_name());
+    if (it != param_ops.end()) {
+      if (it->second.validator) {
+        std::string reason;
+        if (!it->second.validator(param, reason)) {
```
mojomex, line 399: Now that I look at it again, this is only ever called when params are changed at runtime - not at startup.
Since we have such nice validators, it would be good to use them at startup too.

I think the same issue might be in the CPU version @drwnz?

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_outlier_filter/cuda_polar_voxel_outlier_filter.cu
```diff
@@ -0,0 +1,949 @@
...
+      ret = false;
+    } else {  // rhs has value
+      if (!lhs) {
+        ret = true;
+      } else {
+        ret = lhs.value() != rhs.value();
+      }
+    }
+    return ret;
+  }
```
mojomex: nit: personal preference:
```suggestion
    return rhs && (!lhs || lhs.value() != rhs.value());
  }
```

mojomex: or:
```suggestion
    if (!rhs) {
      return false;
    } 
    
    if (!lhs) {
      return true;
    }

    return lhs.value() != rhs.value();
  }
```

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_outlier_filter/cuda_polar_voxel_outlier_filter.cu
```diff
@@ -0,0 +1,949 @@
...
+{
+  if (!primary_return_type_dev_) {
+    return FilterReturn{nullptr, nullptr, 0., 0.};
+  }
+
+  size_t num_points = input_cloud->width * input_cloud->height;
+  if (num_points == 0) {
+    // sometimes topic might contain zero point even the pointer is valid
+    // For such cases, this filter do nothing
+    return FilterReturn{nullptr, nullptr, 0., 0.};
```
mojomex: I think in this case, at least an empty cloud should be published for filtered and noise cloud.

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_outlier_filter/cuda_polar_voxel_outlier_filter.cu
```diff
@@ -0,0 +1,949 @@
...
+  // } else {
+  //   if (!filter_secondary_returns) {
+  //     is_valid = true;
+  //   } else {
+  //     is_valid = is_primary_returns[point_index.value()];
+  //   }
+  // }
+  // valid_points_mask[point_index.value()] = is_valid;
+  // -----
+
```
mojomex: I think the statement below would still be clear, even when this comment is removed.

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_outlier_filter/cuda_polar_voxel_outlier_filter.cu
```diff
@@ -0,0 +1,949 @@
...
+    return;
+  }
+
+  if constexpr (::cuda::std::is_same<TFieldData, double>::value) {
+    output[point_index] = static_cast<int32_t>(floor(field_data / resolutions[field_index]));
+  } else {
+    output[point_index] = static_cast<int32_t>(floorf(field_data / resolutions[field_index]));
+  }
+}
+
```
mojomex: While I like the idea of `FieldDataIndex` and `FieldDataComposer`, I feel like the code would be simpler if it just operated on a `float azimuth, float elevation, float distance` directly. Checks like `is_finite` could still be done in a loop over the three fields, like this: https://godbolt.org/z/h3KfKvMsY

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_outlier_filter/cuda_polar_voxel_outlier_filter.cu
```diff
@@ -0,0 +1,949 @@
...
+{
+  bool is_finite = isfinite(field_data);
+  bool is_within_radius_range = (field_index == FieldDataIndex::radius)
+                                  ? (min_radius <= field_data) && (field_data <= max_radius)
+                                  : true;
+  bool has_sufficient_radius =
+    (field_index == FieldDataIndex::radius)
+      ? ::cuda::std::abs(field_data) >= ::cuda::std::numeric_limits<TFieldData>::epsilon()
+      : true;
+
```
mojomex: Especially this part would be better to be in a separate function.

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_outlier_filter/cuda_polar_voxel_outlier_filter.cu
```diff
@@ -0,0 +1,949 @@
...
+        field_data = sqrtf(x * x + y * y + z * z);
+        break;
+      case FieldDataIndex::azimuth:
+        field_data = atan2f(y, x);
+        break;
+      case FieldDataIndex::elevation:
+        field_data = atan2f(z, sqrtf(x * x + y * y));
+        break;
+    }
+  }
```
mojomex, line 273: Is this strictly necessary? Isn't float good enough for all cases? Given that LiDAR has at best 0.1deg angular and ~30mm range accuracy, float should be sufficient for all practical purposes.

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_outlier_filter/cuda_polar_voxel_outlier_filter.cu
```diff
@@ -0,0 +1,949 @@
...
+  CHECK_CUDA_ERROR(cudaStreamSynchronize(stream_));
+
+  types_dev = ReturnTypeCandidates{return_type, types.size()};
+}
+
+std::tuple<int, CudaPooledUniquePtr<::cuda::std::optional<int>>, CudaPooledUniquePtr<int>>
+CudaPolarVoxelOutlierFilter::calculate_voxel_index(
+  const FieldDataComposer<::cuda::std::optional<int32_t> *> & polar_voxel_indices,
+  const size_t & num_points)
+{
```
mojomex, line 745: This function is quite long (~100 lines). Consider breaking it up.

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_outlier_filter/cuda_polar_voxel_outlier_filter.cu
```diff
@@ -0,0 +1,949 @@
...
+    // Because implicit data conversion from cuda::std::optional<int> to bool is not supported by
+    // nvcc (even though std::optional has a operator bool() to check it contains a value),
+    // cub::DeviceAdjacentDifference::SubtractLeftCopy cannot be applied here. To handle this case
+    // execute dedicated CUDA kernel that accepts cuda::std::optional<int> input and bool output
+    dim3 block_dim(512);
+    dim3 grid_dim((num_points + block_dim.x - 1) / block_dim.x);
+    subtract_left_optional_kernel<<<grid_dim, block_dim, 0, stream_>>>(
+      voxel_indices_raw.get(), num_points, voxel_indices_bool.get());
+
+    // calculate mapped index (from geometry-based liner voxel indices to memory-based indices)
```
mojomex:
```suggestion
    // calculate mapped index (from geometry-based linear voxel indices to memory-based indices)
```

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_outlier_filter/cuda_polar_voxel_outlier_filter.cu
```diff
@@ -0,0 +1,949 @@
...
+      stream_);
+
+    // get the number of valid voxels
+    CHECK_CUDA_ERROR(cudaMemcpyAsync(
+      &valid_voxel_num,
+      voxel_indices.get() +
+        (num_points - 1),  // the end of array contains the number of valid voxels
+      sizeof(int), cudaMemcpyDeviceToHost, stream_));
+
+    // Since the current polar_voxel_index_seq contains index value starting from 1,
```
mojomex:
```suggestion
    // Since the current voxel_indices contain index values starting from 1,
```

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_outlier_filter/cuda_polar_voxel_outlier_filter.cu
```diff
@@ -0,0 +1,949 @@
...
+  std::unique_ptr<cuda_blackboard::CudaPointCloud2> & output_cloud)
+{
+  auto [filtered_indices, count] = calculate_filtered_point_indices(points_mask, num_points);
+
+  output_cloud->header = input_cloud->header;
+  output_cloud->fields = input_cloud->fields;
+  output_cloud->is_bigendian = input_cloud->is_bigendian;
+  output_cloud->point_step = input_cloud->point_step;
+  output_cloud->is_dense = input_cloud->is_dense;
+  output_cloud->height = point_cloud_height_organized;
```
mojomex, line 925: Wouldn't a height of `1` be _unorganized_ instead of organized? Maybe this constant needs renaming.

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_outlier_filter/cuda_polar_voxel_outlier_filter.cu
```diff
@@ -0,0 +1,949 @@
...
+  }
+
+  return {std::move(filtered_cloud), std::move(noise_cloud), filter_ratio, visibility};
+}
+
+void CudaPolarVoxelOutlierFilter::set_return_types(
+  const std::vector<int> & types, std::optional<ReturnTypeCandidates> & types_dev)
+{
+  if (types_dev) {
+    // Reset allocated region previously to reflesh the parameters
```
mojomex:
```suggestion
    // Reset previously allocated region to refresh the parameters
```

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_outlier_filter/cuda_polar_voxel_outlier_filter.cu
```diff
@@ -0,0 +1,949 @@
...
+  size_t max_mem_pool_size_in_byte = 1e9;  // 1GB
+  mem_pool_ =
+    autoware::cuda_utils::create_memory_pool(max_mem_pool_size_in_byte, current_device_id);
+}
+
+CudaPolarVoxelOutlierFilter::FilterReturn CudaPolarVoxelOutlierFilter::filter(
+  const cuda_blackboard::CudaPointCloud2::ConstSharedPtr & input_cloud,
+  const CudaPolarVoxelOutlierFilterParameters & params, const PolarDataType polar_type)
+{
+  if (!primary_return_type_dev_) {
```
mojomex, line 542: This is guaranteed to be set in the node constructor, and when changing params at runtime. Thus, this should probably be handled as an error here.

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_outlier_filter/cuda_polar_voxel_outlier_filter.cu
```diff
@@ -0,0 +1,949 @@
...
+  FieldDataComposer<size_t> offsets{};
+  switch (polar_type) {
+    case PolarDataType::PreComputed:
+      offsets.radius = get_offset(input_cloud->fields, "distance");
+      offsets.azimuth = get_offset(input_cloud->fields, "azimuth");
+      offsets.elevation = get_offset(input_cloud->fields, "elevation");
+      break;
+    case PolarDataType::DeriveFromCartesian:
+      // Though struct member names assume polar coordinates one,
+      // fill offset for cartesian coordinates to compute polar coordinates
```
mojomex, line 564: 🥸

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_outlier_filter/cuda_polar_voxel_outlier_filter.cu
```diff
@@ -0,0 +1,949 @@
...
+        cartesian_to_polar_voxel_kernel<float, float><<<grid_dim, block_dim, 0, stream_>>>(
+          input_cloud->data.get(), num_points, input_cloud->point_step, offsets, resolutions,
+          params.min_radius_m, params.max_radius_m, polar_voxel_indices);
+        break;
+      default:
+        throw std::runtime_error("undefined polar_data_type is specified");
+    }
+
+    // calculate unique voxel index that each point belongs to
+    std::tie(num_total_voxels, point_indices, voxel_indices) =
```
mojomex, line 615: std::tie is really cool, I didn't know it existed!

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_outlier_filter/cuda_polar_voxel_outlier_filter.cu
```diff
@@ -0,0 +1,949 @@
...
+    grid_dim = dim3((num_points + block_dim.x - 1) / block_dim.x);
+    classify_point_by_return_type_and_intensity_kernel<uint8_t, uint8_t>
+      <<<grid_dim, block_dim, 0, stream_>>>(
+        input_cloud->data.get(), num_points, num_total_voxels, input_cloud->point_step,
+        return_type_offset, intensity_offset, primary_return_type_dev_.value(),
+        static_cast<uint8_t>(params.intensity_threshold), point_indices.get(), voxel_indices.get(),
+        radius_idx.get(), resolutions.radius, params.visibility_estimation_max_range_m,
+        num_primary_returns.get(), num_secondary_returns.get(), is_in_visibility_range.get(),
+        is_primary_returns.get(), is_secondary_returns.get());
+  }
```
mojomex, line 640: That's a lot of arguments for a single function. Probably breaking it down into smaller functions would help.

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_outlier_filter/cuda_polar_voxel_outlier_filter.cu
```diff
@@ -0,0 +1,949 @@
...
+{
+  return count <= static_cast<size_t>(threshold);
+}
+
+__device__ [[nodiscard]] inline bool meets_intensity_threshold(
+  const uint8_t intensity, const uint8_t threshold)
+{
+  return intensity <= threshold;
+}
+
```
mojomex, line 150: The naming is a bit confusing for me. As far as I understand, secondary returns below the threshold get filtered out. So, I would expect that anything **above** the threshold is considered to "meet the threshold". This naming suggests the opposite. Maybe something like `is_below_threshold` would be clearer?

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_outlier_filter/cuda_polar_voxel_outlier_filter_node.cpp
manato: Agree! Addressed at a271740

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_outlier_filter/cuda_polar_voxel_outlier_filter_node.cpp
manato: Thanks! Because this part is common with the CPU version, and I would like to align with it, let me leave this point as a future task.

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_outlier_filter/cuda_polar_voxel_outlier_filter_node.cpp
manato: Agree! Addressed at 7e208

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_outlier_filter/cuda_polar_voxel_outlier_filter.cu
manato: Thanks for your suggestion. Addressed at b8f795b

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_outlier_filter/cuda_polar_voxel_outlier_filter.cu
manato: I modified the filter to return empty results at 78bfbb2

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_outlier_filter/cuda_polar_voxel_outlier_filter.cu
manato: Thanks! I addressed this at a1c27b4

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_outlier_filter/cuda_polar_voxel_outlier_filter.cu
manato: Thanks for your suggestion. 
Actually, this function is called from CUDA threads that treat each field (i.e., `azimuth`, `elevation`, and `distance`) in parallel. For changing the structure to use them directly requires a large construction, so forgive me for leaving the item as future consideration.

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_outlier_filter/cuda_polar_voxel_outlier_filter.cu
manato: Addressed at ccc6cd1

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor_node.cpp
manato: I guess enqueueing GPU tasks from the same thread using the same CUDA stream can minimize some overhead for better performance. But yes, from the viewpoint of portability and package usability, I agree to separate the node. Let me put the discussion into our backlog!

sensing/autoware_cuda_utils/include/autoware/cuda_utils/cuda_memory_pool.hpp
manato: Thanks for pointing this out. Addressed at b1a835e

manato:
@mojomex 
I really appreciate your detailed review.
I treated some of your suggestions, but some are still unresolved. My apologies, but let me treat them in the future refactoring/updating :pray:

drwnz (approved):
LGTM!

mojomex (approved):

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_outlier_filter/cuda_polar_voxel_outlier_filter.cu
```diff
@@ -0,0 +1,940 @@
...
+      break;
+    case PolarDataType::DeriveFromCartesian:
+      // Though struct member names assume polar coordinates one,
+      // fill offset for cartesian coordinates to compute polar coordinates
+      offsets.radius = get_offset(input_cloud->fields, "x");
+      offsets.azimuth = get_offset(input_cloud->fields, "y");
+      offsets.elevation = get_offset(input_cloud->fields, "z");
+      break;
+    default:
+      throw std::runtime_error("undefined polar_data_type is specified");
```
amadeuszsz, line 570: nit:
Isn't it `std::invalid_argument` more appropriate? Please look also on other throws in this file.

PR #11123, opened 2025-08-06
feat(system_monitor): add host name for system monitor

zhiwango:
## Description

This PR will add a hostname for the system monitor. The reason for these changes is that the system monitor will be run on the other ECUs (such as Sub ECU, Perception ECU, etc).

We have implemented this feature in the old version of Autoware, and it works fine, so I want to resubmit the PR this time.

## Related links

- https://github.com/tier4/autoware_universe/pull/1636

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


zhiwango:
launch fileでsystem_monitorを管理するため、こちらの変更は不要です。Closeさせていただきます。

PR #11132, opened 2025-08-07
fix(multi object tracker): tracker type to associate

technolojin:
## Description

Association was set by the tracker's class. However, the class label and the tracker type is not the same.
This PR explicitly defines the tracker type.

If an "unknown_tracker" has update in classification, it could be a CAR (for example)
If an vehicle detection comes, the vehicle detection will be associated to the unknown_tracker, which cannot estimate the vehicle motion model properly.
By defining tracker type and matrix of detection label -> tracker matrix, make sure given detection is associated to intended tracker type.

* before

https://github.com/user-attachments/assets/7550a913-7dcb-4592-80cb-55080fb54ff8

* after

https://github.com/user-attachments/assets/cdab6330-d18c-451e-b2cb-05f9ba1721c0

* enum `TrackerType` is defined
* association configuration `can_assign_matrix` (label-to-label) is interpreted to `can_assign_map` (label-to-tracker type)
* misc: remove unused `confident_count_threshold`

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_multi_object_tracker/src/multi_object_tracker_node.cpp
+++ b/perception/autoware_multi_object_tracker/src/multi_object_tracker_node.cpp
@@ -195,6 +195,8 @@ MultiObjectTracker::MultiObjectTracker(const rclcpp::NodeOptions & node_options)
         std::make_pair(
           Label::MOTORCYCLE,
           getTrackerType(this->declare_parameter<std::string>("motorcycle_tracker"))));
+      config.tracker_map.insert(
+        std::make_pair(Label::UNKNOWN, TrackerType::UNKNOWN));  // Default for unknown objects
 
       // Declare parameters
       config.tracker_lifetime = declare_parameter<double>("tracker_lifetime");
@@ -228,17 +230,31 @@ MultiObjectTracker::MultiObjectTracker(const rclcpp::NodeOptions & node_options)
     AssociatorConfig associator_config;
     {
       auto initializeMatrixInt = [](const std::vector<int64_t> & vector) {
-        const int label_num = static_cast<int>(std::sqrt(vector.size()));
+        const int label_num = types::NUM_LABELS;
+        if (vector.size() != label_num * label_num) {
+          throw std::runtime_error("Invalid can_assign_matrix size");
+        }
         std::vector<int> converted_vector(vector.begin(), vector.end());
-        Eigen::Map<Eigen::MatrixXi> matrix_tmp(converted_vector.data(), label_num, label_num);
-        // transpose to make it row-major
-        return matrix_tmp.transpose();
+        // Use row-major mapping to match the YAML layout
+        using RowMajorMatrixXi =
+          Eigen::Matrix<int, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>;
+        Eigen::Map<RowMajorMatrixXi> matrix_tmp(converted_vector.data(), label_num, label_num);
+
+        // Convert to column-major (Eigen's default) for consistency
+        return Eigen::MatrixXi(matrix_tmp);
       };
       auto initializeMatrixDouble = [](const std::vector<double> & vector) {
-        const int label_num = static_cast<int>(std::sqrt(vector.size()));
-        Eigen::Map<const Eigen::MatrixXd> matrix_tmp(vector.data(), label_num, label_num);
-        // transpose to make it row-major
-        return matrix_tmp.transpose();
+        const int label_num = types::NUM_LABELS;
+        if (vector.size() != label_num * label_num) {
+          throw std::runtime_error("Invalid max_dist_matrix size");
+        }
+        // Use row-major mapping to match the YAML layout
+        using RowMajorMatrixXd =
+          Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>;
+        Eigen::Map<const RowMajorMatrixXd> matrix_tmp(vector.data(), label_num, label_num);
+
+        // Convert to column-major (Eigen's default) for consistency
+        return Eigen::MatrixXd(matrix_tmp);
       };
       associator_config.can_assign_matrix =
         initializeMatrixInt(this->declare_parameter<std::vector<int64_t>>("can_assign_matrix"));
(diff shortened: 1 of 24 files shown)


perception/autoware_multi_object_tracker/src/multi_object_tracker_node.cpp
```diff
@@ -159,22 +159,44 @@ MultiObjectTracker::MultiObjectTracker(const rclcpp::NodeOptions & node_options)
...
+        if (tracker_name == "pedestrian_and_bicycle_tracker")
+          return TrackerType::PEDESTRIAN_AND_BICYCLE;
+        if (tracker_name == "normal_vehicle_tracker") return TrackerType::NORMAL_VEHICLE;
+        if (tracker_name == "pedestrian_tracker") return TrackerType::PEDESTRIAN;
+        if (tracker_name == "big_vehicle_tracker") return TrackerType::BIG_VEHICLE;
+        if (tracker_name == "bicycle_tracker") return TrackerType::BICYCLE;
+        if (tracker_name == "pass_through_tracker") return TrackerType::PASS_THROUGH;
+        return TrackerType::UNKNOWN;
+      };
+
```
ogrex, line 175: Using a static const std::unordered_map<std::string, TrackerType>  seems better and more scalable

perception/autoware_multi_object_tracker/src/multi_object_tracker_node.cpp
```diff
@@ -159,22 +159,44 @@ MultiObjectTracker::MultiObjectTracker(const rclcpp::NodeOptions & node_options)
...
-        std::make_pair(Label::BICYCLE, this->declare_parameter<std::string>("bicycle_tracker")));
+        std::make_pair(
+          Label::BICYCLE, getTrackerType(this->declare_parameter<std::string>("bicycle_tracker"))));
       config.tracker_map.insert(
         std::make_pair(
-          Label::MOTORCYCLE, this->declare_parameter<std::string>("motorcycle_tracker")));
+          Label::MOTORCYCLE,
+          getTrackerType(this->declare_parameter<std::string>("motorcycle_tracker"))));
+      config.tracker_map.insert(
+        std::make_pair(Label::UNKNOWN, TrackerType::UNKNOWN));  // Default for unknown objects
```
ogrex, line 200: Just a discussion/question:

Currently, the code uses a mapping like label → string → string → label, which involves multiple conversions to support configurability.
Is this really necessary?

- Changing this mapping seems to usually imply developers modifying the tracking algorithm itself, rather than just tuning parameters.
- Do users actually need to tune this mapping in practice.

perception/autoware_multi_object_tracker/src/multi_object_tracker_node.cpp
```diff
@@ -221,19 +230,33 @@ MultiObjectTracker::MultiObjectTracker(const rclcpp::NodeOptions & node_options)
...
+        return Eigen::MatrixXi(matrix_tmp);
       };
       auto initializeMatrixDouble = [](const std::vector<double> & vector) {
-        const int label_num = static_cast<int>(std::sqrt(vector.size()));
-        Eigen::Map<const Eigen::MatrixXd> matrix_tmp(vector.data(), label_num, label_num);
-        // transpose to make it row-major
-        return matrix_tmp.transpose();
+        const int label_num = types::NUM_LABELS;
+        if (vector.size() != label_num * label_num) {
+          throw std::runtime_error("Invalid max_dist_matrix size");
```
ogrex: not just used for `max_dist_matrix` only

ogrex (commented):
Some comments about the code

perception/autoware_multi_object_tracker/src/multi_object_tracker_node.cpp
technolojin: 1. In the simulation,  `pass_through_tracker` could be used for all types of objects. Therefore, the tracker selection is needed to be configurable.
2. However, in the autoware_launch, Planning simulation do not use this.

If nobody is using the `pass_through_tracker`, we may can remove the tracker per class from configurable parameter.

perception/autoware_multi_object_tracker/src/multi_object_tracker_node.cpp
technolojin: fixed https://github.com/autowarefoundation/autoware_universe/pull/11132/commits/b8f2ff5b6c323a2d05daefb9d9ed1d29d38e0e79

perception/autoware_multi_object_tracker/src/multi_object_tracker_node.cpp
technolojin: fixed https://github.com/autowarefoundation/autoware_universe/pull/11132/commits/b8f2ff5b6c323a2d05daefb9d9ed1d29d38e0e79

perception/autoware_multi_object_tracker/src/multi_object_tracker_node.cpp
ogrex:
```suggestion

      static const std::unordered_map<std::string, TrackerType> TRACKER_TYPE_MAP = {
          {"multi_vehicle_tracker", TrackerType::MULTIPLE_VEHICLE},
          {"pedestrian_and_bicycle_tracker", TrackerType::PEDESTRIAN_AND_BICYCLE},
          {"normal_vehicle_tracker", TrackerType::NORMAL_VEHICLE},
          {"pedestrian_tracker", TrackerType::PEDESTRIAN},
          {"big_vehicle_tracker", TrackerType::BIG_VEHICLE},
          {"bicycle_tracker", TrackerType::BICYCLE},
          {"pass_through_tracker", TrackerType::PASS_THROUGH}
      };
      auto getTrackerType = [](const std::string & tracker_name) -> TrackerType {
        auto it = TRACKER_TYPE_MAP.find(tracker_name);
        return it != TRACKER_TYPE_MAP.end() ? it->second : TrackerType::UNKNOWN;
      };
```

technolojin: Now I understood. fixed https://github.com/autowarefoundation/autoware_universe/pull/11132/commits/92b563a3bbd839c16609a78242f02b9b6cfa0f6e

ogrex (approved):
LGTM

PR #11133, opened 2025-08-07
build: add GitHub workflow to generate the build cache

esteve:
## Description

This PR adds a new workflow to manually trigger the generation of the build cache

## Related links

**Parent Issue:**

- https://github.com/autowarefoundation/autoware_universe/issues/10932

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/.github/workflows/generate-build-cache.yaml
+++ b/.github/workflows/generate-build-cache.yaml
@@ -7,7 +7,7 @@ on:
 
 jobs:
   generate-cache:
-    runs-on: "['self-hosted', 'Linux', 'X64']"
+    runs-on: [self-hosted, Linux, X64]
     container: ghcr.io/autowarefoundation/autoware:universe-devel
     env:
       rosdistro: humble


esteve:
@mitsudome-r @xmfcx can you have a look at this PR when you have a moment? I want to make sure it matches what you had in mind. If this is what you expected, I'll submit changes to the `build-and-test` workflows to retrieve the cached folder before they start the build.

xmfcx:
This workflow currently builds the entire `autoware_universe` repository and caches it **once per day**.
That approach has two major drawbacks:

1. **Low update frequency** -> daily runs are not frequent enough to keep up with upstream dependency changes.
2. **Risk of stale or conflicting artifacts** -> if a package in `autoware_universe` is renamed, restoring a cached build will leave both the old and new packages in place, making it harder for CI to detect build issues accurately.

### Why not cache the whole repository?

Caching the main `autoware_universe` workspace directly bakes in whatever was in that branch at the time of caching.
This makes it brittle against:

* Package renames
* Build system changes
* Files left over from previous package structures

### What we should cache instead

The dependencies defined by combining:

* [`build_depends_humble.repos`](https://github.com/autowarefoundation/autoware_universe/blob/main/build_depends_humble.repos)
* [`build_depends_nightly.repos`](https://github.com/autowarefoundation/autoware_universe/blob/main/build_depends_nightly.repos)

Notes:

* `build_depends_humble.repos` is only used for PR validation against the `humble` branch. Therefore there’s no need to cache that on its own.
* PRs against `main` are tested against the **merged** dependency file:
  `build_depends_humble.repos` + `build_depends_nightly.repos`.

`build_depends_nightly.repos` points to **branches** that change frequently. That means our cache must be invalidated when:

* The `.repos` files themselves change, or
* Any of the branches/tags they reference move to a new commit.

## How to detect changes in GitHub Actions

**Key idea:** use `vcs export --exact-with-tags` to freeze all moving references to immutable tags or commit SHAs, and use that frozen manifest as the cache key.

### Cache update workflow (scheduled, **every hour**)

1. Generate the combined `build_depends.repos`.
2. Import it into a temporary workspace.
3. Export an exact manifest (`build-depends-exact.repos`) with pinned versions.
4. Use `hashFiles('build-depends-exact.repos')` as the cache key.
5. If the hash matches the last run, exit early (nothing to rebuild).
6. Otherwise, build dependencies and save the `install/` and `build/` directories to the cache.

### PR workflow (build & test differential)

1. Generate the combined `build_depends.repos`.
2. Import and export an exact manifest as above.
3. Restore the dependency cache **only if the exact key exists** (no restore-keys fallback).
4. If no exact match, build dependencies from scratch (but do not save cache as only the scheduled workflow updates it).

---

**Example keying logic:**

```yaml
- name: Prepare build_depends.repos (combine)
  uses: ./.github/actions/combine-repos-action
  with:
    base_file: build_depends_humble.repos
    overlay_file: build_depends_nightly.repos
    output_file: build_depends.repos

- name: Resolve exact dependencies
  run: |
    vcs import dependency_ws < build_depends.repos
    vcs export dependency_ws --exact-with-tags > build-depends-exact.repos

- name: Restore deps cache
  uses: actions/cache/restore@v4
  with:
    path: |
      install
      build
    key: deps-ws-${{ runner.arch }}-${{ env.rosdistro }}-${{ env.BUILD_TYPE_CUDA_STATE }}-${{ hashFiles('build-depends-exact.repos') }}

- name: Save deps cache
  uses: actions/cache/save@v4
  with:
    path: |
      install
      build
    key: deps-ws-${{ runner.arch }}-${{ env.rosdistro }}-${{ env.BUILD_TYPE_CUDA_STATE }}-${{ hashFiles('build-depends-exact.repos') }}
```

---

This way, the cache represents **only the exact dependency set** for a given build, and it automatically updates when either the `.repos` files change or any moving branch/tag they reference is updated upstream.

esteve:
@xmfcx 

> This workflow currently builds the entire `autoware_universe` repository and caches it **once per day**.

The original ticket mentions running this daily.

It also says to add a manual trigger, so the `workflow_dispatch` approach should also be part of the workflow.

In any case, thanks for your explanation, though it should go in the ticket itself instead so it can be referenced in the future. IMHO, PR work better with more specific feedback (e.g. line per line)

.github/workflows/generate-build-cache.yaml
```diff
@@ -0,0 +1,88 @@
...
+    env:
+      rosdistro: humble
+      container-suffix: ""
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 1
+
+      - name: Remove exec_depend
```
esteve, line 21: It's better to remove this, we don't need to cache the entire repository, only the build depends.

xmfcx:
Left some comments under the issue as well:
- https://github.com/autowarefoundation/autoware_universe/issues/10932#issuecomment-3192444530

esteve:
@mitsudome-r @xmfcx I've included the feedback in the latest commit. I couldn't see a way to only build packages from a given workspace using the `autowarefoundation/autoware-github-actions/colcon-build@v1`, so I added an option to it (see https://github.com/autowarefoundation/autoware-github-actions/pull/353), but if there's another way of doing it, I'm happy to update this PR.

.github/workflows/generate-build-cache.yaml
```diff
@@ -0,0 +1,86 @@
...
+          overlay_file: build_depends_nightly.repos
+          output_file: build_depends.repos
+
+      - name: Resolve exact dependencies
+        run: |
+          vcs import dependency_ws < build_depends.repos
+          vcs export dependency_ws --exact-with-tags > build-depends-exact.repos
+
+      - name: Build dependencies
+        uses: esteve/autoware-github-actions/colcon-build@colcon-build-base-paths
```
xmfcx, line 74:
```suggestion
        uses: autowarefoundation/autoware-github-actions/colcon-build@v1
```

https://github.com/autowarefoundation/autoware-github-actions/releases/tag/v1.57.0 is pushed, we can update it.

esteve:
@mitsudome-r @xmfcx do you have any more feedback or is this ready to be merged? Thanks.

.github/workflows/generate-build-cache.yaml
```diff
@@ -0,0 +1,86 @@
+name: Generate build cache
+
+on:
+  schedule:
+    - cron: 0 0 * * * # every day at midnight UTC
+  workflow_dispatch:
+
+jobs:
+  generate-cache:
+    runs-on: "['self-hosted', 'Linux', 'X64']"
```
xmfcx:
```suggestion
    runs-on: [self-hosted, Linux, X64]
```

We can simplify like this, the other format was used for passing the contents as a parameter.

xmfcx (approved):
Once that small change is made, we can merge this and test if it will succeed. The new workflow should not affect the existing workflows for now.

Also a new pre-step is now added to build the dependencies first with this PR.

esteve:
@xmfcx thanks for the feedback, I've just applied the change you suggested. I think `build-and-test-packages-above-differential` test is failing because there are no modified packages and DCO is failing because when applying changes via the website, the commit is not signed off. Shall we merge this PR anyway?

xmfcx:
https://github.com/autowarefoundation/autoware_universe/actions/runs/19411914835/job/55534497715?pr=11133#step:9:36

`fatal: origin/main...HEAD: no merge base` this is some flaky git issue we have since we limited the checkout depth, generally gets fixed when it is a fresh rebase. Dont mind the dco for this PR, DCO is mostly important for outside contributors. We can set it to pass for this PR.

I will rebase to latest main.

xmfcx:
Once the CI passes, it should auto merge 🙏

xmfcx:
Now that it is merged, I was able to test it: https://github.com/autowarefoundation/autoware_universe/actions/runs/19412299645/job/55535395535

```
Run vcs import dependency_ws < build_depends.repos
  vcs import dependency_ws < build_depends.repos
  vcs export dependency_ws --exact-with-tags > build-depends-exact.repos
  shell: sh -e {0}
  env:
    rosdistro: humble
    container-suffix: 
    BUILD_TYPE_CUDA_STATE: nocuda
usage: vcs import [-h] [--input FILE_OR_URL] [--force] [--shallow]
                  [--recursive] [--retry N] [--skip-existing] [--debug] [-w N]
                  [--repos]
                  [path]
vcs import: error: argument path: Path 'dependency_ws' does not exist.
```

At least now we can debug it in further PRs.

PR #11138, opened 2025-08-08
fix(planning_evaluaotr): prevent abnormal value for ttc

kosuke55:
## Description

obstacle_ttc sometimes becomes abnormally large.
The exact reason is unknown (maybe issue on application side. refer to Related Links) , but it may become large when p0 is not initialized, so a check has been added.
In addition, division by extremely small values has been prevented.

## Related links

`obstacle_ttc` osd

**Parent Issue:**

- Link

## Effects on system behavior

None.


xtk8532704 (approved):
Thank you! LGTM.

maxime-clem (approved):

PR #11151, opened 2025-08-12
feat(rear_collision_checker): add parameter to make collision detection behavior configurable

satoshi-ota:
### Change1https://github.com/autowarefoundation/autoware_universe/commit/b594caba4cb275cd31339bcbba82180496240819

This PR adds a new parameter `yaw_th` to the RearCollisionChecker, addressing the limitation that previously only the lanelet’s turn direction was considered, which meant the collision detection logic could run even in locations without an actual risk of turn-related entrapment. By introducing a yaw difference check between the ego vehicle’s orientation and the lane end, the system now operates the collision detection function only in scenarios with a genuine risk, reducing unnecessary detections.

For example, previously, if the turn direction was set to left, the system would still perform collision checks on the vehicle’s left side even when no leftward shift was planned, which could lead to false positives.

image

### Change2 https://github.com/autowarefoundation/autoware_universe/commit/5b4059cba262c44ed038164359e65d3b5fb37cf4

This PR adds a `check_on_unstoppable` parameter to the RearCollisionChecker, allowing control over whether collision checks are performed when the ego cannot stop before the conflict area. By default (false), such checks are skipped, reducing unnecessary detections; when enabled, checks still run even if stopping is infeasible.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

- [x] [[PR check (OTA)][Lexus][CommonScenario_Basic] planning test](https://evaluation.tier4.jp/evaluation/reports/565cf1d5-4e3e-52aa-8fb0-74eadaa88dc8?project_id=prd_jt)
- [x] [[PR check (OTA)][Lexus][FuncVerificationScenario_Basic] planning test](https://evaluation.tier4.jp/evaluation/reports/487024b5-419c-532f-bd7a-fd8b2d8ea287?project_id=prd_jt)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/planning_validator/autoware_planning_validator_rear_collision_checker/README.md
+++ b/planning/planning_validator/autoware_planning_validator_rear_collision_checker/README.md
@@ -1,0 +1,123 @@
+# Rear Collision Checker
+
+The `rear_collision_checker` is a plugin module of the `autoware_planning_validator` node. It validates the planned trajectory by verifying that it does **not** lead to a collision with other road users (primarily vehicles, but also pedestrian/cyclists/motorcycles) approaching from lateral/rear directions. In particular, it is designed to detect potential rear-end collisions and entrapment risks when the ego vehicle is making a right/left turn or merging into an adjacent lane.
+
+## Inner Workings
+
+The module operates by:
+
+1. **Identifying conflict lanes**
+   - Uses the current route and lanelet topology to determine lanes that the ego vehicle’s trajectory may intersect with during right/left turns or lane changes.
+   - Focuses on lanes where vehicles, bicycles, or motorcycles could approach from behind or from a lateral direction, posing a potential rear-end or entrapment risk.
+   - Considers the `turn_direction` of the approaching lanelets to filter out irrelevant lanes.
+
+2. **Filtering perception data**
+   - Subscribes to obstacle pointcloud and first filters by the conflict region, keeping only points that fall within relevant spatial bounds.
+   - Performs **clustering** on the filtered points to form obstacle candidates.
+   - Associates clusters with nearby lanelets and computes the **nearest face of each obstacle along the lane direction**.
+   - Applies configurable range gates (forward/backward, lateral, height) and basic outlier rejection to discard distant or irrelevant clusters.
+
+3. **Estimating motion**
+   - For each selected object, determines its motion relative to the lane direction.
+   - Estimates speed and direction based on frame-to-frame position changes along the lane’s centerline.
+   - Maintains a per-lane tracking list to stabilize velocity estimation and reduce noise from perception flicker.
+
+4. **Calculating metrics**
+   - **Depending on the selected metric (e.g., TTC, RSS), the calculation method and decision logic differ.** The chosen metric is computed, and the result is used to determine whether there is a potential collision risk.
+
+5. **Judging collision risk**
+   - Evaluates the selected metric results against predefined thresholds to determine whether a collision is likely.
+   - Applies hysteresis logic:
+     - **`on_time_buffer`**: Hazardous state must persist for this duration before marking the trajectory unsafe.
+     - **`off_time_buffer`**: Safe state must persist for this duration before clearing the unsafe flag.
+   - If `check_on_unstoppable=false`, the module will skip collision warnings in cases where the ego vehicle cannot realistically stop before the conflict area (to prevent unnecessary alerts when stopping is already infeasible).
+
+### Flowchart
+
+WIP
+
+## General Parameters
+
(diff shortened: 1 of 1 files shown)


planning/planning_validator/autoware_planning_validator_rear_collision_checker/src/utils.cpp
```diff
@@ -358,6 +359,16 @@ auto check_turn_behavior(
     debug.stoppable_line = stoppable_point.value().first;
   }
 
+  const auto get_yaw_diff_between_ego_and_lane_end =
+    [&ego_pose](const auto & lanelet) -> std::optional<double> {
+    if (lanelet.centerline().size() < 1) return std::nullopt;
```
maxime-clem:
```suggestion
    if (lanelet.centerline().size() < 2) return std::nullopt;
```

planning/planning_validator/autoware_planning_validator_rear_collision_checker/src/utils.cpp
```diff
@@ -386,10 +397,23 @@ auto check_turn_behavior(
...
     if (turn_direction == "left" && p.check.left) {
       const auto sibling_straight_lanelet = get_sibling_straight_lanelet(lane, routing_graph_ptr);
+      const auto yaw_diff = get_yaw_diff_between_ego_and_lane_end(lane);
 
-      if (exceed_dead_line(sibling_straight_lanelet, distance, false)) {
+      if (yaw_diff.has_value()) {
+        if (yaw_diff.value() > -1.0 * p.check.yaw_th) {
+          continue;
+        }
+      }
```
maxime-clem:
```suggestion
      if (yaw_diff.has_value() && yaw_diff.value() > -1.0 * p.check.yaw_th) {
        continue;
      }
```

planning/planning_validator/autoware_planning_validator_rear_collision_checker/src/utils.cpp
```diff
@@ -405,8 +429,17 @@ auto check_turn_behavior(
...
     if (turn_direction == "right" && p.check.right) {
       const auto sibling_straight_lanelet = get_sibling_straight_lanelet(lane, routing_graph_ptr);
+      const auto yaw_diff = get_yaw_diff_between_ego_and_lane_end(lane);
 
-      if (exceed_dead_line(sibling_straight_lanelet, distance, true)) {
+      if (yaw_diff.has_value()) {
+        if (yaw_diff.value() < p.check.yaw_th) {
+          continue;
+        }
+      }
```
maxime-clem:
```suggestion
      if (yaw_diff.has_value() && yaw_diff.value() < p.check.yaw_th) {
        continue;
      }
```

satoshi-ota:
@maxime-clem Thanks for the review. I’m planning to add a minimal README that includes explanations for each parameter in this PR.

planning/planning_validator/autoware_planning_validator_rear_collision_checker/src/utils.cpp
satoshi-ota: Fixed in https://github.com/autowarefoundation/autoware_universe/pull/11151/commits/412d08e3af4b8c33e1d1a53838f1255775f88e4e.

planning/planning_validator/autoware_planning_validator_rear_collision_checker/src/utils.cpp
satoshi-ota: Fixed in https://github.com/autowarefoundation/autoware_universe/pull/11151/commits/412d08e3af4b8c33e1d1a53838f1255775f88e4e.

planning/planning_validator/autoware_planning_validator_rear_collision_checker/src/utils.cpp
satoshi-ota: Fixed in https://github.com/autowarefoundation/autoware_universe/pull/11151/commits/412d08e3af4b8c33e1d1a53838f1255775f88e4e.

satoshi-ota:
> Please add some explanation of the `check_on_unstoppable` parameter in the README. It is a bit confusing because the check is enabled when the parameter is set to `false`.

@maxime-clem I added README in https://github.com/autowarefoundation/autoware_universe/pull/11151/commits/13527b88febccb641a532f5c81028adf6ed7e83d.

`check_on_unstoppable` is disabled by default, so the behavior remains unchanged. However, depending on future discussions, this flag may be set to `true`.

maxime-clem (approved):

Issue #11159, opened 2025-08-12
hdd_monitor fails to retrieve statistics data from eMMC storage devices


nishikawa-masaki:
### Checklist

- [x] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [x] I've searched other issues and no duplicate issues were found.
- [x] I'm convinced that this is not my fault but a bug.

### Description

The "hdd_monitor" node of  the "autoware_system_monitor" component fails to retrieve statistics information (ex. read/write data rate, S.M.A.R.T. data) from eMMC storages, which are used by NVIDIA Jetson AGX Orin.

### Expected behavior

All of the HDD statistics data entries are retrieved from the storage devices on the supported platforms.

Currently, there is no problem with SCSI I/F storages (/dev/sd*) and NVMe SSD (/dev/nvme*) , while NVIDIA Jetson AGX Orin has only an eMMC storage device (/dev/mmcblk*), which the "hdd_monitor" node can't handle.

### Actual behavior

The "/diagnostics" topic published by the "hdd_monitor" node on NVIDIA Jetson AGX Orin includes the following entries:
- Reported as “hdd_reader error” with the reason "No such file or directory”
  - HDD Temperature
  - HDD PowerOnHours
  - HDD TotalDataWritten
  - HDD RecoveredError
- Reported as “stat file read error”
  - HDD ReadDataRate
  - HDD WriteDataRate
  - HDD ReadIOPS
  - HDD WriteIOPS
- Works fine
  - HDD Usage
  - HDD Connection

### Steps to reproduce

1. Execute "sudo hdd_reader" to run the "hdd_reader" application in the background, which retrieves S.M.A.R.T. data from storage devices.
2. On terminal 1, execute "ros2 launch autoware_system_monitor system_monitor.launch.xml" .
3. On terminal 2, execute "ros2 topic echo -f /diagnostics > /tmp/log" .
4. Wait 10 seconds and terminate the  "ros2 topic" command on terminal 2 with Ctrl-C.
5. Investigate the "/tmp/log" file. The "hdd_monitor" reports errors in "/diagnostics" topic.
6. Be sure to terminate the "hdd_reader" process running in the background and the "ros2 launch" process.

### Versions

- OS: Ubunts 22.4
- ROS: Humble
- Autoware: autoware_universe 0.47.0

### Possible causes

[...]

PR #11160, opened 2025-08-12
Fix/hdd monitor emmc storage type support

nishikawa-masaki:
## Description
Add "/dev/mmcblk" as a base for storage device file names, in addition to "/dev/sd" and "/dev/nvme" currently supported.
This addition enables the "hdd_monitor" node to retrieve statistics data about eMMC storage and SD cards.

WIthout this change, the "hdd_monitor" can't work on NVIDIA Jetson AGX Orin.

## Related links

**Parent Issue:**

- [hdd_monitor fails to retrieve statistics data from eMMC storage devices #11159](https://github.com/autowarefoundation/autoware_universe/issues/11159)

**Private Links:**

- [TIER IV internal link](https://tier4.atlassian.net/browse/RT1-10979)

## How was this PR tested?

- Tested on NVIDIA Jetson AGX Orin with JetPack 6.2 (Ubuntu 22.4)
  - Statistics information about eMMC storage device is retrieved as expected.
- Tested on Intel PC with Ubuntu 22.4 (NVMe SSDs are installed)
  - There is no change related to the behavior of the "hdd_monitor" node.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/system/autoware_system_monitor/src/hdd_monitor/hdd_monitor.cpp
+++ b/system/autoware_system_monitor/src/hdd_monitor/hdd_monitor.cpp
@@ -38,6 +38,20 @@
 #include <utility>
 #include <vector>
 
+namespace {
+
+bool is_non-scsi_device(const std::string & device_name)
+{
+  // cspell:disable
+  // clang-format off
+  return (boost::starts_with(device_name, "/dev/nvme") ||   // NVMe SSD
+          boost::starts_with(device_name, "/dev/mmcblk"));  // SD card, eMMC
+  // cspell:enable
+  // clang-format on
+}
+
+}  // namespace
+
 namespace bp = boost::process;
 
 HddMonitor::HddMonitor(const rclcpp::NodeOptions & options)
@@ -848,7 +862,7 @@ void HddMonitor::updateHddConnections()
           const std::regex pattern("\\d+$");
           hdd_param.second.disk_device_ =
             std::regex_replace(hdd_param.second.part_device_, pattern, "");
-        } else if (boost::starts_with(hdd_param.second.part_device_, "/dev/nvme")) {
+        } else if (is_non_scsi_device(hdd_param.second.part_device_)) {
           const std::regex pattern("p\\d+$");
           hdd_param.second.disk_device_ =
             std::regex_replace(hdd_param.second.part_device_, pattern, "");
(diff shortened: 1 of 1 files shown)


system/autoware_system_monitor/CMakeLists.txt
```diff
@@ -216,6 +216,13 @@ if(BUILD_TESTING)
 
   target_link_libraries(test_cpu_monitor cpu_monitor_lib ${Boost_LIBRARIES} ${LIBRARIES})
 
+  set_tests_properties(
+    test_process_monitor
+    test_cpu_monitor
+    PROPERTIES
+      RESOURCE_LOCK system_monitor
+  )
+
```
nishikawa-masaki: Self comment:
This PR fails in CI test with test_arm_cpu_monitor.cpp, which is irrelevant to this PR.
These lines are added to avoid parallel execution of unit tests, which was found **not** to be the cause of the problem.
They will be removed later when the CI test problem is fixed.

nishikawa-masaki: This change didn't fix the CI test problem.
Therefore, it was reverted.

system/autoware_system_monitor/src/hdd_monitor/hdd_monitor.cpp
```diff
@@ -287,8 +302,12 @@ void HddMonitor::checkUsage(diagnostic_updater::DiagnosticStatusWrapper & stat)
...
 
     // Invoke shell to use shell wildcard expansion
+    // because the very initial version of this code used wildcard expansion
+    // to handle incomplete device file names,
+    // but we don't need wildcard expansion anymore after a specification change.
+    // The string "part_device_" has a full device file name derived from the mount point.
     bp::child c(
-      "/bin/sh", "-c", fmt::format("df -Pm {}*", itr->second.part_device_.c_str()),
+      "/bin/sh", "-c", fmt::format("df -Pm {}", itr->second.part_device_.c_str()),
       bp::std_out > is_out, bp::std_err > is_err);
```
nishikawa-masaki, line 311: Self comment:
The asterisk in the script caused a problem on NVIDIA Jetson AGX Orin.
The device file /dev/mmcblk0p10 (/blk/efi) was reported in addition to the expected device file /dev/mmcblkp1 (/) because the asterisk expanded the string.
Please see the related issue tickets for the detail.

TetsuKawa (approved):
Approve for now

PR #11164, opened 2025-08-13
fix(autoware_radar_objects_adapter): object orientation availability

technolojin:
## Description

The object orientation availability was missing. 
This PR adds the orientation availability to AVAILABLE

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


sensing/autoware_radar_objects_adapter/src/radar_objects_adapter.cpp
```diff
@@ -263,8 +263,18 @@ void RadarObjectsAdapter::parse_as_detections(
...
+    // Populate common fields
     const auto & yaw = input_object.orientation;
     populate_common_fields(input_object, output_object, yaw);
+
+    // Set flags for kinematics
+    output_object.kinematics.has_position_covariance = true;
+    output_object.kinematics.orientation_availability =
+      autoware_perception_msgs::msg::DetectedObjectKinematics::AVAILABLE;
+    output_object.kinematics.has_twist = true;
+    output_object.kinematics.has_twist_covariance = true;
```
YoshiRi, line 275: [just a comment]
This part could highly depend on original radar message definition, this is now correct for our continental radar but may change for the future.

sensing/autoware_radar_objects_adapter/src/radar_objects_adapter.cpp
```diff
@@ -263,8 +263,18 @@ void RadarObjectsAdapter::parse_as_detections(
   for (const auto & input_object : input_msg.objects) {
     autoware_perception_msgs::msg::DetectedObject output_object;
 
+    // Populate common fields
     const auto & yaw = input_object.orientation;
     populate_common_fields(input_object, output_object, yaw);
+
+    // Set flags for kinematics
+    output_object.kinematics.has_position_covariance = true;
```
YoshiRi, line 271: [Just a comment]
We only define array number in `RadarObject` msg, so I prefer to do sanity check in this node. (ex: if all of array number is 0, covariance is unavailable)

```msg
float32[6] position_covariance
float32[6] velocity_covariance
float32[6] acceleration_covariance
float32[6] size_covariance
```

sensing/autoware_radar_objects_adapter/src/radar_objects_adapter.cpp
YoshiRi, line 275: same comment as above

PR #11169, opened 2025-08-13
feat(mission_planner): manual lane selection

PanConChicharron:
## Description
Allows the user to force a lane change using an rviz UI.

## Related links

**Parent Issue:**

- [Link](https://tier4.atlassian.net/browse/RT1-10751)

## How was this PR tested?
[Left, Right, and Auto mode](https://github.com/user-attachments/assets/f19b8e78-b6d1-443d-9cec-f4ecfa1193af)
[Goal and turn reachability](https://github.com/user-attachments/assets/4973ac62-ce8b-4af4-89dc-2e82abe7f08d)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/autoware_mission_planner_universe/src/mission_planner/mission_planner.cpp
+++ b/planning/autoware_mission_planner_universe/src/mission_planner/mission_planner.cpp
@@ -248,7 +248,7 @@ void MissionPlanner::on_modified_goal(const PoseWithUuidStamped::ConstSharedPtr 
   }
 
   original_route_ = std::nullopt;
-  
+
   change_route(route);
   change_state(RouteState::SET);
   RCLCPP_INFO(get_logger(), "Changed the route with the modified goal");


planning/autoware_mission_planner_universe/README.md
PanConChicharron, line 1: @takayuki5168 here is the README update

isamu-takagi (approved):
The question I asked [here](https://github.com/autowarefoundation/autoware_universe/pull/11043#issuecomment-3141279624) has now been clarified. I'm not familiar with the detailed routing algorithm, so I'll leave that to other reviewers.

planning/autoware_mission_planner_universe/CMakeLists.txt
```diff
@@ -16,6 +16,7 @@ rclcpp_components_register_node(goal_pose_visualizer_component
 ament_auto_add_library(${PROJECT_NAME}_component SHARED
   src/mission_planner/arrival_checker.cpp
   src/mission_planner/service_utils.cpp
+  src/mission_planner/manual_lane_change_handler.cpp
```
takayuki5168: Currently, both `lane_change` and `lane_selection` are used in this PR. Please unify the name.

PanConChicharron: Done.

visualization/tier4_manual_lane_change_rviz_plugin/src/tier4_manual_lane_change_rviz_plugin.hpp
mitsudome-r, line 1: Does this header file have to be exported for other packages? Otherwise, I think we can put it under src directory.

PanConChicharron: Done.

planning/autoware_manual_lane_change_handler/package.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0"?>
+<package format="3">
+  <name>autoware_manual_lane_change_handler</name>
+  <version>1.0.0</version>
+  <description>Manual lane change handler node extracted from autoware_mission_planner_universe</description>
+  <maintainer email="you@autoware.org">Your Name</maintainer>
```
TaikiYamada4: @PanConChicharron 
Could you add your information here?

PanConChicharron: Done

visualization/tier4_manual_lane_change_rviz_plugin/package.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+  <name>tier4_manual_lane_change_rviz_plugin</name>
+  <version>0.0.0</version>
+  <description>TODO: Package description</description>
+  <maintainer email="arjun.ram@tier4.jp">arjunram</maintainer>
+  <license>TODO: License declaration</license>
```
TaikiYamada4: Please fill the information here.

planning/autoware_manual_lane_change_handler/README.md
```diff
@@ -0,0 +1,76 @@
+# Manual Lane Change Handler
+
+## Purpose
+
+`Manual Lane Change Handler` helps shift the preferred-primitives of a route to the left or right.
+
```
zulfaqar-azmi-t4, line 14: It would be helpful to include a section explaining how to use the feature, so
1. Example of the command that should be used to call it (there is already `Service` subsection, but it's nicer to have a command that we can copy directly as well).
2. Since there's RVIZ plugin, at least mention it here.

planning/autoware_manual_lane_change_handler/src/manual_lane_change_handler.cpp
```diff
@@ -0,0 +1,248 @@
...
+
+#include <deque>
+#include <string>
+#include <vector>
+
+namespace autoware::manual_lane_change_handler
+{
+
+std::vector<autoware_planning_msgs::msg::LaneletPrimitive>
+ManualLaneChangeHandler::sortPrimitivesLeftToRight(
```
zulfaqar-azmi-t4:
```suggestion
ManualLaneChangeHandler::sort_primitives_left_to_right(
```

nits: snake_case

planning/autoware_mission_planner_universe/README.md
```diff
@@ -198,6 +199,10 @@ The interface for the MRM that pulls over the road shoulder. It has to be stoppe
 This is a goal change to pull over, avoid parked vehicles, and so on by a planning component. If the modified goal is outside the calculated route, a reroute is required. This goal modification is executed by checking the local environment and path safety as the vehicle actually approaches the destination. And this modification is allowed for both normal_route and mrm_route.
 The new route generated here is sent to the AD API so that it can also be referenced by the application. Note, however, that the specifications here are subject to change in the future.
 
+#### Setting Preferred Lane
+
+This service allows external modification to the preferred-lanes of the current-route.
```
zulfaqar-azmi-t4: nits: instead of `This service`, `manual lane change service` is more explicit? I'll leave this to you.

planning/autoware_manual_lane_change_handler/launch/manual_lane_change_handler.launch.xml
```diff
@@ -0,0 +1,25 @@
+<launch>
+  <arg name="modified_goal_topic_name" default="/planning/scenario_planning/modified_goal"/>
+  <arg name="map_topic_name" default="/map/vector_map"/>
+  <arg name="route_topic_name" default="/planning/route"/>
+  <arg name="route_state_topic_name" default="/planning/route_state"/>
+  <arg name="clear_route_service_name" default="/planning/clear_route"/>
+  <arg name="set_lanelet_route_service_name" default="/planning/set_lanelet_route"/>
+  <arg name="set_waypoint_route_service_name" default="/planning/set_waypoint_route"/>
```
mitsudome-r: These arguments are unused. Make sure you only specify used arguments.

planning/autoware_manual_lane_change_handler/launch/manual_lane_change_handler.launch.xml
```diff
@@ -0,0 +1,25 @@
...
+  <arg name="visualization_topic_name" default="/planning/mission_planning/route_marker"/>
+  <arg name="mission_planner_param_path" default="$(find-pkg-share autoware_mission_planner_universe)/config/mission_planner.param.yaml"/>
```
mitsudome-r: You should change the parameter name to something like manual_lane_change_handler

planning/autoware_manual_lane_change_handler/launch/manual_lane_change_handler.launch.xml
```diff
@@ -0,0 +1,25 @@
...
+
+  <node_container pkg="rclcpp_components" exec="component_container_mt" name="mission_planner_container" namespace="">
```
mitsudome-r: I think the node container is already used by mission_planner. Please check

launch/tier4_planning_launch/launch/planning.launch.xml
```diff
@@ -21,6 +21,13 @@
 
   <group>
     <push-ros-namespace namespace="planning"/>
+    <!-- manual lane change handler module -->
+    <group>
+      <push-ros-namespace namespace="manual_lane_change_handler"/>
+      <include file="$(find-pkg-share autoware_manual_lane_change_handler)/launch/manual_lane_change_handler.launch.xml">
+        <arg name="mission_planner_param_path" value="$(var mission_planner_param_path)"/>
+      </include>
+    </group>
```
mitsudome-r, line 30: I would like to have this under mission_planning namespace.

planning/autoware_manual_lane_change_handler/README.md
```diff
@@ -0,0 +1,76 @@
...
+### Services
```
mitsudome-r, line 15: There are server and client for the ndoe please make sure that you have both of them described.

planning/autoware_mission_planner_universe/src/mission_planner/mission_planner.cpp
```diff
@@ -400,7 +452,7 @@ void MissionPlanner::change_route()
   // pub_marker_->publish();
 }
 
-void MissionPlanner::change_route(const LaneletRoute & route)
+void MissionPlanner::change_route(const LaneletRoute & route, bool emphasise_goal_lanes)
```
mitsudome-r: Could you explain why we have to emphasise the goal lane only when we change route once again?

I thought there were discussion about making this a configurable parameter regardless of reset state.
(cc. @takayuki5168)

planning/autoware_manual_lane_change_handler/src/manual_lane_change_handler.cpp
```diff
@@ -0,0 +1,248 @@
...
+    std::vector<LaneletPrimitive> preferred_primitives;
+    std::transform(
+      original_route_.value()->segments.begin(), original_route_.value()->segments.end(),
+      std::back_inserter(preferred_primitives),
+      [](const auto & segment) { return segment.preferred_primitive; });
+    original_route_ = std::nullopt;
+
+    return {
+      preferred_primitives, true,
+      "Manual lane selection to AUTO is commanded and executed successfully."};
```
mitsudome-r: I rather would like to have this original route handling to be done in the mission_planner side rather than in manual lane change handler.

Manual lane change handler would set "reset" field to to true in the service and send that to mission_planner. mission_planner will reset back to original route whenever it receives the reset request.

planning/autoware_mission_planner_universe/src/mission_planner/mission_planner.cpp
takayuki5168: @mitsudome-r 
This is because the default lane visualization is very hard to see since it's too light. During the manual lane selection, the user wanna see the currently selected lane, so that's why.

But I aslo feel this implementation a bit weird, so if you have any better idea, please give us.

planning/autoware_mission_planner_universe/src/mission_planner/mission_planner.cpp
```diff
@@ -329,6 +331,56 @@ void MissionPlanner::on_set_lanelet_route(
   print_pose_log("set_lanelet_route", odometry_->pose.pose, req->goal_pose);
 }
 
+void MissionPlanner::on_set_preferred_primitive(
+  const autoware_planning_msgs::srv::SetPreferredPrimitive::Request::SharedPtr req,
+  const autoware_planning_msgs::srv::SetPreferredPrimitive::Response::SharedPtr res)
+{
+  if (!current_route_) {
+    using ResponseCode = autoware_adapi_v1_msgs::msg::ResponseStatus;
```
TaikiYamada4: This `using` line should be outside of this `if` block.

planning/autoware_manual_lane_change_handler/src/manual_lane_change_handler.cpp
```diff
@@ -0,0 +1,248 @@
...
+    if (auto match = find_primitive(left->id())) {
+      sorted_primitives.push_front(*match);
+    }
+  }
+
+  sorted_primitives.push_back(preferred_primitive);
+
+  // Walk right lanes
+  for (auto right = route_handler.getRightLanelet(current, true); right;
+       right = route_handler.getRightLanelet(*right, true, true)) {
```
TaikiYamada4: Why only this `getRightLanelet` has a third argument?
This argument is `true` by default anyways, but is it OK for this feature to allow `road_shoulder` lanelets as left or right lanelets?

planning/autoware_manual_lane_change_handler/src/manual_lane_change_handler.hpp
```diff
@@ -0,0 +1,216 @@
...
+      res->status.success = false;
+      res->status.message = lane_change_request_result.message;
+      return;
+    }
+
+    std::shared_ptr<SetPreferredPrimitive::Request> set_preferred_primitive_req =
+      std::make_shared<SetPreferredPrimitive::Request>();
+    set_preferred_primitive_req->preferred_primitives =
+      lane_change_request_result.preferred_primitives;
+    set_preferred_primitive_req->reset = req->lane_change_direction != 2;
```
TaikiYamada4:
```suggestion
    set_preferred_primitive_req->reset = req->lane_change_direction != SetPreferredLane::Request::AUTO;
```

I prefer using enum constants since a plain integer will lose its information.

planning/autoware_manual_lane_change_handler/src/manual_lane_change_handler.hpp
```diff
@@ -0,0 +1,216 @@
...
+  rclcpp::Subscription<LaneletRoute>::SharedPtr sub_route_;
+  rclcpp::Publisher<autoware_internal_debug_msgs::msg::Float64Stamped>::SharedPtr
+    pub_processing_time_;
+
+  pluginlib::ClassLoader<PlannerPlugin> plugin_loader_;
+  std::shared_ptr<PlannerPlugin> planner_;
+
+  nav_msgs::msg::Odometry::ConstSharedPtr odometry_;
+  std::shared_ptr<LaneletRoute> current_route_;
+  std::function<lanelet::ConstLanelet(const int64_t)> get_lanelet_by_id_;
```
TaikiYamada4: I think its better to define `get_lanelet_by_id` as a simple function for readability because it seems there is no merits to keep this as a function object.

planning/autoware_manual_lane_change_handler/src/manual_lane_change_handler.hpp
```diff
@@ -0,0 +1,216 @@
...
+
+    sub_route_ = create_subscription<LaneletRoute>(
+      "/planning/mission_planning/route", rclcpp::QoS(1).transient_local(),
+      [this](const LaneletRoute::ConstSharedPtr msg) {
+        RCLCPP_INFO(logger_, "Received new route with %zu segments", msg->segments.size());
+        auto route = *msg;
+        planner_->updateRoute(*msg);
+
+        std::for_each(route.segments.begin(), route.segments.end(), [&](auto & segment) {
+          const auto & route_handler = planner_->getRouteHandler();
```
TaikiYamada4: Do we need to redefine the `route_handler` every time? I think we can bring this outside of `std::for_each`.

planning/autoware_manual_lane_change_handler/src/manual_lane_change_handler.cpp
```diff
@@ -0,0 +1,248 @@
...
+      current_segment.primitives.begin(), current_segment.primitives.end(),
+      [&](const autoware_planning_msgs::msg::LaneletPrimitive & p) {
+        return p.id == current_segment.preferred_primitive.id;
+      });
+    if (current_it == current_segment.primitives.end()) {
+      throw std::runtime_error(
+        "ManualLaneChangeHandler: Preferred primitive not found in current segment.");
+    }
+
+    // Find the index of the current preferred primitive
```
TaikiYamada4:
```suggestion
    // Find the index of the next preferred primitive
```

planning/autoware_manual_lane_change_handler/src/manual_lane_change_handler.hpp
```diff
@@ -0,0 +1,216 @@
...
+
+    get_lanelet_by_id_ = [&](const int64_t id) {
+      return planner_->getRouteHandler().getLaneletMapPtr()->laneletLayer.get(id);
+    };
+
+    sub_odometry_ = create_subscription<nav_msgs::msg::Odometry>(
+      "~/input/odometry", rclcpp::QoS(1),
+      [this](const nav_msgs::msg::Odometry::ConstSharedPtr msg) { odometry_ = msg; });
+
+    sub_route_ = create_subscription<LaneletRoute>(
```
TaikiYamada4: This subscription callback is too long to write in the `create_subscription` function. You should better independently define a callback function and bind it to here.

TaikiYamada4 (commented):
I will leave my review comments now but I haven't seen the visualization part yet.

Besides these small comments below, I think it is better to write the constructor of `ManualLaneChangeHandler` in the cpp file. There is too much implementation to write in a hpp file. All other implementation of functions (like `publish_processing_time` and `set_preferred_lane` should go to the cpp file.

launch/tier4_planning_launch/launch/planning.launch.xml
PanConChicharron: Done.

planning/autoware_manual_lane_change_handler/launch/manual_lane_change_handler.launch.xml
PanConChicharron: Done.

planning/autoware_manual_lane_change_handler/launch/manual_lane_change_handler.launch.xml
PanConChicharron: Removed this since it's unused.

planning/autoware_manual_lane_change_handler/src/manual_lane_change_handler.cpp
PanConChicharron: Done.

planning/autoware_manual_lane_change_handler/src/manual_lane_change_handler.cpp
PanConChicharron: Done.

planning/autoware_manual_lane_change_handler/src/manual_lane_change_handler.hpp
PanConChicharron: Done.

planning/autoware_manual_lane_change_handler/src/manual_lane_change_handler.hpp
PanConChicharron: Done.

planning/autoware_manual_lane_change_handler/src/manual_lane_change_handler.hpp
PanConChicharron: Done.

planning/autoware_manual_lane_change_handler/src/manual_lane_change_handler.hpp
PanConChicharron: Done

planning/autoware_mission_planner_universe/src/mission_planner/mission_planner.cpp
PanConChicharron: Done.

planning/autoware_mission_planner_universe/README.md
PanConChicharron: Done.

planning/autoware_manual_lane_change_handler/README.md
PanConChicharron: Done.

planning/autoware_manual_lane_change_handler/README.md
PanConChicharron: Done.

visualization/tier4_manual_lane_change_rviz_plugin/package.xml
PanConChicharron: Done.

planning/autoware_manual_lane_change_handler/launch/manual_lane_change_handler.launch.xml
PanConChicharron: Done.

planning/autoware_manual_lane_change_handler/launch/manual_lane_change_handler.launch.xml
PanConChicharron: Done

planning/autoware_manual_lane_change_handler/src/manual_lane_change_handler.cpp
PanConChicharron: Done.

planning/autoware_manual_lane_change_handler/config/manual_lane_change_handler.param.yaml
TaikiYamada4, line 1: Is this file ever used? If not, please delete this file.

planning/autoware_manual_lane_change_handler/launch/manual_lane_change_handler.launch.xml
```diff
@@ -0,0 +1,12 @@
+<launch>
+  <arg name="map_topic_name" default="/map/vector_map"/>
+  <arg name="manual_lane_change_handler_param_path" default="$(find-pkg-share autoware_manual_lane_change_handler)/config/manual_lane_change_handler.param.yaml"/>
+
+  <node_container pkg="rclcpp_components" exec="component_container_mt" name="manual_lane_change_handler_container" namespace="">
```
TaikiYamada4: So currently this is just a node_container that only contains `manual_lane_change_handler_container` but is this necessary? I think we can just load this as a simple node.

planning/autoware_manual_lane_change_handler/src/manual_lane_change_handler.hpp
```diff
@@ -0,0 +1,111 @@
...
+private:
+  std::vector<autoware_planning_msgs::msg::LaneletPrimitive> sort_primitives_left_to_right(
+    const route_handler::RouteHandler & route_handler,
+    autoware_planning_msgs::msg::LaneletPrimitive preferred_primitive,
+    std::vector<autoware_planning_msgs::msg::LaneletPrimitive> primitives);
+
+  lanelet::ConstLanelet get_lanelet_by_id(const int64_t id)
+  {
+    return planner_->getRouteHandler().getLaneletMapPtr()->laneletLayer.get(id);
+  }
```
TaikiYamada4, line 84: Thanks for making the hpp file clean!
Could you also bring `get_lanelet_by_id` to cpp too?

TaikiYamada4 (commented):
@PanConChicharron 
Thank you for reflecting the review comments one by one.
Could you also take a look to these comments?

Actually when I tried this feature by myself, the "AUTO" command leads to different kinds of crashes in my environment. I'm still investigating what is going wrong but I will just leave these comments in advance.

planning/autoware_manual_lane_change_handler/src/manual_lane_change_handler.hpp
PanConChicharron: Actually, since it's a small getter, it might be best to keep it in the hpp: https://stackoverflow.com/questions/23366826/implement-getters-and-setters-in-header-file

If that's okay!

planning/autoware_manual_lane_change_handler/config/manual_lane_change_handler.param.yaml
PanConChicharron: This one is imported by src/universe/autoware_universe/planning/autoware_manual_lane_change_handler/launch/manual_lane_change_handler.launch.xml:

> <launch>
>   <arg name="map_topic_name" default="/map/vector_map"/>
>   <arg name="manual_lane_change_handler_param_path" default="$(find-pkg-share autoware_manual_lane_change_handler)/config/manual_lane_change_handler.param.yaml"/>
> 
>   <node pkg="autoware_manual_lane_change_handler" exec="autoware_manual_lane_change_handler_node" name="manual_lane_change_handler" namespace="">
>     <param from="$(var manual_lane_change_handler_param_path)"/>
>     <remap from="~/input/vector_map" to="$(var map_topic_name)"/>
>     <remap from="~/input/odometry" to="/localization/kinematic_state"/>
>   </node>
> </launch>

planning/autoware_manual_lane_change_handler/launch/manual_lane_change_handler.launch.xml
PanConChicharron: Done.

PanConChicharron:
> @PanConChicharron Thank you for reflecting the review comments one by one. Could you also take a look to these comments?
> 
> Actually when I tried this feature by myself, the "AUTO" command leads to different kinds of crashes in my environment. I'm still investigating what is going wrong but I will just leave these comments in advance.

Rectified the issue, a small bug had crept in.

planning/autoware_mission_planner_universe/src/mission_planner/mission_planner.cpp
PanConChicharron: Changed the UI to a parametrised transparency implementation, as per @TaikiYamada4  -san's suggestion.

planning/autoware_manual_lane_change_handler/config/manual_lane_change_handler.param.yaml
TaikiYamada4: It is imported but it is not used in the code.

TaikiYamada4: It is imported in the launch file but the parameters are not used in the code. Is that intended?

planning/autoware_mission_planner_universe/src/mission_planner/mission_planner.cpp
```diff
@@ -329,6 +331,74 @@ void MissionPlanner::on_set_lanelet_route(
...
+    RCLCPP_INFO(get_logger(), "Cleared the saved original route after reset.");
+
+    change_route(**original_route_);
+    res->status.message = "Successfully set preferred primitive.";
+    res->status.success = true;
+
+    original_route_ = std::nullopt;
+
+    return;
+  }
```
TaikiYamada4, line 388: The mission planner crashes when I press the "AUTO" service twice in a row because it tries to `change_route` against a `nullopt` route. Add a protection for that.

PanConChicharron: Done.

zulfaqar-azmi-t4 (approved):
LGTM.

Note:
Manually changing lanes can break some assumptions in the lane change module, such as the distance to the dead end based on the original preferred lane.
So when a manual lane change is triggered, please be aware that returning to the original preferred lane may fail.

launch/tier4_planning_launch/launch/planning.launch.xml
```diff
@@ -21,6 +21,13 @@
 
   <group>
     <push-ros-namespace namespace="planning"/>
+    <!-- manual lane change handler module -->
+    <group>
+      <push-ros-namespace namespace="mission_planning"/>
+      <include file="$(find-pkg-share autoware_manual_lane_change_handler)/launch/manual_lane_change_handler.launch.xml">
+        <arg name="mission_planner_param_path" value="$(var mission_planner_param_path)"/>
```
mitsudome-r, line 28: This should be `manual_lane_change_handler_param_path` instead of mission_planner_param_path

planning/autoware_manual_lane_change_handler/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
+cmake_minimum_required(VERSION 3.14)
+project(autoware_manual_lane_change_handler)
+
+find_package(ament_cmake_auto REQUIRED)
+find_package(rclcpp REQUIRED)
+find_package(pluginlib REQUIRED)
+find_package(autoware_planning_msgs REQUIRED)
+
+ament_auto_find_build_dependencies()
```
mitsudome-r, line 9: For autoware packages, we recommend using [autoware_cmake](https://github.com/autowarefoundation/autoware_cmake/tree/main/autoware_cmake)

You would have to update package.xml to have  `<buildtool_depend>autoware_cmake</buildtool_depend>` as dependency as well.

```suggestion
find_package(autoware_cmake REQUIRED)
autoware_package()
```

planning/autoware_mission_planner_universe/src/mission_planner/mission_planner.cpp
```diff
@@ -329,6 +331,81 @@ void MissionPlanner::on_set_lanelet_route(
...
+      autoware_adapi_v1_msgs::srv::SetRoute::Response::ERROR_INVALID_STATE,
+      fmt::format(
+        "The size of preferred_primitives ({}) is different from that of the current route ({}).",
+        req->preferred_primitives.size(), current_route_->segments.size()));
+  }
+  if (req->uuid != current_route_->uuid) {
+    throw service_utils::ServiceException(
+      autoware_adapi_v1_msgs::srv::SetRoute::Response::ERROR_INVALID_STATE,
+      "Route UUID does not match the current route.");
+  }
```
mitsudome-r, line 364: UUID check should come before size check.

If the size difference is due to the difference in route ID, the user won't get the error message for ID because exception is already thrown by size check.

PanConChicharron: Done.

planning/autoware_manual_lane_change_handler/CMakeLists.txt
PanConChicharron: Done.

launch/tier4_planning_launch/launch/planning.launch.xml
PanConChicharron: Done.

planning/autoware_mission_planner_universe/src/mission_planner/mission_planner.cpp
```diff
@@ -329,6 +331,83 @@ void MissionPlanner::on_set_lanelet_route(
...
+      autoware_adapi_v1_msgs::srv::SetRoute::Response::ERROR_INVALID_STATE,
+      fmt::format(
+        "The size of preferred_primitives ({}) is different from that of the current route ({}).",
+        req->preferred_primitives.size(), current_route_->segments.size()));
+  }
+
+  if (!req->reset && !original_route_) {
+    RCLCPP_INFO(get_logger(), "Saved the original route for future resets.");
+    original_route_ = std::make_shared<LaneletRoute>(*current_route_);
+  }
```
mitsudome-r, line 369: I was playing around with planning simulation and found a bug with the following procedure:
1. set intial pose and goal so that mission planner calculates route. (Let's assume this to be the route with id=1)
2. trigger manual lane change (either left or right.)
3. give a new goal to mission planner to create new route with id=2
4. trigger reset from manual lane change plugin by clicking Auto
5. route with id=1 is reloaded

This happens because the current code assumes that reset is always triggered before mission planner receives a new goal, which is not promised. 

Instead, we could reset original route whenever we detect original route has different uuid from the current route. (Please be aware that I haven't tested the code myself so additional debugging might be needed)
```suggestion
  // Only update the original_route_ if the current route is different or not yet set
  if (!original_route_ || (current_route_->uuid != original_route_->uuid)) {
    RCLCPP_DEBUG(get_logger(), "New route detected. Saving original route for possible reset.");
    original_route_ = std::make_shared<LaneletRoute>(*current_route_);
  }

```

PanConChicharron: Done.

planning/autoware_mission_planner_universe/src/mission_planner/mission_planner.cpp
```diff
@@ -322,13 +326,93 @@ void MissionPlanner::on_set_lanelet_route(
       ResponseCode::ERROR_REROUTE_FAILED, "New route is not safe. Reroute failed.");
   }
 
+  original_route_ = std::nullopt;
```
mitsudome-r, line 331: Okay, so you are now resetting original_route whenever set_route service is called. 
You might want to reset the original_route_ when on_modified_goal is called as well in that case

PanConChicharron: Done

mitsudome-r (approved):

PR #11170, opened 2025-08-13
feat(autoware_freespace_planner): enhance freespace planner search algorithm and add reparking feature for improved parking accuracy

YCLiu-TD:
### 1. Optimize the heuristic function of the A star searching algorithm
Refine the A* algorithm’s heuristic function so that the generated trajectory becomes progressively straighter as it approaches the final goal, reducing the controller’s tracking burden and thereby minimizing the final pose error.
image

> ### Main Improvements:

- **Introduction of the concept of a final segment​**
The yaw weight (yaw_weight), the distance to the goal extension line (distance_to_goal_extension_weight), and the curvature of the planned path (extra_steering_penalty_factor) will all have increasing penalty weights as the planning distance gets closer to the goal.
This approach produces a straighter path near the goal, enabling the vehicle controller to approach the goal in as straight a manner as possible during actual driving.
- **Introduction of the goal extension line​**
The distance difference between the vehicle and the goal’s extension line, used to help the vehicle align with the goal earlier and reverse in a straight line.
  image
- **Early alignment with the target**
In addition to the distance from the goal’s extension line, the goal’s orientation angle should also be aligned in advance to reduce the yaw difference (yaw_diff) and ensure minimal skew when parking.

### 2. Introduce a reparking function
Emulate human behavior by adding a reparking feature: if the vehicle’s final parking orientation exceeds a specified deviation threshold, trigger a secondary parking maneuver that generates a combined pull‑out-and‑alignment trajectory to correct the angle. Allow up to three corrective attempts to iteratively reduce the error, while alleviating path constraints imposed by nearby obstacles.

https://github.com/user-attachments/assets/a3706994-1693-4c15-a49d-e78998ba9114

> ### Main Improvements:
- **Added misalignment check within the current architecture​**
The Arrival Checker determines whether the vehicle has reached its destination. However, it does not take into account that in the Parking state, the completion of the parking process must first be verified by the Freespace Planner. As a result, when the vehicle is close enough to the destination and stationary, it may be forcibly switched to a stopped state due to this check.
A new condition has now been added: if the current state is in the Parking procedure, the system must wait for the parking_completed message.​
  image
- **Triggered reparking mechanism​**
Once the vehicle has stopped at the goal, if the degree of misalignment exceeds the acceptable threshold (parking_accuracy_tolerance), the reparking function will be activated.
It will replan a straight path moving outward, increasing the weights for alignment (reparking_alignment_weight) and deviation (reparking_deviation_penalty) to slightly correct the angle, align the vehicle, and then approach the goal again.
This process will run for a maximum of three cycles, and if any cycle results in a misalignment smaller than parking_accuracy_tolerance, the parking procedure will be considered complete.
  image

## Related links
**Launch File:**

- PR [Launch](https://github.com/autowarefoundation/autoware_launch/pull/1605)

**Parent Issue:**
- Discussion [6320](https://github.com/orgs/autowarefoundation/discussions/6320)

## How was this PR tested?
This PR has been tested in Psim.
## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_probabilistic_occupancy_grid_map/launch/laserscan_based_occupancy_grid_map.launch.py
+++ b/perception/autoware_probabilistic_occupancy_grid_map/launch/laserscan_based_occupancy_grid_map.launch.py
@@ -114,33 +114,6 @@ def launch_setup(context, *args, **kwargs):
         ),
     ]
 
-    additional_node = ComposableNode(
-        package="autoware_probabilistic_occupancy_grid_map",
-        plugin="autoware::occupancy_grid_map::LaserscanBasedOccupancyGridMapNode",
-        name="finer_occupancy_grid_map_node",
-        remappings=[
-            ("~/input/laserscan", LaunchConfiguration("output/laserscan")),
-            ("~/input/obstacle_pointcloud", LaunchConfiguration("input/obstacle_pointcloud")),
-            ("~/input/raw_pointcloud", LaunchConfiguration("input/raw_pointcloud")),
-            ("~/output/occupancy_grid_map", "finer_occupancy_grid_map"),
-        ],
-        parameters=[
-            laserscan_based_occupancy_grid_map_node_params,
-            occupancy_grid_map_updater_params,
-            {
-                "input_obstacle_pointcloud": LaunchConfiguration("input_obstacle_pointcloud"),
-                "input_obstacle_and_raw_pointcloud": LaunchConfiguration(
-                    "input_obstacle_and_raw_pointcloud"
-                ),
-                "updater_type": LaunchConfiguration("updater_type"),
-                "map_resolution": 0.25,
-            },
-        ],
-        extra_arguments=[{"use_intra_process_comms": LaunchConfiguration("use_intra_process")}],
-    )
-
-    composable_nodes.append(additional_node)
-
     occupancy_grid_map_container = ComposableNodeContainer(
         name=LaunchConfiguration("individual_container_name"),
         namespace="",
--- a/perception/autoware_probabilistic_occupancy_grid_map/launch/laserscan_based_occupancy_grid_map.launch.py
+++ b/perception/autoware_probabilistic_occupancy_grid_map/launch/laserscan_based_occupancy_grid_map.launch.py
@@ -129,9 +129,11 @@ def launch_setup(context, *args, **kwargs):
             occupancy_grid_map_updater_params,
             {
                 "input_obstacle_pointcloud": LaunchConfiguration("input_obstacle_pointcloud"),
-                "input_obstacle_and_raw_pointcloud": LaunchConfiguration("input_obstacle_and_raw_pointcloud"),
+                "input_obstacle_and_raw_pointcloud": LaunchConfiguration(
+                    "input_obstacle_and_raw_pointcloud"
+                ),
                 "updater_type": LaunchConfiguration("updater_type"),
-                "map_resolution": 0.25
+                "map_resolution": 0.25,
             },
         ],
         extra_arguments=[{"use_intra_process_comms": LaunchConfiguration("use_intra_process")}],
--- a/perception/autoware_probabilistic_occupancy_grid_map/launch/laserscan_based_occupancy_grid_map.launch.py
+++ b/perception/autoware_probabilistic_occupancy_grid_map/launch/laserscan_based_occupancy_grid_map.launch.py
@@ -114,6 +114,31 @@ def launch_setup(context, *args, **kwargs):
         ),
     ]
 
+    additional_node = ComposableNode(
+        package="autoware_probabilistic_occupancy_grid_map",
+        plugin="autoware::occupancy_grid_map::LaserscanBasedOccupancyGridMapNode",
+        name="finer_occupancy_grid_map_node",
+        remappings=[
+            ("~/input/laserscan", LaunchConfiguration("output/laserscan")),
+            ("~/input/obstacle_pointcloud", LaunchConfiguration("input/obstacle_pointcloud")),
+            ("~/input/raw_pointcloud", LaunchConfiguration("input/raw_pointcloud")),
+            ("~/output/occupancy_grid_map", "finer_occupancy_grid_map"),
+        ],
+        parameters=[
+            laserscan_based_occupancy_grid_map_node_params,
+            occupancy_grid_map_updater_params,
+            {
+                "input_obstacle_pointcloud": LaunchConfiguration("input_obstacle_pointcloud"),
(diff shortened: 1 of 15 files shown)


perception/autoware_probabilistic_occupancy_grid_map/launch/laserscan_based_occupancy_grid_map.launch.py
```diff
@@ -114,6 +114,33 @@ def launch_setup(context, *args, **kwargs):
         ),
     ]
 
+    additional_node = ComposableNode(
```
technolojin: This function may requires finer grid map.

Can you make it as optional, so that other user not to run not-necessary high cost grid-map?

planning/autoware_freespace_planner/include/autoware/freespace_planner/freespace_planner_node.hpp
```diff
@@ -142,6 +144,8 @@ class FreespacePlannerNode : public rclcpp::Node
   std::vector<size_t> reversing_indices_;
   size_t prev_target_index_;
   size_t target_index_;
+  size_t endpoint_index_;
```
maxime-clem: From what I can see this variable is never used. Can it be removed ?

planning/autoware_freespace_planner/src/autoware_freespace_planner/freespace_planner_node.cpp
```diff
@@ -406,6 +472,9 @@ void FreespacePlannerNode::planTrajectory()
     prev_target_index_ = 0;
     target_index_ = utils::get_next_target_index(
       trajectory_.points.size(), reversing_indices_, prev_target_index_);
+
+    endpoint_index_ = trajectory_.points.size() - 1;
+    reversing_indices_.push_back(trajectory_.points.size() - 1);
```
maxime-clem: From what I understand, `get_next_target_index()`  already returns `trajectory_.points.size() - 1` by default so I do not think this line is necessary.

planning/autoware_freespace_planner/src/autoware_freespace_planner/freespace_planner_node.cpp
```diff
@@ -184,6 +187,29 @@ bool FreespacePlannerNode::checkCurrentTrajectoryCollision()
...
+
+  if (angle > 180.0) {
+    angle -= 360.0;
+  } else if (angle < -180.0) {
+    angle += 360.0;
+  }
+
+  return angle;
+}
+
```
maxime-clem: Please use this function instead: https://github.com/autowarefoundation/autoware_utils/blob/3b0094fa0f4455e63418fb1f86b338e3f7578644/autoware_utils_geometry/include/autoware_utils_geometry/pose_deviation.hpp#L36

planning/autoware_freespace_planner/src/autoware_freespace_planner/freespace_planner_node.cpp
```diff
@@ -200,12 +226,43 @@ void FreespacePlannerNode::updateTargetIndex()
     utils::get_next_target_index(trajectory_.points.size(), reversing_indices_, target_index_);
 
   if (new_target_index == target_index_) {
-    // Finished publishing all partial trajectories
-    is_completed_ = true;
-    RCLCPP_INFO_THROTTLE(get_logger(), *get_clock(), 1000, "Freespace planning completed");
-    std_msgs::msg::Bool is_completed_msg;
-    is_completed_msg.data = is_completed_;
-    parking_state_pub_->publish(is_completed_msg);
+    double angle_goal_current =
```
maxime-clem: (optional) the name of this variable can be improved, e.g., `yaw_error`, `goal_yaw_deviation`, ...

planning/autoware_freespace_planning_algorithms/src/astar_search.cpp
```diff
@@ -539,11 +654,98 @@ bool AstarSearch::isGoal(const AstarNode & node) const
...
+  while (current->parent != nullptr) {
+    const AstarNode * parent = current->parent;
+    Pose parent_pose = node2pose(*parent);
+
+    double dist = autoware_utils::calc_distance2d(current_pose, parent_pose);
+    total_length += dist;
+
+    current = parent;
+    current_pose = parent_pose;
+  }
```
maxime-clem: This function is not used so it can be removed.

perception/autoware_probabilistic_occupancy_grid_map/launch/laserscan_based_occupancy_grid_map.launch.py
maxime-clem: This change looks like a mistake that can be removed.
The node output  (`finer_occupancy_grid_map`) is not used by any Autoware node.

maxime-clem (changes requested):
I tested the reparking with both A* and RRT* and it works well.
Some code improvements are required.

planning/autoware_freespace_planner/src/autoware_freespace_planner/freespace_planner_node.cpp
YCLiu-TD: Removed this line as suggested.

planning/autoware_freespace_planner/include/autoware/freespace_planner/freespace_planner_node.hpp
YCLiu-TD: Removed this variable as suggested.

planning/autoware_freespace_planning_algorithms/src/astar_search.cpp
YCLiu-TD: Removed this function as suggested.

perception/autoware_probabilistic_occupancy_grid_map/launch/laserscan_based_occupancy_grid_map.launch.py
YCLiu-TD: The unused finer_occupancy_grid_map node has been removed since it was not referenced by any other Autoware module.

planning/autoware_freespace_planner/src/autoware_freespace_planner/freespace_planner_node.cpp
YCLiu-TD: I replaced the custom quaternionAngularDifference with autoware_utils_geometry::calc_yaw_deviation.

planning/autoware_freespace_planner/src/autoware_freespace_planner/freespace_planner_node.cpp
YCLiu-TD: I have renamed the variable angle_goal_current to yaw_error for better readability.

PR #11179, opened 2025-08-14
fix(rear_collision_checker): collision detection not triggered when no stop point before conflict area

satoshi-ota:
## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/planning_validator/autoware_planning_validator_rear_collision_checker/src/utils.cpp
+++ b/planning/planning_validator/autoware_planning_validator_rear_collision_checker/src/utils.cpp
@@ -216,6 +216,7 @@ auto check_shift_behavior(
   const auto & points = context->data->current_trajectory->points;
   const auto & ego_pose = context->data->current_kinematics->pose.pose;
   const auto & vehicle_width = context->vehicle_info.vehicle_width_m;
+  const auto & max_longitudinal_offset = context->vehicle_info.max_longitudinal_offset_m;
 
   const auto combine_lanelet = lanelet::utils::combineLaneletsShape(lanelets);
   const auto nearest_idx =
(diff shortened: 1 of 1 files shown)


planning/planning_validator/autoware_planning_validator_rear_collision_checker/src/utils.cpp
```diff
@@ -285,16 +295,28 @@ auto check_shift_behavior(
...
+    }
+
     const auto is_left_shift = boost::geometry::intersects(
       axle, lanelet::utils::to2D(combine_lanelet.leftBound()).basicLineString());
-    if (is_left_shift && std::abs(points.at(i).longitudinal_velocity_mps) > 1e-3) {
-      return std::make_pair((i < nearest_idx ? Behavior::NONE : Behavior::SHIFT_LEFT), distance);
+    if (is_left_shift) {
+      const auto has_stop_point_before_conflict_area =
+        distance_to_stop_point.has_value() &&
+        distance_to_stop_point.value() + max_longitudinal_offset < distance;
```
maxime-clem, line 333: I do not think you need to add `max_longitudinal_offset` here since everything is calculated relative to the base_link of ego, not its front point.

satoshi-ota: I’m sorry for the confusion, but the variable distance represents the distance traveled by the `base_link` plus the distance to the front of the vehicle. Therefore, it is necessary to add `max_longitudinal_offset`.

To avoid confusion, I have documented the function’s specification in the following commit. (https://github.com/autowarefoundation/autoware_universe/pull/11179/commits/1594e33dfd379a35d8cc40d62a82a29ac591c98e)

PR #11198, opened 2025-08-14
fix(boundary_departure_checker): update poses for the new trajectory

maxime-clem:
## Description

Properly recalculate the poses used to insert slowdown intervals in the trajectory.
Previously, we could insert poses calculated on a previous trajectory into a new trajectory, causing the points to not be aligned on the new trajectory which could cause high deceleration.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

Psim

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/motion_velocity_planner/autoware_motion_velocity_boundary_departure_prevention_module/src/utils.cpp
+++ b/planning/motion_velocity_planner/autoware_motion_velocity_boundary_departure_prevention_module/src/utils.cpp
@@ -156,6 +156,8 @@ void update_departure_intervals_poses(
         utils::is_point_shifted(prev_pose, curr_pose, th_pt_shift_dist_m, th_pt_shift_angle_rad)) {
       return true;
     }
+    interval.start.pose =
+      motion_utils::calcInterpolatedPose(raw_ref_traj, interval.start_dist_on_traj);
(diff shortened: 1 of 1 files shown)


planning/motion_velocity_planner/autoware_motion_velocity_boundary_departure_prevention_module/src/utils.cpp
```diff
@@ -141,6 +143,7 @@ void update_departure_intervals_poses(
 
     dpt_pt.end_dist_on_traj =
       motion_utils::calcSignedArcLength(raw_ref_traj, 0UL, dpt_pt.end.pose.position);
+    dpt_pt.end.pose = motion_utils::calcInterpolatedPose(raw_ref_traj, dpt_pt.end_dist_on_traj);
```
zulfaqar-azmi-t4: interval.end.pose is updated in line 162

maxime-clem: Thank you. I fixed it and moved the update of the start pose to be done together with the update of the end pose.

PR #11222, opened 2025-08-15
feat(autoware_calibration_status_classifier): add ML-based miscalibration detection module

amadeuszsz:
## Description

Add a new ML-based module for miscalibration detection.

[Screencast from 09-18-2025 04:50:48 PM.webm](https://github.com/user-attachments/assets/5cc6b9c2-4732-4ed4-8dc8-ae9d736ef2af)

Some interesting demos are in [TIER IV INTERNAL LINK](https://star4.slack.com/archives/C057JDFS5M2/p1756873865887939?thread_ts=1750246394.824509&cid=C057JDFS5M2).

## Related links

* https://github.com/tier4/AWML/tree/main/projects/CalibrationStatusClassification

## How was this PR tested?

* Unit tests
* Rosbags

## Notes for reviewers

* Copy `ml_package_calibration_status_classifier.param.yaml` to `~/autoware_data/calibration_status_classifier/`.
* Copy ONNX file from [TIER IV INTERNAL LINK](https://drive.google.com/file/d/1t0f2W1VHxMraYJO602mpaOT2_MojjKfw/view?usp=drive_link) to `~/autoware_data/calibration_status_classifier/`.
* Build package:
  ```bash
  colcon build --symlink-install --cmake-args -DCMAKE_BUILD_TYPE=Release -DCMAKE_EXPORT_COMPILE_COMMANDS=On --packages-select autoware_calibration_status_classifier
  ```
* For ROS 2 node:
  * Configure param files with relevant topics.
  * Update image decompressor node in launch file if needed.
  * Run rosbag, example for pilot.auto.x2:
  ```bash
  # terminal 1
  webauto data rosbag pull --project-id x2_dev --environment-id b4448f17-3a93-4e5b-b332-ace03e6b690f --rosbag-id e4c77db5-5700-4442-9317-9be5e85a77ca
  ros2 bag play ~/.webauto/data/data/rosbag/e4c77db5-5700-4442-9317-9be5e85a77ca/cfa23601-97c4-4d47-a37e-edfc7e080d8c_2025-05-29-17-00-12_p0900_11.db3 --remap /diagnostics:=/null --clock 100 -r 0.1 --loop
  # terminal 2
  export VEHICLE_ID=j6_gen2_03
  ./cmd_helper.sh --autoware launch_system:=false launch_map:=false launch_sensing_driver:=false launch_localization:=false launch_perception:=false launch_planning:=false launch_control:=false launch_api:=false launch_v2x:=false launch_l4_toolkit:=false use_sim_time:=true launch_vehicle_interface:=false
  ```
  * `ros2 launch autoware_calibration_status_classifier calibration_status_classifier.launch.xml`
* For unit tests with test data:
  *  `colcon test-result --delete-yes && colcon test --packages-select autoware_calibration_status_classifier --event-handlers console_cohesion+`
  * You can preview the output comparison (reference from training pipeline v.s. result from this library) in `~/autoware/install/autoware_calibration_status_classifier/share/autoware_calibration_status_classifier/data`.
  * You can generate custom data for test:
    * Set any path in [binary_save_dir](https://github.com/tier4/AWML/blob/0a7a03140ee2750dc1d7bab0981cdfce163feb1c/projects/CalibrationStatusClassification/configs/t4dataset/resnet18_5ch_1xb8-25e_j6gen2.py#L14).
    * Run deploy script as described in [documentation](https://github.com/tier4/AWML/tree/main/projects/CalibrationStatusClassification#5-deploy).
    * Copy data to `sensing/autoware_calibration_status_classifier/test/data` and add new samples [here](https://github.com/autowarefoundation/autoware_universe/blob/aaec2dd0d217a3e4b782c2457dfdf699a1a94fc2/sensing/autoware_calibration_status/test/test_calibration_status.cpp#L54), [here](https://github.com/autowarefoundation/autoware_universe/blob/aaec2dd0d217a3e4b782c2457dfdf699a1a94fc2/sensing/autoware_calibration_status/test/test_model_inference.cpp#L54) and and [here](https://github.com/autowarefoundation/autoware_universe/blob/aaec2dd0d217a3e4b782c2457dfdf699a1a94fc2/sensing/autoware_calibration_status/test/test_preprocessing.cpp#L70).
    * Rebuild package and run test again.

## Interface changes

None.

## Effects on system behavior

None.


sensing/autoware_calibration_status/include/autoware/calibration_status/config.hpp
```diff
@@ -0,0 +1,90 @@
...
+      static_cast<uint32_t>(height[2])};
+    this->width = {
+      static_cast<uint32_t>(width[0]), static_cast<uint32_t>(width[1]),
+      static_cast<uint32_t>(width[2])};
+  }
+  double max_depth;
+  uint32_t dilation_size;
+  std::vector<uint32_t> height;
+  std::vector<uint32_t> width;
+  uint32_t channels{5};
```
manato:
```suggestion
      static_cast<int32_t>(height[0]), static_cast<int32_t>(height[1]),
      static_cast<int32_t>(height[2])};
    this->width = {
      static_cast<int32_t>(width[0]), static_cast<int32_t>(width[1]),
      static_cast<int32_t>(width[2])};
  }
  double max_depth;
  uint32_t dilation_size;
  std::vector<int32_t> height;
  std::vector<int32_t> width;
  int32_t channels{5};
```

Since `nvinfer1::Dims` is defined using `int32_t`, using `uint32_t` causes `narrowing conversion of ‘...'{aka ‘unsigned int’} to ‘int’` in my environment.

sensing/autoware_calibration_status/lib/calibration_status.cpp
```diff
@@ -0,0 +1,176 @@
...
+      preview_img_data, image_undistorted_d_.get(),
+      sizeof(InputImageBGR8Type) * image_msg->height * image_msg->width, cudaMemcpyDeviceToHost,
+      stream_));
+  }
+  CHECK_CUDA_ERROR(cudaStreamSynchronize(stream_));
+  auto t2 = std::chrono::steady_clock::now();
+
+  // Run inference
+  network_trt_ptr_->setTensor(
+    "input", in_d_.get(), {4, {1, config_.channels, image_msg->height, image_msg->width}});
```
manato:
```suggestion
    "input", in_d_.get(), {4, {1, config_.channels, static_cast<int32_t>(image_msg->height), static_cast<int32_t>(image_msg->width)}});
```
Same as a comment on `include/autoware/calibration_status/config.hpp`, this cast seems to be needed due to `int32_t` usage in  `nvinfer1::Dims` definition

sensing/autoware_calibration_status/lib/preprocess/preprocess_cuda.cu
```diff
@@ -0,0 +1,379 @@
...
+
+cudaError_t PreprocessCuda::projectPoints_launch(
+  const InputPointType * input_points, InputImageBGR8Type * undistorted_image,
+  const double * tf_matrix, const double * projection_matrix, const size_t num_points,
+  const size_t width, const size_t height, float * output_array, uint32_t * num_points_projected)
+{
+  dim3 threads(256);
+  dim3 blocks((num_points + threads.x - 1) / threads.x);
+
+  float * metric_depth_buffer;
```
manato: For allocating a temporal memory region on the GPU memory, I would suggest using `autowaree_cuda_utils::make_unique` with memory pool ([ref](https://github.com/autowarefoundation/autoware_universe/blob/d49316343991a0a1fda85cb7d7e6075182a8954d/sensing/autoware_cuda_utils/include/autoware/cuda_utils/cuda_unique_ptr.hpp#L59-L79)), because it reuses allocated memory regions internally and can avoid issuing expensive actual allocation operations.

You can also see an example of how to create a memory pool [here](https://github.com/autowarefoundation/autoware_universe/blob/d49316343991a0a1fda85cb7d7e6075182a8954d/sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_outlier_filter/cuda_polar_voxel_outlier_filter.cu#L530-L535)

sensing/autoware_calibration_status/lib/preprocess/preprocess_cuda.cu
```diff
@@ -0,0 +1,379 @@
...
+    if (p_cam_z < 0.0 || p_cam_z >= max_depth) {
+      return;
+    }
+
+    // Project 3D point onto the 2D undistorted image plane
+    const int64_t u = lround((fx_new * p_cam_x + cx_new * p_cam_z) / p_cam_z);
+    const int64_t v = lround((fy_new * p_cam_y + cy_new * p_cam_z) / p_cam_z);
+
+    // Projection ROI validation
+    if (u >= 0 && u < static_cast<int64_t>(width) && v >= 0 && v < static_cast<int64_t>(height)) {
```
manato: (Not mandatory to be handled now)
I would prefer to use early return instead of wrapping up the whole process by `if` to avoid deep nesting.
i.e., 
```cuda
if ( u < 0 || width <= u || v < 0 || height <= v) {
    return;
}
// remain process
```
instead of 
```cuda
if (...) {
  // remain process
}
```

sensing/autoware_calibration_status_classifier/lib/preprocess/preprocess_cuda.cu
manato, line 1: For pointer arguments of all CUDA kernels, you can add [`__restrict__` keyword](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html?highlight=__restrict__#restrict) if they are surely in separate regions.
e.g.
```cuda
__global__ void copyImage_kernel(
  const InputImageBGR8Type * __restrict__ input_image, const size_t width, const size_t height,
  InputImageBGR8Type * __restrict__ output_image, float * __restrict__ output_array)
```

sensing/autoware_calibration_status/lib/preprocess/preprocess_cuda.cu
```diff
@@ -0,0 +1,379 @@
...
+  const InputImageBGR8Type * input_image, const double * dist_coeffs, const double * camera_matrix,
+  const double * projection_matrix, const size_t width, const size_t height,
+  InputImageBGR8Type * output_image, float * output_array)
+{
+  dim3 threads(16, 16);
+  dim3 blocks((width + threads.x - 1) / threads.x, (height + threads.y - 1) / threads.y);
+
+  undistortImage_kernel<<<blocks, threads, 0, stream_>>>(
+    input_image, dist_coeffs, camera_matrix, projection_matrix, width, height, output_image,
+    output_array);
```
manato: This kernel call may be replaced by some NVIDIA-provided API, which runs fast in most cases and is flexible for various inputs, in particular, the camera distortion model
- zero-filling all elements in the output buffer
- [`nppiRemap_8u_C3R`](https://docs.nvidia.com/cuda/archive/9.2/npp/group__image__remap.html#ga62d5fb5fcd3e3bc67e06830c41ad339d) for undistortion
- [`nppiCopy_8u_C3P3R`](https://docs.nvidia.com/cuda/archive/10.2/npp/group__image__packed__to__planar__channel__copy.html#gac5f584f230216c6b62e9c5b7e4b7a18b) for divide channel plane


Since nppi operations require aligned memory regions (i.e., it may include a padding area), you may need to handle such a padding area properly when copying the actual data onto the network input

amadeuszsz: Unfortunately, Autoware does not consist of `NPP` and we cannot link against it. Checked with `find /usr/local/cuda* -name "*npp*"`. Can be fixed with `sudo apt-get install libnpp-dev-12-4` but it takes 510 MB additional space. If we can update our ansible scripts with this extra dependency, we will need to wait up to few weeks for next image deployment to make it available in our CI.

Personally, I think we should use NPP as recently we push Autoware towards GPU optimized implementation. I can add this dependency to Autoware and when available, use NPP instead of custom kernels.
https://github.com/autowarefoundation/autoware_universe/pull/11222/commits/2bcf7712ff9b7057fabb9e894b8fbf9513d46866

sensing/autoware_calibration_status/lib/preprocess/preprocess_cuda.cu
amadeuszsz: Addressed in https://github.com/autowarefoundation/autoware_universe/pull/11222/commits/a22c19551dfedfcee7320af7de6ab1c56c31c6fd

sensing/autoware_calibration_status_classifier/lib/preprocess/preprocess_cuda.cu
amadeuszsz: Addressed in https://github.com/autowarefoundation/autoware_universe/pull/11222/commits/059a9e4164de25b0ac769e108a7f28debd1eaac1

sensing/autoware_calibration_status/lib/preprocess/preprocess_cuda.cu
amadeuszsz: Nice catch! I think a cleaner approach will be to use class member.
Addressed in https://github.com/autowarefoundation/autoware_universe/pull/11222/commits/8a8c54c78d4976d06929e9e7093bcb0b8e296b70 and https://github.com/autowarefoundation/autoware_universe/pull/11222/commits/6ff6a5739ce019770a959c896f9ddc9d9057f9fc and https://github.com/autowarefoundation/autoware_universe/pull/11222/commits/4c07af8b1f8b27a58a2fd65ca127bc6f08244142

sensing/autoware_calibration_status/src/calibration_status_node.cpp
```diff
@@ -0,0 +1,478 @@
...
+    this->declare_parameter<std::vector<std::string>>("input.image_topics"),
+    this->declare_parameter<std::vector<double>>("input.approx_deltas"),
+    this->declare_parameter<std::vector<bool>>("input.already_rectified"));
+
+  auto camera_lidar_info_collector =
+    std::make_shared<CameraLidarInfoCollector>(this, camera_lidar_in_out_info_);
+  camera_lidar_info_ = camera_lidar_info_collector->get_cameras_lidars_info();
+
+  // Diagnostic publisher
+  diagnostic_pub_ =
```
vividf: Please use autoware_utils::DiagnosticsInterface for consistency

sensing/autoware_calibration_status_classifier/src/calibration_status_classifier_node.cpp
```diff
@@ -0,0 +1,478 @@
...
+  preview_image_pubs_.resize(num_pairs);
+  synchronizers_.resize(num_pairs);
+  synchronized_data_.resize(num_pairs);
+
+  for (size_t i = 0; i < num_pairs; ++i) {
+    cloud_subs_.at(i) =
+      std::make_shared<message_filters::Subscriber<sensor_msgs::msg::PointCloud2>>(
+        this, camera_lidar_in_out_info_.at(i).lidar_topic, rmw_qos_profile_sensor_data);
+    image_subs_.at(i) = std::make_shared<message_filters::Subscriber<sensor_msgs::msg::Image>>(
+      this, camera_lidar_in_out_info_.at(i).camera_topic, rmw_qos_profile_sensor_data);
```
vividf, line 179: Any reason why these used rmw_qos_profile_sensor_data but all other used rclcpp::SensorDataQoS()

sensing/autoware_calibration_status/lib/preprocess/preprocess_cuda.cu
```diff
@@ -0,0 +1,384 @@
...
+    output_array[3 * height * width + y * width + x] = 0.0f;
+    output_array[4 * height * width + y * width + x] = 0.0f;
+  }
+}
+
+cudaError_t PreprocessCuda::copyImage_launch(
+  const InputImageBGR8Type * input_image, const size_t width, const size_t height,
+  InputImageBGR8Type * output_image, float * output_array)
+{
+  dim3 threads(16, 16);
```
vividf: I think we could replace this to constant variable.

sensing/autoware_calibration_status/lib/preprocess/preprocess_cuda.cu
```diff
@@ -0,0 +1,384 @@
...
+    output_array[4 * height * width + y * width + x] = 0.0f;
+  }
+}
+
+cudaError_t PreprocessCuda::undistortImage_launch(
+  const InputImageBGR8Type * input_image, const double * dist_coeffs, const double * camera_matrix,
+  const double * projection_matrix, const size_t width, const size_t height,
+  InputImageBGR8Type * output_image, float * output_array)
+{
+  dim3 threads(16, 16);
```
vividf: Same :)

sensing/autoware_calibration_status/lib/preprocess/preprocess_cuda.cu
```diff
@@ -0,0 +1,384 @@
...
+ * @param v The input value, normalized between 0.0 and 1.0.
+ * @return An InputImageBGR8Type pixel representing the color.
+ */
+__device__ inline InputImageBGR8Type jet_colormap(float v)
+{
+  v = fminf(fmaxf(v, 0.0f), 1.0f);
+
+  unsigned char r = 0, g = 0, b = 0;
+
+  if (v < 0.33f) {
```
vividf: Could we replace this magic number with a named constant (e.g., constexpr) for better readability and understanding?

sensing/autoware_calibration_status/lib/preprocess/preprocess_cuda.cu
```diff
@@ -0,0 +1,384 @@
...
+    while (old_metric_depth == 0.0f || current_metric_depth < fabsf(old_metric_depth)) {
+      unsigned int assumed_int = __float_as_uint(old_metric_depth);
+      unsigned int returned_int =
+        atomicCAS(target_addr_int, assumed_int, __float_as_uint(current_metric_depth));
+
+      if (returned_int == assumed_int) {  // Race condition resolved
+        InputImageBGR8Type color = jet_colormap(static_cast<float>(p.intensity) / 255.0f);
+        size_t pixel_idx = v * width + u;
+
+        output_array[3 * height * width + pixel_idx] = current_metric_depth / max_depth;
```
vividf: There are lots of height * width operation, maybe we can reuse it to reduce the cost

sensing/autoware_calibration_status/lib/preprocess/preprocess_cuda.cu
```diff
@@ -0,0 +1,384 @@
...
+  const size_t y = blockIdx.y * blockDim.y + threadIdx.y;
+
+  if (x < width && y < height) {
+    const InputImageBGR8Type px = input_image[y * width + x];
+    output_image[y * width + x] = px;
+    output_array[0 * height * width + y * width + x] = static_cast<float>(px.r) / 255.0f;
+    output_array[1 * height * width + y * width + x] = static_cast<float>(px.g) / 255.0f;
+    output_array[2 * height * width + y * width + x] = static_cast<float>(px.b) / 255.0f;
+    output_array[3 * height * width + y * width + x] = 0.0f;
+    output_array[4 * height * width + y * width + x] = 0.0f;
```
vividf: We could define an enum for the channel indices to avoid using raw numbers like 3 and 4.
For example:

```cpp
enum Channel { channel_r = 0, channel_g = 1, channel_b = 2, channel_depth = 3, channel_intensity = 4 };
```

This would make the intent clearer when accessing output_array (e.g., output_array[channel_depth * stride + idx]) and improve readability.

sensing/autoware_calibration_status/lib/preprocess/preprocess_cuda.cu
```diff
@@ -0,0 +1,384 @@
...
+    const auto current_metric_depth = static_cast<float>(p_cam_z);
+
+    // Main Projection Z-Buffer (using the separate metric buffer)
+    auto target_addr_int = reinterpret_cast<uint32_t *>(&metric_depth_buffer[v * width + u]);
+    float old_metric_depth = __uint_as_float(*target_addr_int);
+
+    while (old_metric_depth == 0.0f || current_metric_depth < fabsf(old_metric_depth)) {
+      unsigned int assumed_int = __float_as_uint(old_metric_depth);
+      unsigned int returned_int =
+        atomicCAS(target_addr_int, assumed_int, __float_as_uint(current_metric_depth));
```
vividf: We could replace this custom atomicCAS loop with atomicMin(float*) for cleaner semantics and potentially better performance.

This would also require initializing the depth buffer to FLT_MAX and adjusting how dilation is handled (since it currently uses negative depths). Leaving it to you to decide if this refactor is worth doing here！

sensing/autoware_calibration_status/lib/preprocess/preprocess_cuda.cu
```diff
@@ -0,0 +1,384 @@
...
+ * @param num_points The total number of points in the input cloud.
+ * @param width The width of the output image and data array.
+ * @param height The height of the output image and data array.
+ * @param max_depth The maximum range to use for depth normalization.
+ * @param metric_depth_buffer A temporary buffer of size (width*height) initialized to 0.0f, used
+ * for Z-buffering.
+ * @param output_array The final output 2D grid where uint8 data is stored.
+ * @param num_points_projected A counter for the number of successfully projected points.
+ */
+__global__ void projectPoints_kernel(
```
vividf: This kernel is doing several things at once.
To improve readability, consider splitting it into small device helpers

sensing/autoware_calibration_status/lib/calibration_status.cpp
```diff
@@ -0,0 +1,179 @@
...
+  projection_matrix_d_ = cuda_utils::make_unique<double[]>(projection_matrix_size);
+  tf_matrix_d_ = cuda_utils::make_unique<double[]>(tf_matrix_size);
+  num_points_projected_d_ = cuda_utils::make_unique<uint32_t>();
+
+  CHECK_CUDA_ERROR(cudaStreamCreate(&stream_));
+  preprocess_ptr_ = std::make_unique<PreprocessCuda>(
+    config.max_depth, config.dilation_size, config_.width.at(2), config_.height.at(2), stream_);
+}
+
+CalibrationStatusResult CalibrationStatus::process(
```
vividf: Please seperate them into smaller functions like preprocess_image, project_points, infer.

sensing/autoware_calibration_status/src/calibration_status_node.cpp
```diff
@@ -0,0 +1,478 @@
...
+  } else if (is_calibrated) {
+    status.level = diagnostic_msgs::msg::DiagnosticStatus::OK;
+    status.message = "Calibration is valid";
+  } else {
+    status.level = diagnostic_msgs::msg::DiagnosticStatus::ERROR;
+    status.message = "Calibration is invalid.";
+  }
+
+  diagnostic_msgs::msg::KeyValue key_value;
+  key_value.key = "Source image topic";
```
vividf: It would be nice if we could use a helper function to make it more clear
```cpp
auto add_kv = [&](const std::string& key, const std::string& value) {
  diagnostic_msgs::msg::KeyValue kv;
  kv.key = key;
  kv.value = value;
  status.values.push_back(kv);
};
```

amadeuszsz: Thanks, good idea!
Addressed in https://github.com/autowarefoundation/autoware_universe/pull/11222/commits/f21247977997acf74a63e6d7b2956335a6cd93e5

sensing/autoware_calibration_status/lib/calibration_status.cpp
amadeuszsz: Leaving as is.

We usually want to have a separated library and make a single function call from ROS node level. Please check centerpoint, transfusion or bevfusion for reference.

sensing/autoware_calibration_status/lib/preprocess/preprocess_cuda.cu
amadeuszsz: Addressed in https://github.com/autowarefoundation/autoware_universe/pull/11222/commits/4eeb0e1952c7b454d1570ed9070e964c7adba21f

sensing/autoware_calibration_status_classifier/src/calibration_status_classifier_node.cpp
amadeuszsz: Because message filters has an argument of `rmw_qos_profile_t` type instead of `rclcpp::QoS`.

sensing/autoware_calibration_status/lib/preprocess/preprocess_cuda.cu
amadeuszsz: Leaving as is (please check about common subexpression elimination).

sensing/autoware_calibration_status/lib/preprocess/preprocess_cuda.cu
amadeuszsz: Part of https://github.com/autowarefoundation/autoware_universe/pull/11222/commits/c2a6cfe14a65f92116fdbae4c1d708c9c3516ceb

sensing/autoware_calibration_status/lib/preprocess/preprocess_cuda.cu
amadeuszsz: Part of https://github.com/autowarefoundation/autoware_universe/pull/11222/commits/c2a6cfe14a65f92116fdbae4c1d708c9c3516ceb

sensing/autoware_calibration_status/lib/preprocess/preprocess_cuda.cu
amadeuszsz: Part of https://github.com/autowarefoundation/autoware_universe/pull/11222/commits/c2a6cfe14a65f92116fdbae4c1d708c9c3516ceb

sensing/autoware_calibration_status/lib/preprocess/preprocess_cuda.cu
amadeuszsz: Part of https://github.com/autowarefoundation/autoware_universe/pull/11222/commits/c2a6cfe14a65f92116fdbae4c1d708c9c3516ceb

sensing/autoware_calibration_status/lib/preprocess/preprocess_cuda.cu
amadeuszsz: Part of https://github.com/autowarefoundation/autoware_universe/pull/11222/commits/c2a6cfe14a65f92116fdbae4c1d708c9c3516ceb

No performance improvement, maybe negligibly slower but for sure we improved code readability 👌🏻 
Also I decided to split one kernel for readability purpose.

sensing/autoware_calibration_status/src/calibration_status_node.cpp
amadeuszsz: Thanks, I forgot about this class! Addressed in https://github.com/autowarefoundation/autoware_universe/pull/11222/commits/f354bc72f1b18c431b9a1c3f9bbb0a55796ab518

`autoware_utils::DiagnosticsInterface` implementation does not allow for custom message if status is OK... I can accept it 😢 
```c++
void DiagnosticsInterface::update_level_and_message(const int8_t level, const std::string & message)
{
  if ((level > diagnostic_msgs::msg::DiagnosticStatus::OK)) {
    if (!diagnostics_status_msg_.message.empty()) {
      diagnostics_status_msg_.message += "; ";
    }
    diagnostics_status_msg_.message += message;
  }
  if (level > diagnostics_status_msg_.level) {
    diagnostics_status_msg_.level = level;
  }
}
```

sensing/autoware_calibration_status_classifier/src/calibration_status_classifier_node.cpp
```diff
@@ -0,0 +1,454 @@
...
+}
+
+bool CalibrationStatusClassifierNode::run(std::size_t pair_idx)
+{
+  if (
+    synchronized_data_.at(pair_idx).first == nullptr ||
+    synchronized_data_.at(pair_idx).second == nullptr) {
+    return false;
+  }
+  const auto & [cloud_msg, image_msg] = synchronized_data_.at(pair_idx);
```
manato:
```suggestion
  const auto  [cloud_msg, image_msg] = synchronized_data_.at(pair_idx);
```
Since `synchronized_data_` contains shared_ptr, I would suggest using copy constructors rather than referencing to increase the reference count of the shared_ptr explicitly. My concern is `cloud_msg` and/or `image_msg` become invalid during processing due to the new assignment to the `synchronized_data_.at(pair_idx)` in `synchronized_callback()`.

I guess this kind of race condition may happen when the node runs with `RuntimeMode::MANUAL` or `RuntimeMode::PERIODIC`, which `synchronized_callback` (data insertion) and `run` (data consumption) are called asynchronously.

sensing/autoware_calibration_status_classifier/src/calibration_status_classifier_node.cpp
```diff
@@ -0,0 +1,454 @@
...
+  auto result = calibration_status_classifier_->process(
+    cloud_msg, image_msg, camera_lidar_info_.at(pair_idx), preview_img_data);
+
+  publish_diagnostic_status(input_metadata, pair_idx, result);
+
+  if (is_preview_subscribed) {
+    preview_image_pubs_.at(pair_idx)->publish(*preview_img_msg);
+  }
+
+  synchronized_data_.at(pair_idx) = {nullptr, nullptr};
```
manato: Since read and write operations possibly occur asynchronously when the node runs with `RuntimeMode::MANUAL` or `RuntimeMode::PERIODIC`, I would suggest using mutex for each element of `synchronized_data_` to prevent race condition.

sensing/autoware_calibration_status_classifier/src/calibration_status_classifier_node.cpp
amadeuszsz: Thank you for this additional check! Since callbacks are in the same mutually exclusive group by default, I believe none of them will be processed at the same time. Could you please confirm if in this circumstances race condition can occur? To make it safe for sure, I addressed potential issue in https://github.com/autowarefoundation/autoware_universe/pull/11222/commits/2ede2989bb2cce1738c330940d2aff1da87a535a a bit differently. Please, take a look on it 🙏🏻

sensing/autoware_calibration_status_classifier/src/calibration_status_classifier_node.cpp
amadeuszsz: Same thoughts as in comment before. For clearance addressed in https://github.com/autowarefoundation/autoware_universe/pull/11222/commits/24c180a4d278fc05de541060aa8c9a03fc23ee46

PR #11224, opened 2025-08-15
fix(goal_planner): generate goal candidates using lane boundary to fix wrong validation

kosuke55:
## Description

In the existing method, goal candidates were generated based on the centerline, and the orientation was aligned with the boundary. However, there were inconsistencies in the margin and within-area checks.

image

image

This PR changes the logic to regenerate goal poses based on the Lanelet boundary and fix https://github.com/autowarefoundation/autoware_universe/pull/11086 issue.

Ensure that the position closest to the footprint's boundary maintains a sufficient margin.
The offset distance from the boundary is determined by the following four patterns:

- Case 1-1: Left boundary is on the right side of rear pose. Rear Pose is closer to left boundary than Front Pose.
- Case 1-2: Left boundary is on the left side of rear pose.  Front pose is closer to left boundary than Rear Pose.
- Case 2-1: Right boundary is on the right side of rear pose. Front pose is closer to right boundary than Rear Pose.
- Case 2-2: Right boundary is on the left side of rear pose. Rear Pose is closer to right boundary than Front Pose.

image
image

## Related links

**Parent Issue:**

- depends https://github.com/autowarefoundation/autoware_lanelet2_extension/pull/77

## How was this PR tested?

psim

left_side

bus stop area

image

image

image

right_side

image

image

Evaluator
- [[Daily][Lexus][ControlDLRCommon_Basic] DLR test](https://evaluation.tier4.jp/evaluation/reports/fb3803e0-6230-5c03-952a-c598b4d4656f?project_id=prd_jt) 
- [[PR check][Lexus][ControlDLRCommon_Basic] DLR test](https://evaluation.tier4.jp/evaluation/reports/cede9220-d177-5743-8d94-b0bec12b8a51?project_id=prd_jt)
- [[PR check][Lexus][CommonScenario_Basic] planning test](https://evaluation.tier4.jp/evaluation/reports/b157f322-3a2d-5680-affb-a13a68093087?project_id=prd_jt)
  - 1 test failed, but not relating goal planner

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_path_planner/autoware_behavior_path_goal_planner_module/src/util.cpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_goal_planner_module/src/util.cpp
@@ -613,8 +613,9 @@ double calcLateralDeviationBetweenPaths(
       reference_path.points, target_point.point.pose.position);
     lateral_deviation = std::max(
       lateral_deviation,
-      std::abs(autoware_utils::calc_lateral_deviation(
-        reference_path.points[nearest_index].point.pose, target_point.point.pose.position)));
+      std::abs(
+        autoware_utils::calc_lateral_deviation(
+          reference_path.points[nearest_index].point.pose, target_point.point.pose.position)));
   }
   return lateral_deviation;
 }
@@ -1072,8 +1073,9 @@ bool hasPreviousModulePathShapeChanged(
       // p.point.pose.position is not within the segment, skip lateral distance check
       continue;
     }
-    const double lateral_distance = std::abs(autoware::motion_utils::calcLateralOffset(
-      last_upstream_module_output.path.points, p.point.pose.position, nearest_seg_idx));
+    const double lateral_distance = std::abs(
+      autoware::motion_utils::calcLateralOffset(
+        last_upstream_module_output.path.points, p.point.pose.position, nearest_seg_idx));
     if (lateral_distance > LATERAL_DEVIATION_THRESH) {
       return true;
     }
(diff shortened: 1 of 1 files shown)


planning/behavior_path_planner/autoware_behavior_path_goal_planner_module/src/goal_searcher.cpp
```diff
@@ -31,16 +34,167 @@
...
+    // NOTE:
+    // align_orientation_with_trajectory_direction only calculates the azimuth for the input to
+    // build, and the interpolated points have their orientation interpolated as well.
+    // Therefore, we recalculate the azimuth here.
+    pose.orientation =
+      autoware_utils_geometry::create_quaternion_from_yaw(pose_trajectory->azimuth(s));
+    cropped_poses.push_back(pose);
+  }
+
+  for (double s = start_s; s < end_s; s += interval) {
```
soblin: these points already added in above block, aren't they ?

kosuke55: ah thanks, this is mistake. fixed

kosuke55:
@soblin 
To accurately calculate the signed distance to the boundary in the y-axis direction of the pose, I implemented calcSignedLateralDistanceToBoundary instead of calcClosestPose. The old implementation calculated the signed y-distance to neighboring point.
Take a quick look again.

fb33cbe

unit test case. (blue x, green y)
image

PR #11236, opened 2025-08-19
feat(start_planner): add custom priority configuration for planner selection

kyoichi-sugahara:
## Description

**Parameter change [PR](https://github.com/autowarefoundation/autoware_launch/pull/1613) should be merged together**

This PR adds a new `custom` search priority mode to the start planner module that allows users to specify a custom priority order for different planner types. This enhancement provides flexibility in controlling which planners are tried first when searching for pull-out paths.

**New Features:**

- **Custom planner priority**: Users can now specify exactly which planners to use and in what order
- **Selective planner exclusion**: Ability to exclude specific planners (e.g., exclude CLOTHOID for safety-critical applications)

**Behavior Changes:**

- Proper fallback to `efficient_path` mode if custom configuration fails or is incomplete
- Only enabled planners are used, even if listed in the priority list

**Use Cases:**

- Conservative approach: `["SHIFT", "GEOMETRIC"]` - excludes CLOTHOID planner
- Testing scenarios: `["GEOMETRIC"]` - tests only the geometric planner
- Custom optimization: Any combination based on specific requirements

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

## Notes for reviewers

The implementation adds comprehensive documentation and examples in the README, along with proper error handling and fallback mechanisms when custom priority configuration fails.

## Interface changes

None.

### ROS Parameter Changes

#### Additions and removals

| Change type | Parameter Name          | Type           | Default Value | Description                                                                                                |
| :---------- | :---------------------- | :------------- | :------------ | :--------------------------------------------------------------------------------------------------------- |
| Added       | `planner_priority_list` | `string array` | `[]`          | List of planner types in priority order for custom mode. Available types: "SHIFT", "GEOMETRIC", "CLOTHOID" |

#### Modifications

| Version | Parameter Name    | Type     | Default Value    | Description                                                                                                                                                                                                                                                          |
| :------ | :---------------- | :------- | :--------------- | :------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| Old     | `search_priority` | `string` | `efficient_path` | priority strategy for planner selection. "efficient_path" or "short_back_distance"                                                                                                                                                                                   |
| New     | `search_priority` | `string` | `efficient_path` | priority strategy for planner selection. `efficient_path`: prioritize efficient paths regardless of back distance. `short_back_distance`: prioritize paths with minimal back distance. `custom`: specify custom planner priority order using `planner_priority_list` |

## Effects on system behavior


planning/behavior_path_planner/autoware_behavior_path_start_planner_module/src/start_planner_module.cpp
```diff
@@ -1143,7 +1143,25 @@ PriorityOrder StartPlannerModule::determinePriorityOrder(
...
+  if (search_priority == "custom") {
+    if (parameters_->planner_priority_list.empty()) {
+      RCLCPP_ERROR(
+        getLogger(),
+        "Custom priority mode requested but planner_priority_list is empty. Falling back to "
+        "efficient_path");
+    } else {
+      try {
+        return generatePriorityOrderFromList(
+          parameters_->planner_priority_list, start_pose_candidates_num);
```
danielsanchezaran: Why not just handle the parameters_->planner_priority_list.empty() == true case inside the generatePriorityOrderFromList() function? just add a check at the beggining of the function and throw? that way you can get rid of the overly complex if-else and just do a try-catch with `return generatePriorityOrderFromList(
          parameters_->planner_priority_list, start_pose_candidates_num);`

planning/behavior_path_planner/autoware_behavior_path_start_planner_module/src/start_planner_module.cpp
```diff
@@ -1143,7 +1143,25 @@ PriorityOrder StartPlannerModule::determinePriorityOrder(
...
+      } catch (const std::runtime_error & e) {
+        RCLCPP_ERROR(
+          getLogger(), "Custom priority failed: %s. Falling back to efficient_path", e.what());
+      }
+    }
+  }
+
+  if (search_priority == "efficient_path" || search_priority == "custom") {
```
danielsanchezaran: While not strictly part of this PR, would you consider refactoring this to use lambdas? this part of the code is very confusing for first timers. I recommend something like: 
```
// Define the different ordering strategies as lambdas
auto order_by_planner_first = [&]() {
  for (const auto &planner : start_planners_) {
    for (size_t i = 0; i < start_pose_candidates_num; i++) {
      order_priority.emplace_back(i, planner);
    }
  }
};

auto order_by_index_first = [&]() {
  for (size_t i = 0; i < start_pose_candidates_num; i++) {
    for (const auto &planner : start_planners_) {
      order_priority.emplace_back(i, planner);
    }
  }
};

if (search_priority == "efficient_path" || search_priority == "custom") {
  order_by_planner_first();
} else if (search_priority == "short_back_distance") {
  order_by_index_first();
}
```
You can go further and do an early return for the case were search_priority  is not equal to any recognized option, instead of having that fallback at the end as an else clause

planning/behavior_path_planner/autoware_behavior_path_start_planner_module/src/start_planner_module.cpp
kyoichi-sugahara: thanks! I will address the complex problem later. for now I convert this pr to draft bacause I feel that this change is not somewhat good to begin with.

kyoichi-sugahara:
opened new PR in https://github.com/autowarefoundation/autoware_universe/pull/11275
@danielsanchezaran 
If you have comments for new PR, please feel free to share your opinion.

PR #11244, opened 2025-08-21
fix(autoware_multi_object_tracker): update isConfident method signature and logic for time handling

technolojin:
## Description

isConfident is a method to determine whether the tracker is confident enough to be used for the following purpose.
- merge trackers
- publish

When extrapolation time is long (only when `enable_delay_compensation` is true), the position covariance becomes larger.
To make a fair and stable outcome, the function is enabled to choose given time, depends on its purpose.
- specific time (for merge trackers)
- last tracker time + constant interval (for publish)

The constant interval is used to consider motion (velocity) uncertainty into the position uncertainty.

## Related links

Related with PR https://github.com/autowarefoundation/autoware_universe/pull/11230

## How was this PR tested?

This test is over https://github.com/autowarefoundation/autoware_universe/pull/11230

Before

https://github.com/user-attachments/assets/4d7fca0d-c890-472c-a2d5-9bc5d6fdeadb

After

https://github.com/user-attachments/assets/697be767-593c-4595-b628-e31cc14d4ca6

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


linick-kang (approved):
LGTM

PR #11257, opened 2025-08-22
feat(stop_mode_operator): shift gear to park on stop mode

isamu-takagi:
## Description

Shift gear to park on stop mode. This is to be consistent with shift_decider behavior. 

## Related links

**Private Links:**

- [TIER IV internal link](https://tier4.atlassian.net/browse/RT0-38870)

## How was this PR tested?

1. Drive the vehicle in autonomous mode using planning simulator.
1. Check that the gear is parking when the vehicle goals.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/control/autoware_stop_mode_operator/src/continuous_condition.cpp
+++ b/control/autoware_stop_mode_operator/src/continuous_condition.cpp
@@ -1,0 +1,38 @@
+//  Copyright 2025 The Autoware Contributors
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+#include "continuous_condition.hpp"
+
+namespace autoware::stop_mode_operator
+{
+
+void ContinuousCondition::update(const rclcpp::Time & stamp, bool condition)
+{
+  if (!condition) {
+    start_stamp_ = std::nullopt;
+  }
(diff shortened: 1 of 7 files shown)


mkuri (approved):

control/autoware_stop_mode_operator/src/continuous_condition.cpp
```diff
@@ -0,0 +1,38 @@
...
+
+void ContinuousCondition::update(const rclcpp::Time & stamp, bool condition)
+{
+  if (!condition) {
+    start_stamp_ = std::nullopt;
+  }
+  if (!start_stamp_) {
+    start_stamp_ = stamp;
+  }
+}
```
yuki-takagi-66: Even if the condition is false, start_stamp_ will have some value. Is it desired behavior?

isamu-takagi: If `condition` is false, `start_stamp_` is assigned nullopt. Are you concerned about case where this function is not called? I will fix that.

yuki-takagi-66: Yes, If condition arg is false, `start_stamp_` is assigned nullopt by line 23.
However, the condition in line 25 is always satisfied if the line 23 is executed.
So, Isn't `start_stamp_` being updated?

I feel that one of the following should be added
- add else to line 25
- add return to line 23

isamu-takagi: Oh, that's right. I forgot else. Thank you for nice checking.

isamu-takagi: Fix: https://github.com/autowarefoundation/autoware_universe/pull/11257/commits/4687bdeef692f0337d5180d84c80c6b32f416c19
And add timeout check: https://github.com/autowarefoundation/autoware_universe/pull/11257/commits/bfdc9bfba76b25d7ea0a37717624c46d0ee1a62d, https://github.com/autowarefoundation/autoware_universe/pull/11257/commits/cade9137ff31edab424af127d3a24068b988318c

PR #11258, opened 2025-08-22
fix(fusion node): subscribe from concatenation info

vividf:
## Description
This PR eliminates the subscription to `/diagnostic`, which was increasing CPU usage in advanced mode. The node now subscribes directly to `concatenation_info` for improved efficiency.

Note that the algorithm itself has not changed. This PR only moves the value into the defined config field within concatenation_info, which the fusion node then subscribes to directly.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?
* Test with sample rosbag.
* Test with a large vehicle rosbag. 

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_image_projection_based_fusion/config/fusion_common.param.yaml
+++ b/perception/autoware_image_projection_based_fusion/config/fusion_common.param.yaml
@@ -30,7 +30,6 @@
 
     publish_previous_but_late_output_msg: false
     rosbag_length: 10.0
-
     # matching strategy
     matching_strategy:
       type: advanced
--- a/perception/autoware_image_projection_based_fusion/config/fusion_common.param.yaml
+++ b/perception/autoware_image_projection_based_fusion/config/fusion_common.param.yaml
@@ -31,12 +31,10 @@
     publish_previous_but_late_output_msg: false
     rosbag_length: 10.0
 
-    # concatenation info topic for advanced matching strategy
-    concatenation_info_topic: "/sensing/lidar/concatenated/pointcloud_info"
-
     # matching strategy
     matching_strategy:
       type: advanced
+      concatenation_info_topic: "/sensing/lidar/concatenated/pointcloud_info"
       msg3d_noise_window: 0.02
       rois_timestamp_noise_window: [0.01, 0.01, 0.01, 0.01, 0.01, 0.01]
       # type: naive
--- a/perception/autoware_image_projection_based_fusion/config/fusion_common.param.yaml
+++ b/perception/autoware_image_projection_based_fusion/config/fusion_common.param.yaml
@@ -33,7 +33,6 @@
     # matching strategy
     matching_strategy:
       type: advanced
-      concatenation_info_topic: "/sensing/lidar/concatenated/pointcloud_info"
       msg3d_noise_window: 0.02
       rois_timestamp_noise_window: [0.01, 0.01, 0.01, 0.01, 0.01, 0.01]
       # type: naive
--- a/perception/autoware_image_projection_based_fusion/config/fusion_common.param.yaml
+++ b/perception/autoware_image_projection_based_fusion/config/fusion_common.param.yaml
@@ -30,6 +30,10 @@
 
     publish_previous_but_late_output_msg: false
     rosbag_length: 10.0
+
+    # concatenation info topic for advanced matching strategy
+    concatenation_info_topic: "/sensing/lidar/concatenated/pointcloud_info"
+
     # matching strategy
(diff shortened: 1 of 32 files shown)


perception/autoware_image_projection_based_fusion/config/fusion_common.param.yaml
```diff
@@ -33,6 +33,7 @@
     # matching strategy
     matching_strategy:
       type: advanced
+      concatenation_info_topic: "/sensing/lidar/concatenated/pointcloud_info"
```
technolojin: the topic shall not be in config file, but in launchers
```xml
<launch>
  <arg name="input/concatenation_info" default="/camera_concatenation_info"/>
...
  <group>
    <node pkg="autoware_image_projection_based_fusion" exec="roi_cluster_fusion_node" name="roi_cluster_fusion" output="screen">
      <remap from="input/concatenation_info" to="$(var input/concatenation_info)"/>
...
```

technolojin (changes requested):
1. I observed the following warning messages (not because of this PR, but after advanced strategy is implied)

```
[roi_cluster_fusion_node-61] [WARN] [1756096746.490698077] [perception.object_recognition.detection.clustering.camera_lidar_fusion.roi_cluster_fusion]: Reset the oldest collector because the number of processing collectors (10) equal to the limit (10).
[roi_cluster_fusion_node-61] [WARN] [1756096756.505832196] [perception.object_recognition.detection.clustering.camera_lidar_fusion.roi_cluster_fusion]: Reset the oldest collector because the number of processing collectors (10) equal to the limit (10).
```

~~Can you also see the message and adjust hard-coded parameter for this?~~

-> my launcher setting was bad


2. I suggest to set the concat info topic move to launcher. If necessary, please update the autoware_launch repo and share the PR link on this PR.

perception/autoware_image_projection_based_fusion/config/fusion_common.param.yaml
vividf: Thanks for the suggestion!
Addressed in https://github.com/autowarefoundation/autoware_universe/pull/11258/commits/5b931211728a7f087fae0ca766186537bc72f44e

technolojin (approved):
LGTM

perception/autoware_image_projection_based_fusion/config/fusion_common.param.yaml
vividf: @technolojin 
Thanks, would you mind checking https://github.com/autowarefoundation/autoware_universe/pull/11258/commits/439d9897c18198d3796b11648b6f2902f699b09d and https://github.com/autowarefoundation/autoware_universe/pull/11258/commits/5b931211728a7f087fae0ca766186537bc72f44e

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_concatenate_data/cuda_combine_cloud_handler.cpp
```diff
@@ -281,7 +282,28 @@ CombineCloudHandler<CudaPointCloud2Traits>::combine_pointclouds(
...
+      static_cast<uint32_t>((reference_timestamp_max - reference_timestamp_max_msg.sec) * 1e9);
+
+    StrategyAdvancedConfig strategy_config(
+      reference_timestamp_min_msg, reference_timestamp_max_msg);
+    auto serialized_config = strategy_config.serialize();
+    ConcatenationInfoManager::set_config(
+      serialized_config, *concatenate_cloud_result.concatenation_info_ptr);
+  }
+
+  concatenation_info_manager_.set_result(
```
mojomex, line 306: (nit) Duplicate from pointcloud_preprocessor

mojomex (approved):
LGTM

badai-nguyen (changes requested):
@vividf 
I think `input/concatenation_info` subscription adding is also needed for irregular object launch, Could you confim reconfirm that? 
https://github.com/autowarefoundation/autoware_universe/blob/dafe7af359572ec69d392a00a019031a43d2cf45/launch/tier4_perception_launch/launch/object_recognition/detection/detector/camera_lidar_irregular_object_detector.launch.py#L30

vividf:
@badai-nguyen
Thanks, I added in https://github.com/autowarefoundation/autoware_universe/pull/11258/commits/2b5328f7103667726ab8145c888326cf7589841a

badai-nguyen (approved):

PR #11265, opened 2025-08-25
fix(blind_spot): filter non-vru based on lateral clearance

zulfaqar-azmi-t4:
## Description

The blind spot module currently slows down the vehicle whenever an object overlaps with the attention area.
This can result in false positives when an object’s width is larger than the available lateral clearance.

VRUs (pedestrians, bicycles, motorcycles): Their size is generally small, and they have more flexibility to squeeze through narrow gaps. Prediction noise may also cause their detected size to fluctuate, but this is acceptable. The current logic continues to handle VRUs as before.

Non-VRUs (e.g., cars): Their maneuverability is limited and they cannot realistically pass through gaps smaller than their width, which leads to unnecessary stop in the current approach.

image

### Fix

This PR updates the blind spot check only for non-VRUs by comparing the lateral clearance with the object’s width.

If the clearance is smaller than the object’s width, the object is assumed unable to pass through and is ignored for safety decisions.

VRU handling remains unchanged.

This reduces unnecessary slowdowns for non-VRUs, while keeping conservative behavior for VRUs.

## Related links

[TIER IV Internal Link - Issue's ticket](https://tier4.atlassian.net/browse/RT0-39023)
[TIER IV Internal Link - Issue's ticket](https://tier4.atlassian.net/browse/RT0-39024)

### Test 1: Perception reproducers

Testing when the gap is smaller than the width of non-vru targets

https://github.com/user-attachments/assets/297b3e10-56e4-446e-8ad4-824e42338be2

When there's sufficient gap for non-vru targets to pass through

https://github.com/user-attachments/assets/4172f316-2d0f-418d-ad9e-3ef5ca743358

2. [Evaluator]

[TIER IV Internal Link - ControlDLR Test](https://evaluation.tier4.jp/evaluation/reports/400cdad4-c9f7-508d-8bd4-31ee73a9579b?project_id=prd_jt)
[TIER IV Internal Link - FuncVerification Test](https://evaluation.tier4.jp/evaluation/reports/52bbf013-e911-5cf9-b3b8-4bfe2afd2318?project_id=prd_jt)
[TIER IV Internal Link - CommonScenario_Basic Test](https://evaluation.tier4.jp/evaluation/reports/b73a1489-19b5-5d35-8a96-e095bea30964?project_id=prd_jt)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/include/autoware/behavior_velocity_blind_spot_module/scene.hpp
+++ b/planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/include/autoware/behavior_velocity_blind_spot_module/scene.hpp
@@ -144,7 +144,7 @@ private:
    * @brief filter objects whose position is inside the attention_area and whose type is target type
    */
   std::vector<autoware_perception_msgs::msg::PredictedObject> filter_attention_objects(
-    const lanelet::BasicPolygon2d & attention_area) const;
+    const lanelet::BasicPolygon2d & attention_area, const double lateral_gap) const;
 
   /**
    * @brief Check if object is belong to targeted classes
@@ -152,6 +152,8 @@ private:
    * @return True when object belong to targeted classes
    */
   bool isTargetObjectType(const autoware_perception_msgs::msg::PredictedObject & object) const;
+
+  static bool isVRUObjectType(const autoware_perception_msgs::msg::PredictedObject & object);
 
   /**
    * @brief compute the deceleration and jerk for collision stop from `ttc`
--- a/planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/include/autoware/behavior_velocity_blind_spot_module/scene.hpp
+++ b/planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/include/autoware/behavior_velocity_blind_spot_module/scene.hpp
@@ -153,7 +153,7 @@ private:
    */
   bool isTargetObjectType(const autoware_perception_msgs::msg::PredictedObject & object) const;
 
-  static bool isVRUObjectType(const autoware_perception_msgs::msg::PredictedObject & object);
+  static bool is_vru_object_type(const autoware_perception_msgs::msg::PredictedObject & object);
 
   /**
    * @brief compute the deceleration and jerk for collision stop from `ttc`
--- a/planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/src/util.cpp
+++ b/planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/src/util.cpp
@@ -685,7 +685,7 @@ std::optional<StopPoints> generate_stop_points(
     std::nullopt, stop_points_list.instant_stopline, stop_points_list.critical_stopline};
 }
 
-tl::expected<double, std::string> calc_ego_to_blind_spot_lanelet_lateral_gap(
+std::optional<double> calc_ego_to_blind_spot_lanelet_lateral_gap(
   const autoware_utils::LinearRing2d & ego_footprint,
   const lanelet::ConstLanelets & last_lanelets_before_turning,
   const autoware::experimental::lanelet2_utils::TurnDirection & turn_direction)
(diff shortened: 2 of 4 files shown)


planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/include/autoware/behavior_velocity_blind_spot_module/scene.hpp
```diff
@@ -153,6 +153,8 @@ class BlindSpotModule : public SceneModuleInterfaceWithRTC
    */
   bool isTargetObjectType(const autoware_perception_msgs::msg::PredictedObject & object) const;
 
+  static bool isVRUObjectType(const autoware_perception_msgs::msg::PredictedObject & object);
```
soblin:
```suggestion
  static bool is_vru_object_type(const autoware_perception_msgs::msg::PredictedObject & object);
```

planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/src/util.cpp
```diff
@@ -681,4 +684,57 @@ std::optional<StopPoints> generate_stop_points(
...
+  const autoware_utils::LinearRing2d & ego_footprint,
+  const lanelet::ConstLanelets & last_lanelets_before_turning,
+  const autoware::experimental::lanelet2_utils::TurnDirection & turn_direction)
+{
+  if (ego_footprint.size() != 7) {
+    return tl::make_unexpected("Unexpected ego_footprint's size.");
+  }
+
+  // left front is 6, left back is 4, right_front is 1, right_back is 3
+  const auto front_idx = (turn_direction == TurnDirection::Left) ? 6 : 1;
```
soblin: please use these constants

https://github.com/autowarefoundation/autoware_core/blob/d01219ddfd26d38ad865b01d150cfc98579e1047/common/autoware_vehicle_info_utils/include/autoware/vehicle_info_utils/vehicle_info.hpp#L50-L53

planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/src/util.cpp
```diff
@@ -681,4 +684,57 @@ std::optional<StopPoints> generate_stop_points(
...
+    }
+  }
+
+  if (line.size() < 2) {
+    return tl::make_unexpected("Not enough points in lanelet boundary.");
+  }
+
+  std::vector<autoware_utils::Segment2d> segments;
+  segments.reserve(line.size() - 1);
+  for (size_t i = 0; i + 1 < line.size(); ++i) {
```
soblin:
```suggestion
  for ([point1, point2) = ranges::view::zip(line, line | ranges::views::drop(1)) {
```

planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/src/util.cpp
```diff
@@ -681,4 +684,57 @@ std::optional<StopPoints> generate_stop_points(
     std::nullopt, stop_points_list.instant_stopline, stop_points_list.critical_stopline};
 }
 
+tl::expected<double, std::string> calc_ego_to_blind_spot_lanelet_lateral_gap(
```
soblin: failure reason is almost obvious, so optional is enough for me

planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/src/util.cpp
zulfaqar-azmi-t4: done in https://github.com/autowarefoundation/autoware_universe/pull/11265/commits/a4023d8cad8bffc9bdc68c0f6120cbf176a0d3a7

planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/src/util.cpp
zulfaqar-azmi-t4: done in https://github.com/autowarefoundation/autoware_universe/pull/11265/commits/894fcf90ae086af6c888b1b597b064023a7a70b1

planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/src/util.cpp
zulfaqar-azmi-t4: done in https://github.com/autowarefoundation/autoware_universe/pull/11265/commits/894fcf90ae086af6c888b1b597b064023a7a70b1

PR #11266, opened 2025-08-25
fix[autoware_trajectory_follower_node]: change plot label of data[25] from calculated to feedback

XiaoyuWang0601:
## Description
Change the label name of `longitudinal/diagnostic/data[25]` in the Longitudinal Controller Info section.
Because this will mislead the user into thinking that this is an acc command.
`longitudinal/diagnostic/data[25]` is actually the acc feedback (measured acc).

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


xmfcx:
@TakaHoribe @takayuki5168 could you check this very simple cosmetic PR if it's ok to merge?

PR #11272, opened 2025-08-26
feat(ground segmentation): add cuda based ground_segmentation

badai-nguyen:
## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


perception/autoware_ground_segmentation_cuda/src/cuda_scan_ground_segmentation/cuda_scan_ground_segmentation_filter.cu
```diff
@@ -0,0 +1,1330 @@
...
+    last_gnd_cells_dev[num_latest_gnd_cells++] = cell_index_in_sector;
+  }
+  if (cell_index_in_sector == 0) {
+    return;  // Base case: no more cells to check
+  }
+  // Continue searching in the previous cell
+  RecursiveGndCellSearch(
+    sector_cells_list_dev, last_gnd_cells_num_threshold, cell_index_in_sector - 1,
+    last_gnd_cells_dev, num_latest_gnd_cells);
+}
```
anhnv3991, line 312: Recursive device functions are expensive, since parameters are passed through the local memory. The recursive functions should be avoided.

perception/autoware_ground_segmentation_cuda/src/cuda_scan_ground_segmentation/cuda_scan_ground_segmentation_filter.cu
```diff
@@ -0,0 +1,1330 @@
...
+ * sector's cells (device memory).
+ * @param last_gnd_cells_indices_dev Pointer to the array of indices indicating which cells are used
+ * for fitting (device memory).
+ * @param num_fitting_cells Number of cells to use for fitting.
+ * @param filter_parameters_dev Pointer to the FilterParameters structure containing fitting
+ * thresholds (device memory).
+ * @return The slope (a) of the fitted line. If fitting is not possible, returns 0.0f.
+ */
+
+__device__ float fitLineFromGndCell(
```
anhnv3991, line 331: Each time we move to a new cell, we have to re-calculate the slope of the previous ground cells from scratch. I think there may be some other way to update the slope value. That should be faster than recalculation.

perception/autoware_ground_segmentation_cuda/src/cuda_scan_ground_segmentation/cuda_scan_ground_segmentation_filter.cu
```diff
@@ -0,0 +1,1330 @@
...
+  if (idx >= filter_parameters_dev->num_sectors) {
+    return;  // Out of bounds
+  }
+  // For each sector, find the ground reference points if points exist
+  // otherwise, use the previous sector ground reference points
+  // initialize the previous cell centroid
+
+  // Process the first cell of the sector
+  SegmentationMode mode = SegmentationMode::UNINITIALIZED;
+  auto sector_start_cell_index = idx * filter_parameters_dev->max_num_cells_per_sector;
```
anhnv3991, line 777: As I understand, the cells in the same sectors are put in the continuous addresses. Something like this
cell 0.0, cell 0.1, cell 0.2; cell 1.0, cell 1.1, cell 1.2; cell 2.0, cell 2.1, cell 2.2... (cell i.j is the cell j at sector i)
When the kernel is launched, each thread accesses cells in ONE sector, i.e. thread 0 accesses cell 0.0, 0.1, 0.2; thread 1 handles cell 1.0, 1.1, 1.2, etc. This access is not optimal, as it cannot leverage the coalesce constraint of the GPU global memory.
I suggest changing the cell array to something like this
cell 0.0, cell 1.0, cell 2.0; cell 0.1, cell 1.1, cell 2.1, cell 0.2, cell 1.2, cell 2.2. i.e. the cells in the same ring are put close together.

perception/autoware_ground_segmentation_cuda/src/cuda_scan_ground_segmentation/cuda_scan_ground_segmentation_filter.cu
```diff
@@ -0,0 +1,1330 @@
...
+  CellCentroid * sector_cells_list_dev = &centroid_cells_list_dev[sector_start_cell_index];
+  for (int cell_index_in_sector = 0;
```
anhnv3991, line 779: Currently, each thread handles points in one sector. But because the number of sectors is quite small (360), we only have 360 threads to process a point cloud. That wastes the computing resource. I think we should let multiple threads to handles one sector. That may yield much better performance.

perception/autoware_ground_segmentation_cuda/src/cuda_scan_ground_segmentation/cuda_scan_ground_segmentation_filter.cu
```diff
@@ -0,0 +1,1330 @@
...
+{
+  uint32_t idx = blockIdx.x * blockDim.x + threadIdx.x;
+  if (idx >= max_num_cells) {
+    return;  // Out of bounds
+  }
+  centroid_cells_list_dev[idx].gnd_radius_avg = 0.0f;
+  centroid_cells_list_dev[idx].gnd_height_avg = 0.0f;
+  centroid_cells_list_dev[idx].gnd_height_min = 1e6f;
+  centroid_cells_list_dev[idx].num_points = 0;
+  centroid_cells_list_dev[idx].cell_id = 0;  // Initialize cell_id to -1
```
anhnv3991, line 857: Is this correct? The cell_id is being initialized to 0, not -1.

perception/autoware_ground_segmentation_cuda/src/cuda_scan_ground_segmentation/cuda_scan_ground_segmentation_filter.cu
```diff
@@ -0,0 +1,1330 @@
...
+  ClassifiedPointTypeStruct * classified_points_dev, CellCentroid * centroid_cells_list_dev,
+  const FilterParameters * filter_parameters_dev)
+{
+  const uint32_t num_sectors = filter_parameters_.num_sectors;
+  if (num_sectors == 0) {
+    return;
+  }
+
+  dim3 block_dim(1);
+  dim3 grid_dim((num_sectors + block_dim.x - 1) / block_dim.x);
```
anhnv3991, line 1077: This is not an efficient size of the kernel. With this setting, each block contains only one thread. However, the minimum scheduling unit on GPU is a wrap (32 threads). This mean the system still create a warp for a block, but only one thread is active, 31 other threads are idle. The computing resource is wasted. 
This should be changed to something like
```
  dim3 grid_dim(32);   // Or 64, 128, 256, multiple of 32
  dim3 grid_dim((num_sectors + block_dim.x - 1) / block_dim.x);
```

perception/autoware_ground_segmentation_cuda/src/cuda_scan_ground_segmentation/cuda_scan_ground_segmentation_filter.cu
```diff
@@ -0,0 +1,1330 @@
...
+  extractNonGroundPoints(
+    input_points_dev, classified_points_dev, ground_points_dev, num_ground_points,
+    PointType::GROUND);
+
+  returnBufferToPool(cell_counts_dev);
+  returnBufferToPool(num_points_per_cell_dev);
+  returnBufferToPool(cell_first_point_indices_dev);
+  returnBufferToPool(classified_points_dev);
+  returnBufferToPool(filter_parameters_dev);
+  returnBufferToPool(centroid_cells_list_dev);
```
anhnv3991, line 1319: I think it is better to define structures like std::vector, but for GPU memory to hold these kinds of arrays. By doing so, you don't have to manually free GPU buffers (it is done by destructors), and the code is much more readable.

perception/autoware_ground_segmentation_cuda/src/cuda_scan_ground_segmentation/cuda_scan_ground_segmentation_filter.cu
badai-nguyen: @anhnv3991 Thanks for your comment, I'm trying to keeps the algorithm the same as before to compare the peformance but I agree with you. I will try to remove this expensive function.

badai-nguyen:
close this PR as 
https://github.com/autowarefoundation/autoware_universe/pull/11371 merged

Issue #11269, opened 2025-08-25
gpu_monitor causes a failure in CI workflow cppcheck_differential


nishikawa-masaki:
### Checklist

- [x] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [x] I've searched other issues and no duplicate issues were found.
- [x] I'm convinced that this is not my fault but a bug.

### Description

When a Pull Request changes some codes in gpu_monitor_base.cpp, the CI workflow "cppcheck_differential" triggered by the change fails as the following lines.

> Run cat cppcheck-report.txt
> Error: /home/runner/work/autoware_universe/autoware_universe/system/autoware_system_monitor/src/gpu_monitor/gpu_monitor_base.hpp:70:16: style: Virtual function 'shut_down' is called from destructor '~GPUMonitorBase()' at line 60. Dynamic binding is not used. [virtualCallInConstructor]
>   virtual void shut_down();
>                ^
> /home/runner/work/autoware_universe/autoware_universe/system/autoware_system_monitor/src/gpu_monitor/gpu_monitor_base.cpp:60:3: note: Calling shut_down
>   shut_down();
>   ^
> /home/runner/work/autoware_universe/autoware_universe/system/autoware_system_monitor/src/gpu_monitor/gpu_monitor_base.hpp:70:16: note: shut_down is a virtual function
>   virtual void shut_down();
>                ^

This problem blocks the PR #10888 and needs a quick and independent fix because it may block other PRs.

### Expected behavior

The CI workflow "cppcheck_differential" should pass with no error.

### Actual behavior

The CI workflow "cppcheck_differential" fails and PRs related to gpu_monitor_base.cpp can't be merged.

### Steps to reproduce

1. Submit a PR which contains changes in gpu_monitor_base.cpp. Example: #10888 
2. The CI workflows start execution.
3. The workflow "cppcheck_differentials" fails.

### Possible causes

The error is reported because the destructor of the base class GPUMonitorBase calls the virtual function shut_down() defined by its derived class, which does not exist after destruction of the derived class.

The change was introduced by PR #11022.

### Additional context

Why didn't the CI workflow "cppcheck_differential" fail for PR #11022?
~~Was there any update in the CI workflow itself?~~
~~The CI workflow "cppcheck_differential" was introduced recently after PR #11022.~~ <- It was introduced 1 year ago.

The reason is still unknown, but it isn't worth the effort to find it out.

[...]

PR #11279, opened 2025-08-26
fix(gpu_monitor): explicitly qualify a function call to stop error from the cppcheck-differential CI workflow 

nishikawa-masaki:
## Description
Fix the problem with which the CI workflow "cppcheck_differential" complains about calling of a virtual function in a base-class destructor.

## Related links

**Parent Issue:**

- [gpu_monitor causes a failure in CI workflow cppcheck_differential #11269](#11269)

## How was this PR tested?

- All of CI workflows pass.
- gpu_monitor works in the same way as before.
- nvmlShutdown() is called in the process of termination.

## Notes for reviewers

- There is no change of behavior.
- The changes in this PR are:
  - Explicit call of GPUMonitorBase::shut_down() from the destructor "~GPUMonitorBase()" to stop "cppcheck_differential" from complaining.
    - Even without the "GPUMonitorBase::" qualifier, GPUMonitorBase::shut_down() were called.
  - Additional comments.
- Issue #11269 was introduced by the PR #11022. Its author iwatake2222 is invited for reviewing.

## Interface changes

None.

## Effects on system behavior

None.


iwatake2222 (approved):
LGTM!
Thank you for fixing it.

TetsuKawa (approved):
Approve for now

PR #11281, opened 2025-08-27
fix(autoware_probabilistic_occupancy_grid_map): skip frame when TF is unavailable to avoid crash

multimeters:
### Context
On tag v0.45.0, when transforming laser/obstacle/raw pointclouds in `laserscan_based_occupancy_grid_map_node`, TF gaps could lead to crashes or undefined behavior. This PR ensures that if TF is temporarily unavailable during pointcloud transforms, the current frame is skipped and a warning is logged, preventing a crash.

- ROS 2: Humble
- Build: CPU
- Reproduction:  Start the complete Autoware stack, including localization,perception and simulate transient TF loss between `base_link` -> `map` by continually initializing the localization system, which will temporarily lose TF between these frames.

### Root cause
The issue arises when **TF transformations between the `base_link` and `map` frames** are temporarily unavailable or delayed. When the node attempts to process pointcloud data without valid transformations, it proceeds with invalid or missing data, leading to crashes or undefined behavior. Specifically, when a transform error occurs, the system attempts to access data that doesn't exist, leading to errors like:

```
terminate called after throwing an instance of 'std::runtime_error'
what(): Field x does not exist
```

### Fix
- Check the return values of `transformPointcloud` for laser, obstacle, and raw pointclouds.
- If any of the transforms fail, log a warning with the source/target frames and timestamp, then skip processing the current frame.

Sample warning:
```
Skip frame: TF unavailable (ls:%d obs:%d raw:%d). from [%s] to [%s] at %.3f  
```
### Behavior change
- **Before:** Node would crash or block when TF was unavailable.
- **After:** The node continues running; frames with missing TF are skipped and logged as warnings.

### Tests
- `pre-commit` passed.
- `colcon build/test --packages-select autoware_probabilistic_occupancy_grid_map` passed.
- Verified runtime by introducing TF gaps: no crash; frames were skipped as expected with corresponding warnings.

### Risk & Impact
- Low risk; this change only affects frames where TF is missing.
- No changes to ROS interface or parameters.

### Performance
- Negligible overhead from boolean checks and early returns.

### Checklist
- [x] DCO signed (`-s`)
- [x] Conventional commit title
- [x] Pre-commit hooks pass
- [x] Package builds and tests pass

### DIFF
--- a/perception/autoware_probabilistic_occupancy_grid_map/src/laserscan_based_occupancy_grid_map/laserscan_based_occupancy_grid_map_node.cpp
+++ b/perception/autoware_probabilistic_occupancy_grid_map/src/laserscan_based_occupancy_grid_map/laserscan_based_occupancy_grid_map_node.cpp
@@ -192,9 +192,22 @@ void LaserscanBasedOccupancyGridMapNode::onLaserscanPointCloud2WithObstacleAndRa
       inner_st_ptr = std::make_unique<ScopedTimeTrack>("transformPointcloud", *time_keeper_);
 
     try {
-      utils::transformPointcloud(*laserscan_pc_ptr, *tf2_, map_frame_, trans_laserscan_pc);
-      utils::transformPointcloud(filtered_obstacle_pc, *tf2_, map_frame_, trans_obstacle_pc);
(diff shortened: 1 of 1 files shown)


technolojin:

PR #11282, opened 2025-08-27
fix(blind_spot): cut virtual blind lines up to intersection lanelet's bound

zulfaqar-azmi-t4:
## Description

cut `virtual_blind_lane_boundary_after_turning` and ``virtual_ego_straight_path_after_turning` so that it doesn't exceed intersection lanelet's boundary.

| Before | After |
|:---:|:----:|
|  image |  image |
| image    | image |
| image  | image |

## Related links

**Parent Issue:**

- [TIER IV Internal Link - Issue's ticket](https://tier4.atlassian.net/browse/RT0-39021) 

## How was this PR tested?

- Perception reproducer

| Before | After |
|:---:|:----:|
| cap- 2025-08-27-11-27-45   | cap- 2025-08-27-11-29-15 |
| Object became blind spot target | Object is excluded from blind spot target |

- Internal evaluation

[TIER IV Internal Link - ControlDLR Test](https://evaluation.tier4.jp/evaluation/reports/369e033b-3f4e-5552-9b25-5b43ab8a6455?project_id=prd_jt)
[TIER IV Internal Link - FuncVerification Test](https://evaluation.tier4.jp/evaluation/reports/c1b33777-37d7-5f6f-a9ce-58e3ca06211d?project_id=prd_jt)
[TIER IV Internal Link - CommonScenario_Basic Test](https://evaluation.tier4.jp/evaluation/reports/8a7093dc-554a-53e7-ac5a-1ea900d6b8d7?project_id=prd_jt)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/src/util.cpp
+++ b/planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/src/util.cpp
@@ -115,7 +115,8 @@ Eigen::Vector3d linestring_normal_direction(
  * @brief extend the last part of given `line` by some length
  */
 lanelet::LineString3d generate_segment_beyond_linestring_end(
-  const lanelet::ConstLineString3d & line, const lanelet::ConstLanelet & intersection_lanelet)
+  const lanelet::ConstLineString3d & line, const lanelet::ConstLanelet & intersection_lanelet,
+  const autoware::experimental::lanelet2_utils::TurnDirection & turn_direction)
 {
   const auto extend_length = lanelet::utils::getLaneletLength3d(intersection_lanelet);
   const auto size = line.size();
@@ -123,7 +124,7 @@ lanelet::LineString3d generate_segment_beyond_linestring_end(
   const auto & p2 = line[size - 1];
   const auto p3 = autoware::experimental::lanelet2_utils::extrapolate_point(p1, p2, extend_length);
   return clip_virtual_line_to_intersection_bound(
-    remove_const(p2), remove_const(p3), intersection_lanelet);
+    remove_const(p2), remove_const(p3), intersection_lanelet, turn_direction);
 };
 
 template <typename L1, typename L2>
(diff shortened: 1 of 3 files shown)


planning/behavior_velocity_planner/autoware_behavior_velocity_blind_spot_module/src/util.cpp
```diff
@@ -528,9 +527,51 @@ std::optional<lanelet::LineString3d> generate_virtual_ego_straight_path_after_tu
...
+    for (size_t i = 0; i + 1 < linestring.size(); ++i) {
+      const auto & p1 = linestring[i];
+      const auto & p2 = linestring[i + 1];
+
+      auto linestring_segment = to_bg2d(std::vector{p1, p2});
+      boost::geometry::intersection(virtual_line, linestring_segment, intersected_points);
+    }
+  };
+
+  get_intersected_points(intersection_lanelet.leftBound());
```
soblin: I think it is better to pass `turn_direction` as the argument, and if it is `Left(Right)` use `rightBound() (leftBound())`.
Is there some reason for not doing so ?

zulfaqar-azmi-t4: Because be have the following case

image

where we have `virtual_blind_lane_boundary_after_turning` intersects with both left and right boundaries. If we just take the turn direction, then the virtual line will be cut short.

soblin: But we want to find the further intersection point, and isn't it obvious that rightBound() gives the farthest point for Left lanelet ?

zulfaqar-azmi-t4: yes, that actually makes total sense.
I have revised the code https://github.com/autowarefoundation/autoware_universe/pull/11282/commits/78902ed08607e91306278cb252b39aefa245e1c2

soblin (approved):

PR #11289, opened 2025-08-28
feat(autoware_compare_map_segmentation): add lanelet elevation filter

kaancolak:
### Description

The new lanelet_elevation_filter package filters point clouds based on lanelet elevation information. It first creates a grid-based elevation map from lanelet data, then filters out points that deviate significantly from the expected road surface height.

#### Why do we need this package?

After the compare map filter operation, and just before the euclidean clustering operation, the low_height_cropbox operation ([link](https://github.com/autowarefoundation/autoware_universe/blob/main/perception/autoware_euclidean_cluster/launch/voxel_grid_based_euclidean_cluster.launch.py?utm_source=chatgpt.com#L37)
) is applied to the point cloud. This useful step removes unnecessary points left over from the voxel-based compare map filter, such as dynamic changes (e.g., branches of trees). Otherwise, these objects can cause undesired braking.

However, the problem is that the crop box operation is performed on the static local base_link frame. If the road plane changes along the direction of travel, the algorithm cannot handle these changes properly.

When the road is sloped, especially for long-range points, valid points may be incorrectly removed by the crop box filter. To address this, I designed a filter based on lanelet data that adapts to road surface changes.

As shown in the illustration below, there is a gap between the crop box polygon and the actual road surface.

#### Basic illustration of the problem

Points in the gap between the crop box polygon and the actual road surface are removed.

Untitled Diagram drawio

#### Real-world samples

In some cases, the crop box polygon does not align with the road surface, resulting in the removal of valid vehicle points.

For example, in the opposite slope case compared to the previous illustration, the real road surface lies above the crop box polygon.

[https://github.com/user-attachments/assets/67ff897e-5ab0-4b96-ba8d-45a643c490f6](https://github.com/user-attachments/assets/67ff897e-5ab0-4b96-ba8d-45a643c490f6?utm_source=chatgpt.com)

#### Sample image
fail_case_2

#### Debug marker visualization
Screenshot from 2025-08-28 13-01-39

#### Future steps

Replace the low_height_cropbox operation ([link](https://github.com/autowarefoundation/autoware_universe/blob/main/perception/autoware_euclidean_cluster/launch/voxel_grid_based_euclidean_cluster.launch.py?utm_source=chatgpt.com#L37)) with this package.

Update autoware_launch accordingly.

#### Performance

Euclidean clustering takes around 10k points before operation, overall processing time for this package is approximately 1 ms.

- Large input size: ~200,000 points → processing time: ~17 ms

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

Adjust input point cloud and lanelet topics. 

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

No effect observed.

None.

### DIFF
--- a/perception/autoware_compare_map_segmentation/src/lanelet_elevation_filter/grid_processor.cpp
+++ b/perception/autoware_compare_map_segmentation/src/lanelet_elevation_filter/grid_processor.cpp
@@ -1,0 +1,268 @@
+// Copyright 2025 Tier IV, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "grid_processor.hpp"
+
+#include <lanelet2_core/geometry/Lanelet.h>
+
+#include <algorithm>
+#include <cmath>
+#include <limits>
+
+namespace autoware::compare_map_segmentation
+{
+
+GridProcessor::GridProcessor(double grid_resolution) 
+: grid_resolution_(grid_resolution), inv_grid_resolution_(1.0 / grid_resolution)
+{
+}
+
+void GridProcessor::processLanelets(const lanelet::LaneletMapPtr & lanelet_map)
+{
+  reset();
+  
+  if (!lanelet_map) {
+    return;
+  }
+
+  // Process all lanelets in the map
+  for (const auto & lanelet : lanelet_map->laneletLayer) {
+    processLaneletBoundary(lanelet);
+  }
+  
+  calculateGridStatistics();
+}
+
+GridIndex GridProcessor::getGridIndex(double x, double y) const
+{
+  // Use precomputed inverse resolution for speed
+  GridIndex index;
+  index.x = static_cast<int>(x * inv_grid_resolution_);
+  index.y = static_cast<int>(y * inv_grid_resolution_);
+  
+  // Handle negative coordinates correctly
+  if (x < 0.0 && (x * inv_grid_resolution_) != index.x) index.x--;
+  if (y < 0.0 && (y * inv_grid_resolution_) != index.y) index.y--;
+  
+  return index;
+}
+
+void GridProcessor::addPointToGrid(double x, double y, double z)
+{
+  GridIndex index = getGridIndex(x, y);
+  
+  auto it = grid_cells_.find(index);
+  if (it == grid_cells_.end()) {
+    grid_cells_[index] = GridCell();
+  }
+  
+  geometry_msgs::msg::Point point;
+  point.x = x;
+  point.y = y;
+  point.z = z;
+  
+  grid_cells_[index].points.push_back(point);
+}
+
+void GridProcessor::processLaneletBoundary(const lanelet::Lanelet & lanelet)
+{
+  double sampling_distance = grid_resolution_ * 0.5;  // Sample at half grid resolution
+  
+  // Sample points from left boundary
+  auto left_points = samplePointsFromLineString(lanelet.leftBound(), sampling_distance);
+  for (const auto & point : left_points) {
+    addPointToGrid(point.x, point.y, point.z);
+  }
+  
+  // Sample points from right boundary
+  auto right_points = samplePointsFromLineString(lanelet.rightBound(), sampling_distance);
+  for (const auto & point : right_points) {
+    addPointToGrid(point.x, point.y, point.z);
+  }
+  
+  // Sample points from centerline if available
+  if (lanelet.hasAttribute(lanelet::AttributeName::Subtype)) {
+    try {
+      auto centerline = lanelet.centerline();
+      auto center_points = samplePointsFromLineString(centerline, sampling_distance);
+      for (const auto & point : center_points) {
+        addPointToGrid(point.x, point.y, point.z);
+      }
+    } catch (...) {
+      // Centerline might not be available, continue without it
+    }
+  }
+}
+
+std::vector<geometry_msgs::msg::Point> GridProcessor::samplePointsFromLineString(
+  const lanelet::ConstLineString3d & linestring, double sampling_distance) const
+{
+  std::vector<geometry_msgs::msg::Point> sampled_points;
+  
+  if (linestring.size() < 2) {
+    return sampled_points;
+  }
+  
+  for (size_t i = 0; i < linestring.size() - 1; ++i) {
+    const auto & start_point = linestring[i];
+    const auto & end_point = linestring[i + 1];
+    
+    double dx = end_point.x() - start_point.x();
+    double dy = end_point.y() - start_point.y();
+    double dz = end_point.z() - start_point.z();
+    double segment_length = std::sqrt(dx * dx + dy * dy);
+    
+    if (segment_length < 1e-6) {
+      continue;
+    }
+    
+    int num_samples = static_cast<int>(segment_length / sampling_distance) + 1;
+    
+    for (int j = 0; j <= num_samples; ++j) {
+      double ratio = static_cast<double>(j) / num_samples;
+      
+      geometry_msgs::msg::Point point;
+      point.x = start_point.x() + ratio * dx;
+      point.y = start_point.y() + ratio * dy;
+      point.z = start_point.z() + ratio * dz;
+      
+      sampled_points.push_back(point);
+    }
+  }
+  
+  return sampled_points;
+}
+
+void GridProcessor::calculateGridStatistics()
+{
+  for (auto & [index, cell] : grid_cells_) {
+    if (cell.points.empty()) {
+      continue;
+    }
+    
+    // Calculate average height in single pass
+    double sum_z = 0.0;
+    const size_t num_points = cell.points.size();
+    
+    for (size_t i = 0; i < num_points; ++i) {
+      sum_z += cell.points[i].z;
+    }
+    
+    cell.average_height = sum_z / num_points;
+    cell.is_valid = true;
+    
+    // Clear points vector to save memory after statistics calculation
+    cell.points.clear();
+    cell.points.shrink_to_fit();
+  }
+}
+
+double GridProcessor::getElevationAtPoint(double x, double y) const
+{
+  GridIndex index = getGridIndex(x, y);
+  
+  // Check cache first
+  auto cache_it = elevation_cache_.find(index);
+  if (cache_it != elevation_cache_.end()) {
+    return cache_it->second;
+  }
+  
+  auto it = grid_cells_.find(index);
+  if (it != grid_cells_.end() && it->second.is_valid) {
+    // Cache the direct result
+    double elevation = it->second.average_height;
+    elevation_cache_[index] = elevation;
+    return elevation;
+  }
+  
+  // For missing cells, use fast nearest neighbor instead of expensive interpolation
+  double min_distance = std::numeric_limits<double>::max();
+  double nearest_elevation = 0.0;
+  bool found_neighbor = false;
+  
+  // Check immediate neighbors only (3x3 grid)
+  for (int dx = -1; dx <= 1; ++dx) {
+    for (int dy = -1; dy <= 1; ++dy) {
+      if (dx == 0 && dy == 0) continue;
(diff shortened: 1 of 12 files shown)


badai-nguyen:
@kaancolak Thanks for your PR. I think this is the right approach to address the current issue with the cropbox filter.

Regarding your estimation:  `Large input size: ~200,000 points → processing time: ~17 ms` — could you please share more details about the environment you used for testing?

Also, from my quick look, it seems that the filter processing time may also depend on the complexity (details) and the size of the lanelet. Am I understanding this correctly?

kaancolak:
@badai-nguyen Thank you very much for your quick review.

I tested it on my local machine with a 12th Gen Intel® Core™ i7-12700H × 20. I checked the results while Autoware was running in radar–lidar–camera modes.

Lanelet to grid map generation takes around 200 ms in my environment and it happens only once at launch. Complex environments do not directly affect filter processing time, but they do affect grid map generation.

Alternatively, it is possible to save the grid map for future use. While 200 ms seems acceptable, it could be higher in more complex environments, which needs to be confirmed, but  I do not think it would exceed a couple of seconds.

perception/autoware_compare_map_segmentation/src/lanelet_elevation_filter/grid_processor.cpp
```diff
@@ -0,0 +1,260 @@
...
+    double segment_length = std::sqrt(dx * dx + dy * dy);
+
+    if (segment_length < 1e-6) {
+      continue;
+    }
+
+    int num_samples = static_cast<int>(segment_length / sampling_distance) + 1;
+
+    for (int j = 0; j <= num_samples; ++j) {
+      double ratio = static_cast<double>(j) / num_samples;
```
badai-nguyen:
```suggestion
    double inv_num_samples = 1 / num_samples;
    for (int j = 0; j <= num_samples; ++j) {
      double ratio = static_cast<double>(j) * inv_num_samples;
```

perception/autoware_compare_map_segmentation/src/lanelet_elevation_filter/grid_processor.cpp
```diff
@@ -0,0 +1,260 @@
...
+  auto cache_it = elevation_cache_.find(index);
+  if (cache_it != elevation_cache_.end()) {
+    return cache_it->second;
+  }
+
+  auto it = grid_cells_.find(index);
+  if (it != grid_cells_.end() && it->second.is_valid) {
+    // Cache the direct result
+    double elevation = it->second.average_height;
+    elevation_cache_[index] = elevation;
```
badai-nguyen: Q: Is the cache size will be increased when the length of operation path is large? If so, adding CACHE_SIZE_THRESHOLD might better?

perception/autoware_compare_map_segmentation/src/lanelet_elevation_filter/grid_processor.cpp
```diff
@@ -0,0 +1,260 @@
...
+      if (neighbor_it != grid_cells_.end() && neighbor_it->second.is_valid) {
+        double distance = std::abs(dx) + std::abs(dy);
+        if (distance < min_distance) {
+          min_distance = distance;
+          nearest_elevation = neighbor_it->second.average_height;
+          found_neighbor = true;
+        }
+      }
+    }
+  }
```
badai-nguyen: IMO, depend on the gradient of road surface, the high level point's grid index might quite different compared with lanelet grids that need to compare, Am I correct?
If so Could you please estimation the error of this assumption? Or the number of the necessary neighbors grids to extend related to point.z, road surface gradient?
My understanding as this picture.  

image

badai-nguyen:
@kaancolak Thank you for your clarify. It's very nice idea. 
I leave some other questions and comments

badai-nguyen:
> I tested it on my local machine with a 12th Gen Intel® Core™ i7-12700H × 20. I checked the results while Autoware was running in radar–lidar–camera modes.

@kaancolak 
Sorry for my ambigious question, 
My question is above the operation environment (lanelet), could you give me a rough information?
- How large the map is?
- How long the operation path? 
- Does it have any slope road pattern?

perception/autoware_compare_map_segmentation/src/lanelet_elevation_filter/node.cpp
```diff
@@ -0,0 +1,245 @@
...
+    throw std::invalid_argument("height_threshold must be positive");
+  }
+  if (params_.sampling_distance <= 0.0) {
+    throw std::invalid_argument("sampling_distance must be positive");
+  }
+}
+
+void LaneletElevationFilterComponent::filter(
+  const PointCloud2ConstPtr & input, [[maybe_unused]] const IndicesPtr & indices,
+  PointCloud2 & output)
```
badai-nguyen, line 149: @kaancolak Which `frame_id` are you supposed to process the pointcloud? 
IMO, it is better to documentation it?

perception/autoware_compare_map_segmentation/src/lanelet_elevation_filter/grid_processor.cpp
kaancolak: Yes, you are right. As a sample error estimation, road gradients are usually designed with a maximum of about 9 degrees, and rarely up to 12 degrees. With the current height threshold set to 2 meters, the expected basic calculation would be as follows: with a 2-degree axis error, there is approximately a 0.06-meter height error

Başlıksız Diyagram drawio

kaancolak:
@badai-nguyen you can reach related maps from these links:
- https://console.fms.tier4.jp/erga/b7d2e260-5417-4fc4-a3f3-c7654843844a/driving-environment/
- https://console.fms.tier4.jp/erga/b4d3890f-49c2-4f42-897a-bb8c36280ccb/driving-environment/

I tested two environments: one approximately 1 km × 1.5 km with an operational path of around 4.5 km, and another one is test area which was partially tested. Both areas contain roads with partially varying gradients.

I refactored the code and made some modifications to the algorithm;  filled empty grids during the first map generation for faster processing and implemented a basic mechanism to save cached maps and reuse them later.

kaancolak:
@badai-nguyen , do you have any consideration or concern about this feature?

perception/autoware_compare_map_segmentation/src/lanelet_elevation_filter/grid_processor.cpp
```diff
@@ -0,0 +1,672 @@
...
+  calculateGridStatistics();
+
+  // Fill empty grids around original points with specified extension count
+  fillEmptyGrids(extension_count);
+}
+
+GridIndex GridProcessor::getGridIndex(double x, double y) const
+{
+  GridIndex index;
+  index.x = static_cast<int>(x * inv_grid_resolution_);
```
badai-nguyen, line 49: Question: since x, y is coordinate in map frame, do you think there might be overflow issue?

perception/autoware_compare_map_segmentation/src/lanelet_elevation_filter/grid_processor.cpp
```diff
@@ -0,0 +1,672 @@
...
+              radius_limit_reached = true;
+            }
+          }
+        }
+      }
+    }
+
+    // After each radius, calculate statistics for the newly added cells
+    // This makes them valid for the next radius iteration
+    calculateGridStatistics();
```
badai-nguyen: IMO, since `calculateGridStatics()` includes `shrink_to_fit`, is better to avoid calling it multiple times?
Shall we mark added cell as valid and skip call this function each radius increment?

perception/autoware_compare_map_segmentation/src/lanelet_elevation_filter/grid_processor.cpp
```diff
@@ -0,0 +1,672 @@
...
+    return;
+  }
+
+  // Process all lanelets to generate map from original points
+  for (const auto & lanelet : lanelet_map->laneletLayer) {
+    processLaneletBoundary(lanelet);
+  }
+
+  // Calculate statistics for original points
+  calculateGridStatistics();
```
badai-nguyen: `calculateGridsStatics` is repetedly called inside `fillEmptyGrids`, is better to call `calculateGridsStatics()` only onetime at the end of processLanelets?

perception/autoware_compare_map_segmentation/src/lanelet_elevation_filter/grid_processor.cpp
```diff
@@ -0,0 +1,672 @@
...
+    for (const auto & [elevation, weight] : neighbor_elevations_with_weight) {
+      weighted_sum += elevation * weight;
+      weight_sum += weight;
+    }
+
+    return weighted_sum / weight_sum;
+  }
+
+  // No neighbors found - return default elevation
+  return 0.0;
```
badai-nguyen: IMO, return some kind of no neighbors status and handle it as skipping filtering pointcloud might be better since the default elevation as 0.0 seems not be appropriate. How do you think?

perception/autoware_compare_map_segmentation/src/lanelet_elevation_filter/grid_processor.cpp
```diff
@@ -0,0 +1,672 @@
...
+  // First, process all lanelets to generate map from original points
+  for (const auto & lanelet : lanelet_map->laneletLayer) {
+    processLaneletBoundary(lanelet);
+  }
+
+  // Calculate statistics for original points
+  calculateGridStatistics();
+
+  // Then, fill empty grids around original points with specified extension
+  fillEmptyGrids(extension_size);
```
badai-nguyen: The same comment with `ProcessLanelets`.

perception/autoware_compare_map_segmentation/src/lanelet_elevation_filter/grid_processor.cpp
kaancolak: In realistic scenarios, it’s unlikely to reach the integer limit. However, with unnecessarily small resolution parameters, it could happen. The maximum value for a 32-bit signed integer is about 2.1 billion.

UTM maximum values are roughly 834,000 m (Easting) and 10,000,000 m (Northing).

If Northing were stored with a 0.005 m resolution, it would require around 2.0 billion steps, which is close to the 32-bit limit and could cause an overflow. However, a 0.005 m resolution is not practical for real-time applications, though it may still be useful to validate the parameter during the initial load process.

I updated the validation rule so that the smallest allowed resolution is 0.1 m. Otherwise, it will trigger RCLCPP_ERROR: "Grid resolution must be greater than 0.1 meters."

kaancolak:
@badai-nguyen , Thank you very much for your valuable comments and reviews. I agree with your points and have applied your suggestions.

perception/autoware_compare_map_segmentation/src/lanelet_elevation_filter/grid_processor.cpp
badai-nguyen: @kaancolak Thank you for your clarification. It's clear for me now.

badai-nguyen:
@kaancolak I'm sorry for late response. The overal looks good to me. Could you solve the required CI check?

kaancolak:
@badai-nguyen, thank you for your review. I’ve fixed the required CI checks, so everything should be fine now

badai-nguyen:
I tested this node together with other ground_segmentation node as below and comfirmed this worked as desired. Please checked attached video below.
- white pointcloud: input pointcloud 
- green: output of `/lanelet_elevation_filter`

| Test case 1 | Test case 2| 
| --- | ---|
| input poincloud topic `obstacle_segmentation/pointcloud` | input pointcloud topic: `pointcloud_map_filtered/pointcloud` |
| `ros2 launch autoware_compare_map_segmentation lanelet_elevation_filter.launch.xml input_pointcloud:=/perception/obstacle_segmentation/pointcloud`| `ros2 launch autoware_compare_map_segmentation lanelet_elevation_filter.launch.xml input_pointcloud:=/perception/object_recognition/detection/pointcloud_map_filtered/pointcloud`|
| number pointcloud: ~ 168496 points | number pointcloud: ~ 15062|
| processing time: ~ 20ms | processing time: ~ 3ms|
|image| Screenshot from 2025-09-18 22-16-07 |
| [Screencast from 2025年09月18日 22時01分02秒.webm](https://github.com/user-attachments/assets/0c0a2bc5-d4fb-4a1f-9214-6ccf42cf7663)| [Screencast from 2025年09月18日 22時16分45秒.webm](https://github.com/user-attachments/assets/308229e3-d6e6-4016-99db-ec02e2db1524)|

badai-nguyen (commented):
NITs:
- It would be great to add some unit tests to cover the changes.
- Consider using the schema to auto-generate the parameters section in README.md for consistency

kaancolak:
@badai-nguyen Thanks for the suggestions! I’ll check and address them when I have time.

PR #11290, opened 2025-08-28
fix(autoware_trajectory_optimizer): reduce excessive optimizer logging

danielsanchezaran:
## Description
Changes warning and error messages in the optimizer to be of throttle type as to not pollute the terminal excessively.
## Related links

**Parent Issue:**

- Link

## How was this PR tested?
PSim
## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


SakodaShintaro (approved):
:+1:

Issue #11285, opened 2025-08-27
gpu_monitor reports a false warning every second when GPU is idle


nishikawa-masaki:
### Checklist

- [x] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [x] I've searched other issues and no duplicate issues were found.
- [x] I'm convinced that this is not my fault but a bug.

### Description

gpu_monitor of autoware_system_monitor reports a false warning message every second when the GPU is completely idle.

> [component_container_mt-1] [WARN 1756285485.062733682] [gpu_monitor]: Failed to nvmlDeviceGetProcessUtilization(2nd) NVML: Not Found (addProcessUsage() at /home/masakinishikawa/work/fix_issue_11269/autoware/src/universe/autoware_universe/system/autoware_system_monitor/src/gpu_monitor/nvml_gpu_monitor.cpp:242)

According to [the specification of nvmlDeviceGetProcessUtilization()](https://docs.nvidia.com/deploy/nvml-api/group__nvmlDeviceQueries.html#group__nvmlDeviceQueries_1gb0ea5236f5e69e63bf53684a11c233bd:~:text=nvmlReturn_t%20nvmlDeviceGetProcessUtilization%20(%20nvmlDevice_t%C2%A0device%2C%20nvmlProcessUtilizationSample_t*%C2%A0utilization%2C%20unsigned%20int*%C2%A0processSamplesCount%2C%20unsigned%20long%20long%C2%A0lastSeenTimeStamp%20)), the function returns [NVML_ERROR_NOT_FOUND](https://docs.nvidia.com/deploy/nvml-api/group__nvmlDeviceEnumvs.html#group__nvmlDeviceEnumvs_1gg06fa9b5de08c6cc716fbf565e93dd3d0a1bfb13c6d0f00249b20f556c2d5d23f) if sample entries are not found.

It means that, when the GPU is completely idle, nvmlDeviceGetProcessUtilization() finds no sample entry and returns NVML_ERROR_NOT_FOUND, because there is **no** process that uses the GPU.

This is a **false warning** and gpu_monitor should report "GPU Usage 0%" with no error or warning.
The false warning messages are recorded in log and they will lower the quality of the log data.

In addition to the main issue, gpu_monitor dynamically allocates/frees heap memory every second.
```
  std::unique_ptr<nvmlProcessUtilizationSample_t[]> utils;
[...]

PR #11291, opened 2025-08-28
fix(gpu_monitor): stop false warning messages reported by gpu_monitor when the GPU is idle

nishikawa-masaki:
## Description
"Gpu_monitor" of "autoware_system_monitor" is implemented by nvml_gpu_monitor.cpp for NVML based platforms.
Nvml_gpu_monitor.cpp uses nvmlDeviceGetProcessUtilization() to retrieve the name, the PID and the GPU usage of the process(es) that uses the GPU.
When there is no such process (i.e. the GPU is idle), nvmlDeviceGetProcessUtilization() returns NVML_ERROR_NOT_FOUND.

This error should not be reported because it is one of the expected results.

## Related links

**Parent Issue:**

- #11285 

**Private Links:**

- [TIER IV internal link](https://tier4.atlassian.net/browse/RT1-11131)

## How was this PR tested?

- All CI workflows passed.
- When the GPU is idle, no false warning is reported on the terminal or in ROS2 log files.
- When the GPU is busy, the name, the PID and the GPU usage of the GPU user is reported in the "/diagnostics" topic.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


sasakisasaki (approved):
The test looks performed for both cases: GPU is idle, or not. I understand this is for noisy log suppression :+1:

PR #11293, opened 2025-08-29
fix(traffic_light_module): fix traffic light

yhisaki:
## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_velocity_planner/autoware_behavior_velocity_traffic_light_module/src/scene.hpp
+++ b/planning/behavior_velocity_planner/autoware_behavior_velocity_traffic_light_module/src/scene.hpp
@@ -78,7 +78,8 @@ public:
 public:
   TrafficLightModule(
     const int64_t lane_id, const lanelet::TrafficLight & traffic_light_reg_elem,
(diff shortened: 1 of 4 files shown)


planning/behavior_velocity_planner/autoware_behavior_velocity_traffic_light_module/src/scene.hpp
```diff
@@ -121,6 +124,7 @@ class TrafficLightModule : public SceneModuleInterfaceWithRTC
   // Key Feature
   const lanelet::TrafficLight & traffic_light_reg_elem_;
   lanelet::ConstLanelet lane_;
+  lanelet::ConstLineString3d stop_line_;
```
soblin: rename or add comment for this variable to make it clear that it could be a stopline associated with other lanelet than this

yhisaki: @soblin 
I added comment in ce4701d

PR #11297, opened 2025-09-01
fix(traffic_light_camera_fusion): change group fusion algorithm

toki-1441:
## Description
---
### Problem
The primary issue we are currently addressing is the methodology for consolidating detection results from multiple traffic signals allocated to a single intersection. The current specification operates by first filtering signals based on two criteria: whether a signal is located at the edge of the Region of Interest (ROI) and whether its timestamp is the most recent. Following this filtration, the system selects the detection with the highest confidence score.

A key challenge with this approach is that the confidence score for signal clustering is consistently high, typically in the vicinity of 0.99. As a consequence, in situations where different signals at the same intersection yield conflicting detection results, the system outputs an incorrect state, a behavior we have confirmed via video analysis.

### Change
To enhance stability and accuracy, we propose the following modification: when multiple signal detections for a single intersection persist after the initial filtering stage, the system will aggregate the confidence scores for each distinct color. The color that achieves the highest total score will be selected as the definitive result for that intersection.

This revised logic will enable the system to produce a more robust and reliable detection outcome, as it can mitigate the impact of a single false detection and instead rely on a consensus from all available signals.

[Before](https://github.com/user-attachments/assets/4de38cd0-3603-405d-875d-b95e32fce941)

[After](https://github.com/user-attachments/assets/4cb486d2-36a5-406e-b4ed-004412ed364d)

## Related links
- [Ticket](https://tier4.atlassian.net/browse/RT0-38651)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


perception/autoware_traffic_light_multi_camera_fusion/src/traffic_light_multi_camera_fusion_node.cpp
```diff
@@ -308,19 +310,40 @@ void MultiCameraFusion::groupFusion(
...
+    const auto & record = p.second;
+    if (record.signal.elements.empty()) {
+      continue;
+    }
+    const uint8_t color = record.signal.elements[0].color;
+    const double confidence = record.signal.elements[0].confidence;
     const auto reg_ele_id_vec =
       traffic_light_id_to_regulatory_ele_id_[p.second.roi.traffic_light_id];
     for (const auto & reg_ele_id : reg_ele_id_vec) {
+      group_fusion_info_map[reg_ele_id].accumulated_scores[color] += confidence;
```
Owen-Liuyuxuan: Thanks for the proposal! 

Would you consider applying Bayesian updates instead of direct score summation?
And also, please help us summarize the mathematical model of the fusion algorithm in the documentation.

toki-1441: I have changed the mechanism to process using Bayesian updating.
### result
[bayesian.webm](https://github.com/user-attachments/assets/440346cb-2e0f-4b6a-b9f8-858060f39e05)

perception/autoware_traffic_light_multi_camera_fusion/src/traffic_light_multi_camera_fusion_node.hpp
```diff
@@ -124,6 +130,18 @@ class MultiCameraFusion : public rclcpp::Node
...
+  /**
+   * @brief The prior log-odds for a traffic light state.
+   *
+   * This corresponds to a prior probability of 0.5, representing a state of
+   * complete uncertainty before any evidence is considered.
+   *
+   * Example of a biased prior: If we wanted to assume a slight bias
+   * towards "Red" by default (e.g., 60% probability), we would set this
+   * to log(0.6 / 0.4) ≈ 0.405.
+   * static constexpr double PRIOR_LOG_ODDS_BIASED_TOWARDS_RED = 0.405;
```
Owen-Liuyuxuan: Thank you very much for adding this example. This is crucial for the future. But is it possible to make this a configurable parameter in the configuration file?

toki-1441: Thank you. I've parameterized the variables so they can be modified via YAML. I've also updated the schema accordingly.

toki-1441:
this is autoware launch PR https://github.com/autowarefoundation/autoware_launch/pull/1639

```diff_yaml:traffic_light_multi_camera_fusion.param.yaml
  ros__parameters:
    message_lifespan: 0.09
    approximate_sync: false
+    prior_log_odds: 0.0
```

shmpwk:
@tzhong518 @MasatoSaeki 
Could you check this PR as a code owner?

perception/autoware_traffic_light_multi_camera_fusion/src/traffic_light_multi_camera_fusion_node.cpp
```diff
@@ -308,19 +334,56 @@ void MultiCameraFusion::groupFusion(
...
+       * confidence scoring systems. If the confidence metric is not a true probability, this
+       * conversion may be invalid. Future maintainers should verify that the confidence values used
+       * here are indeed probabilities, or update this logic if the scoring system changes.
+       */
+      double evidence_log_odds = probabilityToLogOdds(confidence);
+
+      // We assume the prior probability (with no information) is 0.5, meaning the log odds = 0, and
+      // then add evidence to it.
+      group_fusion_info_map[reg_ele_id].accumulated_log_odds[color] +=
+        evidence_log_odds + prior_log_odds_;
```
a-maumau: I think this part is a bug. For each loop it will add the prior log-odds too. It should be added (initialized) once.
Or is this intended to add some bias for whole predictions? Then, the name should be made more explicit.

perception/autoware_traffic_light_multi_camera_fusion/src/traffic_light_multi_camera_fusion_node.cpp
```diff
@@ -308,19 +334,56 @@ void MultiCameraFusion::groupFusion(
...
 
-    /*
-    keep the best record for every regulatory element id
-    */
+    const auto & record = p.second;
+    if (record.signal.elements.empty()) {
+      continue;
+    }
+    const uint8_t color = record.signal.elements[0].color;
+    const double confidence = record.signal.elements[0].confidence;
```
MasatoSaeki: I think we should consider like `{red, circle}, {red, straight_arrow}, {red, right_arrow}, {green, circle}, ......`, it is not good to consider `{red}, {green}, ......`
What do you think?

perception/autoware_traffic_light_multi_camera_fusion/src/traffic_light_multi_camera_fusion_node.cpp
toki-1441: Your observation is correct. This is indeed a bug. I've made the necessary fixes. Thank you.

perception/autoware_traffic_light_multi_camera_fusion/src/traffic_light_multi_camera_fusion_node.cpp
toki-1441: I made it compatible with both color and shape. This was a critical issue. Thank you.

perception/autoware_traffic_light_multi_camera_fusion/src/traffic_light_multi_camera_fusion_node.cpp
```diff
@@ -308,19 +334,67 @@ void MultiCameraFusion::groupFusion(
...
+        * confidence scoring systems. If the confidence metric is not a true probability, this
+        * conversion may be invalid. Future maintainers should verify that the confidence values used
+        * here are indeed probabilities, or update this logic if the scoring system changes.
+        */
+
+        // Get a reference to the log-odds map for the current regulatory element ID.
+        auto & log_odds_map = group_fusion_info_map[reg_ele_id].accumulated_log_odds;
+        // The prior should only be applied once at the beginning of the belief accumulation.
+        if (log_odds_map.find(state_key) == log_odds_map.end()) {
+          log_odds_map[state_key] = prior_log_odds_;
```
MasatoSaeki: @toki-1441  @Owen-Liuyuxuan 
This implementation adds `prior_log_odds_` to all log_odds_map which has `state_key` so there is no bias between `state_key`, right? I mean `prior_log_odds_` has no meaning.
And, it is difficult for user to get to have bias between `state_key`. So I thought it might be okay to delete `prior_log_odds_`. What do you think?

MasatoSaeki (commented):
Thank you for fixing!

Just to confirm, the PR description states the following, but it's not “over time,” right?
> To enhance stability and accuracy, we will replace the simple aggregation of confidence scores with a more principled approach: Bayesian updating. This method provides a theoretically sound framework for integrating multiple pieces of evidence over time.

perception/autoware_traffic_light_multi_camera_fusion/src/traffic_light_multi_camera_fusion_node.cpp
Owen-Liuyuxuan: @MasatoSaeki 
The idea is that this makes the output the correct logit of the posterior log-odds.  (Please check the formula in the description)

1. This does not have effect on deciding the final signal.
2. Because it is the correct logit, this could enable further processing if needed.
3. For people with bayesian background, this could be easier to understand instead.

MasatoSaeki: @Owen-Liuyuxuan 
I see. I got it. So [this description](https://github.com/autowarefoundation/autoware_universe/pull/11297/files#diff-2924990215cdcc8f861dfdb5e9e50d60a960d43ffe77f4d800f432039c4ef689R18-R22) is wrong ?

perception/autoware_traffic_light_multi_camera_fusion/src/traffic_light_multi_camera_fusion_node.cpp
Owen-Liuyuxuan: @a-maumau  @toki-1441 
I believe the prior should be added for each measurement from my computation.

Owen-Liuyuxuan: @a-maumau  @toki-1441 
https://www.cs.cmu.edu/~16831-f14/notes/F14/16831_lecture06_agiri_dmcconac_kumarsha_nbhakta.pdf

Please check the material on log-odd probability update for occupancy grid map. We are using a similar mathematic model except that we are integrating over multiple measurement source instead of multiple time-stamp.

image

perception/autoware_traffic_light_multi_camera_fusion/src/traffic_light_multi_camera_fusion_node.cpp
Owen-Liuyuxuan: @MasatoSaeki 
Sorry, I believe they are correct. May I ask which part do you think it is wrong?  -- and yes, I realize they should be added to each element's states instead of only being initialized once.

As the discussion here: https://github.com/autowarefoundation/autoware_universe/pull/11297#discussion_r2354320101

MasatoSaeki: @Owen-Liuyuxuan 
The wrong point I think is a part of below description `positive values bias towards 'Red', and negative values bias towards 'Green'.`
> "The prior log-odds to be added to each traffic light state before fusing evidence. A value of 0.0 means no bias, positive values bias towards 'Red', and negative values bias towards 'Green'."

That is because for all `state_keys`, the same `prior_log_odds_` are added, so there is no bias toward `state_keys`. Therefore, I believe this explanation is wrong.

Owen-Liuyuxuan: @MasatoSaeki 
Ah, I believe you are correct :+1:  , if there is only one uniform value for different shapes/colors, then it essentially means the signal is on or not (or some a uniform probabilistic meaning instead of a color-biased meaning)

perception/autoware_traffic_light_multi_camera_fusion/src/traffic_light_multi_camera_fusion_node.cpp
a-maumau: @Owen-Liuyuxuan Thank you for the correction. Sorry for my lack of study and seems you are right. we need to use the prior term for each update with log form in inverse model.

But, I think we should subtract the prior_log_odds instead of adding it (depending on how we set the prior_log_odds, though):
$$\log{\frac{p(\bar{x})}{p(x)}} = - \log{\frac{p(x)}{p(\bar{x})}},$$
and we still need the $l_0(x) = prior\ log\ odds$, I think.

(Also as Saeki-san mentioned, we need to treat this prior term for each light color)

technolojin:
Updating the base branch may needed to pass `build-test-tidy-pr`

perception/autoware_traffic_light_multi_camera_fusion/src/traffic_light_multi_camera_fusion_node.cpp
toki-1441: thanks for catching that. I apologize, I overlooked updating the comment.

perception/autoware_traffic_light_multi_camera_fusion/src/traffic_light_multi_camera_fusion_node.cpp
toki-1441: @Owen-Liuyuxuan I'm not very familiar with this theory. I apologize. However, after reviewing it, I feel that applying Bayesian estimation theory would be more appropriate than Bayesian updating this time. What do you think?
image

Owen-Liuyuxuan: @toki-1441 
Yes, at this case, you might need to examine the detailed expression for gathering observations from multiple independent input sources. I believe the result will be similar to the shape of the Bayesian update, because the pre-requisite is the same - total expression computation with multiple independent inputs

technolojin:
[cppcheck-differential](https://github.com/autowarefoundation/autoware_universe/actions/runs/17965870697/job/51098302330?pr=11297#logs) indicates 

Screenshot from 2025-09-24 13-05-15

perception/autoware_traffic_light_multi_camera_fusion/src/traffic_light_multi_camera_fusion_node.cpp
toki-1441: @Owen-Liuyuxuan Sorry if I'm missing something, but I think the main point here is how we should handle prior_log_odds_ inside the for loop. I'd appreciate it if you could clarify what the recommended approach is.

MasatoSaeki: @Owen-Liuyuxuan @toki-1441 @a-maumau 

Let me summarize.
**Conclusion**
I think we can proceed with the Bayesian update you proposed this time. However, we will modify it to subtract `prior_log_odds` each time. (I believe what @a-maumau  said is correct.)
Bayesian estimation is also good, but we need to design the likelihood, and if we treat each color as a cell, we can easily apply the formula in this document. (However, it's unclear whether the colors are independent.)
**Checklist**  (to @toki-1441  mainly)
- This is only sensor-to-sensor integration, not temporal integration, correct?
- `prior_log_odds` should be defined for each state, but this PR will use a single definition (as-is). (This initial value is likely unnecessary anyway, and I want to merge this PR for now.)

@toki-1441 
When refactoring this package, conflicts occurred. I am planning to commit my fixes and the remaining PR tasks. Is it okay for me to proceed after getting everyone's input?

technolojin:
@toki-1441 
Passing the CI tests is required.
You can check the code in your local environment.
https://autowarefoundation.github.io/autoware-documentation/main/contributing/pull-request-guidelines/ci-checks/#pre-commit

shmpwk:
@a-maumau @MasatoSaeki @technolojin Could you review this again when you have time?

MasatoSaeki:
@shmpwk  @toki-1441 

I am so sorry for delay. I will check this PR :bow:

MasatoSaeki:
@toki-1441 @shmpwk 

LGTM, thank you for working.

I will approve and merge when finishes the test.

https://evaluation.ci.tier4.jp/evaluation/reports/905abc06-01f1-5366-b6ef-c5c0cef27d42?project_id=prd_jt

a-maumau:
LGTM too.

MasatoSaeki (approved):

PR #11304, opened 2025-09-02
fix(bpp_common, lane_change): generate extended objects directly with highest confidence path.

zulfaqar-azmi-t4:
## Description

When the lane change module filters objects for safety checks, it first converts the object into extended object, then checks whether any of the object's predicted paths cross the target lanes.

Currently all predicted paths are considered, which can lead to false negative for certain objects.

This PR addresses the issue by considering only the highest confidence predicted path. (option use all predicted path still remains).

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


planning/behavior_path_planner/autoware_behavior_path_planner_common/include/autoware/behavior_path_planner_common/utils/path_safety_checker/objects_filtering.hpp
```diff
@@ -227,6 +227,60 @@ std::pair<PredictedObjects, PredictedObjects> separateObjectsByLanelets(
...
+ * @brief Get the predicted path with the highest confidence.
+ *
+ * @tparam PredictedPathType The container type for predicted paths. The elements
+ * of this container must have a public member named `confidence`.
+ * @param predicted_paths The container of predicted paths to search.
+ * @return PredictedPathType A new container holding only the predicted path with
+ * the highest confidence.
+ */
+template <typename PredictedPathType>
+PredictedPathType get_highest_confidence_path(const PredictedPathType & predicted_paths)
```
maxime-clem: There are many situations where there are multiple paths with highest confidence (e.g., 3 paths with `0.333`).
The current function will return only one. Is this okay ?

zulfaqar-azmi-t4: fixed in https://github.com/autowarefoundation/autoware_universe/pull/11304/commits/0f45752f17686bda8f24c85c5179482330866f6e

PR #11306, opened 2025-09-02
fix: use the correct jerk computation formula

Owen-Liuyuxuan:
## Description

Following this PR:

https://github.com/autowarefoundation/autoware_universe/pull/10440

Given acceleration: 

$$a_{lat} = v_{lon}^2 \cdot \kappa$$

Jerk as the acceleration derivatives w.r.t. time: 

$$j_{lat} = v_{lon}^3 \cdot \frac{d\kappa}{ds} + 2 \cdot v_{lon} \cdot a_{lon} \cdot \kappa$$

It does not look like $3 v^2$ in the original formula.

From this formula, the code should be `2 * v_lon * a_lon * curvature`.

Gen-AI gave the same results, but that is not a proof. I derived this equation though.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


kyoichi-sugahara (approved):
Thank you for fixing my mistake :bow: 
Change looks good to me. so approved!

PR #11307, opened 2025-09-02
feat(bpp, lane_change): enable prepare phase check even when ego is preparing

zulfaqar-azmi-t4:
## Description

1. Currently, the safety check is disabled during the prepare phase, except in cases such as when the ego is in turn-direction lanes or intersections. When disabled, the first few seconds of safety checks are skipped.
   - This is usually not an issue if both vehicles are moving. However, if the ego is stopped while the target vehicle is moving, the skipped safety checks can create a potential near-miss situation, since the initial seconds are incorrectly considered “safe.”
   - This PR fixes the issue by enabling the safety check during the prepare phase.

3. Previously, the prepare phase reused the same RSS parameters as the lane change phase. This made lane changes difficult to execute, as the extended ego polygon (both laterally and longitudinally) frequently overlapped during preparation.
   - In this PR, the RSS parameters for the prepare phase are explicitly defined, allowing parameter tuning and correct calculation of the lateral gap.

Launcher's PR: https://github.com/autowarefoundation/autoware_launch/pull/1631

## Related links

**Parent Issue:**

[TIER IV Internal link - Issue's link](https://tier4.atlassian.net/browse/RT0-39121)

### Before PR

Notice that the target vehicle with **green** bounding box is considered safe, even when ego is close by. 

screenshot-20250902-073549Z

https://github.com/user-attachments/assets/bb2223ef-d8fc-4e63-b53f-1243c176fb8c

### After PR 

Object remains unsafe 

image

https://github.com/user-attachments/assets/288a901e-6179-4d8b-8a7a-43231f408a03

### 2. Internal evaluator

1. [TIER IV Internal Link - ControlDLR Test](https://evaluation.tier4.jp/evaluation/reports/e96e6c26-a63d-54fd-9faa-1242a13b4f0f?project_id=prd_jt)
2. [TIER IV Internal Link - FuncVerification Test](https://evaluation.tier4.jp/evaluation/reports/da8f7335-92ec-5101-bb27-2564fd26d540?project_id=prd_jt)
3. [TIER IV Internal Link - CommonScenario_Basic Test](https://evaluation.tier4.jp/evaluation/reports/17486faa-5ed2-5e95-82a3-855a096e4edf?project_id=prd_jt) 

## Notes for reviewers

None.

## Interface changes

None.

#### Additions and removals

| Removed | `enable_for_prepare_phase.general_lanes`   | `bool` | `false`         | Param description |
| Removed | `enable_for_prepare_phase.intersection`   | `bool` | `true`         | Param description |
| Removed | `enable_for_prepare_phase.turns`   | `bool` | `true`         | Param description |

| Added | `safety_check.prepare.expected_front_deceleration`   | `double` | `-1.0`         | The front object's maximum deceleration when the front vehicle perform sudden braking. |
| Added | `safety_check.prepare.expected_rear_deceleration`   | `double` | `-1.0`         | The rear object's maximum deceleration when the rear vehicle perform sudden braking. |
| Added | `safety_check.prepare.rear_vehicle_reaction_time`   | `double` | `1.0`         | The reaction time of the rear vehicle driver which starts from the driver noticing the sudden braking of the front vehicle until the driver step on the brake. |
| Added | `safety_check.prepare.rear_vehicle_safety_time_margin`   | `double` | `0.8`         | The time buffer for the rear vehicle to come into complete stop when its driver perform sudden braking. |
| Added | `safety_check.prepare.lateral_distance_max_threshold`   | `double` | `0.5`         | The lateral distance threshold that is used to determine whether lateral distance between two object is enough and whether lane change is safe. |
| Added | `safety_check.prepare.longitudinal_distance_min_threshold`   | `double` | `1.0`         | The longitudinal distance threshold that is used to determine whether longitudinal distance between two object is enough and whether lane change is safe. |
| Added | `safety_check.prepare.longitudinal_velocity_delta_time`   | `double` | `0.0`         | The time multiplier that is used to compute the actual gap between vehicle at each predicted points (not RSS distance) |
| Added | `safety_check.prepare.extended_polygon_policy`   | `string` | `along-path`         | Policy used to determine the polygon shape for the safety check. Available options are: `rectangle` or `along-path`. |

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/scene.cpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/scene.cpp
@@ -1735,18 +1735,10 @@ PathSafetyStatus NormalLaneChange::isLaneChangePathSafe(
   constexpr auto is_safe = true;
   constexpr auto is_object_behind_ego = true;
 
-  const auto is_check_prepare_phase = check_prepare_phase();
-
   const auto all_paths_collide = [&](const auto & objects) {
-    const auto stopped_obj_vel_th = lane_change_parameters_->safety.th_stopped_object_velocity;
     return ranges::all_of(ego_predicted_paths, [&](const auto & ego_predicted_path) {
       const auto check_for_collision = [&](const auto & object) {
-        const auto selected_rss_param = (object.initial_twist.linear.x <= stopped_obj_vel_th)
-                                          ? lane_change_parameters_->safety.rss_params_for_parked
-                                          : rss_params;
-        return is_colliding(
-          lane_change_path, object, ego_predicted_path, selected_rss_param, is_check_prepare_phase,
-          debug_data);
+        return is_colliding(lane_change_path, object, ego_predicted_path, rss_params, debug_data);
       };
       return ranges::any_of(objects, check_for_collision);
     });
@@ -1765,8 +1757,7 @@ PathSafetyStatus NormalLaneChange::isLaneChangePathSafe(
 
 bool NormalLaneChange::is_colliding(
   const LaneChangePath & lane_change_path, const ExtendedPredictedObject & obj,
-  const std::vector<PoseWithVelocityStamped> & ego_predicted_path,
-  const RSSparams & selected_rss_param, const bool check_prepare_phase,
+  const std::vector<PoseWithVelocityStamped> & ego_predicted_path, const RSSparams & rss_param,
   CollisionCheckDebugMap & debug_data) const
 {
   constexpr auto is_colliding{true};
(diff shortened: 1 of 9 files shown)


planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/scene.cpp
```diff
@@ -1790,31 +1781,44 @@ bool NormalLaneChange::is_colliding(
...
-      });
-
-    const auto collided_polygons = utils::path_safety_checker::get_collided_polygons(
-      lane_change_path.path, ego_predicted_path, obj, predicted_obj_path, bpp_param.vehicle_info,
-      selected_rss_param, hysteresis_factor, safety_check_max_vel,
-      common_data_ptr_->lc_param_ptr->safety.collision_check.th_yaw_diff,
-      current_debug_data.second);
+      CollisionCheckDebug * debug_ptr = collided_polygons.empty() ? &debug : nullptr;
+      const auto & selected_rss_param =
+        obj_pose_with_poly.time < prepare_duration
```
maxime-clem, line 1807: CPPCheck found that this condition cannot be true since we `continue` at the beginning of the loop if it is true.

planning/behavior_path_planner/autoware_behavior_path_planner_common/src/utils/path_safety_checker/safety_check.cpp
```diff
@@ -594,100 +720,13 @@ std::vector<Polygon2d> get_collided_polygons(
...
-        debug.expected_ego_pose = ego_pose;
-        debug.expected_obj_pose = obj_pose;
-        debug.extended_ego_polygon = ego_polygon;
-        debug.extended_obj_polygon = obj_polygon;
-      }
-      collided_polygons.push_back(obj_polygon);
-
-      continue;
-    }
+    CollisionCheckDebug * debug_ptr = collided_polygons.empty() ? &debug : nullptr;
```
maxime-clem, line 723: (optional) We will only have debug information for one calculation (the last one that started before a collision was found).
Is this the intended behavior ? How about storing each debug information in a vector instead ?

maxime-clem (changes requested):

planning/behavior_path_planner/autoware_behavior_path_planner_common/src/utils/path_safety_checker/safety_check.cpp
zulfaqar-azmi-t4: Yes, it is intended. We want the first unsafe information. Storing each debug information is unnecessary as it will make it difficult to associate which one that is responsible for unsafe or cancel event.

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/scene.cpp
zulfaqar-azmi-t4: done in https://github.com/autowarefoundation/autoware_universe/pull/11307/commits/5c0170f5792327b41abfb0824e31e6a817544c0f

planning/behavior_path_planner/autoware_behavior_path_planner_common/src/utils/path_safety_checker/safety_check.cpp
maxime-clem: Now I understand thank you.
Since this is not so obvious from the code, maybe some comment can be added ?

PR #11309, opened 2025-09-02
feat(shape_estimation): correcting cluster yaw

emuemuJP:
## Description
- Correcting cluster yaw angle estimated with fitLShape function when a cluster exists in  front of or behind the own behicle

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

test with the following rosbag

## Notes for reviewers

None.

## Interface changes

None.

<!-- ⬇️🔴

### Topic changes

None.

## Effects on system behavior

None.


perception/autoware_shape_estimation/lib/corrector/utils.cpp
```diff
@@ -69,178 +67,99 @@ bool correctWithDefaultValue(
...
+      correction_vector = v_point.at(second_most_distant_index);
+      applyCorrectionVector(correction_vector, param.default_length);
+    }
+    // Case 5: Only second point is within width range
+    else if (second_in_width_range) {
+      correction_vector = v_point.at(first_most_distant_index);
+      applyCorrectionVector(correction_vector, param.default_length);
+    }
+    // Case 6: First point is within length range, second point fits width constraint
+    else if (first_in_length_range && (second_point_distance < param.max_width)) {
```
Owen-Liuyuxuan: Please check the cppcheck-differential

perception/autoware_shape_estimation/lib/model/bounding_box.cpp
```diff
@@ -159,6 +166,28 @@ bool BoundingBoxShapeModel::fitLShape(
...
+    if ((M_PI_2 - YAW_DEG_TOLERANCE) < raw_yaw && raw_yaw <= M_PI_2) {
+      raw_yaw -= M_PI_2;
+      double temp = shape_output.dimensions.x;
+      shape_output.dimensions.x = shape_output.dimensions.y;
+      shape_output.dimensions.y = temp;
+    }
+  } else if ((-M_PI + FRONT_BACK_ANGLE_THRESHOLD) > cluster_angle_rad || 
+              cluster_angle_rad >= (M_PI - FRONT_BACK_ANGLE_THRESHOLD)) { //back
+    if ((M_PI_2 - YAW_TOLERANCE) < raw_yaw && raw_yaw <= M_PI_2) {
+      raw_yaw -= M_PI_2;
```
Owen-Liuyuxuan: please check the cppcheck-differential

emuemuJP: I fixed the problem

perception/autoware_shape_estimation/lib/corrector/utils.cpp
emuemuJP: I can't see the problem.

emuemuJP:
@Owen-Liuyuxuan Is there any rosbag to check that problem is happening or not.

perception/autoware_shape_estimation/lib/model/bounding_box.cpp
```diff
@@ -159,6 +166,28 @@ bool BoundingBoxShapeModel::fitLShape(
   pose_output.position.x = (intersection_x_1 + intersection_x_2) * 0.5;
   pose_output.position.y = (intersection_y_1 + intersection_y_2) * 0.5;
   pose_output.position.z = min_z + shape_output.dimensions.z * 0.5;
+  double cluster_angle_rad = std::atan2(pose_output.position.y, pose_output.position.x);
+
+  if (-FRONT_BACK_ANGLE_THRESHOLD <= cluster_angle_rad || 
+      cluster_angle_rad < FRONT_BACK_ANGLE_THRESHOLD) { // front
+    if ((M_PI_2 - YAW_DEG_TOLERANCE) < raw_yaw && raw_yaw <= M_PI_2) {
```
Owen-Liuyuxuan: `YAW_DEG_TOLERANCE` Pleasefix this variable

perception/autoware_shape_estimation/lib/model/bounding_box.cpp
Owen-Liuyuxuan: `YAW_TOLERANCE` variable does not seem to be available.

perception/autoware_shape_estimation/lib/model/bounding_box.cpp
```diff
@@ -159,6 +166,28 @@ bool BoundingBoxShapeModel::fitLShape(
...
+      raw_yaw -= M_PI_2;
+      double temp = shape_output.dimensions.x;
+      shape_output.dimensions.x = shape_output.dimensions.y;
+      shape_output.dimensions.y = temp;
+    }
+  } else if ((-M_PI + FRONT_BACK_ANGLE_THRESHOLD) > cluster_angle_rad || 
+              cluster_angle_rad >= (M_PI - FRONT_BACK_ANGLE_THRESHOLD)) { //back
```
Owen-Liuyuxuan: I am wondering if 
- we need to rectify for the back objects.
- we need to rectify for only vehicle-like objects perhaps

perception/autoware_shape_estimation/lib/corrector/utils.cpp
```diff
@@ -69,178 +67,99 @@ bool correctWithDefaultValue(
...
+  // Calculate distances for clearer condition checking
+  double first_point_distance = (v_point.at(first_most_distant_index) * 2.0).norm();
+  double second_point_distance = (v_point.at(second_most_distant_index) * 2.0).norm();
+  double third_point_distance = (v_point.at(third_most_distant_index) * 2.0).norm();
+  // Check if points are within parameter ranges
+  bool first_in_width_range = (first_point_distance < param.max_width);
+  bool first_in_length_range = (param.min_length < first_point_distance && first_point_distance < param.max_length);
+  bool second_in_width_range = (second_point_distance < param.max_width);
+  bool second_in_length_range = (param.min_length < second_point_distance && second_point_distance < param.max_length);
+  bool third_in_width_range = (third_point_distance < param.max_width);
```
Owen-Liuyuxuan: Thanks for the refactoring! That is very helpful.

perception/autoware_shape_estimation/lib/model/bounding_box.cpp
emuemuJP: I fixed this variable.

perception/autoware_shape_estimation/lib/model/bounding_box.cpp
emuemuJP: > we need to rectify for the back objects.

I thought correction might be needed since the object behind could also overtake, but what do you think?

> we need to rectify for only vehicle-like objects perhaps

The corrector parameters are called and adjusted according to each pre-estimated class.

- https://github.com/arayabrain/autoware.universe.origin/blob/bf8bb30bb41e64cdd5c97a61d878ca9a493e5da9/perception/autoware_shape_estimation/lib/shape_estimator.cpp#L116

Therefore, if the vehicle is not present, NoCorrector should be called and no correction should be performed.

- https://github.com/arayabrain/autoware.universe.origin/blob/feat/shape_estimation_correcting_yaw/perception/autoware_shape_estimation/include/autoware/shape_estimation/corrector/no_corrector.hpp

technolojin:
@emuemuJP 

Suggestion:

Can we separate the refactoring and the logic change?

- refactoring is always welcomed
- for the logic change, it need to tested and verified considering under various situations, so it may takes a lot of time

For the logic change, can you provide any visual comparison before and after this PR? 
Example: https://github.com/autowarefoundation/autoware_universe/pull/11230

perception/autoware_shape_estimation/lib/corrector/utils.cpp
```diff
@@ -69,178 +67,99 @@ bool correctWithDefaultValue(
...
+
+  // Check if first and second most distant points are on opposite edges (0,2 pair or 1,3 pair)
+  bool are_opposite_edges = (static_cast<int>(std::abs(
+    static_cast<int>(first_most_distant_index) - static_cast<int>(second_most_distant_index))) % 2 == 0);
+  // Calculate distances for clearer condition checking
+  double first_point_distance = (v_point.at(first_most_distant_index) * 2.0).norm();
+  double second_point_distance = (v_point.at(second_most_distant_index) * 2.0).norm();
+  double third_point_distance = (v_point.at(third_most_distant_index) * 2.0).norm();
+  // Check if points are within parameter ranges
+  bool first_in_width_range = (first_point_distance < param.max_width);
```
a-maumau: I might misunderstanding something here, but why we don't need to check the param.min_width for `*_in_width_range` part?  
In the original implementation, it was checking like `param.min_width < (...).norm() && (...).norm() < param.max_width`.

perception/autoware_shape_estimation/lib/model/bounding_box.cpp
```diff
@@ -159,6 +166,28 @@ bool BoundingBoxShapeModel::fitLShape(
   pose_output.position.x = (intersection_x_1 + intersection_x_2) * 0.5;
   pose_output.position.y = (intersection_y_1 + intersection_y_2) * 0.5;
   pose_output.position.z = min_z + shape_output.dimensions.z * 0.5;
+  double cluster_angle_rad = std::atan2(pose_output.position.y, pose_output.position.x);
+
+  if (-FRONT_BACK_ANGLE_THRESHOLD <= cluster_angle_rad || 
+      cluster_angle_rad < FRONT_BACK_ANGLE_THRESHOLD) { // front
```
a-maumau: I think this is a bug.  
Since `||` is used here, all the conditions will go in this branch.
`&&` should be used.

a-maumau (commented):
@emuemuJP  Hi, thanks for the refactoring! I also wanted to do something for this package.
I have some points to point out.

shmpwk:
@emuemuJP Could you add your great idea on README?

perception/autoware_shape_estimation/lib/model/bounding_box.cpp
a-maumau: fixed at [f022e2b](https://github.com/autowarefoundation/autoware_universe/pull/11309/commits/f022e2b07497b9e0cae51ed23b8682b0754aa15e)

emuemuJP:
@technolojin @a-maumau @shmpwk 
Sorry, after retesting, I found that the only effective improvement for ShapeEstimation was removing the param.min_width condition.
When the point cloud shape is small, if one edge is within {min_width, max_width} and another is below min_width,
it falls under one of the following conditions:
https://github.com/arayabrain/autoware.universe.origin/blob/83ef38195526121b2d6f58b46425c4b739d41015/perception/autoware_shape_estimation/lib/corrector/utils.cpp#L180-L210
This causes the longer-appearing edge to be treated as the width. 

Therefore, by removing the min condition under the following conditions,
https://github.com/arayabrain/autoware.universe.origin/blob/f022e2b07497b9e0cae51ed23b8682b0754aa15e/perception/autoware_shape_estimation/lib/corrector/utils.cpp#L161-L178
it will extend along the x-axis.

I confirmed the before and after as follows:

 **Before** 
Screenshot from 2025-09-10 21-30-49

[shape_estimation_before_10s_mini.webm](https://github.com/user-attachments/assets/3de6f2f6-2662-4a9f-bbc7-04532ddd6b51)

**After**
Screenshot from 2025-09-10 21-18-05

[shape_estimation_after_10s_mini.webm](https://github.com/user-attachments/assets/8aab7eff-70a9-4a6e-9fc9-5460e5ffb764)

perception/autoware_shape_estimation/lib/corrector/utils.cpp
a-maumau: > @technolojin @a-maumau @shmpwk Sorry, after retesting, I found that the only effective improvement for ShapeEstimation was removing the param.min_width condition. When the point cloud shape is small, if one edge is within {min_width, max_width} and another is below min_width, it falls under one of the following conditions: https://github.com/arayabrain/autoware.universe.origin/blob/83ef38195526121b2d6f58b46425c4b739d41015/perception/autoware_shape_estimation/lib/corrector/utils.cpp#L180-L210 This causes the longer-appearing edge to be treated as the width.
> 
> Therefore, by removing the min condition under the following conditions, https://github.com/arayabrain/autoware.universe.origin/blob/f022e2b07497b9e0cae51ed23b8682b0754aa15e/perception/autoware_shape_estimation/lib/corrector/utils.cpp#L161-L178 it will extend along the x-axis.
> 
> I confirmed the before and after as follows:
> 
> **Before** Screenshot from 2025-09-10 21-30-49 [shape_estimation_before_10s_mini.webm](https://github.com/user-attachments/assets/3de6f2f6-2662-4a9f-bbc7-04532ddd6b51)
> 
> **After** Screenshot from 2025-09-10 21-18-05 [shape_estimation_after_10s_mini.webm](https://github.com/user-attachments/assets/8aab7eff-70a9-4a6e-9fc9-5460e5ffb764)

a-maumau:
@emuemuJP I am not sure whether we need the `min_width` or not, but removing the `param.min_width` should be almost equal to set `param.min_width` to 0 in each `*_corrector.hpp`. Do you think it is worth to leave it?

If we don't need it, we also need delete the min_width param in `*_corrector.hpp`.

FYI, I tested on some rosbag, but in some cases, it seems it still shows the wrong correction.  
May be this is not related to your PR.
wrong_correction

a-maumau:
Also, if you create the refactor PR, could you link it here too?

kaancolak:
I believe that without prior information, it’s not possible to handle all cases in rule-based shape estimation. Simply checking the outside dimensions does not provide enough information to decide which side is the width and which side is the length, especially due to occlusions / FOV of LIDAR where width and length can be mixed up. In some cases, a small one could be the length and a big one could be the width

At first glance, I considered proposing a basic idea as a maybe possible solution. It would be easy to implement but requires a quite big changes:

1- Add the camera ID to the DetectedObjectsWithFeature->Feature message.
2- Inside the corrector function, to decide which side is width and which is length, project the 8 corners of bounding box for both possible width–length and choose the pair with the best IoU.

emuemuJP:
I thought it would be acceptable to retain the reference values for the expected size of each vehicle.
I believe there is potential to further refine this algorithm in the future.
I'll include the refactoring link below.
https://github.com/autowarefoundation/autoware_universe/pull/11353

After discussing with @Owen-Liuyuxuan , the current logic changes could lead to some conditions being missed. 
Therefore, I want to add the changes as an extension to the existing logic.
Additionally, while the current logic considers the positional relationship with vehicles, it does not account for angles. I want to add such logic to improve it.

a-maumau:
@emuemuJP 
> I thought it would be acceptable to retain the reference values for the expected size of each vehicle.

So, Are you going to revert the change and set the default parameter to 0?

emuemuJP:
@Owen-Liuyuxuan @shmpwk @a-maumau 
I've organized the relevance of the conditions for the correctWithDefaultValue function here.
https://github.com/arayabrain/autoware.universe.origin/blob/180332a4957e9434d7a4723a8957b7651ac0edde/perception/autoware_shape_estimation/resource/correctWithDefaultValue.svg

The primary issue causing rotation occurs within the range indicated by the red box. Within this range, when the visible point clouds are extremely small, the farthest edge is frequently swapped, causing the edge being corrected to change often.

**Before**
image

Therefore, I unified the behavior so that when the number of point clouds is extremely low, the correction is directed toward the farthest edge while maintaining the current conditions as much as possible, as follows.

**After**
image

image

[shape_estimation_after_mini.webm](https://github.com/user-attachments/assets/31cd6b23-a794-4262-8c9c-6fa158b3705b)

emuemuJP:
@shmpwk @Owen-Liuyuxuan Do you have any comments?

Owen-Liuyuxuan:
@emuemuJP 
Please update your branch with the latest main so that we can run build test.
And also help fix the DCO problem

Please @ me when you finished you work.

emuemuJP:
@Owen-Liuyuxuan 
I fixed the DCO problem.
Please check.

a-maumau:
@emuemuJP Sorry for the late reply.
Is this PR meld in to the https://github.com/autowarefoundation/autoware_universe/pull/11353 ?

Owen-Liuyuxuan:
@a-maumau 
This PR is not melt into #11353

a-maumau:
since the successor PR https://github.com/autowarefoundation/autoware_universe/pull/11634 has been merged, I will close this PR.

PR #11311, opened 2025-09-03
fix(start_planner): check collision from start_pose

kyoichi-sugahara:
## Description

This PR refactors the creation of `PullOutPath` object in the clothoid pull-out planner to use aggregate initialization and fixes a safety-related issue in collision detection logic.

- Fixed collision detection logic that was causing unnecessary path generation failures
  - Previously, collision checking was performed on sections behind the start_pose, which sometimes prevented path generation even when it was safe to proceed
  - The empty check was deemed unnecessary as this validation is already performed upstream in the processing pipeline

**Comparison:**

| Before                        | After                        |
| ----------------------------- | ---------------------------- |
| Screenshot from 2025-09-03 10-51-55 | Screenshot from 2025-09-03 10-52-51 |

## Related links

**Private Links:**
- [TIERIV internal link](https://tier4.atlassian.net/browse/RT0-38846)

## How was this PR tested?

- [x] run psim

skip scenarios evaluation because default setting doesn't enable clothoid planner

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


planning/behavior_path_planner/autoware_behavior_path_start_planner_module/src/clothoid_pull_out.cpp
```diff
@@ -1422,11 +1422,7 @@ std::optional<PullOutPath> ClothoidPullOut::plan(
     // STEP 5-7: Collision check
     // ===================================================================
     // Create PullOutPath for collision check
-    PullOutPath temp_pull_out_path;
-    temp_pull_out_path.partial_paths.push_back(clothoid_path);
-    temp_pull_out_path.start_pose =
-      clothoid_path.points.empty() ? start_pose : clothoid_path.points.front().point.pose;
-    temp_pull_out_path.end_pose = target_pose;
+    PullOutPath temp_pull_out_path{{clothoid_path}, {}, start_pose, target_pose};
```
kosuke55:
```suggestion
    const PullOutPath temp_pull_out_path{{clothoid_path}, {}, start_pose, target_pose};
```
?

kyoichi-sugahara: addressed in https://github.com/autowarefoundation/autoware_universe/pull/11311/commits/acb436b4a38600e9ac33b3a3fff531b294c934c7 :pray:

kosuke55 (approved):

PR #11317, opened 2025-09-03
fix(autoware_pointcloud_preprocessor): update transform frame for cor…

mertyavuz41:
…rect functionality

## Description
The cropbox filter doesn't filter the self points properly.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

Before;
issue_before

After;
issue_after

None.


mertyavuz41:
I just notice i should leave the output_frame parameter blank.

mertyavuz41:
I misinterpreted input_frame and output_frame

PR #11327, opened 2025-09-05
feat(diagnostic_graph_utils): use raw level for leaf node

isamu-takagi:
## Description

The graph aggregator treats STALE as an ERROR in intermediate nodes. So use raw level instead when logging.

## Related links

**Private Links:**

- [TIER IV internal link](https://star4.slack.com/archives/CH6FSPRM1/p1757034761650159)

## How was this PR tested?

Launch planning simulator and check that the error tree on terminal contains STALE level (at the immediately after startup).

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


h-ohta (approved):
I checked with planning simulator in my environment and it looks good.

sasakisasaki:
@isamu-takagi Thank you for doing refactoring! As I have read the PR description and changed code but could not understood, could you kindly provide steps to verify if the applied change is working? :pray:

PR #11330, opened 2025-09-08
refactor(diffusion_planner): fix to use `lane_segments_`

SakodaShintaro:
## Description

This PR refactors the diffusion planner to use the `lane_segments_` member variable directly instead of extracting data from the matrix representation, improving code clarity and reducing redundant data storage.

- Store lane segments as a member variable and access segment properties directly from the original data structure
- Remove redundant storage of speed limit, lane ID, and turn direction in the matrix representation
- Update matrix dimension constants to reflect the reduced data storage requirements

## How was this PR tested?

Run `autoware_diffusion_planner`

### Processing Time

* Before
  * create_input_data        1.0 ± 0.4 msec
* After
  * create_input_data        1.0 ± 0.5 msec

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/autoware_diffusion_planner/src/preprocessing/lane_segments.cpp
+++ b/planning/autoware_diffusion_planner/src/preprocessing/lane_segments.cpp
@@ -91,13 +91,14 @@ std::pair<std::vector<float>, std::vector<float>> LaneSegmentContext::get_route_
       0, added_route_segments * POINTS_PER_SEGMENT, SEGMENT_POINT_DIM, POINTS_PER_SEGMENT) =
       map_lane_segments_matrix_.block(0, row_idx, SEGMENT_POINT_DIM, POINTS_PER_SEGMENT);
 
-    const int64_t turn_direction = lane_segments_[row_idx / POINTS_PER_SEGMENT].turn_direction;
+    const int64_t seg_idx = row_idx / POINTS_PER_SEGMENT;
+    const int64_t turn_direction = lane_segments_[seg_idx].turn_direction;
     add_traffic_light_one_hot_encoding_to_segment(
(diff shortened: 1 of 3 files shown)


planning/autoware_diffusion_planner/src/preprocessing/lane_segments.cpp
```diff
@@ -92,12 +91,13 @@ std::pair<std::vector<float>, std::vector<float>> LaneSegmentContext::get_route_
...
 
-    const int64_t turn_direction = map_lane_segments_matrix_(TURN_DIRECTION, row_idx);
+    const int64_t turn_direction = lane_segments_[row_idx / POINTS_PER_SEGMENT].turn_direction;
     add_traffic_light_one_hot_encoding_to_segment(
       traffic_light_id_map, full_route_segment_matrix, row_idx, added_route_segments,
       turn_direction);
 
-    speed_limit_vector[added_route_segments] = map_lane_segments_matrix_(SPEED_LIMIT, row_idx);
+    speed_limit_vector[added_route_segments] =
+      lane_segments_[row_idx / POINTS_PER_SEGMENT].speed_limit_mps.value_or(0.0f);
```
go-sakayori: nits: I would rather store it in a variable than calculating `row_idx / POINTS_PER_SEGMENT` twice.

go-sakayori (approved):
One small comment, but LGTM

planning/autoware_diffusion_planner/src/preprocessing/lane_segments.cpp
SakodaShintaro: https://github.com/autowarefoundation/autoware_universe/pull/11330/commits/7e850037dc158b23602d69d3e02ac758768632d5

PR #11350, opened 2025-09-11
fix(start_planner): increase lateral gap to furthest bound when checking target object passing

zulfaqar-azmi-t4:
## Description

The system calculates the distance from the right edge of the vehicle's footprint to the right boundary line of the lane, calculates whether there is room for the vehicle behind to pass, and if there is no room for the vehicle to pass, skips the safety judgment.

In this scene, it is determined that there is not enough space for the vehicle behind to pass, so it starts moving.

image

The corresponding policy is to calculate the distance of the red line below,

**
image
**

If the calculation is changed to calculate the distance shown by the red line below, that is, the distance to the rightmost boundary in the same direction of travel, the vehicle will be able to stop.

image

## Related links

**Parent Issue:**

- [TIER IV Internal Link - Issue's ticket](https://tier4.atlassian.net/browse/RT0-39052) 

#### Before

https://github.com/user-attachments/assets/d8cf5872-2b4b-4702-a85f-855e328dd72b

#### After

https://github.com/user-attachments/assets/9db5735f-9769-4f34-b47d-063ce7d6dcf9

### Internal Degradation check

[TIER IV Internal Link - ControlDLR Test](https://evaluation.tier4.jp/evaluation/reports/e4a54d33-78e3-577e-9bb3-18046373e465?project_id=prd_jt)
[TIER IV Internal Link - FuncVerification Test](https://evaluation.tier4.jp/evaluation/reports/c0bfc3ef-9213-561d-8466-375a595d9ba4?project_id=prd_jt)
[TIER IV Internal Link - CommonScenario_Basic Test](https://evaluation.tier4.jp/evaluation/reports/0b2bf6ff-6c70-59e6-96a4-568d5f746dc5?project_id=prd_jt)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_path_planner/autoware_behavior_path_start_planner_module/src/start_planner_module.cpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_start_planner_module/src/start_planner_module.cpp
@@ -654,7 +654,7 @@ StartPlannerModule::getGapBetweenEgoAndLaneBorder(
     point_pose.position = autoware_utils::to_msg(point.to_3d());
     point_pose.orientation = ego_pose.orientation;
 
-    lanelet::ConstLanelets closest_lanelet;
+    lanelet::ConstLanelet closest_lanelet;
     if (!lanelet::utils::query::getClosestLanelet(target_lanes, point_pose, &closest_lanelet)) {
       return std::nullopt;
     }


planning/behavior_path_planner/autoware_behavior_path_start_planner_module/src/start_planner_module.cpp
```diff
@@ -541,83 +569,35 @@ bool StartPlannerModule::isPreventingRearVehicleFromPassingThrough(const Pose &
...
-    if (!is_closest_lanelet) return std::nullopt;
-    lanelet::ConstLanelet closest_lanelet_const(closest_lanelet.constData());
+    lanelet::ConstLanelet closest_lanelet;
+    if (!lanelet::utils::query::getClosestLanelet(
+          target_lanes, ego_overhang_point_as_pose, &closest_lanelet))
+      return std::nullopt;
     // Check backwards just in case the Vehicle behind ego is in a different lanelet
-    constexpr double backwards_length = 200.0;
-    const auto prev_lanes = autoware::behavior_path_planner::utils::getBackwardLanelets(
+    constexpr auto backwards_length = 200.0;
```
kyoichi-sugahara:
```suggestion
    constexpr double backwards_length = 200.0;
```

planning/behavior_path_planner/autoware_behavior_path_start_planner_module/src/start_planner_module.cpp
```diff
@@ -655,6 +635,72 @@ bool StartPlannerModule::isPreventingRearVehicleFromPassingThrough(const Pose &
...
+  std::optional<double> smallest_lateral_gap_between_ego_and_farthest_border;
+  auto corresponding_lateral_gap_with_other_lane_bound = std::numeric_limits<double>::max();
+
+  const auto ego_is_merging_from_the_left = (starting_pose_lateral_offset > 0.0);
+  for (const auto & point : vehicle_footprint) {
+    geometry_msgs::msg::Pose point_pose;
+    point_pose.position = autoware_utils::to_msg(point.to_3d());
+    point_pose.orientation = ego_pose.orientation;
+
+    lanelet::Lanelet closest_lanelet;
```
kyoichi-sugahara:
```suggestion
    lanelet::ConstLanelets closest_lanelet;
```

PR #11358, opened 2025-09-12
fix(out_of_lane): properly apply the objects.extra_width parameter

maxime-clem:
## Description

This PR fixes the way the `objects.extra_width` parameter is applied. Previously the feature was not working.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


satoshi-ota (approved):
LGTM

PR #11359, opened 2025-09-12
fix(tensorrt_yolox): parameterize models

badai-nguyen:
## Description
- To make it's easier to launch multi-yolox model together with `logging_simulator` with customized `model_name` and `model_dir`

## Related links

**Parent Issue:**

- [TIER IV Internal Link](https://star4.slack.com/archives/CRUE57C30/p1757662245863059?thread_ts=1757639076.334349&cid=CRUE57C30)

## How was this PR tested?

- Tesing by [internal evaluator](https://evaluation.tier4.jp/evaluation/reports/ee302b82-1635-54b6-a7b9-2bf9d64c1e01?project_id=x2_dev) 

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_tensorrt_yolox/launch/yolox_s_plus_opt.launch.xml
+++ b/perception/autoware_tensorrt_yolox/launch/yolox_s_plus_opt.launch.xml
@@ -4,20 +4,20 @@
   <arg name="yolox_node_name" default="tensorrt_yolox"/>
   <arg name="image_transport_decompressor_node_name" default="image_transport_decompressor_node"/>
 
-  <arg name="data_path" default="$(env HOME)/autoware_data" description="packages data and artifacts directory path"/>
-
+  <arg name="model_dir" default="$(env HOME)/autoware_data/tensorrt_yolox" description="packages data and artifacts directory path"/>
+  <arg
+    name="model_name"
+    default="yolox-sPlus-opt-pseudoV2-T4-960x960-T4-seg16cls"
+    description="options `yolox-sPlus-T4-960x960-pseudo-finetune.onnx` if only detection is needed, `yolox-sPlus-opt-pseudoV2-T4-960x960-T4-seg16cls.onnx` if sematic segmentation is also needed"
+  />
   <arg name="input/image" default="/sensing/camera/camera0/image_rect_color"/>
   <arg name="output/objects" default="/perception/object_recognition/detection/rois0"/>
   <arg name="output/mask" default="/perception/object_recognition/detection/mask0"/>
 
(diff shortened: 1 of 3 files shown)


perception/autoware_tensorrt_yolox/launch/yolox_s_plus_opt.launch.xml
```diff
@@ -4,20 +4,20 @@
...
-  <arg
-    name="model_path"
-    default="$(var data_path)/tensorrt_yolox/yolox-sPlus-opt-pseudoV2-T4-960x960-T4-seg16cls.onnx"
-    description="options `yolox-sPlus-T4-960x960-pseudo-finetune.onnx` if only detection is needed, `yolox-sPlus-opt-pseudoV2-T4-960x960-T4-seg16cls.onnx` if sematic segmentation is also needed"
-  />
-  <arg name="label_path" default="$(var data_path)/tensorrt_yolox/label.txt" description="label path"/>
-  <arg name="color_map_path" default="$(var data_path)/tensorrt_yolox/semseg_color_map.csv" description="color map path"/>
+  <arg name="model_path" default="$(var model_dir)/$(var model_name).onnx"/>
+  <arg name="label_path" default="$(var model_dir)/label.txt" description="label path"/>
+  <arg name="color_map_path" default="$(var model_dir)/semseg_color_map.csv" description="color map path"/>
```
MasatoSaeki: I think do not need to change the args name. So why do you change it?

badai-nguyen: @MasatoSaeki I rolled back the argument name changes in https://github.com/autowarefoundation/autoware_universe/pull/11359/commits/a1399d8132a5f47cd9f228715cbaf3c7b880405b

MasatoSaeki:
@badai-nguyen 
I apologize for the corrections you made, but how about doing it as follows?
https://github.com/autowarefoundation/autoware_universe/pull/11364

I think it is good to handle with "path" instead of "directory" and "name" because it becomes unnecessarily complicated. (So traffic light recognition defines it that way.)
https://github.com/autowarefoundation/autoware_launch/blob/main/autoware_launch/launch/components/tier4_perception_component.launch.xml#L265-L272

badai-nguyen:
> @badai-nguyen I apologize for the corrections you made, but how about doing it as follows? #11364
> 
> I think it is good to handle with "path" instead of "directory" and "name" because it becomes unnecessarily complicated. (So traffic light recognition defines it that way.) https://github.com/autowarefoundation/autoware_launch/blob/main/autoware_launch/launch/components/tier4_perception_component.launch.xml#L265-L272

@MasatoSaeki 
Thank you for you suggestion, I think your approach looks simpler. I will close this PR. Please open your PR for reviewing.

PR #11362, opened 2025-09-12
ci: use github hosted runner for arm

mitsudome-r:
## Description
This updates the workflows for arm64 to use GitHub hosted runners instead of self-hosted runners since we are no longer using self-hosted runners.

## Related links

**Parent Issue:**

- https://github.com/autowarefoundation/autoware_universe/issues/11334

## How was this PR tested?

https://github.com/autowarefoundation/autoware_universe/actions/runs/17765450327

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/.github/workflows/build-and-test-daily.yaml
+++ b/.github/workflows/build-and-test-daily.yaml
@@ -33,7 +33,7 @@ jobs:
   build-and-test-daily-arm64:
     uses: ./.github/workflows/build-and-test-reusable.yaml
     with:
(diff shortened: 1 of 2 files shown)


nishikawa-masaki (approved):
LGTM.
I confirmed that only the 2 files fixed in this PR need the fix by running:
```
$ cd autoware_universe
$ find . -name '*.yaml' -exec grep ARM64 {} \; -print
```

PR #11363, opened 2025-09-12
ci: use Valkey (Redis-compatible server) as storage backend for ccache. Add Tailscale to access the private Valkey server

esteve:
## Description

This PR adds support for using Valkey (a Redis-compatible implementation) with ccache. Additionally, it uses Tailscale to access the Valkey server, as it's not recommended to expose it to the internet, not even with a password.

## Related links

**Parent Issue:**

- https://github.com/autowarefoundation/autoware_universe/issues/11295

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/.github/workflows/build-and-test-differential.yaml
+++ b/.github/workflows/build-and-test-differential.yaml
@@ -50,7 +50,9 @@ jobs:
   build-and-test-differential:
     if: ${{ inputs.run-condition }}
     runs-on: ${{ fromJson(inputs.runner) }}
-    container: ${{ inputs.container }}${{ inputs.container-suffix }}
+    container:
+      image: ${{ inputs.container }}${{ inputs.container-suffix }}
+      options: --cap-add=NET_ADMIN --device=/dev/net/tun
     steps:
       - name: Tailscale
         uses: tailscale/github-action@v3
--- a/.github/workflows/build-and-test-packages-above-differential.yaml
+++ b/.github/workflows/build-and-test-packages-above-differential.yaml
@@ -48,7 +48,9 @@ jobs:
   build-and-test-packages-above-differential:
     if: ${{ inputs.run-condition }}
     runs-on: ${{ fromJson(inputs.runner) }}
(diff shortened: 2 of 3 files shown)


esteve:
The approach taken in this PR involves the following components:

- Valkey (https://valkey.io/) as a remote backend storage for ccache. Valkey is a Redis-compatible server that offers extra functionality such as multithreaded IO. Valkey started as a fork of Redis when the latter changed its license to a non-opensource license (Redis has reverted to an OSI-approved license since then).
- Tailscale (https://tailscale.com/) as a mechanism for connecting the public GitHub runners with our private server. Tailscale is a VPN service that provides really low latency and flexible workflows for ephermeral clients (e.g. GitHub runners).

The diagram is as follows:

valkey_tailscale drawio

The rationale for using Tailscale is that the developers of Redis themselves advise strongly against exposing Redis (and Valkey) to the internet, even if they are protected by a password (see https://valkey.io/topics/security/ and https://redis.io/docs/latest/operate/oss_and_stack/management/security/).

On our server, Tailscale runs as a Docker container to both provide an extra layer of isolation and as straightforward way to deploy it without installing extra packages. The Docker container for Tailscale is set up as a systemd service, so that systemd can monitor it and restart it when necessary. Additionally, the Tailscale service is not running on the same network as the host, but on a separate Docker network, to provide extra isolation.

Valkey also runs as a Docker container, and instead of exposing its port (6379) to the host, it is running on the same network as Tailscale, so that only VPN clients can actually access Valkey and not any untrusted client. This is achieved by passing the option `--network="container:docker.tailscale.service"`, where `docker.tailscale.service` is the name of the systemd service running the Tailscale container.

On the GitHub side, Tailscale provides a GitHub action (https://tailscale.com/kb/1276/tailscale-github-action) that can be used to connect to a Tailscale network. This action requires either the use of an `auth_key` or OAuth authentication, and the nodes marked as ephemeral so that the GitHub runners' keys are deleted when no longer in use.

esteve:
Closing this PR as we're not any longer exploring the use of Valkey

PR #11369, opened 2025-09-16
feat(pid_longitudinal_controller): don't switch to DRIVE if the state conditions are not met

MertClk:
## Description
Closes https://github.com/autowarefoundation/autoware_universe/issues/11368

This PR aims to prevent the oscillation in the control state when Autoware control is not enabled. Before this PR, control mode switches between DRIVE and STOPPED until Autoware control is enabled.

control_state_oscillation

After this PR, STOPPED state will be kept as long as Autoware control is disabled and stopped state conditions are met.

control_state_oscillation_fix

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


Owen-Liuyuxuan:
image

I agree that this update is a robust update because it uses a full condition in the drive->stopped transition to keep stopped locked.

Please check the image above for the state machine problem.

PR #11371, opened 2025-09-16
feat: gpu ground segmentation

anhnv3991:
## Description
@badai-nguyen created an interesting PR that uses CUDA to accelerate the ground-segmentation but the code is not quite optimized and well-organized. I re-organized his code a little bit and fix some minor bugs. Besides, I added some utility functions for allocating/freeing GPU memory.

## Related links

**Parent Issue:**

- [Dai's draft PR](https://github.com/autowarefoundation/autoware_universe/pull/11272)

## How was this PR tested?
- Same as Dai's PR. 
- Execute the cuda_scan_ground_segmenation_node for example:
` ros2 launch autoware_ground_segmentation_cuda cuda_scan_ground_filter.launch.py input/pointcloud:=/sensing/lidar/concatenated/pointcloud`
- Open Rviz2, add the topics `/cuda_scan_ground_segmentation_filter/output/ground_pointcloud` and `/perception/obstacle_segmentation/pointcloud`. Those publish the ground and non-ground points on rviz. Set the global options Fixed Frame to base_link. Set the Reliability Policy of Point Cloud topics to Best Effort.
- Play rosbag which include `/sensing/lidar/concatenated/pointcloud` topic
A sample t4_dataset whose input_bag with /sensing/lidar/concatenated/pointcloud topic available could be downloaded from [here](https://github.com/tier4/tier4_perception_dataset/blob/main/docs/t4_format_3d_detailed.md#dataset-sample)
- [Download link](https://drive.google.com/file/d/1UjMWZj5Yc55O7BZiGHa0ikZGhwmcfPiS/view)

- Testing on internal evaluator: [eva/pr11371_cuda_segmentation_segm](https://evaluation.tier4.jp/evaluation/reports/62296088-9c1b-5994-8f2b-37a45c05b97e?project_id=x2_dev) 
  - Comparison result on both dataset shows GPU version is a little bit higher than current CPU algorithm. 
  - 2ad1ca45-ec8d-43ee-bcff-491e8434f443 dataset: 
62296088-9c1b-5994-8f2b-37a45c05b97e_0005_ground_segmentation-sample-test_2 pkl z_*_2ad1ca45-ec8d-43ee-bcff-491e8434f443
  - 14be49dc-c5f4-4496-932c-60c83500b825 dataset
62296088-9c1b-5994-8f2b-37a45c05b97e_0005_ground_segmentation-sample-test_2 pkl z_*_14be49dc-c5f4-4496-932c-60c83500b825

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_ground_segmentation_cuda/CMakeLists.txt
+++ b/perception/autoware_ground_segmentation_cuda/CMakeLists.txt
@@ -34,8 +34,6 @@ endif()
 
 if(BUILD_TESTING)
     list(APPEND AMENT_LINT_AUTO_EXCLUDE ament_cmake_uncrustify)
-    find_package(ament_lint_auto REQUIRED)
-    ament_lint_auto_find_test_dependencies()
 endif()
 
 include_directories(


perception/autoware_ground_segmentation_cuda/src/cuda_scan_ground_segmentation/cuda_scan_ground_segmentation_filter.cu
```diff
@@ -0,0 +1,848 @@
...
+  extractPoints<NonGroundChecker>(classified_points, *dev_input_points_, *dev_output_points_);
+
+  // Extract ground points
+  extractPoints<GroundChecker>(classified_points, *dev_input_points_, *dev_ground_points_);
+
+  dev_output_points_->to_point_cloud2(non_ground);
+  non_ground.header.frame_id = "map";
+  dev_ground_points_->to_point_cloud2(ground);
+  ground.header.frame_id = "map";
+  CHECK_CUDA_ERROR(cudaStreamSynchronize(stream_->get()));
```
badai-nguyen: the ground_segmentation algorithm is assumed to process in base_link, why you fix the `frame_id` to `map`?
I think the `header.stamp` assignment is also needed.

```suggestion
  dev_output_points_->to_point_cloud2(non_ground);
  non_ground.header = input.header;
  dev_ground_points_->to_point_cloud2(ground);
  ground.header = input.header;
  CHECK_CUDA_ERROR(cudaStreamSynchronize(stream_->get()));
```

perception/autoware_ground_segmentation_cuda/include/autoware/cuda_scan_ground_segmentation/device_vector.hpp
```diff
@@ -0,0 +1,332 @@
...
+  ~device_vector();
+
+private:
+  // Return false if allocation fail
+  void allocate(size_t ele_num);
+  void release();
+  void copyDtoD(T * dst, const T * src, size_t ele_num);
+  void copyHtoD(T * dst, const T * src, size_t ele_num);
+  void copyDtoH(T * dst, const T * src, size_t ele_num) const;
+
```
badai-nguyen, line 105: since most IMO, it would be better to reuse the existing function from the [cuda_blackboard library](https://github.com/autowarefoundation/cuda_blackboard/blob/842d763e71a6bbe5ab513e04e5b45dc9ec8d2209/include/cuda_blackboard/cuda_adaptation.hpp#L59-L95) instead of re-implementing it here.

badai-nguyen (commented):
@anhnv3991 Thank you very much for your suggestion on my PR and also for your contribution PR. I sincerely apologize for the late response.

I’ve added some comments above, but the main point is that `cuda_blackboard` has already been implemented and tested in sensing’s pointcloud_preprocessor, so there are many functions we can reuse.

IMO, it’s better to take advantage of `cuda_blackboard` and improve the library itself if needed, rather than re-implement it here. This will make maintenance more efficient overall.

perception/autoware_ground_segmentation_cuda/src/cuda_scan_ground_segmentation/cuda_point_cloud2.cu
```diff
@@ -0,0 +1,227 @@
...
+      point_num, offset_x, offset_y, offset_z, (int)(msg.point_step)));
+}
+
+__global__ void convert_to_point_cloud2(
+  PointCloud2::Ptr input, uint8_t * output, int point_num, int offset_x, int offset_y, int offset_z,
+  int point_step)
+{
+  int index = threadIdx.x + blockIdx.x * blockDim.x;
+  int stride = blockDim.x * gridDim.x;
+  PointXYZ p;
```
badai-nguyen: it seems you fixed the output pointcloud type to pointXYZ? 
since all pointcloud used in perception pipeline is followed the pointcloud type of sensing's pointcloud_preprocessor as described [here](https://autowarefoundation.github.io/autoware-documentation/main/design/autoware-architecture/sensing/data-types/point-cloud/), so do you have any reason for this change?

.cspell.json
```diff
@@ -4,5 +4,5 @@
     "planning/behavior_velocity_planner/autoware_behavior_velocity_intersection_module/scripts/**"
   ],
   "ignoreRegExpList": [],
-  "words": ["dltype", "tvmgen", "fromarray", "soblin", "brkay54", "libtensorrt", "TRTBEV"]
+  "words": ["dltype", "tvmgen", "fromarray", "soblin", "brkay54", "libtensorrt", "TRTBEV", "CUDAH"]
```
Owen-Liuyuxuan: Question:
I am not an engineering expert in CUDA coding, but I am curious about why you need to define this `CUDAH` name?

This name is not a good name, and it is only defined in the new codes, where we can probably find a better name for it.

perception/autoware_ground_segmentation_cuda/src/cuda_scan_ground_segmentation/cuda_scan_ground_segmentation_filter.cu
anhnv3991: @badai-nguyen Sorry for my late response.
This was fixed. The header part is now copied from the input message by a function called copyPointCloud2Metadata

perception/autoware_ground_segmentation_cuda/include/autoware/cuda_scan_ground_segmentation/device_vector.hpp
anhnv3991: @badai-nguyen The cuda blackboard library you refered to use a unique_ptr to hold a GPU buffer. It has some drawbacks
- Simple functions like size(), resize(), clear(), etc are not provided. 
- Data copy between host and GPU memory must be done by calling cudaMemcpy explicitly a lot.
- Mixing smart pointers and data container. Imo, smart pointers should handle the ownership, while data containers handles the buffer management.
Hence, I prefer a std::vector-like class for such simple operations. I may integrate these classes to cuda_blackboard in the future, though.

perception/autoware_ground_segmentation_cuda/src/cuda_scan_ground_segmentation/cuda_point_cloud2.cu
anhnv3991: @badai-nguyen This is fixed in my latest commit.

.cspell.json
anhnv3991: @Owen-Liuyuxuan Sorry for my late response. CUDAH is defined in the cuda_common.hpp
```
#ifndef CUDAH
#define CUDAH __forceinline__ __host__ __device__
#endif 
```
It is used for declaring some class methods that can be called from both CPU and GPU code. I just do not want to re-type that long qualifier a lot. However, the spell checker keeps reporting that macro so I had to add it to the word list.
I don't like that name, either. Can you suggest a better one?

Owen-Liuyuxuan: @anhnv3991 

image

I think AI is good at naming variables. I think CUDA_HD/HOST_DEV/BOTH are good names. Please select the one you like.

anhnv3991: @Owen-Liuyuxuan Thanks. That CUDA_HOSTDEV sounds good. I'll take it.

perception/autoware_ground_segmentation_cuda/src/cuda_scan_ground_segmentation/cuda_scan_ground_segmentation_filter.cu
```diff
@@ -0,0 +1,841 @@
...
+
+  device_vector<int> cell_id(point_num, stream_, mempool_);
+  device_vector<ClassifiedPointType> tmp_classified_points(point_num, stream_, mempool_);
+
+  CHECK_CUDA_ERROR(cuda::fill(starting_pid, 0));
+  CHECK_CUDA_ERROR(
+    cuda::launchAsync<BLOCK_SIZE_X>(
+      point_num, 0, stream_->get(), computeCellId,
+      reinterpret_cast<const PointTypeStruct *>(dev_input_points_->data.get()),
+      (int)(dev_input_points_->height * dev_input_points_->width), filter_parameters_,
```
badai-nguyen, line 590:
```suggestion
      point_num, filter_parameters_,
```

perception/autoware_ground_segmentation_cuda/src/cuda_scan_ground_segmentation/cuda_scan_ground_segmentation_filter.cu
```diff
@@ -0,0 +1,841 @@
...
+  int cell_num = filter_parameters_.max_num_cells;
+
+  cell_list.resize(cell_num);
+
+  CHECK_CUDA_ERROR(
+    cuda::launchAsync<BLOCK_SIZE_X>(
+      (int)(cell_list.size()), 0, stream_->get(), cellInit, cell_list.data(),
+      (int)(cell_list.size())));
+
+  starting_pid.resize(cell_num + 1);
```
badai-nguyen, line 580: Shall we use a more self-explanatory variable name here?
```suggestion
  cell_starting_process_id.resize(cell_num + 1);
```

perception/autoware_ground_segmentation_cuda/src/cuda_scan_ground_segmentation/cuda_scan_ground_segmentation_filter.cu
```diff
@@ -0,0 +1,841 @@
...
+ * @param centroid_cells_list_dev Pointer to the array of cell centroid structures (device memory),
+ *        where each cell maintains a count of points assigned to it.
+ *
+ * @note Each thread processes one point. Atomic operations are used to safely increment the
+ *       point count in each cell when multiple threads access the same cell concurrently.
+ * @note Points that fall outside the defined grid (cell_id >= max_num_cells) are ignored.
+ */
+__global__ void computeCellId(
+  const PointTypeStruct * __restrict__ input_points, int point_num, FilterParameters param,
+  int * __restrict__ cell_id, int * __restrict__ count,
```
badai-nguyen, line 73:
```suggestion
  int * __restrict__ cell_id, int * __restrict__ cell_point_nb_count_ptr,
```

perception/autoware_ground_segmentation_cuda/src/cuda_scan_ground_segmentation/cuda_scan_ground_segmentation_filter.cu
```diff
@@ -0,0 +1,841 @@
...
+  float slope,                 // Slope of the line connect the previous ground cells
+  int start_pid, int end_pid,  // Start and end indices of points in the current cell
+  float prev_cell_gnd_radius_avg, float prev_cell_gnd_height_avg, const SegmentationMode & mode)
+{
+  if (start_pid >= end_pid) {
+    return;
+  }
+
+  // Here I assume each thread handles no more than 64 points
+  int64_t ground_mask = 0, recheck_mask = 0, backup_mask = 0;
```
badai-nguyen: I'm sorry for basic question. What factor decides the point numbers which handled by each thread? The max number of points in each cell?

perception/autoware_ground_segmentation_cuda/src/cuda_scan_ground_segmentation/cuda_scan_ground_segmentation_filter.cu
```diff
@@ -0,0 +1,841 @@
...
+  recheck_mask = backup_mask = ground_mask;
+
+  if (
+    param.use_recheck_ground_cluster && cell_gnd_point_num > 1 &&
+    cell_gnd_radius_sum / (float)(cell_gnd_point_num) > param.recheck_start_distance) {
+    // Now recheck the points using the height_min
+    // This is to record the remaining ground point in each thread
+    recheck_mask = ground_mask;
+    // After looping, ground_mask will be cleared, so we need to back it up
+    backup_mask = ground_mask;
```
badai-nguyen: These duplicated lines could be removed?
```suggestion
  recheck_mask = backup_mask = ground_mask;

  if (
    param.use_recheck_ground_cluster && cell_gnd_point_num > 1 &&
    cell_gnd_radius_sum / (float)(cell_gnd_point_num) > param.recheck_start_distance) {
    // Now recheck the points using the height_min
```

perception/autoware_ground_segmentation_cuda/src/cuda_scan_ground_segmentation/cuda_scan_ground_segmentation_filter.cu
```diff
@@ -0,0 +1,841 @@
...
+
+    while (ground_mask != 0) {
+      // Get the offset of the ground point
+      int pos = __ffs(ground_mask) - 1;
+      int gnd_pid = pos * WARP_SIZE + start_pid + wid;
+
+      // Clear the bit
+      ground_mask &= (ground_mask - 1);
```
badai-nguyen: I think ground_mask holds mask for all points (max 64 points) process by threads, why only last bit are cleared here?

perception/autoware_ground_segmentation_cuda/src/cuda_scan_ground_segmentation/cuda_scan_ground_segmentation_filter.cu
```diff
@@ -0,0 +1,841 @@
...
+      p = classified_points[gnd_pid];
+
+      if (p.z > cell_gnd_height_min + param.non_ground_height_threshold && cell_gnd_point_num > 1) {
+        minus_t_gnd_height_sum += p.z;
+        minus_t_gnd_radius_sum += p.radius;
+        ++minus_t_gnd_point_num;
+
+        // If the point is no longer ground, clear the bit
+        recheck_mask &= (recheck_mask - 1);
```
badai-nguyen: The same question for recheck_mask

perception/autoware_ground_segmentation_cuda/src/cuda_scan_ground_segmentation/cuda_scan_ground_segmentation_filter.cu
```diff
@@ -0,0 +1,841 @@
...
+      }
+    }
+
+    // Now use ballot sync to see if there are any ground points remaining
+    uint32_t recheck_res = __ballot_sync(FULL_MASK, recheck_mask > 0);
+    uint32_t backup_res = __ballot_sync(FULL_MASK, backup_mask > 0);
```
badai-nguyen: IMO, at least, the lowest points should be remained recheck_res always > 0, am I correct?

.cspell.json
badai-nguyen: Could you refactor the code according to the agreement here?

badai-nguyen:
Note: 
I add 35c6835cdd08146831b9afe1ce82e86397c986da to add option testing this node together with perception pipeline by enable [use_cuda_ground_segmentation](https://github.com/anhnv3991/autoware.universe/blob/35c6835cdd08146831b9afe1ce82e86397c986da/launch/tier4_perception_launch/launch/perception.launch.xml#L151)

perception/autoware_ground_segmentation_cuda/src/cuda_scan_ground_segmentation/cuda_scan_ground_segmentation_filter.cu
```diff
@@ -0,0 +1,841 @@
...
+{
+  int index = threadIdx.x + blockIdx.x * blockDim.x;
+  int stride = blockDim.x * gridDim.x;
+
+  for (int i = index; i < point_num; i += stride) {
+    auto p = cloud[i];
+    float radius = hypotf(p.x - param.center_x, p.y - param.center_y);
+
+    mark[i] = (radius < max_radius) ? 1 : 0;
+  }
```
badai-nguyen, line 763: @anhnv3991
Since the current ground_segmentation already applies crop_box_filter before processing (as defined in autoware_launch/config/perception/obstacle_segmentation/ground_segmentation/ground_segmentation.param.yaml),
I’d like to update the OutlierFilter to align with the existing crop_box_filter: https://github.com/anhnv3991/autoware.universe/pull/1/files
This will make it easier to compare the results during evaluation.

perception/autoware_ground_segmentation_cuda/src/cuda_scan_ground_segmentation/cuda_scan_ground_segmentation_filter.cu
badai-nguyen: @anhnv3991 
As I understand from implementation and our internal discussion, the points in each cell are divided into 32 WARPs of threads, and each thread can process a maximum of 64 points (since only 64 bits are used to hold the point mask).
This means the number of points that the algorithm can process per cell is limited — currently up to 32 × 64 = 2048 points.

This could be a limitation of the current implementation, as there might be cases (e.g., around poles) where a large number of points are concentrated in a very small area.
Is there any possibility to avoid or mitigate this limitation?

perception/autoware_ground_segmentation_cuda/include/autoware/cuda_scan_ground_segmentation/cuda_mempool_wrapper.hpp
manato, line 1: Similar mempool functions are provided by [`autoware::cuda_utils`](https://github.com/autowarefoundation/autoware_universe/blob/main/sensing/autoware_cuda_utils/include/autoware/cuda_utils/cuda_memory_pool.hpp), though it returns a raw `cudaMempool_t` handler for full control. Do you think `autoware_ground_segmentation_cuda` can also reuse it (or you can update it if needed)? The reason why I ask you this consideration is I'd prefer to avoid code fragmentation of this kind of mempool usage.

perception/autoware_ground_segmentation_cuda/include/autoware/cuda_scan_ground_segmentation/device_vector.hpp
```diff
@@ -0,0 +1,332 @@
...
+  mempool_.reset();
+}
+
+template <typename T>
+void device_vector<T>::allocate(size_t ele_num)
+{
+  if (ele_num > 0) {
+    if (mempool_) {
+      CHECK_CUDA_ERROR(
+        cudaMallocFromPoolAsync(&data_, sizeof(T) * ele_num, mempool_->get(), stream_->get()));
```
manato, line 285: I would recommend zero-clearing the memory region returned by `cudaMallocFromPoolAsync` using `cudaMemsetAsync` because the region may be dirty, and it can cause unexpected behavior.
(Or you can also reuse [pool supported `make_unique` in `autoware::cuda_utils`](https://github.com/autowarefoundation/autoware_universe/blob/cf396aedbe97368a89f8a96dd24b1e6a7b0cdf31/sensing/autoware_cuda_utils/include/autoware/cuda_utils/cuda_unique_ptr.hpp#L59-L79))

perception/autoware_ground_segmentation_cuda/include/autoware/cuda_scan_ground_segmentation/device_vector.hpp
manato, line 1: This class looks very similar to [thrust::device_vector](https://nvidia.github.io/cccl/thrust/api/classthrust_1_1device__vector.html) for me.  Do you have a strong purpose of (re)implementing this class by yourself, or can we consider replacing this with thrust?

perception/autoware_ground_segmentation_cuda/src/cuda_scan_ground_segmentation/cuda_scan_ground_segmentation_filter.cu
```diff
@@ -0,0 +1,819 @@
...
+
+// Initialize the cell list
+__global__ void cellInit(Cell * __restrict__ cell_list, int max_num_cells)
+{
+  int index = threadIdx.x + blockIdx.x * blockDim.x;
+  int stride = blockDim.x * gridDim.x;
+  Cell new_cell;
+
+  for (int i = index; i < max_num_cells; i += stride) {
+    new_cell.gnd_height_min = 1e6f;
```
manato: nits: If `1e6f` means super large value, you can also use `::cuda::std::numeric_limits<float>::max()` or `::cuda::std::numeric_limits<float>::infinity()`

perception/autoware_ground_segmentation_cuda/src/cuda_scan_ground_segmentation/cuda_scan_ground_segmentation_filter.cu
```diff
@@ -0,0 +1,819 @@
...
+  }
+}
+
+__global__ void markEmptyCells(int * starting_pid, int cell_num, int * mark)
+{
+  int index = threadIdx.x + blockIdx.x * blockDim.x;
+  int stride = blockDim.x * gridDim.x;
+
+  for (int i = index; i < cell_num; i += stride) {
+    mark[i] = (starting_pid[i] >= starting_pid[i + 1]) ? 1 : 0;
```
manato, line 539: nits: Might it be better to use `==` rather than `>=` logically because `starting_pid` is zero initialized and is exclusive-scaned?
nits: You can also use `cub::DeviceAdjacentDifference::SubtractRight(Copy)` for similar purpose.

.cspell.json
anhnv3991: @badai-nguyen It is done.

perception/autoware_ground_segmentation_cuda/src/cuda_scan_ground_segmentation/cuda_scan_ground_segmentation_filter.cu
anhnv3991: @badai-nguyen The limit was removed.

perception/autoware_ground_segmentation_cuda/include/autoware/cuda_scan_ground_segmentation/device_vector.hpp
badai-nguyen: I fixed here e5c3de9d2605ac3a29b280ddd623860ebaf7a6df

perception/autoware_ground_segmentation_cuda/src/cuda_scan_ground_segmentation/cuda_scan_ground_segmentation_filter.cu
badai-nguyen: I fixed here 0549db2d82cc1356d16ccbc4ea6053e7c7cd66dd

perception/autoware_ground_segmentation_cuda/include/autoware/cuda_scan_ground_segmentation/device_vector.hpp
badai-nguyen: @manato 
Thank you for your review, and I'm really sorry for the late response.
Since @anhnv3991  seems to be on leave for a while, I’ve made a minor change based on your suggestion until he returns.
Regarding this comment, I actually had the same opinion and raised the same question earlier here: https://github.com/autowarefoundation/autoware_universe/pull/11371#discussion_r2394042042

Does this address your concern?

badai-nguyen:
@manato cc @anhnv3991 
- I evaluated this change by Evaluator and confirmed there no degradation result (even a little bit improved), the comparison results was added in description. 
- Since it seems there will be resource confict if this run on GPU the same time with Centerpoint which use the same topic `/sensing/lidar/concatenated/pointcloud` as input. I compare the processsing time of ground_segmentation in both cases [here](https://tier4.atlassian.net/wiki/spaces/~617fae0b16119e00697e5ada/pages/4619665411/Odaiba+2025-12-02+03#%E3%83%AD%E3%83%BC%E3%82%AB%E3%83%AB%E3%81%A7%E5%86%8D%E6%A4%9C%E8%A8%BC)

I think we need time to address the GPU resource conflict as well as refactor as [your comment](https://github.com/autowarefoundation/autoware_universe/pull/11371#discussion_r2490362792), so as our internal discussion, shall we merge this PR as inital change for cuda version and address remain issue in separate PR?

perception/autoware_ground_segmentation_cuda/include/autoware/cuda_scan_ground_segmentation/device_vector.hpp
manato: @badai-nguyen 
Thanks for taking a look at this. I believe the items listed in that comment are covered by `thrust::device_vector`. I'm sure that the current std::vector-like implementation is well-structured and looks elegant, but I'd still like to consider using thrust from the viewpoint of a well-tested library.

manato (approved):
@badai-nguyen @anhnv3991 
Thanks for the great effort from both of you guys. Since introducing this accelerated function would be a good first step for future development, I approve this PR. Let's treat the remaining topics for further improvement as future tasks!

PR #11378, opened 2025-09-17
refactor(diffusion_planner): refactor lane_segment

SakodaShintaro:
## Description

This PR refactors the lane segment structure in the diffusion planner by replacing the complex polyline system with simpler vector-based geometry. The change removes polyline wrappers and replaces them with direct vector collections for centerlines and boundaries.

- Removes the Polyline class and simplifies lane segment structure
- Updates lane segment interface to use direct vector access instead of polyline methods
- Simplifies test code by removing complex polyline-based logic

## How was this PR tested?

(1) Run `autoware_diffusion_planner`

(2) Confirm that the data converter matches the Python code  (https://github.com/SakodaShintaro/autoware_universe/pull/7)

## Notes for reviewers

None.

#### Additions and removals

| Added | `use_route_handler`   | `bool` | `true`         | Use route handler for selecting route segment indices |

## Effects on system behavior

None.


planning/autoware_diffusion_planner/src/conversion/lanelet.cpp
```diff
@@ -199,24 +112,25 @@ std::vector<LaneSegment> convert_to_lane_segments(
...
-    const auto left_bound = lanelet.leftBound3d();
-    const auto left_points = from_geometry(left_bound);
-    left_boundary_segments.emplace_back(
-      MapType::Unused, interpolate_points(left_points, num_lane_points));
-    const auto right_bound = lanelet.rightBound3d();
-    const auto right_points = from_geometry(right_bound);
-    right_boundary_segments.emplace_back(
-      MapType::Unused, interpolate_points(right_points, num_lane_points));
+    const auto lanelet_subtype = lanelet.attribute("subtype").as<std::string>();
+    if (!lanelet_subtype || ACCEPTABLE_LANE_SUBTYPES.count(lanelet_subtype.value()) == 0) {
```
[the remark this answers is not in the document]
SakodaShintaro: It is optional type.

planning/autoware_diffusion_planner/src/conversion/lanelet.cpp
```diff
@@ -99,97 +79,30 @@ std::vector<LanePoint> interpolate_points(const std::vector<LanePoint> & input,
...
+    const double seg_end = arc_lengths[seg_idx + 1];
+    const double seg_length = seg_end - seg_start;
 
     // Calculate interpolation parameter, handling zero-length segments
-    float safe_seg_length = std::max(seg_length, 1e-6f);
-    float t = (target - seg_start) / safe_seg_length;
-    // Clamp t to [0, 1] to ensure we don't extrapolate
-    t = std::max(0.0f, std::min(1.0f, t));
-    result.push_back(input[seg_idx].lerp(input[seg_idx + 1], t));
+    const double safe_seg_length = std::max(seg_length, 1e-6);
```
go-sakayori: nits:
define something like `constexpr double epsilon = 1e-6;`

planning/autoware_diffusion_planner/include/autoware/diffusion_planner/conversion/lanelet.hpp
```diff
@@ -71,16 +78,15 @@ struct LaneSegment
   static constexpr int64_t TRAFFIC_LIGHT_ID_NONE = -1;
 
   LaneSegment(
-    int64_t id, Polyline polyline, bool is_intersection,
-    const std::vector<BoundarySegment> & left_boundaries,
-    const std::vector<BoundarySegment> & right_boundaries, LineType left_line_type,
+    int64_t id, const Polyline & centerline, const Polyline & left_boundary,
+    const Polyline & right_boundary, const LanePoint & mean_point, LineType left_line_type,
```
go-sakayori:
```suggestion
    const Polyline & right_boundary, const LanePoint & mean_point, const LineType left_line_type,
```
Same can be applied for right_boundary.

planning/autoware_diffusion_planner/include/autoware/diffusion_planner/constants.hpp
```diff
@@ -22,15 +22,15 @@ namespace autoware::diffusion_planner::constants
 constexpr int WEIGHT_MAJOR_VERSION = 1;
 
 // Velocity thresholds
-constexpr float MOVING_VELOCITY_THRESHOLD_MPS = 0.2f;
+constexpr double MOVING_VELOCITY_THRESHOLD_MPS = 0.2;
```
go-sakayori: Why did you change it to double? It's being compared with [linear_vel](https://github.com/autowarefoundation/autoware_universe/blob/main/planning/autoware_diffusion_planner/src/conversion/ego.cpp#L31) which is calculated as float, so I think it is more consistent to set this variable back to float.

planning/autoware_diffusion_planner/src/conversion/lanelet.cpp
```diff
@@ -99,97 +79,30 @@ std::vector<LanePoint> interpolate_points(const std::vector<LanePoint> & input,
...
-    } else {
-      dx = static_cast<float>(itr->x() - (itr - 1)->x());
-      dy = static_cast<float>(itr->y() - (itr - 1)->y());
-      dz = static_cast<float>(itr->z() - (itr - 1)->z());
-      normalize_direction(dx, dy, dz);
-    }
-    output.emplace_back(
-      itr->x(), itr->y(), itr->z(), dx, dy, dz, 0.0);  // TODO(danielsanchezaran): Label ID
+  for (const lanelet::Point3d::ConstType & point : line_string) {
+    output.emplace_back(point.x(), point.y(), point.z());
```
go-sakayori, line 104: nits: Adding `output.reserve(line_string.size());` before the for loop might increase the performance.

go-sakayori:
Not related to this PR, but I found out that [VisualizationParams](https://github.com/autowarefoundation/autoware_universe/blob/main/planning/autoware_diffusion_planner/include/autoware/diffusion_planner/constants.hpp#L38-L43) is not used. Maybe remove it since you have made modification to constants.hpp.

planning/autoware_diffusion_planner/src/conversion/lanelet.cpp
SakodaShintaro: https://github.com/autowarefoundation/autoware_universe/pull/11378/commits/bdfb4ec9a4a66c618a1dbd7a17e38389b268dcd6

planning/autoware_diffusion_planner/include/autoware/diffusion_planner/conversion/lanelet.hpp
SakodaShintaro: https://github.com/autowarefoundation/autoware_universe/pull/11378/commits/7d3a55394ae3051f95159defe1a4bf2d0f868704

planning/autoware_diffusion_planner/include/autoware/diffusion_planner/constants.hpp
SakodaShintaro: https://github.com/autowarefoundation/autoware_universe/pull/11378/commits/390788bd5a4af86e5d5879bfe0b1a1b26a19c997

planning/autoware_diffusion_planner/src/conversion/lanelet.cpp
SakodaShintaro: https://github.com/autowarefoundation/autoware_universe/pull/11378/commits/5c75efa646925035e2c6fcf5797070a310e38f31

PR #11379, opened 2025-09-17
fix(autoware_command_mode_decider): fix operation mode timeout to manual

isamu-takagi:
## Description

Fix the mode is not roll back to manual when the mode transition is timed out. During implementation, a change was made to send manual mode request separately, but this was not reflected in the cancel function.

## Related links

**Private Links:**

- [TIER IV internal link](https://tier4.atlassian.net/browse/T4DEV-12478)

## How was this PR tested?

Disable operation mode transition manager flag node.
Send available=true and completed=false topics manually.
Run planning simulation.
Change to autonomous mode.
Disable autoware control.
Enable autoware contorl.
Check that the autoware control disabled when timeout.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


shtokuda (approved):
LGTM

PR #11386, opened 2025-09-19
feat(mult object tracker): publish merged object if it is multi-channel mode

technolojin:
## Description
In case of the multi-channel mode, merge process of the detected object is not needed.
However, some evaluation process runs base on assumption that `/perception/object_detection/detection/objects` topic is the overall detected objects.

This PR solves issues that publishing merged `detected objects` topic if the multi-channel mode is used.

## Related links
This PR need to be merged after this PR
https://github.com/autowarefoundation/autoware_launch/pull/1647

## How was this PR tested?

[TIER IV INTERNAL](https://evaluation.ci.tier4.jp/evaluation/reports/119f4f48-ce76-5ea7-9822-87b3f8611d09?project_id=prd_jt)

Indicating that evaluation tests are running with the multi-channel mode is enabled.

## Notes for reviewers

This PR does not change any runtime result.

## Effects on system behavior

None.


perception/autoware_multi_object_tracker/config/multi_object_tracker_node.param.yaml
```diff
@@ -36,6 +36,7 @@
       [     9.0, 5.0,   9.0, 9.0,     9.0,        4.0,     3.0,        2.0]
 
     # debug parameters
+    publish_merged_objects: false
```
YoshiRi: [comment]
This parameter will be overrided in mot node when `is_simulator` is true.
I think at least we need to write some comments for readers.

perception/autoware_multi_object_tracker/src/multi_object_tracker_node.cpp
```diff
@@ -143,6 +144,28 @@ MultiObjectTracker::MultiObjectTracker(const rclcpp::NodeOptions & node_options)
   input_manager_->setTriggerFunction(
     std::bind(&MultiObjectTracker::onTrigger, this));  // Set trigger function
 
+  // ROS interface - Publisher
+  tracked_objects_pub_ = create_publisher<autoware_perception_msgs::msg::TrackedObjects>(
+    "output/objects", rclcpp::QoS{1});
+  if (input_channel_size_ < 2) publish_merged_objects_ = false;
```
YoshiRi: [imho]
This part is not friendly for readers. Also, I personally think it's okay for single channel tracker to output "merged" detection.

perception/autoware_multi_object_tracker/src/processor/processor.cpp
```diff
@@ -544,6 +544,71 @@ void TrackerProcessor::getTentativeObjects(
...
 }
 
+void TrackerProcessor::getMergedObjects(
+  const rclcpp::Time & time, const geometry_msgs::msg::Transform & tf_base_to_world,
+  autoware_perception_msgs::msg::DetectedObjects & merged_objects) const
+{
+  std::unique_ptr<ScopedTimeTrack> st_ptr;
+  if (time_keeper_) st_ptr = std::make_unique<ScopedTimeTrack>(__func__, *time_keeper_);
+
+  merged_objects.header.stamp = time;
```
YoshiRi, line 554: [imho]
Reserving array memory before push back is better in terms of memory allocation performance. (Of course it won't affect many case.)

```suggestion
  merged_objects.header.stamp = time;
  merged_objects.objects.clear();
  merged_objects.objects.reserve(list_tracker_.size());
```

technolojin:
For Automatic test, https://github.com/tier4/pilot-auto/pull/1859 need to be merged in the same time.

PR #11396, opened 2025-09-22
fix(autoware_lidar_apollo_instance_segmentation): remove invalid key and change variable int to double

MasatoSaeki:
## Description

This PR fix the issue caused by https://github.com/autowarefoundation/autoware_universe/pull/10097 and https://github.com/autowarefoundation/autoware_universe/pull/11357.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


technolojin (approved):
Thanks for the fix

PR #11399, opened 2025-09-22
fix(simple_object_merger): replace unexpected std::const_pointer_cast() into std::make_shared()

ktro2828:
## Description

Currently, `std::const_pointer_cast()` is used to get rid of const qualifier for the shared pointer, but it causes unexpected mutation of subscribed source object data.

This pull request makes a minor but important change to the way shared pointers are created in the `getTransfor` method of `SimpleObjectMergerBase`. Instead of using `std::const_pointer_cast`, it now creates a new shared pointer by copying the contents of `objects`. This improves safety and ensures the output pointer is not just a cast of the input.

* Changed shared pointer creation in `getTransfor` to use `std::make_shared<ObjsMsgType>(*objects)` instead of `std::const_pointer_cast`, ensuring a new copy is made rather than just casting the original pointer. (`perception/autoware_simple_object_merger/src/simple_object_merger_base.cpp`)

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


a-maumau:
@ktro2828 Thank you for pointing out the issue!
LGTM.

ktro2828:
@technolojin @YoshiRi Friendly ping. Could you take a look as the maintainers?

a-maumau:
@YoshiRi @technolojin Friendly ping

PR #11407, opened 2025-09-25
fix(out_of_lane): ignore collisions inside the trajectory lanelets

maxime-clem:
## Description

By definition the `out_of_lane` module should only care about collisions happening outside the ego lane.
There is a bug in the current implementation where areas inside a lanelet followed by the ego trajectory can be incorrectly checked for collisions. This happens when another lanelet overlap the trajectory lanelet.
This PR fixes the issue by making sure that the areas checked for collisions are not completely inside a trajectory lanelet.

## Related links

**Private Links:**

- [TIER IV internal link](https://tier4.atlassian.net/browse/RT0-39409)

## How was this PR tested?

Psim + reproducer.
Evaluator: TODO

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


satoshi-ota (approved):
LGTM

PR #11412, opened 2025-09-26
feat(run_out): support specifying polygon subtypes for map filtering

maxime-clem:
## Description

The `run_out` module uses parameters to define map geometries (linestring, polygon, lanelets) that can be used to filter the predicted paths or collisions used by the module.
Previously, polygons could only be selected based on their `type`. This PR adds a way to also define their `subtype`.

## Related links
**Private Links:**

- [TIER IV internal link](https://tier4.atlassian.net/browse/RT0-39491)

## How was this PR tested?

Psim after adding an item in the `polygon_types` parameter in the format `type.subtype` (e.g., `area.vegetation`).

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/motion_velocity_planner/autoware_motion_velocity_run_out_module/src/map_data.cpp
+++ b/planning/motion_velocity_planner/autoware_motion_velocity_run_out_module/src/map_data.cpp
@@ -37,9 +37,22 @@ namespace autoware::motion_velocity_planner::run_out
 
 namespace
 {
+/// @brief return true if the given type string is contained in the vector of strings
 bool contains_type(const std::vector<std::string> & types, const std::string & type)
 {
   return std::find(types.begin(), types.end(), type) != types.end();
+}
+
+/// @brief return true if the given type and subtype strings match one of the string in the vector
+/// @details the strings in the vector are assumed to have format "type.subtype" or just "type"
+bool contains_type(
+  const std::vector<std::string> & types, const std::string & type, const std::string & subtype)
+{
+  if (subtype.empty()) {
+    return contains_type(types, type);
+  }
+  const auto full_type_str = type + "." + subtype;
+  return std::find(types.begin(), types.end(), full_type_str) != types.end();
 }
 }  // namespace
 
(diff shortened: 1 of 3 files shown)


satoshi-ota (approved):
LGTM

planning/motion_velocity_planner/autoware_motion_velocity_run_out_module/src/map_data.cpp
```diff
@@ -37,10 +37,33 @@ namespace autoware::motion_velocity_planner::run_out
...
+    if (main_type_str != type) {
+      continue;
+    }
+    const auto sub_type_str = type_str.substr(dot_pos + 1, type_str.size());
+    if (sub_type_str == subtype) {
+      return true;
+    }
+  }
+  return false;
+}
```
kyoichi-sugahara, line 56:
```suggestion
bool contains_type(
  const std::vector<std::string> & types, const std::string & type, const std::string & subtype)
{
  if (subtype.empty()) {
    return std::find(types.begin(), types.end(), type) != types.end();
  }

  const std::string full_type = type + "." + subtype;

  for (const auto & type_str : types) {
    if (type_str == full_type) {
      return true;
    }

    const auto dot_pos = type_str.find('.');

    if (dot_pos == std::string::npos) {
      continue;
    }

    if (type_str.compare(0, dot_pos, type) != 0) {
      continue;
    }

    if (type_str.compare(dot_pos + 1, std::string::npos, subtype) == 0) {
      return true;
    }
  }
  return false;
}
```

The main purpose is to appeal that I've checked the code :sweat_smile: , but I'd like to propose some minor improvements. The current implementation is fine as is.
- By changing `substr()` to `compare()`, we can avoid string copying.
- By performing the exact match check first, we might improve the flow (though I'm not entirely confident about this).

kyoichi-sugahara (approved):

planning/motion_velocity_planner/autoware_motion_velocity_run_out_module/src/map_data.cpp
maxime-clem: Thank you ! Your comment helped me simplify the function even more.

PR #11441, opened 2025-10-03
fix(trajectory_optimizer): add acceleration recalculation

SakodaShintaro:
## Description

This PR fixes an issue in the trajectory optimizer by adding acceleration recalculation after velocity changes. When the maximum velocity is applied to trajectory points, the acceleration values need to be recalculated to maintain consistency between velocity and acceleration values in the trajectory.

## How was this PR tested?

Run `autoware_diffusion_planner`

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/autoware_trajectory_optimizer/src/utils.cpp
+++ b/planning/autoware_trajectory_optimizer/src/utils.cpp
@@ -138,6 +138,22 @@ void set_max_velocity(TrajectoryPoints & input_trajectory_array, const float max
     [max_velocity](TrajectoryPoint & point) {
       point.longitudinal_velocity_mps = std::min(point.longitudinal_velocity_mps, max_velocity);
     });
+
+  // recalculate acceleration after velocity change
+  const int64_t size = input_trajectory_array.size();
+  for (int64_t i = 0; i + 1 < size; ++i) {
+    const float curr_time_from_start =
+      static_cast<float>(input_trajectory_array[i].time_from_start.sec) +
+      static_cast<float>(input_trajectory_array[i].time_from_start.nanosec) * 1e-9f;
+    const float next_time_from_start =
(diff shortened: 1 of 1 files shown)


yukkysaito (approved):

planning/autoware_trajectory_optimizer/src/utils.cpp
```diff
@@ -138,6 +138,22 @@ void set_max_velocity(TrajectoryPoints & input_trajectory_array, const float max
...
+    const float curr_time_from_start =
+      static_cast<float>(input_trajectory_array[i].time_from_start.sec) +
+      static_cast<float>(input_trajectory_array[i].time_from_start.nanosec) * 1e-9f;
+    const float next_time_from_start =
+      static_cast<float>(input_trajectory_array[i + 1].time_from_start.sec) +
+      static_cast<float>(input_trajectory_array[i + 1].time_from_start.nanosec) * 1e-9f;
+    const float dt = next_time_from_start - curr_time_from_start;
+    const float dv = input_trajectory_array[i + 1].longitudinal_velocity_mps -
+                     input_trajectory_array[i].longitudinal_velocity_mps;
+    input_trajectory_array[i].acceleration_mps2 = dv / (dt + 1e-5f);
```
go-sakayori, line 154: I think it is better to check if it exceeds a certain threshold, but how do you think? For instance if it exceeds 10m/s^2 , an upper limit should be applied.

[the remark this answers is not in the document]
go-sakayori: I think Copilot is right.

go-sakayori: The control side would handle it, so it is okay

Issue #11370, opened 2025-09-16
external_velocity_limit_selector: Bug - Velocity Limit Not Updated on Empty Table Clear


marvsee:
### Checklist

- [x] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [x] I've searched other issues and no duplicate issues were found.
- [x] I'm convinced that this is not my fault but a bug.

### Description

There is a logical bug in the `ExternalVelocityLimitSelectorNode` implementation:

- When a velocity limit clear command is received, `onVelocityLimitClearCommand` calls `clearVelocityLimit`, then `getCurrentVelocityLimit` and `publishVelocityLimit`.

- If `velocity_limit_table_` is empty, `clearVelocityLimit` returns early and does not call `updateVelocityLimit()`.

- This means `hardest_limit_` is not updated, and subsequent calls to `getCurrentVelocityLimit()` may return invalid or stale data.

### Expected behavior

After a clear command (even when the table is empty), the node should always update its internal state and publish a valid/default velocity limit.

### Actual behavior

When a clear command is send before any velocity limit command, an undefined value will be published as the external velocity limit. In RVIZ the speed limit is immediately set to 0 km/h.

### Steps to reproduce

1. Start the planning simulation with the external_velocity_limit_selector.
2. Do not send any external velocity limit
3. Send a clear command (VelocityLimitClearCommand). E.g. via Terminal: ``ros2 topic pub -1 /planning/scenario_planning/clear_velocity_limit tier4_planning_msgs/msg/VelocityLimitClearCommand "{command: true, sender: "examplesender"}"``
4. Observe that the published velocity limit may be inconsistent or default-initialized.

### Possible causes

The undefined state of the `hardest_limit_` can be avoided, if `updateVelocityLimit()` is always called. My proposal to fix this is:

```
void ExternalVelocityLimitSelectorNode::clearVelocityLimit(const std::string & sender)
{
  if (velocity_limit_table_.empty()) {
    RCLCPP_WARN(get_logger(), "no velocity limit has been set from internal.");
    updateVelocityLimit(); // <-- this should be added, so the limit is always updated, even if the table is empty
    return;
  }

  velocity_limit_table_.erase(sender);
  updateVelocityLimit();
}
```

### Additional context

[...]

PR #11463, opened 2025-10-08
fix(external_velocity_limit_selector): correctly handle clear command on empty table

marvsee:
## Description
This PR fixes a logical bug in the ExternalVelocityLimitSelectorNode.

When a VelocityLimitClearCommand is received while the internal velocity_limit_table_ is empty, the node fails to update its internal state (hardest_limit_). This causes a stale or uninitialized velocity limit to be published, which can lead to incorrect behavior (e.g., the speed limit dropping to 0 km/h in RViz).

The fix ensures that updateVelocityLimit() is always called after a clear command, even if the table is empty. This guarantees that the node's state is reset to a valid default and the correct limit is published.
## Related links

**Parent Issue:**

#11370 

## How was this PR tested?
- Started the planning simulation including the external_velocity_limit_selector node.
- Sent a VelocityLimitClearCommand before any other velocity limit was set, using the command: `ros2 topic pub -1 /planning/scenario_planning/clear_velocity_limit tier4_planning_msgs/msg/VelocityLimitClearCommand "{command: true, sender: "examplesender"}"`
- Check in RViz, if the velocity behaves differently.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/autoware_external_velocity_limit_selector/test/test_external_velocity_limit_selector_node_launch.py
+++ b/planning/autoware_external_velocity_limit_selector/test/test_external_velocity_limit_selector_node_launch.py
@@ -136,8 +136,8 @@ class TestExternalVelocityLimitSelector(unittest.TestCase):
         clear_cmd = VelocityLimitClearCommand(command=True)
         self.pub_clear_limit_.publish(clear_cmd)
         self.wait_for_output()
-        # velocity limit is 0 before any limit is set
-        self.assertEqual(self.velocity_limit_output_.max_velocity, 0.0)
+        # velocity limit should be max_vel_param, before any external limit is set
+        self.assertEqual(self.velocity_limit_output_.max_velocity, 15.0)
 
         # Send velocity limits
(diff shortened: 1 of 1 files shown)


xmfcx:
https://github.com/autowarefoundation/autoware_universe/actions/runs/18591153418/job/53006568678?pr=11463#step:17:390
Here the launch test failed.

https://github.com/autowarefoundation/autoware_universe/blob/892e8c59a31d1b1049690672b72d5d4ffea24863/planning/autoware_external_velocity_limit_selector/test/test_external_velocity_limit_selector_node_launch.py#L140

> `AssertionError: 15.0 != 0.0`

Could you evaluate this and if necessary update the test file too?

You can test the single package with:
`colcon test --event-handlers console_cohesion+ --packages-select autoware_external_velocity_limit_selector`

marvsee:
@xmfcx
I updated the one line in the test file too, because the expected max velocity of 0.0 does not make sense. When now external velocity limit is requested, it should go back to the max_vel_param value.

PR #11467, opened 2025-10-09
feat(image_object_locator): add fov validation

a-maumau:
## Description
This PR will add a validation/filtering option for ROI / bbox in `autowre_image_object_locator` package.

For cameras with a large FOV (typically large horizontal FOV), points near the edge tend to have lower precision when estimating object positions. This option filters out the noisy regions in the image.

When `roi_fov_validation` is set to `true`, input ROIs are checked to ensure they fall within the valid horizontal FOV.
For the vertical FOV, `detection_max_range` serves a similar purpose under normal camera settings.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


a-maumau:
In some cases, the FOV filter may not reasonable for some camera calibration result, and removing the object may not a good idea.
Modeling the uncertainty (covariance) and do the work in the object tracker should be more reasonable.
So I will close this PR.

I create a new PR that calculate the covariance and handle some object removal in https://github.com/autowarefoundation/autoware_universe/pull/11522

PR #11468, opened 2025-10-09
feat(autoware_bevfusion): separate image backbone from fusion model and add lidar intensity option

SamratThapa120:
## Description

This pull request introduces significant updates to the BEVFusion perception package, mainly to support a fusion model architecture with a separate image backbone. A separate image backbone opens  possibilities for various optimizations to the inference pipeline in the future such as post-training quantization of image backbone, image feature caching, etc.

In addition, the roi resolution is updated to better match the training dataset, and a new parameter `use_intensity` is added to enable or disable lidar intensity use during inference. The training dataset does not consist of intensity so this feature was needed.

We also refactor various parts of the code to reduce code complexity.

## How was this PR tested?

Tested with various TIER IV internal rosbags.

#### Additions and removals

| Change type | Parameter Name                 | Type     | Default Value                                                | Description                                      |
|:-------------|:-------------------------------|:---------|:-------------------------------------------------------------|:-------------------------------------------------|
| Added        | `use_intensity`                | `bool`   | `false`                                                      | Whether to use LiDAR intensity in prediction.     |
| Added        | `image_backbone_onnx_path`     | `string` | `"/home/autoware/autoware_data/bevfusion/image_backbone.onnx"`      | Path to the image backbone ONNX model.           |
| Added        | `image_backbone_engine_path`   | `string` | `"/home/autoware/autoware_data/bevfusion/image_backbone.engine"`    | Path to the TensorRT engine file for inference.  |
| Added        | `image_backbone_trt_precision` | `string` | `"fp16"`                                                     | Precision type for TensorRT inference.           |

## Notes for Reviewers

You can follow the following steps to test BEVFusion:

1. Download the onnx checkpoints[TIER IV internal link](https://drive.google.com/drive/u/0/folders/16ZqWTd51LjunN9T_jjrbWkHaRVMLhBnf)
2. Pull pilot-auto docker image[TIER IV internal link](https://evaluation.tier4.jp/evaluation/reports/d3bbe869-f126-5b07-bc7a-f5bbe41fa438/builds/c830fa71-4137-5182-9ad7-af9f719edec0?project_id=prd_jt). Or you can build autoware locally with the same commit id.
3. Download the rosbags [TIER IV internal link](https://console.data-search.tier4.jp/projects/prd_jt/t4datasets/e50d0da8-ce92-4ba5-9657-90ddad9cc5c2?version=2&tab=info)
```
webauto data annotation-dataset pull --project-id x2_dev --annotation-dataset-id 057e4039-54fd-4257-ac2f-1e447b2f2f71 --annotation-dataset-version-id 1 --asset-dir  /path/to/dataset

# The rosbag is located under /path/to/dataset/{dataset_id}/input_bag
```
4. Start the container while mounting the the rosbags and onnx checkpoints
```bash
rocker --x11 \
  --gpus all \
  --nocleanup \
  --volume  /path/to/onnx_files:/home/autoware/autoware_data/bevfusion\
 --volume /path/to/dataset:/home/autoware/workspace/data
  --name bevfusion_detection pilot_auto_docker_image
```

The images in the rosbags are not rectified, so you will need to update the images with the attached script.
[inject_uncompressed_image.py](https://github.com/user-attachments/files/23157200/inject_uncompressed_image.py)

```bash
python3 inject_uncompressed_image.py /path/to/dataset/{dataset_id}/input_bag /path/to/dataset/{dataset_id}/injected_input_bag

```
5. Clone the branch and build bevfusion
```
git clone -b chore/bevfusion/model-separate git@github.com:SamratThapa120/autoware_universe.git
source $AUTOWARE_PATH/install/setup.bash
cd autoware_universe
colcon build --symlink-install --packages-up-to autoware_bevfusion --cmake-args -DCMAKE_BUILD_TYPE=Release
```
6. Launch Bevfusion

```
source $AUTOWARE_PATH/install/setup.bash
source /path/to/autoware_universe/install/setup.bash
ros2 autoware_bevfusion bevfusion.launch.xml model_name:=bevfusion_camera_lidar #change to bevfusion_lidar for lidar-only model
```

## NOTES

changes the model weights (onnx) files, so autoware needs to be updated with new model artifacts when we merge this PR.


perception/autoware_bevfusion/config/bevfusion_camera_lidar.param.yaml
```diff
@@ -11,6 +11,12 @@
     cloud_capacity: 2000000
     onnx_path: "$(var model_path)/bevfusion_camera_lidar.onnx"
     engine_path: "$(var model_path)/bevfusion_camera_lidar.engine"
+
+    # image backbone network (fusion model always uses separate image backbone)
+    image_backbone_onnx_path: "$(var model_path)/image_backbone.onnx"
+    image_backbone_engine_path: "$(var model_path)/image_backbone.engine"
+    image_backbone_trt_precision: fp16
+
```
amadeuszsz, line 17: I think we should directly point model's name in file path, therefore I suggest to add `bevfusion` prefix. CenterPoint does same way (suffix). Please populate this change in all the BEVFusion's code as well (schema, param name in cpp).
```suggestion
    # image backbone network (fusion model always uses separate image backbone)
    image_backbone_onnx_path: "$(var model_path)/bevfusion_image_backbone.onnx"
    image_backbone_engine_path: "$(var model_path)/bevfusion_image_backbone.engine"
    image_backbone_trt_precision: fp16
```

perception/autoware_bevfusion/lib/bevfusion_trt.cpp
```diff
@@ -115,6 +141,11 @@ void BEVFusionTRT::initPtr()
           config_.raw_image_height_ * config_.raw_image_width_ * 3));
     }
     camera_masks_d_ = autoware::cuda_utils::make_unique<float[]>(config_.num_cameras_);
+
+    // buffers for fusion model with separate image backbone
+    image_feats_d_ = autoware::cuda_utils::make_unique<float[]>(
+      config_.num_cameras_ * 256 * config_.features_height_ * config_.features_width_);
+    img_aug_matrix_d_ = autoware::cuda_utils::make_unique<float[]>(config_.num_cameras_ * 4 * 4);
```
amadeuszsz: Please replace magic numbers with relevant field from config if value is relevant to the model. `constexpr` eventually. I see similar issue in this file introduced before, please adjust them as well if you can 🙏🏻

perception/autoware_bevfusion/schema/ml_package_bevfusion.schema.json
```diff
@@ -162,6 +162,11 @@
           "description": "Number of depth features used in the view transform.",
           "default": 118,
           "minimum": 0
+        },
+        "use_intensity": {
```
amadeuszsz, line 172: `bevfusion.schema.json` missing `image_backbone_onnx_path`, `image_backbone_engine_path` and `image_backbone_trt_precision`.

perception/autoware_bevfusion/src/bevfusion_node.cpp
```diff
@@ -45,6 +45,14 @@ BEVFusionNode::BEVFusionNode(const rclcpp::NodeOptions & options)
...
     this->declare_parameter<std::string>("trt_precision", descriptor);
 
+  // Image backbone parameters (for fusion model)
+  const std::string image_backbone_onnx_path =
+    this->declare_parameter<std::string>("image_backbone_onnx_path", "", descriptor);
+  const std::string image_backbone_engine_path =
+    this->declare_parameter<std::string>("image_backbone_engine_path", "", descriptor);
+  const std::string image_backbone_trt_precision =
+    this->declare_parameter<std::string>("image_backbone_trt_precision", trt_precision, descriptor);
+
```
amadeuszsz: Please remove default params. LiDAR only config should consists of these parameters too, empty strings in this case.

perception/autoware_bevfusion/src/bevfusion_node.cpp
```diff
@@ -139,9 +147,31 @@ BEVFusionNode::BEVFusionNode(const rclcpp::NodeOptions & options)
   detection_class_remapper_.setParameters(
     allow_remapping_by_area_matrix, min_area_matrix, max_area_matrix);
 
-  auto trt_config =
-    tensorrt_common::TrtCommonConfig(onnx_path, trt_precision, engine_path, 1ULL << 32U);
-  detector_ptr_ = std::make_unique<BEVFusionTRT>(trt_config, densification_param, config);
+  // Initialize detector based on architecture
+  if (sensor_fusion_) {
```
amadeuszsz: You proposed _"fusion model always uses separate image backbone"_ which is totally fine. With this design `sensor_fusion` parameter is redundant, therefore I suggest a slight refactor:
- Remove `sensor_fusion` parameter (param files, cpp, schema).
- Handle all three `image_backbone*` parameters:
  - All empty strings --> `sensor_fusion_ = false`.
  - All non-empty strings --> `sensor_fusion_ = true`.
  - Otherwise throw `std::invalid_argument` exception.
  - Schema file should inform about this behavior.

perception/autoware_bevfusion/src/bevfusion_node.cpp
```diff
@@ -221,6 +257,9 @@ void BEVFusionNode::cloudCallback(
   const auto objects_sub_count =
     objects_pub_->get_subscription_count() + objects_pub_->get_intra_process_subscription_count();
   if (objects_sub_count < 1) {
+    RCLCPP_WARN_THROTTLE(
+      this->get_logger(), *this->get_clock(), 10000,
+      "No subscribers found for detection objects topic. Skipping detection.");
```
amadeuszsz: I think this is unnecessary log.

perception/autoware_bevfusion/include/autoware/bevfusion/bevfusion_trt.hpp
```diff
@@ -75,6 +75,13 @@ class BEVFUSION_PUBLIC BEVFusionTRT
   explicit BEVFusionTRT(
     const tensorrt_common::TrtCommonConfig & trt_config,
     const DensificationParam & densification_param, const BEVFusionConfig & config);
+
+  // Constructor for fusion model with separate image backbone and main body
+  explicit BEVFusionTRT(
+    const tensorrt_common::TrtCommonConfig & main_trt_config,
+    const tensorrt_common::TrtCommonConfig & image_backbone_trt_config,
+    const DensificationParam & densification_param, const BEVFusionConfig & config);
+
```
amadeuszsz: Looking deep inside implementation there is too much boilerplate code. Please rework the implementation of how the LiDAR only and Camera + LiDAR instances are initialized. I suggest to introduce an extra struct:
```cpp
struct TrtBEVFusionConfig {
tensorrt_common::TrtCommonConfig common;
std::optional<tensorrt_common::TrtCommonConfig> image_backbone;
};
```
and pass this to constructor. Later on I believe you can remove `initTrtFusion` and initialize required members base on `sensor_fusion_` value.

perception/autoware_bevfusion/lib/bevfusion_trt.cpp
```diff
@@ -392,6 +518,8 @@ bool BEVFusionTRT::preProcess(
   }
 
   if (!vg_ptr_->enqueuePointCloud(pc_msg_ptr, tf_buffer)) {
+    RCLCPP_ERROR(
```
amadeuszsz: This is redundant. In node file we already have a log.

perception/autoware_bevfusion/lib/bevfusion_trt.cpp
```diff
@@ -414,6 +542,7 @@ bool BEVFusionTRT::preProcess(
 
     for (std::int64_t camera_id = 0; camera_id < config_.num_cameras_; camera_id++) {
       int start_y = roi_start_y_vector_[camera_id];
+
```
amadeuszsz: Was it pre-commit? If not and this PR does not address code style adjustments, please avoid such a change for a clean history purpose.

perception/autoware_bevfusion/lib/bevfusion_trt.cpp
```diff
@@ -488,16 +628,64 @@ bool BEVFusionTRT::preProcess(
...
+  if (config_.sensor_fusion_) {
+    // Fusion model uses separate image backbone and main body
+    return inferenceFusion();
+  } else {
+    // Lidar-only model uses single network
+    auto status = network_trt_ptr_->enqueueV3(stream_);
+    CHECK_CUDA_ERROR(cudaStreamSynchronize(stream_));
+
+    if (!status) {
+      RCLCPP_ERROR(rclcpp::get_logger("bevfusion"), "Fail to enqueue and skip to detect.");
```
amadeuszsz: Redundant log (same as pointed before).

perception/autoware_bevfusion/lib/bevfusion_trt.cpp
```diff
@@ -488,16 +628,64 @@ bool BEVFusionTRT::preProcess(
...
+  }
+
+  // Then run main network with image features
   auto status = network_trt_ptr_->enqueueV3(stream_);
   CHECK_CUDA_ERROR(cudaStreamSynchronize(stream_));
 
   if (!status) {
-    RCLCPP_ERROR(rclcpp::get_logger("bevfusion"), "Fail to enqueue and skip to detect.");
+    RCLCPP_ERROR(
+      rclcpp::get_logger("bevfusion"), "Fail to enqueue main network and skip to detect.");
```
amadeuszsz: Redundant log. Looks like it was not catch before. In general we only need to return false and call logger in node file. Please check all the logger calls in this file and remove them if needed.

perception/autoware_bevfusion/lib/bevfusion_trt.cpp
```diff
@@ -369,6 +480,21 @@ void BEVFusionTRT::setIntrinsicsExtrinsics(
...
+    for (const auto & matrix : img_aug_matrices_) {
+      for (int i = 0; i < 4; ++i) {
+        for (int j = 0; j < 4; ++j) {
+          img_aug_matrix_flattened.push_back(matrix(i, j));
+        }
+      }
+    }
+    cudaMemcpy(
+      img_aug_matrix_d_.get(), img_aug_matrix_flattened.data(),
+      config_.num_cameras_ * 4 * 4 * sizeof(float), cudaMemcpyHostToDevice);
```
amadeuszsz: Instead of copying to the auxiliary host vector, could you please copy `matrix(i, j)` directly to expected position in img_aug_matrix_d_? Or going further, I think `img_aug_matrices_` memory representation is same as you try to fit it now, therefore you can just do `cudaMemcpy(img_aug_matrix_d_.get(), img_aug_matrices_.data(), ...` but please confirm it first.

perception/autoware_bevfusion/lib/bevfusion_trt.cpp
```diff
@@ -23,6 +23,9 @@
 #include <autoware/cuda_utils/cuda_utils.hpp>
 #include <autoware/point_types/memory.hpp>
 #include <autoware/universe_utils/math/constants.hpp>
+#include <opencv2/imgcodecs.hpp>
+#include <opencv2/imgproc.hpp>
+#include <opencv2/opencv.hpp>
```
amadeuszsz: Are these headers necessary?

amadeuszsz (changes requested):
Thank you for your contribution! Apart from comments:
- For testing purposes, I would like to ask you for sharing to me target LiDAR only and Camera+LiDAR ONNX files which will match this new configuration.
- Note that cuda_blackboard is not addressed for camera subscribers. If we aim to use Camera + LiDAR model, I suggest to adjust this implementation. It can be also addressed in separate PR.

SamratThapa120:
@amadeuszsz Thanks for the review. I have made all the requested changes, and uploaded the onnx files to 
[TIER IV internal link](https://drive.google.com/drive/u/0/folders/16ZqWTd51LjunN9T_jjrbWkHaRVMLhBnf)

perception/autoware_bevfusion/include/autoware/bevfusion/bevfusion_config.hpp
```diff
@@ -38,10 +41,36 @@ class BEVFusionConfig
...
+        "Image backbone parameters must be either all empty (lidar-only mode) or all non-empty "
+        "(fusion mode). Got: image_backbone_onnx_path='" +
+        image_backbone_onnx_path + "', image_backbone_engine_path='" + image_backbone_engine_path +
+        "', image_backbone_trt_precision='" + image_backbone_trt_precision + "'");
+    }
+
+    sensor_fusion_ = all_non_empty;
+
+    if (use_intensity) {
+      // x, y, z, intensity, lag
```
KSeangTan: `timestamp_lag` is more accurate

perception/autoware_bevfusion/include/autoware/bevfusion/bevfusion_config.hpp
```diff
@@ -128,7 +164,8 @@ class BEVFusionConfig
   std::int64_t min_num_voxels_{};
   std::int64_t max_num_voxels_{};
   std::int64_t max_points_per_voxel_;
-  const std::int64_t num_point_feature_size_{5};  // x, y, z, intensity, lag
+
+  std::int64_t num_point_feature_size_{5};  // x, y, z, intensity, lag
```
KSeangTan: Should it be `4` by default?

perception/autoware_bevfusion/config/ml_package_bevfusion_camera_lidar.param.yaml
```diff
@@ -7,18 +7,20 @@
...
+    img_aug_scale_y: 0.4
     roi_height: 384
-    roi_width: 704
+    roi_width: 576
     features_height: 48
-    features_width: 88
-    num_depth_features: 118
+    features_width: 72
+    num_depth_features: 95
+    image_feature_dim: 256
```
KSeangTan: Is this feature channels or dim?

perception/autoware_bevfusion/lib/preprocess/preprocess_kernel.cu
```diff
@@ -55,7 +55,8 @@ PreprocessCuda::PreprocessCuda(
 
 __global__ void generateSweepPoints_kernel(
   const InputPointType * __restrict__ input_points, std::size_t points_size, float time_lag,
-  const float * transform_array, int num_features, float * __restrict__ output_points)
+  const float * transform_array, int num_features, bool use_intensity,
```
KSeangTan: Can we use `template` to add intensity?

https://github.com/autowarefoundation/autoware_universe/blob/main/perception/autoware_lidar_centerpoint/lib/preprocess/preprocess_kernel.cu#L58

perception/autoware_bevfusion/config/ml_package_bevfusion_camera_lidar.param.yaml
```diff
@@ -7,18 +7,20 @@
...
+    d_bound: [1.0, 134.0, 1.4]
     x_bound: [-122.4, 122.4, 0.68]
     y_bound: [-122.4, 122.4, 0.68]
     z_bound: [-10.0, 10.0, 20.0]
     num_cameras: 6
     raw_image_height: 1080
     raw_image_width: 1440
-    img_aug_scale_x: 0.489
-    img_aug_scale_y: 0.489
+    img_aug_scale_x: 0.4
```
KSeangTan, line 19: Can you remind me why do we need this again?

KSeangTan (changes requested):
LGTM overall, just some minor changes
Thanks for the great work!

perception/autoware_bevfusion/lib/bevfusion_trt.cpp
```diff
@@ -369,6 +453,21 @@ void BEVFusionTRT::setIntrinsicsExtrinsics(
...
+    const std::size_t matrix_size = 
+      BEVFusionConfig::kTransformMatrixDim * BEVFusionConfig::kTransformMatrixDim;
+    
+    for (std::int64_t i = 0; i < config_.num_cameras_; i++) {
+      cudaMemcpy(
+        img_aug_matrix_d_.get() + i * matrix_size,
+        img_aug_matrices_[i].data(),
+        matrix_size * sizeof(float),
+        cudaMemcpyHostToDevice);
+    }
```
amadeuszsz: I think you don't need for loop - data is continuous across whole the vector. Example code:
```cpp
    const std::size_t matrices_size = BEVFusionConfig::kTransformMatrixDim *
                                      BEVFusionConfig::kTransformMatrixDim * config_.num_cameras_;
    cudaMemcpy(
      img_aug_matrix_d_.get(), img_aug_matrices_.data(), matrices_size * sizeof(float),
      cudaMemcpyHostToDevice);
  }
```
Please confirm it.

perception/autoware_bevfusion/config/ml_package_bevfusion_camera_lidar.param.yaml
SamratThapa120: This is the scaling ratio before ROI extraction along x axis. Since this is model-specific, it had to be updated to work with the newly trained model that uses different image sizes than the previously released models. it had to be updated with the The descriptions can be found in https://github.com/autowarefoundation/autoware_universe/blob/eb4950d17d78430aaed70badd2120a6305e3ba09/perception/autoware_bevfusion/schema/ml_package_bevfusion.schema.json#L124

perception/autoware_bevfusion/lib/bevfusion_trt.cpp
SamratThapa120: Thanks. 
Each `Matrix4fRowM `object is individually contiguous due to `Eigen::RowMajor` storage. However, each of these objects are created separately, which does not guaranteed the data to be contiguous across different matrix objects
https://github.com/autowarefoundation/autoware_universe/blob/12cfe8b9cf5f8ec9713bd15a4af5b0a0d48858d2/perception/autoware_bevfusion/lib/bevfusion_trt.cpp#L401

perception/autoware_bevfusion/lib/preprocess/preprocess_kernel.cu
SamratThapa120: fixed 
3763ceb

perception/autoware_bevfusion/config/ml_package_bevfusion_camera_lidar.param.yaml
SamratThapa120: fixed
https://github.com/autowarefoundation/autoware_universe/commit/3763ceb1cb01c2bf40f48307ed745bf0ff8ced1f

perception/autoware_bevfusion/include/autoware/bevfusion/bevfusion_config.hpp
SamratThapa120: fixed
https://github.com/autowarefoundation/autoware_universe/commit/3763ceb1cb01c2bf40f48307ed745bf0ff8ced1f

perception/autoware_bevfusion/include/autoware/bevfusion/bevfusion_config.hpp
SamratThapa120: fixed
https://github.com/autowarefoundation/autoware_universe/commit/3763ceb1cb01c2bf40f48307ed745bf0ff8ced1f

amadeuszsz (approved):
LGTM!

SamratThapa120:
@KSeangTan Could you re-review when you have time

KSeangTan (approved):
LGTM

S-XiaoHa:
@SamratThapa120 Hi ,

Thank you for this excellent PR! I noticed that this update introduces the image_backbone_onnx_path parameter to support a separate image backbone architecture. This seems to be exactly what I need to address the latency bottlenecks I'm currently facing on the NVIDIA Jetson Orin (32GB).

I am currently running a standalone BEVFusion model (exported through mmdeploy with the spconv plugin) on a relatively old branch of Autoware Humble. Current Benchmarks (FP16, Orin 32G):

LiDAR-only: ~80 ms

Fusion (Monolithic): ~290 ms

Bottleneck: The camera branch adds ~210 ms overhead.

Based on your PR description regarding the "separate image backbone," I have a few specific questions to help me migrate to this new architecture for better performance:

(1) Benchmarks on Orin & Parallelism: Have you benchmarked this "split architecture" on Orin? Since the image backbone is now decoupled, does the current bevfusion_node implementation automatically schedule the Image Backbone and LiDAR Backbone on parallel CUDA streams? I am hoping this split will significantly hide the latency of the image branch.

(2) INT8 Quantization Workflow: You mentioned that "A separate image backbone opens possibilities for... post-training quantization." This is critical for Orin. Since I can now provide a standalone image_backbone.onnx, can I simply export a standard ResNet50 ONNX, run it through trtexec --int8 --saveEngine=..., and then pass this INT8 Engine path to image_backbone_engine_path?

Does the node require any specific IO node names (binding names) for the image backbone ONNX?

Are there any reference scripts or configs for calibrating this specific separated backbone?

(3) LiDAR Branch Optimization: Even with the LiDAR-only mode, I am seeing ~80ms latency. Is this purely limited by the generic ImplicitGemm TensorRT plugin efficiency compared to NVIDIA's native CUDA SCN kernels (~20ms)? Are there any plans to integrate faster sparse convolution kernels in future updates?

SamratThapa120:
Hello @S-XiaoHa, thank you for the detailed feedback! Here is a deeper dive into your questions:

### (1) Benchmarks & Parallelism
While I haven't benchmarked on the Orin 32GB yet, here are the baseline results from an **RTX 3090 (FP16)** to help you calibrate expectations:

| Mode | Inference (ms) | Preprocess (ms) | Total (ms) |
| :--- | :--- | :--- | :--- |
| **Camera-LiDAR** | 40 | 30 | 70 |
| **LiDAR Only** | 24 | <1 | 25 |

* **CUDA Stream Scheduling:** Currently, all forward passes are executed on the **same CUDA stream**. While decoupling the backbone allows for potential parallelism, simply moving them to separate streams on the Orin often leads to **resource contention**. Since both branches are compute-heavy, they tend to saturate the GPU's SMs (Streaming Multiprocessors), resulting in serialized execution regardless of the stream.
* **Preprocessing Latency:** Our data shows that image preprocessing (2880x1860) is a major bottleneck (~30ms). In your case, this might be higher or lower depending on image size. I am planning to implement **asynchronous preprocessing**, where images are processed immediately upon arrival rather than waiting for the LiDAR scan to finish. This should effectively "hide" the preprocessing time.

### (2) INT8 Quantization Workflow
We are starting to look into INT8 post-training quantization (PTQ) now (cc @vividf ). 
* **Workflow:** Your approach (ONNX -> `trtexec` with `--int8`) is fundamentally correct, but i havent tried it so I cannot assure you that it works with current codebase.
* **IO Bindings:** The node requires specific naming for input/output tensors to map correctly. Please refer to our deployment configuration for the exact binding names: [BEVFusion Camera Backbone Config](https://github.com/tier4/AWML/blob/5deba7010b8346dc14409039b83d0539534e1ec7/projects/BEVFusion/configs/deploy/bevfusion_camera_backbone_tensorrt_dynamic.py) (changes to be merged in PR: https://github.com/tier4/AWML/pull/98).

### (3) Sparse Convolution Kernels
Sorry, but as of now we havent thought about integrating faster sparse convolution kernels, but I have noted it for future exploration. (cc @KSeangTan ).

---

S-XiaoHa:
Hi @SamratThapa120,

Thank you for the detailed explanation regarding CUDA stream scheduling and resource contention! That clarifies why simply splitting the backbone might not yield immediate parallelism gains on constrained hardware like Orin.

To provide more context on the performance gap we are observing between Desktop and Embedded platforms, here is a comparison table based on our own benchmarks. We are using a 5-camera setup (1920x1536 raw resolution).

**Benchmark Comparison: RTX A4000 vs. AGX Orin 32G (FP16)**

| Platform | Mode | Inference (ms) | Preprocess (ms) | Total (ms) |
| :--- | :--- | :--- | :--- | :--- |
| **RTX A4000** | **Camera-LiDAR** | ~29 | ~12 | ~42 |
| | LiDAR Only | - | - | ~17 |
| **AGX Orin** | **Camera-LiDAR** | **~265** | **~10** | **~275** |
| | LiDAR Only | - | - | ~80 |

*(Note: The Orin inference time includes the overhead of `ImplicitGemm` plugins and the image backbone.)*

**Observations & Follow-up Questions:**

1. **The "Orin Gap":** While the A4000 performance is excellent (<50ms), the latency explosion on Orin (jumping from ~29ms to ~265ms for inference) is quite severe, almost a **10x slowdown**.
    * Since you mentioned that CUDA streams might saturate SMs, do you think the bottleneck on Orin is strictly **compute-bound** (Backbone calculation) or **memory-bound** (LSS/View Transform feature reading/writing)?
    * If it is compute-bound, this strongly reinforces the need for **INT8 quantization** on the image backbone.

2. **Path to <100ms on Orin:**
    Given the current architecture and your insights, do you see a viable path to bring the Camera-LiDAR fusion latency under **100ms** on Orin *without* switching to NVIDIA's native C++ kernels?
    * Would **Async Preprocessing** (as you planned) plus **INT8 Image Backbone** be enough to bridge this ~170ms gap?
    * Or should we also investigate optimizing the **View Transform (LSS)** operator specifically for TensorRT on Orin?

3. **INT8 Collaboration:**
    We are very excited about the INT8 PTQ work mentioned (cc @vividf). We will try to experiment with `trtexec --int8` using the binding names from tier4/AWML#98 as you suggested. If we manage to get a working INT8 backbone running on this node, we would be happy to share our findings/config.

Looking forward to the async preprocessing feature!

KSeangTan:
Hi @S-XiaoHa 
Thanks for the detailed report and discussions, as @SamratThapa120 mentioned, I believe the main reason is due to the saturation of SMs in ORIN, and thus, it should be compute-bound according to this table:

| Feature        | NVIDIA RTX A4000 (Desktop) | NVIDIA Jetson AGX Orin 64GB | NVIDIA Jetson AGX Orin 32GB |
|---------------|-----------------------------|-----------------------------|-----------------------------|
| SM Count      | 48                          | 16                          | 14                          |
| Architecture  | Ampere                      | Ampere                      | Ampere                      |
| CUDA Cores    | 6,144                       | 2,048                       | 1,792                       |


IMO, i dont think quantization can help reducing the latency a lot without sacrificing too much performance, especially, we might want to avoid do the quantization in the BEVFusion head since it's generally more sensitive to precision. 
And the parallelism we are working generally improves preprocessing time instead of inference time. 

My suggestion is that we should probably first replace the image backbone with a smaller image backbone, and consider to use [BEVPoolV2](https://arxiv.org/pdf/2211.17111) in the View Transform

PR #11469, opened 2025-10-09
fix(autoware_trajectory_optimizer): optimizer spline and orientation

danielsanchezaran:
## Description

This PR improves the trajectory optimization pipeline by refactoring the spline interpolation implementation and adjusting the optimizer execution order to address stability and performance issues.

**Key Changes:**

1. **Replaced trajectory spline interpolation library**: Switched from `autoware::trajectory::interpolator::AkimaSpline` to `motion_utils::resampleTrajectory` for better control over the resampling process. The new implementation uses a two-stage approach:
   - First, resample at 2x resolution with linear interpolation to avoid ill-conditioned spline fitting
   - Then apply Akima spline smoothing at the target resolution

2. **Added outlier detection for spline endpoints**: Introduced two new parameters to detect and handle unstable spline behavior at trajectory endpoints:
   - `spline_interpolation_max_yaw_discrepancy_deg` (2.5°): Maximum yaw deviation allowed
   - `spline_interpolation_max_distance_discrepancy_m` (5.0m): Maximum position deviation allowed

3. **Disabled elastic band trajectory smoothing**: Set `smooth_trajectories` to `false` due to jittering issues near path endpoints that cause violent steering behavior when approaching the end of the path.

4. **Disabled lateral acceleration limiting**: Set `limit_lateral_acceleration` to `false` due to buggy behavior causing unintended speed reduction. This will be addressed in a separate PR.

5. **Simplified trajectory point fixer**: Removed orientation adjustment logic (`insertOrientation` and `removeFirstInvalidOrientationPoints`) that was generating incorrect yaw values. The spline smoothing step now handles orientation calculation.

Before (Optimizer (in red) creates weird yaws for the output path of the diffusion planner (yellow): 
image

After (Optimizer YELLOW and Diffusion Planner RED):

https://github.com/user-attachments/assets/5b1e6cd6-cfd5-4080-a137-c72c70969f6b

**Bug related to lateral acceleration limiter** At around 45 seconds, the ego stops with zero velocity even though the output of the diffusion planner gives a non zero target velocity. The ego only moves after the lateral acceleration limit is removed. There is no reason for the target velocity to be set to 0 at that place, so there must be a bug on the plugin, so I will disable it for now.

https://github.com/user-attachments/assets/9932782f-db4e-4a26-a55c-1d9966e44c9b

## Related links

**Parent Issue:**

- N/A

## How was this PR tested?

- Testing in simulation/vehicle recommended to verify:
  - Trajectory smoothness improvements
  - No jittering at path endpoints
  - Correct trajectory orientation throughout the path
  - Proper handling of spline outliers

## Notes for reviewers

- The parameter `use_akima_spline_for_xy` in `motion_utils::resampleTrajectory` has counter-intuitive naming: setting it to `false` actually **enables** Akima spline interpolation, while `true` uses linear interpolation. This is a known issue in the library documentation.
- The two-stage resampling approach (linear first, then spline) helps avoid numerical instability with Akima splines on trajectories with irregular point spacing.
- The outlier detection parameters may need tuning based on real-world testing results.

## Interface changes

None.

#### Additions

| Change type | Parameter Name                                   | Type     | Default Value | Description                                                    |
|:------------|:-------------------------------------------------|:---------|:--------------|:---------------------------------------------------------------|
| Added       | `spline_interpolation_max_yaw_discrepancy_deg`   | `double` | `2.5`         | Maximum yaw deviation allowed for spline outlier detection     |
| Added       | `spline_interpolation_max_distance_discrepancy_m`| `double` | `5.0`         | Maximum position deviation allowed for spline outlier detection|

#### Modifications

| Version | Parameter Name       | Type   | Default Value | Description                                      |
|:--------|:---------------------|:-------|:--------------|:-------------------------------------------------|
| Old     | `smooth_trajectories`| `bool` | `true`        | Enable elastic band trajectory smoothing         |
| New     | `smooth_trajectories`| `bool` | `false`       | Disabled due to jittering issues at path endpoints|

## Effects on system behavior

- **Trajectory smoothness**: Improved path smoothness due to refined spline implementation, but elastic band smoothing is now disabled by default
- **Endpoint behavior**: Reduced steering oscillations when approaching the end of trajectories
- **Orientation accuracy**: More accurate trajectory headings throughout the path
- **Safety**: Better handling of spline interpolation edge cases through outlier detection

### DIFF
--- a/planning/autoware_trajectory_optimizer/src/utils.cpp
+++ b/planning/autoware_trajectory_optimizer/src/utils.cpp
@@ -245,14 +245,13 @@ void filter_velocity(
 
 bool validate_point(const TrajectoryPoint & point)
 {
-  return std::isfinite(point.longitudinal_velocity_mps) && std::isfinite(point.acceleration_mps2) &&
-         std::isfinite(point.pose.position.x) && std::isfinite(point.pose.position.y) &&
-         std::isfinite(point.pose.position.z) && std::isfinite(point.pose.orientation.x) &&
-         std::isfinite(point.pose.orientation.y) && std::isfinite(point.pose.orientation.z) &&
-         std::isfinite(point.pose.orientation.w) && !std::isnan(point.pose.position.x) &&
-         !std::isnan(point.pose.position.y) && !std::isnan(point.pose.position.z) &&
-         !std::isnan(point.pose.orientation.x) && !std::isnan(point.pose.orientation.y) &&
-         !std::isnan(point.pose.orientation.z) && !std::isnan(point.pose.orientation.w);
+  auto is_valid = [](auto value) { return std::isfinite(value) && !std::isnan(value); };
+
+  return is_valid(point.longitudinal_velocity_mps) && is_valid(point.acceleration_mps2) &&
+         is_valid(point.pose.position.x) && is_valid(point.pose.position.y) &&
+         is_valid(point.pose.position.z) && is_valid(point.pose.orientation.x) &&
+         is_valid(point.pose.orientation.y) && is_valid(point.pose.orientation.z) &&
+         is_valid(point.pose.orientation.w);
 }
 
 void apply_spline(TrajectoryPoints & traj_points, const TrajectoryOptimizerParams & params)


planning/autoware_trajectory_optimizer/src/trajectory_optimizer_plugins/trajectory_spline_smoother.cpp
```diff
@@ -24,9 +25,10 @@ void TrajectorySplineSmoother::optimize_trajectory(
   TrajectoryPoints & traj_points, [[maybe_unused]] const TrajectoryOptimizerParams & params)
 {
   // Apply spline to smooth the trajectory
-  if (params.use_akima_spline_interpolation) {
-    utils::apply_spline(traj_points, params);
+  if (!params.use_akima_spline_interpolation) {
+    return;
   }
+  utils::apply_spline(traj_points, params);
```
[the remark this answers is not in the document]
go-sakayori: Refer to this line https://github.com/autowarefoundation/autoware_core/blob/277925b1b9d340ce44088c706a91828b23ce24df/common/autoware_motion_utils/src/resample/resample.cpp#L77-L78

planning/autoware_trajectory_optimizer/src/trajectory_optimizer_plugins/trajectory_spline_smoother.cpp
```diff
@@ -14,6 +14,7 @@
 
 #include "autoware/trajectory_optimizer/trajectory_optimizer_plugins/trajectory_spline_smoother.hpp"
 
+#include "autoware/motion_utils/resample/resample.hpp"
```
go-sakayori: nits: Make the include statements consistent
```suggestion
#include <autoware/motion_utils/resample/resample.hpp>
```

planning/autoware_trajectory_optimizer/src/utils.cpp
```diff
@@ -14,15 +14,13 @@
 
 #include "autoware/trajectory_optimizer/utils.hpp"
 
-#include "autoware/trajectory/interpolator/akima_spline.hpp"
-#include "autoware/trajectory/interpolator/interpolator.hpp"
-#include "autoware/trajectory/pose.hpp"
-#include "autoware/trajectory/trajectory_point.hpp"
+#include "autoware/motion_utils/resample/resample.hpp"
```
go-sakayori: nits: Make the include statements consistent
```suggestion
#include <autoware/motion_utils/resample/resample.hpp>
```

planning/autoware_trajectory_optimizer/src/utils.cpp
```diff
@@ -14,15 +14,13 @@
...
 #include "autoware/trajectory_optimizer/trajectory_optimizer_structs.hpp"
 
 #include <autoware/motion_utils/trajectory/trajectory.hpp>
 #include <autoware_utils/geometry/geometry.hpp>
 #include <autoware_utils/math/normalization.hpp>
```
go-sakayori: nits:
```suggestion
#include <autoware_utils_math/normalization.hpp>
```
If you are going to fix it, please fix https://github.com/autowarefoundation/autoware_universe/blob/main/planning/autoware_trajectory_optimizer/src/utils.cpp#L405 too

planning/autoware_trajectory_optimizer/src/utils.cpp
```diff
@@ -269,58 +252,52 @@ bool validate_point(const TrajectoryPoint & point)
          std::isfinite(point.pose.orientation.w) && !std::isnan(point.pose.position.x) &&
          !std::isnan(point.pose.position.y) && !std::isnan(point.pose.position.z) &&
          !std::isnan(point.pose.orientation.x) && !std::isnan(point.pose.orientation.y) &&
-         !std::isnan(point.pose.orientation.z) && !std::isnan(point.pose.orientation.w) &&
-         !std::isnan(point.longitudinal_velocity_mps) && !std::isnan(point.acceleration_mps2);
+         !std::isnan(point.pose.orientation.z) && !std::isnan(point.pose.orientation.w);
```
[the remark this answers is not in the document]
danielsanchezaran: Those conditions were repeated

planning/autoware_trajectory_optimizer/src/utils.cpp
```diff
@@ -269,58 +252,62 @@ bool validate_point(const TrajectoryPoint & point)
...
```
go-sakayori:
```suggestion
```
I think std::isfinite would be enough. Same for `point.pose.position`

danielsanchezaran: I would like to leave it just because we dont know what output we can be getting from a model. Better safe that sorry

go-sakayori: In that case I think you can keep this line as well?
https://github.com/autowarefoundation/autoware_universe/pull/11469/files#diff-913dad8903cf2f6b6735c7722d9df71e00daa1b9dc21ce7ac5c07c918dc9e615L273

planning/autoware_trajectory_optimizer/src/trajectory_optimizer_plugins/trajectory_spline_smoother.cpp
danielsanchezaran: Done in https://github.com/autowarefoundation/autoware_universe/pull/11469/commits/6981725b0be37ad1de215dc34d01f4405242b0d3

planning/autoware_trajectory_optimizer/src/utils.cpp
danielsanchezaran: done in https://github.com/autowarefoundation/autoware_universe/pull/11469/commits/6981725b0be37ad1de215dc34d01f4405242b0d3

planning/autoware_trajectory_optimizer/src/utils.cpp
danielsanchezaran: Done here https://github.com/autowarefoundation/autoware_universe/pull/11469/commits/ebfd6abf03ce2a20092a5498e9c5597595cecd9b I simplified the logic with a lambda, I thought I had deleted repeated lines, but they were not repeated

PR #11476, opened 2025-10-10
feat(start planner): replace zero velocity with stop line on candidate path

zulfaqar-azmi-t4:
## Description

This PR prevents sudden deceleration during `RUNNING` -> `WAITING_APPROVAL` transition.

Currently, the start planner sets the velocity of candidate path to zero:

```c++
for (auto &p : stop_path.points) {
  p.point.longitudinal_velocity_mps = 0.0;
}
```

In Autoware, the usual practice is to add a stop point instead.
This PR replaces the zero-velocity assignment with a stop line.

It also includes some minor refactoring.

## Related links

[TIER IV internal issue link](https://tier4.atlassian.net/browse/T4DEV-9639)

## How was this PR tested?

[TIER IV Internal Link - ControlDLR Test](https://evaluation.tier4.jp/evaluation/reports/31a836a8-51fb-56f7-8c17-d48b5a6d5a1e?project_id=prd_jt)
[TIER IV Internal Link - FuncVerification Test](https://evaluation.tier4.jp/evaluation/reports/cd1cadf8-adee-5fba-8669-6c0aef5a8e6d?project_id=prd_jt)
[TIER IV Internal Link - CommonScenario_Basic Test](https://evaluation.tier4.jp/evaluation/reports/dfc17712-a85d-5ae0-bea2-d9acff760634?project_id=prd_jt)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_path_planner/autoware_behavior_path_planner_common/src/utils/parking_departure/utils.cpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_planner_common/src/utils/parking_departure/utils.cpp
@@ -31,6 +31,12 @@ std::optional<double> calcFeasibleDecelDistance(
   std::shared_ptr<const PlannerData> planner_data, const double acc_lim, const double jerk_lim,
   const double target_velocity)
 {
+  const auto v_now = planner_data->self_odometry->twist.twist.linear.x;
+
+  if (v_now < target_velocity) {
+    return 0.0;
+  }
+
   return calc_feasible_decel_distance(planner_data, -acc_lim, jerk_lim, target_velocity);
 }
 
--- a/planning/behavior_path_planner/autoware_behavior_path_planner_common/test/test_parking_departure_utils.cpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_planner_common/test/test_parking_departure_utils.cpp
@@ -252,7 +252,7 @@ TEST(BehaviorPathPlanningParkingDepartureUtil, generateFeasibleStopPath)
 
   // condition: valid condition
   maximum_jerk = 5.0;
-  maximum_deceleration = -3.0;
+  maximum_deceleration = 3.0;
   stop_path =
     generateFeasibleStopPath(path, planner_data, stop_pose, maximum_deceleration, maximum_jerk);
   size_t i = 0;


planning/behavior_path_planner/autoware_behavior_path_planner_common/src/utils/utils.cpp
```diff
@@ -1500,4 +1502,55 @@ bool checkOriginalGoalIsInShoulder(const std::shared_ptr<RouteHandler> & route_h
   const Pose & goal_pose = route_handler->getOriginalGoalPose();
   return !route_handler->getShoulderLaneletsAtPose(goal_pose).empty();
 }
+
+std::optional<double> calc_feasible_decel_distance(
```
kosuke55, line 1506: can we use `autoware::behavior_path_planner::utils::parking_departure::calcFeasibleDecelDistance;` ?

planning/behavior_path_planner/autoware_behavior_path_planner_common/src/utils/utils.cpp
```diff
@@ -1500,4 +1502,55 @@ bool checkOriginalGoalIsInShoulder(const std::shared_ptr<RouteHandler> & route_h
...
+    v_now, target_velocity, a_now, acc_lim, jerk_lim, -1.0 * jerk_lim);
+
+  if (min_stop_distance) {
+    return std::max(*min_stop_distance, 0.0);
+  }
+
+  return std::nullopt;
+}
+
+PoseWithDetailOpt insert_feasible_stop_point(
```
kosuke55, line 1531: can we use `autoware::behavior_path_planner::utils::parking_departure::generateFeasibleStopPath` ?

zulfaqar-azmi-t4: It’s basically the same function, with only minor refactoring.

In the original function:

```cpp
std::optional<PathWithLaneId> generateFeasibleStopPath(
  PathWithLaneId & current_path, std::shared_ptr<const PlannerData> planner_data,
  PoseWithDetailOpt & stop_pose, const double maximum_deceleration,
  const double maximum_jerk, const std::string & stop_reason)
```

Both `current_path` and `stop_pose` are passed by reference, and then `current_path` is returned again.
I think this is a bit confusing. Returning the stop pose instead would make it cleaner.

Also I feel like it is more beneficial to have this function in utils instead.

planning/behavior_path_planner/autoware_behavior_path_planner_common/src/utils/utils.cpp
zulfaqar-azmi-t4: I needed this function for `insert_feasible_stop_point`, and I think it makes more sense to have this function in utils instead, in case other modules would like to use it.

PR #11477, opened 2025-10-10
feat(start planner): support transit from running state to waiting approval state when unsafe

zulfaqar-azmi-t4:
## Description

To support force approval, the RTC needs an unsafe candidate path.
However, the start planner only transitions to the running state, so no candidate path is available at that point.

Sometimes, the operator may override Autoware’s decision by pressing **Force Approval**.
In the start planner’s case, if it inserts a stop point but the operator believes it’s safe to proceed, they may want to cancel that stop.

This PR allows the start planner to transition from the running state back to the waiting-for-approval state when an unsafe condition is detected due to dynamic objects.

| Before | After |
|:------:|:-----:|
|  |  |

## Related links

[TIER IV internal issue link](https://tier4.atlassian.net/browse/T4DEV-9639)

#### 1. Driving forward

https://github.com/user-attachments/assets/a58c062a-dd7a-4bb5-9c50-ccc5099c3c24

#### 2. Driving backward

https://github.com/user-attachments/assets/1cf02f24-223b-42fd-ba04-f38833163f06

### Internal evaluator

[TIER IV Internal Link - ControlDLR Test](https://evaluation.tier4.jp/evaluation/reports/90e78a67-e34c-59db-9a50-f9638c24283a?project_id=prd_jt)
[TIER IV Internal Link - FuncVerification Test](https://evaluation.tier4.jp/evaluation/reports/883047d7-4e5e-57de-8a6c-8ff1832c1a14?project_id=prd_jt)
[TIER IV Internal Link - CommonScenario_Basic Test](https://evaluation.tier4.jp/evaluation/reports/1a25a3ae-e5be-5598-b667-0ecc03549158?project_id=prd_jt)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_path_planner/autoware_behavior_path_start_planner_module/src/start_planner_module.cpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_start_planner_module/src/start_planner_module.cpp
@@ -860,7 +860,7 @@ BehaviorModuleOutput StartPlannerModule::plan()
   path_candidate_ = std::make_shared<PathWithLaneId>(getFullPath());
   path_reference_ = std::make_shared<PathWithLaneId>(getPreviousModuleOutput().reference_path);
 
-  if (!status_.prev_approved_path) {
+  if (!status_.prev_approved_path && status_.driving_forward) {
     status_.prev_approved_path = std::make_shared<PathWithLaneId>(output.path);
   }
 


planning/behavior_path_planner/autoware_behavior_path_start_planner_module/src/start_planner_module.cpp
```diff
@@ -1145,9 +1115,82 @@ PathWithLaneId StartPlannerModule::getCurrentPath() const
...
+  if (status_.is_safe_dynamic_objects) {
+    return getCurrentPath();
+  }
+
+  waitApproval();
+  removeRTCStatus();
+
+  auto current_path = getCurrentPath();
+  const auto stop_path =
+    autoware::behavior_path_planner::utils::parking_departure::generateFeasibleStopPath(
```
kosuke55, line 1165: minor: 
Is it better to use your new function `insert_feasible_stop_point` (impletmented in https://github.com/autowarefoundation/autoware_universe/pull/11476/files) for uniformity?

zulfaqar-azmi-t4: yes, actually it is. But because it is a part of the original logic, so i didn't make the changes yet. I think making another PR to refactor it is a better option.

PR #11483, opened 2025-10-14
feat(autoware_trajectory_optimizer): copy nearest point orientation from original trajectory

danielsanchezaran:
## Description

This PR adds an optional feature to preserve orientation values from the original trajectory after spline interpolation in the trajectory optimizer. This is particularly useful for planners like the Diffusion Planner that output trajectories with accurate orientation values that are **not** derived from position-to-position vectors.

**Problem:** The Diffusion Planner outputs spatially noisy trajectories but with smooth, accurate orientation values. When spline interpolation is applied, the default behavior calculates new orientation values based on the tangent vectors between consecutive points. For noisy trajectories, this produces noisy yaw outputs that degrade trajectory quality.

**Solution:** A new parameter `spline_copy_original_orientation` allows the optimizer to copy orientation values from the nearest point in the original trajectory after spline smoothing, preserving the planner's smooth orientation profile while still benefiting from spatial smoothing.

## Related links

**Parent Issue:**

None - this is a proactive enhancement for noisy trajectory handling.

## How was this PR tested?

Tested in Planning Simulator (Psim) with Diffusion Planner trajectories, validating that:
- Orientation values are preserved from the original noisy trajectory
- Spatial smoothing via spline interpolation still functions correctly

Comparison of the feature OFF vs ON

[Screencast from 2025年10月14日 09時16分35秒.webm](https://github.com/user-attachments/assets/ba22cec2-05a3-4b4e-a3c5-de9a665f57bf)

## Notes for reviewers

- The feature is **enabled by default** (`spline_copy_original_orientation: true`)
- This is applied in `apply_spline()` at `utils.cpp:323-325` after all spline resampling is complete
- **Important:** Using the Elastic Band (EB) smoother (`smooth_trajectories: true`, disabled by default) will effectively undo the orientation copying since EB recalculates orientation from the smoothed path geometry

#### Additions and removals

| Change type | Parameter Name                     | Type   | Default Value | Description                                                                                          |
|:------------|:-----------------------------------|:-------|:--------------|:-----------------------------------------------------------------------------------------------------|
| Added       | `spline_copy_original_orientation` | `bool` | `true`        | Flag to indicate if the orientation from the original trajectory should be copied to the spline-interpolated trajectory |

## Effects on system behavior

When `spline_copy_original_orientation` is enabled:
- Orientation values will be copied from the nearest point in the pre-spline trajectory rather than computed from spline tangent vectors
- This produces smoother yaw profiles for planners that output noisy spatial trajectories with accurate orientation
- Spatial path smoothing is unaffected - only orientation handling changes

### DIFF
--- a/planning/autoware_trajectory_optimizer/schema/trajectory_optimizer.schema.json
+++ b/planning/autoware_trajectory_optimizer/schema/trajectory_optimizer.schema.json
@@ -108,7 +108,7 @@
         },
         "spline_copy_original_orientation": {
           "type": "boolean",
-          "default": false,
+          "default": true,
           "description": "Flag to indicate if the orientation from the original trajectory should be copied to the spline-interpolated trajectory"
         },
         "jerk_filter_params": {
--- a/planning/autoware_trajectory_optimizer/include/autoware/trajectory_optimizer/utils.hpp
+++ b/planning/autoware_trajectory_optimizer/include/autoware/trajectory_optimizer/utils.hpp
@@ -65,6 +65,16 @@ void smooth_trajectory_with_elastic_band(
 bool validate_point(const TrajectoryPoint & point);
 
 /**
+ * @brief Copies the orientation of the input trajectory points to the output trajectory points.
+ * @param input_trajectory The input trajectory points.
+ * @param output_trajectory The output trajectory points.
+ * @param params The parameters for trajectory optimization.
+ */
+void copy_trajectory_orientation(
+  const TrajectoryPoints & input_trajectory, TrajectoryPoints & output_trajectory,
+  const TrajectoryOptimizerParams & params);
+
+/**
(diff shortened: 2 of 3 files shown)


planning/autoware_trajectory_optimizer/schema/trajectory_optimizer.schema.json
```diff
@@ -106,6 +106,11 @@
           "default": true,
           "description": "Flag used to indicate if the ego's trajectory should be extended backward"
         },
+        "spline_copy_original_orientation": {
+          "type": "boolean",
+          "default": false,
```
go-sakayori:
```suggestion
          "default": true,
```

danielsanchezaran: Fixed

PR #11494, opened 2025-10-15
feat(autoware_traffic_light_arbiter): priority switch

dkoldaev:
## Description

In some cases it is advisable to use perception as the priority source of traffic light data. This PR replaces the flag for prioritizing external signal with an open choice of the priority source. 

Note that this changes won't affect the predictions and will only be used to judge the current state of the elements. 

## Related links

**Parent Issue:**

**Private Links:**

- [TIER IV internal link](https://tier4.atlassian.net/browse/T4DEV-32430)

## How was this PR tested?

Unit test is added that tests the option to prioritize the perception over external sources

## Notes for reviewers

None.

## Interface changes

external_priority boolean flag replace with string source_priority

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_simple_object_merger/src/simple_object_merger_base.cpp
+++ b/perception/autoware_simple_object_merger/src/simple_object_merger_base.cpp
@@ -118,7 +118,7 @@ typename ObjsMsgType::SharedPtr SimpleObjectMergerBase<ObjsMsgType>::getTransfor
   typename ObjsMsgType::ConstSharedPtr objects, const std::string & target_frame_id,
   geometry_msgs::msg::TransformStamped::ConstSharedPtr transform)
 {
-  typename ObjsMsgType::SharedPtr output_objects = std::const_pointer_cast<ObjsMsgType>(objects);
+  typename ObjsMsgType::SharedPtr output_objects = std::make_shared<ObjsMsgType>(*objects);
 
   if (objects->header.frame_id == target_frame_id) {
     return output_objects;
--- a/planning/behavior_path_planner/autoware_behavior_path_goal_planner_module/src/goal_planner_module.cpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_goal_planner_module/src/goal_planner_module.cpp
@@ -1063,7 +1063,9 @@ BehaviorModuleOutput GoalPlannerModule::plan()
         getLogger(), *clock_, 5000, " [pull_over] plan() is called without valid context_data");
     } else {
       auto & context_data_mut = context_data_.value();
-      return planPullOver(context_data_mut);
+      const auto output = planPullOver(context_data_mut);
+      set_longitudinal_planning_factor(output.path);
+      return output;
     }
   }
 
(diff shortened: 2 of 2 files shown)


dkoldaev:
@YoshiRi could you please review this PR when you have time?

perception/autoware_traffic_light_arbiter/src/signal_match_validator.cpp
```diff
@@ -155,10 +155,11 @@ std::unordered_set<lanelet::Id> create_signal_id_set(
...
 
-// Returns the signal with the highest confidence elements, considering a external priority
+// Returns the signal with the highest confidence elements, considering source priority
 TrafficSignal get_highest_confidence_signal(
   const std::optional<TrafficSignal> & perception_signal,
-  const std::optional<TrafficSignal> & external_signal, const bool external_priority)
+  const std::optional<TrafficSignal> & external_signal,
+  const autoware::traffic_light::SourcePriority source_priority)
 {
   // Returns the existing signal if only one of them exists
```
YoshiRi, line 163: [imho]
Though I know there is escape before the usage of this function , I personally want protection or assertion to prevent case which both signals are nullopt. `if (!perception_signal && !external_signal) ...`

perception/autoware_traffic_light_arbiter/src/signal_match_validator.cpp
```diff
@@ -168,9 +169,16 @@ TrafficSignal get_highest_confidence_signal(
...
-  if (external_priority) {
-    return *external_signal;
+  // Apply source priority
+  switch (source_priority) {
+    case autoware::traffic_light::SourcePriority::EXTERNAL:
+      return *external_signal;
+    case autoware::traffic_light::SourcePriority::PERCEPTION:
+      return *perception_signal;
+    case autoware::traffic_light::SourcePriority::CONFIDENCE:
+    default:
```
YoshiRi: [imho]
Using a default case here makes it harder to detect newly added SourcePriority values, since they would silently fall through to the CONFIDENCE branch. It’s safer to remove the default and handle each enum explicitly, so that future extensions trigger a compile-time error.

dkoldaev: @YoshiRi thank you for your comment, it makes perfect sense! Here (https://github.com/autowarefoundation/autoware_universe/pull/11494/commits/3207ed26d344aac36fe38b168b856717114e9524) I removed the default and used builtin_unreachable to raise compiler error in case some of the cases are not handled well. 
Note that the parsing will still default to confidence in case an unknown value is added without proper handling.

perception/autoware_traffic_light_arbiter/src/signal_match_validator.cpp
dkoldaev: Thank you! This makes sense. However, I think this should be done in another PR. Assertion won't make much difference in release build and proper protection will require using optional in the return type or something of that kind.

dkoldaev:
@YoshiRi can you please check the update to the code I made? and also the corresponding [launcher PR](https://github.com/autowarefoundation/autoware_launch/pull/1672). Thank you!

perception/autoware_traffic_light_arbiter/src/signal_match_validator.cpp
YoshiRi: Agree. Set this conversation as resolved

PR #11498, opened 2025-10-15
feat(obstacle_slow_down): add obstacle filtter

yuki-takagi-66:
## Description
Previously, the obstacle_slow_down module determined whether to slow down for an object based on the obstacle_filtering.min/max_lat_margin parameters and a satisfaction count for this condition.

While this logic worked well when the range between min_lat_margin and max_lat_margin was narrow, it caused unnecessary slow-downs in two specific cases: when min_lat_margin was set to a small value (mainly due to https://github.com/autowarefoundation/autoware_launch/pull/1638) and when max_lat_margin was set to a large value. Real-world driving tests confirmed that this issue was particularly noticeable, causing unnecessary deceleration for cross-traffic vehicles.

To address this, we are introducing two new filtering logic:

1. Exclude Objects on the Path's Center Line: An object is now excluded from the slow-down process if it overlaps with the center line of the path. This change primarily addresses issues arising from a small min_lat_margin. We determined that it is not necessary to trigger obstacle_slow_down for objects located directly on the path's center line.
2. Filter by Low Lateral Velocity: Only objects with low lateral velocity are now considered targets for slowing down. This change is mainly to handle cases with a large max_lat_margin. We concluded that responding to objects that has high lateral speeds is the responsibility of the obstacle_stop module.

## Related links
launch PR: https://github.com/autowarefoundation/autoware_launch/pull/1673

## How was this PR tested?
psim and tier4 scenario test

## Notes for reviewers
The logic for clearing the count still has room for improvement.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/motion_velocity_planner/autoware_motion_velocity_obstacle_slow_down_module/src/obstacle_slow_down_module.cpp
+++ b/planning/motion_velocity_planner/autoware_motion_velocity_obstacle_slow_down_module/src/obstacle_slow_down_module.cpp
@@ -20,6 +20,7 @@
 #include <autoware/motion_utils/distance/distance.hpp>
 #include <autoware/motion_utils/marker/marker_helper.hpp>
 #include <autoware/motion_utils/marker/virtual_wall_marker_creator.hpp>
+#include <autoware/motion_utils/resample/resample.hpp>
 #include <autoware/motion_utils/trajectory/trajectory.hpp>
 #include <autoware/signal_processing/lowpass_filter_1d.hpp>
 #include <autoware_utils/geometry/geometry.hpp>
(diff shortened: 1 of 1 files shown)


planning/motion_velocity_planner/autoware_motion_velocity_obstacle_slow_down_module/src/obstacle_slow_down_module.cpp
```diff
@@ -510,16 +510,48 @@ ObstacleSlowDownModule::create_slow_down_obstacle_for_predicted_object(
...
   }
 
+  const auto obstacle_poly = autoware_utils::to_polygon2d(
+    object->predicted_object.kinematics.initial_pose_with_covariance.pose,
+    object->predicted_object.shape);
+
+  // Create linestring from trajectory points
+  const auto traj_line = [&]() {
+    constexpr double resample_interval = 2.0;  // [m]
+    const auto resample_traj = autoware::motion_velocity_planner::utils::resample_trajectory_points(
```
maxime-clem: `resample_trajectory_points()` will resample all fields of the trajectory (orientation, velocity, etc), but you only need the X,Y points.
Therefore, I think you can use the `autoware::motion_utils::resamplePoseVector()` function instead.

yuki-takagi-66: Thank you. I'll use autoware::motion_utils::resamplePointVector()

maxime-clem (approved):

PR #11499, opened 2025-10-15
fix(autoware_pointcloud_preprocessor): prevent race condition

kaancolak:
## Description

We are using previous version of the concatenate_data module (software version v4.2). During one of the tests, a segmentation fault occurred, likely related to a potential race condition. Although an older version of the package is in use, a detailed investigation of the current implementation was performed after observing this issue.

Screenshot from 2025-10-14 15-37-07

The point cloud nodes run inside a  point cloud multi-threaded component container that utilizes the container’s multi-threaded executor. Callbacks (timers, subs, etc.) may run in parallel.

The possibility of occurrence is very low, but in the current implementation, the following cases might cause a race condition:

1. `topic_to_cloud_map_` Data Race
**Problem:** Multiple threads accessing the pointcloud map concurrently:
- **Thread A**: Adding pointclouds via `process_pointcloud()`
- **Thread B**: Reading/copying map in `concatenate_callback()` (timer thread)
- **Result**: Map corruption, iterator invalidation, undefined behavior

2. Collector Status Race Condition
**Problem:** `status_` member modified without synchronization:
- **Thread A:** Checking status in main callback thread
- **Thread B:** Changing status in timer callback thread
- **Result:** Inconsistent status reads, duplicate processing

3. `cloud_collectors_` List Race Condition
**Problem:** Collector list accessed by multiple callbacks without protection:
- **Thread A:** `cloud_callback()` searching/modifying collectors
- **Thread B:** `manage_collector_list()` resetting finished collectors
- **Result:** List corruption, memory access violations

4. Initialization Race Condition
**Problem:** Non-thread-safe collector list initialization:
- Multiple threads could initialize collectors simultaneously
- **Result:** Duplicate initialization, resource leaks

5. `collector_info_` Access Race
**Problem:** Shared pointer accessed without synchronization:
- **Thread A:** Setting collector info via` set_info()`
- **Thread B:** Reading info via `get_info()`
- **Result:** Use-after-free, null pointer dereference

Relevant mutex/lock mechanisms have been added, and logging operations have been moved outside the lock to prevent potential deadlocks.

I may have missed some parts, could you please give me feedback after reviewing it?

## Related links

**Parent Issue:**

- Link

- [TIERIV internal link](https://tier4.atlassian.net/browse/ISSJP-3236)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/concatenate_data/cloud_collector.hpp
+++ b/sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/concatenate_data/cloud_collector.hpp
@@ -59,7 +59,8 @@ public:
   void set_info(std::shared_ptr<CollectorInfoBase> collector_info);
   [[nodiscard]] std::shared_ptr<CollectorInfoBase> get_info() const;
   void show_debug_message(
-    const std::unordered_map<std::string, typename MsgTraits::PointCloudMessage::ConstSharedPtr> & topic_to_cloud_map,
+    const std::unordered_map<std::string, typename MsgTraits::PointCloudMessage::ConstSharedPtr> &
+      topic_to_cloud_map,
     const std::shared_ptr<CollectorInfoBase> & collector_info);
   void reset();
 
--- a/sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/concatenate_data/cloud_collector.ipp
+++ b/sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/concatenate_data/cloud_collector.ipp
@@ -77,10 +77,10 @@ void CloudCollector<MsgTraits>::process_pointcloud(
 {
   bool should_concatenate = false;
   bool should_warn_duplicate = false;
-  
+
   {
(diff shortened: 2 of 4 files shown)


veqcc:
@kaancolak 
Hi, thank you for investigating the bug and contributing!

I have a question:
How did you conclude that this is a concurrent bug? As you say, these variables are acccessed by many threads, but the ROS 2 executor usually runs the callbacks mutually execlusively.

kaancolak:
@veqcc - san, yes, I also expected the callbacks to behave mutually exclusively, but after observing the shared issue, I got confused.
If you have time, please take a quick look at the internal issue.
When the timer callback of the DiagnosticUpdater tries to read shared variables that are also accessed by another callback, Autoware crashes with a segmentation fault.

The modules share a callback function with DiagnosticUpdater via a function pointer — could this be causing concurrent access behavior?

veqcc:
@kaancolak 
> We are using previous version of the concatenate_data module (software version v4.2).

Could you tell me the detail of your software version? (which software is v4.2?)

kaancolak:
We talked with @veqcc -san. I will close this issue because previous versions of concat/agnocast mt could cause a concurrency bug, but after the updates, this issue now seems outdated

PR #11502, opened 2025-10-16
feat(obstacle_slow_down): update velocity calclation

yuki-takagi-66:
## Description
**Feature Change: Modified the Calculation Formula for the Deceleration Point**
Previously, the calculated deceleration point for an approaching vehicle would move closer to the ego vehicle with each calculation cycle. This occurred because as the ego vehicle began to slow down, subsequent calculations would use this new, lower speed, pushing the required deceleration point closer. This often led to abrupt deceleration or a failure to slow down in time.

To address this, the logic now anticipates the ego vehicle's speed drop from its current value to the target speed. From this potential speed range, it selects the speed that establishes a stable, worst-case deceleration point from the outset.

This change mitigates the issue of the deceleration point creeping closer and now creeping far. Note: The deceleration point may still shift due to the acceleration or deceleration of other vehicles.

**Fix: Used the Correct Parameter**
Corrected an instance where common_param_.min_accel was used instead of the appropriate parameter, slow_down_planning_param_.slow_down_min_acc. This did not affect behavior when using default parameters, as both have the same value.

**Fix/Improvement: Improved the Value-Holding Mechanism for Deceleration Constraint Logic**
Previously, the value-holding mechanism within the deceleration-constrained speed planning logic could not account for shifts in the deceleration point's position. A formula has been added to adjust the previously calculated maximum deceleration speed based on changes in the deceleration point's location.

**Fix: Fixed Bug in Low-Pass Filter for Lateral Distance Calculation**
Fixed an issue where a section intended to function as a low-pass filter (LPF) was incorrectly outputting a simple weighted average of the previous and current values. This was caused by internal values not being updated correctly. Although this is a bug fix, it is also a breaking change, as existing default parameter values will no longer function appropriately. Consequently, the parameter name has been changed.

## Related links
launch PR: https://github.com/autowarefoundation/autoware_launch/pull/1674

## How was this PR tested?
psim and tier4 scenario test

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/motion_velocity_planner/autoware_motion_velocity_obstacle_slow_down_module/src/obstacle_slow_down_module.cpp
+++ b/planning/motion_velocity_planner/autoware_motion_velocity_obstacle_slow_down_module/src/obstacle_slow_down_module.cpp
@@ -860,7 +860,8 @@ std::vector<SlowdownInterval> ObstacleSlowDownModule::plan_slow_down(
     new_prev_slow_down_output.push_back(
       SlowDownOutput{
         obstacle.uuid, slow_down_traj_points, slow_down_start_idx, slow_down_end_idx,
-        stable_slow_down_vel, feasible_slow_down_vel, obstacle.dist_to_traj_poly, obstacle_motion});
+        stable_slow_down_vel, feasible_slow_down_vel, obstacle.stable_dist_to_traj_poly.value(),
+        obstacle_motion});
   }
 
   // update prev_slow_down_output_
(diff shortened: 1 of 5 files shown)


planning/motion_velocity_planner/autoware_motion_velocity_obstacle_slow_down_module/src/obstacle_slow_down_module.cpp
```diff
@@ -1031,8 +1039,14 @@ ObstacleSlowDownModule::calculate_distance_to_slow_down_with_constraints(
       // NOTE: If longitudinal controllability is not good, one_shot_slow_down_vel may be getting
       // larger since we use actual ego's velocity and acceleration for its calculation.
       //       Suppress one_shot_slow_down_vel getting larger here.
+      const double start_point_diff =
+        filtered_dist_to_slow_down_start -
+        motion_utils::calcSignedArcLength(traj_points, 0, prev_output->start_point->position);
+      const double prev_feasible_slow_down_vel = std::sqrt(
+        std::pow(prev_output->feasible_target_vel, 2) +
+        2 * slow_down_planning_param_.slow_down_min_acc * start_point_diff);
```
Kotakku: Please add a safeguard like `std::max(0.0, ...)` to prevent the argument of `std::sqrt` from becoming negative.

yuki-takagi-66: @Kotakku 
Thank you. I've updated

Kotakku (approved):
LGTM

Kotakku (approved):

PR #11524, opened 2025-10-20
fix: resolve clock type mismatch in tf2 lookups with simulation time - part2 

ralwing:
## Description
Replace rclcpp::Time(0) with tf2::TimePointZero in lookupTransform calls to fix clock type conflicts when using simulation time.

This is a second part of changes in Autoware Universe. 

## Related links
First part of changes:
https://github.com/autowarefoundation/autoware_universe/pull/11523

## Effects on system behavior

None.


ogrex (approved):

technolojin (approved):

control/autoware_pure_pursuit/include/autoware/pure_pursuit/util/tf_utils.hpp
```diff
@@ -36,12 +36,13 @@ namespace autoware::pure_pursuit
 namespace tf_utils
 {
 rclcpp::Logger logger = rclcpp::get_logger(TF_UTILS_LOGGER);
+
+template <typename... TransformArgs>
 inline boost::optional<geometry_msgs::msg::TransformStamped> getTransform(
-  const tf2_ros::Buffer & tf_buffer, const std::string & from, const std::string & to,
-  const rclcpp::Time & time, const rclcpp::Duration & duration)
+  const tf2_ros::Buffer & tf_buffer, TransformArgs &&... args)
```
yukkysaito: This change increases flexibility, but I have some concerns.
Because the from and to arguments have the same type, even if they are swapped, the code would still compile.
For readability and safety, I think it’s better to define the arguments explicitly instead of using a variadic template.

ralwing: This update makes it possible to call all overloaded `tf_buffer` methods, providing future-proofing against signature changes. Because the method is short and defined in the `.hpp` file, the required arguments can be inferred directly from the `tf_buffer` implementation.

Let me know if you would prefer me to restore the explicit arguments instead.

yukkysaito: Thanks for the clarification that makes sense.
I agree that the template version improves flexibility and makes it future-proof. :+1: 
However, considering that the from/to mix-up can easily happen and is difficult to detect at compile time, I’d still prefer to keep the explicit argument version for readability and safety.

Especially since this code is part of Autoware, where multiple contributors maintain the utilities, I believe it’s safer for the team to keep explicit arguments as a general rule.

If you’re okay with that, I think reverting to the explicit argument form would be clearer for maintainers.

ralwing: ok i restored old version and added a new overload for [here](https://github.com/autowarefoundation/autoware_universe/pull/11524/commits/86e099e455324684ab539c08e413a9be24fa55f0)

ralwing:
@TakaHoribe kindly ping :-)

shmpwk:
@takayuki5168 Could you review this PR as a code owner?

ralwing:
@yukkysaito  @TakaHoribe i'm kindly pinging :-)

PR #11528, opened 2025-10-21
fix(run_out): always ignore non-target objects

maxime-clem:
## Description

In the `run_out` module, objects can be ignored based on multiple criteria, including their classification label.
For consistent and safe decision making, once a collision has been detected with an object, the module does not ignore it in subsequent iterations.
This PR adjusts this behavior to always ignore an object when its classification label changes to a non-target label. 

## Related links

**Private Links:**

- [TIER IV internal link](https://tier4.atlassian.net/browse/RT1-11406)

## How was this PR tested?

Psim
Reproducer

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

The `run_out` module no longer checks collision with non-target objects, even if in a previous iteration the object was labelled as a target and was predicted to collide with ego.


zulfaqar-azmi-t4 (approved):
LTGM!!

Note:
I'm curious if the target label chatters (bicycle ↔ non VRU), but if we want to consider something like hysteresis it will complicate matters a lot.

PR #11536, opened 2025-10-21
fix(autoware_crosswalk_traffic_light_estimator): remove old traffic light groups

MasatoSaeki:
## Description

This PR prevents temporary duplication caused by [this feature](https://github.com/autowarefoundation/autoware_universe/pull/11397).

There are two fixed points.

1. Use `std::stable_sort` instead of `std::sort` in `removeDuplicateIds` to store the data uniquely.
2. Avoid duplicate about `output` in setCrosswalkTrafficSignal. (Actually it is sufficient to fix 1. point, but good to keep the data clean)

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

https://evaluation.ci.tier4.jp/evaluation/reports/022d9bbf-33c5-5868-9ec8-68f26c2006ae?project_id=prd_jt

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_crosswalk_traffic_light_estimator/src/node.cpp
+++ b/perception/autoware_crosswalk_traffic_light_estimator/src/node.cpp
@@ -472,8 +472,8 @@ void CrosswalkTrafficLightEstimatorNode::setCrosswalkTrafficSignal(
             .empty()) {  // unnecessary check because msg has detection but for safety
         out_signal.elements.push_back(base_traffic_signal_element);
       }
-      out_signal.elements[0].color =
-        updateAndGetColorState(detected);  // TODO(MasatoSaeki): determine what value is good for confidence
+      out_signal.elements[0].color = updateAndGetColorState(
+        detected);  // TODO(MasatoSaeki): determine what value is good for confidence
       continue;
     }
 


satoshi-ota (approved):
LGTM

perception/autoware_crosswalk_traffic_light_estimator/src/node.cpp
```diff
@@ -409,12 +409,17 @@ void CrosswalkTrafficLightEstimatorNode::setCrosswalkTrafficSignal(
   const auto tl_reg_elems = crosswalk.regulatoryElementsAs<const lanelet::TrafficLight>();
 
   std::unordered_map<lanelet::Id, size_t> valid_id2idx_map;  // detected traffic light
-
   for (size_t i = 0; i < msg.traffic_light_groups.size(); ++i) {
     auto signal = msg.traffic_light_groups[i];
     valid_id2idx_map[signal.traffic_light_group_id] = i;
   }
 
+  std::unordered_map<lanelet::Id, size_t> output_id2idx_map;  // to check duplicate
```
YoshiRi, line 417: I agree we need to add this map to handle output control.

perception/autoware_crosswalk_traffic_light_estimator/src/node.cpp
```diff
@@ -409,12 +409,17 @@ void CrosswalkTrafficLightEstimatorNode::setCrosswalkTrafficSignal(
   const auto tl_reg_elems = crosswalk.regulatoryElementsAs<const lanelet::TrafficLight>();
 
   std::unordered_map<lanelet::Id, size_t> valid_id2idx_map;  // detected traffic light
-
   for (size_t i = 0; i < msg.traffic_light_groups.size(); ++i) {
     auto signal = msg.traffic_light_groups[i];
```
YoshiRi: [minor comment]
It's better to use `const &`. Same comments to many other parts. I think we need many rewrite but we only focus on these part.

perception/autoware_crosswalk_traffic_light_estimator/src/node.cpp
```diff
@@ -409,12 +409,17 @@ void CrosswalkTrafficLightEstimatorNode::setCrosswalkTrafficSignal(
   const auto tl_reg_elems = crosswalk.regulatoryElementsAs<const lanelet::TrafficLight>();
 
   std::unordered_map<lanelet::Id, size_t> valid_id2idx_map;  // detected traffic light
-
   for (size_t i = 0; i < msg.traffic_light_groups.size(); ++i) {
```
YoshiRi, line 412: [main comment]
I think this for loop is now too messy. IMHO we need total rewrite to handle this part.

perception/autoware_crosswalk_traffic_light_estimator/src/node.cpp
```diff
@@ -423,14 +428,21 @@ void CrosswalkTrafficLightEstimatorNode::setCrosswalkTrafficSignal(
   for (const auto & tl_reg_elem : tl_reg_elems) {
     auto id = tl_reg_elem->id();
     if (crosswalk_traffic_signal_overrides.count(id)) {
-      TrafficSignal output_traffic_signal;
       TrafficSignalElement output_traffic_signal_element = base_traffic_signal_element;
       output_traffic_signal_element.color = crosswalk_traffic_signal_overrides.at(id);
-      output_traffic_signal.elements.push_back(output_traffic_signal_element);
-      output_traffic_signal.traffic_light_group_id = id;
-      output.traffic_light_groups.push_back(output_traffic_signal);
+      if (output_id2idx_map.count(id)) {
```
YoshiRi, line 433: Good. I think we also need same check in `else` statement in L457. But it will make this for loop too messy, I suggest refactor whole part.

YoshiRi (commented):
The part you settled seems to be very complicated now. I think we need refactoring.

perception/autoware_crosswalk_traffic_light_estimator/src/node.cpp
YoshiRi: @MasatoSaeki 
How about totally rewrite this part like following code?
This fixes multiple structural issues:

1. Duplicate TrafficSignal entries caused by unconditional push_back.
1. Potential index mismatch between input and output arrays.
1. Unsafe access to empty .elements[0].
1. Inflexible override logic always taking precedence.
1. Unnecessary copies of message data.
1. Early continue statements skipping internal state updates.
1. Output ordering inconsistencies.


```cpp
for (const auto &tl_reg_elem : tl_reg_elems) {
  const lanelet::Id id = tl_reg_elem->id();

  // helper lambda to get or create output signal
  auto get_or_create_output_signal = [&](lanelet::Id id) -> TrafficSignal & {
    if (output_id2idx_map.count(id)) {
      return output.traffic_light_groups[output_id2idx_map[id]];
    } else {
      TrafficSignal new_signal;
      new_signal.traffic_light_group_id = id;
      new_signal.elements.push_back(base_traffic_signal_element);
      output.traffic_light_groups.push_back(new_signal);
      output_id2idx_map[id] = output.traffic_light_groups.size() - 1;
      return output.traffic_light_groups.back();
    }
  };

  TrafficSignal &out_signal = get_or_create_output_signal(id);

  // 1. Map-based override (highest priority)
  if (auto it = crosswalk_traffic_signal_overrides.find(id);
      it != crosswalk_traffic_signal_overrides.end()) {
    out_signal.elements[0].color = it->second;
    continue;
  }
  // 2. Use detected pedestrian signal if valid
  if (auto it = valid_id2idx_map.find(id); it != valid_id2idx_map.end()) {
    const auto &detected = msg.traffic_light_groups[it->second];

    if (!use_pedestrian_signal_detect_ || isInvalidDetectionStatus(detected)) {
      // Replace detection with estimated base color
      out_signal.elements.clear();
      out_signal.elements.push_back(base_traffic_signal_element);
      continue;
    }

    // Update flashing state and apply the most recent color
    updateFlashingState(detected);
    if (out_signal.elements.empty()) { 
      out_signal.elements.push_back(base_traffic_signal_element);
    }
    out_signal.elements[0].color = updateAndGetColorState(detected);
    continue;
  }

  // 3. No detection available → use estimated vehicle-based color
  if (out_signal.elements.empty()) {
    out_signal.elements.push_back(base_traffic_signal_element);
  }
  out_signal.elements[0].color = base_traffic_signal_element.color;
}
```

maxime-clem (approved):

perception/autoware_crosswalk_traffic_light_estimator/src/node.cpp
MasatoSaeki: @YoshiRi 

True. I fixed this part based on your snipet. Could you check it?
https://github.com/autowarefoundation/autoware_universe/pull/11536/commits/5e36f6abb5447ea231e9fe019c4a8504654b54f6

perception/autoware_crosswalk_traffic_light_estimator/src/node.cpp
MasatoSaeki: I answered [here](https://github.com/autowarefoundation/autoware_universe/pull/11536#discussion_r2450398453).

perception/autoware_crosswalk_traffic_light_estimator/src/node.cpp
MasatoSaeki: Fixed in https://github.com/autowarefoundation/autoware_universe/pull/11536/commits/ebc3e7dd89e84d980798789c443db0b4dece6882.

PR #11540, opened 2025-10-22
fix: use `main` for `autoware_adapi_msgs`

sasakisasaki:
## Description

When running CI build, we frequently use `main` for `autoware_adapi_msgs`. Recently we faced the CI issue in the following PR which can be resolved by using `main` for `autoware_adapi_msgs`.
- https://github.com/autowarefoundation/autoware_core/pull/670

## How was this PR tested?

We will tests by running CI. Will be updated after the tests.

## Notes for reviewers

Same as "Description"

## Interface changes

None.

## Effects on system behavior

CI build will use `main` for `autoware_adapi_msgs`.

### DIFF
--- a/build_depends_humble.repos
+++ b/build_depends_humble.repos
@@ -8,7 +8,7 @@ repositories:
   core/autoware_adapi_msgs:
     type: git
     url: https://github.com/autowarefoundation/autoware_adapi_msgs.git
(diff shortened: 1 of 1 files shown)


sasakisasaki:
It seems the issue looks existing only in `autoware_core`. Let me apply the fix for the `autoware_core` side

PR #11546, opened 2025-10-22
fix(default_adapi_universe): add energy normalize parameter

isamu-takagi:
## Description

AD API defines the range of energy level ​as 0.0 to 1.0.
However, in some existing implementations, the ranges from 0 to 100, so add a parameter for normalization.

## Related links

[TIER IV internal link](https://star4.slack.com/archives/C05BEU7GFA9/p1761029390639529)

## How was this PR tested?

```
ros2 topic pub /vehicle/status/battery_charge tier4_vehicle_msgs/msg/BatteryStatus "energy_level: 30"
ros2 topic echo /api/vehicle/metrics
```

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


system/autoware_default_adapi_universe/config/default_adapi.param.yaml
```diff
@@ -20,6 +20,7 @@
 /adapi/node/vehicle_metrics:
   ros__parameters:
     update_rate: 0.1
+    max_energy_level: 100.0
```
mitsudome-r: It's better to set default to 1.0 so we won't disrupt users that are already using this interface.

isamu-takagi: Modified https://github.com/autowarefoundation/autoware_universe/pull/11546/commits/125b74a7d2d326c01a8f0ec446acd4f11a2924d1

system/autoware_default_adapi_universe/src/vehicle_metrics.cpp
```diff
@@ -37,7 +39,7 @@ void VehicleMetricsNode::on_timer()
 
   VehicleMetrics::Message msg;
   msg.stamp = now();
-  msg.energy = energy_ ? energy_->energy_level : f_nan;
+  msg.energy = energy_ ? energy_->energy_level / max_energy_level_ : f_nan;
```
sasakisasaki, line 42: To avoid divide-by-zero, perhaps can we use a similar approach like [this](https://github.com/autowarefoundation/autoware_core/blob/4ff43d292bde668054a6efe5b05ed3cd45b326c5/planning/autoware_path_generator/param/path_generator_parameters.yaml#L5)?

Ref. https://github.com/PickNikRobotics/generate_parameter_library

(ADDED) Sorry, the link was wrong and I put a correct URL

isamu-takagi: Currently, there are no plans to use that library in the ADAPI implementation. This is such an obvious case that I want to cover it with a static check.

sasakisasaki, line 42:
```suggestion
  msg.energy = (energy_ && max_energy_level_ > 0)
    ? energy_->energy_level / max_energy_level_
    : f_nan;
```

isamu-takagi: As I said before, I'm not going to check this here. This should be checked static like JSON schema check, not at runtime.

PR #11549, opened 2025-10-23
feat: limit area for visibility estimation

manato:
## Description
This PR adds the capability of limiting the area used for estimating visibility in `autoware_pointcloud_preprocessor/polar_voxel_outlier_filter`. This PR consists of the following modifications:

- Introducing new parameters named `visibility_estimation_(min|max)_azimuth_rad` and `visibility_estimation_(min|max)_elevation_rad` to limit the area used for visibility estimation
  - the used area can be visualized by setting `publish_area_marker:=true`
- Introducing a hysteresis filter so that a single (or a few) frame error(s) can be suppressed
  - The default values for this filter are set to perform pass-through

The following capture depicts an example output after merging this PR
Screenshot from 2025-10-23 15-00-50

  

## Related links
- [TIER IV INTERNAL](https://tier4.atlassian.net/browse/RT0-39844)

## How was this PR tested?
- Download rosbag
  ```bash
  # false alert case due to low vegetation (TIER IV INTERNAL)
  webauto data log-file pull-filtered-rosbag --project-id x2_dev \
    --log-file-ids "83c9c6b8-7013-4658-ac69-efd2be17a738" \
    --target-dir . \
    --topics "/sensing/lidar/right_upper/pandar_packets"

  # false alert case due to large wind shield (TIER IV INTERNAL)
  webauto data log-file pull-filtered-rosbag --project-id x2_dev \
    --log-file-ids "fe73e739-4459-46e6-8363-2b9c69634167" \
    --target-dir . \
    --topics "/sensing/lidar/right_upper/pandar_packets"

  # true alert case under rain environment (TIER IV INTERNAL)
  webauto data log-file pull-filtered-rosbag --project-id x2_dev \
    --log-file-ids "75d6df26-b3b3-4274-bfb3-5aae3a2cf894" \
    --target-dir . \
    --topics "/sensing/lidar/right_upper/pandar_packets"
  ```
- play rosbag 
  ```bash
  ros2 bag play <bag name> --clock 100
  ```
- run Nebula to decode pointcloud
- launch polar_voxel_outlier_filter
  ```bash
  ros2 launch autoware_pointcloud_preprocessor polar_voxel_outlier_filter_node.launch.xml \ 
    input_topic_name:=/sensing/lidar/right_upper/pandar_points \
    polar_voxel_outlier_filter_param_file:=./polar_voxel_outlier_filter_node.param.yaml
  ```
  - parameters other than those newly introduced by this PR were cited from [aip_x2_gen2_launch](https://github.com/tier4/aip_launcher/blob/a52dfdebcd0aac94bf48e2d732f93ac75a36d816/aip_x2_gen2_launch/config/polar_voxel_outlier_filter_node.param.yaml)

## Notes for reviewers

The area visualization is created under the assumption of specific coordinates as follows:
- azimuth: starts from the y-axis, increasing in counter-corkscrew rule around the z-axis. The range domain is [0, 2pi]
- elevation: starts from the x-axis, increasing in counter-corkscrew rule around the y-axis. The range domain is [-pi/2, pi/2]

These assumptions may not be applicable to certain LiDARs.

#### Additions and removals

| Change type | Parameter Name                            | Type     | Default Value | Description                                                                                                              |
|:------------|:------------------------------------------|:---------|:--------------|:-------------------------------------------------------------------------------------------------------------------------|
| Added       | `visibility_estimation_min_azimuth_rad`   | `double` | `0.78`        | The miminum range to consider for visibility estimation [rad] (limits visibility calculations to reliable sensor range). |
| Added       | `visibility_estimation_max_azimuth_rad`   | `double` | `2.35`        | The maximum range to consider for visibility estimation [rad] (limits visibility calculations to reliable sensor range). |
| Added       | `visibility_estimation_min_elevation_rad` | `double` | `-0.26`       | The minimum range to consider for visibility estimation [rad] (limits visibility calculations to reliable sensor range). |
| Added       | `visibility_estimation_max_elevation_rad` | `double` | `1.04`        | The maximum range to consider for visibility estimation [rad] (limits visibility calculations to reliable sensor range). |
| Added       | `publish_area_marker`                     | `bool`   | `false`       | Publish a marker to visualize region used for visibility estimation                                                      |
| Added       | `num_frames_hysteresis_transition`        | `int`    | `1`           | The number of frames to be required to transition judgement                                                              |
| Added       | `immediate_report_error`                  | `bool`   | `false`       | Whether to report error state immediately if a single frame categorized as error                                         |
| Added       | `immediate_relax_state`                   | `bool`   | `false`       | Whether to relax state if observed state is better than the current one                                                  |

## Effects on system behavior

None.

### DIFF
--- a/sensing/autoware_pointcloud_preprocessor/docs/polar-voxel-outlier-filter.md
+++ b/sensing/autoware_pointcloud_preprocessor/docs/polar-voxel-outlier-filter.md
@@ -356,7 +356,7 @@ visibility_estimation_only: true # Diagnostics only
 - **Supported formats**: PointXYZIRC and PointXYZIRCAEDT only
 - **Finite coordinates required**: Automatically filters out NaN/Inf points
 - **Return type dependency**: Advanced filtering effectiveness depends on accurate return type classification
-- **Visibility range dependency**: Visibility accuracy depends on appropriate `visibility_estimation_max_range_m`, `visibility_estimation_(min|max)_azimutha_rad`, and `visibility_estimation_(min|max)_elevation_rad` settings. Besides, for azimuth and elevation range, the following definitions are assumed
+- **Visibility range dependency**: Visibility accuracy depends on appropriate `visibility_estimation_max_range_m`, `visibility_estimation_(min|max)_azimuth_rad`, and `visibility_estimation_(min|max)_elevation_rad` settings. Besides, for azimuth and elevation range, the following definitions are assumed
   - azimuth: starts with the y-axis, increasing in counter-corkscrew rule around the z-axis. The range domain is $`[0, 2\pi]`$
   - elevation: starts with the x-axis, increasing in counter-corkscrew rule around the y-axis. The range domain is $`[-\frac{\pi}{2}, \frac{pi}{2}]`$
 - **Secondary voxel limiting**: Visibility estimation can be tuned via `visibility_estimation_max_secondary_voxel_count`


sensing/autoware_pointcloud_preprocessor/docs/polar-voxel-outlier-filter.md
```diff
@@ -254,19 +258,23 @@ This implementation inherits `autoware::pointcloud_preprocessor::Filter` class,
...
-| `filter_ratio_warn_threshold`  | double | Warning threshold for filter ratio      | 0.7     |
-| `visibility_error_threshold`   | double | Error threshold for visibility metric   | 0.8     |
-| `visibility_warn_threshold`    | double | Warning threshold for visibility metric | 0.9     |
+| Parameter                          | Type   | Description                                                                                               | Default |
+| ---------------------------------- | ------ | --------------------------------------------------------------------------------------------------------- | ------- |
+| `filter_ratio_error_threshold`     | double | Error threshold for filter ratio                                                                          | 0.5     |
+| `filter_ratio_warn_threshold`      | double | Warning threshold for filter ratio                                                                        | 0.7     |
+| `visibility_error_threshold`       | double | Error threshold for visibility metric                                                                     | 0.8     |
```
badai-nguyen: NIT: Replacing the manual parameter description with 
{{ json_to_markdown("sensing/autoware_pointcloud_preprocessor/schema/polar_voxel_outlier_filter_node.schema.json") }}
could help reduce maintenance cost.

sensing/autoware_pointcloud_preprocessor/src/outlier_filter/polar_voxel_outlier_filter_node.cpp
```diff
@@ -737,6 +780,28 @@ bool PolarVoxelOutlierFilterComponent::validate_normalized(
...
+bool PolarVoxelOutlierFilterComponent::validate_negative_half_pi_to_half_pi(
+  const rclcpp::Parameter & param, std::string & reason)
+{
+  double val = param.as_double();
+  if (val < -HALF_PI || val > HALF_PI) {
+    reason = param.get_name() + " must be between -PI/2 and PI/2";
+    return false;
+  }
+  return true;
+}
```
badai-nguyen, line 842: these function are quite similar, shall we combine these 2 functions into 1 like: validate_in_range?

sensing/autoware_pointcloud_preprocessor/config/polar_voxel_outlier_filter_node.param.yaml
```diff
@@ -9,17 +9,25 @@
...
+    visibility_estimation_min_elevation_rad: -0.26 # Minimum elevation for estimating point cloud visibility, in radius
+    visibility_estimation_max_elevation_rad: 1.04  # Maximum elevation for estimating point cloud visibility, in radius
 
     # Return type classification parameters
     use_return_type_classification: true         # Whether to use return type classification
     filter_secondary_returns: false              # Whether to filter secondary returns (keep only primary when true)
     secondary_noise_threshold: 4                 # Threshold for classifying primary vs secondary returns
     visibility_estimation_max_secondary_voxel_count: 500  # Maximum secondary voxel count for visibility estimation
     primary_return_types: [1, 6, 8, 10]          # Return types considered as primary returns (all others are secondary)
+    num_frames_hysteresis_transition: 1          # The number of frames to be required to transition judgement
```
drwnz: Nit: these parameters should probably be grouped with diagnostic thresholds rather than return type classification.

sensing/autoware_pointcloud_preprocessor/docs/polar-voxel-outlier-filter.md
manato: @badai-nguyen 
Thanks for your comment. I addressed [35aa337](https://github.com/autowarefoundation/autoware_universe/pull/11549/commits/35aa337c292c4c0dc8fe81743dea97b64f6e389f)

sensing/autoware_pointcloud_preprocessor/src/outlier_filter/polar_voxel_outlier_filter_node.cpp
manato: Thanks for your detailed check! Since these functions are called to validate parameters like [these](https://github.com/autowarefoundation/autoware_universe/blob/35aa337c292c4c0dc8fe81743dea97b64f6e389f/sensing/autoware_pointcloud_preprocessor/src/outlier_filter/polar_voxel_outlier_filter_node.cpp#L856-L861) lines without arguments, I would prefer these as they are

sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/diagnostics/hysteresis_state_machine.hpp
```diff
@@ -0,0 +1,195 @@
+// Copyright 2025 TIER IV, Inc.
```
drwnz, line 1: Is the location of this library appropriate? I feel it should be in the root include directory for pointcloud preprocessor, or even in a common utils directory in Autoware, since none of the contents are specific to outlier filtering.

sensing/autoware_pointcloud_preprocessor/src/outlier_filter/polar_voxel_outlier_filter_node.cpp
```diff
@@ -993,6 +1102,107 @@ PolarVoxelOutlierFilterComponent::extract_polar_from_xyz(float x, float y, float
...
+  marker.scale.x = 1.0;
+  marker.scale.y = 1.0;
+  marker.scale.z = 1.0;
+  marker.color.a = 0.1;
+  marker.color.r = 0.0;
+  marker.color.g = 1.0;  // transparent green
+  marker.color.b = 0.0;
+  marker.pose.orientation.w = 1.0;
+
+  auto polar_to_xyz = [](auto r, auto az, auto el) {
```
drwnz: super nit: variable names for az, el are not commonly used. Suggest using radius, azimuth, elevation.

sensing/autoware_pointcloud_preprocessor/config/polar_voxel_outlier_filter_node.param.yaml
manato: Thanks for pointing this out. addressed in [26ef0f2](https://github.com/autowarefoundation/autoware_universe/pull/11549/commits/26ef0f2acea5f5d03b692fe39efa4933e97abdd9)

sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/diagnostics/hysteresis_state_machine.hpp
manato: Agree. SInce I found `include/autoware/pointcloud_preprocessor/diagnostics`, I moved `hysteresis_state_machine.hpp` there! [34289fb](https://github.com/autowarefoundation/autoware_universe/pull/11549/commits/34289fb65c56800f34ed9888973155ddb5050875)

sensing/autoware_pointcloud_preprocessor/src/outlier_filter/polar_voxel_outlier_filter_node.cpp
manato: Renamed in [52ff3e8](https://github.com/autowarefoundation/autoware_universe/pull/11549/commits/52ff3e816c6c8cb1fd0a30a1dc420cf6d6ab4595).  Although other lines still have variables like `az1`, I left them as they are because they are short-scoped local variables, and avoiding the codes getting longer unintentionally. :pray:

manato:
@drwnz 
Thank you very much for your detailed review!

> Other suggestion: I think the defaults for the node are a bit arbitrary for most sensors. I would propose that the defaults are full range (no effect).

Regarding this point, I set full range (no effect) as default values in [442f6e6](https://github.com/autowarefoundation/autoware_universe/pull/11549/commits/442f6e6236dc2adb75ab8a58c5083aa1a2371e3f)

> It may also be better to test for the case when min_[azimuth, elevation] > max_[azimuth, elevation] as this can cause problems if wraparound is not handled.

I appreciate your suggestion. I completely missed such cases. I updated the range conditions according to `min < max` or `min >= max` in [12902f1](https://github.com/autowarefoundation/autoware_universe/pull/11549/commits/12902f1313345e6235b9e101a600f2dee0851e31). 
The resulting visualizations become as follows:

- case1: max_azimuth < min_azimuth
  ```yaml
    visibility_estimation_min_azimuth_rad: 2.35
    visibility_estimation_max_azimuth_rad: 0.78  # <- smalleer than min!
    visibility_estimation_min_elevation_rad: -0.26
    visibility_estimation_max_elevation_rad: 1.04
  ```
  Screenshot from 2025-10-29 01-06-57

- case2: max_elevation < min_elevation
  ```yaml
    visibility_estimation_min_azimuth_rad: 0.78
    visibility_estimation_max_azimuth_rad: 2.35
    visibility_estimation_min_elevation_rad: 1.04
    visibility_estimation_max_elevation_rad: -0.26  # <- smaller than min!
  ```
  Screenshot from 2025-10-29 01-07-34
  Note: elevation is defined within $[ -\frac{\pi}{2}, \frac{\pi}{2}]$

If you feel this range treatment is inappropriate, please let me know!

drwnz:
Thanks for the changes! Looks good. I'd like to test with another sensor (like Velodyne) for this feature and the whole module at a later time as I think their may be some reference frame dependencies.

drwnz (approved):

PR #11550, opened 2025-10-23
feat(autoware_trajectory_optimizer): customization of point constraining

danielsanchezaran:
## Description
This PR adds the option to customize point constraining in the QP solver to prioritize starting / ending acceleration and velocity with respect to the original trajectory. 

Reasoning: 
Since acceleration is calculated using a 3 point window, we can preserve / constrain the first 3 points of the original trajectory to preserve the original acceleration value, making the trajectory smoother but also constrained to the initial conditions, same for the ending points.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?
PSIM 

[Screencast from 10-23-2025 01:16:29 PM.webm](https://github.com/user-attachments/assets/62005fca-6e98-406b-9c85-7f70a6bd2993)

[Screencast from 10-23-2025 01:26:17 PM.webm](https://github.com/user-attachments/assets/6cfb6d79-a08e-4cf7-a399-adc413a23042)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/autoware_trajectory_optimizer/docs/qp_smoother.md
+++ b/planning/autoware_trajectory_optimizer/docs/qp_smoother.md
@@ -109,7 +109,6 @@ After QP optimization solves for smoothed positions, the following are recalcula
 ### Core Optimization Weights
 
 - `weight_smoothness` (default: 10.0)
-
   - Controls smoothness penalty
   - Higher values → smoother but more deviation from original path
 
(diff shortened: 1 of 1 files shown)


planning/autoware_trajectory_optimizer/schema/trajectory_optimizer.schema.json
```diff
@@ -194,10 +194,17 @@
...
-              "description": "Fix last point as hard constraint (if false, only first point fixed)"
+            "num_constrained_points_start": {
+              "type": "integer",
+              "default": 3,
+              "minimum": 0,
+              "description": "Number of points from the start of the trajectory to constrain as hard constraints (preserves initial state)"
+            },
+            "num_constrained_points_end": {
+              "type": "integer",
+              "default": 0,
```
go-sakayori:
```suggestion
              "default": 3,
```

planning/autoware_trajectory_optimizer/src/trajectory_optimizer_plugins/trajectory_qp_smoother.cpp
```diff
@@ -145,11 +156,16 @@ void TrajectoryQPSmoother::optimize_trajectory(
     return;
   }
 
-  constexpr size_t min_points_for_optimization = 5;
```
go-sakayori, line 148: Why did you delete it? Better to have it as a variable rather than putting hard coded 5.

danielsanchezaran: I added a variable instead of having the 5 as a magic number, thanks

PR #11563, opened 2025-10-28
feat(object_sorter): add per-axis min-max range filter for each class label

a-maumau:
## Description
This PR will add a min-max range filter for each x and y axis for each class label.
If the objects are outside this min-max range, it will be filtered out.

Also remove the parameter update function, which is considered to be not in use.

## Related links

autoware_launch: https://github.com/autowarefoundation/autoware_launch/pull/1684

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_map_based_prediction/include/map_based_prediction/predictor_vru.hpp
+++ b/perception/autoware_map_based_prediction/include/map_based_prediction/predictor_vru.hpp
@@ -147,6 +147,7 @@ private:
   /// @brief return true if the given predicted path crosses a fence before arriving at the
   /// crosswalk
   bool doesPathCrossAnyFenceBeforeCrosswalk(const PredictedPathWithArrivalIndex & predicted_path);
+  PredictedPath cutPathBeforeFences(const PredictedPath & predicted_path) const;
   std::optional<TrafficLightElement> getTrafficSignalElement(const lanelet::Id & id);
 };
 
--- a/perception/autoware_map_based_prediction/src/predictor_vru.cpp
+++ b/perception/autoware_map_based_prediction/src/predictor_vru.cpp
@@ -17,6 +17,7 @@
 #include "map_based_prediction/path_generator.hpp"
 #include "map_based_prediction/utils.hpp"
 
+#include <autoware/motion_utils/trajectory/trajectory.hpp>
 #include <autoware_utils/geometry/geometry.hpp>
 #include <autoware_utils/ros/uuid_helper.hpp>
 
@@ -24,8 +25,10 @@
 
 #include <algorithm>
 #include <deque>
+#include <functional>
 #include <limits>
 #include <memory>
+#include <numeric>
 #include <string>
 #include <unordered_map>
 #include <utility>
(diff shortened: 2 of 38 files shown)


perception/autoware_object_sorter/src/object_sorter_base.cpp
```diff
@@ -97,27 +137,27 @@ void ObjectSorterBase<ObjsMsgType>::objectCallback(
...
+
   for (const auto & object : input_msg->objects) {
-    // Filter by velocity
-    if (
-      // NOTE: use the raw velocity in the topic for now
-      std::abs(
-        autoware_utils_geometry::calc_norm(object.kinematics.twist_with_covariance.twist.linear)) <
-      node_param_.min_velocity_threshold) {
-      // Low velocity object
+    const LabelSettings & label_settings = label_settings_[object.classification.front().label];
```
technolojin: not always the first class largest

technolojin (approved):
see the comment.

other then that, LGTM

perception/autoware_object_sorter/src/object_sorter_base.cpp
a-maumau: @technolojin Fixed at https://github.com/autowarefoundation/autoware_universe/pull/11563/commits/58693d38174b8363233df05c8c68feab5a00d9c8 .
Using the `autoware::object_recognition_utils::getHighestProbLabel` to take the highest prob. label.

YoshiRi: We have `getHighestProbClassification` function in utils

a-maumau: @YoshiRi You mean the one in the autoware_core's `autoware::object_recognition_utils::getHighestProbLabel` (fixed in https://github.com/autowarefoundation/autoware_universe/commit/58693d38174b8363233df05c8c68feab5a00d9c8), right?
Or is there other utils?

YoshiRi: Sorry that was my mistypo :bow:

perception/autoware_object_sorter/src/object_sorter_base.cpp
```diff
@@ -72,6 +52,68 @@ ObjectSorterBase<ObjsMsgType>::ObjectSorterBase(
...
+          // Outside x
+          dx < min_x || dx > max_x ||
+          // Outside y
+          dy < min_y || dy > max_y) {
+          // Outside the target area
+          return false;
+        } else {
+          return true;
+        }
+      };
```
YoshiRi, line 101: [minor comment]
You can simplify this part:

```suggestion
      label_settings.isInTargetRange = [min_x, max_x, min_y, max_y](double dx, double dy) {
        return (min_x <= dx && dx <= max_x && min_y <= dy && dy <= max_y);
      };
```

perception/autoware_object_sorter/src/object_sorter_base.cpp
```diff
@@ -97,27 +139,29 @@ void ObjectSorterBase<ObjsMsgType>::objectCallback(
     RCLCPP_WARN(this->get_logger(), "Transform lookup failed: %s", ex.what());
   }
 
+  double dist_origin_x = ego_pos.x + range_calc_offset_x_;
```
YoshiRi: [memo]
remember ego_pos means object_frame -> ego vector in object_frame coordinate.

YoshiRi: [imho]

We should completely rewrite this transformation and validation using pure coordinate transformation like:

```cpp
try {
  // object frame -> ego frame
  const auto tf_obj_to_ego = tf_buffer_.lookupTransform(
    range_calc_frame_id_, input_msg->header.frame_id,
    input_msg->header.stamp, rclcpp::Duration::from_seconds(0.5));

  for (const auto & object : input_msg->objects) {
    geometry_msgs::msg::PoseStamped pose_in, pose_out;
    pose_in.header = input_msg->header;
    pose_in.pose = object.kinematics.pose_with_covariance.pose;

    tf2::doTransform(pose_in, pose_out, tf_obj_to_ego);

    const double dx = pose_out.pose.position.x - range_calc_offset_x_;
    const double dy = pose_out.pose.position.y - range_calc_offset_y_;

    const uint8_t label =
      autoware::object_recognition_utils::getHighestProbLabel(object.classification);
    const LabelSettings & settings = label_settings_[label];

    if (!settings.publish) continue;
    if (!settings.isInTargetVelocity(object.kinematics.twist_with_covariance.twist.linear)) continue;
    if (!settings.isInTargetRange(dx, dy)) continue;

    output_objects.objects.push_back(object);
  }

  pub_output_objects_->publish(output_objects);

} catch (const tf2::TransformException & ex) {
  RCLCPP_WARN(this->get_logger(), "Transform lookup failed: %s", ex.what());
}
```

perception/autoware_object_sorter/src/object_sorter_base.cpp
a-maumau: Yes, you are right. I thought the `||` one is faster, but it is the same.

perception/autoware_object_sorter/src/object_sorter_base.cpp
a-maumau: Big thanks for pointing out the bug.
The original (current) implementation only needed to calculate the distance, so it was intended to do some inverse transform to avoid all objects to transform.

I fixed the issue at https://github.com/autowarefoundation/autoware_universe/pull/11563/commits/d8c46bff4dc3ee97230e53e34c353337e745d9d8 .

a-maumau:
Visualization

## distance mode (vs original implementation)

[compare_distance_with_original.webm](https://github.com/user-attachments/assets/f81c1b52-882e-47e8-b43d-61af6e546e43)

Top: current, bottom: this PR.
I think there is no difference. (radar topic was used with velocity threahsold: 0, min_range: 30.0)


## area mode (target frame: baselink)

[test_area.webm](https://github.com/user-attachments/assets/0de9be18-3345-4503-9cac-ecf40067b776)

Grid cell size is 10 m.
Parameters:
```
min_velocity_threshold: 0.0
 
max_x: 30.0
min_x: -10.0
max_y: 30.0
min_y: -20.0
```
Objects that is outside of this min-max area will be removed.

## area mode (target frame: radar)

[test_other_frame.webm](https://github.com/user-attachments/assets/e71ca8c8-7daa-4a33-8e7f-3ae084e855f1)

Grid cell size is 10 m.
Parameters:
```
min_velocity_threshold: 0.0
 
max_x: 20.0
min_x: -20.0
max_y: 20.0
min_y: -20.0
```
Objects that is outside of this min-max area will be removed.

YoshiRi (approved):
LGTM

PR #11568, opened 2025-10-29
feat(map_based_prediction): trim default crosswalk user path at fence

soblin:
## Description

Modified the default VRU path to be trimmed just before the crossing with fence

### before

image

### after

https://github.com/user-attachments/assets/3f281eb0-6bd2-4d30-9b35-4d463d4491b5

## Related links

https://tier4.atlassian.net/browse/T4DEV-39370

**Parent Issue:**

- Link

## How was this PR tested?

- [Common](https://evaluation.tier4.jp/evaluation/reports/0ea39323-0689-5b0b-8ced-236f68b4de37?project_id=prd_jt)
- [DLR](https://evaluation.tier4.jp/evaluation/reports/70d93922-2017-54ba-8dbd-9d07d49aa2de?project_id=prd_jt)
- [FuncVerification](https://evaluation.tier4.jp/evaluation/reports/f79195e6-75a1-5710-83fa-f3e73d48e723?project_id=prd_jt)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/motion_velocity_planner/autoware_motion_velocity_run_out_module/src/slowdown.cpp
+++ b/planning/motion_velocity_planner/autoware_motion_velocity_run_out_module/src/slowdown.cpp
@@ -40,12 +40,15 @@ geometry_msgs::msg::Point interpolated_point_at_time(
     [&](const autoware_planning_msgs::msg::TrajectoryPoint & t) {
       return rclcpp::Duration(t.time_from_start).seconds() >= time;
     });
+  if (prev_it == trajectory.end()) {
+    return trajectory.back().pose.position;
+  }
   const auto prev_time = rclcpp::Duration(prev_it->time_from_start).seconds();
-  if (prev_time == time) {
+  if (std::next(prev_it) == trajectory.end() || std::abs(prev_time - time) < 1e-3) {
     return prev_it->pose.position;
   }
   const auto next_time = rclcpp::Duration(std::next(prev_it)->time_from_start).seconds();
-  if (next_time == prev_time) {
+  if (std::abs(next_time - prev_time) < 1e-3) {
     return std::next(prev_it)->pose.position;
   }
   const auto t_delta = next_time - prev_time;


perception/autoware_map_based_prediction/include/map_based_prediction/predictor_vru.hpp
```diff
@@ -147,6 +147,7 @@ class PredictorVru
   /// @brief return true if the given predicted path crosses a fence before arriving at the
   /// crosswalk
   bool doesPathCrossAnyFenceBeforeCrosswalk(const PredictedPathWithArrivalIndex & predicted_path);
+  bool doesPathCrossAnyFenceBeforeCrosswalk(const PredictedPath & predicted_path) const;
```
maxime-clem:
```suggestion
  bool doesPathCrossAnyFence(const PredictedPath & predicted_path) const;
```

perception/autoware_map_based_prediction/src/predictor_vru.cpp
```diff
@@ -257,6 +257,23 @@ bool PredictorVru::doesPathCrossAnyFenceBeforeCrosswalk(
   return false;
 }
 
+bool PredictorVru::doesPathCrossAnyFenceBeforeCrosswalk(const PredictedPath & predicted_path) const
+{
+  lanelet::BasicLineString2d predicted_path_ls;
+  for (auto i = 0UL; i <= predicted_path.path.size(); ++i) {
```
maxime-clem:
```suggestion
  for (auto i = 0UL; i < predicted_path.path.size(); ++i) {
```

perception/autoware_map_based_prediction/src/predictor_vru.cpp
```diff
@@ -257,6 +257,23 @@ bool PredictorVru::doesPathCrossAnyFenceBeforeCrosswalk(
...
+    const auto & pt = predicted_path.path[i];
+    predicted_path_ls.emplace_back(pt.position.x, pt.position.y);
```
maxime-clem:
```suggestion
  for (const auto & pt : predicted_path.pathi) {
    predicted_path_ls.emplace_back(pt.position.x, pt.position.y);
```

maxime-clem (changes requested):
the new function checks the whole path so I propose to rename it.
It also contains a bug (out of array index).

I am also not sure about the core logic.
We probably do not want to ignore the full default predicted path just because one of its segment crosses a fence. Instead I propose to cut the predicted path at the first index that crosses a fence.

perception/autoware_map_based_prediction/src/predictor_vru.cpp
```diff
@@ -257,6 +260,64 @@ bool PredictorVru::doesPathCrossAnyFenceBeforeCrosswalk(
...
+      crossed_fences.push_back(candidate);
+      break;
+    }
+  }
+  if (crossed_fences.empty()) {
+    return predicted_path;
+  }
+
+  const auto & path = predicted_path.path;
+  auto closest_cross_index = path.size() - 1;
```
maxime-clem: please add a guard against `path.size() == 0`. The best is probably an early `return` if `predicted_path.path` is empty.

perception/autoware_map_based_prediction/src/predictor_vru.cpp
```diff
@@ -257,6 +260,64 @@ bool PredictorVru::doesPathCrossAnyFenceBeforeCrosswalk(
...
+    const double distance_to_nearest_intersection = std::inner_product(
+      path.begin(), path.begin() + nearest_intersection_index, path.begin() + 1, 0.0 /* init */,
+      std::plus<double>{}, [](const auto & a, const auto & b) {
+        return autoware_utils_geometry::calc_distance3d(a, b);
+      });
+    if (distance_to_nearest_intersection < dist_to_closest_cross) {
+      dist_to_closest_cross = distance_to_nearest_intersection;
+      closest_cross_index = nearest_intersection_index;
+    }
+  }
```
maxime-clem: I think the search for the `closest_cross_index` can be simplified.
```suggestion
  std::optional<size_t> closest_cross_index;
  for (auto i = 0UL; i + 1 < predicted_path_ls.size() && !closest_cross_index.has_value(); ++i) {
    autoware_utils_geometry::Segment2d path_segment(predicted_path_ls[i], predicted_path_ls[i+1]);
    for (const auto & fence : crossed_fences) {
      if(boost::geometry::intersects(path_segment, lanelet::utils::to2D(fence).basicLineString())) {
        closest_cross_index = i;
      }
    }
  }
```

maxime-clem (approved):
Looks good but the code can be simplified I think.

perception/autoware_map_based_prediction/src/predictor_vru.cpp
soblin: fixed in 3f3a70658288e3deaa17a664ef70ae6bbba31cd7

perception/autoware_map_based_prediction/src/predictor_vru.cpp
soblin: fixed in 3f3a70658288e3deaa17a664ef70ae6bbba31cd7

PR #11584, opened 2025-11-03
feat(behavior_velocity_no_stopping_area_module): use Trajectory instead of PathWithLaneId

mitukou1109:
## Description

This PR replaces `PathWithLaneId` in `behavior_velocity_no_stopping_area_module` with `Trajectory` as much as possible.

## How was this PR tested?

Psim

Evaluator: https://evaluation.tier4.jp/evaluation/reports/740642b8-ad8d-5338-8724-3aa048729ce4?project_id=autoware_dev

## Related links

**Parent Issue:**

- https://github.com/autowarefoundation/autoware_core/issues/322

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_velocity_planner/autoware_behavior_velocity_no_stopping_area_module/src/experimental/scene_no_stopping_area.cpp
+++ b/planning/behavior_velocity_planner/autoware_behavior_velocity_no_stopping_area_module/src/experimental/scene_no_stopping_area.cpp
@@ -205,17 +205,25 @@ bool NoStoppingAreaModule::modifyPathVelocity(
     return true;
   }
 
-  debug_data_.stuck_vehicle_detect_area = toGeomPoly(*stuck_vehicle_detect_area);
-  debug_data_.stop_line_detect_area = toGeomPoly(*stop_line_detect_area);
-  // Find stuck vehicle in no stopping area
-  const bool is_entry_prohibited_by_stuck_vehicle =
-    check_stuck_vehicles_in_no_stopping_area(*stuck_vehicle_detect_area, predicted_obj_arr_ptr);
-  // Find stop line in no stopping area
-  const bool is_entry_prohibited_by_stop_line =
-    no_stopping_area::check_stop_lines_in_no_stopping_area(
-      path, *stop_line_detect_area, debug_data_);
-  const bool is_entry_prohibited =
-    is_entry_prohibited_by_stuck_vehicle || is_entry_prohibited_by_stop_line;
+  const auto is_entry_prohibited = [&]() {
+    if (stuck_vehicle_detect_area) {
+      debug_data_.stuck_vehicle_detect_area = toGeomPoly(*stuck_vehicle_detect_area);
+      // Find stuck vehicle in no stopping area
+      if (check_stuck_vehicles_in_no_stopping_area(
+            *stuck_vehicle_detect_area, predicted_obj_arr_ptr)) {
+        return true;
+      }
+    }
+    if (stop_line_detect_area) {
+      debug_data_.stop_line_detect_area = toGeomPoly(*stop_line_detect_area);
+      // Find stop line in no stopping area
+      if (no_stopping_area::check_stop_lines_in_no_stopping_area(
+            path, *stop_line_detect_area, debug_data_)) {
+        return true;
+      }
+    }
+    return false;
+  }();
 
   pass_judge_.check_if_stoppable(distance_to_stop_point, ego_data, logger_, *clock_);
   if (!pass_judge_.is_stoppable) {


planning/behavior_velocity_planner/autoware_behavior_velocity_no_stopping_area_module/src/utils.cpp
```diff
@@ -47,6 +50,197 @@ bool is_vehicle_type(const autoware_perception_msgs::msg::PredictedObject & obje
...
+    if (!slow_velocity) {
+      // pass through
+      pass_judge.is_stoppable = false;
+      RCLCPP_WARN_THROTTLE(
+        logger, clock, 1000, "[NoStoppingArea] can't stop in front of no stopping area");
+    } else {
+      pass_judge.is_stoppable = true;
+    }
+  }
+  return pass_judge.is_stoppable;
```
soblin:
```suggestion
  return true;
```

planning/behavior_velocity_planner/autoware_behavior_velocity_no_stopping_area_module/src/utils.cpp
```diff
@@ -47,6 +50,197 @@ bool is_vehicle_type(const autoware_perception_msgs::msg::PredictedObject & obje
...
+    pass_judge.pass_judged = true;
+    // can't stop using maximum brake consider jerk limit
+    if (!slow_velocity) {
+      // pass through
+      pass_judge.is_stoppable = false;
+      RCLCPP_WARN_THROTTLE(
+        logger, clock, 1000, "[NoStoppingArea] can't stop in front of no stopping area");
+    } else {
+      pass_judge.is_stoppable = true;
+    }
```
soblin:
```suggestion
      return pass_judge.is_stoppable;
    }
```

planning/behavior_velocity_planner/autoware_behavior_velocity_no_stopping_area_module/src/experimental/scene_no_stopping_area.cpp
```diff
@@ -208,11 +214,12 @@ bool NoStoppingAreaModule::modifyPathVelocity(
...
   const bool is_entry_prohibited_by_stop_line =
     no_stopping_area::check_stop_lines_in_no_stopping_area(
-      path_msg, stop_line_detect_area, debug_data_);
+      path, stop_line_detect_area, debug_data_);
   const bool is_entry_prohibited =
     is_entry_prohibited_by_stuck_vehicle || is_entry_prohibited_by_stop_line;
+
   if (!no_stopping_area::is_stoppable(
-        pass_judge_, current_pose->pose, stop_point->second, ego_data, logger_, *clock_)) {
+        pass_judge_, distance_to_stop_point, ego_data, logger_, *clock_)) {
```
soblin: Can you define `is_stoppable` as member function of `PassJudge` returning `void`, call it before `if` sentence, and then check `pass_judge.is_stoppable` in `if`'s condition branch ?

planning/behavior_velocity_planner/autoware_behavior_velocity_no_stopping_area_module/src/utils.cpp
```diff
@@ -47,6 +50,197 @@ bool is_vehicle_type(const autoware_perception_msgs::msg::PredictedObject & obje
...
+{
+  for (const auto & no_stopping_area : no_stopping_areas) {
+    auto basic_polygon = lanelet::utils::to2D(no_stopping_area).basicPolygon();
+    basic_polygon.push_back(basic_polygon.front());  // close polygon
+    const auto collision_points = experimental::trajectory::crossed(path, basic_polygon);
+    if (collision_points.empty()) {
+      continue;
+    }
+
+    const auto closest_collision_point =
```
soblin: You don't need to call `min_element` because it is ensured the output is sorted: https://autowarefoundation.github.io/autoware_core/main/common/autoware_trajectory/#autowaretrajectoryutilscrossedhpp

planning/behavior_velocity_planner/autoware_behavior_velocity_no_stopping_area_module/src/utils.cpp
```diff
@@ -47,6 +50,197 @@ bool is_vehicle_type(const autoware_perception_msgs::msg::PredictedObject & obje
...
+
+Polygon2d generate_ego_no_stopping_area_lane_polygon(
+  const Trajectory & path, const geometry_msgs::msg::Pose & ego_pose,
+  const lanelet::autoware::NoStoppingArea & no_stopping_area_reg_elem, const double margin,
+  const double max_polygon_length, const double path_expand_width)
+{
+  const auto ego_s =
+    experimental::trajectory::find_first_nearest_index(path, ego_pose, 3.0, M_PI_4);
+  if (!ego_s) {
+    return {};
```
soblin: do not return {} for Polygon. Return optional instead

kosuke55:
Do you need to keep old functions? 
```
bool check_stop_lines_in_no_stopping_area(
  const autoware_internal_planning_msgs::msg::PathWithLaneId & path,
```

kosuke55:
> Do you need to keep old functions?

ah, I see. so we're keeping the non-experimental scene modules as well.

planning/behavior_velocity_planner/autoware_behavior_velocity_no_stopping_area_module/src/experimental/scene_no_stopping_area.cpp
```diff
@@ -157,62 +158,67 @@ bool NoStoppingAreaModule::modifyPathVelocity(
...
-  const Polygon2d stop_line_detect_area =
-    no_stopping_area::generate_ego_no_stopping_area_lane_polygon(
-      path_msg, current_pose->pose, no_stopping_area_reg_elem_, ego_space_in_front_of_stop_line,
-      planner_param_.detection_area_length, planner_param_.path_expand_width, logger_, *clock_);
-  if (stuck_vehicle_detect_area.outer().empty() && stop_line_detect_area.outer().empty()) {
+  const auto stop_line_detect_area = no_stopping_area::generate_ego_no_stopping_area_lane_polygon(
+    path, current_pose, no_stopping_area_reg_elem_, ego_space_in_front_of_stop_line,
+    planner_param_.detection_area_length, planner_param_.path_expand_width);
+
+  if (!stuck_vehicle_detect_area && !stop_line_detect_area) {
```
kosuke55, line 203: Should you handle the case where only one of them is nullopt?

mitukou1109: I simply followed the original implementation, the case where both are empty was not covered before the changes either. I could modify the condition to return early when either one is empty.

mitukou1109: > I could modify the condition to return early when either one is empty.

This seems wrong, we should continue if either one is not empty.

mitukou1109: I fixed this in https://github.com/autowarefoundation/autoware_universe/pull/11584/commits/626682573ead78afea202dd38e42a61d6f09ba1e, could you check it?

kosuke55: sorry for the dealy, approved!

PR #11590, opened 2025-11-05
fix(obstacle_slow_down): fix left/right judgment, update of deceleration constraint calculation

yuki-takagi-66:
## Description
**Change 1: Update Logic for Determining Object's Lateral Position**

Modified the logic for determining whether a target object (for deceleration) is on the left or right side.

Before: The determination was based on the ego-vehicle's pose (orientation) within the slow-down section.
After: This has been changed to determine the object's side based on the path waypoints.

Reason: In situations like the attached figure (e.g., on a curve), determining the side relative to the ego-pose can be ambiguous or unstable. Using the waypoint as a reference provides a consistent and reliable result.

image

**Change 2: Use LPF-processed Value for Slow-Down Start Point Calculation**

The calculation used to meet deceleration constraints will now use a different reference for the slow-down start point.

Before: Used the value before the LPF (Low-Pass Filter) was applied.
After: Now uses the value after the LPF is applied.

Reason: Calculations based on the pre-LPF value could potentially violate acceleration constraints in some cases.

Related Change: Concurrently, the LPF applied directly to the "slow-down velocity" has been disabled. (We assume this is acceptable, as the lateral distance to the object and the slow-down section calculations themselves are still processed by an LPF.)

## Related links
launch PR: https://github.com/autowarefoundation/autoware_launch/pull/1689
This is related to this Pull Request, but not dependent on it.

## How was this PR tested?
engage available in psim and tier4 scenario test

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/motion_velocity_planner/autoware_motion_velocity_obstacle_slow_down_module/src/obstacle_slow_down_module.cpp
+++ b/planning/motion_velocity_planner/autoware_motion_velocity_obstacle_slow_down_module/src/obstacle_slow_down_module.cpp
@@ -464,8 +464,8 @@ std::vector<SlowDownObstacle> ObstacleSlowDownModule::filter_slow_down_obstacle_
     }
     const auto & front_collision_point = *slow_down_point_data.front;
     const auto & back_collision_point = slow_down_point_data.back.value_or(front_collision_point);
-    const auto signed_lateral_distance = autoware_utils_geometry::calc_lateral_deviation(
-      traj_points[ego_idx].pose, front_collision_point);
+    const auto signed_lateral_distance =
+      motion_utils::calcLateralOffset(traj_points, front_collision_point);
     const auto side = signed_lateral_distance > 0.0 ? Side::Left : Side::Right;
 
     const auto slow_down_obstacle = create_slow_down_obstacle_for_point_cloud(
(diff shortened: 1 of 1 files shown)


planning/motion_velocity_planner/autoware_motion_velocity_obstacle_slow_down_module/src/obstacle_slow_down_module.cpp
```diff
@@ -464,8 +464,8 @@ std::vector<SlowDownObstacle> ObstacleSlowDownModule::filter_slow_down_obstacle_
     }
     const auto & front_collision_point = *slow_down_point_data.front;
     const auto & back_collision_point = slow_down_point_data.back.value_or(front_collision_point);
-    const auto signed_lateral_distance = autoware_utils_geometry::calc_lateral_deviation(
-      traj_points[ego_idx].pose, front_collision_point);
+    const auto signed_lateral_distance =
+      motion_utils::calcLateralOffset(traj_points, front_collision_point);
```
maxime-clem, line 468: (note) the docstring of the function says that the returned distance is unsigned, but this looks like a mistake.

maxime-clem (approved):

Kotakku (approved):

PR #11597, opened 2025-11-06
feat(autoware_pointcloud_preprocessor): crop box ray intersection mode

mojomex:
## Description

The current crop box filter checks whether a given point is *inside* the crop box.
This PR adds an optional mode where the filter checks whether the line segment from the LiDAR to the point *intersects* the crop box.

Please have a look at the updated [docs](https://autowarefoundation.github.io/autoware_universe/pr-11597/sensing/autoware_pointcloud_preprocessor/docs/crop-box-filter/).

### Motivation

Some LiDAR rays hit the vehicle, causing points on the vehicle. These points can be filtered out by the current filter.
However, if the vehicle has reflective surfaces (glass, mirrors, metal), the following can happen:
1. the LiDAR shoots a ray in some direction A
2. the ray reflects off the vehicle, now travelling in direction B
3. it hits a real object somewhere in the scene
4. the LiDAR assumes that the object is in the direction it sent the ray (A)
5. points show up behind the vehicle, causing a ghost of the object

image

This cannot be filtered by the current implementation, because the ghost points are outside of the crop box.
However, the crop box can be used to solve this: notice how the ray from LiDAR to ghost object has to travel *through* the vehicle - or through the crop box as an over-approximation.
Instead of checking whether a given point is *inside* the box, checking whether the line segment from LiDAR to the point *intersected* the box thus solves the issue.

### Justification

**Why should this be in CropBoxFilter?**

* The ray approach uses the same concept - use boxes to approximate the vehicle shape
* The ray approach has the same purpose - filter out points ~on the~ caused by the vehicle
* The ray approach is a generalization of the current functionality, enabling more use cases
* The functionality is completely optional

## Related links

**Parent Issue:**

- successor of https://github.com/autowarefoundation/autoware_core/pull/661

## How was this PR tested?

Two test methods:

* unit tests
* visual confirmation 

A cube-shaped pointcloud ($P = [-5; 5]^3 $) moving side to side ( $x = 5 \sin (\frac{t}{\pi})$ ).
A crop box ($B = [-1;1]^3$) in ray intersection mode. `negative: false`.

[Screencast from 2025年11月06日 13時51分21秒.webm](https://github.com/user-attachments/assets/187eddd8-6429-4560-88dd-5d68a562a023)

A cube-shaped pointcloud ($P = [-5; 5]^3 $) moving side to side ($x = 5 \sin (\frac{t}{\pi})$) and rotating ( $qx = \cos (\frac{t}{4 \pi})$, $qy = \sin (\frac{t}{2 \pi})$ ).
A crop box ($B = [-1;1]^3$) in ray intersection mode. `negative: false`.

[Screencast from 2025年11月06日 13時50分33秒.webm](https://github.com/user-attachments/assets/5e8c0d30-4904-411e-b8aa-52174dc41ab4)

The same setup, but with `negative: true`.

[a](https://github.com/user-attachments/assets/0c917e7b-e58c-441a-8bc1-a367e2b36d60)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/sensing/autoware_pointcloud_preprocessor/src/crop_box_filter/crop_box_filter_node.cpp
+++ b/sensing/autoware_pointcloud_preprocessor/src/crop_box_filter/crop_box_filter_node.cpp
@@ -81,8 +81,7 @@ bool does_line_segment_intersect_crop_box(
 
   Eigen::Vector3f ray_origin = from_point.head<3>();
   Eigen::Vector3f ray_direction = to_point.head<3>() - from_point.head<3>();
-  Eigen::Vector3f ray_direction_inv = {
-    1.0f / ray_direction.x(), 1.0f / ray_direction.y(), 1.0f / ray_direction.z()};
+  Eigen::Vector3f ray_direction_inv = ray_direction.cwiseInverse();
 
   Eigen::Vector3f box_min = {box.min_x, box.min_y, box.min_z};
   Eigen::Vector3f box_max = {box.max_x, box.max_y, box.max_z};


mojomex:
While the approach works as designed, it's challenging to approximate the (non-cuboid) vehicle using only axis aligned bounding boxes. Thus, we have chosen to achieve the same functionality using [Nebula's mask filter](https://tier4.github.io/nebula/filters/#downsample-mask-filter).

PR #11600, opened 2025-11-06
feat(autoware_cuda_pointcloud_preprocessor): crop box ray intersection mode

mojomex:
## Description

This PR is an exact copy of 

- #11597, 

for the CUDA version of pointcloud preprocessor.

Subtle differences:

- CUDA preprocessor supports multiple crop boxes, and each of them can have ray mode enabled/disabled
- CUDA preprocessor does not output a polygon visualizing the crop box(es), so the extension of that code in the CPU version to display the ray FoV has not been added here.
- No unit tests for now, because
  - They don't run in CI right now anyway
  - Would be a verbatim copy of the CPU version
  - I am planning on unifying our preprocessor testing methodology so all (CPU, CUDA, ...) implementations can be checked against the same tests

For design and docs, please refer to the above PR.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

Visual confirmation with two crop boxes:

1. a crop box ($B_1 = [-1;1]^3$) in ray intersection mode. `negative: false`.
2. a crop box ($B_2 = [-4;4]^3$) in standard mode. `negative: true`.

First scenario: a cube-shaped pointcloud ($P = [-5; 5]^3 $) moving side to side ( $x = 5 \sin (\frac{t}{\pi})$ ).

[](https://github.com/user-attachments/assets/db13b6ee-c50e-4272-a91f-b875d0ce93af)

Second scenario: a cube-shaped pointcloud ($P = [-5; 5]^3 $) moving side to side ( $x = 5 \sin (\frac{t}{\pi})$ ) and rotating ( $qx = \cos (\frac{t}{4 \pi})$, $qy = \sin (\frac{t}{2 \pi})$ ).

[](https://github.com/user-attachments/assets/e6469310-a98c-49f7-a87b-c3eb0ca90059)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_pointcloud_preprocessor/common_kernels.cu
+++ b/sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_pointcloud_preprocessor/common_kernels.cu
@@ -15,6 +15,7 @@
 #include "autoware/cuda_pointcloud_preprocessor/common_kernels.hpp"
 #include "autoware/cuda_pointcloud_preprocessor/point_types.hpp"
 #include "autoware/cuda_pointcloud_preprocessor/types.hpp"
+
 #include <Eigen/src/Core/Matrix.h>
 
 namespace autoware::cuda_pointcloud_preprocessor


mojomex:
While the approach works as designed, it's challenging to approximate the (non-cuboid) vehicle using only axis aligned bounding boxes. Thus, we have chosen to achieve the same functionality using [Nebula's mask filter](https://tier4.github.io/nebula/filters/#downsample-mask-filter).

mojomex:
While the approach works as designed, it's challenging to approximate the (non-cuboid) vehicle using only axis aligned bounding boxes. Thus, we have chosen to achieve the same functionality using[Nebula's mask filter](https://tier4.github.io/nebula/filters/#downsample-mask-filter).

PR #11632, opened 2025-11-13
feat(autoware_pointcloud_preprocessor): empty cloud is valid for cloud info

amadeuszsz:
## Description
So far we considered empty cloud as invalid input. However, this does not mean that cloud is corrupt. This PR considers empty clouds as correct input and therefore set `concatenation_success` as true. Note that algorithm of concatenation was not modified, but only info message.

## Related links

- [TIER IV INTERNAL LINK](https://star4.slack.com/archives/C057JDFS5M2/p1762936061651919)

## How was this PR tested?

- Added unit test.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


SamratThapa120 (approved):
Thanks for the quick support 🤗 🤗 . LGTM

YoshiRi:
Update source branch to rerun CI.

PR #11634, opened 2025-11-13
feat(shape_estimation): correcting yaw (#11309)

Owen-Liuyuxuan:
## Description

- Correcting cluster yaw angle estimated with fitLShape function when a cluster exists in front of or behind the own behicle

https://github.com/autowarefoundation/autoware_universe/pull/11309

Re-Created in my fork for continuous development.

## To Users

Please check the python script:

https://gist.github.com/Owen-Liuyuxuan/55da9fa9061522fc35841b345795fd3d

to experience with the updated logics.

### Contents

https://github.com/autowarefoundation/autoware_universe/pull/11309#issuecomment-3332393418

> I've organized the relevance of the conditions for the correctWithDefaultValue function here.
> https://github.com/arayabrain/autoware.universe.origin/blob/180332a4957e9434d7a4723a8957b7651ac0edde/perception/autoware_shape_estimation/resource/correctWithDefaultValue.svg
> 
> The primary issue causing rotation occurs within the range indicated by the red box. Within this range, when the visible point clouds are extremely small, the farthest edge is frequently swapped, causing the edge being corrected to change often.
> 
> **Before**
> image
> 
> Therefore, I unified the behavior so that when the number of point clouds is extremely low, the correction is directed toward the farthest edge while maintaining the current conditions as much as possible, as follows.
> 
> **After**
> image
> 
> image
> 
> [shape_estimation_after_mini.webm](https://github.com/user-attachments/assets/31cd6b23-a794-4262-8c9c-6fa158b3705b)
> 

## Related links

**Parent Issue:**

- [Link](https://github.com/autowarefoundation/autoware_universe/pull/11309)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/concatenate_data/concatenate_and_time_sync_node.ipp
+++ b/sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/concatenate_data/concatenate_and_time_sync_node.ipp
@@ -340,9 +340,9 @@ void PointCloudConcatenateDataSynchronizerComponentTemplated<MsgTraits>::publish
   const double now_sec = this->get_clock()->now().seconds();
 
   for (const auto & [topic, stamp] : concatenated_cloud_result.topic_to_original_stamp_map) {
-    const double latency = (now_sec - stamp) * 1000.0;  // ms
-    topic_to_pipeline_latency_map[topic] = latency;
-    max_pipeline_latency = std::max(max_pipeline_latency, latency);
+    const double latency_ms = (now_sec - stamp) * 1000.0;  // ms
+    topic_to_pipeline_latency_map[topic] = latency_ms;
+    max_pipeline_latency = std::max(max_pipeline_latency, latency_ms);
   }
 
   diagnostic_info.publish_pointcloud = publish_pointcloud;
@@ -484,7 +484,8 @@ void PointCloudConcatenateDataSynchronizerComponentTemplated<MsgTraits>::check_c
     }
     const auto latency_it = diagnostic_info.topic_to_pipeline_latency_map.find(topic);
     if (latency_it != diagnostic_info.topic_to_pipeline_latency_map.end()) {
-      diagnostics_interface_->add_key_value("Latency (s): " + topic, latency_it->second);
+      double latency_ms = latency_it->second;
+      diagnostics_interface_->add_key_value("Latency (ms): " + topic, latency_ms);
     }
   }
 


Owen-Liuyuxuan:
@a-maumau @technolojin @shmpwk 
Because of the business problem, I re-implemented the PRs from their repos into my personal forks. Please mind considering continue the test in this PR.

Owen-Liuyuxuan:
Please click to view larger mermaid graph

[](https://mermaid.live/edit#pako:eNqdl99u4kYUxl9lNFV7UUHqfxiwdlMdN0SqlE1WaVZVC6jy2mPwLnjQMDSkbN6iV71ZtZe97hP1CfoInbENPjaYkDgXxJzznXNm-M0H3tCQR4x6NJ7x-3AaCEnu_FFKyHL1fiKCxZRcsQlLo-GIkhEd64i-rszhf59__4fcsjn_lUXkik-SsIxaKvr5LwJRtB-zdexP8oZHSZzsh53hv3_8Td6lapZ0UgmrKfRLZTifxVwwNZw_uLy5HZB3by_gbkDa5Gax4MtEMvI2SMSS-CJQFdECfHNYS7lgkoWSRa_ei2_Ov08jtiZqwph8SSzy-jUxkJS02-fEtzbZFsyT9Jf7JJJT8oqYS_m1pV7nwTp_79vHncrSqk8_seUn4tsbW0S7zBlLJ5VUG6U6w8FaqqUTpSCSk4jFwWomC9G4prnmStIZ3jK5Eim5DGZLNq5NkKW4Gz12XqM29940LpqmWxm8vsQuyuw1zJ1pxjVJNlO_YWy3TDGN_ZxDWEAsmVBUwOXd4PZ0KODlUEAOBWgoFPn7IGQVrm8QLuGUhR93mweIDzjOByA-4EQ-oOQDGviAkg94Bh-A-ICjfADiA07jA0o-oIEPKPmAU_nIbQOiDwecQ70bhCyVjc6hKKnlbCnBOblHQGESPlf7qHaRfEWsNKfpPpHTJCU5C0KbHTpJmAYf7O1e6QpHPAB9gj44G52Nu-D6Dq7f2dZXsx2t76D67kZnN9XHpgHdU-dHRx162fyqdkFhtnMNxgM93K7fsJya9RSivJtpZMvB7XT7hnamgfuZZsP66g0LWdHReoaX5ajW7OwpUuEEUmFrXFvnOkDq4cNcMSzlWJl6oHBgItNzsStwXroeLoBtrPSx44wANjIFbtYV72Pus4MfPTIIxOyBCPZBrTjh6bg-eV7C3WQ_R2rf4viDzyrC9UWG0_nhvcAeWOJ-_DgBdi6FOx6hEXXAqEOJ-lO71kOtFOsnHy2osA6I9aOHCzDr6sY6_XTp7EpH-6TTtZMVHZ3jp0s-zJj-RRQns5n3RRyH6motpeAfmb411FXctrM-nr1Yl0LYCsNQS3dCw9DSI0IfXtwSKWOnogyCJ5T2C6dVB-zFSvfZ05bqK7OySShQ3XcUsCvtUMApAqyn_0apDtEWnTMxD5JIPfhsst5U-dacjain_i0AG9FR-qhSg5XkPzykIfWkWLEWFXw1mVIv1ly16GoRBZJdJIFy6_k2ZRGkP3O-u2VRIrl4kz9oZc9bWQr1NnRNPdswzvrdbqdj95xux-y16AP12qZld89s1-h0-z3Xtfq289iiv2VFjbOO7dqOaZhmx3Isw3BadCL0YooBFehMfMdXqaRe__F_hwpk4g)

image
image


I think the update is larger than it claimed

technolojin:
@a-maumau Can you test this update?

a-maumau:
I checked this PR with some rosbags and confirmed the results are almost the same. It might handle some cases better than the original one.

[shape_estimation_cmp_red_original_blue_new_one.webm](https://github.com/user-attachments/assets/304c3fff-9605-48a0-a41e-f82ee4e5a473)

red: original implementation, blue: this PR, purple: the same result between original implementation and this PR.
I use the same input to the nodes.

technolojin:
after improving the readability, I think this change could be introduced!

perception/autoware_shape_estimation/lib/corrector/utils.cpp
```diff
@@ -148,24 +149,37 @@ bool correctWithDefaultValue(
   }
   // Case: Adjacent edges (different scenarios based on which points fit ranges)
   else {
-    // Case 3: Both first and second points are within width range
-    if (first_in_width_range && second_in_width_range) {
-      correction_vector = v_point.at(most_distant_index);
+    // Case 3: Both first and second points are below max width
```
Owen-Liuyuxuan, line 152: This part correponding to the original descriptions "After" charts.

Owen-Liuyuxuan:
@technolojin  Please check the current state of the PR. I believe it has been well-adapted to the new re-factor and the updated part is well-documented. It works well in avoiding aggressive expansion when both edges are small.

perception/autoware_shape_estimation/lib/corrector/utils.cpp
```diff
@@ -134,8 +134,9 @@ bool correctWithDefaultValue(
      0);
 
   if (are_opposite_edges) {
-    // Case 1: First point fits width range, third point is below max length
-    if (first_in_width_range && third_below_max_length) {
+    // Case 1: First point below max width (no min constraint), third point is below max length
+    // Bug Fix 1: Removed min_width requirement to allow correction when edges are very small
+    if (first_below_max_width && third_below_max_length) {
```
technolojin, line 139: case when are_opposite_edges=True

1. object at side: case 1, adjust length seems OK

Screenshot from 2025-12-01 20-30-10



2. object in front:  (low probability than 1)
   - case 1, in case of leading object in lane, faulty extension of length(width in real) can happen. 
   - case 2 may not happen unless it is actually crossing vehicle

Screenshot from 2025-12-01 20-30-16

Owen-Liuyuxuan: @technolojin 
Let's please align the code behavior now.

For your case 1, notice that we will "extend the third vector to the width" (which is the mid point of the ③ edge), which will be the vector of 
`Eigen::Vector2d(shape.dimensions.x / 2.0, 0.0)` .

image


image

Owen-Liuyuxuan: 2. For object in front, most of the cases, it will fall into are_opposite_edges=False. 

First_vector = ③
second_vector= ①
`not first_in_width_range && second_in_width_range`

It should enlarge to length size
image

technolojin: > 2. For object in front, most of the cases, it will fall into are_opposite_edges=False.
> 
> 
> First_vector = ③ second_vector= ① `not first_in_width_range && second_in_width_range`
> 
> It should enlarge to length size image

The number I indicated is order of farthest edge. the farthest edge is not always opposite edge (it usually happen aspect ratio is large at in certain position & angle, which are picked-out).

`are_opposite_edges` is necessary because the logic is based on sorted edges. the top two edges are not always perpendicular.

replacing it to "closest/farthest corner and edges connected", it can be avoided.

Owen-Liuyuxuan: @technolojin 
The update happens in your case should look like this one?
image

technolojin: how about when 3 is in width range? 

ex) following trailer behind stops close to ego vehicle. side edges are invisible, so cluster is very thin.

technolojin: by the way, its very good simulator for this PR

Owen-Liuyuxuan: > how about when 3 is in width range?
> 
> ex) following trailer behind stops close to ego vehicle. side edges are invisible, so cluster is very thin.'

I guess you are talking about this one

- 3 is in width range (1.4m)
image

- 3 is below min-width(0.9m) -- same result
image

technolojin: DIM X: 0.3m
DIM Y: 1.7m
POS X: 1.5m
POS Y: 0m

Owen-Liuyuxuan: I manually trimmed the min-dim to 0.3m for this case. But it looks normal, you can see from the code too it is case 1
image

PR #11660, opened 2025-11-18
docs(autoware_lidar_frnet): add UML diagram for processing flow

kyoichi-sugahara:
## Description

I have added a processing flow diagram to visualize the logic flow. This addition is intended to facilitate future discussions regarding code improvements, such as multi-threading and batch processing optimization.

**Note:**
The diagram is designed for a high-level understanding and does not aim to be strictly comprehensive or cover every single processing step; therefore, some details may be omitted.

I am open to adjusting the granularity of the processes described in the flow. If there are specific policies or preferences regarding how detailed these steps should be, please provide feedback in the comments.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_lidar_frnet/README.md
+++ b/perception/autoware_lidar_frnet/README.md
@@ -10,6 +10,114 @@ The implementation is based on the FRNet [1] project. It uses TensorRT library f
 
 We trained the models using AWML [2].
 
+### Processing Flow
+
+The following diagram shows the processing flow of the `autoware_lidar_frnet` package using UML Activity Diagram notation:
+
+```plantuml
+@startuml
+!theme plain
+skinparam activity {
+    BackgroundColor #FFFFFF
+    BorderColor #000000
+    FontSize 12
+}
+
+start
+
+note right
+  **Entry Point: cloudCallback**
+  Triggered when PointCloud2 message arrives
+end note
+
+:Receive Input Point Cloud\n:PointCloud2;
+
+if (Has active subscribers?) then (yes)
+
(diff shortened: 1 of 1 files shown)


amadeuszsz (commented):
@kyoichi-sugahara 
Thank you for your PR!

We definitely need such a **inner-algorithm** description. However, since proposed diagram is not C++ UML architecture design (not enough details), in some places it seems to me too detailed which increases maintenance cost. 

For example, you could just use a single block "Copy outputs from device to host" instead of pointing each tensor separately.

On the other hand, diagram is not detailed enough, or let's say - too ambiguous. The greatness of such a diagrams is that most of users can take a look and understand the whole pipeline. Let me start from beginning of the diagram:
- Input validation does not show what happens if input is invalid.
- I don't what "against profile" means. What is the profile? (I know, but any user even with technical background will get confused). What happens if input is invalid? Wouldn't be better to write it as "Check the input point cloud’s point count against the min/max thresholds". 
- What yaw and pitch angles refer to? You could write that it is about sensor's specification, but having in mind previously described detail level, "Project 3D points to 2D frustum" is enough.
- And so on.

Please consider to update this diagram.

kyoichi-sugahara:
@amadeuszsz 
Thank you so much for your comment!
I agree with your feedback!

Can I try updating diagram based on your feedback? Or, if you don't think this diagram is strictly necessary, I'm fine with closing this PR :sweat_smile: because the main purpose of this PR was to clarify my understanding, and I'm already satisfied.

PR #11675, opened 2025-11-19
fix(centerpoint): fix insufficient buffer clearing

technolojin:
## Description

the centerpoint had issue on lower detection score on the left side of the ego vehicle.

the reason may due to
* no time to clear memory in high gpu usage
* buffer clear before generateSweepPoints missing

when the number of point fluctuate, the remained point data could be fed to the inference.

1. frame 1
  - 100000 points generated
  - All written to points_aux_d_[0...99,999]
2. frame 2
  - Only 80,000 points generated (maybe due to occlusion, sensor angle, etc.)
  - Written to points_aux_d_[0...79,999]
points_aux_d_[80,000...99,999] STILL HAS OLD POINTS FROM FRAME 1!

## Related links

**Parent Issue:**

- Link

## How was this PR tested?
visualize feature layer as grid https://github.com/autowarefoundation/autoware_universe/pull/11673

* before

Screenshot from 2025-11-19 09-37-43

* after

Screenshot from 2025-11-19 10-54-11

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_lidar_centerpoint/lib/centerpoint_trt.cpp
+++ b/perception/autoware_lidar_centerpoint/lib/centerpoint_trt.cpp
@@ -40,14 +40,14 @@ CenterPointTRT::CenterPointTRT(
   const DensificationParam & densification_param, const CenterPointConfig & config)
 : config_(config)
 {
+  cudaStreamCreate(&stream_);
+
   vg_ptr_ = std::make_unique<VoxelGenerator>(densification_param, config_, stream_);
   pre_proc_ptr_ = std::make_unique<PreprocessCuda>(config_, stream_);
   post_proc_ptr_ = std::make_unique<PostProcessCUDA>(config_, stream_);
 
   initPtr();
   initTrt(encoder_param, head_param);
-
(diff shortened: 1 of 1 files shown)


ktro2828 (approved):
LGTM

KSeangTan:
Sorry to add another unrelated comment, for safety reason, do you think should we move 
https://github.com/autowarefoundation/autoware_universe/blob/main/perception/autoware_lidar_centerpoint/lib/centerpoint_trt.cpp#L50

to the line before line 43?

technolojin:
> Sorry to add another unrelated comment, for safety reason, do you think should we move https://github.com/autowarefoundation/autoware_universe/blob/main/perception/autoware_lidar_centerpoint/lib/centerpoint_trt.cpp#L50
> 
> to the line before line 43?

I agree.
fixed https://github.com/autowarefoundation/autoware_universe/pull/11675#issuecomment-3550379827

SamratThapa120 (approved):
LGTM 💯

PR #11679, opened 2025-11-20
fix(multi object tracker): fix covariance orientation bug

technolojin:
## Description

Orientation of position covariance conversion has a bug.

* Before

Screenshot from 2025-11-20 10-43-12

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


technolojin:
fixed https://github.com/autowarefoundation/autoware_universe/pull/11683

PR #11683, opened 2025-11-20
fix(multi-object-tracker): correct sign of lateral and longitudinal velocity terms in state prediction matrix

technolojin:
## Description

Bug has been introduced by https://github.com/autowarefoundation/autoware_universe/pull/11230

Correct sign of lateral and longitudinal velocity terms in state prediction matrix

* before

Screenshot from 2025-11-20 18-39-33

* after

Screenshot from 2025-11-20 18-40-08

To visualize object position covariance correctly, https://github.com/autowarefoundation/autoware_rviz_plugins/pull/28 is required.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


YoshiRi (approved):
LGTM

PR #11699, opened 2025-11-28
fix(crosswalk): update the previous stop pose before using it

maxime-clem:
## Description

Fix an issue with the `crosswalk` that could insert a stop pose that was calculated for a previous path, leading to yaw or distance deviation issues.
This PR fixes the issue by always re-projecting the stop pose on the updated path.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

Psim + rosbag
Evaluator (TIER IV INTERNAL LINK): https://evaluation.tier4.jp/evaluation/reports/be5ee509-7353-53a5-b828-0d950fbea994?project_id=prd_jt

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_velocity_planner/autoware_behavior_velocity_crosswalk_module/src/scene_crosswalk.cpp
+++ b/planning/behavior_velocity_planner/autoware_behavior_velocity_crosswalk_module/src/scene_crosswalk.cpp
@@ -1268,7 +1268,7 @@ CrosswalkModule::getNearestStopFactorAndReason(
     return {std::nullopt, ""};
   }
   constexpr auto previous_stop_reuse_margin =
-    3.0;  // [m] reuse the previous stop pose if it is within the margin
+    1.0;  // [m] reuse the previous stop pose if it is within the margin
   const auto dist_to_stop = get_distance_to_stop(nearest_stop_and_reason->first);
   if (previous_stop_pose_) {
     const auto previous_stop_arc_length = motion_utils::calcSignedArcLength(


planning/behavior_velocity_planner/autoware_behavior_velocity_crosswalk_module/src/scene_crosswalk.cpp
```diff
@@ -1269,12 +1270,19 @@ CrosswalkModule::getNearestStopFactorAndReason(
...
-  if (use_previous_stop_pose) {
-    previous_stop_pose_->target_object_ids = nearest_stop_and_reason->first->target_object_ids;
-    return {previous_stop_pose_, nearest_stop_and_reason->second};
+  if (previous_stop_pose_) {
+    const auto previous_stop_arc_length = motion_utils::calcSignedArcLength(
+      ego_path.points, 0UL, previous_stop_pose_->stop_pose.position);
+    previous_stop_pose_->stop_pose =
+      motion_utils::calcInterpolatedPose(ego_path.points, previous_stop_arc_length);
+    const auto dist_to_previous_stop = get_distance_to_stop(previous_stop_pose_);
+    const auto use_previous_stop_pose =
```
yuki-takagi-66, line 1279: I don't think this good enough.
Keeping the first position is not works well in some situation.
I'll comments soon.

maxime-clem: the logic to use the previous stop pose and to keep the first stop pose among the previous and the new ones is already implemented. This PR is only to fix an issue when using the previous stop pose.

yuki-takagi-66: Sorry for the late comment.
I realize this is an issue from the previous PR, not this one, and I think this PR improves the situation.

yuki-takagi-66 (approved):

PR #11702, opened 2025-11-28
feat(crosswalk): update obstruction prevention

N-Iemura:
## Description

The naming of the `enable_obstruction_prevention` parameter was misleading. While it suggested control over all obstruction prevention, its effect was limited to intersections. Specifically, setting enable_obstruction_prevention to false would only skip the main obstruction prevention logic at intersections with traffic lights.

This led to an issue where, if a location was not classified as an intersection, the obstruction prevention logic would not be skipped, regardless of the parameter's setting.

To address this, the `enable_obstruction_prevention` parameter has been repurposed to globally disable the entire feature. A new parameter, `ignore_in_intersection`, has been introduced to replicate the previous behavior of skipping the logic specifically at intersections.

autoware_launch: https://github.com/autowarefoundation/autoware_launch/pull/1713

## How was this PR tested?

A bag file was run on a non-intersection crosswalk, confirming that when the parameter is set to false, the crosswalk module's obstruction prevention feature does not activate, even in the presence of a stopped vehicle.

<table>
   <tr>
    <td align="center"><strong>Before</strong></td>
    <td align="center"><strong>After</strong></td>
  </tr>
  <tr>
    <td align="center">
      <video src="https://github.com/user-attachments/assets/9fc34700-b123-46d3-85e7-b72412b3317d" width="300" controls></video>
    </td>
    <td align="center">
      <video src="https://github.com/user-attachments/assets/6e9203e6-be18-49da-86cd-bec828da75e7" width="300" controls></video>
    </td>
  </tr>
</table>

Internal Evaluator Test 
- Original: [3769/3800](https://evaluation.tier4.jp/evaluation/reports/ce5dc448-d575-5fba-b8da-e1616ac34fdb?project_id=prd_jt)
- With this PR: [3771/3800](https://evaluation.tier4.jp/evaluation/reports/b03599f9-089b-5472-b358-c42945c91c51?project_id=prd_jt)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_velocity_planner/autoware_behavior_velocity_crosswalk_module/config/crosswalk.param.yaml
+++ b/planning/behavior_velocity_planner/autoware_behavior_velocity_crosswalk_module/config/crosswalk.param.yaml
@@ -30,7 +30,7 @@
       # params to prevent stopping on crosswalks due to another vehicle ahead
       obstruction_prevention:
         enable_obstruction_prevention: true # [-] flag to enable obstruction prevention feature
-        ignore_in_intersection: true # [-] if true, skip obstruction prevention for crosswalks in intersection        
+        ignore_in_intersection: true # [-] if true, skip obstruction prevention for crosswalks in intersection
         target_vehicle_velocity: 1.0 # [m/s] threshold velocity whether other vehicles are assumed to be stuck or not.
         max_target_vehicle_lateral_offset: 2.0 # [m] The feature does not handle the vehicles farther than this value to the ego's path.
         required_clearance: 6.0 # [m] clearance to be secured between the ego and the ahead vehicle


planning/behavior_velocity_planner/autoware_behavior_velocity_crosswalk_module/config/crosswalk.param.yaml
```diff
@@ -29,7 +29,8 @@
 
       # params to prevent stopping on crosswalks due to another vehicle ahead
       obstruction_prevention:
-        enable_obstruction_prevention: false # [-] flag to enable obstruction prevention checking for crosswalk which is in intersection
+        enable_obstruction_prevention: true # [-] flag to enable obstruction prevention feature
```
shmpwk, line 32:
```suggestion
        enable_obstruction_prevention: false # [-] flag to enable obstruction prevention feature
```
Please keep the default setting

PR #11711, opened 2025-12-01
fix(lidar_centerpoint): correct index check in shufflePoints_kernel

jakor97:
## Description
This PR fixes a bug in the shufflePoints_kernel function in the `autoware_lidar_centerpoint` preprocessing kernel.

The condition check was incorrectly validating `dst_idx >= points_size` instead of `src_idx >= points_size`, since `src_idx` is the index used to read from the source points array.

This bug caused incorrect copying of point data instead of setting it to INFINITY. It was covered up by https://github.com/autowarefoundation/autoware_universe/pull/11675

The same bug can occur in `autoware_lidar_transfusion`: https://github.com/autowarefoundation/autoware_universe/blob/343a70f84246a2ad9a5ad2ec301f3a714ebad3fd/perception/autoware_lidar_transfusion/lib/preprocess/preprocess_kernel.cu#L109 fix https://github.com/autowarefoundation/autoware_universe/pull/11712

## Related links

**Parent Issue:**

https://github.com/autowarefoundation/autoware_universe/pull/11675

## How was this PR tested?
Before, without the https://github.com/autowarefoundation/autoware_universe/pull/1167 , when visualizing the shuffled points `points_d_` ( :red_circle: red) and input point cloud ( :orange_circle: orange), it could be observed that after driving near obstacles, some incorrect points from the buffer were passed on for processing. 

https://github.com/user-attachments/assets/4af0a1a8-a3a9-4d90-9761-675dc5e0aff2

After this fix, these points no longer appear.
## Notes for reviewers

Fix for the `lidar_transfusion` :
https://github.com/autowarefoundation/autoware_universe/pull/11712
## Interface changes

None.

## Effects on system behavior

None.


YoshiRi (approved):
@jakor97 
Thanks for your work!
It makes sense to me, I will ask maintainers to check it.

BTW can I update your branch so that it can pass some CI?

ktro2828 (approved):

jakor97:
> @jakor97 Thanks for your work! It makes sense to me, I will ask maintainers to check it.
> 
> BTW can I update your branch so that it can pass some CI?

Yes, go ahead. Feel free to update anything thats needed. Thanks!

technolojin (approved):
This change will improve invalid point filling. 

Before: invalid points are filled as zero
After: invalid points are filled as INFINITY

PR #11712, opened 2025-12-01
fix(lidar_transfusion): correct index check in shufflePoints_kernel

jakor97:
## Description
This PR fixes a bug in the shufflePoints_kernel function in the `autoware_lidar_transfusion` preprocessing kernel.

The condition check was incorrectly validating `dst_idx >= points_size` instead of `src_idx >= points_size`. Since `src_idx` is the index used to read from the source points array.

## Related links

**Parent Issue:**

https://github.com/autowarefoundation/autoware_universe/pull/11711

## How was this PR tested?
The same bug as in `lidar_centerpoint` `shufflePoints_kernel` https://github.com/autowarefoundation/autoware_universe/pull/11711
## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


Owen-Liuyuxuan:
Thank you for your update, I believe the update is correct and should fix critical bugs when input point clouds are smaller than `cloud_capacity_`

jakor97:
@YoshiRi Shouldn't this PR be merged before closing?

PR #11721, opened 2025-12-03
feat(hazard_status_converter): use latch level for holding status

isamu-takagi:
## Description

Add option to calculate the emergency_holding status from latch levels instead of topic.
Add option of emergency level threshold.

## Related links

[TIER IV internal link](https://star4.slack.com/archives/C017VB9UG1L/p1764755261143539?thread_ts=1729675438.504999&cid=C017VB9UG1L)

## How was this PR tested?

Run planning simulation and check `/system/emergency/hazard_status` is output as before.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


system/autoware_hazard_status_converter/src/converter.cpp
```diff
@@ -23,6 +24,19 @@ namespace autoware::hazard_status_converter
...
+  } else {
+    emergency_threshold_ = HazardStatus::SINGLE_POINT_FAULT;
+  }
+
+  use_external_emergency_holding_ = declare_parameter<bool>("use_external_emergency_holding");
+  if (use_external_emergency_holding_) {
+    sub_emergency_holding_ =
+      autoware_utils_rclcpp::InterProcessPollingSubscriber<EmergencyHolding>::create_subscription(
+        this, "~/input/emergency_holding");
+  }
```
sasakisasaki: Looking [these lines](https://github.com/autowarefoundation/autoware_universe/pull/11721/files#diff-6786e819257c3cc22d4d79b2106364a5fd3d29174283b147b2ca7d64700fe261R142-R145) too, it seems the `sub_emergency_holding_` is used only when `use_external_emergency_holding_ = true`.

Perhaps using `std::optional<...>` for `sub_emergency_holding_` might be one of the answers (then perform `take_data()` if `sub_emergency_holding_.has_value() == true`). But personally it is okay with the current implementation :+1:

sasakisasaki (approved):
Reviewed, I agree with merging this PR :+1:

system/autoware_hazard_status_converter/src/converter.cpp
isamu-takagi: Replaced with a nullptr check. https://github.com/autowarefoundation/autoware_universe/pull/11721/commits/ef0d62e7975c7f5fab1df48a4782d68da67b0f77

PR #11724, opened 2025-12-04
feat(stream_petr): use dynamic triangle filter for image downsampling  

SamratThapa120:
## Description

When the original image is more than twice as large as the image input to the model, the `resizeAndExtractRoi_kernel` introduces aliasing in the downsampled images. The current implementation of  `resizeAndExtractRoi_kernel` does simple bilinear filtering which is the cause of such aliasing. The aliasing results in degradation of detection performance leading to False positives and poor detection of small/distant objects.

This PR updates the  `resizeAndExtractRoi_kernel` to downsample using a triangle filer which prevents any aliasing artifacts in the downsampled image.

Additionally, the PR makes change to ensure that all image pre-processing streams are synchronized before forward pass is done.

Example: In the following short-clip, there is a fatal false positive (high confidence above 90%) that can be fixed by using the new `resizeAndExtractRoi_kernel`

https://github.com/user-attachments/assets/269b4d2a-0644-40a5-9b5d-8b15bf9f524c

## How was this PR tested?

Tested using various rosbags locally.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

For image of resolution 2880x1860, the kernel time increases from 15 **micro**seconds to 90 **micro**seconds. But this is still negligible compared to the image copy time (host to device: ~5millisecond) and model forward pass time of about 15-20 milliseconds.

### DIFF
--- a/perception/autoware_camera_streampetr/lib/network/preprocess_kernel.cu
+++ b/perception/autoware_camera_streampetr/lib/network/preprocess_kernel.cu
@@ -101,7 +101,8 @@ __global__ void resizeAndExtractRoi_kernel(
   float sum_weight = 0.0f;
 
   // 7. Convolution Loop (Triangle/Bilinear Filter)
-  // TODO(SamratThapa120): If necessary, optimize with im2col, separate horizontal and vertical filtering, etc.
+  // TODO(SamratThapa120): If necessary, optimize with im2col, separate horizontal and vertical
+  // filtering, etc.
   for (int y = y_min; y <= y_max; ++y) {
     // Y-weight: Triangle filter (1 - distance / support)
     float dy = (center_y - y);


ogrex:
Locally confirmed FP detection case solved

Before:Red After:Yellow

https://github.com/user-attachments/assets/8b91fc2b-201e-4d2c-8dd8-b0dfe44b7e1c

ogrex (approved):

perception/autoware_camera_streampetr/lib/network/preprocess_kernel.cu
```diff
@@ -38,73 +38,122 @@
 namespace autoware::camera_streampetr
 {
 
+// Helper to ceil division
+inline int divUp(int a, int b)
```
YoshiRi: This function seems not used?
Is `divup` correct?
ex: https://github.com/autowarefoundation/autoware_universe/blob/7b61e259cada7d788851bf527bcb37ab7d6499ac/perception/autoware_camera_streampetr/lib/network/preprocess_kernel.cu#L170

SamratThapa120: Thanks for pointing out. Fixed in [f3dc8b5](https://github.com/autowarefoundation/autoware_universe/pull/11724/commits/f3dc8b5beab109dccf46d226e2010ec29fe75364)

YoshiRi (approved):
LGTM

perception/autoware_camera_streampetr/lib/network/preprocess_kernel.cu
```diff
@@ -37,74 +37,116 @@
...
+  y_min = max(0, y_min);
+  y_max = min(in_h - 1, y_max);
+
+  // Accumulators
+  float sum_r = 0.0f;
+  float sum_g = 0.0f;
+  float sum_b = 0.0f;
+  float sum_weight = 0.0f;
+
+  // 7. Convolution Loop (Triangle/Bilinear Filter)
```
KSeangTan, line 103: TODO: Maybe we can optimize it by `im2col`

perception/autoware_camera_streampetr/lib/network/preprocess_kernel.cu
```diff
@@ -37,74 +37,116 @@
...
+    sum_b /= sum_weight;
   }
+
+  // 9. Normalize (Mean/Std) and Write to Output (Planar CHW)
+  //    Output layout: [Batch/Camera, Channel, Height, Width]
+  int area = roi_h * roi_w;
+  int out_idx = out_y * roi_w + out_x;
+
+  // Channel 0 (R)
+  output_img[camera_offset + (0 * area + out_idx)] = (sum_r - mean[0]) / std[0];
```
KSeangTan, line 147: Just wondering, these are the values from network input normalization right? So we normalize them in here instead of graph?

SamratThapa120: Yes, they are normalized in preprocessing instead of network graph.

perception/autoware_camera_streampetr/lib/network/preprocess_kernel.cu
SamratThapa120: added 
892769a

PR #11731, opened 2025-12-06
refactor(freespace_planning_algorithms): reduce autoware_utils deps

xmfcx:
## Description

**Parent Issue:**
- https://github.com/autowarefoundation/autoware_universe/issues/10467

This PR reduces the `autoware_utils` dependency into `autoware_utils_geometry` and `autoware_utils_math` so that it can compile faster.

## How was this PR tested?

CI.


xmfcx:
@kosuke55 @TakaHoribe @takayuki5168 @isamu-takagi Hi!
This is a very short and easy review, could you take a look at it?

Issue #11737, opened 2025-12-08
fix: Add empty data guards to perception packages (HIGH SEVERITY)


youtalk:
## Checklist

- [x] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [x] I've searched other issues and no duplicate issues were found.
- [x] I've agreed with the maintainers that I can plan this task.

## Description

Add empty data guards to perception packages that process point clouds and other sensor data without checking for empty inputs. Missing guards can cause undefined behavior or crashes when empty data is received.

## Purpose

Processing empty point clouds without validation can lead to:
- Undefined behavior in PCL functions
- Potential crashes or segmentation faults
- Wasted computation on empty data
- Difficult-to-debug issues in production

**Affected files (14 locations):**

| File | Line | Issue |
|------|------|-------|
| `perception/autoware_euclidean_cluster/src/euclidean_cluster_node.cpp` | 55 | No check before `pcl::fromROSMsg` |
| `perception/autoware_euclidean_cluster/src/voxel_grid_based_euclidean_cluster_node.cpp` | 66-74 | Warns but doesn't early return |
| `perception/autoware_lidar_apollo_instance_segmentation/src/node.cpp` | 57 | No validation |
| `sensing/livox/autoware_livox_tag_filter/src/livox_tag_filter_node.cpp` | 65 | No check |
| `perception/autoware_elevation_map_loader/src/elevation_map_loader_node.cpp` | 214, 298 | No check |
| `perception/autoware_compare_map_segmentation/src/compare_elevation_map_filter/node.cpp` | 75 | No check |
| `perception/autoware_compare_map_segmentation/src/distance_based_compare_map_filter/node.cpp` | 34 | No check |
| `perception/autoware_compare_map_segmentation/src/voxel_distance_based_compare_map_filter/node.cpp` | 34 | No check |
| `perception/autoware_ground_segmentation/src/ray_ground_filter/node.cpp` | 341 | No check |
| `perception/autoware_ground_segmentation/src/ransac_ground_filter/node.cpp` | 234 | No check |
| `perception/autoware_shape_estimation/src/shape_estimation_node.cpp` | 122 | No check |
| `perception/autoware_detected_object_validation/src/obstacle_pointcloud/obstacle_pointcloud_validator.cpp` | 79, 182 | No check |

## Possible approaches

1. **Add early return guards**
   - Add empty data checks at the beginning of callbacks
[...]

PR #11744, opened 2025-12-08
fix(autoware_euclidean_cluster): add empty point cloud guards

youtalk:
## Description

Add validation to check for empty point clouds before processing to prevent undefined behavior in PCL functions and potential crashes.

- Add empty data guards in `euclidean_cluster_node.cpp`
- Add early return in `voxel_grid_based_euclidean_cluster_node.cpp` (previously only warned but continued processing)

## Related links

**Parent Issue:**

- #11737

## How was this PR tested?

- Built successfully in Docker environment with `ghcr.io/autowarefoundation/autoware:universe-devel`
- All 38 tests passed (0 errors, 0 failures)

## Notes for reviewers

The fix follows the recommended pattern from issue #11737:
```cpp
if (msg->data.empty() || msg->width == 0 || msg->height == 0) {
  RCLCPP_DEBUG(get_logger(), "Empty point cloud received, skipping processing");
  return;
}
```

## Interface changes

None.

## Effects on system behavior

When empty point clouds are received, the node will now skip processing and return early instead of potentially causing undefined behavior in PCL functions.

### DIFF
--- a/common/autoware_boundary_departure_checker/include/autoware/boundary_departure_checker/boundary_departure_checker.hpp
+++ b/common/autoware_boundary_departure_checker/include/autoware/boundary_departure_checker/boundary_departure_checker.hpp
@@ -17,7 +17,7 @@
 
 #include "autoware/boundary_departure_checker/parameters.hpp"
 
-#include <autoware_utils/system/time_keeper.hpp>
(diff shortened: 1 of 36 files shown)


youtalk:
@ogrex @technolojin @yukkysaito Please review this.

PR #11745, opened 2025-12-08
fix(autoware_lidar_apollo_instance_segmentation): add empty point cloud guard

youtalk:
## Description

Add validation to check for empty point clouds before processing to prevent undefined behavior and potential crashes.

## Related links

**Parent Issue:**

- #11737

## How was this PR tested?

- Built successfully in Docker environment with `ghcr.io/autowarefoundation/autoware:universe-devel`
- Package has no unit tests

## Notes for reviewers

The fix follows the recommended pattern from issue #11737:
```cpp
if (msg->data.empty() || msg->width == 0 || msg->height == 0) {
  RCLCPP_DEBUG(get_logger(), "Empty point cloud received, skipping processing");
  return;
}
```

## Interface changes

None.

## Effects on system behavior

When empty point clouds are received, the node will now skip processing and return early instead of potentially causing undefined behavior.

### DIFF
--- a/common/autoware_boundary_departure_checker/include/autoware/boundary_departure_checker/boundary_departure_checker.hpp
+++ b/common/autoware_boundary_departure_checker/include/autoware/boundary_departure_checker/boundary_departure_checker.hpp
@@ -17,7 +17,7 @@
 
 #include "autoware/boundary_departure_checker/parameters.hpp"
 
-#include <autoware_utils/system/time_keeper.hpp>
+#include <autoware_utils_debug/time_keeper.hpp>
 #include <autoware_vehicle_info_utils/vehicle_info_utils.hpp>
 #include <rosidl_runtime_cpp/message_initialization.hpp>
 #include <tl_expected/expected.hpp>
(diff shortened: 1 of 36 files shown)


perception/autoware_lidar_apollo_instance_segmentation/src/node.cpp
```diff
@@ -52,7 +52,14 @@ LidarInstanceSegmentationNode::LidarInstanceSegmentationNode(
 void LidarInstanceSegmentationNode::pointCloudCallback(
   const sensor_msgs::msg::PointCloud2::ConstSharedPtr msg)
 {
+
```
youtalk:
```suggestion
```

perception/autoware_lidar_apollo_instance_segmentation/src/node.cpp
```diff
@@ -52,7 +52,13 @@ LidarInstanceSegmentationNode::LidarInstanceSegmentationNode(
 void LidarInstanceSegmentationNode::pointCloudCallback(
   const sensor_msgs::msg::PointCloud2::ConstSharedPtr msg)
 {
+  // check for empty point cloud
+  if (msg->data.empty() || msg->width == 0 || msg->height == 0) {
+    RCLCPP_DEBUG(get_logger(), "Empty point cloud received, skipping processing");
+    return;
+  }
   stop_watch_ptr_->toc("processing_time", true);
+
```
youtalk:
```suggestion

stop_watch_ptr_->toc("processing_time", true);
```

youtalk:
@technolojin @yukkysaito Please review this.

Issue #11738, opened 2025-12-08
refactor: Address Node/Callback Explosion anti-pattern (HIGH SEVERITY)


youtalk:
## Checklist

- [x] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [x] I've searched other issues and no duplicate issues were found.
- [x] I've agreed with the maintainers that I can plan this task.

## Description

The perception pipeline currently runs 14-20+ individual processes for what could be consolidated into fewer processes using ComposableNodeContainers. This causes excessive inter-process communication overhead and resource consumption.

## Purpose

Running many separate processes instead of composable nodes leads to:
- Increased memory usage (each process has its own address space)
- IPC overhead for message passing between nodes
- More complex deployment and debugging
- Increased latency in the perception pipeline

**Affected areas:**

| Component | Issue |
|-----------|-------|
| `sensing/autoware_pointcloud_preprocessor/launch/` | Individual nodes for each filter |
| `perception/autoware_object_range_splitter/` | 15 lines of logic as independent process |
| `perception/autoware_object_velocity_splitter/` | Pass-through filter as independent process |

**Good example (reference):**
- `perception/autoware_ground_segmentation/launch/ground_segmentation.launch.py` - Uses ComposableNodeContainer

## Possible approaches

1. **Consolidate into ComposableNodeContainers**
   - Group related nodes into single container processes
   - Use `ComposableNodeContainer` in launch files
   - Reduces IPC overhead significantly

2. **Convert simple nodes to components**
   - Make standalone nodes available as components
   - Allow flexible composition in launch files
   - Maintain backward compatibility

3. **Merge pass-through filters into parent nodes**
   - For trivial operations like range/velocity splitting
   - Add as optional preprocessing step in consuming nodes
   - Reduces unnecessary node boundaries

## Definition of done

- [ ] Identify all nodes suitable for composition
- [ ] Convert key perception nodes to ComposableNodes
- [ ] Create consolidated launch files using ComposableNodeContainer
- [ ] Verify no regression in functionality
- [ ] Measure and document performance improvement
- [ ] Build and test pass successfully

youtalk:
## Analysis: Current `pointcloud_container` Architecture and Limitations of Related PRs

I've analyzed the relationship between `autoware_launch`'s `pointcloud_container` and the related PRs (#11750, #11751, #11752). Here are my findings:

### Current Architecture

[...]

PR #11752, opened 2025-12-08
feat(autoware_pointcloud_preprocessor): add ComposableNode support to filter launch files

youtalk:
## Description

Add `container_name` argument to filter launch files to support loading nodes into an existing ComposableNodeContainer, addressing the Node/Callback Explosion anti-pattern identified in #11738.

Updated launch files:
- `voxel_grid_downsample_filter_node.launch.xml`
- `crop_box_filter_node.launch.xml`
- `ring_outlier_filter_node.launch.xml`

When `container_name` is empty (default), the node is launched as a standalone node (same behavior as before). When `container_name` is specified, the node is loaded into the existing container using `load_composable_node`.

## Related links

**Parent Issue:**

- #11738

## How was this PR tested?

- Added launch tests for `voxel_grid_downsample_filter_node.launch.xml` as a representative:
  - `test_voxel_grid_launch.py`: Tests standalone mode (container_name is empty)
  - `test_voxel_grid_launch_composable.py`: Tests composable mode (loads into existing container)
- Built and tested with Docker using `colcon build` and `colcon test`

## Notes for reviewers

None.

#### Additions and removals

| Change type | Parameter Name   | Type     | Default Value | Description                                                                 |
|:------------|:-----------------|:---------|:--------------|:----------------------------------------------------------------------------|
| Added       | `container_name` | `string` | `""`          | Container name for composable node. If empty, standalone node is launched. |

## Effects on system behavior

None. Default behavior (standalone node) is preserved when `container_name` is not specified.

### DIFF
--- a/sensing/autoware_pointcloud_preprocessor/CMakeLists.txt
+++ b/sensing/autoware_pointcloud_preprocessor/CMakeLists.txt
@@ -327,4 +327,8 @@ if(BUILD_TESTING)
     TIMEOUT "50"
   )
 
+  find_package(launch_testing_ament_cmake REQUIRED)
+  add_launch_test(test/test_voxel_grid_launch.py)
+  add_launch_test(test/test_voxel_grid_launch_composable.py)
+
 endif()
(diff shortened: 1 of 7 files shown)


youtalk:
Note that in `nebula_node_container.launch.py`, the execution of the composable node is hard-coded.
https://github.com/autowarefoundation/autoware_launch/blob/main/sensor_kit/single_lidar_sensor_kit_launch/single_lidar_common_launch/launch/nebula_node_container.launch.py

PR #11756, opened 2025-12-09
feat(streampetr): class wise confidence threshold

SamratThapa120:
## Description
This pull request updates the per-class confidence threshold handling for object detection in the StreamPetr network, improving the flexibility and accuracy of post-processing. The main changes include switching from a single global threshold to a vector of per-class thresholds, updating configuration, data structures, and CUDA post-processing logic to support this feature. The PR also modifies the `PostprocessCuda` class to allocate the parameter vectors only once in GPU memory.

## How was this PR tested?

Tested locally with various rosbags.

## Notes for reviewers

The new class-wise confidence threshold were computed using AWML
https://github.com/tier4/AWML/blob/515070656aef169ad6ac9518d6f1606eade79b50/autoware_ml/detection3d/evaluation/t4metric/t4metric_v2.py

#### Modifications

| Old     | `confidence_threshold` | `double` | `0.35`         | post processing object confidence threshold|
| New     | `confidence_threshold` | `vector<double>` | ` [0.36, 0.39, 0.38, 0.41, 0.43]`         | post processing pre-class object confidence threshold|

### DIFF
--- a/perception/autoware_camera_streampetr/lib/postprocess/postprocess_kernel.cu
+++ b/perception/autoware_camera_streampetr/lib/postprocess/postprocess_kernel.cu
@@ -15,6 +15,8 @@
 #include "autoware/camera_streampetr/postprocess/circle_nms_kernel.hpp"
 #include "autoware/camera_streampetr/postprocess/postprocess_kernel.hpp"
 
+#include <autoware/cuda_utils/thrust_utils.hpp>
+
 #include <thrust/count.h>
 #include <thrust/device_vector.h>
 #include <thrust/execution_policy.h>
@@ -155,7 +157,7 @@ cudaError_t PostprocessCuda::generateDetectedBoxes3D_launch(
   auto boxes3d_ptr = thrust::device_pointer_cast(boxes3d_d_.get());
 
   // Create stream-aware execution policy
-  auto policy = thrust::cuda::par.on(stream);
+  auto policy = autoware::cuda_utils::thrust_on_stream(stream);
 
   // suppress by score
   const auto num_det_boxes3d = thrust::count_if(


perception/autoware_camera_streampetr/lib/postprocess/postprocess_kernel.cu
```diff
@@ -121,21 +121,25 @@ cudaError_t PostprocessCuda::generateDetectedBoxes3D_launch(
   auto detection_range_d =
     thrust::device_vector<float>(config_.detection_range_.begin(), config_.detection_range_.end());
 
+  auto score_thresholds_d = thrust::device_vector<float>(
```
YoshiRi: I noticed that `score_thresholds_d` is initialized locally inside this function. This triggers `cudaMalloc` and a Host-to-Device copy every frame.

Although the impact might be small at low frequencies, `cudaMalloc` can introduce unpredictable latency (jitter). Since these threshold values are static, could you move `score_thresholds_d` to a member variable and initialize it once in the constructor? This would avoid unnecessary allocation overhead and improve stability.

```cpp
// member variable
thrust::device_vector<float> score_thresholds_d_;
```

YoshiRi: FYI: I noticed that the following existing variables in `generateDetectedBoxes3D_launch` follow the same pattern (local initialization causing repeated allocation/copy):

* `boxes3d_d`
* `yaw_norm_thresholds_d`
* `detection_range_d`

While strictly outside the scope of this PR, refactoring these into member variables (initialized in the constructor) alongside `score_thresholds_d` would further optimize the post-processing step by eliminating per-frame `cudaMalloc` overhead.

KSeangTan: @YoshiRi 

Another note is that we don't need to process every box in `generateBoxes3D_kernel`, for example, we should just stop processing for every box3d confidence lower than the threshold

KSeangTan: FYI @SamratThapa120 

You can refer to this for the initialization 
https://github.com/autowarefoundation/autoware_universe/blob/main/perception/autoware_lidar_centerpoint/lib/postprocess/postprocess_kernel.cu#L140

SamratThapa120: Thanks. Fixed [2e706c0](https://github.com/autowarefoundation/autoware_universe/pull/11756/commits/2e706c06916cf63e0e45df7c9e8a632f84a944bc)

perception/autoware_camera_streampetr/lib/postprocess/postprocess_kernel.cu
```diff
@@ -115,27 +145,25 @@ cudaError_t PostprocessCuda::generateDetectedBoxes3D_launch(
...
-    thrust::raw_pointer_cast(yaw_norm_thresholds_d.data()),
-    thrust::raw_pointer_cast(detection_range_d.data()), thrust::raw_pointer_cast(boxes3d_d.data()));
+    yaw_norm_thresholds_d_.get(), score_thresholds_d_.get(), detection_range_d_.get(),
+    boxes3d_d_.get());
+
+  // Wrap raw pointer with thrust device pointer for thrust algorithms
+  auto boxes3d_ptr = thrust::device_pointer_cast(boxes3d_d_.get());
 
   // suppress by score
   const auto num_det_boxes3d = thrust::count_if(
```
YoshiRi, line 163: It seems that the `thrust` algorithms (`count_if`, `copy_if`, `sort`) are invoked without specifying an execution policy. This causes them to run on the default stream (or implicitly synchronize), effectively ignoring the `cudaStream_t stream` argument passed to this function.

To ensure these operations run asynchronously on the specified stream, please use the `thrust::cuda::par.on(stream)` execution policy.

**Reference:**
- copy if example: [nvidia ref](https://nvidia.github.io/cccl/thrust/api/group__stream__compaction_1ga291bd49963dc7d3d0529fb8ed05de66e.html)
- [example usage in autoware](https://github.com/autowarefoundation/autoware_universe/blob/5cade4321a6523963b08dce3a50116abe0afcbf2/perception/autoware_ptv3/lib/postprocess/postprocess_kernel.cu#L130-L131) 
- There are also [autoware cuda utils implementation](https://github.com/autowarefoundation/autoware_universe/blob/5cade4321a6523963b08dce3a50116abe0afcbf2/sensing/autoware_cuda_utils/include/autoware/cuda_utils/thrust_utils.hpp#L27-L28) on this feature. It seems to be used in [cuda pointcloud preprocessor](https://github.com/autowarefoundation/autoware_universe/blob/5cade4321a6523963b08dce3a50116abe0afcbf2/sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor.cu#L432-L433)

**Example fix:**
```cpp
#include <thrust/execution_policy.h> // Required header

// ...
auto policy = thrust::cuda::par.on(stream);

// Pass the policy as the first argument
thrust::count_if(policy, ...);
thrust::copy_if(policy, ...);
thrust::sort(policy, ...);

File: perception/autoware_camera_streampetr/lib/postprocess/postprocess_kernel.cu [hunk 3]
```diff
@@ -115,27 +145,25 @@ cudaError_t PostprocessCuda::generateDetectedBoxes3D_launch(
...
-    thrust::device, boxes3d_d.begin(), boxes3d_d.end(), is_score_greater(config_.score_threshold_));
+    thrust::device, boxes3d_ptr, boxes3d_ptr + config_.num_proposals_,
+    is_score_greater_classwise(score_thresholds_d_.get()));
   if (num_det_boxes3d == 0) {
     return cudaGetLastError();
   }
   thrust::device_vector<Box3D> det_boxes3d_d(num_det_boxes3d);
   thrust::copy_if(
```
YoshiRi, line 170: same comment with above. use policy to control stream.

perception/autoware_camera_streampetr/lib/postprocess/postprocess_kernel.cu
```diff
@@ -115,27 +145,25 @@ cudaError_t PostprocessCuda::generateDetectedBoxes3D_launch(
...
-    thrust::device, boxes3d_d.begin(), boxes3d_d.end(), det_boxes3d_d.begin(),
-    is_score_greater(config_.score_threshold_));
+    thrust::device, boxes3d_ptr, boxes3d_ptr + config_.num_proposals_, det_boxes3d_d.begin(),
+    is_score_greater_classwise(score_thresholds_d_.get()));
 
   // sort by score
   thrust::sort(det_boxes3d_d.begin(), det_boxes3d_d.end(), score_greater());
```
YoshiRi: same comment with above. use policy to control stream.

YoshiRi (approved):
Though we have some space to be improved, IMO it is LGTM for the scope of this PR.

perception/autoware_camera_streampetr/lib/postprocess/postprocess_kernel.cu
SamratThapa120: Thanks, I agree
[de83b15](https://github.com/autowarefoundation/autoware_universe/pull/11756/commits/de83b1553dc7559f34f8b07aa873c2cd807dcabf)

PR #11764, opened 2025-12-10
feat: dag implementation of cuda point cloud preprocessor

Owen-Liuyuxuan:
## Description

**YAML-defined processing pipeline** that executes GPU-accelerated pointcloud filters in a directed acyclic graph (DAG) structure.

In general, my target is to

1. Accomplish what we have done in cuda_preprocessing_node using a flexible dynamic configuration.
2. Allow more complicated branching management so that we can further centralize the point cloud processing into a single component and launch fewer ROS nodes for similar functions.

## Example Workflow

```text
Input (ROS Topic)
  ↓
[INTERNALIZED] Organize
  ↓
Transform Filter
  ├─→ [COPY] → Downsample Filter → [Finalize + Publish] → Topic A
  │
  └─→ [NO COPY] → CropBox Filter
                    ↓
                  Distortion Filter
                    ├─→ [COPY] → [Finalize + Publish] → Topic B
                    │
                    └─→ [NO COPY] → Ring Outlier Filter
                                      ↓
                                    [Finalize + Publish] → Topic C
```

### 1. **Configuration-Driven Pipeline**

- Define entire processing workflow in YAML
- No recompilation needed to change pipeline

### 2. **Intellegent Branching**

- Only produce copy when there is output (finalize filter will copy to create unique_ptr) or multiple nodes connected to same input source (branch out)
- Configurable topic publications at any nodes of the graph.

### 3. **Zero-Copy GPU Processing**

- Data stays on GPU throughout pipeline
- Filters pass `PointcloudProcessingState` (raw GPU pointer + metadata)
- Only sync at entry (organize) and exit (publish)
- **Same performance** of the current hard-coded node. But allow much flexibility.

### 4. **Immediate Publishing**

- Outputs published **RIGHT AFTER** production
- Prevents modification by downstream filters

### 5. **Smart Finalization**

- **Regular filters** (cropbox, outlier): Apply masks at finalize
- **Downsample filter**: Already compacted, skip masks
- Uses `is_finalized` flag to determine path

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

Tested with JPNTaxi pointcloud processing bags.

- When the configuration contains the same workflow of the existing autoware_cuda_processing_node, we can produce same point clouds in output while the performance is the same.
- I also tested branching out intermediate topics for the down-sampling processings in the same node. I tested the downsampling produce expected pointclouds, and the branching - multi publisher mechanism also works as expected.

## Interface changes

None.

## Effects on system behavior

None.


Owen-Liuyuxuan:
At the current stage, the code was tested in the JPTaxi project codes.
And moved to Autoware main code base, run local pre-commit, and `colcon build` passed.

Owen-Liuyuxuan:
Same workload on JPTaxi data

Time is faster because DAG is on Desktop and the raw output is on vehicle ECU.
image

Number of points from the original CUDA workflow (different from the CPU workflow) and the DAG workflow
image

The outputs are fully aligned.
image

mojomex (commented):
Thank you for the PR! I agree that CUDA preprocessor should be split into individual components to make it flexible.

I think your approach goes in the right direction, but it seems to me that the DAG configuration, filter registry, and executor are fulfilling very similar roles to ROS2 launch, topics, and executors.

Given that the added DAG functionality makes this PR over 6000 lines, I would suggest re-using what's already there:
* make each filter a node (all with same input/output topics)
* turn FilterContext into a ROS2 message type or publish pointcloud+mask separately and use message_filters to reunite
* use CUDA blackboard pub/sub for GPU zero-copy
* use ROS2 launch to let the user configure the pipeline

I'd like to hear your opinion, and that of the other maintainers though!

Owen-Liuyuxuan:
@mojomex  
Thanks for the suggestion and your patience in reading through the code.

Regarding the current PR's design goals:
- **Flexible configuration** – This aligns well with both the current approach and your proposed ROS2 launch-based architecture. Proper filter design would make ROS2 node launching viable.
- **Limiting the number of ROS2 nodes** – I'm uncertain about its priority, but splitting into separate nodes would inevitably increase the number of nodes/topics.

Regarding your suggestion, I see the main technical challenges as:

> - Turn FilterContext into a ROS2 message type or publish pointcloud+mask separately and use message_filters to reunite
> - Use CUDA blackboard pub/sub for GPU zero-copy

- FilterContext serialization: Since FilterContext contains GPU memory, converting it to a single joint ROS2 message would be non-trivial. I'm not yet clear on how to achieve this without triggering GPU-to-CPU copies, which would defeat the performance benefits. (Do we need to add new cudablackboard types?)
  
- message_filters performance: While I initially had concerns about synchronization overhead, the implementation appears reasonable upon closer inspection. However:
  - We'd likely need a custom CUDA blackboard type for mask data to maintain zero-copy semantics.
  - Launch files would become significantly more complex, requiring negotiated topic remapping for both point clouds and masks, which could impact maintainability.

I'd appreciate your thoughts on handling GPU memory in ROS2 messages, or whether there are existing patterns.

I hope to have more in-depth communication if you have the time.

PR #11765, opened 2025-12-10
fix(autoware_probabilistic_occupancy_grid_map): prevent local_map leak in occupancy grid map updateOrigin

multimeters:
Why: updateOrigin would copy uninitialized costmap data on the first frame and, when LaserscanBasedOccupancyGridMapNode creates a per-frame OccupancyGridMap (single-frame mode), the CPU path allocated a temporary buffer with new[] that was not freed on early return, causing a leak.
What: Guard the first iteration to only set the origin and return; switch the temporary CPU buffer to std::vector<unsigned char> so it’s always owned and freed.
How: Early-return before any window copy on the first call, replace raw allocation with RAII, leaving CUDA/CPU behavior otherwise unchanged.

### DIFF
--- a/perception/autoware_camera_streampetr/config/tensorrt_stream_petr.param.yaml
+++ b/perception/autoware_camera_streampetr/config/tensorrt_stream_petr.param.yaml
@@ -11,7 +11,7 @@
       num_proposals: 5400
       rois_number: 6
     post_process_params: # These are post-processing parameters to improve the predictions.
-      confidence_threshold: 0.35
+      confidence_threshold: [0.36, 0.39, 0.38, 0.41, 0.43]  # Per-class thresholds: [CAR, TRUCK, BUS, BICYCLE, PEDESTRIAN]
       circle_nms_dist_threshold: 0.0   # turn off circle NMS by setting 0.0, otherwise set to around 0.5
       iou_nms_search_distance_2d: 10.0
       iou_nms_threshold: 0.5           # set to 0 to turn off iou NMS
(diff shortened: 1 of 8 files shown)


perception/autoware_probabilistic_occupancy_grid_map/lib/costmap_2d/occupancy_grid_map_base.cpp
```diff
@@ -148,11 +149,12 @@ void OccupancyGridMapInterface::updateOrigin(double new_origin_x, double new_ori
     return;
 #endif
   } else {
-    local_map = new unsigned char[cell_size_x * cell_size_y];
+    local_map.resize(static_cast<size_t>(cell_size_x) * cell_size_y);
```
YoshiRi: [minor comment]
Using resize with static_cast seems to be good idea. But why not also cast `cell_size_y` ?

YoshiRi (commented):
Thanks for your PR. The idea looks good. Please check a minor comment.

perception/autoware_probabilistic_occupancy_grid_map/lib/costmap_2d/occupancy_grid_map_base.cpp
multimeters: Thanks for the catch! Updated to cast both cell_size_x and cell_size_y to size_t and ran clang-format.

YoshiRi:
@multimeters 
Please fix following DCO error. You may need to add sign off to each commit.

```
Commit sha: [af4fc77](https://github.com/autowarefoundation/autoware_universe/pull/11765/commits/af4fc77f9b5654c160cffdde9ce1a41583aaed79), Author: Libpet-Henry, Committer: Libpet-Henry; The sign-off is missing.
Commit sha: [e0e6397](https://github.com/autowarefoundation/autoware_universe/pull/11765/commits/e0e63973ae31bfe6ce48e8d27b9f9a6ec8b85452), Author: Libpet-Henry, Committer: Libpet-Henry; The sign-off is missing.
```

multimeters:
@YoshiRi Hi! All required checks (including DCO) have passed, and the PR is up to date with main. Please let me know if you’d like any changes. Thanks!

PR #11768, opened 2025-12-11
feat(feature_environment_recognizer)!: add feature environment recognizer based on polygon area of LL2

Motsu-san:
This package determines which area (polygon) in a Lanelet2 map the current position belongs to and identifies the environment ID based on the area's subtype. It can be used to identify environments with rich/poor features for map matching.

By using area-based determination instead of lanelet IDs, information can be directly embedded in the map, avoiding issues where lanelet IDs change when the map is regenerated.

Features:
- Reads areas from the polygon layer of Lanelet2 maps
- Determines which area the current position is within (point-in-polygon check)
- Determines environment ID based on the area's subtype
- Publishes environment ID (integer)

The package helps identify environments where map matching may be challenging due to lack of features (e.g., tunnel straight sections, uniform roads).

## Related links

https://github.com/autowarefoundation/autoware_launch/pull/1717

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/launch/tier4_localization_launch/launch/feature_environment_recognizer/feature_environment_recognizer.launch.xml
+++ b/launch/tier4_localization_launch/launch/feature_environment_recognizer/feature_environment_recognizer.launch.xml
@@ -1,0 +1,11 @@
+<?xml version="1.0"?>
+<launch>
+  <arg name="param_file" description="Path to the parameter file"/>
+  <group>
+    <include file="$(find-pkg-share autoware_feature_environment_recognizer)/launch/feature_environment_recognizer.launch.xml">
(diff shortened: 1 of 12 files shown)


paulsohn:
@Motsu-san cc @TaikiYamada4 @YamatoAndo 

As we decided to migrate tier4 launchers into autoware_launch repository,
a separate PR for tier4 launchers must be created unless this PR is merged before 2026/02/10 07:00 UTC (16:00 JST)

* https://github.com/orgs/autowarefoundation/discussions/6780
* https://github.com/autowarefoundation/autoware/issues/6774

PR #11769, opened 2025-12-11
feat: allow the latest frame from each tlr camera to stay for a little longer

Owen-Liuyuxuan:
## Description

There is a fundamental flaw in TLR fusion when we have two cameras triggered with a discrepancy of less than 0.01s while looking in the same direction at a 10Hz system.

Current fact:
- Fusion node outputs at every single callback.
- Each camera is running at 10Hz, and two cameras on the same direction are triggered in the same phase.

Assuming the triggering time is 0.05, 0.15, 0.25 second, cases like the following graph will happen
image

There will always be cases where only one camera is fused and the other camera in the same direction, **even though triggered correctly,** will be lost for a frame.

### The new design

- If all cameras are triggered correctly, the response time for the traffic light change will not change more than 0.01s for a 10Hz system.
**- Keep the latest frame of each camera for slightly longer (0.11s or 0.12s as proposed here), so that the system will not break easily.**

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

#### Additions and removals

| Added | `message_lifespan_unique`   | `double` | `0.12`         | The maximum timestamp span to be fused if the message is the latest one from a certain camera." |

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_camera_streampetr/config/tensorrt_stream_petr.param.yaml
+++ b/perception/autoware_camera_streampetr/config/tensorrt_stream_petr.param.yaml
@@ -11,7 +11,7 @@
       num_proposals: 5400
       rois_number: 6
     post_process_params: # These are post-processing parameters to improve the predictions.
-      confidence_threshold: 0.35
+      confidence_threshold: [0.36, 0.39, 0.38, 0.41, 0.43]  # Per-class thresholds: [CAR, TRUCK, BUS, BICYCLE, PEDESTRIAN]
       circle_nms_dist_threshold: 0.0   # turn off circle NMS by setting 0.0, otherwise set to around 0.5
       iou_nms_search_distance_2d: 10.0
       iou_nms_threshold: 0.5           # set to 0 to turn off iou NMS
--- a/perception/autoware_camera_streampetr/include/autoware/camera_streampetr/network/network.hpp
+++ b/perception/autoware_camera_streampetr/include/autoware/camera_streampetr/network/network.hpp
@@ -154,7 +154,7 @@ struct NetworkConfig
   double search_distance_2d = 0.0;
   double circle_nms_dist_threshold = 0.0;
   double iou_threshold = 0.0;
-  double confidence_threshold = 0.0;
+  std::vector<double> confidence_thresholds;
   std::vector<std::string> class_names;
   int32_t num_proposals = 0;
   std::vector<double> yaw_norm_thresholds;
--- a/perception/autoware_camera_streampetr/include/autoware/camera_streampetr/postprocess/postprocess_kernel.hpp
+++ b/perception/autoware_camera_streampetr/include/autoware/camera_streampetr/postprocess/postprocess_kernel.hpp
@@ -17,6 +17,8 @@
 
 #include "autoware/camera_streampetr/utils.hpp"
 
+#include <autoware/cuda_utils/cuda_unique_ptr.hpp>
+
 #include <cuda.h>
 #include <cuda_runtime_api.h>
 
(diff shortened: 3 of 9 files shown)


Owen-Liuyuxuan:
Test result in internal  ROSBags (bottom graph is for on-vehicle old results, and upper graph is for re-implementation result)
image

MasatoSaeki:
@Owen-Liuyuxuan 

Thank you for your PR.

I assume like below:
- There are two cameras, camera0, camera2.
- There are two frames, frameA, frameB. (frameB is newer than frameA)

In my understanding, the following occurred:
1. In frameB, assume that arrive camera0 first. Then, publish only frameB info of camera0. (Of course frameA info is old.)
2. After that, arrive camera2. Then publish frameB info of camera0 and camera2. (already removed frameA info.)

The problem you mentioned is `1.` ? 
So publish information from both cameras(= `2.`) right after that, but is it generally not a good idea to publish information from just one camera in the first place?

Owen-Liuyuxuan:
@MasatoSaeki 

> The problem you mentioned is 1. ?

Yes. In order to avoid that, one solution is to directly increase the lifespan (I saw some projects increase the lifespan to 0.2s), but that will make the response to signal change slower because old frames will be preserved like a low-pass filter.

This PR proposes to avoid a time-span low-pass filter while waiting for cameras in the same direction to work, and it works much better in full-synchronized camera setting too.

>  but is it generally not a good idea to publish information from just one camera in the first place?

I think the current code does not have the concept of an entire frame; it is dynamically adapting. If you have a better idea for a more overall refactoring, then it will also be good, but it involves a synchronization strategy implementation similar to point cloud concatenation. The current PR is just considering a patch that preserve the original coding ideas.

MasatoSaeki:
@Owen-Liuyuxuan 

Thank you for your replying!

I have some questions.

1. This PR gets to reduce the opportunity to infer only one camera introducing `message_lifespan_unique` which is bigger than camera frequency. So in the image you attached, we can solve the problem fusing frameB of camera2 and frameA of camera0, right?

2. Actually, it is okay to leave only latest result on each camera. And we do not need the `message_lifespan` I think. What do you think? (I think this implementation is acceptable for now, given that we'll gradually change it over time.)

3. It is not easy to understand `message_lifespan_unique` How about `message_lifespan_for_latest`? (I welcome other suggestions as well.)

---
Note:
Pros:
assumes 10Hz camera and existing multiple cameras which is same phase.
- If one of them estimates `unknown`, it can use the other one's result. So will be stable (=Output stable results without chattering even at the absolute limit of its `message_lifespan`.)
  - set the `message_lifespan_unique` bigger than `0.1`.
  - Setting the `message_lifespan` smaller than `0.1`, we can delete the not latest result on each camera.

Owen-Liuyuxuan:
@MasatoSaeki 

Great, I think we have reached some levels of consensus/common understanding on the current problem and the this PR.

> Actually, it is okay to leave only latest result on each camera. And we do not need the message_lifespan I think. What do you think? (I think this implementation is acceptable for now, given that we'll gradually change it over time.)

In our internal project, that will be OK. But I am not sure, maybe some projects will be considering applying fusion filters over time. If we hardcoded the `keep only latest logic` in code without a parameter, this could break some parameter tuning. For, example, I remember JPTaxi also had TLR life_span for 0.2s in sometime before.

> It is not easy to understand message_lifespan_unique How about message_lifespan_for_latest? (I welcome other suggestions as well.)

I agree, will be fixing the name.

MasatoSaeki:
@Owen-Liuyuxuan 

I am so sorry, this node filter the old frame on each camera in [here](https://github.com/autowarefoundation/autoware_universe/blob/main/perception/autoware_traffic_light_multi_camera_fusion/src/traffic_light_multi_camera_fusion_process.cpp#L84).

So there is no problem to be over the `message_lifespan` `0.1` without delay the response time I think. Could you create like [this image](https://github.com/autowarefoundation/autoware_universe/pull/11769#issuecomment-3640707339) using `message_lifespan = 0.12`?

Owen-Liuyuxuan:
image

You are right, it did improve. I think overall slightly increasing the lifespan can solve the problem.

Issue #11788, opened 2025-12-17
non-monotonic trajectory generated by freespace planner


ralwing:
### Checklist

- [x] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [x] I've searched other issues and no duplicate issues were found.
- [x] I'm convinced that this is not my fault but a bug.

### Description

When `freeespace planner` is generating backwards trajectories, they are sometimes non-monotonic, e.g. can have points (x,y) (1.0 ,0.0); (2.0, 0.0); **(1.5, 0.0)**; (3.0, 0.0); 
This behavior causes the velocity_smoother to sometimes change the point orientation to the opposite, as mentioned in an old issue "[motion velocity smoother generate invalid trajectory from freespace planner output](https://github.com/autowarefoundation/autoware_universe/issues/5938)"

### Expected behavior

Freespace planner trajectory points are monotonic when driving backwards and forwards. 

### Actual behavior

Sample trajectory generated by freespace_planner:

<img width="1422" height="985" alt="Image" src="https://github.com/user-attachments/assets/89585269-7e48-4201-a06d-dda8353bd60c" />

Different trajectory orientations in deg:

<img width="500" height="300" alt="Image" src="https://github.com/user-attachments/assets/686fead0-e769-4f26-9841-d4f98def271f" />

### Steps to reproduce

1. Spawn an ego vehicle on parking lot
2. Park the vehicle in the parking space frontally (observe the trajectory points)
3. Reverse out of the parking space (observe the trajectory points)

### Versions

OS: Ubuntu 24.04
ROS2: humble
Autoware Version: 0.47

kosuke55:
@ralwing 
Thank you for fixing the issue! 
https://github.com/autowarefoundation/autoware_universe/pull/11789

Issue #5938, opened 2023-12-22
motion velocity smoother generate invalid trajectory from freespace planner output:


VRichardJP:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I'm convinced that this is not my fault but a bug.

### Description

freespace_planner trajectory:
![image](https://github.com/autowarefoundation/autoware.universe/assets/18645627/c91fa34e-fc9f-450d-b53f-4dc86b3d51ac)

```bash
clear; ros2 topic echo /planning/scenario_planning/parking/trajectory --once
header:
  stamp:
    sec: 1703230235
    nanosec: 694643853
  frame_id: map
points:
- time_from_start:
    sec: 0
    nanosec: 0
  pose:
    position:
      x: 3704.4996056985747
      y: 73740.54148407953
      z: 0.0
    orientation:
      x: 0.0
      y: 0.0
      z: 0.8876844177650136
      w: 0.46045235850974725
  longitudinal_velocity_mps: -1.3888888359069824
  lateral_velocity_mps: 0.0
  acceleration_mps2: 0.0
  heading_rate_rps: 0.0
  front_wheel_angle_rad: 0.0
  rear_wheel_angle_rad: 0.0
- time_from_start:
    sec: 0
    nanosec: 0
  pose:
    position:
      x: 3704.4930889210755
      y: 73740.55103609574
      z: 0.0
    orientation:
      x: 0.0
      y: 0.0
      z: 0.8882163516239227
      w: 0.45942541582708296
  longitudinal_velocity_mps: -1.3888888359069824
  lateral_velocity_mps: 0.0
  acceleration_mps2: 0.0
  heading_rate_rps: 0.0
  front_wheel_angle_rad: 0.0
  rear_wheel_angle_rad: 0.0
- time_from_start:
    sec: 0
    nanosec: 0
  pose:
    position:
      x: 3704.5292234154686
      y: 73740.50098668145
      z: 0.0
    orientation:
      x: 0.0
      y: 0.0
      z: 0.8910379629250171
      w: 0.45392879246246953
  longitudinal_velocity_mps: -1.3888888359069824
  lateral_velocity_mps: 0.0
  acceleration_mps2: 0.0
  heading_rate_rps: 0.0
  front_wheel_angle_rad: 0.0
[...]

Issue #4490, opened 2023-08-01
Autoware sometimes stuck in the middle of a parking scenario


TakaHoribe:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I'm convinced that this is not my fault but a bug.

### Description

Even though the Planning module is still in the middle of the sequence, the goal decision is made with a large yaw angle error, and the state of Autoware is transited to STOP.

The essential cause is that the goal decision is made without considering the status of the Planning and Control.

In the following picture, the planning is still making the trajectory to fix the yaw angle, but the goal decision is made and the Autoware will not move.

Example 1
![image](https://github.com/autowarefoundation/autoware.universe/assets/21360593/a03b9b82-d819-45ee-a75e-2a6738f520ce)

Example 2
![image](https://github.com/autowarefoundation/autoware.universe/assets/21360593/9a06fa62-e914-40ee-86e9-4e9768b8ea04)

### Expected behavior

The goal decision can appropriately consider the status of the planning process.

### Actual behavior

Goal decision is made in the middle of the planning process.

### Steps to reproduce

1. run Psim
2. test the freespace planner
3. the issue sometimes happen

### Versions

autoware.universe version: 2212b003ad5efb7f0baba221c976d0c8801ad236

TakaHoribe:
cc. @yukkysaito @isamu-takagi

isamu-takagi:
@TakaHoribe The mission_planner receives the goal pose and does arrival check with [its own parameters](https://github.com/autowarefoundation/autoware.universe/blob/main/planning/mission_planner/config/mission_planner.param.yaml#L4-L6). Since this parameter does not reflect the control parameter, I think the arrival check should be moved to a path planning or control module that knows the control parameter.

ralwing:
From my observations, the main reason for such divergence is the trajectories generated by freespace_planner. These trajectories after smoothing may be unfeasible for control: mainly trajectory points orientation shows may accidentally flip:
freespace_trajectory:

<img width="681" height="994" alt="Image" src="https://github.com/user-attachments/assets/18ad6240-802c-4c0b-b4f0-a1b53927a37e" />
smoother trajectory:


Smoother orientation flips:
[...]

PR #11789, opened 2025-12-17
fix(astar_search): non-monotonic trajectories fix. Ensure that points between nodes are added in the correct order

ralwing:
## Description

This fixes issue `non-monotonic trajectory generated by freespace planner`  #11788 

When the planner reconstructs the path from Goal to Start, it interpolates points between A* nodes.
The code was generating these interpolated points from "Parent to Node" (Start-side to Goal-side)
but pushing them into the list in a way that, after the final reversal, resulted in local reversals
within each segment.
Fix Applied:
I reversed the interpolation loop in astar_search.cpp. This ensures that points between nodes are
added in the correct order (Node-side to Parent-side during the backward collection), so the final
path is perfectly monotonic (Parent -> Interpolated -> Node).

## Related links
Parent issue:
https://github.com/autowarefoundation/autoware_universe/issues/11788

#4490: `Autoware sometimes stuck in the middle of a parking scenari` - Most of the situations like this are fixed.

#5938: `motion velocity smoother generate invalid trajectory from freespace planner output:` 
Motion velocity smoother doesn't get invalid backward trajectories, so even though the motion valocity planner still should be fixed, this fix removes the main cause of it.

This is also one of the PRs required to allow backwards driving:
https://github.com/orgs/autowarefoundation/discussions/6559

**Parent Issue:**
https://github.com/autowarefoundation/autoware_universe/issues/11788

## How was this PR tested?
Added regression unit tests that cover the issue. Also tested it in carla and on the real vehicle. 

TIER IV Internal Evaluator :100: 
- [[kosuke55][Lexus][MasterScenario] PR check](https://evaluation.tier4.jp/evaluation/reports/ff6e0610-828c-5466-9605-2f99c87eaa7e?project_id=autoware_dev)
- [[kosuke55][Lexus][DLR_Basic] PR check](https://evaluation.tier4.jp/evaluation/reports/e378d934-588c-501b-a1e1-41dd86c48bbe?project_id=prd_jt)
- [[kosuke55][Lexus][FuncVerif_Basic] PR check](https://evaluation.tier4.jp/evaluation/reports/25d44888-c218-5afa-9568-cc37bdb92f54?project_id=prd_jt)

image

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/autoware_freespace_planning_algorithms/src/astar_search.cpp
+++ b/planning/autoware_freespace_planning_algorithms/src/astar_search.cpp
@@ -444,7 +444,7 @@ void AstarSearch::setPath(const AstarNode & goal_node)
     const auto parent_pose = node2pose(*node.parent);
     const double distance_2d = calc_distance2d(node2pose(node), parent_pose);
     const int n = static_cast<int>(distance_2d / min_expansion_dist_);
-    for (int i = 1; i < n; ++i) {
+    for (int i = n - 1; i >= 1; --i) {
       const double dist =
         ((distance_2d * i) / n) * (node.is_back == is_backward_search_ ? 1.0 : -1.0);
       const double steering = node.steering_index * steering_resolution_;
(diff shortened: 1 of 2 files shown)


satoshi-ota:
@kosuke55  Could you please review this PR as a package maintainer?

kosuke55:
@ralwing 
I'm so soory for the delay. I just approved. thank you for your contribution.

PR #11795, opened 2025-12-19
chore(yabloc_image_processing, autoware_pose_estimator_arbiter): resolve cmake targets

paulsohn:
## Description

Add explicit `find_package` for build failing packages on my local: 
* yabloc_image_processing: Sophus
* autoware_pose_estimator_arbiter: magic_enum

## Related links

None.

## How was this PR tested?

`colcon build --packages-up-to yabloc_image_processing` and `colcon build --packages-up-to autoware_pose_estimator_arbiter` passes.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/localization/autoware_pose_estimator_arbiter/CMakeLists.txt
+++ b/localization/autoware_pose_estimator_arbiter/CMakeLists.txt
@@ -4,6 +4,8 @@ project(autoware_pose_estimator_arbiter)
 find_package(autoware_cmake REQUIRED)
 autoware_package()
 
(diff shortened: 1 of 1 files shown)


KYabuuchi (approved):
I don' know why you failed to build these packages. But some other packages have the same commands in thier CMakeLists.txt. So I think this PR is OK. 

I have confirmed that the build works in my environment regardless of whether this PR is present or not.

PR #11816, opened 2025-12-26
fix(imu_scale_bias_estimator): ekf initializes only when it has in bounds samples

SergioReyesSan:
## Description

This PR adds an initial check to the initialization process. Observations from the Mobility Observer indicated that the scale could exceed the limits during the initialization phase (immediately following Autoware initialization). It was found that scale estimation may fail if localization is unstable during this period. This issue was successfully reproduced in the simulator by perturbing the ego vehicle's pose.

This fix was tested both in the simulator and on the physical bus. The PR does not modify the core algorithm; instead, it adds a check to verify that the estimated initial scale falls within the allowed range. If the value is out of bounds, reinitialization is attempted. If all reinitialization attempts fail, an error is triggered via the diagnostics.

Once this PR is accepted, i will use the commit of this one to submit it to the tier4 universe branch, including the aip launcher modifications.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?
This fix was tested on the simulator and the bus directly

## Notes for reviewers
Also is necessary to modify the launcher config files to add the new parameters
None.

## Interface changes

None.

## Effects on system behavior

None.


YoshiRi:
@SergioReyesSan IMO It's better to add JIRA/confluence link so that the reviewer can understand what you have confirmed.

sensing/autoware_imu_corrector/src/gyro_bias_estimator.cpp
```diff
@@ -133,18 +133,26 @@ GyroBiasEstimator::GyroBiasEstimator(const rclcpp::NodeOptions & options)
...
+  ekf_angle_.x_state_(1) = ekf_angle_.initial_scale_;  // estimated scale
+  ekf_angle_.estimated_scale_angle_ = ekf_angle_.initial_scale_;
   ekf_angle_.max_variance_p_angle_ = declare_parameter<double>("ekf_angle.variance_p_angle");
-  double noise_ekf_r_angle = declare_parameter<double>("ekf_angle.measurement_noise_r_angle");
+  ekf_angle_.noise_ekf_r_angle_ = declare_parameter<double>("ekf_angle.measurement_noise_r_angle");
   ekf_angle_.p_angle_ << ekf_angle_.max_variance_p_angle_, 0, 0, ekf_angle_.max_variance_p_angle_;
-  ekf_angle_.q_angle_ << 0, 0, 0, declare_parameter<double>("ekf_angle.process_noise_q_angle");
-  ekf_angle_.r_angle_ << noise_ekf_r_angle * noise_ekf_r_angle;
-  ekf_angle_.filtered_scale_angle_ = initial_scale;
+  ekf_angle_.r_angle_ << ekf_angle_.noise_ekf_r_angle_ * ekf_angle_.noise_ekf_r_angle_;
```
TaikiYamada4, line 144: Why the initialization of `ekf_angle_.q_angle_` is deleted here?

sensing/autoware_imu_corrector/README.md
```diff
@@ -65,7 +65,7 @@ Currently, it is possible to use methods other than NDT as a `pose_source` for A
 
 In the future, with careful implementation for pose errors, the IMU bias estimated by NDT could potentially be used not only for validation but also for online calibration.
 
-The Extended Kalman Filter (EKF) is used for scale estimation. The NDT pose is used as ground truth, and we assume it's accurate enough to provide long-term convergence for the correct scale observation.
+The Extended Kalman Filter (EKF) is used for scale estimation. The NDT pose is used as ground truth, and we assume it's accurate enough to provide long-term convergence for the correct scale observation, the EFK filter includes an initial checking and it requires N amount of consecutive in bounds samples to initialize `samples_in_bounds_to_init` , otherwise retries up to the value set on `max_reinitialization_retries` before triggering an error.
```
TaikiYamada4:
```suggestion
The Extended Kalman Filter (EKF) is used for scale estimation. The NDT pose is used as ground truth, and we assume it's accurate enough to provide long-term convergence for the correct scale observation, the EKF filter includes an initial checking and it requires N amount of consecutive in bounds samples to initialize `samples_in_bounds_to_init` , otherwise retries up to the value set on `max_reinitialization_retries` before triggering an error.
```

SergioReyesSan: @TaikiYamada4 Thank you for noticing this.

Adressed here

d230c8eae3478727a071f1cab11c7185def35cf7

sensing/autoware_imu_corrector/src/gyro_bias_estimator.cpp
SergioReyesSan: @TaikiYamada4 Sorry for my late response, and thank you for noticing that, appreciate it, already fixed and verified that it works.

Adressed here

d230c8eae3478727a071f1cab11c7185def35cf7

PR #11827, opened 2026-01-05
refactor(tier4_planning_factor_rviz_plugin): use service call to retrieve vehicle info

yhisaki:
## Description

This PR refactors the `tier4_planning_factor_rviz_plugin` to retrieve vehicle info via service call instead of using `autoware_vehicle_info_utils`.
By this PR, it is now possible to open only RViz.

## Related links

- https://github.com/autowarefoundation/autoware_universe/pull/9999#issuecomment-3240544746

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/visualization/tier4_planning_factor_rviz_plugin/src/planning_factor_rviz_plugin.cpp
+++ b/visualization/tier4_planning_factor_rviz_plugin/src/planning_factor_rviz_plugin.cpp
@@ -178,7 +178,7 @@ void PlanningFactorRvizPlugin::start_vehicle_info_request()
 
       // Wait for service to be ready
       while (rclcpp::ok() && !client->wait_for_service(std::chrono::seconds(1))) {
-        RCLCPP_INFO(
+        RCLCPP_WARN_ONCE(
           rclcpp::get_logger("PlanningFactorRvizPlugin"), "Waiting for vehicle_info service...");
       }
 


Owen-Liuyuxuan:
@yhisaki Please correct me if I am wrong. By the current design, if there is no parameter service, the system will remain using `baselink2front = 0.0` for the computation in the plugin.

Owen-Liuyuxuan:
@yhisaki Gemini suggests that logging every second of `Waiting for vehicle_info service...` could be troubling. It suggests using RCLCPP_WARN_ONCE or other commands to control the frequency of logging outputs

yhisaki:
> @yhisaki Gemini suggests that logging every second of `Waiting for vehicle_info service...` could be troubling. It suggests using RCLCPP_WARN_ONCE or other commands to control the frequency of logging outputs

@Owen-Liuyuxuan 
Thanks! I modified it at d9d1131659e46517e705ef74b84a2d213bd44c29

yhisaki:
> @yhisaki Please correct me if I am wrong. By the current design, if there is no parameter service, the system will remain using `baselink2front = 0.0` for the computation in the plugin.

@Owen-Liuyuxuan 
Yes! If there is no service, this plugin uses baselink2front = 0.0. However, there should generally be a service when playing rosbag with logsim or psim. So, I guess it may not be a problem.

PR #11846, opened 2026-01-09
feat(intersection): add custom pass judge line

Kotakku:
## Description
In this PR, the intersection module will use the reference position for the pass judge line if it is specified in the map data.

default (green line is pass judge line): 
image

a stop line is embedded in the map:
image

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_velocity_planner/autoware_behavior_velocity_intersection_module/src/scene_intersection_prepare_data.cpp
+++ b/planning/behavior_velocity_planner/autoware_behavior_velocity_intersection_module/src/scene_intersection_prepare_data.cpp
@@ -187,10 +187,12 @@ std::optional<size_t> IntersectionModule::getStopLineIndexFromMap(
     assigned_lanelet.regulatoryElementsAs<lanelet::autoware::RoadMarking>();
   lanelet::ConstLineStrings3d stopline;
   for (const auto & road_marking : road_markings) {
-    const std::string type =
-      road_marking->roadMarking().attributeOr(lanelet::AttributeName::Type, "none");
-    if (type == lanelet::AttributeValueString::StopLine) {
-      stopline.push_back(road_marking->roadMarking());
+    const auto & line = road_marking->roadMarking();
+    const std::string type = line.attributeOr(lanelet::AttributeName::Type, "none");
+    const std::string stopline_type =
+      line.attributeOr("intersection_pass_judge_line_type_v1", "default");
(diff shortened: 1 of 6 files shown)


planning/behavior_velocity_planner/autoware_behavior_velocity_intersection_module/src/scene_intersection_prepare_data.cpp
```diff
@@ -187,10 +187,12 @@ std::optional<size_t> IntersectionModule::getStopLineIndexFromMap(
...
-    const std::string type =
-      road_marking->roadMarking().attributeOr(lanelet::AttributeName::Type, "none");
-    if (type == lanelet::AttributeValueString::StopLine) {
-      stopline.push_back(road_marking->roadMarking());
+    const auto & line = road_marking->roadMarking();
+    const std::string type = line.attributeOr(lanelet::AttributeName::Type, "none");
+    const std::string stopline_type =
+      line.attributeOr("intersection_pass_judge_line_type_v1", "default");
+    if (type == lanelet::AttributeValueString::StopLine && stopline_type != "manual") {
+      stopline.push_back(line);
```
mitsudome-r, line 195: I have two concerns about this modification:
### 1. Backward compatibility
~~This is going to be a breaking changes to map format and would require the original users of the function to modify their map. (It used to work with roadmarking with just lanelet::AttributeValueString::StopLine, but now requires intersection_pass_judge_line_type_v1 to be "manual").~~

~~Is it possible for you to briefly explain why you would like to add this new attribute?
I would like to consider if we could have a way to keep backward compatibility as well.~~

Never mind with the first concern. I misread the condition as `stopline_type == "manual"` instead of `stopline_type != "manual"`. The code shouldn't affect most of the users. 

### 2. Naming of the attribute
Do you have any reasons behind the naming of the attribute name `intersection_pass_judge_line_type_v1`? We already have the [map format version](https://github.com/autowarefoundation/autoware_lanelet2_extension/blob/15a50aff0771ae64262be4ea037f718b9ccf71d2/autoware_lanelet2_extension/include/autoware_lanelet2_extension/version.hpp#L30), and I don't feel comfortable defining versions per attribute. Can't it be something like stopline_type instead?

mitsudome-r (changes requested):

yhisaki (approved):

PR #11847, opened 2026-01-09
fix(autoware_agnocast_wrapper): fix message_ptr to respect unique semantics

bdm-k:
## Description
Previously, `AUTOWARE_MESSAGE_UNIQUE_PTR` could be copied, even though it is a *unique* pointer. This PR fixes the internal implementation and prohibits copying of `AUTOWARE_MESSAGE_UNIQUE_PTR`.

### Implementation details
First, the pointer type that `message_ptr` uses to point to a `message_interface` object is changed from `std::shared_ptr` to `std::unique_ptr`. In the initial implementation, `std::shared_ptr` was used to give `message_ptr` copy semantics. However, this enabled copying not only for the *shared* `message_ptr` but also for the *unique* `message_ptr`, which is not desirable. Using `std::unique_ptr` also has the advantage of eliminating the overhead associated with control block management.

The copy ctor/assign are deleted for the *unique* `message_ptr`. On the other hand, custom copy ctor/assign are defined for the *shared* `message_ptr`. As described above, the pointer type has been changed to `std::unique_ptr`, so the default copy ctor/assign do not work in this case. The copy ctor/assign are defined such that the inner `message_interface` object—which is also a smart pointer btw—is cloned and used as the copy.

## Related links

None.

## How was this PR tested?

I have verified that Autoware compiles successfully with the `ENABLE_AGNOCAST` environment variable set to 1. Moreover, I have confirmed that attempting to copy `AUTOWARE_MESSAGE_UNIQUE_PTR` results in a compile error.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


bdm-k:
autoware_agnocast_wrapper has been migrated to autoware_core.

PR #11851, opened 2026-01-09
feat(pointcloud_preprocessor): simplify is_valid

xmfcx:
## Description

This code block came from:
https://github.com/tier4/AutowareArchitectureProposal.iv/pull/114/files#diff-07b20aa2d5779278a53c437245ddaaa091aa84d10486f1095e567153aad8070cR216-R240
as it is and was preserved over time.

Probably the initial intent was to add these checks later on.

But right now they just add false sense of security by returning `true`.

So this PR removes these return-true functions and remove unused undefined inputs.

## How was this PR tested?

CI.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/filter.hpp
+++ b/sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/filter.hpp
@@ -243,8 +243,7 @@ protected:
 
   std::unique_ptr<managed_transform_buffer::ManagedTransformBuffer> managed_tf_buffer_{nullptr};
 
-  inline bool is_valid(
-    const PointCloud2ConstPtr & cloud, const std::string & /*topic_name*/ = "input")
+  inline bool is_valid(const PointCloud2ConstPtr & cloud)
   {
     if (cloud->width * cloud->height * cloud->point_step != cloud->data.size()) {
       RCLCPP_WARN(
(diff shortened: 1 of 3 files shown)


sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/filter.hpp
```diff
@@ -258,18 +257,6 @@ class Filter : public rclcpp::Node
     return true;
   }
 
-  static inline bool is_valid(
```
mitsudome-r, line 261: Should we delete the function?

Maybe we could have some simple implementation like this. 

```
static inline bool is_valid(
    const PointIndicesConstPtr & indices, size_t num_points)
  {
    if (!indices) return false;
    for (const auto& idx : indices->indices) {
      if (idx >= num_points) return false;
    }
    return true;
  }
```

sensing/autoware_pointcloud_preprocessor/src/filter.cpp
```diff
@@ -232,11 +232,6 @@ void autoware::pointcloud_preprocessor::Filter::input_indices_callback(
     RCLCPP_ERROR(this->get_logger(), "[input_indices_callback] Invalid input!");
     return;
   }
-  // If indices are given, check if they are valid
-  if (indices && !is_valid(indices)) {
-    RCLCPP_ERROR(this->get_logger(), "[input_indices_callback] Invalid indices!");
-    return;
-  }
```
xmfcx, line 239: Removing because `!is_valid(indices)` is always false and this code can never execute.

sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/filter.hpp
xmfcx: We could do something like that, do you want me to implement it? (with std::algorithm)

sensing/autoware_pointcloud_preprocessor/src/filter.cpp
mitsudome-r: I'm also okay with removing as long as we create an issue to make sure that we remind ourselves that we have to implement the function.

xmfcx:
I'll merge this. Then will create another PR that will implement a simple indices sanity check.

PR #11853, opened 2026-01-09
feat(pointcloud_preprocessor): improve cloud validation

xmfcx:
## Description

- Related: https://github.com/autowarefoundation/autoware_universe/pull/11852

This PR strengthens PointCloud2 input validation across pointcloud preprocessing and compare-map segmentation components. It introduces more robust consistency checks, improves logging clarity, and adds reusable utilities to verify PointXYZ memory layout compatibility.

### Key Changes

* **Enhanced `is_valid()` validation**

  * Now checks for:

    * Null PointCloud2 pointers
    * Zero `point_step`
    * `row_step` consistency (`width × point_step`)
    * Data buffer size consistency (`height × row_step`)
    * Compatibility with a `PointXYZ` memory layout
  * Validation failures emit **clear, structured warning logs** with frame ID and timestamp.

* **Added PointXYZ layout compatibility utilities**

  * New helper functions:

    * `is_data_layout_compatible_with_point_xyz(const std::vector<PointField>&)`
    * `is_data_layout_compatible_with_point_xyz(const PointCloud2&)`
  * Ensures the first three fields are tightly packed `x`, `y`, `z` as `FLOAT32`.
  * Allows additional fields after XYZ.

* **API update for safer logging**

  * `is_valid()` now takes an explicit `rclcpp::Logger` argument.
  * Removes hidden logger dependencies and improves reuse.

* **Updated call sites**

  * All `is_valid()` callers updated to pass the node logger explicitly.
  * Affects:

    * `autoware_pointcloud_preprocessor::Filter`
    * `VoxelBasedCompareMapFilterComponent`

### Motivation

Downstream pointcloud processing often assumes:

* Valid buffer sizes
* Correct stride values
* A standard `PointXYZ` prefix layout

Invalid or malformed PointCloud2 messages can otherwise lead to:

* Undefined behavior
* Hard-to-trace crashes
* Silent data corruption

This PR makes such failures **explicit, early, and diagnosable**.

### Impact

* Improves robustness and debuggability of pointcloud pipelines
* Prevents unsafe assumptions about PointCloud2 memory layout
* No functional changes for valid, well-formed PointCloud2 messages

### Notes

* Additional point fields (e.g. intensity, ring, time) remain supported
* Behavior is backward-compatible for existing valid inputs

## How was this PR tested?

- CI
- https://github.com/autowarefoundation/autoware_universe/pull/11853#issuecomment-3731889518

### DIFF
--- a/sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/filter.hpp
+++ b/sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/filter.hpp
@@ -245,27 +245,105 @@ protected:
 
   std::unique_ptr<managed_transform_buffer::ManagedTransformBuffer> managed_tf_buffer_{nullptr};
 
-  static bool is_valid(const PointCloud2ConstPtr & cloud, const rclcpp::Logger & logger)
+  /**
+   * @brief Validate a sensor_msgs::msg::PointCloud2 message for structural consistency and layout.
+   *
+   * This function performs a series of lightweight sanity checks to determine whether a
+   * PointCloud2 message is safe to consume by algorithms expecting an XYZ-compatible,
+   * organized (dense) point cloud.
+   *
+   * Validation errors and warnings are reported using throttled ROS logging to avoid
+   * excessive log spam when invalid data is repeatedly received.
+   *
+   * @details
+   * The following checks are performed in order:
+   *
+   * 1. **Null pointer check**
+   *    - Ensures the input PointCloud2 pointer is valid.
+   *
+   * 2. **Point step validation**
+   *    - Verifies that `point_step` is non-zero.
+   *
+   * 3. **Dense (organized) cloud requirement**
+   *    - Warns if `is_dense` is false.
+   *    - Currently does *not* fail validation, but will become a hard error
+   *      starting **July 2026**.
+   *
+   * 4. **Point field layout compatibility**
+   *    - Ensures the point data layout is compatible with `PointXYZ`
+   *      (fields must begin with `x`, `y`, `z` of type `FLOAT32`).
+   *
+   * 5. **Row step consistency**
+   *    - Validates that `row_step == width * point_step`.
+   *    - Currently logs a warning on mismatch.
+   *    - Will become a hard error starting **July 2026**.
+   *
+   * 6. **Data buffer size consistency**
+   *    - Ensures `data.size() == height * row_step`.
+   *
+   * @param cloud
+   *   Shared pointer to the input PointCloud2 message to validate.
+   *
+   * @param logger
+   *   ROS 2 logger used for emitting validation warnings.
+   *
+   * @param clock
+   *   ROS 2 clock used for throttled logging.
+   *
+   * @return true
+   *   If the point cloud passes all mandatory validation checks.
+   *
+   * @return false
+   *   If a critical validation failure is detected (e.g., null pointer,
+   *   incompatible data layout, or inconsistent buffer size).
+   *
+   * @note
+   * - All warnings are throttled with a fixed duration of 5000 ms.
+   * - Some warnings are marked as *future errors* and are expected to
+   *   become fatal after July 2026.
+   *
+   * @warning
+   * This function assumes consumers expect an XYZ-only point layout.
+   * Additional fields (e.g., intensity, color) are allowed only if they
+   * appear *after* the XYZ fields and do not alter the layout compatibility.
+   */
+  static bool is_valid(
+    const PointCloud2ConstPtr & cloud, const rclcpp::Logger & logger, rclcpp::Clock & clock)
   {
+    static constexpr rcutils_duration_value_t throttle_duration_ms = 5000;
+
     if (!cloud) {
-      RCLCPP_WARN(logger, "Invalid PointCloud: Null pointer received.");
+      RCLCPP_WARN_THROTTLE(
+        logger, clock, throttle_duration_ms, "Invalid PointCloud: Null pointer received.");
       return false;
     }
 
     // Check: Point Step
     if (cloud->point_step == 0) {
-      RCLCPP_WARN(
-        logger,
+      RCLCPP_WARN_THROTTLE(
+        logger, clock, throttle_duration_ms,
         "Invalid PointCloud: point_step is 0. "
         "Frame: '%s', Stamp: %d.%09u",
         cloud->header.frame_id.c_str(), cloud->header.stamp.sec, cloud->header.stamp.nanosec);
       return false;
     }
 
+    // Check: Only accept organized (dense) point clouds
+    if (!cloud->is_dense) {
+      RCLCPP_WARN_THROTTLE(
+        logger, clock, throttle_duration_ms,
+        "Invalid PointCloud: is_dense is false. "
+        "The point cloud should be organized (dense) and contain only valid points. "
+        "This will be an ERROR starting in 2026 July"
+        "Frame: '%s', Stamp: %d.%09u",
+        cloud->header.frame_id.c_str(), cloud->header.stamp.sec, cloud->header.stamp.nanosec);
+      // TODO(mfc): return false; After 2026 July
+    }
+
     // Check: Point field layout compatibility
     if (!utils::is_data_layout_compatible_with_point_xyz(*cloud)) {
-      RCLCPP_WARN(
-        logger,
+      RCLCPP_WARN_THROTTLE(
+        logger, clock, throttle_duration_ms,
         "The pointcloud layout is not compatible with PointXYZ. Aborting. "
         "Make sure Point fields start with x, y, z as FLOAT32.");
       return false;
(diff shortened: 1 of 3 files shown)


mitsudome-r:
I tested with [the sample rosbag](https://autowarefoundation.github.io/autoware-documentation/main/demos/rosbag-replay-simulation/#how-to-run-a-rosbag-replay-simulation), and it seems like some pipeline fails to publish pointcloud with valid row_step info. We might want to fix the issue before merging so that users could at least run demos without an issue.

xmfcx:
Thanks for testing!

> it seems like some pipeline fails to publish pointcloud with valid row_step info.

I totally knew this would happen... Most people just fill in width and height and call it a day. (if they are not using the pcl <-> ros converter)

xmfcx:
I've rebased it now.

I am even considering checking the old `cloud->width * cloud->height * cloud->point_step != cloud->data.size()` with 64-bit arithmetic to be on the safe side for the first step.

And only warn if `row_step` is malformed. What do you think?

> We might want to fix the issue before merging so that users could at least run demos without an issue.

I agree.

mitsudome-r:
> I am even considering checking the old cloud->width * cloud->height * cloud->point_step != cloud->data.size() with 64-bit arithmetic to be on the safe side for the first step.

This could be an option.

mitsudome-r:
At least the error that I saw was due to concatenate node, but there could be some other nodes that fails to calculate row_steps in the downstream pipeline.
We can set it as warning for now and fix those warnings in other PRs.

xmfcx:
Turned it into a warning only in https://github.com/autowarefoundation/autoware_universe/pull/11853/commits/f66794a96aa79778df570932c0458a2c609bcfae

And added a deprecation warning. Do you think 2026 July is a good deadline?

Also checked that rest of the code (added in this PR) doesn't need to be modified since `expected_row_step` is reused so https://github.com/autowarefoundation/autoware_universe/pull/11853/commits/f66794a96aa79778df570932c0458a2c609bcfae should suffice, rest of the code acts like `cloud->width * cloud->height * cloud->point_step != cloud->data.size()` validation.

xmfcx:
https://github.com/ros2/common_interfaces/blob/jazzy/sensor_msgs/msg/PointCloud2.msg

Basically the rule is:
- If cloud is unorganized (expected by Autoware code):
   - `width = points_count`
   - `height = 1`
   - `row_step = points_count * point_step`
   - `data.size() = row_step`
   - `is_dense = true` (no NaN points)
- If cloud is organized (similar to a depth image):
   - `width = horizontal_points_count`
   - `height = vertical_points_count`
   - `row_step = width * point_step`
   - `data.size() = height * row_step`
   - `is_dense = false` (can have NaN points, not space efficient)

We might want to reject **non-dense** point clouds by default in autoware too.

I didn't investigate how Autoware handles NaN or Inf in point cloud fields throughout the codebase.

Reference: https://robotics.stackexchange.com/a/73982

mitsudome-r:
> And added a deprecation warning. Do you think 2026 July is a good deadline?

Yes, I think so.

> We might want to reject non-dense point clouds by default in autoware too.

Maybe we could add a warning with non-dense point clouds and make it an error with the same timing with row_steps.

mitsudome-r:
image
At least rosbag simulation demo publishes the perception results now.

However, warning got quite annoying
image

xmfcx:
I will throttle it

mitsudome-r:
It looks like concatenation was the only issue (at least with rosbag demo).
https://github.com/autowarefoundation/autoware_universe/pull/11855 will remove the warning messages, but we could still throttle the message to avoid contamination of the logs if there are any other nodes with invalid row_steps.

xmfcx:
Throttled and added the density enforcement with https://github.com/autowarefoundation/autoware_universe/pull/11853/commits/f56915542e9bd83e9f0d5aa9daa7b7114863d613

image

Tested with AWSIM-Lightweight 2.0.1 and I only get:

```
[WARN 1768033994.311136343] [localization.util.crop_box_filter_measurement_range]: Invalid PointCloud: is_dense is false. The point cloud should be organized (dense) and contain only valid points. This will be an ERROR starting in 2026 JulyFrame: 'base_link', Stamp: 216.089846874 (is_valid() at /home/mfc/projects/autoware/src/universe/autoware_universe/sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/filter.hpp:333)
```

One warning per 5s as expected.

We should also fix `crop_box_filter_measurement_range` for density.
This one didn't complain for concatenation. (I didn't test your PR)

The actual annoying warning is **(Not related to this PR)**:

```
[WARN 1768033994.125431232] [sensing.lidar.concatenate_data]: transformed_raw_points[/sensing/lidar/left/pointcloud_before_sync] is nullptr, skipping pointcloud publish. (publish_clouds() at /home/mfc/projects/autoware/src/universe/autoware_universe/sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/concatenate_data/concatenate_and_time_sync_node.ipp:327)
```

This is tripping newcomers https://github.com/orgs/autowarefoundation/discussions/6694 by confusing them with this benign warning. (Autoware works just fine only with top lidar even when the setup is configured with 3 lidars)

I will ask AWSIM team to publish all 3 top lidars in the demo for the lightweight one. So the launch files are in line with the demo.

- Addressed in https://github.com/autowarefoundation/autoware_launch/pull/1733

xmfcx:
- I created https://github.com/autowarefoundation/autoware_universe/pull/11856 to fix cropbox is_dense issue.

Let's merge this PR first so we will have basis for the fix PRs (for ease of testing)

xmfcx:
@drwnz @mojomex @YoshiRi -san, We have tested in rosbag replay sim and AWSIM so we will go ahead and merge it.

Please take a look at this change (even after merging) to be up to date.

If an issue arises we can revert it.

sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/filter.hpp
```diff
@@ -244,17 +245,141 @@ class Filter : public rclcpp::Node
...
+    size_t expected_data_size = static_cast<size_t>(cloud->height) * expected_row_step;
+
+    if (expected_data_size != cloud->data.size()) {
+      RCLCPP_WARN_THROTTLE(
+        logger, clock, throttle_duration_ms,
+        "Invalid PointCloud: data size mismatch. "
+        "Expected: %zu (height %u * row_step %u), Got: %zu. "
+        "Frame: '%s', Stamp: %d.%09u",
+        expected_data_size, cloud->height, cloud->row_step, cloud->data.size(),
+        cloud->header.frame_id.c_str(), cloud->header.stamp.sec, cloud->header.stamp.nanosec);
```
yukkysaito, line 379: @xmfcx This calculation depends on size_t, and depending on the environment,
width * point_step or height * row_step may overflow.

Instead of only comparing the multiplication results, it would be safer to add
overflow-safe checks before performing the multiplication.

For example, one option is to introduce a helper like safe_mul:
```
template <typename T>
std::optional<T> safe_mul(T a, T b) {
  static_assert(std::is_unsigned_v<T>, "use unsigned types");
  if (a == 0 || b == 0) return T{0};
  if (a > std::numeric_limits<T>::max() / b) return std::nullopt;
  return a * b;
}
```

Then, validating row_step == width * point_step and
data.size() == height * row_step step-by-step using safe_mul
would prevent overflow-based validation bypasses.

xmfcx: - @yukkysaito what about this: https://github.com/autowarefoundation/autoware_universe/pull/11861 ?

xmfcx: @yukkysaito first I planned to do it like this: https://gist.github.com/xmfcx/cad5841007a44bf4f023a88496f8e469#file-fix-hpp-L38-L52

```cpp
  size_t expected_size;
  if (__builtin_mul_overflow(
        static_cast<size_t>(cloud->width),
        static_cast<size_t>(cloud->height),
        &expected_size) ||
      __builtin_mul_overflow(
        expected_size,
        static_cast<size_t>(cloud->point_step),
        &expected_size)) {
    RCLCPP_WARN(
      this->get_logger(),
      "Invalid PointCloud (size computation overflow)");
    return false;
  }
```

but then size_t felt cleaner. now made a pr to cast to uint64 instead.

xmfcx: Reference: https://gcc.gnu.org/onlinedocs/gcc/Integer-Overflow-Builtins.html

yukkysaito: > @yukkysaito what about this: https://github.com/autowarefoundation/autoware_universe/pull/11861 ?
Looks Good :+1:

xmfcx: - Related: https://github.com/autowarefoundation/autoware_universe/issues/11858#issuecomment-3738796489

PR #11855, opened 2026-01-10
fix(autoware_pointcloud_preprocessor): recalculate row_step and width after concatenation

mitsudome-r:
## Description
This PR will fix the output of concatenate_pointclouds_node to have a valid width and row_step values after concatenating pointclouds.

## Related links

This is a follow up from https://github.com/autowarefoundation/autoware_universe/pull/11853.

## How was this PR tested?

I tested with rosbag simulation demo. 
You can play the rosbag and echo `/sensing/lidar/concatenated/pointcloud` and check width, point_step, and row_step values from a random message. 

Before:
 * width = 211049
 * point_step = 16
 * row_step = 3132272 (which is a wrong value and is supposed to be 211049 * 16 = 3376784)

image

After: 
 * width = 210241
 * point_step = 16
 * row_step = 3363856 (matches with `width * point_step`)
image

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


sensing/autoware_pointcloud_preprocessor/src/concatenate_data/combine_cloud_handler.cpp
```diff
@@ -231,6 +231,14 @@ CombineCloudHandler<PointCloud2Traits>::combine_pointclouds(
   }
   concatenate_cloud_result.concatenate_cloud_ptr->header.stamp = oldest_stamp;
 
+  // concatenated cloud is no longer structured so recalculate the height, width, and row_step
+  concatenate_cloud_result.concatenate_cloud_ptr->height = 1;
+  concatenate_cloud_result.concatenate_cloud_ptr->row_step =
+    concatenate_cloud_result.concatenate_cloud_ptr->data.size();
+  concatenate_cloud_result.concatenate_cloud_ptr->width =
+    concatenate_cloud_result.concatenate_cloud_ptr->data.size() /
+    concatenate_cloud_result.concatenate_cloud_ptr->point_step;
```
xmfcx:
```suggestion
  if (data_size % point_step != 0) {
      throw std::runtime_error("PointCloud2 data size is not divisible by point_step");
  }
  concatenate_cloud_result.concatenate_cloud_ptr->width =
    concatenate_cloud_result.concatenate_cloud_ptr->data.size() /
    concatenate_cloud_result.concatenate_cloud_ptr->point_step;
```

What about something like this to make sure the division is reliable?

mitsudome-r: Good catch!
I will fix the code.

mitsudome-r: Fixed in [`daaf4ef` (#11855)](https://github.com/autowarefoundation/autoware_universe/pull/11855/commits/daaf4ef3bd99ee2e83c1c2f3d9ef1d68f5569a92)

yukkysaito (approved):
LGTM

PR #11857, opened 2026-01-10
fix(autoware_pointcloud_preprocessor): inherit is_dense for concatenated pointcloud

mitsudome-r:
## Description
Related PR https://github.com/autowarefoundation/autoware_universe/pull/11853.

This PR fixes the warning about is_dense being false when running the [rosbag replay demo](https://autowarefoundation.github.io/autoware-documentation/main/demos/rosbag-replay-simulation/).

In original code, published topic from concatenate_pointclouds_node always have is_dense=false because [pcl::ConcatenatePointCloud](https://github.com/autowarefoundation/autoware_universe/blob/a1e9993a93bbda315566aa6ede82766f2be0245c/sensing/autoware_pointcloud_preprocessor/src/concatenate_data/combine_cloud_handler.cpp#L192) function always returns the field with false. After this fix, the node will publish topic with is_dense=true only when all the input pointclouds have is_dense=true. 

## Related links

Related PR https://github.com/autowarefoundation/autoware_universe/pull/11853

## How was this PR tested?
Tested with rosbag replay demo and made sure that the is_dense field in concatenated pointcloud is true.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


xmfcx:
@YoshiRi @drwnz @amc-nu @badai-nguyen @kyoichi-sugahara @kaancolak @mojomex @vividf @yukkysaito could you review please?

PR #11861, opened 2026-01-12
fix(point_cloud_preprocessor): use uint64_t to avoid size_t variance

xmfcx:
## Description

- Follow up from: https://github.com/autowarefoundation/autoware_universe/pull/11853#discussion_r2682461425

- Related: https://github.com/autowarefoundation/autoware_universe/issues/11858#issuecomment-3738796489

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/filter.hpp
+++ b/sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/filter.hpp
@@ -367,7 +367,8 @@ protected:
     }
 
     // Check: Data buffer size consistency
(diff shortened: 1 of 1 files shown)


xmfcx:
Ready for review!

yukkysaito:
Thanks for the update.
I think this PR is fine as it is.

Casting to uint64_t significantly improves robustness in practice, and given the current code structure and scope, this looks like a reasonable and pragmatic solution.

That said, the [safe_mul-style](https://github.com/autowarefoundation/autoware_universe/pull/11853#discussion_r2682461425) approach I mentioned earlier could also be useful as a more generic pattern, since it detects overflows without relying on explicit casts and may be easier to reuse in other places in the future.

In any case, I’m happy with the direction of this PR.

PR #11863, opened 2026-01-13
feat(behavior_path_side_shift): add drivable area check to prevent lane departure

k-uta:
## Description
This PR implements a boundary check mechanism in the `side_shift` module to prevent the ego vehicle from deviating from the drivable area (Lanelet) when a lateral offset is requested.

Previously, the module would execute the requested shift amount regardless of lane boundaries, which could lead to unsafe lane departures. With this update, the module calculates the maximum allowable lateral offset based on the distance to the lanelet boundaries and clamps the requested value to ensure the vehicle stays within the safe area.

## Changes
* **Added `calcMaxLateralOffset` function**: Calculates the safe shift limit based on the vehicle width, configured margin, and distance to lane boundaries.
* **Added Check Modes**:
    * **Mode 0**: No check (Disable).
    * **Mode 1**: Current lane only (Strict).
    * **Mode 2**: Current lane + Neighbor lanes (Allows changing lanes if direction matches).
* **Added Parameters**: Introduced `drivable_area_check_mode` and `min_drivable_area_margin`.
* **Logic Update**: Modified `SideShiftModule` to use the clamped `final_shift_length` instead of the raw requested offset.
* **Safety Improvements**: Added checks to abort shifting if the path length is insufficient or if generation fails.

### ROS Parameters
The following parameters have been added to `side_shift.param.yaml`:

| Change type | Parameter Name | Type | Default | Description |
| :--- | :--- | :--- | :--- | :--- |
| Added | `drivable_area_check_mode` | int | `1` | **0**: No check<br>**1**: Check current lane only<br>**2**: Check current and same-direction neighbor lanes |
| Added | `min_drivable_area_margin` | double | `0.3` | Minimum margin [m] to keep between the vehicle footprint and the lane boundary. |

## Effects on system behavior
* **Mode 0**: Behaves as before; shifts exactly by the requested amount regardless of boundaries.
* **Mode 1**: The vehicle will shift towards the target but stop *before* crossing the white line of the current lane.
* **Mode 2**: The vehicle can cross the white line into an adjacent lane (if it exists and has the same direction) but will stop before leaving the road boundary.

## How was this PR tested?
1.  Launch the planning simulator.
2.  Set the parameter `min_drivable_area_check_mode` to `1` or `2`.
3.  Publish a large lateral offset (e.g., -5.0m) that exceeds the lane width.

**Command:**
```bash
sec=$(($(date +%s))); ros2 topic pub /planning/scenario_planning/lane_driving/behavior_planning/behavior_path_planner/input/lateral_offset tier4_planning_msgs/msg/LateralOffset "{stamp: {sec: $sec, nanosec: 999999999}, lateral_offset: -5.0}" -1

### DIFF
--- a/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/include/autoware/behavior_path_lane_change_module/base_class.hpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/include/autoware/behavior_path_lane_change_module/base_class.hpp
@@ -33,6 +33,7 @@
 #include <geometry_msgs/msg/twist.hpp>
 
 #include <memory>
+#include <optional>
 #include <string>
 #include <utility>
 
@@ -94,7 +95,7 @@ public:
 
   virtual bool hasFinishedAbort() const = 0;
 
-  virtual bool isLaneChangeRequired() = 0;
+  virtual std::optional<std::string> isLaneChangeRequired() = 0;
 
   virtual bool isAbortState() const = 0;
 
@@ -149,6 +150,11 @@ public:
   const lanelet::ConstLanelets & get_current_lanes() const
   {
     return common_data_ptr_->lanes_ptr->current;
+  }
+
+  const lanelet::ConstLanelets & get_target_lanes() const
+  {
+    return common_data_ptr_->lanes_ptr->target;
   }
 
   const BehaviorPathPlannerParameters & getCommonParam() const { return planner_data_->parameters; }
--- a/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/include/autoware/behavior_path_lane_change_module/interface.hpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/include/autoware/behavior_path_lane_change_module/interface.hpp
@@ -138,7 +138,7 @@ protected:
 
   bool is_abort_approval_requested_{false};
 
-  lane_change::InterfaceDebug interface_debug_;
+  mutable lane_change::InterfaceDebug interface_debug_;
 };
 }  // namespace autoware::behavior_path_planner
 
--- a/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/include/autoware/behavior_path_lane_change_module/scene.hpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/include/autoware/behavior_path_lane_change_module/scene.hpp
@@ -109,7 +109,7 @@ public:
 
   bool isAbortState() const override;
 
-  bool isLaneChangeRequired() override;
+  std::optional<std::string> isLaneChangeRequired() final;
 
   bool isStoppedAtRedTrafficLight() const override;
 
--- a/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/include/autoware/behavior_path_lane_change_module/structs/debug.hpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/include/autoware/behavior_path_lane_change_module/structs/debug.hpp
@@ -108,6 +108,7 @@ struct InterfaceDebug
 {
   std::string_view failing_reason;
   LaneChangeStates lc_state;
+  std::string request_info;
 };
 }  // namespace autoware::behavior_path_planner::lane_change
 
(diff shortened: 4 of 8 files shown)

### CODE REVIEW

File: planning/behavior_path_planner/autoware_behavior_path_side_shift_module/src/scene.cpp [hunk 1]
```diff
@@ -214,8 +221,19 @@ void SideShiftModule::updateData()
...
+  // resamplePathWithSpline may not preserve velocities correctly, so we interpolate from original
+  for (size_t i = 0; i < reference_path_.points.size(); ++i) {
+    const auto & pos = reference_path_.points.at(i).point.pose.position;
+    const size_t nearest_idx =
+      autoware::motion_utils::findNearestIndex(centerline_path.points, pos);
+    if (nearest_idx < centerline_path.points.size()) {
+      reference_path_.points.at(i).point.longitudinal_velocity_mps =
+        centerline_path.points.at(nearest_idx).point.longitudinal_velocity_mps;
+    }
+  }
```
TaikiYamada4: Maybe not using indices make things a bit simple.
```suggestion
  for (auto & point : reference_path_.points) {
    const auto & pos = point.point.pose.position;
    const size_t nearest_idx =
      autoware::motion_utils::findNearestIndex(centerline_path.points, pos);
    if (nearest_idx < centerline_path.points.size()) {
      point.point.longitudinal_velocity_mps =
        centerline_path.points.at(nearest_idx).point.longitudinal_velocity_mps;
    }
  }
```

planning/behavior_path_planner/autoware_behavior_path_side_shift_module/config/side_shift.param.yaml
```diff
@@ -7,4 +7,6 @@
       min_shifting_distance: 5.0
       min_shifting_speed: 5.56
       shift_request_time_limit: 1.0
+      drivable_area_check_mode: 2 # 0: no check, 1: current lane, 2: current lane + neighbor lane
```
TaikiYamada4: Could you set 1 as default? Since this is the most tight and safe option.

planning/behavior_path_planner/autoware_behavior_path_side_shift_module/src/scene.cpp
```diff
@@ -214,8 +221,19 @@ void SideShiftModule::updateData()
     getPreviousModuleOutput().reference_path);
 
   constexpr double resample_interval = 1.0;
-  const auto backward_extened_path = extendBackwardLength(getPreviousModuleOutput().path);
```
TaikiYamada4, line 217: Why did you remove `extendBackwardLength` here?

TaikiYamada4 (commented):
Thank you for your contribution! I have some review comments below so please check them out. 

Besides, please also check the DCO is failing so please sign your commits as written in the DCO workflow instructions.

planning/behavior_path_planner/autoware_behavior_path_side_shift_module/config/side_shift.param.yaml
k-uta: I updated the default value to 1. Thank you.
[[2a1155b](https://github.com/autowarefoundation/autoware_universe/pull/11863/commits/2a1155b27a2cda2a0e7392be5aa1bac5ff2ca2cc)]

planning/behavior_path_planner/autoware_behavior_path_side_shift_module/src/scene.cpp
k-uta: I removed `extendBackwardLength` to fix an issue where the vehicle would suddenly decelerate during the shift.

The `extendBackwardLength` function concatenates the previous path with the current path. This may cause `resamplePathWithSpline` to generate unnatural curvature at the joint.

It is likely that this high curvature is interpreted by the velocity planner as a sharp turn, triggering unexpected braking. My intention was to avoid this stitching artifact by using `centerline_path` directly.

k-uta:
> @k-uta I think `drivable_area_margin` should be `min_drivable_area_margin`.

Agreed. Fixed it to min_drivable_area_margin in [0c76986](https://github.com/autowarefoundation/autoware_universe/pull/11863/commits/0c769866400a4081036c5cac92e142d0a7d21e01).

PR #11864, opened 2026-01-13
refactor(boundary_departure): reduce call to boundary_departure_checker_ptr and class renaming

zulfaqar-azmi-t4:
## Description

Refactoring `autoware_boundary_departure_checker` and `autoware_motion_velocity_boundary_departure_prevention` packages.

The goals of this refactoring are
- (In `boundary_departure_prevention`) unify `get_abnormalities_data` function to reduce call to `boundary_departure_checker_ptr_`. 
- (In `boundary_departure_checker`) move member functions and variable from `BoundaryDepartureChecker` class to `UncrossableBoundaryDepartureChecker` class to distinguish the different roles of the class.
- (In `boundary_departure_checker`) Separate build libraries in CMakeLists.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/common/autoware_boundary_departure_checker/include/autoware/boundary_departure_checker/boundary_departure_checker.hpp
+++ b/common/autoware_boundary_departure_checker/include/autoware/boundary_departure_checker/boundary_departure_checker.hpp
@@ -105,7 +105,6 @@ private:
   Param param_;
   lanelet::LaneletMapPtr lanelet_map_ptr_;
   std::shared_ptr<VehicleInfo> vehicle_info_ptr_;
-  std::unique_ptr<UncrossableBoundRTree> uncrossable_boundaries_rtree_ptr_;
 
   bool willLeaveLane(
     const lanelet::ConstLanelets & candidate_lanelets,


common/autoware_boundary_departure_checker/src/boundary_departure_checker_abnormality.cpp
maxime-clem, line 1: As far as I understand the `boundary_departure_checker_abnormality.cpp` is no longer used and the new `uncrossable_boundary_departure_checker.cpp` is just a copy of it.

Simply renaming the file( using `git mv boundary_departure_checker_abnormality.cpp uncrossable_boundary_departure_checker.cpp`) should greatly simplify the diff.

common/autoware_boundary_departure_checker/include/autoware/boundary_departure_checker/uncrossable_boundary_departure_checker.hpp
```diff
@@ -0,0 +1,185 @@
...
+   */
+  tl::expected<ClosestProjectionsToBound, std::string> get_closest_projections_to_boundaries(
+    const Abnormalities<ProjectionsToBound> & projections_to_bound, const double curr_vel,
+    const double curr_acc);
+
+private:
+  Param param_;
+  lanelet::LaneletMapPtr lanelet_map_ptr_;
+  std::shared_ptr<VehicleInfo> vehicle_info_ptr_;
+  std::unique_ptr<UncrossableBoundRTree> uncrossable_boundaries_rtree_ptr_;
```
mkquda, line 180: whats the difference between this rtree and the one in BoundaryDepartureChecker class?
https://github.com/autowarefoundation/autoware_universe/pull/11864/files#diff-0a708bfc69f530484d332d67bc996d0091f6723555056bb92783b75325ac7be6R108

zulfaqar-azmi-t4: The one in BoundaryDepartureChecker creates a bounding box around ego vehicle and extract uncrossable boundary at each iteration.

The one ins UncrossableBoundaryDeparture checker extract all uncrossable boundaries from vector map (just one off).

common/autoware_boundary_departure_checker/src/boundary_departure_checker_abnormality.cpp
zulfaqar-azmi-t4: Fixed. I had to force push because for unknown reason i had trouble with `git mv`

common/autoware_boundary_departure_checker/include/autoware/boundary_departure_checker/uncrossable_boundary_departure_checker.hpp
mkquda: I see, but I don't see why BoundaryDepartureChecker needs to care about uncrossable boundaries now that there is a dedicated class UncrossableBoundaryDepartureChecker.

I guess its because some modules are still using the original class, but I think its better to change usage side as well to avoid confusion

PR #11867, opened 2026-01-14
fix: resolve GPU memory leak issue in E2E module

YSH-research:
## Description

Fixed CUDA memory leak in VAD model inference by reusing `prev_bev` tensor memory instead of allocating new memory every frame.

**Problem:**
- `save_prev_bev()` allocated a new `Tensor` object every inference frame
- `Tensor` struct lacks a destructor, so CUDA memory (`cudaMalloc`) was never freed
- Memory leaked ~10MB per frame, causing ~300MB/sec leak at 30Hz

**Root Cause:**
- Cannot add destructor to `Tensor` because TensorRT shares tensor pointers between networks
- Adding `cudaFree()` in destructor would cause illegal memory access

**Solution:**
- Modified `save_prev_bev()` to reuse existing `saved_prev_bev_` memory when dimensions match
- Only allocate new memory on first frame or when BEV dimensions change
- Use `cudaMemcpyAsync` to update data in-place without reallocation

**Changed File:**
- `src/universe/autoware_universe/e2e/autoware_tensorrt_vad/src/vad_model.hpp`

## Notes for reviewers

Screenshot from 2026-01-14 14-17-10

- The fix is minimal and focused: only modified the `save_prev_bev()` function
- Memory reuse pattern checks dimension compatibility before reusing
- No functional changes to inference logic - only memory management optimization

## Interface changes
None

## Effects on system behavior
**Positive:**
- GPU memory usage stabilized - no more memory leak
- Prevents out-of-memory crashes during long-running operations

**Neutral:**
- No changes to inference output or trajectory generation
- Minimal performance impact (one additional dimension check per frame)

### DIFF
--- a/e2e/autoware_tensorrt_vad/src/vad_node.cpp
+++ b/e2e/autoware_tensorrt_vad/src/vad_node.cpp
@@ -302,7 +302,8 @@ bool VadNode::initialize_vad_model()
   try {
     // load configs
     VadConfig vad_config = load_vad_config();
-    auto [backbone_trt_config, head_trt_config, head_no_prev_trt_config] = load_trt_common_configs();
+    auto [backbone_trt_config, head_trt_config, head_no_prev_trt_config] =
+      load_trt_common_configs();
 
     // Initialize VAD interface and model
     auto tf_buffer_shared = std::shared_ptr<tf2_ros::Buffer>(&tf_buffer_, [](tf2_ros::Buffer *) {});
@@ -318,8 +319,7 @@ bool VadNode::initialize_vad_model()
       "VAD model and interface initialized successfully");
     return true;
   } catch (const std::exception & e) {
-    RCLCPP_ERROR(
-      this->get_logger(), "Error during VAD model initialization: %s", e.what());
+    RCLCPP_ERROR(this->get_logger(), "Error during VAD model initialization: %s", e.what());
     return false;
   }
 }


YSH-research:
> ## Description
> Fixed CUDA memory leak in VAD model inference by reusing `prev_bev` tensor memory instead of allocating new memory every frame.
> 
> **Problem:**
> 
> * `save_prev_bev()` allocated a new `Tensor` object every inference frame
> * `Tensor` struct lacks a destructor, so CUDA memory (`cudaMalloc`) was never freed
> * Memory leaked ~10MB per frame, causing ~300MB/sec leak at 30Hz
> 
> **Root Cause:**
> 
> * Cannot add destructor to `Tensor` because TensorRT shares tensor pointers between networks
> * Adding `cudaFree()` in destructor would cause illegal memory access
> 
> **Solution:**
> 
> * Modified `save_prev_bev()` to reuse existing `saved_prev_bev_` memory when dimensions match
> * Only allocate new memory on first frame or when BEV dimensions change
> * Use `cudaMemcpyAsync` to update data in-place without reallocation
> 
> **Changed File:**
> 
> * `src/universe/autoware_universe/e2e/autoware_tensorrt_vad/src/vad_model.hpp`
> 
> ## Notes for reviewers
> Screenshot from 2026-01-14 14-17-10
> * The fix is minimal and focused: only modified the `save_prev_bev()` function
> * Memory reuse pattern checks dimension compatibility before reusing
> * No functional changes to inference logic - only memory management optimization
> 
> ## Interface changes
> None
> 
> ## Effects on system behavior
> **Positive:**
> 
> * GPU memory usage stabilized - no more memory leak
> * Prevents out-of-memory crashes during long-running operations
> 
> **Neutral:**
> 
> * No changes to inference output or trajectory generation
> * Minimal performance impact (one additional dimension check per frame)

PR #11868, opened 2026-01-14
docs(goal_planner): use same terminology from bpp main docs

zulfaqar-azmi-t4:
## Description

Replace collision check explanation to have a standardized term.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


kosuke55 (approved):

kosuke55:
if you delete the word `RSS` from document,  I think it’s worth mentioning that RSS is also a valid option.

related to : 
https://github.com/autowarefoundation/autoware_universe/pull/11843#discussion_r2688864860

PR #11884, opened 2026-01-16
refactor: use boost version instead of ros version

felixf4xu:
## Description

This PR use boost version (instead of ROS version) to include boost file of `hull_graham_andrew.hpp`

## Related links

https://github.com/autowarefoundation/autoware_universe/pull/11556#issuecomment-3758094925

## How was this PR tested?

colcon build

## Notes for reviewers

The same code is already used inside this repo, like:

https://github.com/autowarefoundation/autoware_universe/blob/1eecaa86f1959a4464f801b053cfb53e22255523/common/autoware_universe_utils/src/geometry/random_concave_polygon.cpp#L31-L33

## Interface changes

None.

## Effects on system behavior

The packages can be build with differenct boost versions.


kyoichi-sugahara (approved):
Thank you! LGTM for autoware_autonomous_emergency_braking package's change

PR #11890, opened 2026-01-16
chore(autoware_camera_streampetr): remove cudnn dependency

amadeuszsz:
## Description

Remove CUDNN as it is unused dependency.

## Related links

**Parent Issue:**

- https://github.com/autowarefoundation/autoware/issues/6729#issuecomment-3757777221

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


SamratThapa120 (approved):
Thanks for the PR. LGTM
I was able to successfully build and test that the node is functioning normally.

PR #11901, opened 2026-01-16
chore(autoware_diffusion_planner): remove cudnn dependency

amadeuszsz:
## Description

Remove CUDNN as it is unused dependency.

## Related links

**Parent Issue:**

- https://github.com/autowarefoundation/autoware/issues/6729#issuecomment-3757777221

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


SakodaShintaro (approved):
Thank you very much.
LGTM

Note: I have confirmed that `planning_simulator` works properly.

PR #11906, opened 2026-01-19
ci: adjust checkout depth

amadeuszsz:
## Description

"Differentials" CI workflows fail because checkout depth is not set.

Sample error: https://github.com/autowarefoundation/autoware_universe/actions/runs/21116746716/job/60723420240?pr=11898

This PR applies checkout as we do in other workflows. 

## Related links

**Parent Issue:**

- Link

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/.github/workflows/build-test-tidy-pr.yaml
+++ b/.github/workflows/build-test-tidy-pr.yaml
@@ -24,7 +24,14 @@ jobs:
     outputs:
       cuda_job_is_needed: ${{ steps.check.outputs.any_changed }}
     steps:
-      - uses: actions/checkout@v4
+      - name: Set PR fetch depth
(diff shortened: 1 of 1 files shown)


xmfcx (changes requested):
image

You are modifying `check-if-cuda-job-is-needed:` even though another job has failed.

xmfcx:
This issue is common and it is a weird one.

It is sort of flaky and I still didn't figure out why it happens.

The solution is simple: rebase the PR branch to the latest main and it will be fixed.

amadeuszsz:
@xmfcx 
I don't understand why build-and-test-differential didn't fail... It uses only different runner and that's it.

amadeuszsz:
Closing as proposed change is irrelevant.

xmfcx:
it is flaky and somehow related to where the tip of remote main is.

PR #11936, opened 2026-01-20
fix(behavior_path_start_planner_module): add missing dependencies to CMakeLists

xmfcx:
## Description

- Similar to: https://github.com/autowarefoundation/autoware_core/blob/646ce7b30b9be1b127df12f4f9676a516a6bfabc/common/autoware_lanelet2_utils/CMakeLists.txt#L71-L84

### Before

So many linking errors to `autoware_pyplot`.

### After

Compiles successfully with GCC 13 and ROS 2 Jazzy.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_path_planner/autoware_behavior_path_start_planner_module/CMakeLists.txt
+++ b/planning/behavior_path_planner/autoware_behavior_path_start_planner_module/CMakeLists.txt
@@ -18,9 +18,6 @@ ament_auto_add_library(${PROJECT_NAME} SHARED
 )
 
 if(BUILD_TESTING)
-  find_package(ament_lint_auto REQUIRED)
-  ament_lint_auto_find_test_dependencies()
-
   find_package(ament_cmake_ros REQUIRED)
   ament_auto_find_test_dependencies()
 


planning/behavior_path_planner/autoware_behavior_path_start_planner_module/CMakeLists.txt
```diff
@@ -21,6 +21,16 @@ if(BUILD_TESTING)
   find_package(ament_lint_auto REQUIRED)
   ament_lint_auto_find_test_dependencies()
 
+  find_package(ament_cmake_ros REQUIRED)
+  ament_auto_find_test_dependencies()
```
mitsudome-r, line 22: Did you have to explicitly find ament_cmake_ros?
I think it should be automatically found by ament_auto_find_test_dependencies()

```suggestion
  ament_auto_find_test_dependencies()
```

xmfcx: https://github.com/ament/ament_cmake/blob/jazzy/ament_cmake_auto/cmake/ament_auto_find_test_dependencies.cmake

`ament_auto_find_test_dependencies()` cmake macro is part of `ament_cmake_ros` package.

Regardless of that, apparently, 
- https://github.com/ament/ament_cmake/blob/jazzy/ament_cmake_auto/cmake/ament_auto_find_test_dependencies.cmake
- https://github.com/ament/ament_lint/blob/jazzy/ament_lint_auto/cmake/ament_lint_auto_find_test_dependencies.cmake

These 2 files are identical.
I thought `ament_lint_auto_find_test_dependencies()` would find the linter dependencies like `ament_cmake_cppcheck`, `ament_cmake_clang_format` etc.

But they are the same so I will remove the lint version and only keep the  `ament_cmake_ros` variant to avoid confusion in the future.

xmfcx: https://github.com/autowarefoundation/autoware_universe/pull/11936/commits/c20dbb1f6d77673ab1a6084d5b4864c6410cde80 with this it is updated.

PR #11937, opened 2026-01-20
fix(behavior_velocity_intersection_module): update test dependencies

xmfcx:
## Description

Just added missing dependencies.

The code uses
https://github.com/autowarefoundation/autoware_universe/blob/0446229b26093de8535aeb1bda9a0810ce9bcfa8/planning/behavior_velocity_planner/autoware_behavior_velocity_intersection_module/test/test_planning_factor.cpp#L70

But doesn't declare in package.xml and doesn't refer to it in cmakelists.

This PR fixes it.

### Before

```bash
/usr/bin/ld: CMakeFiles/test_autoware_behavior_velocity_intersection_module.dir/test/test_planning_factor.cpp.o: undefined reference to symbol '_ZN4YAML6detail9node_data5beginEv'
/usr/bin/ld: /lib/x86_64-linux-gnu/libyaml-cpp.so.0.8: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
gmake[2]: *** [CMakeFiles/test_autoware_behavior_velocity_intersection_module.dir/build.make:827: test_autoware_behavior_velocity_intersection_module] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:185: CMakeFiles/test_autoware_behavior_velocity_intersection_module.dir/all] Error 2
---
Failed   <<< autoware_behavior_velocity_intersection_module [9.30s, exited with code 2]
```

### After

Compiles successfully with GCC 13 and ROS 2 Jazzy.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_velocity_planner/autoware_behavior_velocity_intersection_module/CMakeLists.txt
+++ b/planning/behavior_velocity_planner/autoware_behavior_velocity_intersection_module/CMakeLists.txt
@@ -40,9 +40,6 @@ target_link_libraries(${PROJECT_NAME}
 ament_auto_package(INSTALL_TO_SHARE config test_data)
 
 if(BUILD_TESTING)
-  find_package(ament_lint_auto REQUIRED)
-  ament_lint_auto_find_test_dependencies()
-
   find_package(ament_cmake_ros REQUIRED)
   ament_auto_find_test_dependencies()
 


planning/behavior_velocity_planner/autoware_behavior_velocity_intersection_module/CMakeLists.txt
```diff
@@ -42,11 +42,18 @@ ament_auto_package(INSTALL_TO_SHARE config test_data)
 if(BUILD_TESTING)
   find_package(ament_lint_auto REQUIRED)
   ament_lint_auto_find_test_dependencies()
+
+  find_package(ament_cmake_ros REQUIRED)
+  ament_auto_find_test_dependencies()
```
mitsudome-r, line 44: Did you have to explicitly find ament_cmake_ros?
I think it should be automatically found by ament_auto_find_test_dependencies()

```suggestion
  ament_auto_find_test_dependencies()
```

xmfcx: - https://github.com/autowarefoundation/autoware_universe/pull/11936#discussion_r2716926127

Similar to that, with https://github.com/autowarefoundation/autoware_universe/pull/11937/commits/d752c383c2697d1e9f135645c8666bf3d8c46cf5 only kept `ament_auto_find_test_dependencies()`

PR #11941, opened 2026-01-21
fix(lane_change): validate linestring tag for all path candidates

zulfaqar-azmi-t4:
## Description

Adds `lane_change=no` linestring validation when generating candidate path.

## Related links

**Parent Issue:**

- [TIER IV internal slack inquiry](https://star4.slack.com/archives/C074CCKN35E/p1768550804165509)

## How was this PR tested?

1. Planning Simulator

|Before|After|
|:-:|:-:|
| <video src="https://github.com/user-attachments/assets/b24fa8f5-0dc9-4f9d-a2b3-94eafaf84a46"> | <video src="https://github.com/user-attachments/assets/08e6b000-7da5-441f-8186-ae4beec8086d"> |
| Candidate path crosses lane change prohibited area | Candidate path doesn't cross lane change prohibited area | 

2. TIER IV internal evaluation
[TIER IV Internal Link - ControlDLR Test](https://evaluation.tier4.jp/evaluation/reports/a20bbf98-72ac-5460-aab9-460c8520890e?project_id=prd_jt)
[TIER IV Internal Link - FuncVerification Test](https://evaluation.tier4.jp/evaluation/reports/ee085071-6058-5c2b-9eb5-7217908935b2?project_id=prd_jt)
[TIER IV Internal Link - CommonScenario_Basic Test](https://evaluation.tier4.jp/evaluation/reports/4697efc6-1c8e-50c2-82c3-fac80389efed?project_id=autoware_dev)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/include/autoware/behavior_path_lane_change_module/utils/utils.hpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/include/autoware/behavior_path_lane_change_module/utils/utils.hpp
@@ -520,7 +520,7 @@ std::vector<lanelet::ConstLineString3d> get_no_lane_change_lines(
  * @param ego_pose              Current ego vehicle pose used as distance reference point.
  * @return Vector of distance intervals [start, end] representing no-lane-change zones.
  */
-std::vector<std::pair<double, double>> get_inverval_dist_no_lane_change_lines(
+std::vector<std::pair<double, double>> get_interval_dist_no_lane_change_lines(
   const std::vector<lanelet::ConstLineString3d> & no_lane_change_lines,
   const PathWithLaneId & centerline_path, const Pose & ego_pose);
 
@@ -534,7 +534,7 @@ std::vector<std::pair<double, double>> get_inverval_dist_no_lane_change_lines(
  * @return True if the distance point falls within at least one buffered interval, false otherwise.
  */
 bool is_intersecting_no_lane_change_lines(
-  const std::vector<std::pair<double, double>> & inverval_dist_no_lane_change_lines,
+  const std::vector<std::pair<double, double>> & interval_dist_no_lane_change_lines,
   const double expected_intersecting_dist, const double buffer);
 }  // namespace autoware::behavior_path_planner::utils::lane_change
 #endif  // AUTOWARE__BEHAVIOR_PATH_LANE_CHANGE_MODULE__UTILS__UTILS_HPP_
--- a/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/utils/utils.cpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/utils/utils.cpp
@@ -1284,7 +1284,7 @@ std::vector<lanelet::ConstLineString3d> get_no_lane_change_lines(
   return no_lane_change_lines;
 }
 
-std::vector<std::pair<double, double>> get_inverval_dist_no_lane_change_lines(
+std::vector<std::pair<double, double>> get_interval_dist_no_lane_change_lines(
   const std::vector<lanelet::ConstLineString3d> & no_lane_change_lines,
   const PathWithLaneId & centerline_path, const Pose & ego_pose)
 {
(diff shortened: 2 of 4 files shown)


planning/behavior_path_planner/autoware_behavior_path_lane_change_module/include/autoware/behavior_path_lane_change_module/utils/utils.hpp
```diff
@@ -501,5 +501,12 @@ bool is_lanelet_in_lanelet_collections(
  */
 void trim_preferred_after_alternative(
   lanelet::ConstLanelets & base_lanes, const lanelet::ConstLanelets & preferred_lanes);
+
+std::vector<lanelet::ConstLineString3d> get_no_lane_change_lines(
+  const lanelet::ConstLanelets & target_lanes, const Direction direction);
+
+bool is_intersecting_no_lane_change_lines(
+  const std::vector<lanelet::ConstLineString3d> & no_lane_change_lines,
+  const std::vector<PathPointWithLaneId> & lane_change_path);
```
maxime-clem: (optional) some short docstrings would be nice

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/utils/utils.cpp
```diff
@@ -1255,4 +1255,63 @@ void trim_preferred_after_alternative(
...
+
+    if ((is_solid && !explicit_yes) || explicit_no) {
+      no_lane_change_lines.push_back(ls);
+    }
+  }
+
+  return no_lane_change_lines;
+}
+
+bool is_intersecting_no_lane_change_lines(
```
maxime-clem: Is there no concern about performance ? If there are many "no lane change lines"  in the map this function may become expensive.
(optional) store the "no lane change lines" segments in a Rtree for faster intersection checks.

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/utils/utils.cpp
```diff
@@ -1255,4 +1255,63 @@ void trim_preferred_after_alternative(
...
+  using ranges::views::sliding;
+
+  // Create a sliding view over the path
+  auto path_segments = lane_change_path | ranges::views::sliding(2);
+
+  return ranges::any_of(path_segments, [&](const auto & p_seg) {
+    // FIX 1: Use iterators for the path segment
+    auto p_it = p_seg.begin();
+    const auto & path_p1 = p_it->point.pose.position;
+    const auto & path_p2 = (++p_it)->point.pose.position;
```
maxime-clem: (FYI) there is a ["for each segment" function](https://www.boost.org/doc/libs/latest/libs/geometry/doc/html/geometry/reference/algorithms/for_each/for_each_segment.html).

mkquda:
@zulfaqar-azmi-t4 
1- In the "After" video the stop line say "no safe path" and terminal path is generated, which means the LC node is setting the terminal point there, which is not correct ...

2- Instead of checking for geometric intersection, which can cause performance issues with too many candidate paths, we can use distance measurements:
- compute distance to start of no lane change once
- for each candidate path compare it to prepare length + shift length

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/include/autoware/behavior_path_lane_change_module/utils/utils.hpp
zulfaqar-azmi-t4: https://github.com/autowarefoundation/autoware_universe/pull/11941/commits/95d4b9d49076dcf4797e9c36135bfac603162dca

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/utils/utils.cpp
zulfaqar-azmi-t4: function no longer used.

planning/behavior_path_planner/autoware_behavior_path_lane_change_module/src/utils/utils.cpp
zulfaqar-azmi-t4: I'm using distance based to verify, which should be a bit faster compared to checking intersection.

zulfaqar-azmi-t4:
@mkquda 

> 1- In the "After" video the stop line say "no safe path" and terminal path is generated, which means the LC node is setting the terminal point there, which is not correct ...

It's a mistake when setting the goal pose. It is replaced with new exampple.

> 2- Instead of checking for geometric intersection, which can cause performance issues with too many candidate paths, we can use distance measurements:
> 
> compute distance to start of no lane change once
> for each candidate path compare it to prepare length + shift length

done in https://github.com/autowarefoundation/autoware_universe/pull/11941/commits/97e9f566e6df01b5d846e9e27891d772526f1d89 and https://github.com/autowarefoundation/autoware_universe/pull/11941/commits/dd23d22289d053ef6bde64b3c472d740e81b4054

PR #11954, opened 2026-01-22
fix(hazard_status_converter): use root output level for emergency holding

isamu-takagi:
## Description

Fix an issue where the emergency_holding was not calculated correctly when latch was not set to leaf units only.
The latch level does not propagate up the parent, so it must refer to the output level.

## Related links

**Private Links:**

- [TIER IV link](https://star4.slack.com/archives/C017VB9UG1L/p1768889847290329?thread_ts=1729675438.504999&cid=C017VB9UG1L)

## How was this PR tested?

1. Set latch to any leaf unit in diagnostic graph config.
2. Raise the error that set the latch above
3. Check `emergency_holding` is true after latch second above.
    `ros2 topic echo /system/emergency/hazard_status --no-arr`

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


sasakisasaki:
Updated the branch. Waiting the CIs.

PR #11959, opened 2026-01-22
fix(bevfusion): suppress -Werror for precomputed_features.cpp

xmfcx:
## Description

- Part of: https://github.com/autowarefoundation/autoware_universe/issues/11418

This was compiling before but somehow started giving these errors after:
- https://github.com/autowarefoundation/autoware_universe/pull/11887
- https://github.com/autowarefoundation/autoware_universe/pull/11953
- Or maybe any of the cuda related dependency PRs

Anyway this PR just disables that warning for this single file.

I tried to fix it too but it has gotten too complex. I will leave that to future.

### Before

Full error logs: https://gist.github.com/xmfcx/d872d111014a239714e28c8b2f454bbb

### After

Compiles with Jazzy & GCC 13.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_bevfusion/lib/preprocess/precomputed_features.cpp
+++ b/perception/autoware_bevfusion/lib/preprocess/precomputed_features.cpp
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// See https://github.com/autowarefoundation/autoware_universe/pull/11959 for details.
 #pragma GCC diagnostic ignored "-Warray-bounds"
 
 #include "autoware/bevfusion/preprocess/precomputed_features.hpp"


perception/autoware_bevfusion/lib/preprocess/precomputed_features.cpp
```diff
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#pragma GCC diagnostic ignored "-Warray-bounds"
```
mitsudome-r, line 16: Should we leave some comments why we are adding this ignore "-Warray-bounds"?

xmfcx: Sure, added https://github.com/autowarefoundation/autoware_universe/pull/11959/commits/8f1bd661bd880346e7ebb0be38b8638ff2fdf61d for this 👍

PR #11960, opened 2026-01-22
fix(agnocast): build on jazzy, remove from ground_segmentation_cuda

xmfcx:
## Description

- Part of https://github.com/autowarefoundation/autoware_universe/issues/11418

Because Agnocast is not available for ROS 2 Jazzy, we have to skip building it until the Agnocast team makes it ready.

This PR skips building `autoware_agnocast_wrapper` and `autoware_cuda_pointcloud_preprocessor` packages if `ROS_DISTRO != humble`.

And finally [autoware_ground_segmentation_cuda](https://github.com/autowarefoundation/autoware_universe/tree/9cc6de76b99eec6d708684b6e3e8a935fd9ab3db/perception/autoware_ground_segmentation_cuda) references agnocast in its CMakeLists.txt file but doesn't use it anywhere. So I removed the unused dependency from that package.

## How was this PR tested?

During build, the following packages skip correctly:
- `autoware_agnocast_wrapper` 
- `autoware_cuda_pointcloud_preprocessor`

Entire autoware_universe compiles with this PR on Jazzy.

## Notes for reviewers

I thought about making `autoware_cuda_pointcloud_preprocessor` build and disable some pathways in the code if agnocast was referred. But it made use of Agnocast macros like:
- https://github.com/autowarefoundation/autoware_universe/blob/9cc6de76b99eec6d708684b6e3e8a935fd9ab3db/sensing/autoware_cuda_pointcloud_preprocessor/include/autoware/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor_node.hpp#L87
- https://github.com/autowarefoundation/autoware_universe/blob/9cc6de76b99eec6d708684b6e3e8a935fd9ab3db/sensing/autoware_cuda_pointcloud_preprocessor/include/autoware/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor_node.hpp#L132

So I gave up and just disabled its compilation until Agnocast is made available (probably within a month).

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/common/autoware_agnocast_wrapper/CMakeLists.txt
+++ b/common/autoware_agnocast_wrapper/CMakeLists.txt
@@ -2,7 +2,6 @@ cmake_minimum_required(VERSION 3.14)
 project(autoware_agnocast_wrapper)
 
 find_package(autoware_cmake REQUIRED)
-
 autoware_package()
 
 find_package(rclcpp REQUIRED)
--- a/sensing/autoware_cuda_pointcloud_preprocessor/CMakeLists.txt
+++ b/sensing/autoware_cuda_pointcloud_preprocessor/CMakeLists.txt
@@ -2,7 +2,6 @@ cmake_minimum_required(VERSION 3.14)
 project(autoware_cuda_pointcloud_preprocessor)
 
 find_package(ament_cmake_auto REQUIRED)
-
 find_package(CUDA)
 
 if("$ENV{ENABLE_AGNOCAST}")


badai-nguyen (approved):
LGTM regarding `autowaer_ground_segmentation_cuda`

sensing/autoware_cuda_pointcloud_preprocessor/package.xml
```diff
@@ -15,7 +15,7 @@
   <buildtool_depend>ament_cmake_auto</buildtool_depend>
   <buildtool_depend>autoware_cmake</buildtool_depend>
 
-  <depend>agnocastlib</depend>
+  <build_depend condition="$ROS_DISTRO == humble">agnocastlib</build_depend>
```
xmfcx:
```suggestion
  <depend condition="$ROS_DISTRO == humble">agnocastlib</depend>
```

I didn't intend to change the dependency type.

common/autoware_agnocast_wrapper/package.xml
```diff
@@ -10,7 +10,7 @@
   <buildtool_depend>ament_cmake_auto</buildtool_depend>
   <buildtool_depend>autoware_cmake</buildtool_depend>
 
-  <build_depend>agnocastlib</build_depend>
+  <build_depend condition="$ROS_DISTRO == humble">agnocastlib</build_depend>
```
amadeuszsz: > Because Agnocast is not available for ROS 2 Jazzy, we have to skip building it until the Agnocast team makes it ready.

Could you please take a look on this?
```xml
  <depend condition="$ENABLE_AGNOCAST == 1">agnocastlib</depend>
```

The idea is taken from https://github.com/tier4/nebula/pull/397

I believe we should follow one approach for handling `agnocastlib` if possible.

common/autoware_agnocast_wrapper/CMakeLists.txt
mojomex, line 1: This has to be built even when Agnocast is not available (ensuring that only the non-Agnocast versions of the macros are compiled)

common/autoware_agnocast_wrapper/package.xml
xmfcx: Looks like a cleaner solution, thx! Applied in https://github.com/autowarefoundation/autoware_universe/pull/11960/commits/f720f19c42cbc24644924d51f190870038999d57

common/autoware_agnocast_wrapper/CMakeLists.txt
xmfcx: I didnt know it was meant to be buildable without agnocast. It is just exporting headers after all so it turned out to be fine 👍

Applied with: https://github.com/autowarefoundation/autoware_universe/pull/11960/commits/3a3d31b901cab49ee02082b8f4350b7395afbf1d

Both packages build in jazzy now (as long as agnocast enable flag is not set ofc) ✅

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor_node.cpp
```diff
@@ -21,7 +21,7 @@
 #include "autoware/pointcloud_preprocessor/diagnostics/latency_diagnostics.hpp"
 #include "autoware/pointcloud_preprocessor/diagnostics/pass_rate_diagnostics.hpp"
 
-#include <agnocast/agnocast_subscription.hpp>
+#include <autoware/agnocast_wrapper/autoware_agnocast_wrapper.hpp>
```
xmfcx, line 24: @mojomex I had to update this header, it was previously using agnocast directly. With wrapper it doesn't error out anymore.

xmfcx:
Good thing with this latest implementation is that it brings cleaner separation whether you want to use agnocast or not.

And smaller diff footprint.

Also I saw this PR as something to be reverted when Agnocast support came to jazzy. But now this doesn't need to be reverted at all.

Only `  <depend condition="$ENABLE_AGNOCAST == 1">agnocastlib</depend>` these conditions can be removed, that's it.

common/autoware_agnocast_wrapper/CMakeLists.txt
mojomex: Thanks! I was worried that the `include` directory wouldn't be made available for other packages but now it's definitely fine :+1:

PR #11971, opened 2026-01-26
feat(autoware_system_monitor): change diagnostic logic for mem_monitor

takam5f2:
### Motivation and current issues
This PR modifies the logic used to determine memory exhaustion. The current logic has several issues outlined below, prompting the need for this change.

1. **Ambiguity of the monitored metric:**
    * The current metrics are based on empirical values, making their definitions vague.
    * It introduces a custom, proprietary metric called `used_plus`, which is difficult for general developers to understand.

2. **Reliance on Swap existence:**
    * If Swap is removed from `used_plus`, the status almost never shifts away from `OK` (potentially causing false negatives).
    * The logic assumes Swap access is permissible. However, this is incompatible with real-time software requirements, where avoiding access to non-volatile storage is critical.

3. **Weak rationale for threshold settings:**
    * The current thresholds are based on short-term experiments, lacking sufficient validity.

## Proposal
I propose transitioning to a logic based on the `available` metric from the standard `free` command.

For systems relying solely on main memory, `available` alone is a sufficient indicator. However, for systems with Swap enabled, we must account for Swap usage, as utilizing Swap while memory is "available" is undesirable.

### Implementation Details
Memory available and Swap usage are checked independently:

1. **Memory available check**: Compares `available` against `error_available_size` and `warning_available_size` thresholds.
2. **Swap usage check**: Any Swap usage (`swap_used > 0`) generates a Warning. If Swap usage exceeds `swap_error_threshold`, it generates an Error.

### Threshold Configuration
The thresholds are determined by the following external parameters: `available_size`, `warning_margin`, and `swap_error_threshold`.

* **`available_size`**: Inherited from the existing configuration. This parameter alone functions as the threshold for the **Error** state.
* **`warning_margin`**: When this parameter is set, the sum of `available_size` and `warning_margin` functions as the threshold for the **Warning** state. If `warning_margin` is not set, the Warning state is disabled.
* **`swap_error_threshold`**: Threshold for Swap usage to generate the **Error** state. Any Swap usage above zero generates a Warning.

## Concerns
I have also updated the unit tests. However, there is a concern regarding their low reproducibility (flakiness), which may potentially cause CI/CD failures.

While I have ensured they pass in this PR, if this risk is unacceptable, please let me know so we can discuss a solution.

## Related links

N/A

**Parent Issue:**

- Link

## How was this PR tested?

Please prepare the branch.
```bash
cd /path/to/your/autoware_universe
git remote add takam5f2 git@github.com:takam5f2/autoware.universe.git 
git fetch takam5f2 
git checkout takam5f2/feat/change-diagnostic-logic-for-memory
```

Unit test with the following command.

```bash
colcon build --symlink-install --cmake-args -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_FLAGS="-w" -DBUILD_TESTING=true --packages-select autoware_system_monitor
colcon test --packages-select autoware_system_monitor  --event-handlers console_direct+
```

Launch sotftware with ros2 launch

```bash
ros2 launch autoware_system_monitor system_monitor.launch.xml 
```

You'll find the diagnostic result with the runtime monitor.

```bash
```

## Notes for reviewers
The documentation for [ROS parameters](https://github.com/autowarefoundation/autoware_universe/pull/11971) is outdated.

The current document states that memory saturation criteria are based on percentage. usage_warn and uasge_error are threshold parameters for the previous design. However, this logic was updated a few years ago to use memory consumption in "MiB" instead, but the documentation was not updated accordingly.

#### Additions and removals

| Added | `warning_margin`   | `int` | `0`         |  Margin MiB for generating warning for memory saturation |
| Added | `swap_error_threshold`   | `int` | `1024`         |  Swap usage MiB threshold for generating error |

### DIFF
--- a/build_depends_humble.repos
+++ b/build_depends_humble.repos
@@ -24,7 +24,7 @@ repositories:
   core/autoware_lanelet2_extension:
     type: git
     url: https://github.com/autowarefoundation/autoware_lanelet2_extension.git
-    version: 0.11.0
+    version: 0.12.0
   core/autoware_core:
     type: git
     url: https://github.com/autowarefoundation/autoware_core.git
--- a/planning/autoware_diffusion_planner/CMakeLists.txt
+++ b/planning/autoware_diffusion_planner/CMakeLists.txt
@@ -99,6 +99,9 @@ if(TRT_AVAIL AND CUDA_AVAIL)
 
     target_link_libraries(test_diffusion_planner autoware_diffusion_planner_component)
     ament_target_dependencies(test_diffusion_planner
+      autoware_lanelet2_extension
+      autoware_map_msgs
+      autoware_perception_msgs
       autoware_test_utils
     )
   endif()
--- a/planning/autoware_diffusion_planner/README.md
+++ b/planning/autoware_diffusion_planner/README.md
@@ -189,14 +189,15 @@ Parameters can be set via YAML (see `config/diffusion_planner.param.yaml`).
 
 ## Outputs
 
-| Topic                        | Message Type                                              | Description                                |
-| ---------------------------- | --------------------------------------------------------- | ------------------------------------------ |
-| `~/output/trajectory`        | autoware_planning_msgs/msg/Trajectory                     | Planned trajectory for the ego vehicle     |
-| `~/output/trajectories`      | autoware_internal_planning_msgs/msg/CandidateTrajectories | Multiple candidate trajectories            |
-| `~/output/predicted_objects` | autoware_perception_msgs/msg/PredictedObjects             | Predicted future states of dynamic objects |
-| `~/output/turn_indicators`   | autoware_vehicle_msgs/msg/TurnIndicatorsCommand           | Planned turn indicator command             |
-| `~/debug/lane_marker`        | visualization_msgs/msg/MarkerArray                        | Lane debug markers                         |
-| `~/debug/route_marker`       | visualization_msgs/msg/MarkerArray                        | Route debug markers                        |
+| Topic                           | Message Type                                              | Description                                                |
+| ------------------------------- | --------------------------------------------------------- | ---------------------------------------------------------- |
+| `~/output/trajectory`           | autoware_planning_msgs/msg/Trajectory                     | Planned trajectory for the ego vehicle                     |
+| `~/output/trajectories`         | autoware_internal_planning_msgs/msg/CandidateTrajectories | Multiple candidate trajectories                            |
+| `~/output/predicted_objects`    | autoware_perception_msgs/msg/PredictedObjects             | Predicted future states of dynamic objects                 |
+| `~/output/turn_indicators`      | autoware_vehicle_msgs/msg/TurnIndicatorsCommand           | Planned turn indicator command                             |
+| `~/output/debug/traffic_signal` | autoware_perception_msgs/msg/TrafficLightGroup            | First traffic light on route (ego forward) for RViz/ad_api |
+| `~/debug/lane_marker`           | visualization_msgs/msg/MarkerArray                        | Lane debug markers                                         |
+| `~/debug/route_marker`          | visualization_msgs/msg/MarkerArray                        | Route debug markers                                        |
 
 ---
 
--- a/planning/autoware_diffusion_planner/include/autoware/diffusion_planner/diffusion_planner_node.hpp
+++ b/planning/autoware_diffusion_planner/include/autoware/diffusion_planner/diffusion_planner_node.hpp
@@ -41,6 +41,7 @@
 #include <autoware_map_msgs/msg/lanelet_map_bin.hpp>
 #include <autoware_perception_msgs/msg/predicted_objects.hpp>
 #include <autoware_perception_msgs/msg/tracked_objects.hpp>
+#include <autoware_perception_msgs/msg/traffic_light_group.hpp>
 #include <autoware_perception_msgs/msg/traffic_signal.hpp>
 #include <autoware_planning_msgs/msg/lanelet_route.hpp>
 #include <autoware_planning_msgs/msg/trajectory.hpp>
@@ -214,6 +215,12 @@ private:
     const rclcpp::Time & timestamp) const;
 
   /**
+   * @brief Publish the first traffic light on the route (from ego forward) for debug.
+   * @param frame_context Context of the current frame (ego pose).
+   */
+  void publish_first_traffic_light_on_route(const FrameContext & frame_context) const;
+
+  /**
    * @brief Publish model predictions.
    * @param predictions Output from the model.
    * @param frame_context Context of the current frame.
@@ -271,6 +278,8 @@ private:
   rclcpp::Publisher<MarkerArray>::SharedPtr pub_lane_marker_{nullptr};
   rclcpp::Publisher<MarkerArray>::SharedPtr pub_route_marker_{nullptr};
   rclcpp::Publisher<TurnIndicatorsCommand>::SharedPtr pub_turn_indicators_{nullptr};
+  rclcpp::Publisher<autoware_perception_msgs::msg::TrafficLightGroup>::SharedPtr
+    pub_traffic_signal_{nullptr};
   mutable std::shared_ptr<autoware_utils::TimeKeeper> time_keeper_{nullptr};
   autoware_utils::InterProcessPollingSubscriber<Odometry> sub_current_odometry_{
     this, "~/input/odometry"};
--- a/planning/autoware_diffusion_planner/include/autoware/diffusion_planner/preprocessing/lane_segments.hpp
+++ b/planning/autoware_diffusion_planner/include/autoware/diffusion_planner/preprocessing/lane_segments.hpp
@@ -20,6 +20,7 @@
 #include "autoware/diffusion_planner/preprocessing/traffic_signals.hpp"
 #include "autoware/traffic_light_utils/traffic_light_utils.hpp"
 
+#include <autoware_perception_msgs/msg/traffic_light_group.hpp>
 #include <autoware_perception_msgs/msg/traffic_light_group_array.hpp>
 #include <autoware_planning_msgs/msg/lanelet_route.hpp>
 #include <geometry_msgs/msg/detail/point__struct.hpp>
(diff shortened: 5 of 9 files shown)


system/autoware_system_monitor/src/mem_monitor/mem_monitor.cpp
```diff
@@ -34,8 +34,17 @@ namespace bp = boost::process;
 MemMonitor::MemMonitor(const rclcpp::NodeOptions & options)
 : Node("mem_monitor", options),
   updater_(this),
-  available_size_(declare_parameter<int>("available_size", 1024) * 1024 * 1024)
+  error_available_size_(declare_parameter<int>("available_size", 1024) * 1024 * 1024),
+  warning_available_size_()
 {
+  // Define warning_available_size_
+  int warning_margin = declare_parameter<int>("warning_margin", 0);
```
nishikawa-masaki: If you add the new parameter "warning_margin", you need to mention it in the "Interface changes" part of the PR.

system/autoware_system_monitor/docs/ros_parameters.md
```diff
@@ -52,10 +52,10 @@ hdd_monitor:
...
-| :---------- | :---: | :-----: | :-----: | :-------------------------------------------------------------------------------- |
-| usage_warn  | float | %(1e-2) |  0.95   | Generates warning when physical memory usage reaches a specified value or higher. |
-| usage_error | float | %(1e-2) |  0.99   | Generates error when physical memory usage reaches a specified value or higher.   |
+| Name           | Type | Unit | Default |                                                                Notes                                                                 |
+| :------------- | :--: | :--: | :-----: | :----------------------------------------------------------------------------------------------------------------------------------: |
+| available_size | int  | byte |  1024   |                  Generates error when physical memory usage reaches a specified value defined by `available_size`.                   |
+| warning_margin | int  | byte |    0    | Generates warning when physical memory usage reaches a specified value defined by addition of `available_size` and `warning_margin`. |
```
nishikawa-masaki: I think it would be better if you leave the following comment in the "Notes for reviwers" part of the PR.

The parameter description for mem_monitor in ros_parameters.md has not been maintained since the very early version.
It listed the outdated parameters "usage_warn" and "usage_error" while it failed to describe the current parameter "available_size".

system/autoware_system_monitor/docs/ros_parameters.md
```diff
@@ -52,10 +52,10 @@ hdd_monitor:
 
 mem_monitor:
 
-| Name        | Type  |  Unit   | Default | Notes                                                                             |
-| :---------- | :---: | :-----: | :-----: | :-------------------------------------------------------------------------------- |
-| usage_warn  | float | %(1e-2) |  0.95   | Generates warning when physical memory usage reaches a specified value or higher. |
-| usage_error | float | %(1e-2) |  0.99   | Generates error when physical memory usage reaches a specified value or higher.   |
+| Name           | Type | Unit | Default |                                                                Notes                                                                 |
+| :------------- | :--: | :--: | :-----: | :----------------------------------------------------------------------------------------------------------------------------------: |
+| available_size | int  | byte |  1024   |                  Generates error when physical memory usage reaches a specified value defined by `available_size`.                   |
```
nishikawa-masaki: The Unit for "available_size" and "warning_margin" is "MiB" (1024 * 1024 byte), not byte.
"MiB" is preferable to "MB" for clarity.

system/autoware_system_monitor/config/mem_monitor.param.yaml
```diff
@@ -1,3 +1,4 @@
 /**:
   ros__parameters:
     available_size: 1024 # MB
+    warning_maring: 0 # MB
```
nishikawa-masaki: Typo: warning_margin

system/autoware_system_monitor/src/mem_monitor/mem_monitor.cpp
nishikawa-masaki: For consistency with line 37, the following would be preferable, I think.
` int warning_margin = declare_parameter<int>("warning_margin", 0) * 1024 * 1024;`

I mean "1024 * 1024" should be here.

system/autoware_system_monitor/src/mem_monitor/mem_monitor.cpp
```diff
@@ -34,8 +34,17 @@ namespace bp = boost::process;
...
+  if (warning_margin < 0) {
+    warning_margin = 0;
+  }
+  warning_available_size_ =
+    error_available_size_ + static_cast<size_t>(warning_margin) * 1024 * 1024;
```
nishikawa-masaki: Please read the previous comment.

system/autoware_system_monitor/src/mem_monitor/mem_monitor.cpp
takam5f2: Sorry for my late reply, I updated the description.

system/autoware_system_monitor/docs/ros_parameters.md
takam5f2: I updated the description.

system/autoware_system_monitor/docs/ros_parameters.md
takam5f2: Thank you for checking. I updated at https://github.com/autowarefoundation/autoware_universe/pull/11971/commits/5958219a27d0f26f29c9cd7d309d14beb28b1139

system/autoware_system_monitor/config/mem_monitor.param.yaml
takam5f2: Thank you for checking. I updated at https://github.com/autowarefoundation/autoware_universe/commit/5958219a27d0f26f29c9cd7d309d14beb28b1139

system/autoware_system_monitor/src/mem_monitor/mem_monitor.cpp
takam5f2: Thank you for checking. I updated at https://github.com/autowarefoundation/autoware_universe/commit/5958219a27d0f26f29c9cd7d309d14beb28b1139

system/autoware_system_monitor/src/mem_monitor/mem_monitor.cpp
takam5f2: Thank you for checking. I updated at https://github.com/autowarefoundation/autoware_universe/commit/5958219a27d0f26f29c9cd7d309d14beb28b1139

system/autoware_system_monitor/src/mem_monitor/mem_monitor.cpp
```diff
@@ -34,8 +34,16 @@ namespace bp = boost::process;
 MemMonitor::MemMonitor(const rclcpp::NodeOptions & options)
 : Node("mem_monitor", options),
   updater_(this),
-  available_size_(declare_parameter<int>("available_size", 1024) * 1024 * 1024)
+  error_available_size_(declare_parameter<int>("available_size", 1024) * 1024 * 1024),
+  warning_available_size_()
```
nishikawa-masaki: I believe empty () initializes an int value with 0.
However, "warning_available_size_(0)" would be explicit and clear.

system/autoware_system_monitor/src/mem_monitor/mem_monitor.cpp
```diff
@@ -34,8 +34,16 @@ namespace bp = boost::process;
...
 {
+  // Define warning_available_size_
+  int warning_margin = declare_parameter<int>("warning_margin", 0) * 1024 * 1024;
```
nishikawa-masaki: Sorry. I failed to submit the following comment last time.
If you don't have time to fix and test or this is irrelevant to this PR, you can ignore this comment.

--- 

The specification of ROS2 requires the following:
* If not defined as "read-only", parameters can be modified during operation.
  * This is a bit demanding, but the specification says so.
  * [ParameterDescriptor.html](https://docs.ros.org/en/humble/p/rcl_interfaces/msg/ParameterDescriptor.html)

[cpu_monitor_base.cpp](https://github.com/autowarefoundation/autoware_universe/blob/main/system/autoware_system_monitor/src/cpu_monitor/cpu_monitor_base.cpp) can be an example.

system/autoware_system_monitor/src/mem_monitor/mem_monitor.cpp
```diff
@@ -142,24 +152,24 @@ void MemMonitor::checkUsage(diagnostic_updater::DiagnosticStatusWrapper & stat)
       stat.add(fmt::format("{} shared", list[0]), toHumanReadable(list[4]));
       stat.add(fmt::format("{} buff/cache", list[0]), toHumanReadable(list[5]));
       stat.add(fmt::format("{} available", list[0]), toHumanReadable(list[6]));
-    } else if (index == 3) {
-      // Total:used + Mem:shared
-      used_plus = std::atoll(list[2].c_str()) + mem_shared;
-      double giga = static_cast<double>(used_plus) / (1024 * 1024 * 1024);
-      stat.add(fmt::format("{} used+", list[0]), fmt::format("{:.1f}{}", giga, "G"));
```
nishikawa-masaki, line 149: Removing the misleading "used+" is reasonable, but don't you report "swap_used" which is going to be an important entry?

If you report "swap_used", line 156 should be "// Do nothing for total".

takam5f2: @nishikawa-masaki  
Thank you for feedback.

mem_monitor reports `swap_used` according to the following lines.
https://github.com/takam5f2/autoware.universe/blob/98234cbdc605f5021de736a0567b5b9130c20322/system/autoware_system_monitor/src/mem_monitor/mem_monitor.cpp#L145-L148

image

I added clearer comment.

system/autoware_system_monitor/src/mem_monitor/mem_monitor.cpp
takam5f2: I added `ParameterDescriptor`. 
https://github.com/autowarefoundation/autoware_universe/pull/11971/commits/bbd6547e1b96164124127326610efee0115a54fa

system/autoware_system_monitor/src/mem_monitor/mem_monitor.cpp
takam5f2: @nishikawa-masaki 
I added the initial value. https://github.com/autowarefoundation/autoware_universe/pull/11971/commits/bbd6547e1b96164124127326610efee0115a54fa

PR #11980, opened 2026-01-27
fix(occupancy_grid_map_outlier_filter): fix funcArgNamesDifferent

veqcc:
## Description

Fixed `funcArgNamesDifferent`
```
perception/autoware_occupancy_grid_map_outlier_filter/src/occupancy_grid_map_outlier_filter_node.cpp:550:23: style: inconclusive: Function 'transformToBaseLink' argument 1 names different: declaration 'input' definition 'ros_input'. [funcArgNamesDifferent]
  const PointCloud2 & ros_input, [[maybe_unused]] const Header & header, PointCloud2 & output)
                      ^
perception/autoware_occupancy_grid_map_outlier_filter/src/occupancy_grid_map_outlier_filter_node.hpp:99:27: note: Function 'transformToBaseLink' argument 1 names different: declaration 'input' definition 'ros_input'.
      const PointCloud2 & input, const Header & header, PointCloud2 & output);
                          ^
perception/autoware_occupancy_grid_map_outlier_filter/src/occupancy_grid_map_outlier_filter_node.cpp:550:23: note: Function 'transformToBaseLink' argument 1 names different: declaration 'input' definition 'ros_input'.
  const PointCloud2 & ros_input, [[maybe_unused]] const Header & header, PointCloud2 & output)
                      ^
```

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


perception/autoware_occupancy_grid_map_outlier_filter/src/occupancy_grid_map_outlier_filter_node.hpp
```diff
@@ -96,7 +96,7 @@ class OccupancyGridMapOutlierFilterComponent : public rclcpp::Node
 
   private:
     void transformToBaseLink(
-      const PointCloud2 & input, const Header & header, PointCloud2 & output);
+      const PointCloud2 & ros_input, const Header & header, PointCloud2 & output);
```
yukkysaito: Since the header is also a ROS input, I think it could be misleading. It would be better to rename it to something like pointcloud_input.

veqcc: I just followed the name in .cpp implementation:
https://github.com/autowarefoundation/autoware_universe/blob/e852fba9e018eb99793647a13125b68f0d1cae17/perception/autoware_occupancy_grid_map_outlier_filter/src/occupancy_grid_map_outlier_filter_node.cpp#L549-L553

Do you think the implementation should also be refactored?

yukkysaito: Yes, I understand. Aligning it with ros_input would be misleading, so if we are going to modify it, we should change it to a more appropriate name.

veqcc: Fixed in https://github.com/autowarefoundation/autoware_universe/pull/11980/commits/f8bb97e918d10170759bcf17ea20482e0fd530da

PR #11983, opened 2026-01-27
fix(traffic_rule_filter): improve traffic light filter

maxime-clem:
## Description

Improve the `TrafficLightFilter` module in the `traffic_rule_filter` package.

## Related links

**Private Links:**

- [TIER IV JIRA link](https://tier4.atlassian.net/browse/T4DEV-46071)

## How was this PR tested?

Psim.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/autoware_trajectory_traffic_rule_filter/src/filters/traffic_light_filter.cpp
+++ b/planning/autoware_trajectory_traffic_rule_filter/src/filters/traffic_light_filter.cpp
@@ -16,9 +16,6 @@
 
 #include <autoware/traffic_light_utils/traffic_light_utils.hpp>
 
-#include <autoware_internal_planning_msgs/msg/path_point_with_lane_id.hpp>
-#include <autoware_internal_planning_msgs/msg/path_with_lane_id.hpp>
-
 #include <boost/geometry.hpp>
 
 #include <lanelet2_core/LaneletMap.h>
@@ -31,9 +28,6 @@
 
 namespace autoware::trajectory_traffic_rule_filter::plugin
 {
-using autoware_internal_planning_msgs::msg::PathPointWithLaneId;
-using autoware_internal_planning_msgs::msg::PathWithLaneId;
-
 TrafficLightFilter::TrafficLightFilter() : TrafficRuleFilterInterface("TrafficLightFilter")
 {
 }
@@ -62,7 +56,7 @@ std::vector<lanelet::BasicLineString2d> TrafficLightFilter::get_red_stop_lines(
 
 bool TrafficLightFilter::is_feasible(const TrajectoryPoints & trajectory_points)
 {
-  if (!lanelet_map_ || !traffic_lights_ || trajectory_points.empty()) {
+  if (!lanelet_map_ || !traffic_lights_ || trajectory_points.empty() || !vehicle_info_ptr_) {
     return true;  // Allow if no data available
   }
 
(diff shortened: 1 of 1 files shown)


planning/autoware_trajectory_traffic_rule_filter/src/filters/traffic_light_filter.cpp
```diff
@@ -37,62 +42,49 @@ TrafficLightFilter::TrafficLightFilter() : TrafficRuleFilterInterface("TrafficLi
...
-          }
-        }
+  lanelet::BasicLineString2d trajectory_ls;
+  for (const auto & p : trajectory_points) {
+    trajectory_ls.emplace_back(p.pose.position.x, p.pose.position.y);
+  }
+  const auto bbox = boost::geometry::return_envelope<lanelet::BoundingBox2d>(trajectory_ls);
+  for (const auto & ll : lanelet_map_->laneletLayer.search(bbox)) {
+    for (const auto & stop_line : get_red_stop_lines(ll)) {
+      if (boost::geometry::intersects(trajectory_ls, stop_line)) {
```
go-sakayori, line 83: Just to confirm, this mean we’re not allowed to cross the stop line even by a mm? I’m concerned this could be a pretty strict requirement for a learning-based planner.

yukkysaito:
Can you record a video in PSim showing that this feature is successfully filtering things as intended?

planning/autoware_trajectory_traffic_rule_filter/src/filters/traffic_light_filter.cpp
maxime-clem: It is my understanding of the current requirement for this feature.
The assumption is that if the Diffusion planner trajectory is rejected, then the Minimal rule-based planner trajectory will be used.

maxime-clem: Additionally, I did not take the front overhang into consideration yet, only the trajectory points, so there is quite a big margin at the moment.

maxime-clem:
> Can you record a video in PSim showing that this feature is successfully filtering things as intended?

https://github.com/user-attachments/assets/5da9c424-250e-40d2-9846-549d27e49b15

- the traffic signals information is not sent to the Diffusion planner such that its trajectory crosses the stop line.
- there is no backup planner, so once the Diffusion planner trajectory is rejected, it just stops being updated.

planning/autoware_trajectory_traffic_rule_filter/src/filters/traffic_light_filter.cpp
```diff
@@ -70,29 +44,55 @@ void TrafficLightFilter::set_traffic_lights(
...
   if (!lanelet_map_ || !traffic_lights_ || trajectory_points.empty()) {
     return true;  // Allow if no data available
   }
 
-  const auto lanes = get_lanelets_from_trajectory(trajectory_points);
+  lanelet::BasicLineString2d trajectory_ls;
+  for (const auto & p : trajectory_points) {
+    trajectory_ls.emplace_back(p.pose.position.x, p.pose.position.y);
+  }
+  RCLCPP_WARN(
```
danielsanchezaran: Maybe this should be a debug msg? if not it would pollute the terminal

planning/autoware_trajectory_traffic_rule_filter/src/filters/traffic_light_filter.cpp
```diff
@@ -70,29 +44,55 @@ void TrafficLightFilter::set_traffic_lights(
...
+    rclcpp::get_logger("TrafficLightFilter"), "Front overhang = %2.2fm",
+    vehicle_info_ptr_->max_longitudinal_offset_m);
+  if (
```
danielsanchezaran: if vehicle_info_ptr_->max_longitudinal_offset_m is less than 0.0 I think we should just print a WARN message and return with true, same for the case when the pointer is null no?

planning/autoware_trajectory_traffic_rule_filter/src/filters/traffic_light_filter.cpp
```diff
@@ -70,29 +44,55 @@ void TrafficLightFilter::set_traffic_lights(
...
+    vehicle_info_ptr_ && vehicle_info_ptr_->max_longitudinal_offset_m > 0.0 &&
+    trajectory_ls.size() > 1) {
+    const lanelet::BasicSegment2d last_segment(
+      trajectory_ls[trajectory_ls.size() - 2], trajectory_ls.back());
+    const auto last_vector = last_segment.second - last_segment.first;
+    const auto last_length = boost::geometry::length(last_segment);
+    const auto ratio = (last_length + vehicle_info_ptr_->max_longitudinal_offset_m) / last_length;
+    lanelet::BasicPoint2d front_vehicle_point = last_segment.first + last_vector * ratio;
+    trajectory_ls.emplace_back(front_vehicle_point);
+  }
```
danielsanchezaran: As I understand, you are setting the last point to be equal to the last position, plus the max long offset, but it took me some time to understand, can you add a couple comments explaining this part?

planning/autoware_trajectory_traffic_rule_filter/src/filters/traffic_light_filter.cpp
maxime-clem: I removed it 857d9ab611

planning/autoware_trajectory_traffic_rule_filter/src/filters/traffic_light_filter.cpp
maxime-clem: I moved the nullptr check to the beginning of the function with the other checks. https://github.com/autowarefoundation/autoware_universe/commit/857d9ab611c3fd986217d71ab7a607cc2e957957
The case `vehicle_info_ptr_->max_longitudinal_offset_m <= 0.0`  does not cause any crash or bad behavior so it is fine to ignore I think.

planning/autoware_trajectory_traffic_rule_filter/src/filters/traffic_light_filter.cpp
maxime-clem: I added a comment https://github.com/autowarefoundation/autoware_universe/commit/857d9ab611c3fd986217d71ab7a607cc2e957957

PR #11988, opened 2026-01-28
fix(autoware_tensorrt_vad): remove unnecessary USE_SCOPED_HEADER_INST…

Max-Bin:
…ALL_DIR

The USE_SCOPED_HEADER_INSTALL_DIR flag is unnecessary for this package because:
- The package has no include/ directory (no public headers)
- All headers are private in src/ directory
- This flag only affects public header installation paths
- It can cause build failures in certain configurations

This flag was the only usage in autoware_universe and provided no benefit.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


yukkysaito (approved):
LGTM

PR #11989, opened 2026-01-28
feat(autoware_traffic_light_multi_camera_fusion): use valid color and use whole elements as state

MasatoSaeki:
## Description

This PR contains following things;
- use valid color (before; if exist only unknown in roi, compare the some valid color about confidence, actually want to use valid color without comparing the confidence)
- use whole elements as state (before; compare the each one of elements. elements have the meaning if check overall it-self.)

TODO; Refactor and add some unit tests

## Related links

https://github.com/autowarefoundation/autoware_universe/pull/11297

**Parent Issue:**

- Link

## How was this PR tested?

Tested on AWSIM.

https://evaluation.ci.tier4.jp/evaluation/reports/1542081e-310a-532e-b5e1-39225463633a?project_id=prd_jt

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -15,7 +15,7 @@ common/autoware_polar_grid/** yukihiro.saito@tier4.jp
 common/autoware_time_utils/** christopherj.ho@gmail.com shumpei.wakabayashi@tier4.jp tomoya.kimura@tier4.jp
 common/autoware_traffic_light_recognition_marker_publisher/** shumpei.wakabayashi@tier4.jp takeshi.miura@tier4.jp tomoya.kimura@tier4.jp
 common/autoware_traffic_light_utils/** kotaro.uetake@tier4.jp masato.saeki@tier4.jp satoshi.ota@tier4.jp yoshi.ri@tier4.jp
-common/autoware_universe_utils/** mamoru.sobue@tier4.jp takamasa.horibe@tier4.jp takayuki.murooka@tier4.jp
+common/autoware_universe_utils/** alqudah.mohammad@tier4.jp junya.sasaki@tier4.jp mamoru.sobue@tier4.jp taiki.yamada@tier4.jp takamasa.horibe@tier4.jp takayuki.murooka@tier4.jp
 common/tier4_api_utils/** isamu.takagi@tier4.jp
 control/autoware_autonomous_emergency_braking/** alqudah.mohammad@tier4.jp daniel.sanchez@tier4.jp kyoichi.sugahara@tier4.jp mamoru.sobue@tier4.jp takamasa.horibe@tier4.jp tomoya.kimura@tier4.jp
 control/autoware_collision_detector/** go.sakayori@tier4.jp kyoichi.sugahara@tier4.jp tomohito.ando@tier4.jp
@@ -33,6 +33,7 @@ control/autoware_predicted_path_checker/** berkay@leodrive.ai junya.sasaki@tier4
 control/autoware_pure_pursuit/** takamasa.horibe@tier4.jp takayuki.murooka@tier4.jp
 control/autoware_shift_decider/** takamasa.horibe@tier4.jp takayuki.murooka@tier4.jp
 control/autoware_smart_mpc_trajectory_follower/** kosuke.takeuchi@tier4.jp masayuki.aino@proxima-ai-tech.com takamasa.horibe@tier4.jp takayuki.murooka@tier4.jp
+control/autoware_spheric_collision_detector/** simon.ranjith@lx.it.pt
 control/autoware_stop_mode_operator/** isamu.takagi@tier4.jp
 control/autoware_trajectory_follower_base/** takamasa.horibe@tier4.jp takayuki.murooka@tier4.jp
 control/autoware_trajectory_follower_node/** takamasa.horibe@tier4.jp takayuki.murooka@tier4.jp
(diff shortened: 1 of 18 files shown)


perception/autoware_traffic_light_multi_camera_fusion/src/traffic_light_multi_camera_fusion_node.hpp
```diff
@@ -45,7 +45,28 @@ namespace autoware::traffic_light
...
 namespace mf = message_filters;
 
-using StateKey = std::pair<uint8_t, uint8_t>;
+using StateKey = std::vector<std::pair<
+  tier4_perception_msgs::msg::TrafficLightElement::_color_type,
+  tier4_perception_msgs::msg::TrafficLightElement::_shape_type>>;
+
+inline bool isUnknown(const StateKey & state_key)
+{
+  return state_key.size() == 1 &&
```
YoshiRi, line 54: Do we really want to keep `unknown + right arrow` case as it is within this function?

perception/autoware_traffic_light_multi_camera_fusion/src/traffic_light_multi_camera_fusion_node.hpp
```diff
@@ -45,7 +45,28 @@ namespace autoware::traffic_light
...
+         state_key[0].first == tier4_perception_msgs::msg::TrafficLightElement::UNKNOWN;
+}
+
+inline bool compareStateKeyLogOdds(
+  const std::pair<StateKey, double> & key1, const std::pair<StateKey, double> & key2)
+{
+  // True is key1 has lower log-odds than key2, False otherwise
```
YoshiRi: I think following comment would be better

```cpp
// Ordering rule:
// 1. Unknown StateKey is always lower priority
// 2. Otherwise, smaller log-odds comes first
```

perception/autoware_traffic_light_multi_camera_fusion/src/traffic_light_multi_camera_fusion_node.hpp
```diff
@@ -45,7 +45,28 @@ namespace autoware::traffic_light
...
+  const bool key1_is_unknown = isUnknown(key1.first);
+  const bool key2_is_unknown = isUnknown(key2.first);
+  // If either is unknown, the unknown is of lower priority
+  if (key1_is_unknown != key2_is_unknown) {
+    return key1_is_unknown;
+  }
```
YoshiRi: Hard to read.  I prefer straightforward way

```
if (key1_is_unknown && !key2_is_unknown) {
  return true;
}
if (!key1_is_unknown && key2_is_unknown) {
  return false;
}
```

perception/autoware_traffic_light_multi_camera_fusion/src/traffic_light_multi_camera_fusion_node.cpp
```diff
@@ -273,11 +271,13 @@ void MultiCameraFusion::processFusedRecord(
...
-  const tier4_perception_msgs::msg::TrafficLightElement & element,
   const utils::FusionRecord & record)
 {
-  const StateKey state_key = {element.color, element.shape};
-  const double confidence = element.confidence;
+  StateKey state_key;
+  for (const auto & element : record.signal.elements) {
+    state_key.emplace_back(std::make_pair(element.color, element.shape));
+  }
+  const double confidence = record.signal.elements[0].confidence;  // Same value for all elements
```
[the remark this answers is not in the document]
YoshiRi: @MasatoSaeki  Do you know if signals size should have valid number?
Hmm, I think this issue can be discussed in another PR.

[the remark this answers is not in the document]
YoshiRi: IMO we can skip this issue for now.

perception/autoware_traffic_light_multi_camera_fusion/src/traffic_light_multi_camera_fusion_node.hpp
a-maumau: Also adding some comment would be better to clarify that we only consider a prediction as unknown when it only contains unknown.

perception/autoware_traffic_light_multi_camera_fusion/src/traffic_light_multi_camera_fusion_node.cpp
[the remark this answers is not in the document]
MasatoSaeki: Actually, the length of signals is more than 0 (A value must exist.). (I want to use functions from a common module(traffic_light_utils) for operations on msg.)
- https://github.com/autowarefoundation/autoware_universe/tree/main/perception/autoware_traffic_light_fine_detector#inner-workings--algorithms
- https://github.com/autowarefoundation/autoware_universe/tree/main/perception/autoware_traffic_light_classifier#inner-workings--algorithms

perception/autoware_traffic_light_multi_camera_fusion/src/traffic_light_multi_camera_fusion_node.hpp
MasatoSaeki: `unknown` exists only as singular. (As mentioned above, I'd like to handle this using something like a common function going forward.)

MasatoSaeki: `elements` is the vector. But it as a whole has meaning, not each individual element.

MasatoSaeki: And Even if it happens to be empty, I believe it's designed not to throw an error.

perception/autoware_traffic_light_multi_camera_fusion/src/traffic_light_multi_camera_fusion_node.hpp
MasatoSaeki: Fixed in https://github.com/autowarefoundation/autoware_universe/pull/11989/commits/b926027eae1e3eb6c398ba8390a8dbb25b7c3f23.

perception/autoware_traffic_light_multi_camera_fusion/src/traffic_light_multi_camera_fusion_node.hpp
MasatoSaeki: Fixed in https://github.com/autowarefoundation/autoware_universe/pull/11989/commits/b926027eae1e3eb6c398ba8390a8dbb25b7c3f23.

perception/autoware_traffic_light_multi_camera_fusion/src/traffic_light_multi_camera_fusion_node.hpp
YoshiRi: Understood. I had some misunderstandings.

YoshiRi (approved):
LGTM

PR #11995, opened 2026-01-28
feat(dummy_perception publisher): use object z as baselink coordinates

yuki-takagi-66:
## Description
Currently, the object spawn position is fixed to the ego vehicle's height, and the Z position set in RViz is ignored.
This update changes the behavior to interpret the Z position set in RViz as the height relative to the base_link frame.
Note that this has no impact on behavior when the base_link relative height is not used.

## Related links
Should be merged after the launch PR is merged.
https://github.com/autowarefoundation/autoware_launch/pull/1739

## How was this PR tested?
Before
image

After
image

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/simulator/autoware_dummy_perception_publisher/src/node.cpp
+++ b/simulator/autoware_dummy_perception_publisher/src/node.cpp
@@ -336,7 +336,7 @@ void DummyPerceptionPublisherNode::objectCallback(
       try {
         ros_map2base_link = tf_buffer_.lookupTransform(
           "map", "base_link", rclcpp::Time(0), rclcpp::Duration::from_seconds(0.5));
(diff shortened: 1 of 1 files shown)


sasakisasaki (approved):
I understood the effect of changes and agree with merging this PR.

(Note to myself) It seems the `objectCallback` is called by every "input/object" subscription as [these lines](https://github.com/autowarefoundation/autoware_universe/pull/11995/changes#diff-576c9bb989d60f155db30babdba65da32cd4b27b10fa1772c64684f9e656d8c7R101-R103).

If this callback is called multiple times for the _**unique**_ object, the z-position would be wrongly accumulated by every call. But it seems this callback is called only once for each object.

PR #12000, opened 2026-01-29
feat(autoware_lidar_frnet): add multi-format input, ego crop box, sensor-specific config, docs

amadeuszsz:
## Description

Adds support for multiple point cloud input formats (`XYZIRCAEDT`, `XYZIRADRT`, `XYZIRC`, `XYZI`) with automatic format detection, and an optional ego crop box filter with a one-time static TF lookup. Introduces sensor-specific configs (OT128 / QT128) and parameter layout (filter, ego_crop_box, model path by sensor). Fixes output buffer sizing so seg/viz use the profile max instead of input count. Docs and param schemas updated for the new options and limits.

[Screencast from 03-02-2026 05:04:52 PM.webm](https://github.com/user-attachments/assets/04037230-6947-4b5c-a004-c1437193722f)

## Related links

**Parent Issue:**

- https://github.com/autowarefoundation/autoware_universe/issues/10481

## How was this PR tested?

1. Download artifact from [TIER IV INTERNAL LINK](https://drive.google.com/drive/folders/1KZDa85mz0jD54vNuxIHPznd4fMN5GJQb?usp=sharing) and place them in `autoware_data/lidar_frnet`.
2. Play any rosbag with point cloud data.
3. Run node for specific LiDAR, e.g.:
```shell
ros2 launch autoware_lidar_frnet lidar_frnet.launch.xml sensor_model:=ot128 input/pointcloud:=/sensing/lidar/front_upper/pointcloud_raw_ex
```
4. Check the preview of:
- `/lidar_frnet/visualization`
- `/lidar_frnet/filtered`

## Notes for reviewers

- Make sure you use appropriate model for given input point cloud (available options are `ot128` and `qt128`).
- ~~For internal TIER IV data, `front_lower` underperforms, I'm currently investigating training pipeline. Anyway, functional node can be reviewed and I will try to update artifacts as soon as problem is solved.~~ Solved, valid artifacts added in https://github.com/autowarefoundation/autoware/pull/6928

## Interface changes

None.

## Effects on system behavior

None.


perception/autoware_lidar_frnet/lib/lidar_frnet.cpp
```diff
@@ -110,12 +148,17 @@ bool LidarFRNet::process(
     return false;
   }
 
-  cuda_utils::clear_async(cloud_in_d_.get(), network_params_.num_points_profile.max, stream_);
+  const auto max_buffer_size = network_params_.num_points_profile.max * input_point_step_;
```
manato:
```suggestion
 const auto max_buffer_size = network_params_.num_points_profile.max * get_num_fields(input_format_);
```

[nit]: Because the 2nd argument of `cuda_utils::clear_async` expects "the number of elements" of the array rather than the total size in byte, I would suggest to multiplying the number of fields. (Though the code behavior would not change because cloud_in_d_ is the array of `uint_8`!)

perception/autoware_lidar_frnet/lib/lidar_frnet.cpp
```diff
@@ -162,8 +209,12 @@ bool LidarFRNet::preprocess(const uint32_t input_num_points)
...
   CHECK_CUDA_ERROR(cudaStreamSynchronize(stream_));
   CHECK_CUDA_ERROR(preprocess_ptr_->projectPoints_launch(
-    cloud_in_d_.get(), input_num_points, num_points_d_.get(), points_d_.get(), coors_d_.get(),
-    coors_keys_d_.get(), proj_idxs_d_.get(), proj_2d_d_.get()));
+    cloud_in_d_.get(), input_num_points, input_format_, num_points_d_.get(), points_d_.get(),
+    coors_d_.get(), coors_keys_d_.get(), proj_idxs_d_.get(), proj_2d_d_.get(),
+    cloud_compact_d_.get()));
+  CHECK_CUDA_ERROR(cudaStreamSynchronize(stream_));
+  CHECK_CUDA_ERROR(
+    cudaMemcpy(&num_points_raw_, num_points_d_.get(), sizeof(uint32_t), cudaMemcpyDeviceToHost));
```
manato:
```suggestion
  CHECK_CUDA_ERROR(
    cudaMemcpyAsync(&num_points_raw_, num_points_d_.get(), sizeof(uint32_t), cudaMemcpyDeviceToHost, stream_));
```

Because the execution order of the operations on the same stream is guaranteed in the operation-issued order, I think we can omit `cudaStreamSynchronize` here as long as the previous `projectPoints_launch` and the latter `cudaMemcpy` (, and `interpolatePoints_launch` as well) are issued on the same stream

perception/autoware_lidar_frnet/lib/lidar_frnet.cpp
```diff
@@ -172,6 +223,7 @@ bool LidarFRNet::preprocess(const uint32_t input_num_points)
   CHECK_CUDA_ERROR(cudaStreamSynchronize(stream_));
   CHECK_CUDA_ERROR(
     cudaMemcpy(&num_points, num_points_d_.get(), sizeof(uint32_t), cudaMemcpyDeviceToHost));
+  num_points_ = num_points;
```
manato: [nit]: it would be perhaps possible to copy the data on `num_points_d_.get()` to the `num_points_` directly(?)

amadeuszsz: Yeah, we can copy directly to `num_points_`.
https://github.com/autowarefoundation/autoware_universe/pull/12000/commits/ec676c5959ce9a1890a3d121648f0f6caf85a71e

perception/autoware_lidar_frnet/lib/lidar_frnet.cpp
amadeuszsz: `get_num_fields` returns number of fields regardless type, which would result with insufficient buffer size. I renamed variable in https://github.com/autowarefoundation/autoware_universe/pull/12000/commits/ec587bf23bc946b79fb216b2e25e327ccbf8398f to make it more readable as `cuda_utils::clear_async` argument is called `num_elem`.

perception/autoware_lidar_frnet/lib/lidar_frnet.cpp
```diff
@@ -162,31 +209,33 @@ bool LidarFRNet::preprocess(const uint32_t input_num_points)
...
+    &num_points_raw_, num_points_d_.get(), sizeof(uint32_t), cudaMemcpyDeviceToHost, stream_));
   CHECK_CUDA_ERROR(preprocess_ptr_->interpolatePoints_launch(
     proj_idxs_d_.get(), proj_2d_d_.get(), num_points_d_.get(), points_d_.get(), coors_d_.get(),
     coors_keys_d_.get()));
 
-  uint32_t num_points{};
   CHECK_CUDA_ERROR(cudaStreamSynchronize(stream_));
   CHECK_CUDA_ERROR(
-    cudaMemcpy(&num_points, num_points_d_.get(), sizeof(uint32_t), cudaMemcpyDeviceToHost));
+    cudaMemcpy(&num_points_, num_points_d_.get(), sizeof(uint32_t), cudaMemcpyDeviceToHost));
```
manato:
```suggestion
    cudaMemcpyAsync(&num_points_, num_points_d_.get(), sizeof(uint32_t), cudaMemcpyDeviceToHost, stream_));
    
    CHECK_CUDA_ERROR(cudaStreamSynchronize(stream_));
```

Sorry for pointing out a small thing again, but would you mind changing here to Async copy? `cudaMemcpy` triggers implicit synchronization using the default CUDA stream, which can make other CUDA-related processing stuck here (especially in cases where this node runs in a composable container).

Since you use the copy result on the CPU side very soon, you also need to add stream sync. (And you can omit the stream sync before this copy!)

amadeuszsz:
> LGTM! Thank you for handling my concerns!

@manato Thank you!
I will merge this PR as soon as the new artifacts for Autoware are ready.

PR #10541, opened 2025-04-23
feat(tier4_perception_launch): added bevfusion to the launchers

knzo25:
## Description
BEVFusion was added in https://github.com/autowarefoundation/autoware_universe/pull/10024
but was not added to the launchers. This PRs implements that integration

## Related links

**Parent Issue:**

- Link

## How was this PR tested?
Launched BEVFusion by setting the model to `bevfusion/bevfusion_lidar` instead of centerpoint

Related PR: https://github.com/autowarefoundation/autoware_launch/pull/1416

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

PR #12005, opened 2026-01-30
chore(tier4_perception_launch): make lidar_detection_model_name non-exhaustive

paulsohn:
## Description

* remove choices from lidar detection model name argument, make it non-exhaustive and expandable (type is still choice)
* remove duplicate definition of lidar detection model type and name, and update them 'let' instead of 'arg' (they are parsed internally, to prevent excessive DoF given to end user)

## Related links

* [TIER IV internal discussion](https://star4.slack.com/archives/CRUE57C30/p1769140996406339)
* This is a partial revert of #10541 .

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

* `lidar_detection_model_name` can accept custom model names.
* The new `perception.launch.xml` now fails if the `lidar_detection_model_type` and `lidar_detection_model_name` is (accidently) passed into. Pass `lidar_detection_model:=<lidar_detection_model_type>/<lidar_detection_model_name>` instead.

### DIFF
--- a/perception/autoware_traffic_light_multi_camera_fusion/src/traffic_light_multi_camera_fusion_node.cpp
+++ b/perception/autoware_traffic_light_multi_camera_fusion/src/traffic_light_multi_camera_fusion_node.cpp
@@ -23,6 +23,7 @@
 #include <memory>
 #include <string>
 #include <unordered_map>
+#include <utility>
 #include <vector>
 
 namespace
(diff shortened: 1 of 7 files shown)


paulsohn:
@YoshiRi Sorry, I found this involves more work: I would like to sort of revert all choices introduced in #10541 .

YoshiRi:
> @YoshiRi Sorry, I found this involves more work: I would like to sort of revert all choices introduced in #10541 .

make sense. SRY, I should point out related fields.

PR #12011, opened 2026-02-02
fix(start_planner): reuse previous stop pose to stop for dynamic objects

maxime-clem:
## Description

Reuse the previous stop pose in the `start_planner` when the path is unsafe against dynamic objects. This allows the stop pose to go behind the ego vehicle, allowing other systems to detect when the stop pose is ignored (e.g., in the case of control issues).

## Related links

**Private Links:**

- [TIER IV internal link](https://tier4.atlassian.net/browse/T4DEV-46481)

## How was this PR tested?

Evaluator: https://evaluation.tier4.jp/evaluation/reports/e79a85ed-37ad-5823-8024-f3ad83d93d6e?project_id=prd_jt

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

The stop pose generated by the `start_planner` to avoid dynamic objects is no longer updated and can now go behind the ego vehicle.

### DIFF
--- a/planning/behavior_path_planner/autoware_behavior_path_start_planner_module/src/start_planner_module.cpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_start_planner_module/src/start_planner_module.cpp
@@ -1022,6 +1022,10 @@ BehaviorModuleOutput StartPlannerModule::planWaitingApproval()
     previous_stop_pose_ = stop_pose_ = utils::insert_feasible_stop_point(
       stop_path, planner_data_, -parameters_->maximum_deceleration_for_stop,
       parameters_->maximum_jerk_for_stop, stop_reason);
+  } else {
+    const auto stop_length =
+      motion_utils::calcSignedArcLength(stop_path.points, 0UL, stop_pose_->pose.position);
+    utils::insertStopPoint(stop_length, stop_path);
   }
 
   const auto drivable_lanes = generateDrivableLanes(stop_path);


planning/behavior_path_planner/autoware_behavior_path_start_planner_module/src/start_planner_module.cpp
```diff
@@ -1172,12 +1184,12 @@ PathWithLaneId StartPlannerModule::getCurrentOutputPath()
...
   removeRTCStatus();
 
-  stop_pose_ = utils::insert_feasible_stop_point(
+  previous_stop_pose_ = stop_pose_ = utils::insert_feasible_stop_point(
     current_path, planner_data_, -parameters_->maximum_deceleration_for_stop,
     parameters_->maximum_jerk_for_stop, "unsafe against dynamic objects");
 
   if (stop_pose_) {
-    RCLCPP_DEBUG_THROTTLE(
+    RCLCPP_WARN_THROTTLE(
```
kyoichi-sugahara: @maxime-clem 
Is this change for remporal debug purpose?

maxime-clem: That was a mistake. It is now fixed.

planning/behavior_path_planner/autoware_behavior_path_start_planner_module/src/start_planner_module.cpp
```diff
@@ -799,7 +799,14 @@ bool StartPlannerModule::isExecutionReady() const
   }
 
   if (!is_safe) {
-    stop_pose_ = PoseWithDetail(planner_data_->self_odometry->pose.pose, stop_reason);
+    if (previous_stop_pose_) {
```
kyoichi-sugahara, line 802: If only the first if condition is met, it seems that the insertion of the stop point into stop_path (which was originally handled inside insert_feasible_stop_point) is skipped.

Therefore, when using `previous_stop_pose_` as the stop_pose, please use `insertStopPoint` to explicitly insert the stop point.

PR #12021, opened 2026-02-03
feat(autoware_trajectory_optimizer): velocity optimizer refactor

Owen-Liuyuxuan:
# Lateral Acceleration Limit Bug: Analysis & Solution

## Summary

A critical bug in the `TrajectoryVelocityOptimizer` caused **90% velocity drops** (~7.3 m/s → ~0.8 m/s) at trajectory boundaries when lateral acceleration limiting was enabled. The root cause was **orientation discontinuities** introduced by the `TrajectorySplineSmoother`. The solution evolved from a simple boundary-skipping fix to a complete **architecture refactor** with a new QP-based smoother that preserves the DP planner's acceleration profile.

---

## Part 1: Lateral Acceleration Limit Analysis

> [!NOTE] 
> 1. Some special process happens at the start/end of the input trajectory.
> 2. We disable spline smoother's function on adding original goal point

````mermaid
sequenceDiagram
    participant SP as SplineSmoother
    participant Traj as Trajectory
    participant VO as VelocityOptimizer
    participant LAL as limit_lateral_acceleration
    
    Note over SP: trajectory_spline_smoother_utils.cpp
    SP->>Traj: Copy first point directly<br/>output_points{traj_points.front()}
    SP->>Traj: Interpolate points [ds, 2ds, ..., L]<br/>with Akima spline
    
    alt Distance to last > 1cm
        SP->>Traj: Append original last point<br/>output_points.push_back(original_last)
        Note over Traj: ⚠️ Orientation discontinuity created
    end
    
    Traj->>VO: Smoothed trajectory with discontinuities
    VO->>LAL: Process pairs (i, i+1) for i ∈ [0, size-2]
    
    Note over LAL: At i = size-2 (e.g., 140)
    LAL->>LAL: Δθ = yaw[141] - yaw[140]<br/>⚠️ Large due to discontinuity
    LAL->>LAL: ω = Δθ / Δt → Very high
    LAL->>LAL: lat_accel = v × ω → Exceeds limit
    LAL->>Traj: Reduce velocity drastically ❌
````
---

### Problem with Direct Velocity Modification

> [!NOTE] 
> 1. The original JerkFilter treats the input trajectory speed as the maximum speed allowed.
> 2. The original JerkFilter will lose the information of the original acceleration.
> 3. If we use a significantly different acceleration, the diffusion planner will **overreact** and further decelerate.
> 4. Original Filter requires the end of the path as a stop point

````mermaid
graph TB
    subgraph "Old Architecture Problems"
        A[Input Trajectory] --> B[limit_lateral_acceleration<br/>Directly modifies velocity]
        B --> C[set_max_velocity<br/>Directly modifies velocity]
        C --> D[JerkFilteredSmoother<br/>Requires stop point]
        
        B -.->|Triggers| E[Acceleration recalculation]
        C -.->|Triggers| F[Acceleration recalculation]
        E -.-> G[Cascading Errors]
        F -.-> G
        
        D --> H[Output with<br/>discontinuities ❌]
        
        style G fill:#f99
        style H fill:#f99
    end
````

### New Architecture: Constraint-Based Approach

````mermaid
graph TB
    subgraph "New Architecture: Constraint Propagation"
        A[Input Trajectory<br/>v, a from DP planner] --> B[Initialize<br/>max_velocity_per_point = max_speed]
        
        B --> C[limit_lateral_acceleration<br/>Updates max_velocity_per_point ONLY]
        C --> D[set_max_velocity<br/>Updates max_velocity_per_point ONLY]
        
        D --> E[ContinuousJerkSmoother<br/>QP Optimization]
        
        A -.->|Reference acceleration| E
        C -.->|Velocity constraints| E
        D -.->|Velocity constraints| E
        
        E --> F[Output Trajectory<br/>Smooth, kinematic ✓]
        
        style C fill:#aaf
        style D fill:#aaf
        style E fill:#afa
        style F fill:#afa
    end
    
    subgraph "QP Optimization Details"
        E1[Objective:<br/>• Minimize jerk<br/>• Track acceleration<br/>• Penalize violations]
        E2[Constraints:<br/>• v² ≤ max_velocity_per_point²<br/>• a_min ≤ a ≤ a_max<br/>• j_min ≤ j ≤ j_max<br/>• v²' = 2a dynamics]
        
        E1 --> E
        E2 --> E
    end
````

### ContinuousJerkSmoother vs JerkFilteredSmoother

```mermaid
graph LR
    subgraph "JerkFilteredSmoother Old"
        A1[Requires stop point] --> B1[Search for v=0 index]
        B1 --> C1[Optimize up to stop]
        C1 --> D1[Set all points after<br/>stop to v=0]
        D1 --> E1[❌ Fails for continuous<br/>trajectories]
        
        style E1 fill:#f99
    end
    
    subgraph "ContinuousJerkSmoother New"
        A2[Works with/without<br/>stop point] --> B2{Has v=0 points?}
        B2 -->|Yes| C2[Optimize up to stop]
        B2 -->|No| D2[Optimize full trajectory]
        C2 --> E2[✓ Handles continuous<br/>trajectories]
        D2 --> E2
        
        style E2 fill:#afa
    end
```
---

### ContinuousJerkSmoother Class Structure

````mermaid
classDiagram
    class ContinuousJerkSmoother {
        -Params params_
        -std::shared_ptr~QPInterface~ qp_solver_
        -rclcpp::Logger logger_
        
        +ContinuousJerkSmoother(params, logger)
        +apply(input, max_velocity_per_point, v0, a0) TrajectoryPoints
        -optimize(N, ds, v0, a0, v_ref, a_ref, v_max) OptimizationResult
        -buildQPMatrices(N, ds, v0, a0, v_ref, a_ref, v_max) QPMatrices
        -extractTrajectory(input, result) TrajectoryPoints
    }
    
    class Params {
        +double jerk_weight
        +double velocity_tracking_weight
        +double accel_tracking_weight
        +double over_v_weight
        +double over_a_weight
        +double over_j_weight
        +double max_accel_mps2
        +double min_accel_mps2
        +double max_jerk_mps3
        +double min_jerk_mps3
    }
    
    class OptimizationResult {
        +std::vector~double~ velocities
        +std::vector~double~ accelerations
        +bool success
    }
    
    ContinuousJerkSmoother --> Params
    ContinuousJerkSmoother --> OptimizationResult
    ContinuousJerkSmoother --> QPInterface
````

---

## How was this PR tested?

PSim and perception reproducer test

## Notes for reviewers

None.

#### Additions and removals

| Change Type | Parameter Name | Type | Default Value | Description |
|-------------|----------------|------|---------------|-------------|
| Added | `publish_debug_trajectories` | bool | true | Enable publishing of debug trajectories for visualization and debugging |
| Added | `trajectory_velocity_optimizer.min_limited_speed_mps` | double | 4.0 | Minimum speed when applying lateral acceleration limit [m/s] |
| Added | `trajectory_velocity_optimizer.continuous_jerk_smoother.jerk_weight` | double | 10.0 | Weight for jerk minimization (smoothness) |
| Added | `trajectory_velocity_optimizer.continuous_jerk_smoother.over_v_weight` | double | 50000.0 | Weight for velocity limit violation (slack) |
| Added | `trajectory_velocity_optimizer.continuous_jerk_smoother.over_a_weight` | double | 3000.0 | Weight for acceleration limit violation (slack) |
| Added | `trajectory_velocity_optimizer.continuous_jerk_smoother.over_j_weight` | double | 10.0 | Weight for jerk limit violation (slack) |
| Added | `trajectory_velocity_optimizer.continuous_jerk_smoother.velocity_tracking_weight` | double | 0.0 | Weight for tracking reference velocity from input trajectory (disabled) |
| Added | `trajectory_velocity_optimizer.continuous_jerk_smoother.accel_tracking_weight` | double | 3000.0 | Weight for tracking reference acceleration from input trajectory |

## Effects on system behavior

None.

### DIFF
--- a/planning/.pages
+++ b/planning/.pages
@@ -85,6 +85,7 @@ nav:
       - 'Trajectory Optimizer': planning/autoware_trajectory_optimizer
       - 'Trajectory Ranker': planning/autoware_trajectory_ranker
       - 'Trajectory Safety Filter': planning/autoware_trajectory_safety_filter
+      - 'Continuous Jerk Filter': planning/autoware_trajectory_optimizer/docs/continuous_jerk_filter
 
   - 'Scenario Selector': planning/autoware_scenario_selector
 

### CODE REVIEW

File: planning/autoware_trajectory_optimizer/config/plugins/trajectory_velocity_optimizer.param.yaml [hunk 1]
```diff
@@ -6,17 +6,33 @@
       nearest_yaw_threshold_deg: 60.0  # [deg] Yaw threshold for trajectory matching
 
       # Engage behavior
-      target_pull_out_speed_mps: 1.0  # [m/s] Target speed during pull-out maneuver
-      target_pull_out_acc_mps2: 1.0   # [m/s²] Target acceleration during pull-out
```
danielsanchezaran, line 10: Why were these changed?

planning/autoware_trajectory_optimizer/config/plugins/trajectory_velocity_optimizer.param.yaml
```diff
@@ -6,17 +6,33 @@
...
+      target_pull_out_speed_mps: 0.5  # [m/s] Target speed during pull-out maneuver
+      target_pull_out_acc_mps2: 0.5   # [m/s²] Target acceleration during pull-out
       set_engage_speed: false         # Set minimum speed during engage
 
       # Speed limiting
-      max_speed_mps: 13.88             # [m/s] Maximum allowed speed (30 km/h default)
```
danielsanchezaran: Why is this value changed? I guess the comment should change to be 40 kmph instead?

planning/autoware_trajectory_optimizer/config/plugins/trajectory_velocity_optimizer.param.yaml
```diff
@@ -6,17 +6,33 @@
...
+      max_speed_mps: 11.11             # [m/s] Maximum allowed speed (30 km/h default)
       limit_speed: true               # Enable speed limiting
 
       # Lateral acceleration limiting
-      max_lateral_accel_mps2: 1.5     # [m/s²] Maximum lateral acceleration
-      limit_lateral_acceleration: false  # Enable lateral acceleration limiting
+      max_lateral_accel_mps2: 1.0     # [m/s²] Maximum lateral acceleration
+      limit_lateral_acceleration: true  # Enable lateral acceleration limiting
```
danielsanchezaran:
```suggestion
      limit_lateral_acceleration: false  # Enable lateral acceleration limiting
```
This feature is not used by default due to lack of testing. This should be looked as a safety measure and not a way to solve Model output issues.

planning/autoware_trajectory_optimizer/config/plugins/trajectory_velocity_optimizer.param.yaml
```diff
@@ -6,17 +6,33 @@
...
+      min_limited_speed_mps: 4.0        # [m/s] Minimum speed when applying lateral acceleration limit
```
danielsanchezaran: wouldnt this force accelerations when the ego is moving at low speeds. This should not be necessary when we have a lateral acceleration_mps2 param

planning/autoware_trajectory_optimizer/config/plugins/trajectory_velocity_optimizer.param.yaml
```diff
@@ -6,17 +6,33 @@
...
+      skip_lateral_accel_limit_head_points: 2  # Number of head points to skip (handles orientation discontinuities at start)
```
danielsanchezaran: If there are issues due to point yaw discontinuities, then that should be fixed, skipping points makes little sense, specially the first points

planning/autoware_trajectory_optimizer/config/plugins/trajectory_velocity_optimizer.param.yaml
```diff
@@ -6,17 +6,33 @@
...
+      skip_lateral_accel_limit_tail_points: 2  # Number of tail points to skip (handles orientation discontinuities at end)
 
       # Velocity smoothing
-      smooth_velocities: false        # Enable jerk-filtered velocity smoothing
+      smooth_velocities: true        # Enable jerk-filtered velocity smoothing
```
danielsanchezaran:
```suggestion
      smooth_velocities: false       # Enable jerk-filtered velocity smoothing
```

planning/autoware_trajectory_optimizer/include/autoware/trajectory_optimizer/trajectory_optimizer_plugins/plugin_utils/continuous_jerk_smoother.hpp
```diff
@@ -0,0 +1,126 @@
...
+   */
+  bool apply(
+    const double v0, const double a0, const TrajectoryPoints & input, TrajectoryPoints & output,
+    const std::vector<double> & max_velocity_per_point = {});
+
+  /**
+   * @brief Set smoother parameters
+   * @param params New parameters
+   */
+  void setParams(const ContinuousJerkSmootherParams & params);
```
danielsanchezaran: Please use snake_case 
```suggestion
  void set_params(const ContinuousJerkSmootherParams & params);
```

planning/autoware_trajectory_optimizer/include/autoware/trajectory_optimizer/trajectory_optimizer_plugins/plugin_utils/continuous_jerk_smoother.hpp
```diff
@@ -0,0 +1,126 @@
...
+
+  /**
+   * @brief Get current smoother parameters
+   * @return Current parameters
+   */
+  ContinuousJerkSmootherParams getParams() const;
```
danielsanchezaran: snake_case

planning/autoware_trajectory_optimizer/config/plugins/trajectory_velocity_optimizer.param.yaml
Owen-Liuyuxuan: If smooth_velocity & lateral limit are both false, we do not need much update, and the current code is nice enough.

planning/autoware_trajectory_optimizer/config/plugins/trajectory_velocity_optimizer.param.yaml
Owen-Liuyuxuan: `min_limited_speed_mps: 4.0.` This essentially means that I do not want to over-supress the turning speed. when it has been slow enough.

planning/autoware_trajectory_optimizer/config/plugins/trajectory_velocity_optimizer.param.yaml
Owen-Liuyuxuan: It is a test case problem; sometimes, the velocity smoother will drive the vehicle to drive too fast. the original speed limit should be correct overall.

planning/autoware_trajectory_optimizer/config/plugins/trajectory_velocity_optimizer.param.yaml
Owen-Liuyuxuan: This will destroy the smoothing result with velocity smoothing. If we insist the first point to have large accleration and speed, and we should also keep acceleration later on, then the velocity smoother will produce a large acceleration/speed at the first point, then sudden de-celeration on the following points -- which fed to the delay-sensitive PID and result in the vehicle keep stopping.

planning/autoware_trajectory_optimizer/config/plugins/trajectory_velocity_optimizer.param.yaml
danielsanchezaran: Both parameters are false by default, as I mentioned previously I was planning on removing the velocity smoothing and the lateral acceleration limit, but if this PR makes their behavior better, I would like to enable it. However, we cannot enable these features without testing in real vehicles first, so they should be false

Owen-Liuyuxuan: @danielsanchezaran You are right. Sorry I am busy on the refactoring works on release works so it will take some times before I make further improvements.

planning/autoware_trajectory_optimizer/config/plugins/trajectory_velocity_optimizer.param.yaml
Owen-Liuyuxuan: @danielsanchezaran After my debug, I have confidence that the value here can be restored.

planning/autoware_trajectory_optimizer/config/plugins/trajectory_velocity_optimizer.param.yaml
Owen-Liuyuxuan: @danielsanchezaran Thank you for your suggestion, I will start woking on the lateral acc part after I fixed the problem in optimizer.

Owen-Liuyuxuan:
@danielsanchezaran 
Thanks for you review!.

1. I resolved some of your review concerns related to the optimizer and fixed a structural bug in my optimization setting. Now the optimizer can basically safely run in my PSim system.
   - I have also carried out a sanity check, which with all other optimization parameters weights fixed to 0, the result can successfully tract the input velocity or acceleration.
3. I manually resolved the conflict with your recent updates in the commit https://github.com/autowarefoundation/autoware_universe/pull/12021/commits/e4b31ddcaf26c4ad56f3cf5aed32928cc9058486. Please also help checking that.
4. I will try to investigate the discontinuous points problem further after some time.

planning/autoware_trajectory_optimizer/src/trajectory_optimizer_plugins/plugin_utils/continuous_jerk_smoother.cpp
```diff
@@ -175,24 +177,30 @@ bool ContinuousJerkSmoother::apply(
...
 
   // Reference acceleration tracking: minimize weight * (a - a_ref)^2
   // Expands to: weight * a^2 - 2*weight*a_ref*a + const
   // P term: weight on a^2
   // q term: -2*weight*a_ref on a
-  const double accel_tracking_weight = params_.accel_tracking_weight;
   for (size_t i = 0; i < N; ++i) {
     const double a_ref = input.at(i).acceleration_mps2;
     P(IDX_A0 + i, IDX_A0 + i) += accel_tracking_weight;
-    q.at(IDX_A0 + i) += -2.0 * accel_tracking_weight * a_ref;
```
Owen-Liuyuxuan: This "2.0" is the structural bugs I have created before.
The QP definition implicitly add the factor 0.5 to the P matrix

planning/autoware_trajectory_optimizer/src/trajectory_optimizer_plugins/plugin_utils/trajectory_velocity_optimizer_utils.cpp
```diff
@@ -163,11 +162,36 @@ void limit_lateral_acceleration(
 
   const auto closest_index =
     motion_utils::findNearestIndex(input_trajectory_array, current_position);
+
+  // Calculate start and end indices, excluding head and tail points
+  const size_t effective_skip_head = static_cast<size_t>(std::max(0, skip_head_points));
+  const size_t effective_skip_tail = static_cast<size_t>(std::max(0, skip_tail_points));
```
danielsanchezaran: Please verify what is the issue with the trajectory points, I think it is not correct to skip points but to fix the big yaw changes. I have not been able to see the big yaw discontinuities you mentioned in my tests, so I would like to know where is the bug. It should be located in the "first point" and "last_point" insertion in the splinesmootheroptimizer by your description. If that is that case, I would recommend 1: starting the loop that creates the spline trajectory with s = 0.0 instead of s = ds and 2) checking if there is a big yaw discontinuity between the last point of the smoothed trajectory and the last point of the original trajectory and then normalizing the last point of the original trajectory before inserting it to the smoothed trajectory vector.

planning/autoware_trajectory_optimizer/src/trajectory_optimizer.cpp
```diff
@@ -184,8 +204,16 @@ void TrajectoryOptimizer::on_traj([[maybe_unused]] const CandidateTrajectories::
...
-      plugin->optimize_trajectory(trajectory.points, params_, data);
+    for (size_t i = 0; i < plugins_.size(); ++i) {
+      plugins_[i]->optimize_trajectory(trajectory.points, params_, data);
+
+      // Publish debug trajectory after each plugin (only for first candidate)
+      if (i < debug_trajectory_pubs_.size() && debug_trajectory_pubs_[i]) {
+        Trajectory debug_traj;
+        debug_traj.header = trajectory.header;
+        debug_traj.points = trajectory.points;
+        debug_trajectory_pubs_[i]->publish(debug_traj);
```
danielsanchezaran: no need to copy the trajectory points, you can just publish them, also please add a flag to publish or not publish the debug trajs.

Something like

`if (!publish_debug_trajectories) {continue;}
 if (i < debug_trajectory_pubs_.size() && debug_trajectory_pubs_[i]) {
debug_trajectory_pubs_[i]->publish(trajectory); } 
`

planning/autoware_trajectory_optimizer/src/utils.cpp
```diff
@@ -124,4 +128,47 @@ void copy_trajectory_orientation(
   }
 }
 
+void debug_log_trajectory_tail_orientations(
```
danielsanchezaran: This function seems to be unused?

planning/autoware_trajectory_optimizer/include/autoware/trajectory_optimizer/utils.hpp
```diff
@@ -105,6 +104,20 @@ void recalculate_longitudinal_acceleration(
...
+ * This function logs the yaw angle of each of the last N points and the delta_theta
+ * (orientation change) between consecutive points. Useful for debugging orientation
+ * discontinuities at the end of trajectories.
+ *
+ * @param traj The trajectory points to analyze.
+ * @param stage_name A descriptive name for the pipeline stage (e.g., "AFTER_SPLINE_SMOOTHER").
+ * @param num_tail_points Number of tail points to log (default: 5).
+ */
+void debug_log_trajectory_tail_orientations(
+  const TrajectoryPoints & traj, const std::string & stage_name, size_t num_tail_points = 5);
```
danielsanchezaran: Unused function?

planning/autoware_trajectory_optimizer/src/trajectory_optimizer_plugins/plugin_utils/trajectory_velocity_optimizer_utils.cpp
```diff
@@ -145,13 +145,12 @@ void set_max_velocity(TrajectoryPoints & input_trajectory_array, const float max
...
-  TrajectoryPoints & input_trajectory_array, const double max_lateral_accel_mps2,
-  const Odometry & current_odometry)
+std::vector<double> limit_lateral_acceleration(
+  TrajectoryPoints & input_trajectory_array, std::vector<double> & max_velocity_per_point,
+  const double max_lateral_accel_mps2, const double min_limited_speed_mps,
+  const Odometry & current_odometry, const int skip_head_points, const int skip_tail_points)
 {
-  if (input_trajectory_array.empty()) {
-    return;
-  }
```
danielsanchezaran, line 154: why remove this early return? if the trajectory is empty we can safely return

  if (input_trajectory_array.empty()) {
    return;
  }

planning/autoware_trajectory_optimizer/src/trajectory_optimizer_plugins/plugin_utils/trajectory_velocity_optimizer_utils.cpp
```diff
@@ -189,29 +213,40 @@ void limit_lateral_acceleration(
...
+    const double limited_velocity =
+      std::max(max_lateral_accel_mps2 / yaw_rate, min_limited_speed_mps);
 
-  // recalculate acceleration after velocity change
-  autoware::trajectory_optimizer::utils::recalculate_longitudinal_acceleration(
-    input_trajectory_array);
+    if (current_speed < limited_velocity) {
+      // No modification needed
+      continue;
+    }
```
danielsanchezaran: You can move this check right after 

`  const double current_speed = std::abs(itr->longitudinal_velocity_mps);`

if the velocity is too low we dont need to check for lateral acceleration values

planning/autoware_trajectory_optimizer/src/trajectory_optimizer.cpp
```diff
@@ -66,9 +68,27 @@ void TrajectoryOptimizer::initialize_optimizers()
...
+    }
+
+    // Create debug publisher with format: ~/debug/{index:02d}_{name}/trajectory
+    std::ostringstream topic_name;
+    topic_name << "~/debug/" << short_name << "_" << std::setfill('0') << std::setw(2) << i
+               << "/trajectory";
+    auto pub = create_publisher<Trajectory>(topic_name.str(), 1);
+    debug_trajectory_pubs_.push_back(pub);
+
+    RCLCPP_INFO_STREAM(get_logger(), "Created debug publisher: " << topic_name.str());
```
danielsanchezaran: Please make this part a member method of the class, something like "void init_debug_publishers()"

planning/autoware_trajectory_optimizer/src/trajectory_optimizer.cpp
```diff
@@ -184,8 +204,16 @@ void TrajectoryOptimizer::on_traj([[maybe_unused]] const CandidateTrajectories::
   CandidateTrajectories output_trajectories = *msg;
   for (auto & trajectory : output_trajectories.candidate_trajectories) {
     // Apply optimizations - plugins execute in order from plugin_names parameter
-    for (auto & plugin : plugins_) {
-      plugin->optimize_trajectory(trajectory.points, params_, data);
+    for (size_t i = 0; i < plugins_.size(); ++i) {
+      plugins_[i]->optimize_trajectory(trajectory.points, params_, data);
+
+      // Publish debug trajectory after each plugin (only for first candidate)
+      if (i < debug_trajectory_pubs_.size() && debug_trajectory_pubs_[i]) {
```
danielsanchezaran: This is a weird check as we create a publisher per plugin, there is no other way around it, I think it is not needed.

planning/autoware_trajectory_optimizer/config/plugins/trajectory_velocity_optimizer.param.yaml
danielsanchezaran: I agree, but this functionality is currently not used, I think it is not necessary to change it, I will make a PR to remove it in the near future

planning/autoware_trajectory_optimizer/src/trajectory_optimizer_plugins/plugin_utils/trajectory_velocity_optimizer_utils.cpp
Owen-Liuyuxuan: Yes. After I correctly handle the trajectory optimizer, the front points are fine, but the last point sometimes significantly drops slow.

I will check your "second" suggestion.

planning/autoware_trajectory_optimizer/src/utils.cpp
Owen-Liuyuxuan: @danielsanchezaran 
Yes.. This is the debugging function I was using. Very sorry.

planning/autoware_trajectory_optimizer/src/trajectory_optimizer.cpp
Owen-Liuyuxuan: @danielsanchezaran  The trajectory is a "CandidateTrajectories" message, instead of the Trajectory type, for debugging should we using the "Trajectory"?

planning/autoware_trajectory_optimizer/config/plugins/trajectory_velocity_optimizer.param.yaml
Owen-Liuyuxuan: @danielsanchezaran 
Unfortunately, at the moment, these two parameters determine the `initial_motion` for the velocity smoother. However, after fixing the bug in the velocity smoother, it has become robust enough to handle these parameters correctly. Therefore, I will revert this change.

Owen-Liuyuxuan: @danielsanchezaran  Fixed in https://github.com/autowarefoundation/autoware_universe/pull/12021/changes/dc575c24656a41dca69bd4e6bfa6f8e0e559c486

planning/autoware_trajectory_optimizer/include/autoware/trajectory_optimizer/trajectory_optimizer_plugins/plugin_utils/continuous_jerk_smoother.hpp
Owen-Liuyuxuan: @danielsanchezaran  This already resolved in https://github.com/autowarefoundation/autoware_universe/pull/12021/changes/0d2cad94eeb1842e735fb51b3581bbfd7041e761

planning/autoware_trajectory_optimizer/include/autoware/trajectory_optimizer/trajectory_optimizer_plugins/plugin_utils/continuous_jerk_smoother.hpp
Owen-Liuyuxuan: https://github.com/autowarefoundation/autoware_universe/pull/12021/changes/0d2cad94eeb1842e735fb51b3581bbfd7041e761

planning/autoware_trajectory_optimizer/config/plugins/trajectory_velocity_optimizer.param.yaml
Owen-Liuyuxuan: @danielsanchezaran  This is fixed in https://github.com/autowarefoundation/autoware_universe/pull/12021/changes/4f6a677ef3bfb3a0c102712889b9b75e436fe442

&&

https://github.com/autowarefoundation/autoware_universe/pull/12021/changes/3ea7e677df268c8c1190c322e3141b80ae4ff6bc

As long as the spline smoother does not add the final points, the result will be nice.

planning/autoware_trajectory_optimizer/src/trajectory_optimizer_plugins/plugin_utils/trajectory_velocity_optimizer_utils.cpp
Owen-Liuyuxuan: @danielsanchezaran 

https://github.com/autowarefoundation/autoware_universe/pull/12021/changes/4f6a677ef3bfb3a0c102712889b9b75e436fe442

https://github.com/autowarefoundation/autoware_universe/pull/12021/changes/3ea7e677df268c8c1190c322e3141b80ae4ff6bc

As long as the spline smoother does not add the final points, the result will be nice. So I delete the logic of adding final points, then I can also clean up the skip point logics.

planning/autoware_trajectory_optimizer/src/trajectory_optimizer.cpp
Owen-Liuyuxuan: @danielsanchezaran 

https://github.com/autowarefoundation/autoware_universe/pull/12021/changes/665bec1038c9534db87a1d7e8a813661a6c7cc4a

I added a parameter to control the debug trajectories. But it seems a class change is still needed.

planning/autoware_trajectory_optimizer/include/autoware/trajectory_optimizer/utils.hpp
Owen-Liuyuxuan: @danielsanchezaran 
Very sorry they are cleared here: 

https://github.com/autowarefoundation/autoware_universe/pull/12021/changes/bc081f6c8b1ca1b251868f6f963e58751ff13c85

planning/autoware_trajectory_optimizer/src/trajectory_optimizer_plugins/plugin_utils/trajectory_velocity_optimizer_utils.cpp
Owen-Liuyuxuan: https://github.com/autowarefoundation/autoware_universe/pull/12021/changes/4f6a677ef3bfb3a0c102712889b9b75e436fe442

In this update, I added back the empty return to return the `max_velocity_per_point`. At empty case, this array will be initialized as a vector of size 0, and the code will not die here.

planning/autoware_trajectory_optimizer/src/trajectory_optimizer_plugins/plugin_utils/trajectory_velocity_optimizer_utils.cpp
Owen-Liuyuxuan: https://github.com/autowarefoundation/autoware_universe/pull/12021/changes/4f6a677ef3bfb3a0c102712889b9b75e436fe442

I clean up the logic. 

- **The current_index check is not needed because we do not skip tail.**
- if the current speed is low, it will shortcut the check.
- if the current speed is high, we will try to clamp it to `std::max(max_lateral_accel_mps2 / yaw_rate, min_limited_speed_mps);` if `lateral_acceleration` is high

image

planning/autoware_trajectory_optimizer/src/trajectory_optimizer.cpp
Owen-Liuyuxuan: @danielsanchezaran 

https://github.com/autowarefoundation/autoware_universe/pull/12021/changes/665bec1038c9534db87a1d7e8a813661a6c7cc4a

In this commit, I split the debug publisher into a member method so that `initialize_optimizers` will be simpler.

planning/autoware_trajectory_optimizer/src/trajectory_optimizer.cpp
Owen-Liuyuxuan: @danielsanchezaran 

https://github.com/autowarefoundation/autoware_universe/pull/12021/changes/665bec1038c9534db87a1d7e8a813661a6c7cc4a

I cleaned up this check and we only need to check the debugging parameter.

Owen-Liuyuxuan:
@danielsanchezaran 

1. I noticed that by not letting spline_smoother append the last point, we can make the input to `limit_lateral_accleration` smooth.
2. I tried to merge some coding conflicts between the "max_velocity_mps" setting.  https://github.com/autowarefoundation/autoware_universe/pull/12021/changes/4f6a677ef3bfb3a0c102712889b9b75e436fe442

danielsanchezaran:
> @danielsanchezaran
> 
> 1. I noticed that by not letting spline_smoother append the last point, we can make the input to `limit_lateral_accleration` smooth.
> 2. I tried to merge some coding conflicts between the "max_velocity_mps" setting.  [4f6a677](https://github.com/autowarefoundation/autoware_universe/commit/4f6a677ef3bfb3a0c102712889b9b75e436fe442)

@Owen-Liuyuxuan instead of not appending the last point of the trajectory, can you try this PR: https://github.com/autowarefoundation/autoware_universe/pull/12043 ? I think it should solve the yaw discontinuity issue without sacrificing the first or last point

Owen-Liuyuxuan:
> @Owen-Liuyuxuan instead of not appending the last point of the trajectory, can you try this PR: #12043 ? I think it should solve the yaw discontinuity issue without sacrificing the first or last point

@danielsanchezaran Great, I tried in bus PSim and see no degrade on basic rides.

I will wait for your update to merge

danielsanchezaran:
> > @Owen-Liuyuxuan instead of not appending the last point of the trajectory, can you try this PR: #12043 ? I think it should solve the yaw discontinuity issue without sacrificing the first or last point
> 
> @danielsanchezaran Great, I tried in bus PSim and see no degrade on basic rides.

Thank you, I will get that PR merged and then you can rebase to have it too. I think this PR is almost done too, I will need to do some extra checks on simulation but in general LGTM

planning/autoware_trajectory_optimizer/config/plugins/trajectory_velocity_optimizer.param.yaml
```diff
@@ -14,8 +14,22 @@
       limit_speed: true               # Enable speed limiting
 
       # Lateral acceleration limiting
-      max_lateral_accel_mps2: 1.5     # [m/s²] Maximum lateral acceleration
-      limit_lateral_acceleration: false  # Enable lateral acceleration limiting
+      max_lateral_accel_mps2: 1.0     # [m/s²] Maximum lateral acceleration
```
danielsanchezaran: In my simulation tests this value seems quite low, I think 1.5 or even 2.0 is better for a normal passenger vehicle I believe

planning/autoware_trajectory_optimizer/config/trajectory_optimizer.param.yaml
```diff
@@ -21,3 +21,4 @@
     use_trajectory_extender: false
     use_kinematic_feasibility_enforcer: true
     use_mpt_optimizer: false
+    publish_debug_trajectories: true
```
danielsanchezaran: please set false as default

planning/autoware_trajectory_optimizer/include/autoware/trajectory_optimizer/utils.hpp
```diff
@@ -33,6 +32,7 @@
 #include <nav_msgs/msg/detail/odometry__struct.hpp>
 
 #include <memory>
+#include <string>
```
danielsanchezaran: do we need this include?

planning/autoware_trajectory_optimizer/src/trajectory_optimizer_plugins/plugin_utils/trajectory_velocity_optimizer_utils.cpp
danielsanchezaran:
```suggestion
const double yaw_current = autoware_utils_geometry::get_rpy(current_pose).z;
    const double yaw_next = autoware_utils_geometry::get_rpy(next_pose).z;
    const double delta_yaw = autoware_utils_geometry::normalize_radian(yaw_next - yaw_current);

    constexpr double epsilon_yaw_rate = 1.0e-5;
    const double yaw_rate = std::max(std::abs(delta_yaw / delta_time), epsilon_yaw_rate);
    const double current_speed = std::abs(itr->longitudinal_velocity_mps);
    const double lateral_acceleration = current_speed * yaw_rate;

```
This might be more accurate and easier to read and uses autoware's inner tools for consistency

planning/autoware_trajectory_optimizer/src/trajectory_optimizer.cpp
```diff
@@ -184,8 +211,15 @@ void TrajectoryOptimizer::on_traj([[maybe_unused]] const CandidateTrajectories::
...
-    for (auto & plugin : plugins_) {
-      plugin->optimize_trajectory(trajectory.points, params_, data);
+    for (size_t i = 0; i < plugins_.size(); ++i) {
+      plugins_[i]->optimize_trajectory(trajectory.points, params_, data);
+
+      if (params_.publish_debug_trajectories) {
+        Trajectory debug_traj;
+        debug_traj.header = trajectory.header;
+        debug_traj.points = trajectory.points;
+        debug_trajectory_pubs_[i]->publish(debug_traj);
```
danielsanchezaran: Hmm the issue is that this code only really works well if we have one input candidate trajectory, if we have more than 1, we will be overwriting the debug trajectory of the previous trajectories. We would need a publisher per plugin and per candidate trajectory, I recommend you remove the debugging part of this PR and we think about it for a separate PR

danielsanchezaran (commented):
I have made some more comments, in particular, I think the debug publishers will need some extra though as they dont really support the case when candidate_trajectories.trajectories.size() > 1.  Maybe we should remove the debug publishers for now and implement them on a different PR?

Also if I understand correctly, if the parameter smooth_velocities is false and limit_lateral_acceleration is false, then the ego vehicle behavior is unchanged when compared to current OSS Autoware right? (this is what I want, as I want to further test the performance on a real vehicle before enabling these features). 

Finally, I have tested the smoothing feature with lateral acceleration limit, it seems to work quite well and I could not find issues in simulation:

- Here is a plot of the velocity and acceleration with the smoother and it can be seen that the spikes in the curve are softened:

https://github.com/user-attachments/assets/81503e41-f1c1-4202-b592-b9ae990329f2

- A video also shows that speed limit and lateral acceleration limits are preserved: 

https://github.com/user-attachments/assets/73151bbe-fca8-41ff-b814-866a9ccb170d

- However, the new smoother comes with high processing times (though the last one was slow too I think), adding an extra 30 ms, sometimes more to the processing time, which might be too much for multimodal processing:

https://github.com/user-attachments/assets/69016163-baef-4a48-9e7b-803f63cd55ba 

Finally, since the JerkConstrained smoother is being removed, the folder config/velocity_smoother should be removed too

planning/autoware_trajectory_optimizer/config/plugins/trajectory_velocity_optimizer.param.yaml
Owen-Liuyuxuan: https://github.com/autowarefoundation/autoware_core/blob/main/planning/autoware_velocity_smoother/config/default_velocity_smoother.param.yaml

@danielsanchezaran  actually, in rule-based planner, the limitation is even higher. I believe we are using 1.0 in taxi and 0.8 for bus.

In DP, I chose 1.0 so that DP will not over-react.

planning/autoware_trajectory_optimizer/config/trajectory_optimizer.param.yaml
Owen-Liuyuxuan: @danielsanchezaran  Should I set it to false, or should I completely revert this?

danielsanchezaran: remove the debug publishing and lets discuss how to do it better for another PR please

planning/autoware_trajectory_optimizer/include/autoware/trajectory_optimizer/utils.hpp
Owen-Liuyuxuan: @danielsanchezaran 
Thank you! This include was for the deleted un-used function

planning/autoware_trajectory_optimizer/config/plugins/trajectory_velocity_optimizer.param.yaml
danielsanchezaran: And the rule-based planner tends to be quite sluggish when taking curves. I see this feature more as a safety precaution than an actual enforcement that is supposed to be used all the time. So I think it is OK to use higher values

Owen-Liuyuxuan: @danielsanchezaran 
I updated the value in bf40323a6ac7819aa667d1ab59144b29237261f3.

But I think we would need to further tune it in a real vehicle for projects.

planning/autoware_trajectory_optimizer/config/trajectory_optimizer.param.yaml
Owen-Liuyuxuan: @danielsanchezaran 

f69a4ae67b3ac3a07c18c94a8096d0b909aaed6b

Directly cleaned up the debug publishers.

planning/autoware_trajectory_optimizer/include/autoware/trajectory_optimizer/utils.hpp
Owen-Liuyuxuan: f3d0282189565c18a857027481e93f91f77de5f1

@danielsanchezaran

planning/autoware_trajectory_optimizer/src/trajectory_optimizer_plugins/plugin_utils/trajectory_velocity_optimizer_utils.cpp
Owen-Liuyuxuan: @danielsanchezaran Thank you
10d92c4cd549459da31aee45e8ae1ab997360489

planning/autoware_trajectory_optimizer/src/trajectory_optimizer.cpp
Owen-Liuyuxuan: @danielsanchezaran 
Cleaned up in f69a4ae67b3ac3a07c18c94a8096d0b909aaed6b

planning/autoware_trajectory_optimizer/include/autoware/trajectory_optimizer/trajectory_optimizer_plugins/plugin_utils/continuous_jerk_smoother.hpp
```diff
@@ -0,0 +1,129 @@
...
+   */
+  ContinuousJerkSmootherParams get_params() const;
+
+private:
+  /**
+   * @brief Calculate interval distances between trajectory points
+   * @param trajectory Input trajectory
+   * @return Vector of distances between consecutive points
+   */
+  std::vector<double> calcTrajectoryIntervalDistance(const TrajectoryPoints & trajectory) const;
```
danielsanchezaran: please change to snake_case, all functions should be in snake_case to maintain consistency

planning/autoware_trajectory_optimizer/src/trajectory_optimizer_plugins/plugin_utils/continuous_jerk_smoother.cpp
```diff
@@ -0,0 +1,294 @@
...
+
+  const uint32_t l_variables = 5 * N;    // gamma has N-1 elements
+  const uint32_t l_constraints = 4 * N;  // N + N + (N-1) + (N-1) + 2 = 4N - 2
+
+  // Allocate matrices
+  Eigen::MatrixXd A = Eigen::MatrixXd::Zero(l_constraints, l_variables);
+  std::vector<double> lower_bound(l_constraints, 0.0);
+  std::vector<double> upper_bound(l_constraints, 0.0);
+  Eigen::MatrixXd P = Eigen::MatrixXd::Zero(l_variables, l_variables);
+  std::vector<double> q(l_variables, 0.0);
```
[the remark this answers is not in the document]
Owen-Liuyuxuan: sounds quite difficult, lets skip this first.

planning/autoware_trajectory_optimizer/src/trajectory_optimizer_plugins/plugin_utils/continuous_jerk_smoother.cpp
```diff
@@ -0,0 +1,294 @@
...
+  std::vector<double> v_max_arr(N, 0.0);
+  if (max_velocity_per_point.empty()) {
+    // Fallback: use a large value if max_velocity_per_point is not provided
+    // (should not happen in practice as it's always initialized in trajectory_velocity_optimizer)
+    std::fill(v_max_arr.begin(), v_max_arr.end(), 1000.0);
+  } else {
+    for (size_t i = 0; i < N; ++i) {
+      v_max_arr.at(i) = max_velocity_per_point.at(i);
+    }
+  }
```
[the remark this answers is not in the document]
Owen-Liuyuxuan: The vector is initialized to the same size.

planning/autoware_trajectory_optimizer/include/autoware/trajectory_optimizer/trajectory_optimizer_plugins/plugin_utils/trajectory_velocity_optimizer_utils.hpp
```diff
@@ -64,34 +64,46 @@ void set_max_velocity(TrajectoryPoints & input_trajectory_array, const float max
...
+ * @return Vector of max velocities per point based on lateral acceleration limits.
+ *         Points that were limited will have their constrained velocity;
+ *         other points will have their original velocity as the max.
  */
-void limit_lateral_acceleration(
-  TrajectoryPoints & input_trajectory_array, const double max_lateral_accel_mps2,
+std::vector<double> limit_lateral_acceleration(
+  TrajectoryPoints & input_trajectory_array, std::vector<double> & max_velocity_per_point,
+  const double max_lateral_accel_mps2, const double min_limited_speed_mps,
   const Odometry & current_odometry);
```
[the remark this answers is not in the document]
Owen-Liuyuxuan: Yes, I should only mutate in-place.

planning/autoware_trajectory_optimizer/src/trajectory_optimizer.cpp
```diff
@@ -66,11 +67,10 @@ void TrajectoryOptimizer::initialize_optimizers()
   // Get plugin names from parameter
   const auto plugin_names = get_parameter("plugin_names").as_string_array();
 
-  // Load each plugin in order
-  for (const auto & plugin_name : plugin_names) {
-    load_plugin(plugin_name);
+  // Load each plugin in order and create debug publishers
```
danielsanchezaran:
```suggestion
  // Load each plugin in order
```

planning/autoware_trajectory_optimizer/src/trajectory_optimizer.cpp
```diff
@@ -184,8 +184,8 @@ void TrajectoryOptimizer::on_traj([[maybe_unused]] const CandidateTrajectories::
   CandidateTrajectories output_trajectories = *msg;
   for (auto & trajectory : output_trajectories.candidate_trajectories) {
     // Apply optimizations - plugins execute in order from plugin_names parameter
-    for (auto & plugin : plugins_) {
-      plugin->optimize_trajectory(trajectory.points, params_, data);
+    for (size_t i = 0; i < plugins_.size(); ++i) {
+      plugins_[i]->optimize_trajectory(trajectory.points, params_, data);
```
danielsanchezaran: This change is no longer needed as we are not using the debug publishers anymore

planning/autoware_trajectory_optimizer/src/trajectory_optimizer_plugins/plugin_utils/trajectory_velocity_optimizer_utils.cpp
```diff
@@ -161,57 +165,70 @@ void limit_lateral_acceleration(
...
-    // Handle wrap-around
-    if (delta_theta > M_PI) {
-      delta_theta -= 2.0 * M_PI;
-    } else if (delta_theta < -M_PI) {
-      delta_theta += 2.0 * M_PI;
-    }
 
+    const double yaw_current = autoware_utils_geometry::get_rpy(current_pose).z;
+    const double yaw_next = autoware_utils_geometry::get_rpy(next_pose).z;
+    const double delta_yaw = autoware_utils_geometry::normalize_radian(yaw_next - yaw_current);
```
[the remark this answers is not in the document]
Owen-Liuyuxuan: I believe there was some API changes?

danielsanchezaran (commented):
This PR is looking great, I have added some final (I think) comments. Also please do git rm config/velocity_smoothing to remove the config files used by the JerkFilteredSmoother as it is no longer used (I think you need to remove it from the optimizer .launch file too)

planning/autoware_trajectory_optimizer/src/trajectory_optimizer_plugins/trajectory_velocity_optimizer.cpp
```diff
@@ -91,26 +90,32 @@ void TrajectoryVelocityOptimizer::optimize_trajectory(
...
                                  ? static_cast<float>(external_velocity_limit->max_velocity)
                                  : static_cast<float>(velocity_params_.default_max_velocity_mps);
     trajectory_velocity_optimizer_utils::set_max_velocity(traj_points, max_speed_mps);
+    for (auto & v : max_velocity_per_point) {
+      v = std::min(v, static_cast<double>(max_speed_mps));
+    }
     if (external_velocity_limit) {
       pub_velocity_limit_->publish(*external_velocity_limit);
     }
   }
```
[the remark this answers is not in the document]
Owen-Liuyuxuan: @danielsanchezaran  I think I should systematically handle this problem, 
1. When smooth is false, directly update the max_velocity_per_point by the `limit_lateral_accel` and `limit_speed`
2. When smooth is true, centralize the process and only update the threshold?

planning/autoware_trajectory_optimizer/src/trajectory_optimizer_plugins/trajectory_velocity_optimizer.cpp
```diff
@@ -74,9 +66,16 @@ void TrajectoryVelocityOptimizer::optimize_trajectory(
...
+
+  // Apply lateral acceleration limiting and update max velocity constraints
   if (velocity_params_.limit_lateral_acceleration) {
-    trajectory_velocity_optimizer_utils::limit_lateral_acceleration(
-      traj_points, velocity_params_.max_lateral_accel_mps2, data.current_odometry);
+    // limit_lateral_acceleration returns per-point max velocities based on curvature
+    max_velocity_per_point = trajectory_velocity_optimizer_utils::limit_lateral_acceleration(
+      traj_points, max_velocity_per_point, velocity_params_.max_lateral_accel_mps2,
+      velocity_params_.min_limited_speed_mps, data.current_odometry);
   }
```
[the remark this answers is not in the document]
danielsanchezaran: @Owen-Liuyuxuan this is true, what do you think can be a good solution?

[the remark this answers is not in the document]
Owen-Liuyuxuan: @danielsanchezaran 
This is also a problem at the limit_velocity.

- I think I should systematically handle this problem,

When smooth is false, directly update the max_velocity_per_point by the limit_lateral_accel and limit_speed.
When smooth is true, centralize the process and only update the threshold.

I believe this can produce the best performance.

planning/autoware_trajectory_optimizer/src/trajectory_optimizer_plugins/plugin_utils/trajectory_velocity_optimizer_utils.cpp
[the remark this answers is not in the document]
danielsanchezaran: I think Copilot is wrong here, we should use autoware_utils_geometry I believe

planning/autoware_trajectory_optimizer/src/trajectory_optimizer_plugins/plugin_utils/trajectory_velocity_optimizer_utils.cpp
```diff
@@ -161,57 +165,70 @@ void limit_lateral_acceleration(
...
+    const double current_speed = std::abs(itr->longitudinal_velocity_mps);
+    if (current_speed < min_limited_speed_mps) {
+      // No modification needed for low speeds
+      continue;
+    }
 
-  for (auto itr = start_itr; itr < std::prev(input_trajectory_array.end()); ++itr) {
     const auto current_pose = itr->pose;
     const auto next_pose = std::next(itr)->pose;
     const auto delta_time = get_delta_time(std::next(itr), itr);
```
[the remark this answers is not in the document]
Owen-Liuyuxuan: 3a9803fce5b9144d1c8085227f73ac111d1dd511

planning/autoware_trajectory_optimizer/src/trajectory_optimizer_plugins/plugin_utils/trajectory_velocity_optimizer_utils.cpp
```diff
@@ -161,57 +165,70 @@ void limit_lateral_acceleration(
...
 
+    const double yaw_current = autoware_utils_geometry::get_rpy(current_pose).z;
+    const double yaw_next = autoware_utils_geometry::get_rpy(next_pose).z;
+    const double delta_yaw = autoware_utils_geometry::normalize_radian(yaw_next - yaw_current);
     constexpr double epsilon_yaw_rate = 1.0e-5;
-    const double yaw_rate = std::max(std::abs(delta_theta / delta_time), epsilon_yaw_rate);
-    const double current_speed = std::abs(itr->longitudinal_velocity_mps);
-    // Compute lateral acceleration
+    const double yaw_rate = std::max(std::abs(delta_yaw / delta_time), epsilon_yaw_rate);
     const double lateral_acceleration = current_speed * yaw_rate;
```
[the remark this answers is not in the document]
Owen-Liuyuxuan: 3a9803fce5b9144d1c8085227f73ac111d1dd511

planning/autoware_trajectory_optimizer/src/trajectory_optimizer.cpp
Owen-Liuyuxuan: 903e51fc1141f4c6b749b592ea9f34d13261e5d8

Owen-Liuyuxuan:
@danielsanchezaran 

9cb475cf254d2bee9c05042955116a0f6d1e5507  This commit is to implement explicit choosing between inplace update or not.

9d958328c83c8c78268d5c3573717782130bea1f  & 89618873e27cd78138caca4a183636eea0170a23  These twos are for jerkfilter clean up

903e51fc1141f4c6b749b592ea9f34d13261e5d8 plugin loops restoring

3a9803fce5b9144d1c8085227f73ac111d1dd511  Delta time fix

danielsanchezaran (approved):
I think everything is in order. Thank you so much for this great PR, lets see if we can further test on real vehicles and set these functions to default in the future. Note that I will approve this PR, but please update the documentation before merging

Owen-Liuyuxuan:
Deploy docs are automatically cancelled...
I can not check the documentation update directly. Forgive me to skip this check further.

PR #12023, opened 2026-02-03
feat(autoware_trajectory_optimizer): add external velocity limit subscription

danielsanchezaran:
## Description
This PR adds a subscription to the "/planning/scenario_planning/max_velocity_default" inside the VelocityOptimizer plugin. With this PR, if the "limit_speed" parameter is true, the VelocityOptimizer will use the default velocity limit  "max_vel" value, normally found in the [common.param.yaml ](https://github.com/autowarefoundation/autoware_launch/blob/6589f099e1917b1b904e1ea03ea7d24d3ef709cb/autoware_launch/config/planning/scenario_planning/common/common.param.yaml#L3) file. If a velocity change request comes from the  "/planning/scenario_planning/max_velocity_default" topic, the node will use said speed as the new limit. 

A diagram of the implementation can be seen here: 
API-diagram drawio

A video example here: 

https://github.com/user-attachments/assets/e693aeaf-c068-452c-b564-b5de8b9abb05

## Related links

**Parent Issue:**

- Link

## How was this PR tested?
PSim
## Notes for reviewers

None.

#### Additions and removals

| Added| Sub | `/planning/scenario_planning/max_velocity_default` | `autoware_internal_planning_msgs/VelocityLimit`   | Velocity Limit |
| Added| Pub | `/planning/scenario_planning/current_max_velocity` | `autoware_internal_planning_msgs/VelocityLimit`   | Velocity Limit |

None.

## Effects on system behavior

External velocity limit setters can be used to cap the speed.

### DIFF
--- a/planning/autoware_trajectory_optimizer/package.xml
+++ b/planning/autoware_trajectory_optimizer/package.xml
@@ -35,6 +35,8 @@
   <depend>tf2_geometry_msgs</depend>
   <depend>visualization_msgs</depend>
 
+  <exec_depend>autoware_core_planning</exec_depend>
+
   <test_depend>ament_cmake_ros</test_depend>
   <test_depend>ament_lint_auto</test_depend>
   <test_depend>autoware_lint_common</test_depend>


planning/autoware_trajectory_optimizer/src/trajectory_optimizer_plugins/trajectory_velocity_optimizer.cpp
```diff
@@ -71,8 +85,13 @@ void TrajectoryVelocityOptimizer::optimize_trajectory(
   }
 
   if (velocity_params_.limit_speed) {
-    trajectory_velocity_optimizer_utils::set_max_velocity(
-      traj_points, static_cast<float>(max_speed_mps));
+    const auto max_param_speed_mps = static_cast<float>(velocity_params_.max_speed_mps);
+    const auto max_speed_mps =
+      (latest_external_velocity_limit_opt_.has_value())
+        ? std::min(latest_external_velocity_limit_opt_->max_velocity, max_param_speed_mps)
+        : max_param_speed_mps;
```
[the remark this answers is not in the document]
danielsanchezaran: Since there is no way for latest_external_velocity_limit_opt_ to be set to nullopt once set, this is not an issue, there are no value requirements

planning/autoware_trajectory_optimizer/include/autoware/trajectory_optimizer/trajectory_optimizer_plugins/trajectory_velocity_optimizer.hpp
```diff
@@ -67,6 +74,8 @@ class TrajectoryVelocityOptimizer : public TrajectoryOptimizerPluginBase
 private:
   std::shared_ptr<JerkFilteredSmoother> jerk_filtered_smoother_{nullptr};
   TrajectoryVelocityOptimizerParams velocity_params_;
+  rclcpp::Subscription<VelocityLimit>::SharedPtr sub_planning_velocity_;
```
go-sakayori: I assume you are implementing something similar to velocity smoother. Velocity smoother publishes the current velocity limit. Don't we need it as well?
https://github.com/autowarefoundation/autoware_core/blob/main/planning/autoware_velocity_smoother/src/node.cpp#L62-L63  

Please refer to [TIER IV internal link](https://star4.slack.com/archives/CEV8XMJBV/p1766396991888579?thread_ts=1766137536.389439&cid=CEV8XMJBV).

yukkysaito: Right, that makes sense.
I remember there was an arbitration node. Do you know which component is responsible for publishing the maximum velocity?

go-sakayori: I think [External Velocity Limit Selector](https://github.com/autowarefoundation/autoware_universe/tree/main/planning/autoware_external_velocity_limit_selector) and [/awapi_awiv_adapter](https://github.com/tier4/tier4_ad_api_adaptor/blob/tier4/universe/awapi_awiv_adapter/launch/awapi_awiv_adapter.launch.xml#L67) is related to this.

danielsanchezaran: @go-sakayori @yukkysaito The external Velocity Limit Selector is in charge of choosing the velocity limit. It decides between the hard coded "max_vel" value, normally found in the [common.param.yaml ](https://github.com/autowarefoundation/autoware_launch/blob/6589f099e1917b1b904e1ea03ea7d24d3ef709cb/autoware_launch/config/planning/scenario_planning/common/common.param.yaml#L3) file and the input given by the API, like FOA. So I think we dont need to worry about the external velocity limiter since it comes before the optimizer in the pipeline. The velocity optimizer receives the output of the external_velocity_limit_selector as an input to set the velocity limit of the trajectory (the same way the velocity smoother does it), finally I added a publisher that re-publishes the velocity limit to the "/planning/scenario_planning/current_max_velocity" topic, which is what the velocity_smoother also does. 

Regarding [/awapi_awiv_adapter](https://github.com/tier4/tier4_ad_api_adaptor/blob/tier4/universe/awapi_awiv_adapter/launch/awapi_awiv_adapter.launch.xml#L67): 

This is a topic I would like to discuss: the /planning/scenario_planning/current_max_velocity topic is [used by the awapi_awiv_adapter](https://github.com/tier4/tier4_ad_api_adaptor/blob/e272bef2f88df1c3a36074f23c9cd800eee5459a/awapi_awiv_adapter/launch/awapi_awiv_adapter.launch.xml#L40), so I made the velocity optimizer also output this topic. 

HOWEVER,  in [this line](https://github.com/tier4/tier4_ad_api_adaptor/blob/e272bef2f88df1c3a36074f23c9cd800eee5459a/awapi_awiv_adapter/launch/awapi_awiv_adapter.launch.xml#L67) we can see how the /awapi_awiv_adapter depends on the velocity_smoother node, in [this line](https://github.com/tier4/tier4_ad_api_adaptor/blob/e272bef2f88df1c3a36074f23c9cd800eee5459a/awapi_awiv_adapter/src/awapi_awiv_adapter_core.cpp#L32)  we can see how the awapi adapter "copies" the "max_velocity" parameter from the velocity smoother node, however, I think this makes little sense as I believe that we should be requesting for the "max_vel" parameter, normally found in the [common.param.yaml ](https://github.com/autowarefoundation/autoware_launch/blob/6589f099e1917b1b904e1ea03ea7d24d3ef709cb/autoware_launch/config/planning/scenario_planning/common/common.param.yaml#L3) I am currently consulting about this issue

go-sakayori (approved):
The code looks good! Can you modify the interface changes, since you added a publisher.

PR #12042, opened 2026-02-09
refactor(tensorrt_yolox): split utility functions

a-maumau:
## Description
This PR moves the utility functions from `tensorrt_yolox.cpp` into `utils.cpp/.hpp`.
It prepares for parameterizing the label remapping, which is currently [hard-coded in the source.](https://github.com/autowarefoundation/autoware_universe/blob/256f769d4a2ea9d2de5b41a8319bd108d9de0324/perception/autoware_tensorrt_yolox/src/tensorrt_yolox_node.cpp#L252-L263).

## Related links

**Parent Issue:**

- Link

**Private Links:**

- related to [TIERIV internal link](https://tier4.atlassian.net/browse/T4DEV-48449)

## How was this PR tested?

Tested with `yolox_traffic_light_detector.launch.xml` and `yolox.launch.xml`
## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/.github/workflows/pre-commit-lite.yaml
+++ b/.github/workflows/pre-commit-lite.yaml
@@ -14,7 +14,7 @@ jobs:
 
       - uses: actions/setup-python@v5
         with:
-          python-version: 3.x
+          python-version: "3.10"
 
       - uses: pre-commit/action@v3.0.1
 
--- a/packages_above.repos
+++ b/packages_above.repos
@@ -7,10 +7,22 @@ repositories:
     type: git
     url: https://github.com/autowarefoundation/autoware_launch.git
     version: main
+  universe/external/eagleye:
+    type: git
+    url: https://github.com/MapIV/eagleye.git
+    version: 575136ebba99892946d36d8398f228aee2136af0
+  universe/external/rtklib_ros_bridge:
+    type: git
+    url: https://github.com/MapIV/rtklib_ros_bridge.git
+    version: ef094407bba4f475a8032972e0c60cbb939b51b8
+  universe/external/llh_converter:
+    type: git
+    url: https://github.com/MapIV/llh_converter.git
+    version: 4fc2a2e1bc9dcf3e6ab0a8085d8257168e160342
   sensor_component/external/nebula:
     type: git
     url: https://github.com/tier4/nebula.git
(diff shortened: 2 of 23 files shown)


ktro2828 (changes requested):
@a-maumau Thank you for your PR! I'd like to avoid naming `utils` as much as possible, use concrete words instead.
In this case, for example, `label` or `text` sounds reasonable.

a-maumau:
@ktro2828 Thank you for reviewing!
> I'd like to avoid naming utils as much as possible, use concrete words instead.

Does this mean, not `label_utils` but `label_...`? In that case I think I will go with `label_processing`.

ktro2828:
> @ktro2828 Thank you for reviewing!
> 
> > I'd like to avoid naming utils as much as possible, use concrete words instead.
> 
> Does this mean, not `label_utils` but `label_...`? In that case I think I will go with `label_processing`.

@a-maumau I meant just `label` or `text` is fine. `xxx_utils` sounds redandunt.

perception/autoware_tensorrt_yolox/src/tensorrt_yolox.cpp
```diff
@@ -36,74 +37,8 @@
...
-        file = prefixed;
-      else
-        std::cerr << "WARNING: couldn't find: " << prefixed << " while loading: " << filename
-                  << std::endl;
-    }
-  }
-  return fileList;
-}
+using autoware::tensorrt_yolox::loadListFromTextFile;
+using autoware::tensorrt_yolox::trim;
```
ktro2828: @a-maumau How about wrapping unnamed namespace with `namespace autoware::tensorrt_yolox`? It allows not to declare `using` directive.

```cpp
namespace autoware::tensorrt_yolox
{
namespace
{
  // ...
} // anonymous namespace
} // autoware::tensorrt_yolox
```

a-maumau: Fixed in https://github.com/autowarefoundation/autoware_universe/pull/12042/commits/30f95000e0d464e80189429d1827306a547cf63d .

ktro2828 (approved):
LGTM

PR #12043, opened 2026-02-10
fix(autoware_trajectory_optimizer): prevent yaw spikes when creating splines 

danielsanchezaran:
## Description
This PR removes the direct copying of the starting and final trajectory points of the original trajectory. Instead, these points are also calculated using the smoothing util 
## Related links
Related to issue reported here: https://github.com/autowarefoundation/autoware_universe/pull/12021
**Parent Issue:**

- Link

## How was this PR tested?
PSim
## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/autoware_trajectory_optimizer/src/trajectory_optimizer_plugins/plugin_utils/trajectory_spline_smoother_utils.cpp
+++ b/planning/autoware_trajectory_optimizer/src/trajectory_optimizer_plugins/plugin_utils/trajectory_spline_smoother_utils.cpp
@@ -64,16 +64,17 @@ void apply_spline(
 
   TrajectoryPoints output_points{};
   output_points.reserve(static_cast<size_t>(1 + trajectory_interpolation_util->length() / ds));
-  double s{};
-  for (s = 0.0; s <= trajectory_interpolation_util->length(); s += ds) {
+  double last_s{0.0};
+  for (auto s = 0.0; s <= trajectory_interpolation_util->length(); s += ds) {
     auto p = trajectory_interpolation_util->compute(s);
     if (!autoware::trajectory_optimizer::utils::validate_point(p)) {
       continue;
     }
     output_points.push_back(p);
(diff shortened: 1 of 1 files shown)


planning/autoware_trajectory_optimizer/src/trajectory_optimizer_plugins/plugin_utils/trajectory_spline_smoother_utils.cpp
```diff
@@ -59,44 +59,34 @@ void apply_spline(
     return;
   }
   trajectory_interpolation_util->align_orientation_with_trajectory_direction();
-  TrajectoryPoints output_points{traj_points.front()};
```
Owen-Liuyuxuan, line 62: First point no-longer copied.

planning/autoware_trajectory_optimizer/src/trajectory_optimizer_plugins/plugin_utils/trajectory_spline_smoother_utils.cpp
```diff
@@ -59,44 +59,34 @@ void apply_spline(
...
 
-  if (output_points.size() < 2) {
-    auto clock = rclcpp::Clock::make_shared(RCL_ROS_TIME);
-    RCLCPP_WARN_THROTTLE(
-      rclcpp::get_logger("trajectory_spline_smoother"), *clock, 5000,
-      "Not enough points in trajectory after akima spline interpolation");
-    return;
+  if (trajectory_interpolation_util->length() - last_s > min_interpolation_step) {
+    output_points.push_back(
+      trajectory_interpolation_util->compute(trajectory_interpolation_util->length()));
```
Owen-Liuyuxuan, line 79: Compute the interpolation result at the final end of the path, making sure that the output path has the same spatial length as the input.

PR #12044, opened 2026-02-10
feat(autoware_tensorrt_vad): update nvcc flags

amadeuszsz:
## Description

- Adjust gencode flags for SM 86-120 to avoid JIT overhead and enable architecture-specific optimizations on modern GPUs.
- Handle CUDA 13.0 SM101 -> SM110 rename.

## Related links

**Parent Issue:**

- https://github.com/autowarefoundation/autoware/issues/6789

## Notes for reviewers

Selected SMs assures trade-off between binary size and potential target devices (skipped server-like SMs).

## Interface changes

None.

## Effects on system behavior

None.


Max-Bin (approved):
LGTM

PR #12062, opened 2026-02-10
fix(out_of_lane): clarify boolean argument name to avoid semantic ambiguity

leon-nie-pnc:
## Summary

This PR renames a boolean argument in
`calculate_collisions_to_avoid` to reduce semantic ambiguity.

The original name `is_stopping` could be misinterpreted as the ego vehicle's
current motion state, while the actual meaning is whether a previous
avoidance / stop decision already exists.

## Changes

- Rename `is_stopping` to `has_prev_avoidance_decision`
- Update all related call sites
- Add a clarifying comment in the function declaration
- No functional or behavioral change

## Motivation

Clear naming is important for safety-related decision logic.
This change improves readability and reduces the risk of incorrect usage
during future maintenance or extension.

## Impact

- No behavior change
- Improves maintainability and code clarity

### DIFF
--- a/planning/motion_velocity_planner/autoware_motion_velocity_out_of_lane_module/src/out_of_lane_collisions.cpp
+++ b/planning/motion_velocity_planner/autoware_motion_velocity_out_of_lane_module/src/out_of_lane_collisions.cpp
@@ -276,7 +276,8 @@ void calculate_collisions_to_avoid(
   for (auto & p : out_of_lane_data.outside_points) {
     calculate_min_max_arrival_times(p, trajectory);
     if (params.mode == "ttc") {
-      const auto threshold = has_prev_avoidance_decision ? params.ttc_release_threshold : params.ttc_threshold;
+      const auto threshold =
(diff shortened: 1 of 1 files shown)


maxime-clem (approved):
This makes sense 👍 
It would also be good to rename the variable here:
https://github.com/leon-nie-pnc/autoware_universe/blob/decdf732b812c4be4fb179b129ad463e904bb231/planning/motion_velocity_planner/autoware_motion_velocity_out_of_lane_module/src/out_of_lane_module.cpp#L466

leon-nie-pnc:
> This makes sense 👍 It would also be good to rename the variable here: https://github.com/leon-nie-pnc/autoware_universe/blob/decdf732b812c4be4fb179b129ad463e904bb231/planning/motion_velocity_planner/autoware_motion_velocity_out_of_lane_module/src/out_of_lane_module.cpp#L466

Good catch, thanks 👍  
I've renamed the variable there as well to keep the semantics consistent.

PR #12063, opened 2026-02-10
feat(autoware_probabilistic_occupancy_grid_map): use ApproximateTime synchronizer for pointcloud-based OGM

shmpwk:
### Summary
This PR replaces the strict timestamp equality check between `obstacle_pointcloud` and `raw_pointcloud` with a `message_filters::Synchronizer` using the `ApproximateTime` policy in the pointcloud-based probabilistic occupancy grid map node. This improves robustness against small timestamp jitter and reduces missed processing triggers, helping stabilize the output rate (e.g., 10 Hz).

### Why
- Exact timestamp matching is fragile in real systems where sensor and transport timing jitter exists.
- Approximate time sync reduces “dropouts” where processing never triggers because stamps are not identical.
- Queue size fixed at 10 provides a simple, stable default without adding tuning surface.

## Related links

[TIER IV internal ticket](https://tier4.atlassian.net/browse/T4DEV-44446)

## How was this PR tested?

I verified in the real taxi environment:

**Before** : The output of pointcloud-based OGM map was sometimes delayed relative to the input pointclouds.
- Output: `/perception/occupancy_grid_map/*/map`
- Input: `/perception/occupancy_grid_map/*/downsample/pointcloud`

OGM sync

**After** : The delay in the OGM output has been eliminated, and it is now synchronized with the input.

OGM sync (2)
OGM sync (1)

## Notes for reviewers
This part came from https://github.com/autowarefoundation/autoware_universe/pull/10060 by @knzo25.
~~I believe there is no intention to synchronize two input pointclouds exactly by looking at that PR.~~

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_probabilistic_occupancy_grid_map/src/pointcloud_based_occupancy_grid_map/pointcloud_based_occupancy_grid_map_node.cpp
+++ b/perception/autoware_probabilistic_occupancy_grid_map/src/pointcloud_based_occupancy_grid_map/pointcloud_based_occupancy_grid_map_node.cpp
@@ -64,14 +64,18 @@ PointcloudBasedOccupancyGridMapNode::PointcloudBasedOccupancyGridMapNode(
     this->declare_parameter<bool>("filter_obstacle_pointcloud_by_raw_pointcloud");
   const double map_length = this->declare_parameter<double>("map_length");
   const double map_resolution = this->declare_parameter<double>("map_resolution");
+  constexpr int approx_sync_queue_size = 10;
 
   /* Subscriber and publisher */
-  obstacle_pointcloud_sub_ptr_ = this->create_subscription<PointCloud2>(
-    "~/input/obstacle_pointcloud", rclcpp::SensorDataQoS{}.keep_last(1),
-    std::bind(&PointcloudBasedOccupancyGridMapNode::obstaclePointcloudCallback, this, _1));
-  raw_pointcloud_sub_ptr_ = this->create_subscription<PointCloud2>(
-    "~/input/raw_pointcloud", rclcpp::SensorDataQoS{}.keep_last(1),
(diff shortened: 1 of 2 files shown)


shmpwk:
@YoshiRi 
This part came from https://github.com/autowarefoundation/autoware_universe/pull/10060 by @knzo25. But he was not TIER IV member anymore.
Could you select the reviewer from your team?

technolojin:
`fromROSMsgAsync` may have asynchronized behavior, so there is chance of racing condition between two `fromROSMsgAsync`s and `onPointcloudWithObstacleAndRaw`. In other word, `onPointcloudWithObstacleAndRaw` can be triggered before `fromROSMsgAsync` is done, so the OGM runs with old messages.

we have to make the message sync explicitly. the PR implementation does not.

technolojin (changes requested):
This change uses approximate-time sync and then immediately starts processing when a pair of pointcloud messages arrives.
The code first calls fromROSMsgAsync() for both clouds, which may start work (conversion/upload) that finishes later.
Even with a single-threaded node, “async” work can still be incomplete when processing begins.
That can cause the algorithm to read stale/unfinished pointcloud data and produce wrong or unstable output.
Fix by making data readiness explicit (wait/barrier) or by processing directly from the synced messages.

YoshiRi:
This OGM generation node is designed to use the exact same timestamp pointclouds so that it can compare raw/obstacle pointcloud in its rigid definition. So, if topic hz drop is caused by obstacle pointcloud delay, we theoretically need to fix update logic to keep output OGM hz.

shmpwk:
Thank you for your review.
Because of the following reason, I will close this PR. Instead, the product side will handle it if needed.
- the current implementation is opposed to the design
- there is chance of racing condition

PR #12064, opened 2026-02-10
chore(stream_petr): remove invalid thrust stream policy for copy operations

SamratThapa120:
## Description

Using stream-aware execution policy was causing the following error when multiple processes that use CUDA are running in the same GPU .
```
[autoware_camera_streampetr_node-1] terminate called after throwing an instance of 'thrust::system::system_error'
[autoware_camera_streampetr_node-1]   what():  __copy:: D->D: failed: cudaErrorInvalidValue: invalid argument
```
This issue was partially fixed in the previous PR https://github.com/autowarefoundation/autoware_universe/pull/11800, but it was not fully resolved. This PR resolves the remaining issues.

This error arises because the current code tries to copy from device to host memory using device execution policy. 

## How was this PR tested?

Tested using various rosbags locally.

## Interface changes

None.


YoshiRi (approved):
LGTM

PR #12074, opened 2026-02-12
chore(autoware_pointcloud_preprocessor): add `GLOBAL_SECONDS` increment for test so that future pre-commit works

paulsohn:
## Description

In tier4 fork of autoware_universe, I have discovered:
* [pre-commit fails](https://github.com/tier4/autoware_universe/actions/runs/21932363694/job/63338966797?pr=2674) because there is no setuptools installed. Have a separate PR for this (https://github.com/tier4/pre-commit-hooks-ros/pull/107)
* After resolving the above issue, [pre-commit fails](https://github.com/tier4/autoware_universe/actions/runs/21938159103/job/63356900888?pr=2677) notifying that we don't have an unused global for some test cases.

```
flake8 ros...............................................................Failed
- hook id: flake8-ros
- exit code: 1
sensing/autoware_pointcloud_preprocessor/test/test_concatenate_node_component.py:431:9: F824 `global GLOBAL_SECONDS` is unused: name is never assigned in scope
        global GLOBAL_SECONDS
        ^
sensing/autoware_pointcloud_preprocessor/test/test_concatenate_node_component.py:601:9: F824 `global GLOBAL_SECONDS` is unused: name is never assigned in scope
        global GLOBAL_SECONDS
        ^
2     F824 `global GLOBAL_SECONDS` is unused: name is never assigned in scope
```

I could've just suppressed F824 alternatively, but from the other test cases, I suspect that the intention of these test would be adding `GLOBAL_SECONDS += 1`.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

CI of this PR itself

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None. This commit only affects for unit test.


paulsohn:
It seems like I managed the version poorly and had the flake8 version too recent, but I think it is worth updating the test scripts.

paulsohn:
@vividf can you check if this update agrees with your intention?

PR #12079, opened 2026-02-13
fix: update pre-commit-hooks-ros version

mitsudome-r:
## Description
This fixes the error from pre-commit-lite. ([example](http://github.com/autowarefoundation/autoware_universe/actions/runs/21968099266/job/63462950826?pr=12004))

It seems like flake8-ros started to fail in the latest versions of python since pkg_resources module is deprecated.

## Related links
https://github.com/autowarefoundation/autoware_universe/pull/12072

## How was this PR tested?

Tested with this PR: 

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -47,7 +47,7 @@ repos:
       - id: check-package-depends
 
   - repo: https://github.com/tier4/pre-commit-hooks-ros
-    rev: v0.10.1
+    rev: v0.10.0
     hooks:
       - id: flake8-ros
       - id: prettier-xacro


.pre-commit-config.yaml
```diff
@@ -47,7 +47,7 @@ repos:
       - id: check-package-depends
 
   - repo: https://github.com/tier4/pre-commit-hooks-ros
-    rev: v0.10.0
+    rev: v0.10.1
```
paulsohn:
```suggestion
    rev: v0.10.0
```

This will be overwritten by sync-file workflow, and I have [confirmed](https://github.com/tier4/autoware_universe/actions/runs/21971716472/job/63474552004?pr=2677) that v0.10.1 is not necessary.

mitsudome-r: I have reverted the change

paulsohn (approved):

PR #12088, opened 2026-02-13
refactor(trajectory_safety_filter): replace bool to optional

zulfaqar-azmi-t4:
## Description

This PR replace `bool` return types with `tl::expected`.

By moving to `tl::expected`, error handling and logging can be improved. Instead of functions returning a generic false upon failure, they now return a descriptive error type, allowing the caller to understand why a trajectory was rejected (e.g., collision, out of bounds, or invalid input).

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/autoware_trajectory_safety_filter/src/trajectory_safety_filter_node.cpp
+++ b/planning/autoware_trajectory_safety_filter/src/trajectory_safety_filter_node.cpp
@@ -98,15 +98,15 @@ void TrajectorySafetyFilter::process(const CandidateTrajectories::ConstSharedPtr
     }
 
     // Apply each filter to the trajectory
-    bool is_safe = true;
+    bool is_feasible = true;
     for (const auto & plugin : plugins_) {
       if (const auto res = plugin->is_feasible(trajectory.points, context); !res) {
-        is_safe = false;
+        is_feasible = false;
         RCLCPP_WARN(get_logger(), "Not feasible: %s", res.error().c_str());
       }
     }
 
-    if (is_safe) filtered_msg->candidate_trajectories.push_back(trajectory);
+    if (is_feasible) filtered_msg->candidate_trajectories.push_back(trajectory);
   }
 
   // Also filter generator_info to match kept trajectories


planning/autoware_trajectory_safety_filter/src/filters/collision_filter.cpp
```diff
@@ -259,11 +261,11 @@ void CollisionFilter::set_parameters(const std::unordered_map<std::string, std::
   get_value("min_value", params_.min_ttc);
 }
 
-bool CollisionFilter::is_feasible(
+std::optional<std::string> CollisionFilter::validate(
```
yukkysaito: Nice improvement 👍
Returning the reason instead of just a bool will be very useful for debugging and evaluation.

However, in the current form the API may be easy to misread or misuse, so I have a concern.

std::optional<std::string> means:
- nullopt → OK
- value → NG (reason)

But from the name validate(), readers would normally expect either:

- a bool return value, or
- an exception on failure

So at the call site, code like:

```
if (filter.validate(...)) {
  ...
}
```

could easily be misinterpreted as “true means success.”

zulfaqar-azmi-t4: @yukkysaito 

Thank you. Yes, that make sense. 
How about `has_violation`?

yukkysaito: has_violation tends to make people expect a boolean, and they will naturally try to write something like if (has_violation).
How about using a pattern like the following instead?

Example: define a Result type inside CollisionFilter (or a common header)
```
struct FeasibleResult
{
  bool ok;
  std::string reason;

  explicit operator bool() const { return ok; }

  static FeasibleResult success() { return {true, ""}; }
  static FeasibleResult fail(std::string r) { return {false, std::move(r)}; }
};
```

Revised implementation (change return type to FeasibleResult)
```
FeasibleResult CollisionFilter::is_feasible(
  const TrajectoryPoints & traj_points, const FilterContext & context)
{
  if (!context.predicted_objects || context.predicted_objects->objects.empty()) {
    return FeasibleResult::success();  // No objects to check collision with
  }

  const auto cache =
    precompute_object_positions(*context.predicted_objects, params_.max_check_time);

  // Check each trajectory point within time horizon
  for (const auto & point : traj_points) {
    const double time_from_start = rclcpp::Duration(point.time_from_start).seconds();
    if (time_from_start > params_.max_check_time) {
      break;
    }

    // Check collision with each cached object
    for (size_t obj_idx = 0; obj_idx < cache.size(); ++obj_idx) {
      if (check_collision(point, cache, obj_idx, time_from_start)) {
        return FeasibleResult::fail(
          fmt::format("Collision with object at time {:.2f}s", time_from_start));
      }
    }
  }

  return FeasibleResult::success();
}
```

Call site example
```
const auto res = collision_filter.is_feasible(traj_points, context);
if (!res) {
  RCLCPP_WARN(logger, "Not feasible: %s", res.reason.c_str());
}
```

zulfaqar-azmi-t4: In that case, just as suggestion, I'd like to propose using `tl::expected<void, std::string>`, or `tl::expected<bool, std::string>` which provides the same behavior, instead of creating a custom `FeasibleResult` struct. 


```c++
tl::expected<bool, std::string> CollisionFilter::is_feasible(
    const TrajectoryPoints & traj_points, const FilterContext & context)
{
  if (!context.predicted_objects || context.predicted_objects->objects.empty()) {
    return true;  // Feasible - no collision possible
  }

  const auto cache = 
    precompute_object_positions(*context.predicted_objects, params_.max_check_time);

  for (const auto & point : traj_points) {
    const double time_from_start = rclcpp::Duration(point.time_from_start).seconds();
    if (time_from_start > params_.max_check_time) {
      break;
    }

    for (size_t obj_idx = 0; obj_idx < cache.size(); ++obj_idx) {
      if (check_collision(point, cache, obj_idx, time_from_start)) {
        return tl::unexpected{
          fmt::format("Collision with object at time {:.2f}s", time_from_start)};
      }
    }
  }

  return true;  // Feasible - no collision found
}

```

And in call site is the same as your suggestion
```c++
const auto res = collision_filter.is_feasible(traj_points, context);
if (!res) {
  RCLCPP_WARN(logger, "Not feasible: %s", res.reason.c_str());
}
```

The advantage `expected`
- less custom type to follow.

The disadvantage `expected`
- `expected` normally used to handle exceptions, so it might be overkill in this case. 
- The bool is always `true` when the check succeeds. There is no `false`
  - The bool itself has no special meaning - it just makes the code easier to read.

yukkysaito: I agree for the suggestion.
I think moving to expected makes sense.

Since this function does not have any meaningful value to return on success, tl::expected<void, std::string> might better represent the intent than including a bool.
In that case, the call site would use res.error() to access the reason.

go-sakayori:
@zulfaqar-azmi-t4 
Would bool be replaced as well in the traffic rule filter? May be better to keep it the same architecture.

planning/autoware_trajectory_safety_filter/src/trajectory_safety_filter_node.cpp
```diff
@@ -100,9 +100,9 @@ void TrajectorySafetyFilter::process(const CandidateTrajectories::ConstSharedPtr
     // Apply each filter to the trajectory
     bool is_feasible = true;
     for (const auto & plugin : plugins_) {
-      if (!plugin->is_feasible(trajectory.points, context)) {
+      if (const auto res = plugin->is_feasible(trajectory.points, context); !res) {
         is_feasible = false;
-        break;
```
go-sakayori, line 105: Removing `break` adds unnecessary compute and log noise: once a trajectory is already infeasible, running remaining filters does not change the result, only increases cost. If the intention is to collect all failure reasons, then it's fine.

zulfaqar-azmi-t4: @go-sakayori 
I understand the concern regarding compute cost and log noise, but there are three main reasons of why evaluating all filters rather than returning early:

1. A trajectory might fail multiple checks. If we return early, we lose visibility into subsequent failures. Having the complete set of failure reasons in the logs significantly speeds up the debugging process.
2. There's plan to merge the filters and ranker in the future. In that architecture, the ranker may use a weighted scoring system to choose the best path rather than a binary pass/fail. Keeping the filters flexible now ensures we have the necessary data for a smarter ranker later.
3.  In ideal case, a feasible path needs to run all checks anyway, if we extend the same understanding to infeasible path, by evaluating all paths against all filters, we can identify potential computation issues or edge cases across the entire pipeline. It is better to catch these performance or logic issues now during the development phase rather than after the architecture is finalized.

zulfaqar-azmi-t4:
@go-sakayori 

> Would bool be replaced as well in the traffic rule filter? May be better to keep it the same architecture.

Yes, most probably. I will discuss with @maxime-clem and make a PR later.

planning/autoware_trajectory_safety_filter/src/trajectory_safety_filter_node.cpp
go-sakayori: Understood! Seems reasonable for me

zulfaqar-azmi-t4:
@yukkysaito 
Can I get your approval for this PR as well?

yukkysaito (approved):

PR #12116, opened 2026-02-17
feat(crosswalk_traffic_light_estimator): remove /route topic to simplify estimation logic

interimadd:
## Description

Remove the route (`~/input/route`) dependency from `autoware_crosswalk_traffic_light_estimator` and switch to a signal-driven crosswalk estimation approach.

Previously, the node required `~/input/route` to determine which crosswalks to estimate signals for, by finding crosswalks that conflict with route lanelets. This had limitations:
- Estimation only worked for crosswalks on the current route
- Required waiting for both map and route to be available

With this change, the node precomputes mappings between vehicle traffic light IDs and crosswalks at map loading time. At runtime, it identifies relevant crosswalks based on the received (and recently detected) vehicle traffic signal IDs, removing the need for route information entirely.

Key changes:
- Remove `~/input/route` subscription and `onRoute` callback
- Remove `autoware_planning_msgs` dependency
- Precompute `traffic_light_id_to_crosswalks_` and `crosswalk_to_vehicle_lanelets_` mappings in `onMap`
- In `onTrafficLightArray`, collect crosswalks to process based on known vehicle traffic light IDs (from current and last detected signals)
- Add integration tests to verify estimation behavior

## Related links

**Private Links:**

- [TIER IV internal link](https://star4.slack.com/archives/C03S84LDJGG/p1770355034477919)

### Integration Test

- Added integration tests (`test_crosswalk_traffic_light_estimator_node.cpp`):
  - `StraightGreenEstimatesRed`: Verifies that when a vehicle traffic light on a straight lanelet is GREEN, the crosswalk signal is estimated as RED
  - `AllRedEstimatesUnknown`: Verifies that when all vehicle traffic lights are RED, the crosswalk signal is estimated as UNKNOWN
- Built and ran tests with:
  ```
  colcon build --packages-select autoware_crosswalk_traffic_light_estimator
  colcon test --packages-select autoware_crosswalk_traffic_light_estimator --event-handlers console_direct+
  ```

### Performance Test

- CPU usage decreased from 24.7% to 0.82%.
- Processing time decreased from 12.3ms to 0.2ms.

image

The scripts and rosbag used for the performance test are available in the following Google Drive folder:
https://drive.google.com/drive/folders/1IMtNCt523U4wK5pwzVYFczq0U4Wxg6Bs?usp=sharing

Measurement can be run with `./run_crosswalk_estimator.sh`.

### Integration test with rosbag
Pedestrian signal estimation results before and after the changes were compared using the Odaiba map and a rosbag.

image

The left shows the signal topic before entering the node, the middle shows the estimation results before the change, and the right shows the results after the change. 

The main difference is that the "unknown" duration is shorter because the route is no longer referenced (as it stops predicting upcoming signals and estimating signals already passed), but the red light estimation results remain consistent. That difference will not cause problem for actual use cases as discussed in following threads.

https://star4.slack.com/archives/C03S84LDJGG/p1771561939278539?thread_ts=1770355034.477919&cid=C03S84LDJGG

Detailed Measurement methodology:
- Play only /judged/traffic_light_signal, /map, and /route from the Odaiba rosbag.
- Launch only the crosswalk_traffic_light_estimator node.
- Record the topics output from the crosswalk_traffic_light_estimator node.
- Plot the time-series transitions of the pedestrian signal estimation results using a script.

The scripts and rosbag used for the performance test are available in the following Google Drive folder:
https://drive.google.com/drive/folders/1IMtNCt523U4wK5pwzVYFczq0U4Wxg6Bs?usp=sharing
Actual measurement commands are belows.
```
./run_crosswalk_estimator.sh
python3 plot_traffic_light_signals.py output/rosbag_20260219_201744/  --topic /crosswalk_traffic_light_estimator/output/traffic_signals --save traffic_light_after.png
```

## Notes for reviewers

The following line consumed almost all of the processing time:

```cpp
const auto conflict_lls = overall_graphs_ptr_->conflictingInGraph(crosswalk, VEHICLE_GRAPH_ID);
```

This function is called to get lanelets that conflict with crosswalks existing on the subscribed route.

## Interface changes

None.

#### Additions and removals

| Change type | Topic Type | Topic Name      | Message Type                                 | Description                       |
|:------------|:-----------|:----------------|:---------------------------------------------|:----------------------------------|
| Removed     | Sub        | `~/input/route` | `autoware_planning_msgs::msg::LaneletRoute`  | Route input is no longer required |

## Effects on system behavior

- The node no longer requires a route to estimate crosswalk traffic signals. Estimation is now performed for all crosswalks associated with detected vehicle traffic lights, regardless of the current route.


perception/autoware_crosswalk_traffic_light_estimator/src/node.cpp
```diff
@@ -296,9 +293,32 @@ void CrosswalkTrafficLightEstimatorNode::onTrafficLightArray(
...
+      continue;
+    }
+    for (const auto & crosswalk : crosswalks_iter->second) {
+      crosswalk_ids_to_process.insert(crosswalk.id());
+    }
+  }
+
+  for (const auto & crosswalk_id : crosswalk_ids_to_process) {
+    const auto & crosswalk = lanelet_map_ptr_->laneletLayer.get(crosswalk_id);
+    const auto & conflict_lls = crosswalk_to_vehicle_lanelets_.at(crosswalk_id);
```
YoshiRi, line 321: [very minor comment]
While the logic ensures the key exists for now, using .at() could potentially crash the node if future changes break this invariant. To make it more robust against future refactors, consider using find() and adding a check.

interimadd: Thank you for the review! I'm planning to do more refactoring to extract the logic, so I'll make sure to address your comments in upcoming PRs.

MasatoSaeki (approved):
LGTM
This PR works well in my local PC.

PR #12131, opened 2026-02-18
fix(autoware_cuda_pointcloud_preprocessor): add conditional include for CUDA 13.0 and above

yhisaki:
## Description

The thrust inclusion path has changed in CUDA 13.0.

https://developer.nvidia.com/blog/whats-new-and-important-in-cuda-toolkit-13-0/#cccl_headers_have_moved_in_cuda_130

This PR has been adjusted to reflect this change.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/sensing/autoware_cuda_pointcloud_preprocessor/CMakeLists.txt
+++ b/sensing/autoware_cuda_pointcloud_preprocessor/CMakeLists.txt
@@ -48,6 +48,10 @@ include_directories(
     ${CUDA_INCLUDE_DIRS}
 )
 
+if(NOT CUDA_VERSION VERSION_LESS "13.0")
+  include_directories(SYSTEM ${CUDA_INCLUDE_DIRS}/cccl)
+endif()
+
 # cSpell: ignore expt
 list(APPEND CUDA_NVCC_FLAGS "--expt-relaxed-constexpr -diag-suppress 20012")
 list(APPEND CUDA_NVCC_FLAGS "-gencode arch=compute_86,code=sm_86")
--- a/sensing/autoware_cuda_pointcloud_preprocessor/CMakeLists.txt
+++ b/sensing/autoware_cuda_pointcloud_preprocessor/CMakeLists.txt
@@ -48,7 +48,7 @@ include_directories(
     ${CUDA_INCLUDE_DIRS}
 )
 
-if(NOT CUDA_VERSION VERSION_LESS "13.0")
+if(CUDA_VERSION VERSION_GREATER_EQUAL "13.0")
   include_directories(SYSTEM ${CUDA_INCLUDE_DIRS}/cccl)
 endif()
 


sensing/autoware_cuda_pointcloud_preprocessor/CMakeLists.txt
```diff
@@ -48,6 +48,10 @@ include_directories(
     ${CUDA_INCLUDE_DIRS}
 )
 
+if(NOT CUDA_VERSION VERSION_LESS "13.0")
+  include_directories(SYSTEM ${CUDA_INCLUDE_DIRS}/cccl)
```
mojomex, line 52: From [NVIDIA's blog](https://developer.nvidia.com/blog/whats-new-and-important-in-cuda-toolkit-13-0/#cccl_headers_have_moved_in_cuda_130):

> ❌ Do not write #include <cccl/...> — this will break.
If including CCCL headers only in files compiled with nvcc
✅ No action needed. (This is the common case for most users)
If using CCCL headers in files compiled exclusively with a host compiler (e.g., GCC, Clang, MSVC):
Using CMake and linking CCCL::CCCL
✅ No action needed. (This is the recommended path. See [example](https://github.com/NVIDIA/cccl/tree/main/examples/basic))
Other build systems
⚠️ Add ${CTK_ROOT}/include/cccl to your compiler’s include search path (e.g., with -I)

I think moving to `target_link_libraries(<target> PRIVATE CCCL::CCCL)` would be the best path forward.

If we indeed add the `.../cccl` include dir, we should ensure that it doesn't mess with linters as

```cmake
include_directories(
    include
    SYSTEM
    ${CUDA_INCLUDE_DIRS}
)
``` 

is still also there.

sensing/autoware_cuda_pointcloud_preprocessor/CMakeLists.txt
```diff
@@ -48,6 +48,10 @@ include_directories(
     ${CUDA_INCLUDE_DIRS}
 )
 
+if(NOT CUDA_VERSION VERSION_LESS "13.0")
```
mojomex: Probably more readable:
```suggestion
if(CUDA_VERSION VERSION_GREATER_EQUAL "13.0")
```

mojomex (commented):
Thanks for tackling this. I've left a few comments about NVIDIA's recommended solution, what are your thoughts? Tbh, as long as it works, we don't have to do it "the perfect way". Your solution is clean enough.

sensing/autoware_cuda_pointcloud_preprocessor/CMakeLists.txt
yhisaki: I can't find the CCCL module in the cmake, so I'll investigate. 🤔

mojomex: Hm, it seems it's installed as part of CUDAToolkit on Ubuntu. This snippet seems to work:

```cmake
  find_package(CCCL CONFIG QUIET)
  if(NOT CCCL_FOUND)
    find_package(CUDAToolkit REQUIRED)
    set(CCCL_DIR "${CUDAToolkit_TARGET_DIR}/targets/${CMAKE_SYSTEM_PROCESSOR}-linux/lib/cmake/cccl")
    find_package(CCCL CONFIG REQUIRED)
  endif()
```

then target_link_libraries(... CCCL::CCCL) works. This should also work on other systems/CCCL installations.

xmfcx:
- @yhisaki -san, I've created https://github.com/autowarefoundation/autoware_universe/pull/13302 that encompasses and fixes this issue too. I took your and @mojomex 's solution and iterated on it a bit and created that PR. With that, universe compiles fine now.

Could you review that? Also we can close this PR if you don't mind.

PR #12134, opened 2026-02-19
chore(streampetr): update configs and launch file to reflect autoware artifacts

SamratThapa120:
## Description

This PR updates the streampetr launch file to match the first model artifacts released in autoware

https://github.com/autowarefoundation/autoware/pull/6826

https://github.com/autowarefoundation/autoware/pull/6841

- The launch file and configs were updated to match the number of camera inputs for the released model which is 5
-  The README was updated to include model download links
- Unused lines (localization topic remap) from launch file was removed.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_camera_streampetr/config/camera_streampetr.param.yaml
+++ b/perception/autoware_camera_streampetr/config/camera_streampetr.param.yaml
@@ -15,7 +15,7 @@
       head_path: "$(var model_path)/simplify_pts_head_memory.onnx" # Path to the detection head onnx file, relative to the model_param_path
       position_embedding_path: "$(var model_path)/simplify_position_embedding.onnx" # Path to the position embedding onnx file, relative to the model_param_path
 
-      backbone_engine_path: ""   # Path to the tensorRT engine file for the backbone. If empty, the path will be set to bachbone_path with .engine extension, and the engine will be built if it does not exist.
+      backbone_engine_path: ""   # Path to the tensorRT engine file for the backbone. If empty, the path will be set to backbone_path with .engine extension, and the engine will be built if it does not exist.
       head_engine_path: ""       # Path to the tensorRT engine file for the head. If empty, the path will be set to head_path with .engine extension, and the engine will be built if it does not exist.
       position_embedding_engine_path: "" # Path to the tensorRT engine file for the position embedding. If empty, the path will be set to position_embedding_path with .engine extension, and the engine will be built if it does not exist.
 


perception/autoware_camera_streampetr/config/tensorrt_stream_petr.param.yaml
```diff
@@ -9,7 +9,7 @@
       input_image_height: 480
       input_image_width: 640
       num_proposals: 5400
-      rois_number: 6
+      rois_number: 5
```
amadeuszsz: Slightly out of related PR context, but please consider it:
- Our naming convention for config used to be <package-name-without-"autoware"-prefix>.param.yaml, do you think we can rename it to `camera_streampetr.param.yaml`?
- We split config on two separate files: node related and model related. Therefore I suggest to have:
  - `camera_streampetr.param.yaml` with node configuration e.g. trt_precision etc. Users are free to change it.
    - Any other node config will be called accordingly e.g. `camera_streampetr_mini.param.yaml`, so json schema configuration named `camera_streampetr.schema.json` can parse all the node configs as in our CI it will look for `camera_streampetr*.param.yaml` files.
  - `ml_package_camera_streampetr.param.yaml` with deployed model configuration e.g. num_proposals. This config is deployed alongside onnx file and by default shouldn't be changed by user - any change may result runtime error due to model incompatibility.
    - Any other model config will be called accordingly e.g. `ml_package_camera_streampetr_mini.param.yaml`, so json schema configuration named `ml_package_camera_streampetr.schema.json` can parse all the model configs as in our CI it will look for `ml_package_camera_streampetr*.param.yaml` files.

Even though artifact are already deployed, this is very early stage of this package so we can consider these changes. Sorry if this topic has already been discussed and I wasn't aware of it.

SamratThapa120: Thanks for the comments, I agree with your suggestions. Please wait while I update the artifacts and filenames.

SamratThapa120: @amadeuszsz Fixed in [4b5f940](https://github.com/autowarefoundation/autoware_universe/pull/12134/commits/4b5f940d0b2cde356a1200eda3fc2334c1f6569b)

perception/autoware_camera_streampetr/launch/streampetr.launch.xml
```diff
@@ -1,41 +1,42 @@
 <launch>
```
amadeuszsz, line 1: Please rename launch file to `streampetr.launch.xml` and update documentation. 

Also do we need two launch files? Please check if "compression" option can be handled by argument and appropriate `if` block condition.

I noticed you are setting parameter in launch file which does not exists in yaml file. Please take a look how it is resolved in other package e.g. CenterPoint. In other words, all the parameters shall be defined in yaml file and covered in schema file.

Sorry for pointing it that late 🙏🏻

SamratThapa120: Thanks for the feedback. [43d13f6](https://github.com/autowarefoundation/autoware_universe/pull/12134/commits/43d13f6042b321259a50dfabb04dde298b68b6ee)
- I have renamed he launch file as you suggested
- I have merged the launch files to one using `is_compressed_image` param for branching.
- I moved `max_camera_time_diff` and `multithreading` to the `camera_streampetr.param.yaml` file, but I left `anchor_camera_id`, `is_compressed_image`, `is_distorted_image` in the launch file under the camera topic remap section because these 3 parameters heavily depend upon the camera topic specified and should be decided after the camera topics to be subscribed to are added.
- Currently streampetr package does not have a schema file, I will leave it as a task for future PR.

amadeuszsz: For Autoware, the rule of thumb is any declared parameter has no default value. The only exception is "build_only" (for CI purposes). There is still a lot of missing parameter in your node's parameter file. Here are your parameters:

> logger_name
> multithreading
> model_params.rois_number
> max_camera_time_diff
> anchor_camera_id
> debug_mode
> model_params.input_image_width
> model_params.input_image_height
> model_params.backbone_path
> model_params.head_path
> model_params.position_embedding_path
> model_params.backbone_engine_path
> model_params.head_engine_path
> model_params.position_embedding_engine_path
> model_params.trt_precision
> model_params.workspace_size
> model_params.use_temporal
> post_process_params.iou_nms_search_distance_2d
> post_process_params.circle_nms_dist_threshold
> post_process_params.iou_nms_threshold
> post_process_params.confidence_threshold
> model_params.class_names
> model_params.num_proposals
> post_process_params.yaw_norm_thresholds
> model_params.detection_range
> model_params.pre_memory_length
> model_params.post_memory_length
> build_only  ---> can be with default values and not placed in param file
> is_compressed_image
> is_distorted_image

- Some of parameters are not necessary and can be hardcoded, e.g. "logger_name".
- Parameters needed only for launch file which are not declared in node source code shouldn't be stored in yaml file.
- If you need to override, you can resolve it using substitution (`allow_substs`) as on shown example:
```yaml
/**:
  ros__parameters:
    model_params:
      backbone_path:
        "$(var model_path)/simplify_extract_img_feat.onnx"
```
Or maybe even `"$(var model_path)/$(var model_name)_img_feat.onnx"` if we want to standarize naming convention for future models.

Having:
- default values,
- some of parameters set in yaml files
- some of parameters set in launch file
- some of parameters declared only in node

confuse users. Please take a look on it.

SamratThapa120: @amadeuszsz Thanks for the rigorous checks 💯, it makes sense. I have updated the config files to include all defined parameters and removed some unnecessary parameters too [4576854](https://github.com/autowarefoundation/autoware_universe/pull/12134/commits/457685413a02b90308c1628dc5b31253a841fa24). Please re-check when you have time and do not hesitate to request for changes again.

amadeuszsz: @SamratThapa120 
Sorry, my previous comment didn't clearly explain all needed changes:
Probably the last one - please remove default value for parameters from node's cpp file 🙇🏻‍♂️

SamratThapa120: @amadeuszsz THanks. Fixed in 
83181ee

SamratThapa120:
@amadeuszsz Is it okay if I merge the PR once I get approval?

KSeangTan (approved):
LGTM

amadeuszsz:
> @amadeuszsz Is it okay if I merge the PR once I get approval?

Before merge I wanted to test it, but if you already run latest version, please feel free to merge

PR #12155, opened 2026-02-20
fix(boundary_departure_checker): link pybind11 properly

xmfcx:
## Description

- https://github.com/autowarefoundation/autoware_universe/pull/12093 added pyplot support for the package.

It compiles fine for humble but for jazzy it needs explicit linking like similar packages that use pyplot in the autoware workspace.

Compilation errors (on jazzy) before this PR:
https://gist.github.com/xmfcx/ea033facd6f4ab610f81f5c0ee02fad7

## Related links

 - **Parent Issue:** https://github.com/autowarefoundation/autoware_universe/issues/11418
 - **Similar to:** https://github.com/autowarefoundation/autoware_universe/pull/11936

## How was this PR tested?

CI and locally.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/common/autoware_boundary_departure_checker/CMakeLists.txt
+++ b/common/autoware_boundary_departure_checker/CMakeLists.txt
@@ -35,7 +35,10 @@ target_link_libraries(boundary_departure_checker
 if(BUILD_TESTING)
   find_package(ament_lint_auto REQUIRED)
   ament_lint_auto_find_test_dependencies()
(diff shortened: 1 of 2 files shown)


xmfcx:
We are trying to get jazzy ci working, after that we shouldn't have these kind of regressions hopefully.

mitsudome-r (approved):

PR #12160, opened 2026-02-24
feat(autoware_detected_object_feature_remover):  add convex hull process

technolojin:
## Description

1. add convex-hull process (no convex hull process by default)
2. separate business logic from node implementation. the business logic can be used in different nodes. 
3. update test not to rely on the ROS implementation

## Related links

related issue raised https://github.com/autowarefoundation/autoware_universe/issues/11740

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_detected_object_feature_remover/CMakeLists.txt
+++ b/perception/autoware_detected_object_feature_remover/CMakeLists.txt
@@ -5,7 +5,6 @@ find_package(autoware_cmake REQUIRED)
 autoware_package()
 
 find_package(PCL REQUIRED COMPONENTS common)
-find_package(pcl_conversions REQUIRED)
 find_package(OpenCV REQUIRED)
 
 set(DETECTED_OBJECT_FEATURE_REMOVER_DEPENDENCIES
--- a/perception/autoware_detected_object_feature_remover/package.xml
+++ b/perception/autoware_detected_object_feature_remover/package.xml
@@ -15,6 +15,9 @@
 
   <depend>autoware_perception_msgs</depend>
   <depend>autoware_utils</depend>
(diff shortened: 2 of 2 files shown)


perception/autoware_detected_object_feature_remover/include/autoware/detected_object_feature_remover/convert.hpp
```diff
@@ -0,0 +1,44 @@
...
+{
+
+struct ConvertParams
+{
+  bool run_convex_hull_conversion;
+};
+
+void convertToDetectedObjects(
+  const tier4_perception_msgs::msg::DetectedObjectsWithFeature & input,
+  autoware_perception_msgs::msg::DetectedObjects & output, const ConvertParams & params);
```
ktro2828, line 36: [NIT] I prefer to return an instance rather than specify writable reference as an input from the perspective of testability.
I feel it's OK to specify a writable reference if we want to **modify instance** or need to **care about its performance**.

```suggestion
autoware_perception_msgs::msg::DetectedObjects convertToDetectedObjects(
  const tier4_perception_msgs::msg::DetectedObjectsWithFeature & input,
  const ConvertParams & params);
```

ktro2828 (commented):
@technolojin Thank you for your dedicated refactoring!

perception/autoware_detected_object_feature_remover/include/autoware/detected_object_feature_remover/convert.hpp
technolojin: It is to generalized node implementation. In the future, other multiple output functions also need to be separated from node implementations.

PR #12197, opened 2026-02-26
feat(trajectory_safety_filter): add new safety filter considering vehicle constraints

ktro2828:
## Description

This pull request adds a new vehicle constraint safety filter to the `autoware_trajectory_safety_filter` package. The filter checks if planned trajectories violate vehicle-specific constraints such as maximum speed, acceleration, deceleration, steering angle, and steering rate. The PR also updates configuration files, documentation, and build/test infrastructure to support this new filter.

**Key changes:**

**1. New Vehicle Constraint Filter Implementation**
- Added `VehicleConstraintFilter` class with logic to check trajectory feasibility against speed, acceleration, deceleration, steering angle, and steering rate constraints. Includes helper functions for each check and integrates with the safety filter plugin interface. (`vehicle_constraint_filter.hpp`, `vehicle_constraint_filter.cpp`) [[1]](diffhunk://#diff-5acd55a13661402d5029c006c89aa273a7ab870c0068e0b751f2e4d1af9b3c2eR1-R124) [[2]](diffhunk://#diff-8ca4fe14c95817d9b4f0a4e1e4428a2a59a44defff81ca9f65302a999f583defR1-R264)
- Registered the new filter as a plugin in `plugins.xml`.

**2. Configuration and Parameter Updates**
- Updated the default filter list in `trajectory_safety_filter.param.yaml` to include `VehicleConstraintFilter`, and documented all available filters.
- Added default values for all vehicle constraint parameters (max speed, acceleration, deceleration, steering angle, steering rate) to configuration files. [[1]](diffhunk://#diff-2ac68ff4199031a55f7abc6d5e9cdeaf42989124c863958649910d21cf80e568R21-R27) [[2]](diffhunk://#diff-7aca0889eeac19675cedb1b5f7942d2b9c23131efd5011d4e25bbe9794ea283dR31-R57)
- Updated the CMake build to install the new filter source file and add a unit test for it.
- Declared a new dependency on `autoware_utils_geometry` for geometric calculations.

**3. Documentation**
- Updated the `README.md` to describe the new vehicle constraint filter in the list of applied checks and documented the new parameters in the parameters table. [[1]](diffhunk://#diff-4fabc379a30d9b3a55558ba0432e5f7e0daccb986633dfff3afcb3bcf546153aR16) [[2]](diffhunk://#diff-4fabc379a30d9b3a55558ba0432e5f7e0daccb986633dfff3afcb3bcf546153aL34-R45)

**4. System Integration**
- Modified the node logic to load vehicle constraint parameters when the `VehicleConstraintFilter` is used.

These changes collectively introduce a robust mechanism to ensure that only trajectories compliant with physical vehicle limits are considered safe, improving the reliability of trajectory selection.

## Related links

**Parent Issue:**

https://github.com/autowarefoundation/autoware/issues/6292

**Private Links:**

- [TIER IV internal link](https://tier4.atlassian.net/wiki/spaces/AIP/pages/4922049605/PoC+Step1#SF-FRQ-01.-%E8%BB%8A%E4%B8%A1%E3%81%AE%E6%80%A7%E8%83%BD%E5%88%B6%E7%B4%84%E3%83%AA%E3%82%B9%E3%82%AF%E3%82%92%E8%A9%95%E4%BE%A1%E3%81%99%E3%82%8B)

## How was this PR tested?

- [x] All unit testings passed
- [x] Confirmed it works and logs error if a trajectory constraint violates
  -  The following video shows the filter outputs error when trajectory speed exceeds `max_speed=5.5 (~20km/h)`, otherwise allows.

[Screencast from 02-26-2026 11:20:41 PM.webm](https://github.com/user-attachments/assets/600bf5ba-fa43-4b60-be23-8d26a061f87f)

## Notes for reviewers

None.

## Interface changes

None.

#### Additions and removals

| Added | `vehicle_constraint.max_speed`   | `double` | `16.7`         | Max allowed speed [m/s] |
| Added | `vehicle_constraint.max_acceleration` | `double` | `5.0`     | Max allowed acceleration [m/s^2] |
| Added | `vehicle_constraint.max_deceleration` | `double` | `5.0`    | Max allowed deceleration [m/s^2; positive but represents deceleration] |
| Added | `vehicle_constraint.max_steering_angle` | `double` | `0.8` | Max allowed steering angle [rad] |
| Added | `vehicle_constraint.max_steering_rate`   | `double` | `0.3` | Max allowed steering rate [rad/s] |

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_predicted_path_postprocessor/CMakeLists.txt
+++ b/perception/autoware_predicted_path_postprocessor/CMakeLists.txt
@@ -1,10 +1,6 @@
 cmake_minimum_required(VERSION 3.14)
 project(autoware_predicted_path_postprocessor)
 
-if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
-    add_compile_options(-Wall -Wextra -Wpedantic)
-endif()
-
 # --- find dependencies ---
 find_package(autoware_cmake REQUIRED)
 autoware_package()
@@ -18,10 +14,6 @@ target_include_directories(
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
     $<INSTALL_INTERFACE:include/${PROJECT_NAME}>)
 
-target_link_libraries(${PROJECT_NAME}
-  ament_index_cpp::ament_index_cpp
-)
-
 rclcpp_components_register_node(
     ${PROJECT_NAME} PLUGIN
     "autoware::predicted_path_postprocessor::PredictedPathPostprocessorNode"
@@ -32,11 +24,12 @@ if(BUILD_TESTING)
     ament_auto_find_test_dependencies()
 
     ament_add_ros_isolated_gtest(test_${PROJECT_NAME}
-        test/test_refine_by_speed.cpp)
+        test/test_collision.cpp
+        test/test_refine_by_speed.cpp
+        test/test_refine_penetration_by_static_objects.cpp)
 
     target_link_libraries(test_${PROJECT_NAME}
       ${PROJECT_NAME}
-      ament_index_cpp::ament_index_cpp
     )
 endif()
 
--- a/perception/autoware_predicted_path_postprocessor/README.md
+++ b/perception/autoware_predicted_path_postprocessor/README.md
@@ -10,6 +10,8 @@ The following processors are supported:
 
 - [RefineBySpeed](./docs/refine_by_speed.md)
   - Refine the paths of objects based on their current speed.
+- [RefinePenetrationByStaticObjects](./docs/refine_penetration_by_static_objects.md)
+  - Refine the penetrated paths of objects based on the other static objects.
 
 ## Inputs / Outputs
 
@@ -33,7 +35,7 @@ The following processors are supported:
 To use this package, you can launch it with the following command:
 
 ```bash
-ros2 launch autoware_predicted_path_postprocessor autoware_predicted_path_postprocessor.launch.xml
+ros2 launch autoware_predicted_path_postprocessor predicted_path_postprocessor.launch.xml
 ```
 
 ### Leverage Processor in Your Codebase
@@ -106,8 +108,9 @@ As an example, let's see how to add a new processor by using a processor called 
        }
 
     private:
-       result_type check_context(const Context & context)
+       result_type check_context(const Context & context) const noexcept override
        {
+         // Override if necessary!!
          // ...Check context if it contains the required information
          return make_ok<error_type>();
        }
@@ -131,7 +134,7 @@ As an example, let's see how to add a new processor by using a processor called 
    {
      std::vector<ProcessorInterface::UniquePtr> outputs;
      for (const auto & name : processor_names) {
-       if ( /* ... */) {
+       if ( /* ... */ ) {
          // ...
        } else if (name == "filter_by_something") {
          outputs.push_back(std::make_unique<FilterBySomething>(node_ptr, name));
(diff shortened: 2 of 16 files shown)


planning/autoware_trajectory_safety_filter/test/vehicle_constraint_filter.cpp
```diff
@@ -0,0 +1,297 @@
...
+    create_trajectory_point(0.0, 0.0, 0.0, 5.0, 0.0, 0.0),
+    create_trajectory_point(1.0, 1.0, 0.0, 5.5, 0.1, 1.0),
+    create_trajectory_point(2.0, 2.0, 0.0, 6.0, 0.2, 2.0)};
+
+  VehicleInfo vehicle_info;
+  vehicle_info.wheel_base_m = 2.5;  // Example wheelbase
+
+  VehicleConstraintFilter filter;
+  filter.set_parameters(
+    {{"max_velocity", 10.0},
```
go-sakayori: `VehicleConstraintFilter::set_parameters()` expects `vehicle_constraint.*` keys, but `test/vehicle_constraint_filter.cpp` passes unprefixed keys (e.g. `max_velocity`), so those test parameters are not applied.

I implemented this part with string-based parameter keys, and that made it easy to introduce a mismatch between runtime keys (`vehicle_constraint.*`) and test keys (`max_*`). That’s on me sry.

planning/autoware_trajectory_safety_filter/src/filters/vehicle_constraint_filter.cpp
```diff
@@ -0,0 +1,244 @@
...
+  double dt = to_seconds(curr_point.time_from_start) - to_seconds(prev_point.time_from_start);
+  return dt > 0 ? (curr_vel - prev_vel) / dt : 0.0;
+}
+
+/**
+ * @brief Convert TrajectoryPoint to steering angle (rad)
+ */
+double to_steering_angle(const TrajectoryPoint & point, const VehicleInfo & vehicle_info)
+{
+  return std::atan2(point.lateral_velocity_mps, point.longitudinal_velocity_mps) *
```
go-sakayori: 1. Both rule-based planners and the [diffusion planner](https://github.com/autowarefoundation/autoware_universe/blob/main/planning/autoware_diffusion_planner/src/postprocessing/postprocessing_utils.cpp#L249-L342) often do not explicitly compute `lateral_velocity_mps`, so it is usually left as zero.
2. This does not calculate the steering angle. Should be something like below:
    ``` const double steering_angle = atan2(vehicle_info.wheel_base,turning_radius);```

planning/autoware_trajectory_safety_filter/test/vehicle_constraint_filter.cpp
ktro2828: @go-sakayori Thank you for catching, I addressed in https://github.com/autowarefoundation/autoware_universe/pull/12197/commits/f6436ecfc09b283c2256a34a11518b59c277110f.

planning/autoware_trajectory_safety_filter/src/filters/vehicle_constraint_filter.cpp
ktro2828: @go-sakayori Thank you for catching and advising, I overlooked this. I fixed in https://github.com/autowarefoundation/autoware_universe/pull/12197/commits/51af7b52cdf565bd146e76e91cf227b1bad27952, but I'm not sure this is the common way to calculate a steering angle in planning module.

go-sakayori: May be need smoothing for the curvature calculation, but the equation itself to calculate the steering angle LGTM.  

> common way to calculate a steering angle in planning module
Maybe better to ask other reviewers if there is a better solution.

PR #12219, opened 2026-03-02
refactor(boundary_departure_prevention): move nearest projection function to utils

zulfaqar-azmi-t4:
## Description

Move function out of class into utils. This reduces direct dependency on class state, making the logic easier to reason about and unit test in isolation.

## How was this PR tested?

Build success. Unit test all pass.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/common/autoware_boundary_departure_checker/include/autoware/boundary_departure_checker/utils.hpp
+++ b/common/autoware_boundary_departure_checker/include/autoware/boundary_departure_checker/utils.hpp
@@ -421,15 +421,14 @@ std::optional<ProjectionsToBound> get_closest_projections_for_side(
   const double min_braking_dist, const double max_braking_dist, const SideKey side_key);
 
 /**
- * @brief Evaluates multiple footprint candidates and selects the
- * projection with the highest departure priority.
+ * @brief Evaluates multiple footprint candidates and selects the projection with the highest
+ * departure priority.
  *
  * This function determines the worst-case boundary departure scenario for a specific point in time
  * by applying a strict safety hierarchy:
- * 1. Higher severity states (CRITICAL > APPROACHING > NEAR) always override lower
- * ones.
- * 2. If candidates share the exact same departure type, the one with the
- * shortest lateral distance to the boundary is selected.
+ * 1. Higher severity states (CRITICAL > APPROACHING > NEAR) always override lower ones.
+ * 2. If candidates share the exact same departure type, the one with the shortest lateral distance
+ * to the boundary is selected.
  *
  * @param candidate_projections List of projections from various footprint types
  * @param param Configuration parameters containing safety thresholds


common/autoware_boundary_departure_checker/src/utils.cpp
```diff
@@ -772,4 +802,130 @@ std::optional<std::pair<double, double>> is_point_shifted(
...
+
+      candidate.footprint_type_opt = footprint_type;
+      candidate_projections.push_back(candidate);
+    }
+
+    std::optional<double> previous_longitudinal_distance =
+      min_to_bound.empty() ? std::nullopt
+                           : std::make_optional(min_to_bound.back().lon_dist_on_pred_traj);
+
+    auto closest_projection = get_closest_projection_at_index(
```
maxime-clem: (nitpick) the name of the function is confusing as it does not use any index.
`get_closest_projection` would be better I think.

common/autoware_boundary_departure_checker/src/utils.cpp
```diff
@@ -101,6 +101,36 @@ std::vector<SegmentWithIdx> create_local_segments(const lanelet::ConstLineString
...
+    return false;
+  }
+
+  const auto & fr_proj_to_bound = projections_to_bound[footprint_type_order.front()][side_key];
+  const auto check_size = [&](const auto footprint_type) {
+    return fr_proj_to_bound.size() != projections_to_bound[footprint_type][side_key].size();
+  };
+
+  return !std::any_of(
+    std::next(footprint_type_order.begin()), footprint_type_order.end(), check_size);
```
maxime-clem, line 132: This code is a hard to understand.
Are we making sure that, for the given side, all ProjectionsToBound have the same size over all footprint types ?

common/autoware_boundary_departure_checker/src/utils.cpp
```diff
@@ -772,4 +802,130 @@ std::optional<std::pair<double, double>> is_point_shifted(
...
+      closest_projection = candidate;
+      break;
+    }
+
+    // Ignore candidates that are too far away
+    if (candidate.lat_dist > th_dist_near) continue;
+
+    // Keep the candidate that is closest to the boundary
+    if (!closest_projection || candidate.lat_dist < closest_projection->lat_dist) {
+      candidate.departure_type_opt = DepartureType::NEAR_BOUNDARY;
```
maxime-clem, line 892: the order of the candidates will change whether they are marked as `NEAR_BOUNDARY`. Is this okay ?

common/autoware_boundary_departure_checker/src/utils.cpp
```diff
@@ -772,4 +802,130 @@ std::optional<std::pair<double, double>> is_point_shifted(
...
+
+  if (!closest_projection || !closest_projection->departure_type_opt) {
+    return std::nullopt;
+  }
+
+  // 2. Filter out non-critical points that are longitudinally too close to the previously accepted
+  // point
+  if (
+    previous_longitudinal_distance && !closest_projection->is_critical_departure() &&
+    std::abs(*previous_longitudinal_distance - closest_projection->lon_dist_on_pred_traj) < 0.5) {
```
maxime-clem: magic number.
Please add at least a `constexpr` variable for the `0.5` with a descriptive name.

common/autoware_boundary_departure_checker/src/utils.cpp
zulfaqar-azmi-t4: Yes, we are making sure that all projection have the same size.

common/autoware_boundary_departure_checker/src/utils.cpp
zulfaqar-azmi-t4: no problem. We are choosing the one  that has minimum lateral gap.

common/autoware_boundary_departure_checker/src/utils.cpp
zulfaqar-azmi-t4: How about `get_closest_projection_by_departure_priority`? or `get_closest_projection_by_departure_severity`?

common/autoware_boundary_departure_checker/src/utils.cpp
zulfaqar-azmi-t4: Parameterized the value https://github.com/autowarefoundation/autoware_universe/pull/12219/commits/840c2682fbffd7da634058135d32fe356962453f

common/autoware_boundary_departure_checker/src/utils.cpp
maxime-clem: Both sounds good :+1:

maxime-clem (approved):

common/autoware_boundary_departure_checker/src/utils.cpp
zulfaqar-azmi-t4: renamed https://github.com/autowarefoundation/autoware_universe/pull/12219/commits/ef4c7bedb41c9fc0e80a214c55dbdf84d986117f

PR #12224, opened 2026-03-03
fix(crosswalk): improve logic to detect stuck vehicle

satoshi-ota:
## Description

This PR refactors and enhances the obstruction prevention logic in the crosswalk module.

Previously, obstruction detection for stuck vehicles relied on a simple lateral offset threshold (`max_target_vehicle_lateral_offset`) relative to the ego path. This approach only evaluated the lateral distance from the path centerline and did not consider the actual vehicle footprint or occupied area along the passing region.

This PR introduces the following improvements:

- Replaces `max_target_vehicle_lateral_offset` with `required_lateral_clearance`.
- Expands the ego vehicle polygon laterally by a configurable margin.
- Generates an attention area by sweeping the (optionally expanded) ego footprint along the relevant crosswalk passing range.
- Uses polygon intersection (`boost::geometry::disjoint`) between the object polygon and the computed attention area instead of a simple lateral offset check.
- Extends the attention range to include:
  - The crosswalk section.
  - Additional longitudinal space equal to `vehicle_length + required_clearance`.

As a result, obstruction detection is now based on geometric overlap with the ego passing area rather than a scalar lateral distance threshold. This makes the behavior more consistent with actual vehicle footprint and better aligned with obstacle stop logic.

The new parameter `required_lateral_clearance` is recommended to be set greater than or equal to the lateral stop margin used in the obstacle_stop module.

## Related links

[TIER IV INTERNAL LINK](https://tier4.atlassian.net/browse/T4DEV-39671)

## How was this PR tested?

- [PASS TIER IV INTERNAL SCENARIOS](https://evaluation.tier4.jp/evaluation/reports/a4aa37d8-5fd0-52e1-a16f-f84d0e04a167?project_id=autoware_dev)
- [PASS TIER IV INTERNAL SCENARIOS](https://evaluation.tier4.jp/evaluation/reports/4ea2ac88-7bf2-5c9b-abcd-d5fba02bc388?project_id=prd_jt)
- [PASS TIER IV INTERNAL SCENARIOS](https://evaluation.tier4.jp/evaluation/reports/d2ec022e-f6be-56ab-899d-ee98944dcdb5?project_id=prd_jt)

## Notes for reviewers

None.

## Interface changes

| Change type | Parameter Name                                                     | Type     | Default Value | Description |
|-------------|--------------------------------------------------------------------|----------|---------------|------------|
| Removed     | `obstruction_prevention.max_target_vehicle_lateral_offset`         | `double` | `2.0`         | Maximum allowed lateral offset from ego path for obstruction check |
| Added       | `obstruction_prevention.required_lateral_clearance`                | `double` | `0.3`         | Lateral margin added to ego passing area; objects inside the expanded region trigger STOP |

## Effects on system behavior

Although no interface change is introduced, system behavior is improved as follows:

- Obstruction detection is now footprint-based instead of centerline-distance-based.
- STOP decisions become more geometrically accurate and consistent with actual ego occupancy.
- Reduces false negatives where vehicles were inside the effective passing region but outside the lateral offset threshold.
- Behavior becomes more consistent with obstacle stop module lateral margin configuration.

No impact on topics or external interfaces.

### DIFF
--- a/planning/behavior_velocity_planner/autoware_behavior_velocity_crosswalk_module/src/scene_crosswalk.cpp
+++ b/planning/behavior_velocity_planner/autoware_behavior_velocity_crosswalk_module/src/scene_crosswalk.cpp
@@ -1012,7 +1012,7 @@ void CrosswalkModule::applySlowDownByOcclusion(
 }
 
 Polygon2d CrosswalkModule::getAttentionArea(
-  const PathWithLaneId & sparse_resample_path, const std::pair<double, double> & attension_range,
+  const PathWithLaneId & sparse_resample_path, const std::pair<double, double> & attention_range,
   const double lateral_margin, std::vector<std::vector<geometry_msgs::msg::Point>> & polygons) const
 {
   const auto & ego_pos = planner_data_->current_odometry->pose.position;
@@ -1027,11 +1027,11 @@ Polygon2d CrosswalkModule::getAttentionArea(
     const auto front_length = length_sum.at(j) - backward_path_length;
     const auto back_length = length_sum.at(j + 1) - backward_path_length;
 
-    if (back_length < attension_range.first) {
+    if (back_length < attention_range.first) {
       continue;
     }
 
-    if (attension_range.second < front_length) {
+    if (attention_range.second < front_length) {
       break;
     }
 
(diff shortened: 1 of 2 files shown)


sasakisasaki:
@satoshi-ota Sorry if I'm misunderstanding. According to [this launch file](https://github.com/tier4/autoware_launch/blob/9a915f8340d6d73786cbac33b355ad89be29aac7/tier4_universe_launch/tier4_planning_launch/launch/scenario_planning/lane_driving/behavior_planning/behavior_planning.launch.xml#L109-L113), it looks we use the plugin that of `experimental`. If so, I guess we also need to fix [this part](https://github.com/autowarefoundation/autoware_universe/blob/be3408346ec6865f3b98f2d175d6bf3911416827/planning/behavior_velocity_planner/autoware_behavior_velocity_crosswalk_module/src/experimental/manager.cpp#L81-L82) too.

PR #12235, opened 2026-03-04
chore(autoware_multi_object_tracker): refactor association interface and flow

technolojin:
## Description
* refactored node and core: 
  - node handles subscription, publication, and callbacks
  - core handles processes
* refactored association interface to contain association results base on UUID
* moved association process after message receive. the association process becomes distributed timings. Iin multi-channel mode, it may reduces overall latency.
* (fix) in case of delay compensation mode, fix pruning time from last publish time to last tracker (update) time

this PR does not change any logic.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

tested with logging-simulator

tracker simulation [TIER IV INTERNAL](https://evaluation.tier4.jp/evaluation/reports/85c8d0d0-79c9-57e4-a928-2c7b0ce00c69?project_id=prd_jt)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


perception/autoware_multi_object_tracker/lib/object_model/types.cpp
```diff
@@ -107,6 +111,16 @@ DynamicObjectList toDynamicObjectList(
   return dynamic_objects;
 }
 
+int DynamicObjectList::getObjectIndexByUuid(const unique_identifier_msgs::msg::UUID & uuid) const
```
ogrex: getObjectIndexByUuid performs a O(n) linear search.   an unordered_map can do O(1) lookup 


some reference code

```
    std::vector<DynamicObject> objects_;
    std::unordered_map<UUID, size_t> uuid_to_index_;  // Built once
```

```
    // O(1) lookup
    std::optional<size_t> getObjectIndexByUuid(const UUID& uuid) const {
        auto it = uuid_to_index_.find(uuid);
        if (it != uuid_to_index_.end()) {
            return it->second;
        }
        return std::nullopt;
    }

```

perception/autoware_multi_object_tracker/src/multi_object_tracker_core.cpp
```diff
@@ -292,9 +235,166 @@ std::optional<autoware_perception_msgs::msg::DetectedObjects> get_merged_objects
...
+{
+  OptionalPublishingData result;
+
+  // Get merged objects
+  if (params.publish_merged_objects) {
+    result.merged_objects = get_merged_objects_(object_time, params, state, logger);
+  }
+
+  // Calculate min_extrapolation_time
+  result.min_extrapolation_time = (object_time - state.last_tracker_time).seconds();
```
YoshiRi: [minor comment]
Previously we had negative check. I think it is better to keep that check.

```
result.min_extrapolation_time = dt > 0.0 ? dt : 0.0;
```

perception/autoware_multi_object_tracker/lib/object_model/types.cpp
technolojin: thank you for advice

implemented with size-check
https://github.com/autowarefoundation/autoware_universe/pull/12235/changes/0c148f0a3a3d63144f21f8de56a7886db943e2e8

perception/autoware_multi_object_tracker/src/multi_object_tracker_core.cpp
technolojin: fixed https://github.com/autowarefoundation/autoware_universe/pull/12235/changes/523b7172a07b51b58f838c845d84a67b45973635

PR #12243, opened 2026-03-05
fix(roi_pointcloud_fusion): add fusion option per class

badai-nguyen:
## Description
- This PR replaces the global “fuse unknown only” flag in ROI pointcloud fusion with per-class fusion options
- Need merging after https://github.com/autowarefoundation/autoware_launch/pull/1783 merged
## Related links

**Parent Issue:**

- [TIERIV internal link](https://tier4.atlassian.net/browse/T4DEV-50154)

## How was this PR tested?

- Run test and compare with based model on [Internal Evaluator](https://evaluation.tier4.jp/evaluation/reports/tables/new?catalog_id=14b1d54b-5c9f-4cbf-a7e1-0eebceb1d30f&filter=&job_id=5cf0cc05-d02c-50de-b43b-182860c8e7ba&project_id=x2_dev&table_config=date&target_ids=32441c3a-f351-5dfa-9320-bc8c49351517): 

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_image_projection_based_fusion/config/roi_pointcloud_fusion.param.yaml
+++ b/perception/autoware_image_projection_based_fusion/config/roi_pointcloud_fusion.param.yaml
@@ -6,7 +6,7 @@
     roi_scale_factor: 1.0
     max_object_size: 2.0
     override_class_with_unknown: false
-    enable_fusion_per_class:
+    fusion_enabled_classes:
       UNKNOWN: true
       CAR: false
       BUS: false
--- a/perception/autoware_image_projection_based_fusion/include/autoware/image_projection_based_fusion/roi_pointcloud_fusion/node.hpp
+++ b/perception/autoware_image_projection_based_fusion/include/autoware/image_projection_based_fusion/roi_pointcloud_fusion/node.hpp
@@ -47,7 +47,7 @@ private:
   int min_cluster_size_{1};
   int max_cluster_size_{20};
   double max_object_size_{2.0};
-  std::unordered_map<uint8_t, bool> enable_fusion_per_class_;
+  std::unordered_map<uint8_t, bool> fusion_enabled_classes_;
   double cluster_2d_tolerance_;
   double roi_scale_factor_{1.0};
   bool override_class_with_unknown_{false};
--- a/perception/autoware_image_projection_based_fusion/src/roi_pointcloud_fusion/node.cpp
+++ b/perception/autoware_image_projection_based_fusion/src/roi_pointcloud_fusion/node.cpp
@@ -48,8 +48,8 @@ RoiPointCloudFusionNode::RoiPointCloudFusionNode(const rclcpp::NodeOptions & opt
      {"BICYCLE", Classification::BICYCLE},
      {"PEDESTRIAN", Classification::PEDESTRIAN}}};
   for (const auto & [class_name, label] : fusion_class_names) {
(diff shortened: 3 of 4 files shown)


perception/autoware_image_projection_based_fusion/src/roi_pointcloud_fusion/node.cpp
```diff
@@ -17,24 +17,35 @@
...
 {
+using autoware::object_recognition_utils::getHighestProbLabel;
 using autoware_utils::ScopedTimeTrack;
 using Classification = autoware_perception_msgs::msg::ObjectClassification;
 
 RoiPointCloudFusionNode::RoiPointCloudFusionNode(const rclcpp::NodeOptions & options)
 : FusionNode<PointCloudMsgType, RoiMsgType, ClusterMsgType>("roi_pointcloud_fusion", options)
 {
-  fuse_unknown_only_ = declare_parameter<bool>("fuse_unknown_only");
+  const std::vector<std::pair<std::string, uint8_t>> fusion_class_names = {
```
ktro2828: [NIT] It seems better to use `std::array` because it's fixed size array.
```suggestion
  const std::array<std::pair<std::string, uint8_t>, 8> fusion_class_names = {
```

perception/autoware_image_projection_based_fusion/include/autoware/image_projection_based_fusion/roi_pointcloud_fusion/node.hpp
```diff
@@ -45,7 +47,7 @@ class RoiPointCloudFusionNode : public FusionNode<PointCloudMsgType, RoiMsgType,
   int min_cluster_size_{1};
   int max_cluster_size_{20};
   double max_object_size_{2.0};
-  bool fuse_unknown_only_{true};
+  std::map<uint8_t, bool> enable_fusion_per_class_;
```
ktro2828: [NIT] I think `std::unordered_map` is sufficient considering how it's used.
```suggestion
  std::unordered_map<uint8_t, bool> enable_fusion_per_class_;
```

perception/autoware_image_projection_based_fusion/config/roi_pointcloud_fusion.param.yaml
```diff
@@ -1,9 +1,17 @@
 /**:
   ros__parameters:
-    fuse_unknown_only: true
     min_cluster_size: 2
     max_cluster_size: 20
     cluster_2d_tolerance: 0.5
     roi_scale_factor: 1.0
     max_object_size: 2.0
     override_class_with_unknown: false
+    enable_fusion_per_class:
```
a-maumau: The name `enable_fusion_per_class` sounds like it has its own toggle option, like:
```
enable_fusion_per_class: true
```

I prefer something like `fusion_target_class` (classes?).

badai-nguyen: > fusion_target_class (classes) 

that seems a better name but it means like a lst of classes will be fused `enable_fusion_per_class = {"CAR", "BUS"..}`, not a togged map for each class, so I prefer `fusion_enabled_classes` and fixed here:https://github.com/autowarefoundation/autoware_universe/pull/12243/changes/021262bc4b32edcd54827d1b2905db20570ded65

PR #12252, opened 2026-03-06
refactor(crosswalk_traffic_light_estimator): extract flashing signal detector and unit tests

interimadd:
## Description
The `crosswalk_traffic_light_estimator` previously handled two primary responsibilities:
1. Estimating crosswalk signals based on vehicle signals.
2. Detecting flashing crosswalk signals.
To improve maintainability and readability, I have decoupled these functions.

In this PR, I refactored the flashing signal detection logic into a standalone `FlashingDetector` class, complete with unit tests.

Key changes:
- Refactored `updateFlashingState`, `updateAndGetColorState`, and `updateLastDetectedSignals` from `node.cpp` into the new `FlashingDetector` class.
- Introduced `estimate_stable_color` as the primary public API to manage history updates, flashing detection, and color resolution.
- Added comprehensive unit tests in `test_flashing_detection.cpp` utilizing the AAA pattern.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

- Added unit tests for `FlashingDetector` in `test_flashing_detection.cpp`:
  - `FirstCall_ReturnsDetectedColor`: First observation returns the detected color as-is
  - `FlashingDetected_UnknownAfterGreen`: UNKNOWN after GREEN triggers flashing, maintains GREEN
  - `FlashingTransition_GreenToRed`: During flashing, RED input transitions state from GREEN to RED
  - `EstimateStableColor_PrunesOldEntries`: Old history entries beyond hold time are pruned, resetting flashing state
  - `ClearState_RemovesTracking`: `clear_state` removes all tracking, subsequent call behaves as first observation
- Built and ran tests with:
  ```
  PACKAGE_NAME=autoware_crosswalk_traffic_light_estimator && colcon build --packages-select $PACKAGE_NAME && colcon test --packages-select $PACKAGE_NAME --event-handlers console_cohesion+
  ```

## Notes for reviewers

- `FlashingDetector` is independent of `rclcpp::Node` and `rclcpp::Logger`. It only relies on `rclcpp::Time` and ROS2 message types.
- The unit tests provide a clear overview of the class's intended behavior and logic.

## Interface changes

None.

## Effects on system behavior

None. This is an internal refactoring only; no changes to external interfaces (topics, parameters).

### DIFF
--- a/.github/workflows/lint-autoware-system-design-format.yml
+++ b/.github/workflows/lint-autoware-system-design-format.yml
@@ -1,0 +1,30 @@
+name: Lint Autoware System Design Format
+
+on:
+  pull_request:
+    paths:
+      - "**/*.node.yaml"
+      - "**/*.module.yaml"
+      - "**/*.system.yaml"
+      - "**/*.parameter_set.yaml"
+  push:
+    branches:
+      - main
+    paths:
+      - "**/*.node.yaml"
+      - "**/*.module.yaml"
+      - "**/*.system.yaml"
+      - "**/*.parameter_set.yaml"
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Lint system design format files
+        uses: autowarefoundation/autoware_system_designer@v0.3.0
+        with:
+          design_files_path: .
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -108,6 +108,12 @@ repos:
         files: .svg$
         additional_dependencies: [prettier@2.7.1, "@prettier/plugin-xml@2.2.0"]
 
+  - repo: https://github.com/autowarefoundation/autoware_system_designer
+    rev: v0.3.0
+    hooks:
+      - id: lint-system-design-format
+        args: [--format=github-actions]
+
   - repo: https://github.com/AleksaC/hadolint-py
     rev: v2.14.0
     hooks:
--- a/perception/autoware_euclidean_cluster/src/euclidean_cluster_node.cpp
+++ b/perception/autoware_euclidean_cluster/src/euclidean_cluster_node.cpp
@@ -53,6 +53,9 @@ void EuclideanClusterNode::onPointCloud(
   // check for empty point cloud
   if (input_msg->data.empty() || input_msg->width == 0 || input_msg->height == 0) {
     RCLCPP_DEBUG(get_logger(), "Empty point cloud received, skipping processing");
+    tier4_perception_msgs::msg::DetectedObjectsWithFeature output;
+    output.header = input_msg->header;
+    cluster_pub_->publish(output);
     return;
   }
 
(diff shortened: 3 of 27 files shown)


perception/autoware_crosswalk_traffic_light_estimator/src/flashing_detection.cpp
```diff
@@ -0,0 +1,155 @@
...
+  return update_and_get_color_state(signal);
+}
+
+bool is_skippable_signal(const TrafficSignal & signal)
+{
+  const auto & elements = signal.elements;
+  if (elements.empty()) return true;
+  // Occluded signal: UNKNOWN with confidence=1
+  return elements.front().color == TrafficSignalElement::UNKNOWN &&
+         elements.front().confidence == 1;
```
a-maumau: I think occluded signal is confidence = 0: https://github.com/autowarefoundation/autoware_universe/blob/cd25ae4cc3b47d5aca3a9bbdcdb3b1b7c78601a8/perception/autoware_traffic_light_occlusion_predictor/src/node.cpp#L173-L175
(It seems original code was doing with `confidence ==  1` ...)

Also this method is doing the same thing as `CrosswalkTrafficLightEstimatorNode::isInvalidDetectionStatus`.
We could move both of them into some shared function.

---
https://github.com/autowarefoundation/autoware_universe/pull/5871 introduce the part of `confidence == 1`, but I didn't find the reason to set this as ==1, so I am considering this as a bug (typo).
At least, recent traffic light classifier's output should be rarely to be unknown with 1.0, I think.
And, even it was intended, I think there is almost no reason to use `confidence == 1` with unknown.

perception/autoware_crosswalk_traffic_light_estimator/include/autoware_crosswalk_traffic_light_estimator/flashing_detection.hpp
```diff
@@ -0,0 +1,71 @@
...
+
+private:
+  void update_signal_history(const TrafficSignal & signal, const rclcpp::Time & current_time);
+  void remove_expired_entries(lanelet::Id id, const rclcpp::Time & current_time);
+  void update_flashing_state(const TrafficSignal & signal);
+  uint8_t update_and_get_color_state(const TrafficSignal & signal);
+
+  FlashingDetectionConfig config_{};
+  std::map<lanelet::Id, bool> is_flashing_;
+  std::map<lanelet::Id, uint8_t> current_color_state_;
```
a-maumau: we don't need the order so we can use `std::unordered_map`.

a-maumau (commented):
Overall, LGTM.
I have some comments. Could you check it?

perception/autoware_crosswalk_traffic_light_estimator/include/autoware_crosswalk_traffic_light_estimator/flashing_detection.hpp
interimadd: Thank you for the suggstion! Addressed in https://github.com/autowarefoundation/autoware_universe/pull/12252/commits/4169ffd16cd1bd8c791db8925e21bc06ce455d21

perception/autoware_crosswalk_traffic_light_estimator/src/flashing_detection.cpp
interimadd: Thanks for the catch!
That could have caused a critical bug, so I'm glad we caught it during the review.
Fixed in https://github.com/autowarefoundation/autoware_universe/pull/12252/commits/13776f2bf67097b1604c55a3db209f68323697a3

interimadd:
@a-maumau 
Thank you for your review!
I have fixed the points you pointed out.
Also, I found an odd behavior where flashing states are kept even after traffic light signal history is cleared. That behavior has been fixed in:
https://github.com/autowarefoundation/autoware_universe/pull/12252/changes/c7c4686e4f84cac8c4e2d41659910585e94507bc

PR #12269, opened 2026-03-10
fix(autoware_trajectory_modifier): use wildcard to prevent namespacing issues

danielsanchezaran:
## Description
The param file has a specific namespace that does not match autoware convention, this PR fixes that and also sets the default value to match the param file for the velocity threshold. 
## Related links

**Parent Issue:**

- Link

## How was this PR tested?
PSim
## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/autoware_trajectory_modifier/schema/trajectory_modifier.schema.json
+++ b/planning/autoware_trajectory_modifier/schema/trajectory_modifier.schema.json
@@ -52,7 +52,7 @@
     }
   },
   "properties": {
-    "/trajectory_modifier": {
+    "/**": {
       "type": "object",
       "properties": {
         "ros__parameters": {
@@ -63,6 +63,6 @@
       "additionalProperties": false
     }
   },
-  "required": ["/trajectory_modifier"],
+  "required": ["/**"],
   "additionalProperties": false
 }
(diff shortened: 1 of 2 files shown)


planning/autoware_trajectory_modifier/src/trajectory_modifier_plugins/stop_point_fixer.cpp
```diff
@@ -116,7 +116,7 @@ void StopPointFixer::set_up_params()
   rcl_interfaces::msg::ParameterDescriptor velocity_desc;
   velocity_desc.description = "Velocity threshold below which ego vehicle is considered stationary";
   params_.velocity_threshold_mps =
-    node->declare_parameter<double>("stop_point_fixer.velocity_threshold_mps", 0.1, velocity_desc);
+    node->declare_parameter<double>("stop_point_fixer.velocity_threshold_mps", 0.25, velocity_desc);
```
yukkysaito, line 119: Is this correct parameter?

danielsanchezaran: Yes, that is the default parameter on the param file. It might be changed in the future if tuning is required

PR #12272, opened 2026-03-10
feat(autoware_calibration_status_classifier): add vehicle's ego raycasting filter

amadeuszsz:
## Description

Add per-pair ego vehicle occlusion mask via CPU raycasting to prevent LiDAR points passing through the ego chassis from being projected onto the camera image.

Changes:

- New `ego_box` parameter [x_min, y_min, z_min, x_max, y_max, z_max] in lidar frame (all zeros = disabled).
- `generate_ego_occlusion_mask()` computes a binary image mask using slab intersection raycasting from the camera center through each pixel against the ego bounding box.
- If the camera is geometrically inside the box (roof-mounted), the nearest X/Y walls are pushed past the camera so only far walls occlude.
- Mask is generated once per camera-lidar pair, allocated directly on GPU, and reused with zero-copy on subsequent frames.
- CUDA kernels (project_points_kernel, dilate_kernel) skip masked pixels to filter ego-chassis projections and prevent dilation bleed.

## Related links

**Parent Issue:**

- Link

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/sensing/autoware_calibration_status_classifier/lib/calibration_status_classifier.cpp
+++ b/sensing/autoware_calibration_status_classifier/lib/calibration_status_classifier.cpp
@@ -152,7 +152,8 @@ CalibrationStatusClassifierResult CalibrationStatusClassifier::process(
   const uint8_t * ego_mask_ptr = nullptr;
   if (!ego_box_.empty()) {
     const std::string mask_key =
-      camera_lidar_info.camera_frame_id + "_" + camera_lidar_info.lidar_frame_id;
+      camera_lidar_info.camera_frame_id + "_" + camera_lidar_info.lidar_frame_id + "_" +
+      std::to_string(image_msg->width) + "x" + std::to_string(image_msg->height);
 
     auto it = ego_masks_d_.find(mask_key);
     if (it == ego_masks_d_.end()) {  // lazy init, only generate mask if not already stored


sensing/autoware_calibration_status_classifier/config/ml_package_calibration_status_classifier.param.yaml
```diff
@@ -4,3 +4,4 @@
     dilation_size: 1
     height: [1080, 1860, 2160]  # [min, opt, max]
     width: [1920, 2880, 3840]  # [min, opt, max]
+    ego_box: [-1.53, -1.15, -0.1, 5.72, 1.15, 3.1]
```
n-patiphon: nit: adding a comment at the end of the line could help the readers and prevent an accidental misconfigure. `[x_min, y_min, z_min, x_max, y_max, z_max]`

sensing/autoware_calibration_status_classifier/schema/ml_package_calibration_status_classifier.schema.json
```diff
@@ -35,9 +35,19 @@
...
           "default": [1920, 2880, 3840]
+        },
+        "ego_box": {
+          "type": "array",
+          "items": {
+            "type": "number"
+          },
+          "minItems": 6,
+          "maxItems": 6,
+          "description": "Ego vehicle bounding box in lidar frame [x_min, y_min, z_min, x_max, y_max, z_max]. Used to generate an occlusion mask that prevents projecting LiDAR points through the vehicle chassis onto the image. Set all values to 0.0 to disable.",
```
n-patiphon: nit: Is this `base_link` frame?

sensing/autoware_calibration_status_classifier/src/calibration_status_classifier_node.cpp
```diff
@@ -49,7 +49,8 @@ CalibrationStatusClassifierNode::CalibrationStatusClassifierNode(
   const CalibrationStatusClassifierConfig calibration_status_classifier_config(
     this->declare_parameter<double>("max_depth"), this->declare_parameter<int64_t>("dilation_size"),
     this->declare_parameter<std::vector<int64_t>>("height"),
-    this->declare_parameter<std::vector<int64_t>>("width"));
+    this->declare_parameter<std::vector<int64_t>>("width"),
+    this->declare_parameter<std::vector<double>>("ego_box"));
```
n-patiphon: Wouldn't this make `ego_box` mandatory and not optional config?

sensing/autoware_calibration_status_classifier/test/test_preprocessing.cpp
```diff
@@ -141,7 +141,7 @@ TEST_F(PreprocessingTest, TestPreprocessing)
     CHECK_CUDA_ERROR(preprocess_ptr->project_points_launch(
       cloud_d.get(), image_undistorted_d.get(), tf_matrix_d.get(), projection_matrix_d.get(),
       sample.pointcloud->width * sample.pointcloud->height, sample.image_bgr->width,
-      sample.image_bgr->height, in_d.get(), num_points_projected_d.get()));
+      sample.image_bgr->height, in_d.get(), num_points_projected_d.get(), nullptr));
```
n-patiphon, line 144: note: this means the `ego_mask` is excluded from the test.

sensing/autoware_calibration_status_classifier/src/calibration_status_classifier_node.cpp
n-patiphon: Based on recent commit, I think the `ego_box` were movvd from model param to node param. Would this need to be updated as `input.ego_box`

sensing/autoware_calibration_status_classifier/lib/calibration_status_classifier.cpp
```diff
@@ -129,11 +133,30 @@ CalibrationStatusClassifierResult CalibrationStatusClassifier::process(
...
+  // Generate ego occlusion mask on first call per pair (lazy init)
+  const uint8_t * ego_mask_ptr = nullptr;
+  if (!config_.ego_box.empty()) {
+    const std::string mask_key =
+      camera_lidar_info.camera_frame_id + "_" + camera_lidar_info.lidar_frame_id;
+
+    auto it = ego_masks_d_.find(mask_key);
+    if (it == ego_masks_d_.end()) {  // lazy init, only generate mask if not already stored
+      auto mask =
+        generate_ego_occlusion_mask(camera_lidar_info, image_msg->width, image_msg->height);
```
n-patiphon, line 161: This is just a question out of my curiosity. Apologies for the lack of my knowledge about our system.

There's no way that the first image from camera will have weird resolution, e.g., empty message with 0 width and heigh or anything like that, right?

The reason I'm wondering is that if the first frame is somehow corrupted, this mask will forever be based on that first frame.

sensing/autoware_calibration_status_classifier/config/ml_package_calibration_status_classifier.param.yaml
n-patiphon: This is outdated in the current HEAD. Yet, I think some kind of inline comments would still be helpful.

amadeuszsz: Yeah, parameter has been moved already, sorry for late commit

sensing/autoware_calibration_status_classifier/schema/ml_package_calibration_status_classifier.schema.json
amadeuszsz: Nope, we read input `frame_id` attached to sensor. For concatenated point cloud it can be `base_link`.

sensing/autoware_calibration_status_classifier/test/test_preprocessing.cpp
amadeuszsz: This was intentionally 😄

sensing/autoware_calibration_status_classifier/src/calibration_status_classifier_node.cpp
amadeuszsz: Refined parameter placement in https://github.com/autowarefoundation/autoware_universe/pull/12272/commits/ed148b53ae6a7aad95738fbdc9ec5e4600f033f9

sensing/autoware_calibration_status_classifier/lib/calibration_status_classifier.cpp
amadeuszsz: AFAIR, camera_info topic is not composed dynamically, but only during initialization, using [camera info manager](https://github.com/ros-perception/image_common/tree/humble/camera_info_manager) where configuration file defines content of this topic.
Maybe there is a use case where these parameters are redefined during the runtime. If this will become an issue, we can add extra check of camera info content to see if any update occurred.

n-patiphon: I see. Correct me if I'm wrong, but `image_msg->width` is actually the image width from current message not from `camera_info`. 

But I agree, it's not likely that this issue will occur. Only if you want to be extra careful, one easy guard is to add resolution to cache key:

```
const std::string mask_key =
  camera_lidar_info.camera_frame_id + "_" + camera_lidar_info.lidar_frame_id + "_" +
  std::to_string(image_msg->width) + "x" + std::to_string(image_msg->height);
```

To summarize, it's ok to leave as is. If you're going to modify, I would propose the suggestion above. 💪

n-patiphon (approved):
The changes look good to me!

sensing/autoware_calibration_status_classifier/lib/calibration_status_classifier.cpp
amadeuszsz: Thanks, added it https://github.com/autowarefoundation/autoware_universe/pull/12272/commits/62b025d10153d29c035c3bf247cfc132dc129ce8

PR #12278, opened 2026-03-11
fix(behavior_path_side_shift): preserve shifted path shape to prevent chattering

k-uta:
# fix(behavior_path_side_shift): preserve shifted path shape to prevent chattering

## Description

This PR addresses an issue where the side shift path chatters every planning cycle even without a new shift request. 

Previously, the side shift path was regenerated every cycle via `path_shifter_.generate()`, which depends on the reference path resampled from the upstream module output. When the upstream output fluctuated (e.g., due to object detection chattering), the shifted path shape changed constantly.

To resolve this, this commit introduces the `keepPrevPathShape()` method to retain the previously generated shifted path shape when no new shift line is inserted via `replaceShiftLine()`. 

**Details of changes:**
* Added the `keepPrevPathShape()` method.
* The method trims already-traversed points behind the ego position, keeping a small backward margin of 3 points for continuity.
* It appends new forward points from the freshly generated path to extend the path coverage.
* Updated the `plan()` function to apply this logic using a `shift_line_replaced` boolean flag.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

* Evaluated side shift behavior in simulation to confirm the path shape no longer chatters.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

The side shift path will remain stable and retain its shape during planning cycles when no new shift request is made, eliminating unwanted steering or path tracking chattering.


planning/behavior_path_planner/autoware_behavior_path_side_shift_module/src/scene.cpp
```diff
@@ -459,6 +468,46 @@ PathWithLaneId SideShiftModule::extendBackwardLength(const PathWithLaneId & orig
...
 
+ShiftedPath SideShiftModule::keepPrevPathShape(const ShiftedPath & new_shifted_path) const
+{
+  const auto & ego_position = getEgoPose().position;
+
+  // Find the ego nearest segment in the previous output path
+  const size_t prev_ego_seg_idx = findNearestSegmentIndex(prev_output_.path.points, ego_position);
+
+  // Trim already-traversed portion, keeping a small backward margin for continuity
+  constexpr size_t backward_margin_points = 3;
```
TaikiYamada4: Could you use `backward_path_length` and calculate length-wise?
https://github.com/autowarefoundation/autoware_universe/blob/e0db0e0904f746617c4630bc0a76595573c355ce/planning/behavior_path_planner/autoware_behavior_path_planner/config/behavior_path_planner.param.yaml#L11

planning/behavior_path_planner/autoware_behavior_path_side_shift_module/src/scene.cpp
```diff
@@ -459,6 +468,46 @@ PathWithLaneId SideShiftModule::extendBackwardLength(const PathWithLaneId & orig
...
+
+  // Copy remaining (not-yet-traversed) points from the previous output to preserve shape
+  for (size_t i = trim_start_idx; i < prev_output_.path.points.size(); ++i) {
+    result.path.points.push_back(prev_output_.path.points.at(i));
+    result.shift_length.push_back(prev_output_.shift_length.at(i));
+  }
+
+  if (result.path.points.empty()) {
+    return new_shifted_path;
+  }
```
TaikiYamada4: Isn't this equivalent to `prev_output_.path.points.empty()`?
If so, we can bring this earlier to reduce computation.

planning/behavior_path_planner/autoware_behavior_path_side_shift_module/src/scene.cpp
```diff
@@ -265,20 +265,28 @@ void SideShiftModule::replaceShiftLine()
...
     RCLCPP_DEBUG(getLogger(), "change is not requested");
   }
 
-  // Refine path
+  // Preserve the shape during shifting by reusing the last generated path unless re-generation
+  // is explicitly required.
   ShiftedPath shifted_path;
-  path_shifter_.generate(&shifted_path);
+  if (should_regenerate_shifted_path || prev_output_.path.points.empty()) {
+    path_shifter_.generate(&shifted_path);
```
TaikiYamada4, line 287: @k-uta 
Maybe bringing `replaceShiftLine()` to here might be better?

k-uta: Thank you for the suggestion! I totally agree, moving it here makes the logic more cohesive and easier to read.
To preserve the original behavior, I will move `replaceShiftLine()` inside the if block, guarded by `if (should_regenerate_shifted_path)`.

Fixed in [ce8c3e9](https://github.com/autowarefoundation/autoware_universe/pull/12278/commits/ce8c3e9e0b4f0cf232eb74a8cd64c3b21ff61221)

PR #12286, opened 2026-03-12
feat(autoware_calibration_status_classifier): add angular velocity filter and refine filters design

amadeuszsz:
## Description

- Add `AngularVelocityFilter` that blocks calibration check when angular velocity axis exceeds a threshold.
- Make the filter library ROS-agnostic which can be reused in the downstream module. 
- Overall refactor for standardized filters integration.
- Add per-pair thresholds.
- Update config, schema, launch, README, and tests accordingly.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/sensing/autoware_calibration_status_classifier/include/autoware/calibration_status_classifier/calibration_status_classifier_filters.hpp
+++ b/sensing/autoware_calibration_status_classifier/include/autoware/calibration_status_classifier/calibration_status_classifier_filters.hpp
@@ -18,6 +18,7 @@
 #include "autoware/calibration_status_classifier/filter.hpp"
 
 #include <memory>
+#include <mutex>
 #include <shared_mutex>
 #include <utility>
 #include <vector>
@@ -35,8 +36,8 @@ struct FiltersResult
 /**
  * @brief Thread-safe container for prerequisite filters
  *
- * Manages a collection of Filter instances with thread-safe update (shared lock)
- * and evaluate (exclusive lock) operations. Filters are stored as unique_ptr<Filter>
+ * Manages a collection of Filter instances with thread-safe update (exclusive lock)
+ * and evaluate (shared lock) operations. Filters are stored as unique_ptr<Filter>
  * and accessed by type via dynamic_cast.
  */
 class CalibrationStatusClassifierFilters
@@ -60,11 +61,11 @@ public:
 
   explicit CalibrationStatusClassifierFilters(std::vector<std::unique_ptr<Filter>> filters);
 
-  /// Thread-safe update: shared lock (concurrent updates to different filters are safe)
+  /// Thread-safe update: exclusive lock (writes filter state)
   template <typename FilterT, typename... Args>
   bool update(Args &&... args)
   {
-    std::shared_lock lock(mutex_);
+    std::unique_lock lock(mutex_);
     if (auto * f = get_impl<FilterT>()) {
       f->update(std::forward<Args>(args)...);
       return true;
@@ -72,7 +73,7 @@ public:
     return false;
   }
 
-  /// Thread-safe evaluate: exclusive lock (consistent snapshot of all filter states)
+  /// Thread-safe evaluate: shared lock (read-only access to filter states)
   [[nodiscard]] FiltersResult evaluate(double stamp) const;
 
   /// Non-locked accessor - use only during single-threaded setup (e.g. constructor)
--- a/sensing/autoware_calibration_status_classifier/lib/calibration_status_classifier_filters.cpp
+++ b/sensing/autoware_calibration_status_classifier/lib/calibration_status_classifier_filters.cpp
@@ -14,7 +14,7 @@
 
 #include "autoware/calibration_status_classifier/calibration_status_classifier_filters.hpp"
 
-#include <mutex>
+#include <shared_mutex>
 #include <utility>
 #include <vector>
 
@@ -47,7 +47,7 @@ CalibrationStatusClassifierFilters::CalibrationStatusClassifierFilters(
 
 FiltersResult CalibrationStatusClassifierFilters::evaluate(double stamp) const
 {
-  std::unique_lock lock(mutex_);
+  std::shared_lock lock(mutex_);
   FiltersResult result;
   result.all_passed = true;
   result.filter_results.reserve(filters_.size());


sensing/autoware_calibration_status_classifier/include/autoware/calibration_status_classifier/filter.hpp
```diff
@@ -0,0 +1,211 @@
...
+  ThresholdFilter(const ThresholdFilter &) = delete;
+  ThresholdFilter & operator=(const ThresholdFilter &) = delete;
+  ThresholdFilter(ThresholdFilter &&) = delete;
+  ThresholdFilter & operator=(ThresholdFilter &&) = delete;
+
+  void update(const ValueT value, const double stamp)
+  {
+    current_value_ = value;
+    last_update_time_ = stamp;
+    if (last_exceeded_time_ > stamp) {
```
n-patiphon: `ThresholdFilter` currently measures recovery from the last threshold exceed, not from when the signal actually drops back below threshold. It won't make that much of a difference with high frequency incoming messages, but it contradicts with the comment:

```
the filter remains blocked until it has been continuously below the threshold
 * for at least `timeout_sec` seconds.
```

sensing/autoware_calibration_status_classifier/include/autoware/calibration_status_classifier/calibration_status_classifier_filters.hpp
```diff
@@ -0,0 +1,116 @@
...
+    (filters_.emplace_back(std::move(filters)), ...);
+  }
+
+  explicit CalibrationStatusClassifierFilters(std::vector<std::unique_ptr<Filter>> filters);
+
+  /// Thread-safe update: shared lock (concurrent updates to different filters are safe)
+  template <typename FilterT, typename... Args>
+  bool update(Args &&... args)
+  {
+    std::shared_lock lock(mutex_);
```
n-patiphon: Correct me if I'm wrong here, but isn't `shared_lock` safe if it's read-only?  
Would `f->update(xxx)` breaks this assumption and can potentially cause a data update race?

sensing/autoware_calibration_status_classifier/CMakeLists.txt
```diff
@@ -60,6 +60,17 @@ if(TRT_AVAIL AND CUDA_AVAIL)
     set(CUDA_NVCC_FLAGS "-g -G")
   endif()
 
+  ament_auto_add_library(${PROJECT_NAME}_filters_lib SHARED
```
n-patiphon: 🤖(reviewer)
This new filters library is introduced as reusable and ROS agnostic, but it is still built only inside the TRT_AVAIL && CUDA_AVAIL block. In CPU-only or TensorRT-less builds there is no filters_lib target to reuse, which undercuts the stated downstream reuse goal.

sensing/autoware_calibration_status_classifier/include/autoware/calibration_status_classifier/calibration_status_classifier_filters.hpp
```diff
@@ -0,0 +1,116 @@
...
+#include <memory>
+#include <shared_mutex>
+#include <utility>
+#include <vector>
+
+namespace autoware::calibration_status_classifier
+{
+
+/// @brief Aggregated result of evaluating all prerequisite filters
+struct FiltersResult
```
n-patiphon, line 30: nit: Shouldn't this be `FilterResults`

amadeuszsz: Oh, I'm not linguistic expert, so please let me use LLM to decide which combination of struct name (`FilterResults`, `FiltersResult`, `FiltersResults`) + class' member (`filter_results`, `filters_result`, `filters_results`) is most appropriate:
> For this specific use case, the most natural naming pair is:
> 
> Struct: `FiltersResult`
> Member vector: `filter_results`
> Why this is linguistically and semantically consistent:
> 
> 1. Result of an aggregate operation
> - The struct represents one aggregate evaluation outcome of the whole filter set (one call to evaluate), not multiple independent result objects.
> - So singular Result at the end is appropriate: it is one container/result object.
> 2. Plural modifier vs plural head noun
> - In English compound naming, the left word often acts as a classifier/modifier.
> - `FiltersResult` reads as “result for filters” (aggregate context = filters).
> - `FilterResults` would imply “multiple filter results” as the primary meaning, which is more suitable for the vector itself, not the wrapper object.
> 3. Member naming should reflect contents directly
> - The vector stores many per-filter entries, so plural result(s) is correct.
> - `filter_results` is ideal: it means “results of filters” and matches the element type `FilterResult`.
> - `filters_result` is awkward grammatically (“filters result” as singular phrase), and `filters_results` is usually redundant/noisier in code style.
> 4. Type-to-member symmetry
> - Element type: `FilterResult` (single item)
> - Collection member: `filter_results` (many items)
> - Aggregate wrapper: `FiltersResult` (single summary object containing many item results + `all_passed`)

The current naming convention seems to be reasonable.

sensing/autoware_calibration_status_classifier/include/autoware/calibration_status_classifier/calibration_status_classifier_filters.hpp
amadeuszsz: Good point! Theoretically if we will use multi-threading in near future in this node, each thread would be dedicated for independent vector's index (filter), so writing is thread-safe. The problem is with reading, where we can encounter race condition. The optimal solution would be to have independent mutexes per each filter. However, this will make implementation a bit messy and in my opinion it is better to have strict locks, as performance loss will be negligible.
Addressed in https://github.com/autowarefoundation/autoware_universe/pull/12286/commits/c606014ed50436166eeee2f589af26594ec19bee

sensing/autoware_calibration_status_classifier/include/autoware/calibration_status_classifier/filter.hpp
amadeuszsz: Not sure if understood your concerns correctly, but here is clearer logic: https://github.com/autowarefoundation/autoware_universe/pull/12286/commits/9d243027da48317ac786e176aa6755dbb2904215

n-patiphon: Yes, this fixes the off-by-one issue. However, now that you mention preventing time going backward. I think the solution may be incomplete.

Correct me if I'm wrong, but if an older stamped message arrives after recovery already started, this clears `recovery_start_time_`, and `is_passed()` may become `true` too early. 

For example, if recovery started at `t=101` with a `2.0s` timeout, a late below-threshold message stamped `t=100.5` can clear the recovery timer, and then `is_passed(102)` may return `true` even though it should still be waiting until `103`.

Would it be safer to just ignore stale samples? Something like this?

```
void update(const ValueT value, const double stamp)
  {
    if (last_update_time_.has_value() && stamp < last_update_time_.value())
    {
      return;
    }
  
    current_value_ = value;
    last_update_time_ = stamp;
    ....
```

amadeuszsz: > if an older stamped message arrives after recovery already started, this clears recovery_start_time_, and is_passed() may become true too early.

Yes, but the time does not go backward. If older stamp occurred, that means the rosbag has been loaded again, therefore we want to treat this input as a first sample. Note that filter is handled by a single source (callback) so you don't expect stamp misalignment as it could occur if you had two sources for single filter.


Your proposed change provides runtime issue. Consider you play rosbag and you replay again. If threshold exceed before rosbag restarted, you will never recover from latest saved stamp.

PR #12287, opened 2026-03-12
feat(autoware_trajectory_optimizer): force 3 point trajectory

danielsanchezaran:
force the optimizer to output a 3 point trajectory with 0 velocity if the output trajectory has less than 3 points. This is because Autoware's downstream nodes cannot properly cope with 2 or fewer point trajectories. 

## Related links

**Parent Issue:**

- Link

## How was this PR tested?
PSim
## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

Instead of a 2 point trajectory a 3 point trajectory with 0 acceleration and speed along the ego axis is returned.

### DIFF
--- a/planning/autoware_trajectory_optimizer/src/utils.cpp
+++ b/planning/autoware_trajectory_optimizer/src/utils.cpp
@@ -36,14 +36,37 @@ rclcpp::Logger get_logger()
   return rclcpp::get_logger("trajectory_optimizer");
 }
 
+void log_debug_throttle(const std::string & message)
+{
+  auto clock = rclcpp::Clock::make_shared(RCL_ROS_TIME);
+  RCLCPP_DEBUG_THROTTLE(get_logger(), *clock, 5000, "%s", message.c_str());
+}
+
+void log_error_throttle(const std::string & message)
+{
+  auto clock = rclcpp::Clock::make_shared(RCL_ROS_TIME);
+  RCLCPP_ERROR_THROTTLE(get_logger(), *clock, 5000, "%s", message.c_str());
+}
+
+void log_warn_throttle(const std::string & message)
+{
+  auto clock = rclcpp::Clock::make_shared(RCL_ROS_TIME);
+  RCLCPP_WARN_THROTTLE(get_logger(), *clock, 5000, "%s", message.c_str());
+}
+
 TrajectoryPoints generate_three_point_stopped_trajectory(
   const TrajectoryPoints & input_traj, const Odometry & odom)
 {
   // ego vehicle is moving, returning a dummy trajectory might be dangerous
   constexpr double moving_threshold_mps{0.1};
   if (odom.twist.twist.linear.x > moving_threshold_mps) {
+    log_debug_throttle(
+      "Skip generating three point stopped trajectory since the vehicle is moving");
     return input_traj;
   }
+
+  log_debug_throttle("Generated three-point stopped trajectory");
+
   TrajectoryPoint base_link_point;
   base_link_point.time_from_start = rclcpp::Duration::from_seconds(0.);
   base_link_point.pose = odom.pose.pose;
@@ -61,18 +84,6 @@ TrajectoryPoints generate_three_point_stopped_trajectory(
   offset_point_2.pose = autoware_utils_geometry::calc_offset_pose(
     base_link_point.pose, /*offset_x*/ offset_x2, /*offset_y*/ 0., /*offset_z*/ 0.);
   return {base_link_point, offset_point_1, offset_point_2};
-}
-
-void log_error_throttle(const std::string & message)
-{
-  auto clock = rclcpp::Clock::make_shared(RCL_ROS_TIME);
-  RCLCPP_ERROR_THROTTLE(get_logger(), *clock, 5000, "%s", message.c_str());
-}
-
-void log_warn_throttle(const std::string & message)
-{
-  auto clock = rclcpp::Clock::make_shared(RCL_ROS_TIME);
-  RCLCPP_WARN_THROTTLE(get_logger(), *clock, 5000, "%s", message.c_str());
 }
 
 void smooth_trajectory_with_elastic_band(


planning/autoware_trajectory_optimizer/include/autoware/trajectory_optimizer/utils.hpp
```diff
@@ -83,6 +83,12 @@ void copy_trajectory_orientation(
  */
 rclcpp::Logger get_logger();
 
+/**
+ * @brief generates a 3 point trajectory to prevent downstream issues
```
go-sakayori, line 87: nits: Can we have the `param` written too

planning/autoware_trajectory_optimizer/src/trajectory_optimizer.cpp
```diff
@@ -186,6 +188,10 @@ void TrajectoryOptimizer::on_traj([[maybe_unused]] const CandidateTrajectories::
     for (auto & plugin : plugins_) {
       plugin->optimize_trajectory(trajectory.points, params_, data);
     }
+    if (trajectory.points.size() < 3) {
```
yukkysaito, line 194: It would be helpful to include a brief comment explaining why three points are necessary.

planning/autoware_trajectory_optimizer/src/utils.cpp
```diff
@@ -33,6 +36,30 @@ rclcpp::Logger get_logger()
...
+  base_link_point.longitudinal_velocity_mps = 0.0;
+  base_link_point.acceleration_mps2 = 0.0;
+
+  TrajectoryPoint offset_point_1 = base_link_point;
+  TrajectoryPoint offset_point_2 = base_link_point;
+
+  offset_point_1.pose =
+    autoware_utils_geometry::calc_offset_pose(base_link_point.pose, 0.5, 0., 0.);
+  offset_point_2.pose =
+    autoware_utils_geometry::calc_offset_pose(base_link_point.pose, 1.0, 0., 0.);
```
yukkysaito: How about this?
```suggestion
  offset_point_1.pose =
    autoware_utils_geometry::calc_offset_pose(base_link_point.pose, /*offset_x*/ 0.5, /*offset_y*/ 0., /*offset_z*/ 0.);
  offset_point_2.pose =
    autoware_utils_geometry::calc_offset_pose(base_link_point.pose, /*offset_x*/ 1.0, /*offset_y*/ 0., /*offset_z*/ 0.);
```

planning/autoware_trajectory_optimizer/src/utils.cpp
```diff
@@ -33,6 +36,30 @@ rclcpp::Logger get_logger()
   return rclcpp::get_logger("trajectory_optimizer");
 }
 
+TrajectoryPoints generate_three_point_stopped_trajectory(
+  const TrajectoryPoints & input_traj, const Odometry & odom)
+{
+  constexpr double moving_threshold{0.1};
+  if (odom.twist.twist.linear.x > moving_threshold) {
+    // ego vehicle is moving
+    return input_traj;
```
go-sakayori, line 65: This return a trajectory with less than 3 points. Is it okay?

planning/autoware_trajectory_optimizer/src/utils.cpp
```diff
@@ -33,6 +36,30 @@ rclcpp::Logger get_logger()
...
+  base_link_point.acceleration_mps2 = 0.0;
+
+  TrajectoryPoint offset_point_1 = base_link_point;
+  TrajectoryPoint offset_point_2 = base_link_point;
+
+  offset_point_1.pose =
+    autoware_utils_geometry::calc_offset_pose(base_link_point.pose, 0.5, 0., 0.);
+  offset_point_2.pose =
+    autoware_utils_geometry::calc_offset_pose(base_link_point.pose, 1.0, 0., 0.);
+  return {base_link_point, offset_point_1, offset_point_2};
```
go-sakayori, line 86: You are overwriting the original trajectory, but is it okay?

planning/autoware_trajectory_optimizer/src/utils.cpp
danielsanchezaran: I dont want to ouput a sudden stop trajectory when the ego is in movement.

planning/autoware_trajectory_optimizer/src/utils.cpp
danielsanchezaran: if the original trajectory has less than 3 point we need to send a 3 point one, no point in keeping the original

planning/autoware_trajectory_optimizer/src/trajectory_optimizer.cpp
danielsanchezaran: I will add a comment.

planning/autoware_trajectory_optimizer/include/autoware/trajectory_optimizer/utils.hpp
danielsanchezaran: Done in 20bd353

planning/autoware_trajectory_optimizer/src/utils.cpp
danielsanchezaran: Done in 20bd353 , though I personally believe it causes visual pollution

planning/autoware_trajectory_optimizer/src/trajectory_optimizer.cpp
danielsanchezaran: added a comment in 20bd353

planning/autoware_trajectory_optimizer/src/utils.cpp
go-sakayori: okay. What do you think of filling the time_from_start field with other values?

danielsanchezaran: what would they mean? isnt it more dangerous if I add time_from_Start values that are impossible?

go-sakayori: Thats true my bad

PR #12299, opened 2026-03-13
feat(pid_longitudinal_controller): migrate pid_longitudinal_controller to use continous Trajectory

mraditya01:
## Description

Converted `autoware_pid_longitudinal_controller` from discrete `autoware_planning_msgs::msg::Trajectory` handling to continuous `TrajectoryExperimental` handling.

The controller still receives `autoware_planning_msgs::msg::Trajectory`, then converts it to `TrajectoryExperimental` in `run()` and uses the continuous path internally. 

### Migration Status
| Function Name | Location/File | Status |
|:--------------|:--------------|:-------|
| `run` | `pid_longitudinal_controller.cpp` | ✅ Done - switched live runtime path to continuous workflow with discrete fallback |
| `setTrajectory` | `pid_longitudinal_controller.cpp` | ✅ Done - added `TrajectoryExperimental` |
| `getControlData` | `pid_longitudinal_controller.cpp` | ✅ Done - added `getExperimentalControlData(...)` |
| `updateControlState` | `pid_longitudinal_controller.cpp` | ✅ Done - added `ExperimentalControlData` |
| `calcCtrlCmd` | `pid_longitudinal_controller.cpp` | ✅ Done - added `ExperimentalControlData` |
| `publishDebugData` | `pid_longitudinal_controller.cpp` | ✅ Done - added `ExperimentalControlData` |
| `getCurrentShift` | `pid_longitudinal_controller.cpp` | ✅ Done - added continuous  |
| `keepBrakeBeforeStop` | `pid_longitudinal_controller.cpp` | ✅ Done - added continuous |
| `applyVelocityFeedback` | `pid_longitudinal_controller.cpp` | ✅ Done - added continuous |
| `updateDebugVelAcc` | `pid_longitudinal_controller.cpp` | ✅ Done - added continuous |
| `buildTrajectoryWithSamples` | `pid_longitudinal_controller.cpp` | ✅ Done - rebuilds continuous trajectory from sampled points using `pretty_build(...)` |
| `calcStopDistance` | `longitudinal_controller_utils.cpp` | ✅ Done - added continuous |
| `getPitchByTraj` | `longitudinal_controller_utils.cpp` | ✅ Done - added continuous |

## Related links

**Related PR:**  
- None

## How was this PR tested?
[4288/4289 (unrelated failed test)
](https://evaluation.tier4.jp/evaluation/reports/de195d58-c229-50d5-9489-85e02290fbca?project_id=autoware_dev)


control/autoware_pid_longitudinal_controller/src/longitudinal_controller_utils.cpp
```diff
@@ -53,6 +57,62 @@ bool isValidTrajectory(const Trajectory & traj)
   return true;
 }
 
+bool isValidTrajectory(const TrajectoryExperimental & trajectory)
```
soblin: You don't need this function, because the TrajectoryClass itself validates that in the constructor !

control/autoware_pid_longitudinal_controller/src/longitudinal_controller_utils.cpp
```diff
@@ -53,6 +57,62 @@ bool isValidTrajectory(const Trajectory & traj)
...
+
+    if (base < -eps || base > trajectory_length + eps) {
+      return false;
+    }
+  }
+
+  return true;
+}
+
+double findFirstNearestIndexWithSoftConstraints(
```
soblin: You can use `autoware::trajectory::find_first_nearest_index` so you won't need this

control/autoware_pid_longitudinal_controller/src/longitudinal_controller_utils.cpp
```diff
@@ -71,6 +131,21 @@ double calcStopDistance(
...
+double calcStopDistance(
+  const Pose & current_pose, const TrajectoryExperimental & traj, const double max_dist,
+  const double max_yaw)
+{
+  const double nearest_s =
+    findFirstNearestIndexWithSoftConstraints(traj, current_pose, max_dist, max_yaw);
+
+  const auto stop_s = autoware::experimental::trajectory::search_zero_velocity_position(traj);
+  if (!stop_s) {
+    return traj.length() - nearest_s;
```
soblin: Shouldn't we return 0.0 instead ?
https://github.com/autowarefoundation/autoware_universe/blob/main//control/autoware_pid_longitudinal_controller/src/longitudinal_controller_utils.cpp?plain=1#L68

control/autoware_pid_longitudinal_controller/src/pid_longitudinal_controller.cpp
```diff
@@ -571,6 +624,92 @@ PidLongitudinalController::ControlData PidLongitudinalController::getControlData
   return control_data;
 }
 
+std::optional<PidLongitudinalController::ExperimentalControlData>
+PidLongitudinalController::getExperimentalControlData(const geometry_msgs::msg::Pose & current_pose)
```
soblin, line 474: NOTE: Seems OK from my side

control/autoware_pid_longitudinal_controller/src/pid_longitudinal_controller.cpp
```diff
@@ -796,6 +935,156 @@ void PidLongitudinalController::updateControlState(const ControlData & control_d
   return;
 }
 
+void PidLongitudinalController::updateControlState(const ExperimentalControlData & control_data)
```
soblin: NOTE: This would be duplicate

control/autoware_pid_longitudinal_controller/src/pid_longitudinal_controller.cpp
```diff
@@ -906,6 +1190,101 @@ PidLongitudinalController::Motion PidLongitudinalController::calcCtrlCmd(
   return ctrl_cmd_as_pedal_pos;
 }
 
+PidLongitudinalController::Motion PidLongitudinalController::calcCtrlCmd(
```
soblin, line 744: NOTE: Seems OK from my side

soblin (approved):

control/autoware_pid_longitudinal_controller/include/autoware/pid_longitudinal_controller/longitudinal_controller_utils.hpp
```diff
@@ -15,44 +15,26 @@
...
 using geometry_msgs::msg::Pose;
 using geometry_msgs::msg::Quaternion;
 
-/**
- * @brief check if trajectory is invalid or not
- */
-bool isValidTrajectory(const Trajectory & traj);
-
 /**
  * @brief calculate distance to stopline from current vehicle position where velocity is 0
```
go-sakayori, line 33: nits: Other functions have `@param` filed so would be better to align

control/autoware_pid_longitudinal_controller/include/autoware/pid_longitudinal_controller/pid_longitudinal_controller.hpp
```diff
@@ -288,11 +292,8 @@ class PidLongitudinalController : public trajectory_follower::LongitudinalContro
   trajectory_follower::LongitudinalOutput run(
     trajectory_follower::InputData const & input_data) override;
 
-  /**
-   * @brief calculate data for controllers whose type is ControlData
-   * @param [in] current_pose current ego pose
-   */
-  ControlData getControlData(const geometry_msgs::msg::Pose & current_pose);
+  std::optional<ExperimentalControlData> getExperimentalControlData(
```
go-sakayori, line 301: nits: please add doxygen

go-sakayori (approved):

control/autoware_pid_longitudinal_controller/src/longitudinal_controller_utils.cpp
```diff
@@ -14,61 +14,39 @@
...
-#include <limits>
 #include <utility>
 
 namespace autoware::motion::control::pid_longitudinal_controller
 {
 namespace longitudinal_utils
 {
 
-bool isValidTrajectory(const Trajectory & traj)
+double calcStopDistance(
```
yuki-takagi-66, line 34: I think calculating the positional relationship using the index nearest to the ego-pose and the stop point index might not be sufficient here.
Could you update it to use the relationship between the actual ego-pose and the stop point instead?

1. Since this value is used for control rather than planning, it requires a continuous value.
2. We need to ensure it returns an appropriate value even if the vehicle overruns the goal (e.g., goes 1m past the goal).

control/autoware_pid_longitudinal_controller/src/longitudinal_controller_utils.cpp
```diff
@@ -82,29 +60,21 @@ double getPitchByPose(const Quaternion & quaternion_msg)
...
-    }
-    // NOTE: The ego pose is close to the goal.
-    return std::make_pair(
-      std::min(start_idx, trajectory.points.size() - 2), trajectory.points.size() - 1);
-  }();
-
-  return autoware_utils::calc_elevation_angle(
-    trajectory.points.at(prev_idx).pose.position, trajectory.points.at(next_idx).pose.position);
+  const double clamped_start_base = std::clamp(start_base, 0.0, trajectory.length());
+  const auto [pitch_start_base, end_base] =
```
yuki-takagi-66, line 71: Is the current handling sufficient when the vehicle is near the goal?
For example, shouldn't we need something like the following logic?

```
const auto pitch_start_base = min(clamped_start_base, bases.at(bases.size() - 2))
const auto end_base = min(clamped_start_base + wheel_base, bases.at(bases.size() - 1) 
```

control/autoware_pid_longitudinal_controller/include/autoware/pid_longitudinal_controller/pid_longitudinal_controller.hpp
```diff
@@ -82,18 +85,19 @@ class PidLongitudinalController : public trajectory_follower::LongitudinalContro
   };
   enum class Shift { Forward = 0, Reverse };
 
-  struct ControlData
+  struct ExperimentalControlData
```
yuki-takagi-66, line 88: I feel like we might not need the experimental prefix for this type name. What do you think?

control/autoware_pid_longitudinal_controller/src/pid_longitudinal_controller.cpp
```diff
@@ -440,101 +466,63 @@ trajectory_follower::LongitudinalOutput PidLongitudinalController::run(
...
-  control_data.nearest_idx = current_interpolated_pose.second + 1;
-  control_data.target_idx = control_data.nearest_idx;
-  const auto nearest_point = current_interpolated_pose.first;
-  auto target_point = current_interpolated_pose.first;
-
-  // Delay compensation - Calculate the distance we got, predicted velocity and predicted
-  // acceleration after delay
+  control_data.interpolated_traj = m_trajectory_experimental;
+
+  const auto current_s = autoware::experimental::trajectory::find_first_nearest_index(
```
yuki-takagi-66, line 483: Please ensure it works even if the vehicle overruns the goal.

yuki-takagi-66 (changes requested):
Thank you for the PR

As a basic requirement, the pid_long controller must continue to function even in the event of a goal overrun. Could you please revise the code based on this assumption?

I've only made it through about half of the review, but I'd like to take another look at the entire PR after these updates are applied.

control/autoware_pid_longitudinal_controller/src/pid_longitudinal_controller.cpp
yuki-takagi-66: I'm note sure at this point.

control/autoware_pid_longitudinal_controller/src/longitudinal_controller_utils.cpp
```diff
@@ -31,40 +33,86 @@ namespace autoware::motion::control::pid_longitudinal_controller
...
-  const size_t end_idx = stop_idx_opt ? *stop_idx_opt : traj.points.size() - 1;
-  const size_t seg_idx = autoware::motion_utils::findFirstNearestSegmentIndexWithSoftConstraints(
-    traj.points, current_pose, max_dist, max_yaw);
-  const double signed_length_on_traj = autoware::motion_utils::calcSignedArcLength(
-    traj.points, current_pose.position, seg_idx, traj.points.at(end_idx).pose.position,
-    std::min(end_idx, traj.points.size() - 2));
+    const double dx = end_point.pose.position.x - prev_point.pose.position.x;
+    const double dy = end_point.pose.position.y - prev_point.pose.position.y;
+    const double dz = end_point.pose.position.z - prev_point.pose.position.z;
+    const double seg_len = std::sqrt(dx * dx + dy * dy + dz * dz);
```
soblin:
```suggestion
    const double seg_len = last_base - prev_base
```
is sufficient, right ?

control/autoware_pid_longitudinal_controller/src/longitudinal_controller_utils.cpp
```diff
@@ -31,40 +33,86 @@ namespace autoware::motion::control::pid_longitudinal_controller
...
+    const double first_base = bases[0];
+    const double second_base = bases[1];
+
+    const auto first_point = traj.compute(first_base);
+    const auto second_point = traj.compute(second_base);
+
+    const double dx = second_point.pose.position.x - first_point.pose.position.x;
+    const double dy = second_point.pose.position.y - first_point.pose.position.y;
+    const double dz = second_point.pose.position.z - first_point.pose.position.z;
+    const double seg_len = std::sqrt(dx * dx + dy * dy + dz * dz);
```
soblin: Same

control/autoware_pid_longitudinal_controller/src/longitudinal_controller_utils.cpp
```diff
@@ -82,29 +130,23 @@ double getPitchByPose(const Quaternion & quaternion_msg)
 }
 
 double getPitchByTraj(
-  const Trajectory & trajectory, const size_t start_idx, const double wheel_base)
+  const TrajectoryExperimental & trajectory, const double start_base, const double wheel_base)
 {
-  // cannot calculate pitch
-  if (trajectory.points.size() <= 1) {
+  const auto bases = trajectory.get_underlying_bases();
+  if (bases.size() <= 1) {
```
soblin:
```suggestion
  if (bases.size() <= 2) {
```

control/autoware_pid_longitudinal_controller/src/pid_longitudinal_controller.cpp
```diff
@@ -440,101 +466,59 @@ trajectory_follower::LongitudinalOutput PidLongitudinalController::run(
...
+  const double current_s = autoware::experimental::trajectory::find_nearest_index(
+    control_data.interpolated_traj, current_pose.position);
+
+  double target_s = current_s;
+
   control_data.state_after_delay =
     predictedStateAfterDelay(control_data.current_motion, m_delay_compensation_time);
 
-  // calculate the target motion for delay compensation
   constexpr double min_running_dist = 0.01;
```
soblin, line 491: I think we should define this as a constant used both here and in `calcStopDistance`

PR #12301, opened 2026-03-13
feat(mpc_lateral_controller): add integrator friendly parameters to mpc lateral control

mkquda:
## Description

Add parameter to limit change in steering offset, and implement logic to apply change limit in mpc_lateral_controller

## Related links

**Parent Issue:**

- Link

## How was this PR tested?
PSIM

## Notes for reviewers

requires https://github.com/autowarefoundation/autoware_launch/pull/1792

### ROS Parameter Changes

| Added | `steering_offset.max_update_th`   | `double` | `0.01`         | threshold for maximum change in steering offset [rad] |

## Effects on system behavior

None.

### DIFF
--- a/control/autoware_mpc_lateral_controller/include/autoware/mpc_lateral_controller/mpc_lateral_controller.hpp
+++ b/control/autoware_mpc_lateral_controller/include/autoware/mpc_lateral_controller/mpc_lateral_controller.hpp
@@ -72,6 +72,7 @@ private:
 
   std::shared_ptr<Butterworth2dFilter> lpf_steer_offset_;
   double m_steering_offset_{0.0};
+  double m_steering_offset_target_{0.0};
   double m_steering_offset_filtered_{0.0};
 
   std::shared_ptr<diagnostic_updater::Updater>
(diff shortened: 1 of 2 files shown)


mkquda:
> @mkquda I believe this PR clamps the update amount of the `steer_offset_target` between two executions of `run()` in the mpc_lateral_controller, which only limits the update amount between `ctrl_period:=0.03` seconds. Therefore, the controller looks like trying to compensate the steering offset as usual even if the `steering_offset_update` is larger than expected.
> 
> Since this change is aimed to make safe control by preventing the effect of drastic changes of `steering_offset_update`, I think clamping the value in [`set_steering_offset`](https://github.com/autowarefoundation/autoware_universe/blob/89371f70859b2ddadb0cd87fcf9b620ff5f105fd/control/autoware_mpc_lateral_controller/include/autoware/mpc_lateral_controller/mpc_lateral_controller.hpp#L63) makes more sense to me. What do you think?

@TaikiYamada4 
apologies for the late response.
The intention of this PR is to limit the change of the steering_offset_update used in the controller per execution, however it should eventually reach the target steering_offset.

`set_steering_offset` is only executed when a new msg is received on `steering_offet_update` topic, so if we clamp it, we cannot update the value anymore until the next msg.

we already have the parameter [max_offset_limit](https://github.com/autowarefoundation/autoware_universe/blob/main/vehicle/autoware_steer_offset_estimator/config/steer_offset_estimator.param.yaml#L26) on the steering_offset_estimator side to limit the value of the offset update.

TaikiYamada4 (approved):
Confirmed the improvement via PSim.
LGTM!

PR #12302, opened 2026-03-13
feat(traffic_light_classifier): add regression architecture based classifier

badai-nguyen:
## Description

- This PR adds support for a new regression-based traffic light recognition (TLR) inference to the `autoware_traffic_light_classifier` package.
- The architecture of CNN model is based on lightnet model which is implemented and described here: https://github.com/tier4/trt-lightnet
- This PR merging must be done after https://github.com/autowarefoundation/autoware/pull/6968 and https://github.com/autowarefoundation/autoware/pull/6986 merged 
- This PR need merge together with https://github.com/autowarefoundation/autoware_launch/pull/1807
## Related links

**Parent Issue:**

- [TIER IV internal link](https://tier4.atlassian.net/browse/T4DEV-48279)

## How was this PR tested?
- Run node's launcher standalone and confirm there is no error and the new engine file is created successfully:
  - New added artifacts [new onnx](https://awf.ml.dev.web.auto/perception/models/traffic_light_classifier/v4/traffic_light_lamp_recognizer_comlops.onnx)  and [lamp_recognizer_ml.param.yaml](https://awf.ml.dev.web.auto/perception/models/traffic_light_classifier/v4/lamp_recognizer_ml.param.yaml) should be downloaded into `$HOME/autoware_data/traffic_light_classifier/` manually or by running `./setup-dev-env.sh`
  - Run launch with new model option and confirm no error from launch side, engine file is created:
    - `ros2 launch autoware_traffic_light_classifier pedestrian_traffic_light_classifier.launch.xml classifier_type:=2`
    - `ros2 launch autoware_traffic_light_classifier car_traffic_light_classifier.launch.xml classifier_type:=2`

- Intergration test by internal Evaluator: WIP

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


perception/autoware_traffic_light_classifier/src/classifier/cnn_lamp_recognizer.cpp
```diff
@@ -0,0 +1,579 @@
...
+  const cv::Size input_size(input_width_, input_height_);
+  cv::Mat blob =
+    cv::dnn::blobFromImages(images, scale, input_size, cv::Scalar(0, 0, 0), false, false, CV_32F);
+  if (!blob.isContinuous()) {
+    blob = blob.clone();
+  }
+  input_h_.assign(blob.ptr<float>(), blob.ptr<float>() + blob.total());
+  const size_t copy_size = input_h_.size() * sizeof(float);
+  CHECK_CUDA_ERROR(
+    cudaMemcpyAsync(input_d_.get(), input_h_.data(), copy_size, cudaMemcpyHostToDevice, *stream_));
```
YoshiRi: Since you already ensure blob is continuous, input_h_ is completely redundant. You can copy directly from the OpenCV blob to the GPU, saving CPU cycles and memory bandwidth:

```
const size_t copy_size = blob.total() * sizeof(float);
cudaMemcpyAsync(input_d_.get(), blob.ptr<float>(), copy_size, cudaMemcpyHostToDevice, *stream_);
```

YoshiRi: So I think we can remove input_h_ safely.

perception/autoware_traffic_light_classifier/src/classifier/cnn_lamp_recognizer.cpp
```diff
@@ -0,0 +1,579 @@
...
+constexpr int TLR_NUM_ANCHORS = 3;
+constexpr int TLR_CHANS_PER_ANCHOR = 16;
+constexpr int TLR_X_INDEX = 0, TLR_Y_INDEX = 1, TLR_W_INDEX = 2, TLR_H_INDEX = 3;
+constexpr int TLR_OBJ_INDEX = 4;
+constexpr int TLR_COLOR_START = 5, TLR_TYPE_START = 8;
+constexpr int TLR_NUM_TYPES = 6, TLR_NUM_COLORS = 3;
+constexpr int TLR_COS_INDEX = 14, TLR_SIN_INDEX = 15;
+constexpr float TLR_SCALE_X_Y = 2.0f;
+constexpr float TLR_BBOX_OFFSET = 0.5f * (TLR_SCALE_X_Y - 1.0f);
+constexpr float TLR_ANCHORS[TLR_NUM_ANCHORS * 2] = {7.0f, 7.0f, 14.0f, 14.0f, 42.0f, 42.0f};
```
YoshiRi: While it's standard to hardcode some YOLO/regression grid math, consider moving these to a struct `CnnLampRecognizerConfig` or declaring them as node parameters. If the ONNX model is ever retrained with different anchor scales or class counts, hardcoded values will silently cause bounding boxes to decode incorrectly, which is a nightmare to debug.

perception/autoware_traffic_light_classifier/src/classifier/cnn_lamp_recognizer.cpp
```diff
@@ -0,0 +1,577 @@
...
+#include <vector>
+namespace autoware::traffic_light
+{
+
+using MsgTE = tier4_perception_msgs::msg::TrafficLightElement;
+
+namespace
+{
+
+constexpr float kPi = 3.14159265358979f;
```
a-maumau: we can use `autoware_utils_math::pi` https://github.com/autowarefoundation/autoware_utils/blob/2c663f0274f5370b507a3ee238f49bc570745abd/autoware_utils_math/include/autoware_utils_math/constants.hpp#L21
if you want to avoid the use of `M_PI`

perception/autoware_traffic_light_classifier/src/classifier/cnn_lamp_recognizer.cpp
```diff
@@ -0,0 +1,577 @@
...
+  if (angle_rad >= -7.0f * kPi / 8.0f && angle_rad < -5.0f * kPi / 8.0f)
+    return ArrowDirection::DOWN_LEFT_ARROW;
+  if (angle_rad >= -5.0f * kPi / 8.0f && angle_rad < -3.0f * kPi / 8.0f)
+    return ArrowDirection::LEFT_ARROW;
+  return ArrowDirection::UP_LEFT_ARROW;
+}
+
+}  // namespace
+
+static float boxIou(const BBox & bbox1, const BBox & bbox2)
```
a-maumau: I think `boxIou` should be renamed to use a verb.
e.g. calculateIou or something

perception/autoware_traffic_light_classifier/src/classifier/cnn_lamp_recognizer.cpp
```diff
@@ -0,0 +1,577 @@
...
+  const int expand_h =
+    std::max(static_cast<int>((kDebugImageWidth * debug_image.rows) / debug_image.cols), 1);
+  cv::resize(debug_image, debug_image, cv::Size(kDebugImageWidth, expand_h));
+  cv::Mat text_img(cv::Size(kDebugImageWidth, kDebugTextHeight), CV_8UC3, cv::Scalar(0, 0, 0));
+  cv::putText(
+    text_img, text, cv::Point(5, 25), cv::FONT_HERSHEY_COMPLEX, 0.5, cv::Scalar(0, 255, 0), 1);
+  cv::vconcat(debug_image, text_img, debug_image);
+}
+
+static void cvtBBoxInfo2LampElement(const BBoxInfo & d, LampElement & element)
```
a-maumau: for me, it is too shortened. I prefer `convertBBoxInfoToLampElement`.
also the name for the argument variable `d`, it doesn't sounds like it is bound to `BBoxInfo`.

a-maumau (commented):
Other than the part YoshiRi-san mentioned, the code overall LGTM.

One point that I am concerned about is the confidence for each traffic light bulb.
This PR's implementation may have different confidence of each bulbs, but for the CNN classifier, all the bulbs are the same (since the model classify as a single traffic light image).
This assumption, all bulbs confidence is same, is used in the [multi camera fusion](https://github.com/autowarefoundation/autoware_universe/blob/587617ac86939482db59aa952fee6700f4e80f39/perception/autoware_traffic_light_multi_camera_fusion/src/traffic_light_multi_camera_fusion_node.cpp#L280).
We might need to use the mean, min, or max to unified the results.

Also, I have some minor comments.

a-maumau:
Also, this is not only for this PR, but we should add what kind of model outputs (and inputs) are required in the documents.

perception/autoware_traffic_light_classifier/src/classifier/cnn_lamp_recognizer.cpp
badai-nguyen: @YoshiRi @a-maumau I working on refactoring. Since this parameters are determined during training and cannot be changed during operation, so I think it better to keep them as model parameters (which are managed together with ONNX file).

perception/autoware_traffic_light_classifier/src/classifier/cnn_lamp_recognizer.cpp
badai-nguyen: Thank you for your comment. I fixed at c3589043de4aa525214a3c3ed240adcb3dc6f530

perception/autoware_traffic_light_classifier/src/classifier/cnn_lamp_recognizer.cpp
badai-nguyen: As your recommendation and internal discussion, I've changed those constants into parameter file at https://github.com/autowarefoundation/autoware_universe/pull/12302/changes/cbe94ab22cb02786848260b9636f965a36254b7b which is managed together with onnx. An additional PR for updated that file on autoware side is needed. I will open PR after the file is uploaded to webauto.

perception/autoware_traffic_light_classifier/src/classifier/cnn_lamp_recognizer.cpp
badai-nguyen: I fixed at https://github.com/autowarefoundation/autoware_universe/pull/12302/changes/483b59d1ceea43a18b80dd0b373a2a715713e269

perception/autoware_traffic_light_classifier/src/classifier/cnn_lamp_recognizer.cpp
badai-nguyen: I updated here https://github.com/autowarefoundation/autoware_universe/pull/12302/changes/83da59012e7df29912f194873ad0e6e2d7d16ec8

perception/autoware_traffic_light_classifier/src/classifier/cnn_lamp_recognizer.cpp
badai-nguyen: I updated here https://github.com/autowarefoundation/autoware_universe/pull/12302/changes/0f28ee898f03b276e5284f353bd4fc702a93b867

perception/autoware_traffic_light_classifier/src/classifier/cnn_lamp_recognizer.cpp
a-maumau: thank you, but is there a reason to use `cvt` instead of `convert`? (I am assuming it is convert)

badai-nguyen: I'm sorry, I missed that. It's fixed at 3d301c93bf7166d36c3a373cb3e408530adb9cb5

perception/autoware_traffic_light_classifier/src/classifier/cnn_lamp_recognizer.cpp
badai-nguyen: @a-maumau @YoshiRi 
the PR is created: could you check this one also https://github.com/autowarefoundation/autoware/pull/6986

perception/autoware_traffic_light_classifier/src/classifier/cnn_lamp_recognizer.cpp
```diff
@@ -0,0 +1,589 @@
...
+  dim_max.d[2] = input_height_;
+  dim_max.d[3] = input_width_;
+
+  auto profile_dims = std::make_unique<std::vector<autoware::tensorrt_common::ProfileDims>>();
+  profile_dims->push_back(autoware::tensorrt_common::ProfileDims(0, dim_min, dim_opt, dim_max));
+
+  if (!trt_common_->setup(std::move(profile_dims), nullptr)) {
+    throw std::runtime_error("CnnLampRecognizer: Failed to setup TensorRT engine");
+  }
+
```
YoshiRi, line 331: To prevent potential out-of-bounds memory access or silent logical bugs, we should validate that the `model_params` loaded from the YAML file match the actual TensorRT engine output dimensions.

Consider adding an assertion or a runtime error check right after `trt_common_->setup(...)` and getting the output dimensions:

```cpp
const int expected_channels = model_params_.num_anchors * model_params_.chans_per_anchor;
if (out_c_ != expected_channels) {
  throw std::runtime_error("LampRecognizer: Model output channels (" + std::to_string(out_c_) + 
                           ") do not match YAML configuration (" + std::to_string(expected_channels) + ").");
}

File: perception/autoware_traffic_light_classifier/src/classifier/cnn_lamp_recognizer.cpp [hunk 7]
```diff
@@ -0,0 +1,589 @@
...
+    for (int32_t j = 1; j < dims.nbDims; ++j) {
+      if (dims.d[j] <= 0) dims.d[j] = 1;
+    }
+    const size_t output_size = std::accumulate(
+      dims.d, dims.d + dims.nbDims, static_cast<size_t>(1), std::multiplies<size_t>());
+    CHECK_CUDA_ERROR(cudaMemcpyAsync(
+      output_h_.at(i).get(), output_d_.at(i).get(), output_size * sizeof(float),
+      cudaMemcpyDeviceToHost, *stream_));
+  }
+  CHECK_CUDA_ERROR(cudaStreamSynchronize(*stream_));
```
YoshiRi, line 419: [Just a comment]
Just a heads-up regarding the `cudaStreamSynchronize(*stream_)` call here. 
Calling this essentially blocks the host thread (and thus the ROS 2 callback thread) until the GPU inference and memory copy are fully complete. While this might be unavoidable with the current synchronous callback design, it prevents the CPU from doing any other work (like preparing the next batch or handling other ROS messages) while the GPU is busy. 

No immediate action is required for this PR, but we should keep this in mind as a potential bottleneck for future latency optimizations (e.g., fully asynchronous inference pipelines).

perception/autoware_traffic_light_classifier/src/classifier/cnn_lamp_recognizer.cpp
```diff
@@ -0,0 +1,589 @@
...
+}
+
+static void convertBBoxInfoToLampElement(const BBoxInfo & box_info, LampElement & element)
+{
+  element.color = static_cast<Color>(std::min(2, std::max(0, box_info.subClassId)));
+  element.confidence = box_info.prob;
+  element.box = box_info.box;
+  element.shape = static_cast<Shape>(box_info.classId);
+  if (element.shape == Shape::ARROW) {
+    element.arrow_direction = angleToArrowDirection(std::atan2(box_info.sin, box_info.cos));
```
YoshiRi: Let's add a numerical fail-safe for `std::atan2`. If the network outputs values very close to `0.0` for both `sin` and `cos`, `atan2(0, 0)` is technically implementation-defined (usually 0) and represents an uncertain prediction.

It would be safer to check for tiny values before computing the angle, for example:

```cpp
if (std::abs(box_info.sin) < 1e-6f && std::abs(box_info.cos) < 1e-6f) {
  element.arrow_direction = ArrowDirection::UNKNOWN; // Or fallback to a default
} else {
  element.arrow_direction = angleToArrowDirection(std::atan2(box_info.sin, box_info.cos));
}

File: perception/autoware_traffic_light_classifier/src/classifier/cnn_lamp_recognizer.cpp [hunk 9]
```diff
@@ -0,0 +1,589 @@
...
+      std::vector<LampElement> unique_lamps;
+      for (const auto & lamp : traffic_lamps) {
+        auto it =
+          std::find_if(unique_lamps.begin(), unique_lamps.end(), [&](const LampElement & e) {
+            return e.shape == lamp.shape && e.arrow_direction == lamp.arrow_direction;
+          });
+        if (it == unique_lamps.end()) {
+          unique_lamps.push_back(lamp);
+        } else {
+          it->confidence = std::max(it->confidence, lamp.confidence);
```
YoshiRi: [bug?]
If the new lamp has a higher confidence than the existing *it, we are only updating the confidence value. This means we keep the bounding box (box) of the lower-confidence detection, but give it the higher score.
We should replace the entire element if the new confidence is higher:

```cpp
} else if (lamp.confidence > it->confidence) {
  *it = lamp; // Overwrite the whole object to keep the better bounding box
}
```

perception/autoware_traffic_light_classifier/src/classifier/cnn_lamp_recognizer.cpp
a-maumau: Adding some small value will help the issue without adding if-branch, it should be more faster.

a-maumau: It would be add some small bias to the result, but we can add small value (we don't need the high precision for this part, I guess). this could remove the if-branch, more faster.

badai-nguyen: @a-maumau As my understand,  @YoshiRi concern is about the undesired output of network, not about the std::atan2 function itself. So I agree with Ri-san and prefer ArrowDirection::UNKNOWN in that case than add bias to make std::atan2 to be determined (= some thingk like std::atan2(0, 1e-6).

perception/autoware_traffic_light_classifier/src/classifier/cnn_lamp_recognizer.cpp
badai-nguyen: @YoshiRi thanks for your comment, Yes it was a bug.  I fixed https://github.com/autowarefoundation/autoware_universe/pull/12302/commits/114c4b1b13e19247f14b7c7d8b140ba5d861dd0c

perception/autoware_traffic_light_classifier/src/classifier/cnn_lamp_recognizer.cpp
badai-nguyen: I fixed at 224f0df8a745021ea52b93d6d81f21b72aee714d

perception/autoware_traffic_light_classifier/src/classifier/cnn_lamp_recognizer.cpp
badai-nguyen: I fixed at 799710fe400b316cb4e53fb323373b3a88d69010

PR #12303, opened 2026-03-14
feat(euclidean_cluster): apply `agnocast_wrapper::Node` to voxel_grid_based_euclidean_cluster

Koichi98:
## Description
Apply `autoware_agnocast_wrapper` to `voxel_grid_based_euclidean_cluster` node using the Node Wrapper approach (`agnocast_wrapper::Node`).

Please refer to https://github.com/autowarefoundation/autoware_core/blob/main/common/autoware_agnocast_wrapper/docs/review_guide.md for the review guide of `autoware_agnocast_wrapper` application PR.

  For details on `autoware_agnocast_wrapper`, see:
  - https://github.com/autowarefoundation/autoware_core/tree/main/common/autoware_agnocast_wrapper
  - https://github.com/orgs/autowarefoundation/discussions/5835

### Launch behavior when `ENABLE_AGNOCAST=1`
When `ENABLE_AGNOCAST=1`, the `euclidean_cluster` node is launched as a standalone `Node` (with `LD_PRELOAD`) instead of a `ComposableNode` inside component conatiner. Since Agnocast enables zero-copy transfer of pointcloud data between processes, there is no longer a need to keep nodes in the same process for performance. Separating them is preferable from the perspectives of startup time and fault isolation. Applying agnocast_wrapper to low_height_cropbox_filter is planned as a follow-up PR.

The `low_height_cropbox_filter` (used when `use_low_height_cropbox` is true) remains as a `ComposableNode` loaded into the container, as it does not use the Agnocast wrapper.

## Related links
- Apply agnocast publisher to the topic which publishes to this node.
    - https://github.com/autowarefoundation/autoware_universe/pull/12304
- Apply agnocast subscription to the topic which subscribes this node's publisher
    - https://github.com/autowarefoundation/autoware_universe/pull/12308

  Parent Issue:

  - Link

## How was this PR tested?
By running logging simulator with sample ros2 bag, with 

- `ENABLE_AGNOCAST=0` when build, and `ENABLE_AGNOCAST=0` at runtime
- `ENABLE_AGNOCAST=1` when build, and `ENABLE_AGNOCAST=0` at runtime
- `ENABLE_AGNOCAST=1` when build, and `ENABLE_AGNOCAST=1` at runtime

## Notes for reviewers
I would really appreciate it if you could also check the following:
- `ENABLE_AGNOCAST=0` when build, and `ENABLE_AGNOCAST=0` at runtime
- `ENABLE_AGNOCAST=1` when build, and `ENABLE_AGNOCAST=0` at runtime

CodeScene flags `launch_setup` for exceeding the function length threshold (131 lines vs 70-line limit). This is intentional — extracting the agnocast path into a separate function would make the diff harder to review and obscure what was actually changed. Happy to refactor if reviewers prefer it.

## Interface changes

  None.

## Effects on system behavior
None when `ENABLE_AGNOCAST` is not set or set to 0 on runtime. When `ENABLE_AGNOCAST=1`, CPU Time used in this node will be reduced, and also the startup time of the whole autoware will be reduced by keep applying `agnocast::Node` to each nodes in autoware. (`ENABLE_AGNOCAST=1`)


technolojin (approved):
LGTM

PR #12307, opened 2026-03-15
fix(tier4_system_rviz_plugin): update vel limit I/F from autoware_state_panel

Autumn60:
## Description

Change the topic used by autoware_state_panel (Rviz) for speed limit settings from `/planning/scenario_planning/max_velocity_default` to `/planning/scenario_planning/max_velocity_candidates`.

### Background / Problem

- `/planning/scenario_planning/max_velocity_default` was being published by both **autoware_state_panel** and **ad_api_adaptor**.
- **external_velocity_limit_selector** treats `max_velocity_default` as a single "API" source: [it always fixes the sender to `"api"` and overwrites a single table entry.](https://github.com/autowarefoundation/autoware_universe/blob/0a309acc11555591ed7844d36b5c0b9818abe225/planning/autoware_external_velocity_limit_selector/src/external_velocity_limit_selector_node.cpp#L197-L210)
- As a result, when autoware_state_panel publishes on UI interaction, ad_api_adaptor soon publishes based on `current_max_velocity`, overwriting the "api" entry. The panel’s speed limit therefore only takes effect briefly; the sustained limit in practice comes from ad_api_adaptor subscribing to `current_max_velocity` and republishing to `max_velocity_default`.

### Changes

- Switch autoware_state_panel’s speed-limit publisher to **max_velocity_candidates**.
- On `max_velocity_candidates`, the selector uses the message **sender** as-is, so the UI is stored under a different sender (e.g. `"xxx(autoware_state_panel)"`) from the API.
  - Since the node name for RvizPlugin is the same as the Rviz2 node name (e.g., `/rviz2`), so I added `(autoware_state_panel)` suffix.
- The selector chooses the strictest limit (minimum `max_velocity`) across all senders, so a limit set via the panel is no longer overwritten by ad_api_adaptor’s continuous publish.
- Update the README Output topic description to match this interface change.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

- In Rviz, operated the autoware_state_panel speed limit slider and confirmed that the set speed limit is maintained.
- Confirmed that speed limits from ad_api_adaptor (API) and from the panel coexist, and the stricter one is applied.

## Notes for reviewers

The fact that `autoware_state_panel` uses the internal interface directly rather than via the API is an issue that should be addressed in the future.

#### Modifications

| Version | Topic Type | Topic Name | Message Type | Description |
|:--------|:-----------|:-----------|:-------------|:------------|
| Old | Pub | `/planning/scenario_planning/max_velocity_default` | `autoware_internal_planning_msgs::msg::VelocityLimit` | Topic previously used by autoware_state_panel for max speed (shared with API) |
| New | Pub | `/planning/scenario_planning/max_velocity_candidates` | `autoware_internal_planning_msgs::msg::VelocityLimit` | Topic used by autoware_state_panel for max speed (internal multi-source; identified by sender) |

- **Scope**: Only the topic published by autoware_state_panel in `tier4_state_rviz_plugin`. ad_api_adaptor continues to publish to `max_velocity_default`.
- **Message**: Unchanged. The panel publishes `VelocityLimit` with `sender` set to the node name plus `"(autoware_state_panel)"`.

## Effects on system behavior

- **autoware_state_panel**: Speed limits set in the Rviz panel are no longer overwritten by ad_api_adaptor’s continuous publish and are maintained as intended.
- **external_velocity_limit_selector**: Unchanged behavior; it still subscribes to both `max_velocity_default` (API) and `max_velocity_candidates` (internal sources) and selects the strictest limit. autoware_state_panel is treated as one sender on the candidates side.


visualization/tier4_state_rviz_plugin/README.md
```diff
@@ -20,18 +20,18 @@ This plugin also can engage from the panel.
...
+| `/api/operation_mode/change_to_remote`                | `autoware_adapi_v1_msgs::srv::ChangeOperationMode`    | The service to change operation mode to remote     |
+| `/api/operation_mode/enable_autoware_control`         | `autoware_adapi_v1_msgs::srv::ChangeOperationMode`    | The service to enable vehicle control by Autoware  |
+| `/api/operation_mode/disable_autoware_control`        | `autoware_adapi_v1_msgs::srv::ChangeOperationMode`    | The service to disable vehicle control by Autoware |
+| `/api/routing/clear_route`                            | `autoware_adapi_v1_msgs::srv::ClearRoute`             | The service to clear route state                   |
+| `/api/motion/accept_start`                            | `autoware_adapi_v1_msgs::srv::AcceptStart`            | The service to accept the vehicle to start         |
+| `/api/autoware/set/emergency`                         | `tier4_external_api_msgs::srv::SetEmergency`          | The service to set external emergency              |
+| `/planning/scenario_planning/max_velocity_candidates` | `autoware_internal_planning_msgs::msg::VelocityLimit` | The topic to set maximum speed of the vehicle      |
```
sasakisasaki, line 34:
```suggestion
| `/planning/scenario_planning/max_velocity_candidates` | `autoware_internal_planning_msgs::msg::VelocityLimit` | The topic to set maximum velocity candidates of the vehicle      |
```

sasakisasaki (approved):
Reviewed, I agree with merging this PR while put a minor comment 👍 

visualization/tier4_state_rviz_plugin/README.md
Autumn60: In this topic, the parameter that can be set is "speed" (scalar), not "velocity" (vector).
In that sense, the message name and topic name should be corrected.

厳密な単語の定義として、この機能で設定可能なのは、速さ(スカラー。マグニチュード。絶対量であり方向の概念無し)であり、速度(ベクトル。方向の概念を表す符号付き)ではないです。
そういう意味では、本来修正されるべきなのはメッセージ名( `VelocityLimit` )やトピック名( `max_velocity` ) だと思います。

sasakisasaki: @Autumn60 Thank you for your kind feedback. Yes, you are right. Let's keep the current `README.md`.

PR #12308, opened 2026-03-15
feat(autoware_shape_estimation): replace to agnocast wrapper

TetsuKawa:
## Description
Apply `autoware_agnocast_wrapper` to `autoware_shape_estimation` node using the Node Wrapper approach (`agnocast_wrapper::Node`).

Please refer to https://github.com/autowarefoundation/autoware_core/blob/main/common/autoware_agnocast_wrapper/docs/review_guide.md for the review guide of `autoware_agnocast_wrapper` application PR.

  For details on `autoware_agnocast_wrapper`, see:
  - https://github.com/autowarefoundation/autoware_core/tree/main/common/autoware_agnocast_wrapper
  - https://github.com/orgs/autowarefoundation/discussions/5835

### Launch behavior when `ENABLE_AGNOCAST=1`
When `ENABLE_AGNOCAST=1`, the `shape_estimation` node is configured to load the Agnocast heap-hook library via `LD_PRELOAD` through the inclusion of `agnocast_env.launch.xml`. 

Since Agnocast enables zero-copy transfer of large perception messages (like `DetectedObjectsWithFeature`) between processes, it reduces CPU overhead significantly while maintaining modularity.

## Related links
- Apply agnocast publisher to the topic which publishes to this node (input: `labeled_clusters`).
    - https://github.com/autowarefoundation/autoware_universe/pull/12309
    - https://github.com/autowarefoundation/autoware_universe/pull/12310
- Apply agnocast subscription to the topic which subscribes this node's publisher (output: `shape_estimated_objects`).
    - https://github.com/autowarefoundation/autoware_universe/pull/12313
    - https://github.com/autowarefoundation/autoware_universe/pull/12314

## How was this PR tested?
By running logging simulator with sample ros2 bag, with 

- `ENABLE_AGNOCAST=0` when build, and `ENABLE_AGNOCAST=0` at runtime
- `ENABLE_AGNOCAST=1` when build, and `ENABLE_AGNOCAST=0` at runtime
- `ENABLE_AGNOCAST=1` when build, and `ENABLE_AGNOCAST=1` at runtime

## Notes for reviewers
- **Debug Publisher Template**: In `shape_estimation_node.cpp`, I updated `DebugPublisher` and `PublishedTimePublisher` to use `BasicDebugPublisher<autoware::agnocast_wrapper::Node>` and `BasicPublishedTimePublisher<autoware::agnocast_wrapper::Node>`. This is necessary because the Agnocast Node wrapper requires the node type to be explicitly passed to these utility classes to access the proper `now()` and `create_publisher()` methods.
- **Executor**: The node is registered using `AgnocastOnlySingleThreadedExecutor` as specified in `CMakeLists.txt`.
- **Test Guard**: Added a guard in `CMakeLists.txt` to skip certain tests when `ENABLE_AGNOCAST=1` to avoid environment-related failures during build-time testing.

## Interface changes
  None.

## Effects on system behavior
None when `ENABLE_AGNOCAST` is not set or set to 0 on runtime. When `ENABLE_AGNOCAST=1`, CPU Time used in this node for message serialization/deserialization will be reduced, especially for the `DetectedObjectsWithFeature` message which contains pointcloud-like features.

### DIFF
--- a/.github/workflows/build-and-test-daily.yaml
+++ b/.github/workflows/build-and-test-daily.yaml
@@ -48,7 +48,7 @@ jobs:
     uses: ./.github/workflows/build-and-test-reusable.yaml
     with:
       rosdistro: ${{ matrix.rosdistro }}
-      container: ${{ matrix.container-base }}${{ matrix.container-suffix }}
+      container: ${{ matrix.container-base }}
       container-suffix: ${{ matrix.container-suffix }}
       runner: ${{ matrix.runner }}
       build-pre-command: ${{ matrix.build-pre-command }}
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -15,10 +15,10 @@ jobs:
         include:
           # ROS distribution specific configurations
           - rosdistro: humble
-            container: ghcr.io/autowarefoundation/autoware:universe-devel-cuda
+            container: ghcr.io/autowarefoundation/autoware:universe-devel
 
           - rosdistro: jazzy
-            container: ghcr.io/autowarefoundation/autoware:universe-devel-jazzy-amd64-cuda
+            container: ghcr.io/autowarefoundation/autoware:universe-devel-jazzy-amd64
 
     uses: ./.github/workflows/build-and-test-reusable.yaml
     with:


perception/autoware_shape_estimation/package.xml
```diff
@@ -18,11 +18,14 @@
 
   <buildtool_export_depend>tensorrt_cmake_module</buildtool_export_depend>
 
+  <depend>agnocastlib</depend>
```
Koichi98: We might not need this. (Same for CMakeLists.txt)

perception/autoware_shape_estimation/package.xml
```diff
@@ -18,11 +18,14 @@
...
+  <depend>autoware_agnocast_wrapper</depend>
   <depend>autoware_cuda_dependency_meta</depend>
   <depend>autoware_cuda_utils</depend>
   <depend>autoware_perception_msgs</depend>
   <depend>autoware_tensorrt_common</depend>
   <depend>autoware_utils</depend>
+  <depend>autoware_utils_debug</depend>
```
Koichi98: We also might not need this. (Could you check it by building with and without `ENABLE_AGNOCAST=1` to make sure? NOTE: Please don't forget to delete the build cache of this package before building)

perception/autoware_shape_estimation/package.xml
TetsuKawa: Fixed in https://github.com/autowarefoundation/autoware_universe/pull/12308/commits/331b15e25dc51526eea028dd4409265941462bf5

perception/autoware_shape_estimation/package.xml
TetsuKawa: Fixed in https://github.com/autowarefoundation/autoware_universe/pull/12308/commits/dc6a51ce9ed727c2e437eeb34139e3456ba6ca97

technolojin (approved):
LGTM

PR #12311, opened 2026-03-16
fix(steer_offset_estimator): fix offset update publishing condition

mkquda:
## Description

use latest_reliable_result for publishing offset update which takes into account covariance and steady driving duration.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


TaikiYamada4 (approved):
LGTM

PR #12322, opened 2026-03-16
feat(autoware_raw_vehicle_cmd_conveter): adopt cie

TetsuKawa:
## Description

Apply `autoware_agnocast_wrapper` to enable `CallbackIsolatedAgnocastExecutor` (CIE) for this package, as part of the effort to achieve middleware-transparent scheduling and optimize end-to-end response time across Autoware. By going through `autoware_agnocast_wrapper`, CIE activation is controlled by the `ENABLE_AGNOCAST` environment variable — when unset or `0`, the node runs with the standard ROS 2 executor as before; when set to `1`, it switches to `CallbackIsolatedAgnocastExecutor`.

### Changes

- CMakeLists.txt: Replaced rclcpp_components_register_node with autoware_agnocast_wrapper_register_node, adding ROS2_EXECUTOR SingleThreadedExecutor and AGNOCAST_EXECUTOR CallbackIsolatedAgnocastExecutor options.
- package.xml: Added <depend>autoware_agnocast_wrapper</depend>.

## Related links

- [What is autoware_agnocast_wrapper?](https://github.com/autowarefoundation/autoware_core/blob/main/common/autoware_agnocast_wrapper/README.md)

## How was this PR tested?

- Evaluator passed.

## Notes for reviewers

**Please review the following aspects:**

- Verify that all relevant `CMakeLists.txt` and launch files in this package have been updated (no missed files).
- If the node meets **all** of the following conditions, please check for potential race conditions (see [Effects on system behavior](#effects-on-system-behavior) for the reason):
  1. The node was originally running on a `SingleThreadedExecutor`
  2. The node has multiple callback groups
  3. Shared variables are accessed across callback groups without proper mutex protection

## Interface changes

None.

## Effects on system behavior

When `ENABLE_AGNOCAST=0` (default), there is no change in behavior at all — the node runs with the standard ROS 2 executor exactly as before.

When `ENABLE_AGNOCAST=1`, the executor switches to `CallbackIsolatedAgnocastExecutor` (CIE). Internally, CIE creates a dedicated `rclcpp::SingleThreadedExecutor` for each callback group, so the fundamental scheduling behavior within each group remains almost identical. The key difference is that, for nodes originally using `SingleThreadedExecutor`, callback groups now run in parallel across separate threads, similar to `MultiThreadedExecutor`.

### DIFF
--- a/common/autoware_boundary_departure_checker/package.xml
+++ b/common/autoware_boundary_departure_checker/package.xml
@@ -18,7 +18,6 @@
   <depend>autoware_adapi_v1_msgs</depend>
   <depend>autoware_internal_debug_msgs</depend>
   <depend>autoware_interpolation</depend>
-  <depend>autoware_lanelet2_extension</depend>
   <depend>autoware_map_msgs</depend>
   <depend>autoware_motion_utils</depend>
   <depend>autoware_planning_msgs</depend>
--- a/common/autoware_traffic_light_utils/package.xml
+++ b/common/autoware_traffic_light_utils/package.xml
@@ -17,8 +17,9 @@
   <test_depend>ament_lint_auto</test_depend>
   <test_depend>autoware_lint_common</test_depend>
 
-  <depend>autoware_lanelet2_extension</depend>
   <depend>autoware_perception_msgs</depend>
+  <depend>lanelet2_core</depend>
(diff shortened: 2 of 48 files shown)


satoshi-ota (approved):
LGTM

PR #12347, opened 2026-03-18
fix(autoware_multi_object_tracker): enhance EKF stability with symmetry enforcement and improved covariance handling

technolojin:
## Description

This PR improves EKF numerical stability in the multi object tracker to prevent invalid/negative covariance values that can trigger warnings like “Covariance is not valid”.

- Stabilize EKF update math in kalman_filter_template.hpp
  - Replace explicit `S.inverse()` with an `LDLT` solve
  - Use **Joseph-form** covariance update + covariance **symmetrization** to better preserve PSD(Positive Semi-Definite)-ness and reduce numerical drift

- Harden association inverse-covariance in association/association.cpp
  - Guard against near-zero/negative 2×2 covariance determinant
  - Fall back to a safe diagonal inverse to avoid `inf/NaN` scores

- Robustify covariance eigen extraction in tracker/model/tracker_base.cpp
  - Explicitly symmetrize the XY block (handles small `X_Y != Y_X` drift)
  - Clamp tiny negative discriminants from floating-point round-off to avoid NaNs

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/kalman_filter_template.hpp
+++ b/perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/kalman_filter_template.hpp
@@ -222,7 +222,8 @@ public:
    */
   bool update(const MeasVec & y, const MeasVec & y_pred, const MeasModelMat & C, const MeasMat & R)
   {
-    if (!x_.allFinite() || !P_.allFinite() || !y.allFinite() || !y_pred.allFinite() ||
+    if (
+      !x_.allFinite() || !P_.allFinite() || !y.allFinite() || !y_pred.allFinite() ||
       !C.allFinite() || !R.allFinite()) {
       return false;
     }


perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/kalman_filter_template.hpp
```diff
@@ -221,16 +222,33 @@ class KalmanFilterTemplate
...
-    P_ -= K * (C * P_);
-    return true;
+
+    // Joseph stabilized covariance update: preserves PSD better than P = (I-KC)P.
+    const StateMat I_state = StateMat::Identity();
+    const StateMat I_KC = I_state - K * C;
+    P_ = I_KC * P_ * I_KC.transpose() + K * R_sym * K.transpose();
+    P_ = Scalar(0.5) * (P_ + P_.transpose());
+
+    return P_.allFinite();
```
ogrex: should x and P_ still be updated if P_.allFinite() is false?

technolojin: let me update the update functions

1. local copy the states
2. update
3. check
4. if checks are passed, apply to the states

technolojin: implemented https://github.com/autowarefoundation/autoware_universe/pull/12347/changes/6f3fa7be5acb216516445a16194d3523a91099ac

technolojin:
@mitsudome-r 
it seems CI is failing by autoware_path_optimizer build

PR #12354, opened 2026-03-18
ci: fix handling of container suffix for build-and-test workflows

mitsudome-r:
## Description

I've noticed that build-and-test CI check was failing due to wrong image name.
https://github.com/autowarefoundation/autoware_universe/actions/runs/23234272107/job/67534366005

The build-and-test-reusable.yaml takes `-cuda` as ${container-suffix} argument and attach this internally when pulling the image so there is no need to attach it when passing the container name

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/.github/workflows/build-and-test-daily.yaml
+++ b/.github/workflows/build-and-test-daily.yaml
@@ -48,7 +48,7 @@ jobs:
     uses: ./.github/workflows/build-and-test-reusable.yaml
     with:
       rosdistro: ${{ matrix.rosdistro }}
-      container: ${{ matrix.container-base }}${{ matrix.container-suffix }}
+      container: ${{ matrix.container-base }}
       container-suffix: ${{ matrix.container-suffix }}
       runner: ${{ matrix.runner }}
       build-pre-command: ${{ matrix.build-pre-command }}
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -15,10 +15,10 @@ jobs:
         include:
           # ROS distribution specific configurations
           - rosdistro: humble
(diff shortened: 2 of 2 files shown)


mitsudome-r:
Testing in https://github.com/autowarefoundation/autoware_universe/actions/runs/23237975679

xmfcx:
It seems I shouldn't have enabled auto-merge.

https://github.com/autowarefoundation/autoware/pkgs/container/autoware/742880377?tag=universe-devel-cuda-jazzy-amd64

this is the image.

https://github.com/autowarefoundation/autoware_universe/actions/runs/23246447670/job/67635300576#step:3:25

It is trying to pull `ghcr.io/autowarefoundation/autoware:universe-devel-jazzy-amd64-cuda`


- **Image:** `universe-devel-cuda-jazzy-amd64`
- **Pulling:** `universe-devel-jazzy-amd64-cuda`

We either need to update the manifests or how we refer to them.

Also we dont need to specify the cpu arch at all. We have multi arch tags as `universe-cuda-jazzy`: https://github.com/autowarefoundation/autoware/pkgs/container/autoware/743325854?tag=universe-cuda-jazzy

It all went wrong when we introduced this change like this:

https://github.com/autowarefoundation/autoware/commit/e53832e802607e5736de5a4bce6031606d31748c#diff-0a93724fe21d2c6fc898fa5254a9998c58b6bfdaa346e03759dfb3b0645fb94fR97-R99

We should have left the cuda suffix at the end.

I'll try to fix it for the universe images on the autoware main repository.

xmfcx:
- Created https://github.com/autowarefoundation/autoware/pull/6916

to address this. We don't need to update universe side for this fix, your PR has already made it correct.

PR #12355, opened 2026-03-18
fix: bug(`dummy_perception_publisher`): remove a wrong dependency

sasakisasaki:
## Description
Remove the runtime dependency on `autoware_shape_estimation` from `autoware_dummy_perception_publisher`, since it is not used by the simulator package.

## Related links
- https://github.com/orgs/autowarefoundation/discussions/6892

## How was this PR tested?

Based on this PR, I built all the autoware packages based on [the nightly installation instructions](https://autowarefoundation.github.io/autoware-documentation/main/installation/autoware/source-installation/#prerequisites).

Note that I edited `autoware-nightly.repos` and replaced the remote URL and branch to point to my fork and branch for this PR, so that the source code of this PR is included in the build.

```bash
```

After build, I followed the instruction in [this planning simulator demo](https://autowarefoundation.github.io/autoware-documentation/main/demos/planning-sim/), and confirmed that the simulator still works as follows.
```bash
$ source ~/autoware/install/setup.bash
$ ros2 launch autoware_launch planning_simulator.launch.xml map_path:=$HOME/autoware_map/sample-map-planning vehicle_model:=sample_vehicle sensor_model:=sample_sensor_kit
```
During the planning simulation, I checked by:
- Initialize pose
- Set route
- Run autonomous driving
- Put dummy objects: pedestrian, vehicle, car, and bus

Also, tested by TIER IV internal scenario test tool.
- https://github.com/tier4/auto-evaluator/actions/runs/23277195466

## Notes for reviewers
The commented shape estimation include remains unchanged; only the package dependency is removed.

## Interface changes
None.

### Topic changes
None.

## Effects on system behavior
No functional changes expected for simulation; this only drops an unused dependency.


simulator/autoware_dummy_perception_publisher/launch/dummy_perception_publisher.launch.xml
Autumn60: Since `/perception/object_recognition/detection/objects_with_feature` will no longer be available in `<group if="$(var real)">`, it should be moved to `<group unless="$(var real)">` .

`/perception/object_recognition/detection/objects_with_feature` が提供されない場合( `<group if="$(var real)">` のとき)、この機能は動作できないため、起動処理は `<group unless="$(var real)">` 内に移すべきです。

sasakisasaki: @Autumn60 Thank you very much! Yes, you are right. Now fixed by [this commit](https://github.com/autowarefoundation/autoware_universe/pull/12355/commits/c48a867d5fb898365cd224b64d73b9ee5dac42df) :+1:

technolojin:
this is not a bug.

https://github.com/orgs/autowarefoundation/discussions/6892#discussioncomment-16304367

sasakisasaki:
@technolojin Thank you, you are right. This is my overlook. Let me close this PR as there is much better approach proposed by @technolojin :+1:

sasakisasaki:
The following PR is the right approach I believe
- https://github.com/autowarefoundation/autoware_universe/pull/12382

PR #12362, opened 2026-03-19
feat(autoware_ptv3): add multi-type input and output & update model

amadeuszsz:
## Description

Align autoware_ptv3 with autoware_lidar_frnet (addressed in https://github.com/autowarefoundation/autoware_universe/pull/12000) for pointcloud API, filter parameters, and naming convention. Also updating the model.

- adding `filter.class_probability_threshold`, `filter.classes`, and `filter.output_format`                                                                                                                                                       
- supporting `XYZI` / `XYZIRC` / `XYZIRADRT` / `XYZIRCAEDT` input cloud formats                                                                                                                                                                     
- allowing filtered output to preserve input format or convert to a compatible requested format                                                                                                                                             
- extending segmentation output with `class_id` and probability                                                                                                                                                                               
- aligning public outputs to output/pointcloud/segmentation, visualization, and filtered                                                                                                                                                    
- moving model-specific metadata to ml_package params and unifying cloud format/type naming

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

This PR can be only merged after Autoware's artifacts (model) merge first in https://github.com/autowarefoundation/autoware/pull/6918.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_ptv3/README.md
+++ b/perception/autoware_ptv3/README.md
@@ -78,7 +78,11 @@ empty string, the filtered output preserves the same format as the input cloud.
 
 The model was trained on the T4Dataset using approximately 4,000 frames and is available in the Autoware artifacts.
 
-### Changelog
+## Troubleshooting
+
+### `Fail to create host memory`
+
+This error may occur when the system runs out of memory during TensorRT engine building. To mitigate this issue, consider reducing the `voxels_num` maximum limit in the model parameters, which controls the maximum number of voxels processed by the model. For the instance, using GPU with 8 GB of memory, setting maximum `voxels_num` to 192'000 may help to avoid this error.
 
 ## References/External links
 


perception/autoware_ptv3/include/autoware/ptv3/ptv3_config.hpp
```diff
@@ -81,44 +81,67 @@ class PTv3Config
...
-      auto g = static_cast<std::uint8_t>(colors_green[i]);
-      auto b = static_cast<std::uint8_t>(colors_blue[i]);
-      std::uint32_t rgb = (r << 16) | (g << 8) | b;
-      float rgb_float;
-      memcpy(&rgb_float, &rgb, sizeof(rgb_float));
-      colors_rgb_.push_back(rgb_float);
+    colors_rgb_ = make_palette(class_names_, palette);
 
-      std::string class_name = class_names_[i];
+    for (auto class_name : class_names_) {
```
kyoichi-sugahara: (I don't fully understand why the lowercase conversion of class names was originally needed.) 
There appears to be an issue where `class_names_` is not modified since a reference is not used. 
Please either 
- use a reference so that the lowercase conversion is properly applied
- or remove the lowercase conversion if it is unnecessary

kyoichi-sugahara:
```suggestion
    for (auto & class_name : class_names_) {
```

PR #12367, opened 2026-03-23
fix(autoware_tensorrt_bevformer): initialize ego2global_rot_cam to identity

xmfcx:
## Summary
- `ego2global_rot_cam` was declared without initialization, causing `-Werror=maybe-uninitialized` when built with `--coverage` (as CI does).
- Initialize it to `Eigen::Quaterniond::Identity()`, matching the pattern already used for `ego2global_trans_cam` on the next line.

## How to reproduce
The default local build (`colcon build`) does not trigger this because `-Werror=maybe-uninitialized` only fires under `--coverage`. CI uses:
```bash
colcon build --cmake-args -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_FLAGS='--coverage'
```
CI failure: https://github.com/autowarefoundation/autoware_universe/actions/runs/23348065420/job/68039152250#step:15:23123

## Test plan
- [ ] CI passes with `--coverage` flag

### DIFF
--- a/perception/autoware_tensorrt_bevformer/src/bevformer_node.cpp
+++ b/perception/autoware_tensorrt_bevformer/src/bevformer_node.cpp
@@ -411,7 +411,7 @@ void TRTBEVFormerNode::calculateSensor2LidarTransformsFromTF(
     Eigen::RowVector3d l2e_t_s = cams2ego_trans_[i].cast<double>().vector().transpose();
 
     // Get ego2global at camera timestamp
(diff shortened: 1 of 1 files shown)


xmfcx:
All passed but this:
image

https://github.com/autowarefoundation/autoware_universe/actions/runs/23428811982/job/68150135043?pr=12367#step:14:639

I guess both clang-tidy humble and jazzy artifacts have the same name so it is failing to upload the second one.

That is not related to this PR though.

Will fix that in a consequent pr.

PR #12370, opened 2026-03-23
fix(crosswalk_traffic_light_estimator): skip lanelets with no signal info in get_non_red_lanelets

interimadd:
## Description

Fix a bug in `get_non_red_lanelets` where lanelets with no current detection (`current_detected_signal`) and no historical detection (`last_detected_signal`) were incorrectly treated as "non-red".

When #12288 removed `if (!last_detected_signal) continue;`, lanelets whose traffic lights had never been detected started being added to `non_red_lanelets` due to the default `current_is_not_red = true` for undetected signals.

In scenarios like right-turn arrow signals, where only some lanes have GREEN detection and other lanes' signals are undetected, undetected straight lanes were incorrectly counted as non-red. This caused the crosswalk pedestrian signal to be estimated as RED instead of UNKNOWN, leading to ego not stopping for pedestrians and colliding with them.

The fix adds a guard to skip lanelets where both `current_detected_signal` and `last_detected_signal` are absent.

## Related links

**Parent Issue:**

- #12288

## How was this PR tested?

- Unit test: added `Estimate_RightArrowGreen_StraightUndetected_CrosswalkUnknown` to verify the crosswalk is estimated as UNKNOWN when only the right-turn signal is GREEN and the straight signal is undetected
- (TIER IV internal) Scenario test: `webauto ci scenario run --scenario-id 45f50680-a544-4a83-a011-1c5c0b0788da --scenario-version-id 3 --scenario-parameters '__tier4_modifier_max_stand_still=100'` confirmed no pedestrian collision
- (TIER IV internal) [Evaluator link](https://evaluation.tier4.jp/evaluation/reports/4a25ff4b-c0f3-5c54-99b0-a3f1b0076fab/?project_id=autoware_dev)

## Notes for reviewers

The map added in this PR represent following intersection.

image

## Interface changes

None.

## Effects on system behavior

In scenarios where only some lanes have GREEN signals (e.g., right-turn arrow), undetected lanes are no longer treated as non-red. The crosswalk estimation changes from RED to UNKNOWN in these cases, causing ego to behave cautiously (e.g., stopping for pedestrians) at the crosswalk.

### DIFF
--- a/perception/autoware_crosswalk_traffic_light_estimator/src/crosswalk_traffic_light_estimator.cpp
+++ b/perception/autoware_crosswalk_traffic_light_estimator/src/crosswalk_traffic_light_estimator.cpp
@@ -526,6 +526,11 @@ lanelet::ConstLanelets CrosswalkTrafficLightEstimator::get_non_red_lanelets(
     const auto last_detected_signal =
       get_highest_confidence_traffic_signal(tl_reg_elem->id(), last_detect_color_);
 
+    const auto has_no_signal_info = !current_detected_signal && !last_detected_signal;
+    if (has_no_signal_info) {
+      continue;
+    }
+
     const auto was_not_red = current_is_unknown_or_none && last_detected_signal &&
                              (last_detected_signal.get() == TrafficSignalElement::GREEN ||
                               last_detected_signal.get() == TrafficSignalElement::AMBER) &&


a-maumau (approved):
Thanks for fixing the bug!
LGTM.

Since this PR is to fix the bug, so I only leave some comments.

---
I think we can now do early return by moving the
```cpp
const auto has_no_signal_info = !current_detected_signal && !last_detected_signal;
if (has_no_signal_info) {
  continue;
}
```
after the (or merge it)
```cpp
if (!current_detected_signal && !config_.use_last_detect_color) {
  continue;
}
```
then we can refactor the parts
```cpp
const auto current_is_not_red =
  current_detected_signal ? current_detected_signal.get() == TrafficSignalElement::GREEN || current_detected_signal.get() == TrafficSignalElement::AMBER
          : true;
```
which is quite hard to read the intention.

interimadd:
@a-maumau 
Thank you for the suggestion! I also feel the `get_non_red_lanelets()` function is difficult to understand.
I would like to prioritize fixing the bug. Therefore, I will create another PR to refactor this logic later.

PR #12373, opened 2026-03-24
fix(autoware_pid_longitudinal_controller): fix test for ROS 2 Jazzy

mitsudome-r:
## Description
autoware_pid_longitudinal_controller is failing in ROS 2 Jazzy colcon test CI check. 
https://github.com/autowarefoundation/autoware_universe/actions/runs/23465272858/job/68275989889

This is because rclcpp::Rate requires ROS context to be initialized (i.e. require `rclcpp::init()` ) before declaring the class. Since the test only use Rate for waiting purpose and doesn't require to look for use_sim_time parameter, I've modified to use sleep_for function instead. 

## Related links

**Parent Issue:**

- https://github.com/autowarefoundation/autoware/issues/6695

## How was this PR tested?
Ran colcon test locally in ROS Jazzy environment

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/control/autoware_pid_longitudinal_controller/test/test_smooth_stop.cpp
+++ b/control/autoware_pid_longitudinal_controller/test/test_smooth_stop.cpp
@@ -16,6 +16,8 @@
 #include "gtest/gtest.h"
 #include "rclcpp/rclcpp.hpp"
 
+#include <chrono>
+#include <thread>
 #include <utility>
 #include <vector>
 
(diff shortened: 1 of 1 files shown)


yuki-takagi-66 (approved):
LGTM. However, if you haven’t already, I’d appreciate it if you could run the scenario tests—you can do so along with the other changes in this issue.

PR #12374, opened 2026-03-24
fix(trajectory_traffic_rule_filter): fix test for ROS Jazzy

mitsudome-r:
## Description
It seems like the test for ROS 2 Jazzy tend to fail due to missing map.
https://github.com/autowarefoundation/autoware_core/actions/runs/23329785827/job/68010108438

I've added a has_map function to make sure that map is available before publishing other topics in the test.

## Related links

**Parent Issue:**

- https://github.com/autowarefoundation/autoware/issues/6695

## How was this PR tested?

I have ran colcon test locally in ROS 2 Jazzy environment.
It is also passing the CI check for this PR: https://github.com/autowarefoundation/autoware_universe/actions/runs/23470248716/job/68291164915?pr=12374
## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


go-sakayori:
@maxime-clem 
As you are going to remove this package, can you apply the same change to the PR below.
https://github.com/autowarefoundation/autoware_universe/pull/12369

PR #12378, opened 2026-03-24
fix(autoware_tensorrt_plugins): avoid tv::zeros and tv::empty

veqcc:
## Description

This PR removes `tv::zeros` and `tv::empty`.
This is because they call `cudaMalloc` inside tensorview, which leads to engine build error on NVIDIA DRIVE AGX Thor.

- in `implicit_gemm_plugin.cpp` , the `tv::zeros` call is just moved to the constructor
- in `get_indices_pairs_implicit_gemm_plugin.cpp` , it is a little bit complex
  - Calc additional workspace size for `tv::zeros` and `tv::empty`
  - Call `tv::from_blob` instead of them at runtime
  - Pre-allocate for thrust workspace

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

I have checked compile/engine-build/execution worked well on both x86-64 machine and DRIVE Thor.
Through [TIER IV internal evaluator](https://star4.slack.com/archives/C09J608H59A/p1775620870785979?thread_ts=1774345163.619479&cid=C09J608H59A), there is no regression on inference precision.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/.clang-tidy-ci
+++ b/.clang-tidy-ci
@@ -1,0 +1,34 @@
+Checks: "
+  -*,
+  bugprone-*,
+  -bugprone-assignment-in-if-condition,
+  -bugprone-branch-clone,
+  -bugprone-easily-swappable-parameters,
+  -bugprone-empty-catch,
+  -bugprone-exception-escape,
+  -bugprone-implicit-widening-of-multiplication-result,
+  -bugprone-infinite-loop,
+  -bugprone-integer-division,
+  -bugprone-macro-parentheses,
+  -bugprone-multi-level-implicit-pointer-conversion,
+  -bugprone-narrowing-conversions,
+  -bugprone-optional-value-conversion,
+  -bugprone-parent-virtual-call,
+  -bugprone-reserved-identifier,
+  -bugprone-signed-char-misuse,
+  -bugprone-switch-missing-default-case,
+  -bugprone-unchecked-optional-access,
+  -bugprone-unused-local-non-trivial-variable,
+  -bugprone-unused-return-value"
+
+WarningsAsErrors: "*"
+
+ExtraArgs:
+  - -std=c++17
+  - -Wno-c11-extensions
+  - -Wno-unknown-warning-option
+  - -Wno-deprecated-declarations
+  - -Wno-delete-non-abstract-non-virtual-dtor
+  - -Wno-unused-private-field
+  - -Wno-unused-but-set-variable
+  - -Wno-deprecated-builtins
--- a/.github/workflows/build-and-test-daily.yaml
+++ b/.github/workflows/build-and-test-daily.yaml
@@ -14,9 +14,14 @@ jobs:
         cuda: [false, true]
         arch: [amd64, arm64]
 
-        # Exclude Jazzy arm64 until autoware docker-build-and-push publishes native arm64 Jazzy images
         exclude:
-          - rosdistro: jazzy
+          # Humble is being phased out; only run CUDA amd64 as a minimal check
+          - rosdistro: humble
+            arch: arm64
+          - rosdistro: humble
+            cuda: false
+          # Non-CUDA arm64 has no deployment target; keep non-CUDA only on amd64 as a sanity check
+          - cuda: false
             arch: arm64
 
         include:
@@ -52,9 +57,11 @@ jobs:
       container-suffix: ${{ matrix.container-suffix }}
       runner: ${{ matrix.runner }}
       build-pre-command: ${{ matrix.build-pre-command }}
+      arch: ${{ matrix.arch }}
       pull-ccache: true
       push-ccache: false
       concurrency-group: build-and-test-daily-${{ matrix.rosdistro }}-${{ matrix.arch }}-${{ matrix.cache-suffix }}-${{ github.ref }}-${{ github.run_id }}
-      codecov-flag: daily-${{ matrix.rosdistro }}-${{ matrix.arch }}-${{ matrix.cache-suffix }}
+      upload-coverage: ${{ matrix.rosdistro == 'jazzy' && matrix.arch == 'amd64' && matrix.cuda }}
+      codecov-flag: daily
     secrets:
       codecov-token: ${{ secrets.CODECOV_TOKEN }}
--- a/.github/workflows/build-and-test-differential-arm64.yaml
+++ b/.github/workflows/build-and-test-differential-arm64.yaml
@@ -99,15 +99,5 @@ jobs:
           target-packages: ${{ steps.get-modified-packages.outputs.modified-packages }}
           build-depends-repos: build_depends.repos
 
-      - name: Upload coverage to CodeCov
-        if: ${{ steps.test.outputs.coverage-report-files != '' }}
-        uses: codecov/codecov-action@v4
-        with:
-          files: ${{ steps.test.outputs.coverage-report-files }}
-          fail_ci_if_error: false
-          verbose: true
-          flags: differential-arm64
-          token: ${{ secrets.CODECOV_TOKEN }}
-
       - name: Show disk space after the tasks
         run: df -h
--- a/.github/workflows/build-and-test-differential.yaml
+++ b/.github/workflows/build-and-test-differential.yaml
@@ -134,14 +134,18 @@ jobs:
           build-depends-repos: build_depends.repos
 
       - name: Upload coverage to CodeCov
-        if: ${{ steps.test.outputs.coverage-report-files != '' }}
-        uses: codecov/codecov-action@v4
+        if: ${{ steps.test.outputs.coverage-report-files != '' && inputs.container-suffix == '-cuda' && inputs.rosdistro == 'jazzy' }}
+        uses: codecov/codecov-action@v6
         with:
           files: ${{ steps.test.outputs.coverage-report-files }}
           fail_ci_if_error: false
           verbose: true
-          flags: differential${{ inputs.container-suffix }}
+          flags: full-suite
           token: ${{ secrets.codecov-token }}
+          disable_search: true
+          # Disable gcov and xcode plugins, keeping only pycoverage.
+          # Default plugins: https://github.com/getsentry/prevent-cli/blob/45e518e5ca344c9fb7a5cbe596a21b7ab85a39ed/codecov-cli/README.md?plain=1#L178
+          plugins: pycoverage
 
       - name: Show disk space after the tasks
         run: df -h
--- a/.github/workflows/build-and-test-packages-above-differential.yaml
+++ b/.github/workflows/build-and-test-packages-above-differential.yaml
@@ -26,9 +26,6 @@ on:
         required: false
         default: ""
         type: string
-    secrets:
-      codecov-token:
-        required: true
 
 env:
   CC: /usr/lib/ccache/gcc
@@ -133,16 +130,6 @@ jobs:
           target-packages: ${{ steps.get-modified-packages.outputs.modified-packages }}
           build-depends-repos: build_depends.repos
           packages-above-repos: packages_above.repos
-
-      - name: Upload coverage to CodeCov
-        if: ${{ steps.test.outputs.coverage-report-files != '' }}
-        uses: codecov/codecov-action@v4
-        with:
-          files: ${{ steps.test.outputs.coverage-report-files }}
-          fail_ci_if_error: false
-          verbose: true
-          flags: differential${{ inputs.container-suffix }}
-          token: ${{ secrets.codecov-token }}
 
       - name: Show disk space after the tasks
         run: df -h
--- a/.github/workflows/build-and-test-reusable.yaml
+++ b/.github/workflows/build-and-test-reusable.yaml
@@ -25,6 +25,10 @@ on:
         type: string
         default: ""
         required: false
+      arch:
+        type: string
+        default: amd64
+        required: false
       push-ccache:
         type: boolean
         default: false
@@ -38,6 +42,10 @@ on:
         default: ${{ github.workflow }}-${{ github.ref }}-${{ github.run_id }} # unique per run
         required: false
       cancel-in-progress:
+        type: boolean
+        default: false
+        required: false
+      upload-coverage:
         type: boolean
         default: false
         required: false
@@ -61,8 +69,6 @@ jobs:
   build-and-test:
     runs-on: ${{ fromJson(inputs.runner) }}
     container: ${{ inputs.container }}${{ inputs.container-suffix }}
-    outputs:
-      coverage-files: ${{ steps.test.outputs.coverage-report-files }}
     steps:
       - name: Check out repository
         uses: actions/checkout@v4
@@ -163,22 +169,26 @@ jobs:
           target-packages: ${{ steps.get-self-packages.outputs.self-packages }}
           build-depends-repos: build_depends.repos
 
-      # Cache lcov/coverage directories for next job
-      - name: Upload coverage artifacts
-        if: ${{ steps.test.outputs.coverage-report-files != '' }}
-        uses: actions/upload-artifact@v4
+      - name: Upload coverage to CodeCov
+        if: ${{ steps.test.outputs.coverage-report-files != '' && inputs.upload-coverage }}
+        uses: codecov/codecov-action@v6
         with:
-          name: coverage-${{ inputs.codecov-flag }}-${{ inputs.rosdistro }}
-          path: |
-            build
-            lcov
+          files: ${{ steps.test.outputs.coverage-report-files }}
+          fail_ci_if_error: false
+          verbose: true
+          flags: ${{ inputs.codecov-flag }}
+          token: ${{ secrets.codecov-token }}
+          disable_search: true
+          # Disable gcov and xcode plugins, keeping only pycoverage.
+          # Default plugins: https://github.com/getsentry/prevent-cli/blob/45e518e5ca344c9fb7a5cbe596a21b7ab85a39ed/codecov-cli/README.md?plain=1#L178
+          plugins: pycoverage
 
       - name: Show disk space after the tasks
         run: df -h
 
   clang-tidy:
     needs: build-and-test
-    if: ${{ inputs.rosdistro == 'jazzy' && inputs.container-suffix == '-cuda' }}
+    if: ${{ inputs.rosdistro == 'jazzy' && inputs.container-suffix == '-cuda' && inputs.arch == 'amd64' }}
     runs-on: ubuntu-latest
     container: ${{ inputs.container }}${{ inputs.container-suffix }}
     steps:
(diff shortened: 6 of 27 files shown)


perception/autoware_tensorrt_plugins/src/get_indices_pairs_implicit_gemm_plugin.cpp
```diff
@@ -294,10 +294,37 @@ std::int32_t GetIndicesPairsImplicitGemmPlugin::enqueue(
...
+  if (!is_subm) {
+    thrust_tmp_offset +=
+      static_cast<std::size_t>(kernel_volume) * static_num_act_in * sizeof(tv::int32);
+    thrust_tmp_offset +=
+      static_cast<std::size_t>(mask_count) * static_num_act_in * sizeof(tv::int32);
+    thrust_tmp_offset +=
+      static_cast<std::size_t>(mask_count) * static_num_act_in * sizeof(tv::int32);
+  }
+
+  tv::Tensor thrust_tmp = tv::from_blob(
```
amadeuszsz, line 369: Looking onto code, `thrust_tmp` seems unused. If we use mechanics that discovers this pre-allocated memory (or anything else), please add a detailed comment block 🙏🏻. That one a few lines above is not enough

perception/autoware_tensorrt_plugins/src/get_indices_pairs_implicit_gemm_plugin.cpp
```diff
@@ -404,12 +443,34 @@ std::size_t GetIndicesPairsImplicitGemmPlugin::getWorkspaceSize(
...
+    // pair_mask_bwd_padded: {mask_count, out_indices_num_limit_}
+    workspace_size +=
+      static_cast<std::size_t>(mask_count) * out_indices_num_limit_ * sizeof(tv::int32);
+    // mask_argsort_bwd_padded: {mask_count, out_indices_num_limit_}
+    workspace_size +=
+      static_cast<std::size_t>(mask_count) * out_indices_num_limit_ * sizeof(tv::int32);
+  }
 
-  return static_cast<std::size_t>(workspace_size);
+  return workspace_size;
```
amadeuszsz, line 516: I need to double check, but potentially we need to include `kThrustTempBytes` to workspace size too. Meantime you can take a look on this too and share your thoughts.

perception/autoware_tensorrt_plugins/src/get_indices_pairs_implicit_gemm_plugin.cpp
```diff
@@ -404,12 +443,34 @@ std::size_t GetIndicesPairsImplicitGemmPlugin::getWorkspaceSize(
...
+
   // query workspace size.
-  int workspace_size = SpconvOps::get_indice_gen_workspace_size(
+  workspace_size += static_cast<size_t>(SpconvOps::get_indice_gen_workspace_size(
     kernel_volume, out_indices_num_limit_, out_indices_num_limit_, out_indices_num_limit_, is_subm,
-    use_int64_hash_k, use_direct_table);
+    use_int64_hash_k, use_direct_table));
+
+  // Additional space for indices_kernel_num tensor (kernel_volume * int32).
+  workspace_size += static_cast<std::size_t>(kernel_volume) * sizeof(tv::int32);
```
amadeuszsz: `sizeof(tv::int32)` does not work as you expected because `tv::int32` is enum, so it always return the same value for any type, which is `4`. It matches `int32_t` size, but just by accident 😄 
Please use helper `tv::detail::sizeof_dtype(tv::int32)`.

perception/autoware_tensorrt_plugins/src/get_indices_pairs_implicit_gemm_plugin.cpp
```diff
@@ -294,10 +294,37 @@ std::int32_t GetIndicesPairsImplicitGemmPlugin::enqueue(
...
+  // Allocate indices_kernel_num from the workspace instead of tv::zeros to avoid cudaMalloc during
+  // CUDA graph capture. The extra region sits right after the spconv workspace area.
+  int spconv_ws_size = SpconvOps::get_indice_gen_workspace_size(
+    kernel_volume, out_indices_num_limit_, out_indices_num_limit_, out_indices_num_limit_, is_subm,
+    use_int64_hash_k, use_direct_table);
+  void * indices_kernel_num_ptr = static_cast<std::uint8_t *>(workspace) + spconv_ws_size;
+
+  tv::Tensor indices_kernel_num =
+    tv::from_blob(indices_kernel_num_ptr, {kernel_volume}, tv::int32, 0);
+  cudaMemsetAsync(indices_kernel_num_ptr, 0, kernel_volume * sizeof(tv::int32), stream);
```
amadeuszsz: No error check. But also I see same problem in few other places in this package. Do you think we can check the return value?

perception/autoware_tensorrt_plugins/src/get_indices_pairs_implicit_gemm_plugin.cpp
```diff
@@ -404,12 +443,34 @@ std::size_t GetIndicesPairsImplicitGemmPlugin::getWorkspaceSize(
...
 
+  size_t workspace_size = 0;
+
   // query workspace size.
-  int workspace_size = SpconvOps::get_indice_gen_workspace_size(
+  workspace_size += static_cast<size_t>(SpconvOps::get_indice_gen_workspace_size(
     kernel_volume, out_indices_num_limit_, out_indices_num_limit_, out_indices_num_limit_, is_subm,
-    use_int64_hash_k, use_direct_table);
+    use_int64_hash_k, use_direct_table));
+
```
amadeuszsz, line 498: What alignment returns `SpconvOps::get_intice_gen_workspace_size`? Maybe for any its call in this file we can add alignment, e.g.:
```
workspace_size = tv::align_up(workspace_size, tv::detail::sizeof_dtype(tv::int32));
```

amadeuszsz (changes requested):

perception/autoware_tensorrt_plugins/src/get_indices_pairs_implicit_gemm_plugin.cpp
veqcc: I think we should.
The function `std::int32_t GetIndicesPairsImplicitGemmPlugin::enqueue(` returns `cudaError_t` , so I will follow it and ealry returns on `cudaMemsetAsync` failure
```
  ...
  cudaError_t const status = cudaMemcpyAsync(
    num_act_out_data, &num_act_out_real, sizeof(std::int32_t), cudaMemcpyHostToDevice, stream);

  return status;
```

perception/autoware_tensorrt_plugins/src/get_indices_pairs_implicit_gemm_plugin.cpp
veqcc: Oops, I'm sorry and thank you for finding it.
I'll check it again whether `thrust` workspace is really necessary

perception/autoware_tensorrt_plugins/src/get_indices_pairs_implicit_gemm_plugin.cpp
veqcc: Thank you! I will fix it.

perception/autoware_tensorrt_plugins/src/get_indices_pairs_implicit_gemm_plugin.cpp
veqcc: Fixed in https://github.com/autowarefoundation/autoware_universe/pull/12378/changes/8f8058455aeccf122f86933ad1b2832a898e0c5f

perception/autoware_tensorrt_plugins/src/get_indices_pairs_implicit_gemm_plugin.cpp
veqcc: Fixed in https://github.com/autowarefoundation/autoware_universe/pull/12378/changes/cacd65d58565cf930933da7ed007fc266406b51d

perception/autoware_tensorrt_plugins/src/get_indices_pairs_implicit_gemm_plugin.cpp
veqcc: I found that `alloc.thrust_tmp_tensor_ = thrust_tmp;` was somehow deleted after my first commit (I deleted it but I don't remember why,,)
I have fixed it in https://github.com/autowarefoundation/autoware_universe/pull/12378/changes/f9abbd52c369d6c8b743f7b70e94da3a960f8640

For a detailed comment, please wait :pray:

perception/autoware_tensorrt_plugins/src/get_indices_pairs_implicit_gemm_plugin.cpp
veqcc: You are right.
Somehow it was deleted in my commit, so I restore it in https://github.com/autowarefoundation/autoware_universe/pull/12378/changes/f9abbd52c369d6c8b743f7b70e94da3a960f8640. Thank you!

perception/autoware_tensorrt_plugins/src/get_indices_pairs_implicit_gemm_plugin.cpp
veqcc: `get_indice_gen_workspace_size` returns at least 4-bytes ( = sizeof(int32) ) aligned size as follows.
Do you think we should add an alignment check (or align_up) in `autoware_tensorrt_plugin` ??
```
std::size_t SpconvOps::get_indice_gen_workspace_size(size_t kv, size_t num_act_in, size_t num_act_out_bound, size_t max_act_out_in_theory, bool subm, bool use_int64_hash_k, bool direct_table)   {
  
  int hash_size = 2 * num_act_out_bound;
  if (direct_table){
      hash_size = tv::align_up(int(1.1 * max_act_out_in_theory), 2);
  }
  size_t res = 0;
  if (subm){
      res = hash_size * (use_int64_hash_k ? 3 : 2) * sizeof(int) + 1 * sizeof(int);
  }else{
      size_t pair_single_size = kv * num_act_in; // 40000
      size_t ind_uniq_and_bkp_size = (pair_single_size + 1) * 2 * (use_int64_hash_k ? sizeof(int64_t) : sizeof(int32_t));
      hash_size = hash_size * (use_int64_hash_k ? 3 : 2) * sizeof(int);
      res = ind_uniq_and_bkp_size + hash_size + 1 * sizeof(int);
  }
  return res;
}
```

perception/autoware_tensorrt_plugins/src/get_indices_pairs_implicit_gemm_plugin.cpp
veqcc: Added a detailed comment in https://github.com/autowarefoundation/autoware_universe/pull/12378/changes/a760abfca82448d0daee7cc77f1135f6e8f4bb0d !!

veqcc:
@amadeuszsz 
Thank you for your reviews!! I have addressed all your comments :+1:

perception/autoware_tensorrt_plugins/src/get_indices_pairs_implicit_gemm_plugin.cpp
amadeuszsz: Thank you for confirmation. Let's keep it as it is.

amadeuszsz:
@veqcc 
Before merging, could you please make sure that BEVFusion works after latest changes? Checking x86 (desktop) is enough.

veqcc:
@amadeuszsz 
> Before merging, could you please make sure that BEVFusion works after latest changes? Checking x86 (desktop) is enough.

I have tested the following on x86-64 machine
- source build success
`colcon build --symlink-install --cmake-args -DCMAKE_BUILD_TYPE=Release --packages-up-to autoware_bevfusion`
- engine build success
`ros2 launch autoware_bevfusion bevfusion.launch.xml build_only:=true`
- launch success
`ros2 launch autoware_bevfusion bevfusion.launch.xml` + `ros2 bag play some_rosbag` and `/perception/detection/objects` is published

veqcc:
@amadeuszsz 
I have also checked there is no regression by this PR through [TIER IV internal evaluator](https://star4.slack.com/archives/C09J608H59A/p1775620870785979?thread_ts=1774345163.619479&cid=C09J608H59A)!

perception/autoware_tensorrt_plugins/include/autoware/tensorrt_plugins/get_indices_pairs_implicit_gemm_plugin.hpp
```diff
@@ -125,6 +125,10 @@ class GetIndicesPairsImplicitGemmPlugin : public IPluginV3,
   // upper bound of number of output indices. needed to bound memory usage.
   static constexpr int out_indices_num_limit_{256000};
 
+  // Pre-allocated workspace size for thrust temporary buffer used by spconv sort operations.
+  // 8 MiB is generous enough for CUB radix sort of up to out_indices_num_limit_ elements.
+  static constexpr std::size_t kThrustTempBytes{8U * 1024U * 1024U};
```
KSeangTan, line 130: Any possible way to configure it instead of a constant value?

perception/autoware_tensorrt_plugins/src/get_indices_pairs_implicit_gemm_plugin.cpp
```diff
@@ -294,10 +336,40 @@ std::int32_t GetIndicesPairsImplicitGemmPlugin::enqueue(
...
+  int spconv_ws_size = SpconvOps::get_indice_gen_workspace_size(
+    kernel_volume, out_indices_num_limit_, out_indices_num_limit_, out_indices_num_limit_, is_subm,
+    use_int64_hash_k, use_direct_table);
+  void * indices_kernel_num_ptr = static_cast<std::uint8_t *>(workspace) + spconv_ws_size;
+
+  tv::Tensor indices_kernel_num =
+    tv::from_blob(indices_kernel_num_ptr, {kernel_volume}, tv::int32, 0);
+
+  {
+    cudaError_t status = cudaMemsetAsync(
```
KSeangTan, line 349: Should we have `checkCudaError` here?

perception/autoware_tensorrt_plugins/src/get_indices_pairs_implicit_gemm_plugin.cpp
```diff
@@ -404,12 +490,30 @@ std::size_t GetIndicesPairsImplicitGemmPlugin::getWorkspaceSize(
   int kernel_volume =
     std::accumulate(params_.ksize.begin(), params_.ksize.end(), 1, std::multiplies<int>());
 
-  // query workspace size.
-  int workspace_size = SpconvOps::get_indice_gen_workspace_size(
+  size_t workspace_size = 0;
```
KSeangTan, line 493: should we simply
```
size_t workspace_size{static_cast<size_t>(SpconvOps::get_indice_gen_workspace_size(
    kernel_volume, out_indices_num_limit_, out_indices_num_limit_, out_indices_num_limit_, is_subm,
    use_int64_hash_k, use_direct_table))};
```

KSeangTan (approved):
LGTM overall, please address comments accordingly

perception/autoware_tensorrt_plugins/include/autoware/tensorrt_plugins/get_indices_pairs_implicit_gemm_plugin.hpp
veqcc: Ummm, I think it is possible to make it a parameter and specify it at launch phase, but it's not realistic.
Indeed, `out_indices_num_limit_` is already hard-coded in the header file, which means `kThrustTempBytes` need not be changed unless `out_indices_num_limit_` is updated.
```
  // upper bound of number of output indices. needed to bound memory usage.
  static constexpr int out_indices_num_limit_{256000};

  // Pre-allocated workspace size for thrust temporary buffer used by spconv sort operations.
  // 8 MiB is generous enough for CUB radix sort of up to out_indices_num_limit_ elements.
  static constexpr std::size_t kThrustTempBytes{8U * 1024U * 1024U};
```

perception/autoware_tensorrt_plugins/src/get_indices_pairs_implicit_gemm_plugin.cpp
veqcc: The interface of `GetIndicesPairsImplicitGemmPlugin::enqueue(` is to return cuda status, so I think this is fine.
```
std::int32_t GetIndicesPairsImplicitGemmPlugin::enqueue(
  PluginTensorDesc const * input_desc, [[maybe_unused]] PluginTensorDesc const * output_desc,
  void const * const * inputs, void * const * outputs, [[maybe_unused]] void * workspace,
  cudaStream_t stream) noexcept
{
  ...
  // added code
  {
    cudaError_t status = cudaMemsetAsync(
      indices_kernel_num_ptr, 0, kernel_volume * tv::detail::sizeof_dtype(tv::int32), stream);
    if (status != cudaSuccess) {
      return status;
    }
  }

  ...
  // existing code
  cudaError_t const status = cudaMemcpyAsync(
    num_act_out_data, &num_act_out_real, sizeof(std::int32_t), cudaMemcpyHostToDevice, stream);

  return status;
}
```

perception/autoware_tensorrt_plugins/src/get_indices_pairs_implicit_gemm_plugin.cpp
veqcc: In my opinion, the former is easy to understand. What do you think?

When the size is `A + B + C`, 

my PR
```
size_t workspace_size = 0;
workspace_size += A;
workspace_size += B;
workspace_size += C;
```

your suggestion
```
size_t workspace_size{A};
workspace_size += B;
workspace_size += C;
```

KSeangTan: In terms of efficiency, they are similar, i believe the difference is about the readability, what about:

```
size_t A
size_t B 
size_t C

size_t workspace_size = A + B + C?
```

veqcc: @KSeangTan 
```
size_t A
size_t B 
size_t C

size_t workspace_size = A + B + C
```

looks fine basically, but this case is like

```
size_t workspace_size = 0;

workspace_size += A;

if (some_condition) {
  workspace_size += B;
}

workspace_size += C;
```

so, if it is written in that way,
```
size_t A = ...;
size_t B = some_condition ? ... : 0;
size_t C = ...;

size_t workspace_size = A + B + C
```

I think they are not so different in terms of readability 🤔

KSeangTan: Make sense, please keep it as it is then

PR #12379, opened 2026-03-24
chore(autoware_multi_object_tracker): convert array based config to explicit keys

technolojin:
## Description

This PR refactors configuration and internal typing in autoware_multi_object_tracker to improve clarity, validation, and maintainability.

Key changes:

* Replaced flat tracker parameters with grouped initial_tracker parameters per class.
* Replaced pruning threshold arrays with explicit per-class objects.
* Replaced association matrices with layered association maps:

* Added strict runtime validation for missing/invalid/inconsistent configuration.
* Migrated processor and association logic to typed label/tracker map access.
* Removed dependency on autoware_object_recognition_utils; added tf2_geometry_msgs.
* Updated tests and schemas to match new parameter and type structure.

Breaking changes:

* configuration format is changed
* the association matrix is simplified to set the parameter per tracker type, instead of tracker's highest classification

## Related links

Launch PR https://github.com/autowarefoundation/autoware_launch/pull/1802

## How was this PR tested?

https://github.com/tier4/auto-evaluator/actions/runs/23521553892

## Notes for reviewers

None.

## Interface changes

None.

### ROS Parameter Changes

data_association_matrix.param
- can_assign
- max_dist
- max_area
- min_area
- min_iou

multi_object_tracker_node.param
- pruning_generalized_iou_thresholds
- pruning_distance_thresholds

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_multi_object_tracker/src/processor/processor.cpp
+++ b/perception/autoware_multi_object_tracker/src/processor/processor.cpp
@@ -162,7 +162,7 @@ void TrackerProcessor::spawn(const types::AssociatedObjects & associated_objects
 std::shared_ptr<Tracker> TrackerProcessor::createNewTracker(
   const types::DynamicObject & object, const rclcpp::Time & time) const
 {
-  const Label label = classes::getHighestProbLabel(object.classification);
+  const classes::Label label = classes::getHighestProbLabel(object.classification);
   const auto tracker_type_opt = get_map_value_if_exists(config_.tracker_map, label);
   if (tracker_type_opt) {
     const auto tracker_type = tracker_type_opt->get();
@@ -265,7 +265,7 @@ void TrackerProcessor::mergeOverlappedTracker(const rclcpp::Time & time)
   {
     std::shared_ptr<Tracker> tracker;
     types::DynamicObject object;
-    Label label;
+    classes::Label label;
     bool is_unknown;
     int tracker_priority;
     int measurement_count;


ogrex:
Looks good.
Switching from a matrix-based configuration makes it easier to understand what is set, although applying it to the product may require some additional work. Testing is needed to ensure it works exactly as before.
When adding new tracker in the future (possbile?), updates will be required in multiple places such as the YAML, trackedLabels(), NUM_LABELS,  ALL_TRACKER_TYPES etc.. maybe it is better to document these somewhere

perception/autoware_multi_object_tracker/src/processor/processor.cpp
```diff
@@ -420,7 +416,7 @@ void TrackerProcessor::mergeOverlappedTracker(const rclcpp::Time & time)
     nearby.reserve(16);  // Reasonable initial capacity
 
     Point p1(data1.object.pose.position.x, data1.object.pose.position.y);
-    double max_search_dist_sq = search_distance_sq_per_label[data1.label];
+    const double max_search_dist_sq = search_distance_sq_per_label.at(data1.label);
```
YoshiRi: This is a solid refactoring, but there is a critical runtime safety concern here.
The Label enum in classes.hpp includes labels like ANIMAL and HAZARD, but TRACKED_LABELS only defines 8 classes. In mergeOverlappedTracker, the following code is used:

```cpp
const double max_search_dist_sq = search_distance_sq_per_label.at(data1.label);
```

If the perception module outputs an unsupported label like Label::ANIMAL, std::unordered_map::at() will throw a std::out_of_range exception, which will crash the tracker node.

Suggestion:
To ensure safety, please consider either:

Using .find() instead of .at() here, and falling back to a default value (like the threshold for UNKNOWN) if the label is not found.

perception/autoware_multi_object_tracker/lib/association/association.cpp
YoshiRi, line 1: The hierarchical configuration makes the parameters much easier to read. However, there is a performance concern in the hot path.

Previously, association thresholds (like max_dist) were stored in an Eigen::MatrixXd, allowing fast O(1) array access. With this change, the parameters are stored in a nested map (LabelToTrackerDoubleMap / std::unordered_map). Since DataAssociation::calculateScore is called heavily every frame (num_measurements * num_trackers times), repeatedly performing nested std::unordered_map::find() lookups will likely cause cache misses and significantly increase latency on the vehicle.

How about keep matrices as inside logic and manage parameter with your dict?

technolojin:
@ogrex 
thank you for reviewing

yes, we have to make unified code about the classification.
one of purpose of this work is to know requirement of unification.

perception/autoware_multi_object_tracker/lib/association/association.cpp
technolojin: I updated to use hash in unordered_map. the cost is go down to O(1)

perception/autoware_multi_object_tracker/src/processor/processor.cpp
technolojin: the access method is unified. if the given label is not exist, the process will passed. 

current out-of-range labels can cause unexpected behavior. 
further consider and development will continue.

PR #12380, opened 2026-03-24
refactor(autoware_crosswalk_traffic_light_estimator): improve readability of get_non_red_lanelets

interimadd:
## Description

Refactor `get_non_red_lanelets` in `autoware_crosswalk_traffic_light_estimator` to improve readability.

- Extract `is_lanelet_non_red` free function from the complex loop body in `get_non_red_lanelets`, reducing it to a simple filter loop.
- Extract `is_green_or_amber` and `is_unknown_or_absent` helper functions that accept `std::optional<uint8_t>`, encapsulating null-handling and color classification.
- Replace `boost::optional` with `std::optional` in `get_highest_confidence_traffic_signal`, removing the `boost/optional.hpp` dependency.
- Fix a latent bug where absent `current_color` was unconditionally treated as non-red (bypassing the `last_detect_color` check), by aligning absent and UNKNOWN handling.

## Related links

https://github.com/autowarefoundation/autoware_universe/pull/12370

## How was this PR tested?

- Built the package successfully with `colcon build --packages-select autoware_crosswalk_traffic_light_estimator`
- All 7 tests passed (3 gtest + 4 linter) with `colcon test --packages-select autoware_crosswalk_traffic_light_estimator`
- (TIER IV internal) Evaluator

## Notes for reviewers

- The bug fix changes behavior when `current_color` is absent and `use_last_detect_color` is true: previously the lanelet was always considered non-red regardless of `last_color`, now it falls through to check whether `last_color` is green or amber.

## Interface changes

None.

## Effects on system behavior

- When `use_last_detect_color` is enabled and no current detection exists, lanelets are now only considered non-red if the last detected color was green or amber (previously any last color was accepted).


perception/autoware_crosswalk_traffic_light_estimator/src/crosswalk_traffic_light_estimator.cpp
```diff
@@ -494,53 +493,61 @@ void CrosswalkTrafficLightEstimator::set_crosswalk_traffic_signal(
...
+  if (!color) {
+    return true;
+  }
+  return *color == TrafficSignalElement::UNKNOWN;
+}
 
-    const auto current_is_unknown_or_none =
-      current_detected_signal ? current_detected_signal.get() == TrafficSignalElement::UNKNOWN
-                              : true;
+bool is_lanelet_non_red(
```
a-maumau, line 521: I love to see some comment how this logic behave with the inputs.

| current color | last color | result |
|---|---|---|
| red | - | false |
| green / amber | - | true |
| unknown / NA | green / amber | true |
| ... | ... | false |
| ... | ... | ... |

interimadd: Thank you for comment! I have addressed it in https://github.com/autowarefoundation/autoware_universe/pull/12380/commits/694f8a5873be6f1a5d0018f067a441fbaddf9d9f

a-maumau: thank you!

PR #12383, opened 2026-03-25
refactor(`simple_planning_simulator`): remove `tier4_api_utils` dependency from `autoware_simple_planning_simulator`

sasakisasaki:
## Description
- Drop `tier4_api_utils` usage in `simple_planning_simulator`.

Let me quote @isamu-takagi san's comment. We do not need to set pose via simulator.
> The service `/api/simulator/set/pose` is probably not being used, so it should be possible to simply delete it. According to [this diagram](https://github.com/autowarefoundation/autoware_universe/blob/main/system/autoware_default_adapi_universe/document/localization.md), the simulator receives `/initialpose3d` from the pose_initializer. Also, since the service `/localization/initialize` is provided by pose_initializer, it should not be provided by the simulator.

## Related links
- https://github.com/orgs/autowarefoundation/discussions/6892

## How was this PR tested?

Based on this PR, I built all the autoware packages based on [the nightly installation instructions](https://autowarefoundation.github.io/autoware-documentation/main/installation/autoware/source-installation/#prerequisites).

Note that I edited `autoware-nightly.repos` and replaced the remote URL and branch to point to my fork and branch for this PR, so that the source code of this PR is included in the build.

```bash
```

After build, I followed the instruction in [this planning simulator demo](https://autowarefoundation.github.io/autoware-documentation/main/demos/planning-sim/), and confirmed that the simulator still works as follows.
```bash
$ source ~/autoware/install/setup.bash
$ ros2 launch autoware_launch planning_simulator.launch.xml map_path:=$HOME/autoware_map/sample-map-planning vehicle_model:=sample_vehicle sensor_model:=sample_sensor_kit
```
During the planning simulation, I checked by:
- Initialize pose
- Set route
- Run autonomous driving
- Put dummy objects: pedestrian, vehicle, car, and bus

#### Additions and removals

| Change type | Topic Type | Topic Name               | Message Type                                           | Description                         |
|:------------|:-----------|:-------------------------|:-------------------------------------------------------|:------------------------------------|
| Removed     | Srv        | `/api/simulator/set/pose` | `tier4_external_api_msgs/srv/InitializePose`           | Legacy simulator pose initialization |

#### Modifications

None.

#### Additions and removals

None.

#### Modifications

None.

## Effects on system behavior
No functional changes expected for simulation; this only drops a deprecated dependency.

### DIFF
--- a/simulator/autoware_simple_planning_simulator/include/autoware/simple_planning_simulator/simple_planning_simulator_core.hpp
+++ b/simulator/autoware_simple_planning_simulator/include/autoware/simple_planning_simulator/simple_planning_simulator_core.hpp
@@ -15,14 +15,11 @@
 #ifndef AUTOWARE__SIMPLE_PLANNING_SIMULATOR__SIMPLE_PLANNING_SIMULATOR_CORE_HPP_
 #define AUTOWARE__SIMPLE_PLANNING_SIMULATOR__SIMPLE_PLANNING_SIMULATOR_CORE_HPP_
 
-#include "autoware/component_interface_specs/localization.hpp"
 #include "autoware/simple_planning_simulator/vehicle_model/sim_model_interface.hpp"
 #include "autoware/simple_planning_simulator/visibility_control.hpp"
 #include "rclcpp/rclcpp.hpp"
 
-#include "autoware_common_msgs/msg/response_status.hpp"
 #include "autoware_control_msgs/msg/control.hpp"
-#include "autoware_localization_msgs/srv/initialize_localization.hpp"
 #include "autoware_map_msgs/msg/lanelet_map_bin.hpp"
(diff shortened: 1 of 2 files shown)


isamu-takagi:
The service `/api/simulator/set/pose` is probably not being used, so it should be possible to simply delete it. According to [this diagram](https://github.com/autowarefoundation/autoware_universe/blob/main/system/autoware_default_adapi_universe/document/localization.md), the simulator receives `/initialpose3d` from the pose_initializer. Also, since the service `/localization/initialize` is provided by pose_initializer, it should not be provided by the simulator.

sasakisasaki:
@isamu-takagi Exactly, you are right. In actual, when I applied the change as this PR, the planning simulator still worked.

Could you do review again? Thank you in advance.

PR #12392, opened 2026-03-30
refactor(autoware_ground_segmentation/launch, autoware_probabilistic_occupancy_grid_map/launch): use launch substitution instead of get_package_share_directory

paulsohn:
## Description

Use high-level API: `FindPackageShare` from `launch.substitutions` and `PathJointSubstitution` from `launch_ros.substitutions` in `launch.py` files, instead of raw `ament_index_python` and `os.path.join`.

Also, obsolote default of `updater_param_file` in  `perception/autoware_probabilistic_occupancy_grid_map/launch/laserscan_based_occupancy_grid_map.launch.py` is updated to `"binary_bayes_filter_updater.param.yaml"`.

This PR aims to follow proper semantics of launcher system which is open to the launcher-specific tooling / analysis, without handling the full python semantics itself.

Will make a separate PR for sensing-related refactor.

## Related links

https://github.com/autowarefoundation/autoware_launch/pull/1762
https://github.com/autowarefoundation/autoware_launch/pull/1799

## How was this PR tested?

Built and ran all four: no launcher-side error was found.

```
$ ros2 launch autoware_ground_segmentation scan_ground_filter.launch.py
$ ros2 launch autoware_probabilistic_occupancy_grid_map laserscan_based_occupancy_grid_map.launch.py
$ ros2 launch autoware_probabilistic_occupancy_grid_map multi_lidar_pointcloud_based_occupancy_grid_map.launch.py
$ ros2 launch autoware_probabilistic_occupancy_grid_map pointcloud_based_occupancy_grid_map.launch.py
```

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None. When launched with standard `ros2 launch`, this should give identical result before and after the PR. 
The only difference is that default value of `updater_param_file` is no longer `"updater.param.yaml"` (which is incorrect, as the file does not exist) but existing `"binary_bayes_filter_updater.param.yaml"`.

### DIFF
--- a/evaluator/autoware_planning_evaluator/src/planning_evaluator_node.cpp
+++ b/evaluator/autoware_planning_evaluator/src/planning_evaluator_node.cpp
@@ -149,8 +149,12 @@ PlanningEvaluatorNode::~PlanningEvaluatorNode()
     json output_json;
     for (OutputMetric metric : metrics_for_output_) {
       const json j = metrics_accumulator_.getOutputJson(metric);
-      if (!j.empty()) {
-        output_json[output_metric_to_str.at(metric)] = j;
+      if (j.empty()) {
+        continue;
+      }
+      const std::string base_name = output_metric_to_str.at(metric) + "/";
+      for (const auto & item : j.items()) {
+        output_json[base_name + item.key()] = item.value();
       }
     }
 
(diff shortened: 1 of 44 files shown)


badai-nguyen:
@paulsohn PR still is blocked because missing approval from codeowner of `autoware_probabilistic_occupancy_grid_map` 
@YoshiRi could you review this PR?

PR #12393, opened 2026-03-30
feat(blockage_diag): apply agnocast subscription to `blockage_diag`

Koichi98:
## Description
Applied `autoware_agnocast_wrapper` to the subscriptions of `BlockageDiagComponent` and `PolarVoxelOutlierFilterComponent` for the `pointcloud_raw_ex` topic. Both nodes subscribe to `pointcloud_raw_ex` published by the Nebula driver (HesaiRosWrapper), which resides in the same `lidar_specific_container` component container.                                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                                                                                                                                                                         
### Changes                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        
  **BlockageDiagComponent:**                                
  - Replaced `rclcpp::Subscription` with `AUTOWARE_CREATE_SUBSCRIPTION` / `AUTOWARE_SUBSCRIPTION_PTR` / `AUTOWARE_MESSAGE_CONST_SHARED_PTR` wrapper macros.
                                                                                                                                                                                                                                                                                                                                                                                                                                         
  **PolarVoxelOutlierFilterComponent:**                                                                                                                                                                                                                                                                                                                                                                                                  
  - Since this node inherits from the `Filter` base class which manages its own `rclcpp::Subscription` (`sub_input_`), the base class subscription is reset and replaced with an agnocast-aware subscription in the derived class constructor.                                                                                                                                                                                           
  - An aliasing `shared_ptr` bridge is used to convert `AUTOWARE_MESSAGE_CONST_SHARED_PTR` to `PointCloud2ConstPtr` without copying, maintaining compatibility with the existing `Filter` callback chain.                                                                                                                                                                                                                                
  - This workaround will be unnecessary once the `Filter` base class itself is migrated to `agnocast_wrapper`.                                                                                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                                                                                                                                                                         
### Motivation: Avoiding the Agnocast-ROS 2 Bridge for intra-container communication                                                                                                                                                                                                                                                                                                                                                   
The [Agnocast-ROS 2 Bridge](https://github.com/autowarefoundation/agnocast/blob/main/docs/agnocast_ros2_bridge.md) now **always performs inter-process communication** (serialization/deserialization) when forwarding messages between Agnocast and ROS 2 endpoints. This means that even when the publisher and subscriber reside in the **same component container** — where rclcpp would normally use zero-copy intra-process pointer passing — the bridge introduces unnecessary serialization overhead.
                                                                                                                                                                                                                                                                                                                                                                                                                                         
By migrating these subscribers to `agnocast_wrapper`, both the publisher (Nebula) and subscribers (`blockage_diag`, `polar_voxel_outlier_filter`) communicate via Agnocast's shared memory, bypassing the bridge entirely. The current setup with rclcpp subscribers does work, but it forces the message through the bridge path, which is suboptimal for same-container pub/sub pairs.                                                   

## Related links

**Parent Issue:**

- Link

<!-- ⬇️🟢
**Private Links:**

- [CompanyName internal link]()
⬆️🟢 -->

## How was this PR tested?
I've tested the following with the internal product.                                                      
  - Build with `ENABLE_AGNOCAST=1` and verify blockage_diag and polar_voxel_outlier_filter subscribe via Agnocast (check `agnocast topic list`)                                                                                                                                                                                                                                                                                      
  - Build with `ENABLE_AGNOCAST=0` and verify fallback to standard rclcpp subscription

## Notes for reviewers

Currently, I've suppressed the `unknownMacro` warning from cppcheck. I will suppress this for `autowrare_agnocast_wrapper` related macro somehow by fixing `.github/workflows`. Will delete `cppcheck-suppress` from these files once I've done it.

## Interface changes

None.

<!-- ⬇️🔴

### Topic changes

#### Additions and removals

#### Modifications

### ROS Parameter Changes

#### Additions and removals

#### Modifications

| Old     | `old_param_name` | `double` | `1.0`         | Param description |
| New     | `new_param_name` | `double` | `1.0`         | Param description |

🔴⬆️ -->

## Effects on system behavior

None.

### DIFF
--- a/sensing/autoware_pointcloud_preprocessor/src/blockage_diag/blockage_diag_node.cpp
+++ b/sensing/autoware_pointcloud_preprocessor/src/blockage_diag/blockage_diag_node.cpp
@@ -249,8 +249,8 @@ void BlockageDiagComponent::publish_blockage_debug_info(
   }
 }
 
-// cppcheck-suppress unknownMacro
 void BlockageDiagComponent::update_diagnostics(
+  // cppcheck-suppress unknownMacro
   AUTOWARE_MESSAGE_CONST_SHARED_PTR(sensor_msgs::msg::PointCloud2) input)
 {
   try {


vividf:
kindly ping @manato @badai-nguyen @mojomex

sensing/autoware_pointcloud_preprocessor/CMakeLists.txt
```diff
@@ -229,9 +233,19 @@ rclcpp_components_register_node(pointcloud_preprocessor_filter
   EXECUTABLE distortion_corrector_node)
 
 # ========== Blockage Diagnostics ===========
-rclcpp_components_register_node(pointcloud_preprocessor_filter
+target_include_directories(pointcloud_preprocessor_filter PRIVATE
+  ${agnocastlib_INCLUDE_DIRS}
+  ${autoware_agnocast_wrapper_INCLUDE_DIRS}
+)
+ament_target_dependencies(pointcloud_preprocessor_filter
+  autoware_agnocast_wrapper
```
mojomex, line 241: This should make the above `_INCLUDE_DIRS` obsolete since the internally-called target_link_libraries also handles headers. Consider removing above.

mojomex: Note that `ament_target_dependencies` links ALL targets of a CMake package, so `target_link_libraries` with `autoware_agnocast_wrapper::autoware_agnocast_wrapper` might be preferred.
I haven't looked into whether Agnocast packages export modern CMake targets, but most packages nowadays do without modification.

sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/outlier_filter/polar_voxel_outlier_filter_node.hpp
```diff
@@ -256,6 +257,10 @@ class PolarVoxelOutlierFilterComponent : public autoware::pointcloud_preprocesso
   diagnostic_updater::Updater updater_;
   OnSetParametersCallbackHandle::SharedPtr set_param_res_;
 
+  // TODO(Koichi98): Remove once Filter base class supports agnocast_wrapper.
+  // USE_AGNOCAST_ENABLED should not appear in application code; this is a temporary workaround.
```
mojomex: `USE_AGNOCAST_ENABLED` is not appearing here, is this an old comment?

Koichi98: Thanks you. Deleted!

sensing/autoware_pointcloud_preprocessor/CMakeLists.txt
Koichi98: `autoware_agnocast_wrapper` doesn't seem to have modern CMake targets.
I will work on to support this ASAP, and will fix this CMakeLitst.txt too.

mojomex: Super low priority though! Just something to think about moving forward. Since you have a deadline to meet and this approach is not wrong per se, I'd suggest ignoring this one.

Koichi98: Let me do this in the follow up PR.

PR #12401, opened 2026-03-31
fix(autoware_fault_injection): wait for endpoint discovery before publishing in tests

xmfcx:
- Replace single `spin_once` with explicit DDS endpoint discovery polling in launch tests
- Add `_wait_for_discovery` helper that checks `get_subscription_count()` and `count_publishers()` before proceeding
- Apply discovery wait to both `test_node_link` and `test_receive_multiple_message_simultaneously`

## Why

`test_receive_multiple_message_simultaneously` fails on Jazzy because messages are published before DDS endpoint discovery completes. The previous single `spin_once` was insufficient for Jazzy's middleware timing. Polling for actual endpoint matches makes the test deterministic across distros.

---

- Related CI failure: https://github.com/autowarefoundation/autoware_universe/actions/runs/23807891783/job/69386438893

## Test plan

- [x] `colcon build --symlink-install --packages-select autoware_fault_injection && colcon test --packages-select autoware_fault_injection --return-code-on-test-failure && colcon test-result --verbose --all` - all 6 tests pass, 0 failures
- [x] CI passes on both Humble and Jazzy
- [ ] Full build test: https://github.com/autowarefoundation/autoware_universe/actions/runs/23817370303

### DIFF
--- a/simulator/autoware_fault_injection/test/test_fault_injection_node.test.py
+++ b/simulator/autoware_fault_injection/test/test_fault_injection_node.test.py
@@ -68,11 +68,25 @@ class TestFaultInjectionLink(unittest.TestCase):
         self.test_node = rclpy.create_node("test_node")
         self.event_name = "cpu_temperature_is_high"
         self.evaluation_time = 0.5  # 500ms
+        self.discovery_timeout = 5.0  # seconds
         self.input_diagnostics_topic = "/diagnostics"
         self.output_diagnostics_topic = "/diagnostics/fault_injection"
 
     def tearDown(self):
         self.test_node.destroy_node()
+
+    def _wait_for_discovery(self, publishers, subscribers=None):
+        """Wait until all publishers have matched with at least one subscriber and vice versa."""
+        end_time = time.time() + self.discovery_timeout
+        while time.time() < end_time:
+            pubs_matched = all(pub.get_subscription_count() > 0 for pub in publishers)
+            subs_matched = subscribers is None or all(
+                self.test_node.count_publishers(topic) > 0 for topic in subscribers
+            )
+            if pubs_matched and subs_matched:
+                return
+            rclpy.spin_once(self.test_node, timeout_sec=0.1)
+        self.fail("Timed out waiting for endpoint discovery")
 
     @staticmethod
     def print_message(stat):
@@ -108,6 +122,12 @@ class TestFaultInjectionLink(unittest.TestCase):
         msg_buffer = []
         self.test_node.create_subscription(
             DiagnosticArray, self.output_diagnostics_topic, lambda msg: msg_buffer.append(msg), 10
+        )
+
+        # Wait for endpoint discovery before testing.
+        self._wait_for_discovery(
+            publishers=[pub_events, pub_diagnostics],
+            subscribers=[self.output_diagnostics_topic],
         )
 
         # Test init state.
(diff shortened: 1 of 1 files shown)


xmfcx:
# DDS Configuration Investigation in CI Containers

## Background

These issues stem from DDS not being properly set up for the CI machines. Neither of the following environment variables are set correctly in the images run directly by the GitHub Actions runner (which mounts its own custom home directory):

- `CYCLONEDDS_URI=file:///home/mfc/cyclonedds.xml`
- `RMW_IMPLEMENTATION=rmw_cyclonedds_cpp`

### Verification

Running the following command confirms the variables are unset:

```bash
docker run --rm ghcr.io/autowarefoundation/autoware:universe-devel-jazzy \
  bash -c 'echo "RMW_IMPLEMENTATION=$RMW_IMPLEMENTATION"; \
           echo "CYCLONEDDS_URI=$CYCLONEDDS_URI"; \
           dpkg -l | grep rmw-cyclone; \
           echo "---"; \
           cat ~/.bashrc 2>/dev/null | grep -i "rmw\|cyclone" || echo "No RMW in .bashrc"'
```

### Root Cause

The exports live in `/etc/skel/.bashrc`, but that file only matters when new users are created — the skeleton is copied to the new user's home directory. Since the GitHub Actions runner mounts its own home directory, the runner user never receives the skel-based `.bashrc`.

## Default RMW Middleware per Distro

Both Humble and Jazzy default to **`rmw_fastrtps_cpp`** (FastRTPS). Since `RMW_IMPLEMENTATION` is never actually set in CI, both distros run on this default rather than CycloneDDS.

Despite both using FastRTPS, the versions differ significantly: Humble ships with FastRTPS **6.2.10** while Jazzy ships with **8.4.3**. The newer version on Jazzy has different endpoint discovery timing. The test's single `spin_once` before publishing was enough time for discovery on Humble's older FastRTPS, but not on Jazzy's — so messages were published before the fault injection node's subscribers matched, and they were silently dropped.

## Investigation Commands

### Check Default RMW on Humble

```bash
docker run --rm ghcr.io/autowarefoundation/autoware:universe-devel \
  bash -c 'source /opt/ros/*/setup.bash && \
           echo "ROS_DISTRO=$ROS_DISTRO" && \
           ros2 doctor --report 2>/dev/null | grep -i "middleware\|rmw\|dds"'
```

Output:

```
ROS_DISTRO=humble
rmw_dds_common                                                   : latest=1.6.0, local=1.6.0
rmw                                                              : latest=6.1.2, local=6.1.2
rmw_implementation                                               : latest=2.8.5, local=2.8.5
rmw_fastrtps_shared_cpp                                          : latest=6.2.10, local=6.2.10
rmw_implementation_cmake                                         : latest=6.1.2, local=6.1.2
rmw_fastrtps_cpp                                                 : latest=6.2.10, local=6.2.10
rmw_cyclonedds_cpp                                               : latest=1.3.4, local=1.3.4
   RMW MIDDLEWARE
middleware name    : rmw_fastrtps_cpp
```

### Check Default RMW on Jazzy

```bash
docker run --rm ghcr.io/autowarefoundation/autoware:universe-devel-jazzy \
  bash -c 'source /opt/ros/*/setup.bash && \
           echo "ROS_DISTRO=$ROS_DISTRO" && \
           ros2 doctor --report 2>/dev/null | grep -i "middleware\|rmw\|dds"'
```

Output:

```
ROS_DISTRO=jazzy
rmw_dds_common                                                   : latest=3.1.1, local=3.1.1
rmw                                                              : latest=7.3.3, local=7.3.3
rmw_implementation                                               : latest=2.15.6, local=2.15.6
rmw_fastrtps_shared_cpp                                          : latest=8.4.3, local=8.4.3
rmw_implementation_cmake                                         : latest=7.3.3, local=7.3.3
rmw_fastrtps_cpp                                                 : latest=8.4.3, local=8.4.3
rmw_cyclonedds_cpp                                               : latest=2.2.3, local=2.2.3
   RMW MIDDLEWARE
middleware name    : rmw_fastrtps_cpp
```

## Results Summary

|                            | Humble (`universe-devel`) | Jazzy (`universe-devel-jazzy`) |
| -------------------------- | ------------------------- | ------------------------------ |
| ROS distro                 | humble                    | jazzy                          |
| Default middleware         | `rmw_fastrtps_cpp`        | `rmw_fastrtps_cpp`             |
| rmw_fastrtps_cpp version   | 6.2.10                    | 8.4.3                          |
| rmw_cyclonedds_cpp version | 1.3.4                     | 2.2.3                          |

## Conclusion

Both distros use `rmw_fastrtps_cpp` as the default middleware — not CycloneDDS. The `RMW_IMPLEMENTATION=rmw_cyclonedds_cpp` export exists in the container's `.bashrc` and `/etc/skel/.bashrc`, but neither is sourced in non-interactive shells (which is how GitHub Actions runs commands). CI therefore falls back to the default `rmw_fastrtps_cpp` on both distros.

The test flakiness is caused by discovery timing differences between FastRTPS 6.2.10 (Humble) and 8.4.3 (Jazzy). The newer version has different endpoint discovery timing, causing published messages to be silently dropped when the test doesn't wait for discovery to complete.

KeisukeShima:
@xmfcx Thank you for your thorough investigation and explanation :100:

PR #12403, opened 2026-04-01
fix(autoware_freespace_planning_algorithms): check collision at intermediate arc points during A* expansion

xmfcx:
## Summary
- When `adapt_expansion_distance` is enabled, A* node expansion only checked collision at arc endpoints. `setPath()` later interpolates waypoints between nodes, but these intermediate poses were never validated during search.
- On arm64, floating-point differences caused the planner to produce paths where interpolated waypoints clipped obstacles near tight gaps (between parked cars), failing the post-hoc `hasObstacleOnTrajectory` check in `AstarSearchTestSuite.MultiCurvature`.
- This adds intermediate collision checks along the arc in `expandNodes`, using the same `min_expansion_dist_` spacing that `setPath` uses for interpolation. Expansions where any intermediate point collides are now rejected, ensuring the output trajectory is collision-free on all architectures.

## Test plan
- [x] `colcon build --packages-select autoware_freespace_planning_algorithms` passes
- [x] `colcon test --packages-select autoware_freespace_planning_algorithms` passes (all 9 tests, 0 failures)
- [x] `pre-commit` passes
- [x] Verify arm64 CI passes: https://github.com/autowarefoundation/autoware_universe/actions/runs/23863866816
  - It actually passed for arm finally 🥳 https://github.com/autowarefoundation/autoware_universe/actions/runs/23863866816/job/69577508424#step:17:9771
  - The failures are not related to this PR's scope.
    
image

### DIFF
--- a/planning/autoware_freespace_planning_algorithms/src/astar_search.cpp
+++ b/planning/autoware_freespace_planning_algorithms/src/astar_search.cpp
@@ -338,6 +338,23 @@ void AstarSearch::expandNodes(AstarNode & current_node, const bool is_back)
     AstarNode * next_node = &graph_[getKey(next_index)];
     if (next_node->status == NodeStatus::Closed || detectCollision(next_index)) continue;
 
+    // Check intermediate points along the arc for collision
+    const double abs_distance = std::abs(distance);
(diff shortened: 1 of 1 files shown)


xmfcx:
fix_before_after

This should explain the fix with a before & after.

PR #12406, opened 2026-04-02
perf(lidar_centerpoint): remove atomic functions by using warp shuffle primitives

nishikawa-masaki:
## Description

This PR improves performance of pre-processing for autoware_lidar_centerpoint by removing unnecessary atomic read/modify/write functions and replacing some of them with CUDA shuffle warp-primitives.

## Related links

**Parent Issue:**

- None

**Private Links:**

- [TIER IV internal link](https://tier4.atlassian.net/browse/T4DEV-51246)

## How was this PR tested?
- [ ] Build passed
- [ ] Existing unit tests passed
  - unit tests that require GPU are disabled and obsolete
- [ ] LiDAR data are processed correctly
  - Objects can be identified in the same way as before

## Notes for reviewers

This PR can be an example of performance improvement of data processing with CUDA.
The code looks a bit tricky, but the techniques used here are orthodox and proven.

Other functions than generateFeatures_kernel() can be improved more, but changes are kept minimum in this PR.

## Interface changes

None.

## Effects on system behavior

None.

The performance of pre-processing for autoware_lidar_centerpoint is improved.

### DIFF
--- a/perception/autoware_lidar_centerpoint/lib/preprocess/preprocess_kernel.cu
+++ b/perception/autoware_lidar_centerpoint/lib/preprocess/preprocess_kernel.cu
@@ -291,7 +291,8 @@ __global__ void generateFeatures_kernel(
   float y_val = (point_idx < points_num) ? pillarSM[pillar_idx_inBlock][1][point_idx] : 0.0f;
   float z_val = (point_idx < points_num) ? pillarSM[pillar_idx_inBlock][2][point_idx] : 0.0f;
   // 2. Parallel Reduction using Warp Shuffle
-  // Note that all threads in the warp participate in the reduction as there is no thread divergence.
+  // Note that all threads in the warp participate in the reduction as there is no thread
+  // divergence.
 #pragma unroll
   for (int offset = 16; offset > 0; (offset = offset >> 1)) {
     x_val += __shfl_down_sync(ALL_THREADS_IN_WARP_MASK, x_val, offset);


perception/autoware_lidar_centerpoint/lib/preprocess/preprocess_kernel.cu
```diff
@@ -38,8 +38,10 @@
 
 namespace
 {
-const std::size_t MAX_POINT_IN_VOXEL_SIZE = 32;  // the same as max_point_in_voxel_size_ in config
-const std::size_t WARPS_PER_BLOCK = 4;
+const std::size_t NUM_THREADS_IN_WARP = 32;
+const std::size_t MAX_POINT_IN_VOXEL_SIZE =
```
nishikawa-masaki, line 42: Self-comment:
To get better performance, the maximum number of points in a voxel is 32 (the number of threads in a warp) inherited from the former implementation.
If it was required to change the value, implementation would need considerable changes.

perception/autoware_lidar_centerpoint/lib/preprocess/preprocess_kernel.cu
```diff
@@ -95,7 +97,9 @@ __global__ void shufflePoints_kernel(
   const std::size_t points_size, const std::size_t max_size, const std::size_t offset)
 {
   int point_idx = blockIdx.x * blockDim.x + threadIdx.x;
-  if (point_idx >= max_size) return;
+  if (point_idx >= max_size) {
```
nishikawa-masaki, line 100: Self-comment:
The recent version of Google C++ style guide allows the previous style (without braces), but I prefer this style (with braces even if only 1 line inside) because it is more robust for future changes.
(Google changed the rule in 2004)

perception/autoware_lidar_centerpoint/lib/preprocess/preprocess_kernel.cu
```diff
@@ -126,40 +130,45 @@ __global__ void generateVoxels_random_kernel(
...
   int voxel_idx = floorf((x - min_x_range) / pillar_x_size);
   int voxel_idy = floorf((y - min_y_range) / pillar_y_size);
   voxel_idx = voxel_idx < 0 ? 0 : voxel_idx >= grid_x_size ? grid_x_size - 1 : voxel_idx;
   voxel_idy = voxel_idy < 0 ? 0 : voxel_idy >= grid_y_size ? grid_y_size - 1 : voxel_idy;
   unsigned int voxel_index = (grid_x_size - 1 - voxel_idx) * grid_y_size + voxel_idy;
 
-  unsigned int point_id = atomicAdd(&(mask[voxel_index]), 1);
-
-  if (point_id >= MAX_POINT_IN_VOXEL_SIZE) return;
+  // point_id must be in the range of [0, MAX_POINT_IN_VOXEL_SIZE)
```
nishikawa-masaki: Self-comment:
The previous implementation keeps incrementing "mask[voxel_index]".
It leaves a room for "wrapping around" (not likely to happen).

For clarity, atomicInc() is more preferable.

perception/autoware_lidar_centerpoint/lib/preprocess/preprocess_kernel.cu
```diff
@@ -126,40 +130,45 @@ __global__ void generateVoxels_random_kernel(
...
+  }
 
   float * address =
     voxels + (voxel_index * MAX_POINT_IN_VOXEL_SIZE + point_id) * POINT_NUM_FEATURES;
-  atomicExch(address, x);
-  atomicExch(address + 1, y);
-  atomicExch(address + 2, z);
+  address[0] = x;
+  address[1] = y;
+  address[2] = z;
```
nishikawa-masaki, line 163: Self-comment:
As the voxel is exclusively reserved by the previous atomic function, there is no need to use quite costly "atomicExch()".

perception/autoware_lidar_centerpoint/lib/preprocess/preprocess_kernel.cu
```diff
@@ -206,9 +222,6 @@ __global__ void generateBaseFeatures_kernel(
       voxel_features[outIndex * 5 + 4] = voxels[inIndex * 5 + 4];
     }
   }
-
-  // clear buffer for next infer
-  atomicExch(mask + voxel_index, 0);
```
nishikawa-masaki, line 211: Self-comment:
There is no need to clear the buffer with quite costly atomicExch() here.

The caller should clear the buffer and it does.
In addition to that, if this function returns earlier, the buffer is left uncleared.

perception/autoware_lidar_centerpoint/lib/preprocess/preprocess_kernel.cu
```diff
@@ -220,128 +233,161 @@ __global__ void generateFeatures_kernel(
...
   }
-
+  constexpr uint32_t ALL_THREADS_IN_WARP_MASK = 0xffffffff;  // 32 threads
+  points_num = __shfl_sync(ALL_THREADS_IN_WARP_MASK, points_num, 0);
+  cords.x = __shfl_sync(ALL_THREADS_IN_WARP_MASK, cords.x, 0);
+  cords.y = __shfl_sync(ALL_THREADS_IN_WARP_MASK, cords.y, 0);
+  cords.z = __shfl_sync(ALL_THREADS_IN_WARP_MASK, cords.z, 0);
+
+  // Copy data from global memory to shared memory with transposition for coalesced access.
+  {
```
nishikawa-masaki, line 272: Self-comment:
This brace pair is to show the range of life of the variable explicitly to the compiler so that it can allocate registers efficiently.

It might be better to implement this block as a separated \_\_device\_\_ function, but there are pros and cons.

perception/autoware_lidar_centerpoint/lib/preprocess/preprocess_kernel.cu
```diff
@@ -220,128 +233,161 @@ __global__ void generateFeatures_kernel(
...
+      const int dst_point = point_offset / point_dim;
+      const int dst_dim = point_offset % point_dim;
+      const int dst_index = dst_dim * MAX_POINT_IN_VOXEL_SIZE + dst_point;
+      const int src_index = i * MAX_POINT_IN_VOXEL_SIZE + point_idx;
+      ((float *)pillarSM)[dst_offset + dst_index] =
+        ((float *)voxel_features)[src_offset + src_index];
+    }
   }
-  __syncthreads();
+  __syncwarp();
```
nishikawa-masaki, line 286: Self-comment:
__syncthreads() is more expensive than __syncwarp() because more threads are involved.

The previous implementation needed __syncthreads() because it worked in "thread block context".
This version of generateFeatures_kernel() works in "warp context" and requires only __syncwarp().

perception/autoware_lidar_centerpoint/lib/preprocess/preprocess_kernel.cu
```diff
@@ -220,128 +233,161 @@ __global__ void generateFeatures_kernel(
...
+  for (int offset = 16; offset > 0; offset /= 2) {
+    x_val += __shfl_down_sync(ALL_THREADS_IN_WARP_MASK, x_val, offset);
+    y_val += __shfl_down_sync(ALL_THREADS_IN_WARP_MASK, y_val, offset);
+    z_val += __shfl_down_sync(ALL_THREADS_IN_WARP_MASK, z_val, offset);
   }
-  __syncthreads();
+  // 3. Broadcast warp reduction results from lane 0.
+  const float sum_x = __shfl_sync(ALL_THREADS_IN_WARP_MASK, x_val, 0);
+  const float sum_y = __shfl_sync(ALL_THREADS_IN_WARP_MASK, y_val, 0);
+  const float sum_z = __shfl_sync(ALL_THREADS_IN_WARP_MASK, z_val, 0);
```
nishikawa-masaki, line 305: Self-comment:
atomicAdd() is quite expensive and it requires __syncthreads().
__shfl_down_sync() and __shfl_sync() do the same work faster.

perception/autoware_lidar_centerpoint/lib/preprocess/preprocess_kernel.cu
```diff
@@ -220,128 +233,161 @@ __global__ void generateFeatures_kernel(
...
+        pillarOutSM[pillar_idx_inBlock][9][point_idx] = center.z;
       }
     }
-
   } else {
-    pillarOutSM[pillar_idx_inBlock][point_idx][0] = 0;
-    pillarOutSM[pillar_idx_inBlock][point_idx][1] = 0;
-    pillarOutSM[pillar_idx_inBlock][point_idx][2] = 0;
-    pillarOutSM[pillar_idx_inBlock][point_idx][3] = 0;
+    pillarOutSM[pillar_idx_inBlock][0][point_idx] = 0.0f;
```
nishikawa-masaki, line 360: Self-comment:
Explicit "0.0f" (obviously float) is bettern than "0" (integer to be cast to float).

"0.0" is treated as a "double" and it may cause compiler warning for "lost precision".

perception/autoware_lidar_centerpoint/lib/preprocess/preprocess_kernel.cu
```diff
@@ -220,128 +233,161 @@ __global__ void generateFeatures_kernel(
...
-  __syncthreads();
-
-  for (int i = 0; i < ENCODER_IN_FEATURE_SIZE; i++) {
-    int outputSMId = pillar_idx_inBlock * MAX_POINT_IN_VOXEL_SIZE * ENCODER_IN_FEATURE_SIZE +
-                     i * MAX_POINT_IN_VOXEL_SIZE + point_idx;
-    int outputId = pillar_idx * MAX_POINT_IN_VOXEL_SIZE * ENCODER_IN_FEATURE_SIZE +
-                   i * MAX_POINT_IN_VOXEL_SIZE + point_idx;
-    features[outputId] = ((float *)pillarOutSM)[outputSMId];
+  // Copy data from shared memory to global memory with transposition for coalesced access.
+  {
```
nishikawa-masaki, line 382: Self-comment:
This brace pair is to show the life-spans of the variables explicitly to the compiler so that it can allocate registers efficiently.

It might be better to implement this block as a separated \_\_device\_\_ function, but there are pros and cons.

perception/autoware_lidar_centerpoint/lib/preprocess/preprocess_kernel.cu
```diff
@@ -220,128 +233,161 @@ __global__ void generateFeatures_kernel(
...
 
   // point dimemension is 5 if feature size in encoder is 11, otherwise 4
   constexpr int point_dim =
     (ENCODER_IN_FEATURE_SIZE >= ENCODER_NUM_FEATURES_11) ? POINT_DIM_XYZIT : POINT_DIM_XYZT;
 
   // load src
-  __shared__ float pillarSM[WARPS_PER_BLOCK][MAX_POINT_IN_VOXEL_SIZE][point_dim];
-  __shared__ float3 pillarSumSM[WARPS_PER_BLOCK];
-  __shared__ int3 cordsSM[WARPS_PER_BLOCK];
-  __shared__ int pointsNumSM[WARPS_PER_BLOCK];
```
nishikawa-masaki, line 238: Self-comment:
This version of generateFeatures_kernel() works in "warp context".
These values do not need to be held in the shared memory any more.
In this implementation, they are replaced by temporary variables for each thread whose values are broadcast among warp members with __shfl_sync().

KSeangTan:
Thanks for the PR, and I know it's still in draft, it would be nice to attach some reports/numbers if the performance really improves with the changes when the PR is ready to be reviewed.

nishikawa-masaki:
Some changes in this draft PR does not seem to contribute to performance improvement while the code lines become a bit tricky.
They will need further investigation for the reasons why they are not effective (coalesced access to global memory are expected to improve the performance).

Changes with clear improvement are extracted and moved to PR #12419.

This draft PR is closed.

PR #12408, opened 2026-04-03
feat(autoware_universe_designs): autware system designer planning and control nodes

technolojin:
## Description

add node descriptions of planning and control nodes 

for modules and system, see https://github.com/autowarefoundation/autoware_launch/pull/1810

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

tested by autoware_sample_designs https://github.com/autowarefoundation/autoware_launch/pull/1810, build and launch planning simulation mode

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


maxime-clem (approved):
Tested and it works well 👍 The only difference I could notice is that Autoware stops at traffic lights if no signal is published, whereas with the default launch system it does not.

PR #12412, opened 2026-04-06
refactor(traffic_light_map_based_detector): extract core logic

interimadd:
## Description

Extract core detection logic from `MapBasedDetector` ROS node into a standalone `TrafficLightMapBasedDetector` class to improve testability and separation of concerns.

The extraction was done as a pure move of logic — the detection behavior (transform selection, ROI computation, visibility filtering) is unchanged. The node now acts as a thin wrapper responsible only for ROS I/O (subscriptions, TF lookup, publishing).

### Key changes

- **New `TrafficLightMapBasedDetector` class** (`traffic_light_map_based_detector.hpp/cpp`): Contains all map processing, route processing, and detection logic with no `rclcpp` dependency.
  - `setMap()` — extracts traffic lights from lanelet map
  - `setRoute()` — filters traffic lights by route, returns `SetRouteResult` (Error as Value pattern)
  - `detect()` — computes rough/expect ROIs and visualization markers
  - `getVisibleTrafficLights()`, `getTrafficLightRoi()`, `createTrafficLightMarkers()` — moved from node as-is
- **Slimmed `MapBasedDetector` node** (`traffic_light_map_based_detector_node.hpp/cpp`): Delegates to core class. Retains only TF lookup, parameter loading, and pub/sub wiring.
- **`TransformSamplingConfig` struct**: Groups `min_timestamp_offset`, `max_timestamp_offset`, and `timestamp_sample_len` to clarify their purpose as TF sampling parameters.

## Related links

None

## How was this PR tested?

- `colcon build --packages-select autoware_traffic_light_map_based_detector` — build succeeds
- `colcon test --packages-select autoware_traffic_light_map_based_detector` — all 6 tests pass (including existing integration test and linters)

## Notes for reviewers

This PR intentionally keeps logic changes to a minimum — the core class methods are nearly identical to the original node methods, with only variable name expansions and Error as Value conversion as differences. This makes it possible to verify correctness by side-by-side comparison with the original code. Simplification and cleanup of the extracted class will be done in follow-up PRs.

## Interface changes

None.

## Effects on system behavior

None. The detection logic, topic interfaces, and parameter interfaces are all preserved.

### DIFF
--- a/perception/autoware_traffic_light_map_based_detector/src/traffic_light_map_based_detector_node.cpp
+++ b/perception/autoware_traffic_light_map_based_detector/src/traffic_light_map_based_detector_node.cpp
@@ -33,24 +33,32 @@ namespace autoware::traffic_light
 MapBasedDetector::MapBasedDetector(const rclcpp::NodeOptions & node_options)
 : Node("traffic_light_map_based_detector", node_options),
   tf_buffer_(this->get_clock()),
-  tf_listener_(tf_buffer_),
-  detector_(
-    TrafficLightMapBasedDetectorConfig{
-      this->declare_parameter<double>("max_vibration_pitch"),
-      this->declare_parameter<double>("max_vibration_yaw"),
-      this->declare_parameter<double>("max_vibration_height"),
-      this->declare_parameter<double>("max_vibration_width"),
-      this->declare_parameter<double>("max_vibration_depth"),
-      this->declare_parameter<double>("max_detection_range"),
-      this->declare_parameter<double>("car_traffic_light_max_angle_range"),
-      this->declare_parameter<double>("pedestrian_traffic_light_max_angle_range")}),
-  transform_sampling_config_{
-    this->declare_parameter<double>("min_timestamp_offset"),
-    this->declare_parameter<double>("max_timestamp_offset"),
-    this->declare_parameter<double>("timestamp_sample_len")}
+  tf_listener_(tf_buffer_)
 {
   using std::placeholders::_1;
 
+  // detector config
+  TrafficLightMapBasedDetectorConfig config{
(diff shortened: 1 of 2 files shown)


a-maumau:
It's not directly about this PR, but I have one consideration about logging.
When we try to separate the logic and ROS2 part, we may face creating a boilerplate code for logging every time in each package.
We might need to consider about this part in somewhere.

interimadd:
@a-maumau 
Thank you for review and comment!

I agree that the current design of the return value for `setRoute()` is not ideal. In this PR, I prioritized not changing the logic, which resulted in this implementation, but I believe it needs improvement.

My fundamental stance is that logging is the responsibility of the Node, not the logic. The logic should return errors as values, and the design of those error types should be left to the logic itself.
I believe we should focus on designing functions so they don't fail in the first place, which reduces the need to return a "success" status.

In this specific case, the reason `setRoute()` has to return a warning log is due to a design that requires `setMap()` to be executed before `setRoute()`. The problem lies in this implicit calling order, which is currently being addressed with logs. If the design were such that the calling order didn't matter, no special case handling would be necessary.
Specifically, if the map is set within the constructor of `TrafficLightMapBasedDetector`, first-time users won't have to worry about how to handle the `setRoute()` process. If the Node receives a Route before the `TrafficLightMapBasedDetector` instance is created, the Node can simply handle the logging if necessary.

I would like to proceed with this Node by improving the class design to move away from having the logic output logs itself.

interimadd:
Sorry I lost validation for `max_detection_range`. fixed it in https://github.com/autowarefoundation/autoware_universe/pull/12412/commits/389cb7b76dd357cfbe26c896fb403a534e0b6a4f

PR #12419, opened 2026-04-07
perf(lidar_centerpoint): remove unnecessary atomic functions from lidar_centerpoint

nishikawa-masaki:
## Description

This PR improves performance of pre-processing for autoware_lidar_centerpoint by removing unnecessary atomic read/modify/write functions.

Atomic read/modify/write to GPU global memory is quite expensive because it stops all threads to guarantee atomic operation.

## Related links

**Parent Issue:**

- None

**Private Links:**

- [TIER IV internal link](https://tier4.atlassian.net/browse/T4DEV-51246)

## How was this PR tested?
- [x] Build passed
- [x] Existing unit tests passed
  -  Note: The unit tests that require GPU are disabled and obsolete
- [x] LiDAR data are processed correctly
  - Objects can be identified in the same way as before

## Notes for reviewers

The following table shows that the performance of the fixed functions is improved.

generateVoxels_random_kernel()
|                    | Ave.          | Med.         | Min.          | Max.        | Std.        |
|----------------:|--------------:|--------------:|--------------:|--------------:|------------:|
| before (ns)  | 3,310,904 | 3,049,872 | 2,185,184 | 6,028,544 | 835,129 |
|    after (ns)  | 1,213,708 |    956,144 |    858,432 | 3,909,888 | 480,662 |

generateBaseFeatures_kernel()
|                    | Ave.       | Med       | Min.       | Max.      | Std.      |
|----------------:|------------:|-----------:|------------:|-----------:|----------:|
| before (ns)  | 179,357 | 176,480 | 163,456 | 466,368 | 22,593 |
|    after (ns)  | 176,343 | 174,080 | 162,432 | 204,096 |   7,955 |

Test conditions:
Platform: ANVIL (NVIDIA Jetson AGX Orin)
Data playback by rosbag with x1.0 speed
The values are extracted from the result of "nsys stats" after "nsys profile".

## Interface changes

None.

## Effects on system behavior

None.

The performance of pre-processing for autoware_lidar_centerpoint is improved.


perception/autoware_lidar_centerpoint/lib/preprocess/preprocess_kernel.cu
```diff
@@ -149,17 +149,17 @@ __global__ void generateVoxels_random_kernel(
 
   float * address =
     voxels + (voxel_index * MAX_POINT_IN_VOXEL_SIZE + point_id) * POINT_NUM_FEATURES;
-  atomicExch(address, x);
-  atomicExch(address + 1, y);
-  atomicExch(address + 2, z);
+  address[0] = x;
+  address[1] = y;
+  address[2] = z;
```
nishikawa-masaki, line 154: Self-comment:
As the voxel is exclusively reserved by the previous atomic function, there is no need to use costly "atomicExch()".

perception/autoware_lidar_centerpoint/lib/preprocess/preprocess_kernel.cu
```diff
@@ -206,9 +206,6 @@ __global__ void generateBaseFeatures_kernel(
       voxel_features[outIndex * 5 + 4] = voxels[inIndex * 5 + 4];
     }
   }
-
-  // clear buffer for next infer
-  atomicExch(mask + voxel_index, 0);
```
nishikawa-masaki, line 211: Self-comment:
There is no need to clear the buffer with costly atomicExch() here.

The caller (CenterPointTRT::preprocess() in centerpoint_trt.cpp) should clear the buffer and it does.
In addition to that, if this function returns earlier, some parts of the buffer is left uncleared.

nishikawa-masaki (commented):
I left a few self-comments for description of the changes.

KSeangTan:
Thanks for the work! @nishikawa-masaki 
Did you run any rosbags to test if it generates bounding boxes as expected?

nishikawa-masaki:
@KSeangTan 
> Did you run any rosbags to test if it generates bounding boxes as expected?

If there is a defined testing procedure for centerpoint, please let me know.
I'm not sure if my current testing is thorough enough.

My current testing method is:
* I have an experiment environment which consists of 2 boxes of ANVIL (Jetson AGX Orin) .
* I playback rosbag data on another PC with speed 1.0 to feed the ANVILs and observe the behavior of Autoware with Rviz running on the same PC.
* The bounding boxes and the recognition results displayed on Rviz **looks similar** between the two versions of autoware_lidar_centerpoint (with/without the changes in this PR).

I understand that this is not a very strict and deterministic way of testing.
If there is more precise testing procedure, I'll conduct a test with it.

KSeangTan:
Hi @nishikawa-masaki 

That sounds good to me.

Do you guys have any extra procedure to test the code? I reviewed it quickly and it looks good to me overall. Thanks 
@technolojin @amadeuszsz @YoshiRi @ktro2828

perception/autoware_lidar_centerpoint/lib/preprocess/preprocess_kernel.cu
```diff
@@ -149,17 +149,17 @@ __global__ void generateVoxels_random_kernel(
...
-  atomicExch(address, x);
-  atomicExch(address + 1, y);
-  atomicExch(address + 2, z);
+  address[0] = x;
+  address[1] = y;
+  address[2] = z;
   if (POINT_NUM_FEATURES == POINT_DIM_XYZT) {
     const float t = points[point_idx * POINT_NUM_FEATURES + 3];
-    atomicExch(address + 3, t);  // Time_lag
+    address[3] = t;  // Time_lag
```
KSeangTan, line 157: Do you think we should define these numbers as enum somewhere?

KSeangTan (approved):
LGTM overall, thanks for the fix.

amadeuszsz (approved):
LGTM

ktro2828 (approved):
Thank you, LGTM!

perception/autoware_lidar_centerpoint/lib/preprocess/preprocess_kernel.cu
nishikawa-masaki: @KSeangTan 
I think that replacing "magic numbers" with meaningful enums will improve readability.
For this case, I think "3" should be named as "TIME_LAG" or something similar.

However, there are some "cons":
* The actual value of TIME_LAG depends on POINT_NUM_FEATURES.
* Enums hide sequential order and magnitude of values.
  * If you see "0, 1, 2, 5, 4", you can easily find that "3 is missing, 5 and 4 are swapped".

I can't judge whether it is better or not at this point and I think it should be addressed in another PR.

PR #12445, opened 2026-04-09
feat(trajectory_validator): add evaluation tables to propagate filtering results

zulfaqar-azmi-t4:
## Description

This PR introduces structured, per-trajectory evaluation results in `trajectory_validator` and exposes them to downstream consumers. Previously the validator reduced each plugin's verdict to a boolean and only logged a free-form error string via diagnostics, so there was no way for downstream components or operators reading rosbags to see which metric from which plugin caused a trajectory to be rejected or what the underlying value was.

This PR:
1. Adds three new ROS 2 messages (`MetricReport`, `ValidationReport`, `ValidationReportArray`) that carry the validator name, validator category, metric name, measured value, and severity level (`OK` / `WARN` / `ERROR`).
2. Updates the plugin contract so each `is_feasible` call returns a `ValidationResult { is_feasible, metrics[] }`, letting plugins emit one or more `MetricReport`s per trajectory.
3. Rewrites the built-in filters (`CollisionFilter`, `OutOfLaneFilter`, `VehicleConstraintFilter`, `TrafficLightFilter`) to produce metric reports alongside their feasibility decision.
4. Introduces an internal `EvaluationTable` per trajectory that aggregates per-plugin results bucketed by plugin **category**, and publishes a `ValidationReportArray` on `~/debug/validation_reports` every cycle.
5. Generalizes and simplifies the diagnostic level logic.

#### New messages (`msg/`)
- **`MetricReport`** — `validator_name`, `validator_category`, `metric_name`, `metric_value`, `level` (`OK` / `WARN` / `ERROR`).
- **`ValidationReport`** — `trajectory_stamp`, `generator_id`, `generator_name`, `level`, `metrics[]`.
- **`ValidationReportArray`** — `reports[]`, one entry per candidate trajectory per cycle.

#### Plugin interface (`validator_interface.hpp`)
- New `ValidationResult { bool is_feasible; std::vector<MetricReport> metrics; }`.
- New alias `result_t = tl::expected<ValidationResult, std::string>`.
- `virtual is_feasible(...)` return type changed from `tl::expected<void, std::string>` to `result_t`. Plugins now return a feasibility flag **and** the metrics that led to it.
- Added `set_category()` / `category()` and protected `category_` so evaluation results can be grouped per plugin category (e.g. `safety`, `traffic_rule`).

#### Node (`trajectory_validator_node.{hpp,cpp}`)
- Added `PluginEvaluation { plugin_name, is_feasible, is_shadow_mode, reason }` and `EvaluationTable { generator_id, evaluations }` with two predicates: `all_feasible()` (every plugin passed) and `all_acceptable()` (every plugin either passed or is in shadow mode).
- `process()` builds one `EvaluationTable` per candidate trajectory; a trajectory is kept iff `all_acceptable()` is true. The full set of tables is kept in `evaluation_tables_` for the cycle.
- A `ValidationReport` is assembled per trajectory and the full `ValidationReportArray` is published on `~/debug/validation_reports` via `autoware_utils_debug::DebugPublisher`.
- `update_diagnostic` was simplified: `num_feasible == input_size` → `OK`; `num_feasible == 0` → `ERROR` ("No feasible trajectories found"); otherwise → `WARN` ("At least one trajectory is infeasible"). The previous Diffusion-specific WARN branch and its `has_trajectory_from_generator()` helper were removed.

#### Built-in filters
- **`CollisionFilter`** — no longer early-exits on first collision; runs over the full trajectory and reports a single `check_collision` metric.
- **`OutOfLaneFilter`** — reports `check_out_of_lane`.
- **`VehicleConstraintFilter`** — runs **all** checks per trajectory (previously short-circuited) so every trajectory yields five metrics (`check_speed`, `check_acceleration`, `check_deceleration`, `check_steering_angle`, `check_steering_rate`) with the maximum observed value in `metric_value`. Acceleration/deceleration are now taken from `TrajectoryPoint::acceleration_mps2` directly instead of a finite difference across adjacent points. Steering angle and rate are computed on a 5-point moving-average smoothed series (`to_steering_angles(..., smoothing_window_size=5)`). `is_*_ok` helpers return `std::pair<double, bool>` (max observed, pass/fail) rather than `bool`.
- **`TrafficLightFilter`** — reports `check_crossing_red_light` and `check_crossing_amber_light`.

#### Build (`CMakeLists.txt`, `package.xml`)
- Library target renamed from `${PROJECT_NAME}` to `${PROJECT_NAME}_lib` to avoid a name clash with the rosidl interface target.
- Added `rosidl_generate_interfaces` for the three new messages with `builtin_interfaces` and `unique_identifier_msgs` as dependencies.
- Added `rosidl_default_generators` (build), `rosidl_default_runtime` (exec), and `<member_of_group>rosidl_interface_packages</member_of_group>`.
- `unset(PYTHON_EXECUTABLE CACHE)` before `rosidl_generate_interfaces` as a workaround for stale CMake cache of the Python interpreter.

#### Default configuration (`config/trajectory_validator.param.yaml`)
- `filter_names` is now `[""]` (no enforcing filters by default).
- `safety::OutOfLaneFilter`, `safety::CollisionFilter`, `safety::VehicleConstraintFilter`, and `traffic_rule::TrafficLightFilter` are moved to `shadow_mode_filter_names`.
- Net effect: with default params the validator **observes and reports** on every plugin but does not reject any trajectory. Operators must explicitly promote a filter to `filter_names` to have it gate output.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Notes for reviewers

- Suggested review order: (1) new `msg/` files, (2) changes in `validator_interface.hpp`, (3) one filter end-to-end to see the refactor pattern (e.g. `VehicleConstraintFilter`), (4) node processing loop, (5) `CMakeLists.txt` and `package.xml`.
- `VehicleConstraintFilter` now always runs every sub-check instead of short-circuiting. This is intentional so that `MetricReport`s are emitted for every metric regardless of overall feasibility. Worth confirming this is acceptable for per-cycle cost on target hardware.
- Acceleration/deceleration now read `acceleration_mps2` directly and steering uses a 5-point smoother — same input trajectory may produce different accept/reject results compared to `main`. Reviewers who rely on specific planner/validator combinations should re-validate on a representative bag.
- Diagnostic level logic is now generator-agnostic. The previous Diffusion-specific WARN ("All diffusion planner trajectories are infeasible") no longer fires. Any downstream monitor that matched on that string needs updating.
- `unset(PYTHON_EXECUTABLE CACHE)` in `CMakeLists.txt` is a workaround for stale CMake state during `rosidl_generate_interfaces`; open to a cleaner alternative.
- `publish_validation_reports` uses `uuid_to_name.at(hex_generator_id)`; this throws if a trajectory's generator UUID is missing from `generator_info`. Happy to switch to `find()` with a fallback if reviewers prefer a defensive approach.

## Interface changes

- **Plugin ABI (`plugin::ValidatorInterface`)**: `is_feasible` return type changed from `tl::expected<void, std::string>` to `tl::expected<ValidationResult, std::string>` (aliased as `result_t`). **Out-of-tree plugins must be updated.** New public methods `set_category` / `category` and new protected member `category_`.
- **New published topic**: `~/debug/validation_reports` (`autoware_trajectory_validator/msg/ValidationReportArray`).
- **Package interface**: `autoware_trajectory_validator` is now an `rosidl_interface_packages` member and exports three new messages.

#### Additions and removals

| Change type | Topic Type | Topic Name                  | Message Type                                                  | Description                                                                                       |
|:------------|:-----------|:----------------------------|:--------------------------------------------------------------|:--------------------------------------------------------------------------------------------------|
| Added       | Pub        | `~/debug/validation_reports` | `autoware_trajectory_validator/msg/ValidationReportArray`    | Per-cycle array of `ValidationReport`, one per candidate trajectory, carrying per-plugin metrics. |

#### Modifications

| Version | Parameter Name             | Type           | Default Value                                                                                                   | Description                                                                                                                  |
|:--------|:---------------------------|:---------------|:----------------------------------------------------------------------------------------------------------------|:-----------------------------------------------------------------------------------------------------------------------------|
| Old     | `filter_names`             | `string_array` | `["safety::OutOfLaneFilter", "safety::CollisionFilter", "safety::VehicleConstraintFilter", "traffic_rule::TrafficLightFilter"]` | Filters that gate output (an infeasible verdict rejects the trajectory).                                                     |
| New     | `filter_names`             | `string_array` | `[""]`                                                                                                          | Filters that gate output. Default is now empty; operators must explicitly list filters here to restore the previous gating. |
| Old     | `shadow_mode_filter_names` | `string_array` | `[]`                                                                                                            | Filters run in shadow mode — results are reported but do not gate output.                                                   |
| New     | `shadow_mode_filter_names` | `string_array` | `["safety::OutOfLaneFilter", "safety::CollisionFilter", "safety::VehicleConstraintFilter", "traffic_rule::TrafficLightFilter"]` | Filters run in shadow mode. The four previous gating filters are now shadow-mode by default.                                 |

## Effects on system behavior

- **Default deployment no longer rejects trajectories.** All previously enforcing filters were moved to `shadow_mode_filter_names`; operators must promote filters back to `filter_names` to retain the previous gating behavior.
- `VehicleConstraintFilter`, `CollisionFilter`, and `TrafficLightFilter` now run their full set of checks per trajectory rather than short-circuiting. Expect a modest increase in per-cycle CPU cost and one metric entry per sub-check in the published reports.
- `VehicleConstraintFilter` acceleration/deceleration and steering semantics changed (see above); may produce different accept/reject decisions on the same trajectory data.
- Diagnostic WARN messages are generic now; the Diffusion-specific WARN is removed.
- New high-frequency publisher on `~/debug/validation_reports` — bandwidth scales with `num_plugins × num_candidate_trajectories`.


planning/autoware_trajectory_validator/src/filters/traffic_rule/traffic_light_filter.cpp
```diff
@@ -124,7 +125,7 @@ TrafficLightFilter::get_stop_lines(
   return {red_stop_lines, amber_stop_lines};
 }
 
-tl::expected<void, std::string> TrafficLightFilter::is_feasible(
+TrafficLightFilter::result_t TrafficLightFilter::is_feasible(
```
maxime-clem, line 141: This function looks wrong, with some "unfeasible" branches returning a `tl::make_unexpected(...)` instead of a `ValidationResult`.

zulfaqar-azmi-t4: Yeah, it is difficult because there difference between main branch and e2e.

There are two options:
1. I can just copy as paste the code as is.
2. Merge the traffic light changes first and rebase.

zulfaqar-azmi-t4: I have manually copy-pasted the changes, but in case  you would like merge the changes first, then i will revert the commit https://github.com/autowarefoundation/autoware_universe/pull/12445/commits/855e89b3fa6fe5e57b695a65348b65d4d35db3dd.

maxime-clem (approved):

ktro2828 (approved):
@zulfaqar-azmi-t4 LGTM. I checked messages, interface/node, and vehicle constraint filer.

PR #12467, opened 2026-04-10
refactor(autoware_trajectory_validator): consolidate test targets

yhisaki:
## Description

Before this PR, `autoware_trajectory_validator` built its tests as multiple executables and kept redundant setup code in the traffic light filter test.

This PR refactors the trajectory validator test configuration.
- Consolidates the trajectory validator tests into a single test executable
- Replaces hard-coded target names in `CMakeLists.txt` with `${PROJECT_NAME}`-based names
- Simplifies `test_traffic_light_filter.cpp` by removing unused node setup, redundant `main`, and an unused helper header

## Related links
**Parent Issue:**
- None.

## Notes for reviewers
The current diff also switches the aggregated test target to `ament_auto_add_gtest(...)`, so reviewers should confirm that the required link dependencies for the node-oriented test remain satisfied in CI.

#### Additions and removals
| None | None | None | None | None |

#### Modifications
| None | None | None | None | None |

#### Additions and removals
| None | None | None | None | None |

#### Modifications
| None | None | None | None | None |

## Effects on system behavior

None. This change only affects test code and test build configuration.


planning/autoware_trajectory_validator/CMakeLists.txt
```diff
@@ -71,29 +52,15 @@ if(BUILD_TESTING)
...
-
-  ament_add_gtest(test_traffic_light_filter
     test/plugins/test_traffic_light_filter.cpp
-  )
-  target_link_libraries(test_traffic_light_filter
-    autoware_trajectory_validator_plugins
+    test/plugins/test_vehicle_constraint_filter.cpp
   )
 
+  ament_auto_add_gtest(test_${PROJECT_NAME}
```
maxime-clem, line 63: Since the tests use ros2 communication they need to be added with `ament_add_ros_isolated_gtest`.

maxime-clem (changes requested):

zulfaqar-azmi-t4 (approved):

planning/autoware_trajectory_validator/CMakeLists.txt
yhisaki: @maxime-clem 
Thanks!
I found that the only difference between ament_auto_add_gtest and ament_add_ros_isolated_gtest is just a RUNNER.

https://github.com/ros2/ament_cmake_ros/blob/1ebf5450912a971d70638a0713a2c18b649a4d1d/ament_cmake_ros/cmake/ament_add_ros_isolated_test.cmake

So, I set RUNNER manually ad601acbf1b243e92e21bd2b84576a1c240dc897.

maxime-clem: I had missed that line. However, I believe it is better to use `ament_add_ros_isolated_gtest` instead of setting the `RUNNER` manually.
The main reason is that in the future the way to run isolated tests may change and the macro may be updated, and this code may miss the update.

yhisaki: I see. At first, I wanted to use the automatic dependency resolution feature in `ament_auto_xxx`, but it can't be helped😢
If there's no better way, I'll revert it.

PR #12473, opened 2026-04-13
fix(autoware_behavior_path_side_shift): stabilize side shift path

TaikiYamada4:
## Description

This PR stabilizes the path of the side_shift module.
Currently the path of the side_shift module changes dynamically during the shift because `path_shifter_.generate()` keeps changing the path while the side_shift module is in `SHIFTING` status.

## How was this PR tested?

- Checked that `colcon test` passes
- Checked the difference between PSim 
  - Published topics with an order of `-0.5 m` -> `-1.5` -> `0.0` -> `1.0` -> `0.0` (meters)

### Current Autoware

https://github.com/user-attachments/assets/d148c01c-f339-494f-809e-5ae8cf257bd0

### After PR applied

https://github.com/user-attachments/assets/108888a8-0a7f-49d5-abc5-cf2f638f8041

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

The side_shift modules quits editing the path while shifting.

### DIFF
--- a/planning/behavior_path_planner/autoware_behavior_path_side_shift_module/src/scene.cpp
+++ b/planning/behavior_path_planner/autoware_behavior_path_side_shift_module/src/scene.cpp
@@ -275,7 +275,7 @@ BehaviorModuleOutput SideShiftModule::plan()
     RCLCPP_DEBUG(getLogger(), "change is not requested");
   }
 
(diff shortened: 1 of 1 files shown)


kyoichi-sugahara (approved):
LGTM!

PR #12488, opened 2026-04-17
refactor(traffic_light_map_based_detector): unify detect() transform args into StampedTransform vector

interimadd:
## Description

The `detect()` method previously took two separate transform arguments:
- `std::vector<tf2::Transform> tf_map2camera_vec` (sampled transforms over a time window)
- `tf2::Transform tf_map2camera` (exact camera timestamp transform)

This made it unclear what each argument represented and how they related to each other.

This PR introduces a `StampedTransform` struct that pairs `builtin_interfaces::msg::Time` with `tf2::Transform`, and replaces the two arguments with a single `std::vector<StampedTransform>`. The `detect()` method now internally selects the transform closest to the camera timestamp for expect ROI calculation using `findClosestTransform()`.

Additionally:
- The node now always includes the exact camera timestamp transform in the samples, ensuring `detect()` always has the best match available.
- An early return guard was added for the case when the samples vector is empty, preventing undefined behavior.
- Debug markers now use the closest-to-camera-time transform instead of the first sample in the time window.

## Related links

None

## How was this PR tested?

- `colcon test --packages-select autoware_traffic_light_map_based_detector` passes all 6 tests (unit tests, integration test, linters).
- Evaluator[(internal link)](https://evaluation.tier4.jp/evaluation/reports/4576c902-08c0-5e54-925f-734e6d10b6cb/?project_id=autoware_dev)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

- The exact camera timestamp transform is now always included in the sampled transforms passed to `detect()`. Previously it was only added as a fallback when the time-window sampling returned no results. This means visibility filtering and rough ROI computation now consider one additional sample, but the practical impact is negligible since the exact timestamp falls within the sampling range.
- Debug visualization markers now use the transform closest to the camera timestamp instead of the first transform in the sampling window, making the markers more accurate.


perception/autoware_traffic_light_map_based_detector/src/traffic_light_map_based_detector.hpp
```diff
@@ -96,13 +103,13 @@ class TrafficLightMapBasedDetector
    * @brief Filter traffic lights that are visible from the camera
    *
    * @param all_traffic_lights      all the traffic lights in the route or in the map
-   * @param tf_map2camera_vec       the transformation sequences from map to camera
+   * @param tf_map2camera_samples    the stamped transformation samples from map to camera
```
a-maumau: I noticed it, so I will mention it.

```suggestion
   * @param tf_map2camera_samples   the stamped transformation samples from map to camera
```

a-maumau (approved):
LGTM!

PR #12498, opened 2026-04-21
fix(dummy_diag_publisher): support dummy diags with at most one period when `default` is specified

riki-ukyo:
## Description
The dummy diag publisher was designed to publish a dummy diagnostic by truncating the diagnostic name at the first period (`.`) if one existed in the diagnostic name. This PR updates this behavior to allow publishing of diagnostics that contain at most one period.

#### Key Changes
- Modified the logic to support diagnostic names containing at most one period so that they can be properly published instead of being truncated.

- Added special handling for `.is_active` and `.status` suffixes: Since only these two strings can be used to indicate the state of a dummy diagnostic, if a diagnostic name ends with either of them, the part preceding the suffix is recognized as the diagnostic name.

#### Limitations
It has been confirmed that specifying `is_active` or `status` attributes for diagnostics containing a period results in the publishing of incorrect diagnostic names. Since this issue does not occur when `default` is specified, the following two constraints are applied when a diagnostic name includes a period:

- A diagnostic name must contain no more than one period.

- If a diagnostic name contains a period, only the `default` setting must be used.

#### Notes
This PR is specifically intended to enable the publishing dummy diagnostics for diagnostic names that already contain a period in their existing implementation.

Please note that using periods for sub-components is considered a violation of ROS naming conventions ([reference](https://ros.org/reps/rep-0107.html)). This practice can lead to compatibility issues with various tools and internal specifications. Therefore, it is highly recommended to use `diagnostic_updater` and define diagnostic names without periods for standard diagnostic outputs.

## Related links

**Parent Issue:**

- [TIER IV internal link](https://star4.slack.com/archives/CRUE57C30/p1775624637363469?thread_ts=1775122298.158069&cid=CRUE57C30)

## How was this PR tested?
To verify this PR, I performed the following steps:

1. Ran the planning-simulator with this PR applied.

2. Configured `dummy_diag_publisher.param.yaml` to include a diagnostic name containing a period: `"test1.test2": default`.

3. Monitored the diagnostic output using rqt_runtime_monitor.

4. Confirmed that the diagnostic `test1.test2` was output with the expected OK status message under the `default` setting. Additionally, I have verified that this change does not affect any other diagnostics.

image

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/system/autoware_dummy_diag_publisher/src/dummy_diag_publisher_core.cpp
+++ b/system/autoware_dummy_diag_publisher/src/dummy_diag_publisher_core.cpp
@@ -84,7 +84,8 @@ void DummyDiagPublisher::loadRequiredDiags()
   static const std::string is_active_suffix = ".is_active";
   static const std::string status_suffix = ".status";
   const auto ends_with = [](const std::string & s, const std::string & suffix) {
-    return s.size() > suffix.size() && s.compare(s.size() - suffix.size(), suffix.size(), suffix) == 0;
+    return s.size() > suffix.size() &&
+           s.compare(s.size() - suffix.size(), suffix.size(), suffix) == 0;
   };
 
   for (const auto & param_name : param_names) {
--- a/system/autoware_dummy_diag_publisher/src/dummy_diag_publisher_core.cpp
+++ b/system/autoware_dummy_diag_publisher/src/dummy_diag_publisher_core.cpp
@@ -26,17 +26,6 @@
 
 namespace autoware::dummy_diag_publisher
 {
-std::vector<std::string> split(const std::string & str, const char delim)
-{
-  std::vector<std::string> elems;
-  std::stringstream ss(str);
-  std::string item;
-  while (std::getline(ss, item, delim)) {
-    elems.push_back(item);
-  }
-  return elems;
-}
-
 std::optional<DummyDiagPublisher::Status> DummyDiagPublisher::convertStrToStatus(
   const std::string & status_str)
 {
@@ -91,12 +80,27 @@ void DummyDiagPublisher::loadRequiredDiags()
 
   std::set<std::string> diag_names;
 
-  for (const auto & param_name : param_names) {
-    const auto split_names = split(param_name, '.');
-    const auto & param_required_diags = split_names.at(0);
-    const auto & param_diag = split_names.at(1);
+  const std::string required_diags_prefix = param_key + std::string(".");
+  static const std::string is_active_suffix = ".is_active";
+  static const std::string status_suffix = ".status";
+  const auto ends_with = [](const std::string & s, const std::string & suf) {
+    return s.size() > suf.size() &&
+           s.compare(s.size() - suf.size(), suf.size(), suf) == 0;
+  };
 
-    const auto diag_name_with_prefix = fmt::format("{0}.{1}", param_required_diags, param_diag);
+  for (const auto & param_name : param_names) {
+    if (param_name.rfind(required_diags_prefix, 0) != 0) {
+      continue;
+    }
+
+    std::string param_diag = param_name.substr(required_diags_prefix.size());
+    if (ends_with(param_diag, is_active_suffix)) {
+      param_diag.resize(param_diag.size() - is_active_suffix.size());
+    } else if (ends_with(param_diag, status_suffix)) {
+      param_diag.resize(param_diag.size() - status_suffix.size());
+    }
+
+    const auto diag_name_with_prefix = required_diags_prefix + param_diag;
 
     if (diag_names.count(diag_name_with_prefix) != 0) {
       continue;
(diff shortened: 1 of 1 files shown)


system/autoware_dummy_diag_publisher/src/dummy_diag_publisher_core.cpp
sasakisasaki: I guess we can use already defined variable as this:

```suggestion
    const auto is_active_key = diag_name_with_prefix + is_active_suffix;
```

sasakisasaki: Same here too:

```suggestion
    const auto status_key = diag_name_with_prefix + status_suffix;
```

system/autoware_dummy_diag_publisher/src/dummy_diag_publisher_core.cpp
```diff
@@ -91,12 +80,26 @@ void DummyDiagPublisher::loadRequiredDiags()
 
   std::set<std::string> diag_names;
 
+  const std::string required_diags_prefix = param_key + std::string(".");
+  static const std::string is_active_suffix = ".is_active";
+  static const std::string status_suffix = ".status";
+  const auto ends_with = [](const std::string & s, const std::string & suf) {
+    return s.size() > suf.size() && s.compare(s.size() - suf.size(), suf.size(), suf) == 0;
+  };
```
sasakisasaki: Let's use `suffix` without abbreviation 👍 
```suggestion
  const auto ends_with = [](const std::string & s, const std::string & suffix) {
    return s.size() > suffix.size() && s.compare(s.size() - suffix.size(), suffix.size(), suffix) == 0;
  };
```

system/autoware_dummy_diag_publisher/src/dummy_diag_publisher_core.cpp
```diff
@@ -91,12 +80,26 @@ void DummyDiagPublisher::loadRequiredDiags()
...
+    if (param_name.rfind(required_diags_prefix, 0) != 0) {
+      continue;
+    }
+
+    std::string param_diag = param_name.substr(required_diags_prefix.size());
+    if (ends_with(param_diag, is_active_suffix)) {
+      param_diag.resize(param_diag.size() - is_active_suffix.size());
+    } else if (ends_with(param_diag, status_suffix)) {
+      param_diag.resize(param_diag.size() - status_suffix.size());
+    }
```
sasakisasaki, line 101: Could you add comment(s) why this operation for string is needed? I guess this is because we'll attach the suffix at [these lines](https://github.com/autowarefoundation/autoware_universe/pull/12498/changes#diff-361e6038f3d3b140e91c5934fddd914216d36b6c83e8405312dd35e932f376d7R110-R113). Thus this operation looks to prevent from doubled suffix attachment.

system/autoware_dummy_diag_publisher/src/dummy_diag_publisher_core.cpp
```diff
@@ -26,17 +26,6 @@
...
-{
-  std::vector<std::string> elems;
-  std::stringstream ss(str);
-  std::string item;
-  while (std::getline(ss, item, delim)) {
-    elems.push_back(item);
-  }
-  return elems;
-}
-
```
sasakisasaki, line 39: (Not a change request, just a note to me) This function is within its namespace. No impact scope for other packages.

sasakisasaki (commented):
@riki-ukyo Thank you so much for your contribution. The PR description provides enough information and the applied changes are almost fine. Only minor comments from my side just before approval 👍

sasakisasaki:
@isamu-takagi If you have time, could you kindly review this PR? Thank you in advance!

system/autoware_dummy_diag_publisher/src/dummy_diag_publisher_core.cpp
riki-ukyo: Thank you for the comment. As you suspected, this operation is intended to prevent from doubled suffix attachment.
If diagnostic values are set manually in `dummy_diag_publisher.param.yaml` (e.g., `{is_active: "true", status: "OK"}`), `param_diag` in [line 95](https://github.com/riki-ukyo/autoware_universe/blob/41edf31db58d30dc47448993e81da0a385367ead/system/autoware_dummy_diag_publisher/src/dummy_diag_publisher_core.cpp#L95) includes suffixes like `.is_active`. This operation ensures those suffixes aren't treated as the base diagnostic name.

system/autoware_dummy_diag_publisher/src/dummy_diag_publisher_core.cpp
riki-ukyo: ☑️

system/autoware_dummy_diag_publisher/src/dummy_diag_publisher_core.cpp
riki-ukyo: ☑️

riki-ukyo: ☑️

riki-ukyo:
As discussed in [this TIER IV internal link](https://star4.slack.com/archives/C07K9RBUQR3/p1776850396485159?thread_ts=1776389439.767639&cid=C07K9RBUQR3), I will recreate this PR in [tier4/autoware_universe](https://github.com/tier4/autoware_universe), so I’m closing this one.

PR #12504, opened 2026-04-24
feat(behavior_path_side_shift): add drivable area check to prevent lane departure

k-uta:
## Description

This PR adds a lane boundary check to the `side_shift` module to prevent the ego vehicle from departing the drivable area when a lateral offset is requested.

Previously, the module executed the requested shift amount unconditionally, which could cause unsafe lane departures. With this change, the module calculates the maximum allowable lateral offset based on the distance to the lanelet boundaries and clamps the requested value so that the vehicle remains within the safe area.

## Changes

- **Added `calcMaxLateralOffset` and `calcOffsetLimitsFromLanelets` methods**: Introduced directly into the `SideShiftModule` class to encapsulate the boundary-check logic. These methods calculate the safe shift limit based on the vehicle width, configured margin, and distance to lane boundaries. The scan range extends from the ego position to the end of the shift to detect any lane narrowing.
- **Continuous offset clamping**: Added logic in `updateData()` to continuously clamp requested offsets against the current drivable boundary. This detects narrowing conditions (e.g., when adjacent lanes disappear at intersections) and dynamically reduces the inserted lateral offset to keep the vehicle safe.
- **Added check modes**: 
  - **Disabled**: No boundary check; behaves the same as before.
  - **Current lane only**: Limits the offset to the current lane boundaries by evaluating the left and right bounds of the closest lanelet.
  - **Adjacent Lanes (`use_adjacent`)**: Extends the allowable drivable area into the most left or most right same-direction neighbor lanes.
- **Added parameters**: `drivable_area_check_mode` and `min_drivable_area_margin` have been added to the `SideShiftModuleManager` for initialization and dynamic parameter updates.

### ROS Parameters

The following parameters have been added to the module's configuration:

| Change Type | Parameter Name | Type | Description |
| :--- | :--- | :--- | :--- |
| Added | `drivable_area_check_mode` | int | `0`: Disabled / `1`: Current lane only / `2`: Current + same-direction adjacent lanes |
| Added | `min_drivable_area_margin` | double | Minimum margin [m] between the vehicle footprint edge and the lane boundary |

## Effects on System Behavior

- **Mode 0 (Disabled)**: Behaves identically to the previous implementation; the vehicle shifts by the exact requested amount.
- **Mode 1 (Current Lane)**: The vehicle shifts toward the target but is strictly clamped before crossing the boundary of the current routing lanelet.
- **Mode 2 (Adjacent Lanes)**: The vehicle can cross the lane marking into an adjacent lane (if one exists in the same direction) but stops before leaving the combined outer boundaries.

## How Was This PR Tested?

1. Launch the planning simulator.
2. Set `drivable_area_check_mode` to `1` or `2`.
3. Publish a large lateral offset (e.g., −5.0 m) that exceeds the lane width and verify the vehicle stays within the boundary and dynamically adjusts if the lane narrows.

**Command:**
```bash
sec=$(($(date +%s))); ros2 topic pub /planning/scenario_planning/lane_driving/behavior_planning/behavior_path_planner/input/lateral_offset tier4_planning_msgs/msg/LateralOffset "{stamp: {sec: $sec, nanosec: 999999999}, lateral_offset: -5.0}" -1
```


planning/behavior_path_planner/autoware_behavior_path_side_shift_module/src/scene.cpp
```diff
@@ -265,11 +288,9 @@ void SideShiftModule::replaceShiftLine()
 BehaviorModuleOutput SideShiftModule::plan()
 {
   // Replace shift line
-  if (
-    lateral_offset_change_request_ && ((shift_status_ == SideShiftStatus::BEFORE_SHIFT) ||
-                                       (shift_status_ == SideShiftStatus::AFTER_SHIFT))) {
+  if (lateral_offset_change_request_) {
```
TaikiYamada4: This change is not equivalent. Please revert it to the original condition, since it is necessary to prevent chattering.

planning/behavior_path_planner/autoware_behavior_path_side_shift_module/src/scene.cpp
```diff
@@ -469,6 +494,96 @@ PathWithLaneId SideShiftModule::extendBackwardLength(const PathWithLaneId & orig
...
+std::pair<double, double> SideShiftModule::calcOffsetLimitsFromLanelets(
+  const double requested_offset) const
+{
+  if (!planner_data_ || !planner_data_->route_handler || reference_path_.points.empty()) {
+    return {0.0, 0.0};
+  }
+
+  const auto & route_handler = planner_data_->route_handler;
+  const auto & p = planner_data_->parameters;
+  const double vehicle_half_width = p.vehicle_width / 2.0;
```
TaikiYamada4: If `p` is used only once, maybe we should merge these two lines.

TaikiYamada4 (commented):
@k-uta 

Thank you for your contributions!
I found a case that the following doesn't work. Could your fix this problem?

1. Set a large lateral offset when check mode is `CURRENT_LANE=1`. The path will be shifted to the possible end.
2. Set that same large lateral offset when the road got wider than the previous path at the middle of route.

Plus, I added two nits comments too.

planning/behavior_path_planner/autoware_behavior_path_side_shift_module/config/side_shift.param.yaml
```diff
@@ -7,4 +7,6 @@
       min_shifting_distance: 5.0
       min_shifting_speed: 5.56
       shift_request_time_limit: 1.0
+      drivable_area_check_mode: 1
```
TaikiYamada4:
```suggestion
      drivable_area_check_mode: 1 # 0 = No check, 1 = Only current lane, 2 = Adjacent lanes included
```

Could you add a comment like this?

k-uta: Thank you for the suggestion. I've added the comment to explain the modes as you suggested.

Fixed in [037206d](https://github.com/autowarefoundation/autoware_universe/pull/12504/commits/037206dde6b635a4534edc067f0f8854a267d746)

planning/behavior_path_planner/autoware_behavior_path_side_shift_module/src/scene.cpp
k-uta: I have reverted it to the original logic.

Fixed in [7af6cdf](https://github.com/autowarefoundation/autoware_universe/pull/12504/commits/7af6cdf69b5554c5baf34c19fae5553189ab91bc)

planning/behavior_path_planner/autoware_behavior_path_side_shift_module/src/scene.cpp
k-uta: Since `p` is indeed only used once, I have merged them into a single line.

Fixed in [84bf375](https://github.com/autowarefoundation/autoware_universe/pull/12504/commits/84bf3756a967aeebf45eabf376aebdb01902358c)

TaikiYamada4 (approved):
LGTM

technolojin:
@TaikiYamada4 @k-uta 
You shall not merge before https://github.com/autowarefoundation/autoware_launch/pull/1828 is approved. User will fail the system due to the missing parameters.

TaikiYamada4:
@technolojin I'm so sorry for overlooking the launch PR. Merged it now.

PR #12525, opened 2026-04-30
feat(autoware_carla_interface): add light weight sensor config

mitsudome-r:
## Description

Adds an optional light-weight sensor configuration for `autoware_carla_interface` to make CARLA simulation usable on machines with limited GPU/CPU resources.

- New `use_light_weight_sensor_mapping` launch argument (default `False`). When `True`, the launch file selects [`config/sensor_mapping_light_weight.yaml`](simulator/autoware_carla_interface/config/sensor_mapping_light_weight.yaml) and skips the `image_transport` republish nodes for the disabled cameras and the `multi_camera_combiner` node.
- The light-weight mapping uses a single front camera (`CAM_FRONT`, 120° FOV, 1080x720) instead of the default 6-camera 360° setup, and lowers LiDAR/camera frequency to 10 Hz. LiDAR/IMU/GNSS configuration is otherwise unchanged.
- README updated with a new "Light-Weight Sensor Mapping" subsection and parameter table entries.
- `SensorKitLoader` now logs a warning when it falls back to the source-tree default sensor mapping path, to make the fallback visible during debugging.

## Related links

**Parent Issue:**

- https://github.com/autowarefoundation/autoware/issues/7079

## How was this PR tested?

- [ ] Launched `e2e_simulator.launch.xml` with `simulator_type:=carla` and `use_light_weight_sensor_mapping:=True`, confirmed only `CAM_FRONT` is published and `multi_camera_combiner` is not started.
- [ ] Launched the same with `use_light_weight_sensor_mapping:=False` (default) and confirmed all 6 cameras and the combiner still come up as before.
- [ ] Verified LiDAR, IMU, and GNSS topics are published in both modes.

## Notes for reviewers

The default behavior is unchanged — `use_light_weight_sensor_mapping` defaults to `False` and the existing `sensor_mapping.yaml` is selected. Reviewers who don't use the flag should see no functional difference.

## Interface changes

None.

## Effects on system behavior

None.


maxime-clem (approved):
Tested with both `use_light_weight_sensor_mapping:=False` and `use_light_weight_sensor_mapping:=True` and in both cases Autoware could correctly localize.

PR #12532, opened 2026-05-01
feat(trajectory_selector): combine validator and concatenator

zulfaqar-azmi-t4:
## Description
Merges the previously separate `autoware_trajectory_concatenator` node and `autoware_trajectory_validator` library into a single `TrajectorySelectorNode`. The node collects candidate trajectories from multiple upstream planners, concatenates them into a single set, runs the configured validator plugins, and publishes the surviving trajectories.
Clean public wrappers are introduced for both subsystems (`TrajectoryConcatenatorWrapper`, `TrajectoryValidatorWrapper`) so each remains independently readable and testable. The old `ValidationStage` class is removed. Internal implementation types (`TrajectoryValidator`, `ValidatorContext`, `TrajectoryValidatorReport`) are moved under `detail/`.
## Related links
**Parent Issue:**
- Link
## How was this PR tested?
- `colcon build` — success
- `colcon test` — all tests pass, including:
  - Existing plugin unit tests (`test_vehicle_constraint_filter`, `test_traffic_light_filter`)
  - Node integration tests (`FiltersTrajectoriesViaPlugin`, `HandlesPluginRejection`)
  - New tests for missing-input early returns in `take_validator_data()` (missing odometry, acceleration, predicted objects)
## Notes for reviewers
The `trajectory_validator_node` ROS node name and its associated topic namespace (`~/input/trajectories`) are replaced by the ROS node name `trajectory_selector_node` (executable: `autoware_trajectory_selector_node`) and two separate input topics (see Interface changes below). Downstream launch files that remap `~/input/trajectories` will need to be updated to remap `~/input/trajectories_generative` and/or `~/input/trajectories_backup`.
#### Additions and removals
|:---|:---|:---|:---|:---|
| Added | Sub | `~/input/trajectories_generative` | `autoware_internal_planning_msgs/msg/CandidateTrajectories` | Candidate trajectories from generative planners |
| Added | Sub | `~/input/trajectories_backup` | `autoware_internal_planning_msgs/msg/CandidateTrajectories` | Candidate trajectories from backup planners |
| Added | Pub | `~/output/trajectories` | `autoware_internal_planning_msgs/msg/CandidateTrajectories` | Trajectories that passed all enforced validator plugins |
| Added | Pub | `~/debug/processing_time_detail_ms/trajectory_selector` | `autoware_internal_debug_msgs/msg/ProcessingTimeTree` | Per-function processing time breakdown |
| Removed | Sub | `~/input/trajectories` | `autoware_internal_planning_msgs/msg/CandidateTrajectories` | Replaced by the two topics above |
#### Modifications
| Version | Topic Type | Topic Name | Description |
|:---|:---|:---|:---|
| Old | All | `/trajectory_validator_node/...` | Node was named `trajectory_validator_node` |
| New | All | `/trajectory_selector_node/...` | Node renamed to `trajectory_selector_node` |
## Effects on system behavior
The node now concatenates trajectories from multiple sources before validation, rather than validating a single pre-concatenated input. Trajectories that arrive on `trajectories_generative` or `trajectories_backup` are merged into one candidate set each timer tick before being passed through the validator plugins. Behavior of the validator plugins themselves is unchanged.


planning/autoware_trajectory_validator/CMakeLists.txt
```diff
@@ -2,6 +2,7 @@ cmake_minimum_required(VERSION 3.14)
 project(autoware_trajectory_validator)
 
 find_package(autoware_cmake REQUIRED)
+find_package(autoware_trajectory_concatenator REQUIRED)
```
maxime-clem: This should not be necessary (handled by `autoware_package()` + `package.xml`).

planning/autoware_trajectory_validator/CMakeLists.txt
```diff
@@ -17,12 +18,17 @@ include_directories(
...
+  src/validation_stage.cpp
 )
 
 target_link_libraries(${trajectory_validator_lib_target}
   ${PROJECT_NAME}_param
 )
 
+ament_target_dependencies(${trajectory_validator_lib_target}
+  autoware_trajectory_concatenator
+)
```
maxime-clem: This also should be not be necessary

planning/autoware_trajectory_validator/include/autoware/trajectory_validator/filter_context.hpp
maxime-clem, line 1: This file is only needed for backward compatibility ?

planning/autoware_trajectory_validator/src/trajectory_validator_node.cpp
```diff
@@ -83,9 +74,13 @@ void TrajectoryValidator::subscribers()
...
-  sub_trajectories_ = create_subscription<CandidateTrajectories>(
-    "~/input/trajectories", 1,
-    std::bind(&TrajectoryValidator::process, this, std::placeholders::_1));
+  sub_trajectories_generative_ = create_subscription<CandidateTrajectories>(
+    "~/input/trajectories_generative", 1,
+    std::bind(&TrajectoryValidator::on_trajectories, this, std::placeholders::_1));
+
+  sub_trajectories_backup_ = create_subscription<CandidateTrajectories>(
+    "~/input/trajectories_backup", 1,
+    std::bind(&TrajectoryValidator::on_trajectories, this, std::placeholders::_1));
```
maxime-clem: It would be better if these input topics were defined as parameters to be more flexible. It would also enable handling the special case with only one input topic where we no longer need to wait for `on_timer()`.

maxime-clem (approved):

planning/autoware_trajectory_validator/src/trajectory_validator_node.cpp
zulfaqar-azmi-t4: do you mean something like this?


```
input_trajectory = {input_topic1, input_topic2}
for(traj:input_traj)
   sub_trajectory.push_back(create_subscription(traj))
```

planning/autoware_trajectory_validator/include/autoware/trajectory_validator/filter_context.hpp
zulfaqar-azmi-t4: yes, in this PR, I would prefer not to make any changes to the plugin side code.
It allows us to cherry-pick more easily in experimental branch. Once this is merged, i will make a new PR to change the type.

planning/autoware_trajectory_validator/src/trajectory_validator_node.cpp
maxime-clem: Yes, where the input topics can be set as a string array node parameter.

maxime-clem: This is fine to keep it for a future PR

planning/autoware_trajectory_validator/CMakeLists.txt
zulfaqar-azmi-t4: done in https://github.com/autowarefoundation/autoware_universe/pull/12532/commits/ef2b8cea22b6ffb0278d4333f8bf9db299c71086

planning/autoware_trajectory_validator/CMakeLists.txt
zulfaqar-azmi-t4: done in https://github.com/autowarefoundation/autoware_universe/pull/12532/commits/ef2b8cea22b6ffb0278d4333f8bf9db299c71086

danielsanchezaran (changes requested):
I dont see how the inclusion of paralelism and mutex is worth it over the removal of a node that is quite simple (concatenator), can you please illustrate the advantages of this new layout? I dont think it is worth it, especially when it is not obvious that the validator should be the one concatenating trajectories

zulfaqar-azmi-t4:
@danielsanchezaran 

> I dont see how the inclusion of paralelism and mutex

The mutex is not new in this PR. It's already present in the original standalone concatenator node, and I ported that logic directly into the validator without modification.

> can you please illustrate the advantages of this new layout? I dont think it is worth it

There are three things to highlight

1. (Main reason) Node count reduction. This is part of the plan to reduce the number of running nodes in the planning pipeline. There is no clear requirement for concatenator to be its own node. If it is for separation of responsibility, it is satisfied just as well by keeping it as a class inside the validator. Unless there's a specific reason for it to be a node, a library called from the validator is the simpler design.
2.  The standalone concat node publishes on a fixed 100ms timer. I assume that the diffusion planner has high latency which requires a longer time buffer. The side effect, however, is that the validator's callback, which based on concat's output subscription, is also capped at ~100ms cadence. Of course we can also add timer based publisher in the validator, but that doesn't make sense if the reason is because we can't wait for concatenator's subscription. Merging the nodes makes it easier to handle this.
3. The validator node is planned to be renamed to trajectory_selector, and each existing planning-pipeline node will be converted into a library called by the selector. This PR prioritizes the node merge first; the rename and the moving validator out of the package will be done later.

danielsanchezaran:
> @danielsanchezaran
> 
> > I dont see how the inclusion of paralelism and mutex
> 
> The mutex is not new in this PR. It's already present in the original standalone concatenator node, and I ported that logic directly into the validator without modification.
> 
> > can you please illustrate the advantages of this new layout? I dont think it is worth it
> 
> There are three things to highlight
> 
> 1. (Main reason) Node count reduction. This is part of the plan to reduce the number of running nodes in the planning pipeline. There is no clear requirement for concatenator to be its own node. If it is for separation of responsibility, it is satisfied just as well by keeping it as a class inside the validator. Unless there's a specific reason for it to be a node, a library called from the validator is the simpler design.
> 2. The standalone concat node publishes on a fixed 100ms timer. I assume that the diffusion planner has high latency which requires a longer time buffer. The side effect, however, is that the validator's callback, which based on concat's output subscription, is also capped at ~100ms cadence. Of course we can also add timer based publisher in the validator, but that doesn't make sense if the reason is because we can't wait for concatenator's subscription. Merging the nodes makes it easier to handle this.
> 3. The validator node is planned to be renamed to trajectory_selector, and each existing planning-pipeline node will be converted into a library called by the selector. This PR prioritizes the node merge first; the rename and the moving validator out of the package will be done later.

The points have been addressed in a private conversation, I am ok with this PR changes given the proposed future changes. @zulfaqar-azmi-t4 please take a look at the copilot review comments too

danielsanchezaran (commented):
@zulfaqar-azmi-t4 please also update the READMEs

danielsanchezaran (approved):

maxime-clem (approved):

PR #12533, opened 2026-05-01
ci(bump-version-pr): support custom base version and target branch

mitsudome-r:
### Summary
Current bump-version-pr workflow fails when there are packages with different versions described in package.xml versions([example](https://github.com/autowarefoundation/autoware_universe/actions/runs/25214254408/job/73931085269)). This reworks the `bump-version-pr` workflow so it can bump from any specified base version, target any branch, and produces a clean single-commit PR with a more informative description.

#### New `workflow_dispatch` inputs
- **`base_version`** — git tag to bump from. Defaults to the latest semantic version tag (`MAJOR.MINOR.PATCH`) found in the repo.
- **`target_branch`** — branch to open the PR against. Defaults to `main`.

The previous behavior was hardcoded to check out the `humble` branch and target `main`. Both are now configurable.

#### Auto-align diverged packages before `catkin_prepare_release`
Packages whose `<version>` in `package.xml` does not match `base_version` would otherwise cause `catkin_generate_changelog` / `catkin_prepare_release` to fail. The workflow now, in a single pass over all `package.xml` files:
- rewrites the `<version>` to `base_version`, and
- deletes the package's sibling `CHANGELOG.rst`
- add PR description about deleted `CHANGELOG.rst`
## Related links

**Parent Issue:**

- https://github.com/autowarefoundation/autoware/issues/7047

## How was this PR tested?
Tested on my fork:
 - Action: https://github.com/mitsudome-r/autoware_universe/actions/runs/25223075552/job/73959844920
 - created PR: https://github.com/mitsudome-r/autoware_universe/pull/48

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/.github/workflows/bump-version-pr.yaml
+++ b/.github/workflows/bump-version-pr.yaml
@@ -77,8 +77,10 @@ jobs:
 
       - name: Resolve base version
         id: resolve_base_version
+        env:
+          INPUT_BASE_VERSION: ${{ inputs.base_version }}
         run: |
-          base_version="${{ inputs.base_version }}"
+          base_version="${INPUT_BASE_VERSION}"
           if [ -z "${base_version}" ]; then
             base_version=$(git tag --list --sort=-v:refname | grep -E '^[0-9]+\.[0-9]+\.[0-9]+$' | head -n 1)
           fi
@@ -98,6 +100,7 @@ jobs:
         env:
           BASE_VERSION: ${{ steps.resolve_base_version.outputs.base_version }}
           TARGET_BRANCH: ${{ inputs.target_branch }}
+          VERSION_PART: ${{ inputs.version_part }}
         run: |
           git checkout -b tmp/bot/bump_version_base "refs/tags/${BASE_VERSION}"
           git fetch origin "${TARGET_BRANCH}"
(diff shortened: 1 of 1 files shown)


.github/workflows/bump-version-pr.yaml
```diff
@@ -44,37 +75,128 @@ jobs:
...
+            echo "Bump version from \`${BASE_VERSION}\` to \`${NEW_VERSION}\`."
+            echo
+            if [ -s "${DELETED_CHANGELOG_PACKAGES_FILE}" ]; then
+              echo "## Packages whose \`CHANGELOG.rst\` was reset"
+              echo
+              echo "The following packages had their \`CHANGELOG.rst\` deleted before"
+              echo "regenerating changelogs (e.g. their version had diverged from"
+              echo "\`${BASE_VERSION}\`). Please double-check the regenerated changelog"
+              echo "entries for these packages:"
+              echo
```
sasakisasaki: Same as this my proposal on `autoware_core` side
```suggestion
              printf "%s\n\n%s\n" \
                     "## Packages whose \`CHANGELOG.rst\` was reset" \
                     "The following packages had their \`CHANGELOG.rst\` deleted before regenerating changelogs (e.g. their version had diverged from \`${BASE_VERSION}\`). Please double-check the regenerated changelog entries for these packages:"
```

sasakisasaki (approved):
Reviewed, actually this PR's change is already reviewed on [autoware_core side](https://github.com/autowarefoundation/autoware_core/pull/1060).

But during re-review, I found a quick win issue. Please check before merging this PR 👍

.github/workflows/bump-version-pr.yaml
sasakisasaki: Related link:
- https://github.com/autowarefoundation/autoware_core/pull/1060#discussion_r3208454351

PR #12536, opened 2026-05-01
ci: use underlay workspace for build-and-test workflows

mitsudome-r:
# Description
Resolves issue https://github.com/autowarefoundation/autoware_universe/issues/12535

## Related links

**Parent Issue:**

- https://github.com/autowarefoundation/autoware_universe/issues/12535

## How was this PR tested?
I added a [test commit](https://github.com/mitsudome-r/autoware_universe/commit/49a4eafdb28866fbfe8ada4eed9fc8eeb168a88a) temporarily to trigger the build of autoware_system_designs.

**Without underlay workspace it fails:**
https://github.com/autowarefoundation/autoware_universe/actions/runs/25227669086/job/73975189683?pr=12536
image

**With underlay workspace sourced:**
https://github.com/autowarefoundation/autoware_universe/actions/runs/25226017986/job/73969661204?pr=12536
image

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/.github/workflows/build-and-test-differential.yaml
+++ b/.github/workflows/build-and-test-differential.yaml
@@ -113,6 +113,8 @@ jobs:
           build-depends-repos: build_depends.repos
           cache-key-element: ${{ env.BUILD_TYPE_CUDA_STATE }}
           build-pre-command: ${{ inputs.build-pre-command }}
+          underlay-workspace: /opt/autoware
+
       - name: Show ccache stats after build
         run: du -sh ${CCACHE_DIR} && ccache -s
         shell: bash
@@ -125,6 +127,7 @@ jobs:
           rosdistro: ${{ inputs.rosdistro }}
           target-packages: ${{ steps.get-modified-packages.outputs.modified-packages }}
           build-depends-repos: build_depends.repos
+          underlay-workspace: /opt/autoware
 
       - name: Upload coverage to CodeCov
         if: ${{ steps.test.outputs.coverage-report-files != '' && contains(inputs.container, '-cuda') && inputs.rosdistro == 'jazzy' }}
--- a/.github/workflows/build-and-test-packages-above-differential.yaml
+++ b/.github/workflows/build-and-test-packages-above-differential.yaml
@@ -109,6 +109,7 @@ jobs:
           packages-above-repos: packages_above.repos
           cache-key-element: ${{ env.BUILD_TYPE_CUDA_STATE }}-above
           build-pre-command: ${{ inputs.build-pre-command }}
+          underlay-workspace: /opt/autoware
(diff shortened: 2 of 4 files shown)


xmfcx:
autoware_core's latest pinned version **_(from autoware.repos)_** is in the underlay workspaces due to way our docker build and push workflow works, it imports autoware.repos only.

https://github.com/autowarefoundation/autoware_universe/blob/2ad15b151803b0201c15301f2729c438f466d541/build_depends_nightly.repos#L9-L12

But this repo depends on main branch of `autoware_core`.

Adding it as an underlay workspace would cause conflicts.

- To resolve https://github.com/autowarefoundation/autoware_universe/issues/12535 we could update https://github.com/autowarefoundation/autoware_universe/blob/main/build_depends_stable.repos since autoware_system_designer is an autoware core package.

xmfcx:
Of course, we can go with underlay_ws option too.

But to do that, we should remove all mentions of `core` folder references from:
https://github.com/autowarefoundation/autoware_universe/blob/2ad15b151803b0201c15301f2729c438f466d541/build_depends_stable.repos#L2-L31

And also remove the concept of https://github.com/autowarefoundation/autoware_universe/blob/main/build_depends_nightly.repos altogether (from this repository).

It'll work. It will reduce the flexibility and will require either having to make non-breaking changes as we develop. But it will work.

---

Thinking more on this:

If we remove https://github.com/autowarefoundation/autoware_universe/blob/main/build_depends_nightly.repos then it also means that 
[autoware-nightly.repos](https://github.com/autowarefoundation/autoware/blob/889843a40a6cc9ce7b2c77ecfc81163775b157f9/repositories/autoware-nightly.repos#L26-L29) from the main repository might be left untested and unreliable.

So I'm conflicted about this, needs further discussion.

Only way I can see this working is that if we leave only autoware_universe as main branch in [autoware-nightly](https://github.com/autowarefoundation/autoware/blob/main/repositories/autoware-nightly.repos) and remove all else. This is the only way we can promise `autoware-nightly` will work if we do this modification.

mitsudome-r:
@xmfcx I also mentioned the approach in the "How to fix" section in the [linked issue](https://github.com/autowarefoundation/autoware_universe/issues/12535).

At least, autoware_core main branch is always tested with `/opt/autoware` workspace so it should be always compatible, but I agree that we should remove all core dependencies if we are making this change.

I will create another PR for now to just update the repos files instead. I will change this PR to draft and maybe bring it as a discussion topic in the next Software WG call.

mitsudome-r:
Already done in https://github.com/autowarefoundation/autoware_universe/pull/12735

PR #12547, opened 2026-05-06
feat(autoware_ptv3): source cloud reconstruction & entropy

amadeuszsz:
## Description

This PR refines the PTv3 runtime point cloud handling and output reconstruction behavior. It adds configurable source reconstruction modes so PTv3 can optionally publish predictions back on the original input point layout instead of only voxel representatives. Changes:
- Added source_reconstruction node parameter with modes:
  - none: existing behavior, publish voxelized output only.
  - partial: unvoxelize predictions back to cropped in-range input points.
  - full: unvoxelize and uncrop predictions back to the full input cloud.
    - Out-of-range or clipped points receive unknown label 255 and 0.0 class probabilities.
- Updated segmentation output to include normalized Shannon entropy for uncertainty inspection.
- Updated visualization and filtered point cloud generation to work correctly with reconstructed outputs.
- Added schema/config support for the new reconstruction mode.
- Extra:
  - Remap blackboard's (cuda) topics with dummy suffix. This let's topics negotiation fail explicitly which is expected outcome for nodes in separate process.
  - Fixes an out-of-bounds initialization of hash index buffers by matching the Thrust sequence range to the allocated cloud capacity.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

Average latency in milliseconds for sample rosbag input with RTX 3060 laptop.

| Metric  | Original | Mode `none` | Mode `partial` | Mode `full` |
| ------------- | ------------- | ------------- | ------------- | ------------- |
| Preprocessing  | 1.23 | 1.43 | 2.92 | 2.56 |  
| Inference  | 31.11  | 32.55 | 29.63 | 31.01 |
| Postprocessing  | 1.19  | 1.38 | 3.88 | 4.05 |
| Total  | 33.55  | 35.38 | 36.44 | 37.63 |

Original implementation (before this PR) and mode `none` seem to have equivalent latency. `partial` and `full` come with expected overhead.

I would suggest to keep mode `full` as default, because intuitively we expect same output points sequence as input.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


perception/autoware_ptv3/src/ptv3_node.cpp
```diff
@@ -59,6 +59,8 @@ PTv3Node::PTv3Node(const rclcpp::NodeOptions & options) : Node("ptv3", options)
     this->declare_parameter<std::vector<std::string>>("filter.classes", descriptor);
   const auto filter_output_format =
     this->declare_parameter<std::string>("filter.output_format", descriptor);
+  const auto source_reconstruction =
+    this->declare_parameter<std::string>("source_reconstruction", descriptor);
```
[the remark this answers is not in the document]
amadeuszsz: This is the expected behavior. No parameters with default value.

perception/autoware_ptv3/config/ptv3.param.yaml
```diff
@@ -5,6 +5,7 @@
     # network
     trt_precision: fp16
     cloud_capacity: 2000000
+    source_reconstruction: full
     onnx_path: "$(var model_path)/ptv3.onnx"
```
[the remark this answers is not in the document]
amadeuszsz: Default behavior assumes same output point cloud length as input.

perception/autoware_ptv3/schema/ptv3.schema.json
```diff
@@ -48,6 +48,12 @@
           "default": 2000000,
           "minimum": 1
         },
+        "source_reconstruction": {
+          "type": "string",
+          "description": "Controls whether voxel predictions are mapped back to source points. 'none' publishes voxel representative points; 'partial' publishes cropped source points inside point_cloud_range; 'full' publishes the full input cloud, assigning unknown label 255 and zero class probabilities to points outside point_cloud_range.",
+          "default": "full",
+          "enum": ["none", "partial", "full"]
+        },
```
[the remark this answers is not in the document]
amadeuszsz: This is correct.

perception/autoware_ptv3/lib/preprocess/preprocess_kernel.cu
```diff
@@ -311,37 +365,70 @@ std::size_t PreprocessCuda::generateFeatures(
...
     return 0;
   }
+  *output_num_cropped_points = num_cropped_points;
 
   extractIndicesKernel<<<num_blocks, config_.threads_per_block_, 0, stream_>>>(
     reinterpret_cast<float4 *>(points_d_.get()), crop_mask_d_.get(), crop_indices_d_.get(),
     reinterpret_cast<float4 *>(cropped_points_d_.get()), num_points);
 
+  if (
+    config_.source_reconstruction_ == SourceReconstruction::PARTIAL && source_features != nullptr) {
```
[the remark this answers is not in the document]
amadeuszsz: It was already handled, but cleaned it in https://github.com/autowarefoundation/autoware_universe/pull/12547/commits/4fc27a3e143e419ae13727dd68b198db7130cc0f

perception/autoware_ptv3/lib/preprocess/preprocess_kernel.cu
```diff
@@ -265,9 +278,43 @@ __global__ void computeGridCoordsAndSerializationKernel(
...
+    cudaMemsetAsync(hashes32_d_.get(), 0, config_.cloud_capacity_ * sizeof(std::uint32_t), stream_);
+    cudaMemsetAsync(
+      sorted_hashes32_d_.get(), 0, config_.cloud_capacity_ * sizeof(std::uint32_t), stream_);
+    cudaMemsetAsync(
+      sorted_hash_indexes32_d_.get(), 0, config_.cloud_capacity_ * sizeof(std::uint32_t), stream_);
+    cudaMemsetAsync(
+      unique_mask32_d_.get(), 0, config_.cloud_capacity_ * sizeof(std::uint32_t), stream_);
+    cudaMemsetAsync(
+      unique_indices32_d_.get(), 0, config_.cloud_capacity_ * sizeof(std::uint32_t), stream_);
+  }
```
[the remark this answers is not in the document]
amadeuszsz: Actually don't need to clear the memory as we process indices up to number of input points. https://github.com/autowarefoundation/autoware_universe/pull/12547/commits/49bc420e33d579144c0c2bcc81e0ce005003cbf2

perception/autoware_ptv3/lib/ptv3_trt.cpp
```diff
@@ -352,15 +369,51 @@ bool PTv3TRT::preProcess(const std::shared_ptr<const cuda_blackboard::CudaPointC
...
+  }
+
+  std::size_t num_cropped_points = 0;
   num_voxels_ = pre_ptr_->generateFeatures(
     msg_ptr->data.get(), input_format_, num_points, feat_d_.get(), grid_coord_d_.get(),
-    serialized_code_d_.get(), compact_points_d_.get());
+    serialized_code_d_.get(), compact_points_d_.get(),
+    reconstruct_source ? reconstructed_features_d_.get() : nullptr,
+    reconstruct_source ? cropped_source_points_d_.get() : nullptr,
+    reconstruct_source ? inverse_map_d_.get() : nullptr, &num_cropped_points);
```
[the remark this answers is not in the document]
amadeuszsz: https://github.com/autowarefoundation/autoware_universe/pull/12547/commits/5894e2da6b9b3ef22334cd083acce9e51efecef5

perception/autoware_ptv3/launch/ptv3.launch.xml
mojomex, line 1: Temporary changes made it into the PR

perception/autoware_ptv3/lib/postprocess/postprocess_kernel.cu
```diff
@@ -26,19 +26,21 @@ struct OutputSegmentationPointType
   float z;
   std::uint8_t class_id;
   float probability;
+  float entropy;
```
mojomex, line 29: Is this expected to be used by downstream modules? If it's only for debug/evaluation it should ideally be disabled/excluded in a production context.

amadeuszsz: @ktro2828 do you plan to use `~/ptv3/segmentation` topic? Are you ok with this change?

Anyway, if `segmentation` topic will be used in any downstream module, I think `entropy` can be useful as `probability` can be overconfident. So far PTv3 does not support real confidence output, therefore maybe we can keep `entropy` as extra measurement. Please note we do not use this publisher if not subscribed, so the overhead of this change is negligible.

perception/autoware_ptv3/lib/postprocess/postprocess_kernel.cu
```diff
@@ -55,11 +57,71 @@ __global__ void createSegmentationPointcloudKernel(
...
-  output_points[idx].probability = pred_probs[idx * num_classes + label];
+  output_points[idx].class_id = has_valid_label ? static_cast<std::uint8_t>(label) : 255U;
+  output_points[idx].probability = has_valid_label ? pred_probs[idx * num_classes + label] : 0.0f;
+  output_points[idx].entropy = entropy;
+}
+
+__global__ void reconstructPartialKernel(
+  const std::int64_t * inverse_map, const std::int64_t * voxel_labels, const float * voxel_probs,
+  std::int64_t * output_labels, float * output_probs, std::size_t num_classes,
+  std::size_t num_cropped_points, std::size_t num_voxels)
```
manato:
```suggestion
  const std::int64_t * __restrict__ inverse_map, const std::int64_t * __restrict__ voxel_labels, const float * __restrict__ voxel_probs,
  std::int64_t * __restrict__ output_labels, float * __restrict__ output_probs, std::size_t num_classes,
  std::size_t num_cropped_points, std::size_t num_voxels)
```

If (and only if) you are sure that the region specified by any argument pointers is never overlapped, you can add `__restrict__` for all argument pointers for better compiler optimization.
ref: https://developer.nvidia.com/blog/cuda-pro-tip-optimize-pointer-aliasing/

perception/autoware_ptv3/lib/postprocess/postprocess_kernel.cu
```diff
@@ -55,11 +57,71 @@ __global__ void createSegmentationPointcloudKernel(
...
+    return;
+  }
+
+  const auto voxel_idx = inverse_map[idx];
+  const bool has_valid_voxel = voxel_idx >= 0 && static_cast<std::size_t>(voxel_idx) < num_voxels;
+  output_labels[idx] = has_valid_voxel ? voxel_labels[voxel_idx] : 255;
+  for (std::size_t class_idx = 0; class_idx < num_classes; ++class_idx) {
+    output_probs[idx * num_classes + class_idx] =
+      has_valid_voxel ? voxel_probs[voxel_idx * num_classes + class_idx] : 0.0f;
+  }
```
manato: At a glance, it looks like this loop still can be parallelized because each iteration is independent of the others. Do you think we can add one more (i.e., `y`) block dimension for more parallelization?

manato (commented):
@amadeuszsz 
I left a few comments on the CUDA code. Your consideration would be appreciated!

perception/autoware_ptv3/lib/postprocess/postprocess_kernel.cu
amadeuszsz: I tested it and postprocessing went up from 4 ms to 7 ms:
- preprocessing: 2.65 ms
- inference: 35.48 ms
- postprocessing: 7.03 ms
- total: 45.18 ms

Each block does trivially little work, so the GPU block scheduler overhead far exceeds the cost of the original 20~ iteration `for` loop.

perception/autoware_ptv3/lib/postprocess/postprocess_kernel.cu
ktro2828: @amadeuszsz From the perspective of the downstream module, it's fine to change the field unless the existing fields are not deleted.
Currently, the downstream module refers to `x/y/z` and optionally `class_id/probability` by field name using `PointCloud2ConstIterator`, so adding some other field does not affect iteration order.

perception/autoware_ptv3/lib/postprocess/postprocess_kernel.cu
manato: Thank you for testing. If you have spare time, it would be great if you also try adding `y` dimension to the grid instead of the block (i.e., `auto y = blockIdx.y;` instead of `auto y = threadIdx.y;`) because this generates a separate executable unit that would be executed on the different SMs, which requires less overhead.

This is not mandatory. I can approve this PR if there are no other unhandled comments from other reviews.

amadeuszsz: I tried multiple configuration, seems [this](https://github.com/autowarefoundation/autoware_universe/pull/12547/commits/f6433aa6af1a76687058429d522b6588d3dae2ce) is the best. Results are very close to these reported in PR description.

Note: I assumed more than 32 classes (soon) and this is important due to warp size.

manato: Thank you very much for your consideration! It looks good to me.

manato (approved):
Since my concerns have been treated, and it seems there are no remaining untreated comments from other reviews, I approve this PR.

PR #12550, opened 2026-05-07
refactor(autoware_trajectory_modifier): split per-frame inputs from long-lived context

interimadd:
## Description

Refactor the plugin interface of `autoware_trajectory_modifier` to make data flow explicit at the call boundary, by separating short-lived per-frame inputs from long-lived shared state.

Previously, all data consumed by plugins lived inside a single `TrajectoryModifierData` struct that was held as a `std::shared_ptr` member (`data_`) on every plugin. It mixed two categories of data with different lifetimes:

- Per-frame ROS messages (`current_odometry`, `current_acceleration`, `predicted_objects`, `obstacle_pointcloud`)
- Long-lived shared resources (`vehicle_info`, `tf_buffer`, `tf_listener`)

Plugin methods (`modify_trajectory`, `is_trajectory_modification_required`) took no input arguments — they implicitly read mutable per-frame state via the shared pointer. This obscured which inputs each plugin actually consumed.

A further motivation is testability. With per-frame inputs hidden behind a shared `TrajectoryModifierData` pointer, exercising a plugin in a unit test required constructing and mutating the shared struct via the plugin's `initialize()` path. After this refactor, `InputData` is a plain value passed as a method argument, so a unit test can populate it directly and call `modify_trajectory` / `is_trajectory_modification_required` without standing up the full plugin lifecycle. This PR is intentionally scoped to the interface change; follow-up PRs will add unit tests for `ObstacleStop` on top of this foundation.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

- `colcon build --packages-select autoware_trajectory_modifier --cmake-args -DCMAKE_BUILD_TYPE=Release` succeeds.
- `colcon test --packages-select autoware_trajectory_modifier --event-handlers console_cohesion+` passes, including the existing `stop_point_fixer` integration tests that were updated alongside the interface change.
- [Evaluator](https://evaluation.tier4.jp/evaluation/reports/4cd9c3cf-9bc7-5f04-88bc-cf7290693a9a/?project_id=autoware_dev)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


planning/autoware_trajectory_modifier/include/autoware/trajectory_modifier/trajectory_modifier_plugins/obstacle_stop.hpp
```diff
@@ -45,10 +47,10 @@ class ObstacleStop : public TrajectoryModifierPluginBase
 public:
   ObstacleStop() = default;
 
-  bool modify_trajectory(TrajectoryPoints & traj_points) override;
+  bool modify_trajectory(TrajectoryPoints & traj_points, const InputData & inputs) override;
```
mkquda: (**Nitpick**) I think using the singular `input` is more standard for passing a struct, plural usually is used with list types (array, vector, etc)

mkquda (approved):
LGTM.

planning/autoware_trajectory_modifier/include/autoware/trajectory_modifier/trajectory_modifier_plugins/obstacle_stop.hpp
interimadd: Thank you for the suggestion. I have addressed it in https://github.com/autowarefoundation/autoware_universe/pull/12550/commits/3464e5ce0cb83626a0c2c83c5556b9ec624f9f17

planning/autoware_trajectory_modifier/include/autoware/trajectory_modifier/trajectory_modifier_context.hpp
```diff
@@ -12,10 +12,9 @@
...
 // limitations under the License.
 
-#ifndef AUTOWARE__TRAJECTORY_MODIFIER__TRAJECTORY_MODIFIER_STRUCTS_HPP_
-#define AUTOWARE__TRAJECTORY_MODIFIER__TRAJECTORY_MODIFIER_STRUCTS_HPP_
+#ifndef AUTOWARE__TRAJECTORY_MODIFIER__TRAJECTORY_MODIFIER_CONTEXT_HPP_
+#define AUTOWARE__TRAJECTORY_MODIFIER__TRAJECTORY_MODIFIER_CONTEXT_HPP_
 #include <autoware_vehicle_info_utils/vehicle_info_utils.hpp>
-#include <tl_expected/expected.hpp>
 
 #include <autoware_perception_msgs/msg/predicted_objects.hpp>
```
go-sakayori: nitpick: After moving the per-frame message pointers to `InputData`, these message includes and aliases in `trajectory_modifier_context.hpp` look unused. Could we remove them so the context header only depends on the long-lived resources it actually owns?

go-sakayori (approved):

planning/autoware_trajectory_modifier/include/autoware/trajectory_modifier/trajectory_modifier_context.hpp
interimadd: Thank you for pointing that out! I have removed unused header include in https://github.com/autowarefoundation/autoware_universe/pull/12550/commits/d3f92a4dca65834597e5664e39351c3b8e8e8c9b

PR #12554, opened 2026-05-07
perf(autoware_tensorrt_plugins): remove Thrust from sort kernels

mojomex:
## Summary

Removes Thrust, syncs and host/device copies from the TensorRT `Argsort` and `CustomUnique` plugins.

Benchmarks below are made using an internal version of PTv3.

## PR Stack

- [perf(autoware_tensorrt_plugins): remove Thrust from sort kernels](https://github.com/autowarefoundation/autoware_universe/pull/12554) (`ptv3-t18-no-thrust-c8f76ed-20260506`)
- [perf(autoware_tensorrt_plugins): keep SegmentCSR allocation-free](https://github.com/autowarefoundation/autoware_universe/pull/12555) (`ptv3-t18-no-thrust-no-alloc-e9515b790-20260506`)
- [perf(autoware_tensorrt_plugins): keep plugin outputs stream-ordered](https://github.com/autowarefoundation/autoware_universe/pull/12556) (`ptv3-t18-no-thrust-no-alloc-no-sync-13f3672a0-20260506`)
- [perf(autoware_tensorrt_plugins): cache max sort workspace bounds](https://github.com/autowarefoundation/autoware_universe/pull/12557) (`ptv3-t18-no-thrust-no-alloc-no-sync-maxnumel-47bf5656f-20260506`)
- [fix(ptv3): set TensorRT max auxiliary streams to 1](https://github.com/autowarefoundation/autoware_universe/pull/12558) (`ptv3-t18-no-thrust-no-alloc-no-sync-maxnumel-maxauxstreams1`)

### CustomUnique

* replaced Thrust with raw CUB calls
* cleaned up and documented workspace layout
* removed unnecessary result size copy to host and back to device, keeping result on GPU directly instead

| Description | Nsys |
|-|-|
| Before | Screenshot from 2026-05-26 14-36-57 |
| After (CUDA) | Screenshot from 2026-05-26 14-37-07 |
| After (GPU) | Screenshot from 2026-05-26 14-38-33 |

### Argsort

* replaced Thrust with raw CUB calls
* cleaned up and documented workspace layout

| Description | Nsys |
|-|-|
| Before | Screenshot from 2026-05-26 14-42-22|
| After (CUDA) | Screenshot from 2026-05-26 14-42-27|
| After (GPU) | Screenshot from 2026-05-26 14-42-51|

## Benchmarks

| Variant | Measurement | CPU mean (ms) | CPU p95 (ms) | CPU faster vs baseline | GPU mean (ms) | GPU p95 (ms) | GPU faster vs baseline | Mean voxels |
| --- | --- | --- | --- | --- | --- | --- | --- | --- |
| ptv3-t18 | series (7 x 50) | 30.542 | 33.195 | +0.0% | 29.419 | 32.023 | +0.0% | 120572 |
| ptv3-t18-pr12554 | series (7 x 50) | 28.611 | 29.198 | +6.7% | 28.599 | 29.184 | +2.9% | 120572 |

### Relative Performance

relative_to_baseline

### DIFF
--- a/perception/autoware_tensorrt_plugins/include/autoware/tensorrt_plugins/plugin_utils.hpp
+++ b/perception/autoware_tensorrt_plugins/include/autoware/tensorrt_plugins/plugin_utils.hpp
@@ -16,6 +16,7 @@
 #define AUTOWARE__TENSORRT_PLUGINS__PLUGIN_UTILS_HPP_
 
 #include <NvInferRuntime.h>
+#include <cuda_runtime_api.h>
 
 #include <cstdint>
 #include <exception>
@@ -25,6 +26,11 @@ void caughtError(std::exception const & e);
 void logDebug(char const * msg);
 
 void logWarning(char const * msg);
+
+cudaError_t reportCudaStatus(
+	cudaError_t status, char const * msg, char const * file, std::int32_t line);
+
+#define PLUGIN_CUDA_CHECK(val) reportCudaStatus((val), #val, __FILE__, __LINE__)
 
 #define PLUGIN_ASSERT(val) reportAssertion((val), #val, __FILE__, __LINE__)
 void reportAssertion(bool success, char const * msg, char const * file, std::int32_t line);
--- a/perception/autoware_tensorrt_plugins/include/autoware/tensorrt_plugins/plugin_utils.hpp
+++ b/perception/autoware_tensorrt_plugins/include/autoware/tensorrt_plugins/plugin_utils.hpp
@@ -28,7 +28,7 @@ void logDebug(char const * msg);
 void logWarning(char const * msg);
 
 cudaError_t reportCudaStatus(
-	cudaError_t status, char const * msg, char const * file, std::int32_t line);
+  cudaError_t status, char const * msg, char const * file, std::int32_t line);
 
 #define PLUGIN_CUDA_CHECK(val) reportCudaStatus((val), #val, __FILE__, __LINE__)
 
--- a/perception/autoware_tensorrt_plugins/src/argsort_ops/argsort.cu
+++ b/perception/autoware_tensorrt_plugins/src/argsort_ops/argsort.cu
@@ -17,6 +17,8 @@
 
 #include <cub/cub.cuh>
 
+#include <cstddef>
+
 namespace
 {
 
@@ -35,7 +37,7 @@ cudaError_t argsort(
   const auto scratch_offset =
     autoware::tensorrt_plugins::align_up(argsort_workspace_size, alignof(std::int64_t));
   auto * input_idx_d =
-    reinterpret_cast<std::int64_t *>(reinterpret_cast<char *>(workspace) + scratch_offset);
+    reinterpret_cast<std::int64_t *>(reinterpret_cast<std::byte *>(workspace) + scratch_offset);
   auto * input_sorted_d = input_idx_d + num_elements;
 
   const auto num_blocks =
--- a/perception/autoware_tensorrt_plugins/src/argsort_ops/argsort.cu
+++ b/perception/autoware_tensorrt_plugins/src/argsort_ops/argsort.cu
@@ -54,8 +54,7 @@ cudaError_t argsort(
   const auto num_blocks =
     static_cast<unsigned int>((num_elements + kThreadsPerBlock - 1U) / kThreadsPerBlock);
   fill_iota<<<num_blocks, kThreadsPerBlock, 0, stream>>>(input_idx_d, num_elements);
-  cudaError_t status = cudaGetLastError();
-  if (status != cudaSuccess) {
+  if (const auto status = cudaPeekAtLastError(); status != cudaSuccess) {
     return status;
   }
 
--- a/perception/autoware_tensorrt_plugins/src/argsort_ops/argsort.cu
+++ b/perception/autoware_tensorrt_plugins/src/argsort_ops/argsort.cu
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include "autoware/argsort_ops/argsort.hpp"
+#include "autoware/tensorrt_plugins/kernel_utils.hpp"
 
 #include <cub/cub.cuh>
 
@@ -20,21 +21,6 @@ namespace
 {
 
 constexpr int kThreadsPerBlock = 256;
-
-std::size_t align_up(const std::size_t size, const std::size_t alignment)
-{
-  return ((size + alignment - 1U) / alignment) * alignment;
-}
-
-__global__ void fill_iota(std::int64_t * output, const std::size_t num_elements)
-{
-  const auto index = static_cast<std::size_t>(blockIdx.x) * blockDim.x + threadIdx.x;
-  if (index >= num_elements) {
-    return;
-  }
-
-  output[index] = static_cast<std::int64_t>(index);
-}
 
 }  // namespace
 
@@ -46,14 +32,16 @@ cudaError_t argsort(
     return cudaSuccess;
   }
 
-  const auto scratch_offset = align_up(argsort_workspace_size, alignof(std::int64_t));
+  const auto scratch_offset =
+    autoware::tensorrt_plugins::align_up(argsort_workspace_size, alignof(std::int64_t));
   auto * input_idx_d =
     reinterpret_cast<std::int64_t *>(reinterpret_cast<char *>(workspace) + scratch_offset);
   auto * input_sorted_d = input_idx_d + num_elements;
 
   const auto num_blocks =
     static_cast<unsigned int>((num_elements + kThreadsPerBlock - 1U) / kThreadsPerBlock);
-  fill_iota<<<num_blocks, kThreadsPerBlock, 0, stream>>>(input_idx_d, num_elements);
+  autoware::tensorrt_plugins::fill_iota<<<num_blocks, kThreadsPerBlock, 0, stream>>>(
+    input_idx_d, num_elements);
   if (const auto status = cudaPeekAtLastError(); status != cudaSuccess) {
     return status;
   }
--- a/perception/autoware_tensorrt_plugins/src/argsort_ops/argsort.cu
+++ b/perception/autoware_tensorrt_plugins/src/argsort_ops/argsort.cu
@@ -16,25 +16,52 @@
 
 #include <cub/cub.cuh>
 
-#include <thrust/device_ptr.h>
-#include <thrust/execution_policy.h>
-#include <thrust/sequence.h>
+namespace
+{
+
+constexpr int kThreadsPerBlock = 256;
+
+std::size_t align_up(const std::size_t size, const std::size_t alignment)
+{
+  return ((size + alignment - 1U) / alignment) * alignment;
+}
+
+__global__ void fill_iota(std::int64_t * output, const std::size_t num_elements)
+{
+  const auto index = static_cast<std::size_t>(blockIdx.x) * blockDim.x + threadIdx.x;
+  if (index >= num_elements) {
+    return;
+  }
+
+  output[index] = static_cast<std::int64_t>(index);
+}
+
+}  // namespace
 
 cudaError_t argsort(
   const std::int64_t * input_d, std::int64_t * output_d, void * workspace, std::size_t num_elements,
   std::size_t argsort_workspace_size, cudaStream_t stream)
 {
-  int workspace_offset = (argsort_workspace_size + sizeof(std::int64_t) - 1) / sizeof(std::int64_t);
-  thrust::device_ptr<std::int64_t> idx_ptr(
-    &reinterpret_cast<std::int64_t *>(workspace)[workspace_offset]);
+  if (num_elements == 0U) {
+    return cudaSuccess;
+  }
 
-  thrust::sequence(thrust::cuda::par.on(stream), idx_ptr, idx_ptr + num_elements, 0);
+  const auto scratch_offset = align_up(argsort_workspace_size, alignof(std::int64_t));
+  auto * input_idx_d =
+    reinterpret_cast<std::int64_t *>(reinterpret_cast<char *>(workspace) + scratch_offset);
+  auto * input_sorted_d = input_idx_d + num_elements;
 
-  std::int64_t * input_sorted_d = thrust::raw_pointer_cast(idx_ptr) + num_elements;
+  const auto num_blocks =
+    static_cast<unsigned int>((num_elements + kThreadsPerBlock - 1U) / kThreadsPerBlock);
+  fill_iota<<<num_blocks, kThreadsPerBlock, 0, stream>>>(input_idx_d, num_elements);
+  cudaError_t status = cudaGetLastError();
+  if (status != cudaSuccess) {
+    return status;
+  }
 
   return cub::DeviceRadixSort::SortPairs(
-    workspace, argsort_workspace_size, input_d, input_sorted_d, thrust::raw_pointer_cast(idx_ptr),
-    output_d, num_elements, 0, 64, stream);
+    workspace, argsort_workspace_size, input_d, input_sorted_d, input_idx_d, output_d, num_elements,
+    0, 64, stream);
 }
 
 std::size_t get_argsort_workspace_size(std::size_t num_elements)
--- a/perception/autoware_tensorrt_plugins/src/unique_ops/unique.cu
+++ b/perception/autoware_tensorrt_plugins/src/unique_ops/unique.cu
@@ -185,9 +185,9 @@ UniqueWorkspaceLayout make_unique_workspace_layout(
     reinterpret_cast<std::int32_t *>(sorted_input_positions + num_input_elements_in);
   auto * run_ids = reinterpret_cast<std::int32_t *>(num_unique + 1U);
 
-  return UniqueWorkspaceLayout{
-    workspace_inout, cub_temp_storage_size_in, input_positions, sorted_input,
-    sorted_input_positions, num_unique, run_ids};
+  return UniqueWorkspaceLayout{workspace_inout, cub_temp_storage_size_in, input_positions,
+                               sorted_input,    sorted_input_positions,   num_unique,
+                               run_ids};
 }
 
 __global__ void mark_run_starts(
--- a/perception/autoware_tensorrt_plugins/src/unique_ops/unique.cu
+++ b/perception/autoware_tensorrt_plugins/src/unique_ops/unique.cu
@@ -257,14 +257,13 @@ cudaError_t unique(
   const std::int64_t * input_in, std::int64_t * unique_values_out,
   std::int64_t * inverse_indices_out, std::int64_t * unique_counts_out,
   std::int64_t * num_unique_elements_out, void * workspace_inout, std::size_t num_input_elements_in,
-  std::size_t workspace_size_in, cudaStream_t stream_in)
+  [[maybe_unused]] std::size_t workspace_size_in, cudaStream_t stream_in)
 {
   if (num_input_elements_in == 0U) {
     return cudaMemsetAsync(num_unique_elements_out, 0, sizeof(std::int64_t), stream_in);
   }
 
   assert(workspace_size_in >= get_unique_workspace_size(num_input_elements_in));
-  (void)workspace_size_in;
 
   const auto cub_temp_storage_size = get_unique_temp_storage_size(num_input_elements_in);
   auto layout =
--- a/perception/autoware_tensorrt_plugins/src/unique_ops/unique.cu
+++ b/perception/autoware_tensorrt_plugins/src/unique_ops/unique.cu
@@ -101,77 +101,153 @@
 
 #include <cub/cub.cuh>
 
-#include <thrust/adjacent_difference.h>
-#include <thrust/device_ptr.h>
-#include <thrust/execution_policy.h>
-#include <thrust/scan.h>
-#include <thrust/scatter.h>
-#include <thrust/sequence.h>
-#include <thrust/sort.h>
-#include <thrust/unique.h>
+#include <algorithm>
+#include <cstdint>
+
+namespace
+{
+
+constexpr int kThreadsPerBlock = 256;
+
+std::size_t align_up(const std::size_t size, const std::size_t alignment)
+{
+  return ((size + alignment - 1U) / alignment) * alignment;
+}
+
+std::size_t query_unique_temp_storage_size(const std::size_t num_elements)
+{
+  std::size_t sort_temp_size = 0;
+  std::size_t scan_temp_size = 0;
+  std::size_t unique_temp_size = 0;
+
+  std::int64_t * int64_nullptr = nullptr;
+  std::int32_t * int32_nullptr = nullptr;
+
+  cub::DeviceRadixSort::SortPairs(
+    nullptr, sort_temp_size, int64_nullptr, int64_nullptr, int64_nullptr, int64_nullptr,
+    num_elements, 0, 64, nullptr);
+  cub::DeviceScan::InclusiveSum(
+    nullptr, scan_temp_size, int32_nullptr, int32_nullptr, num_elements, nullptr);
+  cub::DeviceSelect::UniqueByKey(
+    nullptr, unique_temp_size, int64_nullptr, int64_nullptr, int64_nullptr, int64_nullptr,
+    int64_nullptr, num_elements, nullptr);
+
+  return std::max(sort_temp_size, std::max(scan_temp_size, unique_temp_size));
+}
+
+__global__ void mark_run_starts(
+  const std::int64_t * sorted_input, std::int32_t * run_ids, const std::size_t num_input_elements)
+{
+  const auto index = static_cast<std::size_t>(blockIdx.x) * blockDim.x + threadIdx.x;
+  if (index >= num_input_elements) {
+    return;
+  }
+
+  run_ids[index] = (index == 0U || sorted_input[index] != sorted_input[index - 1U]) ? 1 : 0;
+}
+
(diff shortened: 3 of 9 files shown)


mojomex:
@codex review. Is the implementation memory safe and has no race conditions? Do the unit tests guarantee correct behavior? Is the implementation as concise and readable as possible?

mojomex:
@codex Code isn't documented and behavior is not immediately obvious to the reader, e.g. `write_unique_offset_sentinel`. Document function contracts/pre/post conditions concisely, annotate argsort and unique flows so user can easily follow. For design choices, e.g. where to put sentinels etc., explain the "why".

ktro2828 (approved):
confirmed it worked and code looks good to me!

perception/autoware_tensorrt_plugins/include/autoware/tensorrt_plugins/plugin_utils.hpp
```diff
@@ -24,6 +25,11 @@ void caughtError(std::exception const & e);
 
 void logDebug(char const * msg);
 
+cudaError_t reportCudaStatus(
```
KSeangTan, line 30: Should we move this function to a higher-level? Or is it specifically only for plugins?

perception/autoware_tensorrt_plugins/src/argsort_ops/argsort.cu
```diff
@@ -16,25 +16,51 @@
 
 #include <cub/cub.cuh>
 
-#include <thrust/device_ptr.h>
-#include <thrust/execution_policy.h>
-#include <thrust/sequence.h>
+namespace
+{
+
+constexpr int kThreadsPerBlock = 256;
```
KSeangTan, line 25: Is this better to be a hard requirement?

perception/autoware_tensorrt_plugins/src/argsort_ops/argsort.cu
```diff
@@ -16,25 +16,51 @@
...
+}  // namespace
 
 cudaError_t argsort(
   const std::int64_t * input_d, std::int64_t * output_d, void * workspace, std::size_t num_elements,
   std::size_t argsort_workspace_size, cudaStream_t stream)
 {
-  int workspace_offset = (argsort_workspace_size + sizeof(std::int64_t) - 1) / sizeof(std::int64_t);
-  thrust::device_ptr<std::int64_t> idx_ptr(
-    &reinterpret_cast<std::int64_t *>(workspace)[workspace_offset]);
+  if (num_elements == 0U) {
```
KSeangTan, line 33: I am thinking should we throw some warnings/errors when there's no elements. Or the function calling it should do the checking beforehand.

perception/autoware_tensorrt_plugins/src/argsort_ops/argsort.cu
```diff
@@ -16,25 +16,51 @@
...
-  thrust::sequence(thrust::cuda::par.on(stream), idx_ptr, idx_ptr + num_elements, 0);
+  const auto scratch_offset = align_up(argsort_workspace_size, alignof(std::int64_t));
+  auto * input_idx_d =
+    reinterpret_cast<std::int64_t *>(reinterpret_cast<char *>(workspace) + scratch_offset);
+  auto * input_sorted_d = input_idx_d + num_elements;
 
-  std::int64_t * input_sorted_d = thrust::raw_pointer_cast(idx_ptr) + num_elements;
+  const auto num_blocks =
+    static_cast<unsigned int>((num_elements + kThreadsPerBlock - 1U) / kThreadsPerBlock);
+  fill_iota<<<num_blocks, kThreadsPerBlock, 0, stream>>>(input_idx_d, num_elements);
```
KSeangTan: Do you think is it cleaner if we use `CHECK_CUDA_ERROR` instead of `cudaPeekAtLastError` afterwards.

perception/autoware_tensorrt_plugins/src/unique_ops/unique.cu
```diff
@@ -101,77 +101,259 @@
...
+#include <cassert>
+#include <cstddef>
+#include <cstdint>
+
+namespace
+{
+
+constexpr int kThreadsPerBlock = 256;
+
+std::size_t align_up(const std::size_t size, const std::size_t alignment)
```
KSeangTan: Should we reuse the same function above?

perception/autoware_tensorrt_plugins/src/unique_ops/unique.cu
```diff
@@ -101,77 +101,259 @@
...
+  const std::int64_t * sorted_input_in, std::int32_t * run_ids_out,
+  const std::size_t num_input_elements_in)
+{
+  const auto index = static_cast<std::size_t>(blockIdx.x) * blockDim.x + threadIdx.x;
+  if (index >= num_input_elements_in) {
+    return;
+  }
 
-  std::int64_t num_out;
+  run_ids_out[index] =
```
KSeangTan, line 202: is it cleaner if we write `if-else`?

Do you think if `warp divergence` might happen here?

perception/autoware_tensorrt_plugins/src/unique_ops/unique.cu
```diff
@@ -101,77 +101,259 @@
...
+    (index == 0U || sorted_input_in[index] != sorted_input_in[index - 1U]) ? 1 : 0;
+}
 
-  std::int64_t * range_ptr = idx_ptr.get();
-  num_out =
-    thrust::unique_by_key(policy, sorted_input, sorted_input + num_input_elements, range_ptr, equal)
-      .first -
-    sorted_input;
+__global__ void fill_iota(std::int64_t * output_out, const std::size_t num_input_elements_in)
```
KSeangTan: Possible to reuse the same function above?

perception/autoware_tensorrt_plugins/src/unique_ops/unique.cu
```diff
@@ -101,77 +101,259 @@
...
+  const std::int64_t * sorted_idx_in, const std::int32_t * run_ids_in,
+  std::int64_t * inverse_indices_out, const std::size_t num_input_elements_in)
+{
+  const auto index = static_cast<std::size_t>(blockIdx.x) * blockDim.x + threadIdx.x;
+  if (index >= num_input_elements_in) {
+    return;
+  }
 
-  return num_out;
+  inverse_indices_out[sorted_idx_in[index]] = static_cast<std::int64_t>(run_ids_in[index] - 1);
```
KSeangTan, line 216: Comment why it reduces 1.

perception/autoware_tensorrt_plugins/src/unique_ops/unique.cu
```diff
@@ -101,77 +101,259 @@
...
+__global__ void write_unique_counts(
+  const std::int64_t * unique_offsets_in, const std::int32_t * num_unique_in,
+  const std::int64_t * unique_offsets_end_in, std::int64_t * unique_counts_out)
+{
+  const auto index = static_cast<std::size_t>(blockIdx.x) * blockDim.x + threadIdx.x;
+  if (index >= static_cast<std::size_t>(*num_unique_in)) {
+    return;
+  }
+
+  const auto next_offset = (index + 1U == static_cast<std::size_t>(*num_unique_in))
```
KSeangTan: Is it cleaner if we write it in `if-else` instead?

perception/autoware_tensorrt_plugins/src/unique_ops/unique.cu
```diff
@@ -101,77 +101,259 @@
...
+  std::int64_t * inverse_indices_out, std::int64_t * unique_counts_out,
+  std::int64_t * num_unique_elements_out, void * workspace_inout, std::size_t num_input_elements_in,
+  std::size_t workspace_size_in, cudaStream_t stream_in)
+{
+  if (num_input_elements_in == 0U) {
+    return cudaMemsetAsync(num_unique_elements_out, 0, sizeof(std::int64_t), stream_in);
+  }
+
+  assert(workspace_size_in >= get_unique_workspace_size(num_input_elements_in));
+  (void)workspace_size_in;
```
KSeangTan: Sorry, would you mind explaining the purpose of this line?

perception/autoware_tensorrt_plugins/src/unique_ops/unique.cu
```diff
@@ -101,77 +101,259 @@
...
+  const auto cub_temp_storage_size = get_unique_temp_storage_size(num_input_elements_in);
+  auto layout =
+    make_unique_workspace_layout(workspace_inout, num_input_elements_in, cub_temp_storage_size);
 
-  return temp_size + (4 * num_elements + 1) * sizeof(std::int64_t);
+  const auto num_blocks =
+    static_cast<unsigned int>((num_input_elements_in + kThreadsPerBlock - 1U) / kThreadsPerBlock);
+
+  // 1. Sort values while carrying their original input positions.
+  fill_iota<<<num_blocks, kThreadsPerBlock, 0, stream_in>>>(
```
KSeangTan: Maybe we should use `CHECK_CUDA_ERROR` instead.

KSeangTan (approved):
Overall, it looks good to me. Would you mind addressing the comments related to style? Thanks.

KSeangTan:
It would be also good if you could attach some results to verify if the outputs are same before/after changes.

perception/autoware_tensorrt_plugins/src/unique_ops/unique.cu
mojomex: `assert` is to document developer assumptions, and to fail unit tests if those assumptions are broken.

As `assert` is not compiled in most release configurations, `workspace_size_in` would be an unused parameter and the compiler would complain, so putting `(void)workspace_size_in` explicitly ignores the param.

There's [discussions](https://discourse.llvm.org/t/maybe-unused-vs-attribute-unused-vs-void-var-for-assert-variables/64212) online whether there's nicer ways of doing this, such as `[[maybe_unused]]` which I'm open for as well.

perception/autoware_tensorrt_plugins/include/autoware/tensorrt_plugins/plugin_utils.hpp
mojomex: It could be re-used in other places, but I feel it's small and understandable enough that it's fine to have a few duplicates in other modules.

Putting it in a common utils package would force every module to depend on that package, adding probably more code and confusion than duplicating here.

perception/autoware_tensorrt_plugins/src/argsort_ops/argsort.cu
mojomex: Not sure if I understand what you mean by hard requirement.

This is pretty arbitrary ([here some NVIDIA advice](https://forums.developer.nvidia.com/t/how-to-choose-how-many-threads-blocks-to-have/55529)) and does not affect computational outcome nor causes any restrictions.

perception/autoware_tensorrt_plugins/src/argsort_ops/argsort.cu
mojomex: For this specific algorithm, 0 is a valid input size, and since it can be reused in all kinds of models, I'd like to keep such restrictions out of it.

As you suggested, I think the caller should do checking if necessary.

perception/autoware_tensorrt_plugins/src/argsort_ops/argsort.cu
mojomex: `CHECK_CUDA_ERROR` throws, and that's not a good thing to do in a TensorRT plugin. There's not much documentation, but it seems that TensorRT wants us to return `-1` on error in `enqueue()`. TensorRT also seems to provide an `IErrorRecorder` and `ILogger` for more advanced error reporting.

perception/autoware_tensorrt_plugins/src/unique_ops/unique.cu
mojomex: See comment above

perception/autoware_tensorrt_plugins/src/argsort_ops/argsort.cu
```diff
@@ -16,25 +16,51 @@
...
-  thrust::device_ptr<std::int64_t> idx_ptr(
-    &reinterpret_cast<std::int64_t *>(workspace)[workspace_offset]);
+  if (num_elements == 0U) {
+    return cudaSuccess;
+  }
 
-  thrust::sequence(thrust::cuda::par.on(stream), idx_ptr, idx_ptr + num_elements, 0);
+  const auto scratch_offset = align_up(argsort_workspace_size, alignof(std::int64_t));
+  auto * input_idx_d =
+    reinterpret_cast<std::int64_t *>(reinterpret_cast<char *>(workspace) + scratch_offset);
```
manato:
```suggestion
   reinterpret_cast<std::int64_t *>(reinterpret_cast<std::byte *>(workspace) + scratch_offset);
```
[nits] You can utilize `std::byte` to represent your intention of pointer arithmetic explicitly.

manato (commented):
@mojomex 
Thank you for your contribution! Overall, it looks good to me. I left some comments suggesting alternative implementations, but handling them is not mandatory!

perception/autoware_tensorrt_plugins/src/unique_ops/unique.cu
```diff
@@ -101,77 +101,259 @@
...
+  if (const auto status = cub::DeviceRadixSort::SortPairs(
+        layout.cub_temp_storage, layout.cub_temp_storage_size, input_in, layout.sorted_input,
+        layout.input_positions, layout.sorted_input_positions, num_input_elements_in, 0, 64,
+        stream_in);
+      status != cudaSuccess) {
+    return status;
+  }
+
+  // 2. Convert sorted run starts into sorted-position -> unique-index ids.
+  mark_run_starts<<<num_blocks, kThreadsPerBlock, 0, stream_in>>>(
```
manato, line 261: Because input array (`layout.sorted_input`) is already sorted, I believe what you want to do here can be realize by `cub::DeviceAdjacentDifference::SubtractLeft` with the custom difference operator like

``` cuda
struct CustomDifference
{
    template <typename RunIdsType, typename SortedInputType>
    __host__ RunIdsType operator()(SortedInputType &lhs, SortedInputType &rhs)
    {
        return static_cast<RunIdsType>(lhs < rhs);
    }
}
```
ref: https://nvidia.github.io/cccl/unstable/cub/api/structcub_1_1DeviceAdjacentDifference.html#_CPPv4N3cub24DeviceAdjacentDifferenceE

perception/autoware_tensorrt_plugins/src/unique_ops/unique.cu
```diff
@@ -101,77 +101,259 @@
...
+  }
+  if (const auto status = cub::DeviceScan::InclusiveSum(
+        layout.cub_temp_storage, layout.cub_temp_storage_size, layout.run_ids, layout.run_ids,
+        num_input_elements_in, stream_in);
+      status != cudaSuccess) {
+    return status;
+  }
+
+  // 3. Scatter unique ids back to original input order.
+  scatter_inverse_indices<<<num_blocks, kThreadsPerBlock, 0, stream_in>>>(
```
manato, line 274: this can also seem to be replaced using `cub::DeviceFor::ForEach()`. 
ref: https://nvidia.github.io/cccl/unstable/cub/api/structcub_1_1DeviceFor.html#_CPPv4I000_N4cudaSt11enable_if_tIXntN4cudaSt16is_convertible_vI4EnvT12cudaStream_tEEEiEEEN3cub9DeviceFor7ForEachE11cudaError_t21RandomAccessIteratorT21RandomAccessIteratorT3OpT4EnvT

perception/autoware_tensorrt_plugins/src/unique_ops/unique.cu
```diff
@@ -101,77 +101,259 @@
...
+  if (const auto status = cudaPeekAtLastError(); status != cudaSuccess) {
+    return status;
+  }
+  write_unique_counts<<<num_blocks, kThreadsPerBlock, 0, stream_in>>>(
+    unique_offsets_inout, layout.num_unique, layout.unique_offsets_end, unique_counts_out);
+  if (const auto status = cudaPeekAtLastError(); status != cudaSuccess) {
+    return status;
+  }
+
+  write_num_unique_elements<<<1, 1, 0, stream_in>>>(layout.num_unique, num_unique_elements_out);
```
manato, line 289: This can also be replaced by `cub::DeviceAdjacentDifference::SubtractRight`, I believe
ref: https://nvidia.github.io/cccl/unstable/cub/api/structcub_1_1DeviceAdjacentDifference.html#_CPPv4I000EN3cub24DeviceAdjacentDifference13SubtractRightE11cudaError_tPvR6size_t21RandomAccessIteratorT9NumItemsT13DifferenceOpT12cudaStream_t

perception/autoware_tensorrt_plugins/src/unique_ops/unique.cu
mojomex: Thanks, agreed. I moved the duplicated helper into a package-local CUDA helper header and reused it from both argsort and unique in 3ccea2c54.

perception/autoware_tensorrt_plugins/src/unique_ops/unique.cu
mojomex: Thanks. I kept this one as-is because rewriting this ternary as `if`/`else` would not remove the data-dependent branch/divergence; the divergence is inherent in detecting sorted run starts. The current expression keeps the per-element flagging compact.

perception/autoware_tensorrt_plugins/src/unique_ops/unique.cu
mojomex: Thanks, agreed. `fill_iota` is now shared through the package-local kernel helper in 3ccea2c54.

perception/autoware_tensorrt_plugins/src/unique_ops/unique.cu
mojomex: Thanks, fixed in f7b3f87ca. The comment now explains that the inclusive scan produces 1-based run ids, while inverse indices are 0-based.

perception/autoware_tensorrt_plugins/src/unique_ops/unique.cu
mojomex: Thanks, changed in eb1309826. The final-run case now uses an explicit `if`/`else` before writing the count.

perception/autoware_tensorrt_plugins/src/unique_ops/unique.cu
mojomex: Changed to `[[maybe_unused]]` in  3ba7273f1

perception/autoware_tensorrt_plugins/src/argsort_ops/argsort.cu
mojomex: Thanks, good suggestion. Applied in ac99447dc to make the byte-offset arithmetic explicit.

perception/autoware_tensorrt_plugins/src/unique_ops/unique.cu
mojomex: `DeviceAdjacentDifference::SubtractLeft` can't be used as a drop-in replacement because the inputs/outputs are slightly different, but this prompted another CUB-native replacement: `cub::DeviceRunLengthEncode::Encode` now handles the unique-values/counts compaction in 9474b7a18. End-to-end `unique()` microbenchmarking shows the full pipeline is mostly sort-bound, but the new path is generally faster overall.

perception/autoware_tensorrt_plugins/src/unique_ops/unique.cu
mojomex: `cub::DeviceFor`could indeed be used. I kept the custom scatter kernel for now because it is small and removing it doesn't improve performance or maintainability that much. Let me know if you disagree.

perception/autoware_tensorrt_plugins/src/unique_ops/unique.cu
mojomex: Good direction. I did not use `DeviceAdjacentDifference::SubtractRight` directly, but I replaced the count-generation path with `cub::DeviceRunLengthEncode::Encode` in 9474b7a18. That removes `UniqueByKey`, the sentinel write kernel, and the custom count kernel while producing unique values, counts, and `num_unique` directly.

mojomex:
I re-benchmarked after applying the CUB RunLengthEncoding change in 9474b7a18.

Benchmark method: standalone CUDA 12.8 microbenchmark comparing the previous full `unique()` pipeline against the new full `unique()` pipeline. This includes iota fill, radix sort, run-start marking, inclusive scan, inverse-index scatter, unique-values/counts generation, and num-unique output.

Device: NVIDIA RTX PRO 6000 Blackwell, CUDA runtime 12.8.

Results, average CUDA-event time per `unique()` call:

```text
kind              n        old ms   new ms   speedup
high-cardinality  1024     0.0444   0.0396   1.121x
low-cardinality   1024     0.0449   0.0415   1.083x
nearly-unique     1024     0.0458   0.0414   1.105x
clustered         1024     0.0450   0.0405   1.109x
high-cardinality  65536    0.1870   0.1810   1.033x
low-cardinality   65536    0.1806   0.1756   1.028x
nearly-unique     65536    0.1817   0.1789   1.015x
clustered         65536    0.1767   0.1744   1.013x
high-cardinality  1048576  0.4450   0.4308   1.033x
low-cardinality   1048576  0.4266   0.4206   1.014x
nearly-unique     1048576  0.4415   0.4298   1.027x
clustered         1048576  0.4233   0.4267   0.992x
```

Interpretation: the full operation is mostly radix-sort-bound, so end-to-end gains are modest, but the RLE version removes two custom kernels and is neutral-to-faster overall in this benchmark.

Validation also passes:
- `colcon build --packages-select autoware_tensorrt_plugins`
- `reference_kernels_test`

mojomex:
Rebased onto `main` as #12631 caused conflicts.
@manato Could you have one more look regarding my changes (1df4a9d) related to your review? :bow:

manato:
> I re-benchmarked after applying the CUB RunLengthEncoding change in 9474b7a18.
> 
> Benchmark method: standalone CUDA 12.8 microbenchmark comparing the previous full `unique()` pipeline against the new full `unique()` pipeline. This includes iota fill, radix sort, run-start marking, inclusive scan, inverse-index scatter, unique-values/counts generation, and num-unique output.
> 
> Device: NVIDIA RTX PRO 6000 Blackwell, CUDA runtime 12.8.
> 
> Results, average CUDA-event time per `unique()` call:
> 
> ```text
> kind              n        old ms   new ms   speedup
> high-cardinality  1024     0.0444   0.0396   1.121x
> low-cardinality   1024     0.0449   0.0415   1.083x
> nearly-unique     1024     0.0458   0.0414   1.105x
> clustered         1024     0.0450   0.0405   1.109x
> high-cardinality  65536    0.1870   0.1810   1.033x
> low-cardinality   65536    0.1806   0.1756   1.028x
> nearly-unique     65536    0.1817   0.1789   1.015x
> clustered         65536    0.1767   0.1744   1.013x
> high-cardinality  1048576  0.4450   0.4308   1.033x
> low-cardinality   1048576  0.4266   0.4206   1.014x
> nearly-unique     1048576  0.4415   0.4298   1.027x
> clustered         1048576  0.4233   0.4267   0.992x
> ```
> 
> Interpretation: the full operation is mostly radix-sort-bound, so end-to-end gains are modest, but the RLE version removes two custom kernels and is neutral-to-faster overall in this benchmark.
> 
> Validation also passes:
> - `colcon build --packages-select autoware_tensorrt_plugins`
> - `reference_kernels_test`
> 

@mojomex 
Thank you for your detailed investigation. RLE looks smart alternative for the purpose.
In the case that we suffer from radix-sort, we can consider the following for speedup.

In the current code:

``` c++
cudaError_t unique(
  const std::int64_t * input_in, std::int64_t * unique_values_out,
  std::int64_t * inverse_indices_out, std::int64_t * unique_counts_out,
  std::int64_t * num_unique_elements_out, void * workspace_inout, std::size_t num_input_elements_in,
  [[maybe_unused]] std::size_t workspace_size_in, cudaStream_t stream_in)
{
    /*
     * Do some setup
     */

    if (const auto status = cub::DeviceRadixSort::SortPairs(
        layout.cub_temp_storage, layout.cub_temp_storage_size, input_in, layout.sorted_input,
        layout.input_positions, layout.sorted_input_positions, num_input_elements_in, 0, 64,
        stream_in);
    status != cudaSuccess) {
        return status;
    }
```

If (and only if) we can estimate the maximum value of `input_in` (i.e., we can say `max(input_in) < num_input_elements_in`),
we can bound the most-significant bit index needed for key comparison.
Currently, the most-significant bit index is set to `64`, which is full size of int64, but if we can set smaller value to it,
I guess radix-sort gets faster.

i.e.,

``` c++
    int most_significant_bit = std::floor(std::log2(num_input_elements_in));
    if (const auto status = cub::DeviceRadixSort::SortPairs(
        layout.cub_temp_storage, layout.cub_temp_storage_size, input_in, layout.sorted_input,
        layout.input_positions, layout.sorted_input_positions, num_input_elements_in, 0, most_significant_bit,
        stream_in);
    status != cudaSuccess) {
        return status;
    }
```

mojomex:
@manato Really nice optimization idea, let's definitely do it. Is it okay to do in a follow-up PR? This one is already blocking a few other PRs, so I'd like to ship it first :bow:

manato:
@mojomex 
Sure, no problem at all. Let's leave that point for future PR!

manato (approved):
LGTM. Thanks for your tons of contributions!

PR #12555, opened 2026-05-07
perf(autoware_tensorrt_plugins): keep SegmentCSR allocation-free

mojomex:
## Summary

Fourth PR in the SegmentCSR split stack.

After the test, contract-cleanup, and n-dimensional `indptr` removal commits, this final commit keeps `segment_csr_launch` allocation-free by filling the output buffer directly instead of allocating, filling, copying, and freeing a scratch base buffer on every launch.

## Stack

1. #12740: add SegmentCSR kernel tests.
2. #12741: document/clarify the contract and clean up in/out API names.
3. #12739: remove the dead n-dimensional `indptr` path.
4. This PR: keep SegmentCSR allocation-free.

## Benchmark context

This keeps the original #12555 optimization isolated. The prior measurements showed an isolated SegmentCSR-path improvement of about 2.7% on top of #12554 for PTv3-T18, with standalone kernel microbenchmarks showing the per-call allocation removal saving roughly 14 microseconds.


mojomex:
Microbenchmark of `segment_csr_launch` before vs. after this PR.

Method: standalone CUDA 12.8 microbenchmark comparing the previous `segment_csr_launch` from #12554 against the new allocation-free version in this PR. Each variant runs the full launch path: output zero-fill, the segment-reduction kernel (warp-aggregating kernel for `num_cols == 1`, broadcast kernel otherwise), plus the previous version's per-call `cudaMallocAsync`/`cudaMemcpyAsync`/`cudaFreeAsync` of the `base` output scratch and the `indptr_size_dev` shape vector. SUM reduction, `arg_indices_out = nullptr` (the path the plugin actually takes). 200 iterations per variant after 5 warmup iterations, timed via `cudaEventElapsedTime` on a dedicated stream. `indptr` generated as a randomly jittered cumulative sum.

Device: NVIDIA RTX PRO 6000 Blackwell, CUDA 12.8, driver 580.95.05.

Results, average CUDA-event time per `segment_csr_launch` call:

```text
kind         segments   K    avg_seg     old ms     new ms    speedup
small_K=1    1024       1    8           0.0205     0.0062    3.31x
small_K=32   1024       32   8           0.0206     0.0065    3.19x
med_K=1      16384      1    8           0.0212     0.0060    3.54x
med_K=32     16384      32   8           0.0202     0.0103    1.97x
large_K=1    131072     1    8           0.0199     0.0127    1.57x
large_K=32   131072     32   8           0.1234     0.1016    1.21x
```

Interpretation: the new path saves a consistent ~14 µs per call — the cost of the two `cudaMallocAsync`/`cudaFreeAsync` pairs, the H2D copy of the indptr-shape vector, and the D2D copy from `base` to `out`. At small and medium sizes this overhead dominates the kernel work, so the speedup is large (3.2–3.5×). At the largest workload (131k segments × 32 cols) the broadcast kernel dwarfs the overhead and the speedup shrinks to 1.21×, but the new path is strictly faster across every shape tested.

Validation also passes:
- `colcon build --packages-up-to autoware_tensorrt_plugins`
- `reference_kernels_test` (including the new `SegmentCsrMeanMatchesCpuReference` and `SegmentCsrMaxMatchesCpuReference`)

mojomex:
Tested the following:

```bash
source /opt/ros/humble.bash
colcon build --symlink-install --mixin rel-with-deb-info compile-commands --packages-up-to autoware_ptv3
colcon test --packages-select autoware_tensorrt_plugins --event-handlers console_cohesion+
colcon test-result --verbose
```

Then launched ptv3 with a custom rosbag that already contains a concat pointcloud, and visually confirmed correctness.

[segmentcsr.webm](https://github.com/user-attachments/assets/a3d669f9-5c42-43bd-a94b-a1780700a902)

perception/autoware_tensorrt_plugins/src/scatter_ops/segment_csr.cu
```diff
@@ -81,114 +72,72 @@ __global__ void segment_csr_kernel(
...
 
   int thread_idx = blockIdx.x * blockDim.x + threadIdx.x;
-  int row_idx = thread_idx / K;
-  int lane_idx = thread_idx % K;
-  if (thread_idx >= N * K) return;
+  int row_idx = thread_idx / num_cols_in;
+  int lane_idx = thread_idx % num_cols_in;
+  if (thread_idx >= num_segments_in * num_cols_in) return;
 
-  int offset = indptr_to_offset(indptr_size, indptr_dim, row_idx);
```
manato: [ASK]
I confirmed semantically that we can consider `int offset = indptr_to_offset(indptr_size, indptr_dim, row_idx)` is identical to `row_idx` in the case of `segment_csr_kernel` because `row_idx` is guaranteed to be smaller than `num_segments_in`.

On the other hand, in the `segment_csr_broadcast_kernel` case, I'm wondering if we may still need to calculate `offset` because it might not be identical to `row_idx`. Would you ind checking this point? If you already confirmed that they are identical, please le me know!

manato (commented):
@mojomex 
Thank you for your contribution! I left a comment regarding the index calculation where I was not confident (though I might not fully understand the code flow). I'd appreciate it if you could take a look at it :pray:

perception/autoware_tensorrt_plugins/src/scatter_ops/segment_csr.cu
mojomex: Thanks for the comment, and sorry for the long response time. As it turns out, the plugin was written with multi-dimensional `indptr` in mind, but PTv3 is the only model using the plugin (in fact, the plugin was added specifically for PTv3), and only ever uses 1D `indptr`s. I simplified the plugin to the 1D case, making `indptr_to_offset` obsolete.

Would you be more comfortable with me
* removing support for n-dimensional `indptr`
* adding back the `indptr_to_offset` functionality?

For context, I asked Codex about the nD functionality:

> Was nD support already broken? For the TensorRT plugin: **effectively yes**, or at least not part of the public contract. Even before this PR, SegmentCSRPlugin::getOutputShapes() asserted:
> 
>   - src has nbDims == 2
>   - indptr has nbDims == 1
>   - output has nbDims == 2
> 
>   And enqueue() only ever constructed:
> 
>   - src_size = {rows, cols}
>   - indptr_size = {indptr_len}

manato: [Note]
After discussion, we concluded that we omit nD support for the segmentCSR plugin at this moment since nD functionality has already been corrupted before this PR, and (at least, at this moment) no package requires its support

mojomex: (future readers: also see #12739)

manato (approved):
@mojomex
LGTM! Thank you for your series of contributions to improving TRT plugins!

PR #12559, opened 2026-05-08
fix(clang-tidy): re-enable bugprone-unused-return-value

vishalchauhan-hi:
## Description

Re-enables `bugprone-unused-return-value` in `.clang-tidy-ci`.

The check was suppressed in #12431. This PR addresses the 2 reported errors in `system/autoware_system_monitor` listed in #12450.

Refs #12450

## Approach

Both errors came from Boost.Asio calls where the returned `boost::system::error_code` was ignored, even though the code already had an `error_code` variable and checked it immediately afterward.

This PR assigns the returned value back to `error_code`, preserving the existing logic while satisfying `bugprone-unused-return-value`.

Changed calls:

```cpp
error_code = acceptor_->accept(*socket_, error_code);
error_code = socket_->connect(endpoint, error_code);
```

Affected files:

- `system/autoware_system_monitor/reader/traffic_reader/traffic_reader_service.cpp`
- `system/autoware_system_monitor/src/net_monitor/net_monitor.cpp`

## How to reproduce

From the colcon workspace root:

```bash
cd ~/autoware_clang_tidy

mkdir -p /tmp/clang-tidy-result

colcon build --packages-up-to autoware_system_monitor \

run-clang-tidy -p build/ \
  -j $(nproc) autoware_system_monitor \
```

To check the reported errors:

```bash
grep -n "bugprone-unused-return-value" /tmp/clang-tidy-result/report.log | grep "error:" || echo "0 unused-return-value errors"
```

## Verification

Before the fix, the following 2 errors were reproduced:

```text
system/autoware_system_monitor/reader/traffic_reader/traffic_reader_service.cpp:83
system/autoware_system_monitor/src/net_monitor/net_monitor.cpp:667
```

After the fix, `autoware_system_monitor` builds successfully and `run-clang-tidy` reports:

```text
0 unused-return-value errors
```

## Notes

The only `.clang-tidy-ci` change is removing the suppression line for this check:

```diff
-  -bugprone-unused-return-value"
```


system/autoware_system_monitor/reader/traffic_reader/traffic_reader_service.cpp
```diff
@@ -80,7 +80,7 @@ void TrafficReaderService::run()
 
     // Accept a new connection
     boost::system::error_code error_code;
-    acceptor_->accept(*socket_, error_code);
+    error_code = acceptor_->accept(*socket_, error_code);
```
mitsudome-r, line 83: Did you check what was actually returned in this function?
the error_code is already passed as an argument so I was wondering if there are any difference between the return value and error_code given within the function.

vishalchauhan-hi: Thank you for the review Sensei

 Since `error_code` is already passed by reference, the function updates that argument directly. I changed it to assign the return value to the same `error_code` variable only to satisfy `bugprone-unused-return-value`.

So the intention is not to change the behavior, but to preserve the existing error handling logic where `error_code` is checked immediately after the call.

nishikawa-masaki: The specification (error code is returned into the second argument while the same value is returned as the return value) looks strange.
I conducted a quick research about it.
The following is what I learned.
* Traditionally, boost library had been employing this function-style as it was convenient for checking the function error-code on the calling line.
* However, it looks odd. They decided to deprecate the style and replace the return-value type with "void" in the future.
* For the time being, the following macro is defined in boost/asio/detail/config.hpp .
  * Note: BOOST_ASIO_NO_DEPRECATE is also effective on other deprecated specifications.
```
// Helper macros to manage transition away from error_code return values.
#if defined(BOOST_ASIO_NO_DEPRECATED)
# define BOOST_ASIO_SYNC_OP_VOID void
# define BOOST_ASIO_SYNC_OP_VOID_RETURN(e) return
#else // defined(BOOST_ASIO_NO_DEPRECATED)
# define BOOST_ASIO_SYNC_OP_VOID boost::system::error_code
# define BOOST_ASIO_SYNC_OP_VOID_RETURN(e) return e
#endif // defined(BOOST_ASIO_NO_DEPRECATED)
```

Therefore, your fix will be invalidated in the future, when the return-value type becomes "void".

How about the following change?
```
// The error code set to the argument "error_code" is also returned as the return-value.
// The return-value is deprecated and will be removed (replaced by void) in the future.
acceptor_->accept(endpoint, error_code);  // NOLINT(bugprone-unused-return-value)
```

In the future, the comment will become outdated, but there won't be any build error.

system/autoware_system_monitor/src/net_monitor/net_monitor.cpp
```diff
@@ -664,7 +664,7 @@ bool NetMonitor::connect_service()
 
   // Connect socket
   boost::system::error_code error_code;
-  socket_->connect(endpoint, error_code);
+  error_code = socket_->connect(endpoint, error_code);
```
nishikawa-masaki, line 667: How about the following change?
```
// The error code set to the argument "error_code" is also returned as the return-value.
// The return-value is deprecated and will be removed (replaced by void) in the future.
socket_->connect(endpoint, error_code);  // NOLINT(bugprone-unused-return-value)
```

nishikawa-masaki (changes requested):
I'm very sorry for not reviewing this PR for long time.

Your fix is OK for now, but I'm afraid that it will be invalidated in the future.

system/autoware_system_monitor/src/net_monitor/net_monitor.cpp
vishalchauhan-hi: Thank you for the detailed comments. You are right that assigning the deprecated return value would break once Boost.Asio switches to void. I have reverted both calls to the original form and added the comment and NOLINT(bugprone-unused-return-value) as you suggested. One small note: I kept *socket_ as the first argument to accept, since that is what the existing code passes. Also merged main to clear the stale CI failures.

.clang-tidy-ci
```diff
@@ -16,8 +16,7 @@ Checks: "
   -bugprone-signed-char-misuse,
   -bugprone-switch-missing-default-case,
   -bugprone-unchecked-optional-access,
-  -bugprone-unused-local-non-trivial-variable,
-  -bugprone-unused-return-value"
+  -bugprone-unused-local-non-trivial-variable"
```
nishikawa-masaki, line 19: Is this change necessary?
The PR title is "fix(clang-tidy): re-enable bugprone-unused-return-value".
You added "// NOLINT(bugprone-unused-return-value)" comments to re-enable the option, didn't you?

PR #12560, opened 2026-05-08
fix(autoware_multi_object_tracker): tracker update path, shape gating, and association data structure

technolojin:
## Description
                                                                                                                                                                                                                                                        

## Bug Fixes

### 1. `shapes.cpp` — `convertConvexHullToBoundingBox` zeroed every object's height

`convertConvexHullToBoundingBox` was tracking `min_z`/`max_z` over the footprint points and writing `max_z - min_z` into `dimensions.z`.
Footprint points are 2D (`z = 0` always), so this silently set every converted object's height to zero.

**Fix:** Removed the z-axis scan entirely. The output `dimensions.z` is now left untouched so the caller's upstream value is preserved.

### 2. `vehicle_tracker.cpp` — `measureWithPose` gradually shrank tracked vehicle length for cluster inputs

Shape update was gated only on `shape.type == BOUNDING_BOX`. Cluster measurements with `trust_extension = false` are converted to bounding boxes in baselink frame with unreliable dimensions. These were being fed into the EKF length update, causing the tracked vehicle length to drift smaller over time.

**Fix:** Added `can_update_shape = channel_info.trust_extension && is_bbox` guard. When `can_update_shape` is false, shape update is skipped and the EKF length input is taken from `motion_model_.getLength()` (the current state estimate) rather than the noisy measurement.

### 3. `vehicle_tracker.cpp` — `determineUpdateStrategy` alignment threshold was too tight for large vehicles and asymmetric for size-mismatched objects

The alignment threshold was `0.09 * predicted_length` (ratio-only). For a 12 m truck the tolerance was ~1.1 m, which caused spurious `WEAK_UPDATE` fallbacks when the vehicle was decelerating. Additionally, comparing against `predicted_length` alone was asymmetric when measurement and prediction sizes differed.

**Fix:**
- Ratio raised from 9 % to 20 % (`ALIGNMENT_RATIO_THRESHOLD = 0.2`).
- Reference length changed from `predicted_length` to `max(predicted_length, measured_length)` so that size-mismatched cases are handled symmetrically.
- Added a 1.0 m absolute floor (`ALIGNMENT_ABSOLUTE_THRESHOLD = 1.0`) to prevent the threshold from collapsing for small objects.
- Combined formula: `threshold = max(0.2 * max_length, 1.0)`.

### 4. `vehicle_tracker.cpp` — `conditionedUpdate` computed wrong edge gaps for cluster inputs

Cluster bounding boxes arrive in baselink frame (orientation ≈ ego heading). The edge-alignment logic in `determineUpdateStrategy` assumed the measurement is oriented in the tracker's heading frame, so near-face distances were computed incorrectly.

**Fix:** Added `alignClusterToTrackerOrientation()`. Before calling `determineUpdateStrategy`, the cluster's polygon footprint is re-projected onto the tracker's current yaw. The aligned object has its pose and dimensions expressed in the tracker's heading frame, so all edge-center calculations become frame-consistent.

---

## Refactoring

### 1. `UpdatePath` enum + `selectUpdatePath()` virtual dispatch (tracker_base, vehicle_tracker, multiple_vehicle_tracker)

`updateWithMeasurement` in `tracker_base.cpp` previously selected between update strategies with an opaque `if (!has_significant_shape_change)` / `else` tree. The branch structure conflated two independent decisions (shape stability and channel trust), making the logic hard to follow and override in subclasses.

**Change:** Introduced `enum class UpdatePath { NORMAL, TRY_EXTENSION, CONDITIONED }` and a `virtual selectUpdatePath(trust_extension, has_significant_shape_change)` method on `Tracker`. The base implementation returns `NORMAL`. `VehicleTracker` and `MultipleVehicleTracker` override to return `CONDITIONED` when `!trust_extension` (clusters always skip the shape filter) and `TRY_EXTENSION` when `has_significant_shape_change`. The `updateWithMeasurement` body becomes a clean three-branch dispatch.

### 2. `TrackerBevEntry` struct (bev_association.hpp / bev_association.cpp)

`PreparationData` previously held four parallel vectors (`tracked_objects`, `tracker_labels`, `tracker_types`, `tracker_inverse_covariances`). Keeping these in sync was fragile; the inverse-covariance vector was also filled in a second loop over `tracked_objects`.

**Change:** Introduced `struct TrackerBevEntry { object, label, type, inv_cov }` and collapsed `PreparationData` to a single `std::vector<TrackerBevEntry> trackers`. The preparation loop now fills all four fields in one pass (inverse covariance is computed immediately after the object is extracted), eliminating both the second loop and the parallel-index bookkeeping.

### 3. `ScoringResult` struct (scoring_types.hpp, bev_assignment_scoring.hpp/.cpp)

`calculateBevAssignmentScore` returned the score as a `double` and communicated `has_significant_shape_change` via a non-const reference output parameter — an error-prone calling convention.

**Change:** Added a new header `scoring_types.hpp` with `struct ScoringResult { double score; bool has_significant_shape_change; }`. `calculateBevAssignmentScore` now returns `ScoringResult` by value. All early-return gates return `{INVALID_SCORE, false}`. Call sites read `result.score` and `result.has_significant_shape_change` directly.

---

## Files Changed

| File | Change |
|------|--------|
| `lib/object_model/shapes.cpp` | Bug fix: remove spurious z-dim computation in `convertConvexHullToBoundingBox` |
| `lib/tracker/model/vehicle_tracker.cpp` | Bug fix: `can_update_shape` guard, `alignClusterToTrackerOrientation`, updated alignment thresholds |
| `lib/tracker/model/tracker_base.cpp` | Refactor: `updateWithMeasurement` → named `UpdatePath` dispatch |
| `lib/association/bev_association.cpp` | Refactor: `PreparationData` → single-pass `TrackerBevEntry` vector |
| `lib/association/scoring/bev_assignment_scoring.cpp` | Refactor: return `ScoringResult` instead of `bool &` out-param |
| `include/.../tracker_base.hpp` | Add `UpdatePath` enum, `selectUpdatePath` virtual |
| `include/.../vehicle_tracker.hpp` | Add `selectUpdatePath` override, updated thresholds, `alignClusterToTrackerOrientation` |
| `include/.../multiple_vehicle_tracker.hpp` | Add `selectUpdatePath` override |
| `include/.../bev_association.hpp` | Add `TrackerBevEntry`; collapse `PreparationData` |
| `include/.../bev_assignment_scoring.hpp` | Change return type to `ScoringResult` |
| `include/.../scoring_types.hpp` | New file: `ScoringResult` struct |

---

## Related links

https://github.com/autowarefoundation/autoware_universe/pull/12475 will base this PR

**Parent Issue:**

- Link

<!-- ⬇️🟢
**Private Links:**

- [CompanyName internal link]()
⬆️🟢 -->

## How was this PR tested?

 using modified perception pipeline https://github.com/technolojin/autoware_launch/tree/feat/mot/association-polar to input polygon shape detections. 

## Notes for reviewers

None.

## Interface changes

None.

<!-- ⬇️🔴

### Topic changes

#### Additions and removals

#### Modifications

### ROS Parameter Changes

#### Additions and removals

#### Modifications

| Old     | `old_param_name` | `double` | `1.0`         | Param description |
| New     | `new_param_name` | `double` | `1.0`         | Param description |

🔴⬆️ -->

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_multi_object_tracker/lib/tracker/model/vehicle_tracker.cpp
+++ b/perception/autoware_multi_object_tracker/lib/tracker/model/vehicle_tracker.cpp
@@ -307,11 +307,16 @@ bool VehicleTracker::getTrackedObject(
   return true;
 }
 
-types::DynamicObject VehicleTracker::alignClusterToTrackerOrientation(
+std::optional<types::DynamicObject> VehicleTracker::alignClusterToTrackerOrientation(
   const types::DynamicObject & cluster, const double tracker_yaw) const
 {
+  // Only re-project polygon (cluster) measurements; bounding boxes are already axis-aligned.
+  const bool is_cluster =
+    cluster.shape.type == autoware_perception_msgs::msg::Shape::POLYGON &&
+    !cluster.shape.footprint.points.empty();
+  if (!is_cluster) return std::nullopt;
+
   const auto & pts = cluster.shape.footprint.points;
-  if (pts.empty()) return cluster;
 
   // footprint.points are in the cluster's local frame (baselink orientation).
   // Transform each point to a map-relative offset, then project onto the tracker's axes.
@@ -361,16 +366,14 @@ bool VehicleTracker::conditionedUpdate(
   const autoware_perception_msgs::msg::Shape & tracker_shape, const rclcpp::Time & measurement_time,
   const types::InputChannel & channel_info)
 {
-  // For cluster measurements, the bounding box orientation is in baselink frame.
-  // Re-project the polygon footprint onto the tracker's current heading so that
-  // determineUpdateStrategy receives correctly-oriented edge centers.
-  const types::DynamicObject & meas_for_strategy =
-    !measurement.shape.footprint.points.empty()
-      ? alignClusterToTrackerOrientation(measurement, motion_model_.getYawState())
-      : measurement;
+  // For cluster measurements, re-project the footprint onto the tracker heading before strategy
+  // selection. nullopt is returned when there are no footprint points.
+  const auto aligned =
+    alignClusterToTrackerOrientation(measurement, motion_model_.getYawState());
+  const types::DynamicObject & meas = aligned ? *aligned : measurement;
 
   // Determine update strategy
-  UpdateStrategy strategy = determineUpdateStrategy(meas_for_strategy, prediction);
+  UpdateStrategy strategy = determineUpdateStrategy(meas, prediction);
 
   // Handle weak update strategy (no edge alignment - use weak update with pseudo measurement)
   if (strategy.type == UpdateStrategyType::WEAK_UPDATE) {
--- a/perception/autoware_multi_object_tracker/lib/tracker/model/vehicle_tracker.cpp
+++ b/perception/autoware_multi_object_tracker/lib/tracker/model/vehicle_tracker.cpp
@@ -139,15 +139,9 @@ bool VehicleTracker::measureWithPose(
 {
   // Shape update is only valid when the channel guarantees reliable size information
   // AND the measurement is a bounding box.
-  // - Polygon (UNKNOWN-label cluster): shape.type != BOUNDING_BOX, dims = (0,0)
-  // - Known-label cluster converted to bbox: trust_extension=false (baselink-frame bbox,
-  // unreliable)
+  // Use the measurement length only when its shape is trustworthy.
   const bool is_bbox = (object.shape.type == autoware_perception_msgs::msg::Shape::BOUNDING_BOX);
   const bool can_update_shape = channel_info.trust_extension && is_bbox;
-
-  // Use the measurement length only when its shape is trustworthy.
-  // Untrusted inputs (polygon or trust_extension=false) have unreliable or zero dimensions;
-  // using them would corrupt the EKF wheel-position states and gradually shrink tracked length.
   constexpr double min_length = 1.0;
   const double length = can_update_shape ? std::max(object.shape.dimensions.x, min_length)
                                          : std::max(motion_model_.getLength(), min_length);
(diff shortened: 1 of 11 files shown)


perception/autoware_multi_object_tracker/lib/tracker/model/vehicle_tracker.cpp
```diff
@@ -310,13 +307,70 @@ bool VehicleTracker::getTrackedObject(
...
   const types::DynamicObject & measurement, const types::DynamicObject & prediction,
   const autoware_perception_msgs::msg::Shape & tracker_shape, const rclcpp::Time & measurement_time,
   const types::InputChannel & channel_info)
 {
+  // For cluster measurements, the bounding box orientation is in baselink frame.
+  // Re-project the polygon footprint onto the tracker's current heading so that
+  // determineUpdateStrategy receives correctly-oriented edge centers.
+  const types::DynamicObject & meas_for_strategy =
+    !measurement.shape.footprint.points.empty()
+      ? alignClusterToTrackerOrientation(measurement, motion_model_.getYawState())
```
YoshiRi: [very minor comment]
The ternary operator here will cause an unintended copy of measurement even when the condition is false, because C++ forces the result type to match the return-by-value of alignClusterToTrackerOrientation. Consider using a pointer or a standard if-else block to avoid this overhead.

YoshiRi (approved):
SRY for late reply. But LGTM.

perception/autoware_multi_object_tracker/lib/tracker/model/vehicle_tracker.cpp
technolojin: fixed https://github.com/autowarefoundation/autoware_universe/pull/12560/changes/e92415cbd2a4c40ef3b3601e50c540f7ac048b10

PR #12561, opened 2026-05-08
test(autoware_tensorrt_plugins): add reference kernel tests

mojomex:
## Summary

Adds CUDA kernel vs. known-good CPU reference tests for `autoware_tensorrt_plugins`:

- verifies `argsort` output against a CPU stable-sort reference
- verifies `unique` values, inverse indices, and counts against a CPU reference

## Compatibility note

While validating the test, the `unique` counts check exposed an existing sentinel write bug, so this PR includes the minimal fix needed for the new test to pass: write the final range sentinel to `range_ptr + num_out`, not `range_ptr + num_out * sizeof(int64_t)` (`range_ptr` is an `int64_t*`).

See 

- https://github.com/autowarefoundation/autoware_universe/pull/12502

for a standalone version of that fix.

## Validation

```shell
colcon build --packages-up-to autoware_tensorrt_plugins && \
colcon test-result --delete-yes &&
colcon test --packages-select autoware_tensorrt_plugins --event-handlers console_cohesion+ && \
colcon test-result --verbose
```

Look out for:
```
[==========] Running 2 tests from 1 test suite.
1: [----------] Global test environment set-up.
1: [----------] 2 tests from ReferenceKernelsTest
1: [ RUN      ] ReferenceKernelsTest.ArgsortMatchesCpuReference
1: [       OK ] ReferenceKernelsTest.ArgsortMatchesCpuReference (277 ms)
1: [ RUN      ] ReferenceKernelsTest.UniqueMatchesCpuReference
1: [       OK ] ReferenceKernelsTest.UniqueMatchesCpuReference (1 ms)
1: [----------] 2 tests from ReferenceKernelsTest (278 ms total)
1: 
1: [----------] Global test environment tear-down
1: [==========] 2 tests from 1 test suite ran. (278 ms total)
1: [  PASSED  ] 2 tests.
```


perception/autoware_tensorrt_plugins/test/argsort_kernel_test.cpp
```diff
@@ -0,0 +1,82 @@
...
+  DeviceBuffer<std::int64_t> output_d(input.size());
+  DeviceBuffer<std::uint8_t> workspace_d(get_argsort_total_workspace_size(input.size()));
+
+  copy_to_device(input_d.get(), input);
+
+  ASSERT_EQ(
+    argsort(
+      input_d.get(), output_d.get(), workspace_d.get(), input.size(),
+      get_argsort_workspace_size(input.size()), stream.get()),
+    cudaSuccess);
```
manato, line 76:
```suggestion
  cudaEvent_t copy_event;
  cudaEventCreate(&copy_event);

  copy_to_device(input_d.get(), input);

  cudaEventRecord(copy_event, 0);  // record event on the default stream
  cudaStreamWaitEvent(stream.get(), copy_event, cudaEventWaitDefault);

  ASSERT_EQ(
    argsort(
      input_d.get(), output_d.get(), workspace_d.get(), input.size(),
      get_argsort_workspace_size(input.size()), stream.get()),
    cudaSuccess);
```

According to the [memcpy synchronous behavior](https://docs.nvidia.com/cuda/cuda-runtime-api/api-sync-behavior.html#api-sync-behavior):

> For transfers from pageable host memory to device memory, a stream sync is performed before the copy is initiated. The function will return once the pageable buffer has been copied to the staging memory for DMA transfer to device memory, but the DMA to final destination may not have completed.

Since `input` is pageable host memory, it would be better to insert explicit synchronization before operating on `input_d`.

Alternatively, we can skip this kind of explicit synchronization if `copy_to_device` takes a CUDA stream as an argument and performs `cudaMemcpyAsync` inside.

perception/autoware_tensorrt_plugins/test/unique_kernel_test.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+  CudaStreamGuard stream;
+  DeviceBuffer<std::int64_t> input_d(input.size());
+  DeviceBuffer<std::int64_t> unique_d(input.size());
+  DeviceBuffer<std::int64_t> inverse_d(input.size());
+  DeviceBuffer<std::int64_t> counts_d(input.size());
+  DeviceBuffer<std::uint8_t> workspace_d(get_unique_workspace_size(input.size()));
+
+  copy_to_device(input_d.get(), input);
+
```
manato, line 84: same as the the case of `argsort-_kernel_test.cpp`. better to insert explicit sync

manato (commented):
@mojomex 
Thank you very much for improving the unit tests. In terms of test strictness, I left small suggestions. I would appreciate it if you consider them!

perception/autoware_tensorrt_plugins/test/argsort_kernel_test.cpp
mojomex: > a stream sync is performed before the copy is initiated
That means that `cudaMemcpy` _already_ syncs afaik, so we are already fine, right? Please correct me if I'm wrong.

perception/autoware_tensorrt_plugins/test/unique_kernel_test.cpp
mojomex: See above reply.

perception/autoware_tensorrt_plugins/test/argsort_kernel_test.cpp
manato: You are right. I only focused on the implicit sync by cudaMemcpy and missed the [legacy CUDA stream behavior](https://docs.nvidia.com/cuda/cuda-runtime-api/stream-sync-behavior.html#stream-sync-behavior).
Don't worry about my previous comment :pray:

manato (approved):
LGTM! Thank you for contributing to test improvement!

ktro2828 (approved):

PR #12562, opened 2026-05-08
fix: imu scale estimation covariance limit for stability

SergioReyesSan:
## Description

This PR fixes an issue in the IMU scale estimation where the covariance matrix (specifically non-diagonal terms) would increase significantly when the vehicle is stationary or moving straight/ slow turning for extended periods. This growth caused instability in scale estimation and triggered MRM.

**Parent Issue:**

- Link

## How was this PR tested?

The fix was validated using Rosbags directly from the vehicle, with prolonged, slow turns causing no updates to the imu scale estimate. Furthermore, a Rosbag containing an NDT pose anomaly (which previously caused system instability due to high accumulated covariance) was tested; the system now remains stable after the fix.

decay coefficient and threshold to estimate scale parameters, also changed to increase scale estimation stability.

No needed changes on this version of Autoware launch, since it uses the default values.
## Related links

TIER IV INTERNAL test.
https://star4.slack.com/archives/C09J2QMQZ1B/p1779250307217069?thread_ts=1777543404.131559&cid=C09J2QMQZ1B

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


sensing/autoware_imu_corrector/src/gyro_bias_estimator.cpp
YoshiRi: How about covariance capping instead of clamping?

```c++
// Update P with F
ekf_angle_.p_angle_ = f_matrix * ekf_angle_.p_angle_ * f_matrix.transpose();

// Covariance Capping: Only add Q if we are below the maximum variance threshold
if (ekf_angle_.p_angle_(0, 0) < ekf_angle_.max_variance_p_angle_ && 
    ekf_angle_.p_angle_(1, 1) < ekf_angle_.max_variance_p_angle_) {
    ekf_angle_.p_angle_ += ekf_angle_.q_angle_;
}
```

sensing/autoware_imu_corrector/src/gyro_bias_estimator.cpp
```diff
@@ -278,6 +278,12 @@ void GyroBiasEstimator::callback_imu(const sensor_msgs::msg::Imu::ConstSharedPtr
...
       ekf_angle_.max_variance_p_angle_);
+    ekf_angle_.p_angle_(0, 1) = std::min(
+      std::max(ekf_angle_.p_angle_(0, 1), -ekf_angle_.max_variance_p_angle_),
+      ekf_angle_.max_variance_p_angle_);
+    ekf_angle_.p_angle_(1, 0) = std::min(
+      std::max(ekf_angle_.p_angle_(1, 0), -ekf_angle_.max_variance_p_angle_),
+      ekf_angle_.max_variance_p_angle_);
     ekf_angle_.p_angle_(1, 1) = std::min(
       std::max(ekf_angle_.p_angle_(1, 1), ekf_angle_.min_covariance_angle_),
       ekf_angle_.max_variance_p_angle_);
```
YoshiRi: I want to discard this clamping and switch to Q skipping because clamping will distort EKF numerical stability.

Do you have any other motivation to keep this part?

sensing/autoware_imu_corrector/src/gyro_bias_estimator.cpp
YoshiRi: [Question]
Setting this to exactly 0 implies we assume zero angle random walk from the gyro, which physically causes the angle variance to collapse over time, making the lower-bound clamping (min_covariance_angle_) strictly necessary to keep the filter alive.
It sounds safer to set small value to first element like 1e-8.

Do you know any intention to this implementation?

YoshiRi (changes requested):
Thanks for tackling this issue! Adding a safety valve to prevent covariance wind-up during long straight/stationary periods is definitely the right direction to stabilize the system.
However, from an EKF design perspective, I have a critical concern regarding the current implementation approach. 

## Concern: Breaking Positive Semi-Definiteness (PSD)
Clamping the off-diagonal elements `(p_angle_(0, 1)` and `p_angle_(1, 0))` independently using `std::min/max` will likely destroy the **Positive Semi-Definite (PSD) property of the covariance matrix** .In a 2x2 covariance matrix, the elements must always satisfy $P_{01}^2 \le P_{00} P_{11}$. This mathematical distortion can lead to singular matrices or NaNs during the Kalman gain inversion as soon as the vehicle starts turning again, potentially causing a new critical crash.

## Proposed Solution: Covariance Capping ($Q = 0$)
Instead of distorting the matrix elements after the prediction step, I prefer the approach to stop adding the process noise ($Q$) when the covariance reaches its upper limit.

sensing/autoware_imu_corrector/src/gyro_bias_estimator.cpp
YoshiRi, line 278: [Bugs?] 
This may be out of scope, but could you check this f_matrix is right?

```suggestion
    f_matrix << 1, dt_imu * gyro.vector.z, 0,
```

IMO we do not need bias value.

$$\begin{bmatrix} \theta_{k} \\\ s_{k} \end{bmatrix} = f(X_{k-1}, u_k) = \begin{bmatrix} \theta_{k-1} + (s_{k-1} \cdot \omega_k - b) \cdot \Delta t \\\ s_{k-1} \cdot \lambda \end{bmatrix}$$

$$F = \frac{\partial f}{\partial X} = \begin{bmatrix} \frac{\partial \theta_k}{\partial \theta_{k-1}} & \frac{\partial \theta_k}{\partial s_{k-1}} \\\ \frac{\partial s_k}{\partial \theta_{k-1}} & \frac{\partial s_k}{\partial s_{k-1}} \end{bmatrix} = \begin{bmatrix} 1 & \omega_k \cdot \Delta t \\\ 0 & \lambda \end{bmatrix}$$

SergioReyesSan: It should work as well, i was thinking to update also only when it's possible to update the filter.

sensing/autoware_imu_corrector/src/gyro_bias_estimator.cpp
SergioReyesSan: i guess that if we stop updating the covariance after a certain boundary, as you mentioned, it might not be necessary this part.

sensing/autoware_imu_corrector/src/gyro_bias_estimator.cpp
SergioReyesSan: Thanks for noticing, I added the bias because when I injected a large artificial bias i noticed an error in the scale estimation. The input of this node is raw IMU data; it might contain a large offset that changes over the time.

SergioReyesSan: Thanks for the comment. When starting the development, I started with simulations and "ideal" conditions, so you're right is better to assume a bit of noise on the process element

YoshiRi: @SergioReyesSan  Could you update the logic and let me know after your commit?

SergioReyesSan: changed here

 [2b18f4c](https://github.com/autowarefoundation/autoware_universe/pull/12562/commits/2b18f4c562ce5f917ec1eb1a89a2362c71e2b2dd)

sensing/autoware_imu_corrector/src/gyro_bias_estimator.cpp
SergioReyesSan: changed here

[2b18f4c](https://github.com/autowarefoundation/autoware_universe/pull/12562/commits/2b18f4c562ce5f917ec1eb1a89a2362c71e2b2dd)

sensing/autoware_imu_corrector/src/gyro_bias_estimator.cpp
SergioReyesSan: changed here

[2b18f4c](https://github.com/autowarefoundation/autoware_universe/pull/12562/commits/2b18f4c562ce5f917ec1eb1a89a2362c71e2b2dd)

SergioReyesSan:
@YoshiRi Thanks for the reminder i checked that is workign with Hihara-San rosbags, added the latest commit with some changes

YoshiRi (changes requested):
This is a brilliant update. I've reviewed Patch 2, and it perfectly addresses all the EKF design concerns we discussed.LGM

YoshiRi (approved):

SergioReyesSan:
@YoshiRi Thanks for your comments and suggestions to make it better1

PR #12565, opened 2026-05-11
refactor(autoware_traffic_light_map_based_detector): simplify detect() and unify ROI computation

interimadd:
## Description

This PR is a behavior-preserving refactor of `TrafficLightMapBasedDetector` that focuses on clarifying the detection pipeline used by `detect()`. No functional change is intended.

Changes:

- Split the previous overload of `get_traffic_light_roi()` into:
  - `project_traffic_light_to_roi()` — single transform → `std::optional<TrafficLightRoi>`.
  - `get_traffic_light_roi()` — `std::vector<StampedTransform>` → `std::optional<TrafficLightRoi>` for the bounding ROI; the single-transform case is now handled as a one-element vector.
- Changed `get_visible_traffic_lights()` to return `std::vector<lanelet::ConstLineString3d>` by value instead of writing through an output parameter.
- Reduced `detect()` to orchestration logic by extracting small helpers for traffic-light-set selection and expect-ROI config construction

## Related links

**Parent Issue:**

- N/A (internal refactor)

## How was this PR tested?

- `colcon build --packages-select autoware_traffic_light_map_based_detector`
- `colcon test --packages-select autoware_traffic_light_map_based_detector --event-handlers console_cohesion+` — all tests pass (gtest, copyright, cppcheck, lint_cmake, xmllint, etc.).

## Notes for reviewers

None

## Interface changes

None

## Effects on system behavior

None. This change is intended to be behavior-preserving and is covered by the existing test suite.

### DIFF
--- a/perception/autoware_traffic_light_map_based_detector/src/traffic_light_map_based_detector.cpp
+++ b/perception/autoware_traffic_light_map_based_detector/src/traffic_light_map_based_detector.cpp
@@ -16,7 +16,7 @@
 
 #include <autoware/lanelet2_utils/conversion.hpp>
 #include <autoware/traffic_light_utils/traffic_light_utils.hpp>
-#include <autoware_lanelet2_extension/visualization/visualization.hpp>
+#include <autoware_lanelet2_extension/utility/query.hpp>
 #include <autoware_utils/math/normalization.hpp>
 #include <autoware_utils/math/unit_conversion.hpp>
 
--- a/perception/autoware_traffic_light_map_based_detector/src/traffic_light_map_based_detector.cpp
+++ b/perception/autoware_traffic_light_map_based_detector/src/traffic_light_map_based_detector.cpp
@@ -199,10 +199,8 @@ DetectionResult TrafficLightMapBasedDetector::detect(
   image_geometry::PinholeCameraModel pinhole_camera_model;
   pinhole_camera_model.fromCameraInfo(camera_info);
 
-  std::vector<lanelet::ConstLineString3d> visible_traffic_lights;
-  get_visible_traffic_lights(
-    select_target_traffic_lights(), tf_map2camera_samples, pinhole_camera_model,
-    visible_traffic_lights);
+  const auto visible_traffic_lights = get_visible_traffic_lights(
+    select_target_traffic_lights(), tf_map2camera_samples, pinhole_camera_model);
 
   const tf2::Transform tf_map2camera_closest =
     find_closest_transform(tf_map2camera_samples, camera_info.header.stamp);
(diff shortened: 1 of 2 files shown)


perception/autoware_traffic_light_map_based_detector/src/traffic_light_map_based_detector.cpp
```diff
@@ -178,55 +200,40 @@ DetectionResult TrafficLightMapBasedDetector::detect(
...
-    get_visible_traffic_lights(
-      *route_traffic_lights_ptr_, tf_map2camera_samples, pinhole_camera_model,
-      visible_traffic_lights);
-  } else {
-    get_visible_traffic_lights(
-      *all_traffic_lights_ptr_, tf_map2camera_samples, pinhole_camera_model,
-      visible_traffic_lights);
-  }
+  const auto visible_traffic_lights = get_visible_traffic_lights(
+    select_target_traffic_lights(), tf_map2camera_samples, pinhole_camera_model);
```
a-maumau, line 204: [nit]
For me, if we do not need to explicitly show which traffic lights to be used, we could move the `select_target_traffic_lights` into `get_visible_traffic_lights`.

perception/autoware_traffic_light_map_based_detector/src/traffic_light_map_based_detector.cpp
```diff
@@ -178,55 +200,40 @@ DetectionResult TrafficLightMapBasedDetector::detect(
...
-  expect_roi_config.max_vibration_pitch = 0;
+  const auto expect_roi_config = make_expect_roi_config(config_);
 
   for (const auto & traffic_light : visible_traffic_lights) {
-    tier4_perception_msgs::msg::TrafficLightRoi rough_roi, expect_roi;
-    if (!get_traffic_light_roi(
-          tf_map2camera_closest, pinhole_camera_model, traffic_light, expect_roi_config,
-          expect_roi)) {
+    const auto expect_roi = get_traffic_light_roi(
+      {tf_map2camera_closest}, pinhole_camera_model, traffic_light, expect_roi_config);
```
a-maumau, line 212: Just for note:
- The result should be the same, but the execution path is slightly changed. it will run `compute_bounding_roi` which should do nothing.

perception/autoware_traffic_light_map_based_detector/src/traffic_light_map_based_detector.cpp
```diff
@@ -278,14 +285,17 @@ void TrafficLightMapBasedDetector::get_visible_traffic_lights(
       break;
     }
   }
+  return visible_traffic_lights;
 }
 
-bool TrafficLightMapBasedDetector::get_traffic_light_roi(
+std::optional<tier4_perception_msgs::msg::TrafficLightRoi>
+TrafficLightMapBasedDetector::project_traffic_light_to_roi(
```
a-maumau, line 292: Ignore this.

The name `project_traffic_light_to_roi` sounds not right to me (it is extracting/creating a ROI from a image by projecting the points of traffic light), but when I asked claude about this, claude says "it should be OK".
So, I will leave this as OK (I also couldn't come up a better name).

a-maumau (approved):
overall LGTM!

perception/autoware_traffic_light_map_based_detector/src/traffic_light_map_based_detector.cpp
interimadd: In my opinion, the current design is more readable because `get_visible_traffic_lights()` takes `TrafficLightSet` as an explicit parameter. It makes clear what data the function operates on.
In addition, since `detect()` is the main orchestration function of this core logic class, I prefer to keep the selection between map-wide and route-restricted traffic lights visible in `detect()`, rather than hiding it inside `get_visible_traffic_lights()`.

interimadd:
@a-maumau 
Thank you for the review and comments!
There is no single correct answer for the implementation, but for the reasons I have provided, I would like to merge the PR as is.

PR #12566, opened 2026-05-11
fix(traffic_light_multi_camera_fusion): use minimum confidence across signal elements

a-maumau:
## Description
This PR will change the confident update from taking the first element to minimum of the confidence.

In the https://github.com/autowarefoundation/autoware_universe/pull/12302 , it introduced a new model that detect the traffic light bulbs. This model will output confidence of each traffic light bulb, so we need to handle this in some plausible manner (in this PR, taking the min).

This PR will not change the behavior of existing method (model) since it was outputting the same confidence among the bulbs in the same traffic light.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


technolojin (approved):
LGTM

PR #12568, opened 2026-05-11
test(autoware_trajectory_modifier): add integration test for obstacle_stop plugin

interimadd:
## Description

Add a node-level integration test for the `ObstacleStop` plugin in `autoware_trajectory_modifier`.

The test exercises the plugin through its public `modify_trajectory` interface together with the real `TrajectoryModifierContext`, so the obstacle-tracking, RSS-based collision detection, and pointcloud clustering code paths are wired up exactly as they are in production. It is intended as a **characterization test** that pins down the current behavior of the obstacle stop pipeline before subsequent refactoring.

### What is covered

| Test case | Scenario |
|---|---|
| `TrajectoryNotModifiedWhenDisabled` | Plugin is disabled via `params_.use_obstacle_stop = false` |
| `TrajectoryNotModifiedForEmptyTrajectory` | Empty input trajectory short-circuits |
| `TrajectoryNotModifiedWhenNoObstaclesDetected` | No predicted objects and no pointcloud |
| `TrajectoryNotModifiedWhenObjectIsBesidePath` | Object lateral to the trajectory polygon |
| `TrajectoryModifiedWhenObjectBlocksPath` | Stopped vehicle directly in front of the ego trajectory |
| `StopPointInsertedBeforeObject` | Verifies the stop pose is inserted upstream of the blocking object and final velocity is reduced |
| `StopPointInsertedForBlockingPointcloudCluster` | Same verification for a synthesized `PointCloud2` cluster |

The obstacle tracker requires `on_time_buffer` of continuous observation before promoting a detection to "active", so each obstacle-detection test invokes `modify_trajectory` twice with a short sleep in between to satisfy that gate.

### Coverage

After adding this test, line coverage of the obstacle_stop sources is:

| File | Line coverage | Function coverage |
|---|---|---|
| `src/trajectory_modifier_plugins/obstacle_stop.cpp` | 75.7% | 69.0% |
| `src/trajectory_modifier_utils/obstacle_stop_utils.cpp` | 79.2% | 76.2% |
| **Total** | **77.4% (598 / 773)** | **72.0% (36 / 50)** |

## Related links

**Parent Issue:**

## How was this PR tested?

- `colcon build --packages-select autoware_trajectory_modifier`
- `colcon test --packages-select autoware_trajectory_modifier --event-handlers console_cohesion+` — all 5 tests pass (gtest + 4 linters).
- `lcov` capture against the built package confirmed the coverage figures above.

## Notes for reviewers

None.
## Interface changes

None.

## Effects on system behavior

None. Test-only change.

### DIFF
--- a/planning/autoware_trajectory_modifier/tests/test_obstacle_stop_integration.cpp
+++ b/planning/autoware_trajectory_modifier/tests/test_obstacle_stop_integration.cpp
@@ -1,0 +1,382 @@
+// Copyright 2026 TIER IV, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "autoware/trajectory_modifier/trajectory_modifier_plugins/obstacle_stop.hpp"
+
+#include <ament_index_cpp/get_package_share_directory.hpp>
+#include <autoware_test_utils/autoware_test_utils.hpp>
+#include <autoware_trajectory_modifier/trajectory_modifier_param.hpp>
+#include <rclcpp/rclcpp.hpp>
+
+#include <autoware_perception_msgs/msg/object_classification.hpp>
+#include <autoware_perception_msgs/msg/predicted_objects.hpp>
+#include <autoware_perception_msgs/msg/shape.hpp>
+#include <geometry_msgs/msg/accel_with_covariance_stamped.hpp>
+#include <nav_msgs/msg/odometry.hpp>
+#include <sensor_msgs/msg/point_cloud2.hpp>
+#include <sensor_msgs/point_cloud2_iterator.hpp>
+
+#include <gtest/gtest.h>
+
+#include <chrono>
+#include <memory>
+#include <string>
+#include <thread>
+#include <utility>
+#include <vector>
+
+namespace
+{
+using autoware::trajectory_modifier::TrajectoryModifierContext;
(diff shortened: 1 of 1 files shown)


planning/autoware_trajectory_modifier/tests/test_obstacle_stop_integration.cpp
```diff
@@ -0,0 +1,382 @@
...
+  //      before becoming active
+  plugin_->modify_trajectory(trajectory, input);
+  std::this_thread::sleep_for(std::chrono::milliseconds(50));
+  const bool modified = plugin_->modify_trajectory(trajectory, input);
+
+  // Assert
+  ASSERT_TRUE(modified);
+  EXPECT_LT(
+    trajectory.back().longitudinal_velocity_mps, trajectory.front().longitudinal_velocity_mps);
+  EXPECT_LT(trajectory.back().pose.position.x, object_x);
```
mkquda, line 359: (**NOT A MUST**) Can also verify distance between `trajectory.back().pose.position.x` and `object_x` is equal to `stop margin + ego_front_offset + object_size_x / 2`

mkquda (approved):
LGTM.

go-sakayori (approved):

planning/autoware_trajectory_modifier/tests/test_obstacle_stop_integration.cpp
interimadd: Thanks for the suggestion. It looks like `trajectory.back().pose.position.x` is being clamped by `min_stopping_distance`, which means it won't necessarily match `stop_margin + ego_front_offset + object_size_x / 2`.
To avoid test fragility when the logic changes, I’d prefer to keep the test conditions as minimal as possible.

interimadd:
Thank you for the review!
As @mkquda pointed out, it is better to make the test conditions more specific. Therefore, I have added this commit: https://github.com/autowarefoundation/autoware_universe/pull/12568/changes/a5520ce8d5e702778026d8239bbfd02274a0e9cf

While updating the test case, I noticed that `trajectory.back().longitudinal_velocity_mps` is not exactly 0, contrary to my expectations. It seems to be due to the interpolation in the calculation logic.
I believe it should be fixed in the other PR.

I will merge this PR once CI passes.

PR #12569, opened 2026-05-11
fix(aeb, shape_estimation): guard pcl::fromROSMsg against empty point clouds

mertyavuz41:
## Description
While executing planning simulator, the some modules has console spam messages that indicates the point cloud message is empty.

## Related links

**Parent Issue:**

- Link: https://github.com/autowarefoundation/autoware/issues/6843#issuecomment-4213412231

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


control/autoware_autonomous_emergency_braking/src/node.cpp
xmfcx, line 341: If you return early, you will have missed setting this class member variable. 🎗️
This should also be set as empty above.

xmfcx (changes requested):

xmfcx (approved):
Thanks for the iteration @mertyavuz41 — setting `obstacle_ros_pointcloud_ptr_` to an empty cloud with the header before the early return addresses my earlier concern. Approving.

PR #12570, opened 2026-05-12
fix(clang-tidy): re-enable clang-diagnostic-unused-private-field

vish0012:
Re-enables `clang-diagnostic-unused-private-field` by removing `-Wno-unused-private-field` from `.clang-tidy-ci`.

The check was suppressed in #12431 with 2 reported errors in `perception/autoware_lidar_apollo_instance_segmentation`. This PR fixes both.

Refs #12450

### `autoware_lidar_apollo_instance_segmentation`

`FeatureGenerator` stored the constructor arguments `use_intensity_feature` and `use_constant_feature` in private member variables, but those member variables were never read after construction.

The constructor already uses the parameters directly to select the correct `FeatureMap` implementation, so this PR removes the unused private fields and their initializer-list entries without changing behavior.

Affected files:

- `perception/autoware_lidar_apollo_instance_segmentation/include/autoware/lidar_apollo_instance_segmentation/feature_generator.hpp`
- `perception/autoware_lidar_apollo_instance_segmentation/src/feature_generator.cpp`

## Verification

```bash
pre-commit run clang-format --files \
  perception/autoware_lidar_apollo_instance_segmentation/include/autoware/lidar_apollo_instance_segmentation/feature_generator.hpp \
  perception/autoware_lidar_apollo_instance_segmentation/src/feature_generator.cpp
```

Result:

```text
clang-format.............................................................Passed
```

```bash
colcon build --packages-up-to autoware_lidar_apollo_instance_segmentation \
```

Result:

```text
Summary: 27 packages finished
```

```bash
run-clang-tidy -p build/ \
  -j $(nproc) \
  autoware_lidar_apollo_instance_segmentation \

grep -n "unused-private-field" /tmp/clang-tidy-result/report.log || echo "0 unused-private-field errors"
```

Result:

```text
0 unused-private-field errors
```


YoshiRi (approved):
Thanks for your work! LGTM

.clang-tidy-ci
```diff
@@ -28,6 +28,6 @@ ExtraArgs:
   - -Wno-c11-extensions
   - -Wno-unknown-warning-option
   - -Wno-deprecated-declarations
-  - -Wno-unused-private-field
+  - -Wno-delete-non-abstract-non-virtual-dtor
```
mitsudome-r: Are there any reasons why you are adding `-Wno-delete-non-abstract-non-virtual-dtor`?

vishalchauhan-hi: Thank you Sensei for your comment

I added `-Wno-delete-non-abstract-non-virtual-dtor` only to keep this PR focused on re-enabling `clang-diagnostic-unused-private-field`, because that diagnostic seemed unrelated to the current task.

However, I agree that adding a new suppression for a different warning in this PR is not ideal. I have removed it now, so this PR only contains the intended change for `clang-diagnostic-unused-private-field`.

mitsudome-r (approved):

PR #12571, opened 2026-05-12
fix(clang-tidy): re-enable clang-diagnostic-deprecated-builtins

vish0012:
Re-enables `clang-diagnostic-deprecated-builtins` by removing `-Wno-deprecated-builtins` from `.clang-tidy-ci`.

The check was suppressed in #12431 with 1 reported error in `sensing/autoware_pointcloud_preprocessor`. This PR fixes it.

Refs #12450

### `autoware_pointcloud_preprocessor`

`robin_hood.h` used the old GCC workaround branch for `__has_trivial_copy` when `__GNUC__ < 5`.

Since clang also defines `__GNUC__`, clang entered this branch and reported:

```text
builtin __has_trivial_copy is deprecated; use __is_trivially_copyable instead
```

This PR keeps the old workaround for real GCC `< 5`, but excludes clang from that branch:

```cpp
#if defined(__GNUC__) && !defined(__clang__) && __GNUC__ < 5
```

So clang uses the existing modern branch:

```cpp
std::is_trivially_copyable<...>::value
```

## Verification

```bash
colcon build --packages-up-to autoware_pointcloud_preprocessor \
```

Result:

```text
Summary: 35 packages finished
```

```bash
run-clang-tidy -p build/ \
  -j $(nproc) \
  autoware_pointcloud_preprocessor \

grep -n "deprecated-builtins" /tmp/clang-tidy-result/report.log || echo "0 deprecated-builtins errors"
```

Result:

```text
0 deprecated-builtins errors
```

### DIFF
--- a/sensing/autoware_pointcloud_preprocessor/src/downsample_filter/robin_hood.h
+++ b/sensing/autoware_pointcloud_preprocessor/src/downsample_filter/robin_hood.h
@@ -229,7 +229,7 @@ static Counts & counts()
 
 // workaround missing "is_trivially_copyable" in g++ < 5.0
 // See https://stackoverflow.com/a/31798726/48181
-#if defined(__GNUC__) && __GNUC__ < 5
+#if defined(__GNUC__) && !defined(__clang__) && __GNUC__ < 5
 #define ROBIN_HOOD_IS_TRIVIALLY_COPYABLE(...) __has_trivial_copy(__VA_ARGS__)
 #else
 #define ROBIN_HOOD_IS_TRIVIALLY_COPYABLE(...) std::is_trivially_copyable<__VA_ARGS__>::value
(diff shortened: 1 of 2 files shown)


sensing/autoware_pointcloud_preprocessor/src/downsample_filter/robin_hood.h
```diff
@@ -229,7 +229,7 @@ static Counts & counts()
 
 // workaround missing "is_trivially_copyable" in g++ < 5.0
 // See https://stackoverflow.com/a/31798726/48181
-#if defined(__GNUC__) && __GNUC__ < 5
+#if defined(__GNUC__) && !defined(__clang__) && __GNUC__ < 5
```
mojomex: I think that no device that is used today is limited to GCC < 5 anymore -- all of our other packages are targeting at least GCC 11.

So, for the sake of cleanliness, I'd prefer if the workaround is removed altogether.

mojomex (commented):
Nice, thank you for contributing to stricter coding standards!

sensing/autoware_pointcloud_preprocessor/src/downsample_filter/robin_hood.h
vishalchauhan-hi: Thank you for the suggestion. I agree that the GCC < 5 workaround is no longer needed for the current Autoware toolchain. I removed the workaround entirely and now use std::is_trivially_copyable directly. I also reran the target build and clang-tidy check; deprecated-builtins reports 0 errors.

mojomex (approved):
LGTM, thank you!

PR #12581, opened 2026-05-13
test(autoware_traffic_light_multi_camera_fusion): add integration test for node

interimadd:
## Description

Adds an integration test for the `autoware_traffic_light_multi_camera_fusion` node. The test exercises the node end-to-end through its ROS interface (camera-info, ROI, signal, and vector-map topics in; traffic-signals topic out) so that the existing fusion behavior is captured before any refactoring is performed on the node implementation.

3 test cases are added:

- `TwoCamerasOneRegulatoryElementOutputsGreenGroup` — two cameras observe two distinct traffic lights bound to the same regulatory element. Per-camera fusion picks a single record per traffic-light ID, and group fusion accumulates evidence across both IDs to publish a single `GREEN` group.
- `HigherConfidenceColorWinsAmongConflictingObservations` — two cameras report conflicting colors (`RED` vs `GREEN`) for traffic lights in the same regulatory element, with the `GREEN` observation reported at a higher confidence. With the default `signal_consistency_check.enable = false`, the node selects the color from the higher-confidence observation and publishes a `GREEN` group.
- `TwoCamerasConflictingColorsOutputsFailsafeUnknown` — two cameras report conflicting colors (`RED` vs `GREEN`) for traffic lights in the same regulatory element. With `signal_consistency_check.enable = true` and `publish_partial_matched_signal = false`, the node detects the conflict and emits a fail-safe `UNKNOWN` group.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

Built and tested with coverage instrumentation against the new integration test on the same package:

```bash
colcon build --packages-select autoware_traffic_light_multi_camera_fusion
colcon test  --packages-select autoware_traffic_light_multi_camera_fusion --event-handlers console_cohesion+
```

Results: `100% tests passed, 0 tests failed out of 6` (2 gtest binaries + 4 linters).

Whole-package line coverage was measured via `lcov` both **before** and **after** introducing this integration test.

| Metric            | Before              | After               | Delta                  |
|-------------------|---------------------|---------------------|------------------------|
| Line coverage     | 24.2% (79 / 327)    | 88.7% (290 / 327)   | **+64.5pp** (+211 lines) |
| Function coverage | 30.3% (10 / 33)     | 100.0% (33 / 33)    | +69.7pp (+23 functions) |

Per-file breakdown:

| File                                              | Before              | After               |
|---------------------------------------------------|---------------------|---------------------|
| `src/signal_validator.cpp`                        | 93.0% (40 / 43)     | 93.0% (40 / 43)     |
| `src/traffic_light_multi_camera_fusion_node.cpp`  | 0.5% (1 / 215)      | 85.6% (184 / 215)   |
| `src/traffic_light_multi_camera_fusion_process.cpp` | 62.5% (30 / 48)   | 97.9% (47 / 48)     |

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None. Test-only addition; no production source files are modified.

### DIFF
--- a/perception/autoware_traffic_light_multi_camera_fusion/CMakeLists.txt
+++ b/perception/autoware_traffic_light_multi_camera_fusion/CMakeLists.txt
@@ -21,9 +21,16 @@ rclcpp_components_register_node(${PROJECT_NAME}
 
 if(BUILD_TESTING)
   find_package(ament_cmake_gtest REQUIRED)
+  find_package(ament_cmake_ros REQUIRED)
   ament_auto_add_gtest(${PROJECT_NAME}_test
     test/test_utils.cpp
     test/test_signal_validator.cpp
+  )
+  ament_add_ros_isolated_gtest(${PROJECT_NAME}_node_test
+    test/test_traffic_light_multi_camera_fusion_node.cpp
+  )
+  target_link_libraries(${PROJECT_NAME}_node_test
+    ${PROJECT_NAME}
   )
 endif()
 
--- a/perception/autoware_traffic_light_multi_camera_fusion/test/test_traffic_light_multi_camera_fusion_node.cpp
+++ b/perception/autoware_traffic_light_multi_camera_fusion/test/test_traffic_light_multi_camera_fusion_node.cpp
@@ -1,0 +1,352 @@
+// Copyright 2026 TIER IV, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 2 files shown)


a-maumau (approved):
Sorry for the late review.
Thank you for adding full node test.
I was a bit lazy when implementing the consistency check feature, so I planned to add the node later. It helped my work.

LGTM!

a-maumau:
This is not related to this PR, but when two signal conflicts with same confidence with no consistency check, I think the current logic's behavior is unpredictable (color can be switched in the same situation). This is not good...

interimadd:
Thank you for your review and comments!

> two signal conflicts with same confidence with no consistency check

I had not noticed that, but since a situation where confidence=1.0 with different colors may occur, it would be better to think about how to handle this situation.
It is also a good point of unit or integration tests that we can discuss the correct behavior of the logic using the implemented test cases. :+1:

PR #12590, opened 2026-05-15
fix(detected_object_feature_remover): update autoware system design format to add parameter for convex hull conversion

technolojin:
## Description

- Introduced existing parameter `run_convex_hull_conversion` to the autoware system design format  configuration.

NO LOGIC CHANGE

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


YoshiRi (approved):
LGTM

PR #12630, opened 2026-05-25
test(autoware_traffic_light_multi_camera_fusion): add unit tests for core logic

interimadd:
## Description

Add unit tests for the `MultiCameraFusion` core logic class in `autoware_traffic_light_multi_camera_fusion`. The class was previously exercised only indirectly through the node integration test (`test_traffic_light_multi_camera_fusion_node.cpp`), which made it hard to pin down individual branches such as unmapped traffic-light handling, message-lifespan filtering, and the signal consistency check.

This PR adds `test/test_multi_camera_fusion.cpp` and registers it in `CMakeLists.txt` so it is built and run as part of the package test suite.

### Test cases added

| Category | Test name | What it verifies |
|---|---|---|
| Fuse - basic | `SingleCameraSingleLightOutputsGroupWithMappedRegulatoryId` | Single ROI/signal pair → a single group keyed by the regulatory element ID. |
| Fuse - empty | `EmptyRoiArrayProducesEmptyTrafficLightGroups` | Empty input → empty output, no unmapped IDs. |
| Fuse - unmapped | `UnknownTrafficLightIdIsRecordedAsUnmapped` | Traffic light ID not in the lanelet map is reported via `unmapped_traffic_light_ids`. |
| Fuse - null map | `NullLaneletMapTreatsAllTrafficLightsAsUnmapped` | A null lanelet map pointer makes every observed ID unmapped. |
| Fuse - mismatch | `RoiWithoutMatchingSignalIsIgnored` | ROI without a matching signal in the same array is discarded. |
| Fuse - fusion | `HigherConfidenceColorIsSelectedAcrossTwoLights` | When two lights of the same regulatory element disagree, the higher-confidence color wins. |
| Fuse - lifespan | `RecordOlderThanMessageLifespanIsDiscarded` | A record older than `message_lifespan` is purged before group fusion. |
| Fuse - lifespan | `RecordWithinMessageLifespanIsKeptAndAccumulated` | A record inside the lifespan window is retained and aggregated. |
| Consistency check | `ConsistencyCheckWithSameColorOutputsNoConflict` | With `use_signal_consistency_check = true`, identical colors produce no conflict. |
| Consistency check | `ConsistencyCheckWithConflictingColorsOutputsUnknownFailsafe` | Conflicting colors trigger the fail-safe `UNKNOWN` group and a `CONFLICT`. |

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

- `colcon build --packages-select autoware_traffic_light_multi_camera_fusion`
- `colcon test --packages-select autoware_traffic_light_multi_camera_fusion --event-handlers console_cohesion+` — all tests pass (including the 10 new gtest cases).
- Coverage measured with `lcov` (results above).

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


a-maumau (commented):
LGTM.

I think overall the usual cases are tested in this new test.
It might be a nitpick, but we might want to check that confidences will get aggregated by taking min.

In the code, it is taking the min of confidences among the detected signal by `utils::get_min_confidence` for handling the case when the confidences differ among the light bulbs.

interimadd:
@a-maumau 
Thank you for the comment!
I have added the case related to `utils::get_min_confidence` in https://github.com/autowarefoundation/autoware_universe/pull/12630/commits/6e1b95ac49c2584db0572c6ce4c7dccfc1a3d04a
Could you check it?

a-maumau (approved):
LGTM!

PR #12631, opened 2026-05-25
feat(autoware_tensorrt_plugins): add do sort attribute in GetIndicePairsImplicitGemmPlugin for faster inference

vividf:
## Description

This PR adds a `do_sort` attribute to `GetIndicePairsImplicitGemmPlugin` to control whether pair-mask argsort is executed during sparse indice-pair generation.

By default, `do_sort` is set to `1`, preserving the existing behavior and maintaining backward compatibility with existing ONNX models.

When `do_sort = 0`, the plugin skips pair-mask sorting and forwards this option to spconv’s `getIndicePairsImplicitGemm()` API.

## Motivation

Profiling shows that pair-mask sorting introduces a non-trivial overhead in sparse indice-pair generation.

Although spconv documents this optimization primarily for INT8 deployment, our measurements show that disabling sorting can also significantly reduce sparse encoder latency in FP16 deployments.

This optimization therefore benefits both:

- FP16 TensorRT deployment
- INT8 TensorRT deployment

while keeping the behavior configurable through a plugin attribute.

## Changes

- Added `do_sort` to `GetIndicesPairsImplicitGemmParameters`
  - `1`: enable pair-mask sorting (default)
  - `0`: disable pair-mask sorting
- Added `do_sort` to plugin serialization/deserialization
- Forwarded `do_sort` to spconv indice-pair generation
- Updated plugin creator to support:
  - Legacy ONNX models (11 attributes)
  - New ONNX models with `do_sort` (12 attributes)

## Related links
* Performance comparison with another PR for BEVFusion bench testing
https://github.com/autowarefoundation/autoware_universe/pull/12658
* PTv3:
https://github.com/tier4/AWML/pull/210

**Parent Issue:**

- Link

## How was this PR tested?
On the bench with BEVFusion

### ONNX Graph Comparison

| Without `do_sort` Attribute (Legacy) | With `do_sort` Attribute |
|--------------------------------------|--------------------------|
| Without do_sort | With do_sort |

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


perception/autoware_tensorrt_plugins/include/autoware/tensorrt_plugins/get_indices_pairs_implicit_gemm_plugin.hpp
```diff
@@ -45,6 +45,10 @@ struct GetIndicesPairsImplicitGemmParameters
   std::vector<std::int32_t> stride;
   std::int32_t subm;  // cSpell:ignore subm
   std::int32_t transpose;
+  // do_sort: 1 = run the pair-mask argsort (default; required for FP16).
```
mojomex: > required for FP16

Thios PR applies to all precisions equally, right? SO having FP16 with `do_sort = 0` should be allowed, right?

perception/autoware_tensorrt_plugins/src/get_indices_pairs_implicit_gemm_plugin_creator.cpp
```diff
@@ -69,7 +70,8 @@ IPluginV3 * GetIndicesPairsImplicitGemmPluginCreator::createPlugin(
       nvinfer1::PluginField const * fields{fc->fields};
       std::int32_t num_fields{fc->nbFields};
 
-      PLUGIN_VALIDATE(num_fields == 11);
+      // Accept 11 (ONNX without do_sort attribute) or 12 (with do_sort attribute)
+      PLUGIN_VALIDATE(num_fields == 11 || num_fields == 12);
```
mojomex, line 74: To notify and push users to think about whether they need sorting, outputting a deprecation notice when used with `11` fields could be an idea. Not sure if that is good practice or not..

mojomex (commented):
Looking good! Just left one small comment for clarification, and one optional one. Please check.

perception/autoware_tensorrt_plugins/src/get_indices_pairs_implicit_gemm_plugin_creator.cpp
vividf: Thanks, added in https://github.com/autowarefoundation/autoware_universe/pull/12631/commits/3a0b72d5192e239f485a75cd0971dfd1edf1ae03

perception/autoware_tensorrt_plugins/include/autoware/tensorrt_plugins/get_indices_pairs_implicit_gemm_plugin.hpp
vividf: Thanks, fixed in https://github.com/autowarefoundation/autoware_universe/pull/12631/commits/3a0b72d5192e239f485a75cd0971dfd1edf1ae03

vividf:
@mojomex 
Thanks for the review! I've addressed the feedback.
Please take another look when you have a chance.

mojomex (approved):
Nice!~

ktro2828 (approved):

PR #12632, opened 2026-05-25
test(autoware_traffic_light_arbiter): add characterization test suite

takam5f2:
## Description

Add `test/test_traffic_arbiter_characteristic.cpp` — a self-contained characterization test suite for `autoware_traffic_light_arbiter` that pins the arbiter's full behavior contract before structural refactoring.

### Background

The existing `test/test_node.cpp` provides smoke-level coverage but is not strong enough as a safety net for upcoming refactoring of `autoware_traffic_light_arbiter`:

- It depends on `autoware_test_utils` and a real osm map under that package, so the test cannot run from a build directory without first installing.
- The test bodies use `isEqual` over the whole `TrafficLightGroupArray`, so a single field mismatch surfaces as a generic equality failure with little diagnostic detail.
- It exercises only a subset of arbitration paths: the Signal Matching mode (`enable_signal_matching=true`), pedestrian-specific behavior, map-boundary cases, and time-tolerance paths are not pinned.

This PR adds a second test file (`test_traffic_arbiter_characteristic.cpp`) targeted at characterization — freezing the current behavior so the upcoming refactoring (extracting the arbitrate logic into a ROS-free `ArbiterCore`, splitting `arbitrateAndPublish`, etc.) can proceed with regression visibility.

### Features

- **24 tests** across three sections covering 22+ behavior contracts
- **Programmatically built minimal map** (no `autoware_test_utils` dependency)
- **Direct `rclcpp::Parameter` overrides** instead of YAML reading at test time
- **Per-spec assertions** (color/shape/predictions count) rather than whole-message equality
- **One spec rule per test** — each observable outcome (matched, mismatch, off-map drop, etc.) is exercised by a dedicated test so a failure points directly to the broken rule

### Architecture

The file is organized into three sections that match the spec matrix below:

| Section | Theme | Tests |
| --- | --- | --- |
| A. Signal Matching mode | Arbitration rules across source agreement (matched / mismatch / single-side / pedestrian path / equivalence) | 10 |
| B. Priority-based mode | Per-shape selection driven by `source_priority` and confidence | 4 |
| C. Focused specifications | Boundary conditions, input validation, time tolerances, priority overrides, predictions | 10 |

Supporting helpers are grouped inside the file by role: signal IDs, map construction, input message builders, timestamp utility, and the test fixture (which owns node setup, publish helpers, and output inspection in one place).

## Related links

**Parent Issue:**

- None

**Relevant code:**

- `perception/autoware_traffic_light_arbiter/test/test_node.cpp` — existing integration test, intentionally left untouched per the characterization-test convention.

## How was this PR tested?

```bash
colcon build --packages-select autoware_traffic_light_arbiter \
  --cmake-args -DCMAKE_BUILD_TYPE=Release -DBUILD_TESTING=ON
colcon test --packages-select autoware_traffic_light_arbiter \
  --event-handlers console_direct+
```

All 24 tests pass (~15 s suite wall-time on the development machine).

Line coverage on a debug build with `--coverage` is **96.7 %** for `src/`; the remaining uncovered lines are dead branches (`__builtin_unreachable`, an `if (!perception && !external) continue;` after a union-of-ids loop) and closing braces.

### Specification coverage matrix

Each row describes a behavior the arbiter is expected to satisfy and the test that pins it.

#### Signal Matching mode

| Specification | Test |
| --- | --- |
| matched (color & shape agree) → perception passes through | `signalMatchingMatchedPassesThrough` |
| color mismatch → UNKNOWN over shared shape | `signalMatchingColorMismatchProducesUnknown` |
| element-count mismatch → UNKNOWN over shape union | `signalMatchingElementCountMismatchProducesUnknown` |
| single-side non-pedestrian → UNKNOWN | `signalMatchingSingleSourceNonPedestrianYieldsUnknown` |
| confidence does not affect equivalence judgement | `signalMatchingIgnoresConfidenceInEquivalence` |
| pedestrian + external priority → external value wins | `signalMatchingPedestrianFallbackUsesSourcePriority` |
| pedestrian + perception priority → perception value wins | `signalMatchingPedestrianPerceptionPriority` |
| pedestrian + CONFIDENCE → higher confidence per shape wins | `signalMatchingPedestrianConfidenceMode` |
| pedestrian + single-side → the present side passes through | `signalMatchingPedestrianSingleSourcePasses` |

#### Priority-based mode

| Specification | Test |
| --- | --- |
| CONFIDENCE picks the higher confidence per shape | `priorityBasedConfidencePicksHigherValue` |
| only external contributes (per id) → passes through (multi-shape preserved) | `priorityBasedExternalOnlyPassesThrough` |
| only perception contributes (per id) → passes through | `priorityBasedPerceptionOnlyPassesThrough` |
| perception priority overrides higher external confidence | `priorityFlagOverridesHigherConfidence` |
| external priority overrides higher perception confidence | `priorityFlagFromExternalOverridesHigherConfidence` |

#### Map / startup

| Specification | Test |
| --- | --- |
| map not yet received → no publish | `perceptionBeforeMapProducesNoOutput` |
| map carries no traffic lights → empty publish | `emptyMapProducesEmptyOutput` |
| off-map id → dropped (WARN+skip) | `signalMatchingOffMapIdDropped`, `priorityBasedOffMapIdDropped` |

#### Time tolerances

| Specification | Test |
| --- | --- |
| `external_delay_tolerance`: stale external message dropped | `externalDelayToleranceDropsStaleMessage` |
| `external_time_tolerance`: stored external cleaned up on perception arrival | `externalTimeToleranceCleanupOnPerception` |
| `perception_time_tolerance`: latest perception cleared on external arrival | `perceptionTimeToleranceClearsLatestPerception` |

#### External accumulation

| Specification | Test |
| --- | --- |
| successive external publishes with different ids → all accumulated and published | `multipleExternalSourcesAccumulate` |

#### Input validation

| Specification | Test |
| --- | --- |
| unknown `source_priority` → falls back to CONFIDENCE | `unknownSourcePriorityFallsBackToConfidence` |

#### Predictions

| Specification | Test |
| --- | --- |
| predictions from both sources → merged | `signalMatchingMatchedPassesThrough` |
| predictions from a single side → propagated | `predictionsFromSingleSidePropagate` |

### Test list

**Section A — Signal Matching mode**

1. `signalMatchingMatchedPassesThrough`
2. `signalMatchingColorMismatchProducesUnknown`
3. `signalMatchingElementCountMismatchProducesUnknown`
4. `signalMatchingOffMapIdDropped`
5. `signalMatchingPedestrianFallbackUsesSourcePriority`
6. `signalMatchingPedestrianPerceptionPriority`
7. `signalMatchingPedestrianConfidenceMode`
8. `signalMatchingPedestrianSingleSourcePasses`
9. `signalMatchingSingleSourceNonPedestrianYieldsUnknown`
10. `signalMatchingIgnoresConfidenceInEquivalence`

**Section B — Priority-based mode**

11. `priorityBasedConfidencePicksHigherValue`
12. `priorityBasedExternalOnlyPassesThrough`
13. `priorityBasedPerceptionOnlyPassesThrough`
14. `priorityBasedOffMapIdDropped`

**Section C — Focused specifications**

15. `perceptionBeforeMapProducesNoOutput`
16. `emptyMapProducesEmptyOutput`
17. `unknownSourcePriorityFallsBackToConfidence`
18. `priorityFlagOverridesHigherConfidence`
19. `priorityFlagFromExternalOverridesHigherConfidence`
20. `multipleExternalSourcesAccumulate`
21. `externalDelayToleranceDropsStaleMessage`
22. `externalTimeToleranceCleanupOnPerception`
23. `perceptionTimeToleranceClearsLatestPerception`
24. `predictionsFromSingleSidePropagate`

### Coexistence with `test_node.cpp`

`test/test_node.cpp` is intentionally **not** modified or removed:

- Every spec exercised by `test_node.cpp` is now also pinned in the new file with finer-grained assertions (the mapping is included in the matrix above).
- `test_node.cpp` plays the role of a smoke test against the real osm map; the new file plays the role of a programmatic characterization layer.
- A future cleanup, once the planned `ArbiterCore` extraction lands and its unit tests subsume these integration-level checks, may remove `test_node.cpp`. That is intentionally out of scope here.

### Design choices worth flagging

- `make_traffic_light_element(color, shape, confidence = 1.0f)` carries a `1.0f` default so test bodies can omit the confidence argument when the arbiter does not consult it. Only the CONFIDENCE / priority-override / equivalence-ignore tests keep an explicit value, and that value is itself the spec (e.g. "the lower-confidence side still wins").
- Topic-name and tolerance constants live inside the fixture as `private` / `protected static constexpr` because no free helper accesses them.
- Each test fixes one observable outcome (matched / mismatch type / off-map drop / passthrough variant) so a single assertion failure points directly to the broken spec rule. Bundled-scenario tests were intentionally split to avoid coupling unrelated outcomes in one test body.

### Topic changes

None — this PR only adds test code.

### ROS Parameter Changes

None.

## Effects on system behavior

None. Test-only change.

### DIFF
--- a/perception/autoware_traffic_light_arbiter/CMakeLists.txt
+++ b/perception/autoware_traffic_light_arbiter/CMakeLists.txt
@@ -18,6 +18,9 @@ if(BUILD_TESTING)
   ament_auto_add_gtest(${PROJECT_NAME}_test
     test/test_node.cpp
   )
+  ament_auto_add_gtest(${PROJECT_NAME}_characteristic_test
+    test/test_traffic_arbiter_characteristic.cpp
+  )
 endif()
 
 autoware_ament_auto_package(INSTALL_TO_SHARE launch config)
--- a/perception/autoware_traffic_light_arbiter/test/test_traffic_arbiter_characteristic.cpp
+++ b/perception/autoware_traffic_light_arbiter/test/test_traffic_arbiter_characteristic.cpp
@@ -404,32 +404,25 @@ protected:
     return group ? group->predictions.size() : SIZE_MAX;
   }
 
-  // Publish-count specs: some contracts hinge on whether the arbiter emits
-  // a message at all, independent of the content. We count callbacks on
-  // the output subscription and expose these helpers so each contract is
-  // named at the call site.
-
-  // Snapshot the running publish count. Capture before an action that
-  // should not trigger a publish, then pair with expect_no_publish_since().
-  std::size_t publish_count() const { return arbiter_publish_count_; }
+  // Publish-count specs: a handful of contracts cannot be expressed by
+  // observing the published content alone, because they hinge on whether
+  // the arbiter emitted any message at all. For those, we count callbacks
+  // on the output subscription. Tests whose spec can be expressed as
+  // "the next publish contains X (and not the value that would have been
+  // emitted if the spec were violated)" should compare content via
+  // observed_* instead — content checks are stronger because they verify
+  // the user-visible behavior.
 
   // Spec: the arbiter has not emitted any TrafficLightGroupArray yet.
-  // Used when an early-return path is expected (e.g., perception arrived
-  // before the vector_map).
+  // Content-only checks cannot distinguish "subscriber never received
+  // anything" from "subscriber received a default-constructed payload",
+  // so the count is the only direct test of the early-return contract.
   void expect_no_publish() { EXPECT_EQ(arbiter_publish_count_, 0u); }
 
   // Spec: the arbiter has emitted at least one TrafficLightGroupArray.
   // Used when the published content is expected to be empty so a
   // content-based check cannot tell "no publish" from "empty publish".
   void expect_publish_happened() { ASSERT_GE(arbiter_publish_count_, 1u); }
-
-  // Spec: the arbiter has not emitted any new TrafficLightGroupArray
-  // since `baseline`. Used after an action whose contract is "this input
-  // must be dropped before reaching arbitrateAndPublish".
-  void expect_no_publish_since(std::size_t baseline)
-  {
-    EXPECT_EQ(arbiter_publish_count_, baseline);
-  }
 
   // Wrapped in std::optional because LaneletMapBin (a generated ROS msg)
   // has no usable default constructor that can be invoked at static storage
@@ -947,22 +940,21 @@ TEST_F(ArbiterCharacteristic, multipleExternalSourcesAccumulate)
 
 TEST_F(ArbiterCharacteristic, externalDelayToleranceDropsStaleMessage)
 {
-  // Arrange: establish a perception baseline so we can detect any extra publish.
+  // Arrange: seed a perception value (RED). If the stale external below were
+  // ever accepted, the output would change to GREEN — the content check at
+  // the bottom is what pins "stale dropped".
   start_arbiter(false, "confidence");  // priority-based mode, "confidence" priority
   publish_map();
   publish_perception(make_signal_array(
     t0_, map_ids::vehicle_signal_a,
     {make_traffic_light_element(TrafficLightElement::RED, TrafficLightElement::CIRCLE)}));
-  const auto baseline = publish_count();
 
   // Act: external_delay_tolerance defaults to 5.0s; 20s in the past is well past it.
   publish_external(make_signal_array(
     offset_time(t0_, -20.0), map_ids::vehicle_signal_a,
     {make_traffic_light_element(TrafficLightElement::GREEN, TrafficLightElement::CIRCLE)}));
 
-  // Assert: stale external message must be dropped before arbitrateAndPublish runs.
-  expect_no_publish_since(baseline);
-  // Last publish must still reflect the earlier perception value.
+  // Assert: the stale GREEN never propagates; the published color stays RED.
   EXPECT_EQ(observed_color(map_ids::vehicle_signal_a), TrafficLightElement::RED);
 }
 
--- a/perception/autoware_traffic_light_arbiter/test/test_traffic_arbiter_characteristic.cpp
+++ b/perception/autoware_traffic_light_arbiter/test/test_traffic_arbiter_characteristic.cpp
@@ -311,8 +311,7 @@ protected:
     executor_ = std::make_shared<rclcpp::executors::SingleThreadedExecutor>();
     executor_->add_node(arbiter_);
     executor_->add_node(test_node_);
-    // Allow discovery to settle before any publish so transient_local replay works.
-    spinFor(std::chrono::milliseconds(200));
+    spinFor(std::chrono::milliseconds(150));
   }
 
   void publishMap() { publishMap(*map_bin_); }
--- a/perception/autoware_traffic_light_arbiter/test/test_traffic_arbiter_characteristic.cpp
+++ b/perception/autoware_traffic_light_arbiter/test/test_traffic_arbiter_characteristic.cpp
@@ -1105,7 +1105,6 @@ TEST_F(ArbiterCharacteristic, externalDelayToleranceDropsStaleMessage)
     {makeTrafficLightElement(TrafficLightElement::RED, TrafficLightElement::CIRCLE)}));
   publishPerception(perception_traffic_signal);
   const auto baseline_count = arbiter_publish_count_;
-  ASSERT_GE(baseline_count, 1u);
 
   // external_delay_tolerance defaults to 5.0s; 20s in the past is well past it.
   TrafficLightGroupArray stale_external_traffic_signal;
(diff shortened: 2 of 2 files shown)


takam5f2:
## Note for reviewers: pedestrian signal handling carries spec ambiguity

While writing these characterization tests I noticed that the pedestrian signal path in Signal Matching mode is in conflict with the documented spec, and I'd like to flag it as a known limitation **before** these tests are accepted as a baseline.

### What the README says

> Note, this option [`source_priority`] will not be effective if signal matching is enabled.
>
> — [`README.md`](https://github.com/autowarefoundation/autoware_universe/blob/test/autoware_traffic_light_arbiter/add-characterization-test/perception/autoware_traffic_light_arbiter/README.md#source-priority)

The Signal Match Validator table in the README treats all colors uniformly (no pedestrian/vehicle distinction).

### What the code does

In `SignalMatchValidator::validateSignals` ([`src/signal_match_validator.cpp:247-254`](https://github.com/autowarefoundation/autoware_universe/blob/test/autoware_traffic_light_arbiter/add-characterization-test/perception/autoware_traffic_light_arbiter/src/signal_match_validator.cpp#L247-L254)), pedestrian signals are routed through `get_highest_confidence_signal(..., source_priority_)` — i.e. **`source_priority` _is_ effective for pedestrian signals**, contradicting the README.

This is intentional in spirit (the inline comment says `TODO(TomohitoAndo): Validate pedestrian signals`), so the deviation is a temporary state, not a stable spec.

### Impact on this PR

Six of the eighteen tests pin this temporary behavior:

- `signalMatchingPedestrianFallbackUsesSourcePriority`
- `signalMatchingPedestrianPerceptionPriority`
- `signalMatchingPedestrianConfidenceMode`
- `signalMatchingPedestrianSingleSourcePasses`
- `signalMatchingSingleSourceNonPedestrianYieldsUnknown` (kept as the explicit contrast case)
- `signalMatchingLongestPath` (the pedestrian id `M2` in the mixed scenario)

These tests freeze the **current, fallback** behavior, not the documented spec. Once the `TODO(TomohitoAndo)` is resolved and pedestrian signals receive proper matching, these tests will need to be updated. That update is **expected**, not a regression.

### Why I'm flagging it here rather than fixing it

- Fixing the contradiction means either a spec change (extend matching to pedestrian signals) or a doc change (revise the README to describe the fallback). Both are out of scope for a test-only PR.
- The point of a characterization suite is to freeze the current behavior; freezing the fallback as well is the right thing to do — otherwise an unrelated refactor could silently flip this behavior.

### Suggested follow-ups (separate work)

1. Decide which side of the contradiction wins (extend matching to pedestrians, or document the fallback in the README).
2. After the planned `ArbiterCore` extraction lands, revisit pedestrian-specific tests and either keep them (if the fallback stays) or update them (if matching is extended).

Happy to take feedback on whether the four pedestrian-priority tests should be condensed or kept as-is.

takam5f2:
## Follow-up: my earlier comment is incorrect — matching mode is legacy, not current policy

**Correction to my earlier comment above.** That comment framed the pedestrian/vehicle asymmetry as a spec ambiguity to be resolved by extending matching to pedestrian signals. After consulting the original author of `SignalMatchValidator` internally, that framing is **wrong**. The matching mode itself is the deprecated direction, not the direction of resolution. The earlier comment is left intact for thread history; this comment supersedes its conclusions.

### Background

`enable_signal_matching=true` is a legacy mode dating from when L4 authorization required cross-checking V2I and perception signals. The policy was later changed: V2I should not be trusted for safety-critical color decisions. Production launches now use:

```yaml
enable_signal_matching: false
source_priority: perception          # perception always wins
```

The pedestrian-specific branch inside `SignalMatchValidator` is similarly legacy: at the time matching was implemented, pedestrian V2I sources did not exist, so pedestrian signals took the "highest confidence" path. Now pedestrian V2I exists, but matching itself was abandoned, so there was never a reason to unify the pedestrian path with the vehicle one.

### What this means for the test suite

The tests in this PR pin all observable behavior including the legacy `enable_signal_matching=true` path. The six pedestrian-related tests freeze the legacy fallback. This is the correct thing for a characterization suite to do — it lets future refactoring (e.g. removing the matching mode entirely) proceed with regression visibility.

### Suggested follow-ups (separate work)

1. Decide whether `enable_signal_matching` should be deprecated or removed in a future release. The matching mode is dead code in current operation.
2. Update the `README.md` Signal Match Validator section, which describes the legacy policy as if it were the active spec.

takam5f2:
@a-maumau

When you have a moment, could I have your review on this PR? This PR will **NOT change** behavior of the ROS Node. It adds the characterization test suite I plan to use as a safety-net before refactoring the arbitration logic out of the ROS Node. 

Best,

a-maumau (approved):
Sorry for the late review.
LGTM!

PR #12655, opened 2026-05-28
feat(autoware_ptv3): split the backbone and head

amadeuszsz:
## Description

Splits the monolithic PTv3 TensorRT engine into a separate backbone engine and segmentation head engine. The backbone is run first and its point features are fed to the segmentation head, which keeps the door open for additional heads on the same backbone in a later PR. Preprocessing, postprocessing, and the segmentation outputs are unchanged, only the engine layout and the parameter structure differ.

Demo shows output with detection 3D head, but this will be introduced in separate PR in order to keep review easier.

[ptv3-multi-head-short-demo.webm](https://github.com/user-attachments/assets/92418f6b-2912-49d7-ac83-4d3be9842800)

Key changes:

- **Engine split:** `PTv3TRT` now loads two engines instead of one fused engine, a backbone engine and a segmentation head engine, taking a `TrtCommonConfig` for the backbone and an `std::optional<TrtCommonConfig>` for the segmentation head. The backbone consumes `grid_coord`, `feat`, and `serialized_code` and produces the intermediate tensors `point_feat`, `point_grid_coord`, and `point_offset`. The segmentation head consumes `point_feat` and produces `pred_labels` and `pred_probs`. The intermediate buffers are allocated once and shared between the two engines.
- **Explicit boundary tensor types:** every `NetworkIO` entry on both engines declares an explicit dtype so TensorRT cannot silently reformat an engine boundary when the engine is built in fp16 mode. Features and probabilities are `kFLOAT`, `grid_coord` and `point_grid_coord` are `kINT32`, and `serialized_code`, `point_offset`, and `pred_labels` are `kINT64`.
- **Preprocess grid coordinate dtype:** `PreprocessCuda::generateFeatures` now writes the voxel grid coordinates as `int32` instead of `int64` to match the backbone engine `grid_coord` input. This is the only functional change to the preprocessing kernel, the rest of the kernel is unchanged.
- **Head selection flag:** `PTv3Config` gains a `use_seg3d_head_` flag and a `backbone_feat_dim_` constant. The node declares the segmentation head parameters and creates the segmentation publishers only when `segmentation3d.use_head` is true. The constructor throws when no head is enabled, which preserves a clear extension point for the forthcoming detection head.
- **Config split:** the single `ml_package_ptv3.param.yaml` is replaced by two artifact files, `ml_package_ptv3_backbone.param.yaml` with the backbone voxelization parameters under `backbone.*`, and `ml_package_ptv3_seg3d_head.param.yaml` with the segmentation class metadata under `segmentation3d.*`. The node parameters in `ptv3.param.yaml` are reorganised into the same `backbone.*` and `segmentation3d.*` namespaces, and the ONNX and engine paths are split into `*_backbone` and `*_seg3d_head` artifacts.
- **Launch:** the single `ml_package_param_path` argument is replaced by `backbone_ml_package_param_path` and `seg3d_head_ml_package_param_path`, a `use_seg3d_head` argument is added, the `model_name` argument is renamed to `model_variant`, and the unused `class_remapper_param_path` argument is removed.
- **Schemas:** `ptv3.schema.json` and `ml_package_ptv3.schema.json` are restructured to validate the new `backbone.*` and `segmentation3d.*` namespaces.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

- This PR depends on `autoware_tensorrt_common` providing the optional `dtype` field on `NetworkIO`, which is already a dependency of this package. The explicit boundary types guarantee the shared `point_feat` boundary stays float32 across the two engines in fp16 mode.
- Delete cached `.engine` files when upgrading. The single `ptv3.engine` is replaced by separate `*_backbone.engine` and `*_seg3d_head.engine` files, and the boundary tensor types changed, so stale engines will fail validation and trigger a rebuild.
- `ml_package_ptv3.param.yaml` is deleted. Any artifact directory or launch override that loads it must be updated to the two new files.
- The kernels keep their original file names and class names, so the only kernel diff is the `int64` to `int32` grid coordinate change.

### Topic changes

No topic additions or removals in this PR.

#### Additions and removals

| Change type | Parameter Name                                      | Type       | Default Value | Description                                                  |
|:------------|:----------------------------------------------------|:-----------|:--------------|:-------------------------------------------------------------|
| Added       | `backbone.onnx_path`                                | `string`   | -             | Backbone ONNX path.                                          |
| Added       | `backbone.engine_path`                              | `string`   | -             | Backbone TRT engine path.                                    |
| Added       | `backbone.voxels_num`                               | `int64[]`  | -             | Voxel count profile `[min, opt, max]`.                       |
| Added       | `backbone.point_cloud_range`                        | `double[]` | -             | Crop range `[x_min, y_min, z_min, x_max, y_max, z_max]`.     |
| Added       | `backbone.voxel_size`                               | `double[]` | -             | Backbone voxel size `[x, y, z]`.                             |
| Added       | `segmentation3d.use_head`                           | `bool`     | `true`        | Enable the segmentation head.                                |
| Added       | `segmentation3d.onnx_path`                          | `string`   | -             | Segmentation head ONNX path.                                 |
| Added       | `segmentation3d.engine_path`                        | `string`   | -             | Segmentation head TRT engine path.                           |
| Added       | `segmentation3d.source_reconstruction`              | `string`   | `"full"`      | Point reconstruction mode: `none`, `partial`, or `full`.     |
| Added       | `segmentation3d.class_names`                        | `string[]` | -             | Segmentation class names.                                    |
| Added       | `segmentation3d.palette`                            | `int64[]`  | -             | Flat RGB palette matching the class names.                   |
| Added       | `segmentation3d.filter.class_probability_threshold` | `double`   | `0.05`        | Probability threshold for the filtered output cloud.         |
| Added       | `segmentation3d.filter.classes`                     | `string[]` | -             | Class names excluded from the filtered cloud above threshold.|
| Added       | `segmentation3d.filter.output_format`               | `string`   | `""`          | Output format for the filtered cloud, empty preserves input. |
| Removed     | `onnx_path`                                         | `string`   | -             | Replaced by `backbone.onnx_path` and `segmentation3d.onnx_path`. |
| Removed     | `engine_path`                                       | `string`   | -             | Replaced by `backbone.engine_path` and `segmentation3d.engine_path`. |
| Removed     | `source_reconstruction`                             | `string`   | -             | Replaced by `segmentation3d.source_reconstruction`.          |
| Removed     | `filter.class_probability_threshold`                | `float`    | -             | Replaced by `segmentation3d.filter.class_probability_threshold`. |
| Removed     | `filter.classes`                                    | `string[]` | -             | Replaced by `segmentation3d.filter.classes`.                 |
| Removed     | `filter.output_format`                              | `string`   | -             | Replaced by `segmentation3d.filter.output_format`.           |
| Removed     | `voxels_num`                                        | `int64[]`  | -             | Replaced by `backbone.voxels_num`.                           |
| Removed     | `point_cloud_range`                                 | `double[]` | -             | Replaced by `backbone.point_cloud_range`.                    |
| Removed     | `voxel_size`                                        | `double[]` | -             | Replaced by `backbone.voxel_size`.                           |
| Removed     | `class_names`                                       | `string[]` | -             | Replaced by `segmentation3d.class_names`.                    |
| Removed     | `palette`                                           | `int64[]`  | -             | Replaced by `segmentation3d.palette`.                        |

## Effects on system behavior

- **Breaking config change:** existing artifact directories with `ml_package_ptv3.param.yaml` must be updated to `ml_package_ptv3_backbone.param.yaml` and `ml_package_ptv3_seg3d_head.param.yaml`, and any launch override of the old single `ml_package_param_path` must be migrated to the two new arguments.
- **Stale engine incompatibility:** the monolithic `ptv3.engine` is replaced by separate backbone and segmentation head engines with explicitly typed boundary tensors. Existing cached `.engine` files must be deleted and rebuilt.
- **Updated backbone and class defaults:** the example parameter files ship the voxelization defaults and class list that match the new split backbone and segmentation head artifacts.
- **Runtime behavior unchanged** for segmentation. The same preprocessing, backbone features, and segmentation outputs are produced, only the engine file layout and the parameter namespaces differ.


mojomex (changes requested):
Thank you for the PR!

Before reviewing I'd like to ask you to split it up into a few ~500 line-ish PRs, as it's impossible to grasp all changes at once.

I'd also like to ask if the det3d and non_maximum_suppression has been copied from, or can be unified with, the Centerpoint head (or other models). In that case, I'd like a separate refactoring PR for that first.

I also see that there's center_head AND trans_head. I'd like to ask you to choose one, as it will otherwise cause explosion of configuration options and will not be clear to users what to use.

Sorry for the harsh initial feedback 🙇 It's very impressive work, I just want to avoid a situation where we can't do proper QA on it due to sheer size.

amadeuszsz:
@mojomex 
Thanks for a quick comment!

> Before reviewing I'd like to ask you to split it up into a few ~500 line-ish PRs, as it's impossible to grasp all changes at once.

Let me summary PR first.

> I'd also like to ask if the det3d and non_maximum_suppression has been copied from, or can be unified with, the Centerpoint head (or other models). In that case, I'd like a separate refactoring PR for that first.

I will take a look.

> I also see that there's center_head AND trans_head. I'd like to ask you to choose one, as it will otherwise cause explosion of configuration options and will not be clear to users what to use.

Valid point. I will remove CenterPoint's head for now. PR we will have a record of its implementation if needed in the future.

> Sorry for the harsh initial feedback 🙇 It's very impressive work, I just want to avoid a situation where we can't do proper QA on it due to sheer size.

I will prepare better description for PR review purposes.

amadeuszsz:
@codex review

perception/autoware_ptv3/schema/ptv3.schema.json
```diff
@@ -1,85 +1,108 @@
...
+      },
+      "required": ["onnx_path", "engine_path"],
+      "additionalProperties": false
+    },
+    "segmentation3d": {
+      "type": "object",
+      "properties": {
+        "use_head": {
+          "description": "Enable the segmentation head."
         },
```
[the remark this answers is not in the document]
amadeuszsz: Then it fails because in config we use substitutions. No thanks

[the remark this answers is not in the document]
amadeuszsz: Then it fails because in config we use substitutions. No thanks

perception/autoware_ptv3/README.md
```diff
@@ -6,7 +6,8 @@ The `autoware_ptv3` package is used for 3D lidar segmentation.
 
 ## Inner-workings / Algorithms
 
-This package implements a TensorRT powered inference node for Point Transformers V3 (PTv3) [1].
+This package implements a TensorRT powered inference node for Point Transformers V3
```
mojomex:
```suggestion
This package implements a TensorRT powered inference node for Point Transformer V3
```

perception/autoware_ptv3/README.md
```diff
@@ -74,9 +76,13 @@ supports:
 The filtered output cloud format is controlled by `filter.output_format`. When it is set to an
 empty string, the filtered output preserves the same format as the input cloud.
 
+If a point cloud exceeds `cloud_capacity` or voxelization exceeds the configured maximum voxel
+count, the input is clipped to the configured capacity and an error is logged.
+
 ## Trained Models
 
-The model was trained on the T4Dataset using approximately 4,000 frames and is available in the Autoware artifacts.
+The segmentation head was trained on the T4Dataset using approximately 4,000 frames. The models are
```
mojomex:
```suggestion
The segmentation head was trained on a proprietary dataset of approximately 4,000 frames. The models are
```

perception/autoware_ptv3/README.md
```diff
@@ -42,8 +43,9 @@ Autoware installs it automatically in its setup script. If needed, the user can
 
 {{ json_to_markdown("perception/autoware_ptv3/schema/ml_package_ptv3.schema.json") }}
 
-`filter.*` parameters are configured in `config/ptv3.param.yaml`, while class metadata and the
-visualization `palette` are configured in `config/ml_package_ptv3.param.yaml`.
+Runtime head enablement and filtering parameters are configured in `config/ptv3.param.yaml`. Model
```
mojomex:
```suggestion
Head and filtering configurations are in `config/ptv3.param.yaml`. Model
```

mojomex (commented):
Thank you for splitting PRs. It's much better to review now, but there's a good amount of unrelated renamings/refactorings and overall code improvements that are separate from the "split backbone and head" change.

I know it's annoying (🙇) but I'd appreciate if you could split into several smaller PRs, one for each separable change. We can review and land those much faster than iterating on a monolithic PR.

amadeuszsz:
@mojomex 

Please take a look once again and let me know if there is anything else we need to do to make this PR ready to review.

perception/autoware_ptv3/README.md
amadeuszsz: Stale, not in current diff

perception/autoware_ptv3/README.md
amadeuszsz: Stale, not in current diff

perception/autoware_ptv3/README.md
amadeuszsz: Stale, not in current diff

perception/autoware_ptv3/include/autoware/ptv3/ptv3_config.hpp
```diff
@@ -43,11 +43,16 @@ class PTv3Config
     const std::vector<float> & voxel_size, const std::vector<std::string> & class_names,
     const std::vector<std::int64_t> & palette, const float filter_class_probability_threshold,
     const std::vector<std::string> & filter_classes, const std::string & filter_output_format,
-    const std::string & source_reconstruction)
+    const std::string & source_reconstruction, const bool use_seg3d_head)
   {
     plugins_path_ = plugins_path;
 
     cloud_capacity_ = cloud_capacity;
+    use_seg3d_head_ = use_seg3d_head;
```
mojomex: (nit) Prefer use of member initializer lists

perception/autoware_ptv3/include/autoware/ptv3/ptv3_config.hpp
```diff
@@ -165,6 +170,9 @@ class PTv3Config
   // TensorRT parameters
   std::string plugins_path_;
 
+  // Head selection
+  bool use_seg3d_head_{true};
```
mojomex: (nit) Don't need to initialize to `true` if member initializer list is used in constructor.

perception/autoware_ptv3/lib/preprocess/preprocess_kernel.cu
```diff
@@ -255,7 +255,7 @@ __global__ void computeGridCoordsAndSerializationKernel(
   const std::int64_t y = static_cast<std::int32_t>(std::floor(point.y / voxel_size_y) - min_y);
   const std::int64_t z = static_cast<std::int32_t>(std::floor(point.z / voxel_size_z) - min_z);
 
-  coords[idx] = make_longlong3(x, y, z);
+  coords[idx] = make_int3(static_cast<int>(x), static_cast<int>(y), static_cast<int>(z));
```
mojomex: x,y,z erroneously still stored as int64, please consider:
```suggestion
const auto x = static_cast<std::int32_t>(std::floor(point.x / voxel_size_x) - min_x);
const auto y = static_cast<std::int32_t>(std::floor(point.y / voxel_size_y) - min_y);
const auto z = static_cast<std::int32_t>(std::floor(point.z / voxel_size_z) - min_z);

  coords[idx] = make_int3(x, y, z);
```

perception/autoware_ptv3/lib/ptv3_trt.cpp
```diff
@@ -439,23 +490,28 @@ bool PTv3TRT::preProcess(const std::shared_ptr<const cuda_blackboard::CudaPointC
...
-  if (!status) {
-    RCLCPP_ERROR(rclcpp::get_logger("ptv3"), "Fail to enqueue and skip to detect.");
+  seg3d_head_trt_ptr_->setInputShape(
+    "point_feat", nvinfer1::Dims{2, {num_voxels_, config_.backbone_feat_dim_}});
+  if (!seg3d_head_trt_ptr_->enqueueV3(stream_)) {
+    RCLCPP_ERROR(rclcpp::get_logger("ptv3"), "Fail to enqueue seg3d head and skip to detect.");
     return false;
   }
 
+  CHECK_CUDA_ERROR(cudaStreamSynchronize(stream_));
```
mojomex: This sync is as far as I can see unnecessary. The above enqueues do error handling correctly already, so any error that occurs has already been read.
If serving no other purpose than error checking, please consider removing.

perception/autoware_ptv3/src/ptv3_node.cpp
```diff
@@ -69,40 +57,72 @@ PTv3Node::PTv3Node(const rclcpp::NodeOptions & options) : Node("ptv3", options)
...
+    seg3d_head_trt_config.emplace(
+      seg3d_head_onnx_path, trt_precision, seg3d_head_engine_path, 1ULL << 30U);
+  }
+
   PTv3Config config(
     plugins_path, cloud_capacity, voxels_num, point_cloud_range, voxel_size, class_names, palette,
-    filter_class_probability_threshold, filter_classes, filter_output_format,
-    source_reconstruction);
+    filter_class_probability_threshold, filter_classes, filter_output_format, source_reconstruction,
+    use_seg3d_head);
```
mojomex, line 93: Consider splitting up this config class: `class_names`, `palette`, `filter_*`, `source_reconstruction` only have meaning if the seg3d head is enabled, but are "dead" otherwise.

mojomex (commented):
Very nice! Thanks for sticking with me with all the PR splitting 🙏 There's not much to criticise at all, I think the main one is the `PTv3Config` that seems like it should be split up.

perception/autoware_ptv3/src/ptv3_node.cpp
amadeuszsz: Overall agree, especially as I already have seen how number of arguments increase while adding detection 3D head. However, this will need to restructure the code which is out of scope of this PR. What if we can apply this refactor in a separate PR, after detection 3D head PR? I would see it more or less this way:
```cpp
common_cfg = PTv3CommonConfig(...);
backbone_cfg = PTv3BackboneConfig(...);
seg3d_cfg = std::optional<PTv3Segmentation3DConfig>(...);
det3d_cfg = std::optional<PTv3Detection3DConfig>(...);
model_ptr_ = std::make_unique<PTv3TRT>(common_cfg, backbone_cfg, seg3d_cfg, det3d_cfg);
```

perception/autoware_ptv3/include/autoware/ptv3/ptv3_config.hpp
amadeuszsz: https://github.com/autowarefoundation/autoware_universe/pull/12655/commits/bf2bf9b337f1f4073e107c3dacd4f8038272a1c0

perception/autoware_ptv3/include/autoware/ptv3/ptv3_config.hpp
amadeuszsz: https://github.com/autowarefoundation/autoware_universe/pull/12655/commits/df7c6bb519a97f52f986092b0ae37db702544728

perception/autoware_ptv3/lib/preprocess/preprocess_kernel.cu
amadeuszsz: https://github.com/autowarefoundation/autoware_universe/pull/12655/commits/8ae44dddb6afc9c5b030ecff90aae05c1a783db1

perception/autoware_ptv3/lib/ptv3_trt.cpp
amadeuszsz: https://github.com/autowarefoundation/autoware_universe/pull/12655/commits/d13ab317b7a27eca8c21f9f0ab2bb7557329384f

mojomex (approved):
LGTM!

Confirmed there is no new overhead between backbone and head in Nsys:
image

Confirmed output with [TIER IV INTERNAL](https://drive.google.com/drive/folders/10NW2sXjPmzok8oGymYazRnI-wdWzSVRc?usp=drive_link) ML package:
```bash
ros2 launch autoware_ptv3 ptv3.launch.xml model_path:="$(pwd)/ml_packages/ptv3_multihead"
```
image

mojomex:
@amadeuszsz  Actually, one question: it looks like both the PTv3 encoder AND decoder are entirely in `backbone`. With the current small model, our decoder has 0 transformer blocks and just unpools/aggregates features across stages, so I suppose that's fine.

However, if we re-introduce transformer blocks in the decoder to e.g. improve semseg performance, that would be unnecessary (or maybe even undesirable) work from the bbox head's view.

Not a blocker right now, so I'll keep this approved. (it also keeps my PR #12727 simple :grin:)

perception/autoware_ptv3/config/ml_package_ptv3_backbone.param.yaml
```diff
@@ -0,0 +1,6 @@
+/**:
+  ros__parameters:
+    backbone:
+      voxels_num: [16000, 128000, 192000]
+      point_cloud_range: [-122.88, -122.88, -3.0, 122.88, 122.88, 5.0]
+      voxel_size: [0.12, 0.12, 0.12]
```
manato, line 6: [ask] Just in case: the default parameters for the backbone do not seem to be identical to those before the modification. Is this intended?

perception/autoware_ptv3/lib/ptv3_trt.cpp
```diff
@@ -439,20 +490,24 @@ bool PTv3TRT::preProcess(const std::shared_ptr<const cuda_blackboard::CudaPointC
...
+  }
 
-  if (!status) {
-    RCLCPP_ERROR(rclcpp::get_logger("ptv3"), "Fail to enqueue and skip to detect.");
+  seg3d_head_trt_ptr_->setInputShape(
+    "point_feat", nvinfer1::Dims{2, {num_voxels_, config_.backbone_feat_dim_}});
+  if (!seg3d_head_trt_ptr_->enqueueV3(stream_)) {
+    RCLCPP_ERROR(rclcpp::get_logger("ptv3"), "Fail to enqueue seg3d head and skip to detect.");
     return false;
   }
```
manato:
```suggestion
if (seg3d_head_trt_ptr_) {
    seg3d_head_trt_ptr_->setInputShape(
      "point_feat", nvinfer1::Dims{2, {num_voxels_, config_.backbone_feat_dim_}});
    if (!seg3d_head_trt_ptr_->enqueueV3(stream_)) {
      RCLCPP_ERROR(rclcpp::get_logger("ptv3"), "Fail to enqueue seg3d head and skip to detect.");
      return false;
    }
  }
```
[imo] Since it is conditional if `seg3d_head_trt_ptr_` holds an instance, I would suggest adding an 'if' guard

manato (commented):
@amadeuszsz 
Thank you very much for your contribution! I left small comments (one is for very just in case to avoid unintentional param chanage), so your consideration is appreciated!

perception/autoware_ptv3/config/ml_package_ptv3_backbone.param.yaml
amadeuszsz: Yes, we will deploy a new model before merging this PR.

perception/autoware_ptv3/lib/ptv3_trt.cpp
amadeuszsz: Right! Can't happen as we throw error if seg3d head is false, but I agree we should put condition here, thanks!

amadeuszsz: Idk why but I can't merge it "Applying suggestions on deleted lines is currently not supported."
Addressed in https://github.com/autowarefoundation/autoware_universe/pull/12655/commits/0b6fb6bffe9194fee4cdc469d095a59b27e7855f

manato (approved):
@amadeuszsz 
Thank you for your prompt handling of my comments. LGTM!

PR #12656, opened 2026-05-29
feat(streampetr): apply mask to input image

tzhong518:
## Description

Implented a masking step for the area of ego vehicle in order to reduce FP caused by reflection.

This changes the distorted image preprocessing order from:
`raw image -> undistort -> resize/crop/normalize`
right: prepocessed image
image
bbox: /perception/object_recognition/detection/streampetr/objects
image

to:
`raw image -> undistort -> mask -> resize/crop/normalize`
image

image
The FP caused by reflection on ego vehicle disappeared.

## Related links

**Parent Issue:**

https://tier4.atlassian.net/browse/T4DEV-52704

autoware_launch PR:
https://github.com/tier4/autoware_launch.x2/pull/2334

## How was this PR tested?

Tested with the issue rosbag: https://tier4.atlassian.net/wiki/spaces/~459732634/pages/5202018574/streampetr_

- The FP disappeared
- The masking preprocess is on GPU, no influence on the average /perception/object_recognition/detection/streampetr/latency/preprocess (tested locally) 
  - before this change: 2.67 ms
  - after this change: 2.66 ms

## Notes for reviewers

The polygon config is set as simple rectangle at present. 

## Interface changes

None.

<!-- ⬇️🔴

#### Additions and removals

None.

#### Modifications

None

#### Additions and removals
| Added | `ego_mask.enabled` | `bool` | `false` | Enable CUDA polygon ego masking in preprocess. |
| Added | `ego_mask.fill_value_bgr` | `double[3]` | `[0.0, 0.0, 0.0]` | BGR fill (0–255) inside mask polygons. |
| Added | `ego_mask.roi_polygons_yaml` | `string[]` | `[]` | One polygon YAML path per model ROI index (`~/input/camera0` …). Empty string disables masking for that ROI. |

#### Modifications

None


YoshiRi:
@tzhong518 Please fix DCO fails.

tzhong518:
I have fixed the DCO fails.
@KSeangTan Could you please review this PR?

perception/autoware_camera_streampetr/config/camera10_polygons.yaml
```diff
@@ -0,0 +1,8 @@
+# Each polygon is [x0, y0, x1, y1, ...] with at least 3 points (6 numbers).
```
KSeangTan: This depends on camera right? Any reason we don't move it to `ml_package_camera_streampetr.param.yaml`? The number of cameras is different, so we need to make the polygon mask config variable in configs.

perception/autoware_camera_streampetr/lib/network/camera_data_store.cpp
```diff
@@ -335,6 +375,65 @@ void CameraDataStore::update_camera_info(
...
+  }
+  if (width <= 0 || height <= 0) {
+    return;
+  }
+  if (is_ego_mask_current(camera_id, width, height)) {
+    return;
+  }
+
+  const auto raster = buildEgoMaskRaster(ego_mask_roi_configs_[camera_id]->polygons, width, height);
+  if (raster.empty()) {
```
KSeangTan: What is the expected return type of `raster`? It does't seem intuitive that we assume `empty()` is available here.

perception/autoware_camera_streampetr/lib/network/camera_data_store.cpp
```diff
@@ -335,6 +375,65 @@ void CameraDataStore::update_camera_info(
...
+    RCLCPP_WARN(logger_, "Empty ego mask raster for camera %d", camera_id);
+    return;
+  }
+
+  upload_ego_mask_gpu(camera_id, raster, width, height);
+
+  RCLCPP_INFO(logger_, "Ego mask GPU buffer built for camera %d (%dx%d)", camera_id, width, height);
+}
+
+bool CameraDataStore::is_ego_mask_current(
```
KSeangTan: Do we need this function if it only calls one line?

perception/autoware_camera_streampetr/lib/network/camera_data_store.cpp
```diff
@@ -335,6 +375,65 @@ void CameraDataStore::update_camera_info(
...
+  const int camera_id, const int width, const int height) const
+{
+  return ego_mask_built_[camera_id] && ego_mask_width_[camera_id] == width &&
+         ego_mask_height_[camera_id] == height;
+}
+
+void CameraDataStore::upload_ego_mask_gpu(
```
KSeangTan: Use `copy` instead of `upload`

perception/autoware_camera_streampetr/lib/network/camera_ego_mask.cpp
```diff
@@ -0,0 +1,265 @@
...
+#include <utility>
+#include <vector>
+
+namespace autoware::camera_streampetr
+{
+
+namespace
+{
+
+std::string loadTextFile(const std::string & path)
```
KSeangTan: `read_config` sounds better

perception/autoware_camera_streampetr/lib/network/camera_ego_mask.cpp
```diff
@@ -0,0 +1,265 @@
...
+  YAML::Node root = YAML::Load(yaml_text);
+  if (!root.IsMap()) {
+    throw std::runtime_error("polygons YAML: root must be a mapping (e.g. 'polygons: [...]').");
+  }
+  return root;
+}
+
+YAML::Node getPolygonsNode(const YAML::Node & root)
+{
+  const YAML::Node polygons = root["polygons"];
```
KSeangTan, line 69: We should move polygon config to ros2 node config and utilize their api instead

KSeangTan (changes requested):
LGTM overall, but we need to make some changes about the structure.

perception/autoware_camera_streampetr/config/camera10_polygons.yaml
tzhong518: I think it's a vehicle param rather than a ml param. Maybe it should be set in individual param.

perception/autoware_camera_streampetr/lib/network/camera_ego_mask.cpp
tzhong518: fixed in https://github.com/autowarefoundation/autoware_universe/pull/12656/changes/3d486fa94ace7cf7d0f9657c130bef64d6e5185a#diff-f3432cf46f5a95586e3c7e5a22cf50bd3d49cdfe5f83fc4b6c9555a71e64be38R218

perception/autoware_camera_streampetr/lib/network/camera_data_store.cpp
tzhong518: fixed in https://github.com/autowarefoundation/autoware_universe/commit/3d486fa94ace7cf7d0f9657c130bef64d6e5185a#diff-48bad7054cfd6763438a5098444305ae7633d56a76a47b59bdbdbe94a795ef96R419

perception/autoware_camera_streampetr/lib/network/camera_data_store.cpp
tzhong518: fixed in https://github.com/autowarefoundation/autoware_universe/commit/3d486fa94ace7cf7d0f9657c130bef64d6e5185a#diff-48bad7054cfd6763438a5098444305ae7633d56a76a47b59bdbdbe94a795ef96R402-R404

perception/autoware_camera_streampetr/lib/network/camera_data_store.cpp
tzhong518: The expected return type is std::vector<std::uint8_t>. Fixed in https://github.com/autowarefoundation/autoware_universe/commit/3d486fa94ace7cf7d0f9657c130bef64d6e5185a#diff-48bad7054cfd6763438a5098444305ae7633d56a76a47b59bdbdbe94a795ef96R409

perception/autoware_camera_streampetr/lib/network/camera_ego_mask.cpp
tzhong518: Moved into camera_streampetr.param.yaml: https://github.com/autowarefoundation/autoware_universe/pull/12656/changes/573a87002493770ec7c0a39d4b3fc475eddf35ee
`<param name="camera_mask.camera_ids" value="[8, 6, 10, 7, 9]"/>` needs to be added to tensorrt_stream_petr.x2.launch.xml as in https://github.com/tier4/autoware_launch.x2/pull/2334

YoshiRi:
@tzhong518 @KSeangTan 
Worked on our bus data

ros2 param visualization
image

actual output 
image

PR #12657, opened 2026-05-29
refactor(traffic_light_multi_camera_fusion): improve readability and maintainability

interimadd:
## Description

Follow-up refactoring of `autoware_traffic_light_multi_camera_fusion` after the core logic was extracted and unit tests were added (#12630). No ROS interfaces, parameters or topics are changed.

The work is split into focused commits:

### 1. Replace the `SignalValidator` class with a free function
`SignalValidator` held only a static `check_conflict` and an unimplemented `merge_partial_match` (dead code).
- Demoted `check_conflict` to a free function in `namespace signal_validator`.
- Dropped the dead `merge_partial_match` declaration.
- Removed the conditionally-created `signal_validator_` `unique_ptr` member, which left a hidden segfault precondition when the consistency check was disabled.

### 2. Move file-local helpers out of the public header
`is_unknown(StateKey)` and `compare_state_key_log_odds` were inline functions in the header but are only used inside the `.cpp`. `convert_output_msg` and `update_best_record` were `static` member functions that touch no member state. All four were demoted into the existing anonymous namespace in the `.cpp`, so the public header now exposes only the types and the class API.

### 3. Rename ambiguous `is_unknown` overloads
Two same-named `is_unknown` functions checked different things (a `StateKey` vs. a `TrafficLight` signal). Renamed to `is_state_key_unknown` and `is_signal_unknown` so the check target is clear from the name alone.

### 4. Replace the three-valued `compare_record` with a boolean predicate
`compare_record` returned `-1 / 0 / 1`, but its only caller used just the sign via `>= 0`. Redesigned into `has_higher_or_equal_priority(candidate, existing)`, which directly answers whether a candidate record should replace the current best one. The fixed priority order (timestamp → recognized-over-unknown → visibility → confidence) and tie handling are preserved.

### 5. Replace `cal_visible_score` (int 0/1) with `is_fully_visible` (bool)
`cal_visible_score` returned `int` `0/1`, which is effectively a boolean. Replaced with the predicate `is_fully_visible(record)` (`true` = not truncated by the image boundary) and simplified the visibility check in `has_higher_or_equal_priority`.

### 6. Refactor the unit tests for readability
- Extracted `make_signal` / `make_record` builder helpers to remove ~20 lines of `FusionRecord` setup duplicated in every test.
- Split multi-assertion tests into single-scenario test cases with intention-revealing names.
- Replaced magic numbers with named constants and the expected results with named values.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

Built and tested with `colcon` on the target package; all tests pass.

```bash
colcon build --packages-select autoware_traffic_light_multi_camera_fusion
colcon test  --packages-select autoware_traffic_light_multi_camera_fusion --event-handlers console_cohesion+
# 100% tests passed, 0 tests failed out of 6 (gtest unit + node tests, copyright, cppcheck, lint_cmake, xmllint)
```

The existing unit tests (`test_utils`, `test_signal_validator`) and the node-level integration tests (`test_multi_camera_fusion`) were updated to follow the renamed/boolean APIs and continue to pass, confirming the refactor is behavior-preserving.

## Notes for reviewers

- This is an internal (non-ROS-interface) refactor. The only API changes are to package-internal helpers declared under `src/` (not installed headers):
  - `compare_record` → `has_higher_or_equal_priority`
  - `cal_visible_score` → `is_fully_visible`
  - `is_unknown` (×2) → `is_state_key_unknown` / `is_signal_unknown`
  - `SignalValidator::check_conflict` → `signal_validator::check_conflict` (free function)
- The large diff in `test/test_utils.cpp` is almost entirely deletion of duplicated setup code.

## Interface changes

None.

## Effects on system behavior

The refactor itself preserves the existing fusion behavior (priority ordering, conflict detection, visibility/truncation handling). The only behavioral change is the bug fix in item 8: an roi that extends past the right or bottom image edge is now correctly classified as truncated (previously it was mis-classified as fully visible due to unsigned underflow), so such a detection is now correctly deprioritized during fusion.

### DIFF
--- a/perception/autoware_traffic_light_multi_camera_fusion/src/multi_camera_fusion.cpp
+++ b/perception/autoware_traffic_light_multi_camera_fusion/src/multi_camera_fusion.cpp
@@ -20,7 +20,6 @@
 #include <algorithm>
 #include <cmath>
 #include <map>
-#include <memory>
 #include <utility>
 #include <vector>
 
@@ -95,9 +94,6 @@ MultiCameraFusion::MultiCameraFusion(const MultiCameraFusionConfig & config)
   traffic_light_id_to_regulatory_ele_id_(
     build_traffic_light_id_to_regulatory_ele_id(config.lanelet_map_ptr))
 {
-  if (config_.use_signal_consistency_check) {
-    signal_validator_ = std::make_unique<SignalValidator>();
-  }
 }
 
 MultiCameraFusionResult MultiCameraFusion::fuse(
@@ -336,7 +332,7 @@ void MultiCameraFusion::determine_best_group_state(
     // check if conflicts exist among the signals within the same regulatory element id
     for (++log_odds_it; log_odds_it != group_info.accumulated_log_odds.end(); ++log_odds_it) {
       const StateKey & competitor_state = (*log_odds_it).first;
-      conflict_result = signal_validator_->check_conflict(running_state, competitor_state);
+      conflict_result = signal_validator::check_conflict(running_state, competitor_state);
       running_state = conflict_result.common_state_key;
 
       if (conflict_result.conflict_type == ConflictType::CONFLICT) {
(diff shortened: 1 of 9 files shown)


a-maumau (approved):
LGTM!

PR #12658, opened 2026-05-29
feat(autoware_tensorrt_plugins): support fused bias/activation in ImplicitGemmPlugin

vividf:
## Description

This PR extends `ImplicitGemmPlugin` to support fused bias and activation in the spconv implicit GEMM path, while also fixing the plugin lifecycle handling between TensorRT build-time and runtime phases.

The main motivation is to reduce extra TensorRT kernels generated after sparse convolution, especially for BN-folded bias/add and activation patterns. When the ONNX graph provides an optional per-channel bias as the 6th input, the plugin now forwards it directly to spconv’s `implicit_gemm()` implementation, allowing the bias and activation to be fused inside the GEMM kernel instead of being executed as separate operations.

## Notes for reviewers
* https://github.com/autowarefoundation/autoware_universe/pull/12631
* https://github.com/autowarefoundation/autoware_universe/pull/12734
### Fused Bias Support

- Added support for an optional 6th plugin input representing a per-channel bias tensor.
  - **5 inputs**: standard implicit GEMM path without bias.
  - **6 inputs**: implicit GEMM path with fused bias.
- Added validation for:
  - bias tensor shape
  - bias channel count
  - bias dtype consistency with activation dtype

### Fused Activation Support

- Added `act_type` to `ImplicitGemmParameters`.

| Value | Activation |
|---------|------------|
| 0 | None |
| 1 | ReLU |
| 2 | Sigmoid |
| 3 | LeakyReLU |

- Added activation conversion helper from serialized integer values to `tv::gemm::Activation`.
- Forward activation parameters (`act_alpha`, `act_beta`, `act_type`) into `ConvGemmOps::implicit_gemm()`.

### Bias Tensor Integration

- Added helper logic to construct a `tv::Tensor` from the optional bias input.
- Ensures the bias tensor uses the same dtype as the activation tensor (`FP16` or `FP32`).
- Avoids unnecessary dtype conversion during inference.

### TensorRT Plugin Lifecycle Fix

- Separated `createPlugin()` handling for:
  - `TensorRTPhase::kBUILD`
  - `TensorRTPhase::kRUNTIME`
- `kBUILD`
  - Parses expanded plugin attributes from ONNX.
- `kRUNTIME`
  - Restores plugin parameters from the serialized plugin state.
- Prevents runtime deserialization from incorrectly expecting ONNX-style plugin fields.

### Serialization Improvements

- Replaced individual field serialization with a packed `ImplicitGemmParameters` blob.
- Added default initialization for all plugin parameters.
- Added serialization support for the new `act_type` field.

### Validation and Debugging Improvements

- Added:
  - `PLUGIN_ASSERT_MSG`
  - `PLUGIN_VALIDATE_MSG`
- Improved error reporting with detailed validation and assertion messages.
- Added a clearer deserialization error message when an incompatible or outdated TensorRT engine is loaded.

## Notes
- ONNX graphs containing BN-folded bias tensors can now leverage fused bias execution inside the sparse convolution GEMM kernel.
- The plugin now correctly follows TensorRT's expected build-time and runtime plugin lifecycle behavior.

## Related links
https://github.com/autowarefoundation/autoware_universe/pull/12631

- Link

## How was this PR tested?
https://drive.google.com/file/d/13-hAjYzrJuKgqYuocvVKMPj9SLCr45CH/view?usp=drive_link

### Validation: BEVFusion ONNX and TensorRT Engine Comparison
The following comparison verifies that the fused implementation produces identical outputs to the original implementation for both ONNX and TensorRT engines.

| Before Fusion | After Fusion |
|---------------|--------------|
| Before Fusion | After Fusion |

### Latency Comparison

| Setting | Preprocess (ms) | Inference (ms) | Postprocess (ms) | Total (ms) | FPS |
|----------|----------:|----------:|----------:|----------:|----------:|
| bevfusion_2_7 | 4.71 | 23.70 | 1.89 | 31.57 | 31.67 |
| bevfusion_2_7_opt | 4.94 | 20.04 (**-15.4%**) | 1.63 | 27.80 | 40.23 |

---

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_tensorrt_plugins/include/autoware/tensorrt_plugins/implicit_gemm_plugin.hpp
+++ b/perception/autoware_tensorrt_plugins/include/autoware/tensorrt_plugins/implicit_gemm_plugin.hpp
@@ -121,15 +121,16 @@ private:
   static constexpr std::int32_t INOUT_PAIR_MASK_FWD_SPLITS_INDEX{3};
   static constexpr std::int32_t INOUT_MASK_ARGSORT_FWD_SPLITS_INDEX{4};
   static constexpr std::int32_t INOUT_OPTIONAL_BIAS_INDEX{5};
-  static constexpr std::int32_t kNumPluginInputsNoBias{5};
-  static constexpr std::int32_t kNumPluginInputsBias{6};
+
+  static constexpr std::int32_t NUM_PLUGIN_INPUTS_NO_BIAS{5};
+  static constexpr std::int32_t NUM_PLUGIN_INPUTS_BIAS{6};
 
   void initFieldsToSerialize();
 
   std::string layer_name_;
   ImplicitGemmParameters params_;
   /// Set in ``configurePlugin`` / ``onShapeChange``.
-  std::int32_t num_plugin_inputs_{kNumPluginInputsNoBias};
+  std::int32_t num_plugin_inputs_{NUM_PLUGIN_INPUTS_NO_BIAS};
   std::tuple<int, int> arch_;
   std::vector<nvinfer1::PluginField> data_to_serialize_;
   nvinfer1::PluginFieldCollection fc_to_serialize_;
--- a/perception/autoware_tensorrt_plugins/include/autoware/tensorrt_plugins/implicit_gemm_plugin.hpp
+++ b/perception/autoware_tensorrt_plugins/include/autoware/tensorrt_plugins/implicit_gemm_plugin.hpp
@@ -43,6 +43,10 @@ struct ImplicitGemmParameters
   std::int32_t is_train{0};
   float output_add_scale{1.0F};
   float output_scale{1.0F};
+
+  /// tv::gemm::Activation as integer:
+  /// kNone=0, kReLU=1, kSigmoid=2, kLeakyReLU=3.
+  std::int32_t act_type{static_cast<std::int32_t>(tv::gemm::Activation::kNone)};
 };
 
 class ImplicitGemmPlugin : public IPluginV3,
@@ -116,12 +120,16 @@ private:
   static constexpr std::int32_t INOUT_PAIR_FWD_INDEX{2};
   static constexpr std::int32_t INOUT_PAIR_MASK_FWD_SPLITS_INDEX{3};
   static constexpr std::int32_t INOUT_MASK_ARGSORT_FWD_SPLITS_INDEX{4};
-  static constexpr std::int32_t INOUT_OUT_FEATURES_INDEX{5};
+  static constexpr std::int32_t INOUT_OPTIONAL_BIAS_INDEX{5};
+  static constexpr std::int32_t kNumPluginInputsNoBias{5};
+  static constexpr std::int32_t kNumPluginInputsBias{6};
 
   void initFieldsToSerialize();
 
   std::string layer_name_;
   ImplicitGemmParameters params_;
+  /// Set in ``configurePlugin`` / ``onShapeChange``.
+  std::int32_t num_plugin_inputs_{kNumPluginInputsNoBias};
   std::tuple<int, int> arch_;
   std::vector<nvinfer1::PluginField> data_to_serialize_;
   nvinfer1::PluginFieldCollection fc_to_serialize_;
--- a/perception/autoware_tensorrt_plugins/src/implicit_gemm_plugin.cpp
+++ b/perception/autoware_tensorrt_plugins/src/implicit_gemm_plugin.cpp
@@ -37,11 +37,12 @@
 namespace
 {
 
-/** Wrap the bias device pointer into a tv::Tensor of the activation dtype (no copy).
+/** Validate the bias inputs, then wrap the bias device pointer into a tv::Tensor of the
+ * activation dtype (no copy).
  *
  * Spconv's fused bias-add needs the bias dtype to match the activation dtype; the ONNX export
  * already emits it that way, so we only assert the match and wrap the pointer. */
-void wrap_bias_tensor(
+void validate_and_wrap_bias_tensor(
   void const * bias_in, std::int64_t c_bias, tv::DType activation_dtype,
   nvinfer1::DataType bias_trt_type, int device, tv::Tensor * bias_out) noexcept
 {
@@ -352,7 +353,7 @@ std::int32_t ImplicitGemmPlugin::enqueue(
   if (num_plugin_inputs_ == NUM_PLUGIN_INPUTS_BIAS) {
     auto const bias_type = input_desc[INOUT_OPTIONAL_BIAS_INDEX].type;
     std::int64_t const c_bias = input_desc[INOUT_OPTIONAL_BIAS_INDEX].dims.d[0];
-    wrap_bias_tensor(
+    validate_and_wrap_bias_tensor(
       inputs[INOUT_OPTIONAL_BIAS_INDEX], c_bias, dtype, bias_type, input_features.device(),
       &bias_tv);
   }
--- a/perception/autoware_tensorrt_plugins/src/implicit_gemm_plugin.cpp
+++ b/perception/autoware_tensorrt_plugins/src/implicit_gemm_plugin.cpp
@@ -37,24 +37,21 @@
 namespace
 {
 
-/** Wrap the optional per-channel bias pointer into a tv::Tensor of the right dtype.
+/** Wrap the bias device pointer into a tv::Tensor of the activation dtype (no copy).
  *
- * Spconv fused bias/add requires the bias dtype to match the activation dtype (see InferenceOps
- * bias_add). This avoids per-enqueue D2H/H2D dtype conversion overhead. */
-void build_bias_tensor_matching_activation(
-  tv::Tensor const & input_features, void const * bias_input, std::int64_t c_bias,
-  tv::DType activation_dtype, nvinfer1::DataType bias_trt_type, tv::Tensor * out_bias) noexcept
+ * Spconv's fused bias-add needs the bias dtype to match the activation dtype; the ONNX export
+ * already emits it that way, so we only assert the match and wrap the pointer. */
+void wrap_bias_tensor(
+  void const * bias_in, std::int64_t c_bias, tv::DType activation_dtype,
+  nvinfer1::DataType bias_trt_type, int device, tv::Tensor * bias_out) noexcept
 {
-  PLUGIN_ASSERT(out_bias != nullptr);
+  PLUGIN_ASSERT(bias_out != nullptr);
+  PLUGIN_ASSERT(bias_in != nullptr);
   PLUGIN_ASSERT(activation_dtype == tv::float16 || activation_dtype == tv::float32);
-  int const dev = input_features.device();
-  if (activation_dtype == tv::float16) {
-    PLUGIN_ASSERT(bias_trt_type == nvinfer1::DataType::kHALF);
-    *out_bias = tv::from_blob(const_cast<void *>(bias_input), {c_bias}, tv::float16, dev);
-    return;
-  }
-  PLUGIN_ASSERT(bias_trt_type == nvinfer1::DataType::kFLOAT);
-  *out_bias = tv::from_blob(const_cast<void *>(bias_input), {c_bias}, tv::float32, dev);
+  PLUGIN_ASSERT(
+    bias_trt_type ==
+    (activation_dtype == tv::float16 ? nvinfer1::DataType::kHALF : nvinfer1::DataType::kFLOAT));
+  *bias_out = tv::from_blob(const_cast<void *>(bias_in), {c_bias}, activation_dtype, device);
 }
 
 tv::gemm::Activation activation_from_int(std::int32_t const v) noexcept
@@ -349,13 +346,15 @@ std::int32_t ImplicitGemmPlugin::enqueue(
 
   // Wrap the optional per-channel bias (BN-folded, 6th ONNX input) into a tv::Tensor.
   // When present, spconv fuses it inside the GEMM kernel instead of a separate bias_add kernel.
-  // bias_tv stays empty (default-constructed) when there are only 5 inputs (no bias).
+  // bias_tv stays empty (default-constructed) when there are only 5 inputs (no bias), in which
+  // case implicit_gemm runs without a bias term.
   tv::Tensor bias_tv{};
   if (num_plugin_inputs_ == NUM_PLUGIN_INPUTS_BIAS) {
     auto const bias_type = input_desc[INOUT_OPTIONAL_BIAS_INDEX].type;
     std::int64_t const c_bias = input_desc[INOUT_OPTIONAL_BIAS_INDEX].dims.d[0];
-    build_bias_tensor_matching_activation(
-      input_features, inputs[INOUT_OPTIONAL_BIAS_INDEX], c_bias, dtype, bias_type, &bias_tv);
+    wrap_bias_tensor(
+      inputs[INOUT_OPTIONAL_BIAS_INDEX], c_bias, dtype, bias_type, input_features.device(),
+      &bias_tv);
   }
 
   std::vector<tv::Tensor> pair_mask_splits;
--- a/perception/autoware_tensorrt_plugins/src/implicit_gemm_plugin.cpp
+++ b/perception/autoware_tensorrt_plugins/src/implicit_gemm_plugin.cpp
@@ -375,8 +375,9 @@ std::int32_t ImplicitGemmPlugin::enqueue(
   [[maybe_unused]] auto const conv_run_status = ConvGemmOps::implicit_gemm(
     alloc2, *tuner_ptr, input_features, weights, pair_fwd, pair_mask_splits, mask_argsort_splits,
     num_act_out, mask_tensor_, arch_, false, params_.is_subm,  // cSpell:ignore subm
-    reinterpret_cast<std::uintptr_t>(stream), tv::CUDAKernelTimer(false), true, false, tv::Tensor(),
-    0.0, 0.0, tv::gemm::Activation::kNone, false, 1.0, tv::Tensor(), tv::Tensor(), 0.0, -1);
+    reinterpret_cast<std::uintptr_t>(stream), tv::CUDAKernelTimer(false), true, false, bias_tv,
+    params_.act_alpha, params_.act_beta, activation_from_int(params_.act_type), false, 1.0,
+    tv::Tensor(), tv::Tensor(), 0.0, -1);
 
   return 0;
 }
--- a/perception/autoware_tensorrt_plugins/src/implicit_gemm_plugin.cpp
+++ b/perception/autoware_tensorrt_plugins/src/implicit_gemm_plugin.cpp
@@ -158,7 +158,7 @@ std::int32_t ImplicitGemmPlugin::configurePlugin(
   // arguments
   PLUGIN_ASSERT(in != nullptr);
   PLUGIN_ASSERT(out != nullptr);
-  PLUGIN_ASSERT(num_inputs == kNumPluginInputsNoBias || num_inputs == kNumPluginInputsBias);
+  PLUGIN_ASSERT(num_inputs == NUM_PLUGIN_INPUTS_NO_BIAS || num_inputs == NUM_PLUGIN_INPUTS_BIAS);
   PLUGIN_ASSERT(num_outputs == 1);
 
   num_plugin_inputs_ = num_inputs;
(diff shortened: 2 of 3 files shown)


perception/autoware_tensorrt_plugins/src/get_indices_pairs_implicit_gemm_plugin_creator.cpp
mojomex, line 67: PR description says this was fixed but it's still here. Please double-check.
IMO, doing this in a separate (non-stacked, independent) PR would be cleaner.

perception/autoware_tensorrt_plugins/include/autoware/tensorrt_plugins/implicit_gemm_plugin.hpp
```diff
@@ -37,12 +37,16 @@ namespace nvinfer1::plugin
...
+  float act_alpha{0.0F};
+  float act_beta{0.0F};
+  std::int32_t is_subm{0};  // cSpell:ignore subm
+  std::int32_t is_train{0};
+  float output_add_scale{1.0F};
+  float output_scale{1.0F};
+
+  /// tv::gemm::Activation as integer:
+  /// kNone=0, kReLU=1, kSigmoid=2, kLeakyReLU=3.
+  std::int32_t act_type{0};
```
mojomex: To make sure the int corresponds to the correct enum value
(haven't tried but similar to below code:)
```suggestion
  std::int32_t act_type{static_cast<int32_t>(tv::gemm::Activation::kNone)};
```

perception/autoware_tensorrt_plugins/include/autoware/tensorrt_plugins/implicit_gemm_plugin.hpp
```diff
@@ -116,12 +120,14 @@ class ImplicitGemmPlugin : public IPluginV3,
...
   static constexpr std::int32_t INOUT_MASK_ARGSORT_FWD_SPLITS_INDEX{4};
-  static constexpr std::int32_t INOUT_OUT_FEATURES_INDEX{5};
+  static constexpr std::int32_t INOUT_OPTIONAL_BIAS_INDEX{5};
 
   void initFieldsToSerialize();
 
   std::string layer_name_;
   ImplicitGemmParameters params_;
+  /// Set in ``configurePlugin`` / ``onShapeChange``: 5 = no bias tensor, 6 = bias at input index 5.
+  std::int32_t num_plugin_inputs_{5};
```
mojomex: Would be nicer as two integer constants `kNumPluginInputsNoBias` and `kNumPluginInputsBias`, instead of only a comment mentioning it.

perception/autoware_tensorrt_plugins/src/implicit_gemm_plugin.cpp
```diff
@@ -34,6 +34,47 @@
...
+  if (activation_dtype == tv::float16) {
+    PLUGIN_ASSERT(bias_trt_type == nvinfer1::DataType::kHALF);
+    *out_bias = tv::from_blob(const_cast<void *>(bias_input), {c_bias}, tv::float16, dev);
+    return;
+  }
+  PLUGIN_ASSERT(bias_trt_type == nvinfer1::DataType::kFLOAT);
+  *out_bias = tv::from_blob(const_cast<void *>(bias_input), {c_bias}, tv::float32, dev);
+}
+
+tv::gemm::Activation activation_from_int(std::int32_t const v) noexcept
```
mojomex, line 58: This can be removed and replaced with `static_cast<tv::gemm::Activation>(v)` (I think).

The error handling is good, so keeping this function, replacing the switch, and bounds checking before is also okay.

perception/autoware_tensorrt_plugins/src/implicit_gemm_plugin.cpp
```diff
@@ -151,33 +193,57 @@ std::int32_t ImplicitGemmPlugin::configurePlugin(
...
     in[INOUT_PAIR_FWD_INDEX].desc.type == in[INOUT_MASK_ARGSORT_FWD_SPLITS_INDEX].desc.type);
 
+  if (num_inputs == 6) {
+    PLUGIN_ASSERT(in[INOUT_OPTIONAL_BIAS_INDEX].desc.dims.nbDims == 1);
+    PLUGIN_ASSERT(
+      in[INOUT_OPTIONAL_BIAS_INDEX].desc.dims.d[0] == in[INOUT_FILTERS_INDEX].desc.dims.d[0]);
+    DataType const bias_t = in[INOUT_OPTIONAL_BIAS_INDEX].desc.type;
+    PLUGIN_ASSERT(bias_t == in[INOUT_IN_FEATURES_INDEX].desc.type);
+  }
+  PLUGIN_ASSERT(params_.act_type >= 0 && params_.act_type <= 3);
```
mojomex: Please replace these numbers with named constants or enum values (in which case they need to be wrapped with `static_cast<int32_t>` or similar.

perception/autoware_tensorrt_plugins/src/implicit_gemm_plugin.cpp
```diff
@@ -151,33 +193,57 @@ std::int32_t ImplicitGemmPlugin::configurePlugin(
...
   std::int32_t num_outputs) noexcept
 {
-  PLUGIN_ASSERT(num_inputs == 5);
+  PLUGIN_ASSERT(in_out != nullptr);
+  PLUGIN_ASSERT(num_inputs == 5 || num_inputs == 6);
   PLUGIN_ASSERT(num_outputs == 1);
+  std::int32_t const n_slots = num_inputs + num_outputs;
+  PLUGIN_ASSERT(pos >= 0 && pos < n_slots);
 
   bool supported = in_out[pos].desc.format == nvinfer1::TensorFormat::kLINEAR;
```
mojomex, line 212: This logic is getting quite complex and there are no comments. What is `in_out`, what determines if supported is true or not? I think it would be best to have one small comment per if statement or case in the switch.

perception/autoware_tensorrt_plugins/src/implicit_gemm_plugin.cpp
```diff
@@ -286,19 +369,21 @@ std::int32_t ImplicitGemmPlugin::enqueue(
 
   auto & tuner_ptr = dtype == tv::float32 ? tuner_fp32_ptr_ : tuner_fp16_ptr_;
 
-  auto conv_run_status = ConvGemmOps::implicit_gemm(
+  [[maybe_unused]] auto const conv_run_status = ConvGemmOps::implicit_gemm(
```
mojomex, line 375: Not sure what this spconv call was supposed to achieve. Prove during build time that the plugin doesn't crash? Or should it pre-warm some internal buffers?

In any case, given the name `conv_run_status`, it looks to me like it should not be ignored with `[[maybe_unused]]` but instead be checked, and if an error status is returned, this function should return an error too.

Doesn't need to be fixed in this PR, but `[[maybe_unused]]` hides the issue instead of "properly" fixing it imho 🙇

perception/autoware_tensorrt_plugins/src/implicit_gemm_plugin_creator.cpp
```diff
@@ -55,79 +56,58 @@ nvinfer1::PluginFieldCollection const * ImplicitGemmPluginCreator::getFieldNames
...
+        PLUGIN_VALIDATE(type == nvinfer1::PluginFieldType::kINT32);
+        parameters.act_type = static_cast<std::int32_t const *>(fields[i].data)[0];
+        PLUGIN_VALIDATE(parameters.act_type >= 0 && parameters.act_type <= 3);
+      }
+    }
+  };
+
   // The build phase and the deserialization phase are handled differently.
-  if (phase == TensorRTPhase::kBUILD || phase == TensorRTPhase::kRUNTIME) {
+  if (phase == TensorRTPhase::kBUILD) {
```
mojomex, line 95: Ideally, this would be done in a separate `fix(autoware_tensorrt_plugins):` PR, since it should be tested on its own to ensure nothing broke.

perception/autoware_tensorrt_plugins/src/implicit_gemm_plugin.cpp
mojomex: Hmmmm, the SpConv examples contain the same piece of code: 

https://github.com/traveller59/spconv/blob/263d6b47425ef843c82f997b12d8b714013d216c/example/libspconv/main.cu#L436

Maybe not important after all....

perception/autoware_tensorrt_plugins/src/plugin_utils.cpp
```diff
@@ -20,6 +20,33 @@
...
+  }
+  stream << file << ':' << line << std::endl << "Aborting..." << std::endl;
+  getLogger()->log(nvinfer1::ILogger::Severity::kINTERNAL_ERROR, stream.str().c_str());
+}
+
+void logValidationFailure(
+  char const * msg, char const * detail, char const * file, std::int32_t line)
+{
+  std::ostringstream stream;
+  stream << "Validation failed: " << msg << std::endl;
```
mojomex: `std::endl` causes a stream flush() which takes unnecessary time. The more modern and efficient way is to just `<< '\n'`. (applies to other callsites as well)

mojomex (commented):
Thank you! Core funcitonality looks good (didn't test yet). I left some style comments, please have a look.

perception/autoware_tensorrt_plugins/include/autoware/tensorrt_plugins/implicit_gemm_plugin.hpp
vividf: Thanks, addressed in https://github.com/autowarefoundation/autoware_universe/pull/12658/commits/a88e09a17498d8507874e442a2b6fb01f90ae820

perception/autoware_tensorrt_plugins/include/autoware/tensorrt_plugins/implicit_gemm_plugin.hpp
vividf: Thanks, addressed in https://github.com/autowarefoundation/autoware_universe/pull/12658/commits/a88e09a17498d8507874e442a2b6fb01f90ae820

perception/autoware_tensorrt_plugins/src/implicit_gemm_plugin.cpp
vividf: Thanks, addressed in https://github.com/autowarefoundation/autoware_universe/pull/12658/commits/a88e09a17498d8507874e442a2b6fb01f90ae820

perception/autoware_tensorrt_plugins/src/implicit_gemm_plugin.cpp
vividf: Thanks, addressed in https://github.com/autowarefoundation/autoware_universe/pull/12658/commits/a88e09a17498d8507874e442a2b6fb01f90ae820

perception/autoware_tensorrt_plugins/src/implicit_gemm_plugin.cpp
vividf: Thanks, addressed in https://github.com/autowarefoundation/autoware_universe/pull/12658/commits/a88e09a17498d8507874e442a2b6fb01f90ae820

perception/autoware_tensorrt_plugins/src/plugin_utils.cpp
vividf: Thanks, addressed in https://github.com/autowarefoundation/autoware_universe/pull/12658/commits/a88e09a17498d8507874e442a2b6fb01f90ae820

perception/autoware_tensorrt_plugins/src/implicit_gemm_plugin_creator.cpp
vividf: Created another PR: https://github.com/autowarefoundation/autoware_universe/pull/12734

perception/autoware_tensorrt_plugins/src/implicit_gemm_plugin.cpp
```diff
@@ -34,6 +34,38 @@
...
+namespace
+{
+
+/** Wrap the optional per-channel bias pointer into a tv::Tensor of the right dtype.
+ *
+ * Spconv fused bias/add requires the bias dtype to match the activation dtype (see InferenceOps
+ * bias_add). This avoids per-enqueue D2H/H2D dtype conversion overhead. */
+void build_bias_tensor_matching_activation(
+  tv::Tensor const & input_features, void const * bias_input, std::int64_t c_bias,
+  tv::DType activation_dtype, nvinfer1::DataType bias_trt_type, tv::Tensor * out_bias) noexcept
```
mojomex: `input_`, `_input`, and `out_` are mixed. I would suggest the C-style convention of `_in` and `_out` for all of these.

perception/autoware_tensorrt_plugins/src/implicit_gemm_plugin.cpp
```diff
@@ -34,6 +34,38 @@
...
+{
+  PLUGIN_ASSERT(out_bias != nullptr);
```
mojomex: `bias_input` is left unchecked, please add an assert

perception/autoware_tensorrt_plugins/src/implicit_gemm_plugin.cpp
```diff
@@ -34,6 +34,38 @@
...
+  PLUGIN_ASSERT(activation_dtype == tv::float16 || activation_dtype == tv::float32);
+  int const dev = input_features.device();
+  if (activation_dtype == tv::float16) {
+    PLUGIN_ASSERT(bias_trt_type == nvinfer1::DataType::kHALF);
+    *out_bias = tv::from_blob(const_cast<void *>(bias_input), {c_bias}, tv::float16, dev);
+    return;
+  }
+  PLUGIN_ASSERT(bias_trt_type == nvinfer1::DataType::kFLOAT);
+  *out_bias = tv::from_blob(const_cast<void *>(bias_input), {c_bias}, tv::float32, dev);
+}
```
mojomex: I'm not quite sure what this function achieves:

1. If activation is FP16 && bias is FP16 -> copy FP16 bias_in to FP16 bias_out
2. If activation is FP32 && bias is FP32 -> copy FP32 bias_in to FP32 bias_out
3. If the two types don't match: error.

So, if this function succeeds, the tensors were already both of the same type, and an unnecessary copy is made.

If the two types don't match, this function will not make them match, it will fail an assertion.

perception/autoware_tensorrt_plugins/src/implicit_gemm_plugin.cpp
```diff
@@ -272,6 +347,17 @@ std::int32_t ImplicitGemmPlugin::enqueue(
...
+  // Wrap the optional per-channel bias (BN-folded, 6th ONNX input) into a tv::Tensor.
+  // When present, spconv fuses it inside the GEMM kernel instead of a separate bias_add kernel.
+  // bias_tv stays empty (default-constructed) when there are only 5 inputs (no bias).
+  tv::Tensor bias_tv{};
+  if (num_plugin_inputs_ == NUM_PLUGIN_INPUTS_BIAS) {
+    auto const bias_type = input_desc[INOUT_OPTIONAL_BIAS_INDEX].type;
+    std::int64_t const c_bias = input_desc[INOUT_OPTIONAL_BIAS_INDEX].dims.d[0];
+    build_bias_tensor_matching_activation(
+      input_features, inputs[INOUT_OPTIONAL_BIAS_INDEX], c_bias, dtype, bias_type, &bias_tv);
+  }
```
mojomex, line 359: Pretty sure this does nothing, see above comment.

mojomex (commented):
Almost there! There's one function I'd like you to confirm.

perception/autoware_tensorrt_plugins/src/implicit_gemm_plugin.cpp
vividf: `from_blob` only wraps the pointer (no copy), and the call is needed to populate bias_tv for the implicit_gemm API (since it requires the bias type to be `tv::Tensor`)

mojomex: I see. In that case the function seems fine, but could I ask you to rename it, e.g. to `type_check_and_wrap_bias_tensor` or similar?

perception/autoware_tensorrt_plugins/src/implicit_gemm_plugin.cpp
vividf: Thanks! addressed in https://github.com/autowarefoundation/autoware_universe/pull/12658/changes/f29bcdbda361c68b722904c6270dfbb17ec240e7#diff-faa2857fbaf93929028db177a1be6dd95d78af06f9573319f931ecdbe5f015ec

perception/autoware_tensorrt_plugins/src/implicit_gemm_plugin.cpp
vividf: Thanks! addressed in https://github.com/autowarefoundation/autoware_universe/pull/12658/changes/f29bcdbda361c68b722904c6270dfbb17ec240e7#diff-faa2857fbaf93929028db177a1be6dd95d78af06f9573319f931ecdbe5f015ec

perception/autoware_tensorrt_plugins/src/implicit_gemm_plugin.cpp
vividf: Thanks! addressed in https://github.com/autowarefoundation/autoware_universe/pull/12658/changes/f29bcdbda361c68b722904c6270dfbb17ec240e7#diff-faa2857fbaf93929028db177a1be6dd95d78af06f9573319f931ecdbe5f015ec

perception/autoware_tensorrt_plugins/src/implicit_gemm_plugin.cpp
```diff
@@ -349,13 +346,15 @@ std::int32_t ImplicitGemmPlugin::enqueue(
...
-  // bias_tv stays empty (default-constructed) when there are only 5 inputs (no bias).
+  // bias_tv stays empty (default-constructed) when there are only 5 inputs (no bias), in which
+  // case implicit_gemm runs without a bias term.
   tv::Tensor bias_tv{};
   if (num_plugin_inputs_ == NUM_PLUGIN_INPUTS_BIAS) {
     auto const bias_type = input_desc[INOUT_OPTIONAL_BIAS_INDEX].type;
     std::int64_t const c_bias = input_desc[INOUT_OPTIONAL_BIAS_INDEX].dims.d[0];
-    build_bias_tensor_matching_activation(
-      input_features, inputs[INOUT_OPTIONAL_BIAS_INDEX], c_bias, dtype, bias_type, &bias_tv);
+    wrap_bias_tensor(
```
mojomex: Thanks for renaming! Logic+comments look good, but I'd like the `validate` or `type_check` part reflected in the name (`wrap` implies that it doesn't check, and cannot fail).

mojomex (commented):
Happy with everything but the one function name 🙇

perception/autoware_tensorrt_plugins/src/implicit_gemm_plugin.cpp
vividf: Thanks! fixed in https://github.com/autowarefoundation/autoware_universe/pull/12658/commits/b5f2c545a91b9bd97c3506d0c3b41d430b405d47

mojomex: 🫶

mojomex:
@vividf ~Do you have a fused-bias ONNX for me to confirm with?~
~Ah, found [TIER IV INTERNAL](nevermind)~
Ah, saw your comment.

vividf:
@mojomex [Internal link](https://drive.google.com/file/d/1cuGSnyBu_UdF5nPx6sUg-WcioXtQfFLW/view?usp=drive_link)

mojomex (approved):
Confirmed visually with the above ONNX:

https://github.com/user-attachments/assets/e15321a5-e06d-45fe-ba2c-477a121416ca

Confirmed that Nsys still looks ok:

Image

I didn't perform in-depth analysis but visually, both look good to me.

vividf:
@amadeuszsz @ktro2828 kingly ping, thanks

ktro2828:
@vividf It seems there are conflicts could you fix them?

vividf:
@ktro2828 
Thanks, I rebased the branch

ktro2828 (approved):
LGTM

PR #12660, opened 2026-05-29
refactor(autoware_traffic_light_arbiter): extract TrafficLightArbiterCore from Node

takam5f2:
## Description

Separate arbitration logic from the ROS node in `autoware_traffic_light_arbiter`. The new `TrafficLightArbiterCore` class owns all arbitration state and decisions and has no `rclcpp` dependency in its function bodies. `TrafficLightArbiter` (the ROS node) becomes a thin adapter: parameter loading, sub/pub wiring, msg ↔ pure-data conversion, publish, and logging. Observable behavior is preserved — all existing tests (`test_node.cpp` + `test_traffic_arbiter_characteristic.cpp`, 31 cases total) pass unchanged.

### Features

- **`TrafficLightArbiterCore` class** — ROS-free in body. Owns arbitration state (map ids, pedestrian validator, perception buffer, external signal cache) and exposes `set_map`, `ingest_perception`, `ingest_external`, `arbitrate`
- **`TrafficLightArbiter` (the Node) reduced to a thin adapter** — covers only param loading, sub/pub wiring, message conversion, publish, and logging
- **Diagnostics flow as return-then-log** — Core returns `ArbitrationResult` (with `output`, `off_map_signal_ids`, `latest_input_time`) and `vector<ExpiredExternalSignal>` (from `ingest_perception` / `ingest_external`); the Node consumes these and emits the corresponding `RCLCPP_WARN_*` / `RCLCPP_DEBUG`
- **`arbitrate()` no longer takes a `stamp` argument** — output stamping (an I/O convention) and staleness detection (a monitoring concern) are both moved to the Node, where they belong. Core's `latest_input_time` field supplies the raw datum the Node needs for the staleness comparison
- **`set_map(const lanelet::LaneletMapConstPtr &)`** — Core takes the parsed LaneletMap and extracts the regulatory-element ID subsets it consumes (vehicle traffic lights, plus pedestrian traffic lights when signal matching is enabled) internally. The Node only parses the `LaneletMapBin` msg and forwards the resulting map; it doesn't need to know which subsets the arbiter cares about.

### Architecture

The Node ↔ Core split:

| Concern | Where |
| --- | --- |
| Param load / sub-pub wiring / `LaneletMapBin` → `LaneletMap` parsing / stamp inheritance / logging | `TrafficLightArbiter` (Node) |
| Arbitration state (map id set extracted from the parsed `LaneletMap`, validator, perception buffer, external cache) | `TrafficLightArbiterCore` |
| `ingest_perception` / `ingest_external` / `arbitrate` | `TrafficLightArbiterCore` |
| Matching / priority routing + per-shape highest-confidence selection | `TrafficLightArbiterCore::arbitrate` |

## Related links

**Parent Issue:**

- None

**Built on:**

- #12632 (`test(autoware_traffic_light_arbiter): add characterization test suite`) — its 24 characterization tests serve as the safety net for this refactor; already merged.

## How was this PR tested?

`colcon build` + `colcon test` against the package. All 31 tests pass without test-side changes: `_test` (7/7) + `_characteristic_test` (24/24), plus 4 lint checks. `RCLCPP_COMPONENTS_REGISTER_NODE` is preserved so the node still loads as a composable component.

### Design choices worth flagging

- `ArbitrationResult.output` is `std::optional<TrafficSignalArray>` so "no map received yet" (`nullopt`) and "map received, no traffic lights" (empty `traffic_light_groups`) can be distinguished. The Node uses this: nullopt → suppress publish + WARN; empty value → publish empty msg silently.
- Core methods that mutate state (`ingest_*`, `set_*`) are non-const; `arbitrate()` is currently non-const because it calls `signal_match_validator_->validate_signals(...)` which is non-const. Marking validator const is queued as a follow-up.

## Interface changes

None. Topics and all five parameters (`external_delay_tolerance`, `external_time_tolerance`, `perception_time_tolerance`, `source_priority`, `enable_signal_matching`) are unchanged in name, type, default, and effect.

## Effects on system behavior

**Downstream-observable**: None. Publish content, timing, and emit count are preserved; parameter handling and log message contents are identical.

**Internal-observable (not affecting downstream)**:

- `RCLCPP_WARN_THROTTLE` macro call sites have moved (editor line numbers change), but startup warning behavior is unchanged.
- DEBUG cleanup log lines flush from a Node-side loop after Core's `ingest_perception` / `ingest_external` return; lag is sub-tick.

### DIFF
--- a/perception/autoware_traffic_light_arbiter/src/traffic_light_arbiter_core.cpp
+++ b/perception/autoware_traffic_light_arbiter/src/traffic_light_arbiter_core.cpp
@@ -55,7 +55,7 @@ void TrafficLightArbiterCore::set_pedestrian_traffic_light_ids(std::unordered_se
 bool TrafficLightArbiterCore::is_external_outdated(
   const rclcpp::Time & current_time, const rclcpp::Time & msg_stamp) const
 {
-  return std::abs((current_time - msg_stamp).seconds()) > external_delay_tolerance_;
+  return (current_time - msg_stamp).seconds() > external_delay_tolerance_;
 }
 
 std::vector<TrafficLightArbiterCore::ExpiredExternalSignal>
@@ -67,7 +67,7 @@ TrafficLightArbiterCore::sweep_expired_external_signals(
   while (it != external_traffic_lights_.end()) {
     const auto & msg_stamp = it->second.first;
     const auto age = (reference_time - msg_stamp).seconds();
-    if (std::abs(age) > tolerance) {
+    if (age > tolerance) {
       expired.push_back({it->first, age});
       it = external_traffic_lights_.erase(it);
     } else {
--- a/perception/autoware_traffic_light_arbiter/src/traffic_light_arbiter_core.cpp
+++ b/perception/autoware_traffic_light_arbiter/src/traffic_light_arbiter_core.cpp
@@ -84,25 +84,27 @@ TrafficLightArbiterCore::ingest_perception(const TrafficSignalArray & msg)
   return sweep_expired_external_signals(rclcpp::Time(msg.stamp), external_time_tolerance_);
 }
 
-std::vector<TrafficLightArbiterCore::DroppedExternalSignal>
-TrafficLightArbiterCore::ingest_external(
+TrafficLightArbiterCore::ExternalIngestResult TrafficLightArbiterCore::ingest_external(
   const TrafficSignalArray & msg, const rclcpp::Time & current_time)
 {
+  const auto msg_time = rclcpp::Time(msg.stamp);
+  if (is_external_outdated(current_time, msg_time)) {
+    return {false, {}};
+  }
+
   // Update external traffic lights map with new information
   for (const auto & signal : msg.traffic_light_groups) {
-    external_traffic_lights_[signal.traffic_light_group_id] =
-      std::make_pair(rclcpp::Time(msg.stamp), signal);
+    external_traffic_lights_[signal.traffic_light_group_id] = std::make_pair(msg_time, signal);
   }
 
   // Clear perception data if too old relative to this external signal
-  const auto msg_time = rclcpp::Time(msg.stamp);
   if (
     std::abs((msg_time - rclcpp::Time(latest_perception_msg_.stamp)).seconds()) >
     perception_time_tolerance_) {
     latest_perception_msg_.traffic_light_groups.clear();
   }
 
-  return sweep_expired_external_signals(current_time, external_delay_tolerance_);
+  return {true, sweep_expired_external_signals(current_time, external_delay_tolerance_)};
 }
 
 TrafficLightArbiterCore::ArbitrationResult TrafficLightArbiterCore::arbitrate()
--- a/perception/autoware_traffic_light_arbiter/src/traffic_light_arbiter_core.cpp
+++ b/perception/autoware_traffic_light_arbiter/src/traffic_light_arbiter_core.cpp
@@ -35,8 +35,7 @@ TrafficLightArbiterCore::TrafficLightArbiterCore(
   perception_time_tolerance_(perception_time_tolerance)
 {
   if (enable_signal_matching_) {
-    signal_match_validator_ = std::make_unique<SignalMatchValidator>();
-    signal_match_validator_->set_source_priority(source_priority_);
+    signal_match_validator_ = std::make_unique<SignalMatchValidator>(source_priority_);
   }
 }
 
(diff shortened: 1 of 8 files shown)


takam5f2:
@a-maumau 

Sorry to ask again right after the last one — could you take a look at this PR as well?
Existing tests pass without any test-side change; this is purely a structural restructure only to separate the logic and the node wrapper.

Review focus areas:

* **Architecture table (in the description)** : is the Node/Core seam drawn at the right granularity? Core depends on rclcpp::Time but stays free of rclcpp::Logger / *_THROTTLE macros — I'd value your read on whether that's the right line to draw.
* **Return-then-log diagnostics**: Core returns diagnostic data (off-map IDs, dropped external entries, latest_input_time) and the Node logs from those returns. Cleaner separation at the cost of one indirection.
* **Design choices (3 bullets at the bottom)**: deliberate decisions left for follow-up PRs — external_time_tolerance lingering on Core, ArbitrationResult.output as optional to distinguish two empty states, and deferred const-ness on arbitrate(). Your perspective on those would help guide the next round.

takam5f2:
@a-maumau  
I'd like to add some simple change so that I convert the PR to draft mode.

takam5f2:
@interimadd  
Could you check the PR before asking the code owners to review that?

interimadd (approved):
LGTM for me

perception/autoware_traffic_light_arbiter/src/traffic_light_arbiter_core.cpp
```diff
@@ -0,0 +1,278 @@
...
+#include <utility>
+#include <vector>
+
+namespace autoware::traffic_light
+{
+
+namespace
+{
+
+std::unordered_set<lanelet::Id> extract_traffic_light_ids(const lanelet::LaneletMapConstPtr & map)
```
interimadd, line 36: This function seems to be ported from the following function.

```cpp
std::vector<TrafficLightConstPtr> filter_traffic_signals(const LaneletMapConstPtr map)
{
  std::vector<TrafficLightConstPtr> signals;
  for (const auto & element : map->regulatoryElementLayer) {
    const auto signal = std::dynamic_pointer_cast<const TrafficLight>(element);
    if (signal) {
      signals.push_back(signal);
    }
  }

  return signals;
}
```

I recommend keeping the implementation the same because this PR's objective is just to extract the core logic.

perception/autoware_traffic_light_arbiter/src/traffic_light_arbiter_core.cpp
```diff
@@ -0,0 +1,278 @@
...
+  if (enable_signal_matching_) {
+    signal_match_validator_->set_pedestrian_traffic_light_ids(
+      extract_pedestrian_traffic_light_ids(map));
+  }
+}
+
+bool TrafficLightArbiterCore::is_external_outdated(
+  const rclcpp::Time & current_time, const rclcpp::Time & msg_stamp) const
+{
+  return (current_time - msg_stamp).seconds() > external_delay_tolerance_;
```
interimadd: Before change, this condition seems to be `std::abs((current_time - msg_stamp).seconds()) > external_delay_tolerance_;
Is this change intended?

interimadd (commented):
Sorry I have noticed some changes related to behavior change.
Could you check if they are intended changes?

perception/autoware_traffic_light_arbiter/src/traffic_light_arbiter_core.cpp
takam5f2: Thanks for the catch. Restored the original `regulatoryElementLayer` + `std::dynamic_pointer_cast` iteration in 55b6c81. Agreed that the typed-accessor swap was out of scope for this PR — it also has a subtly different semantic (`regulatoryElementsAs<>` on each lanelet skips `TrafficLight` regulatory elements not referenced by any lanelet, while the original layer walk picks them up too), so reverting preserves the pre-refactor behavior exactly.

perception/autoware_traffic_light_arbiter/src/traffic_light_arbiter_core.cpp
takam5f2: Good catch — restored in 6ac194e. The `std::abs` removal was an attempt to align the comparison with the function names (`outdated` / `expired` implying one direction), but you're right that any direction-aware tightening belongs in a follow-up PR rather than mixed into the Core extraction. The symmetric check is back, matching the pre-refactor behavior exactly (future-dated msgs are again rejected at admission and evicted at sweep).

takam5f2:
@a-maumau  Sorry for confusing, but could you review the PR?

a-maumau (approved):
LGTM!

PR #12664, opened 2026-05-29
feat(multi_object_tracker): apply `agnocast_wrapper::Node`

Koichi98:
## Description
Apply `autoware_agnocast_wrapper` to `autoware_multi_object_tracker` so the node can run on either the ROS 2 or the Agnocast (zero-copy IPC) backend, selected at runtime via the `ENABLE_AGNOCAST` environment variable.

  Also, please refer to https://github.com/autowarefoundation/autoware_core/blob/main/common/autoware_agnocast_wrapper/docs/review_guide.md for the review guide of `autoware_agnocast_wrapper` application PRs.

  ### Changes
  - Migrate `MultiObjectTracker` to inherit `autoware::agnocast_wrapper::Node` and register it with `AgnocastOnlyCallbackIsolatedExecutor` (ROS 2 fallback: `SingleThreadedExecutor`) via `autoware_agnocast_wrapper_register_node` in `CMakeLists.txt`.
  - Route the subscriptions, publishers and the publish timer through the wrapper macros (`AUTOWARE_SUBSCRIPTION_PTR` / `AUTOWARE_PUBLISHER_PTR` / `AUTOWARE_TIMER_PTR` / `autoware::agnocast_wrapper::create_timer`).
 
  ## Related Links

**Parent Issue:**

- Link

## How was this PR tested?
  - Ran an Autoware-based product with this change applied and confirmed it operates without any regression.

### Only synchronous TF lookups are used
 `Odometry` only does synchronous `canTransform` / `lookupTransform`. The `TransformListener` spins a dedicated thread populating the buffer (`setUsingDedicatedThread(true)`), and the timeout is pure clock polling — no timer entity. So `waitForTransform` and the `CreateTimerInterface` / `setCreateTimerInterface` boilerplate are unnecessary (and wouldn't fire under `AgnocastOnly` anyway); `agnocast::Buffer`
  intentionally omits that async path.

## Interface changes

None.

## Effects on system behavior


technolojin:
we have high priority works in the package. I would like to request this PR after https://github.com/autowarefoundation/autoware_universe/pull/12672

technolojin (approved):
tested with sample ros bag (in case of agnocast disabled)

PR #12670, opened 2026-05-31
fix(aeb, collision_detector): skip empty point clouds to silence PCL warning spam

mertyavuz41:
## Description
fixing similar spamming message problem

## Related links

**Parent Issue:**

- Link: https://github.com/autowarefoundation/autoware/issues/6843#issuecomment-4213412231

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


control/autoware_autonomous_emergency_braking/src/node.cpp
```diff
@@ -964,6 +964,13 @@ void AEB::cropPointCloudWithEgoFootprintPath(
   const std::vector<Polygon2d> & ego_polys, PointCloud::Ptr filtered_objects)
 {
   autoware_utils::ScopedTimeTrack st(__func__, *time_keeper_);
+  if (obstacle_ros_pointcloud_ptr_->width == 0 || obstacle_ros_pointcloud_ptr_->height == 0) {
+    RCLCPP_DEBUG_SKIPFIRST_THROTTLE(
+      get_logger(), *get_clock(), 5000,
+      "[AEB]: Cropping point cloud with ego footprint path. Number of ego polygons: %zu",
+      ego_polys.size());
```
xmfcx: The text says "Cropping point cloud..." but this branch actually *skips* the crop because the cloud is empty, and `ego_polys.size()` isn't relevant to the bail reason. Suggest matching the wording the empty-cloud guard already uses in `onPointCloud` (`"[AEB]: Received empty point cloud"`):

```suggestion
    RCLCPP_DEBUG_SKIPFIRST_THROTTLE(
      get_logger(), *get_clock(), 5000,
      "[AEB]: Received empty obstacle point cloud, skipping crop");
```

xmfcx (changes requested):
Thanks @mertyavuz41! Logic, null-safety and return semantics all check out. One thing to fix before merge: the log message below. Otherwise good to go.

xmfcx (approved):

PR #12672, opened 2026-06-01
feat(autoware_multi_object_tracker): shape-aware tracker spawning and association

technolojin:
## Description

Detection inputs carry a shape type — bounding box, cylinder, or polygon — that reflects how the upstream detector represents an object. Previously the tracker discarded this information: polygon-shaped detections were force-converted to bounding boxes in `InputStream`, and tracker selection (both spawning and association) was keyed only on semantic label.

This PR preserves the detection shape through the pipeline and uses it together with the label to select the right tracker.

---

### New: ShapeType enum (`types.hpp`)

Added `ShapeType` (BOUNDING_BOX, CYLINDER, POLYGON) mirroring the constants in `autoware_perception_msgs::msg::Shape`, with `toString` / `toShapeType` converters and `ALL_SHAPE_TYPES` array.

### Polygon detections are no longer converted to bounding boxes (`input_manager.cpp`)

The block that called `convertConvexHullToBoundingBox` for polygon-shaped objects is removed. Cylinders still get their Y dimension normalized to match X (unchanged behavior), but polygons now pass through as-is.

### Unified tracker assignment configuration

The two separate config concepts — `initial_tracker` (which tracker to spawn) and `association.can_assign` (which trackers to match against) — are merged into a single `tracker_assignment` block, organized by shape then label:

```yaml
tracker_assignment:
  bounding_box:
    car: {create: multi_vehicle_tracker, match: [multi_vehicle_tracker]}
    # ... all eight labels
  polygon:
    car: {create: polygon_tracker, match: [polygon_tracker, multi_vehicle_tracker]}
    # ... all eight labels
  cylinder:
    car: {create: "null"}   # explicitly rejected — no tracker created or associated
    # ... all eight labels
```

All three shapes (`bounding_box`, `polygon`, `cylinder`) must be listed explicitly; omitted `(shape, label)` pairs are implicitly not accepted and emit a throttled runtime error on first encounter. `create: "null"` silently rejects a detection without creating a tracker.

Association thresholds move from a flat four-table layout (`association.max_dist.<label>.<tracker>` etc.) to a `tracker_profiles` block keyed by tracker type, then shape, then label:

```yaml
tracker_profiles:
  multi_vehicle_tracker:
    bounding_box:
      car: {max_dist: 5.0, max_area: 12.10, min_area: 3.6, min_iou: 0.0001}
    polygon:
      car: {max_dist: 4.0, max_area: 60.0, min_area: 1.6, min_iou: 0.0001}
```

Every `(tracker, shape, label)` triple referenced in any `tracker_assignment.<shape>.<label>.match` list must have an explicit profile entry; a missing entry throws at startup with a clear message.

### Internal config structs

`AssociatorConfig` is split and renamed:

- `AssociationProfile` — flat struct holding the four thresholds (`max_dist_sq`, `max_area`, `min_area`, `min_iou`). Promoted to top level.
- `TrackerAssociationConfig` — replaces `AssociatorConfig`. The association map is now keyed by `ShapeLabelKey` (pair of ShapeType + Label) instead of Label alone. Provides `setProfile()` and `buildMaxDistances()` helpers.
- `TrackerCreationConfig::tracker_map` renamed to `shape_tracker_map`, keyed by `ShapeLabelKey`. Provides `setCreation()` helper.
- `TrackedLabelThresholds` helper struct removed; replaced by a generic `parse_label_double_map` lambda in the node.
- `EnumClassHash` promoted from nested to top-level.

### Parameter loading (`multi_object_tracker_node.cpp`)

A helper `parseTrackerType()` extracts the string-to-enum conversion. The constructor iterates over all `(shape, label)` combinations in a single pass and reads `tracker_assignment.<shape>.<label>.{create, match}` directly. Entries absent from the parameter file are implicitly skipped (a throttled runtime error is emitted on first use); `create: "null"` entries are recorded as explicit rejections.

Association profiles are loaded from `tracker_profiles.<tracker>.<shape>.<label>` for every `(tracker, shape, label)` triple referenced in a `match` list. A missing profile throws at startup rather than silently using a default. `buildMaxDistances()` precomputes per-label maximum search distances once at startup and stores them on the config.

### Association logic changes

`BevAssociation` and `PolarAssociation` now build the `ShapeLabelKey` from the measurement's shape type before looking up the profile map. The `updateMaxSearchDistances()` method that was private to `BevAssociation` is replaced by `TrackerAssociationConfig::buildMaxDistances()`, called once at startup; the per-label max-distance table is now stored on the config and shared.

### Processor and creation

`TrackerProcessor::createNewTracker()` looks up `shape_tracker_map[{shape_type, label}]` instead of `tracker_map[label]`, so the spawned tracker type depends on detection shape.

### Schema and config files

`data_association_matrix.schema.json` is updated to validate the new `tracker_assignment` / `tracker_profiles` structure with typed `$ref` definitions, including the three-level `tracker_profiles.<tracker>.<shape>.<label>` hierarchy. `multi_object_tracker_node.schema.json` removes the `initial_tracker` block and adds `diagnostics_warn_extrapolation` / `diagnostics_error_extrapolation` to the required list.

`simulation_tracker.param.yaml` and `simulation_tracker.schema.json` are removed — the simulation tracker relied exclusively on `pass_through_tracker`, which is deleted in this PR.

### Remove pass_through_tracker

Since the pass through tracker is developed https://github.com/autowarefoundation/autoware_universe/issues/655, it was not used.

Remove unused tracker to reduce maintenance cost.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

* Before - cluster of truck converted to very-short bounding box and tracked by vehicle tracker

https://github.com/user-attachments/assets/35a20fa1-ddb7-40b1-a228-d88ed68da3fd

* After - cluster of truck preserved and tracked by the polygon tracker

https://github.com/user-attachments/assets/f1e6a021-775a-4bbb-8e34-eff8809d2998

## Notes for reviewers

https://github.com/autowarefoundation/autoware_launch/pull/1856 will be merged in the same time.

### ROS Parameter Changes

`<label>` is one of: `unknown`, `car`, `truck`, `bus`, `trailer`, `motorcycle`, `bicycle`, `pedestrian`.
`<shape>` is one of: `bounding_box`, `cylinder`, `polygon`.
`<tracker>` is a tracker type name string (e.g. `multi_vehicle_tracker`, `polygon_tracker`, `pedestrian_and_bicycle_tracker`).

#### Additions and removals

| Change type | Parameter name | Type | Default value | Description |
| Removed | `initial_tracker.<label>` | `string` | e.g. `"multi_vehicle_tracker"` | Tracker type to spawn per label. Replaced by `tracker_assignment.<shape>.<label>.create`. |
| Removed | `association.can_assign.<label>` | `string[]` | — | Eligible tracker types per label. Replaced by `tracker_assignment.<shape>.<label>.match`. |
| Removed | `association.max_dist.<label>.<tracker>` | `double` | — | Max distance threshold per (label, tracker). Replaced by `tracker_profiles.<tracker>.<shape>.<label>.max_dist`. |
| Removed | `association.max_area.<label>.<tracker>` | `double` | — | Max area threshold per (label, tracker). Replaced by `tracker_profiles.<tracker>.<shape>.<label>.max_area`. |
| Removed | `association.min_area.<label>.<tracker>` | `double` | — | Min area threshold per (label, tracker). Replaced by `tracker_profiles.<tracker>.<shape>.<label>.min_area`. |
| Removed | `association.min_iou.<label>.<tracker>` | `double` | — | Min IoU threshold per (label, tracker). Replaced by `tracker_profiles.<tracker>.<shape>.<label>.min_iou`. |
| Added | `tracker_assignment.<shape>.<label>.create` | `string` | (omitted = warn at runtime) | Tracker type to spawn for new detections of this (shape, label). `"null"` silently ignores; omitted entries emit a throttled runtime error. |
| Added | `tracker_assignment.<shape>.<label>.match` | `string[]` | `[]` (not associable) | Ordered list of tracker types eligible for association for this (shape, label). Omit for `"null"`-created entries. |
| Added | `tracker_profiles.<tracker>.<shape>.<label>.max_dist` | `double` | — | Max distance threshold [m]. Required for every (tracker, shape, label) triple referenced in any `match` list; missing entry throws at startup. |
| Added | `tracker_profiles.<tracker>.<shape>.<label>.max_area` | `double` | — | Max area threshold [m²] for this (tracker, shape, label) combination. |
| Added | `tracker_profiles.<tracker>.<shape>.<label>.min_area` | `double` | — | Min area threshold [m²] for this (tracker, shape, label) combination. |
| Added | `tracker_profiles.<tracker>.<shape>.<label>.min_iou` | `double` | — | Min IoU threshold for this (tracker, shape, label) combination. |

## Effects on system behavior

For bounding-box detections, runtime behavior is equivalent to before — the same tracker types are spawned and the same thresholds apply. Cylinder detections for vehicle labels (car, truck, bus, trailer, unknown) now have `create: "null"`, reflecting that cylinder shape is not expected for vehicles in practice. For polygon detections (e.g. from LiDAR cluster detectors) the polygon_tracker is now spawned directly instead of routing through bounding-box conversion, and the association profile uses relaxed area constraints appropriate for convex-hull geometry.


perception/autoware_multi_object_tracker/src/multi_object_tracker_node.cpp
```diff
@@ -132,105 +145,144 @@ MultiObjectTracker::MultiObjectTracker(const rclcpp::NodeOptions & node_options)
...
+    }
+
+    const bool has_non_passthrough = std::any_of(
+      params_.creation_config.shape_tracker_map.begin(),
+      params_.creation_config.shape_tracker_map.end(),
+      [](const auto & e) { return e.second != TrackerType::PASS_THROUGH; });
+    if (params_.association_config.association_params_map.empty() && has_non_passthrough) {
+      throw std::invalid_argument(
+        "No tracker_assignment.match entries found — check the parameter file.");
+    }
```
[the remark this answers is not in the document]
technolojin: pass_through_tracker is deprecated. it will be removed later. suggested fix is ignored.

perception/autoware_multi_object_tracker/config/data_association_matrix.param.yaml
YoshiRi, line 1: Concretely, for this PR, I think it would be better to make the `polygon` block in `data_association_matrix.param.yaml` explicit for all labels that we want to support, instead of relying on label-level defaults to fill them implicitly.

For example, the config could spell out all polygon cases like this:

```yaml
tracker_assignment:
  # label-level defaults, mainly representing the normal bounding-box behavior
  car:        {create: multi_vehicle_tracker,          match: [multi_vehicle_tracker]}
  truck:      {create: multi_vehicle_tracker,          match: [multi_vehicle_tracker]}
  bus:        {create: multi_vehicle_tracker,          match: [multi_vehicle_tracker]}
  trailer:    {create: multi_vehicle_tracker,          match: [multi_vehicle_tracker]}
  motorcycle: {create: pedestrian_and_bicycle_tracker, match: [pedestrian_and_bicycle_tracker]}
  bicycle:    {create: pedestrian_and_bicycle_tracker, match: [pedestrian_and_bicycle_tracker]}
  pedestrian: {create: pedestrian_and_bicycle_tracker, match: [pedestrian_and_bicycle_tracker]}

  # polygon-specific behavior, intentionally specified rather than inherited
  polygon:
    car:        {create: polygon_tracker,                match: [polygon_tracker, multi_vehicle_tracker]}
    truck:      {create: polygon_tracker,                match: [polygon_tracker, multi_vehicle_tracker]}
    bus:        {create: polygon_tracker,                match: [polygon_tracker, multi_vehicle_tracker]}
    trailer:    {create: polygon_tracker,                match: [polygon_tracker, multi_vehicle_tracker]}
    motorcycle: {create: pedestrian_and_bicycle_tracker, match: [pedestrian_and_bicycle_tracker]}
    bicycle:    {create: pedestrian_and_bicycle_tracker, match: [pedestrian_and_bicycle_tracker]}
    pedestrian: {create: pedestrian_and_bicycle_tracker, match: [pedestrian_and_bicycle_tracker]}
```

The exact tracker type for polygon pedestrian / bicycle / motorcycle can still be discussed, but I think the important point is that these rows should be intentionally specified as supported polygon inputs. This would make the intended spawn/update behavior much easier to review and test.

YoshiRi (commented):
Regarding `tracker_assignment`, I think it should act as an explicit whitelist of supported `shape × label` combinations, rather than just blindly inheriting class-level defaults. I mean pedestrian/bike should be also defined.

* **Polygon**: Should be explicitly supported for both vehicles and VRUs (pedestrian/bicycle/motorcycle).
* **Cylinder**: Mostly used for pedestrians, so inheriting it for vehicles/bikes feels unintended.

Since this PR is already large, how about we just make the **Polygon** spawn/update matrix explicit (and add tests) for now? We can handle the cleanup/policy for unsupported combinations (like cylinder + vehicles) in a follow-up PR to keep things moving.

perception/autoware_multi_object_tracker/config/data_association_matrix.param.yaml
technolojin: I removed label-level defaults and enforced to address all the shape-class combinations for tracker assignment. 
https://github.com/autowarefoundation/autoware_universe/pull/12672/changes/5a438f4b764bfba5237b54d053bcc620ef889aa8

 Omitted combination (ex. cylinder-car) shall be explicitly addressed.

technolojin:
@YoshiRi since the new format of association matrix, deprecated pass-through tracker is removed this time. https://github.com/autowarefoundation/autoware_universe/pull/12672/changes/84c5719b900260eee95cd6decc9bd862a5fcafc7

YoshiRi (approved):
LGTM
Now I understood these policy as following: 

| Input | Policy |
| --- | --- |
| `bounding_box` | Backward compatible. Basically use the existing tracker as-is. |
| `polygon + vehicle` | Spawn a `polygon_tracker` to preserve the polygon footprint. However, also allow association with the existing `multi_vehicle_tracker` to maintain track continuity. |
| `polygon + pedestrian/bicycle/motorcycle` | Assume that `pedestrian_and_bicycle_tracker` officially accepts polygon input. |
| `polygon + unknown` | Spawn a `polygon_tracker`. Allow broader association with existing trackers. |
| `cylinder + pedestrian` | Allowed. |
| `cylinder + non-pedestrian` | Explicitly reject with `create: "null"`. Do not spawn a tracker and do not associate. |
| omitted `(shape,label)` | Treat as a missing configuration. Raise a throttled runtime error on first encounter. |
| missing profile | If a `(tracker, shape, label)` profile referenced by `match` does not exist, raise a startup error. |

PR #12677, opened 2026-06-01
chore(multi object tracker): remove pass through tracker

technolojin:
## Description

Since the pass through tracker is developed https://github.com/autowarefoundation/autoware_universe/issues/655, it was not used.

Remove unused tracker to reduce maintenance cost.


technolojin:
merged to https://github.com/autowarefoundation/autoware_universe/pull/12672

PR #12682, opened 2026-06-01
feat(perception): add new cluster and merger

ktro2828:
## Description

This pull request introduces a new "label-based" euclidean clustering method to the `autoware_euclidean_cluster` package, along with related configuration, launch files, and documentation. It also implements the actual clustering logic for the voxel grid-based method, and adds a new object fusion merger node to the `autoware_object_merger` package. Below are the most important changes grouped by theme:

### Label-Based Euclidean Clustering Feature

* Added a new node `LabelBasedEuclideanClusterNode` with its implementation and interface (`label_based_euclidean_cluster_node.cpp`/`.hpp`), enabling euclidean clustering of semantic point clouds grouped by label. This node supports label mapping, probability filtering, and shape estimation options. [[1]](diffhunk://#diff-b163b420cfe9b5e075b3664638c9045543696af24f1d96fc6f040e562e0a1823R1-R76) [[2]](diffhunk://#diff-4aa701536d5d499cd04faf6f9d0bd5ed7019daa5b70bd19557a24e053f5e0649R66-R85)
* Introduced configuration (`label_based_euclidean_cluster.param.yaml`) for the new node, including clustering parameters, semantic label mapping, and shape estimation settings.
* Added launch files (`label_based_euclidean_cluster.launch.py` and `.launch.xml`) for easy deployment of the new node, supporting both standalone and containerized execution. [[1]](diffhunk://#diff-ae8f7eee40a93a71b0403ad7b955c144ce09202ffbcf34a57f4689a530e14a6fR1-R94) [[2]](diffhunk://#diff-360dce901ef4c1023cc4c73893cc567e44d5e242da0e65dfa91d6945fd160e2bR1-R16)
* Updated documentation in `README.md` to describe the new clustering method, its parameters, label mapping behavior, and edge cases. [[1]](diffhunk://#diff-72bbae48f354b3fcb200890a3c78c77d73719376e3d22b7a0ddb04eaa26d0b21R48-R61) [[2]](diffhunk://#diff-72bbae48f354b3fcb200890a3c78c77d73719376e3d22b7a0ddb04eaa26d0b21R105-R110)
* Declared the dependency on `autoware_shape_estimation` in `CMakeLists.txt` and `package.xml` to support shape estimation features. [[1]](diffhunk://#diff-4aa701536d5d499cd04faf6f9d0bd5ed7019daa5b70bd19557a24e053f5e0649R7) [[2]](diffhunk://#diff-e7e609c339549549c69d3e531108d2f17d7bde54c8adce319242f48f0de1cb82R18)

### Voxel Grid-Based Clustering Implementation

* Implemented the actual clustering logic in `VoxelGridBasedEuclideanCluster::cluster`, including voxel grid filtering, 2D centroid clustering, and output cluster filtering based on size and voxel constraints.

### Object Merger Enhancements

* Added a new node `ObjectFusionMergerNode` to the `autoware_object_merger` package, with build and registration updates in `CMakeLists.txt` and a corresponding parameter file (`object_fusion_merger.param.yaml`). [[1]](diffhunk://#diff-71a44a08434864435ea07c11c4781195e78d5eb2399fc4625e8bb0322345ad4aR17-R25) [[2]](diffhunk://#diff-71a44a08434864435ea07c11c4781195e78d5eb2399fc4625e8bb0322345ad4aL38-R59) [[3]](diffhunk://#diff-014cb3865f85249fedb2a151ba3fcadc0f5b3866645c38b23b2543ed6949cd4eR1-R5)

These changes collectively enhance the perception pipeline by enabling label-aware clustering, improving clustering implementation, and extending object merging capabilities.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


perception/autoware_euclidean_cluster/launch/label_based_euclidean_cluster.launch.py
```diff
@@ -0,0 +1,94 @@
...
+    def add_launch_arg(name: str, default_value=None):
+        return DeclareLaunchArgument(name, default_value=default_value)
+
+    return launch.LaunchDescription(
+        [
+            add_launch_arg("input_pointcloud", "/perception/ptv3/segmented/pointcloud"),
+            add_launch_arg("output_objects", "objects"),
+            add_launch_arg("use_pointcloud_container", "false"),
+            add_launch_arg("pointcloud_container_name", "pointcloud_container"),
+            add_launch_arg("shape_policy"),
```
YoshiRi: [minor comment]
shape_policy has a default in the XML launch file, but the Python launch argument appears to be declared without a default. If users launch label_based_euclidean_cluster.launch.py directly, this may fail unless they explicitly pass shape_policy. Could we set the same default "0" in the Python launch file as well?

perception/autoware_euclidean_cluster/config/label_based_euclidean_cluster.param.yaml
```diff
@@ -0,0 +1,52 @@
+/**:
+  ros__parameters:
+    # Euclidean clustering
+    use_height: false
```
YoshiRi, line 4: [Just a comment]
`VoxelGridBasedEuclideanCluster::cluster()` still appears to force 2D clustering by using a huge z leaf size and setting the clustering cloud z to 0.0. If `use_height` is intentionally unsupported, I think we can add some comment.

perception/autoware_euclidean_cluster/src/label_based_euclidean_cluster_node.hpp
```diff
@@ -0,0 +1,75 @@
...
+  /// @param options ROS 2 node options.
+  explicit LabelBasedEuclideanClusterNode(const rclcpp::NodeOptions & options);
+
+private:
+  /// @brief Process an input semantic point cloud and publish detected objects.
+  /// @param input_msg Input point cloud containing xyz and optionally class_id / probability.
+  void on_pointcloud(sensor_msgs::msg::PointCloud2::ConstSharedPtr input_msg);
+
+  /// @brief Build the mapping from semantic class IDs to Autoware object labels.
+  /// @param class_mappings Ordered class mappings from original class name to Autoware label.
```
YoshiRi, line 57: [Non-blocking / future improvement:]
The current label mapping relies on YAML declaration order to infer class_id. This is fragile because reordering the YAML, formatting tools, or parameter override behavior can silently change the semantic mapping. 

The semantic label order is fundamentally defined by the ML model artifact, not by this ROS node. Instead of duplicating the full label list in the ROS parameter file and relying on YAML declaration order, it may be more robust to load the model's label file, e.g. `label.txt`, from the ML package/artifact, and only configure which labels should be used and how they map to Autoware object labels.

For this PR, I think the current implementation is acceptable if the final resolved mapping is clearly logged at startup. However, I would prefer not to treat the ROS YAML file as the source of truth for the ML label order in the long term.

perception/autoware_euclidean_cluster/README.md
```diff
@@ -88,3 +102,9 @@ Example:
 ## (Optional) Future extensions / Unimplemented parts
 
 The `use_height` option of `voxel_grid_based_euclidean_cluster` isn't implemented yet.
+
+`label_based_euclidean_cluster` reads `class_names.<original_class_name>` in YAML declaration order and uses that order as `class_id`. Classes mapped to `car`, `bus`, `truck`, `motorcycle`, `bicycle`, or `pedestrian` are kept, and classes mapped to `ignore` are skipped.
+
+If the input pointcloud has no `probability` field, the node skips probability filtering and treats each point as probability `1.0`.
```
YoshiRi: [Suggestion]

```
`label_based_euclidean_cluster` currently computes the output object probability as the average point probability within each semantic label bucket before clustering. Therefore, all clusters generated from the same semantic label in one input pointcloud may share the same probability. This is a temporary implementation trade-off because the current clustering interface does not preserve source point indices. A future improvement is to compute probability per output cluster by preserving point indices through clustering.
```

perception/autoware_euclidean_cluster/src/label_based_euclidean_cluster_node.cpp
```diff
@@ -0,0 +1,472 @@
...
+    pcl::PointCloud<pcl::PointXYZ>::Ptr label_cloud(new pcl::PointCloud<pcl::PointXYZ>);
+    label_cloud->reserve(semantic_points.size());
+    for (const auto & semantic_point : semantic_points) {
+      label_cloud->push_back(semantic_point.point);
+    }
+
+    std::vector<pcl::PointCloud<pcl::PointXYZ>> clusters;
+    cluster_->cluster(label_cloud, clusters);
+
+    const float label_probability = average_probability(semantic_points);
```
YoshiRi, line 446: [Just a comment]
I think there are limitations for averaging whole pointcloud and not cluster-wise probability are calculated due to the design difficulty. So I prefer adding some documentation in both README and codebase.

```
// TODO: This probability is averaged over the whole semantic-label bucket,
// not over each output cluster. As a result, multiple clusters of the same
// label in one frame may share the same probability. Computing probability per
// output cluster requires preserving source point indices through the clustering
// step.
const float label_probability = average_probability(semantic_points);
```

perception/autoware_euclidean_cluster/lib/voxel_grid_based_euclidean_cluster.cpp
```diff
@@ -56,9 +56,99 @@ bool VoxelGridBasedEuclideanCluster::cluster(
...
+    is_extreme_large_cluster[cluster_idx] = cluster_size > max_voxel_cluster_for_output_;
+  }
+
+  // 5) Prepare output clusters
+  clusters.clear();
+  clusters.resize(cluster_indices.size());
+  // Track how many points each voxel has per cluster
+  std::unordered_map<int, std::unordered_map<int, int>> point_counts_per_voxel_per_cluster;
+  std::vector<size_t> random_indices(pointcloud->points.size());
+  static std::default_random_engine rng(42);
```
YoshiRi, line 114: [Future development/Non blocking]
For large clusters, I would ideally prefer voxel-wise deterministic or spatially uniform sampling rather than random sampling, because random sampling may drop edge/outline points and affect the estimated shape. If the more uniform approach is too expensive for this PR, random sampling is acceptable, but could we make the trade-off explicit?

Specifically:
- Add a README limitation that random downsampling may miss edge points and affect box/shape estimation.
- Add a TODO near the sampling implementation for voxel-wise uniform or edge-preserving sampling.
- Emit a throttled warning when random sampling is actually used.
- Seed the RNG deterministically, e.g. from the message header stamp / frame id / cluster id, so rosbag replay remains reproducible.

YoshiRi (commented):
I have a few non-blocking design/maintenance comments.

1. Label mapping
The semantic label order is fundamentally defined by the ML model artifact, not by this ROS node. Instead of duplicating the full label list in the ROS parameter file and relying on YAML declaration order, I think it would be more robust in the long term to load the model's label file, e.g. `label.txt`, from the ML package/artifact. Then this node only needs to configure which labels are used and how they map to Autoware object labels.

For this PR, I think the current approach is acceptable if the final resolved mapping is clearly logged at startup. I would treat the `label.txt` approach as a future improvement rather than a blocker.

2. Object probability
The current probability computation seems acceptable as an implementation trade-off, but the limitation should be explicit. Since the probability is averaged over the semantic-label bucket before clustering, multiple objects of the same label may share the same probability even if their local point confidences differ.

Could we document this in the README and add a TODO near the probability calculation? Ideally, a future implementation would preserve source point indices through clustering and compute probability per output cluster.

3. Random sampling for large clusters
For large clusters, I would ideally prefer voxel-wise deterministic or spatially uniform sampling rather than random sampling, because random sampling may drop edge/outline points and affect the estimated shape. If a more uniform approach is too expensive for this PR, random sampling is acceptable, but could we make the trade-off explicit?

Specifically:
- Add a README limitation that random downsampling may miss edge points and affect shape estimation.
- Add a TODO near the sampling implementation for voxel-wise uniform or edge-preserving sampling.
- Emit a throttled warning when random sampling is actually used.
- Seed the RNG deterministically, e.g. from the message header stamp / frame id / cluster id, so rosbag replay remains reproducible.

4. Size-related parameter names
Could we rename the size-related parameters to make their unit and stage explicit? In the voxel-grid path, some thresholds are applied to voxel clusters while others are applied to restored raw points. Names like `max_cluster_size` / `min_cluster_size` are easy to misinterpret during tuning.

For example:
- `min_cluster_size` -> `min_points_per_cluster`
- voxel-level max cluster limit -> `max_voxels_per_cluster`
- `min_voxel_cluster_size_for_filtering` -> `min_voxels_per_cluster`

If renaming is too disruptive for this PR, please at least document whether each threshold is counted in raw points or voxel centroids.

perception/autoware_euclidean_cluster/src/label_based_euclidean_cluster_node.hpp
ktro2828: That makes sense to me.
Ideally, I think PTv3-defined labels should be closed in PTv3 node itself.
PTv3 node should filter out segmented points to be ignored and publish points after converting ptv3-defined labels to Autoware labels.

Then, LabelBasedEuclideanCluster can perform clustering points into polygons or keeping points as points depending on Autoware labels.

technolojin (approved):
I tested with modified launcher and confirmed the label based euclidean cluster works.

the object fusion merger, however, could be replaced by the multi channel MOT.

Screenshot from 2026-06-04 17-24-39

perception/autoware_euclidean_cluster/src/label_based_euclidean_cluster_node.cpp
```diff
@@ -0,0 +1,472 @@
...
+  }
+  if (mapped_label == "bicycle") {
+    return ObjectClassification::BICYCLE;
+  }
+  if (mapped_label == "pedestrian") {
+    return ObjectClassification::PEDESTRIAN;
+  }
+  if (mapped_label == "animal") {
+    return ObjectClassification::ANIMAL;
+  }
```
technolojin, line 147:
```suggestion
  }
  if (mapped_label == "hazard") {
    return ObjectClassification::HAZARD;
  }
```

perception/autoware_euclidean_cluster/config/label_based_euclidean_cluster.param.yaml
```diff
@@ -0,0 +1,52 @@
...
+      forklift: car
+      kart: car
+      motorcycle: motorcycle
+      bicycle: bicycle
+      pedestrian: pedestrian
+      personal_mobility: pedestrian
+      animal: animal
+      pushable_pullable: unknown
+      traffic_cone: unknown
+      debris: unknown
```
technolojin:
```suggestion
      traffic_cone: hazard
      debris: hazard
```

technolojin (changes requested):
after add hazard, the hazard object could be tracked. https://github.com/autowarefoundation/autoware_universe/pull/12698

Image

perception/autoware_euclidean_cluster/src/label_based_euclidean_cluster_node.cpp
ktro2828: Fixed in https://github.com/autowarefoundation/autoware_universe/pull/12682/commits/90cc1fa7f6cf09970c558f62d755e8cd32f85f75.

perception/autoware_euclidean_cluster/config/label_based_euclidean_cluster.param.yaml
ktro2828: Fixed in https://github.com/autowarefoundation/autoware_universe/pull/12682/commits/90cc1fa7f6cf09970c558f62d755e8cd32f85f75 including `pushable_pullable`.

perception/autoware_euclidean_cluster/launch/label_based_euclidean_cluster.launch.py
ktro2828: Fixed in https://github.com/autowarefoundation/autoware_universe/pull/12682/commits/77b58e504c630e489fd0731184e91057bd93345e.

perception/autoware_euclidean_cluster/src/label_based_euclidean_cluster_node.cpp
ktro2828: Add TODO comment in https://github.com/autowarefoundation/autoware_universe/pull/12682/commits/f27df285bee85f74ef12b5c1e36bc959736d9baf.

perception/autoware_euclidean_cluster/README.md
ktro2828: Add a document for `LabelBasedEuclideanCluster` in https://github.com/autowarefoundation/autoware_universe/pull/12682/commits/f27df285bee85f74ef12b5c1e36bc959736d9baf.

technolojin (approved):

technolojin:
class splits clusters. In case the classes are similar labels (trailer, truck) it causes over-segmentation 

Screenshot from 2026-06-05 13-17-44

remap similar labels to one output label solves the problem
```yaml
      truck: truck
      tractor_unit: truck
      semi_trailer: truck
      construction_vehicle: truck
```

perception/autoware_euclidean_cluster/lib/voxel_grid_based_euclidean_cluster.cpp
ktro2828: Added in https://github.com/autowarefoundation/autoware_universe/pull/12682/commits/7b9bcc56751eb302580b7efd597a0ebe55ec5d8e.

technolojin:
schema.json files are missing.

ktro2828:
NOTE: From the discussion, `pushable_pullable` should be treated as `pedestrian` like `stroller`.

perception/autoware_euclidean_cluster/launch/label_based_euclidean_cluster.launch.py
```diff
@@ -0,0 +1,93 @@
...
+    def add_launch_arg(name: str, default_value=None):
+        return DeclareLaunchArgument(name, default_value=default_value)
+
+    return launch.LaunchDescription(
+        [
+            add_launch_arg("input_pointcloud", "/perception/ptv3/segmented/pointcloud"),
+            add_launch_arg("output_objects", "objects"),
+            add_launch_arg("use_pointcloud_container", "false"),
+            add_launch_arg("pointcloud_container_name", "pointcloud_container"),
+            add_launch_arg("shape_policy", 0),
```
technolojin, line 83:
```suggestion
            add_launch_arg("shape_policy", "0"),
```

technolojin:
```
1780899091.3378823 [ERROR] [launch]: Caught exception in launch (see debug for traceback): Caught multiple exceptions when trying to load file of format [py]:
 - TypeError: 'int' object is not iterable
 - InvalidFrontendLaunchFileError: The launch file may have a syntax error, or its format is unknown
```

PR #12688, opened 2026-06-02
feat(image_projection_based_fusion): apply `agnocast_wrapper::Node` to RoiFusion base node

Koichi98:
## Description
Apply `agnocast_wrapper::Node` to the `FusionNode` base class of `autoware_image_projection_based_fusion`. This covers all of its derived nodes at once: `RoiDetectedObjectFusionNode`, `RoiClusterFusionNode`, `RoiPointCloudFusionNode`, `SegmentPointCloudFusionNode` and `PointPaintingFusionNode`.

Because the base class is now Agnocast-aware, the partial-Agnocast workarounds that were previously needed in the derived nodes are removed: the dedicated `agnocast_pub_ptr_` / `agnocast_msg3d_sub_` members, the `publish()` overrides, the extra ROS 2 → Agnocast message copy in `roi_cluster_fusion`, and the separate `callback_group`. This resolves three `TODO(Koichi98)` comments.

## Dependencies

Depends on https://github.com/autowarefoundation/autoware_core/pull/1294 (`autoware_agnocast_wrapper`).

`FusionNode` hands itself to the matching strategy and the collectors as a shared pointer, so it needs `shared_from_this()`. `agnocast_wrapper::Node` does not provide it yet, so without autoware_core#1294 this PR fails to compile in both modes with `'class FusionNode<...>' has no member named 'shared_from_this'`. Note that the existing `feat/shared_from_this` branch is not sufficient: it patches only the `ENABLE_AGNOCAST=1` class, while the `=0` class needs it too.

## Related Links

  - https://github.com/autowarefoundation/autoware_universe/pull/12439 (planned to be reverted, see *Notes for reviewers*)
  - https://github.com/autowarefoundation/autoware_universe/pull/13011 (planned to be reverted, see *Notes for reviewers*)

## Related links

**Parent Issue:**

- Link

## How was this PR tested?
  - Ran Lsim with sample rosbag.

### 1. `#12439` / `#13011` will be reverted in a follow-up PR

Both were workarounds for the earlier partial (Method 1) integration: a dedicated Agnocast `callback_group` (`#12439`), plus `diagnostic_mutex_` for the contention it caused (`#13011`). Both are unnecessary now that `FusionNode` inherits `agnocast_wrapper::Node`, but are kept as-is here to keep the diff focused; **the revert comes in a separate follow-up PR.**

### 2. `FusionCollector::set_period()` no longer touches the `rcl` timer API

`rcl_timer_get_period()` / `rcl_timer_exchange_period()` are not available for an Agnocast timer, so they are replaced by `autoware::agnocast_wrapper::set_period(timer_, period)`. The `catch` is widened to `const std::exception &` because the wrapper does not necessarily surface `rcl` failures as `RCLError`; behavior on failure is unchanged.

### 3. Debug image topics use plain publishers instead of `image_transport`

`image_transport` requires a real `rclcpp::Node`, so `debug_mode` (`Debugger`) and `is_publish_debug_mask` now publish plain `sensor_msgs/msg/Image`. Both keep working; only the `image_transport` extras are lost, i.e. the `/compressed` (and other transport) side topics and the `format` / `jpeg_quality` / `png_level` parameters they declared.

## Interface changes

The `image_transport` side topics of the debug image outputs (e.g. `~/debug/image_rawN/compressed`) and the `format` / `jpeg_quality` / `png_level` parameters are gone. The raw `Image` topics are unchanged.

## Effects on system behavior

One exception, in both modes: the debug image outputs (`~/debug/image_rawN`, `~/debug/mask`) no longer offer compressed transports, so tools subscribing to the `/compressed` topics must switch to the raw ones. Debug visualization itself and all fusion output topics are unchanged.


perception/autoware_image_projection_based_fusion/src/fusion_collector.cpp
```diff
@@ -55,20 +55,18 @@ template <class Msg3D, class Msg2D, class ExportObj>
 void FusionCollector<Msg3D, Msg2D, ExportObj>::set_info(
   std::shared_ptr<FusionCollectorInfoBase> fusion_collector_info)
 {
-  std::lock_guard<std::mutex> lock(fusion_mutex_);
```
YoshiRi, line 58: [Question]
Is it okay to remove this lock_guard?
We can not prove set/get will called simultaneously, so does it sound safer to keep this guard?

Koichi98: @YoshiRi 
The intent of this PR is to bring the callback group back to a single one, i.e. to restore the application logic to what it was before Agnocast was applied, so I removed the lock as part of that.

But if we want to be robust against the callback groups being split again in the future, the lock should stay as you mentioned. It seems better to keep it, so I've restored it in https://github.com/autowarefoundation/autoware_universe/pull/12688/commits/67a839f37d142d9e7d4ec695a9adb909e309cb5f.

YoshiRi (approved):
Thank you for the fix. LGTM

Koichi98:
@YoshiRi

In 9b76b42832 I added only `autoware_utils_diagnostics` and `autoware_utils_debug` to `package.xml`, since this PR started including those headers directly. The `autoware_utils` umbrella already re-exports them, so this is just an explicit
declaration, not a build fix.

I think it's fine and plan to merge as is — let me know if you see any problem.

PR #12697, opened 2026-06-03
fix(autoware_topic_relay_controller): fix keep publish feature

TetsuKawa:
## Description
Previously, the system provided the following two features:
- **`enable_relay_control`**: Turns topic relaying ON/OFF via external service requests.
- **`enable_keep_publishing`**: Continues publishing the last received value even if the input stops.

### Problem / Previous Behavior
When both features were set to `true`, the previous specification caused the topic forwarding to stop due to the `enable_relay_control` logic.

### Description / Updated Behavior
This change modifies the behavior so that when both features are `true`, the system will continue to publish the topic message received just before the service request

## Related links

**Parent Issue:**
https://github.com/autowarefoundation/autoware_universe/issues/12691

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


mkuri:
@TetsuKawa 
Minor: there is a typo in the PR title. `autoware_topci_relay_controller` should be `autoware_topic_relay_controller`. Could you fix it? The title is used for the squash-merge commit message, so it would be good to have it correct.

mkuri (approved):

PR #12698, opened 2026-06-03
feat(multi object tracker): add animal/hazard class and static tracker

technolojin:
## Description

feature
* add animal/hazard classes to tracker pipelie
* add association parameters
* add static tracker, which tracks hazard class
* convert polygon to bounding box in case of static tracker

bug fix
* polar association: use association_params min_area and max_area

refactor
* vehicle tracker: alignClusterToTrackerOrientation generalized to shape.cpp alignClusterToOrientation function

## Related links

**Parent Issue:**

- https://github.com/autowarefoundation/autoware/issues/7137

## How was this PR tested?

for new classes, ptv3 and label based cluster is used. https://github.com/autowarefoundation/autoware_universe/pull/12682

visualized by modified rviz plugin https://github.com/autowarefoundation/autoware_rviz_plugins/pull/42

https://github.com/user-attachments/assets/355b471c-ba46-4292-9271-463efd5aa28a

clusters are classified by ptv3, and clustered by label based cluster https://github.com/autowarefoundation/autoware_universe/pull/12682 WIP

Poles to hazard class, tracked.

Screenshot from 2026-06-04 18-56-14

## Notes for reviewers

None.

### ROS Parameter Changes

parameter update https://github.com/autowarefoundation/autoware_launch/pull/1864 will be done in the same time

## Effects on system behavior

- `animal` and `hazard` are newly tracked classes.
- `hazard` is currently treated as a static object and tracked by `static_tracker`.
- Polygon-shaped hazard objects may be converted to bounding boxes for static tracking.
- Static tracker outputs set `kinematics.is_stationary = true`.

### DIFF
--- a/perception/autoware_multi_object_tracker/CMakeLists.txt
+++ b/perception/autoware_multi_object_tracker/CMakeLists.txt
@@ -62,6 +62,7 @@ set(${PROJECT_NAME}_lib
   lib/tracker/model/pedestrian_tracker.cpp
   lib/tracker/model/pedestrian_and_bicycle_tracker.cpp
   lib/tracker/model/polygon_tracker.cpp
+  lib/tracker/model/static_tracker.cpp
   lib/uncertainty/uncertainty_processor.cpp
   lib/association/adaptive_threshold_cache.cpp
   lib/tracker/shape_model/unstable_shape_filter.cpp
--- a/perception/autoware_multi_object_tracker/config/data_association_matrix.param.yaml
+++ b/perception/autoware_multi_object_tracker/config/data_association_matrix.param.yaml
@@ -52,11 +52,11 @@
           bus:      {max_dist: 5.0, max_area: 60.00, min_area: 10.0,  min_iou: 0.10}
           trailer:  {max_dist: 5.0, max_area: 60.00, min_area: 10.0,  min_iou: 0.10}
         polygon:
-          unknown:  {max_dist: 4.0, max_area: 60.0,  min_area: 1.6,   min_iou: 0.0001}
-          car:      {max_dist: 4.0, max_area: 60.0,  min_area: 1.6,   min_iou: 0.0001}
-          truck:    {max_dist: 4.0, max_area: 60.0,  min_area: 1.6,   min_iou: 0.0001}
-          bus:      {max_dist: 4.0, max_area: 60.0,  min_area: 1.6,   min_iou: 0.0001}
-          trailer:  {max_dist: 4.0, max_area: 60.0,  min_area: 1.6,   min_iou: 0.0001}
+          unknown:  {max_dist: 4.0, max_area: 75.0,  min_area: 0.01,   min_iou: 0.0001}
+          car:      {max_dist: 4.0, max_area: 75.0,  min_area: 0.01,   min_iou: 0.0001}
+          truck:    {max_dist: 4.0, max_area: 75.0,  min_area: 0.01,   min_iou: 0.0001}
+          bus:      {max_dist: 4.0, max_area: 75.0,  min_area: 0.01,   min_iou: 0.0001}
+          trailer:  {max_dist: 4.0, max_area: 75.0,  min_area: 0.01,   min_iou: 0.0001}
       pedestrian_and_bicycle_tracker:
         bounding_box:
           unknown:    {max_dist: 3.0, max_area: 2.5,  min_area: 0.01, min_iou: 0.0001}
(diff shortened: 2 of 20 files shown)


technolojin:
@YoshiRi The static tracker will export bounding box now. later, I will enable tracker to have options to convert polygon to bounding box or not.

YoshiRi (approved):
No blocking concerns from my side, but I think it would be good to document two design points so that the current behavior is easier to understand and maintain.

## 1. Footprint passthrough and unit tests

Since the current design keeps the footprint available even after polygon-to-bounding-box conversion, I think we should make this contract explicit in the conversion function comment and cover it with unit tests.

My understanding is:

- Polygon hazard objects may be converted to `BOUNDING_BOX` for static tracking.
- This conversion does not discard the original footprint.
- Downstream modules are expected to be able to look at both:
  - `shape.type` / `shape.dimensions` as the bounding box representation
  - `shape.footprint` as the detailed footprint information, shifted relative to the updated object center

If this is the intended contract, I think the following unit tests would be useful:

- Footprint passthrough / preservation
  - Input: polygon footprint.
  - Expected:
    - output `shape.type == BOUNDING_BOX`
    - output footprint has the same number of points as the input
    - footprint geometry is preserved
    - footprint is shifted relative to the updated object center

- Axis-aligned rectangle conversion
  - Input: simple rectangular polygon in the object local frame.
  - Expected:
    - bbox center is correct
    - bbox dimensions are correct
    - bbox orientation is correct or geometrically equivalent
    - footprint remains available

- Rotated rectangle conversion
  - Input: rectangular footprint rotated relative to the object pose.
  - Expected:
    - bbox orientation aligns with the footprint orientation
    - bbox dimensions are correct
    - footprint geometry is preserved

- Ego-pose behavior
  - Input: the same polygon with and without ego pose.
  - Expected:
    - with ego pose, the ego-facing edge preference behaves as intended
    - without ego pose, the function falls back to the full edge search
    - output remains stable for representative hazard shapes such as poles or small static objects

- CW / CCW polygon order
  - Input: the same polygon points in clockwise and counter-clockwise order.
  - Expected:
    - conversion succeeds in both cases
    - bbox center and dimensions are equivalent
    - footprint geometry is preserved
  - Note: bbox yaw may only need to be geometrically equivalent, not exactly identical.

- Degenerate footprint handling
  - Input:
    - fewer than 3 points
    - duplicated points
    - near-zero-length edges
    - nearly collinear points
  - Expected:
    - invalid polygons fail safely
    - no crash or undefined behavior

- StaticTracker integration
  - Input: polygon hazard object.
  - Expected:
    - `StaticTracker::measure()` applies polygon-to-bbox conversion when applicable
    - footprint remains available in the tracked object
    - `getTrackedObject(..., to_publish=true)` sets `kinematics.is_stationary = true`
    - pose, shape, and area are internally consistent after conversion

## 2. Association / transition policy

I also think the current `unknown` / `hazard` / `static_tracker` transition policy can become confusing, so it would be helpful to summarize it somewhere, for example in the config comments, PR description, or a short design note.

Current behavior appears to be:

- `bounding_box + hazard`
  - create: `static_tracker`
  - match: `static_tracker`

- `polygon + hazard`
  - create: `static_tracker`
  - match: `polygon_tracker`, `static_tracker`

- `cylinder + hazard`
  - create: `static_tracker`
  - match: `static_tracker`

- `polygon + unknown`
  - create: `polygon_tracker`
  - match: `polygon_tracker`, `static_tracker`, `multi_vehicle_tracker`, `pedestrian_and_bicycle_tracker`

The following points are still unclear to me, so I think they should be documented as TBD or clarified before this behavior becomes a hidden assumption:

- TBD: If an `unknown` polygon tracker is associated with a `hazard` polygon measurement, should the tracker remain `polygon_tracker`?

- TBD: If a `static_tracker` created from a hazard measurement is later associated with an `unknown` polygon measurement, should it remain `static_tracker`?

- TBD: If a static tracker is associated with an unknown measurement, should it still publish `kinematics.is_stationary = true`?

- TBD: Is the cross-association between `unknown`, `hazard`, `polygon_tracker`, and `static_tracker` intended to tolerate label flicker?

- TBD: What prevents moving objects from being absorbed into `static_tracker` when they are temporarily classified as `hazard` or `unknown`?

- TBD: Should tracker type ever transition after association, or is tracker type fixed at creation time?

- TBD: Should label/classification update policy be documented separately from tracker-type transition policy?

technolojin:
I would like to consider your suggestions and questions to future development. Especially, the tracker output policy shall be clarified.

PR #12710, opened 2026-06-05
feat(traffic_light_multi_camera_fusion): apply agnocast to publisher in `traffic_light_multi_camera_fusion`

Koichi98:
## Description
Apply `agnocast_wrapper::Node` to `autoware_traffic_light_multi_camera_fusion`.
Also, please refer to https://github.com/autowarefoundation/autoware_core/blob/main/common/autoware_agnocast_wrapper/docs/review_guide.md for the review guide of `autoware_agnocast_wrapper` application PR. FYI: This node take [Method1](https://github.com/autowarefoundation/autoware_core/blob/main/common/autoware_agnocast_wrapper/docs/review_guide.md#method-1-macro--free-function-api) of two integrations methods.

### Node specific changes
  - To build the output directly in the agnocast-allocated shared memory, `MultiCameraFusion::fuse()` now takes the output `TrafficLightGroupArray` by reference and writes the result in place. Accordingly, `traffic_light_groups` was removed from `MultiCameraFusionResult`. This also eliminates a pre-existing intermediate copy inside `fuse()`.
  - `test_multi_camera_fusion.cpp` was updated only to match the new `fuse()` signature (pass an output buffer and read the fused groups from it). These are mechanical changes with no test logic modified.

## Related Links

## Related links

**Parent Issue:**

- Link

## How was this PR tested?
  - Ran Autoware based product with this change applied and confirmed it operates without any regression.

## Interface changes

None.

## Effects on system behavior


a-maumau (approved):
LGTM.

PR #12722, opened 2026-06-08
feat(trajectory_selector): anchor the concatenation on the main input

maxime-clem:
## Description
This PR introduces an "anchor" mechanism to the trajectory_selector node to reduce processing latency.

Previously, the node processed buffered trajectories strictly on a fixed 100ms timer. This could introduce up to 100ms of latency if a generative planner finished its computation just after a timer cycle. 

With this change, the `~/input/trajectories_generative` topic acts as an anchor: receiving a message on this topic now triggers an immediate concatenation and validation cycle and resets the internal timer.
Backup trajectories (`~/input/trajectories_backup`) continue to be buffered and processed in the next available cycle (either triggered by a new anchor message or the periodic timer).

## Related links

https://tier4.atlassian.net/browse/PDDP-220

## How was this PR tested?

Unit tests
Psim

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/autoware_trajectory_concatenator/src/detail/trajectory_concatenator.cpp
+++ b/planning/autoware_trajectory_concatenator/src/detail/trajectory_concatenator.cpp
@@ -61,9 +61,6 @@ CandidateTrajectories TrajectoryConcatenator::get_concatenated(
   std::vector<autoware_internal_planning_msgs::msg::GeneratorInfo> generator_info;
 
   const double current_time_sec = to_seconds(current_time);
-  std::cout << "[DEBUG] get_concatenated current_time_sec: " << current_time_sec
-            << ", params_.duration_time: " << params_.duration_time << std::endl;
-
   // Prune expired entries individually
   for (auto it = buffer_.begin(); it != buffer_.end();) {
     auto & pre_combine = it->second;
--- a/planning/autoware_trajectory_concatenator/src/detail/trajectory_concatenator.cpp
+++ b/planning/autoware_trajectory_concatenator/src/detail/trajectory_concatenator.cpp
@@ -52,6 +52,8 @@ void TrajectoryConcatenator::add_candidate(const CandidateTrajectories & msg)
   }
 }
 
+#include <iostream>
+
 CandidateTrajectories TrajectoryConcatenator::get_concatenated(
   const builtin_interfaces::msg::Time & current_time)
 {
@@ -59,6 +61,8 @@ CandidateTrajectories TrajectoryConcatenator::get_concatenated(
   std::vector<autoware_internal_planning_msgs::msg::GeneratorInfo> generator_info;
 
   const double current_time_sec = to_seconds(current_time);
+  std::cout << "[DEBUG] get_concatenated current_time_sec: " << current_time_sec
+            << ", params_.duration_time: " << params_.duration_time << std::endl;
 
   // Prune expired entries individually
   for (auto it = buffer_.begin(); it != buffer_.end();) {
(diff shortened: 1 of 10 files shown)


planning/autoware_trajectory_concatenator/src/detail/trajectory_concatenator.cpp
```diff
@@ -52,13 +52,17 @@ void TrajectoryConcatenator::add_candidate(const CandidateTrajectories & msg)
...
+#include <iostream>
+
 CandidateTrajectories TrajectoryConcatenator::get_concatenated(
   const builtin_interfaces::msg::Time & current_time)
 {
   std::vector<autoware_internal_planning_msgs::msg::CandidateTrajectory> trajectories;
   std::vector<autoware_internal_planning_msgs::msg::GeneratorInfo> generator_info;
 
   const double current_time_sec = to_seconds(current_time);
+  std::cout << "[DEBUG] get_concatenated current_time_sec: " << current_time_sec
```
zulfaqar-azmi-t4: this is accessible from `concatenator_wrapper`, and we can use `rclcpp`'s logging instead.

planning/autoware_trajectory_selector/param/parameter_struct.yaml
```diff
@@ -0,0 +1,6 @@
+selector:
+  fallback_period_ms:
+    type: int
+    default_value: 150
+    description: fallback timer period in milliseconds.
+    read_only: false
```
zulfaqar-azmi-t4, line 6: to keep consistency with json schema minimum value, we should add [validation](https://github.com/pickNikRobotics/generate_parameter_library#built-in-validators). Personally i feel the minimum should be equal to or greater than 100ms.

planning/autoware_trajectory_selector/src/trajectory_selector_node.cpp
```diff
@@ -133,6 +140,27 @@ void TrajectorySelectorNode::on_timer()
   pub_trajectories_->publish(validated_trajectories);
 }
 
+void TrajectorySelectorNode::update_parameters()
+{
+  if (selector_params_listener_.is_old(selector_params_)) {
+    selector_params_ = selector_params_listener_.get_params();
+    update_fallback_timer();
```
zulfaqar-azmi-t4: in case there is any new parameters added and update that parameter, we might ended up recreating a new timer as well. so maybe we can add some guard?

```
  const auto new_params = selector_params_listener_.get_params();
  if (new_params.fallback_period_ms != selector_params_.fallback_period_ms) {
    update_fallback_timer();
  }
  selector_params_ = new_params;

```

zulfaqar-azmi-t4 (approved):
LGTM, just some minor comments on my side.

planning/autoware_trajectory_selector/src/trajectory_selector_node.cpp
maxime-clem: Done: https://github.com/autowarefoundation/autoware_universe/commit/b4c67b0c428e743394a5409ea5782fca87fb73b8

planning/autoware_trajectory_selector/param/parameter_struct.yaml
maxime-clem: Done: https://github.com/autowarefoundation/autoware_universe/commit/b4c67b0c428e743394a5409ea5782fca87fb73b8

planning/autoware_trajectory_concatenator/src/detail/trajectory_concatenator.cpp
maxime-clem: Done: b4c67b0c428

yukkysaito (approved):

PR #9747, opened 2024-12-23
feat(autoware_traffic_light_arbiter): add current time validation

MasatoSaeki:
## Description
This PR enables to validate external messages compared by current time.

## Related links

**Parent Issue:**
- [TIER IV INTERNAL LINK](https://tier4.atlassian.net/jira/software/c/projects/ISSJP/boards/671?assignee=6061477cafe465006afac8aa&selectedIssue=ISSJP-1798)
- https://github.com/autowarefoundation/autoware_launch/pull/1289

## How was this PR tested?
I have checked it on [Evaluator](https://evaluation.ci.tier4.jp/evaluation/reports/8ffde025-0720-59e9-9300-18d76379227b?project_id=prd_jt). 

## Notes for reviewers

None.

## Interface changes

None.

#### Additions and removals

| Change type   | Parameter Name | Type     | Default Value | Description       |
|:--------------|:---------------|:---------|:--------------|:------------------|
| Added | `external_current_time_tolerance`   | double | 5.0           | The duration in seconds an external message is considered valid for merging in comparison with current time                                                                 |

## Effects on system behavior

None.

PR #12723, opened 2026-06-08
test(autoware_traffic_light_arbiter): add unit tests for logic

takam5f2:
## Description

Adds 40 unit tests for `TrafficLightArbiterCore`, the ROS-free core extracted in #12660. The new test file `test_traffic_light_arbiter_core.cpp` instantiates `TrafficLightArbiterCore` directly (no Node, no executor, no ROS pub/sub), pins one observable behaviour per test at the public-contract level, and is organised by `(mode × concern)` fixtures. Production code is untouched; the characterization suite (`test_traffic_arbiter_characteristic.cpp`) and the existing node test (`test_node.cpp`) are unchanged.

### Features

- **New file `test_traffic_light_arbiter_core.cpp`** — 40 tests across 9 fixtures, exercising `TrafficLightArbiterCore` directly without the Node layer.
- **Fixtures organised by `(mode × concern)`** — mode/priority is the primary axis; mode-agnostic concerns (boundary / staleness / latest_input_time / timing) get their own fixtures.
- **Assertions touch only the public API** — `result.output`, `result.latest_input_time`, `result.off_map_signal_ids`, the `accepted` and `expired` fields returned by `ingest_external`, and the `expired` returned by `ingest_perception`.
- **Comments at specification level** — internal identifiers (`validator`, `branch`, `path`, private member names, etc.) do not appear in any comment, so the tests survive future Core-side refactoring.
- **Helpers built on `std::optional<T>`** — `observed_color`, `observed_color_of_shape`, `observed_prediction_source`, `observed_group_count`, and `find_group` absorb the nullopt case internally, so call sites need no `ASSERT_TRUE` guard before observing the result.

### Test structure

| Fixture | Mode | Tests | Pinned behaviour |
| --- | --- | --- | --- |
| `SignalMatchingTest` | matching=true, CONFIDENCE | 7 | matched / color mismatch / shape-set mismatch / single-source (both directions) / off-map / equivalence ignores confidence |
| `PedestrianTest` | matching=true, parametrized | 4 | pedestrian path × 3 priorities + single-source pass-through |
| `ConfidencePriorityTest` | priority, CONFIDENCE | 8 | higher confidence wins / per-shape union / perception-only / off-map / multiple external accumulation / predictions (both / perception-only / external-only) |
| `ExternalPriorityTest` | priority, EXTERNAL | 3 | EXTERNAL setting overrides confidence + single-source fallback (both sides) |
| `PerceptionPriorityTest` | priority, PERCEPTION | 3 | symmetric counterpart of ExternalPriority |
| `BoundaryTest` | (free `TEST()`) | 2 | arbitrate without map (no output) / empty map (engaged empty output) |
| `PerceptionStalenessTest` | priority, CONFIDENCE | 5 | staleness three branches (external presence × tolerance) + message-level scope + prediction merge effect |
| `LatestInputTimeTest` | priority, CONFIDENCE | 3 | most-recent stamp selection in three branches (external wins / perception wins / external-absent fallback) |
| `TimingTest` | priority, CONFIDENCE | 5 | admission three boundaries (within / too old / too far future) + sweep report + sweep's effect on the next arbitrate |

## Related links

**Parent Issue:**

- None

**Built on:**

- #12660 (`refactor(autoware_traffic_light_arbiter): extract TrafficLightArbiterCore from Node`) — extracted the Core class this PR targets directly; this PR is the unit-test sibling.

## How was this PR tested?

`colcon build` + `colcon test` against the package. The new `_core_test` gtest binary runs **40 cases, all pass**. Existing `_test` (7/7) and `_characteristic_test` (24/24) plus the four lint targets remain green. No production code is changed.

```bash
colcon build --packages-select autoware_traffic_light_arbiter
colcon test --packages-select autoware_traffic_light_arbiter
colcon test-result --test-result-base build/autoware_traffic_light_arbiter --verbose
```

#### (1) Perception side has no absolute staleness guard (Node + Core both)

- `arbitrate()`'s staleness check is `has_externals && (max_external_stamp - perception_stamp).seconds() > perception_time_tolerance_`, which **short-circuits when no external stamp is stored**. The Node side likewise has no `now() - msg.stamp` guard on `on_perception_msg`; the message is forwarded straight to `core_->ingest_perception(*msg)`. Only `on_external_msg` goes through wall-clock admission control (via `ingest_external(*msg, this->now())`).
- **This asymmetry predates the Core extraction.** Commit `3dc9605d9 (feat(autoware_traffic_light_arbiter): add current time validation, #9747)` introduced the wall-clock admission for external only; the perception side was never given the same treatment. The Core extraction preserved this exactly.
- Two readings are possible:
  - **Intentional**: the Core is kept as a pure function, and absolute staleness is treated as a Node-layer concern.
  - **Latent production-safety gap**: if the perception node stalls or its publisher disconnects while no external source is publishing, an arbitrarily old perception signal can keep reaching the published output.
- The corresponding test (`perceptionAloneIsNeverStale`) pins the current behaviour, and its in-file comment explicitly says "this test pins it; it does not endorse it." If the reviewer reads this as a gap, a follow-up PR would address it either by:
  - Adding admission to the Node layer (cheapest — Core API unchanged), or
  - Reshaping `arbitrate()` to take `current_time` (loses Core's pure-function property).

#### (2) `ingest_perception`'s `vector<ExpiredExternalSignal>` return value is logging-only

- The only consumer of the return value is `TrafficLightArbiter::log_expired_external_signals`, which `RCLCPP_DEBUG`s each entry.
- If DEBUG logging of expired entries is not required in production, the API can simplify to `void ingest_perception(msg)`. Three of the four assertions in `ingestPerceptionReportsExpiredExternalEntry` (size / id / age) would be dropped along with the API change. The behavioural test (`evictedExternalEntryAbsentFromOutput`) does not depend on the return value and would remain.
- The reviewer can either:
  - Keep the current API (DEBUG logging is considered useful) — no change, both tests stay.
  - Simplify (`void` return) in a follow-up PR — drop `ingestPerceptionReportsExpiredExternalEntry` together with the API change.

## Interface changes

None. No production code is touched. `CMakeLists.txt` only adds a new gtest target (`autoware_traffic_light_arbiter_core_test`); the library, the executable, the component registration, and the parameters are all unchanged.

## Effects on system behavior

**Downstream-observable**: None. Pure test addition.

**Internal-observable (not affecting downstream)**: None.

### DIFF
--- a/perception/autoware_traffic_light_arbiter/test/test_traffic_light_arbiter_core.cpp
+++ b/perception/autoware_traffic_light_arbiter/test/test_traffic_light_arbiter_core.cpp
@@ -1,0 +1,1305 @@
+// Copyright 2026 TIER IV, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "autoware/traffic_light_arbiter/traffic_light_arbiter_core.hpp"
+
+#include <autoware_lanelet2_extension/utility/query.hpp>
+#include <rclcpp/duration.hpp>
+#include <rclcpp/time.hpp>
+
+#include <autoware_perception_msgs/msg/traffic_light_group_array.hpp>
+
+#include <gtest/gtest.h>
+#include <lanelet2_core/Attribute.h>
+#include <lanelet2_core/LaneletMap.h>
+#include <lanelet2_core/primitives/BasicRegulatoryElements.h>
+#include <lanelet2_core/primitives/Lanelet.h>
+#include <lanelet2_core/primitives/LineString.h>
+#include <lanelet2_core/primitives/LineStringOrPolygon.h>
+#include <lanelet2_core/primitives/Point.h>
+
+#include <cstddef>
+#include <cstdint>
+#include <optional>
+#include <string>
+#include <utility>
+#include <vector>
+
+namespace
+{
+
+using autoware::traffic_light::SourcePriority;
+using autoware::traffic_light::TrafficLightArbiterCore;
+using PredictedTrafficLightState = autoware_perception_msgs::msg::PredictedTrafficLightState;
+using TrafficLightElement = autoware_perception_msgs::msg::TrafficLightElement;
+using TrafficLightGroup = autoware_perception_msgs::msg::TrafficLightGroup;
+using TrafficLightGroupArray = autoware_perception_msgs::msg::TrafficLightGroupArray;
+
+// --- Map construction --------------------------------------------------
+
+lanelet::Point3d make_point(lanelet::Id id, double x, double y, double z = 0.0)
+{
+  return lanelet::Point3d(id, x, y, z);
+}
+
+lanelet::Lanelet make_lanelet(double y_left, double y_right, const std::string & subtype)
+{
+  using namespace lanelet;  // NOLINT(build/namespaces)
+  LineString3d left(
+    utils::getId(),
+    {make_point(utils::getId(), 0.0, y_left), make_point(utils::getId(), 10.0, y_left)});
+  LineString3d right(
+    utils::getId(),
+    {make_point(utils::getId(), 0.0, y_right), make_point(utils::getId(), 10.0, y_right)});
+  Lanelet ll(utils::getId(), left, right);
+  ll.attributes()[AttributeName::Subtype] = subtype;
+  ll.attributes()[AttributeName::Type] = AttributeValueString::Lanelet;
+  return ll;
+}
+
+lanelet::LineString3d make_traffic_light_bulb()
+{
+  // Geometry does not affect arbiter logic; pick lanelet midpoint at z=5.
+  constexpr double kBulbX = 5.0;
+  constexpr double kBulbZ = 5.0;
+  using namespace lanelet;  // NOLINT(build/namespaces)
+  LineString3d bulb(
+    utils::getId(), {make_point(utils::getId(), kBulbX, 0.0, kBulbZ),
+                     make_point(utils::getId(), kBulbX, 1.0, kBulbZ)});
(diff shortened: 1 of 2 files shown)


takam5f2:
@interimadd 

Could you review the PR before I ask the code owners to review that. Best,

perception/autoware_traffic_light_arbiter/test/test_traffic_light_arbiter_core.cpp
```diff
@@ -0,0 +1,1242 @@
...
+    utils::getId(),
+    {make_point(utils::getId(), 0.0, y_left), make_point(utils::getId(), 10.0, y_left)});
+  LineString3d right(
+    utils::getId(),
+    {make_point(utils::getId(), 0.0, y_right), make_point(utils::getId(), 10.0, y_right)});
+  Lanelet ll(utils::getId(), left, right);
+  ll.attributes()[AttributeName::Subtype] = subtype;
+  ll.attributes()[AttributeName::Type] = AttributeValueString::Lanelet;
+  return ll;
+}
```
sasakisasaki, line 88: Perhaps we can use `create_safe_lanelet` in `autoware_lanelet2_utils`. Please see [this document](https://github.com/autowarefoundation/autoware_core/tree/e66aabb63028df1263f9e09b835f383b3dbf4b37/common/autoware_lanelet2_utils#conversion) 👍

sasakisasaki (commented):
For lanelet object creation, I guess we can use [`autoware_lanelet2_utils`](https://github.com/autowarefoundation/autoware_core/tree/25b67f276b9721385b9ddfbf2eaa3fbd71b92acd/common/autoware_lanelet2_utils). For example,

takam5f2:
@a-maumau  Could you check the PR? It will **NOT change** the production code instead of just **adding a new test code**.

perception/autoware_traffic_light_arbiter/test/test_traffic_light_arbiter_core.cpp
takam5f2: @sasakisasaki  

Thanks — `create_safe_lanelet` is a clean utility for production-side construction. For this fixture the central need is the `Subtype` attribute (`road` × 2, `crosswalk` × 1 — the crosswalk row is what makes `set_map()` expose the pedestrian id and unlocks the pedestrian-specific tests). Since `create_safe_lanelet` doesn't expose `Subtype`, swapping it in still requires unwrapping the optional, `remove_const`, and assigning attributes by hand on top.

Given that the primary need is the Subtype assignment rather than safe geometry construction, and that I'd like to keep the test self-contained, the inline helper felt like the right fit here.

a-maumau (approved):
LGTM!

PR #12725, opened 2026-06-08
fix(label_based_euclidean_cluster): change shape_policy argument type to string

technolojin:
## Description
fix launcher error
```
1780899091.3378823 [ERROR] [launch]: Caught exception in launch (see debug for traceback): Caught multiple exceptions when trying to load file of format [py]:
 - TypeError: 'int' object is not iterable
 - InvalidFrontendLaunchFileError: The launch file may have a syntax error, or its format is unknown
```

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


technolojin:
fixed https://github.com/autowarefoundation/autoware_universe/pull/12724

PR #12728, opened 2026-06-08
feat(autoware_multi_object_tracker): handle bounding box and polygon

technolojin:
## Description

Extends `autoware_multi_object_tracker` to retain and propagate polygon footprint data through vehicle and pedestrian trackers, even when the primary shape state is a bounding box.

**Before this PR**, `VehicleTracker` silently dropped polygon input — at init it forced the measurement to `BOUNDING_BOX`, and no footprint was stored or exported. `PedestrianTracker` similarly skipped polygon shape updates at measurement time. Tracker merge only kept the winner's shape type; no footprint data was preserved.

### New features

- **Polygon footprint storage in `VehicleTracker`**: the tracker now stores a polygon footprint (from cluster/polygon measurements) in tracker-local frame alongside its bounding-box state. The footprint is time-stamped and expires after 1 s of no update. On export, `getTrackedObject` composes both the bbox dimensions (from the motion model and shape blend) and the stored footprint into the output object — downstream consumers receive both shape representations simultaneously.

- **Footprint transform on each measurement**: when a polygon measurement arrives, its footprint is re-expressed relative to the post-update tracker pose, so the stored polygon remains consistent with the tracker frame through prediction steps.

- **Yaw-flip compensation**: when the yaw-limit correction reverses the tracked heading 180°, the stored footprint is flipped symmetrically to stay aligned.

- **Footprint union on tracker merge**: `TrackerOverlapManager` now calls `mergeFootprintFrom()` when absorbing a tracker. The absorbed tracker's polygon footprint is transformed into the winner's frame and unioned (Boost.Geometry union, convex-hull fallback for disjoint cases). `MultipleVehicleTracker` forwards this call to both inner trackers.

- **Full shape-type handling in `PedestrianShapeModel`**: polygon input is projected onto the tracker heading before blending; cylinder input is normalized to a square diameter; bounding-box input blends directly. Export selects CYLINDER or BOUNDING_BOX based on the length/width aspect ratio.

- **`z` and `orientation` in motion models**: all motion models (bicycle, CTRV, CV, static) now own `z_` and `orientation_` as residual kinematic state (not EKF-estimated). Both are set at init, low-pass-updated on measurement, and exported through every `getPredictedState()` call — removing the separate z/orientation bookkeeping that was scattered across tracker classes.

- **Uniform kinematic interface**: `getMotionState()` and `getStateTime()` implemented on all tracker types, providing a consistent interface for querying kinematic state without going through `getTrackedObject()`.

### Refactors

- Shape management extracted from `VehicleTracker` into `VehicleShapeModel`; from pedestrian/polygon/static trackers into `PedestrianShapeModel`, `PolygonShapeModel`, `StaticShapeModel` — all under a common `ShapeModelBase`
- `determineUpdateStrategy` and `createPseudoMeasurement` moved from `VehicleTracker`'s private methods to a standalone `vehicle_update_strategy.cpp`
- Tracker implementations moved from `tracker/model/` to `tracker/trackers/`; shape managers live under `tracker/shape_model/`
- `multi_vehicle_tracker` is altered by `general_vehicle_tracker` in config, suggesting the best default. launcher will be updated by https://github.com/autowarefoundation/autoware_launch/pull/1872
- `shapes::unionFootprints()` and `shapes::transformFootprint()` added as reusable geometry utilities

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

visualized with rviz plugin update https://github.com/autowarefoundation/autoware_rviz_plugins/pull/43

using ptv3 pipeline is helpful for various lidar cluster inputs. https://github.com/autowarefoundation/autoware_launch/pull/1865

https://github.com/user-attachments/assets/5cd2b360-2674-4299-ba92-cb98d00d0dab

https://github.com/user-attachments/assets/9ad171f3-73a3-4526-be06-1f485d7c363f

[
TESTED TIER IV INTERNAL](https://evaluation.tier4.jp/evaluation/reports/1d04d784-40b2-541d-bb0b-b66f3aa7b985?project_id=autoware_dev)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


perception/autoware_multi_object_tracker/lib/association/tracker_overlap_manager.cpp
```diff
@@ -224,10 +224,14 @@ void TrackerOverlapManager::merge(
...
 
-        // Shape: prefer lower shape type (bounding box < cylinder < convex hull)
+        // Shape: prefer lower shape type (bounding box < cylinder < convex hull).
         if (data1.object.shape.type > data2.object.shape.type) {
           data1.tracker->setObjectShape(data2.object.shape);
         }
+        // If the absorbed tracker carries polygon footprint data, union it into the winner.
+        if (data2.object.shape.type == autoware_perception_msgs::msg::Shape::POLYGON) {
+          data1.tracker->mergeFootprintFrom(data2.object.shape.footprint, data2.object.pose);
+        }
```
YoshiRi: [need to be checked]
Not only Polygon but also BBOX could have footprint.
So we need to check footprint.

```suggestion
        if (!data2.object.shape.footprint.points.empty()) {
          data1.tracker->mergeFootprintFrom(data2.object.shape.footprint, data2.object.pose);
        }
```

perception/autoware_multi_object_tracker/lib/tracker/shape_model/vehicle_shape_model.cpp
```diff
@@ -0,0 +1,157 @@
...
+    std::clamp(height_, object_model_.size_limit.height_min, object_model_.size_limit.height_max);
+}
+
+std::optional<double> VehicleShapeModel::setShape(
+  const autoware_perception_msgs::msg::Shape & shape, const rclcpp::Time & latest_measurement_time)
+{
+  width_ = std::clamp(
+    shape.dimensions.y, object_model_.size_limit.width_min, object_model_.size_limit.width_max);
+  height_ = std::clamp(
+    shape.dimensions.z, object_model_.size_limit.height_min, object_model_.size_limit.height_max);
```
YoshiRi: [question]
This will update width_ and height_ even if shape is polygon/cylindar. Do you think that is intended behaviour? When object has valid width but once polygon cames, it will set 0 as input?

perception/autoware_multi_object_tracker/lib/association/tracker_overlap_manager.cpp
YoshiRi: [concerns]
Could this use data1.object.pose as the destination frame? mergeFootprintFrom() currently only gets
the absorbed pose, so VehicleTracker uses its internal object_.pose as the winner frame. If the
object used by TrackerOverlapManager has been predicted to the merge/evaluation timestamp,
data1.object.pose and object_.pose may differ, and the absorbed footprint will be transformed into
a stale winner frame. Passing winner_pose explicitly would make the frame contract unambiguous.

YoshiRi (changes requested):

perception/autoware_multi_object_tracker/lib/tracker/shape_model/vehicle_shape_model.cpp
technolojin: fixed https://github.com/autowarefoundation/autoware_universe/pull/12728/changes/9f914c788e8fb138e3b688c2271bc4d2a37df313

perception/autoware_multi_object_tracker/lib/association/tracker_overlap_manager.cpp
technolojin: fixed https://github.com/autowarefoundation/autoware_universe/pull/12728/changes/10ec21bb9748af97a4f6f86abe2f881bca244c04 for the concern

technolojin: for the feedback, updated https://github.com/autowarefoundation/autoware_universe/pull/12728/changes/94a58f04c0b1f01ff373aa505fb5226cc380ac62

technolojin:
@ogrex we may need to test by scenarios for performance check

ogrex (commented):
MULTIPLE_VEHICLE is still used in test_bench.cpp.

For other issues, they have already been addressed and fixed. No other issues from my side.

YoshiRi (approved):
Thank you. LGTM

PR #12731, opened 2026-06-09
feat(perception, proximity_hazard_object_checker): enable new module

zulfaqar-azmi-t4:
## Description
Introduces the `autoware_proximity_hazard_object_checker` package — a new ROS 2 component node that publishes per-sector proximity hazard data for the L2 HMI alert function.

The node runs on a configurable timer (default 5 Hz). On each tick it polls the latest predicted objects and vehicle odometry, transforms all object poses into the base_link frame via TF, and evaluates each object against 8 configurable angular sectors around the vehicle. For each sector it reports whether any hazard is present, the distance to the closest object, and the full `PredictedObject` for that closest detection. Objects classified as UNKNOWN can optionally be suppressed, and a polygon-to-footprint distance threshold discards anything beyond the configured detection range.

Detection range (`max_detection_range_m`) is measured as the minimum polygon-to-ego-footprint distance (via `boost::geometry::comparable_distance`), so the threshold applies to the actual surface of the vehicle, not to a point or bounding circle.

The eight sectors tile the full 360° circle and their angular extents are independently configurable. Default extents are narrow fore/aft wedges (±10°) flanked by narrow shoulder wedges (±10–30°) and wide side wedges (±30–150°), with REAR wrapping through ±180°.

A debug topic publishes a `MarkerArray` showing the detection-zone outline (Minkowski sum of the vehicle footprint with a disc of radius `max_detection_range_m`), sector boundary lines, and sector name labels in the vehicle frame.

## Related links

- https://github.com/tier4/tier4_autoware_msgs/pull/222
- https://github.com/autowarefoundation/autoware_launch/pull/1868

## How was this PR tested?
No automated tests are included in this package. The node can be exercised via the provided launch file:

```
ros2 launch autoware_proximity_hazard_object_checker proximity_hazard_object_checker.launch.xml
```

which remaps inputs from `/perception/object_recognition/objects` and `/localization/kinematic_state` and outputs to `/hmi/proximity_hazards`. The debug marker topic (`~/debug/sector_markers`) can be visualized in RViz to confirm sector geometry and detection-zone extent.

## Notes for reviewers
- The sector wedge polygons are built once at construction time and cached; they are **not** recomputed if parameters change at runtime (all `sector_range.*` params are declared `read_only: true`).
- `sector_boundary_tolerance_deg` expands each wedge slightly on both edges to avoid objects that fall exactly on a sector boundary being missed — this means objects near a boundary can appear in two adjacent sectors simultaneously.
- The circumradius pre-filter is a conservative bounding-circle check applied before the more expensive polygon intersection; it cannot produce false negatives.
- The TF lookup uses the odometry `child_frame_id` as the target frame (typically `base_link`), so the node does not hard-code a frame name.

#### Additions and removals
|:------------|:-----------|:-----------|:-------------|:------------|
| Added | Sub | `/perception/object_recognition/objects` | `autoware_perception_msgs/PredictedObjects` | Predicted objects from object recognition pipeline (remappable via `input_objects` launch arg) |
| Added | Sub | `/localization/kinematic_state` | `nav_msgs/Odometry` | Vehicle odometry; `child_frame_id` used as the TF target frame (remappable via `input_odometry` launch arg) |
| Added | Pub | `/hmi/proximity_hazards` | `tier4_external_api_msgs/ProximityHazardObjects` | Per-sector proximity hazard output (remappable via `output_hazards` launch arg) |
| Added | Pub | `~/debug/sector_markers` | `visualization_msgs/MarkerArray` | Debug visualization: detection-zone outline, sector boundary lines, sector labels |

#### Additions and removals
| Added | `node_hz` | `double` | `5.0` | Timer frequency in Hz |
| Added | `max_detection_range_m` | `double` | `2.0` | Max polygon-to-ego-footprint distance (m); objects beyond this threshold are discarded |
| Added | `include_unknown_objects` | `bool` | `true` | If `false`, objects whose classification contains `UNKNOWN` are dropped before sector assignment |
| Added | `sector_boundary_tolerance_deg` | `double` | `1.0` | Angular tolerance (°) added to each side of every sector wedge boundary |
| Added | `sector_range.front` | `double_array` | `[-10.0, 10.0]` | FRONT sector `[from_deg, to_deg]` (REP-103: +x forward, +y left, 0° = ahead) |
| Added | `sector_range.front_left` | `double_array` | `[10.0, 30.0]` | FRONT_LEFT sector `[from_deg, to_deg]` |
| Added | `sector_range.left` | `double_array` | `[30.0, 150.0]` | LEFT sector `[from_deg, to_deg]` |
| Added | `sector_range.rear_left` | `double_array` | `[150.0, 170.0]` | REAR_LEFT sector `[from_deg, to_deg]` |
| Added | `sector_range.rear` | `double_array` | `[170.0, -170.0]` | REAR sector; `to_deg < from_deg` wraps through ±180° |
| Added | `sector_range.rear_right` | `double_array` | `[-170.0, -150.0]` | REAR_RIGHT sector `[from_deg, to_deg]` |
| Added | `sector_range.right` | `double_array` | `[-150.0, -30.0]` | RIGHT sector `[from_deg, to_deg]` |
| Added | `sector_range.front_right` | `double_array` | `[-30.0, -10.0]` | FRONT_RIGHT sector `[from_deg, to_deg]` |

## Effects on system behavior
- A new topic `/hmi/proximity_hazards` (`tier4_external_api_msgs/ProximityHazardObjects`) begins publishing at 5 Hz once the node is running. Downstream HMI consumers must subscribe to this topic to receive proximity alert data.
- The node introduces a 5 Hz TF lookup from the input objects frame to base_link. If the TF tree is incomplete or stale, the node emits a throttled warning and skips that cycle without affecting other nodes.
- Polling subscribers for objects and odometry are non-blocking; if either input is unavailable on a given tick, a throttled warning is emitted and the publish is skipped for that cycle.
- No existing nodes or topics are modified.

### DIFF
--- a/perception/autoware_proximity_hazard_object_checker/config/proximity_hazard.yaml
+++ b/perception/autoware_proximity_hazard_object_checker/config/proximity_hazard.yaml
@@ -1,0 +1,19 @@
+/**:
+  ros__parameters:
+    node_hz: 5.0
+
+    max_detection_range_m: 2.0
+
+    include_unknown_objects: true
+
+    sector_range:
+      front: [-10.0, 10.0]
+      front_left: [10.0, 30.0]
+      left: [30.0, 150.0]
+      rear_left: [150.0, 170.0]
+      rear: [170.0, -170.0]
+      rear_right: [-170.0, -150.0]
+      right: [-150.0, -30.0]
+      front_right: [-30.0, -10.0]
+
+    sector_boundary_tolerance_deg: 1.0
--- a/perception/autoware_proximity_hazard_object_checker/launch/proximity_hazard_object_checker.launch.xml
+++ b/perception/autoware_proximity_hazard_object_checker/launch/proximity_hazard_object_checker.launch.xml
@@ -1,0 +1,13 @@
+<launch>
+  <arg name="input_objects" default="/perception/object_recognition/objects"/>
+  <arg name="input_odometry" default="/localization/kinematic_state"/>
+  <arg name="output_hazards" default="/hmi/proximity_hazards"/>
+  <arg name="param_path" default="$(find-pkg-share autoware_proximity_hazard_object_checker)/config/proximity_hazard.yaml"/>
+
+  <node pkg="autoware_proximity_hazard_object_checker" exec="autoware_proximity_hazard_object_checker_node" name="proximity_hazard_object_checker" output="screen">
+    <remap from="~/input/odometry" to="$(var input_odometry)"/>
+    <remap from="~/input/objects" to="$(var input_objects)"/>
+    <remap from="~/output/proximity_hazards" to="$(var output_hazards)"/>
+    <param from="$(var param_path)"/>
+  </node>
+</launch>
--- a/perception/autoware_proximity_hazard_object_checker/CMakeLists.txt
+++ b/perception/autoware_proximity_hazard_object_checker/CMakeLists.txt
@@ -1,0 +1,28 @@
+cmake_minimum_required(VERSION 3.14)
+project(autoware_proximity_hazard_object_checker)
+
+find_package(autoware_cmake REQUIRED)
+autoware_package()
+
+generate_parameter_library(${PROJECT_NAME}_param
+  param/parameter_struct.yaml
+)
+
+ament_auto_add_library(${PROJECT_NAME}_component SHARED
+  src/proximity_hazard_object_checker.cpp
+  src/proximity_hazard_object_checker_node.cpp
+)
+
+target_link_libraries(${PROJECT_NAME}_component
+  ${PROJECT_NAME}_param
+)
+
+rclcpp_components_register_node(${PROJECT_NAME}_component
+  PLUGIN "autoware::proximity_hazard_object_checker::ProximityHazardObjectCheckerNode"
+  EXECUTABLE ${PROJECT_NAME}_node
+)
+
+ament_auto_package(INSTALL_TO_SHARE
+  launch
+  config
+)
--- a/perception/autoware_proximity_hazard_object_checker/include/autoware/proximity_hazard_object_checker/proximity_hazard_object_checker.hpp
+++ b/perception/autoware_proximity_hazard_object_checker/include/autoware/proximity_hazard_object_checker/proximity_hazard_object_checker.hpp
@@ -1,0 +1,77 @@
+// Copyright 2026 TIER IV, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// cspell : ignore circumradius
+
+#ifndef AUTOWARE__PROXIMITY_HAZARD_OBJECT_CHECKER__PROXIMITY_HAZARD_OBJECT_CHECKER_HPP_
+#define AUTOWARE__PROXIMITY_HAZARD_OBJECT_CHECKER__PROXIMITY_HAZARD_OBJECT_CHECKER_HPP_
+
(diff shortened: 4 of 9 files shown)


technolojin:
@zulfaqar-azmi-t4 

For me, this function seems not a perception component, since this algorithm determines what the hazardous object is.

* definition of hazard
* definition of hazardous situation

Since this kind of consideration requires driving context, it could not be in the perception module. The perception module is responsible only for the perceived world.

perception/autoware_proximity_hazard_object_checker/config/proximity_hazard.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+    sector_range:
+      front: [-10.0, 10.0]
+      front_left: [10.0, 30.0]
+      left: [30.0, 150.0]
+      rear_left: [150.0, 170.0]
+      rear: [170.0, -170.0]
+      rear_right: [-170.0, -150.0]
+      right: [-150.0, -30.0]
+      front_right: [-30.0, -10.0]
```
technolojin, line 17: - is the unit degree?
- is the configuration allows overlap?
- what will happen there is gap between sectors?

perception/autoware_proximity_hazard_object_checker/launch/proximity_hazard_object_checker.launch.xml
```diff
@@ -0,0 +1,13 @@
+<launch>
+  <arg name="input_objects" default="/perception/object_recognition/objects"/>
+  <arg name="input_odometry" default="/localization/kinematic_state"/>
+  <arg name="output_hazards" default="/hmi/proximity_hazards"/>
```
technolojin, line 4: the topics shall not have default.
please see other perception node launchers.

zulfaqar-azmi-t4:
@technolojin 

Thank you for the  feedback! 

I completely agree with your architectural distinction: if this module were performing semantic risk assessment or evaluating driving context (like TTC or trajectory crossing), it would absolutely belong outside of Perception.

However, the current implementation is purely a geometric proximity filter of the perceived world:  
- It has no driving context: It does not look at ego velocity, path planning, lanes, or traffic rules.  
- It is purely spatial: It simply takes the perception outputs and uses boost::geometry to check if any object polygon falls within a static distance threshold relative to the vehicle footprint.  
- It simply sorts these perceived objects into 8 spatial buckets (sectors) based on raw coordinates, regardless of whether the object actually poses a behavioral threat.  Because it is strictly filtering and structuring the 'perceived world' based on proximity, it fits structurally within the perception/object aggregation pipeline.  

That said, I see how the word 'Hazard' in the component and message names implies behavioral risk assessment that isn't actually happening here. Since we are in the review phase, I am completely open to renaming the module to something more accurate to its geometric nature.

technolojin (changes requested):
Dose this algorithm considers object size?
There are various size of objects. In case of trailer, it has 20 m length.

It would be better to have a readme file explaining concept of the algorithm.

As I commented, I do not think this package and node is not a perception component.

technolojin:
This algorithm refers ego vehicle footprint. 
The perception only shall be related with sensor system (sensor position and calibration parameters, record timings). Ego vehicle shape shall not be added to the perception dependency. 

"purely spatial" dose not make this node "perception". the localization is also purely spatial.

zulfaqar-azmi-t4:
@technolojin 

Thank you so much the comment.

> The perception only shall be related with sensor system. Ego vehicle shape shall not be added to the perception dependency.

The statement is a bit inconsistent because `autoware_ground_segmentation`, depend on wheel base.

technolojin:
@zulfaqar-azmi-t4 

ground_segmentation uses vehicle_info_.wheel_base_m to make **virtual sensor position**.
this is limitation of rule-based algorithm. 

By [latest work](https://github.com/autowarefoundation/autoware_launch/pull/1865), the refereed node will be removed from representative systems.

perception/autoware_proximity_hazard_object_checker/config/proximity_hazard.yaml
```diff
@@ -0,0 +1,19 @@
+/**:
+  ros__parameters:
+    node_hz: 5.0
+
+    max_detection_range_m: 2.0
```
technolojin, line 5: max_detection_range_m decides whether given section and objects are in hazard proximity or not. Perception shall not **determine** hazard proximity.

PR #12733, opened 2026-06-09
fix(autoware_compare_map_segmentation): drop stale TF override in voxel_based filter

ksuszka:
## Description

`VoxelBasedCompareMapFilterComponent` overrides `input_indices_callback` and `convert_output_costly` with a private `tf2_ros::Buffer` and a zero-timeout `lookupTransform`. When the transform is not yet available (e.g. "extrapolation into the future" while the latest `map`<-`base_link` transform trails the cloud stamp) the lookup throws immediately and the catch falls back to the original, untransformed cloud (`cloud_tf = cloud`). The map comparison then runs in the sensor frame instead of `map`, silently producing an incorrectly segmented output that downstream consumers act on.

The override is a documented stopgap (`TODO(badai-nguyen)`: delete once `ManagedTransformBuffer` is deployed). It is deployed now: the base `pointcloud_preprocessor::Filter` constructs `managed_tf_buffer_` unconditionally in `setupTF()`, and every other compare-map filter variant already relies on it. The override merely reimplements the base logic with worse TF handling.

Delete both overrides so the node inherits the base behavior, consistent with all other filters:
- 1.0 s lookup timeout that waits for the transform instead of failing instantly;
- drop the frame on transform failure rather than publishing a wrong-frame result;
- empty-cloud handling is preserved natively by `ManagedTransformBuffer::transformPointcloud` (`cloud_in.data.empty()` -> copy and set the target `frame_id`), so the fix from #10572 is retained without the bespoke override code.

Also remove the now-unused `tf_buffer_`/`tf_listener_` members, the unused `PointIndices*` aliases and the `tf2` includes. No `package.xml` change: `autoware_pointcloud_preprocessor` (already a dependency) provides `managed_transform_buffer` transitively.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

A set of internal tests in simulation environment was run using this fixed version and log messages about TF time mismatch were present no more.

## Notes for reviewers

From comments in the code it seems that this change was planned from the beginning.

@badai-nguyen, can you confim that it is the right direction?

## Interface changes

None.

## Effects on system behavior

Small delays in TF propagation are handled in the same way as with other filters.


YoshiRi (approved):
Thanks for you suggest. I agree with you.

PR #12742, opened 2026-06-10
fix(autoware_system_monitor): migrate deprecated NVML APIs for CUDA 13

Max-Bin:
## Description

CUDA 13 bumps the NVML API to version 13, which deprecates `nvmlDeviceGetTemperature()` and `nvmlDeviceGetCurrentClocksThrottleReasons()`. Since `autoware_system_monitor` is compiled with `-Werror=deprecated-declarations`, this turns into a hard build failure on the CUDA 13 toolkit:

```
error: 'nvmlReturn_t nvmlDeviceGetTemperature(...)' is deprecated [-Werror=deprecated-declarations]
error: 'nvmlReturn_t nvmlDeviceGetCurrentClocksThrottleReasons(...)' is deprecated [-Werror=deprecated-declarations]
cc1plus: all warnings being treated as errors
```

This PR migrates the two deprecated NVML calls in the NVML GPU monitor to their replacements:

- `nvmlDeviceGetTemperature()` → `nvmlDeviceGetTemperatureV()` (using the versioned `nvmlTemperature_t` struct)
- `nvmlDeviceGetCurrentClocksThrottleReasons()` → `nvmlDeviceGetCurrentClocksEventReasons()`

Both calls are wrapped in small file-local helpers guarded by `#if defined(NVML_API_VERSION) && NVML_API_VERSION >= 13`, so older CUDA toolkits keep using the original APIs and continue to build unchanged. The returned reason bitmask is identical between the throttle/event variants, so the existing decoding logic and `reasonToString` macro are untouched.

## Related links

**Parent Issue:**

- N/A

## How was this PR tested?

Built `autoware_system_monitor` with `colcon build --cmake-args -DCMAKE_BUILD_TYPE=Release` on a CUDA 13.2 toolkit (NVML API v13, driver 595, TensorRT 10.16) using Autoware's default `-Werror` flags. Before this change the build failed with the deprecation errors shown above; after it, the package compiles cleanly with no deprecation warnings and no workaround flags. The `#else` branch (CUDA < 13) was reviewed to confirm it still calls the original APIs unchanged.

## Notes for reviewers

- Only the two deprecated **function** calls are replaced. The `nvmlClocksThrottleReason*` constants used by `reasonToString` and the throttle switch are preprocessor `#define`s (CUDA 13 aliases them to `nvmlClocksEventReason*`) and produce no compiler warning, so they are intentionally left as-is to keep the diff minimal.
- `nvmlTemperature_t::temperature` is an `int`; it is cast back to `unsigned int` to preserve the existing variable type and downstream formatting.

## Interface changes

None.

## Effects on system behavior

None. The replacement APIs return the same temperature value and the same clock event/throttle reason bitmask, so monitor behavior is identical on supported drivers.


system/autoware_system_monitor/src/gpu_monitor/nvml_gpu_monitor.cpp
```diff
@@ -33,6 +33,38 @@
...
+// keeps building on both CUDA 13+ and older toolkits.
+nvmlReturn_t getGpuTemperature(nvmlDevice_t device, unsigned int * temp)
+{
+#if defined(NVML_API_VERSION) && NVML_API_VERSION >= 13
+  nvmlTemperature_t temperature_info{};
+  temperature_info.version = nvmlTemperature_v1;
+  temperature_info.sensorType = NVML_TEMPERATURE_GPU;
+  const nvmlReturn_t ret = nvmlDeviceGetTemperatureV(device, &temperature_info);
+  if (ret == NVML_SUCCESS) {
+    *temp = static_cast<unsigned int>(temperature_info.temperature);
```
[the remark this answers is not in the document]
Max-Bin: fixed in https://github.com/autowarefoundation/autoware_universe/pull/12742/changes/0a9f89bd6baed407798743567c9b84079683e2cb

system/autoware_system_monitor/src/gpu_monitor/nvml_gpu_monitor.cpp
nishikawa-masaki, line 545: The definitions of numClocksThrottleReasonGpuIdle, nvmlClocksThrottleReasonApplicationsClocksSetting and nvmlClocksThrottleReasonSwPowerCap also have been deprecated.

If this block is left untouched, it will be invalidated in the future.

I think this block should be defined as a private function and the function should have 2 different (but quite similar) implementations depending on the SDK version.

system/autoware_system_monitor/src/gpu_monitor/nvml_gpu_monitor.cpp
```diff
@@ -477,14 +516,14 @@ std::vector<GPUMonitorBase::GpuStatus> GPUMonitor::getGPUStatus() const
...
 
-    unsigned int temp = 0;
-    ret = nvmlDeviceGetTemperature(itr->device, NVML_TEMPERATURE_GPU, &temp);
+    int temp = 0;
+    ret = getGpuTemperature(itr->device, &temp);
     if (ret != NVML_SUCCESS) {
       continue;
     }
 
     unsigned long long clocksThrottleReasons = 0LL;  // NOLINT
```
nishikawa-masaki, line 525: NVIDIA renamed nvmlDeviceGetCurrentClocksThrottleReasons() to nvmlDeviceGetCurrentClocksEventReasons() because some of the reasons are not "throttling".

In that context, "clocksEventReasons" is better, I believe.

nishikawa-masaki (changes requested):
First of all, thank you very much for your contribution and I'm very sorry for not reviewing this pull request. No excuse.

Your changes are just fine to make the source code compilable with CUDA 13, but I want to make it compilable with future SDKs when the deprecated names are deleted.

system/autoware_system_monitor/src/gpu_monitor/nvml_gpu_monitor.cpp
Max-Bin: Addressed in commits 4a63b4309cd and 78f3fa1e5ff. The ignored-reason classification is now handled by the file-local isIgnoredClockEventReason() helper, shared by both call sites. The SDK-specific constants are selected at compile time: NVML API >= 13 uses the nvmlClocksEventReason* names, while the deprecated throttle aliases are confined to the legacy < 13 branch. This keeps the code compatible with future SDKs where those aliases are removed.

system/autoware_system_monitor/src/gpu_monitor/nvml_gpu_monitor.cpp
Max-Bin: Addressed in commits 4a63b4309cd and 78f3fa1e5ff. The local variables and helper are now named clocksEventReasons and getClocksEventReasons(), and the NVML API >= 13 path calls nvmlDeviceGetCurrentClocksEventReasons(). The diagnostic error message was updated to refer to clock event reasons as well.

system/autoware_system_monitor/src/gpu_monitor/nvml_gpu_monitor.cpp
```diff
@@ -33,6 +33,113 @@
...
+  nvmlClocksThrottleReasonApplicationsClocksSetting;
+constexpr unsigned long long kClockEventReasonSwPowerCap =  // NOLINT(runtime/int)
+  nvmlClocksThrottleReasonSwPowerCap;
+constexpr unsigned long long kClockEventReasonSyncBoost =  // NOLINT(runtime/int)
+  nvmlClocksThrottleReasonSyncBoost;
+constexpr unsigned long long kClockEventReasonSwThermalSlowdown =  // NOLINT(runtime/int)
+  nvmlClocksThrottleReasonSwThermalSlowdown;
+constexpr unsigned long long kClockEventReasonDisplayClockSetting =  // NOLINT(runtime/int)
+  nvmlClocksThrottleReasonDisplayClockSetting;
+#endif
```
nishikawa-masaki: There are 9 reasons, but only 6 of them are defined as constants.

system/autoware_system_monitor/src/gpu_monitor/nvml_gpu_monitor.cpp
```diff
@@ -33,6 +33,113 @@
...
+
+constexpr ClockEventReasonName kClockEventReasonNames[] = {
+  {kClockEventReasonGpuIdle, "GpuIdle"},
+  {kClockEventReasonApplicationsClocksSetting, "ApplicationsClocksSetting"},
+  {kClockEventReasonSwPowerCap, "SwPowerCap"},
+  {nvmlClocksThrottleReasonHwSlowdown, "HwSlowdown"},
+  {kClockEventReasonSyncBoost, "SyncBoost"},
+  {kClockEventReasonSwThermalSlowdown, "SwThermalSlowdown"},
+  {nvmlClocksThrottleReasonHwThermalSlowdown, "HwThermalSlowdown"},
+  {nvmlClocksThrottleReasonHwPowerBrakeSlowdown, "HwPowerBrakeSlowdown"},
```
nishikawa-masaki: Line 77, 80, 81 use raw "nvml" constants.
Those "nvml" constants should be replaced with "kClock" constants, which are not defined yet.

P.S.
Now I see that those "nvml" constants are not deprecated and they can be used "as-is".
(I had been understanding that all of "nvmlClocksThrottle" constants were deprecated).

However, I think defining "kClockThrottleReason" (not "kClockEventReason") constants for them will improve readability.

system/autoware_system_monitor/src/gpu_monitor/nvml_gpu_monitor.cpp
nishikawa-masaki, line 65: There are 9 reasons, but only 6 of them are defined as constants.

nishikawa-masaki (changes requested):
It looks like a partial fix.
If I missed anything, please let me know.

system/autoware_system_monitor/src/gpu_monitor/nvml_gpu_monitor.cpp
Max-Bin: Addressed in commit 73ffccc5c35. Added file-local kClockEventReasonHwSlowdown, kClockEventReasonHwThermalSlowdown, and kClockEventReasonHwPowerBrakeSlowdown constants, and updated all three reason-table entries to use them.

system/autoware_system_monitor/src/gpu_monitor/nvml_gpu_monitor.cpp
Max-Bin: Addressed in commit 73ffccc5c35. The reason formatter now defines and uses all nine clock event reasons through kClockEventReason* constants. The six SDK-specific aliases remain selected by NVML_API_VERSION, while the three hardware slowdown constants are shared because their NVML names remain unchanged in the event-reason API. Verified with a CUDA 13.2 -Werror build and a legacy API syntax compile.

nishikawa-masaki:
2 CI checks fail.
[build-test-tidy-pr / build-and-test-packages-above-differential (humble) / build-and-test-p](https://github.com/autowarefoundation/autoware_universe/actions/runs/35082152603/job/104748449561?pr=12742)
[build-test-tidy-pr / build-and-test-packages-above-differential (jazzy) / build-and-test-pack](https://github.com/autowarefoundation/autoware_universe/actions/runs/35082152603/job/104748449383?pr=12742)

These problems will be solved by the following PR in autoware_tools, which has already been merged.
[Autoware Tool PR 465](https://github.com/autowarefoundation/autoware_tools/pull/465)

nishikawa-masaki (approved):
Thank you for the fix.
It looks good to me.

PR #12746, opened 2026-06-10
feat: add mrm reset manager

TetsuKawa:
## Description

Autoware can be deployed in a **redundant configuration**, where the same set of nodes runs more than once — typically in separate ROS 2 domains (e.g. a `main` ECU and a `sub` ECU) so that one side can take over when the other fails.

Two pieces of state in that configuration are *latched* on purpose and therefore need an explicit reset:

- **The diagnostic graph latch.** Once a fault is latched, the graph keeps reporting the error even after the cause is gone, so the error cannot be cleared by simply recovering.
- **The redundancy switching state.** After a switchover (or during start-up, before the vehicle is actually ready), the redundancy switcher and its interface hold a state that must be brought back to a defined baseline.

Until now there was no single owner for that reset sequence: whoever wanted to clear an MRM had to know which services to call, in which order, and which of them only exist in a redundant deployment. The initialization phase had the same problem — the aggregator and the redundancy-switcher interface must be told "I am still initializing" before Autoware is ready, and told the opposite once localization, route and control are all in place.

This PR adds **`autoware_mrm_reset_manager`**, the node that owns exactly this: it forwards a single external `reset_mrm` request to the underlying reset services, and it drives the initializing / ready lifecycle of the aggregator and the redundancy-switcher interface from AD API state.

The node deliberately handles **one ECU path only** — it has no main/sub split logic. Fanning a single call out to the redundant instances is the job of `autoware_generic_service_divider`, which sits behind these output services.

```text
  external          +--------------------------------+
  reset  --------> ~/input/reset_mrm                 |
  request           |   autoware_mrm_reset_manager   |
                    |                                |
 /api/localization/initialization_state --> |        |
 /api/routing/state ----------------------> | ready? |
 /api/operation_mode/state ---------------> |        |
                    |                                |
                    |   +--> ~/output/reset_diag_graph
                    |   +--> ~/output/reset_redundancy_switcher
                    |   +--> ~/output/set_aggregator_initializing
                    |   +--> ~/output/set_redundancy_switcher_interface_initializing
                    +--------------------------------+
```

sub-ecu-block-diagram-mrm-v0 6-service・API_ref drawio

### What is added

A single component node, `autoware::mrm_reset_manager::MrmResetManager` (executable `autoware_mrm_reset_manager_node`, running on a `MultiThreadedExecutor`), plus its launch file, parameter file and README.

The node has three responsibilities.

#### 1. `reset_mrm` forwarding

When `~/input/reset_mrm` (`tier4_external_api_msgs/srv/ResetMrm`) is called, the node calls

1. `~/output/reset_diag_graph` (`tier4_system_msgs/srv/ResetDiagGraph`) — clears the latched diagnostics, and
2. `~/output/reset_redundancy_switcher` (`tier4_system_msgs/srv/ResetRedundancySwitcher`) — resets the redundancy switching state; skipped and treated as success when `is_redundant:=false`.

**Both calls are always attempted**, even if the first one fails, so a partial failure still resets whatever can be reset. The response is `ResponseStatus::SUCCESS` only when both succeeded; otherwise `ERROR` carries the first failure message (or `reset_mrm forwarding failed` if none was produced). The request payload itself is not used — the service is a trigger.

#### 2. Initialization state machine

A 1 s timer advances a linear state machine. States fall through, so several steps can complete within one tick.

| State | Action |
| ----- | ------ |
| `WAIT_SERVICES_READY` | Wait until `reset_diag_graph` and `set_aggregator_initializing` are ready, plus `reset_redundancy_switcher` and `set_redundancy_switcher_interface_initializing` when `is_redundant:=true`. Logs which service it is waiting for once per second. |
| `SET_AGGREGATOR_INIT` | `set_aggregator_initializing(true)` |
| `RESET_SWITCHER` | `reset_redundancy_switcher()` |
| `SET_SWITCHER_IFACE_INIT` | `set_redundancy_switcher_interface_initializing(true)` |
| `DONE` | Cancel the init timer, start the 5 s periodic timer, and evaluate the ready state |

Any failing step returns without advancing, so the same step is retried on the next tick. This makes start-up ordering independent of the launch order of the downstream nodes.

#### 3. Ready-state transition

The node subscribes with `transient_local`, depth 1, to `~/input/localization_initialization_state`, `~/input/route_state` and `~/input/operation_mode_state`, and re-evaluates the ready state on every update as well as from the 5 s timer.

Once the state machine reached `DONE`, all three messages have been received, and at least one initializing flag is still set, the node checks

- localization state `== INITIALIZED`,
- route state `== SET`,
- `is_autoware_control_enabled == true`,

and when all three hold — and `enable_autoware_ready_actions:=true` — it leaves the initializing phase:

1. `set_aggregator_initializing(false)`
2. `reset_redundancy_switcher()`
3. `set_redundancy_switcher_interface_initializing(false)`

While still initializing and not yet ready, the 5 s timer keeps retrying `reset_redundancy_switcher()`, which keeps the switching state at the baseline for as long as the vehicle is not operational.

### Implementation notes

- **Every service call is synchronous** (`future.wait_for(service_timeout_ms)`) inside a callback. To make that safe, the service server and each of the four clients get their **own `MutuallyExclusive` callback group** and the node runs on a `MultiThreadedExecutor`. Changing the executor or merging the callback groups would deadlock.
- `state_mutex_` guards the state machine, the two initializing flags and the three cached AD API messages.
- All external names are remapped in the launch file, so the node itself only knows `~/input/...` and `~/output/...`.

## Related links

**Parent Issue:**

- https://github.com/autowarefoundation/autoware_universe/issues/12691

## How was this PR tested?

- Built with `colcon build` and passed the package lint tests (`ament_lint_auto` / `pre-commit`). The package has no unit tests yet — see *Notes for reviewers*.
- Behaviour to exercise when reviewing: start the node with the downstream services missing (it must stay in `WAIT_SERVICES_READY` and log which service it waits for), then bring them up (it must walk through the state machine to `DONE` and set both initializing flags), then publish `INITIALIZED` / `SET` / `is_autoware_control_enabled=true` (it must clear both flags and call `reset_redundancy_switcher` once), and finally call `~/input/reset_mrm` (it must forward to `reset_diag_graph` and `reset_redundancy_switcher` and return `SUCCESS`).

## Notes for reviewers

Points that deserve a second opinion, all visible in the code as it stands:

1. **`set_redundancy_switcher_interface_initializing` is not gated by `is_redundant`.** `WAIT_SERVICES_READY` only waits for that service when `is_redundant:=true`, but `SET_SWITCHER_IFACE_INIT` calls it unconditionally. With `is_redundant:=false` and no server for it, the state machine stalls there, retrying once per second. Either gate the call like `reset_redundancy_switcher()` does, or document that the service must exist in both configurations.
2. **`service_timeout_ms` has no default in code** (`declare_parameter<int>("service_timeout_ms")`), so the node fails to start unless a parameter file provides it. The shipped `config/mrm_reset_manager.param.yaml` sets `800`, while `is_redundant` and `enable_autoware_ready_actions` are passed as launch arguments rather than through the parameter file.
3. **The README is out of sync with the code** on two points: it documents `service_timeout_ms` default `200` (the config sets `800`, and the code has no default), and its launch-argument table lists `/system/mrm_reset_manager/...` output defaults, whereas `launch/mrm_reset_manager.launch.xml` actually defaults to `/diagnostics_graph/reset`, `/system/redundancy_switcher/reset`, `/system/aggregator/set_initializing` and `/system/redundancy_switcher_interface/set_initializing`. Worth fixing in this PR.
4. **Blocking calls are made while `state_mutex_` is held**, so a slow downstream service stalls the subscription callbacks for up to `service_timeout_ms` per call. Acceptable at the current rates, but it is the reason the timeouts must stay short.
5. **The node publishes no diagnostics**, so a permanent stall in `WAIT_SERVICES_READY` is only visible in the log. Consider whether it should expose readiness like the divider node does.

## Effects on system behavior

None by itself — this PR only adds a new package with its own launch file, and nothing in the existing launch tree starts it. No existing topic, service or parameter is touched.

When a deployment does include the node, the following becomes relevant:

- **`reset_mrm` gains a single entry point.** Callers no longer need to know about `reset_diag_graph` or `reset_redundancy_switcher`, or which of them exists in a redundant deployment.
- **Start-up gains an explicit initializing phase.** From launch until localization is initialized, a route is set and Autoware control is enabled, the aggregator and the redundancy-switcher interface are held in "initializing", and the redundancy switching state is re-reset every 5 s. Downstream nodes must tolerate that flag being asserted for the whole start-up window, and being cleared exactly once when the vehicle becomes ready.
- **Start-up order is tolerated, not enforced.** The node waits for its downstream services instead of failing, so it can be launched before them; the cost is that a permanently missing service leaves it silently stuck in `WAIT_SERVICES_READY` (see *Notes for reviewers*).
- **`enable_autoware_ready_actions:=false` keeps the system in the initializing state indefinitely**, which is intended for bring-up and debugging rather than normal operation.


TetsuKawa:
@mkuri 
## Update

Pushed two fixes and a doc update for the non-redundant (`is_redundant=false`) configuration:

### 1. Guard switcher-interface init when non-redundant

`set_redundancy_switcher_interface_initializing()` was being called unconditionally in both the init state machine (`SET_SWITCHER_INTERFACE_INIT`) and `leave_initializing_phase()`, even though its backing service is only expected to exist when `is_redundant=true` (`are_required_services_ready()` already skips waiting for it otherwise). On a non-redundant configuration, this call timed out every tick, so the init state machine never reached `DONE` and the aggregator's `initializing` flag (diag graph latch) was never cleared.

Added a `set_redundancy_switcher_interface_initializing(bool)` wrapper that short-circuits to success when `is_redundant_` is `false`, mirroring the existing pattern used for `call_reset_redundancy_switcher()`.

### 2. Keep periodic reset firing until actually ready

`on_periodic_reset_check()` gated the periodic `reset_redundancy_switcher` retry on `is_autoware_ready()`, but that only checks whether the localization/route/operation-mode topics have been received at least once — not whether their values indicate readiness. Once all three topics had fired once (even while still uninitialized / route-unset / not in autonomous control), the periodic reset stopped retrying and never resumed, even though the system had not actually left the initializing phase.

Now also gates on `is_ready_for_operation()`, so the periodic reset keeps firing every tick until localization is initialized, the route is set, and autonomous control is enabled.

### 3. README

Documented that `is_redundant=false` skips `set_redundancy_switcher_interface_initializing` in addition to `reset_redundancy_switcher`, in both the init-state-machine table and the ready-state-transition section, and updated the `is_redundant` parameter description to match.

mkuri (approved):

PR #12748, opened 2026-06-10
fix(autoware_carla_interface): decouple sensor publishing from the synchronous tick loop

oguzkaganozt:
## Description

The interface converts and publishes all sensor data inline in the same thread that paces `world.tick()`. A publish call can block on DDS flow control — a 1600x900 BGRA image with `reliable` QoS is ~5.8 MB per frame, far above typical writer watermarks — so transport conditions stall the simulation clock itself. Once the simulation slows, every downstream consumer receives stale data and Autoware's data-freshness gates reject or drop autonomous engagement.

Measured on a cloud VM (28 vCPU EPYC @ 2.45 GHz, NVIDIA L40, CARLA 0.9.16 Town01, OpenADKit e2e demo, CycloneDDS): with a single subscribed camera, tick-loop iterations inflate from 7 ms (no cameras) to ~200 ms, dragging the simulation from 20 fps to ~7 fps. A QoS A/B test isolated the dominant cost to the blocking reliable publish (switching the camera to `best_effort` dropped the loop to 21–67 ms with no other change).

This PR makes camera/lidar handling unable to stall the simulation clock:

- **`SensorPublishWorker`** (new): one daemon thread per heavy sensor with a bounded latest-wins queue. Camera and lidar conversion/publishing run on these workers; the tick thread never blocks on serialization or DDS flow control. When a consumer cannot keep up, stale frames are dropped instead of slowing the simulation (sensor frames are perishable, and the drop count is logged at debug level).
- **Zero-subscriber skip**: conversion and publishing are skipped entirely when a camera/lidar topic has no subscribers, checked per frame so late subscribers start receiving immediately. Spawned-but-unconsumed sensors now cost nothing on the ROS side.
- **Launch consistency**: the always-on CAM_FRONT traffic-light relays and `cam_front_republish` now respect `use_light_weight_sensor_mapping`, like the other republish nodes added in #12525. These nodes subscribe the raw image and force full-rate conversion/publishing of CAM_FRONT even in deployments that disable traffic-light recognition.

Design notes:

- Frequency gating (`checkFrequency`) and registry bookkeeping stay on the tick thread, so the sensor registry is never accessed concurrently.
- Messages are stamped with the timestamp captured when their frame was enqueued, not the publish-time `self.timestamp`.
- Each ROS publisher is used by exactly one worker thread; workers are stopped before publishers are destroyed in `shutdown()`.

## Related links

- Builds on the intent of #12525 (light-weight sensor configuration).
- Found while validating autowarefoundation/openadkit#73 (CARLA e2e sample) on cloud GPU VMs.

## How was this PR tested?

Manually, on the cloud VM described above, running the full OpenADKit modular Autoware stack end to end against CARLA 0.9.16:

| Scenario | Before | After |
|---|---|---|
| 1 camera 1600x900 reliable, 2 subscribers | tick-loop iter ~200 ms, sim ~6.7 fps | **12–28 ms**, sim ~2x faster |
| 6 cameras spawned, no subscribers | n/a (always consumed) | zero ROS-side cost (skip path) |
| Cameras off (regression) | lidar 10 Hz / odom 20 Hz, drives | **identical: 9.7 / 19.4 Hz, autonomous drive verified (3.0 m)** |

## Notes for reviewers

With cameras spawned, CARLA's own per-tick render/stream cost remains the next bottleneck on server-class GPUs (we measured the simulation at ~8–10 fps with HD cameras spawned even with zero ROS-side work). That layer is outside this interface; on such hosts the light-weight sensor mapping remains the right tool. This change removes the interface's own contribution to the stall and makes partial camera consumption (e.g. RViz viewing one stream) safe.

## Interface changes

When `use_light_weight_sensor_mapping:=true`, the traffic-light relay nodes and `cam_front_republish` are no longer launched (previously they always ran). Default behavior (`false`) is unchanged.

## Effects on system behavior

Under camera load the simulation clock stays near real time instead of collapsing; camera frames may be dropped under sustained overload (latest-wins), which replaces the previous behavior of slowing the whole simulation.

### DIFF
--- a/simulator/autoware_carla_interface/launch/autoware_carla_interface.launch.xml
+++ b/simulator/autoware_carla_interface/launch/autoware_carla_interface.launch.xml
@@ -68,9 +68,24 @@
     <!-- <node pkg="tf2_ros" exec="static_transform_publisher" name="carla_velodyne_top" args="0 0 1 -1.5386 -0.015 0.001 velodyne_top velodyne_top_changed"/> -->
     <!-- <node pkg="tf2_ros" exec="static_transform_publisher" name="carla_imu" args="0 0 1 -3.10519265 -0.015 -3.14059265359 tamagawa/imu_link tamagawa/imu_link_changed"/> -->
 
-    <!-- Relay CAM_FRONT to traffic_light namespace for traffic light recognition -->
-    <node pkg="topic_tools" exec="relay" name="traffic_light_image_relay" args="/sensing/camera/CAM_FRONT/image_raw /sensing/camera/traffic_light/image_raw"/>
-    <node pkg="topic_tools" exec="relay" name="traffic_light_camera_info_relay" args="/sensing/camera/CAM_FRONT/camera_info /sensing/camera/traffic_light/camera_info"/>
+    <!-- Relay CAM_FRONT to traffic_light namespace for traffic light recognition.
+         These relays subscribe the raw image, which forces the interface to
+         convert and publish every CAM_FRONT frame; skip them in light-weight
+         mode like the republish nodes below. -->
+    <node
+      pkg="topic_tools"
+      exec="relay"
+      name="traffic_light_image_relay"
+      args="/sensing/camera/CAM_FRONT/image_raw /sensing/camera/traffic_light/image_raw"
+      unless="$(var use_light_weight_sensor_mapping)"
+    />
+    <node
+      pkg="topic_tools"
+      exec="relay"
+      name="traffic_light_camera_info_relay"
+      args="/sensing/camera/CAM_FRONT/camera_info /sensing/camera/traffic_light/camera_info"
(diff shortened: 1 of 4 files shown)


oguzkaganozt:
Validated end-to-end on CARLA 0.9.16 + OpenADKit (NVIDIA L40S): the sensor-publish decoupling and the `use_light_weight_sensor_mapping` launch gating are both exercised, and the stack drives closed-loop with 6 cameras at ~20 fps.

Note: this is a necessary fix and stabilizes the run; for full HD-camera frame rate it should be paired with `best_effort` camera QoS (the residual ceiling is the GIL-bound reliable-QoS publish, not the bridge architecture).

maxime-clem (approved):
Tested locally and it runs fine.

PR #12749, opened 2026-06-11
feat(autoware_universe_designs): add node designs for common packages

vish0012:
## Description

Adds the missing `*.node.yaml` design files (Autoware System Design Format 0.3.0) for all 5 component nodes under `common/`:

- `autoware_path_distance_calculator` / PathDistanceCalculator
- `autoware_goal_distance_calculator` / GoalDistanceCalculatorNode
- `autoware_component_interface_tools` / ServiceLogChecker
- `autoware_traffic_light_recognition_marker_publisher` / TrafficLightRecognitionMarkerPublisher
- `autoware_glog_component` / GlogComponent

All interfaces were derived directly from each node's implementation and CMakeLists.

An inventory of `main` (188 registered component nodes vs. plugins documented in `autoware_universe_designs`) shows ~151 nodes without a central design file. This PR completes the `common` category as the first batch; further batches will follow per category.

## Related links

https://github.com/autowarefoundation/autoware_system_designer
https://github.com/autowarefoundation/autoware_universe/tree/main/common/autoware_universe_designs/design

## How was this PR tested?

- `pre-commit run lint-system-design-format` passes on all 5 files; interfaces were cross-checked against each node's source code.
- Launch test (as recommended in review): created a `CommonDebugTools.module.yaml` instantiating these nodes, added it as a component to `AutowareSample.system.yaml`, built the deployment for all 4 modes, and launched the nodes from the generated launch files — verified with `ros2 node list` and `ros2 param get`. Details in the test report comment below.

## Notes for reviewers

- Adds a new `design/common/` category folder, following the existing per-category layout of `autoware_universe_designs` (categories have been added incrementally as coverage expanded).
- `ServiceLogChecker` outputs only via `diagnostic_updater`; this is documented as a global `/diagnostics` publisher. `GlogComponent` has no runtime interfaces, so its file uses `processes: []` (same as `AutonomousEmergencyBraking.node.yaml`).
- Param file paths use the relative form for the node's own package config (applied from review feedback in b35e0c3).

## Interface changes

None.

## Effects on system behavior

None. Documentation-only change.


common/autoware_universe_designs/design/common/GlogComponent.node.yaml
technolojin, line 1: You can find the same glog component here

https://github.com/autowarefoundation/autoware_launch/blob/main/autoware_sample_designs/design/ros_system/AutowareGlogComponent.node.yaml

common/autoware_universe_designs/design/common/GoalDistanceCalculator.node.yaml
```diff
@@ -0,0 +1,56 @@
...
+    message_type: autoware_internal_debug_msgs/msg/Float64Stamped
+    remap_target: goal_distance_calculator/deviation/yaw
+  - name: deviation/yaw_deg
+    message_type: autoware_internal_debug_msgs/msg/Float64Stamped
+    remap_target: goal_distance_calculator/deviation/yaw_deg
+
+# parameters
+param_files:
+  - name: goal_distance_calculator_param
+    default: $(find-pkg-share autoware_goal_distance_calculator)/config/goal_distance_calculator.param.yaml
```
technolojin:
```suggestion
    default: config/goal_distance_calculator.param.yaml
```
relative path is acceptable for the package itself.

technolojin (commented):

I recommend to make a small module, add to [the sample system ](https://github.com/autowarefoundation/autoware_launch/blob/main/autoware_sample_designs/design/system/AutowareSample.system.yaml) and test if the nodes are actually launched. 

I would like to listen your experience about the autoware system designer.

common/autoware_universe_designs/design/common/GoalDistanceCalculator.node.yaml
vishalchauhan-hi: Thank you for your valuable feedback

common/autoware_universe_designs/design/common/GlogComponent.node.yaml
vishalchauhan-hi: Thank you

vishalchauhan-hi:
> Thank you for your contribution!
> 
> I recommend to make a small module, add to [the sample system ](https://github.com/autowarefoundation/autoware_launch/blob/main/autoware_sample_designs/design/system/AutowareSample.system.yaml) and test if the nodes are actually launched.
> 
> I would like to listen your experience about the autoware system designer.

## Launch test results
Thank you @technolojin 
I tested the nodes through the system designer as you recommended.

Setup: a separate test workspace with `autoware_system_designer` (main), `autoware_launch` (main), `autoware_universe_designs` from this branch, and `nebula` for the sensing entities. I made a small `CommonDebugTools.module.yaml` with the 4 nodes from this PR (I left out GlogComponent until the duplicate question above is decided) and added it to `AutowareSample.system.yaml` as a `common_tools` component, connected to `map.publisher.vector_map`.

Results:
- Deployment built for all 4 modes, and `common_tools.launch.xml` was generated for each mode with the remaps and parameters from the design files.
- I launched the generated PlanningSimulation launcher. All 4 nodes start under `/common_tools/...` and `ros2 param get /common_tools/goal_distance_calculator update_rate` returns `10.0`.
- Node behavior also matched the source: after publishing a static `map→base_link` TF, `goal_distance_calculator` went from waiting for pose to waiting for route, and `path_distance_calculator` from "no pose" to "no path".

One issue came up during the test, related to your param path suggestion. With the `$(find-pkg-share ...)` form, the param file was silently dropped (the package was not in the manifest package_map in my test workspace) and the node died at startup with `UninitializedStaticallyTypedParameterException: 'update_rate' must be initialized`. After changing to the relative `config/goal_distance_calculator.param.yaml` (b35e0c3), the parameters appear in the generated launch and the node runs fine. I will use the relative path style in the next batches too.

<details>
<summary>Terminal output</summary>

```
$ ros2 node list
/common_tools/goal_distance_calculator
/common_tools/path_distance_calculator
/common_tools/service_log_checker
/common_tools/traffic_light_recognition_marker_publisher

$ ros2 param get /common_tools/goal_distance_calculator update_rate
Double value is: 10.0
```

Generated launch entry:

```xml
<node pkg="autoware_goal_distance_calculator" exec="autoware_goal_distance_calculator_node" name="goal_distance_calculator" namespace="/common_tools" output="screen">
  <remap from="~/input/ros_transform" to="/tf"/>
  <remap from="~/input/ros_transform_static" to="/tf_static"/>
  <remap from="~/input/route" to="/planning/mission_planning/route"/>
  <remap from="goal_distance_calculator/deviation/lateral" to="/common_tools/goal_distance_calculator/deviation/lateral"/>
  <param name="update_rate" value="10.0"/>
  <param name="oneshot" value="False"/>
</node>
```

</details>

## My experience with the autoware system designer

The node/module/system structure is easy to understand and the generated launchers are helpful. Some points I struggled with during testing:

1. A file can pass `lint-system-design-format` but still fail at deploy time, because entity and param resolution only happen in `build_deploy`. Some earlier check would be nice.
2. After adding or changing design files, I had to rebuild `autoware_system_designer` itself before the changes were picked up. I couldn't find this documented anywhere and it took me a while to figure out.
3. When a param file path can't be resolved, it is silently dropped and the node crashes at startup with a parameter exception that doesn't mention the file. An explicit error with the unresolved path would have saved me a lot of debugging time.

technolojin:
From my experience, it is better to check the launched node info.
`ros2 node info /common_tools/goal_distance_calculator`
and check the topics are actually remapped.
(it is to check the remap target of all the topics)

Also, please visit visualization pages generated in your <work_directory>/install/systems.html -> Diagram

technolojin (approved):
LGTM

I did not checked the topic remaps are worked.

Contributor understood the system.

technolojin:
> A file can pass lint-system-design-format but still fail at deploy time, because entity and param resolution only happen in build_deploy. Some earlier check would be nice.

the linter only checks the format.

the build process checks combination of config files. you may can trigger the check by re-build the system package (ex. autoware_sample_designs, package that the system config is included) and it will take 1min.

technolojin (changes requested):
please remove GlogComponent.node, since https://github.com/autowarefoundation/autoware_launch/blob/main/autoware_sample_designs/design/ros_system/AutowareGlogComponent.node.yaml already exist.

vishalchauhan-hi:
> please remove GlogComponent.node, since https://github.com/autowarefoundation/autoware_launch/blob/main/autoware_sample_designs/design/ros_system/AutowareGlogComponent.node.yaml already exist.

>  @technolojin Removed in dd28540. Thank you

technolojin (approved):
LGTM

PR #12754, opened 2026-06-11
fix(multi object tracker): merge algorithm fix

technolojin:
## Description

Fixes correctness bugs in `TrackerOverlapManager::merge()` that caused missed merges, order-dependent outcomes, and incorrect fragment absorption. Also enables classification trust for the lidar clustering channel, which is now safe with the fixed merge logic.

### Problems fixed

**P1 — Sort criteria not aligned with win-lose policy**
The global sort used `measurement_count` / `elapsed_time` as tiebreakers, but `canMergeTarget` never read those fields. Pairs were checked in one direction only (`j > i`), so when sort order disagreed with the win-lose policy, merges were silently skipped.

**P2 — Mid-loop mutation causing order-dependent outcomes**
Winner existence-probabilities were updated immediately while later `isRedundant` calls still used stale geometry snapshots. Outcome depended on which pairs were processed first.

**P3 — Both-known `isRedundant` path did not handle fragments**
Used plain 2D IoU > threshold only. A classified fragment fully contained inside a larger bbox has low IoU but high recall, so it was not absorbed. This became the common case with lidar semantic segmentation clusters carrying trusted classification (over-segmented fragments spawn same-priority vehicle trackers, creating tied-priority pairs everywhere).

**P4 — No polygon-shape gate**
Polygon-only trackers could merge with each other, producing incorrect footprint unions.

### Algorithm rewrite (`merge()`)

1. **Snapshot pass** — per-tracker scalars captured once; full `DynamicObject` fetched lazily only for trackers that appear in a gated pair.
2. **Symmetric pair discovery** — R-tree query from each tracker using its own label radius; pairs deduped before evaluation.
3. **`decidePair(a, b)`** — pure, antisymmetric, deterministic function of snapshots:
   - *Shape gate*: both polygon-only → `NoMerge`. Winner must have `BBOX` or `CYLINDER` shape AND be confident.
   - *Lexicographic tiers*: priority → `isConfident` → known-probability band → symmetric per-channel existence dominance → covariance determinant → measurement count → UUID.
4. **Deferred apply** — all mutations batched; each loser merges into its highest-GIoU winner. Eliminates mid-loop geometry/probability inconsistency.

### `isRedundant` both-known path fix

Replaces plain IoU check with `get2dPrecisionRecallGIoU`. Merges when `IoU > threshold` **or** `recall > 0.5` (fragment fully contained inside the larger tracker). Restores classified-fragment absorption that plain IoU missed.

### Config change

`input_channels.param.yaml`: `can_trust_classification: true` for the `lidar_clustering` channel. Previously set to `false` as a workaround because the old merge algorithm could not safely handle same-priority classified fragments. Now enabled.

### Tests

New `test/test_tracker_overlap_manager.cpp` (6 tests):
- `PolygonOnlyPairsDoNotMerge` — two polygon trackers must not be merged.
- `BboxTrackerAbsorbsOverlappingPolygon` — bbox tracker absorbs overlapping polygon tracker.
- `ClassifiedFragmentAbsorbedByContainment` — contained classified fragment absorbed via recall criterion.
- `StarMergeAbsorbsMultipleFragmentsInOneCycle` — one winner absorbs several fragments in a single pass.
- `ChainDoesNotBridgeWithinOneCycle` — transitive chains do not collapse in one cycle (by design).
- `OutcomeIsIndependentOfListOrder` — shuffled tracker list produces identical merge outcome.

Old code fails `PolygonOnlyPairsDoNotMerge` and `ClassifiedFragmentAbsorbedByContainment`; new code passes all 10 package tests.

---

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

- All 10 `autoware_multi_object_tracker` unit tests pass (SimulatedData, RealDataRosbag, UUID, TrackerOverlapManager x6).
- Differential check: old code fails 2 of the 6 new tests; new code passes all.
- Prune-pass latency benchmark (~350 trackers, 3 runs): new avg 0.85 ms vs old 0.87 ms — no regression.

## Notes for reviewers

- **CYLINDER treated as BBOX-equivalent winner** in the shape gate — pedestrian trackers can absorb polygon fragments. Can be tightened to BBOX-only if disputed.
- **Chain convergence**: transitive overlaps (A∩B, B∩C) are not resolved in a single cycle by design — chains converge over 2-3 frames given the merge throttle. One-to-many association is out of scope.
- The R-tree `satisfies`-only query is a linear scan (~0.4 ms at 350 trackers); the `bgi::intersects(box)` spatial pre-filter must not be removed.
- `canMergeTarget()` is removed from the public API; all pair logic is now internal to `decidePair()`.

## Interface changes

None (no topic or service additions or removals).

#### Modifications

| Version | Parameter Name | Type | Default Value | Description |
|:--------|:---------------|:-----|:--------------|:------------|
| Old | `lidar_clustering.can_trust_classification` | `bool` | `false` | Whether to trust classification from lidar clustering channel |
| New | `lidar_clustering.can_trust_classification` | `bool` | `true` | Whether to trust classification from lidar clustering channel |

## Effects on system behavior

1. **Polygon-only tracker pairs no longer merge** — previously an unclassified cluster could absorb another, producing a garbage footprint union.
2. **Classified fragments from lidar semantic segmentation are now correctly absorbed** by the overlapping bbox tracker (containment criterion via recall check).
3. **Merge outcome is independent of tracker list order** — previously tied-priority pairs could merge or not depending on spawn history.
4. **Lidar clustering classification is now trusted** — semantic segmentation labels from the clustering channel are used for tracker type assignment.

### DIFF
--- a/perception/autoware_multi_object_tracker/CMakeLists.txt
+++ b/perception/autoware_multi_object_tracker/CMakeLists.txt
@@ -106,6 +106,7 @@ if(BUILD_TESTING)
     test/test_bench_association.cpp
     test/test_vehicle_tracker.cpp
     test/test_uuid_generator.cpp
+    test/test_tracker_overlap_manager.cpp
   )
   add_definitions(-D_SRC_RESOURCES_DIR_PATH="${PROJECT_SOURCE_DIR}/test/data/")
   ament_add_ros_isolated_gtest(test_multi_object_tracker ${test_files})
--- a/perception/autoware_multi_object_tracker/config/input_channels.param.yaml
+++ b/perception/autoware_multi_object_tracker/config/input_channels.param.yaml
@@ -14,7 +14,7 @@
           can_spawn_new_tracker: true
           can_trust_existence_probability: false
           can_trust_extension: false
-          can_trust_classification: false
+          can_trust_classification: true
           can_trust_orientation: false
         optional:
           name: "clustering"
--- a/perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/association/tracker_overlap_manager.hpp
+++ b/perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/association/tracker_overlap_manager.hpp
@@ -30,14 +30,25 @@
 namespace autoware::multi_object_tracker
 {
 
-/// Detects and merges spatially overlapping trackers produced by different sensors or detectors.
-/// Higher-priority trackers absorb lower-priority ones when sufficient overlap is confirmed.
+/// Detects and merges spatially overlapping trackers produced by different sensors or detectors
+/// (including over-segmented detections that one-to-one association cannot fuse).
+///
+/// Guarantees:
+/// - The merge outcome is a pure function of the tracker states: pair direction comes from a
+///   lexicographic dominance comparison ending in the tracker UUID, so results do not depend on
+///   tracker_list order.
+/// - The winner of a merge must be confident and carry a parametric shape (bounding box or
+///   cylinder); polygon-only trackers may only be absorbed — two polygon-only trackers never
+///   merge.
+/// - Applied merges form a star forest per cycle (no tracker both absorbs and is absorbed);
+///   longer chains converge over subsequent cycles with direct geometry checks.
 class TrackerOverlapManager
 {
 public:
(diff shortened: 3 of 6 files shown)


YoshiRi (approved):
As for code base, LGTM

PR #12758, opened 2026-06-11
feat(autoware_ptv3): add detection3d head

amadeuszsz:
## Description

Adds the PTv3 Detection3D head and publishes decoded bounding boxes as `DetectedObjects`. The existing backbone is shared by the segmentation and detection heads, each head runs only when its output has subscribers.

Key changes:

- **Detection head TensorRT integration:** loads `*_det3d_head.onnx` / `*.engine`, binds the exported point features/grid coordinates and the fixed N-proposal output contract, and shares the backbone output buffers with the segmentation head.
- **GPU postprocessing:** decodes center, dimensions, height, yaw, and velocity, calculates the final query score, applies post-center-range, distance-binned per-class score, and yaw-norm filtering, compacts and sorts retained boxes on the CUDA stream.
- **ROS output:** converts boxes to `autoware_perception_msgs::msg::DetectedObjects`, applies area-based class remapping and BEV IoU NMS, and publishes `~/output/objects`.
- **Class mapping:** publishes `CAR`, `TRUCK`, `BUS`, `BICYCLE`, and `PEDESTRIAN` directly. `TRAFFIC_CONE` and `BARRIER` are represented as `HAZARD`, because the message does not define separate labels for them.
- **Configuration and validation:** adds Detection3D model metadata, score thresholds, yaw thresholds, post-center range, class-remapping matrices, NMS parameters, launch arguments, and JSON schema coverage. Invalid proposal counts, matrix dimensions, grid parameters, and thresholds fail during initialization.
- **Head-aware execution:** detection-only processing skips segmentation inference and large segmentation reconstruction-buffer clears.
- **Segmentation metadata:** updates the segmentation class names and palette to the taxonomy exported by the current PTv3 segmentation artifact. The filtered class changes from `drivable_surface` to `drivable_flat`.

## Related links

**Parent Issue:**

- Requires https://github.com/autowarefoundation/autoware_universe/pull/12655 merge first
- Requires https://github.com/autowarefoundation/autoware_universe/pull/12786

## How was this PR tested?

- Built in the Autoware development container:
  - `colcon build --packages-select autoware_ptv3 --symlink-install --cmake-args -DCMAKE_BUILD_TYPE=Release`
- Validated `ptv3.param.yaml` and all three ML-package parameter files against their JSON schemas.
- Replayed point clouds with the Detection3D output enabled. Bounding boxes are published.

## Notes for reviewers

- PTv3 maps traffic cones and barriers to `HAZARD` because `autoware_perception_msgs::msg::ObjectClassification` does not define these specific classes.

### Model contract

- Detection classes: `CAR`, `TRUCK`, `BUS`, `BICYCLE`, `PEDESTRIAN`, `TRAFFIC_CONE`, `BARRIER`.
- Proposal count: 500.
- Detection output tensors: `dense_heatmap`, `query_heatmap_score`, `query_labels`, `heatmap`, `center`, `height`, `dim`, `rot`, and `vel`.
- The current configuration remaps cars with BEV area `[12.1, 36.0] m²` to trucks.
- Cone and barrier score thresholds are intentionally `0.01`, the other classes use `0.1`.

### Engine compatibility

- A Detection3D engine must be built from `*_det3d_head.onnx`.
- The fixed voxel-origin change is runtime preprocessing and does not by itself require rebuilding an existing compatible backbone or segmentation engine.

### Topic changes

| Change type | Topic               | Type                                                   | Description                              |
|:------------|:--------------------|:-------------------------------------------------------|:-----------------------------------------|
| Added       | `~/output/objects`  | `autoware_perception_msgs::msg::DetectedObjects`       | Detection3D boxes after NMS and remapping. |

#### Additions

| Parameter Name                                                        | Type       | Default / example                         | Description |
|:----------------------------------------------------------------------|:-----------|:------------------------------------------|:------------|
| `detection3d.use_head`                                                | `bool`     | `true`                                    | Enable the Detection3D head. |
| `detection3d.onnx_path`                                               | `string`   | `*_det3d_head.onnx`                       | Detection head ONNX path. |
| `detection3d.engine_path`                                             | `string`   | `*_det3d_head.engine`                     | Detection head TensorRT engine path. |
| `detection3d.bbox_voxel_size`                                        | `double[]` | `[0.96, 0.96, 8.0]`                       | Detection BEV voxel size. |
| `detection3d.bbox_downsample_factor`                                  | `int64`    | `1`                                       | Detection output-grid downsample factor. |
| `detection3d.has_twist`                                               | `bool`     | `true`                                    | Whether the model exports velocity. |
| `detection3d.num_proposals`                                           | `int64`    | `500`                                     | Fixed number of detection queries. |
| `detection3d.post_center_range`                                       | `double[]` | `[-132.88, -132.88, -5.0, 132.88, 132.88, 12.0]` | Accepted decoded-box range. |
| `detection3d.class_names`                                             | `string[]` | Seven model classes                       | Model class order used by output tensors. |
| `detection3d.post_process_params.iou_nms_search_distance_2d`          | `double`   | `10.0`                                    | Center-distance gate for BEV IoU NMS. |
| `detection3d.post_process_params.iou_nms_threshold`                   | `double`   | `0.1`                                     | BEV IoU suppression threshold. |
| `detection3d.post_process_params.yaw_norm_thresholds`                 | `double[]` | `[0.1, 0.1, 0.1, 0.1, 0.0, 0.0, 0.0]`   | Per-class yaw-vector norm threshold. |
| `detection3d.detection_score_thresholds.distance_bin_upper_limits`    | `double[]` | `[50.0, 90.0, 121.0, 200.0]`             | Radial distance-bin upper bounds. |
| `detection3d.detection_score_thresholds.min_confidence_scores.<class>`| `double[]` | One value per distance bin                | Per-class score thresholds. |
| `detection3d.allow_remapping_by_area_matrix`                          | `int64[]`  | 7 x 7 matrix                              | Allowed source-to-target class remaps. |
| `detection3d.min_area_matrix`                                         | `double[]` | 7 x 7 matrix                              | Minimum BEV area for each remap. |
| `detection3d.max_area_matrix`                                         | `double[]` | 7 x 7 matrix                              | Maximum BEV area for each remap. |

### Launch argument changes

| Change type | Argument                              | Default                                | Description |
|:------------|:--------------------------------------|:---------------------------------------|:------------|
| Added       | `output/objects`                      | `~/objects`                            | Detection output remap target. |
| Added       | `use_det3d_head`                      | `true`                                 | Enable the Detection3D head. |
| Added       | `det3d_head_ml_package_param_path`    | `ml_package_*_det3d_head.param.yaml`   | Detection artifact metadata file. |

## Effects on system behavior

- PTv3 publishes 3D detections whenever `~/output/objects` has subscribers.
- Segmentation and detection can run independently while sharing one backbone inference.
- Detection-only execution avoids segmentation inference and reconstruction-buffer clearing.
- Traffic cones and barriers are present in the output as `HAZARD`.
- The segmentation output class IDs and palette change to the current 15-class model taxonomy.


amadeuszsz:
@codex review

perception/autoware_ptv3/launch/ptv3.launch.xml
```diff
@@ -33,9 +36,11 @@
         <remap from="~/output/pointcloud/visualization/cuda" to="$(var output/pointcloud/visualization)/cuda"/>
         <remap from="~/output/pointcloud/filtered" to="$(var output/pointcloud/filtered)"/>
         <remap from="~/output/pointcloud/filtered/cuda" to="$(var output/pointcloud/filtered)/cuda"/>
+        <remap from="~/output/objects" to="$(var output/objects)"/>
         <param from="$(var model_param_path)" allow_substs="true"/>
         <param from="$(var backbone_ml_package_param_path)"/>
         <param from="$(var seg3d_head_ml_package_param_path)"/>
+        <param from="$(var det3d_head_ml_package_param_path)"/>
```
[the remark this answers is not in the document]
amadeuszsz: Nope, we deploy all heads. Setting it as false won't break the runtime. Handling it conditionally would mess up the launch file. Provided launch file is just sample use case, dedicated pipelines will use own configuration.

perception/autoware_ptv3/schema/ptv3.schema.json
```diff
@@ -80,6 +126,63 @@
       ],
       "additionalProperties": false
     },
+    "detection3d": {
+      "type": "object",
+      "properties": {
+        "use_head": {
+          "description": "Enable the detection head."
+        },
```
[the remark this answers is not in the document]
amadeuszsz: Nope, we use substitutions. Schema read it as string unfortunately. We missed `type` field on purpose.

perception/autoware_ptv3/lib/postprocess/detection3d_postprocess.cu
```diff
@@ -0,0 +1,200 @@
...
+  __device__ bool operator()(const Box3D & lhs, const Box3D & rhs) const
+  {
+    return lhs.score > rhs.score;
+  }
+};
+
+// Decode one Detection3D proposal per thread into a Box3D. Rejected proposals keep score 0 and are
+// compacted away by process(). All boundary tensors are pinned to float32 and int64 via NetworkIO,
+// so no runtime dtype dispatch is needed. The `heatmap` input is the per-query class logit, not the
+// dense BEV heatmap.
```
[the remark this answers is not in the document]
amadeuszsz: https://github.com/autowarefoundation/autoware_universe/pull/12758/commits/091edbcedb42c494cfb9ba7dfb539f0d6b86627c

perception/autoware_ptv3/lib/ptv3_trt.cpp
```diff
@@ -143,27 +159,53 @@ void PTv3TRT::initPtr()
     autoware::cuda_utils::make_unique<std::int32_t[]>(config_.max_num_voxels_ * 3);
   bb_point_offset_d_ = autoware::cuda_utils::make_unique<std::int64_t[]>(1);
```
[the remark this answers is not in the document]
amadeuszsz: https://github.com/autowarefoundation/autoware_universe/pull/12758/commits/82fe52cb2dad936356407b33910abf462444ac18

perception/autoware_ptv3/config/ptv3.param.yaml
```diff
@@ -17,3 +17,49 @@
...
+      detection_score_thresholds:
+        distance_bin_upper_limits: [50.0, 90.0, 121.0, 200.0]
+        min_confidence_scores:
+          CAR: [0.1, 0.1, 0.1, 0.1]
+          TRUCK: [0.1, 0.1, 0.1, 0.1]
+          BUS: [0.1, 0.1, 0.1, 0.1]
+          BICYCLE: [0.1, 0.1, 0.1, 0.1]
+          PEDESTRIAN: [0.1, 0.1, 0.1, 0.1]
+          TRAFFIC_CONE: [0.01, 0.01, 0.01, 0.01]
+          BARRIER: [0.01, 0.01, 0.01, 0.01]
```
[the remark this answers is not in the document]
amadeuszsz: Updated description

mojomex:
@amadeuszsz Is it possible to factor out the below modules from bevfusion/centerpoint/ptv3 into a common package?:

It would be nice since it would spare us from reviewing these algorithms as part of this PR.

> BEV IoU NMS: package-local adaptation of [LidarTransfusion NMS](https://github.com/autowarefoundation/autoware_universe/blob/e148902c205a7d207097f3759bf3115d1c5fbf4c/perception/autoware_lidar_transfusion/lib/postprocess/non_maximum_suppression.cpp#L27-L98). The same algorithm is also used by [LidarCenterPoint](https://github.com/autowarefoundation/autoware_universe/blob/e148902c205a7d207097f3759bf3115d1c5fbf4c/perception/autoware_lidar_centerpoint/lib/postprocess/non_maximum_suppression.cpp#L27-L100) and [BEVFusion](https://github.com/autowarefoundation/autoware_universe/blob/e148902c205a7d207097f3759bf3115d1c5fbf4c/perception/autoware_bevfusion/lib/postprocess/non_maximum_suppression.cpp#L29-L100).

> Detection class remapper: uses the same row-major area-matrix interpretation and first-match remapping logic as [LidarTransfusion](https://github.com/autowarefoundation/autoware_universe/blob/e148902c205a7d207097f3759bf3115d1c5fbf4c/perception/autoware_lidar_transfusion/lib/detection_class_remapper.cpp#L23-L72), [LidarCenterPoint](https://github.com/autowarefoundation/autoware_universe/blob/e148902c205a7d207097f3759bf3115d1c5fbf4c/perception/autoware_lidar_centerpoint/lib/detection_class_remapper.cpp#L23-L72), and [BEVFusion](https://github.com/autowarefoundation/autoware_universe/blob/e148902c205a7d207097f3759bf3115d1c5fbf4c/perception/autoware_bevfusion/lib/detection_class_remapper.cpp#L24-L73).

YoshiRi:
Worked on my desktop. I found this multi-head yields >8GB VRAM so I could not test it in the same time. But anyway It worked.

| det3d | seg3d|
|--|--|
|image|image|

perception/autoware_ptv3/config/ml_package_ptv3_det3d_head.param.yaml
```diff
@@ -0,0 +1,9 @@
+/**:
+  ros__parameters:
+    detection3d:
+      bbox_voxel_size: [0.96, 0.96, 8.0]
```
mojomex, line 4: (note to self) I imagine this has to be aligned with/be a multiple of the voxel size of the backbone? Is that explicitly checked somewhere?

perception/autoware_ptv3/config/ml_package_ptv3_det3d_head.param.yaml
```diff
@@ -0,0 +1,9 @@
...
+      bbox_downsample_factor: 1
```
mojomex: What does this option do? Is it needed?

perception/autoware_ptv3/include/autoware/ptv3/postprocess/detection3d_postprocess.hpp
```diff
@@ -0,0 +1,101 @@
...
+   * @return Number of decoded boxes kept after filtering.
+   */
+  [[nodiscard]] std::size_t numBoxes() const { return num_boxes_; }
+
+private:
+  PTv3Config config_;
+
+  thrust::device_vector<Box3D> raw_boxes_d_;
+  thrust::device_vector<Box3D> passing_boxes_d_;
+  thrust::device_vector<float> yaw_norm_thresholds_d_;
```
mojomex, line 90: Using Thrust for memory management, and for algorithms that need a temporary workspace (e.g. sorting) makes it very hard to keep synchronous CUDA calls out of the hot path - and thus should be avoided.

I will point this out when I see it, but it is okay to do this in a follow-up PR (I'm fine to do it too).

perception/autoware_ptv3/schema/ml_package_ptv3.schema.json
```diff
@@ -80,6 +80,49 @@
...
+    "detection3d": {
+      "type": "object",
+      "properties": {
+        "bbox_voxel_size": {
+          "type": "array",
+          "items": { "type": "number", "exclusiveMinimum": 0 },
+          "minItems": 3,
+          "maxItems": 3
+        },
+        "bbox_downsample_factor": {
```
mojomex: Not self-explanatory, please add description.

perception/autoware_ptv3/schema/ml_package_ptv3.schema.json
```diff
@@ -80,6 +80,49 @@
...
+          "type": "integer",
+          "minimum": 1
+        },
+        "has_twist": {
```
mojomex, line 94: Not self-explanatory, please add description.

perception/autoware_ptv3/schema/ml_package_ptv3.schema.json
```diff
@@ -80,6 +80,49 @@
...
+          "type": "boolean"
+        },
+        "num_proposals": {
+          "type": "integer",
+          "minimum": 1
+        },
+        "post_center_range": {
```
mojomex, line 103: Not self-explanatory, please add description.

perception/autoware_ptv3/lib/ros_utils.cpp
```diff
@@ -0,0 +1,105 @@
...
+    return Label::ANIMAL;
+  }
+  if (class_name == "TRAFFIC_CONE") {
+    return Label::HAZARD;
+  }
+  if (class_name == "BARRIER") {
+    return Label::HAZARD;
+  }
+  return Label::UNKNOWN;
+}
```
mojomex: This seems like something that's probably already there in other models, and should be unified. We can do it after this PR.

perception/autoware_ptv3/lib/ptv3_trt.cpp
mojomex, line 1: This file is getting huge - splitting into four:
* this file, as data routing layer
* backbone
* semseg head
* 3d det head

would be cleaner/easier to maintain.

perception/autoware_ptv3/include/autoware/ptv3/ptv3_config.hpp
```diff
@@ -38,22 +39,31 @@ class PTv3Config
...
+    const std::vector<std::int64_t> & palette = {},
+    const float filter_class_probability_threshold = {},
+    const std::vector<std::string> & filter_classes = {},
+    const std::string & filter_output_format = {}, const std::string & source_reconstruction = {},
+    const std::vector<std::string> & detection_class_names = {},
+    const std::vector<float> & bbox_voxel_size = {}, std::size_t bbox_downsample_factor = {},
+    const std::vector<float> & distance_bin_upper_limits = {},
+    const std::vector<float> & detection_score_thresholds = {},
+    const std::vector<float> & yaw_norm_thresholds = {}, const bool has_twist = {},
+    std::size_t num_proposals = {}, const std::vector<float> & post_center_range = {})
```
mojomex, line 57: This also exploded in size.
Let's have separate
* backbone
* semseg
* 3d det
* (common?)
config objects.

perception/autoware_ptv3/include/autoware/ptv3/ptv3_config.hpp
```diff
@@ -205,23 +299,40 @@ class PTv3Config
...
+  std::size_t bbox_downsample_factor_{};
+  bool has_twist_{};
+  std::size_t num_proposals_{};
+  std::vector<float> post_center_range_;
+  std::vector<float> distance_bin_upper_limits_;
+  std::vector<float> detection_score_thresholds_;
+  std::vector<float> yaw_norm_thresholds_;
+  std::size_t det_grid_x_size_{};
+  std::size_t det_grid_y_size_{};
+
```
mojomex, line 341: Refactor these into their respective structs

perception/autoware_ptv3/include/autoware/ptv3/ptv3_config.hpp
```diff
@@ -205,23 +299,40 @@ class PTv3Config
 
   // Head selection
   bool use_seg3d_head_;
+  bool use_det3d_head_;
```
mojomex, line 309: Can be encoded as `std::optional<Seg3dConfig>` when refactored to structs.

perception/autoware_ptv3/include/autoware/ptv3/ptv3_node.hpp
```diff
@@ -60,10 +65,18 @@ class PTV3_PUBLIC PTv3Node : public rclcpp::Node
...
+    detected_objects_pub_;
+
   std::unique_ptr<PTv3TRT> model_ptr_{nullptr};
 
   rclcpp::TimerBase::SharedPtr timer_{nullptr};
 
+  perception_utils::IouBevNms iou_bev_nms_;
+  perception_utils::DetectionClassRemapper detection_class_remapper_;
+  std::vector<std::string> detection_class_names_;
+  bool has_twist_{false};
```
mojomex, line 78: Best have separate structs that group per-head members.

perception/autoware_ptv3/include/autoware/ptv3/ptv3_trt.hpp
```diff
@@ -140,13 +153,25 @@ class PTV3_PUBLIC PTv3TRT
...
+  // Detection3D head outputs
+  CudaUniquePtr<float[]> dense_heatmap_d_{nullptr};
+  CudaUniquePtr<float[]> query_heatmap_score_d_{nullptr};
+  CudaUniquePtr<std::int64_t[]> query_labels_d_{nullptr};
+  CudaUniquePtr<float[]> heatmap_d_{nullptr};
+  CudaUniquePtr<float[]> center_d_{nullptr};
+  CudaUniquePtr<float[]> height_d_{nullptr};
+  CudaUniquePtr<float[]> dim_d_{nullptr};
+  CudaUniquePtr<float[]> rot_d_{nullptr};
+  CudaUniquePtr<float[]> vel_d_{nullptr};
```
mojomex, line 174: This class now has 52 (!) members - the split proposed further up should fix this.

perception/autoware_ptv3/include/autoware/ptv3/ros_utils.hpp
```diff
@@ -41,6 +48,26 @@ CHECK_FIELD(CloudPointType, autoware::point_types::PointXYZIRC, z);
...
+  const Box3D & box3d, const std::vector<std::string> & class_names, bool has_twist,
+  autoware_perception_msgs::msg::DetectedObject & obj);
+
+/**
+ * @brief Map a detection class name to an ObjectClassification label.
+ *
+ * @param class_name Detection class name.
+ * @return Matching ObjectClassification label, or UNKNOWN.
+ */
+std::uint8_t get_semantic_type(const std::string & class_name);
```
mojomex: (nit) The term `semantic` might be confused by people with the `semantic` in semantic segmentation.

perception/autoware_ptv3/src/ptv3_node.cpp
```diff
@@ -91,15 +93,104 @@ PTv3Node::PTv3Node(const rclcpp::NodeOptions & options) : Node("ptv3", options)
       seg3d_head_onnx_path, trt_precision, seg3d_head_engine_path, 1ULL << 30U);
   }
 
+  // Detection head parameters
```
mojomex, line 96: Split into three (four?) functions: one each for backbone, semseg, and 3d det, and maybe for common stuff.

mojomex (commented):
I haven't looked at all the code in detail yet, so I'll do one more round of code review after this one.

Multiple files have grown quite large, and can easily be split into more digestable ones, so most of my feedback is about that.

Functions/members introduced all seem reasonable to me 👍

perception/autoware_ptv3/config/ml_package_ptv3_det3d_head.param.yaml
amadeuszsz: It goes down to the grid size. Wrong configuration will throw exception with clean explanation about invalid setting

perception/autoware_ptv3/include/autoware/ptv3/postprocess/detection3d_postprocess.hpp
amadeuszsz: Let's address "optimization" PR 🙏🏻

perception/autoware_ptv3/lib/ros_utils.cpp
amadeuszsz: We already have [it](https://github.com/autowarefoundation/autoware_core/blob/47d6cf23e40ed363b9a62e8aedd9e3c31cf26135/common/autoware_object_recognition_utils/include/autoware/object_recognition_utils/object_classification.hpp#L94-L119). Imho common utils function shouldn't map class "X" (traffic cone / barrier) to class "Y" (hazard) as we do here. If we add "Z" (debris) to detection3d, we would need to update upstream utils function to map it to "Y" (hazard).

@YoshiRi can we relax [this util function](https://github.com/autowarefoundation/autoware_core/blob/47d6cf23e40ed363b9a62e8aedd9e3c31cf26135/common/autoware_object_recognition_utils/include/autoware/object_recognition_utils/object_classification.hpp#L117) and instead of throwing exception just map custom labels as "UNKNOWN"?

perception/autoware_ptv3/lib/ptv3_trt.cpp
amadeuszsz: I'm getting confused about expectations, because fixing all the issues in this package will result with 5k lines changes PR. 
Not only this file is a problem as [pointed before](https://github.com/autowarefoundation/autoware_universe/pull/12655#discussion_r3378876844). 
We need to introduce new files, change existing file names, update docs, update docstrings, and probably much more, I agree. 
Please let me address a separate PR where we will address pure package restructure without any functional change.
My goal was to provide simplest PR (minimal change) and address a separate PR where we can address code design concerns 🙏🏻

perception/autoware_ptv3/include/autoware/ptv3/ptv3_config.hpp
amadeuszsz: To address in separate PR

perception/autoware_ptv3/include/autoware/ptv3/ptv3_config.hpp
amadeuszsz: To address in separate PR

perception/autoware_ptv3/include/autoware/ptv3/ptv3_config.hpp
amadeuszsz: To address in separate PR

perception/autoware_ptv3/include/autoware/ptv3/ptv3_node.hpp
amadeuszsz: To address in separate PR

perception/autoware_ptv3/include/autoware/ptv3/ptv3_trt.hpp
amadeuszsz: To address in separate PR

perception/autoware_ptv3/src/ptv3_node.cpp
amadeuszsz: To address in separate PR

perception/autoware_ptv3/schema/ml_package_ptv3.schema.json
amadeuszsz: Thanks, part of https://github.com/autowarefoundation/autoware_universe/pull/12758/commits/66f82e642b8b9472b08fff6f4ea6ca2822e30dbc

perception/autoware_ptv3/schema/ml_package_ptv3.schema.json
amadeuszsz: Thanks, part of https://github.com/autowarefoundation/autoware_universe/pull/12758/commits/66f82e642b8b9472b08fff6f4ea6ca2822e30dbc

perception/autoware_ptv3/schema/ml_package_ptv3.schema.json
amadeuszsz: Thanks, part of https://github.com/autowarefoundation/autoware_universe/pull/12758/commits/66f82e642b8b9472b08fff6f4ea6ca2822e30dbc

perception/autoware_ptv3/include/autoware/ptv3/ros_utils.hpp
amadeuszsz: Addressed in https://github.com/autowarefoundation/autoware_universe/pull/12758/commits/80973fd2029d68b9387910c56a212d7e52336dda

perception/autoware_ptv3/config/ml_package_ptv3_det3d_head.param.yaml
amadeuszsz: It comes from CenterPoint convention, which later was adopted to TransFusion in training config. To be honest, we could have only `bev_cell_size: [0.96, 0.96]` and that's what we need.
Since artifacts are already uploaded, maybe let's just keep `bbox_voxel_size: [0.96, 0.96, 8.0]` and thus keep the compatibility.
https://github.com/autowarefoundation/autoware_universe/pull/12758/commits/08fe07ee213b5ffc63b7cf7191d20a4c97e7afa1

perception/autoware_ptv3/lib/ros_utils.cpp
mojomex: Hm, yeah makes sense. No matter how it's done, models will always be hard to keep in sync. Maybe best left as-is.

perception/autoware_ptv3/lib/ptv3_trt.cpp
mojomex: Fair point. For me as a reviewer, such a huge node makes it almost impossible to spot bugs (e.g. something gets used before being set etc.), so that's why I'm kind of pedantic about splitting up.

But you're right to point out that I expressed these expectations in this PR, while not doing so for the last PR which had similar issues. I think it was just under my personal threshold.

Since this PR will not be used in production from day one, having a refactor PR straight after is acceptable to me this time.

Still, I think that doing that refactoring PR _before_ this one would be ideal, but I realize that stuff has to keep moving

mojomex: * Tracking in #12806

perception/autoware_ptv3/include/autoware/ptv3/postprocess/detection3d_postprocess.hpp
mojomex: * Tracking in #12807

perception/autoware_ptv3/include/autoware/ptv3/ptv3_config.hpp
```diff
@@ -93,21 +103,96 @@ class PTv3Config
...
-      });
+    if (use_det3d_head_) {
+      if (detection_class_names.empty()) {
+        throw std::runtime_error("detection_class_names must not be empty when use_det3d_head.");
+      }
+      if (bbox_voxel_size.size() != 3) {
+        throw std::runtime_error("bbox_voxel_size must contain 3 elements.");
+      }
+      if (bbox_voxel_size[0] <= 0.0F || bbox_voxel_size[1] <= 0.0F || bbox_voxel_size[2] <= 0.0F) {
+        throw std::runtime_error("bbox_voxel_size values must be positive.");
```
mojomex, line 131: Should we also see if `bbox_voxel_size % voxel_size == 0`? (it's floats but the idea is the same).
Non-aligning numbers (e.g. `0.12` in backbone and `0.42` in head) would cause aliasing, right? (or not work at all)

perception/autoware_ptv3/lib/ptv3_trt.cpp
```diff
@@ -92,6 +99,10 @@ void PTv3TRT::setPublishFilteredPointcloud(
 
 void PTv3TRT::allocateMessages()
```
mojomex: Probably best to rename to `allocateSegOutputMessages` or similar to make clear this is only for the seg head.

perception/autoware_ptv3/lib/ptv3_trt.cpp
```diff
@@ -598,7 +744,7 @@ bool PTv3TRT::preProcess(const std::shared_ptr<const cuda_blackboard::CudaPointC
   allocateMessages();
 
   const auto num_points = msg_ptr->height * msg_ptr->width;
-  if (config_.source_reconstruction_ == SourceReconstruction::FULL) {
+  if (should_run_seg3d && config_.source_reconstruction_ == SourceReconstruction::FULL) {
```
mojomex, line 747: In line 753, it should be
```
- "Empty pointcloud. Skipping segmentation."
+ "Empty pointcloud. Skipping inference."
```

mojomex: (Github doesn't let me comment on that line)

perception/autoware_ptv3/lib/postprocess/detection3d_postprocess.cu
```diff
@@ -0,0 +1,197 @@
...
+#include <cstdint>
+
+namespace autoware::ptv3
+{
+
+namespace
+{
+constexpr int k_block_size = 256;
+
+struct IsScoreKeep
```
mojomex: Name is a bit weird, `IsScoreNonZero` would be clearer.

perception/autoware_ptv3/lib/postprocess/detection3d_postprocess.cu
```diff
@@ -0,0 +1,197 @@
...
+{
+  __device__ bool operator()(const Box3D & box) const { return box.score > 0.0f; }
+};
+
+struct ScoreGreater
```
mojomex: Since it is a predicate, `IsScoreGreaterThan` would be a clearer name (the `Is` signifies that it's a predicate, the `Than` that it's a binary op)

perception/autoware_ptv3/lib/postprocess/detection3d_postprocess.cu
```diff
@@ -0,0 +1,197 @@
...
+    post_center_range_d_.get(), score_thresholds_d_.get(), dist_bin_limits_d_.get(),
+    static_cast<int>(config_.distance_bin_upper_limits_.size()),
+    thrust::raw_pointer_cast(yaw_norm_thresholds_d_.data()), config_.has_twist_,
+    thrust::raw_pointer_cast(raw_boxes_d_.data()), stream);
+
+  const auto policy = thrust::cuda::par.on(stream);
+
+  const auto num_passing =
+    thrust::count_if(policy, raw_boxes_d_.begin(), raw_boxes_d_.end(), IsScoreKeep{});
+  if (num_passing == 0) {
```
mojomex: `count_if` is redundant here, since `copy_if` checks the same predicate, and returns an iterator which can be used to calculate `num_passing`. Saves 1 kernel call.

perception/autoware_ptv3/lib/postprocess/detection3d_postprocess.cu
```diff
@@ -0,0 +1,197 @@
...
+
+  const int label = static_cast<int>(query_labels[i]);
+  if (label < 0 || label >= num_classes) {
+    return;
+  }
+
+  const int class_offset = label * num_proposals;
+  const float score =
+    (1.0f / (1.0f + expf(-heatmap[class_offset + i]))) * query_heatmap_score[class_offset + i];
+
```
mojomex: Maybe more understandable

```suggestion
  auto sigmoid = [](float logit){ return 1.0f / (1.0f + expf(-logit)); };
  const float score = sigmoid(heatmap[class_offset + i]) * query_heatmap_score[class_offset + i];
```

perception/autoware_ptv3/lib/postprocess/detection3d_postprocess.cu
```diff
@@ -0,0 +1,197 @@
...
+  const float x = center[i] * bbox_voxel_x_size + min_x_range;
+  const float y = center[num_proposals + i] * bbox_voxel_y_size + min_y_range;
+  const float length = expf(dim[i]);
+  const float width = expf(dim[num_proposals + i]);
+  const float box_h = expf(dim[2 * num_proposals + i]);
+  const float z = height[i] - box_h * 0.5f;
```
mojomex: length/width/box_h can be computed further below, after `if (yaw_norm < ...)`. It's faster and imo clearer.

mojomex (commented):
Code is already mostly there! I left some minor comments, please have a look.

perception/autoware_ptv3/include/autoware/ptv3/ptv3_config.hpp
amadeuszsz: Yes!
https://github.com/autowarefoundation/autoware_universe/pull/12758/commits/4ac98192bf82acf8b53bd4282af74588eb608e59

perception/autoware_ptv3/lib/ptv3_trt.cpp
amadeuszsz: Out of PR scope.
To be addressed in refactor PR

perception/autoware_ptv3/lib/ptv3_trt.cpp
amadeuszsz: https://github.com/autowarefoundation/autoware_universe/pull/12758/commits/5e3848d72aa5ab2e4221cf5625e1fbf6823991d9

perception/autoware_ptv3/lib/postprocess/detection3d_postprocess.cu
amadeuszsz: https://github.com/autowarefoundation/autoware_universe/pull/12758/commits/36e6354df26ec68bb21c06b695f8453b4e781668

perception/autoware_ptv3/lib/postprocess/detection3d_postprocess.cu
amadeuszsz: https://github.com/autowarefoundation/autoware_universe/pull/12758/commits/35542706d600db1977a3ef034d846af7c8f8a134

perception/autoware_ptv3/lib/postprocess/detection3d_postprocess.cu
amadeuszsz: Nice catch, https://github.com/autowarefoundation/autoware_universe/pull/12758/commits/2917aa2cbc80a66fe631d295cd02f8e1086d18a6

perception/autoware_ptv3/lib/postprocess/detection3d_postprocess.cu
amadeuszsz: https://github.com/autowarefoundation/autoware_universe/pull/12758/commits/2694f48fa3ade7ee8c1980280a9ef066a2c2a68a

perception/autoware_ptv3/lib/postprocess/detection3d_postprocess.cu
amadeuszsz: Not for `box_h`
https://github.com/autowarefoundation/autoware_universe/pull/12758/commits/17d231204881fdd0f343ec4af58170eea6e95262

mojomex (approved):
LGTM! Tested again with latest ref (17d231204):

https://github.com/user-attachments/assets/e50eb418-6171-4310-89f9-9c4aa05056a4

perception/autoware_ptv3/lib/ptv3_trt.cpp
amadeuszsz: Couldn't look on this sad emoji!
After considerations I think this is the scope.
https://github.com/autowarefoundation/autoware_universe/pull/12758/commits/3192e40ec2be5b772b45b444f097917a970b3680
Plz undo emoji

amadeuszsz:
@mojomex 

Thanks for detailed review, I really appreciate it!

FYI, currently deployed model has some issues with truck class, more specifically cab + trailer. If you see detection (`truck`) flickers between cab and trailer, it is not the NMS bug, but the model problem. I'm currently fixing it and it will be addressed in next model version.

mojomex:
@amadeuszsz Thank you for sticking with me! I think we found good middle ground on this one.

PR #12774, opened 2026-06-12
test(tensorrt_yolox): add integration test

interimadd:
## Description

Add a node-level characterization (integration) test for `autoware_tensorrt_yolox`'s `TrtYoloXNode`, as a safety net before refactoring the YOLOX post-processing logic.

The test drives the node through ROS topics: it publishes a camera image on `~/in/image` and verifies the detections on `~/out/objects` (and, for the multitask model, the segmentation/color masks on `~/out/mask` and `~/out/color_mask`). Two model configurations are covered:

- **Traffic-light detector** (`yolox_s_car_ped_tl_detector`): detection only.
- **General detector + segmentation** (`yolox-sPlus-opt ... seg16cls`): a multitask model, so the node additionally publishes the masks.

This follows the **same approach as the existing test in `sensing/autoware_calibration_status_classifier/test`**: it uses the ONNX model downloaded under `autoware_data`, keeps the test input image next to the test code, and self-skips (`GTEST_SKIP`) when the GPU or model is unavailable. It is additionally gated in CMake behind `TRT_AVAIL AND CUDA_AVAIL`, so it is a no-op on CI machines without a GPU.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

`colcon build` + `colcon test` on a machine with an NVIDIA GPU and TensorRT 10.8; all three cases pass (`tensorrt_yolox_node.cpp`: 85.0% line / 100% function coverage).

Passing the bundled `test/test_image.jpg` through the models produces detections like the following (boxes drawn from the published `~/out/objects`):

- `yolox_viz_tl.png` — traffic-light detector: vehicle / pedestrian traffic lights.
 
yolox_viz_tl
 
- `yolox_viz_seg.png` — multitask model: cars / trucks / pedestrians (plus the segmentation mask).
yolox_viz_seg

## Notes for reviewers

- The PR adds one test file, one test image, and a CMake entry; **no production code is touched.**
- The test input image is committed as a downscaled JPG (~352 KB).

## Interface changes

None.

## Effects on system behavior

None.


ktro2828 (approved):
LGTM!

PR #2370, opened 2022-11-24
feat: add tensorrt_yolox package

manato:
## Description

This PR aims to add the `tensorrt_yolox` package into the series of perception modules in autoware.universe.
The `tensorrt_yolox` package that works with autoware has been published in [separated repository](https://github.com/tier4/tensorrt_yolox). This PR integrates the package's contents into autoware.universe.

## Related links

- [TIER IV INTERNAL LINK](https://tier4.atlassian.net/browse/T4PB-17617)
- https://github.com/tier4/tensorrt_yolox

## Tests performed

```
# build
colcon build --symlink-install --cmake-args -DCMAKE_EXPORT_COMPILE_COMMANDS=ON -DCMAKE_BUILD_TYPE=Release --packages-up-to tensorrt_yolox

# run
ros2 run tensorrt_yolox tensorrt_yolox_node_exe --ros-args -p model_path:=install/tensorrt_yolox/share/tensorrt_yolox/data/yolox-tiny.onnx -p label_path:=install/tensorrt_yolox/share/tensorrt_yolox/data/label.txt -r /tensorrt_yolox/in/image:=/image_raw
```

## Notes for reviewers

I've added small modifications  listed below from [the original repository](https://github.com/tier4/tensorrt_yolox)
- treating labels as case insensitive
- treating the `PERSON`  label in the same way as the `PEDESTRIAN` label for COCO dataset compatibility 

## Pre-review checklist for the PR author

The PR author **must** check the checkboxes below when creating the PR.

- [ ] I've confirmed the [contribution guidelines].
- [ ] The PR follows the [pull request guidelines].

## In-review checklist for the PR reviewers

The PR reviewers **must** check the checkboxes below before approval.

- [ ] The PR follows the [pull request guidelines].
- [ ] The PR has been properly tested.
- [ ] The PR has been reviewed by the code owners.

## Post-review checklist for the PR author

The PR author **must** check the checkboxes below before merging.

- [ ] There are no open discussions or they are tracked via tickets.
- [ ] The PR is ready for merge.

After all checkboxes are checked, anyone who has write access can merge the PR.

[contribution guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/
[pull request guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/pull-request-guidelines/

PR #12775, opened 2026-06-12
fix(autoware_tensorrt_yolox): correct std::clamp argument order

interimadd:
## Description

Fixes an argument-order bug in the `std::clamp` calls that clamp detected bounding-box top-left offsets to the image bounds in `autoware_tensorrt_yolox`.

`std::clamp` has the signature `std::clamp(value, low, high)`, but the code passed the arguments as `std::clamp(0, x1, cols)`.

This bug has existed since the package was first added (commit `03de605b09`, "feat: add tensorrt_yolox package (#2370)", 2022-11-30).

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

- `colcon build --packages-select autoware_tensorrt_yolox` — succeeds (only a pre-existing ament_auto header-install-dir deprecation notice on stderr).
- `colcon test --packages-select autoware_tensorrt_yolox --event-handlers console_cohesion+` — 6/6 tests pass (2 gtest + copyright/cppcheck/lint_cmake/xmllint linters), 0 failed.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

No change for normal in-image detections. For detections whose top-left corner falls outside the image (right/bottom edge), the published ROI top-left offset is now correctly clamped to the image dimensions instead of being passed through un-clamped, and the previous undefined-behavior path is eliminated.


ktro2828 (approved):
Thank you for catching! LGTM.

manato (approved):
@interimadd 
LGTM! Thank you for your contribution!

PR #12777, opened 2026-06-12
refactor(traffic_light_multi_camera_fusion): restore fuse() return-based output

interimadd:
## Description

This PR restores the original input/output of `MultiCameraFusion::fuse()` that was changed in #12710, while **keeping** the `autoware_agnocast_wrapper` adoption on the node side.

Returning the result by value reads better than the output-parameter form.

Memory copy is negligible because `TrafficLightGroupArray` is a small message (well under 1 KB for a typical intersection) published at ~10 Hz, so the copy is far smaller than the fusion computation itself.

## Related links

**Parent Issue:**

- https://github.com/autowarefoundation/autoware/issues/7007 (agnocast wrapper adoption)
- #12710 (the PR whose `fuse()` interface change this reverts)

## How was this PR tested?

```bash
PACKAGE_NAME=autoware_traffic_light_multi_camera_fusion
colcon build --packages-select $PACKAGE_NAME
colcon test --packages-select $PACKAGE_NAME --event-handlers console_cohesion+
```

- Build succeeds.
- All tests pass (6/6: 2 gtest + 4 linters).

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_traffic_light_multi_camera_fusion/src/multi_camera_fusion.cpp
+++ b/perception/autoware_traffic_light_multi_camera_fusion/src/multi_camera_fusion.cpp
@@ -177,8 +177,7 @@ MultiCameraFusion::MultiCameraFusion(const MultiCameraFusionConfig & config)
 }
 
 MultiCameraFusionResult MultiCameraFusion::fuse(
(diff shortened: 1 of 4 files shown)


a-maumau (approved):
LGTM

PR #12779, opened 2026-06-12
feat(planning): apply autoware_agnocast_wrapper to diffussion planner trajectory pipeline nodes for CIE

atsushi421:
## Description

Apply `CallbackIsolatedAgnocastExecutor` (CIE) to the trajectory pipeline node packages, as part of the CIE rollout. By going through `autoware_agnocast_wrapper`, CIE activation is controlled by the `ENABLE_AGNOCAST` environment variable. When unset or `0`, each node runs with the standard ROS 2 executor as before; when set to `1`, it switches to `CallbackIsolatedAgnocastExecutor`.

Target packages (6):

- `autoware_diffusion_planner`
- `autoware_trajectory_modifier`
- `autoware_trajectory_optimizer`
- `autoware_trajectory_adapter`
- `autoware_trajectory_ranker` (both `TrajectoryRanker` and `SimpleTrajectoryRanker`)
- `autoware_trajectory_selector`

### Changes

- `CMakeLists.txt`: replace `rclcpp_components_register_node(...)` with `autoware_agnocast_wrapper_register_node(...)`. Each generated executable uses `CallbackIsolatedAgnocastExecutor` when built with `ENABLE_AGNOCAST=1`, and falls back to a `SingleThreadedExecutor` standalone executable (identical to the previous `rclcpp_components` behavior) otherwise.
- `package.xml`: add the `autoware_agnocast_wrapper` dependency.

No source changes.

## Related links

**Parent Issue:**

- https://github.com/autowarefoundation/autoware_core/issues/968

## How was this PR tested?

Verified in a logging simulator environment based on pilot-auto.x2 (beta/v4.4) with `ENABLE_AGNOCAST=1`:

- All converted nodes were confirmed to be CIE-ized via the `agnocast_cie_thread_configurator` prerun template (their callback groups are announced and captured).
- End-to-end rosbag run: the whole trajectory pipeline (diffusion_planner to trajectory_adapter) publishes at the nominal 10 Hz with no errors.

## Notes for reviewers

If a node meets **all** of the following conditions, please check for potential race conditions (see [Effects on system behavior](#effects-on-system-behavior) for the reason):

1. The node was originally running on a `SingleThreadedExecutor`
2. The node has multiple callback groups (the default is one)
3. Shared variables are accessed across callback groups without proper mutex protection

> **Note:** Nodes already running on `MultiThreadedExecutor` (or `component_container_mt`) are not affected, as their callbacks are already executed concurrently.

## Interface changes

None.

## Effects on system behavior

When `ENABLE_AGNOCAST=0` (default), there is no change in behavior at all. Each node runs with the standard ROS 2 executor exactly as before.

When `ENABLE_AGNOCAST=1`, the executor switches to `CallbackIsolatedAgnocastExecutor` (CIE). Internally, CIE creates a dedicated `rclcpp::SingleThreadedExecutor` for each callback group, so the fundamental scheduling behavior within each group remains almost identical. The key difference is that, for nodes originally using `SingleThreadedExecutor`, callback groups now run in parallel across separate threads, similar to `MultiThreadedExecutor`.


SakodaShintaro (approved):
I have confirmed that the planning simulator with `autoware_diffusion_planner` works well.

## Setup

```bash
sudo add-apt-repository ppa:t4-system-software/agnocast
sudo apt update
sudo apt install agnocast-heaphook-v2.3.5 agnocast-kmod-v2.3.5

sudo rmmod agnocast
sudo modprobe agnocast

lsmod | grep agnocast
dpkg -l | grep agnocast
ls -la /opt/ros/humble/lib/libagnocast_heaphook.so
```

## build

```bash
export ENABLE_AGNOCAST=1
MAKEFLAGS="-j1" colcon build \
  --symlink-install \
  --cmake-args -DCMAKE_BUILD_TYPE=Release -DBUILD_TESTING=OFF \
  --continue-on-error
```

## run

```bash
export ENABLE_AGNOCAST=1
export AGNOCAST_MEMPOOL_SIZE=134217728
ros2 launch autoware_launch planning_simulator.launch.xml \
    map_path:=$MAP_PATH \
    planning_setting:=diffusion_planner \
    vehicle_model:=sample_vehicle \
    sensor_model:=sample_sensor_kit
```

go-sakayori (approved):
Code LGTM! @SakodaShintaro checked the planning simulator so will approve

danielsanchezaran (approved):
LGTM

PR #12780, opened 2026-06-14
style(processing_time_checker): adjust comment indentation

paulsohn:
## Description

Adjust comment indentation to the same level of list entries.

## Related links

This PR is to fix sync-param workflow in autoware_launch downstream. See https://github.com/autowarefoundation/autoware_launch/pull/1882 .

The actual used config file is in `processing_time_checker.param.yaml` placed in `autoware_launch`, watches package config file and syncs it when the field is not marked as `# {OVERRIDE}` marker.

To implement this auto-sync behavior, ruamel behavior and comment styles are crucial.

## How was this PR tested?

In this branch I confirmed that applying this fixes sync-param behavior. Nothing to test for the parameter file itself since this is only a comment style change.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


kosuke55 (approved):

sasakisasaki (approved):
For feature improvement by `sync-params` workflow, it seems this PR is needed for indent-tidy `sync-params`. Currently this PR is required for [this PR](https://github.com/autowarefoundation/autoware_launch/pull/1882).

As `sync-params` workflow is already under operation and it is working as expected, I agree with merging this PR 👍

PR #12781, opened 2026-06-15
feat(autoware_mppi_optimizer): initial implementation of MPPI for diffusion planner

PanConChicharron:
## Description
https://tier4.atlassian.net/wiki/spaces/CB/pages/5206966634/MPPI+for+Static+and+Dynamic+Collision-Avoidance

Adds the Model Predictive Path Integral Control algorithm to autoware as a refinement step after the Diffusion Planner.

https://github.com/user-attachments/assets/8528387c-37ba-4c87-ba5b-a468276b1f79

## Related links

**Parent Issue:**

- [Link](https://tier4.atlassian.net/browse/PDDP-30)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/autoware_mppi_optimizer/CMakeLists.txt
+++ b/planning/autoware_mppi_optimizer/CMakeLists.txt
@@ -1,0 +1,72 @@
+cmake_minimum_required(VERSION 3.14)
+project(autoware_mppi_optimizer LANGUAGES C CXX CUDA)
+
+set(CMAKE_CUDA_ARCHITECTURES OFF)
+
+find_package(autoware_cmake REQUIRED)
+find_package(ament_cmake_auto REQUIRED)
+find_package(MPPI-Generic REQUIRED)
+find_package(Eigen3 REQUIRED)
+find_package(autoware_perception_msgs REQUIRED)
+find_package(autoware_planning_msgs REQUIRED)
+find_package(nav_msgs REQUIRED)
+find_package(rclcpp REQUIRED)
+find_package(tf2_geometry_msgs REQUIRED)
+
+add_library(mppi_extensions INTERFACE)
+add_library(MppiExtensions::extensions ALIAS mppi_extensions)
+target_include_directories(mppi_extensions BEFORE INTERFACE
+  "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>")
+target_link_libraries(mppi_extensions INTERFACE MPPI::MPPI)
+
+add_library(first_order_dubins_mppi SHARED
+  src/first_order_dubins/first_order_dubins_mppi_interface.cu
+)
+target_include_directories(first_order_dubins_mppi PUBLIC
+  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+  $<INSTALL_INTERFACE:include>
+)
+target_link_libraries(first_order_dubins_mppi PUBLIC
+  mppi_extensions
+  rclcpp::rclcpp
+  tf2_geometry_msgs::tf2_geometry_msgs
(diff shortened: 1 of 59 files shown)


shmpwk (commented):
Could we adopt the similar ROS/CUDA domain separation described in [`autoware_tensorrt_vad` design doc](https://github.com/tier4/autoware_universe/blob/0c463a174a6e979ccd6633974cde2a209241962c/planning/autoware_tensorrt_vad/docs/design.md#separation-between-ros-and-cuda-domains)?

Currently `first_order_dubins_mppi` (CUDA target) links against `rclcpp`, `nav_msgs`, and Autoware message packages, and `FirstOrderDubinsMppiInterface::optimizeTrajectory()` takes ROS message types directly (`Trajectory`, `Odometry`, `TrackedObjects`). This couples the CUDA domain to ROS topic/interface changes.

Suggested layering (mirroring VAD):
- **ROS domain** (`MppiOptimizer`): topic I/O and parameter loading only
- **Interface** (`MppiInterface`): convert ROS messages ↔ plain `MppiInputData` / `MppiOutputData`; move `tracked_objects_obstacles.hpp` here
- **CUDA domain** (`first_order_dubins_mppi_cuda`): MPPI inference only, depends on `MPPI-Generic` — no ROS headers or libraries

The existing plain structs (`FirstOrderDubinsMppiState`, `CostParams`, `VehicleParams`) are a good start. The main gaps are (1) removing ROS types from the CUDA library's public API and link dependencies, and (2) splitting debug/output structs so the CUDA side returns plain data and the interface converts to `Trajectory` / markers.

This would also align naturally with using `cuda_add_library` (as in `autoware_tensorrt_vad`) instead of native `LANGUAGES CUDA`.

planning/autoware_mppi_optimizer/CMakeLists.txt
```diff
@@ -0,0 +1,72 @@
...
+  rclcpp::rclcpp
+  tf2_geometry_msgs::tf2_geometry_msgs
+  autoware_perception_msgs::autoware_perception_msgs__rosidl_generator_cpp
+  autoware_planning_msgs::autoware_planning_msgs__rosidl_generator_cpp
+  nav_msgs::nav_msgs__rosidl_generator_cpp
+)
+set_target_properties(first_order_dubins_mppi PROPERTIES
+  POSITION_INDEPENDENT_CODE ON
+  CXX_STANDARD 17
+  CUDA_STANDARD 17
```
shmpwk, line 40: Do we actually need an explicit `CUDA_STANDARD 17` here? Other Autoware CUDA packages like [autoware_tensorrt_vad](https://github.com/autowarefoundation/autoware_universe/blob/e18a5311e509574326ae0ff6256108326a8132d0/e2e/autoware_tensorrt_vad/CMakeLists.txt)  don't set it — they use the legacy `find_package(CUDA)` + `cuda_add_library()` pattern instead of native `LANGUAGES CUDA` + `add_library(.cu)`.

With the native CUDA approach, `autoware_package()`'s `CMAKE_CXX_STANDARD 17` does **not** propagate to `.cu` files, which is why an explicit `CUDA_STANDARD` was added. If we align with the `autoware_tensorrt_vad` pattern (`cuda_add_library` + `CUDA_NVCC_FLAGS` for arch), we can drop the per-target `CUDA_STANDARD` / `CXX_STANDARD` settings and call `autoware_package()` before defining targets.

One caveat: MPPI headers may rely on C++17 features more heavily than the VAD kernels. If we switch to `cuda_add_library` and hit NVCC errors, we may still need `--std=c++17` in `CUDA_NVCC_FLAGS`. Also worth confirming that `MPPI-Generic` doesn't require native CMake CUDA language support.

PR #12785, opened 2026-06-15
feat(traffic_light_multi_camera_fusion): apply `agnocast_wrapper::Node` to `traffic_light_multi_camera_fusion`

Koichi98:
## Description

Apply `agnocast_wrapper::Node` to `autoware_traffic_light_multi_camera_fusion`.

Main changes:

- `MultiCameraFusionNode` now derives from `agnocast_wrapper::Node`; the map
  subscription uses `AUTOWARE_SUBSCRIPTION_PTR`, and the synchronizer callback
  takes Agnocast message pointers.
- The `message_filters` alias is switched from `message_filters` to
  `autoware::agnocast_wrapper::message_filters`, so the per-camera
  `ExactTime` / `ApproximateTime` synchronizers (`CameraInfo`, `TrafficLightRoiArray`,
  `TrafficLightGroupArray`) run through the Agnocast-aware, variadic
  `message_filters` wrapper.
- The diagnostics helper uses the node-templated `BasicDiagnosticsInterface<...>`.
- `AGNOCAST_EXECUTOR` is set to `AgnocastOnlyCallbackIsolatedExecutor`, which is
  required for Method-2 (`agnocast_wrapper::Node`) nodes.
- The node is launched as a standalone node (not inside a component container).

## How was this PR tested?

- **Build**: Confirmed the node builds successfully with both `ENABLE_AGNOCAST=0`
  and `ENABLE_AGNOCAST=1`.
- **Unit tests**: Confirmed the package tests pass.
- **Runtime (synthetic injection)**: On an autoware-based product, confirmed by
  launching the node standalone under `ENABLE_AGNOCAST=1` and injecting synthetic
  inputs for one camera (`camera_info`, `rois`, and `traffic_signals`) with a
  shared timestamp at 10 Hz. The 3-input `ExactTime` synchronizer callback fires
  and the fused traffic signals are published at ~10 Hz.

## Notes for reviewers

- **message_filters wrapper**: The synchronizers now use the variadic
  `autoware::agnocast_wrapper::message_filters` (three-input `ExactTime` /
  `ApproximateTime`). This is available in the Agnocast wrapper and is
  source-compatible for this node.
- **Publish pattern**: The fused signals are published with `publish(const &)`, so
  the message stays in the loaned shared-memory buffer for correct serialization by
  the Agnocast→ROS 2 bridge.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_traffic_light_multi_camera_fusion/CMakeLists.txt
+++ b/perception/autoware_traffic_light_multi_camera_fusion/CMakeLists.txt
@@ -18,7 +18,7 @@ ament_auto_add_library(${PROJECT_NAME} SHARED
 autoware_agnocast_wrapper_register_node(${PROJECT_NAME}
   PLUGIN "autoware::traffic_light::MultiCameraFusionNode"
   EXECUTABLE traffic_light_multi_camera_fusion_node
-  AGNOCAST_EXECUTOR CallbackIsolatedAgnocastExecutor
+  AGNOCAST_EXECUTOR AgnocastOnlyCallbackIsolatedExecutor
   ROS2_EXECUTOR SingleThreadedExecutor
 )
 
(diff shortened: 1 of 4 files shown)


a-maumau (approved):
LGTM

PR #12793, opened 2026-06-16
refactor(tensorrt_yolox): remove hard-coded label mapping

ktro2828:
## Description

This pull request refactors how segmentation mask overlay labels are managed and configured in the `TrtYoloXNode` class. Instead of using hardcoded boolean fields for each label, the code now uses a flexible map-based approach, allowing dynamic configuration of which labels should have overlays via parameters. The most important changes are:

### Refactoring label overlay configuration

* Replaced individual boolean fields for each overlay label in `RoiOverlaySemsegLabel` with an internal `std::unordered_map<uint8_t, bool>`, and added methods to set, clear, and check overlay status for arbitrary label IDs. This change makes the overlay configuration extensible and less error-prone.
* Updated the constructor of `TrtYoloXNode` to remove hardcoded parameter declarations for each label, delegating overlay label setup to a new method.

### Dynamic parameter handling

* Added the `loadOverlaySegmentationLabels()` method, which dynamically loads all overlay label parameters with the prefix `roi_overlay_segmentation_label.`, validates them, and updates the overlay map accordingly. This allows for easier addition or removal of labels through configuration. [[1]](diffhunk://#diff-a46e293f3ec1aa2ddd8253653b9d7cd76cbc1d505d86400fc6de5dcc11ccc21bR96) [[2]](diffhunk://#diff-4519a25e256b218c2807d6f99f5e4b69a3680da45d86c19695909a3897a36526R93) [[3]](diffhunk://#diff-4519a25e256b218c2807d6f99f5e4b69a3680da45d86c19695909a3897a36526R130-R155)

### Code maintenance

* Added necessary includes for `std::unordered_map` and `std::ostream` to support these changes. [[1]](diffhunk://#diff-a46e293f3ec1aa2ddd8253653b9d7cd76cbc1d505d86400fc6de5dcc11ccc21bR45) [[2]](diffhunk://#diff-4519a25e256b218c2807d6f99f5e4b69a3680da45d86c19695909a3897a36526R25)

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_tensorrt_yolox/include/autoware/tensorrt_yolox/tensorrt_yolox_node.hpp
+++ b/perception/autoware_tensorrt_yolox/include/autoware/tensorrt_yolox/tensorrt_yolox_node.hpp
@@ -42,6 +42,7 @@
 #include <memory>
 #include <optional>
 #include <string>
+#include <unordered_map>
 #include <utility>
 #include <vector>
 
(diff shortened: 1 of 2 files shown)


ktro2828:
@a-maumau @badai-nguyen As of this PR, we only have to modify YAML file to change label mappings. Could you review this PR?

a-maumau (commented):
I agree with copilot.
the label mappings are different between Semantic.msg and autoware::object_recognition_utils::toLabel (ObjectClassification.msg).
(This itself is quite confusing...)

We might can add the toLabel for the semantic label in some utils or in the object_recognition_utils

PR #12794, opened 2026-06-16
refactor(autoware_tensorrt_yolox): extract detector

interimadd:
## Description

This PR extracts the per-frame detection pipeline out of `TrtYoloXNode` into a new `rclcpp::Node`-independent `TrtYoloXDetector` class, leaving the node responsible only for ROS I/O. There is **no behavior change**. It is split into three small commits so each diff is easy to review:

- **Commit 1 — `aggregate node parameters into TrtYoloXDetectorConfig`**: Gathers the node parameters into a `TrtYoloXDetectorConfig` struct.
- **Commit 2 — `extract TrtYoloXDetector class within node file`**: Introduces `TrtYoloXDetector` (defined inline in the node files) and makes the node delegate detection to it. **This is the commit to review for the actual change.**
- **Commit 3 — `move TrtYoloXDetector to its own files`**: A pure file move into `tensorrt_yolox_detector.hpp/.cpp`. **Nothing to review here** — to see what actually changed, look at Commit 2.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

```bash
colcon build --packages-select autoware_tensorrt_yolox
colcon test --packages-select autoware_tensorrt_yolox --event-handlers console_cohesion+
```

All 6 tests pass, including the `test_tensorrt_yolox_node` characterization test that exercises the node through its topic pub/sub interface and guards against any behavioral regression.

## Notes for reviewers

Please review Commit 2 for the real change; Commit 1 is parameter plumbing and Commit 3 is a pure move. The cumulative line count looks large, but most of it is the relocation in Commit 3.

## Interface changes

None.

## Effects on system behavior

None.


perception/autoware_tensorrt_yolox/include/autoware/tensorrt_yolox/tensorrt_yolox_detector.hpp
```diff
@@ -0,0 +1,156 @@
...
+           (label == Label::HAZARD && HAZARD);
+  };
+};  // struct RoiOverlaySemsegLabel
+
+/**
+ * @struct TrtYoloXDetectorConfig
+ * @brief Configuration that does not change frame-by-frame. The detector is reconstructed when any
+ * of these values change.
+ */
+struct TrtYoloXDetectorConfig
```
ktro2828, line 67: [IMO] I feel it's better to separate config class into TensorRT related one (e.g. `BackendConfig`) and the other one (e.g. `DetectorConfig`) because it enables us to introduce another runtime backend, such as OnnxRuntime, easily. (...The name of this package is misleading in this case though)
But, we don't have any plan to support the other backend for now, so I just want to know your idea about this!

perception/autoware_tensorrt_yolox/src/tensorrt_yolox_detector.cpp
```diff
@@ -0,0 +1,323 @@
...
+    config_.preprocess_on_gpu, config_.gpu_id, config_.calibration_image_list_path, norm_factor,
+    cache_dir, calib_config);
+}
+
+bool TrtYoloXDetector::isGPUInitialized() const
+{
+  return trt_yolox_->isGPUInitialized();
+}
+
+std::optional<TrtYoloXDetectorResult> TrtYoloXDetector::detect(
```
ktro2828: [NIT] How about using `tl::expected` instead of `std::optional`? It makes more sense to me regarding its meaning.

ktro2828 (commented):
@interimadd Thank you for great PR! I left minor comments, please take a look!

perception/autoware_tensorrt_yolox/include/autoware/tensorrt_yolox/tensorrt_yolox_detector.hpp
interimadd: Thank you for the comment! That's a fair point, and I agree the separation makes sense conceptually.

But I think it's better to decide on the structure once we actually need to introduce another backend. It's hard to design the best structure for a backend that doesn't exist yet, and the cost of introducing a new structure later is low once unit tests are implemented.

So I'd prefer to keep the single config for this PR.

perception/autoware_tensorrt_yolox/src/tensorrt_yolox_detector.cpp
interimadd: @ktro2828 
Thanks for the suggestion! I agree that `tl::expected` is a better fit here.
I've pushed a commit for this: https://github.com/autowarefoundation/autoware_universe/pull/12794/commits/bf6f276b8d540465b2c6ccd9755b770e490bd967

Note that this includes one behavioral change: both the `cv_bridge` exception and `doInference` failures are now logged at the error level. I did this because I don't think it makes sense to treat them differently. Both mean the node failed to produce a result. Please let me know if you'd prefer a different approach.

perception/autoware_tensorrt_yolox/include/autoware/tensorrt_yolox/tensorrt_yolox_detector.hpp
ktro2828: Sounds reasonable and agree with you, thanks!

perception/autoware_tensorrt_yolox/src/tensorrt_yolox_detector.cpp
ktro2828: Thank your for quick commitment! Looks good to me :+1:

ktro2828 (approved):
LGTM, thanks!

PR #12795, opened 2026-06-16
refactor(autoware_pointcloud_preprocessor): decouple distortion corrector core from ROS

mojomex:
## Description

Decouples the `distortion_corrector` core (`DistortionCorrectorBase` / `DistortionCorrector2D` / `DistortionCorrector3D`) from the ROS runtime so it can be reused as a plain library.

ROS message types are still used as data contracts (`TwistWithCovarianceStamped`, `Imu`, `PointCloud2`, `TransformStamped`, ...), but runtime concerns now stay in the ROS node wrapper:

- TF lookup is owned by `DistortionCorrectorComponent` via `ManagedTransformBuffer`.
- The core no longer stores an `rclcpp::Node &`, owns a TF buffer, logs, uses node clocks, or includes `rclcpp`.
- The core exposes FIFO-style sensor input APIs:
  - `insert_twist(...)`
  - `insert_imu(...)`
  - `undistort_pointcloud(..., lidar_to_base_link, pointcloud)`
- The caller is responsible for inserting twist/IMU samples in nondecreasing timestamp order.
- On pointcloud input, the core prunes old twist/IMU samples that can no longer affect future pointclouds while keeping the latest sample before the first point timestamp.
- `initialize()` and pointcloud-transform setup are now internal implementation details of the pointcloud undistortion call.
- Validation/logging outcomes are returned through `PointcloudValidity` / `UndistortionResult`; the ROS node maps those results back to the existing throttled log messages.

## How was this PR tested?

Built and tested locally from the rebased branch head:

```bash
colcon build --packages-up-to autoware_pointcloud_preprocessor --cmake-args -DCMAKE_BUILD_TYPE=Release
colcon test --packages-select autoware_pointcloud_preprocessor --event-handlers console_direct+
```

`autoware_pointcloud_preprocessor` test result: 18/18 tests passed. The updated `test_distortion_corrector_node` suite includes 25 cases, including coverage for FIFO pruning of old twist/IMU samples.


mojomex:
If @vividf could review this first :pray:

sensing/autoware_pointcloud_preprocessor/src/distortion_corrector/distortion_corrector_node.cpp
```diff
@@ -99,35 +103,55 @@ void DistortionCorrectorComponent::pointcloud_callback(PointCloud2::UniquePtr po
...
+      }
     }
-  }
 
-  distortion_corrector_->undistort_pointcloud(use_imu_, angle_conversion_opt_, *pointcloud_msg);
+    const auto undistortion_result = distortion_corrector_->undistort_pointcloud(
+      use_imu_, angle_conversion_opt_, *lidar_to_base_link, *pointcloud_msg);
+    log_undistortion_result(undistortion_result, *pointcloud_msg);
+  }
+  if (!lidar_to_base_link.has_value()) {
```
vividf, line 148: we can use else here?

sensing/autoware_pointcloud_preprocessor/src/distortion_corrector/distortion_corrector_node.cpp
```diff
@@ -99,35 +103,55 @@ void DistortionCorrectorComponent::pointcloud_callback(PointCloud2::UniquePtr po
...
+        managed_tf_buffer_->getTransform<geometry_msgs::msg::TransformStamped>(
+          base_frame_, msg->header.frame_id, this->now(), rclcpp::Duration::from_seconds(1.0),
+          this->get_logger());
+      if (!imu_to_base_link.has_value()) {
+        continue;
+      }
+      distortion_corrector_->set_imu_transform(*imu_to_base_link);
+      distortion_corrector_->insert_imu(msg);
     }
   }
```
vividf, line 122:
```suggestion
if (use_imu_) {
  auto imu_msgs = imu_sub_->take_data();
  if (!imu_msgs.empty()) {
    const auto imu_to_base_link =
      managed_tf_buffer_->getTransform<geometry_msgs::msg::TransformStamped>(
        base_frame_, imu_msgs.front()->header.frame_id, this->now(),
        rclcpp::Duration::from_seconds(1.0), this->get_logger());
    if (imu_to_base_link.has_value()) {
      distortion_corrector_->set_imu_transform(*imu_to_base_link);
      for (const auto & msg : imu_msgs) {
        distortion_corrector_->insert_imu(msg);
      }
    }
  }
}
```

vividf (changes requested):
Thanks for the PR!
I just left some comments. 

Just one question: 
The change drops the previous rosbag-replay / backward-time-jump handling. The old process_twist_message / enqueue_imu cleared the queues when time jumped backwards (// e.g. when a rosbag restarts), but the new insert_twist / insert_imu only push to the queue, and neither the core nor the node re-adds this guard. On a rosbag loop/restart the queues become non-monotonic, which breaks the lower_bound assumption in get_twist_and_imu_iterator and leads to incorrect undistortion (live operation is unaffected).

Could you confirm whether dropping rosbag-replay support is intentional? If not, I think the clear-on-backward-jump logic should be reintroduced in the node before the insert_* calls.

sasakisasaki:
Now I'm working on this PR :wink:

PR #12799, opened 2026-06-16
perf(autoware_cuda_pointcloud_preprocessor): hand node streams to cuda_blackboard

mojomex:
## Summary
Move to new `cuda_blackboard` stream-ordered API.

## Dependency
- Depends on autowarefoundation/cuda_blackboard#22


manato (changes requested):
@mojomex
Thank you very much for aligning with the new cuda_blackboard usage. The modifications made up to now look good.
Would you mind adding similar support to `cuda_outlier_filter/cuda_polar_voxel_noise_filter`, as well?

mojomex:
@manato Good catch! Migrated the remaining `CudaPolarVoxelNoiseFilter` too!

manato (approved):
LGTM! Thank you for fixing!

PR #12803, opened 2026-06-17
refactor(tensorrt_yolox): unify preprocessing on GPU and remove CPU path

interimadd:
## Description

Remove the `use_gpu_preprocess` / `preprocess_on_gpu` option and the entire CPU preprocessing code path from `autoware_tensorrt_yolox`, so that preprocessing is always performed on the GPU.

All shipped configurations already set `preprocess_on_gpu: true`. The CPU path was therefore effectively dead code that doubled the preprocessing implementation and complicated every preprocessing branch. This change unifies on the GPU path and deletes the now-unreachable CPU code.

## Related links

Introducing PR (where the `preprocess_on_gpu` option was originally added): https://github.com/autowarefoundation/autoware_universe/pull/3430

## How was this PR tested?

Built and tested both affected packages on a machine with an NVIDIA RTX 4060 (CUDA 12.4, TensorRT):

- `colcon build --packages-select autoware_tensorrt_yolox` and `colcon test`: 6/6 tests passed.
- `colcon build --packages-select autoware_traffic_light_fine_detector` and `colcon test`: 5/5 tests passed.

The end-to-end characterization test `test_tensorrt_yolox_node` actually ran (not skipped) on the GPU and passed all three cases.

## Notes for reviewers

None.

## Interface changes

The `preprocess_on_gpu` ROS parameter is removed. It was present in all three parameter sets (`yolox_s_plus_opt`, `yolox_tiny`, `yolox_traffic_light_detector`) and always set to `true`, which is now the unconditional behavior.

#### Additions and removals

| Change type | Parameter Name      | Type   | Default Value | Description                                         |
|:------------|:--------------------|:-------|:--------------|:----------------------------------------------------|
| Removed     | `preprocess_on_gpu` | `bool` | `true`        | Selected GPU vs CPU preprocessing; now fixed to GPU |

## Effects on system behavior

None for default deployments: every shipped configuration already ran preprocessing on the GPU, so the runtime behavior is identical. The only difference is that the CPU preprocessing fallback is no longer selectable.

### DIFF
--- a/perception/autoware_tensorrt_yolox/config/yolox_s_plus_opt.param.yaml
+++ b/perception/autoware_tensorrt_yolox/config/yolox_s_plus_opt.param.yaml
@@ -32,6 +32,5 @@
     quantize_last_layer: false # If true, set the operating precision for the last (output) layer to be fp16. This option is valid only when precision==int8.
     profile_per_layer: false # If true, profiler function will be enabled. Since the profile function may affect execution speed, it is recommended to set this flag true only for development purpose.
     clip_value: 6.0 # If positive value is specified, the value of each layer output will be clipped between [0.0, clip_value]. This option is valid only when precision==int8 and used to manually specify the dynamic range instead of using any calibration.
-    preprocess_on_gpu: true # If true, pre-processing is performed on GPU.
     gpu_id: 0 # GPU ID to select CUDA Device
     calibration_image_list_path: "" # Path to a file which contains path to images. Those images will be used for int8 quantization.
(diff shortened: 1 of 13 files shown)


technolojin (approved):
I agree to the statement.

ktro2828:
[Quick note]
We also need to update Autoware launcher side.

ktro2828 (approved):
@interimadd LGTM! But, please give me a moment to prepare launcher PRs 🙏 

- https://github.com/autowarefoundation/autoware_launch/pull/1889

PR #12810, opened 2026-06-17
feat(multi object tracker): optimze execution for diffusion planner

technolojin:
## Description

Optimizes the `multi_object_tracker` publishing/execution path so tracked objects can be exported in a form better suited to the diffusion planner (low-latency, ego-consistent output). The change decouples two concerns that were previously fused into the single `enable_delay_compensation` flag and adds an odometry-based ego-pose backend.

**1. Split the publish trigger from the export-time reference**

The old `enable_delay_compensation` bool controlled *both* the publish trigger (timer vs. on-measurement) *and* the export timestamp (extrapolate to now vs. last detection). These are now two orthogonal parameters:

- `publish_on_timer` (bool): publish trigger only. `false` publishes on measurement (default), `true` publishes from the periodic timer at `publish_rate`.
- `delay_compensation` (enum): which timestamp the published tracks are predicted to, ordered least → most aggressive:
  - `none` — latest detection/update stamp (no compensation)
  - `publish_delay` — detection stamp advanced by the update→publish delay (compensates publish-side latency without over-extrapolating the sensor→tracker pipeline delay)
  - `odometry` — compensate up to the newest buffered odometry stamp
  - `full` — full compensation to the current wall-clock time (old `enable_delay_compensation: true` behavior)

**2. Add an odometry-based ego-pose source**

New `ego_source` parameter selects the backend used to obtain the time-varying ego pose (world ← ego):

- `tf` — look it up from the TF tree (legacy behavior)
- `odometry` — interpolate it from a subscribed `~/input/odometry` buffer (default remapped to `/localization/kinematic_state`, the same source the diffusion planner consumes)

The odometry backend (`lib/odometry.cpp`):
- buffers odometry samples (≤ 2 s history, oldest evicted),
- interpolates pose at the requested time (linear position/twist/covariance, SLERP orientation),
- bounded constant-twist extrapolation (≤ 1 s) beyond buffer edges,
- composes `world ← source = world ← ego(t) [odom] * ego ← source [static TF]` for non-ego frames,
- **falls back to TF on any buffer miss** (throttled warning), so behavior degrades gracefully.

Supporting additions: `EgoSource`/`DelayReference` enums, strict `toEgoSource`/`toDelayReference` string parsers (throw on invalid value), `updateOdometryBuffer()` / `getLatestOdometryTime()`, and the `~/input/odometry` subscriber (only created when `ego_source == odometry`).

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

- Built the package locally.
- No unit tests added yet for the odometry interpolation/extrapolation buffer or the parse helpers — to be added.

## Notes for reviewers

- `enable_delay_compensation` is **removed** and replaced by `publish_on_timer` + `delay_compensation`. Any external launch/config setting `enable_delay_compensation` must be migrated (`true` → `publish_on_timer: true` + `delay_compensation: full`).
- Heads-up on default-value consistency across the layers, please confirm intent:
  - `config/multi_object_tracker_node.param.yaml`: `ego_source: tf`, `delay_compensation: publish_delay`
  - `schema/...schema.json`: `ego_source` default `tf`, `delay_compensation` default `none`
  - `launch/multi_object_tracker.launch.xml`: `ego_source` arg default `odometry` (explicit `<param>` overrides the yaml, so the effective launch default is `odometry`)
  - `design/MultiObjectTracker.node.yaml`: `ego_source` default `odometry`

### Subscribed topics

|:--|:--|:--|:--|:--|
| Added | Sub | `~/input/odometry` | `nav_msgs/msg/Odometry` | Ego pose source when `ego_source: odometry`. Default remap `/localization/kinematic_state`. |

#### Additions and removals

|:--|:--|:--|:--|:--|
| Removed | `enable_delay_compensation` | `bool` | `false` | Replaced by `publish_on_timer` + `delay_compensation`. |
| Added | `ego_source` | `string` | `tf` | Ego-pose backend: `tf` or `odometry`. |
| Added | `publish_on_timer` | `bool` | `false` | Publish trigger: measurement vs. periodic timer. |
| Added | `delay_compensation` | `string` | `none` | Export-time reference: `none` / `publish_delay` / `odometry` / `full`. |

## Effects on system behavior

- Default behavior with `delay_compensation: none` / `publish_on_timer: false` publishes on measurement at the detection stamp (no timer, no extrapolation), differing from the previous default which still published on measurement but without the new explicit reference selection.
- With `ego_source: odometry`, the ego pose is interpolated from `/localization/kinematic_state` instead of TF, with automatic TF fallback on buffer miss.

### DIFF
--- a/perception/autoware_multi_object_tracker/config/multi_object_tracker_node.param.yaml
+++ b/perception/autoware_multi_object_tracker/config/multi_object_tracker_node.param.yaml
@@ -4,13 +4,18 @@
     publish_rate: 10.0
     world_frame_id: map
     ego_frame_id: base_link
-    consider_odometry_uncertainty: false
 
     # ego pose source: "tf" (look up from TF tree) or "odometry" (interpolate from ~/input/odometry)
-    ego_source: tf
+    # publish-trigger side: false publishes on measurement, true publishes from the periodic timer.
     # object-export side: timestamp the published tracks are predicted to.
-    # "detection" (publish on measurement, no timer) | "now" | "latest_odometry"
+    publish_rate: 10.0
+    publish_on_timer: true
+
+    # "detection" (latest detection/update time) | "now" | "latest_odometry"
+    ego_source: odometry
     delay_compensation: detection
+
+    consider_odometry_uncertainty: false
 
     # Per-tracker-type configuration
     tracker_configs:
--- a/perception/autoware_multi_object_tracker/config/multi_object_tracker_node.param.yaml
+++ b/perception/autoware_multi_object_tracker/config/multi_object_tracker_node.param.yaml
@@ -1,7 +1,6 @@
 /**:
   ros__parameters:
     # default tracker node parameters
-    publish_rate: 10.0
     world_frame_id: map
     ego_frame_id: base_link
 
--- a/perception/autoware_multi_object_tracker/config/multi_object_tracker_node.param.yaml
+++ b/perception/autoware_multi_object_tracker/config/multi_object_tracker_node.param.yaml
@@ -12,7 +12,7 @@
     ego_source: tf
     # delay compensation, least -> most:
     # "none" | "publish_delay" (detection + update->publish delay) | "odometry" (up to latest odometry) | "full" (to now)
-    delay_compensation: publish_delay
+    delay_compensation: none
 
     consider_odometry_uncertainty: false
 
--- a/perception/autoware_multi_object_tracker/config/multi_object_tracker_node.param.yaml
+++ b/perception/autoware_multi_object_tracker/config/multi_object_tracker_node.param.yaml
@@ -4,8 +4,13 @@
     publish_rate: 10.0
     world_frame_id: map
     ego_frame_id: base_link
-    enable_delay_compensation: false
     consider_odometry_uncertainty: false
+
+    # ego pose source: "tf" (look up from TF tree) or "odometry" (interpolate from ~/input/odometry)
+    ego_source: tf
+    # object-export side: timestamp the published tracks are predicted to.
+    # "detection" (publish on measurement, no timer) | "now" | "latest_odometry"
+    delay_compensation: detection
 
     # Per-tracker-type configuration
     tracker_configs:
--- a/perception/autoware_multi_object_tracker/config/multi_object_tracker_node.param.yaml
+++ b/perception/autoware_multi_object_tracker/config/multi_object_tracker_node.param.yaml
@@ -9,10 +9,10 @@
     publish_rate: 10.0
 
     # ego pose source: "tf" (look up from TF tree) or "odometry" (interpolate from ~/input/odometry)
-    ego_source: odometry
+    ego_source: tf
     # delay compensation, least -> most:
     # "none" | "publish_delay" (detection + update->publish delay) | "odometry" (up to latest odometry) | "full" (to now)
-    delay_compensation: none
+    delay_compensation: publish_delay
 
     consider_odometry_uncertainty: false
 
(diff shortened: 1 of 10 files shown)


perception/autoware_multi_object_tracker/config/multi_object_tracker_node.param.yaml
```diff
@@ -1,10 +1,19 @@
...
+
+    # publish-trigger
+    publish_on_timer: false
+    publish_rate: 10.0
+
+    # ego pose source: "tf" (look up from TF tree) or "odometry" (interpolate from ~/input/odometry)
+    ego_source: tf
+    # delay compensation, least -> most:
+    # "none" | "publish_delay" (detection + update->publish delay) | "odometry" (up to latest odometry) | "full" (to now)
+    delay_compensation: publish_delay
```
YoshiRi: [non-blocking]
json schema's default is "none". Do you need "publish_delay" as a default? `none` seems to be closer to previous no-compensation settings.

perception/autoware_multi_object_tracker/lib/odometry.cpp
```diff
@@ -18,26 +18,257 @@
...
+  for (auto it = odom_buffer_.begin(); it != odom_buffer_.end();) {
+    if (newest - it->first > max_age) {
+      it = odom_buffer_.erase(it);
+    } else {
+      ++it;
+    }
+  }
+}
+
+std::optional<rclcpp::Time> Odometry::getLatestOdometryTime() const
```
YoshiRi: [just a comment]
This function gather latest tf when `ego_source: tf` and there is no odometry subscriber. So It sounds better to update this function name getLatestEgoPoseTime or something.

perception/autoware_multi_object_tracker/src/multi_object_tracker_core.cpp
```diff
@@ -288,8 +289,27 @@ PublishingData prepare_publishing_data(
...
+      // Advance the detection stamp by the update->publish delay (wall-clock elapsed since the last
+      // tracker update). Compensates the publish-side latency without over-extrapolating by the
+      // sensor->tracker pipeline delay (as FULL would).
+      const auto elapsed = current_time - state.last_updated_time;
+      result.object_time =
+        last_tracker_time + (elapsed.seconds() > 0.0 ? elapsed : rclcpp::Duration(0, 0));
+      break;
+    }
+    case DelayReference::ODOMETRY:
+      result.object_time = state.odometry->getLatestOdometryTime().value_or(current_time);
```
YoshiRi: [just a comment]
For `DelayReference::ODOMETRY`, falling back to `current_time` on missing odometry effectively turns this mode into `full` compensation. That seems unsafe because users selecting `odometry` likely want to avoid extrapolating beyond the newest ego-pose sample. Would `last_tracker_time` or skipping publish be a safer fallback?

YoshiRi (approved):
LGTM with non-blocking comments

perception/autoware_multi_object_tracker/src/multi_object_tracker_core.cpp
technolojin: in some aspect, full compensation could be considered as more safer fallback.

perception/autoware_multi_object_tracker/lib/odometry.cpp
technolojin: fixed https://github.com/autowarefoundation/autoware_universe/pull/12810/changes/8c2d62975702d4e5d049fc331187a62da6eda4dc

perception/autoware_multi_object_tracker/config/multi_object_tracker_node.param.yaml
technolojin: fixed the default to none
https://github.com/autowarefoundation/autoware_universe/pull/12810/changes/76397c17455ed695562d02f5dc92b9197481bef5

technolojin:
parameter sync

https://github.com/autowarefoundation/autoware_launch/pull/1894

PR #12817, opened 2026-06-18
chore: update master `scene_module_manager.param.yaml` to match autoware_launch

paulsohn:
## Description

update `scene_module_manager.param.yaml` to match autoware_launch side. https://github.com/autowarefoundation/autoware_launch/blob/c87d313100095553794b311399fb8b5db4aad9f9/autoware_launch/config/planning/scenario_planning/lane_driving/behavior_planning/behavior_path_planner/scene_module_manager.param.yaml

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None. param files in autoware_universe serves as an example (param files in autoware_launch is real), and my intent is to extend the role to serve as a master file to sync upstream and downstream while allowing downstream to override some of the fields.

See sync-params workflow: https://github.com/autowarefoundation/autoware_launch/pull/1812


planning/behavior_path_planner/autoware_behavior_path_planner/config/scene_module_manager.param.yaml
mkquda, line 92: duplicated.

mkquda (approved):

planning/behavior_path_planner/autoware_behavior_path_planner/config/scene_module_manager.param.yaml
paulsohn: deduplicated in https://github.com/autowarefoundation/autoware_universe/pull/12817/commits/3399497bf78ae3f886253ad21e86b06e5c20efe8 .

planning/behavior_path_planner/autoware_behavior_path_planner/config/scene_module_manager.param.yaml
```diff
@@ -76,10 +76,16 @@
       enable_simultaneous_execution_as_candidate_module: true
 
     sampling_planner:
+      enable_module: true
```
paulsohn, line 79: This seems unnecessary but I'm not sure. Might remove in a future PR.

PR #12821, opened 2026-06-18
feat(ptv3): filter segmented points

ktro2828:
## Description

This pull request introduces significant improvements to the CUDA-based segmentation point cloud postprocessing pipeline in `autoware_ptv3`, focusing on filtering logic, output correctness, and test coverage. 
The main changes add support for filtering out segmented classes specified in `ptv3.param.yaml -> filters: [...]`.

**Filtering and Output Logic Improvements:**

* The `createSegmentationPointcloud` CUDA kernel and its host wrapper now support filtering out points with specified class indices, using a filter list and atomic operations to write only unfiltered points to the output. The function now returns the number of output points after filtering. [[1]](diffhunk://#diff-37a51e10daaf881a8c20e2446dd011d5be5434e34e42ee79e93db1568671a6d5L153-R155) [[2]](diffhunk://#diff-37a51e10daaf881a8c20e2446dd011d5be5434e34e42ee79e93db1568671a6d5R165-R172) [[3]](diffhunk://#diff-37a51e10daaf881a8c20e2446dd011d5be5434e34e42ee79e93db1568671a6d5L174-R192) [[4]](diffhunk://#diff-37a51e10daaf881a8c20e2446dd011d5be5434e34e42ee79e93db1568671a6d5L408-R440) [[5]](diffhunk://#diff-af64bd733046e967f45aafab57e76ffe1143981ee4706186d8eac940e72037caL40-R40)
* The postprocessing pipeline (`PTv3TRT::postProcess`) now sets the output point cloud width based on the actual number of segmented (unfiltered) points, improving downstream correctness.

## Related links

**Parent PRs:**

- https://github.com/autowarefoundation/autoware_universe/pull/12819
- https://github.com/autowarefoundation/autoware_universe/pull/12820

## How was this PR tested?

Visualization result comparison of `ptv3/segmentation`.

left: old, right: new

Screenshot from 2026-07-03 16-28-23

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_ptv3/CMakeLists.txt
+++ b/perception/autoware_ptv3/CMakeLists.txt
@@ -174,7 +174,7 @@ if(TRT_AVAIL AND CUDA_AVAIL)
       include
       ${CUDA_INCLUDE_DIRS}
     )
-        
+
   endif()
 
   ament_auto_package(
--- a/perception/autoware_ptv3/CMakeLists.txt
+++ b/perception/autoware_ptv3/CMakeLists.txt
@@ -163,6 +163,18 @@ if(TRT_AVAIL AND CUDA_AVAIL)
       ${PROJECT_NAME}_cuda_lib
     )
 
+    ament_add_gtest(postprocess_kernel_test
+      test/postprocess_kernel_test.cpp
+    )
+    target_link_libraries(postprocess_kernel_test
+      CUDA::cudart
+      ${PROJECT_NAME}_cuda_lib
+    )
+    target_include_directories(postprocess_kernel_test PRIVATE
+      include
+      ${CUDA_INCLUDE_DIRS}
+    )
+        
   endif()
 
   ament_auto_package(
--- a/perception/autoware_ptv3/config/ptv3.param.yaml
+++ b/perception/autoware_ptv3/config/ptv3.param.yaml
@@ -20,6 +20,7 @@
         class_probability_threshold: 0.05
         classes: ["drivable_flat", "non_drivable_flat", "noise"]
         output_format: ""
+        apply_to_segmentation: false
     detection3d:
       use_head: $(var use_det3d_head)
       onnx_path: "$(var model_path)/$(var model_variant)_det3d_head.onnx"
--- a/perception/autoware_ptv3/config/ptv3.param.yaml
+++ b/perception/autoware_ptv3/config/ptv3.param.yaml
@@ -17,7 +17,6 @@
       workspace_size: 8589934592  # 8 GiB
       source_reconstruction: "full"
       filter:
-        class_probability_threshold: 0.05
         classes: ["drivable_flat", "non_drivable_flat", "noise"]
         output_format: ""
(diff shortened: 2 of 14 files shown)


amadeuszsz:
@codex review

amadeuszsz (changes requested):
Thank you for your PR!

There is some ambiguity. Now we can distinguish two behaviors:
- `~/output/pointcloud/filtered`: drops a point if any filter class probability >= filter.class_probability_threshold
- `~/output/pointcloud/segmentation` (new): drops a point if its argmax label matches, ignoring the threshold

Let's take a look on `~/output/pointcloud/filtered` output cloud and example scenario where we set filtered class to `drivable_flat` and threshold `0.1`. Assume we got a point with `drivable_flat` probability 0.11 and `pedestrian` 0.90. This will filter out `pedestrian` class. On the other hand `~/output/pointcloud/segmentation` will keep this `pedestrian` class as it just uses argmax. Threshold was wrong call, therefore please remove this parameter `segmentation3d.filter.class_probability_threshold` and use only argmax for `~/output/pointcloud/filtered`. Ambiguity gone. We need to accommodate a more suitable (and complex...) strategy for thresholds, but this is out of the scope of this PR.

 Another point is the fact you remove all the information from `~/output/pointcloud/segmentation` about filtered classes. This is not what user may expect as it corrupts the initial sequence of input pointcloud. At the same time I understand this is necessary for downstream module. My proposition is a configuration, which simply let's user to decide about content of `~/output/pointcloud/segmentation`. Can be simply a boolean `segmentation3d.filter.apply_to_segmentation` with default `false`.

ktro2828:
@amadeuszsz Thank you for reviewing! 
I added the following changes:
- Add `filter.apply_to_segmentation` parameter in https://github.com/autowarefoundation/autoware_universe/pull/12821/commits/1e29f67e5546b131258a59aa8f71745a9c6a4357
- Remove `filter.class_probability_threshold` parameter in https://github.com/autowarefoundation/autoware_universe/pull/12821/commits/fc3493619faddc2fbb6f4412f37ac12040e94073

For filtering points with their probability, I think it is OK to consider it in downstream module for now.

Here is the comparison between `filter.apply_to_segmentation:=false(default)` and `true`

|`false(default)` | `true` |
|---------------------|----------|
| Screenshot from 2026-07-23 04-08-00 | Screenshot from 2026-07-23 04-08-37 |

amadeuszsz (approved):
LGTM!

amadeuszsz:
@mojomex (cc: @ktro2828 )
If you can find some time, could you please run the evaluation before merging this PR?

ktro2828:
> @mojomex (cc: @ktro2828 ) If you can find some time, could you please run the evaluation before merging this PR?

@amadeuszsz Sorry! I also need to make PR to remove `experimental` features https://github.com/autowarefoundation/autoware_universe/pull/13044, so please let me evaluate after all PRs have been done!

PR #12822, opened 2026-06-18
feat: visibility method now uses geometric entropy and anisotropy

SergioReyesSan:
## Description

This PR adds a feature to the Polar Voxel Outlier Filter with geometric entropy and anisotropy analysis to improve low-visibility condition detection.

Key Changes

New Feature: Geometric entropy and anisotropy analysis for detecting weather-induced noise (rain, fog) while avoiding false positives on planar surfaces (windows)

Also added a debug feature to visualize the points used for the visibility estimation.

Also, some updates based on previous comments on the polar voxel noise filter helped to improve the pointcloud validation functionality 

Updated documentation and parameters

Benefits
Better discrimination between weather-induced noise and structural features (bus windows)

image
image

image

Tested with some rosbags that were causing problems with low visibility, and now it seems we can avoid triggering those alerts
image

## Related links

**Parent Issue:**
[TIER IV INTERNAL Link](https://tier4.atlassian.net/browse/RDDR-197)

## How was this PR tested?

Tested with the 2 evaluator scenarios and 2 scenarios where a bus was causing low visibility miss detection.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


sensing/autoware_pointcloud_preprocessor/config/polar_voxel_outlier_filter_node.param.yaml
```diff
@@ -1,36 +1,32 @@
...
+    use_return_type_classification: true
+    filter_secondary_returns: false
+    secondary_noise_threshold: 4
+    low_visibility_entropy_threshold: 0.3
+    low_visibility_anisotropy_threshold: 0.95
+    visibility_estimation_max_secondary_voxel_count: 500
+    primary_return_types: [1, 6, 8, 10]
+    visibility_estimation_only: false
+    publish_noise_cloud: false
+    publish_low_visibility_voxels: flase
```
YoshiRi: [typo]
Please fix to "false"

sensing/autoware_pointcloud_preprocessor/src/outlier_filter/polar_voxel_outlier_filter_node.cpp
```diff
@@ -200,19 +211,20 @@ void PolarVoxelOutlierFilterComponent::filter(
     throw std::invalid_argument("Input point cloud is null");
   }
 
-  // Phase 1: Validate inputs
-  validate_filter_inputs(*input, indices);
+  // Phase 1: Validate inputs and detect point cloud format (once)
+  std::call_once(input_format_once_flag_, [this, &input, &indices]() {
```
YoshiRi: [minor comment]
Caching the detected PointCloud2 format with std::call_once is acceptable if the input topic layout is guaranteed to be stable. However, this change also moves validate_filter_inputs() into the call_once block, which means input validation is no longer performed after the first frame. I think the format detection cache and per-frame input validation should be separated.

sensing/autoware_pointcloud_preprocessor/src/outlier_filter/polar_voxel_outlier_filter_node.cpp
```diff
@@ -435,34 +469,51 @@ void PolarVoxelOutlierFilterComponent::publish_noise_cloud(
...
 
-  uint32_t low_visibility_voxels_count = 0;
-  for (const auto & [voxel_idx, counts] : voxel_point_counts) {
+  const auto low_visibility_candidates = collect_low_visibility_candidates(voxel_point_counts);
+
+  selected_low_visibility_voxels_.clear();
+  selected_low_visibility_voxels_.reserve(low_visibility_candidates.size());
+  for (const auto & candidate : low_visibility_candidates) {
+    const auto geometric_metrics =
+      calculate_geometric_shannon_metrics(input, point_voxel_info, candidate);
```
YoshiRi: [performance issue]

This function scans the entire point cloud and is called for each candidate voxel, so the current implementation can be computationally expensive.

I think we should avoid scanning the full point cloud repeatedly. Instead, we could accumulate the required statistics in a single pass using a structure like the following:

```cpp
struct GeometricStatsAccumulator
{
  size_t point_count{0};
  double sum_x{0.0};
  double sum_y{0.0};
  double sum_z{0.0};
  double sum_xx{0.0};
  double sum_xy{0.0};
  double sum_xz{0.0};
  double sum_yy{0.0};
  double sum_yz{0.0};
  double sum_zz{0.0};

  void add(const double x, const double y, const double z)
  {
    ++point_count;
    sum_x += x;
    sum_y += y;
    sum_z += z;
    sum_xx += x * x;
    sum_xy += x * y;
    sum_xz += x * z;
    sum_yy += y * y;
    sum_yz += y * z;
    sum_zz += z * z;
  }

  GeometricShannonMetrics calculate_metrics() const;
};
```

With this approach, we can scan the point cloud once, accumulate per-voxel statistics, and then calculate entropy/anisotropy only for the candidate voxels. This keeps the metric local to each voxel while reducing the complexity from roughly `O(num_candidate_voxels * num_points)` to `O(num_points + num_candidate_voxels)`.

YoshiRi (changes requested):
SRY for my late review. Plz check some comments.

SergioReyesSan:
@YoshiRi Thanks for your comments. I will update them, but I found that the current approach is also not feasible for complex structures with glass, so it is not passing all the MRM cases. It passes most of them, but still not even with entropy and anisotropy is not passing all the cases, so currently I'm modifying it to include other metrics like average intensity and ratio of secondary return appreciate your patience

sensing/autoware_pointcloud_preprocessor/config/polar_voxel_outlier_filter_node.param.yaml
SergioReyesSan: Thanks, fixed here:

e1b4f6d008d89a010e3cb934717d8407411c0122

sensing/autoware_pointcloud_preprocessor/src/outlier_filter/polar_voxel_outlier_filter_node.cpp
SergioReyesSan: I included these changes because that's how we did it on the Polar voxel noise filter suggestion from Amadeuz, so, shall we keep it or shall we separate it?

sensing/autoware_pointcloud_preprocessor/src/outlier_filter/polar_voxel_outlier_filter_node.cpp
SergioReyesSan: Added as you requested; however, to keep the calculations light, I'm only calculating those values if the voxel is in the visibility range, which are the ones of our interest

changed here:

e1b4f6d008d89a010e3cb934717d8407411c0122

SergioReyesSan: Also added those comments specifically for the sections required

SergioReyesSan:
@YoshiRi @manato Sorry for keeping you waiting. I added an extra metric to help us better evaluate voxel noise for rain/fog detection.

This time, I verified it using more rosbags that were causing false detections of low-visibility conditions, mainly due to buses or bus stops with large glass or acrylic surfaces. The low-visibility threshold now seems much clearer. However, it could still fail in challenging scenarios with many car windows around the vehicle, such as in parking lots.

I seriously believe this new method will help reduce unnecessary MRM triggering.

image

Also attached some extra fields that helped me a lot to find conditions where the previous approach was failing; these are only visible when publishing the low_visibility_voxels point cloud and the noise pointcloud

sensing/autoware_pointcloud_preprocessor/src/outlier_filter/polar_voxel_outlier_filter_node.cpp
YoshiRi: Good! set this as resolved

sensing/autoware_pointcloud_preprocessor/src/outlier_filter/polar_voxel_outlier_filter_node.cpp
YoshiRi: Okay, It is not a main issue. So let it be keeped.

sensing/autoware_pointcloud_preprocessor/src/outlier_filter/polar_voxel_outlier_filter_node.cpp
```diff
@@ -14,15 +14,18 @@
 
 #include "autoware/pointcloud_preprocessor/outlier_filter/polar_voxel_outlier_filter_node.hpp"
 
+#include <Eigen/Eigenvalues>
```
manato, line 17: [nits] I would propose adding 

``` xml
<depend>eigen</depend>
```

into `package.xml` to declare Eigen dependency explicitly.

sensing/autoware_pointcloud_preprocessor/src/outlier_filter/polar_voxel_outlier_filter_node.cpp
```diff
@@ -42,6 +45,9 @@ static constexpr double diagnostics_update_period_sec = 0.1;
 static constexpr size_t point_cloud_height_organized = 1;
 static constexpr double TWO_PI = 2.0 * M_PI;
 static constexpr int marker_resolution = 50;
+static constexpr size_t low_visibility_debug_field_count = 9;
+static constexpr size_t low_visibility_sparse_voxel_point_count_threshold = 10;
+static constexpr float low_visibility_sparse_voxel_secondary_return_ratio_threshold = 0.2F;
```
manato: [ask] It looks these thresholds look to be configurable values. Do you think these values will never be changed for any debug scenarios? If there are scenarios where tuning these values are helpful for debugging, I would propose these as ROS params.

sensing/autoware_pointcloud_preprocessor/src/outlier_filter/polar_voxel_outlier_filter_node.cpp
```diff
@@ -200,19 +218,19 @@ void PolarVoxelOutlierFilterComponent::filter(
...
-
-  if (has_polar_coords) {
-    RCLCPP_DEBUG_ONCE(
-      get_logger(), "Processing PointXYZIRCAEDT format with pre-computed polar coordinates");
-  } else {
-    RCLCPP_DEBUG_ONCE(
-      get_logger(), "Processing PointXYZIRC format, computing azimuth and elevation");
-  }
+    input_format_ = InputPointCloudFormat::PointXYZIRC;
+    RCLCPP_DEBUG(get_logger(), "Processing PointXYZIRC format, computing azimuth and elevation");
```
manato:
```suggestion
    } else {
    input_format_ = InputPointCloudFormat::PointXYZIRC;
    RCLCPP_DEBUG(get_logger(), "Processing PointXYZIRC format, computing azimuth and elevation");
    }
```

[nits] Nothing will be changed essentially, but I think we can use `else` to explicitly shows these are mutual exclusive.

sensing/autoware_pointcloud_preprocessor/src/outlier_filter/polar_voxel_outlier_filter_node.cpp
```diff
@@ -435,34 +558,79 @@ void PolarVoxelOutlierFilterComponent::publish_noise_cloud(
...
   // Calculate metrics
-  calculate_visibility_metric(voxel_point_counts);
+  calculate_visibility_metric(voxel_point_counts, geometric_metrics);
   calculate_filter_ratio_metric(valid_points_mask);
 
   // Publish metrics
   publish_visibility_metric();
   publish_filter_ratio_metric();
+  publish_low_visibility_voxels(input, point_voxel_info);
+  updater_.force_update();
```
manato:
```suggestion

```

[ask] This `updater_.force_update();` was deleted in #12006 to avoid duplicated publishing during 1 iteration. Do you intentionally re-add this line?

sensing/autoware_pointcloud_preprocessor/src/outlier_filter/polar_voxel_outlier_filter_node.cpp
```diff
@@ -504,6 +672,90 @@ void PolarVoxelOutlierFilterComponent::publish_filter_ratio_metric()
...
+  // Extra fields provide visual assistance when inspecting low-visibility voxels during debugging.
+  append_field("voxel_radius_idx", sensor_msgs::msg::PointField::FLOAT32);
+  append_field("voxel_azimuth_idx", sensor_msgs::msg::PointField::FLOAT32);
+  append_field("voxel_elevation_idx", sensor_msgs::msg::PointField::FLOAT32);
+  append_field("voxel_entropy", sensor_msgs::msg::PointField::FLOAT32);
+  append_field("voxel_anisotropy", sensor_msgs::msg::PointField::FLOAT32);
+  append_field("voxel_point_count", sensor_msgs::msg::PointField::FLOAT32);
+  append_field("voxel_average_intensity", sensor_msgs::msg::PointField::FLOAT32);
+  append_field("voxel_secondary_return_count", sensor_msgs::msg::PointField::FLOAT32);
+  append_field("voxel_secondary_return_ratio", sensor_msgs::msg::PointField::FLOAT32);
```
manato: The code around here looks same as the one in `publish_noise_cloud()`.
Would you mind considering extracting these into a function?

manato (commented):
@SergioReyesSan
Thank you very much for updating `polar_voxel_outlier_filter`. The result looks promising! The algorithm update looks nice to me. I left some comments where I noticed. Your consideration is appreciated!

manato:
@codex review, please :pray:

sensing/autoware_pointcloud_preprocessor/src/outlier_filter/polar_voxel_outlier_filter_node.cpp
```diff
@@ -247,7 +274,8 @@ PolarVoxelOutlierFilterComponent::collect_voxel_info(const PointCloud2 & input)
   PointVoxelInfoVector point_voxel_info;
   point_voxel_info.reserve(input.width * input.height);
 
-  bool has_polar_coords = has_polar_coordinates(input);
+  const bool has_polar_coords =
+    input_format_ == InputPointCloudFormat::PointXYZIRCAEDT || has_polar_coordinates(input);
```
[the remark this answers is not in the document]
manato: We can assume the input pointcloud format is identical during runtime, so I believe we can skip this.

sensing/autoware_pointcloud_preprocessor/src/outlier_filter/polar_voxel_outlier_filter_node.cpp
SergioReyesSan: You're right. I left it as a constant because I noticed that the behaviour was good with those values, but just in case we need to tune it, it will definitely be easier as a parameter

54fdfa1d0483e8c537f604a794e9a77e3c9047cf

sensing/autoware_pointcloud_preprocessor/src/outlier_filter/polar_voxel_outlier_filter_node.cpp
SergioReyesSan: Ok, adressed here

54fdfa1d0483e8c537f604a794e9a77e3c9047cf

sensing/autoware_pointcloud_preprocessor/src/outlier_filter/polar_voxel_outlier_filter_node.cpp
SergioReyesSan: You are right, not needed anymore

54fdfa1d0483e8c537f604a794e9a77e3c9047cf

sensing/autoware_pointcloud_preprocessor/src/outlier_filter/polar_voxel_outlier_filter_node.cpp
SergioReyesSan: Created a function that appends the same fields to both noise and low_visibility voxels. But the point clouds are different, so I only set  that function as common 

54fdfa1d0483e8c537f604a794e9a77e3c9047cf

SergioReyesSan:
@manato Thank you very much for your feedback. I updated the PR according to your comments.

manato (approved):
@SergioReyesSan
Thank you for handling my comments. Now, all of my concerns have been cleared. LGTM!

SergioReyesSan:
@YoshiRi @manato Thank you very much for the feedback. If there are no more comments or feedback, then we should merge it. Please react or leave a comment.

manato:
@SergioReyesSan 
Thank you for your confirmation. Since all comments from the reviewers are marked as "resolved", I believe all concerns have been handled. 

@YoshiRi 
Because your review status remained "Request changes", I re-request a review. Sorry for taking up your time, but would you mind approving this PR if you don't have any other concerns? :pray:  
(Besides, since I'm still not a maintainer of `autoware_pointcloud_preprocessor`, your approval as a package maintainer is also appreciated. :bow: )

YoshiRi (approved):
thanks for update. LGTM for now.

YoshiRi:
@SergioReyesSan @manato 
Let me merge latest changes to rerun CI test failure.

It seems there are CI test failure.


```
test 10
      Start 10: test_polar_voxel_outlier_filter_node

10: Test command: /usr/bin/python3 "-u" "/opt/ros/humble/share/ament_cmake_test/cmake/run_test.py" "/__w/autoware_universe/autoware_universe/build/autoware_pointcloud_preprocessor/test_results/autoware_pointcloud_preprocessor/test_polar_voxel_outlier_filter_node.gtest.xml" "--package-name" "autoware_pointcloud_preprocessor" "--output-file" "/__w/autoware_universe/autoware_universe/build/autoware_pointcloud_preprocessor/ament_cmake_gtest/test_polar_voxel_outlier_filter_node.txt" "--command" "/__w/autoware_universe/autoware_universe/build/autoware_pointcloud_preprocessor/test_polar_voxel_outlier_filter_node" "--gtest_output=xml:/__w/autoware_universe/autoware_universe/build/autoware_pointcloud_preprocessor/test_results/autoware_pointcloud_preprocessor/test_polar_voxel_outlier_filter_node.gtest.xml"
10: Test timeout computed to be: 60
10: -- run_test.py: invoking following command in '/__w/autoware_universe/autoware_universe/build/autoware_pointcloud_preprocessor':
10:  - /__w/autoware_universe/autoware_universe/build/autoware_pointcloud_preprocessor/test_polar_voxel_outlier_filter_node --gtest_output=xml:/__w/autoware_universe/autoware_universe/build/autoware_pointcloud_preprocessor/test_results/autoware_pointcloud_preprocessor/test_polar_voxel_outlier_filter_node.gtest.xml
10: [==========] Running 4 tests from 1 test suite.
10: [----------] Global test environment set-up.
10: [----------] 4 tests from PolarVoxelOutlierFilterTest
10: [ RUN      ] PolarVoxelOutlierFilterTest.UseReturnTypeClassification_True_PrimaryOnly
10: [INFO] [1784265488.149799067] [PolarVoxelOutlierFilter]: Noise cloud publishing disabled for performance optimization
10: [INFO] [1784265488.150117141] [PolarVoxelOutlierFilter]: Polar Voxel Outlier Filter initialized - supports PointXYZIRC and PointXYZIRCAEDT with advanced two-criteria filtering
10: [       OK ] PolarVoxelOutlierFilterTest.UseReturnTypeClassification_True_PrimaryOnly (11 ms)
10: [ RUN      ] PolarVoxelOutlierFilterTest.UseReturnTypeClassification_False_AllPoints
10: [INFO] [1784265488.155009855] [PolarVoxelOutlierFilter]: Noise cloud publishing disabled for performance optimization
10: [INFO] [1784265488.155240764] [PolarVoxelOutlierFilter]: Polar Voxel Outlier Filter initialized - supports PointXYZIRC and PointXYZIRCAEDT with simple occupancy filtering
10: [       OK ] PolarVoxelOutlierFilterTest.UseReturnTypeClassification_False_AllPoints (5 ms)
10: [ RUN      ] PolarVoxelOutlierFilterTest.IntensityThreshold_SecondaryReturn_NoiseCloud
10: [INFO] [1784265488.159813231] [PolarVoxelOutlierFilter]: Noise cloud publishing enabled
10: [INFO] [1784265488.160075273] [PolarVoxelOutlierFilter]: Polar Voxel Outlier Filter initialized - supports PointXYZIRC and PointXYZIRCAEDT with advanced two-criteria filtering
10: [       OK ] PolarVoxelOutlierFilterTest.IntensityThreshold_SecondaryReturn_NoiseCloud (6 ms)
10: [ RUN      ] PolarVoxelOutlierFilterTest.VisibilityMetric
10: [INFO] [1784265488.165395302] [PolarVoxelOutlierFilter]: Noise cloud publishing disabled for performance optimization
10: [INFO] [1784265488.165541605] [PolarVoxelOutlierFilter]: Polar Voxel Outlier Filter initialized - supports PointXYZIRC and PointXYZIRCAEDT with advanced two-criteria filtering (visibility estimation only - no point cloud output)
10: /__w/autoware_universe/autoware_universe/sensing/autoware_pointcloud_preprocessor/test/test_polar_voxel_outlier_filter_node.cpp:320: Failure
10: The difference between node.visibility_.value() and 0.5 is 0.5, which exceeds 1e-6, where
10: node.visibility_.value() evaluates to 0,
10: 0.5 evaluates to 0.5, and
10: 1e-6 evaluates to 9.9999999999999995e-07.
10: [  FAILED  ] PolarVoxelOutlierFilterTest.VisibilityMetric (5 ms)
10: [----------] 4 tests from PolarVoxelOutlierFilterTest (27 ms total)
10: 
10: [----------] Global test environment tear-down
10: [==========] 4 tests from 1 test suite ran. (27 ms total)
10: [  PASSED  ] 3 tests.
10: [  FAILED  ] 1 test, listed below:
10: [  FAILED  ] PolarVoxelOutlierFilterTest.VisibilityMetric
10: 
10:  1 FAILED TEST
```

PR #12825, opened 2026-06-19
refactor(tensorrt_yolox): remove dead code

interimadd:
## Description

This PR removes dead code from `autoware_tensorrt_yolox`. Every removal was verified to have no caller within the workspace, and the change does not alter runtime behavior.

This PR groups the following commits:

- **remove unused color_masks parameter**
  - `color_masks` argument was threaded through `doInference` and `feedforwardAndDecode` but never read or written. The published color mask is produced separately via `getColorizedMask`, so the parameter was dead.
- **remove dead code (commented-out blocks, unused symbols)**
  - a commented-out block in the constructor and in the batch preprocess function, the header-undeclared `nhwc_to_nchw_gpu` and its kernel, the `load_label_remap_file` declaration (declared but never defined or called; distinct from the live `load_label_id_remap_file`)

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

`colcon build --packages-select autoware_tensorrt_yolox` followed by `colcon test --packages-select autoware_tensorrt_yolox` passes. Integration tests were run in the local test. 

The removed symbols were confirmed to have zero remaining references in the workspace.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


ktro2828 (approved):
LGTM :+1:

PR #12837, opened 2026-06-19
feat(autoware_tensorrt_plugins): bevfusion precompute rulebook in preprocessing stage

vividf:
## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


vividf:
Superseded by #13370, which is the same change rebased onto current `main` plus the work done since this PR went stale: the rulebook geometry and the coordinate column order now travel inside the ONNX (`rulebook_stages` / `rulebook_coors_permutation` in `metadata_props`) instead of being hard-coded, so the runtime no longer assumes the column order the model was exported with — a rulebook built on the wrong order is a valid rulebook for a transposed BEV grid, and the engine would run with wrong detections. #13370 also has the measured before/after (sparse engine 4.015 -> 2.211 ms, 6 -> 0 `[trainStation]` regions, mAP unchanged) and the correct package scope in its title.

PR #12840, opened 2026-06-22
test(autoware_traffic_light_classifier): add characterization test for decoupling the node and the logic

takam5f2:
## Description

This PR supports an upcoming refactor that decouples the ROS 2 node from the classification logic in `autoware_traffic_light_classifier`. It adds a node-level **characterization test suite** for `TrafficLightClassifierNodelet::imageRoiCallback`, plus the CMake wiring to build it, as a safety net for that refactor.
**Test-only change — no runtime/library code is modified (only the test, its CMake wiring, and a test-only dependency).**

### Remarks
- Calls `imageRoiCallback()` directly, bypassing `message_filters` time-sync and the lazy-subscription timer, so the test feeds input to the logic directly rather than through a publisher.
  - This relies on `imageRoiCallback()` being `public`, which is arguably a bug (an internal callback that should be `private`). The test exploits the current visibility; tightening it to `private` later would require a different hook.
- The exact lamp *color* chosen by `ColorClassifier` is its own concern and is intentionally **not** pinned here; only node-level orchestration is characterized.
- Uses the CPU-only HSV backend (`ColorClassifier`) → builds and runs **without CUDA / TensorRT / model files / a GPU**.

### What is characterized (11 cases)
**ROI handling / output shape**
- `EmptyRoisPublishEmptySignalsWithHeader` — empty input → empty signal array, image header copied, no diagnostic (early return).
- `NonMatchingTypeRoiIsDropped` — ROI of a non-target type is dropped.
- `ZeroSizedMatchingRoiAppendedAsUnknown` — zero-sized target ROI appended as UNKNOWN (confidence 0).
- `ValidRoiProducesExactlyOneElement` — valid ROI classified; exactly one element; id/type propagated.
- `MultipleValidRoisMapToOrderedSlots` — multiple valid ROIs map to output slots in input order.
- `ValidThenZeroSizedRoiOrdering` — classified signals lead, appended UNKNOWNs follow (output order ≠ input order).

**Exposure handling + diagnostics**
- `OverExposedRoiOverwrittenWithUnknownAndWarns` — over-exposed ROI → UNKNOWN + WARN diagnostic (`over` flag true).
- `UnderExposedRoiOverwrittenWithUnknownAndWarns` — under-exposed ROI → UNKNOWN + WARN diagnostic (`under` flag true).
- `OnlyExposedSlotIsOverwritten` — only the flagged slot is overwritten; a normally-exposed slot keeps its classifier result.
- `OverAndUnderExposureInSameCall` — both conditions in one call → both slots UNKNOWN, both flags true.

**Defensive guard**
- `UnknownClassifierTypePublishesNothing` — an unknown `classifier_type` leaves the classifier unset and the callback degrades to a silent no-op.

## Related links

**Parent Issue:**
- None.

## How was this PR tested?

Built and tested on ROS 2 Humble (CPU / HSV backend):

```bash
colcon build --packages-select autoware_traffic_light_classifier --cmake-args -DBUILD_TESTING=ON
```

→ 11/11 pass.

## Notes for reviewers
* Test-only; no change to production behavior, topics, or parameters.

## Interface changes
None.

## Effects on system behavior
None. (Test-only; production code is unchanged.)


takam5f2:
@YoshiRi  Could you review the PR?

takam5f2:
@a-maumau  
Sorry for bothering you, but please review the PR after the #12789. Thanks, 
https://github.com/autowarefoundation/autoware_universe/pull/12789

a-maumau (approved):
LGTM!

PR #12843, opened 2026-06-22
feat(diagnostic_graph_aggregator): apply `agnocast_wrapper::Node` to diagnostic_graph_aggregator

Koichi98:
## Description

Apply `agnocast_wrapper::Node` to `autoware_diagnostic_graph_aggregator` (both the `aggregator_node` and `converter_node` executables).
Please read https://github.com/autowarefoundation/autoware/issues/7007 for the context and review guidelines of `autoware_agnocast_wrapper` adoption. Also, please refer to https://github.com/autowarefoundation/autoware_core/blob/main/common/autoware_agnocast_wrapper/docs/review_guide.md for the review guide of `autoware_agnocast_wrapper` application PR. FYI: these nodes take [Method2](https://github.com/autowarefoundation/autoware_core/blob/main/common/autoware_agnocast_wrapper/docs/review_guide.md#method-2-agnocast_wrappernode-inheritance) of the two integration methods.

When built and run with `ENABLE_AGNOCAST=1`, the nodes run on an `AgnocastOnlyCallbackIsolatedExecutor` (CIE), the standard executor for `agnocast_wrapper::Node`. Please check for potential race conditions only if the node meets **all** of the following:

**Conclusion for this PR:** both nodes use only the default callback group (no `create_callback_group`), so condition 2 does not hold. Callbacks remain mutually exclusive even under CIE, so there is no race condition.

## Related Links

`agnocast_wrapper::Node`

## How was this PR tested?
  - Ran LSim with sample rosbag
    - build with `ENABLE_AGNOCAST=0`
    - build with `ENABLE_AGNOCAST=1`, ran with `ENABLE_AGNOCAST=0`
    - build with `ENABLE_AGNOCAST=1`, ran with `ENABLE_AGNOCAST=1`
  - Confirmed both `aggregator_node` and `converter_node` start as standalone executables (not composable nodes), publish their output topics at the same rate in both modes, and that the aggregator's services (e.g. `~/reset`) remain reachable from plain ROS 2 clients via the Agnocast bridge when running with `ENABLE_AGNOCAST=1`.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

When `ENABLE_AGNOCAST=1`, CPU time spent on message serialization/deserialization in these nodes is reduced (Agnocast zero-copy IPC). The executor also switches to CIE, but since CIE is almost equivalent to `MultiThreadedExecutor`, there is effectively no change in scheduling behavior.


system/autoware_diagnostic_graph_aggregator/src/node/aggregator.cpp
```diff
@@ -101,36 +106,46 @@ void AggregatorNode::on_timer()
...
+    pub_unknown_->publish(std::move(unknown_msg));
+  }
 
   // Update plugins.
   if (command_modes_) command_modes_->update(stamp);
   if (driving_modes_) driving_modes_->update(stamp);
 }
 
-void AggregatorNode::on_diag(const DiagnosticArray & msg)
+void AggregatorNode::on_diag(const AUTOWARE_MESSAGE_CONST_SHARED_PTR(DiagnosticArray) & msg)
```
isamu-takagi: Callback using reference types do not use Agnocast-specific pointer, so wouldn't replacement be unnecessary?

atsushi421: Thank you for the review! Unfortunately, keeping the `const DiagnosticArray &` signature does not compile with the wrapper. The wrapper's `create_subscription` only accepts callbacks invocable with `AUTOWARE_MESSAGE_UNIQUE_PTR &&` or `AUTOWARE_MESSAGE_CONST_SHARED_PTR &&`, enforced by `static_assert` on both the Agnocast path and the ROS 2 fallback path ([here](https://github.com/autowarefoundation/autoware_core/blob/2c9f6f6c9b8deb10a34b66d28ddbb8fa503dc872/common/autoware_agnocast_wrapper/include/autoware/agnocast_wrapper/autoware_agnocast_wrapper.hpp#L446-L450) and [here](https://github.com/autowarefoundation/autoware_core/blob/2c9f6f6c9b8deb10a34b66d28ddbb8fa503dc872/common/autoware_agnocast_wrapper/include/autoware/agnocast_wrapper/autoware_agnocast_wrapper.hpp#L483-L487)), and `message_ptr` has no implicit conversion to `const DiagnosticArray &`. So the callback signature needs to be changed regardless of `ENABLE_AGNOCAST`.

isamu-takagi: Isn't it simply that the agnocast wrapper hasn't implemented compatibility with rclcpp? Can it support different callbacks, like in https://github.com/ros2/rclcpp/blob/rolling/rclcpp/include/rclcpp/any_subscription_callback.hpp?

It is generally better to write code that avoids using agnocast wrapper specific syntax wherever possible.

atsushi421: Thank you very much for pointing that out!
I have fixed the agnocast wrapper in https://github.com/autowarefoundation/autoware_core/pull/1228 and incorporated it to this PR in https://github.com/autowarefoundation/autoware_universe/pull/12843/commits/f735978363c57739779531e214f8694da03ce190

isamu-takagi (approved):

PR #1283, opened 2022-07-08
fix(behavior_velocity): fix intersection last point

taikitanaka3:
Signed-off-by: tanaka3 <ttatcoder@outlook.jp>

## Description

fix intersection dies insert stop point at the end
because input path size and *path size is different

I confirmed normal case and intersection module won't die for inserting last stop index point
(Note: test case won't be enough but this won't happen for normal use)

cc: @VRichardJP 

## Related links

resolve
https://github.com/autowarefoundation/autoware.universe/issues/1247

## Tests performed

by psim

https://user-images.githubusercontent.com/65527974/177935298-24dad101-c24b-4708-b79f-ff2a21ba9d5a.mp4

## Notes for reviewers

This PR  has not been tested enough.

## Pre-review checklist for the PR author

The PR author **must** check the checkboxes below when creating the PR.

- [x] I've confirmed the [contribution guidelines].
- [x] The PR follows the [pull request guidelines].

## In-review checklist for the PR reviewers

The PR reviewers **must** check the checkboxes below before approval.

- [ ] The PR follows the [pull request guidelines].
- [ ] The PR has been properly tested.
- [ ] The PR has been reviewed by the code owners.

## Post-review checklist for the PR author

The PR author **must** check the checkboxes below before merging.

- [ ] There are no open discussions or they are tracked via tickets.
- [ ] The PR is ready for merge.

After all checkboxes are checked, anyone who has write access can merge the PR.

[contribution guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/
[pull request guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/pull-request-guidelines/

PR #12844, opened 2026-06-22
feat(mrm_handler): apply `agnocast_wrapper::Node` to `mrm_handler`

Koichi98:
## Description
Apply `agnocast_wrapper::Node` to `autoware_mrm_handler`.
Please read https://github.com/autowarefoundation/autoware/issues/7007 for the context and review guidelines of `autoware_agnocast_wrapper` adoption.  Also, please refer to https://github.com/autowarefoundation/autoware_core/blob/main/common/autoware_agnocast_wrapper/docs/review_guide.md for the review guide of `autoware_agnocast_wrapper` application PR. FYI: This node take [Method2](https://github.com/autowarefoundation/autoware_core/blob/main/common/autoware_agnocast_wrapper/docs/review_guide.md#method-2-agnocast_wrappernode-inheritance) of two integrations methods.

## Related Links

`agnocast_wrapper::Node`

## How was this PR tested?

Verified on 2026-09-03 against `autoware_core` and `autoware_universe` main of the same day, with Agnocast 2.4.0 (userland, kernel module and heaphook all on 2.4.0).

- The generated node main reports `agnocast_only = true` and the node is registered with `AgnocastOnlyCallbackIsolatedExecutor`.
- Standalone startup succeeds with both `ENABLE_AGNOCAST=0` and `ENABLE_AGNOCAST=1`, with no `process has died` and no error log.
- Logging Simulator with the sample rosbag (`-r 0.2`) completes with both `ENABLE_AGNOCAST=0` and `ENABLE_AGNOCAST=1`. No `process has died` and no error while the bag is playing, and the localization output rates match between the two modes (NDT pose 1.905 Hz on both, which is the expected 10 Hz scaled by the 0.2 playback rate).
- The service client path was exercised on purpose, because `mrm_handler` is the only node in this series that calls services. With `operation_mode_availability.autonomous` flipped to `false`, the node requested `EMERGENCY_STOP` and received the response within the 10 ms `timeout_call_mrm_behavior`, in both modes: `NORMAL -> MRM_OPERATING`, `EMERGENCY_STOP is operated.`, `MRM_OPERATING -> MRM_SUCCEEDED`.
- No `rclcpp::ok()`, `rclcpp::shutdown()` or `rclcpp::Rate::sleep()` remains in this package. The single `rclcpp::Rate` use only calls `period()`, which does not depend on the rclcpp context.

## Notes for reviewers
Two commits in this PR are not part of the Agnocast conversion itself. They bring the branch up to the wrapper API as it stands today, because the conversion was written in July.

- `migrate to the current agnocast_wrapper polling API`: `AUTOWARE_POLLING_SUBSCRIBER_PTR` and the `Node` member `create_polling_subscriber()` were removed in autoware_core #1283, so the seven polling subscribers are now declared as `polling::PollingSubscriber<T>::SharedPtr` and obtained from the free function `polling::create_polling_subscriber<T>(this, topic)`, the same way `shift_decider` and `collision_detector` use the wrapper on main.
- `pass rclcpp::QoS to the wrapper's create_client`: `agnocast_wrapper::Node::create_client()` takes `const rclcpp::QoS &`, while `AUTOWARE_DEFAULT_SERVICES_QOS_PROFILE()` expands to an `rmw_qos_profile_t` on Humble, which is what `rclcpp::Node::create_client()` takes. The three MRM operation clients now pass `rclcpp::ServicesQoS()` directly and the unused `autoware_qos_utils` dependency is dropped. This keeps working on Jazzy: the wrapper signature does not depend on the distribution, and the conversion Humble needs is done inside the wrapper.
## Interface changes

None.

## Effects on system behavior

When `ENABLE_AGNOCAST=1`, CPU time spent on message serialization/deserialization in this node is reduced (Agnocast zero-copy IPC). The executor also switches to CIE, but since CIE is almost equivalent to `MultiThreadedExecutor`, there is effectively no change in scheduling behavior.


system/autoware_mrm_handler/launch/mrm_handler.launch.xml
```diff
@@ -1,4 +1,6 @@
 <launch>
+  <include file="$(find-pkg-share autoware_agnocast_wrapper)/launch/agnocast_env.launch.xml"/>
```
sasakisasaki, line 3: Do we need to add this line for all packages that uses agnocast?

system/autoware_mrm_handler/launch/mrm_handler.launch.xml
```diff
@@ -22,6 +24,7 @@
 
   <!-- mrm_handler -->
   <node pkg="autoware_mrm_handler" exec="autoware_mrm_handler_node" name="mrm_handler" output="screen">
+    <env name="LD_PRELOAD" value="$(var ld_preload_value)"/>
```
sasakisasaki, line 29: Perhaps developer cannot understand for which this `LD_PRELOAD` variable is needed. Thus,
```suggestion
    <!-- For using agnocast wrapper -->
    <env name="LD_PRELOAD" value="$(var ld_preload_value)"/>

```

system/autoware_mrm_handler/src/mrm_handler/mrm_handler_core.cpp
```diff
@@ -64,18 +63,16 @@ MrmHandler::MrmHandler(const rclcpp::NodeOptions & options) : Node("mrm_handler"
   // Clients
   client_mrm_pull_over_group_ = create_callback_group(rclcpp::CallbackGroupType::MutuallyExclusive);
   client_mrm_pull_over_ = create_client<tier4_system_msgs::srv::OperateMrm>(
-    "~/output/mrm/pull_over/operate", AUTOWARE_DEFAULT_SERVICES_QOS_PROFILE(),
-    client_mrm_pull_over_group_);
+    "~/output/mrm/pull_over/operate", rclcpp::ServicesQoS(), client_mrm_pull_over_group_);
```
sasakisasaki, line 66: (Note to myself) This fix is needed because:

> This keeps working on Jazzy: the wrapper signature does not depend on the distribution, and the conversion Humble needs is done inside the wrapper.

sasakisasaki (commented):
Minor comment first 👍

system/autoware_mrm_handler/launch/mrm_handler.launch.xml
kobayu858: Yes — a node converted with Method 2 (`agnocast_wrapper::Node`) needs these two lines in its own launch file.

- `agnocast_env.launch.xml` provides `ld_preload_value` (plus the container package/executable variables), and spawns the per-IPC-namespace discovery agent when Agnocast is enabled.
- `LD_PRELOAD` then applies that value to this node only.

This is the shape used by the packages converted so far, for example `autoware_trajectory_processor` (#13242):

```xml
  <include file="$(find-pkg-share autoware_agnocast_wrapper)/launch/agnocast_env.launch.xml"/>

  <node pkg="autoware_trajectory_processor" exec="autoware_trajectory_processor_node" name="trajectory_processor" output="screen">
    <env name="LD_PRELOAD" value="$(var ld_preload_value)"/>
```

and autoware_external_cmd_converter (#13014), in Python:
```python
agnocast_env = IncludeLaunchDescription(... "agnocast_env.launch.py" ...)
node = Node(..., additional_env={"LD_PRELOAD": LaunchConfiguration("ld_preload_value")})
return LaunchDescription(arguments + [agnocast_env, node])
```

system/autoware_mrm_handler/launch/mrm_handler.launch.xml
kobayu858: Applied, thank you. Added  above the LD_PRELOAD line as suggested. https://github.com/autowarefoundation/autoware_universe/pull/12844/commits/5af01f64551e2fdd5933d1083eaa89f13694de68

system/autoware_mrm_handler/launch/mrm_handler.launch.xml
kobayu858: For the full description of `agnocast_env` (provided variables, launch arguments, and XML/Python examples), see [How to Enable Agnocast at Runtime](https://github.com/autowarefoundation/autoware_core/blob/main/common/autoware_agnocast_wrapper/README.md#how-to-enable-agnocast-at-runtime) in the wrapper README.

sasakisasaki, line 3: Thank you for your clear explanation! Now I understand 👍 Finally I like to add for readability:
```suggestion
  <!-- Read `ld_preload_value` -->
  <include file="$(find-pkg-share autoware_agnocast_wrapper)/launch/agnocast_env.launch.xml"/>
```

sasakisasaki (approved):
@kobayu858 Reviewed and I understood the files changed. Thus I agree with merging this PR while put minor comment 👍

system/autoware_mrm_handler/launch/mrm_handler.launch.xml
kobayu858: Applied in https://github.com/autowarefoundation/autoware_universe/pull/12844/commits/3232539bc0752ccb6a7a9cd1aa4978e528c14d9c, thank you 🙏

PR #12851, opened 2026-06-23
refactor(autoware_traffic_light_classifier): extract classification logic from ROS node

takam5f2:
## Description

Extract the classification orchestration out of the ROS node into a ROS-free class.

`TrafficLightClassifierNodelet::imageRoiCallback` previously contained the whole per-ROI pipeline (filter by traffic-light type, exposure detection, crop & classify, append undetected zero-sized ROIs as `UNKNOWN`, overwrite out-of-range exposure slots with `UNKNOWN`). This logic is moved verbatim into a new `TrafficLightClassifier` class (`traffic_light_classifier.{hpp,cpp}`) that owns the classifier backend and the parameters, and exposes a single `classify(image, rois) -> std::optional<Result>`.

The node is now a thin adapter responsible only for I/O: parameter declaration, pub/sub & time-sync, `cv_bridge` conversion, header stamping, and diagnostics.

This is a pure, behavior-preserving extraction — no functional change.

## Related links

**Parent Issue:**

- Follow-up to #12840 (characterization test that decouples the node and the logic).

## How was this PR tested?

- `colcon build --packages-select autoware_traffic_light_classifier` (color-classifier / non-GPU path) — succeeds.
- `colcon test --packages-select autoware_traffic_light_classifier` — green:
  - `test_traffic_light_classifier_characteristics`: 11/11 passed (the safety net from #12840, unchanged).
  - `test_utils`: 13/13 passed.

The characterization test drives the node end-to-end and pins the observable behavior (filtering, ordering, UNKNOWN handling, exposure overwrite, id/type propagation, diagnostics), so a green run demonstrates the extraction preserved behavior.

## Notes for reviewers

- Behavior-preserving refactor only; the diff is intentionally limited to the extraction.
- The now-unused `traffic_light_classifier_process.hpp` include was dropped from the node header (the node no longer touches `utils::compute_brightness`).
- **Node-free unit tests for `TrafficLightClassifier::classify()` are planned as a separate follow-up PR**, to keep this one focused and reduce review cost. The new class is ROS-free and returns a value, so it can be tested directly without the rclcpp executor harness.
- A pre-existing latent issue (a failed `cv_bridge` conversion does not early-return, leading to a null deref) is intentionally left unchanged here — it is out of scope for a behavior-preserving extraction.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_traffic_light_classifier/CMakeLists.txt
+++ b/perception/autoware_traffic_light_classifier/CMakeLists.txt
@@ -60,6 +60,7 @@ if(TRT_AVAIL AND CUDA_AVAIL)
     src/classifier/color_classifier.cpp
     src/classifier/cnn_classifier.cpp
     src/classifier/cnn_lamp_recognizer.cpp
+    src/traffic_light_classifier.cpp
     src/traffic_light_classifier_node.cpp
     src/traffic_light_classifier_process.cpp
   )
@@ -81,6 +82,7 @@ if(TRT_AVAIL AND CUDA_AVAIL)
     src/classifier/cnn_classifier.cpp
     src/classifier/color_classifier.cpp
     src/classifier/cnn_lamp_recognizer.cpp
+    src/traffic_light_classifier.cpp
     src/traffic_light_classifier_node.cpp
     src/traffic_light_classifier_process.cpp
     src/single_image_debug_inference_node.cpp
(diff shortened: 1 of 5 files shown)


takam5f2:
@a-maumau @YoshiRi  

Sorry for bothering you.

This PR is essentially a pure code move: the per-ROI classification logic in `TrafficLightClassifierNodelet::imageRoiCallback` is lifted into a ROS-free `TrafficLightClassifier` class, leaving the node as a thin I/O adapter. No behavior change — the characterization test from #12840 stays green (build + tests all pass locally), so reviewing just means checking the moved logic matches the original. Node-free unit tests will follow in a separate PR.

a-maumau (approved):
LGTM

PR #12853, opened 2026-06-23
feat(trajectory_processor): add tests for the modifier+optimizer pipeline

maxime-clem:
## Description

This PR prepares the merging of the `autoware_trajectory_modifier` and `autoware_trajectory_optimizer` packages.
A new package `autoware_trajectory_processor` is added which only contains unit tests for now. These tests will be used to validate that the logic is unchanged after merging the nodes.

## Related links

**Parent Issue:**

- https://tier4.atlassian.net/browse/PDDP-240

## How was this PR tested?

Unit tests only

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/autoware_trajectory_processor/tests/test_trajectory_processor_integration.cpp
+++ b/planning/autoware_trajectory_processor/tests/test_trajectory_processor_integration.cpp
@@ -241,7 +241,7 @@ protected:
     optimizer_options.append_parameter_override(
       "trajectory_velocity_optimizer.limit_lateral_acceleration", false);
     optimizer_options.append_parameter_override(
-      "trajectory_velocity_optimizer.smooth_velocities", false);
+      "trajectory_velocity_optimizer.smooth_velocities", true);
     optimizer_options.append_parameter_override(
       "trajectory_velocity_optimizer.continuous_jerk_smoother.jerk_weight", 30.0);
     optimizer_options.append_parameter_override(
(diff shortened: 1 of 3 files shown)


planning/autoware_trajectory_processor/tests/test_trajectory_processor_integration.cpp
```diff
@@ -0,0 +1,649 @@
...
+  odom.twist.twist.linear.x = 5.0;
+
+  AccelWithCovarianceStamped acc;
+  acc.header.frame_id = "map";
+
+  // Publish in a loop to ensure the nodes receive it after discovery
+  for (int i = 0; i < 20 && !output_received_; ++i) {
+    odom.header.stamp = sim_time_;
+    acc.header.stamp = sim_time_;
+    publish_mandatory_inputs(traj, odom, acc);
```
interimadd: Setting a timestamp for each message in every iteration seems redundant.
What do you think about handling it inside a private helper function in the test fixture, like publish_mandatory_inputs()?

```cpp
void publish_mandatory_inputs(
  const CandidateTrajectories & traj, Odometry odom, AccelWithCovarianceStamped acc)
{
  tra.header.stamp = sim_time_;
  odom.header.stamp = sim_time_;
  acc.header.stamp = sim_time_;
  pub_tra_->publish(traj);
  pub_odo_->publish(odom);
  pub_acc_->publish(acc);
}
```

planning/autoware_trajectory_processor/tests/test_trajectory_processor_integration.cpp
```diff
@@ -0,0 +1,649 @@
...
+
+TEST_F(TrajectoryProcessorIntegrationTest, BasicPipelineTest)
+{
+  auto traj = create_straight_trajectories(30.0, 5.0, sim_time_);
+
+  Odometry odom;
+  odom.header.frame_id = "map";
+  odom.pose.pose.position.x = 0.0;
+  odom.pose.pose.orientation.w = 1.0;
+  odom.twist.twist.linear.x = 5.0;
```
interimadd: odom is created for each test case, but the only difference is the velocity.
I suggest creating odom using a helper function like `create_odometry(double velocity)`.

interimadd (commented):
Thank you for assigning me as a reviewer!
I've left some comments to improve the readability and maintainability of each test.

I'd like to know whether this integration test is meant to be temporary or permanent.
I'm a bit concerned that this test will be hard to keep green as the node evolves, since it exercises many plugins at once, and this test would need to be updated every time any plugin's behavior changes.

In my opinion, if the node's code are as simple as this:
```cpp
void on_traj(const TrajectoryMessage msg) {
    auto modified_traj = obstacle_stop_->modify(msg);
    modified_traj = stop_point_fixer_->modify(modified_traj);
    modified_traj = velocity_optimizer_->modify(modified_traj);
                         :
    publisher_->publish(msg);
}
```
and each modify() function are well tested, then a node-level integration test would only need a simple case like `BasicPipelineTest` in this PR.
Thanks to your team's effort, each plugin already has its own tests, so this integration test may not need to re-test each plugin's individual behavior again.

However, the new node likely won't be simple as example above, so these tests may still be worth keeping as characterization tests to preserve the current behavior during the merge.

In conclusion, I'm fine with confirming the new node's behavior with these tests in this PR, but as future work, I'd suggest moving the node implementation closer to the example above, so this logic can eventually be unit-tested without a ROS2 runtime.

planning/autoware_trajectory_processor/tests/test_trajectory_processor_integration.cpp
```diff
@@ -0,0 +1,649 @@
...
+    pub_obj_->publish(objects);
+    publish_mandatory_inputs(traj, odom, acc);
+    advance_sim_time_and_spin(std::chrono::milliseconds(100));
+  }
+  wait_for_output_sim();
+
+  ASSERT_TRUE(output_received_);
+  // The trajectory should have a stop point (velocity close to 0) before the object at x=20.0
+  bool found_stop = false;
+  for (const auto & p : latest_output_.points) {
```
mkquda: Technically you don't have to search for the stop point, it should be the last point in the trajectory, so you should check the velocity and position of `latest_output_.points.back()`

planning/autoware_trajectory_processor/tests/test_trajectory_processor_integration.cpp
maxime-clem: Thank you. Done: [5b18198](https://github.com/autowarefoundation/autoware_universe/pull/12853/commits/5b18198b433d7ea3727fbb33fe479dd84a6d27c1)

planning/autoware_trajectory_processor/tests/test_trajectory_processor_integration.cpp
maxime-clem: Thank you. Done: [5b18198](https://github.com/autowarefoundation/autoware_universe/pull/12853/commits/5b18198b433d7ea3727fbb33fe479dd84a6d27c1)

planning/autoware_trajectory_processor/tests/test_trajectory_processor_integration.cpp
maxime-clem: Thank you. Done: [2abc491](https://github.com/autowarefoundation/autoware_universe/pull/12853/commits/2abc4918bc095f6b593965416ddf3f4cd83d966f)

maxime-clem:
@interimadd thank you very much for your review.

> I'd like to know whether this integration test is meant to be temporary or permanent. I'm a bit concerned that this test will be hard to keep green as the node evolves, since it exercises many plugins at once, and this test would need to be updated every time any plugin's behavior changes.

I will do my best to follow your suggestion in the new node to use a ROS-independent interface to make the testing easier.
If some tests are difficult to maintain, I think it is fine to remove them after the initial refactoring is completed.

interimadd (approved):
Thanks for addressing the feedback!
This test will really help with refactoring.

PR #12872, opened 2026-06-25
refactor(autoware_pointcloud_preprocessor): extract matching policy and rename collector matcher

vividf:
## Description

Groundwork for making the pointcloud concatenation core reusable outside the ROS
runtime. This PR isolates the cloud-to-collector matching logic; no behavior changes.

- Extract the pure naive / advanced matching logic into a ROS-runtime-free
  `MatchingPolicy` that operates on plain structs (`IncomingCloudInfo`,
  `CandidateCollectorState`, `CollectorReference`) instead of ROS-side objects.
- Rename `CollectorMatchingStrategy` → `CollectorMatcher`; the matcher classes
  become thin wrappers that adapt the ROS-side collectors to the policy and hold
  the node / logging dependency.
- Rename `MatchingParams` → `IncomingCloudInfo` at the node call sites.
- Rename the CUDA package's matcher in lockstep (it shares the same header).

This is the first of a stack that ends in Python bindings for offline concatenation.

## Related links
* Next PR: https://github.com/autowarefoundation/autoware_universe/pull/12873

**Parent Issue:**

- Link

## How was this PR tested?

- `colcon build --packages-up-to autoware_pointcloud_preprocessor autoware_cuda_pointcloud_preprocessor`
- `colcon test --packages-select autoware_pointcloud_preprocessor` — includes the new
  `test_matching_policy` plus the existing concatenate node tests. Pure refactor, so
  existing behavior is covered by the current test suite.

## Notes for reviewers

Pure rename + extraction, no behavioral change. Reviewing commit-by-commit is easiest.
The CUDA matcher rename is mechanical (shared header).

## Interface changes

None. (Internal C++ types only; no ROS topics, parameters, or public API.)

## Effects on system behavior

None.


sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/concatenate_data/matching_policy.hpp
```diff
@@ -0,0 +1,95 @@
...
+// Candidate collectors needed for matching.
+struct CandidateCollectorState
+{
+  double reference_time;  // the collector's reference time (Naive/Advanced CollectorInfo timestamp)
+  double noise_window;    // the collector's noise window (advanced only; 0 for naive)
+  bool has_topic;         // whether the collector already holds IncomingCloudInfo.topic_name
+};
+
+// The reference a freshly-created collector should be given for an incoming cloud.
+struct CollectorReference
```
mojomex: The name reads like this would be a "reference to a collector".
A name like "MatchingReference" might be clearer.

mojomex (approved):
Nice! One small naming feedback but otherwise great!
Feel free to merge without re-review after renaming :D

jacoblambert (approved):

sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/concatenate_data/matching_policy.hpp
vividf: Thanks, fixed in https://github.com/autowarefoundation/autoware_universe/pull/12872/commits/ccbc913f3f98241fd4901cc0e657c93144f198f3

PR #12873, opened 2026-06-25
refactor(autoware_pointcloud_preprocessor): make the concatenation core ROS-runtime-free

vividf:
## Description

Make the point cloud concatenation core (`CombineCloudHandler`, `CombineCloudHandlerBase`,
the concatenation diagnostics) independent of the ROS runtime, so it can be driven from a
non-node context (the offline/Python pipeline in #13113 / #12874) and unit-tested without
spinning a node. The node keeps exactly the same topics, parameters and outputs.

This PR is the merged result of a 4-part stack that was reviewed step by step on the fork
(links under *Related links*); each step is still a separate commit here.

1. **Type the matching strategy** — `MatchingStrategyType` enum pinned to the
   `ConcatenatedPointCloudInfo` `STRATEGY_*` constants; the string is parsed once, at the
   parameter boundary (`parse_matching_strategy()`); `matching_strategy_name_map` is gone.
2. **Key the handler on the message type** — `CombineCloudHandler<sensor_msgs::msg::PointCloud2>`
   / `CombineCloudHandler<cuda_blackboard::CudaPointCloud2>` instead of the `*Traits` tag types.
3. **Drop pcl and rclcpp time from the core** — stamps are `builtin_interfaces::msg::Time`
   with a small `time_utils.hpp` (`is_earlier`, `add`, `subtract`, `to_seconds`);
   `pcl_ros::transformPointCloud` / `pcl::concatenatePointCloud` are replaced by an Eigen
   transform loop and a byte append on the shared XYZIRC layout.
4. **Inject transforms instead of holding a node** — the handler no longer owns a
   `ManagedTransformBuffer`; the node looks up sensor→output extrinsics and calls
   `set_transform()`. A source whose frame has no transform is **dropped and reported**
   (`dropped_sources_missing_transform`, `STATUS_INVALID`) instead of being placed with an
   identity transform. Motion-compensation warnings became a `MotionCompensationStatus` the
   node logs.
5. **Build the diagnostics outside the node** — `build_diagnostic_status()` in
   `concatenation_diagnostics.{hpp,cpp}` produces the `DiagnosticStatus` from a plain summary
   struct; the node only feeds it and publishes. A source without a transform now counts as
   *not concatenated* in the diagnostics, consistent with `ConcatenatedPointCloudInfo`.

A new `concatenate_core` library target contains only the ROS-runtime-free sources and links
message targets only, to keep it that way.

## Related links

- Previous PR (merged): #12872 — extract matching policy and rename collector matcher
- Reviewed as a stack on the fork, now merged into this branch:
  vividf/autoware.universe#4, #5, #6, #7
- Follow-ups that build on this core: #13113 (`feat/concat_core`), #12874 (Python bindings)

## How was this PR tested?

- `colcon build/test --packages-select autoware_pointcloud_preprocessor autoware_cuda_pointcloud_preprocessor`
  — 519 tests, 0 failures (incl. lint).
  - `test_concatenation_info`: typed strategy.
  - `test_concatenation_diagnostics` (new, 11 cases): key order, level/message rules, missing
    topic, missing transform, late drop, advanced window.
  - `test_cuda_combine_cloud_handler` (new, 2 cases, runs on the GPU and self-skips without one):
    concatenation on device, drop of a source without an extrinsic.
  - `test_concatenate_node_unit`: transforms injected via `set_transform()` instead of a TF broadcaster.
- On-hardware A/B against `main` (same base commit, x86 + RTX, 3 lidars, motion compensation on,
  ~120 frames): output clouds equivalent (max |Δxyz| 23 µm, float32 rounding; identity-rotation
  sensor byte-identical); `ConcatenatedPointCloudInfo` identical; `/diagnostics` level identical on
  every frame; processing time −8 % median.

## Notes for reviewers

- Review commit by commit if the whole diff is too much; the five steps above map to the commits.
- Unknown `matching_strategy.type` now throws `std::invalid_argument` from `parse_matching_strategy()`
  instead of the node's `std::runtime_error`. Both abort startup.
- `transform_pointcloud()` applies `Matrix4f * Vector4f` where pcl used `linear()*v + translation()`;
  results differ only by float32 rounding (≤ 1–8 ULP, zero-mean).

## Interface changes

None. No topic, parameter or message changes. Internal C++ API of the handler changed
(constructor no longer takes a `rclcpp::Node`; `set_transform()` added).

## Effects on system behavior

- **Nominal operation**: none observable (verified as above).
- **Missing TF for a source**: previously the source was concatenated with an identity transform
  and `source_info.status` stayed `OK`. Now its points are dropped, `source_info.status = INVALID`,
  `concatenation_success = false`, and `/diagnostics` reports `ERROR` with
  `Concatenated: <topic> = False` plus a new key `Frames without a transform to the output frame`.
- One diagnostics message string changed: "dropped due to missing topics and its timestamp is
  earlier" → "... and **because** its timestamp is earlier". Exact-match monitoring rules need
  updating.
- Diagnostics timestamps are rendered from `sec + nanosec*1e-9` (single rounding) instead of
  `rclcpp::Time::seconds()`; values may differ in the last digits.
- Misconfigured `matching_strategy.type`: different exception type/message, still aborts startup.


mojomex:
@codex review

sensing/autoware_pointcloud_preprocessor/test/test_concatenation_info.cpp
```diff
@@ -83,7 +83,7 @@ class ConcatenationInfoTest : public ::testing::Test
   }
 
   std::vector<std::string> input_topics_;
-  std::string strategy_name_;
+  autoware::pointcloud_preprocessor::MatchingStrategyType strategy_{};
```
mojomex: This will implicitly initialize to `0` which may or may not hit a valid (`naive`/`advanced`) enum value. Better explicitly initialize to a default value, e.g. `advanced`.

mojomex (commented):
Code looks good, one minor comment only!

sensing/autoware_pointcloud_preprocessor/test/test_concatenation_info.cpp
vividf: Good catch, thanks! Initialized it to `advanced` explicitly (and dropped the now-duplicate assignment in `SetUp()`): 919782a

mojomex (commented):
LGTM!

Please wait with merging until the stack has been merged into this PR.

mojomex:
LGTM. Let's wait with merging until @sasakisasaki's PR #13379 is in place to ensure no behavior is accidentally changed.

PR #12874, opened 2026-06-25
feat(autoware_pointcloud_preprocessor): add Python bindings for the concatenation core

vividf:
## Description

Depends on https://github.com/autowarefoundation/autoware_universe/pull/12872,
https://github.com/autowarefoundation/autoware_universe/pull/12873 and https://github.com/autowarefoundation/autoware_universe/pull/13113.

Expose `CloudConcatenator` to Python so recorded data can be re-concatenated offline with the
*same* C++ core the node runs — no reimplementation, and no wall clock involved.

- Wrap the ROS-free `concatenate_core` library (combine handler + base, diagnostics builder,
  concatenator, matching policy) with pybind11 as `_concatenate_pointclouds_pybind`. Only the
  pybind module links `rclcpp`, for message (de)serialization. ROS messages cross the boundary
  as CDR-serialized bytes, so the binding needs no message-conversion machinery.
- `concatenate_pointclouds.py` is a **thin shell**: it converts between rclpy message objects
  and those bytes and exposes `Concatenator` (`process_cloud` / `close_expired_collectors` /
  `flush`, plus `process_twist` / `process_odometry` / `set_transform`) and
  `build_diagnostics()`. All concatenation logic — matching, collector lifecycle, timeout,
  combining and diagnostics — stays in C++, so offline output matches the node's.
- Because the caller supplies each cloud's arrival time (e.g. the rosbag record stamp), offline
  runs are deterministic and reproducible.
- `build_diagnostics()` produces the node-equivalent `DiagnosticStatus` for an emitted frame,
  including the optional runtime entries (`processing_time_ms`, a publish-time `now`), so
  offline diagnostics can be compared against a bag's `/diagnostics`.

## Related links

* Previous PR 1: https://github.com/autowarefoundation/autoware_universe/pull/12872
* Previous PR 2: https://github.com/autowarefoundation/autoware_universe/pull/12873
* Previous PR 3: https://github.com/autowarefoundation/autoware_universe/pull/13113

**Parent Issue:**

- Link

## How was this PR tested?

- `colcon build --packages-select autoware_pointcloud_preprocessor autoware_cuda_pointcloud_preprocessor`
- `colcon test --packages-select autoware_pointcloud_preprocessor autoware_cuda_pointcloud_preprocessor`
  → **453 tests / 0 failures** (`autoware_pointcloud_preprocessor`) and
  **89 tests / 0 failures** (`autoware_cuda_pointcloud_preprocessor`).
- New `test_concatenate_pointclouds_py.py` (24 cases) covering, through the public Python API:
  the single-collector combine (field layout, motion compensation on/off, synchronized
  per-sensor clouds, missing-transform drops), the collector lifecycle (complete vs timeout,
  in/out-of-window advanced matching, arrival-time vs header-stamp timeout,
  `close_expired_collectors` with no cloud arriving, `flush`), and diagnostics parity with the
  node's entries/level/message.

## Notes for reviewers

- `libconcatenate_core.so` links only ROS message types + header-only math (no `rclcpp` /
  `tf2` / `pcl`); `ldd` on the built library is the quickest way to confirm it.
- The Python layer holds **no** concatenation logic — if you find a rule implemented there,
  that is a bug. Its only jobs are `serialize_message` / `deserialize_message` and packaging
  the C++ frame into a dataclass.
- The `Concatenator` methods mirror the C++ API one-to-one on purpose, so the offline driver
  reads like the node's callbacks.

## Usage Example

```python
from autoware_pointcloud_preprocessor.concatenate_pointclouds import (
    Concatenator,
    FrameStatus,
)

# tf_static: {sensor_frame -> geometry_msgs/TransformStamped}, each with
#   header.frame_id = output_frame, child_frame_id = sensor_frame
# (build these by composing the /tf_static chain from the bag)

concat = Concatenator(
    input_topics=lidar_topics,           # list[str], order matters for advanced offsets/noise
    output_frame="base_link",
    tf_static=tf_static,                 # dict[str, TransformStamped]
    timeout_sec=0.2,                     # per-collector timeout, in arrival time
    is_motion_compensated=True,
    matching_strategy="advanced",        # or "naive"
    lidar_timestamp_offsets=offsets,     # advanced only: one per input topic (seconds), same order
    lidar_timestamp_noise_window=noise,  # advanced only: one per input topic (seconds), same order
)

# Feed messages in ARRIVAL order. arrival_time is in seconds (e.g. bag record ts * 1e-9).
for topic, arrival_time_sec, msg in stream:        # your iterator, in record order
    if topic == twist_topic:
        concat.process_twist(msg)                  # TwistWithCovarianceStamped
        continue
    # msg is a sensor_msgs/PointCloud2 (rclpy object)
    for frame in concat.process_cloud(topic, msg, arrival_time=arrival_time_sec):
        cloud = frame.result.concatenated_cloud    # sensor_msgs/PointCloud2 (or None)
        if frame.status == FrameStatus.COMPLETE:
            ...  # all input topics contributed
        elif frame.status == FrameStatus.TIMEOUT:
            ...  # the collector was closed before all topics arrived
        # frame.result also carries: concatenation_info, no_twist_available,
        # twist_time_gap_too_large, topic_to_original_stamp, dropped_frames_missing_transform

        # The node-equivalent /diagnostics status for this frame:
        status = concat.build_diagnostics(frame, processing_time_ms=None, now=arrival_time_sec)

# Advance the clock when the stream has a gap, without adding a cloud:
for frame in concat.close_expired_collectors(now=arrival_time_sec):
    ...

# Once the stream is exhausted, flush any still-open collectors (emitted as TIMEOUT):
for frame in concat.flush():
    ...
```

## Interface changes

None. (New Python module; no ROS topic or parameter changes.)

## Effects on system behavior

None. The ROS node is untouched by this PR; it only adds a library target, a Python extension
module and the Python package.


sensing/autoware_pointcloud_preprocessor/python/autoware_pointcloud_preprocessor/concatenate_pointclouds.py
```diff
@@ -0,0 +1,500 @@
...
+    # outside the Concatenator). Used by build_diagnostics to reproduce the node's per-collector
+    # window fields: for ``advanced``, ``reference_time +/- noise_window`` is the matching window;
+    # ``first_arrival_time`` is the arrival time of the cloud that opened the collector (the node's
+    # naive "First pointcloud arrival timestamp").
+    reference_time: Optional[float] = None
+    noise_window: Optional[float] = None
+    first_arrival_time: Optional[float] = None
+
+
+class CombineCloudHandler:
```
mojomex: This class is not part of the public Python API, and thus should not be in the Python bindings at all - it should purely live on the C++ side.

sensing/autoware_pointcloud_preprocessor/python/autoware_pointcloud_preprocessor/concatenate_pointclouds.py
```diff
@@ -0,0 +1,500 @@
...
+        for transform in tf_static.values():
+            self._handler.set_transform(transform)
+
+        self._input_topics = list(input_topics)
+        self._timeout_sec = timeout_sec
+        self._collectors: List[_Collector] = []
+        self._is_advanced = matching_strategy == "advanced"
+
+        if matching_strategy == "advanced":
+            if lidar_timestamp_offsets is None or lidar_timestamp_noise_window is None:
```
mojomex: This logic should apply to both C++ and Python, so live entirely on the C++ side.

sensing/autoware_pointcloud_preprocessor/python/autoware_pointcloud_preprocessor/concatenate_pointclouds.py
```diff
@@ -0,0 +1,500 @@
...
+            )
+            self._collectors.append(collector)
+
+        # (4) Add the cloud; emit immediately if the group is now complete.
+        collector.topic_to_cloud[topic] = cloud
+        if len(collector.topic_to_cloud) == len(self._input_topics):
+            self._collectors.remove(collector)
+            outputs.append(self._emit(collector, CollectorStatus.COMPLETE))
+
+        return outputs
```
mojomex: This logic is common to the Python and C++ libs, so shall entierly live in C++.

sensing/autoware_pointcloud_preprocessor/python/autoware_pointcloud_preprocessor/concatenate_pointclouds.py
```diff
@@ -0,0 +1,500 @@
...
+        forwarded unchanged.
+        """
+        return build_diagnostics(frame, self._input_topics, **kwargs)
+
+
+# Default name used for the offline DiagnosticStatus (the online node uses its fully-qualified name).
+_DEFAULT_DIAGNOSTIC_NODE_NAME = "concatenate_data_synchronizer"
+
+
+def build_diagnostics(
```
mojomex: Building diagnostics is already done in C++, so should not be re-implemented in Python.

mojomex (changes requested):
I'm reviewing this first, since the requirements of the Python interface will shape the requirements to the prerequisite PRs.

Currently, the Python wrapper has to duplicate a bunch of behavior that the ROS node already does, hinting that the node-less C++ library split is not optimal yet. Please check if common functionality can be moved to the C++ library instead.

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_concatenate_data/cuda_combine_cloud_handler.cpp
```diff
@@ -228,13 +238,13 @@ CombineCloudHandler<CudaPointCloud2Traits>::combine_pointclouds(
...
-          cloud->header.frame_id, output_frame_, node_.now(), rclcpp::Duration::from_seconds(1.0),
-          node_.get_logger());
+        // Map the (output-frame) cloud back into the sensor frame: inverse of the injected
+        // sensor->output extrinsic.
+        Eigen::Matrix4f transform = Eigen::Matrix4f::Identity();
+        auto transform_opt = get_transform_to_output_frame(cloud->header.frame_id);
 
         if (transform_opt.has_value()) {
-          transform = *transform_opt;
+          transform = transform_opt->inverse();
```
sasakisasaki, line 248: (Note to myself) I need to understand why inverse is used here

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_concatenate_data/cuda_combine_cloud_handler.cpp
```diff
@@ -228,13 +238,13 @@ CombineCloudHandler<CudaPointCloud2Traits>::combine_pointclouds(
...
 
       if (keep_input_frame_in_synchronized_pointcloud_ && need_transform_to_sensor_frame) {
-        Eigen::Matrix4f transform;
-        auto transform_opt = managed_tf_buffer_->getTransform<Eigen::Matrix4f>(
-          cloud->header.frame_id, output_frame_, node_.now(), rclcpp::Duration::from_seconds(1.0),
-          node_.get_logger());
+        // Map the (output-frame) cloud back into the sensor frame: inverse of the injected
+        // sensor->output extrinsic.
+        Eigen::Matrix4f transform = Eigen::Matrix4f::Identity();
+        auto transform_opt = get_transform_to_output_frame(cloud->header.frame_id);
```
sasakisasaki, line 245: Perhaps we can suppress doubled `get_transform_to_output_frame` calculation as this calulation is already performed at [this early stage](https://github.com/autowarefoundation/autoware_universe/pull/12874/changes#diff-d6cb6e6cb50e36cb9ce2d9ae072b55728852243f5287f3ed52d1fbe5b90641edR217).

sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/concatenate_data/cloud_concatenator.ipp
```diff
@@ -0,0 +1,240 @@
...
+  // (3) No match -> start a new collector (matching window from the policy-provided reference,
+  // timeout starting at this cloud's arrival). When the collector limit is reached, make room first
+  // by discarding the collector with the oldest reference (mirrors the node's fixed collector
+  // pool).
+  std::size_t index = 0;
+  if (matched_index.has_value()) {
+    index = *matched_index;
+  } else {
+    if (max_open_collectors_ != 0 && collectors_.size() >= max_open_collectors_) {
+      drop_oldest_collector();
```
sasakisasaki, line 145: If I understand correctly (`collectors_.size()` must be at most `max_open_collectors_`).

Perhaps we need to run this `drop_oldest_collector` multiple times while `collectors_.size() >= max_open_collectors_` because this `drop_oldest_collector` only drops one oldest collector.

I guess something like:
```suggestion
      // Drop until getting one new available room
      while (collectors_.size() >= max_open_collectors_) {
        // Drop one oldest collector
        drop_oldest_collector();
      }
```

sasakisasaki (commented):
Quick comment first 👍 I'll continue my side review until I finish reviewing the entire files changed!

sasakisasaki:
My review approach is as follows for now ✌️

・Do code level review
・Do architecture (interface, ... etc) level retiew
・Do tests on my local

sasakisasaki:
Now reviewing, ...

sasakisasaki:
@vividf @mojomex (CC: @interimadd )

## Review Status

Now I reviewed entire files changed.

## Proposal

I believe we can apply most of source code in this PR. Before starting refactoring, here it would be beneficial to have solid safety guard for refactoring: characterization tests, which enables us to detect any regression immediately before doing tests by using real sensor input.

As the attached picture, current code coverage is not enough. In this situation, doing refactoring might cause unexpected regression and so on.

So I'd like to propose the following approach.

How about separating this PR into some PRs like as follows?

- Add characterization test first as safety guard (prevents regression)
- Core logic separation from ROS2 node logic
- Add unit tests for the core logic only
- Add python interface and its usage examples
- Minor refactoring

If you agree with this idea, I can start from adding the characterization tests. Thank you in advance!

## Current Code Coverage

Screenshot from 2026-09-04 23-16-16

vividf:
@mojomex @sasakisasaki 
Note: will reopen again after https://github.com/autowarefoundation/autoware_universe/pull/12873 and https://github.com/autowarefoundation/autoware_universe/pull/13113.
Please review these two PRs first since it can change pybind code

sasakisasaki:
@vividf Thank you for letting me know! As regression detection in refactoring, I created a [PR that adds characterization test](https://github.com/autowarefoundation/autoware_universe/pull/13379): this PR does not touch anything in production code, it only checks the current node behavior.

As impact scope of characterization test is nothing, perhaps starting from adding the characterization test might be one of the choices :wink:

PR #1395, opened 2022-07-21
fix(ad_service_state_monitor): lock shread variables to prevent race …

ytakano:
`state_machine_` and `state_input_` are shared by multiple callback groups (threads).
So, mutex locks have been applied because it may cause race conditions.

Signed-off-by: Yuuki Takano <yuuki.takano@tier4.jp>

## Description

`AutowareStateMonitorNode::state_machine_` and `AutowareStateMonitorNode::state_input_` are accessed by multiple threads without mutex locks. This may cause race conditions.
So, mutex locks have been applied to the shared variables.

## Pre-review checklist for the PR author

- [x] I've confirmed the [contribution guidelines].
- [x] The PR follows the [pull request guidelines].

## In-review checklist for the PR reviewers

- [x] The PR follows the [pull request guidelines].

## Post-review checklist for the PR author

The PR author **must** check the checkboxes below before merging.

- [x] There are no open discussions or they are tracked via tickets.

After all checkboxes are checked, anyone who has write access can merge the PR.

[contribution guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/
[pull request guidelines]: https://autowarefoundation.github.io/autoware-documentation/main/contributing/pull-request-guidelines/

PR #12876, opened 2026-06-25
feat(radar_objects_adapter): apply `agnocast_wrapper::Node` to `radar_objects_adapter`

Koichi98:
## Description
Apply `agnocast_wrapper::Node` to `autoware_radar_objects_adapter`.

This node has a single callback group, so none of the above applies.

## Related Links

## How was this PR tested?

Verified on 2026-09-03 against `autoware_core` and `autoware_universe` main of the same day, with Agnocast 2.4.0 (userland, kernel module and heaphook all on 2.4.0).

- The generated node main reports `agnocast_only = true` and the node is registered with `AgnocastOnlyCallbackIsolatedExecutor`.
- This node is not part of the standard logging simulator, so it was driven standalone with synthetic input instead: a `RadarInfo` carrying the eight required field names, followed by `RadarObjects` at 10 Hz with one dynamic object.
- Both outputs are published at the input rate with `ENABLE_AGNOCAST=0` and `ENABLE_AGNOCAST=1`: `~/output/detections` at 9.994 Hz and 9.999 Hz, `~/output/tracks` at 9.969 Hz and 9.995 Hz. No `process has died` and no error log in either mode.
- The topic-name hash embedded in `object_id.uuid[4..11]` is byte-identical between the two modes (`205 114 52 221 216 165 73 100`), which shows `Subscription::get_topic_name()` returns the same resolved topic name on the Agnocast path and on the DDS path.
- No `rclcpp::ok()`, `rclcpp::shutdown()` or `rclcpp::Rate` remains in this package.

## Notes for reviewers

This node hashes its input topic name into the UUID of every output object, so it needs the subscription to report its own topic name.
`Subscription::get_topic_name()` was added to the wrapper in autoware_core #1395, so the original line is kept as it is on main and no workaround is needed:

```cpp
std::size_t hash_code = std::hash<std::string>{}(radar_objects_sub_->get_topic_name());
```
This is the only part of the conversion that depends on a wrapper API beyond the Node base class, and it does not appear in the diff because the line itself is unchanged.
The publishers are converted from AUTOWARE_CREATE_PUBLISHER2 to this->create_publisher<T>(), because the macro only accepts rclcpp::Node * and this node no longer derives from it.

## Interface changes

None.

## Effects on system behavior


YoshiRi (approved):
LGTM

PR #12878, opened 2026-06-25
refactor(autoware_traffic_light_arbiter): consolidate external snapshot in arbitrate()

takam5f2:
## Description

Readability refactor of `TrafficLightArbiterCore::arbitrate()` — no behavior change.

The body previously hand-rolled the external-signal aggregation: a loop that simultaneously built the signals array, tracked the freshest stamp with a running max, and set a `has_externals` flag. That same "freshest external stamp" logic was then **recomputed** at the bottom for `latest_input_time`, so there was no single source of truth.

Changes (anonymous-namespace helpers in `traffic_light_arbiter_core.cpp`):

- Add `struct ExternalSnapshot { signals; max_stamp; has_any; }` and `collect_external_snapshot()`. The snapshot is built once; the perception-staleness check and `latest_input_time` now both read `external.max_stamp` / `external.has_any` instead of recomputing the max.
- Add `route_signals()` — routes a whole array via the existing `route_signal()` — and use it to fold the two inner `for` loops in the matching/priority dispatch. The policy decision (which priority flag, whether to validate) stays in `arbitrate()`.

`effective_perception` remains a non-destructive const-ref view (no per-cycle copy). Routing and accumulation semantics are unchanged.

## How was this PR tested?

- `colcon build --packages-select autoware_traffic_light_arbiter` — passes.
- `colcon test ... -R autoware_traffic_light_arbiter_core_test` — 39/39 pass, unchanged (the refactor is behavior-preserving; no test was modified).
- pre-commit hooks (clang-format / cpplint) pass.

## Notes for reviewers

Follow-up to the include/alias cleanup (already merged). Pure structural change scoped to a single file.

### DIFF
--- a/perception/autoware_traffic_light_arbiter/src/traffic_light_arbiter_core.cpp
+++ b/perception/autoware_traffic_light_arbiter/src/traffic_light_arbiter_core.cpp
@@ -35,11 +35,39 @@ namespace
 using autoware_perception_msgs::msg::PredictedTrafficLightState;
 using autoware_perception_msgs::msg::TrafficLightElement;
 using autoware_perception_msgs::msg::TrafficLightGroup;
+using autoware_perception_msgs::msg::TrafficLightGroupArray;
 
 // One element paired with whether it came from the prioritized source.
 // Accumulated per regulatory-element id while routing signals, then reduced to
 // one element per shape by get_highest_confidence_elements().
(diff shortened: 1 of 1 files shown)


takam5f2:
@YoshiRi @a-maumau  
This PR is a simple PR to improve only readability of the `arbitrate()` method, but not include the logic. Best,

a-maumau (approved):
LGTM!

PR #12882, opened 2026-06-25
test(autoware_traffic_light_classifier): add core unit and node integration tests

takam5f2:
## Description

Adds test coverage for `autoware_traffic_light_classifier` at two layers.

**`test/test_traffic_light_classifier.cpp` — ROS-free unit tests for the core**

Exercises `TrafficLightClassifier::classify()` directly over an in-memory `cv::Mat` + `TrafficLightRoiArray`, with a hand-written `FakeClassifier` standing in for the real backend. This isolates the per-ROI orchestration from rclcpp, `message_filters` and the actual color/CNN models, so the assertions are on `classify()`'s return value rather than on what a node publishes. Covered:

- empty ROIs → empty result, backend not invoked
- non-target type ROI dropped
- zero-sized target ROI appended as `UNKNOWN`
- valid ROI classified, header left untouched by the core
- multiple valid ROIs → ordered output slots
- classified-lead / `UNKNOWN`-appended output ordering
- per-ROI crop geometry/order handed to the backend
- over/under exposure overwrite-to-`UNKNOWN` and flags (single, selective slot, combined)
- backend failure → `nullopt`

**`test/test_traffic_light_classifier_integration.cpp` — node-level integration tests**

Covers the topic-driven pub/sub path and diagnostics (the wiring deliberately left out of the unit tests).

Both are registered in `CMakeLists.txt`: a plain `ament_auto_add_gtest` for the ROS-free core, and `ament_add_ros_isolated_gtest` for the node-level tests.

## Effects on system behavior

None. Test-only change — no production sources are modified.

## How was this PR tested

- `pre-commit` (clang-format, cpplint, …) passes.
- Local `colcon` build/test was **not** run in this environment; the gtests are
  intended to be validated by CI. Opening as **draft** for that reason.

## Notes for reviewers

The unit-test file is written as characterization of `classify()`'s observable contract; comments document what each layer intentionally does and does not cover.

### DIFF
--- a/perception/autoware_traffic_light_classifier/CMakeLists.txt
+++ b/perception/autoware_traffic_light_classifier/CMakeLists.txt
@@ -133,6 +133,25 @@ if(BUILD_TESTING)
     test/test_utils.cpp
   )
 
+  # ROS-free unit tests for the classification core.
+  ament_auto_add_gtest(test_traffic_light_classifier
+    test/test_traffic_light_classifier.cpp
+  )
+  target_link_libraries(test_traffic_light_classifier
+    ${PROJECT_NAME}
+    ${OpenCV_LIBRARIES}
+  )
+
+  # Node-level integration tests (topic-driven pub/sub + diagnostics).
+  ament_add_ros_isolated_gtest(test_traffic_light_classifier_integration
+    test/test_traffic_light_classifier_integration.cpp
+  )
+  target_link_libraries(test_traffic_light_classifier_integration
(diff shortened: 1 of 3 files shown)


takam5f2:
@a-maumau @YoshiRi  
I'd like to add an unit test and a simple integration test. The PR will **NEVER change** the production code. 
Please check it if you have some time. Best,

a-maumau (approved):
LGTM

PR #12885, opened 2026-06-26
test(autoware_traffic_light_category_merger): decouple merge core from node and add tests

takam5f2:
## Description

Decouple the core logic of `autoware_traffic_light_category_merger` from its ROS 2 node and cover it with tests.

The merge logic was written inline in the node's sync callback. This PR:

- Adds **node integration tests** that drive the real node over its ROS interface (car/pedestrian input topics → message_filters `ApproximateTime` sync → merged output topic), pinning the current observable behavior before any refactor.
- Extracts the array-concatenation logic into a ROS-free core class `TrafficLightCategoryMerger` (`traffic_light_category_merger.{hpp,cpp}`). The node now holds a `merger_` member and delegates to `merger_.merge()`; the callback is left responsible only for the ROS interface (sync, publish). Behavior is unchanged.
- Adds **unit tests** for the core, exercising `merge()` directly over in-memory messages (car-then-pedestrian ordering, element preservation, car-array header source, and the empty-car / empty-pedestrian / both-empty edge cases).

Follow-up tidy-ups included:

- Use a shared stamp in the core unit tests (the stamp is irrelevant to merge; arrays are told apart by `frame_id`).
- Rename `signalsCallback` → `signals_callback` and the test's `onSignals` → `on_signals` to follow the Autoware C++ snake_case convention.
- Drop unused node includes (`<string>`, `<vector>`, `<chrono>`) left over after the extraction.

## Related links

**Parent Issue:**

- None

## How was this PR tested?

`colcon build` + `colcon test --packages-select autoware_traffic_light_category_merger` on ROS 2 Humble.

- Integration tests: 2 cases (typical car+pedestrian round trip, both-empty) — all pass.
- Core unit tests: 5 cases — all pass.
- The integration tests pass both before and after the core extraction, confirming behavior is unchanged.

## Notes for reviewers

The merge policy preserved from the original node is: concatenate car signals then pedestrian signals, and adopt the **car array's header** for the output. Whether "car wins" for the header is the desired domain policy is intentionally left unchanged here (it would be a behavior change, out of scope for this test/refactor PR).

## Interface changes

None.

## Effects on system behavior

None. This is a test-and-refactor change with no observable behavior change.

### DIFF
--- a/perception/autoware_traffic_light_category_merger/src/traffic_light_category_merger.cpp
+++ b/perception/autoware_traffic_light_category_merger/src/traffic_light_category_merger.cpp
@@ -1,4 +1,4 @@
-// Copyright 2025 TIER IV, Inc.
+// Copyright 2026 TIER IV, Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/perception/autoware_traffic_light_category_merger/src/traffic_light_category_merger.cpp
+++ b/perception/autoware_traffic_light_category_merger/src/traffic_light_category_merger.cpp
@@ -1,0 +1,32 @@
+// Copyright 2025 TIER IV, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "traffic_light_category_merger.hpp"
+
+namespace autoware::traffic_light
+{
+
+TrafficLightArray TrafficLightCategoryMerger::merge(
+  const TrafficLightArray & car_signals, const TrafficLightArray & pedestrian_signals) const
+{
+  TrafficLightArray output;
+  output.header = car_signals.header;
+  output.signals.insert(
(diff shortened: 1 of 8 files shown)


takam5f2:
@a-maumau @YoshiRi  
Could you check the PR? The PR aggregates node/logic separation, unit test, and integration test. It is because the package has a simple function.

perception/autoware_traffic_light_category_merger/src/traffic_light_category_merger.cpp
```diff
@@ -0,0 +1,32 @@
+// Copyright 2025 TIER IV, Inc.
```
a-maumau: Update the year to 2026.

perception/autoware_traffic_light_category_merger/src/traffic_light_category_merger.hpp
```diff
@@ -0,0 +1,39 @@
+// Copyright 2025 TIER IV, Inc.
```
a-maumau: Update the year to 2026.

a-maumau (approved):
LGTM.
I have only minor comment for the year in the license text.

perception/autoware_traffic_light_category_merger/src/traffic_light_category_merger.hpp
takam5f2: @a-maumau  Thank you. Fixed at https://github.com/autowarefoundation/autoware_universe/pull/12885/commits/e22fd7d634eba312ef7d9d6bb2a3e0717077f81c

perception/autoware_traffic_light_category_merger/src/traffic_light_category_merger.cpp
takam5f2: @a-maumau  Thank you. Fixed at https://github.com/autowarefoundation/autoware_universe/pull/12885/commits/e22fd7d634eba312ef7d9d6bb2a3e0717077f81c

PR #12886, opened 2026-06-26
feat(trajectory_follower, pid, mpc): introduce temporal controller

Kotakku:
## Description
This PR adds a time step based controller mode to the PID longitudinal controller and MPC lateral controller. By default, the spatial mode is used, so there is no change in behavior.

## Related links
- https://github.com/tier4/autoware_universe/pull/2860
  - mainly temporal controller implementarion
- https://github.com/tier4/autoware_universe/pull/3138
  - simplify trajectory reference point selection when temporal mode

## How was this PR tested?
- Tested on PSim
- Tested on real vehicle
- [Evaluator PASS (MasterScenario)](https://evaluation.ci.tier4.jp/evaluation/reports/285be772-2bfc-5516-a3a7-8c34299a28da?project_id=autoware_dev)
- [Evaluator PASS (FuncVerif_Basic)](https://evaluation.ci.tier4.jp/evaluation/reports/fbd1741b-fe7f-58f2-94d9-c58603f71b7e?project_id=autoware_dev)
- [Evaluator PASS (DLR_Basic)](https://evaluation.ci.tier4.jp/evaluation/reports/d4260c50-1e26-5661-8fcb-25d350910922?project_id=autoware_dev)

## Notes for reviewers

None.

#### Additions and removals

| Added | `trajectory_reference_mode` | `string` | `"spatial"` | Selects trajectory tracking mode: `spatial` (distance-based, existing behavior) or `temporal` (time-step-based) |

## Effects on system behavior

None, except for a minor fix in spatial mode: `control_cmd_horizon`'s `steering_tire_rotation_rate` now uses the actual per-step `prediction_dt` instead of `m_ctrl_period`, which is more accurate since horizon steps are spaced by `prediction_dt`. The main `control_cmd` output is unaffected.


control/autoware_mpc_lateral_controller/param/lateral_controller_defaults.param.yaml
```diff
@@ -4,6 +4,8 @@
     traj_resample_dist: 0.1         # path resampling interval [m]
     use_steer_prediction: false     # flag for using steer prediction (do not use steer measurement)
     use_delayed_initial_state: true # flag to use x0_delayed as initial state for predicted trajectory
+    # NOTE: trajectory_reference_mode is now set at trajectory_follower_node level
```
go-sakayori: nits:

```suggestion
```

control/autoware_mpc_lateral_controller/include/autoware/mpc_lateral_controller/mpc.hpp
```diff
@@ -156,6 +160,14 @@ struct MPCData
   // Pose (position and orientation) of the nearest point in the trajectory.
   Pose nearest_pose{};
 
+  // Temporal tracking debug values.
```
go-sakayori, line 163: Are you using these debug values. If not I would remove it

control/autoware_mpc_lateral_controller/include/autoware/mpc_lateral_controller/mpc_utils.hpp
go-sakayori, line 160:
```suggestion
 * calculation
 * @param [in]  use_short_segment_protection *****
```

control/autoware_mpc_lateral_controller/include/autoware/mpc_lateral_controller/mpc_utils.hpp
```diff
@@ -175,7 +201,10 @@ std::vector<double> calcTrajectoryCurvature(
  */
 bool calcNearestPoseInterp(
   const MPCTrajectory & traj, const Pose & self_pose, Pose * nearest_pose, size_t * nearest_index,
-  double * nearest_time, const double max_dist, const double max_yaw);
+  double * nearest_time, const double max_dist, const double max_yaw,
```
go-sakayori, line 209: nits: update doxygen

control/autoware_mpc_lateral_controller/src/mpc.cpp
```diff
@@ -35,12 +37,80 @@ using autoware_utils::calc_distance2d;
...
+    return false;
+  }
+  if (interpolated.empty()) {
+    return false;
+  }
+
+  pose->position.x = interpolated.x.front();
+  pose->position.y = interpolated.y.front();
+  pose->position.z = interpolated.z.front();
+  pose->orientation = autoware_utils::create_quaternion_from_yaw(interpolated.yaw.front());
```
go-sakayori:
```suggestion
  pose->orientation = autoware_utils_geometry::create_quaternion_from_yaw(interpolated.yaw.front());
```

mkquda (approved):
LGTM.

control/autoware_mpc_lateral_controller/src/mpc.cpp
```diff
@@ -698,35 +980,60 @@ void MPC::addSteerWeightF(const double prediction_dt, MatrixXd & f) const
...
-  // Calculate the time min_prediction_length ahead from current_pose
-  const auto autoware_traj = MPCUtils::convertToAutowareTrajectory(input);
-  const size_t nearest_idx = autoware::motion_utils::findFirstNearestIndexWithSoftConstraints(
-    autoware_traj.points, current_kinematics.pose.pose, ego_nearest_dist_threshold,
-    ego_nearest_yaw_threshold);
-  double sum_dist = 0;
-  const double target_time = [&]() {
+  if (m_use_temporal_trajectory) {
+    // Temporal mode: use fixed prediction_dt from parameters
+    // Calculate the end time of the prediction horizon
```
go-sakayori: Do we need this comment?

control/autoware_mpc_lateral_controller/src/mpc_utils.cpp
```diff
@@ -40,6 +40,28 @@ double calcLongitudinalOffset(
...
 }
+
+bool isTemporalShortSegment(
+  const double ds, const double dt, const double vx, const bool use_short_segment_protection)
+{
+  if (!use_short_segment_protection) {
+    return ds < 1.0e-3;
+  }
+
+  constexpr double min_distance_threshold = 2.0e-2;
```
go-sakayori, line 51: Are these values going to be fixed, or should it be parameters?

control/autoware_mpc_lateral_controller/include/autoware/mpc_lateral_controller/mpc_utils.hpp
```diff
@@ -130,17 +134,23 @@ bool calcMPCTrajectoryTime(MPCTrajectory & traj);
  * @param [in] acc_lim limit on the acceleration
  * @param [in] tau constant to control the smoothing (high-value = very smooth)
  * @param [inout] traj MPCTrajectory for which to calculate the smoothed velocity
+ * @param [in] use_temporal_trajectory if true, preserve timestamps; if false, recalculate time from
+ * velocity
  */
 void dynamicSmoothingVelocity(
   const size_t start_seg_idx, const double start_vel, const double acc_lim, const double tau,
-  MPCTrajectory & traj);
+  MPCTrajectory & traj, const bool use_temporal_trajectory = true);
```
go-sakayori: Maybe better, as the default is spatial
```suggestion
  MPCTrajectory & traj, const bool use_temporal_trajectory = false);
```

control/autoware_mpc_lateral_controller/include/autoware/mpc_lateral_controller/mpc_utils.hpp
```diff
@@ -68,9 +69,12 @@ double calcLateralError(const Pose & ego_pose, const Pose & ref_pose);
 /**
  * @brief convert the given Trajectory msg to a MPCTrajectory object
  * @param [in] input trajectory to convert
+ * @param [in] use_temporal_trajectory if true, use time_from_start directly; if false, calculate
+ * from distance/velocity
  * @return resulting MPCTrajectory
  */
-MPCTrajectory convertToMPCTrajectory(const Trajectory & input);
+MPCTrajectory convertToMPCTrajectory(
+  const Trajectory & input, const bool use_temporal_trajectory = true);
```
go-sakayori: Default is spatial so should be changed?
```suggestion
  const Trajectory & input, const bool use_temporal_trajectory = false);
```

control/autoware_mpc_lateral_controller/src/mpc.cpp
```diff
@@ -141,7 +223,7 @@ ResultWithReason MPC::calculateMPC(
     lateral.steering_tire_angle = static_cast<float>(std::clamp(*it, -m_steer_lim, m_steer_lim));
     lateral.steering_tire_rotation_rate =
       (lateral.steering_tire_angle - ctrl_cmd_horizon.controls.back().steering_tire_angle) /
-      m_ctrl_period;
+      prediction_dt;
```
go-sakayori, line 205: This change isn't guarded by `m_use_temporal_trajectory`, so it also alters `control_cmd_horizon` output in spatial mode. This fix seems correct but it contradicts the "no behavior change in spatial mode". Please update the desctiption

go-sakayori (commented):
Please update README and schema.

control/autoware_pid_longitudinal_controller/src/pid_longitudinal_controller.cpp
```diff
@@ -457,19 +496,74 @@ PidLongitudinalController::ControlData PidLongitudinalController::getControlData
   control_data.current_motion.vel = m_current_kinematic_state.twist.twist.linear.x;
   control_data.current_motion.acc = m_current_accel.accel.accel.linear.x;
   control_data.interpolated_traj = m_trajectory;
+  const double traj_start_time =
```
go-sakayori, line 479: I think we need an `empty()/size()<2' check before calling these.

control/autoware_pid_longitudinal_controller/src/pid_longitudinal_controller.cpp
```diff
@@ -258,7 +274,30 @@ void PidLongitudinalController::setTrajectory(const autoware_planning_msgs::msg:
     return;
   }
 
+  bool trajectory_stamp_changed = false;
+  if (m_use_temporal_trajectory) {
+    const rclcpp::Time current_stamp(msg.header.stamp);
+    trajectory_stamp_changed =
+      m_prev_trajectory_stamp.has_value() && current_stamp != *m_prev_trajectory_stamp;
```
yuki-takagi-66: Does this function correctly even when the generator is switched?

control/autoware_pid_longitudinal_controller/src/pid_longitudinal_controller.cpp
```diff
@@ -478,7 +572,17 @@ PidLongitudinalController::ControlData PidLongitudinalController::getControlData
 
   // calculate the target motion for delay compensation
   constexpr double min_running_dist = 0.01;
-  if (control_data.state_after_delay.running_distance > min_running_dist) {
+  if (m_use_temporal_trajectory) {
+    const double nearest_time = rclcpp::Duration(nearest_point.time_from_start).seconds();
```
yuki-takagi-66, line 527: On the surface, it appears that `clock.now()` would work fine for this part. Could you help me understand the rationale behind estimating the time with `estimateTrajectoryTimeFromPose()` , `estimateLocalTrajectoryTimeStep()` and `m_prev_nearest_time`? I apologize if I'm simply lacking the historical context.

yuki-takagi-66 (commented):
As a summary comment, I feel like the logic around m_prev_nearest_time looks a bit off. However, I couldn't think of any specific problems it might cause (other than just making the code longer). So, I think we can leave it as it is.

control/autoware_mpc_lateral_controller/src/mpc_utils.cpp
Kotakku: The constant used for the short-segment check is small enough that I don't think it needs per-vehicle tuning, so I'd like to keep it as is.

control/autoware_pid_longitudinal_controller/src/pid_longitudinal_controller.cpp
Kotakku: Indeed, there's no guarantee that this logic gets triggered exactly when the generator switches. However, if the stamp doesn't change, that means the trajectory was planned from the same position as before — so the original problem this logic was introduced to solve (the trajectory's start advancing without that change being detected) doesn't reproduce in that case, and I don't think it becomes an actual issue. That said, we may still need to keep discussing this on the implementation side.

control/autoware_pid_longitudinal_controller/src/pid_longitudinal_controller.cpp
Kotakku: Because of state transitions such as stopping, and tracking-performance limitations, the trajectory's phase doesn't necessarily progress exactly according to time_from_start. So I think the current approach of computing nearest_time using both a prediction-based and an observation-based method is reasonable.

yuki-takagi-66 (approved):

PR #12887, opened 2026-06-26
fix(external_cmd_selector): data race bug

Koichi98:
## Description
`current_selector_mode_` is written by the service callback (`on_select_external_command`) and read by the command-relay subscriptions and the timer. These callbacks belong to two different callback groups, so they can run concurrently under a multi-threaded container executor, making the unguarded access a data race.

This PR adds a mutex to serialize all reads and writes of `current_selector_mode_`.

## Changes
- Add `current_selector_mode_mutex_` guarding every access to `current_selector_mode_`.
- `publish()` calls are kept outside the lock to avoid holding it longer than necessary.

## Tests performed
- `colcon build` for `autoware_external_cmd_selector` passes.
- (Will later check the behavior is preserved by running Psim)

### DIFF
--- a/control/autoware_external_cmd_selector/include/autoware/external_cmd_selector/external_cmd_selector_node.hpp
+++ b/control/autoware_external_cmd_selector/include/autoware/external_cmd_selector/external_cmd_selector_node.hpp
@@ -28,8 +28,8 @@
 #include <tier4_control_msgs/msg/external_command_selector_mode.hpp>
 #include <tier4_control_msgs/srv/external_command_select.hpp>
 
+#include <atomic>
 #include <memory>
-#include <mutex>
 
 namespace autoware::external_cmd_selector
 {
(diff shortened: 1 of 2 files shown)


shmpwk (approved):
LGTM as a data-race fix. 
The mutex scope is appropriate, and keeping publish() outside the lock is the right call. 
The lock_guard usage isn’t really excessive, though the repeated pattern could be simplified with `std::atomic<uint8_t>`. 
Actually, I haven't seen such data-race in `external_cmd_selector`, but I don’t see a practical concern to merge this.

shmpwk:
@Koichi98 I'm just curious about the background of this PR.

Koichi98:
@shmpwk 
> I'm just curious about the background of this PR.

I've been working on a series of PRs to apply Agnocast — and, as part of that effort, the `CallbackIsolatedExecutor` (CIE) — across various nodes. While checking whether enabling CIE could cause data races between callback groups, I happened to run into this bug. So it wasn't the main target of the work; I just stumbled upon it during that verification.

Koichi98:
@shmpwk 
> The lock_guard usage isn’t really excessive, though the repeated pattern could be simplified with std::atomic<uint8_t>.

Thank you for pointing out. I've fixed to use `atomic<uint_8>`.
Could you check the `Files Changed` once again to make sure? Also, this since this fix is a slight one, I'd like to skip the PSIM execution test this time. Is that okay? I've checked that this successfully compiles.

shmpwk (approved):
LGTM

PR #12890, opened 2026-06-26
test(autoware_traffic_light_classifier): add ColorClassifier unit tests

takam5f2:
## Description

Adds focused unit tests for `ColorClassifier` in `autoware_traffic_light_classifier` (`test/test_color_classifier.cpp`).

Coverage:

- HSV-band classification (green / amber / red) from solid in-band images
- out-of-band image → `UNKNOWN` with zero confidence
- confidence is strictly within (0, 1) for a normal match, and clamps to exactly 1.0 once the matched pixel count saturates the formula
- thresholds actually drive classification (narrowing a band flips the result)
- mismatched images/signals count is rejected
- empty batch is a successful no-op

Tests follow Arrange-Act-Assert. `ColorClassifier` currently takes an `rclcpp::Node *` and reads its HSV thresholds from parameters, so the fixture hosts a node and primes the thresholds via `set_parameter()`. The assertions are written so that only this construction/priming step would change if the classifier is later decoupled from ROS.

## Related links

None.

## How was this PR tested?

Built and ran the package test suite locally on a CPU (non-GPU) build:

```
colcon build --packages-select autoware_traffic_light_classifier
```

The new `test_color_classifier` suite passes (7 tests) and the existing tests are unaffected.

## Notes for reviewers

Test-only change; no production code is modified. The tests target the current ROS-coupled `ColorClassifier`, hence the node-hosted fixture and `set_parameter()` priming. A follow-up will decouple the classifier from ROS, at which point only the Arrange step (construction/priming) changes while these assertions stay the same.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_traffic_light_classifier/CMakeLists.txt
+++ b/perception/autoware_traffic_light_classifier/CMakeLists.txt
@@ -142,6 +142,15 @@ if(BUILD_TESTING)
     ${OpenCV_LIBRARIES}
   )
 
+  # ColorClassifier unit tests (ROS-isolated: the classifier currently needs a node).
+  ament_add_ros_isolated_gtest(test_color_classifier
+    test/test_color_classifier.cpp
+  )
+  target_link_libraries(test_color_classifier
+    ${PROJECT_NAME}
+    ${OpenCV_LIBRARIES}
+  )
+
   # Node-level integration tests (topic-driven pub/sub + diagnostics).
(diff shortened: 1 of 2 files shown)


takam5f2:
@a-maumau could you review the PR? this PR will NEVER change the production code.

a-maumau (approved):
LGTM!

PR #12898, opened 2026-06-30
test(tensorrt_yolox): add detector test and reduce test for node

interimadd:
## Description

This PR do refactoring on tests like following.

**Add `test/test_tensorrt_yolox_detector.cpp`:** A new unit test that drives `TrtYoloXDetector::detect()` directly, without any ROS executor or topic plumbing.

**Reduce `test/test_tensorrt_yolox_node.cpp` to a ROS-plumbing smoke test:** Now that the detection / segmentation logic is covered directly by the detector unit test, the node test no longer re-tests that logic. It keeps a single case that publishes a blank image and verifies the full ROS path (parameter loading → lazy subscription via `onConnect` → `onImage` → `detect` → publish).

Both tests require an NVIDIA GPU, TensorRT and the ONNX models under `autoware_data`; they self-skip (`GTEST_SKIP`) when these are unavailable and are gated in CMake behind `TRT_AVAIL AND CUDA_AVAIL`, so they are a no-op on CI machines without a GPU.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

Built and ran the package tests on a machine with an NVIDIA GPU:

```bash
colcon build --packages-select autoware_tensorrt_yolox
colcon test --packages-select autoware_tensorrt_yolox --event-handlers console_cohesion+ \
  --ctest-args -R "test_tensorrt_yolox_node|test_tensorrt_yolox_detector"
```

All tests pass: `test_tensorrt_yolox_detector` (3 cases) and `test_tensorrt_yolox_node` (1 case). 

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_tensorrt_yolox/CMakeLists.txt
+++ b/perception/autoware_tensorrt_yolox/CMakeLists.txt
@@ -191,6 +191,18 @@ if(BUILD_TESTING)
       ${PROJECT_NAME}_node
       ${OpenCV_LIBRARIES}
     )
+
+    # Unit test that drives TrtYoloXDetector::detect() directly (no ROS topics). Same GPU / TensorRT
+    # / ONNX model requirements as the node test, so it is gated here and self-skips at runtime.
+    ament_auto_add_gtest(test_tensorrt_yolox_detector
+      test/test_tensorrt_yolox_detector.cpp
+      TIMEOUT 600
+    )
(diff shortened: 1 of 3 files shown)


ktro2828 (approved):
@interimadd Sorry for late, LGTM.

PR #12900, opened 2026-06-30
feat(diffusion planner): change the classes to ignore

SakodaShintaro:
## Description

This pull request will remove the `ignore_unknown_neighbors` ROS parameter and introduce a fixed class mapping in `get_model_label`. Each `ObjectClassification` is now explicitly mapped to a model label (`VEHICLE/PEDESTRIAN/BICYCLE`), and classes that should not be fed to the model (including `UNKNOWN`) are mapped to the new `AgentLabel::IGNORE`.

Note: Classes added after `autoware_msgs` 1.3.0 are left to the default branch so the code still builds against older message versions.

## How was this PR tested?

- [x] planning simulator

## Notes for reviewers

None.

#### Additions and removals

| Removed | `ignore_unknown_neighbors`   | `bool` | `true`         | Ignore neighbor agents with unknown class |

## Effects on system behavior

None.


yukkysaito:
@SakodaShintaro It looks like our fixes overlap. Which approach do you think we should go with?

https://github.com/autowarefoundation/autoware_universe/pull/12901

go-sakayori (approved):

PR #12907, opened 2026-07-01
fix(trajectory_optimizer): apply engage speed clamp after velocity smoothing

Owen-Liuyuxuan:
## Summary
- Reorder `TrajectoryVelocityOptimizer::optimize_trajectory` so `set_engage_speed` velocity clamping runs **after** lateral-acceleration limits, global speed limits, and continuous-jerk smoothing.
- Smoother is not good at handling cases when the ego velocity is significantly lower than the "set" speed. This PR fix this and make smoother more robust.
- When `set_engage_speed` is enabled and current speed is below the pull-out target, the engage/pull-out speed and acceleration constraints are now applied to the smoothed trajectory instead of being overwritten by the smoother.

Port of [tier4/autoware_universe#3124](https://github.com/tier4/autoware_universe/pull/3124).

## Test plan
- [ ] Not run locally — rely on CI
- [ ] Planning simulator: engage with `set_engage_speed: true` and verify pull-out velocity/acceleration profile matches engage constraints after smoothing
- [ ] Regression: confirm normal driving (above pull-out speed) behavior is unchanged

## Notes for reviewers
This is an ordering-only change — no new parameters or logic. The engage clamp block is moved from before speed limiting/smoothing to immediately after `filter_velocity`.

Made with [Cursor](https://cursor.com)


danielsanchezaran (approved):
LGTM

PR #12908, opened 2026-07-01
fix(mpc_lateral_controller): guard distance resampling against invalid arc length

Owen-Liuyuxuan:
## Summary
- Guard `resampleMPCTrajectoryByDistance()` against invalid arc-length keys before spline/lerp interpolation (non-monotonic or undersampled geometry).
- Return `{false, {}}` instead of throwing, so `MPC::setReferenceTrajectory()` can skip the cycle without aborting the trajectory follower node.
- Log throttled `RCLCPP_ERROR` with exception message and trajectory context when interpolation fails.

## Test plan
- [x] Not run locally — rely on CI
- [ ] Reproduce prior crash with degenerate spatial trajectory; expect throttled error and skipped control instead of `validateKeys` abort

## Notes for reviewers
Fixes abort in `agnocast_component_container_cie` when MPC distance resampling hits `autoware::interpolation::validateKeys()` on trajectories with many points but flat/duplicate arc length.

### DIFF
--- a/control/autoware_mpc_lateral_controller/src/mpc_utils.cpp
+++ b/control/autoware_mpc_lateral_controller/src/mpc_utils.cpp
@@ -14,6 +14,7 @@
 
 #include "autoware/mpc_lateral_controller/mpc_utils.hpp"
 
+#include "autoware/interpolation/interpolation_utils.hpp"
 #include "autoware/interpolation/linear_interpolation.hpp"
 #include "autoware/interpolation/spline_interpolation.hpp"
 #include "autoware/motion_utils/trajectory/trajectory.hpp"
@@ -23,6 +24,7 @@
 #include <algorithm>
 #include <iostream>
 #include <limits>
+#include <stdexcept>
 #include <string>
 #include <utility>
 #include <vector>
@@ -112,7 +114,11 @@ std::pair<bool, MPCTrajectory> resampleMPCTrajectoryByDistance(
   std::vector<double> input_arclength;
   calcMPCTrajectoryArcLength(input, input_arclength);
 
-  if (input_arclength.empty()) {
+  if (input_arclength.size() < 2) {
+    return {false, output};
+  }
+
+  if (!autoware::interpolation::isIncreasing(input_arclength)) {
     return {false, output};
   }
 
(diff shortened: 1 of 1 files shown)


theCodeForgerHQ (commented):
Thank you for this fix. I wrote a unit test that reproduces the abort without a full stack, in case it helps get this merged.

On main the first test fails because the call throws; on this branch both tests pass:

<details>
<summary>Test (append to <code>control/autoware_mpc_lateral_controller/test/test_mpc_utils.cpp</code>)</summary>

```cpp
// https://github.com/autowarefoundation/autoware_universe/pull/12908
// Two consecutive identical positions give equal arc-length keys. Resampling must report failure
// instead of throwing out of the spline interpolation (which aborts the controller container).
TEST(TestMPC, ResampleByDistanceRejectsZeroLengthSegment)
{
  using autoware::motion::control::mpc_lateral_controller::MPCTrajectory;
  // x, y, z, yaw, vx, k, smooth_k, relative_time
  MPCTrajectory input;
  input.push_back(0.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0);
  input.push_back(1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 1.0);
  input.push_back(1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 1.0);  // duplicated position
  input.push_back(2.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 2.0);
  input.push_back(3.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 3.0);

  std::pair<bool, MPCTrajectory> result{true, MPCTrajectory{}};
  EXPECT_NO_THROW(result = MPCUtils::resampleMPCTrajectoryByDistance(input, 0.1, 0, 0.0));
  EXPECT_FALSE(result.first);
}

TEST(TestMPC, ResampleByDistanceStillWorksOnValidInput)
{
  using autoware::motion::control::mpc_lateral_controller::MPCTrajectory;
  MPCTrajectory input;
  for (int i = 0; i < 5; ++i) {
    const auto d = static_cast<double>(i);
    input.push_back(d, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, d);
  }
  const auto [ok, resampled] = MPCUtils::resampleMPCTrajectoryByDistance(input, 0.5, 0, 0.0);
  EXPECT_TRUE(ok);
  EXPECT_GE(resampled.size(), 8UL);
}
```

</details>

<details>
<summary>Output on main (3c7d46e) and on this branch (77559b4)</summary>

```text
# main (3c7d46e)
[ RUN      ] TestMPC.ResampleByDistanceRejectsZeroLengthSegment
Expected: result = MPCUtils::resampleMPCTrajectoryByDistance(input, 0.1, 0, 0.0) doesn't throw an exception.
  Actual: it throws.
[  FAILED  ] TestMPC.ResampleByDistanceRejectsZeroLengthSegment (1 ms)
[ RUN      ] TestMPC.ResampleByDistanceStillWorksOnValidInput
[       OK ] TestMPC.ResampleByDistanceStillWorksOnValidInput (0 ms)

# this branch (77559b4)
[ RUN      ] TestMPC.ResampleByDistanceRejectsZeroLengthSegment
[       OK ] TestMPC.ResampleByDistanceRejectsZeroLengthSegment (0 ms)
[ RUN      ] TestMPC.ResampleByDistanceStillWorksOnValidInput
[       OK ] TestMPC.ResampleByDistanceStillWorksOnValidInput (0 ms)
[  PASSED  ] 2 tests.
```

</details>

The second test passes on both and guards against over-rejecting valid input. The first one covers the new `isIncreasing` guard, so it should also help with the codecov patch report. Feel free to take the test into this PR, or I can open a follow-up PR with it once this one is merged.

One question: with the `isIncreasing` check in place, would you keep the try/catch as a last-resort guard, or narrow it to the specific exception the spline throws? Either is fine by me.

AI usage: the test was drafted with Claude Code; the crash path (`setReferenceTrajectory` → `resampleMPCTrajectoryByDistance` → `autoware::interpolation::spline` → `validateKeys`) follows the description in this PR, and the test was built and run before and after as shown, in `ghcr.io/autowarefoundation/autoware:universe-devel` with `autoware_trajectory_follower_base` also built from main (the image's copy predates `set_steering_offset`).

PR #12912, opened 2026-07-01
ci(cppcheck): define agnocast_wrapper service/client macros

atsushi421:
## Description

The `cppcheck-differential` (and full `cppcheck`) workflows pass a fixed list of `-D` macro definitions so cppcheck can expand the `autoware_agnocast_wrapper` macros (`AUTOWARE_MESSAGE_*_PTR`, `AUTOWARE_SUBSCRIPTION_PTR`, `AUTOWARE_PUBLISHER_PTR`, `AUTOWARE_CREATE_*`, etc.). The list did not include the service/client macros, so the first package that uses them in a `.cpp` file trips cppcheck:

```
error: There is an unknown macro here somewhere. Configuration is required.
If AUTOWARE_SERVER_REQUEST_PTR is a macro then please configure it. [unknownMacro]
  AUTOWARE_SERVER_REQUEST_PTR(tier4_system_msgs::srv::OperateMrm) request,
```

Because cppcheck runs with `--enable=all` and `--error-exitcode=1`, this `[unknownMacro]` diagnostic fails the job with exit code 1.

This PR adds `-D` definitions for the service/client wrapper macros to both `cppcheck-differential.yaml` and `cppcheck.yaml`, mirroring how the existing message macros define both the pointer and the `CREATE_*` forms.

**Callback data pointers** (expansions match the wrapper's actual `message_ptr` ownership: request/received-response are `const`):

- `AUTOWARE_SERVER_REQUEST_PTR(ServiceT)` -> `std::shared_ptr<const typename ServiceT::Request>`
- `AUTOWARE_SERVER_RESPONSE_PTR(ServiceT)` -> `std::shared_ptr<typename ServiceT::Response>`
- `AUTOWARE_CLIENT_REQUEST_PTR(ServiceT)` -> `std::shared_ptr<typename ServiceT::Request>`
- `AUTOWARE_CLIENT_RESPONSE_PTR(ServiceT)` -> `std::shared_ptr<const typename ServiceT::Response>`

**Handle / future types and creation macros** (so a node that creates a service/client or holds a future in a `.cpp` does not reintroduce the same failure):

- `AUTOWARE_SERVICE_PTR` / `AUTOWARE_CLIENT_PTR`
- `AUTOWARE_CLIENT_FUTURE` / `AUTOWARE_CLIENT_SHARED_FUTURE` / `*_AND_REQUEST_ID` variants
- `AUTOWARE_CREATE_SERVICE2..4` / `AUTOWARE_CREATE_CLIENT1..3`

This unblocks PRs that apply `agnocast_wrapper::Node` to nodes with service servers/clients (e.g. #12909).

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


Koichi98 (approved):

mitsudome-r (approved):
I'm okay as a temporarily solution, but we should also think about more sustainable way of fixing this issue. We don't want to hard code all macros in our CI since it reduces readability and maintainability of the workflow.

PR #12916, opened 2026-07-01
refactor(autoware_traffic_light_classifier): extract a Node-free ColorClassifierCore

takam5f2:
## Description

This PR extracts a **Node-free `ColorClassifierCore`** so the classification logic can be
constructed from a plain `HSVConfig` and called directly — no node, no parameters, no RMW.
The change is scoped entirely to `src/classifier/color_classifier.{hpp,cpp}`:

- **`ColorClassifierCore`** depends only on OpenCV + `tier4_perception_msgs` (no `rclcpp`,
  `image_transport`, `cv_bridge`, or logging). Return-based API
  `ClassifierResult classify(const std::vector<cv::Mat> &)`. The HSV bands are built
  in the constructor, fixing a latent issue where they were only built by the reconfigure
  callback.
- **Internal split**: `run_pipeline` (HSV filter → binarize → denoise, per ROI image) →
  `classify_element` (pure decision over the denoised masks) → `make_debug_image` (ROS-free
  mosaic renderer that re-runs `run_pipeline`, invoked only when a debug consumer is attached).
- **`ColorClassifier`** remains a thin ROS adapter with an **unchanged public API** (same
  `rclcpp::Node *` constructor and `getTrafficSignals` override). It keeps the ROS-facing
  concerns: parameter declaration, dynamic reconfigure, `~/debug/image` publishing, logging,
  the images/signals size guard, and element merging (preserving upstream traffic-light
  id/type). The reconfigure lambda was replaced with a named `update_param` helper.

`ClassifierInterface`, the CNN backends, and the node files are intentionally left untouched.

## Related links

**Parent Issue:**

- None

## How was this PR tested?

- `colcon build --packages-select autoware_traffic_light_classifier`
- `colcon test --packages-select autoware_traffic_light_classifier` → **49 tests, 0 failures**
- The existing `test_color_classifier` is kept **unchanged** and passes, confirming the
  refactor preserves behavior.

## Notes for reviewers

- Scope is intentionally minimal and self-contained to `src/classifier/color_classifier.{hpp,cpp}`
  plus a comment reference. The public API of `ColorClassifier` and the `ClassifierInterface`
  contract are unchanged, so all call sites compile without modification.
- A stacked follow-up PR rewrites `test_color_classifier` to construct `ColorClassifierCore`
  directly (ROS-free) and switches it from `ament_add_ros_isolated_gtest` to
  `ament_auto_add_gtest` — where the standalone testability this PR enables is realized.

## Interface changes

None. `ColorClassifier`'s public API and the `ClassifierInterface` abstract interface are
unchanged.

## Effects on system behavior

Behavior-preserving. One intentional refinement: an HSV-filter failure (previously swallowed)
now surfaces through the classifier's failure path — effectively unobservable in practice
since `cv::inRange` only throws on malformed images.

### DIFF
--- a/perception/autoware_traffic_light_classifier/src/classifier/color_classifier.cpp
+++ b/perception/autoware_traffic_light_classifier/src/classifier/color_classifier.cpp
@@ -65,156 +65,174 @@ bool ColorClassifierCore::filter_hsv(
   return true;
 }
 
-ColorClassifierCore::Result ColorClassifierCore::get_traffic_signals(
+ColorClassifierCore::ClassifierResult ColorClassifierCore::get_traffic_signals(
   const std::vector<cv::Mat> & images) const
 {
-  return classify_impl(images, nullptr);
-}
-
-ColorClassifierCore::Result ColorClassifierCore::get_traffic_signals(
-  const std::vector<cv::Mat> & images, std::vector<cv::Mat> & debug_images) const
-{
-  return classify_impl(images, &debug_images);
-}
-
-ColorClassifierCore::Result ColorClassifierCore::classify_impl(
-  const std::vector<cv::Mat> & images, std::vector<cv::Mat> * debug_images) const
-{
-  Result result;
+  ClassifierResult result;
   result.success = true;
   result.signals.signals.resize(images.size());
-  for (size_t image_i = 0; image_i < images.size(); image_i++) {
-    const auto & input_image = images[image_i];
-    auto & traffic_signal = result.signals.signals[image_i];
-    cv::Mat green_image;
-    cv::Mat yellow_image;
-    cv::Mat red_image;
-    if (!filter_hsv(input_image, green_image, yellow_image, red_image)) {
+  for (size_t roi_i = 0; roi_i < images.size(); roi_i++) {
+    const PipelineResult pipeline = run_pipeline(images[roi_i]);
+    result.signals.signals[roi_i].elements.push_back(classify_stages(pipeline.stages));
+    if (!pipeline.filter_ok) {
       result.success = false;
-    }
-    // binarize
-    cv::Mat green_bin_image;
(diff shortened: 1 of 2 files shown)


takam5f2:
@a-maumau @YoshiRi  
Could you check the PR? This PR will decouple the color classifier logic from the ROS 2 node.

perception/autoware_traffic_light_classifier/src/classifier/color_classifier.cpp
```diff
@@ -59,195 +275,57 @@ bool ColorClassifier::getTrafficSignals(
...
+
+  const ColorClassifierCore::ClassifierResult result = core_.classify(images);
+
+  // Publish one debug mosaic per ROI image only when a debug consumer is attached;
+  // make_debug_image re-runs the HSV pipeline, so it stays off the hot path.
+  if (0 < image_pub_.getNumSubscribers()) {
+    for (const auto & image : images) {
       const auto debug_image_msg =
-        cv_bridge::CvImage(std_msgs::msg::Header(), "bgr8", debug_image).toImageMsg();
+        cv_bridge::CvImage(std_msgs::msg::Header(), "bgr8", core_.make_debug_image(image))
```
a-maumau, line 286: Just a comment:
as described in the source code comment, it will run `run_pipeline` twice when there is a subscriber.

a-maumau (approved):
LGTM!

perception/autoware_traffic_light_classifier/src/classifier/color_classifier.cpp
takam5f2: Exactly, it costs twice for debugging. I'd recommend that `make_debug_image()` not run in service. It should only be invoked in testing, debugging, or simulation environments.

PR #12920, opened 2026-07-02
feat(trajectory_selector): apply `agnocast_wrapper::Node` to `autoware_trajectory_selector`

Koichi98:
## Description

Apply `agnocast_wrapper::Node` to `autoware_trajectory_selector`.

Because the node depends on the `trajectory_concatenator` and `trajectory_validator` wrapper libraries, those wrappers are adapted to accept an `agnocast_wrapper::Node` as well, so that the whole trajectory-selection stack builds and runs under both `ENABLE_AGNOCAST=0` and `ENABLE_AGNOCAST=1`.

Main changes:

- `TrajectorySelectorNode` now derives from `agnocast_wrapper::Node`; its subscription/publisher use the Agnocast wrapper types (`AUTOWARE_SUBSCRIPTION_PTR`, `AUTOWARE_PUBLISHER_PTR`), the polling subscribers use the new `autoware::agnocast_wrapper::polling::` free-function API (`polling_subscriber.hpp`), and the timer uses `agnocast_wrapper::create_timer`.
- `AGNOCAST_EXECUTOR` is set to `AgnocastOnlyCallbackIsolatedExecutor`, which is required for Method-2 (`agnocast_wrapper::Node`) nodes.
- The node is launched as a standalone node (not inside a component container).

## How was this PR tested?

- **Build**: Confirmed the trajectory-selection stack (selector, concatenator, validator) builds successfully with both `ENABLE_AGNOCAST=0` and `ENABLE_AGNOCAST=1`.
- **Runtime (PSim)**: On an autoware-based product, confirmed via the planning simulator that the node runs under `ENABLE_AGNOCAST=1`, publishing the validated candidate trajectories at ~10 Hz.
- **Standalone launch**: Confirmed the node starts under `ENABLE_AGNOCAST=1` as a standalone node and resolves its parameters correctly with the `AgnocastOnly` executor.

## Notes for reviewers

- **Polling subscribers**: The polling subscribers use the new `autoware::agnocast_wrapper::polling::create_polling_subscriber<T>()` API (`polling_subscriber.hpp`), whose `take_data()` returns a plain `std::shared_ptr<const T>`. The `ValidatorContext` therefore keeps plain `ConstSharedPtr` fields (unchanged from upstream); no message-pointer wrapping is introduced.
- **Wrapper node type**: The `TrajectoryValidatorWrapper` and `TrajectoryConcatenatorWrapper` constructors now take `agnocast_wrapper::Node &` instead of `rclcpp::Node &`, so the diagnostics/debug helpers use the node-templated `BasicDiagnosticsInterface<...>` / `BasicDebugPublisher<...>`. These are source-compatible for the callers in this PR.
- **Publish pattern**: The selected trajectory is published with `publish(const &)` (rather than allocate-and-`std::move`), so the nested containers of the message are constructed inside the loaned shared-memory buffer. This avoids a cross-process dangling reference when the Agnocast→ROS 2 bridge serializes the message.

## Interface changes

None.

## Effects on system behavior

None.


zulfaqar-azmi-t4 (approved):

kobayu858:
@zulfaqar-azmi-t4 
Sorry for the change after your approval — I moved the polling subscribers back into `node.hpp`. It's just a relocation of the initialization (same topics/types/QoS, no behavior change; builds fine with `ENABLE_AGNOCAST=0/1`). Could you re-review when you have a moment? Thanks!

zulfaqar-azmi-t4 (approved):
@kobayu858 
I checked the new commit and it looks good to me. 👍🏼

kobayu858:
@Kotakku @mkquda @maxime-clem @danielsanchezaran @go-sakayori

Sorry to bother you when you're busy — this PR is already approved by @.zulfaqar-azmi-t4 and CI
is all green, so it just needs one more approval. Could one of you take a look when you have a
moment? Thanks!

maxime-clem (approved):

go-sakayori (approved):

PR #12922, opened 2026-07-02
feat(diffusion_planner,trajectory_ranker,trajectory_adapter): populate turn_indicator field on CandidateTrajectory

Kotakku:
## Description
Following the message change in https://github.com/autowarefoundation/autoware_internal_msgs/pull/89, this PR updates `diffusion_planner` and `trajectory_ranker` as follows:
- `diffusion_planner` now keeps one turn indicator manager per candidate trajectory (batch), and stores the resulting command in each `CandidateTrajectory`. It still publishes the same standalone `TurnIndicatorsCommand` topic as before, so existing behavior does not change for now.
- `autoware_trajectory_ranker` is not currently in use, so only the minimal change needed to populate the new field was applied.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?
- Confirmed via psim that the behavior remains unchanged.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


yhisaki (approved):

zulfaqar-azmi-t4 (approved):

go-sakayori (approved):

planning/autoware_trajectory_adapter/src/trajectory_adapter_node.cpp
go-sakayori, line 61: build pattern should be changed in another PR?

go-sakayori (approved):

PR #12924, opened 2026-07-02
refactor(traffic_light_selector): extract core logic

interimadd:
## Description

Extract the ROI matching logic of `TrafficLightSelectorNode` into a standalone `select` free function (`traffic_light_selector.hpp`/`.cpp`), decoupled from `rclcpp::Node`, and add unit tests covering it directly instead of only through ROS 2 topic-level integration tests.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

- `colcon build --packages-select autoware_traffic_light_selector` succeeds.
- `colcon test --packages-select autoware_traffic_light_selector` passes, including the new `autoware_traffic_light_selector_selector_test` GoogleTest binary (7/7 tests pass).

Integration tests implemented in https://github.com/autowarefoundation/autoware_universe/pull/12892 also pass.

## Notes for reviewers

The `select` function's behavior is unchanged from the previous in-node implementation; this PR is a refactor plus test addition, not a functional change.

## Interface changes

None.

## Effects on system behavior

None.


a-maumau (approved):
LGTM!

PR #12925, opened 2026-07-02
feat(mrm_emergency_stop_operator): apply `agnocast_wrapper::Node` to `autoware_mrm_emergency_stop_operator`

atsushi421:
## Description

Apply `agnocast_wrapper::Node` to `autoware_mrm_emergency_stop_operator`.

FYI: This node uses only the default callback group, so all callbacks remain mutually exclusive and no additional synchronization is required, even though the subscription, service, and timer callbacks share `status_`, `prev_control_cmd_`, and `is_prev_control_cmd_subscribed_`.

## Related Links

`agnocast_wrapper::Node`

## How was this PR tested?

- Ran LSim with sample rosbag
  - build with `ENABLE_AGNOCAST=0`
  - build with `ENABLE_AGNOCAST=1`, ran with `ENABLE_AGNOCAST=0`
  - build with `ENABLE_AGNOCAST=1`, ran with `ENABLE_AGNOCAST=1`

### How to check whether the agnocast wrapper is applied during `ros2 launch`

**1. Node startup log**

The node main generated by `autoware_agnocast_wrapper_register_node()` logs the mode it starts in, so you can check it by grepping the `ros2 launch` log:

```console
$ grep "Starting node" launch.log | grep mrm_emergency_stop_operator
[autoware_mrm_emergency_stop_operator_node-11] [INFO] [1782982777.970801445] [autoware_mrm_emergency_stop_operator_node]: Starting node with agnocast mode (compiled with agnocast support)
```

The line above is the actual output from the LSim run (build and run with `ENABLE_AGNOCAST=1`). The expected log for each configuration is:

| Configuration | Startup log of this node |
|---|---|
| build `1`, run `1` | `Starting node with agnocast mode (compiled with agnocast support)`. The wrapper and `AgnocastOnlyCallbackIsolatedExecutor` are active. |
| build `1`, run `0` | `Starting node with rclcpp mode (compiled with agnocast support)`, followed by `AgnocastOnly executor specified but ENABLE_AGNOCAST is not set, falling back to ROS 2 executor`. |
| build `0` | No such line appears, because the node is registered via the plain `rclcpp_components_register_node()` fallback and runs as a normal ROS 2 component. |

**2. Agnocast topic graph (only when the wrapper is active)**

When built and run with `ENABLE_AGNOCAST=1`, the topics of this node are served by Agnocast, which can be confirmed with the `ros2agnocast` CLI. For example, `ros2 topic echo_agnocast` goes through the Agnocast-to-ROS 2 bridge:

```console
$ ros2 topic echo_agnocast /system/mrm/emergency_stop/status
[*] Triggering A2R bridge for '/system/mrm/emergency_stop/status'. This may take a few seconds...
[*] Starting ros2cli command...
stamp:
  sec: 1733989054
  nanosec: 361725075
state: 1
---
```

`ros2 node list_agnocast` and `ros2 topic list_agnocast` also show this node and its topics only in this configuration.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

When `ENABLE_AGNOCAST=1`, CPU time spent on message serialization/deserialization in this node is reduced (Agnocast zero-copy IPC). The executor also switches to CIE, but since CIE is almost equivalent to `MultiThreadedExecutor`, there is effectively no change in scheduling behavior.


sasakisasaki:
@atsushi421 Thank you for your contribution! I'll review this PR today 👍

system/autoware_mrm_emergency_stop_operator/launch/mrm_emergency_stop_operator.launch.py
```diff
@@ -14,25 +14,44 @@
...
-    with open(config_file_path, "r") as f:
-        params = yaml.safe_load(f)["/**"]["ros__parameters"]
+def generate_launch_description():
+    agnocast_env = IncludeLaunchDescription(
+        PythonLaunchDescriptionSource(
+            PathJoinSubstitution(
+                [
+                    FindPackageShare("autoware_agnocast_wrapper"),
+                    "launch",
+                    "agnocast_env.launch.py",
```
sasakisasaki, line 32: (Note to myself) Reference:
- https://github.com/autowarefoundation/autoware_core/blob/8d210ee4ebf7c4717ac3a7d13da6ae2c38db3e67/common/autoware_agnocast_wrapper/launch/agnocast_env.launch.py

system/autoware_mrm_emergency_stop_operator/src/mrm_emergency_stop_operator_core.cpp
```diff
@@ -80,16 +81,17 @@ rcl_interfaces::msg::SetParametersResult MrmEmergencyStopOperator::onParameter(
...
   if (status_.state != MrmBehaviorStatus::OPERATING) {
-    prev_control_cmd_ = *msg;
+    prev_control_cmd_ = msg;
     is_prev_control_cmd_subscribed_ = true;
   }
 }
 
 void MrmEmergencyStopOperator::operateEmergencyStop(
-  const OperateMrm::Request::SharedPtr request, const OperateMrm::Response::SharedPtr response)
+  const AUTOWARE_SERVER_REQUEST_PTR(OperateMrm) request,
```
sasakisasaki, line 93: (Note to myself) Reference:
- https://github.com/autowarefoundation/autoware_core/blob/8d210ee4ebf7c4717ac3a7d13da6ae2c38db3e67/common/autoware_agnocast_wrapper/include/autoware/agnocast_wrapper/autoware_agnocast_wrapper.hpp#L50-L52

system/autoware_mrm_emergency_stop_operator/src/mrm_emergency_stop_operator_core.cpp
```diff
@@ -142,22 +144,25 @@ void MrmEmergencyStopOperator::publishMrmState() const
   item.mode = driving_mode_id_.value();
   item.state = convert_mrm_state(status_.state);
 
-  DrivingModeMrmState msg;
-  msg.stamp = this->now();
-  msg.items = {item};
-  pub_mrm_state_->publish(msg);
+  auto msg = ALLOCATE_OUTPUT_MESSAGE_UNIQUE(pub_mrm_state_);
```
sasakisasaki, line 147: (Note to myself) Reference:
- https://github.com/autowarefoundation/autoware_core/blob/8d210ee4ebf7c4717ac3a7d13da6ae2c38db3e67/common/autoware_agnocast_wrapper/include/autoware/agnocast_wrapper/autoware_agnocast_wrapper.hpp#L130

sasakisasaki (approved):
Reviewed. The source code, and tests performed looks good to me. Thank you for applying agnocast wrapper for this package. I agree with merging this PR.

Before merging this PR, please add the following information in the PR description.

### Request
I have read "How was this PR tested?". The check looks good. Personally, it would be helpful to add such as command or steps to check if the agnocast wrapper is applied or not during `ros2 launch` (attaching screenshot is also fine). Could you update "how was this PR tested?" in the PR description?

atsushi421:
@sasakisasaki Thank you for the review and approval! I have updated the "How was this PR tested?" section in the PR description: it now includes the commands and expected startup logs to check whether the agnocast wrapper is applied or not during `ros2 launch`, together with the actual outputs from the LSim run. Please take a look.

sasakisasaki:
@atsushi421 Perfect! Thank you very much for adding the detailed description! Now let's merge this PR!

PR #12927, opened 2026-07-02
refactor(multi_object_tracker_node): isolate multiple measurement channels with dynamic callback groups

kobayu858:
## Description
Currently, the introduction of the Callback Isolated Executor into Autoware is underway.
To successfully achieve DAG (Directed Acyclic Graph) Scheduling via the Callback Isolated Executor, each callback recognized as a vertex in the DAG must belong to its own independent CallbackGroup.
This enables a one-to-one mapping between CallbackGroups and OS threads, allowing for fine-grained control over Priority and Affinity on a per-OS-thread basis.

Since isolating all callbacks at once involves a significantly high implementation cost, we are adopting a phased approach—starting the modifications and application with the major nodes that have the highest impact.

**Detailed Changes**

1. Dynamic Generation of Callback Groups Based on the Number of Channels
* Added `cb_groups_measurement_` (a vector for holding the groups) to the header.
* Within the loop processing of the constructor, an independent `MutuallyExclusive` group is dynamically generated for each valid channel and assigned to each subscription via `AUTOWARE_SUBSCRIPTION_OPTIONS`.

2. Complete Protection of Internal State via `std::mutex`
* Newly added `state_mutex_` to the header.
* **Upon Data Reception (`onMeasurement`):** To prevent interference from other threads during processing, the scope executing `core::process_measurement` and `processObjects()` is protected using `std::lock_guard<std::mutex>`.
* **Upon Periodic Timer (`onTimer`):** A lock is similarly applied to the scope handling transmission determination and distribution processing, completely preventing data rewriting (race conditions) during distribution.

## Related links

**Parent Issue:** [Separation of CallbackGroups: Background](https://tier4.atlassian.net/wiki/spaces/~5ed0b4584824b20c18371c06/pages/3009052751/CallbackGroup)

## How was this PR tested?

I ran `lsim` locally using a sample application.
I also measured the latency using CARET and confirmed that there were no significant differences.
The purpose of this PR is to enable scheduler configuration by isolating callback groups, not to reduce latency.

**While the OSS version has only one subscriber, multiple subscribers are actually generated in practice.**

**Before**
image

**After**
image

## Notes for reviewers

- [ ] Ensure that individual callback groups are correctly and dynamically generated and applied to each channel.
- [ ] Verify that the scope of locks using `state_mutex_` within `onMeasurement` and `onTimer` is appropriate, with no risk of deadlocks.
- [ ] Ensure that all accesses to shared resources (`state_` / `debugger_`) are thoroughly protected due to multi-threading.
## Interface changes

None.

## Effects on system behavior

None.


technolojin (commented):
within the given mutex implementation context, odometry callback `updateOdometryBuffer` and it's buffer shall be mutex locked

technolojin (changes requested):

kobayu858:
Closing this as it is out of scope.

PR #12928, opened 2026-07-02
refactor(autoware_traffic_light_selector): shrink node characterization test

interimadd:
## Description

Shrinks the `autoware_traffic_light_selector` node characterization test into a simple integration test. 
It is because now `test_traffic_light_selector.cpp`  covers conversion logic, then `test_traffic_light_selector_node.cpp` have only a role to confirm node can subscribe and publish message.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

- `colcon test --packages-select autoware_traffic_light_selector` passes (7/7 tests, including the shrunk `autoware_traffic_light_selector_node_test`).
- Measured line/function coverage of `traffic_light_selector_node.cpp` with `lcov` before and after the change: unchanged at 100.0% lines (29/29) and 100.0% functions (3/3).

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None. This change only simplifies a test file; no production code changed.

### DIFF
--- a/perception/autoware_traffic_light_selector/test/test_traffic_light_selector_node.cpp
+++ b/perception/autoware_traffic_light_selector/test/test_traffic_light_selector_node.cpp
@@ -17,10 +17,7 @@
 #include <rclcpp/rclcpp.hpp>
 
 #include <sensor_msgs/msg/camera_info.hpp>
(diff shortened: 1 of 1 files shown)


a-maumau (approved):
LGTM.

PR #12929, opened 2026-07-02
feat(perception_analytics_publisher): apply `agnocast_wrapper::Node` to `perception_analytics_publisher`

Koichi98:
## Description
Apply `agnocast_wrapper::Node` to `autoware_perception_analytics_publisher`.

### Node specific change
  - The trigger_node service is registered through the wrapper's `AUTOWARE_CREATE_SERVICE` macro, which takes an `rclcpp::QoS` and absorbs the Humble-only `rmw_qos_profile` signature difference inside the wrapper. As a result, no per-distro / per-build (#ifdef) branch is needed at the call site.

## Related Links

## Related links

**Parent Issue:**

- Link

## How was this PR tested?
  - Ran Autoware based product with this change applied and confirmed it operates without any regression.

## Interface changes

None.

## Effects on system behavior

### DIFF
--- a/evaluator/autoware_perception_online_evaluator/include/autoware/perception_online_evaluator/perception_analytics_calculator.hpp
+++ b/evaluator/autoware_perception_online_evaluator/include/autoware/perception_online_evaluator/perception_analytics_calculator.hpp
@@ -15,7 +15,7 @@
 #ifndef AUTOWARE__PERCEPTION_ONLINE_EVALUATOR__PERCEPTION_ANALYTICS_CALCULATOR_HPP_
 #define AUTOWARE__PERCEPTION_ONLINE_EVALUATOR__PERCEPTION_ANALYTICS_CALCULATOR_HPP_
 
-#include "tf2_ros/buffer.h"
+#include <autoware/agnocast_wrapper/tf2.hpp>
 
 #include "autoware_perception_msgs/msg/predicted_objects.hpp"
 
@@ -78,7 +78,7 @@ public:
    * @param tf_buffer   TF buffer used for transforms
    * @return FrameMetrics
    */
-  FrameMetrics calculate(const tf2_ros::Buffer & tf_buffer) const;
+  FrameMetrics calculate(const autoware::agnocast_wrapper::Buffer & tf_buffer) const;
 
 private:
   mutable std::mutex predicted_objects_mutex_;
--- a/evaluator/autoware_perception_online_evaluator/CMakeLists.txt
+++ b/evaluator/autoware_perception_online_evaluator/CMakeLists.txt
@@ -22,9 +22,11 @@ rclcpp_components_register_node(${PROJECT_NAME}
   EXECUTABLE autoware_perception_online_evaluator_node
 )
 
-rclcpp_components_register_node(${PROJECT_NAME}
+autoware_agnocast_wrapper_register_node(${PROJECT_NAME}
   PLUGIN   "autoware::perception_diagnostics::PerceptionAnalyticsPublisherNode"
   EXECUTABLE autoware_perception_analytics_publisher_node
+  AGNOCAST_EXECUTOR AgnocastOnlyCallbackIsolatedExecutor
+  ROS2_EXECUTOR SingleThreadedExecutor
 )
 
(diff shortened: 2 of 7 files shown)


evaluator/autoware_perception_online_evaluator/include/autoware/perception_online_evaluator/perception_analytics_calculator.hpp
```diff
@@ -78,7 +78,7 @@ class PerceptionAnalyticsCalculator
    * @param tf_buffer   TF buffer used for transforms
    * @return FrameMetrics
    */
-  FrameMetrics calculate(const tf2_ros::Buffer & tf_buffer) const;
+  FrameMetrics calculate(const autoware::agnocast_wrapper::Buffer & tf_buffer) const;
```
sasakisasaki, line 81: (Note to myself) Reference is here:
- https://github.com/autowarefoundation/autoware_core/blob/34800f1b176ec787291372e2850ff460476f4af4/common/autoware_agnocast_wrapper/include/autoware/agnocast_wrapper/tf2.hpp#L45-L54

sasakisasaki: It seems agnocast ON/OFF switch is performed at compile time, not at runtime

sasakisasaki (approved):
Thank you for your contribution. Reviewed, please once read the following my comment before merging this PR. Anyway I agree with merging this PR.

I guess it would be nice to show which part is switch-able via `USE_AGNOCAST_ENABLED` at runtime, and which part is not (= switch-able at compile time). This is for readability and measurement of impact scope.

evaluator/autoware_perception_online_evaluator/include/autoware/perception_online_evaluator/perception_analytics_calculator.hpp
Koichi98: @sasakisasaki 
In agnocast_wrapper, most facilities switch both at compile time and at runtime. E.g. a publisher from `create_publisher()` is a plain rclcpp publisher unless agnocast is active both at compile time and at runtime. Runtime dispatch fits there because the backends are transport-coupled (agnocast shared memory vs. DDS): they behave differently at runtime but share a safe common API.

Buffer (`agnocast::Buffer` / `tf2_ros::Buffer`) is deliberately different — compile-time switch only, and here that is not just sufficient but better:

  1. It gives a compile-time safety we can't get otherwise. The point of `agnocast::Buffer` is to keep the async API (`waitForTransform`, unusable under an AgnocastOnly executor) off the type, so misuse fails to compile. A runtime-dispatched wrapper has one fixed API surface, so it must either re-expose async or hide it while losing the `tf2::BufferCore` identity listeners rely on — neither is safer.
  2. There is nothing to dispatch on. Both are plain `BufferCore` stores with identical lookup semantics; the buffer has no dependency on the communication layer.

Thank you for your comment, and thank you for looking into the agnocast_wrapper code yourself.

Koichi98:
@sasakisasaki 
Thank you for the feedback. I'm sorry for not explicitly mentioning this in the autoware_agnocast_wrapper README. I will update it to clarify which parts are switchable at runtime versus compile time. Do you think I should also include this information in the description of each PR?

sasakisasaki:
@Koichi98 Sorry for being late to reply. Let me answer,

> Do you think I should also include this information in the description of each PR?
Yes, but only once somewhere is enough (Of course you can attach every time. But I guess it would impose you lots of tasks). You can attach the link for the information (which switch-able or not) when if you create a new PR :+1:

sasakisasaki:
As we reviewed and discussed, let me merge this PR

evaluator/autoware_perception_online_evaluator/include/autoware/perception_online_evaluator/perception_analytics_calculator.hpp
sasakisasaki: I see, thank you for your detailed explanation. Though I have need to study the agnocast much more for my deep understanding, the point looks good.

PR #12931, opened 2026-07-02
test(autoware_traffic_light_arbiter): consolidate node tests into a single integration suite

takam5f2:
## Description

Consolidates the two Node-level test files for `autoware_traffic_light_arbiter` into a single, self-contained integration suite.

- **Removed** `test/test_node.cpp` (7 tests) and `test/test_traffic_arbiter_characteristic.cpp` (24 tests).
- **Added** `test/test_traffic_light_arbiter_integration.cpp` (7 tests) verifying the **ROS wiring only**: the three input subscriptions reach the core, `source_priority` / `enable_signal_matching` route in, the output is republished, map presence gates output, and empty input is handled gracefully.
- The exhaustive arbitration **rules** (priority modes, signal-matching truth table, pedestrian handling, time tolerances, staleness) remain covered ROS-free and case-by-case in `test/test_traffic_light_arbiter_core.cpp` (39 tests, unchanged). Each integration test that observes a rule names the owning core test.
- **Dropped** the now-unused `autoware_test_utils` dependency — its only user was `test_node.cpp`; the new suite builds a minimal lanelet map in code with no external map asset.

### Why

`test_node.cpp` and the characterization suite largely re-exercised, through pub/sub, the same logic the ROS-free core test already covers. Collapsing them removes ~2000 lines of duplicated, harness-dependent coverage and clarifies the responsibility split: **the core test owns the rules, the integration test owns the wire.**

## How was this PR tested?

- `colcon test` locally (`RMW_IMPLEMENTATION=rmw_fastrtps_cpp`, `ROS_LOCALHOST_ONLY=1`):
  - integration suite: **7/7 pass** (stable across repeated runs)
  - core suite: **39/39 pass**
- Line coverage of the arbiter production sources is maintained: **95.4%** overall (core 98.7%, validator 94.8%, node 87.3%) — comparable to the removed characterization suite's ~96.7%.
- Lint clean: clang-format, cpplint, ament_copyright, ament_xmllint, ament_lint_cmake.

## Notes for reviewers

- The suite waits on the real publish (`spin_until_publish_count`) instead of fixed sleeps, driving one input at a time — faster and not prone to the multi-publish drop that fixed-sleep node tests hit.
- The `autoware_test_utils` removal is dependency hygiene: verified no other file in the package references it.

## Interface changes

None. No production code, topics, parameters, or message types change — test-only, plus one dependency removal in `package.xml`.

## Effects on system behavior

None.


takam5f2:
@a-maumau  
Could you check the PR? This PR will **NOT CHANGE* the production code.

I'd like to remove the conventional tests using ROS 2 middleware: `test_node.cpp` and `test_traffic_arbiter_characteristic.cpp`. These tests are duplicates of `test_traffic_light_arbiter_core.cpp`.
Since `test_traffic_light_arbiter_core.cpp` does not cover the glue code written with the ROS 2 API, I'd like to add a new, simple test to check ROS 2 communication.

Thank you for reviewing in advance. This is the last PR in the series of core/node decoupling for `autoware_traffic_light_arbiter` package.

Best,

a-maumau (approved):
LGTM

PR #12933, opened 2026-07-03
test(autoware_traffic_light_classifier): drive ColorClassifierCore tests ROS-free and extend coverage

takam5f2:
## Description

Follow-up to #12916, which extracted a Node-free `ColorClassifierCore` but left its unit tests driving through the ROS `ColorClassifier` adapter (hosting an `rclcpp::Node` and priming HSV thresholds via parameters). This PR moves the core　tests onto the core itself and separates the remaining adapter-only concern, so the　classification logic is verified without ROS.

This is a **test-only** change — no production code is touched (`color_classifier.{hpp,cpp}` are byte-identical to `main`).

- **`test_color_classifier.cpp` now exercises `ColorClassifierCore` directly.**
  - The core is constructed from an `HSVConfig` and the tests assert on the `ClassifierResult` returned by `classify()`
  - No node, no parameters, no RMW. Switched from `ament_add_ros_isolated_gtest` to `ament_auto_add_gtest`. The classification / confidence / config assertions are unchanged; only Arrange/Act now talk to the core.
- **Added `make_debug_image` coverage**: mosaic geometry (3 cols × 4 rows, color) and input-dependence (different ROIs produce different mosaics).
- **Extracted the adapter-only concern into `test_color_classifier_adapter.cpp`** (a small ROS-isolated suite): the images/signals size-mismatch guard, which has no meaning at the core layer.
- **Added a new test; an ambiguous-evidence contract test**
  - An image split evenly between two color bands has no strictly-dominant color and must resolve to `UNKNOWN` rather
  than an arbitrary pick — this pins the "strict dominance" branch that no test previously covered.
- Characterization-only assertions (the exact confidence clamp, the mosaic layout) are labelled as such, with a note on what a future contract test would assert instead.

## Related links

**Parent Issue:**

- Follow-up to #12916 (extract a Node-free `ColorClassifierCore`)

## How was this PR tested?

Built and ran the package tests against the latest `main`:

```bash
colcon build --packages-select autoware_traffic_light_classifier --cmake-args -DBUILD_TESTING=ON
```

- `test_color_classifier` (ROS-free core): **9 tests passed**
- `test_color_classifier_adapter` (ROS-isolated): **1 test passed**

## Notes for reviewers

Test-only; production is unchanged from `main`. Reviewing per-commit is recommended — the four commits are self-contained: ROS-free rework + adapter split, fixture removal, ambiguous-evidence coverage, and a comment reword.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_traffic_light_classifier/CMakeLists.txt
+++ b/perception/autoware_traffic_light_classifier/CMakeLists.txt
@@ -142,11 +142,20 @@ if(BUILD_TESTING)
     ${OpenCV_LIBRARIES}
   )
 
-  # ColorClassifier unit tests (ROS-isolated: the classifier currently needs a node).
-  ament_add_ros_isolated_gtest(test_color_classifier
+  # ROS-free unit tests for the color classification core (ColorClassifierCore).
+  ament_auto_add_gtest(test_color_classifier
     test/test_color_classifier.cpp
   )
   target_link_libraries(test_color_classifier
+    ${PROJECT_NAME}
+    ${OpenCV_LIBRARIES}
+  )
+
+  # ColorClassifier ROS adapter tests (the size guard needs the node's logger).
+  ament_add_ros_isolated_gtest(test_color_classifier_adapter
+    test/test_color_classifier_adapter.cpp
+  )
+  target_link_libraries(test_color_classifier_adapter
     ${PROJECT_NAME}
     ${OpenCV_LIBRARIES}
   )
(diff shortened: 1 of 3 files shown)


takam5f2:
@a-maumau  
Could you check the PR? This PR will **NEVER Change** the production codes.

The PR will update unit tests to align the decoupled classifier logic. I will add the test code for `test_color_classifier_adapter.cpp` to cover the current source code structure; `color_classifier.cpp` has the logic and adapter.

Best,

a-maumau (approved):
LGTM

PR #12938, opened 2026-07-03
fix(multi_object_tracker): enforce BOUNDING_BOX output type in PedestrianShapeModel export

technolojin:
## Description

`PedestrianShapeModel::exportTo()` previously switched its output shape type between `CYLINDER` and `BOUNDING_BOX` based on a length/width asymmetry heuristic (nearly-circular footprints were exported as `CYLINDER`). This made the pedestrian tracker's output shape type unstable and inconsistent for downstream consumers.

This change enforces a single, explicit output contract: `PedestrianShapeModel` always exports a `BOUNDING_BOX` with explicit `length` (x) and `width` (y). Input shape types (`BOUNDING_BOX`, `CYLINDER`, `POLYGON`) are still all accepted and normalized internally to a bounding box; only the export path is simplified. The header documentation is updated to match.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

- Built `autoware_multi_object_tracker` locally.
- Existing `multi_object_tracker` unit tests pass.

## Notes for reviewers

The asymmetry-based `CYLINDER` vs `BOUNDING_BOX` branch is removed from `exportTo()`; footprint is cleared and `area` recomputed as before.

## Interface changes

None. (Message/topic/parameter interfaces unchanged; only the exported `shape.type` value produced by the pedestrian tracker changes from a heuristic choice to always `BOUNDING_BOX`.)

## Effects on system behavior

Pedestrian tracks are now always published with `shape.type == BOUNDING_BOX`. Previously, near-symmetric pedestrian footprints could be published as `CYLINDER`. Downstream consumers relying on `CYLINDER` output for pedestrians will now receive equivalent bounding boxes.

### DIFF
--- a/perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/tracker/shape_model/pedestrian_shape_model.hpp
+++ b/perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/tracker/shape_model/pedestrian_shape_model.hpp
@@ -24,16 +24,16 @@
 namespace autoware::multi_object_tracker
 {
 
-// Manages shape extension (length, width, height, output type) for PedestrianTracker.
+// Manages shape extension (length, width, height) for PedestrianTracker.
 // All input shape types (BOUNDING_BOX, CYLINDER, POLYGON) are accepted.
 // Internally always stores BOUNDING_BOX {length, width, height} in tracker heading frame.
-// Output type (BOUNDING_BOX or CYLINDER) is decided at export time based on shape symmetry.
+// Output type is always BOUNDING_BOX with explicit length and width.
 //
 // Data flow:
 //   init()    — force all input types to internal BOUNDING_BOX; apply model-derived sanity bounds
 //   update()  — 3-branch update by input type (BBOX / CYLINDER / POLYGON);
 //               gains differ by type and trust_extension
-//   exportTo() — assemble output with correct shape type
+//   exportTo() — assemble output as a BOUNDING_BOX filling length and width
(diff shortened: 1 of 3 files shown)


perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/tracker/shape_model/pedestrian_shape_model.hpp
ogrex: comment here may also need update 
// Output type is always BOUNDING_BOX with explicit length and width.

ogrex: also in pedestrian_tracker.cpp
currently there is 
```

  // Export shape from extend manager (type selection: CYLINDER vs BOUNDING_BOX)
  assembleShapeTo(object, to_publish);

```

ogrex (changes requested):

perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/tracker/shape_model/pedestrian_shape_model.hpp
technolojin: fixed https://github.com/autowarefoundation/autoware_universe/pull/12938/changes/a8bc7a5497fb4c6cc043c16d3146b4464e1d5345

ogrex (approved):
LGTM

PR #12940, opened 2026-07-03
test(autoware_traffic_light_classifier): add CNN classifier characterization test

takam5f2:
## Description

Add a node-level characterization test for `CNNClassifier` as a safety net **before** refactoring `cnn_classifier.{hpp,cpp}` into a Node-free core and a thin ROS adapter (mirroring the split already done for `ColorClassifier`).

This follows the **same approach as `autoware_tensorrt_yolox`'s integration test (#12774)**: it constructs the real classifier with the MobileNet-v2 ONNX downloaded under `autoware_data`, drives `getTrafficSignals` over the green traffic-light ROI crops in `test/test_data`, and self-skips (`GTEST_SKIP`) when the GPU or model is unavailable. It is additionally gated in CMake behind `TRT_AVAIL AND CUDA_AVAIL`, so it is a no-op on CI machines without a GPU. **No production code is changed.**

Tests (6):

- `ClassifiesNormalGreenCrop`, `ClassifiesWeaklyDimmedGreenCrop`, `ClassifiesMediumDimmedGreenCrop`, `ClassifiesStronglyDimmedGreenCrop` — each green ROI crop decodes to a single `GREEN` `CIRCLE` element with confidence in `(0, 1]`.
- `BatchClassificationScattersPerImageResults` — a bright + a strongly-dimmed crop in one batch: both slots are `GREEN` and the bright slot is strictly more confident than the dimmed one. This pins the static-batch padding/scatter (the batch is padded to the model's static size with copies of the front image); it fails on a swapped slot mapping or a collapse to a single image.
- `MismatchedImageSignalCountReturnsFalse` — the images/signals size guard in `getTrafficSignals`.

## How was this PR tested?

On a machine with an NVIDIA GPU (RTX 5080) + TensorRT 10.8, from the workspace root:

```bash
# 0. Prerequisite: the CAR CNN model + label are downloaded under autoware_data
#    (via the ansible artifacts role). Expected location:
ls ~/autoware_data/ml_models/traffic_light_classifier/traffic_light_classifier_mobilenetv2_batch_6.onnx \
   ~/autoware_data/ml_models/traffic_light_classifier/lamp_labels.txt
#    (the legacy flat ~/autoware_data/traffic_light_classifier/ is also accepted)

# 1. Build (CUDA/TensorRT are auto-detected, which enables the test target)
source install/setup.bash

# 2. Run the characterization test and show the result
colcon test --packages-select autoware_traffic_light_classifier \
  --ctest-args -R test_cnn_classifier
colcon test-result --verbose
```

Result: all 6 tests pass with real inference. `pre-commit run` (clang-format, cpplint) and the full-package `colcon test` are also green.

## Notes for reviewers

- Additive only: one test file + one CMake entry; **no production code is touched.**
- Coverage is intentionally scoped to the CAR classifier and the green crops shipped in `test/test_data` (there are no red/amber/pedestrian/arrow fixtures). This is a regression safety net for the upcoming core/adapter refactor, not exhaustive model coverage.
- `BatchClassificationScattersPerImageResults` hard-codes that a bright green classifies more confidently than a heavily dimmed one (documented in-comment); revisit if a future model equalizes them.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_traffic_light_classifier/CMakeLists.txt
+++ b/perception/autoware_traffic_light_classifier/CMakeLists.txt
@@ -168,6 +168,21 @@ if(BUILD_TESTING)
     ${PROJECT_NAME}
     ${OpenCV_LIBRARIES}
   )
+
+  # GPU/TensorRT characterization test for CNNClassifier (only compiled under
+  # TRT+CUDA, hence the gate). It builds a node, so it uses the ROS-isolated runner;
+  # TIMEOUT covers the first-time TensorRT engine build.
+  if(TRT_AVAIL AND CUDA_AVAIL)
+    ament_add_ros_isolated_gtest(test_cnn_classifier
+      test/test_cnn_classifier.cpp
+      TIMEOUT 600
+    )
+    target_link_libraries(test_cnn_classifier
+      ${PROJECT_NAME}
+      ${OpenCV_LIBRARIES}
+      ${CUDA_LIBRARIES}
+    )
+  endif()
 endif()
 
 ament_auto_package(INSTALL_TO_SHARE
--- a/perception/autoware_traffic_light_classifier/test/test_cnn_classifier.cpp
+++ b/perception/autoware_traffic_light_classifier/test/test_cnn_classifier.cpp
@@ -1,0 +1,335 @@
+// Copyright 2026 TIER IV, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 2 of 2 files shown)


takam5f2:
@a-maumau  
Could you check the PR? This PR will add the characterization test before decoupling the logic and the node.
This PR will **NEVER touch** the production code.

a-maumau (approved):
LGTM.

Note:
It is possible for the model to change the output when the model parameter get updated. So checking the model output should be only remain for the characterization test for following refactor.

takam5f2:
@a-maumau  

> So checking the model output should be only remain for the characterization test for following refactor.

Thank you for leaving the note. I'd like to keep it in mind while refactoring. Best,

PR #12949, opened 2026-07-07
fix(autoware_lidar_centerpoint): sync parameter schemas with actual node parameters

gabrielmanalu:
## Description
This package's parameter schemas had drifted from the parameters actually declared in
`node.cpp`, and part of the drift was invisible because one schema file was misnamed and
because none of these schemas mark individual leaf fields as `required` or set
`additionalProperties: false`, so a mismatch never fails validation.

1. **`centerpoint.schema.json`** (renamed from `centerpoint.schemal.json`): the typo'd
   filename doesn't match the `**/schema/*.schema.json` glob used by the
   `json-schema-check` workflow, so this schema has been silently excluded from CI.
   While unwatched:
   - `encoder_onnx_path` / `encoder_engine_path` / `head_onnx_path` / `head_engine_path`
     are declared by the node and present in `config/centerpoint.param.yaml`, but were
     missing from the schema → added.
   - `has_twist` was present, but nested under `post_process_params` — the node actually
     reads it as `model_params.has_twist`, an entirely different parameter tree, so it
     was documented in the wrong place → removed from here.

2. **`centerpoint_ml_package.schema.json`** / **`centerpoint_tiny_ml_package.schema.json`**:
   - `has_twist` and `has_variance` are declared by the node as `model_params.has_twist`
     / `model_params.has_variance` (used together in `box3DToDetectedObject`, `node.cpp`)
     but were absent from `model_params.properties` in both files → added, matching
     `centerpoint.schema.json`'s corrected location.
   - Both files document the key as `down_sample_factor`, but the node declares and reads
     `model_params.downsample_factor` (no underscore) — matching every real
     `*_ml_package.param.yaml` in this package, the C++/CUDA source, and the unit tests
     → corrected (values/defaults unchanged, including the tiny variant's `default: 2`).

`centerpoint_common.schema.json` was checked and is unaffected (covers `diagnostics.*`
only, no overlap with `model_params`).

## Related links
- `.github/workflows/json-schema-check.yaml` (glob: `**/schema/*.schema.json`)

## How was this PR tested?
- `jsonschema.validate()` of all affected `*.param.yaml` files against their corresponding
  renamed/updated schemas passes locally.
- `git grep` confirms `downsample_factor` (no underscore) is the only spelling used in
  source, tests, and all real parameter files in this package.
- `pre-commit` clean.

## Notes for reviewers
Happy to split into separate PRs per file if preferred — kept together as one PR since
all three fixes share the same root cause (schemas drifted while excluded from effective
validation). Separately noticed `centerpoint_short_range_ml_package` and
`centerpoint_sigma_ml_package` may have no schema file at all — didn't want to expand
scope here, will follow up as a separate issue if that's confirmed.

### DIFF
--- a/perception/autoware_lidar_centerpoint/schema/centerpoint.schema.dummy.json
+++ b/perception/autoware_lidar_centerpoint/schema/centerpoint.schema.dummy.json
@@ -1,6 +1,7 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
   "title": "Parameters for Lidar Centerpoint Node",
+  "$comment": "This schema is not considered in CI workflow due to a basename-prefix collision with sibling ml_package/common parameter files in this directory (see json-schema-check's glob matching in validate_json_schemas.py). See PR #12949 for details. Same pattern as centerpoint_common.schema.json's sibling, transfusion.schema.dummy.json.",
   "type": "object",
   "definitions": {
     "centerpoint": {
--- a/perception/autoware_lidar_centerpoint/schema/centerpoint.schema.json
+++ b/perception/autoware_lidar_centerpoint/schema/centerpoint.schema.json
@@ -6,6 +6,22 @@
     "centerpoint": {
       "type": "object",
       "properties": {
+        "encoder_onnx_path": {
+          "type": "string",
+          "description": "Path to VoxelFeatureEncoder ONNX file."
+        },
+        "encoder_engine_path": {
+          "type": "string",
+          "description": "Path to VoxelFeatureEncoder TensorRT Engine file."
+        },
+        "head_onnx_path": {
+          "type": "string",
+          "description": "Path to DetectionHead ONNX file."
+        },
+        "head_engine_path": {
+          "type": "string",
+          "description": "Path to DetectionHead TensorRT Engine file."
+        },
         "trt_precision": {
           "type": "string",
           "description": "TensorRT inference precision.",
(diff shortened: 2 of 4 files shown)


gabrielmanalu:
Update: renaming `centerpoint.schemal.json` → `centerpoint.schema.json` (my earlier commit)
caused `json-schema-check` to fail. Root cause: the checker matches schemas to param
files via `{schema_basename}*.param.yaml` (a prefix glob, see
`json-schema-check/validate_json_schemas.py`), and `centerpoint` is a literal prefix of
`centerpoint_common`, `centerpoint_ml_package`, `centerpoint_short_range_ml_package`,
`centerpoint_sigma_ml_package`, and `centerpoint_tiny_ml_package` — all structurally
different from `centerpoint.schema.json`. Renaming surfaced this collision; it isn't
introduced by this PR's content changes.

Checked the file's history before deciding how to proceed: this exact rename was
requested in #7421 by @kminoda
(https://github.com/autowarefoundation/autoware_universe/pull/7421#discussion_r1685897281),
and the author confirmed they'd made it
(https://github.com/autowarefoundation/autoware_universe/pull/7421#discussion_r1688113517).
That PR was later closed unmerged after unresolved conflicts and contributor inactivity,
so the rename never reached `main` — the typo isn't an intentional exclusion, it's a
two-year-old known bug that fell through on process, not on purpose.

Given the collision is real regardless of that history, I've followed the pattern already
used for the identical problem in `autoware_lidar_transfusion`
(`transfusion.schema.dummy.json`, excluded from CI on purpose with an explanatory
`$comment`, see #8205) — this finally lands what #7421 agreed on, adapted to a
constraint that a bare rename runs into today. Happy to do something different if
preferred, e.g. addressing the glob matching itself in `autoware-github-actions`.

The `centerpoint_ml_package.schema.json` / `centerpoint_tiny_ml_package.schema.json`
content fixes are unaffected and pass CI as committed.

technolojin (approved):

PR #12950, opened 2026-07-07
feat(lane_departure_checker): apply `agnocast_wrapper::Node` to `lane_departure_checker`

Koichi98:
## Description
Apply `agnocast_wrapper::Node` to `autoware_lane_departure_checker`.

### Note on `route_` / message-pointer handling
  Under `ENABLE_AGNOCAST=1`, `AUTOWARE_MESSAGE_CONST_SHARED_PTR(T)` is an agnocast smart pointer, not `std::shared_ptr<const T>`. The `boundary_departure_checker` library is not migrated and still expects `std::shared_ptr`, so we adapt at that boundary:
  - `getRouteLanelets()` now takes `const LaneletRoute &` instead of `const LaneletRoute::ConstSharedPtr &`, called via `*route_`. This decouples the helper from the pointer type with no extra copy.
  - `Input` fields (`route`, `current_odom`, `reference_trajectory`, `predicted_trajectory`) are assigned by deep-copying into a fresh `std::shared_ptr` (`x_ ? std::make_shared<T>(*x_) : nullptr`), since an agnocast pointer cannot be assigned to a `std::shared_ptr` field. Note this adds one deep copy under `ENABLE_AGNOCAST=0` that main did not have; it is unavoidable without templating `Input`.

## Related Links

## How was this PR tested?
- Build: Confirmed the node builds successfully with both `ENABLE_AGNOCAST=0` and `ENABLE_AGNOCAST=1`.
- LSim (sample rosbag): Confirmed the Logging Simulator runs to completion with the sample rosbag under both `ENABLE_AGNOCAST=0` and `ENABLE_AGNOCAST=1`, with no `process died` or similar failures.

## Interface changes

None.

## Effects on system behavior


zulfaqar-azmi-t4 (approved):
LGTM

PR #12957, opened 2026-07-08
fix(control_validator): declare angles dependency and fix include path

shtokuda:
# Summary
 - Add angles as an explicit dependency in package.xml.
 - Fix the include path for angles in control_validator.cpp.


paulsohn (approved):
Context: https://github.com/tier4/autoware_universe/pull/3135

satoshi-ota (approved):

shtokuda:
https://github.com/autowarefoundation/autoware_universe/pull/13026
で修正済みなのでクローズします

PR #12958, opened 2026-07-08
feat(diffusion_planner): synchronize odometry to tracked-objects stamp

technolojin:
## Description

`autoware_diffusion_planner` previously paired the **latest** odometry sample with the **latest** tracked-objects message independently ("newest of each"). Odometry (localization, 50 Hz) is a **faster, lower-latency stream** than tracked objects (perception, ~10 Hz plus pipeline latency), so the newest odometry available at any instant is systematically **newer** than the newest objects. The planner therefore fused an ego state and an object set that were **~37 ms apart in time**.

This PR makes the **tracked-objects message the timing anchor** and synchronizes the ego state to it, so the ego state and the object set fused into one frame share the same instant in time.

## Changes

- **Odometry buffering**: odometry subscription switched from `Newest` → `All` polling (`QoS{50}`); all samples are buffered into `ego_history_` and pruned by a time window (`ODOMETRY_BUFFER_WINDOW_S = 3.0 s`) instead of a fixed sample count.
- **Object-anchored ego syncing**: `create_frame_context` takes the odometry buffer and selects the ego state at the tracked-objects stamp (`frame_time` = object stamp). This is used only internally, to sync the current ego state and the ego-history resampling to the objects.
- **Output stamping (newest odometry, not object time)**: published outputs (trajectory / candidates / predicted_objects / turn_indicator) are stamped with the **newest odometry timestamp** (`FrameContext.output_time`), *not* the object timestamp. The planner runs on a fixed timer and the same tracked-objects message can be reused by two consecutive cycles; anchoring output stamps to the object time would then emit **duplicate, non-monotonic** stamps. The newest odometry stamp always advances, so output stamps stay strictly increasing.
- **`use_time_interpolation`** (existing param, retained) toggles how the ego state is derived at the object stamp:
  - `true` → interpolate pose (`calc_interpolated_pose`) and twist (linear lerp) between the two bracketing buffer samples, for both the current ego state and the ego-history resampling.
  - `false` → pick the nearest buffer sample (previous match-closest behavior).
- **Gap handling**: if the closest odometry is farther than `MAX_EGO_OBJECT_TIME_DIFF_S = 0.1 s` from the object stamp, proceed with a throttled warning (the time diff is surfaced via `FrameContext.ego_object_time_diff_s`; the core stays ROS-independent).
- **`ignore_neighbors`** mode (may run with no objects message) falls back to the newest odometry stamp for the ego-sync anchor as well.
- Two new `constexpr` in `constants.hpp` (`ODOMETRY_BUFFER_WINDOW_S`, `MAX_EGO_OBJECT_TIME_DIFF_S`); **no new ROS params**. `ego_history_` holds raw odometry; the `base_link → center` `shift_x` is applied only at transform / tensor-build sites.

## Verification

Same-session, same-scenario 30 s captures (sim clock) of the three published header-stamp streams, before vs. after. The metric that matters is the **time gap between the odometry and the objects fused into a single planner frame** — how far apart the two inputs the planner combined actually were.

**Topics:** `/localization/kinematic_state`, `/perception/object_recognition/tracking/objects`, `.../diffusion_planner_node/output/trajectory`.

| | Before (clean HEAD build) | After (this branch) |
|---|---|---|
| **Odom↔object time gap in the fused frame** | **~37 ms** — odometry consistently newer (35.0–39.5 ms, mean ~37) | **~0 ms** — odometry resampled/interpolated to the object stamp |
| Ego state synced to | independent (newest odom) | tracked-objects (scene) stamp |
| Output stamped at | newest odometry time | newest odometry time |

**Why ~37 ms before:** "newest of each" always grabbed an odometry sample ~37 ms ahead of the objects — the systematic latency lead of the faster odometry stream — so the planner steered on an ego state ~37 ms in front of the world it perceived.

**After the fix:** the objects are the timing anchor for the ego state — `frame_time` = object stamp and the ego state is interpolated to that same stamp, so both fused inputs share one instant (gap → 0). Output stamps remain on newest odometry so they stay monotonic even when a tracked-objects message is reused across two planner cycles.

<details>
<summary>Analysis script (<code>stamp_align_analysis.py</code>)</summary>

```python
#!/usr/bin/env python3
"""Record header stamps from odometry, tracked objects, and diffusion_planner output
trajectory, then analyze timestamp alignment.

Checks:
  1. Output trajectory stamp == the tracked-objects stamp that anchored it (new behavior),
     and NOT the odometry stamp (old behavior).
  2. Nearest-odometry-to-object matching quality (|odom - obj| should be small, <= threshold).
  3. Relative latency: odometry stamp vs object stamp.
"""
import bisect
import rclpy
from rclpy.node import Node
from rclpy.qos import QoSProfile, ReliabilityPolicy, HistoryPolicy

from nav_msgs.msg import Odometry
from autoware_perception_msgs.msg import TrackedObjects
from autoware_planning_msgs.msg import Trajectory

ODOM = "/localization/kinematic_state"
OBJS = "/perception/object_recognition/tracking/objects"
TRAJ = ("/planning/trajectory_generator/neural_network_based_planner/"
        "diffusion_planner_node/output/trajectory")

DURATION_S = 30.0
EDGE_MARGIN_S = 0.3  # trim subscription startup/teardown edges


def stamp_to_sec(stamp):
    return stamp.sec + stamp.nanosec * 1e-9


class Recorder(Node):
    def __init__(self):
        super().__init__("stamp_align_recorder")
        qos = QoSProfile(reliability=ReliabilityPolicy.RELIABLE,
                         history=HistoryPolicy.KEEP_LAST, depth=200)
        self.odom = []   # header stamps [s]
        self.objs = []
        self.traj = []
        self.create_subscription(Odometry, ODOM,
                                 lambda m: self.odom.append(stamp_to_sec(m.header.stamp)), qos)
        self.create_subscription(TrackedObjects, OBJS,
                                 lambda m: self.objs.append(stamp_to_sec(m.header.stamp)), qos)
        self.create_subscription(Trajectory, TRAJ,
                                 lambda m: self.traj.append(stamp_to_sec(m.header.stamp)), qos)


def nearest(sorted_list, value):
    """Return (nearest_value, signed_diff = nearest - value) or (None, None)."""
    if not sorted_list:
        return None, None
    i = bisect.bisect_left(sorted_list, value)
    cands = []
    if i < len(sorted_list):
        cands.append(sorted_list[i])
    if i > 0:
        cands.append(sorted_list[i - 1])
    best = min(cands, key=lambda x: abs(x - value))
    return best, best - value


def main():
    rclpy.init()
    node = Recorder()
    start = node.get_clock().now()
    print(f"Recording for {DURATION_S:.0f}s (clock={'sim' if node.get_parameter_or('use_sim_time', None) else 'default'}) ...")
    # spin until duration elapses (use wall time via a simple loop)
    import time
    t0 = time.time()
    while time.time() - t0 < DURATION_S:
        rclpy.spin_once(node, timeout_sec=0.1)

    odom = sorted(node.odom)
    objs = sorted(node.objs)
    traj = sorted(node.traj)

    # Restrict analysis to the window where ALL three streams have data, minus edge margins,
    # so subscription startup/teardown gaps do not masquerade as matching errors.
    win_lo = win_hi = None
    if odom and objs and traj:
        win_lo = max(odom[0], objs[0], traj[0]) + EDGE_MARGIN_S
        win_hi = min(odom[-1], objs[-1], traj[-1]) - EDGE_MARGIN_S
        # keep full odom/objs (needed as match candidates) but only score traj/objs inside window
        traj_w = [t for t in traj if win_lo <= t <= win_hi]
        objs_w = [t for t in objs if win_lo <= t <= win_hi]
    else:
        traj_w, objs_w = traj, objs

    def rate(lst):
        if len(lst) < 2:
            return 0.0
        span = lst[-1] - lst[0]
        return (len(lst) - 1) / span if span > 0 else 0.0

    print("\n==================== CAPTURE ====================")
    print(f"odometry        : {len(odom):4d} msgs, ~{rate(odom):5.1f} Hz, span [{odom[0] if odom else 0:.3f}, {odom[-1] if odom else 0:.3f}]")
    print(f"tracked_objects : {len(objs):4d} msgs, ~{rate(objs):5.1f} Hz, span [{objs[0] if objs else 0:.3f}, {objs[-1] if objs else 0:.3f}]")
    print(f"out trajectory  : {len(traj):4d} msgs, ~{rate(traj):5.1f} Hz, span [{traj[0] if traj else 0:.3f}, {traj[-1] if traj else 0:.3f}]")

    if not traj or not objs or not odom:
        print("\n[!] Missing data on one or more topics; cannot analyze.")
        node.destroy_node(); rclpy.shutdown(); return

    print(f"\nanalysis window (all-streams overlap, -{EDGE_MARGIN_S}s edges): "
          f"[{win_lo:.3f}, {win_hi:.3f}]  traj={len(traj_w)} objs={len(objs_w)}")

    # --- Check 1: output trajectory stamp vs nearest object stamp vs nearest odom stamp ---
    print("\n========= CHECK 1: output stamp source =========")
    print(f"{'traj_stamp':>16} | {'d(obj)[ms]':>11} | {'d(odom)[ms]':>12} | verdict")
    exact_obj = 0
    for ts in traj_w:
        _, d_obj = nearest(objs, ts)
        _, d_odom = nearest(odom, ts)
        d_obj_ms = d_obj * 1e3
        d_odom_ms = d_odom * 1e3
        is_obj = abs(d_obj_ms) < abs(d_odom_ms)
        if abs(d_obj_ms) < 0.5:
            exact_obj += 1
        print(f"{ts:16.3f} | {d_obj_ms:11.3f} | {d_odom_ms:12.3f} | "
              f"{'OBJECT-anchored' if is_obj else 'odom-anchored'}")
    print(f"\n  -> {exact_obj}/{len(traj_w)} output trajectories stamp-match a tracked-objects msg (|diff|<0.5ms)")

    # --- Check 2: nearest-odometry-to-object matching quality ---
    print("\n===== CHECK 2: nearest odom<->object diff =====")
    diffs_ms = []
    for ts in objs_w:
        _, d = nearest(odom, ts)
        diffs_ms.append(d * 1e3)
    if diffs_ms:
        import statistics
        ad = [abs(x) for x in diffs_ms]
        print(f"  per-object |nearest_odom - obj| : "
              f"min={min(ad):.2f}ms  mean={statistics.mean(ad):.2f}ms  "
              f"max={max(ad):.2f}ms  (n={len(ad)})")
        over = sum(1 for x in ad if x > 100.0)
        print(f"  objects with nearest odom > 100ms away: {over}/{len(ad)}")

    # --- Check 3: relative latency of odom vs object stamps ---
    print("\n===== CHECK 3: signed(nearest_odom - obj) =====")
    if diffs_ms:
        pos = sum(1 for x in diffs_ms if x > 0)
        neg = sum(1 for x in diffs_ms if x < 0)
        print(f"  nearest odom AHEAD of object (odom>obj): {pos}, BEHIND: {neg}")
        print(f"  (odometry lower latency => samples typically bracket/lead the object stamp)")

    node.destroy_node()
    rclpy.shutdown()


if __name__ == "__main__":
    main()
```

</details>

---


technolojin:
https://github.com/autowarefoundation/autoware_universe/pull/12979 replaces this PR

PR #12961, opened 2026-07-08
fix(diffusion_planner): filter neighbors by shape type and rename ignore_unknown_neighbors to ignore_polygon_neighbors

technolojin:
## Description

Neighbor-agent filtering in the diffusion planner previously keyed off the object **classification** (`UNKNOWN`). That is fragile: whether an agent is usable for the model actually depends on whether it has a real footprint, which is reflected by the tracker **shape type**, not by its class label.

This PR reworks the filter to be shape-based and renames the parameter accordingly.

### Changes

- **Rename** `ignore_unknown_neighbors` → `ignore_polygon_neighbors` across the param file, JSON schema, core params struct, node param declaration/`on_parameter`, and the integration test.
- **Filter by shape, not class**: replace `is_unknown_object()` (checks highest-prob label `== UNKNOWN`) with `is_polygon_object()` (checks `shape.type == POLYGON`). The tracker output shape type is the robust signal for shapeless polygon agents.
- **Per-shape width/length resolution** (`get_width_and_length()`): each shape type populates the shape message differently, so resolve `{width, length}` accordingly:
  - `BOUNDING_BOX`: `{dimensions.y, dimensions.x}`
  - `CYLINDER`: diameter (`dimensions.x`) for both axes
  - `POLYGON`: derive extent from the footprint polygon bounds (0 if empty)

### Behavior

- Agents whose shape is a polygon (shapeless, no dimensions) are ignored when the flag is enabled, instead of only those classified `UNKNOWN`.
- Cylinder/polygon agents that are kept now get sensible width/length instead of raw (possibly zero) bounding-box dimensions.

## Tests performed

- Existing integration test updated to the new parameter name.

## Notes

Draft: parameter rename is API-visible — downstream launch/config referencing `ignore_unknown_neighbors` must be updated.

### DIFF
--- a/planning/autoware_diffusion_planner/config/diffusion_planner.param.yaml
+++ b/planning/autoware_diffusion_planner/config/diffusion_planner.param.yaml
@@ -24,7 +24,7 @@
         start_time_s: 2.0
     planning_frequency_hz: 10.0
     ignore_neighbors: false
-    ignore_unknown_neighbors: true
+    ignore_polygon_neighbors: true
     traffic_light_group_msg_timeout_seconds: 0.2
     batch_size: 1
     temperature: [0.0]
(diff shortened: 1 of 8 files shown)


technolojin:
replaced by https://github.com/autowarefoundation/autoware_universe/pull/12900

PR #12963, opened 2026-07-08
feat(control_evaluator): apply `agnocast_wrapper::Node` to `control_evaluator`

Koichi98:
## Description
Apply `agnocast_wrapper::Node` to `autoware_control_evaluator`.

## Related Links

## How was this PR tested?
  - Build: `colcon build` succeeds with both `ENABLE_AGNOCAST=1` and `ENABLE_AGNOCAST=0`.
  - Unit tests: 9/9 pass with `ENABLE_AGNOCAST=0`; all 9 are skipped when `ENABLE_AGNOCAST=1` (node-launching tests, per the `autoware_agnocast_wrapper` unit-test policy). The test logic is unchanged.
  - LSim/PSim (sample rosbag): Runs to completion under both `ENABLE_AGNOCAST=0` and `ENABLE_AGNOCAST=1`, with no `process died` or similar failures.

## Interface changes

None.

## Effects on system behavior

### DIFF
--- a/evaluator/autoware_control_evaluator/CMakeLists.txt
+++ b/evaluator/autoware_control_evaluator/CMakeLists.txt
@@ -10,9 +10,11 @@ ament_auto_add_library(control_evaluator_node SHARED
   DIRECTORY src
 )
 
-rclcpp_components_register_node(control_evaluator_node
+autoware_agnocast_wrapper_register_node(control_evaluator_node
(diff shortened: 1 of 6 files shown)


kobayu858:
Made a mistake while trying to fix the DOC. Currently fixing it.

xtk8532704 (approved):

PR #12965, opened 2026-07-08
refactor(detection_by_tracker): isolate subscription callback groups

kobayu858:
## Description
Currently, the introduction of the Callback Isolated Executor into Autoware is underway.
To successfully achieve DAG (Directed Acyclic Graph) Scheduling via the Callback Isolated Executor, each callback recognized as a vertex in the DAG must belong to its own independent CallbackGroup.
This enables a one-to-one mapping between CallbackGroups and OS threads, allowing for fine-grained control over Priority and Affinity on a per-OS-thread basis.
Since isolating all callbacks at once involves a significantly high implementation cost, we are adopting a phased approach—starting the modifications and application with the major nodes that have the highest impact.

**Detailed Changes**

1. Introduction of Independent Callback Groups:

- Newly added `cb_group_trackers_` and `cb_group_objects_` as member variables in `DetectionByTracker`.
- Assigned these mutually exclusive callback groups to the respective subscription options for `trackers_sub_` (standard ROS 2 subscription) and `initial_objects_sub_` (via AUTOWARE_SUBSCRIPTION_OPTIONS for the Autoware macro).

2. Thread-Safety and Minimization of Lock Scopes:

- Introduced `std::mutex mutex_` to safely protect the shared `objects_buffer_` across the now-decoupled concurrent callback groups.
- Implemented mutual exclusion control using `std::lock_guard<std::mutex>` in both the data-receiving side (`onTrackedObjects`) and the estimation side (`estimateTrackedObjects`) to prevent data races and collisions.
- Optimized `estimateTrackedObjects` so that only the buffer min-element search and the copying of objects to a local variable occur inside the locked critical section. This allows the computationally heavy pose estimation loop (for loop) to execute entirely outside the lock, minimizing lock holding time, thread contention, and execution delays.

## Related links

**Parent Issue:** [Separation of CallbackGroups: Background](https://tier4.atlassian.net/wiki/spaces/~5ed0b4584824b20c18371c06/pages/3009052751/CallbackGroup)

## How was this PR tested?

I ran `lsim` locally using a ros2bag.
I also measured the latency using CARET and confirmed that there were no significant differences.
The purpose of this PR is to enable scheduler configuration by isolating callback groups, not to reduce latency.

## Notes for reviewers

- [ ] Verification of Callback Group Creation and Assignment: Ensure Callback Groups (`cb_group_trackers_` and cb_group_objects_) are correctly created in the constructor and properly assigned to their respective SubscriptionOptions and AUTOWARE_SUBSCRIPTION_OPTIONS.
- [ ] Appropriate Scope of Mutex Locks: Verify that the scope of mutex locks (`std::lock_guard<std::mutex>`) within `onTrackedObjects` and `estimateTrackedObjects` is appropriate, particularly ensuring that the lock in estimateTrackedObjects is released before entering the pose estimation loop to eliminate any risk of deadlocks or execution delays.

## Interface changes

None.

## Effects on system behavior

None.


technolojin:
this node is removed from perception pipeline https://github.com/autowarefoundation/autoware_launch/issues/1846, therefore, not maintained, planned to be removed.

are you still request to use this node? please let me know.

technolojin:
I could not test this node. If you tested, please provide evident that the changed implementation still works within the autoware system.

kobayu858:
@technolojin 
Thank you for letting me know. I wasn’t aware of that.
If the node is indeed no longer in use, there is no need to address it.

PR #12966, opened 2026-07-08
feat(planning_evaluator): apply `agnocast_wrapper::Node` to `planning_evaluator`

Koichi98:
## Description
Apply `agnocast_wrapper::Node` to `autoware_planning_evaluator`.

## Related Links

## How was this PR tested?
- Build: Confirmed the node builds successfully with both `ENABLE_AGNOCAST=0` and `ENABLE_AGNOCAST=1`.
- LSim (sample rosbag): Confirmed the Logging Simulator runs to completion with the sample rosbag under both `ENABLE_AGNOCAST=0` and `ENABLE_AGNOCAST=1`, with no `process died` or similar failures.
- Unit tests: Confirmed the unit tests pass with `ENABLE_AGNOCAST=0`. The node-launching test is skipped when `ENABLE_AGNOCAST=1` (per the `autoware_agnocast_wrapper` unit-test policy). The test logic is unchanged — only `rclcpp::spin_some(eval_node)` was adapted to `rclcpp::spin_some(eval_node->get_node_base_interface())` to match the `agnocast_wrapper::Node` interface.

## Interface changes

None.

## Effects on system behavior


xtk8532704 (approved):
LGTM

PR #12967, opened 2026-07-08
refactor(autoware_traffic_light_classifier): extract a Node-free CNNClassifierCore

takam5f2:
## Description

This PR extracts a **Node-free `CNNClassifierCore`** from `CNNClassifier`, mirroring the `ColorClassifierCore` split in #12916. The classification logic can now be constructed from a plain `CNNConfig` and its pure helpers called directly — no node, no parameters, no RMW. The change is scoped to `src/classifier/cnn_classifier.{hpp,cpp}` (plus the CMake/test wiring):

- **`CNNClassifierCore`** depends only on TensorRT + OpenCV + `tier4_perception_msgs` (no `rclcpp`, `image_transport`, `cv_bridge`, or logging). Constructed from a `CNNConfig` (pre-read label lines + model path + precision + mean/std); return-based API `ClassifierResult classify(const std::vector<cv::Mat> &)`. The engine build stays in the constructor, but the two pure helpers `decode_label` and `make_debug_image` are `static`, so they run without a GPU or a model.
- **`CNNClassifier`** remains a thin ROS adapter with an **unchanged public API** (same `rclcpp::Node *` constructor and `getTrafficSignals` override). It keeps the ROS-facing concerns: parameter declaration, label-file reading, `~/output/debug/image` publishing, logging, the images/signals size guard, and element merging (preserving upstream　traffic-light id/type).

`ClassifierInterface`, the color backend, and the node files are intentionally left untouched.

## Related links

**Parent Issue:**

- None

Sibling refactor (same package, color backend): #12916

## How was this PR tested?

- `colcon build --packages-select autoware_traffic_light_classifier` → success
- `colcon test --packages-select autoware_traffic_light_classifier`
  - **`test_cnn_classifier_core`** (new, ROS-free) → 8/8 pass
  - **`test_cnn_classifier`** (existing GPU characterization, kept **unchanged**) → 6/6 pass, confirming the refactor preserves behavior
- `clang-format` / `cpplint` clean (pre-commit)

## Notes for reviewers

- Scope is intentionally minimal and self-contained. `CNNClassifier`'s public API and the `ClassifierInterface` contract are unchanged, so all call sites compile without modification.
- The new `test_cnn_classifier_core` is gated on `TRT_AVAIL AND CUDA_AVAIL` because it links `${PROJECT_NAME}` (the only config that compiles `cnn_classifier.cpp`), but it exercises only the static GPU-free helpers, so it needs no GPU at run time — unlike `test_cnn_classifier`.

## Interface changes

None. `CNNClassifier`'s public API and the `ClassifierInterface` abstract interface are unchanged.

## Effects on system behavior

Behavior-preserving on the success path (verified by the unchanged characterization test).
One intentional hardening, a direct consequence of moving construction into a Node-free core that signals failure by throwing rather than by leaving itself half-built:

- A **missing/unreadable label file** now throws `std::runtime_error` at construction. Before, `readLabelfile`'s return value was ignored, leaving `labels_` empty and causing an out-of-range `labels_[class_index]` at inference time.
- A **wrong-size `mean`/`std`** now throws `std::invalid_argument`. Before, the constructor logged and returned early, leaving `classifier_` null and faulting at inference time.

Both replace a latent silent-failure/UB path with fail-fast node construction.

### DIFF
--- a/perception/autoware_traffic_light_classifier/CMakeLists.txt
+++ b/perception/autoware_traffic_light_classifier/CMakeLists.txt
@@ -169,10 +169,21 @@ if(BUILD_TESTING)
     ${OpenCV_LIBRARIES}
   )
 
-  # GPU/TensorRT characterization test for CNNClassifier (only compiled under
-  # TRT+CUDA, hence the gate). It builds a node, so it uses the ROS-isolated runner;
-  # TIMEOUT covers the first-time TensorRT engine build.
+  # Tests that depend on TensorRT/CUDA code, built only when those libraries are available.
   if(TRT_AVAIL AND CUDA_AVAIL)
+    # ROS-free unit tests for CNNClassifierCore's static helpers (label decode +
+    # debug-image geometry) -- no GPU needed at run time, unlike test_cnn_classifier.
+    ament_auto_add_gtest(test_cnn_classifier_core
+      test/test_cnn_classifier_core.cpp
+    )
+    target_link_libraries(test_cnn_classifier_core
+      ${PROJECT_NAME}
+      ${OpenCV_LIBRARIES}
+      ${CUDA_LIBRARIES}
+    )
+
+    # GPU/TensorRT characterization test for CNNClassifier. It builds a node, so it uses
+    # the ROS-isolated runner; TIMEOUT covers the first-time TensorRT engine build.
     ament_add_ros_isolated_gtest(test_cnn_classifier
       test/test_cnn_classifier.cpp
       TIMEOUT 600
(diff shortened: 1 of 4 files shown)


takam5f2:
@a-maumau  
Could you review the PR? 
This PR will change the exception behavior. A Node-free core can't sensibly leave itself half-built, so it reports construction failures by throwing:
- Missing/unreadable label file → now throws at construction (previous: silently ignored → empty label table → out-of-range crash at inference).
- Wrong-size mean/std → now throws at construction (previous: logged + early return → null classifier → crash at inference).

a-maumau (approved):
LGTM

PR #12974, opened 2026-07-09
test(autoware_traffic_light_classifier): add CnnLampRecognizer characterization test

takam5f2:
## Description

Adds a characterization test that pins the **current** end-to-end behavior of `CnnLampRecognizer::getTrafficSignals` against the real lamp-recognizer TensorRT model, as a safety net **before** a planned Node-free core/adapter split of `cnn_lamp_recognizer.{hpp,cpp}` (mirroring the split already landed for `CNNClassifier` in #12967).

Following the sibling `test/test_cnn_classifier.cpp`, the test:

- loads the real model (`traffic_light_lamp_recognizer_comlops.onnx`) from `autoware_data`;
- self-skips (`GTEST_SKIP`) when the GPU or model is unavailable;
- builds the TensorRT engine once per suite (that build takes minutes);
- asserts coarsely — **shape** pinned exactly, **color** pinned as-observed, **confidence** only bounded — because TensorRT output is not bit-reproducible across GPU / TRT versions.

It is compiled only under `TRT_AVAIL AND CUDA_AVAIL` and uses the ROS-isolated runner (it constructs a node). `model_params.*` are hard-coded from `lamp_recognizer_ml.param.yaml`; the artifact source URLs are referenced in comments.

This is a **test-only** change; no production code is touched.

## How was this PR tested?

`colcon test` on RTX 5080 / TRT 10.8 (`rmw_fastrtps_cpp` + `ROS_LOCALHOST_ONLY=1`): `test_cnn_lamp_recognizer` **6/6 pass with real inference**, rest of the package green. `clang-format` / `cpplint` clean. On machines without the GPU/model the suite self-skips.

## Notes for reviewers

The behavior of `CnnLampRecognizer` is different from that of `CNNClassifier`. It is not blocking problem for refactoring because the characterization test aims to hold the behavior of `CnnLampRecognizer`.

## Effects on system behavior

None (test-only change).


takam5f2:
@a-maumau  
Could you check the PR? This PR will add only characteristic test for `CnnLampRecognizer` to isolate the logic and the glue code.

a-maumau (approved):
LGTM.

I was wondering about adding some dummy classifier to test the all path of the logic.
For me, dummy classifier for testing is acceptable to add in the source code.

PR #12975, opened 2026-07-09
test(autoware_traffic_light_classifier): make CNN classifier core test ROS-free

takam5f2:
## Description

Rewrite the CNN classifier core tests to run without the ROS layer, mirroring the color classifier's core test. Tests + CMake only — no production code changes. Follows up on #12967 (the Node-free `CNNClassifierCore` extraction).

- **`test_cnn_classifier.cpp`** now tests `CNNClassifierCore::classify()` directly (constructed from a `CNNConfig`, no `rclcpp::Node`). This drops the RMW/DDS dependency the previous adapter-level test carried, and its associated flakiness. It stays GPU-gated and self-skips (`GTEST_SKIP`) when the GPU or model is unavailable. The static-helper tests (`decode_label` / `make_debug_image`) previously in `test_cnn_classifier_core.cpp` are folded in, so one file now covers all of `CNNClassifierCore` and the naming matches the color classifier.
- The classification tests are consolidated to two GPU-gated checks — `ClassifiesNormalCropToWellFormedElement` (one crop) and `BatchClassificationYieldsResultPerImage` (the same crop in two batch slots) — and the redundant per-dimming-level variants (weak/medium/strong) were dropped. Following the review note on #12940 (model output can change when the model parameters are updated), these assert only the core's *model-stable* output contract — `success`, result count, one element per signal, confidence in `(0, 1]`, and identical input → identical output — and deliberately do **not** pin the decoded color/shape. That label mapping is covered model-free by the `decode_label` tests, so pinning a specific class against the live model would only make the tests brittle to a model update without adding coverage.

The `getTrafficSignals` images/signals size guard, which needs a node (a ROS adapter concern), is intentionally out of scope here and will be added in a follow-up as a small ROS adapter test.

## Related links

None.

## How was this PR tested?

Environment: ROS 2 Humble, NVIDIA RTX 5080 (CUDA + TensorRT), CNN model + labels present under `~/autoware_data`.

Build the package and run the CNN core test target:

```bash
cd <autoware_workspace>
source install/setup.bash

# compiles the gtest executable and runs the CNN core test target
colcon test --packages-select autoware_traffic_light_classifier \
  --ctest-args -R cnn_classifier
colcon test-result --test-result-base \
  build/autoware_traffic_light_classifier/test_results --verbose
```

Result: `test_cnn_classifier` — 10 tests pass (8 static-helper + 2 GPU-gated `classify()`). The `classify()` tests ran against the real MobileNet-v2 engine, exercising the full preprocess → TensorRT → decode plumbing of `CNNClassifierCore`.

`pre-commit run` (clang-format, cpplint, …) passes on the changed files.

## Notes for reviewers

The two GPU-gated `classify()` tests were reworked to address the review note on #12940: since the model output can shift when the model parameters change, they no longer pin the decoded color/shape and assert only the model-stable output contract.

## Interface changes

None.

## Effects on system behavior

None.


takam5f2:
@a-maumau  
Could you check the PR? This PR will only change the test code rather than the production code.

As you told at https://github.com/autowarefoundation/autoware_universe/pull/12940, the test code will not check the detected color because they depends on the ML model.

a-maumau (approved):
Sorry for my unclear explanation in the last PR. I should make it more clear.

LGTM!

PR #12976, opened 2026-07-09
test(autoware_traffic_light_classifier): add CNN adapter size-guard test

takam5f2:
## Description

Follow-up to #12975 (now merged). Adds the ROS adapter test for `CNNClassifier` that #12975 intentionally left out of scope, restoring the guard coverage that existed before that restructure.

`test_cnn_classifier_adapter.cpp` pins the one adapter concern that cannot move to the Node-free core: the `getTrafficSignals` images/signals size guard. The guard protects the element-merge loop from an out-of-range access when the caller's signal count disagrees with the images, so a regression here is a memory-safety bug, not just a wrong return value.

Constructing `CNNClassifier` builds the TensorRT engine (unlike `ColorClassifier`, whose ctor is cheap), so the test runs under the ROS-isolated runner and self-skips (`GTEST_SKIP`) without a GPU/model. The naming mirrors the color classifier (`test_X` / `test_X_adapter`).

## Related links

- Follow-up to #12975 (merged)

## How was this PR tested?

Environment: ROS 2 Humble, NVIDIA RTX 5080 (CUDA + TensorRT), CNN model + labels present under `~/autoware_data`.

Under the default RMW (CycloneDDS on a loopback-only host) node creation fails, so the test self-skips via `GTEST_SKIP`. To actually exercise the guard, build the package and run the adapter binary directly with the FastRTPS workaround:

```bash
cd <autoware_workspace>
source install/setup.bash

RMW_IMPLEMENTATION=rmw_fastrtps_cpp ROS_LOCALHOST_ONLY=1 \
  ./build/autoware_traffic_light_classifier/test_cnn_classifier_adapter
```

Result: `MismatchedImageSignalCountReturnsFalse` passes (`[ OK ]`), with the expected `image number should be equal to traffic signal number!` warning confirming the guard path is taken.

`pre-commit run` (clang-format, cpplint, …) passes on the changed files.

## Notes for reviewers

The test is GPU-gated because the `CNNClassifier` ctor builds the TensorRT engine; it self-skips cleanly without a GPU/model. It also self-skips under CycloneDDS on a loopback-only host (node creation fails) — see the workaround above to run it locally.

## Interface changes

None.

## Effects on system behavior

None.


takam5f2:
@a-maumau  
Could you check the PR? This PR only adds a test code without changing the production code. This test code only checks if ROS-based interface for `CNNClassifier` works properly.

a-maumau (approved):
LGTM

PR #12978, opened 2026-07-10
refactor(pid_longitudinal_controller): extract smooth stop clock

interimadd:
## Description

Removes `SmoothStop`'s hidden dependency on the ROS clock and its unconfigured-state invalid path, both of which hurt testability.

- `SmoothStop::init`, `calcTimeToStop`, and `calculate` previously called `rclcpp::Clock{RCL_ROS_TIME}.now()` internally, coupling this pure deceleration-profile logic to the ROS runtime and forcing tests to sleep on the real wall clock to exercise its time-based branches (e.g. the 0.5s post-stop hold). 
- `SmoothStop::Params` is now a public struct required by the constructor, so an unconfigured `SmoothStop` can no longer be constructed. This removes the `m_is_set_params` flag and the two `std::runtime_error` throws in `calculate()`/`calcTimeToStop()` that previously guarded against calling them before `setParams()`.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

```bash
colcon build --packages-select autoware_pid_longitudinal_controller
colcon test --packages-select autoware_pid_longitudinal_controller --event-handlers console_direct+
```

All existing tests, including `TestSmoothStop.calculate_stopping_acceleration`, pass. The 0.5s-window portion of that test now runs in 0ms instead of ~750ms of real sleep.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/control/autoware_pid_longitudinal_controller/include/autoware/pid_longitudinal_controller/pid_longitudinal_controller.hpp
+++ b/control/autoware_pid_longitudinal_controller/include/autoware/pid_longitudinal_controller/pid_longitudinal_controller.hpp
@@ -185,7 +185,7 @@ private:
   double m_brake_keeping_acc;
 
   // smooth stop
-  SmoothStop m_smooth_stop;
+  std::optional<SmoothStop> m_smooth_stop;
 
   // stop
   struct StoppedStateParams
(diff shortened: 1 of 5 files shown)


interimadd:
I will submit small PRs to the refactoring branch, and then open a final large PR from there to the main branch.

PR #12979, opened 2026-07-10
feat(diffusion_planner): resample neighbor agent histories onto a constant grid (motion compensation)

technolojin:
## Description

The `neighbor_agents_past` input tensor carries no timestamp / dt / mask channel, so the diffusion planner network implicitly assumes each neighbor history lies on a **uniform 0.1 s grid**. In reality the per-UUID histories buffered from `TrackedObjects` are jittered in time, latent, and start/stop at arbitrary ticks. Feeding those raw buffers means the model reads motion at the wrong effective dt.

This PR adds an **object motion resampling** step that re-times every retained neighbor history onto an odometry-anchored constant 0.1 s grid (newest sample at the ego stamp) before the histories are transformed, sorted, and trimmed for the model. This is the "motion compensation at DP (diffusion planner)" work.

object_history_resampling

### What it does

- **New module `agent_history_resampler`** (`conversion/agent_history_resampler.{hpp,cpp}`), replacing the earlier `agent_history_alignment` prototype:
  - `resample_history()` re-times one history onto the 0.1 s grid anchored at `frame_time` (oldest first, newest at `frame_time`): **interpolates within** observed samples, **extrapolates live agents** past the newest observation, and extrapolates backward from the oldest observation to fill the pre-appearance grid.
  - `propagate_motion()` — a constant-turn-rate (CTRV) core that sub-steps forward/backward with a fixed sub-step to bound forward-Euler error; extrapolation is clamped to `max_extrapolation_time`.
  - `interpolate_yaw()` — shortest-arc heading interpolation that never sweeps the ±π discontinuity.
- **`AgentData` becomes buffer-lifecycle only:**
  - `update_histories()` dedups on the object stamp (the neighbor subscriber polls the latest message, so a stale message can repeat across ticks — only strictly-advancing stamps are ingested), starts new UUIDs from a single real observation (no repeat-fill), and **erases disappeared agents immediately**.
  - `resampled_transformed_and_trimmed_histories()` runs resampling, then the shared `transform_sort_trim()` helper (nearest-first by distance to frame origin, trimmed to `max_num_agent`).
- **Wiring in `diffusion_planner_core`:** when `object_motion_resampling.enable` is true, neighbor histories are anchored to the ego stamp and resampled; otherwise the legacy buffered path is used unchanged.

### Parameters

|:--|:--|:--|:--|:--|
| Added | `object_motion_resampling.enable` | `bool` | `true` | Re-time neighbor histories onto a constant 0.1 s grid anchored at the ego stamp. |
| Added | `object_motion_resampling.max_extrapolation_time` | `double` | `0.5` | Max extrapolation horizon [s], both forward from the newest and backward from the oldest observation. |

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

- New unit tests:
  - `agent_history_resampler_test.cpp` — CTRV propagation (forward/backward/clamping), shortest-arc yaw interpolation, and grid resampling behavior.
  - `agent_lifecycle_test.cpp` — history buffer lifecycle: stamp dedup, new-UUID seeding, immediate erasure of disappeared agents.
- `colcon test` for `autoware_diffusion_planner` (all green locally).

## Notes for reviewers

- The feature is guarded by `object_motion_resampling.enable` (default `true`). Setting it `false` restores the exact legacy buffered path.
- `agent_history_alignment` (the earlier prototype) is fully removed in favor of `agent_history_resampler`.

## Interface changes

Two new ROS parameters (see table above). No topic changes.

## Effects on system behavior

Neighbor agent histories fed to the diffusion planner are now resampled onto a constant 0.1 s grid anchored at the ego stamp, giving the network the temporally consistent input it assumes. When `object_motion_resampling.enable` is `false`, behavior is unchanged.


SakodaShintaro:
@technolojin 
Thank you for the pull request. I have taken a quick look, and this approach seems fine. I will conduct a proper review and a rosbag based test later.

I have two questions:

- Is there a corresponding pull request for `autoware_launch`?
- Although the data distribution might differ slightly from when the model was originally trained, the data is essentially being cleaned up. Is it correct to assume that it will work even without retraining?

technolojin:
@SakodaShintaro 

1. No. recent autoware_launch syncs new parameters by ci (ex. https://github.com/autowarefoundation/autoware_launch/pull/1929)
2. In my guess, maybe not (no need retraining). 
  * If current DP has more wight by simulated data, I think it dose not need re-training. 
  * If the model is trained more by driving data, may DP learned timing jittering. better to re-train with this pre-process. It dose not require additional training data cleaning or re-recording.

SakodaShintaro (approved):
I have confirmed that `autoware_perception_reproducer` works well.

https://github.com/user-attachments/assets/d26bda1b-f641-4f7d-ab36-79990e68f86a

LGTM

PR #12983, opened 2026-07-13
refactor(autoware_pid_longitudinal_controller): refactor smooth stop

interimadd:
## Description

Removes `SmoothStop`'s hidden dependencies on the ROS clock and `DebugValues`, and its unconfigured-state invalid path, all of which hurt testability.

- `SmoothStop::Params` is now a public struct required by the constructor, so an unconfigured `SmoothStop` can no longer be constructed. This removes the `m_is_set_params` flag and the two `std::runtime_error` throws.
- `SmoothStop::init` previously called `rclcpp::Clock{RCL_ROS_TIME}.now()` internally. It now takes the current time as an explicit `rclcpp::Time` parameter.
- Ownership of the velocity history used to predict the time to stop moves from `PidLongitudinalController::m_vel_hist` into `SmoothStop` itself.
- `SmoothStop::calculate()` now returns a `Result{acc, mode}` instead of writing the selected mode into a `DebugValues` reference.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

```bash
colcon build --packages-select autoware_pid_longitudinal_controller
colcon test --packages-select autoware_pid_longitudinal_controller --event-handlers console_direct+
```

All existing tests, including `TestSmoothStop.calculate_stopping_acceleration`, pass. The 0.5s-window portion of that test now runs in 0ms instead of ~750ms of real sleep.

[Evalautor Link (TIER IV internal)](https://evaluation.ci.tier4.jp/evaluation/reports/8ee1bc6e-10af-552c-aef4-ebac81b6244c/?project_id=autoware_dev)

## Notes for reviewers

This PR is made of 3 commits, each a self-contained refactoring step. Reviewing them individually (rather than the squashed diff) is recommended, as each is easier to understand in isolation:

1. https://github.com/autowarefoundation/autoware_universe/pull/12983/commits/47f3ccbe356d0eae0820295e3ccaa6126e6b08f3 require SmoothStop::Params at construction
2. https://github.com/autowarefoundation/autoware_universe/pull/12983/commits/b9bea1608f11fa1a0cca0ddf782dfa84a067c200 decouple SmoothStop from the ROS clock and encapsulate its velocity history
3. https://github.com/autowarefoundation/autoware_universe/pull/12983/commits/f5566c4da567f6bc188be1cdd815add04479047c simplify calculate() and return SmoothStop's mode instead of writing to DebugValues

## Interface changes

None.

## Effects on system behavior

None.


Kotakku (approved):
LGTM

PR #12988, opened 2026-07-13
feat(trajectory_processor): unify parameters handling

maxime-clem:
## Description

This PR unifies the parameter handling between the `trajectory_modifier` and `trajectory_optimizer`, switching to using the `declare_parameters` approach based on a `parameters_struct.yaml` file. The `trajectory_modifier` was already using this approach.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


planning/autoware_trajectory_processor/src/trajectory_optimizer_plugins/trajectory_mpt_optimizer.cpp
```diff
@@ -52,7 +49,28 @@ void TrajectoryMPTOptimizer::initialize(
     debug_data_ptr_ = std::make_shared<DebugData>();
 
     // Set up parameters
-    set_up_params();
+    enabled_ = params.use_mpt_optimizer;
+    mpt_params_.corridor_width_m = params.trajectory_mpt_optimizer.corridor_width_m;
```
mkquda, line 53: why not just define a member `trajectory_optimizer_node_params::Params::TrajectoryMPTOptimizer mpt_params_` ?

If you need to break down subsections its better to do so in the parameter struct yaml in the first place.

planning/autoware_trajectory_processor/src/trajectory_optimizer_plugins/trajectory_eb_smoother_optimizer.cpp
```diff
@@ -52,25 +52,10 @@ void TrajectoryEBSmootherOptimizer::set_up_params()
...
-    ego_nearest_param_.onParam(parameters);
-
-    // parameters for trajectory
-    common_param_.onParam(parameters);
-
-    // parameters for core algorithms
-    eb_path_smoother_ptr_->onParam(parameters);
-    eb_path_smoother_ptr_->initialize(false, common_param_);
-    eb_path_smoother_ptr_->resetPreviousData();
-  }
```
mkquda, line 68: I guess we will lose the ability to update core algorithm params in runtime, but I think its acceptable for now, I doubt we ever need to do it anyway.

planning/autoware_trajectory_processor/src/trajectory_optimizer_plugins/trajectory_mpt_optimizer.cpp
```diff
@@ -79,107 +97,68 @@ void TrajectoryMPTOptimizer::initialize(
...
+        "trajectory_mpt_optimizer.output_backward_traj_length_m",
+        traj_param_.output_backward_traj_length),
+      rclcpp::Parameter(
+        "trajectory_mpt_optimizer.ego_nearest_dist_threshold_m", ego_nearest_param_.dist_threshold),
+      rclcpp::Parameter(
+        "trajectory_mpt_optimizer.ego_nearest_yaw_threshold_deg",
+        params.trajectory_mpt_optimizer.ego_nearest_yaw_threshold_deg),
+      rclcpp::Parameter(
+        "trajectory_mpt_optimizer.acceleration_moving_average_window",
+        mpt_params_.acceleration_moving_average_window)};
```
mkquda, line 139: I don't like the idea that core algorithms/classes are ros dependent, but its okay for now we can change it later in cleanup PR.
But maybe just make a task so that we don't forget.

planning/autoware_trajectory_processor/src/trajectory_optimizer_plugins/trajectory_point_fixer.cpp
```diff
@@ -25,19 +25,16 @@
...
     return;
   }
   auto & semantic_speed_tracker = data.semantic_speed_tracker;
   trajectory_point_fixer_utils::remove_invalid_points(traj_points);
 
-  if (fixer_params_.remove_close_points) {
-    trajectory_point_fixer_utils::remove_close_proximity_points(
-      traj_points, semantic_speed_tracker, fixer_params_.min_dist_to_remove_m);
-  }
+  trajectory_point_fixer_utils::remove_close_proximity_points(
```
mkquda: Was it intentional to remove the condition `if (fixer_params_.remove_close_points) {`?

mkquda (approved):
LGTM.

planning/autoware_trajectory_processor/src/trajectory_optimizer_plugins/trajectory_eb_smoother_optimizer.cpp
maxime-clem: Will be fixed in the future: https://tier4.atlassian.net/browse/PDDP-401

planning/autoware_trajectory_processor/src/trajectory_optimizer_plugins/trajectory_mpt_optimizer.cpp
maxime-clem: Ticket: https://tier4.atlassian.net/browse/PDDP-400

planning/autoware_trajectory_processor/src/trajectory_optimizer_plugins/trajectory_mpt_optimizer.cpp
maxime-clem: You are right I could simplify it: 79712d69889

planning/autoware_trajectory_processor/src/trajectory_optimizer_plugins/trajectory_point_fixer.cpp
maxime-clem: Nice catch ! It was removed by mistakes because the parameter was missing from the schema.
Fixed: dda1fcb1b97

PR #12990, opened 2026-07-14
fix(autoware_multi_object_tracker): lock wheelbase length in partial wheel update to stabilize stationary vehicles

technolojin:
## Description

Stationary vehicles can spin up a spurious yaw when only a **partial cluster** is observed
(e.g. the bbox detection channel drops and only one axle face is seen). The wheel/axle-anchor
update in `BicycleMotionModel::updateStatePoseWheel` measures only the `(x, y)` of a single axle
face center. A partial cluster that overshoots that face **stretches the body length** — the
distance between the two axle points (`X1,Y1` ↔ `X2,Y2`).

Because the bicycle model derives yaw from the axle difference vector

```
yaw = atan2(Y2 - Y1, X2 - X1)
```

a stretched wheelbase acts as a lever: the "enlarged front wheel" is converted into rotation
instead of length. This **bypasses the bicycle motion model's length constraint** and rotates an
otherwise static vehicle.

This PR enforces the object length during the partial update and cleans up the covariance export
so the two axle points stay consistent.

**1. Wheelbase-lock row in the partial wheel update** (primary fix)

`updateStatePoseWheel` gains a third measurement row (`DIM_Y` 2 → 3) that pins the wheelbase to
its current value:

- rows 0,1 — unchanged: face center as the exact linear blend of the two endpoints
  (`face = (1 + γ)·p_near − γ·p_far`)
- row 2 — **wheelbase lock**: projects `(p2 − p1)` onto the current heading `t̂ = (p2 − p1)/L`
  and measures it back to its current value, with a tight `R = 1e-4 m²` (~1 cm std).

The pre-update wheelbase and heading are derived straight from the axle points with a single
`hypot` (no trig). With the length held near-constant, a partial cluster can no longer stretch the
body, so it cannot leak into yaw through the wheelbase lever.

**2. Jacobian-based pose covariance export** in `getPredictedState`

The exported `(x, y, yaw)` covariance block is now the axle covariance `[X1,Y1,X2,Y2]` propagated
through a single `3×4` Jacobian `G` of `[center, yaw] = f(X1,Y1,X2,Y2)`:

- position rides the **center blend** `w_rear·p1 + w_front·p2` instead of copying a single axle
  point;
- yaw uses the difference-vector Jacobian `d(yaw)/d[X1,Y1,X2,Y2] = (1/L)·[sin, −cos, −sin, cos]`.

`M = G · P_sub · Gᵀ` reproduces the exact linearized `YAW_YAW`, restores the position↔yaw
cross-covariance that the previous single-axle-point copy dropped, and is PSD by construction.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

Adds `ExportedPoseCovariance` cases in `test/test_vehicle_tracker.cpp`:

- `MatchesJacobianPropagation` — exported block equals `G · P_sub · Gᵀ` rebuilt from the raw axle
  covariance;
- `PositionYawCorrelated` — position↔yaw coupling is non-zero and the block is symmetric;
- `PositiveSemiDefinite` — exported block is PSD.

* before

https://github.com/user-attachments/assets/bcbd40a9-bdad-4e2a-a0aa-5bc664d9abb6

* after

https://github.com/user-attachments/assets/61b46e7c-090c-4b24-a1f6-c024729bd5ec

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

Stationary vehicles observed as a partial cluster (single axle face, e.g. when the bbox detection
channel drops) no longer accumulate a spurious yaw rotation. The wheelbase lock keeps the tracked
body length near-constant during partial updates, and the Jacobian-based covariance export
restores the position↔yaw cross-covariance for downstream consumers.

### DIFF
--- a/perception/autoware_multi_object_tracker/lib/tracker/motion_model/bicycle_motion_model.cpp
+++ b/perception/autoware_multi_object_tracker/lib/tracker/motion_model/bicycle_motion_model.cpp
@@ -645,9 +645,6 @@ bool BicycleMotionModel::getPredictedState(
   const double wheel_base_inv_sq = wheel_base_inv * wheel_base_inv;
   const double sin_yaw = (X(IDX::Y2) - X(IDX::Y1)) * wheel_base_inv;
   const double cos_yaw = (X(IDX::X2) - X(IDX::X1)) * wheel_base_inv;
-  const double sin_yaw_sq = sin_yaw * sin_yaw;
-  const double cos_yaw_sq = cos_yaw * cos_yaw;
-  const double sin_cos_yaw = sin_yaw * cos_yaw;
 
   // set position
   pose.position.x = (X(IDX::X1) * motion_params_.lf_ratio + X(IDX::X2) * motion_params_.lr_ratio) *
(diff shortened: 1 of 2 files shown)


ogrex (approved):
LGTM

Issue #7724, opened 2024-06-27
Tracking disabled tests due to lack of gpu capable machines


xmfcx:
### Checklist

- [X] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [X] I've searched other issues and no duplicate issues were found.
- [X] I've agreed with the maintainers that I can plan this task.

### Description

Story behind can be tracked from:
- https://github.com/autowarefoundation/autoware.universe/issues/7668#issuecomment-2194415792

Basically the tests disabled here:
- ✅ work on machines with Nvidia GPU
- ❌ fail on machines without Nvidia GPU

When we upgrade our testing infrastructure to have CUDA capable GPUs, we can re-enable these tests.

cc. @knzo25 @mitsudome-r 

### Purpose

To make CI pass on no-gpu machines. (Especially github hosted runners)

### Possible approaches

- Commenting out the failing tests in the respective CMakeLists.txt file.
- Adding a reference to here

### Definition of done

- [ ] `build-and-test` passes on the main branch.

xmfcx:
@mitsudome-r if somehow we gain access to gpu capable machines either by:
- adding more self hosted local machines
- or utilizing https://github.com/philips-labs/terraform-aws-github-runner

We can use this PR as reference:
- https://github.com/autowarefoundation/autoware.universe/pull/7721

I will also create a documentation on how I set up the self hosted runners with custom init scripts.

knzo25:
@xmfcx 
Is there any updates on this issue?
We would like to have the cuda tests enabled (https://github.com/autowarefoundation/autoware.universe/blob/main/perception/autoware_lidar_centerpoint/CMakeLists.txt)

xmfcx:
@knzo25 please have a peek at:
- https://github.com/autowarefoundation/autoware.universe/pull/8498

Through this new system, we have access to `gpu1.small` machines with `15 GB, 4vCPU` and an unknown GPU, I didn't make any tests.

Full list: https://aws.amazon.com/codebuild/pricing/ (select us east n. virginia)

But this machine is pricier.

We only need GPU for tests, not for building with CUDA.

So one option would be to separate build and test in the workflow and test on this new machine.

But it requires someone to work on this task.

xmfcx:
- https://github.com/autowarefoundation/autoware.universe/pull/5174 is tracked too.

mojomex:
[...]

PR #12992, opened 2026-07-14
fix(autoware_lidar_centerpoint): use voxel_size_y when scaling y position variance

gabrielmanalu:
## Description
In `generateBoxes3D_kernel`, the y position variance is scaled by `voxel_size_x` instead
of `voxel_size_y`:

    det_boxes3d[idx].y_variance = voxel_size_x * downsample_factor * expf(offset_y_variance);

The mean decode in the same kernel uses the per-axis sizes (`voxel_size_x` for x,
`voxel_size_y` for y), and the offset head outputs are in downsampled-grid-cell units, so
the y-axis conversion should use the y cell size. This looks like a copy-paste of the
x line. The value flows unmodified into `pose_with_covariance` (Y_Y) via
`convertPoseCovarianceMatrix` and is consumed downstream as measurement noise. This also
affects `pointpainting_fusion`, which reuses this postprocess library.

All shipped configurations use square voxels and set `has_variance: false`, so this has
no effect on any default setup — it only affects custom deployments running
variance-outputting models with anisotropic voxels, where the y covariance is silently
wrong by the factor `voxel_size_x / voxel_size_y`.

## How was this PR tested?
Added `PostprocessKernelAnisotropicVarianceTest` (anisotropic voxels 0.2/0.4 m,
`has_variance: true`, identical log-variance on both offset-variance channels). Built and
run locally on an NVIDIA GPU (WSL2/Ubuntu 22.04, Humble):
- Before the fix: `y_variance` = 0.05000000074505806, expected 0.10000000149011612 —
  off by exactly the `voxel_size_x / voxel_size_y` ratio (0.5), confirming the mechanism.
- After the fix: all 5 tests in the file pass, including the 4 pre-existing ones
  (`EmptyTensorTest`, `SingleDetectionTest`, `InvalidYawTest`, `CircleNMSTest`).

Note: `test_postprocess_kernel`'s build target is currently commented out in
`CMakeLists.txt`, tracked by #7724 (tests disabled due to lack of GPU-capable CI
runners — the same marker also covers `test_voxel_generator` and
`test_preprocess_kernel` in this file). The fixture's `SetUp()` also no longer matched
the current `CenterPointConfig`/`PostProcessCUDA` constructors. I fixed both locally
(GPU-equipped dev machine) to build and run this test, but left the CMake gating
change out of this PR, since a blanket re-enable would break CI on GPU-less runners —
the same constraint #7724 exists to prevent. That's a CI-policy question for
maintainers, not something I want to decide unilaterally in a bug-fix PR.

## Effects on system behavior
None for all shipped configs (square voxels ⇒ bit-identical output). For anisotropic
variance-model deployments, the published y position covariance becomes correct.

## Interface changes
None.


KSeangTan (approved):
LGTM!

PR #12993, opened 2026-07-14
refactor(autoware_traffic_light_classifier): extract a Node-free CnnLampRecognizerCore

takam5f2:
## Description

Splits `CnnLampRecognizer` into a **Node-free `CnnLampRecognizerCore`** plus a thin `CnnLampRecognizer : ClassifierInterface` **ROS adapter**, as well as permed in #12967. Both classes stay in the same
`cnn_lamp_recognizer.{hpp,cpp}` pair (no new files), matching `cnn_classifier.{hpp,cpp}`.

This is part of the ongoing effort to make the traffic-light classifiers callable as Node-free
core libraries (Color and CNN are already done: #12916, #12967).

## What changed

- **`CnnLampRecognizerCore`** (Node-free): owns the TensorRT engine and the whole recognition
  pipeline — `preprocess` / `doInference` / `decodeTlrOutput`, NMS and per-image dedup.
  `classify(images)` returns `DetectionResult{ lamps_per_image, success }`.
  `updateTrafficSignals` and `outputDebugImage` become `static` core helpers (pure logic, like
  `CNNClassifierCore::decode_label` / `make_debug_image`). The core references no `rclcpp` /
  node / `image_transport`.
- **`CnnLampRecognizer`** (adapter): keeps only `node_ptr_`, `image_pub_`, and `core_`.
  `declare_lamp_config(node)` reads the ROS parameters and narrows the anchors
  `double`→`float` (a ROS-param quirk). It declares the same parameter names as before.
- **Invariant ownership**: the `anchors.size() == 2 * num_anchors` validation and the
  `bbox_offset = 0.5 * (scale_x_y - 1)` derivation move from the adapter into the **core ctor**
  (before the engine build, fail-fast). The core is what actually indexes `anchors[...]` and
  uses `bbox_offset` in `decodeTlrOutput`, so a directly-constructed core is now correct on its
  own — mirroring how `CNNClassifierCore` validates its own `mean`/`std` sizes.

The public API — `CnnLampRecognizer(rclcpp::Node*)` + `getTrafficSignals(images, signals)` — is
**unchanged**, so the node/single-image construction sites are untouched.

## Effects on system behavior

Behavior is preserved on the normal path. Two deliberate, non-regressing deltas on the failure path:

1. **On inference failure the output `traffic_signals` is left untouched (all-or-nothing)**
   instead of partially mutated. This mirrors the behavior the CNN core extraction (#12967)
   already introduced (its `getTrafficSignals` likewise moved from per-batch in-place writes to
   build-fresh-then-merge-on-success). It is not a regression: `doInference` failures are
   infrastructure-level (`setInputShape` / `setTensorsAddresses` / `enqueueV3` bool returns, or
   CUDA errors that throw via `CHECK_CUDA_ERROR`) and are not data-dependent, so no meaningful
   "partial success" state exists — a healthy engine processes every batch; a broken one fails
   on the first.
2. `doInference` no longer logs `setInputShape` failures directly; the single adapter-level
   `RCLCPP_ERROR("inference failed")` covers it (same as the CNN core).

## How was this PR tested?

```
colcon build --packages-select autoware_traffic_light_classifier
```

All tests pass, including the `test_cnn_lamp_recognizer` characterization test (#12974).

## Notes for reviewers

- The core's result type is `DetectionResult{ lamps_per_image, success }` rather than the
  family's `ClassifierResult{ TrafficLightArray signals, success }`: the lamp core returns
  geometric detections (the adapter applies `updateTrafficSignals` per signal because it needs
  the caller's `traffic_light_type`), so a distinct name avoids implying it carries built signals.
- Scope is source-only. Deferred to follow-ups (kept out to minimize review surface): a GPU-free
  `CnnLampRecognizerCore` unit test for the decode branches the characterization test cannot
  cover, and trimming the now-unused `<opencv2/highgui/highgui.hpp>` include (as done for the CNN
  classifier).

### DIFF
--- a/perception/autoware_traffic_light_classifier/src/classifier/cnn_lamp_recognizer.cpp
+++ b/perception/autoware_traffic_light_classifier/src/classifier/cnn_lamp_recognizer.cpp
@@ -113,188 +113,28 @@ static void convertBBoxInfoToLampElement(const BBoxInfo & box_info, LampElement 
   }
 }
 
-static void updateTrafficSignals(
-  const std::vector<LampElement> & unique_elements,
-  tier4_perception_msgs::msg::TrafficLight & traffic_signal)
-{
-  traffic_signal.elements.clear();
-  bool is_pedestrian = traffic_signal.traffic_light_type == 1;
-  if (unique_elements.empty()) {
-    MsgTE unknown_elem;
-    unknown_elem.color = MsgTE::UNKNOWN;
-    unknown_elem.shape = MsgTE::UNKNOWN;
-    unknown_elem.confidence = 0.0;
-    traffic_signal.elements.push_back(unknown_elem);
-    return;
-  }
-  for (const auto & e : unique_elements) {
-    MsgTE element;
-    element.confidence = e.confidence;
-    switch (e.color) {
-      case Color::GREEN:
-        element.color = MsgTE::GREEN;
-        break;
-      case Color::AMBER:
-        element.color = MsgTE::AMBER;
-        break;
-      case Color::RED:
-        element.color = MsgTE::RED;
-        break;
-      default:
-        element.color = MsgTE::UNKNOWN;
-        break;
-    }
-    if (is_pedestrian || e.shape == Shape::PED) {
-      element.shape = MsgTE::CIRCLE;
-      traffic_signal.elements.push_back(element);
-      continue;
-    }
-    switch (e.shape) {
-      case Shape::CIRCLE:
-        element.shape = MsgTE::CIRCLE;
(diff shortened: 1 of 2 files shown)


takam5f2:
@a-maumau @badai-nguyen 

Could you check the PR?
The PR attempts to split  the `CnnLampRecognizer` core/adapter, as well as the CNN classifier: source-only, public API unchanged.

The only intentional differences are on the failure path; all-or-nothing output on inference failure + one folded log line.

a-maumau (approved):
LGTM

PR #12996, opened 2026-07-14
fix(autoware_multi_object_tracker): prefer fresh bbox tracker over long partially-updated tracker in overlap pruning

technolojin:
## Description

In overlap pruning, a vehicle tracker that has been **partially updated for a long time** (e.g. tracked only from lidar clusters / corners, `trust_extension=false`) could out-rank and **absorb a fresh, correct bounding-box-spawned tracker**, so the wrong (drifted) state persisted. Nothing in the survival ranking accounted for *how recently a tracker saw a real full observation*, and the merge logic had two structural weaknesses that made targeted fixes fragile: the outcome could depend on tracker-list order, and pair candidacy used fixed per-label distance thresholds that ignored object size.

This PR restructures the overlap merge and fixes the ranking. It is a series of five self-contained commits, each building and passing tests on its own:

1. **`refactor: restructure overlap merge into a staged pipeline`** https://github.com/autowarefoundation/autoware_universe/pull/12996/changes/3c2e6adb348494577f0fcae4539fb760eeb18d97
   Extracts the merge into a `merger/` module (snapshot/decision-context types, spatial gate, survival ranking) driven as four stages: snapshot all trackers once per cycle → decide directed merge edges per candidate pair → group one best winner per loser → apply the star-forest merges. Pair decisions read per-cycle snapshots, not the live trackers, so all later commits inherit order-independent decisions. `redundancy_check` moves under `merger/detail`. No behavior change.

2. **`feat: rank a loser's competing winners by substance`** https://github.com/autowarefoundation/autoware_universe/pull/12996/changes/d9004e6d3fc9d66ceb6327b3a302f5252940a67f
   When several winners contest the same loser, the intrinsically stronger one (priority, confidence, known probability, position covariance, measurement count — `compareWinnerSubstance`) claims it; distance and UUID only break ties. Merges apply strongest-winner-first, so star-forest conflicts resolve toward the stronger edge.

3. **`feat: add tolerance deadband to overlap ranking comparisons`** https://github.com/autowarefoundation/autoware_universe/pull/12996/changes/955f5f4879602e90adfeeee059f56216f2771dd6
   `withinDeadband` (relative-or-absolute tolerance) applied to the continuous comparisons — the position-covariance survival tier and the distance tie-break between equally-strong winners. Values differing only within the band compare equal and defer to the next tie-break, removing knife-edge flips from numerical noise.

4. **`feat: rank overlap survival by full-measurement freshness`** *(the core bugfix)* https://github.com/autowarefoundation/autoware_universe/pull/12996/changes/893b9c6e9f930d3e7c025ca55f01a0e2e573b244
   `Tracker::last_fully_measured_time_` is refreshed on full (`trust_extension`) measurements, including the stabilized TRY_EXTENSION update; partial/conditioned updates do not refresh it. `getElapsedTimeFromFullMeasurement()` is virtual — the base returns 0.0, so non-vehicle trackers always count as fully measured; `VehicleTracker` and `MultipleVehicleTracker` override it. The overlap merge snapshots a `fully_measured_stale` flag (elapsed > `0.35 s`) and adds a survival tier **above** the confidence tier: a stale tracker yields to a fully-measured one.
   Because the winner must still be confident to merge, both directions are safe:
   - **Fresh tracker already confident** → it wins and absorbs the stale partial tracker (state corrected).
   - **Fresh tracker not yet confident** (just spawned) → the merge is deferred and both survive; the stale tracker can no longer remove the fresh detection. It absorbs on a later cycle once confident.

* before: truck detection(yellow) cannot take over car tracker kept by a small polygon

https://github.com/user-attachments/assets/b3705fcb-3033-4540-9e4d-33be7d673488

* after: after a moment, truck detection overtakes car tracker

https://github.com/user-attachments/assets/1ad1634a-6791-4383-9757-2c3af5352673

5. **`feat: gate overlap pairs by size-aware multi-circle cover`** https://github.com/autowarefoundation/autoware_universe/pull/12996/changes/29f3e2613ecf6e52b6912549d2ff617d47a19112
   Replaces the per-label distance-threshold gate: each tracker's extents are covered by a chain of equal circles along its longer axis (up to 8), and a pair is admitted when their circles come within `radius_a + radius_b + margin` (`1.0 m`), found via a bulk-loaded R-tree. Parametric shapes take extents from dimensions; polygon shapes from their footprint's bounds in the object frame, so large polygon-only trackers (e.g. buses seen as clusters) gate correctly. The now-unused `pruning_distance_thresholds` parameter is removed from config, schema, node, and tests.

## Related links

**Parent Issue:**

- None

## How was this PR tested?

Package unit tests (`autoware_multi_object_tracker`, Release build): full gtest suite **30/30 pass**; each of the five commits was built and tested independently.

New overlap-manager tests added in this series:

- `FreshBboxTrackerOutranksStalePartialVehicle` — a confident fresh bbox tracker absorbs an overlapping confident-but-stale partial vehicle tracker.
- `StalePartialVehicleDoesNotAbsorbImmatureFreshTracker` — a stale partial tracker does **not** absorb a just-spawned (not-yet-confident) fresh tracker; both survive. Fails without the freshness tier (stale absorbs fresh → 1 survivor instead of 2).
- `NonVehicleTrackersAlwaysCountFullyMeasured` — the freshness tier never demotes non-vehicle trackers.
- `ChainConflictAppliesStrongerMergeFirst` — star-forest conflicts resolve toward the stronger winner.
- `OutcomeIsIndependentOfListOrder` — merge outcome is identical under reversed tracker-list order.
- `GateAdmitsLargePolygonByFootprintExtent` — the multi-circle gate admits a large polygon tracker whose footprint extends past a fixed-distance gate.

Verified by rosbag logging-simulation, see the videos above.

## Notes for reviewers

- Reviewing commit-by-commit is recommended; the first commit is a pure restructuring and the four behavior changes are isolated on top of it.
- The "fully measured" signal keys off full (`trust_extension`) updates only, so it excludes conditioned/partial updates by construction. Healthy bbox-tracked vehicles refresh every frame and are never stale; a momentary 1–2 frame cluster-only gap stays under the `0.35 s` threshold.
- The staleness window, gate margin, and deadband tolerances are `constexpr` in the `merger/` sources; happy to promote any of them to config if preferred.
- Out of scope: if drift is large enough that `isRedundant` never fires, the two trackers won't merge and a duplicate can remain — that would need a targeted association re-anchoring, gated behind bag evidence.

## Interface changes

ROS parameter removed: `pruning_distance_thresholds` (per-label overlap gate distances) is deleted from `multi_object_tracker_node.param.yaml` and the schema; the spatial gate is now derived from tracker size and needs no tuning parameter.

## Effects on system behavior

- A vehicle tracker without a full measurement within `0.35 s` no longer out-ranks / absorbs an overlapping fully-measured tracker; a fresh bounding-box tracker overlapping a long-partially-updated tracker survives (and, once confident, absorbs the stale one).
- Overlap merge decisions are independent of tracker-list order, and a contested loser goes to the intrinsically stronger winner.
- Pair candidacy scales with object size (multi-circle cover) instead of fixed per-label distances, so long vehicles and large polygon clusters gate by their actual extents.


perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/tracker/trackers/vehicle_tracker.hpp
```diff
@@ -98,6 +98,12 @@ class VehicleTracker : public Tracker
     if (!trust_extension) return UpdatePath::CONDITIONED;
     return has_significant_shape_change ? UpdatePath::TRY_EXTENSION : UpdatePath::NORMAL;
   }
+
+  // Vehicle boxes coast on partial updates; staleness on full measurements is meaningful.
+  double getElapsedTimeFromFullMeasurement(const rclcpp::Time & current_time) const override
```
ogrex, line 103: It is OK here

But currently PedestrianAndBicycleTracker inherits the base getElapsedTimeFromFullMeasurement() implementation, which always returns 0.0.
Since PedestrianAndBicycleTracker can use bicycle_tracker_ as its active inner tracker for bicycle/motorcycle labels, should we also override it?

technolojin: thank you for your finding. 
yes we need to do.

fixed https://github.com/autowarefoundation/autoware_universe/pull/12996/changes/39ca4cf603e28b90a120055aaf83ba5822d661a6

perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/tracker/trackers/pedestrian_and_bicycle_tracker.hpp
```diff
@@ -65,6 +65,12 @@ class PedestrianAndBicycleTracker : public Tracker
     return activeInner().getMotionState(time, pose, pose_cov, twist, twist_cov);
   }
   rclcpp::Time getStateTime() const override { return activeInner().getStateTime(); }
+  // Forward to the active inner tracker: pedestrian reports no staleness, bicycle coasts on partial
+  // updates like a vehicle box.
+  double getElapsedTimeFromFullMeasurement(const rclcpp::Time & current_time) const override
+  {
+    return activeInner().getElapsedTimeFromFullMeasurement(current_time);
```
ogrex: I think the current impl still has a mismatch.

For `PedestrianAndBicycleTracker`, the outer tracker runs `Tracker::updateWithMeasurement()`, and its inner trackers are updated through `pedestrian_tracker_.measure()` / `bicycle_tracker_.measure()` directly. That bypasses activeInner's `updateWithMeasurement()` that refreshes `last_fully_measured_time_`. 

So,  activeInner's last_fully_measured_time_ is not updated

Maybe we can use PedestrianAndBicycleTracker's last_fully_measured_time_?

```cpp
double getElapsedTimeFromFullMeasurement(const rclcpp::Time & current_time) const override
{
  if (getHighestProbLabel() == classes::Label::PEDESTRIAN) {
    return 0.0;
  }
  return elapsedSinceLastFullMeasurement(current_time);
}

ogrex (changes requested):
Maybe still has some problem

perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/tracker/trackers/pedestrian_and_bicycle_tracker.hpp
technolojin: I applied the fix and added a new test.
https://github.com/autowarefoundation/autoware_universe/pull/12996/changes/02f9c0a98fd26fe265ec859bd242b90086150910

ogrex (approved):
LGTM

technolojin:
launcher updated by https://github.com/autowarefoundation/autoware_launch/pull/1907

PR #13001, opened 2026-07-15
test(autoware_traffic_light_classifier): add CnnLampRecognizerCore unit tests

takam5f2:
## Description

Adds unit tests for `CnnLampRecognizerCore`, the Node-free lamp recognition core extracted in #12993. Two groups, both ROS-free:

- **Static helpers (no GPU, no model)** — pin the non-obvious behavior of `updateTrafficSignals()` (empty → single UNKNOWN placeholder, one element per lamp in input order, clear-before-map, pedestrian forcing to CIRCLE, and the confidence zeroing-vs-keeping asymmetry) and `outputDebugImage()` geometry. The plain enum→enum mappings (color / arrow direction / cross) are deliberately left unpinned — a test would only mirror the `switch`.
- **GPU-gated `classify()` fixture** — builds the real TensorRT engine once per suite and self-skips (`GTEST_SKIP`) when the model or a usable GPU is missing. It asserts only model-stable facts: well-formed detections (box coords in the unit interval, confidence in `(0, 1]`), at least one detection on a clear crop, and one result vector per input image (a detected crop paired with a black image, so a scatter swap is detectable). No specific class is pinned, since TensorRT output is not bit-reproducible across GPU / TRT versions.

This is the deferred follow-up to #12993 (which extracted the core, guarded by the existing characterization test from #12974). Production code is unchanged.

## Related links

**Parent Issue:**

- Follow-up to #12993

## How was this PR tested?

- `colcon build` + direct run: the 10 model-free tests pass; the 2 GPU-gated `classify()` tests self-skip cleanly (exit 0) on a host without a usable GPU.
- pre-commit (clang-format / cpplint) pass.
- **Pending before ready-for-review:** a full `colcon test` run on a machine with a usable GPU, to exercise the two `classify()` tests end-to-end against the real model.

## Notes for reviewers

- Before building the engine, the fixture probes CUDA context creation with `cudaFree(0)` and skips on failure. `is_cuda_runtime_available()` only counts devices, so on a present-but-unusable GPU (context-init OOM, driver mismatch) TensorRT would otherwise abort the whole process with an uncatchable `nvinfer1::APIUsageError` during the engine build — which is not a `std::exception` and escapes even a catch-all. The probe turns that case into a clean skip.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_traffic_light_classifier/CMakeLists.txt
+++ b/perception/autoware_traffic_light_classifier/CMakeLists.txt
@@ -206,6 +206,18 @@ if(BUILD_TESTING)
       ${OpenCV_LIBRARIES}
       ${CUDA_LIBRARIES}
     )
+
+    # CnnLampRecognizerCore tests (ROS-free): static helpers run GPU-free; the classify()
+    # tests self-skip without a GPU/model. TIMEOUT covers the engine build.
+    ament_auto_add_gtest(test_cnn_lamp_recognizer_core
+      test/test_cnn_lamp_recognizer_core.cpp
+      TIMEOUT 600
+    )
+    target_link_libraries(test_cnn_lamp_recognizer_core
+      ${PROJECT_NAME}
+      ${OpenCV_LIBRARIES}
+      ${CUDA_LIBRARIES}
+    )
   endif()
 endif()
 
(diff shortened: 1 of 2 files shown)


takam5f2:
@a-maumau @badai-nguyen  

Could you check the PR? This PR only creates a new unit test. The production code will be never changed. 

Best,

a-maumau (approved):
LGTM

PR #13009, opened 2026-07-16
feat(autoware_ptv3): update filtered classes

amadeuszsz:
## Description

Update filtered classes to mimic typical ground segmentation algorithm.
Current model might have wrong classification for flat surfaces (drivable and non-drivable classes) duo to obvious geometrical similarity.
It is safe to filter them out because lanelets are responsible for handling area of interests for planners.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


mojomex (approved):
LGTM.

`non_drivable_flat` being filtered is OK for now, as we are operating on a lanelet HD map anyway.

PR #13014, opened 2026-07-16
feat(external_cmd_converter): apply agnocast_wrapper::Node to external_cmd_converter

kobayu858:
## Description
Apply `agnocast_wrapper::Node` to `autoware_external_cmd_converter`.
Also, please refer to https://github.com/autowarefoundation/autoware_core/blob/main/common/autoware_agnocast_wrapper/docs/review_guide.md for the review guide. FYI: This node takes [Method2](https://github.com/autowarefoundation/autoware_core/blob/main/common/autoware_agnocast_wrapper/docs/review_guide.md#method-2-agnocast_wrappernode-inheritance) of the two integration methods.

## Related links

## How was this PR tested?
- Build: `colcon build` succeeds with both `ENABLE_AGNOCAST=1` and `ENABLE_AGNOCAST=0`.
- Unit tests: Pass with `ENABLE_AGNOCAST=0`; the node-launching test is skipped when `ENABLE_AGNOCAST=1` (build-level guard in `CMakeLists.txt`, per the `autoware_agnocast_wrapper` unit-test policy). Test logic is unchanged.
- Standalone launch: Confirmed the node launches as a standalone node (not inside a component container) under both `ENABLE_AGNOCAST=0` and `ENABLE_AGNOCAST=1`, since `agnocast::Node` does not currently work inside a component container. The node's own `launch.py` always launches standalone and self-contains the `agnocast_env` / `LD_PRELOAD` wiring, so no `autoware_launch`-side change is required.
- LSim (sample rosbag): The Logging Simulator runs to completion under both `ENABLE_AGNOCAST=0` and `ENABLE_AGNOCAST=1` with no `process died` or similar failures, and the converter publishes the expected `control_cmd` output.

## Notes for reviewers
This node is always launched standalone; its `launch.py` self-contains the Agnocast environment setup, so no companion `autoware_launch` PR is needed.

## Interface changes

None.

## Effects on system behavior

None when `ENABLE_AGNOCAST` is not set or set to `0` on runtime. When `ENABLE_AGNOCAST=1`, CPU time used in this node for message serialization/deserialization will be reduced.


Koichi98:
@kobayu858 
Could you delete `rclcpp::QoS{1}` so that it aligns with the original code? The wrapper provides default value.

```
  steering_cmd_sub_ = create_polling_subscriber<SteeringCommand>("in/steering_cmd", rclcpp::QoS{1});
```

vehicle/autoware_external_cmd_converter/launch/external_cmd_converter.launch.py
```diff
@@ -104,42 +103,57 @@ def generate_launch_description():
...
-            _create_mapping_tuple("in/steering_cmd"),
-            _create_mapping_tuple("in/gear_cmd"),
-            _create_mapping_tuple("in/heartbeat"),
-            _create_mapping_tuple("in/current_gate_mode"),
-            _create_mapping_tuple("in/odometry"),
-            _create_mapping_tuple("out/control_cmd"),
-        ],
-        parameters=[
-            dict(  # noqa: C406 for using helper function
+    # Resolves ld_preload_value (LD_PRELOAD with the Agnocast heaphook prepended) and
```
Koichi98: Can we delete these comments? This is a standard application pattern and there seems nothing special for the agnocast application to this node.

kobayu858: fixed in https://github.com/autowarefoundation/autoware_universe/pull/13014/commits/ff3ecd7a2b8ca5d9b1c3493f6e49b953b03fd563

kobayu858:
@Koichi98 
I delete `rclcpp::QoS{1}` in https://github.com/autowarefoundation/autoware_universe/pull/13014/commits/ef0dacf1e3bc4d1b9ef7ff42a6ef731002742fc4

vehicle/autoware_external_cmd_converter/src/node.cpp
```diff
@@ -34,6 +34,16 @@ ExternalCmdConverterNode::ExternalCmdConverterNode(const rclcpp::NodeOptions & n
...
 
+  steering_cmd_sub_ =
+    autoware::agnocast_wrapper::polling::create_polling_subscriber<SteeringCommand>(
+      this, "in/steering_cmd");
+  velocity_sub_ =
+    autoware::agnocast_wrapper::polling::create_polling_subscriber<Odometry>(this, "in/odometry");
+  gear_cmd_sub_ = autoware::agnocast_wrapper::polling::create_polling_subscriber<GearCommand>(
+    this, "in/gear_cmd");
+  gate_mode_sub_ = autoware::agnocast_wrapper::polling::create_polling_subscriber<GateMode>(
+    this, "in/current_gate_mode");
```
Koichi98: @kobayu858 
Could you define these in `node.hpp` as it is before?

kobayu858: fixed in https://github.com/autowarefoundation/autoware_universe/pull/13014/commits/8e505f39469590fdbbade45dec5c63c639dde2a2

Koichi98 (approved):

kobayu858:
@N-Eiki @TakaHoribe

Sorry to bother you when you're busy. 
Could you take a look when you have a moment? Thanks!

N-Eiki (approved):
LGTM

PR #13016, opened 2026-07-17
refactor(autoware_traffic_light_arbiter): rename classes to drop Core suffix

takam5f2:
## Description

Pure rename/move refactor of the traffic light arbiter. It lines up file names with class names by role:

- ROS-free arbitration logic: `TrafficLightArbiterCore` → **`TrafficLightArbiter`**, in `traffic_light_arbiter.{hpp,cpp}`
- Thin ROS node: `TrafficLightArbiter` → **`TrafficLightArbiterNode`**, in `traffic_light_arbiter_node.{hpp,cpp}`

Previously the base-named file held the Node while the core carried a `Core` suffix, so file names and class names were crossed. This drops the `Core` suffix and makes the core the canonical `TrafficLightArbiter`.

This PR also carries a small docs-only commit that rewords the README to describe the arbiter's behavior (merging, signal-match validation) by capability instead of attributing it to "a node", since that behavior lives in the core and the node is only the ROS I/O wrapper — matching the Node/Core split.

## Specification impact

Only the symbol / registration contract changes:

- C++ classes: `TrafficLightArbiterCore` → `TrafficLightArbiter`, and the ROS node `TrafficLightArbiter` → `TrafficLightArbiterNode` (header `traffic_light_arbiter_core.hpp` removed, `traffic_light_arbiter_node.hpp` added).
- Registered composable-node plugin name: `autoware::traffic_light::TrafficLightArbiter` → `autoware::traffic_light::TrafficLightArbiterNode`.

No consumer of either exists in `autoware_universe` or downstream launch (the node runs as the standalone executable `traffic_light_arbiter_node`), so nothing needs updating.

## How was this PR tested?

- `colcon build --packages-select autoware_traffic_light_arbiter` — success (library + node + tests).
- ROS-free core test: 39/39 passed.
- pre-commit (clang-format, cpplint, include-guard, Autoware System Design Format) passed.
- Launch the node via LSim for the T4 Internal project.


takam5f2:
@a-maumau  
This PR will rename class and file name as follows.

* TrafficLightArbiterCore -> TrafficLightArbiter
* TrafficLightArbiter -> TrafficLightArbiterNode

This will not change node name and will change only the symbol.

perception/autoware_traffic_light_arbiter/include/autoware/traffic_light_arbiter/traffic_light_arbiter_node.hpp
```diff
@@ -0,0 +1,54 @@
+// Copyright 2023 The Autoware Contributors
```
a-maumau, line 1: 2023 -> 2026 (or this file exist in the past? in that case, I am not sure which year is appropriate)

(just a comment)
Autoware doesn't do, but it seems when there is some modification, the year should be updated like "2023-2024, 2026" or "2023-2026"

a-maumau (approved):
LGTM.

perception/autoware_traffic_light_arbiter/include/autoware/traffic_light_arbiter/traffic_light_arbiter_node.hpp
takam5f2: @a-maumau  
traffic_light_arbiter_node.{hpp,cpp} isn't new source code. It's the original traffic_light_arbiter.{hpp,cpp} (the ROS node, dating back to 2023), just renamed in this PR. So I kept its original year. Git shows it as an added file only because the base name was swapped with the core, which defeats rename detection.

a-maumau: LGTM!

PR #13018, opened 2026-07-17
test(autoware_traffic_light_classifier): add CnnLampRecognizer adapter test and retire characterization

takam5f2:
## Description

Now that the Node-free `CnnLampRecognizerCore` split has landed and its ROS-free unit test is
merged (#13001), this PR mirrors the CNN classifier test layout for the lamp recognizer:

- **Add `test/test_cnn_lamp_recognizer_adapter.cpp`** (GPU-gated, ROS-isolated) covering the
  concerns that live in the ROS adapter and cannot move to the core:
  - the `images`/`signals` size guard in `getTrafficSignals` (a caller-owned array, so a count
    mismatch would otherwise be an out-of-range write), and
  - the per-image → per-signal-slot scatter of the assembled `getTrafficSignals` path.

  Parameter declaration is exercised implicitly: the adapter ctor reads every `model_params.*`
  parameter, so a wrong name/type fails construction here. Model outputs are **not** pinned; a
  pure-black slot stands in for a guaranteed no-detection (no model detects a lamp in black),
  keeping the assertions model-stable. Mirrors `test_cnn_classifier_adapter.cpp`.

- **Retire the real-model characterization test** (`test/test_cnn_lamp_recognizer.cpp`, added in
  #12974). It was the safety net for the core/adapter split; that split is now covered by the
  ROS-free core unit test (#13001) plus this adapter test, so it is redundant. Dropping it also
  removes its brittle live-model output pins (AMBER/GREEN), which we avoid in permanent tests.

- **Rename the core test to the canonical `test/test_cnn_lamp_recognizer.cpp`** (drop the `_core`
  suffix) so the naming matches the CNN classifier (unsuffixed = core, `_adapter` = adapter), and
  switch it to `ament_auto_add_gtest` since it is ROS-free.

Production code is unchanged; this is a test-only restructuring.

## Related links

**Parent Issue:**

- None.

Related PRs: #13001 (`CnnLampRecognizerCore` unit test, merged), #12993 (core/adapter split),
#12974 (the characterization test this PR retires).

## How was this PR tested?

The adapter test faithfully mirrors the already-merged `test_cnn_classifier_adapter.cpp` and the
`SetUpTestSuite` of the retired characterization test (both of which passed CI), using only APIs
already on `main`. `clang-format` and `cpplint` pass (pre-commit). `cspell` reproduced locally.

Draft: a full `colcon build` + GPU `colcon test` on a machine with a CUDA GPU and the lamp
recognizer ONNX model under `autoware_data` is still pending; the adapter test and the renamed
core test both self-skip when a GPU/model is unavailable. Will be run before marking
ready-for-review.

## Notes for reviewers

- End state is symmetric with the CNN classifier tests: `test_cnn_lamp_recognizer.cpp` (core,
  ROS-free) + `test_cnn_lamp_recognizer_adapter.cpp` (adapter, GPU-gated).
- The adapter's scatter test uses a detected crop + a black image so the two slots are
  asymmetric; a swapped/collapsed mapping is caught while nothing model-specific is pinned.

## Interface changes

None.

## Effects on system behavior

None.


takam5f2:
@a-maumau  
Could you check the PR? 
This PR will adds the adapter test code, `test_cnn_lamp_recognizer_adapter.cpp`, as well as `cnn_classifier.cpp`.

Additionally, `test_cnn_lamp_recognizer_core.cpp` will be renamed as `test_cnn_lamp_recognizer.cpp` because the characterization test is not needed.

a-maumau (approved):
LGTM!

PR #13019, opened 2026-07-17
refactor(autoware_traffic_light_classifier): remove dead code and clean up naming in CnnLampRecognizer

takam5f2:
## Description

Cleanup of `cnn_lamp_recognizer.{hpp,cpp}` (the `CnnLampRecognizer` ROS adapter and the Node-free `CnnLampRecognizerCore`). **No behavior change** — dead-code removal, ROS 2 snake_case naming, and two small non-behavioral refactors (a non-destructive debug-image API and dropping the `MsgTE` alias).

- **Dead code removal:** drop the unreachable `case Shape::PED` in `update_traffic_signals`. The pedestrian guard just above the `switch` always `continue`s for `Shape::PED`, so the switch never sees that value; PED still maps to `CIRCLE` via the guard.
- **ROS 2 snake_case naming:**
  - File-local constants and anonymous-namespace helpers: 
    - `kDebugImageWidth` → `debug_image_width`
    - `kDebugTextHeight` → `debug_text_height`
    - `kDirectionVecNormSqThreshold` → `direction_vec_norm_sq_threshold`
    - `angleToArrowDirection` → `angle_to_arrow_direction`
    - `get2DIoU` → `get_2d_iou`
    - `runNms` → `run_nms`
    - `convertBBoxInfoToLampElement` → `convert_bbox_info_to_lamp_element`
    -  `overlap1D` → `overlap_1d`
  - `CnnLampRecognizerCore` methods (the core is interface-free, so it owns its naming): 
    - `updateTrafficSignals` → `update_traffic_signals`
    - `doInference` → `do_inference`
    - `decodeTlrOutput` → `decode_tlr_output`
  - The other minors: 
    - `BBoxInfo` fields: `classId` → `class_id`
    - `subClassId` → `sub_class_id`
- **Non-destructive debug-image API:** `outputDebugImage(cv::Mat & debug_image, …)`, which mutated its argument in place, becomes `make_debug_image(const cv::Mat & roi_image, …) -> cv::Mat`, which clones internally and returns the new image. The `clone()` moves inside the function, so the now-redundant `images[i].clone()` at the call site is removed. The resulting debug image is identical; only the ownership/signature changes. (This is not a rename — the previous body listed it under naming by mistake.)
- **Drop the `MsgTE` alias:** replace the file-local `using MsgTE = tier4_perception_msgs::msg::TrafficLightElement;` with `using tier4_perception_msgs::msg::TrafficLightElement;`, so the type is referred to by its own name (`TrafficLightElement::…`) rather than the ad-hoc abbreviation.
- **Scope the pi constant:** the file-level `PI` was used only inside `angle_to_arrow_direction`; move it there as a local `constexpr float pi`.

The adapter's `getTrafficSignals` override is intentionally kept in camelCase because it implements `ClassifierInterface`; only the interface-free core adopts snake_case.

## Related links

None.

## How was this PR tested?

`colcon build` + `colcon test --packages-select autoware_traffic_light_classifier` on an RTX 5080 with the real TensorRT engine (`traffic_light_lamp_recognizer_comlops.onnx`): **77 tests, 0 errors, 0 failures, 0 skipped**. Both the GPU decode path and the GPU-free core unit tests run.

## Notes for reviewers

No logic changes; the debug output is unchanged. The diff is dominated by identifier renames, plus the two structural tweaks noted above (the non-destructive `make_debug_image` and the `MsgTE` alias removal). `clang-format` wrapped the `decode_tlr_output` declaration because the longer name pushed it past 100 columns.

## Interface changes

None at the package boundary. `CnnLampRecognizerCore` is an internal class under `src/` (not an installed public header), so its method-name and signature changes are package-internal. The ROS-facing `ClassifierInterface::getTrafficSignals` is unchanged.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_traffic_light_classifier/src/classifier/cnn_lamp_recognizer.cpp
+++ b/perception/autoware_traffic_light_classifier/src/classifier/cnn_lamp_recognizer.cpp
@@ -33,38 +33,38 @@
 namespace autoware::traffic_light
 {
 
-using MsgTE = tier4_perception_msgs::msg::TrafficLightElement;
+using tier4_perception_msgs::msg::TrafficLightElement;
 namespace
 {
-constexpr int kDebugImageWidth = 200;
-constexpr int kDebugTextHeight = 50;
-constexpr float PI = static_cast<float>(autoware_utils::pi);
-ArrowDirection angleToArrowDirection(float angle_rad)
+constexpr int debug_image_width = 200;
+constexpr int debug_text_height = 50;
+ArrowDirection angle_to_arrow_direction(float angle_rad)
 {
-  if (angle_rad >= -PI / 8.0f && angle_rad < PI / 8.0f) return ArrowDirection::UP_ARROW;
-  if (angle_rad >= PI / 8.0f && angle_rad < 3.0f * PI / 8.0f) return ArrowDirection::UP_RIGHT_ARROW;
-  if (angle_rad >= 3.0f * PI / 8.0f && angle_rad < 5.0f * PI / 8.0f)
+  constexpr float pi = static_cast<float>(autoware_utils::pi);
+  if (angle_rad >= -pi / 8.0f && angle_rad < pi / 8.0f) return ArrowDirection::UP_ARROW;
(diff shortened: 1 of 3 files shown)


a-maumau (approved):
LGTM

PR #13031, opened 2026-07-21
feat: update how to calculate distance

TetsuKawa:
## Description

Switch autoware_path_distance_calculator from measuring the remaining distance along the behavior-planner path topic to computing it from the lanelet2 route/map and the self-position.

The motivation is that introducing the Diffusion Planner pipeline stops the publication of `/planning/scenario_planning/lane_driving/behavior_planning/path` (that pipeline does not run the classic behavior planner), so the node could no longer compute the remaining distance to the goal in that configuration.

As a replacement, the node now resolves the shortest path (in lanelets) from the self-position to the route goal using the lanelet2 routing graph built from the vector map, and sums the lanelet lengths along it, following the same approach as `autoware_remaining_distance_time_calculator`. This only depends on the route and the map, both of which are published regardless of which trajectory-generation pipeline is used.

As part of this change, the node's implementation was also split into a ROS-independent calculation class (`RouteDistanceCalculator`, in src/main.cpp) and a thin ROS wrapper (src/ros_interface.cpp) that polls the map/route/self-pose and forwards them to the calculator.

## Related links

**Private Links:**

- [TIER IV internal link](https://star4.slack.com/archives/C0657FNJ5EG/p1780971149044449)

## How was this PR tested?

Tested on tier4 shuttle bus type Pilot-Auto X2 Psim, and confirmed that the remaining distance to the goal is published.

## Notes for reviewers

None.

## Interface changes

None.

#### Additions and removals

| Change type | Topic Type | Topic Name                                                                          | Message Type                                | Description                                            |
| :---------- | :--------- | :----------------------------------------------------------------------------------- | :------------------------------------------- | :------------------------------------------------------ |
| Removed     | Sub        | `~/input/path` (`/planning/scenario_planning/lane_driving/behavior_planning/path`)   | `autoware_planning_msgs::msg::Path`          | Reference path used to compute the remaining distance   |
| Added       | Sub        | `~/input/route` (`/planning/mission_planning/route`)                                 | `autoware_planning_msgs::msg::LaneletRoute`  | Route, used to resolve the goal and the shortest path    |
| Added       | Sub        | `~/input/map` (`/map/vector_map`)                                                    | `autoware_map_msgs::msg::LaneletMapBin`      | Lanelet2 map, used to build the routing graph            |

#### Modifications

None.

#### Additions and removals

None.

#### Modifications

None.

## Effects on system behavior

The published `~/distance` value now represents the remaining arc-length to the route goal along the lanelet2 shortest path, instead of the arc-length to the end of the behavior-planner path. Under normal driving these should be numerically close, but the new calculation is resolved once when the route is received and is not recomputed while driving that same route; see the package's Readme.md for the known limitations.


isamu-takagi (changes requested):
I found a bug where the distance is not calculated according to the route. This is because the route is calculated using `getRoute` based on the current pose and the goal pose. The route has already been calculated and is in LaneletRoute.msg. So get the lanelets from the IDs contained in LaneletRoute.msg.

Image
Image

TetsuKawa:
@isamu-takagi 
Thanks for the detailed explanation — that makes sense.

I've made the fix as suggested: instead of re-deriving the path with our own shortest-path search (`RoutingGraph::getRoute()`), set_route() now looks up the lanelets directly from the IDs in `LaneletRoute::segments[]`.preferred_primitive.id. This should ensure the distance is always calculated along the actual planned route rather than a potentially different lanelet2-derived shortest path.

One note: I haven't been able to reproduce the original bug (distance not following the route) in my own environment — possibly due to a difference in the map version I'm testing with. If you have a reproduction case or a specific map/route where the discrepancy shows up, could you try it on your end to confirm the fix resolves it? Happy to dig further if it's still an issue.

isamu-takagi (approved):
LGTM

Image

PR #13033, opened 2026-07-22
refactor(autoware_control_evaluator): migrate to polling:: API

Koichi98:
## Description

Migrate the polling subscribers from the deprecated macro API (`AUTOWARE_POLLING_SUBSCRIBER_PTR` and the `create_polling_subscriber` node member) to the new free-function API `autoware::agnocast_wrapper::polling::create_polling_subscriber` in `polling_subscriber.hpp`. Pure refactor, no behavior change.

## How was this PR tested?

Built with `ENABLE_AGNOCAST=0` and `=1`. Ran `planning_simulator` with a route set and confirmed `/control/control_evaluator/metrics` publishes at the same rate in both builds.

## Related links

None.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


xtk8532704 (approved):
LGTM

PR #13036, opened 2026-07-22
refactor(autoware_map_based_prediction): migrate to polling:: API

Koichi98:
## Description

Migrate the traffic-signal polling subscriber from the deprecated macro API (`AUTOWARE_POLLING_SUBSCRIBER_PTR` and the `create_polling_subscriber` node member) to the new free-function API `autoware::agnocast_wrapper::polling::create_polling_subscriber` in `polling_subscriber.hpp`. Since the new `take_data()` returns a plain `std::shared_ptr<const T>`, `trafficSignalsCallback` is retyped to take `std::shared_ptr<const T>`. The real subscription callbacks are left unchanged. Pure refactor, no behavior change.

## How was this PR tested?

Built with `ENABLE_AGNOCAST=0` and `=1`. Ran `logging_simulator` with the sample rosbag and confirmed `/perception/object_recognition/objects` publishes at the same rate in both builds.

## Related links

None.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


technolojin (approved):
note: requires latest autoware_core

PR #13039, opened 2026-07-22
perf(crosswalk): skip polygon intersections for disjoint bounds

hugogo1998:
## Description

`CrosswalkModule::getCollisionPoint()` builds a convex hull and runs a polygon intersection for every predicted-path step. Many of those steps are far from the crosswalk attention area and cannot intersect it.

This change adds an AABB broad-phase check before that polygon work. The attention-area bounding box is computed once, and each object step uses a conservative bounding box based on the object half-diagonal. When the boxes are disjoint, the convex hull and exact intersection are skipped.

## Related links

**Parent Issue:**

- None — small, self-contained performance optimization.

## How was this PR tested?

A standalone harness using the relevant functions from `scene_crosswalk.cpp` and real Boost.Geometry types was run across different object counts and predicted-path spans.

| Workload | Base (ns/cycle) | This PR (ns/cycle) | Speedup |
|---|---:|---:|---:|
| 6 objects, 20 m span | 318504 | 155898 | 2.04x |
| 12 objects, 20 m span | 632367 | 311142 | 2.03x |
| 24 objects, 20 m span | 1179034 | 459637 | 2.57x |
| 12 objects, 8 m span | 314793 | 179056 | 1.76x |
| 12 objects, 40 m span | 686740 | 323504 | 2.12x |

The accumulated collision-point state matched the base implementation in every tested workload.

All applicable pre-commit hooks passed on the changed file, including `clang-format` and `cpplint`.

Not run: a full `colcon` build or Autoware scenario test.

## Notes for reviewers

The object box uses the shape half-diagonal, so it remains conservative regardless of object orientation. The broad-phase logic is isolated in a helper to avoid increasing the complexity of `getCollisionPoint()`.

## Interface changes

None.

## Effects on system behavior

No expected change to collision points. Crosswalk collision checks require less computation when predicted object steps are outside the attention area.

## AI usage and verification

AI usage: I used OpenAI Codex to identify and implement this optimization and to help prepare the standalone benchmark.

Self-review: I personally reviewed the full diff and benchmark setup and confirmed that I understand every changed line.

Verification: The benchmark and checks described above were run and reviewed by me.

### DIFF
--- a/planning/behavior_velocity_planner/autoware_behavior_velocity_crosswalk_module/src/scene_crosswalk.cpp
+++ b/planning/behavior_velocity_planner/autoware_behavior_velocity_crosswalk_module/src/scene_crosswalk.cpp
@@ -788,15 +788,35 @@ std::optional<CollisionPoint> CrosswalkModule::getCollisionPoint(
 
   double minimum_stop_dist = std::numeric_limits<double>::max();
   std::optional<CollisionPoint> nearest_collision_point{std::nullopt};
+
+  // Precompute a bounding box of the attention area once, plus the object's bounding radius, for
+  // a cheap broad-phase reject in the per-step loop below (profiling showed most steps are far
+  // from the crosswalk, so their polygon intersection is empty -- wasted convex_hull + intersection).
+  boost::geometry::model::box<Point2d> attention_bbox;
+  boost::geometry::envelope(attention_area, attention_bbox);
+  const double obj_bounding_radius =
+    0.5 * std::hypot(object.shape.dimensions.x, object.shape.dimensions.y);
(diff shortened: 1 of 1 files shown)


hugogo1998:
Hi @satoshi-ota @Kotakku @yuki-takagi-66 , could you please take a look when you have time? I'd appreciate any feedback, and pls let me know if I overlook anything. Thanks

hugogo1998:
Hi @Kotakku @yuki-takagi-66 , could you please take a look when you have time? I'd appreciate any feedback, and pls let me know if I overlook anything. Thanks

PR #13041, opened 2026-07-22
refactor(autoware_traffic_light_classifier): declare classifier params in the node

takam5f2:
## Description

First step of an incremental, by-concern campaign to remove the ROS Node dependency from the three traffic-light classifier implementations (`ColorClassifier`, `CNNClassifier`, `CnnLampRecognizer`). Each classifier already has a Node-free `…Core`; the remaining ROS coupling lives in the thin adapters, which take an `rclcpp::Node *` to declare parameters, publish debug images, log, and (Color only) register dynamic reconfigure.

This PR moves **parameter declaration** out of the classifier files:

- New node-side translation unit `src/classifier_params.{hpp,cpp}` holding `declare_hsv_config`,
  `declare_cnn_config` (+ `read_label_file`), and `declare_lamp_config`, moved verbatim from the
  adapter anonymous namespaces (the CNN/lamp helpers are `#if ENABLE_GPU`-guarded so the non-GPU,
  color-only library build still compiles). It is the only classifier-side file that includes
  `rclcpp`.
- It lives at the node layer (`src/`, next to `traffic_light_classifier_node.cpp`) rather than
  under `src/classifier/`, because declaring parameters is a Node concern and both consumers
  (`traffic_light_classifier_node` and `single_image_debug_inference_node`) are Node classes.
- Each adapter constructor changes from `(rclcpp::Node *)` to `(rclcpp::Node *, const Config &)`;
  the node builds the config via `declare_*_config(this)` and passes it in. The classifier files no
  longer call `declare_parameter` or read the label file.

No behavior change: the same parameters are declared with the same names, defaults, and types.

## Related links

None.

## How was this PR tested?

`colcon build` and `colcon test` for `autoware_traffic_light_classifier` on a GPU environment: 83 tests, 0 failures, 0 skipped. The existing ROS-isolated adapter tests were updated to build the adapter via `(node, declare_*_config(node))`, so parameter declaration is still exercised end to end.

## Notes for reviewers

This is the first of a small series of by-concern PRs that progressively remove the Node dependency from the classifier layer (parameters → debug image → logging/reconfigure → adapter collapse). It is intentionally minimal: only parameter declaration moves here.

## Interface changes

None. (No ROS topics or parameters are added, removed, or renamed; only the C++ constructor signatures of the internal classifier adapters change.)

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_traffic_light_classifier/CMakeLists.txt
+++ b/perception/autoware_traffic_light_classifier/CMakeLists.txt
@@ -57,6 +57,7 @@ if(TRT_AVAIL AND CUDA_AVAIL)
   )
 
   ament_auto_add_library(${PROJECT_NAME} SHARED
+    src/classifier_params.cpp
     src/classifier/color_classifier.cpp
     src/classifier/cnn_classifier.cpp
     src/classifier/cnn_lamp_recognizer.cpp
@@ -79,6 +80,7 @@ if(TRT_AVAIL AND CUDA_AVAIL)
   )
 
   ament_auto_add_library(single_image_debug_inference_node SHARED
+    src/classifier_params.cpp
     src/classifier/cnn_classifier.cpp
     src/classifier/color_classifier.cpp
     src/classifier/cnn_lamp_recognizer.cpp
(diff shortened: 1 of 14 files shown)


a-maumau (approved):
LGTM!

PR #13045, opened 2026-07-22
fix(universe_utils): prevent infinite recursion in `alt_geometry`'s `convex_hull()`

mitukou1109:
## Description

This PR fixes `convex_hull()` and related functions in `alt_geometry`, which were reported to cause infinite recursion in ARM64 environments.

### Changes
* Used epsilon for sign determination and added a tie-break for uniform point selection to resolve the recursion issue.
* Removed an unintended iterator operation from `is_convex()`.
* Added unit tests covering edge cases.

## Related links

**Parent Issue:**

- https://github.com/autowarefoundation/autoware_universe/issues/12270

## How was this PR tested?

Tests on my AMD64 laptop are confirmed to pass.

```
colcon test --event-handlers console_cohesion+ --packages-select autoware_universe_utils
```

ARM64 tests are covered by ~~CI~~ #13116.


mitukou1109:
https://github.com/autowarefoundation/autoware_universe/actions/runs/30606516857/job/91080544053?pr=13116
Tests on ARM64 runner passed

common/autoware_universe_utils/src/geometry/alt_geometry.cpp
```diff
@@ -168,8 +168,13 @@ std::optional<alt::ConvexPolygon2d> convex_hull(const alt::Points2d & points)
 
   // QuickHull algorithm
 
-  const auto p_minmax_itr = std::minmax_element(
-    points.begin(), points.end(), [](const auto & a, const auto & b) { return a.x() < b.x(); });
+  const auto p_minmax_itr =
+    std::minmax_element(points.begin(), points.end(), [](const auto & a, const auto & b) {
+      if (a.x() != b.x()) {
```
sasakisasaki, line 173: I guess we can use `epsilon` similar to [this place](https://github.com/autowarefoundation/autoware_universe/pull/13045/changes#diff-34b267ca053829ae1c98ce684f3a671626c691371cacc4d4eddbdfec23e1dbe7R408-R411)

common/autoware_universe_utils/src/geometry/alt_geometry.cpp
sasakisasaki, line 644: The cases `cross == epsilon` and `cross == -epsilon` are missing. So let's do something like this 👍

- Change `>, <` into `>=, <=`
- Then leave `std::abs(cross) < epsilon` as is

common/autoware_universe_utils/src/geometry/alt_geometry.cpp
```diff
@@ -618,10 +631,10 @@ bool within(const alt::Point2d & point, const alt::ConvexPolygon2d & poly)
     const auto end_vec = point - p2;
     const auto cross = start_vec.cross(end_vec);
 
-    if (is_upward_edge && cross > 0) {  // point is to the left of edge
+    if (is_upward_edge && cross > epsilon) {  // point is to the left of edge
```
sasakisasaki, line 634:
```suggestion
    if (is_upward_edge && cross >= epsilon) {  // point is to the left of edge
```

common/autoware_universe_utils/src/geometry/alt_geometry.cpp
```diff
@@ -618,10 +631,10 @@ bool within(const alt::Point2d & point, const alt::ConvexPolygon2d & poly)
...
       winding_number++;
       continue;
-    } else if (is_downward_edge && cross < 0) {  // point is to the left of edge
+    } else if (is_downward_edge && cross < -epsilon) {  // point is to the left of edge
```
sasakisasaki, line 637:
```suggestion
    } else if (is_downward_edge && cross <= -epsilon) {  // point is to the left of edge
```

common/autoware_universe_utils/test/src/geometry/test_alt_geometry.cpp
```diff
@@ -93,6 +84,45 @@ TEST(alt_geometry, DISABLED_convexHull)
...
+  {  // Duplicate points
+    Points2d points = {{0.0, 0.0}, {0.0, 0.0}, {0.0, 2.0}, {2.0, 0.0}, {2.0, 2.0}, {2.0, 2.0}};
+    const auto result = convex_hull(points);
+    ASSERT_TRUE(result);
+    EXPECT_GE(result->vertices().size(), 4U);
+  }
+
+  {  // Near-collinear points (floating point precision noise)
+    Points2d points = {{0.0, 0.0}, {1.0, 1.0 + 1e-12}, {2.0, 2.0 - 1e-12},
+                       {3.0, 3.0}, {0.0, 3.0},         {3.0, 0.0}};
```
sasakisasaki, line 121: (Just a comment. Not a change request) Effective test, nice 👍

sasakisasaki (approved):
Reviewed. I agree with merging this PR while put some quick win stuffs 👍 

Thank you so much for fixing the infinite test loop bug!

common/autoware_universe_utils/src/geometry/alt_geometry.cpp
mitukou1109: Upon checking, I think the condition here is correct but the one in `find_farthest()` is wrong.
For tie-breaking (deterministically choosing the same point regardless of the environment, e.g., amd64 vs. arm64, when coordinates are nearly identical), using epsilon-based equality checks breaks transitivity. Therefore, strict lexicographical comparison should be used in `find_farthest()` too.

common/autoware_universe_utils/src/geometry/alt_geometry.cpp
mitukou1109: Thank you for pointing it out. I will change `<` in `std::abs(cross) < epsilon` to `<=`.

mitukou1109:
@sasakisasaki Could you check my comments?

PR #13047, opened 2026-07-22
refactor(autoware_traffic_light_classifier): move debug-image publishing to the node

takam5f2:
## Description

Second step of the by-concern campaign to remove the ROS Node dependency from the traffic-light
classifiers (follows #13041, which moved parameter declaration to the node).

This PR moves **debug-image publishing** out of the three classifiers and into the node:

- `ClassifierInterface` gains a value-returning `cv::Mat make_debug_image(const std::vector<cv::Mat> &)`,
  which renders one composite RGB debug frame for the batch from the most recent `getTrafficSignals`
  call. Each classifier stops owning an `image_transport::Publisher`; the publish block is removed.
- The node owns a single `image_transport::Publisher` on `~/output/debug/image` and publishes only
  when a debug consumer is attached (subscriber-gated, so the composite is built only on demand).
- `TrafficLightClassifier` exposes the cropped ROI views via `Result::roi_images` and forwards
  `make_debug_image`, so the node can request the debug view without re-implementing ROI cropping.

Each classifier keeps the per-image debug rendering it already had; only the *publishing* moves.

## Related links

- Follows #13041 (parameter declaration → node).

## How was this PR tested?

`colcon build` and `colcon test` for `autoware_traffic_light_classifier` on GPU environment: 84 tests, 0 failures, 0 skipped. A new ROS-free wrapper test
(`ExposesRoiImagesAndForwardsDebug`) pins that `classify()` exposes the cropped ROIs and that `make_debug_image` is forwarded to the backend.

## Notes for reviewers

Part of a small by-concern series (parameters → debug image → logging/reconfigure → adapter collapse) that progressively removes the Node dependency from the classifier layer.

## Interface changes

No ROS interface change in topic type or name. One debug topic is renamed for consistency (see below).

#### Modifications

| Version | Topic Type | Topic Name              | Message Type        | Description                          |
|:--------|:-----------|:------------------------|:--------------------|:-------------------------------------|
| Old     | Pub        | `~/debug/image`         | `sensor_msgs/Image` | Color (HSV) classifier debug image   |
| New     | Pub        | `~/output/debug/image`  | `sensor_msgs/Image` | Color (HSV) classifier debug image   |

The CNN and lamp backends already published on `~/output/debug/image`; the color backend's
`~/debug/image` is unified onto the same topic (no launch/rviz config referenced the old name).

## Effects on system behavior

Only the debug visualization changes. The primary output `~/output/traffic_signals` is unchanged
in content, type, and timing.

For one camera frame with N target-type ROIs, the debug image on `~/output/debug/image` changes as
follows:

| Backend     | Debug topic (old → new)                  | Messages / frame (old → new) | Encoding (old → new) |
|:------------|:-----------------------------------------|:-----------------------------|:---------------------|
| Color (HSV) | `~/debug/image` → `~/output/debug/image` | N → 1 (stacked vertically)   | `bgr8` → `rgb8`      |
| CNN         | `~/output/debug/image` (unchanged)       | N → 1 (stacked vertically)   | `rgb8` (unchanged)   |
| Lamp        | `~/output/debug/image` (unchanged)       | 1 (unchanged)                | `rgb8` (unchanged)   |

- **Color debug topic is unified** onto `~/output/debug/image`; a viewer still subscribed to the old
  `~/debug/image` receives nothing until it re-subscribes. CNN and lamp topics are unchanged.
- **CNN and Color now publish one composite frame** (the per-ROI views stacked vertically) instead of
  one message per ROI. The lamp backend already composited, so it is unchanged.
- **Color debug colors now display correctly.** The mosaic's raw-ROI pixels are RGB-ordered, so the
  old `bgr8` label swapped R/B in viewers; publishing as `rgb8` corrects the label. The pixel buffer
  is unchanged — this is a label fix, not a pixel change.
- **Debug publishing stays subscriber-gated.** The composite is built and published only when
  `~/output/debug/image` has a subscriber (a cold path); with no subscriber, no work is done. Only
  the publisher owner changed (per-classifier publisher → a single node-owned publisher).
- **Debug rendering can no longer affect the primary output.** The debug build/publish now runs after
  the `~/output/traffic_signals` publish and the diagnostics publish, so a failure there (e.g. an
  OpenCV exception) cannot skip the primary signal output or diagnostics.

### DIFF
--- a/perception/autoware_traffic_light_classifier/src/classifier/classifier_interface.hpp
+++ b/perception/autoware_traffic_light_classifier/src/classifier/classifier_interface.hpp
@@ -29,8 +29,12 @@ class ClassifierInterface
 public:
   virtual ~ClassifierInterface() = default;
   virtual bool getTrafficSignals(
-    const std::vector<cv::Mat> & input_image,
+    const std::vector<cv::Mat> & images,
     tier4_perception_msgs::msg::TrafficLightArray & traffic_signals) = 0;
+
+  // One composite RGB debug view for the batch, rendered from the most recent getTrafficSignals
+  // call. Returns an empty Mat when there is nothing to show.
+  virtual cv::Mat make_debug_image(const std::vector<cv::Mat> & images) const = 0;
 };
 }  // namespace autoware::traffic_light
 
--- a/perception/autoware_traffic_light_classifier/src/classifier/cnn_classifier.cpp
+++ b/perception/autoware_traffic_light_classifier/src/classifier/cnn_classifier.cpp
@@ -151,13 +151,11 @@ cv::Mat CNNClassifierCore::make_debug_image(
 }
 
 // ============================== CNNClassifier ==============================
-// ROS adapter: publishes debug images, logs, and delegates classification to the Node-free core.
+// ROS adapter: logs and delegates classification and debug-image rendering to the Node-free core.
 
 CNNClassifier::CNNClassifier(rclcpp::Node * node_ptr, const CNNConfig & config)
 : node_ptr_(node_ptr), core_(config)
 {
-  image_pub_ = image_transport::create_publisher(
-    node_ptr_, "~/output/debug/image", rclcpp::QoS{1}.get_rmw_qos_profile());
 }
 
 bool CNNClassifier::getTrafficSignals(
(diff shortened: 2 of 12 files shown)


a-maumau (approved):
LGTM

PR #13050, opened 2026-07-23
feat(autoware_multi_object_tracker): orientation sign belief for heading flip decisions

technolojin:
## Description

Vehicle (and bicycle) trackers can lock onto a heading that is 180° off when detections have ambiguous orientation. Previously the only correction was an instantaneous rule in `BicycleMotionModel::limitStates()` that flipped the state when the longitudinal velocity exceeded a reverse-velocity threshold, which reacts to a single noisy velocity estimate and can oscillate.

This PR replaces that with a probabilistic heading-sign belief (`OrientationSignBelief`) that accumulates evidence over time and flips the full tracker state only when the sign is wrong with high confidence:

- **Detection-yaw votes**: each measurement with usable orientation votes on whether its raw yaw agrees with the tracker yaw (before the yaw is normalized away). Votes accumulate in a scalar OU-Kalman agreement estimate that relaxes toward an uninformed prior, with `AVAILABLE` orientation weighted stronger than `SIGN_UNKNOWN`, and a dead zone around ±90° where boxes carry axis ambiguity instead of sign evidence.
- **Velocity evidence**: the tracked longitudinal velocity contributes memorylessly at decision time, weighted by speed and velocity certainty, so only sustained, well-estimated reverse motion can override the yaw evidence.
- **Consistent flip**: `flipOrientation()` on the bicycle/CTRV motion models applies the exact 180° state transform (axle-point swap / yaw rotation, velocity negation, covariance transform `T·P·Tᵀ` including the velocity cross-covariance signs), and the stored footprint and the belief itself are flipped together. Negation of the belief on flip provides hysteresis against oscillation.
- **Multi-layer sync**: the two layers of `MultipleVehicleTracker` are aligned to the layer with the stronger sign belief; the pedestrian layer of `PedestrianAndBicycleTracker` follows the bicycle layer.

The reverse-velocity flip rule in `limitStates()` and the corresponding footprint flip handling in `VehicleTracker::updateKinematics()` are removed.

## Related links

**Parent Issue:**

- None

[TIER IV INTERNAL ](https://tier4.atlassian.net/browse/RDDR-911)

## How was this PR tested?

* before

https://github.com/user-attachments/assets/842fa6aa-13eb-4c17-8fa1-53831e5ada6b

* after

https://github.com/user-attachments/assets/506838b3-e294-4514-9605-a9bd37d4544d

Unit tests added in `test_vehicle_tracker.cpp` covering:

- `OrientationSignBelief`: seeding per orientation availability, vote weighting, dead zone, order-independent fusion of simultaneous votes, OU decay, velocity gating at standstill, velocity override vs. uncertain velocity, resistance of accumulated yaw evidence, flip-negation hysteresis
- `BicycleMotionModel::flipOrientation` / `CTRVMotionModel::flipOrientation`: state and covariance transform correctness
- `VehicleTracker`: end-to-end flip behavior through `measure()`

```bash
colcon test --packages-select autoware_multi_object_tracker
```

## Notes for reviewers

None.

## Interface changes

None. The belief tuning constants are class-independent compile-time defaults (`OrientationSignBeliefParams`), not ROS parameters.

## Effects on system behavior

Tracked vehicle heading no longer flips on a single-frame reverse-velocity threshold; a 180° flip now requires confidently negative fused yaw/velocity evidence, reducing heading-sign oscillation for objects with ambiguous detections.

### DIFF
--- a/perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/tracker/trackers/pedestrian_tracker.hpp
+++ b/perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/tracker/trackers/pedestrian_tracker.hpp
@@ -55,6 +55,15 @@ public:
     geometry_msgs::msg::Twist & twist, std::array<double, 36> & twist_cov) const override;
   rclcpp::Time getStateTime() const override { return motion_model_.getLastPredictionTime(); }
 
+  // Heading state and 180° state flip; composite trackers use these for heading-sign
+  // consensus across their layers.
+  double getYawState() const { return motion_model_.getYawState(); }
+  void flipOrientationSign()
+  {
+    motion_model_.flipOrientation();
+    removeCache();
+  }
+
   ShapeModelBase & getShapeModel() override { return shape_model_; }
   const ShapeModelBase & getShapeModel() const override { return shape_model_; }
   void assembleShapeTo(types::DynamicObject & output, bool /*to_publish*/) const override
--- a/perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/tracker/update/orientation_sign_belief.hpp
+++ b/perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/tracker/update/orientation_sign_belief.hpp
@@ -55,10 +55,13 @@ public:
 
   // Forward motion agrees with the heading sign, reverse motion disagrees. The vote weight grows
   // linearly with speed and matches an AVAILABLE yaw vote at vote_vel_par, so a slow reverse
-  // maneuver (e.g. parking) barely moves the belief while yaw votes dominate.
-  void voteVelocity(const double vel_long)
+  // maneuver (e.g. parking) barely moves the belief while yaw votes dominate. The vote is
+  // discounted by the velocity variance, halving at vote_vel_var; a poorly converged velocity
+  // estimate carries little weight.
+  void voteVelocity(const double vel_long, const double vel_var)
   {
-    log_odds_ += params_.vote_available * vel_long / params_.vote_vel_par;
+    const double confidence = params_.vote_vel_var / (params_.vote_vel_var + vel_var);
+    log_odds_ += params_.vote_available * confidence * vel_long / params_.vote_vel_par;
     log_odds_ = std::clamp(log_odds_, -params_.log_odds_max, params_.log_odds_max);
   }
 
--- a/perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/tracker/update/orientation_sign_belief.hpp
+++ b/perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/tracker/update/orientation_sign_belief.hpp
@@ -18,63 +18,104 @@
 #include "autoware/multi_object_tracker/object_model/object_model.hpp"
 #include "autoware/multi_object_tracker/types.hpp"
 
+#include <rclcpp/time.hpp>
+
 #include <algorithm>
 #include <cmath>
 
 namespace autoware::multi_object_tracker
 {
 
-// Log-odds belief that the tracker's heading sign is correct. Each detection votes by its raw yaw
-// sign and the tracked longitudinal velocity votes by its sign; a strongly negative belief
-// triggers a 180° flip, and negation on flip provides hysteresis against oscillation.
+// Heading-sign belief with two separated evidence sources. Detection-yaw votes accumulate in a
+// scalar OU-Kalman agreement estimate (+1 = sign correct, -1 = wrong) that relaxes toward an
+// uninformed prior over time; the tracked longitudinal velocity contributes memorylessly at
+// decision time, weighted by speed and velocity certainty. A confidently negative fused
+// agreement triggers a 180° flip, and negation on flip provides hysteresis against oscillation.
 class OrientationSignBelief
 {
 public:
   OrientationSignBelief(
     const object_model::OrientationSignBelief & params,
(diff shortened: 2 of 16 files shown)


perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/tracker/trackers/pedestrian_tracker.hpp
```diff
@@ -55,6 +55,15 @@ class PedestrianTracker : public Tracker
     geometry_msgs::msg::Twist & twist, std::array<double, 36> & twist_cov) const override;
   rclcpp::Time getStateTime() const override { return motion_model_.getLastPredictionTime(); }
 
+  // Heading state and 180° state flip; composite trackers use these for heading-sign
+  // consensus across their layers.
+  double getYawState() const { return motion_model_.getYawState(); }
+  void flipOrientationSign()
```
YoshiRi, line 61: [Comment]
VehicleTracker::flipOrientationSign() flips the stored footprint together with yaw, but PedestrianTracker::flipOrientationSign() only flips the CTRV state.
Is it right?

perception/autoware_multi_object_tracker/lib/tracker/motion_model/bicycle_motion_model.cpp
```diff
@@ -367,44 +367,60 @@ bool BicycleMotionModel::updateStateLength(
   return true;
 }
 
+void BicycleMotionModel::flipStateOrientation(StateVec & X, StateMat & P) const
+{
+  // rotate the object orientation by 180 degrees
+  // replace X1 and Y1 with X2 and Y2
```
YoshiRi, line 373: [question]
flipStateOrientation() algebraically reduces to swapping X1/X2 and Y1/Y2; the computed center does not affect the result. With unequal lf_ratio and lr_ratio, this also changes the model-defined body center. Is equal front/rear ratio an invariant?

perception/autoware_multi_object_tracker/lib/tracker/trackers/vehicle_tracker.cpp
```diff
@@ -349,6 +375,8 @@ bool VehicleTracker::conditionedUpdate(
     shape_model_.updateFootprint(
       measurement, measurement_time, has_pose ? std::make_optional(tracker_pose) : std::nullopt);
 
+    evaluateSignFlip(measurement_time);
```
YoshiRi, line 378: [comment]
conditionedUpdate() can be entered not only for channels with untrusted orientation, but also as the TRY_EXTENSION fallback for a trustworthy full-box measurement. In the latter case, the measurement may satisfy the same orientation-availability and trust_orientation conditions used by measure(), but its raw yaw never updates sign_belief_. Since alignClusterToOrientation() may discard the original sign information, should the gated raw-yaw vote be shared between measure() and conditionedUpdate() and applied before alignment?

perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/tracker/trackers/pedestrian_tracker.hpp
technolojin: this is by design. pedestrian tracker has unlinked orientation between tracker and footprint. therefore, footprint only translated in parallel.

perception/autoware_multi_object_tracker/lib/tracker/motion_model/bicycle_motion_model.cpp
technolojin: the following code of line 374-388 is to consider the unequal lf_ratio and lr_ratio

perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/tracker/update/orientation_sign_belief.hpp
```diff
@@ -0,0 +1,141 @@
...
+
+namespace autoware::multi_object_tracker
+{
+
+// Heading-sign belief with two separated evidence sources. Detection-yaw votes accumulate in a
+// scalar OU-Kalman agreement estimate (+1 = sign correct, -1 = wrong) that relaxes toward an
+// uninformed prior over time; the tracked longitudinal velocity contributes memorylessly at
+// decision time, weighted by speed and velocity certainty. A confidently negative fused
+// agreement triggers a 180° flip, and negation on flip provides hysteresis against oscillation.
+// Class-independent tuning constants shared by every vehicle tracker.
```
ogrex: Could you explain why this is class-independent? Or is it just to keep things simple for now?

ogrex (approved):
LGTM. 
Added some non-blocking comments

perception/autoware_multi_object_tracker/lib/tracker/trackers/vehicle_tracker.cpp
technolojin: TRY_EXTENSION path is to split sudden size change. I want to split the paths so that orientation ambiguity filter and size filters are linked and cause unexpected result. let me keep current behavior.

perception/autoware_multi_object_tracker/include/autoware/multi_object_tracker/tracker/update/orientation_sign_belief.hpp
technolojin: it was an old implementation description. revised https://github.com/autowarefoundation/autoware_universe/pull/13050/changes/4a6d17838d42804c1ae6578322566417b98dbee9

PR #13056, opened 2026-07-23
refactor(autoware_traffic_light_classifier): move logging and dynamic reconfigure to the node

takam5f2:
## Description

Third step of the by-concern campaign to remove the ROS Node dependency from the traffic-light classifiers (follows #13041 parameter declaration and #13047 debug-image publishing).

This PR moves **logging and dynamic reconfigure** out of the three classifiers, so the classifier layer no longer depends on the ROS Node at all:

- Each classifier's size-guard / inference-failure paths now just `return false` instead of logging.
  The node's existing single `RCLCPP_ERROR("failed classify image, abort callback")` on a null
  classify result covers the failure path (it previously double-logged: the classifier logged a
  backend-specific message *and* the node logged the generic one).
- Dynamic reconfigure moves to the node: `ColorClassifier` exposes `get_config` / `set_config`
  pass-throughs to its core, and the node registers a named on-set-parameters callback
  (`onSetParametersCallback`) that applies the HSV threshold read-modify-write.
- Each adapter constructor now takes only its plain `Config` (no `rclcpp::Node *`), and
  `#include <rclcpp/rclcpp.hpp>` is dropped from the three classifier headers.

## Related links

- Follows #13041 (parameters → node),
-  and #13047 (debug image → node).

## How was this PR tested?

`colcon build` and `colcon test` for `autoware_traffic_light_classifier` on an NVIDIA-based GPU environment with the
real TensorRT engines: 84 tests, 0 failures, 0 skipped. The ROS-isolated adapter tests were updated
to construct the adapters from a plain `Config` (`declare_*_config(node)`), and dynamic reconfigure
remains exercised end to end by `test_traffic_light_classifier_integration`.

## Notes for reviewers

Part of a small by-concern series (parameters → debug image → logging/reconfigure → adapter
collapse) that progressively removes the Node dependency from the classifier layer. After this PR
the three classifier translation units are free of `rclcpp`.

## Interface changes

None. (No ROS topics or parameters are added, removed, or renamed; the `traffic_light_type` HSV dynamic-reconfigure parameters are declared and handled exactly as before, just from the node.)

## Effects on system behavior

- On a classification failure, a single generic error (`"failed classify image, abort callback"`)
  is logged by the node instead of a backend-specific message. The previous behavior logged twice
  for one failure (the classifier's specific message plus the node's generic one); this PR removes
  that duplication and keeps the node's single log. Backend-specific failure text (e.g. "failed to
  classify … by cnn", "LampRecognizer: inference failed") is no longer emitted.
- The size-mismatch warning (`"image number should be equal to traffic signal number!"`) is
  removed; the guard still returns early, it just no longer logs (the condition is unreachable in
  practice, since the wrapper builds the image and signal arrays together).
- HSV dynamic reconfigure (`ros2 param set` on the color thresholds at runtime) behaves exactly as
  before; only the code that applies it moved from the classifier to the node.
- The `~/output/traffic_signals` output is unchanged.

### DIFF
--- a/perception/autoware_traffic_light_classifier/src/classifier/cnn_classifier.cpp
+++ b/perception/autoware_traffic_light_classifier/src/classifier/cnn_classifier.cpp
@@ -151,10 +151,10 @@ cv::Mat CNNClassifierCore::make_debug_image(
 }
 
 // ============================== CNNClassifier ==============================
-// ROS adapter: logs and delegates classification and debug-image rendering to the Node-free core.
+// Node-free adapter: delegates classification and debug rendering to the core and maps its
+// per-image output into the caller's signals.
 
-CNNClassifier::CNNClassifier(rclcpp::Node * node_ptr, const CNNConfig & config)
-: node_ptr_(node_ptr), core_(config)
+CNNClassifier::CNNClassifier(const CNNConfig & config) : core_(config)
 {
 }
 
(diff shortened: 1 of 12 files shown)


takam5f2:
@a-maumau  
Could you check the PR? This is the simplest PR in the series of `traffic_light_classifier` refactoring.

a-maumau (approved):
LGTM

PR #13057, opened 2026-07-24
feat(autoware_mission_planner_universe): migrate MissionPlanner to agnocast_wrapper::Node

Koichi98:
## Description
Apply `agnocast_wrapper::Node` to `autoware_mission_planner_universe` (`MissionPlanner`).

**Nodes already running on `MultiThreadedExecutor` are unaffected**, since their callbacks already run concurrently. `MissionPlanner` originally ran on `MultiThreadedExecutor` (`component_container_mt`).

## Dependencies

Depends on https://github.com/autowarefoundation/autoware_core/pull/1289 (`autoware_route_handler`).

## Related Links

**Parent Issue:**

- Link

## How was this PR tested?
Built with `ENABLE_AGNOCAST=0` and `ENABLE_AGNOCAST=1`.

- In `planning_simulator`, set an initial pose and requested a route via `/api/routing/set_route_points`. Under `=0`, `mission_planner` planned and published `/planning/mission_planning/route` (2 lanelet segments) and the route state transitioned to `SET`.
- Under `=1`, `mission_planner` launches as a standalone node with the heaphook preloaded (`LD_PRELOAD`), is listed as `(Agnocast enabled)` by `ros2 node info_agnocast` with the expected Agnocast publishers/subscribers, and reaches the route-ready state.
- Unit tests: under `=1` the utility-function tests pass; the node-based `DefaultPlanner` tests are gated out (they create pub/sub, which needs the Agnocast kernel module and heaphook, unavailable in the test environment).

## Notes for reviewers
`mission_planner` is moved out of the `mission_planner_container` and launched as a standalone node (was a `composable_node` in that container), since it now runs its own agnocast-aware executor.

Two unused declarations that existed in `main` were removed while migrating; both were dead code with no behavioral impact:

- `ArrivalChecker::sub_goal_` (`rclcpp::Subscription<PoseWithUuidStamped>`) — declared but never assigned or referenced.
- `MissionPlanner::on_reroute_availability(const RerouteAvailability::ConstSharedPtr)` — declared but never defined or used; `reroute_availability` is read through the polling subscriber's `take_data()`.

Other structural points:

- `PlannerPlugin::initialize()` now takes `autoware::agnocast_wrapper::Node *` instead of `rclcpp::Node *` (the plugin loader is owned by `MissionPlanner`, whose type is now the wrapper node).
- `ArrivalChecker` switches from `VehicleStopChecker` to `VehicleStopCheckerBase`; odometry is fed manually via `ArrivalChecker::add_odometry()` from `MissionPlanner::on_odometry`, so the twist input goes through the wrapper subscription instead of a separate raw `rclcpp` subscription created inside `VehicleStopChecker`. The velocity buffer duration (`10.0 s`) matches the previous default.

## Interface changes

None.

## Effects on system behavior

None.


soblin (approved):

kobayu858:
Closing this PR since it was addressed and merged in https://github.com/autowarefoundation/autoware_universe/pull/13317 instead.

PR #13071, opened 2026-07-27
refactor(pid_logitudinal_controller): cleanup core logic

interimadd:
## Description

This PR refactors `PidLongitudinalController` to remove per-cycle state that was being smuggled through member variables instead of being passed as parameters/return values, and fixes a clock-type bug uncovered in the process.

- Removed unused members: `m_current_time`, `m_prev_nearest_time`, and the unimplemented member function declarations `getCurrentMotion()` / `calcFilteredAcc()`.
- Removed `m_received_invalid_trajectory`, `m_is_steer_converged`, and `m_emergency_stop_reason`, which were only ever set and read within a single `run()` call. These are now local variables / function parameters / return values instead of persistent state.
- Extracted `createDebugMessage()`, `createSlopeMessage()`, and `createCtrlCmdMsg()` into free helper functions in the anonymous namespace, since none of them need access to member state.
- Moved the `m_prev_ctrl_cmd` update out of `createCtrlCmdMsg` (which should only format the output message) into `calcCtrlCmd`, where the state transition belongs. This allowed `createCtrlCmdMsg` and `getTimeUnderControl` to be marked `const`.
- **Bug fix:** `PidLongitudinalController` compared `current_time` (whose clock type is caller-supplied) against timestamps implicitly converted from `builtin_interfaces::msg::Time`, which defaults to `RCL_ROS_TIME`. When `current_time` used a different clock type, `rclcpp::Time` subtraction threw `"can't subtract times with different time sources"`. Fixed by:
  - Replacing the ROS-message-backed `m_ctrl_cmd_vec` with a dedicated `TimestampedAcceleration` struct that stores `rclcpp::Time` directly, so stored stamps always share `current_time`'s clock type.
  - Converting the remaining message-derived timestamp (trajectory header stamp) using `current_time.get_clock_type()` instead of the implicit `RCL_ROS_TIME` default.
  - Updated `test_pid_longitudinal_controller.cpp` accordingly: `make_time()` no longer needs an `rcl_clock_type_t` parameter, since no test needs to force `RCL_ROS_TIME` anymore.

No behavior change is intended other than the clock-type fix, which only affects callers that pass a non-`RCL_ROS_TIME` clock.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

`colcon build --packages-select autoware_pid_longitudinal_controller && colcon test --packages-select autoware_pid_longitudinal_controller --event-handlers console_cohesion+` passes, including the updated unit tests in `test_pid_longitudinal_controller.cpp`.

Evaluator Link (TIER IV internal)
https://evaluation.tier4.jp/evaluation/reports/bbe87235-3f33-59db-820a-0529a24855ac/?project_id=autoware_dev
https://evaluation.tier4.jp/evaluation/reports/e29489d2-7f0c-5d4c-b34c-22906ee6da0d/?project_id=autoware_dev
https://evaluation.tier4.jp/evaluation/reports/723cfcfb-4899-59bb-9884-64d6f508420d/?project_id=autoware_dev

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/control/autoware_pid_longitudinal_controller/include/autoware/pid_longitudinal_controller/pid_longitudinal_controller.hpp
+++ b/control/autoware_pid_longitudinal_controller/include/autoware/pid_longitudinal_controller/pid_longitudinal_controller.hpp
@@ -252,8 +252,6 @@ private:
   // control state
   ControlState m_control_state{ControlState::STOPPED};
 
-  std::optional<double> m_prev_nearest_time{std::nullopt};
-
   // drive
   PIDController m_pid_vel;
   std::shared_ptr<LowpassFilter1d> m_lpf_vel_error{nullptr};
--- a/control/autoware_pid_longitudinal_controller/include/autoware/pid_longitudinal_controller/pid_longitudinal_controller.hpp
+++ b/control/autoware_pid_longitudinal_controller/include/autoware/pid_longitudinal_controller/pid_longitudinal_controller.hpp
@@ -333,14 +333,6 @@ private:
   Motion calcCtrlCmd(const ControlData & control_data);
 
   /**
-   * @brief create the control command message
-   * @param [in] ctrl_cmd calculated control command to control velocity
-   * @param [in] current_time time captured once per control cycle in run()
-   */
-  autoware_control_msgs::msg::Longitudinal createCtrlCmdMsg(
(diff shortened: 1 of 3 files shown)


Kotakku (approved):
LGTM

PR #13073, opened 2026-07-27
refactor(autoware_pointcloud_preprocessor): extract Filter transform/publish helpers

Koichi98:
## Description

Pure prep, no behavior change. Groundwork for a follow-up that templatizes `autoware::pointcloud_preprocessor::Filter` into `FilterBase<NodeT>` so a filter node can run on `agnocast_wrapper::Node`.

Extracts the three places where that follow-up needs to branch on the node type, so each branch lands inside one function instead of being repeated at every call site:

| extracted | wraps | call sites |
| --- | --- | --- |
| `transform_pointcloud()` | `managed_tf_buffer_->transformPointcloud()` | 3 |
| `lookup_transform_matrix()` | `managed_tf_buffer_->getTransform<Eigen::Matrix4f>()` | 1 |
| `make_sync<Policy>()` | the identical exact-time / approximate-time synchronizer setup in `subscribe()` | 2 |

Also adds explicit `this->` to member calls that need it once the class is templated (`declare_parameter`, `create_subscription`, `get_logger`), and fully-qualifies `OnSetParametersCallbackHandle`.

Split out so this diff stays mechanical and reviewable on its own. Each helper ends up with a single `if constexpr` in the follow-up — see the linked PRs if you want to see what lands in them.

## Related links

- Follow-up: Koichi98/autoware.universe#5 — templatizes `Filter` into `FilterBase<NodeT>`.
- Koichi98/autoware.universe#6 — first node to adopt the instantiation.

## How was this PR tested?

Built `autoware_pointcloud_preprocessor` with `ENABLE_AGNOCAST` unset; compiles cleanly with no warnings. No behavior change, so no runtime testing beyond the existing package tests.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/filter.hpp
```diff
@@ -245,6 +246,17 @@ class Filter : public rclcpp::Node
 
   std::unique_ptr<managed_transform_buffer::ManagedTransformBuffer> managed_tf_buffer_{nullptr};
 
+  /** \brief Transform a pointcloud into target_frame via managed_tf_buffer_. Returns false on
+   * lookup failure. */
+  bool transform_pointcloud(
```
mojomex, line 251: Function can be `const` (and then needs to be `[[nodiscard]]`)

sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/filter.hpp
```diff
@@ -245,6 +246,17 @@ class Filter : public rclcpp::Node
...
+    const std::string & target_frame, const sensor_msgs::msg::PointCloud2 & in,
+    sensor_msgs::msg::PointCloud2 & out);
+
+  /** \brief Look up target_frame <- source_frame as an Eigen matrix at the given stamp, via
+   * managed_tf_buffer_. */
+  std::optional<Eigen::Matrix4f> lookup_transform_matrix(
```
mojomex, line 257: Function can be `const` (and then needs to be `[[nodiscard]]`)

sensing/autoware_pointcloud_preprocessor/src/filter.cpp
```diff
@@ -311,6 +314,22 @@ bool autoware::pointcloud_preprocessor::Filter::calculate_transform_matrix(
...
 }
 
+bool autoware::pointcloud_preprocessor::Filter::transform_pointcloud(
+  const std::string & target_frame, const sensor_msgs::msg::PointCloud2 & in,
+  sensor_msgs::msg::PointCloud2 & out)
+{
+  return managed_tf_buffer_->transformPointcloud(
+    target_frame, in, out, in.header.stamp, rclcpp::Duration::from_seconds(1.0),
+    this->get_logger());
+}
```
mojomex, line 324: (nit) This seems to be purely cosmetic and might not be necessary for this PR

sensing/autoware_pointcloud_preprocessor/src/filter.cpp
```diff
@@ -311,6 +314,22 @@ bool autoware::pointcloud_preprocessor::Filter::calculate_transform_matrix(
...
+
+std::optional<Eigen::Matrix4f> autoware::pointcloud_preprocessor::Filter::lookup_transform_matrix(
+  const std::string & target_frame, const std::string & source_frame, const rclcpp::Time & stamp)
+{
+  return managed_tf_buffer_->getTransform<Eigen::Matrix4f>(
+    target_frame, source_frame, stamp, rclcpp::Duration::from_seconds(1.0), this->get_logger());
+}
```
mojomex, line 331: (nit) Same here. I do agree for both that they're more readable tho.

sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/filter.hpp
mojomex: Ah, it logs, so maybe no.

sensing/autoware_pointcloud_preprocessor/include/autoware/pointcloud_preprocessor/filter.hpp
mojomex: Ah, it logs, so maybe no.

mojomex (approved):

sensing/autoware_pointcloud_preprocessor/src/filter.cpp
Koichi98: True on its own. The follow-up (Koichi98/autoware.universe#5 / #6) branches inside these two, so having them extracted keeps that diff small — I'd rather land it here, but happy to move it if you prefer.

PR #13081, opened 2026-07-28
refactor(autoware_traffic_light_classifier): return-based classify and drop the Core suffix

takam5f2:
## Description

This refactor will make the `classify()` methods immutable. After merging the PR, the methods will return `std::optional<tier4_perception_msgs::msg::TrafficLightArray>`-based data.

- Refactor `perception/autoware_traffic_light_classifier` classifier API so `ClassifierInterface::classify()` returns `std::optional<tier4_perception_msgs::msg::TrafficLightArray>` instead of using caller-owned output arguments and a boolean success flag.
- Update concrete classifier implementations to use the new return-based interface:
  - `CNNClassifier`
  - `CnnLampRecognizer`
- Remove legacy `*Core` classifier naming.
- Preserve existing node-level behavior for ROI filtering, exposure handling, UNKNOWN signal generation, and debug image rendering.

## Related links

- https://github.com/autowarefoundation/autoware_universe/pull/13070 is the preceding PR.

## How was this PR tested?

- Unit tests in `perception/autoware_traffic_light_classifier/test/` were updated/verified.
- Tests cover classifier API refactor and `TrafficLightClassifier` wrapper behavior with ROI ordering, exposure handling, and UNKNOWN signal generation.

```bash
colcon test --packages-select autoware_traffic_light_classifier
colcon test --packages-select autoware_traffic_light_classifier --event-handlers console_direct+
```

## Notes for reviewers

The classifier node's externally visible behavior is unchanged (input/output topics, parameters, and published signals). Two things worth calling out:

- The `CnnLampRecognizer` constructor now takes `traffic_light_type`, and `update_traffic_signals()` takes it as a parameter instead of reading it off the signal. `classify()` therefore leaves `traffic_light_id` / `type` unset like the other backends; the `TrafficLightClassifier` wrapper associates them afterwards.
- The wrapper now enforces the one-signal-per-image contract in a single place (the id/type association loop), replacing the size guard that used to be duplicated in each backend.
- `single_image_debug_inference_node` is fixed by this change: on `main` it constructed an empty signals array and always tripped the old size guard, logging `failed to classify image` on every click. This is a user-visible fix, and the exact accident the caller-owned-output design invited — which motivates the return-based API.

## Interface changes

- `ClassifierInterface::classify(const std::vector<cv::Mat> & images)` now returns `std::optional<tier4_perception_msgs::msg::TrafficLightArray>`.
- `CNNClassifier` now implements the return-based classifier API.
- `CnnLampRecognizer` now:
  - accepts `CnnLampRecognizerConfig::traffic_light_type` in its constructor and stores it,
  - passes that configured type to `update_traffic_signals()`, which now takes it as a parameter (instead of reading it off the signal) so pedestrian lamps are forced to `CIRCLE`,
  - leaves `traffic_light_id` / `type` unset on the returned signals, matching the other backends.
- The old `*Core` class name suffix has been removed from classifier implementations.

## Effects on system behavior

None for the classifier node (topics, parameters, and published signals unchanged). The only user-visible change is a bug fix: the `single_image_debug_inference` tool, which was broken on `main` (always logged `failed to classify image`), now works.


takam5f2:
@a-maumau  
Could you check the PR?  There is no behavior change. However, the contract of `CnnLampRecognizer` is changed by the PR. The contract change is based on the assumption that each classifier is provided per `traffic_light_type`.

> - CnnLampRecognizer now:
>  - accepts CnnLampRecognizerConfig::traffic_light_type in its constructor,
>  - stores that configured type internally,
>  - stamps traffic_light_type on each output TrafficLight signal in classify(),

a-maumau (approved):
LGTM

PR #12857, opened 2026-06-24
feat(label_based_euclidean_cluster):  apply `agnocast_wrapper::Node` to `label_based_euclidean_cluster`

Koichi98:
## Description
Apply `agnocast_wrapper::Node` to `autoware_label_based_euclidean_cluster`.
Please read https://github.com/autowarefoundation/autoware/issues/7007 for the context and review guidelines of `autoware_agnocast_wrapper` adoption. 
Also, please refer to https://github.com/autowarefoundation/autoware_core/blob/main/common/autoware_agnocast_wrapper/docs/review_guide.md for the review guide of `autoware_agnocast_wrapper` application PR. FYI: This node take [Method2](https://github.com/autowarefoundation/autoware_core/blob/main/common/autoware_agnocast_wrapper/docs/review_guide.md#method-2-agnocast_wrappernode-inheritance) of two integrations methods.

## Related Links

  - https://github.com/autowarefoundation/autoware/issues/7007
  - https://github.com/orgs/autowarefoundation/discussions/6804

## Related links

**Parent Issue:**

- Link

## How was this PR tested?
  - `colcon build` succeeds with both `ENABLE_AGNOCAST=1` and `ENABLE_AGNOCAST=0`.
  - Ran Autoware based product with this change applied and confirmed it operates without any regression.

## Interface changes

None.

## Effects on system behavior

None when ENABLE_AGNOCAST is not set or set to 0 on runtime. When ENABLE_AGNOCAST=1, CPU Time used in this node for message serialization/deserialization will be reduced.

PR #13082, opened 2026-07-28
feat(euclidean_cluster): apply agnocast_wrapper::Node to label_based_euclidean_cluster

kobayu858:
## Description

Apply `agnocast_wrapper::Node` to `label_based_euclidean_cluster` in `autoware_euclidean_cluster`.

The CMake/`package.xml` wiring for the Agnocast wrapper is already on `main`, so this PR only converts the node itself and adjusts how it is launched.

Changes:

- The publishers are created with `this->create_publisher<T>()` instead of `AUTOWARE_CREATE_PUBLISHER2`, and the two outputs are published with `publish(const &)` instead of `publish(std::move(...))`. See "Notes for reviewers".
- The per-label parameters (`label_cluster_params.*`, `confusable_label_groups.*`) are read from the node's declared parameters via `list_parameters()` instead of scanning `rclcpp::NodeOptions::parameter_overrides()`. See "Notes for reviewers".
- The node is launched as a standalone node instead of a composable node: `label_based_euclidean_cluster.launch.py` now declares a single `launch_ros.actions.Node` and keeps including `agnocast_env.launch.py`, and the container-related arguments are dropped from `label_based_euclidean_cluster.launch.xml`. The topic argument names are unchanged, so callers need no modification.
- The node-level test target is excluded at CMake configuration time when `ENABLE_AGNOCAST=1`; the pure-logic tests are still built and run in both configurations.

## Related links

- Supersedes #12857

## How was this PR tested?

- **Build**: Confirmed the package builds successfully with both `ENABLE_AGNOCAST=0` and `ENABLE_AGNOCAST=1` (clean build for each configuration). Also confirmed that the package and its downstream packages (`autoware_detection_by_tracker`, `autoware_image_projection_based_fusion`, `tier4_perception_launch`, `tier4_simulator_launch`, `autoware_launch`) build with `ENABLE_AGNOCAST=1`.
- **Unit tests**: With `ENABLE_AGNOCAST=0`, the node-level tests pass (8/8) along with the pure-logic tests and the lint tests. With `ENABLE_AGNOCAST=1`, the node-level test target is not generated as intended, and the pure-logic tests pass.
- **Standalone launch**: Confirmed the node starts under `ENABLE_AGNOCAST=1` as a standalone node and resolves the parameters from `--params-file` correctly with the `AgnocastOnly` executor.
- **Runtime (upstream)**: This node is only launched when `use_semantic_segmentation_ptv3` is enabled, so it is not exercised by the default logging simulator pipeline. Instead it was driven directly under `ENABLE_AGNOCAST=1` by injecting a synthetic segmented point cloud (`x`/`y`/`z` plus `class_id` and `probability`) containing four labelled clusters (`CAR`, `CAR`, `PEDESTRIAN`, `VEGETATION`). Both outputs behaved as expected and were published continuously: `output/objects` carried three objects with labels `PEDESTRIAN`, `CAR`, `CAR`, and `output/pointcloud` carried the `VEGETATION` points as segments. `ros2 topic list_agnocast` shows the input and both outputs registered as Agnocast enabled.
- **Regression (LSim)**: Confirmed the logging simulator with the sample rosbag starts and runs under `ENABLE_AGNOCAST=1` with no additional node failures compared to the same run without this PR.

## Notes for reviewers

- **Why nested parameters are read with `list_parameters()`**: As a standalone node, the contents of `--params-file` do not appear in `rclcpp::NodeOptions::parameter_overrides()`; the previous design relied on the container launch script loading the YAML in Python and passing `parameters=[dict]`. Reading the declared parameters instead keeps the same YAML layout working in both configurations.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_euclidean_cluster/launch/label_based_euclidean_cluster.launch.xml
+++ b/perception/autoware_euclidean_cluster/launch/label_based_euclidean_cluster.launch.xml
@@ -5,15 +5,10 @@
 
   <arg name="param_path" default="$(find-pkg-share autoware_euclidean_cluster)/config/label_based_euclidean_cluster.param.yaml"/>
 
-  <arg name="use_pointcloud_container" default="false"/>
-  <arg name="pointcloud_container_name" default="pointcloud_container"/>
-
   <include file="$(find-pkg-share autoware_euclidean_cluster)/launch/label_based_euclidean_cluster.launch.py">
     <arg name="input/pointcloud" value="$(var input/pointcloud)"/>
     <arg name="output/objects" value="$(var output/objects)"/>
     <arg name="output/pointcloud" value="$(var output/pointcloud)"/>
     <arg name="param_path" value="$(var param_path)"/>
-    <arg name="use_pointcloud_container" value="$(var use_pointcloud_container)"/>
-    <arg name="pointcloud_container_name" value="$(var pointcloud_container_name)"/>
   </include>
 </launch>
--- a/perception/autoware_euclidean_cluster/CMakeLists.txt
+++ b/perception/autoware_euclidean_cluster/CMakeLists.txt
@@ -86,7 +86,7 @@ autoware_agnocast_wrapper_setup(${PROJECT_NAME}_label_based_node_core)
 autoware_agnocast_wrapper_register_node(${PROJECT_NAME}_label_based_node_core
   PLUGIN "autoware::euclidean_cluster::LabelBasedEuclideanClusterNode"
   EXECUTABLE label_based_euclidean_cluster_node
-  AGNOCAST_EXECUTOR CallbackIsolatedAgnocastExecutor
+  AGNOCAST_EXECUTOR AgnocastOnlyCallbackIsolatedExecutor
   ROS2_EXECUTOR SingleThreadedExecutor
 )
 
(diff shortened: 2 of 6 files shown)


perception/autoware_euclidean_cluster/launch/label_based_euclidean_cluster.launch.xml
```diff
@@ -5,15 +5,10 @@
...
-  <arg name="use_pointcloud_container" default="false"/>
-  <arg name="pointcloud_container_name" default="pointcloud_container"/>
-
   <include file="$(find-pkg-share autoware_euclidean_cluster)/launch/label_based_euclidean_cluster.launch.py">
     <arg name="input/pointcloud" value="$(var input/pointcloud)"/>
     <arg name="output/objects" value="$(var output/objects)"/>
     <arg name="output/pointcloud" value="$(var output/pointcloud)"/>
     <arg name="param_path" value="$(var param_path)"/>
-    <arg name="use_pointcloud_container" value="$(var use_pointcloud_container)"/>
-    <arg name="pointcloud_container_name" value="$(var pointcloud_container_name)"/>
```
ogrex, line 17: [Confirm]
this change would work with  ENABLE_AGNOCAST=0  as well?

kobayu858: Yes, it works with ENABLE_AGNOCAST=0 as well. `agnocast_env.launch.py` only changes the
`LD_PRELOAD` value; when ENABLE_AGNOCAST is not `1` it falls back to the existing value
(normally empty), so the node runs as a plain ROS 2 node.

Verified with:
- Build with ENABLE_AGNOCAST=0 and =1
- Package tests: 8/8 passed
- Logging simulator run with the sample rosbag: no regression in either configuration

ogrex (approved):
LGTM

PR #13087, opened 2026-07-29
refactor(autoware_lidar_centerpoint)!: consume per-variant model folders via a model manifest

technolojin:
## Description

The `AutowareFoundation/lidar_centerpoint` model bundle is being restructured from a flat directory into one self-contained folder per variant (`base` / `tiny` / `sigma` / `short_range`) with uniform file names:

```text
lidar_centerpoint/
├── base/         # model_name: centerpoint
├── tiny/         # model_name: centerpoint_tiny
├── sigma/        # model_name: centerpoint_sigma
└── short_range/  # model_name: centerpoint_short_range
```

Each folder holds `pts_voxel_encoder.onnx`, `pts_backbone_neck_head.onnx`, `ml_package.param.yaml`, `detection_class_remapper.param.yaml` and `deploy_metadata.yaml`. This PR makes `autoware_lidar_centerpoint` consume that layout.

**`ml_package.param.yaml` becomes a fixed-name model manifest shipped inside each variant folder.** It carries the artifact names (encoder/head ONNX + engine, `class_remapper_param_path`, `trt_precision`) as paths relative to `model_path`, plus `model_params`. Adding a variant no longer means adding files to this repo.

- Node-implementation parameters (`cloud_capacity`, `post_process_params`, `densification_params`) move to `config/centerpoint_common.param.yaml`. The per-variant `config/centerpoint_<variant>.param.yaml` layer is deleted — its only remaining content was those keys plus weight paths that the manifest now owns.
- `node.cpp` declares a `model_path` parameter and resolves relative manifest entries against it; absolute paths pass through unchanged, so a custom model can still be pointed at arbitrary files.
- The class remapper YAML is loaded node-side via `rclcpp::parameter_map_from_yaml_file` from the manifest's `class_remapper_param_path`, so the matrices are no longer declared as node parameters or passed as a launch-level `<param from=…>`.
- `launch/lidar_centerpoint.launch.xml` now takes only `model_path` (+ `common_param_path`) and loads `$(var model_path)/ml_package.param.yaml`. `model_name` is gone — the variant is the folder.
- Schemas: `centerpoint_ml_package.schema.json` → `ml_package.schema.json` (paired by basename with `config/ml_package.param.yaml`); the moved keys are added to `centerpoint_common.schema.json`; `centerpoint.schema.dummy.json` and the duplicate `centerpoint_tiny_ml_package.schema.json` are deleted.

Variant selection now lives in the perception launcher — see the companion `autoware_launch` PR.

## Related links

**Parent Issue:**

- https://github.com/autowarefoundation/autoware_universe/issues/13086

Model bundle PR (draft, awaiting `v4.0` tag): <https://huggingface.co/AutowareFoundation/lidar_centerpoint/discussions/1>

**This PR must land together with:**

- autowarefoundation/autoware_launch#1930 — launcher and param-mirror side of the same restructure
- autowarefoundation/autoware#7236 — ansible artifacts revision `v3.0` → `v4.0`

The `v3.0` bundle layout does not work with this code, and the `v4.0` layout does not work with the current code. Deployments pinned to `v3.0` data plus pre-merge code are unaffected.

## How was this PR tested?

- `pre-commit run --all-files`
- `colcon build --packages-up-to autoware_lidar_centerpoint`
- Launch smoke test against a local staging copy of the `v4.0` bundle: `base` and `tiny` ran the full path (engine build → inference → published objects); `sigma` and `short_range` were verified up to parameter load and manifest resolution.

Not yet done: engine-build smoke test for `sigma` and `short_range`. The test machine's GPU driver got wedged partway through and the run has not been repeated.

## Notes for reviewers

`model_path` is declared with an empty-string default. A manifest that uses absolute paths therefore works with no `model_path` set at all, which keeps custom/out-of-tree models working.

#### Additions and removals

| Change type | Parameter Name                    | Type            | Default Value | Description                                                          |
| :---------- | :-------------------------------- | :-------------- | :------------ | :------------------------------------------------------------------- |
| Added       | `model_path`                      | `string`        | `""`          | Directory containing the model artifacts; relative entries resolve against it |
| Added       | `class_remapper_param_path`       | `string`        | -             | Class remapper parameter file, relative to `model_path`              |
| Removed     | `allow_remapping_by_area_matrix`  | `int64_array`   | -             | Now read from the file at `class_remapper_param_path`                |
| Removed     | `min_area_matrix`                 | `double_array`  | -             | Now read from the file at `class_remapper_param_path`                |
| Removed     | `max_area_matrix`                 | `double_array`  | -             | Now read from the file at `class_remapper_param_path`                |

#### Modifications

| Version | Parameter Name                                                                              | Type     | Default Value                                    | Description                          |
| :------ | :------------------------------------------------------------------------------------------ | :------- | :----------------------------------------------- | :----------------------------------- |
| Old     | `encoder_onnx_path` / `encoder_engine_path` / `head_onnx_path` / `head_engine_path`          | `string` | `$(var model_path)/pts_voxel_encoder_$(var model_name).onnx` etc. | Full path, composed in the param file |
| New     | `encoder_onnx_path` / `encoder_engine_path` / `head_onnx_path` / `head_engine_path`          | `string` | `pts_voxel_encoder.onnx` etc.                    | File name relative to `model_path`; absolute paths still accepted |

### Launch argument changes

`launch/lidar_centerpoint.launch.xml`:

| Change type | Argument                    | Description                                          |
| :---------- | :-------------------------- | :--------------------------------------------------- |
| Removed     | `model_name`                | Variant is the `model_path` folder                   |
| Removed     | `model_param_path`          | Per-variant param layer deleted                      |
| Removed     | `ml_package_param_path`     | Fixed to `$(var model_path)/ml_package.param.yaml`   |
| Removed     | `class_remapper_param_path` | Comes from the manifest, resolved node-side          |
| Modified    | `model_path`                | Now the variant folder, default `$(var data_path)/lidar_centerpoint/tiny` |

## Effects on system behavior

No behavioral change for the `base` and `tiny` variants: the same weights and the same parameter values are loaded, only the file layout and the loading path differ.

`sigma` and `short_range` become usable out of the box. Their weights ship in the bundle for the first time — previously `short_range_centerpoint_model_path` pointed at a directory nothing populated, and `sigma` had no published ONNX at all.


perception/autoware_lidar_centerpoint/config/centerpoint_common.param.yaml
```diff
@@ -1,5 +1,14 @@
 /**:
   ros__parameters:
+    cloud_capacity: 2000000
+    post_process_params:
+      circle_nms_dist_threshold: 0.5
+      iou_nms_search_distance_2d: 10.0
+      iou_nms_threshold: 0.1
+      yaw_norm_thresholds: [0.3, 0.3, 0.3, 0.3, 0.0]
```
YoshiRi, line 8: [comment]
This need to be align with model output size. At least we need some assertion to check class size.

perception/autoware_lidar_centerpoint/src/node.cpp
```diff
@@ -64,10 +75,23 @@ LidarCenterPointNode::LidarCenterPointNode(const rclcpp::NodeOptions & node_opti
...
+    for (const auto & param : entry.second) {
+      if (param.get_name() == "allow_remapping_by_area_matrix") {
+        allow_remapping_by_area_matrix = param.as_integer_array();
+      } else if (param.get_name() == "min_area_matrix") {
+        min_area_matrix = param.as_double_array();
+      } else if (param.get_name() == "max_area_matrix") {
+        max_area_matrix = param.as_double_array();
+      }
+    }
+  }
```
YoshiRi: [comment]
It sounds safer to validate this parameter load.

```c++
if (
  allow_remapping_by_area_matrix.empty() ||
  min_area_matrix.empty() ||
  max_area_matrix.empty()) {
  throw std::invalid_argument(
    "Class remapper parameter file does not contain all required matrices: " +
    class_remapper_param_path);
}
```

YoshiRi (commented):
Overall, the idea looks good.

perception/autoware_lidar_centerpoint/src/node.cpp
technolojin: added https://github.com/autowarefoundation/autoware_universe/pull/13087/changes/a8eb4177eb8e0ab70aff4ac3d543e8b3cee48f5b

perception/autoware_lidar_centerpoint/config/centerpoint_common.param.yaml
YoshiRi: @technolojin 
After [TIER IV INTERNAL DISCUSSION](https://star4.slack.com/archives/C03S84LDJGG/p1785898441220849?thread_ts=1785897531.584829&cid=C03S84LDJGG), would you move these parameter into model?

technolojin: let me move this when https://huggingface.co/AutowareFoundation/lidar_centerpoint is update to v4.1

xmfcx:
When the class remapper file is missing, [`node.cpp:84`](https://github.com/autowarefoundation/autoware_universe/blob/f3b22fd86dad89a0ab64ebde9634c091938fcdaf/perception/autoware_lidar_centerpoint/src/node.cpp#L84) aborts with no file name in the message:

```text
terminate called after throwing an instance of 'rclcpp::exceptions::RCLError'
  what():  Error opening YAML file, at ./src/parser.c:271
```

rclcpp throws from `throw_from_rcl_error` and does not put the path in the message, so the node must add the path. Three lines before [line 84](https://github.com/autowarefoundation/autoware_universe/blob/f3b22fd86dad89a0ab64ebde9634c091938fcdaf/perception/autoware_lidar_centerpoint/src/node.cpp#L84) are enough:

```cpp
if (!std::filesystem::exists(class_remapper_param_path)) {
  throw std::invalid_argument(
    "Class remapper file not found: " + class_remapper_param_path);
}
```

The file already includes `<filesystem>` at [line 29](https://github.com/autowarefoundation/autoware_universe/blob/f3b22fd86dad89a0ab64ebde9634c091938fcdaf/perception/autoware_lidar_centerpoint/src/node.cpp#L29). A `try` and `catch` block around line 84 also covers a malformed file.

The custom model case gives this error: the manifest loads, and the file that it names in `class_remapper_param_path` is missing. A stale `v3.0` bundle fails earlier and more clearly, because the launch file reads `$(var model_path)/ml_package.param.yaml` and names the missing path.

✅ Your empty-matrix guard already handles the wrong-file case well. I set `class_remapper_param_path` to `ml_package.param.yaml` by mistake, and got `Class remapper matrices must be non-empty: <path>`, which is exactly right.

xmfcx:
- https://github.com/autowarefoundation/autoware_core/pull/1354 should fix the broken CI.

perception/autoware_lidar_centerpoint/README.md
```diff
@@ -32,44 +32,57 @@ We trained the models using <https://github.com/open-mmlab/mmdetection3d>.
...
+| `post_process_params.yaw_norm_thresholds`        | list[double] | [0.3, 0.3, 0.3, 0.3, 0.0] | An array of distance threshold values of norm of yaw [rad].                              |
+| `post_process_params.iou_nms_search_distance_2d` | double       | -                         | If two objects are farther than the value, NMS isn't applied.                            |
+| `post_process_params.iou_nms_threshold`          | double       | -                         | IoU threshold for the IoU-based Non Maximum Suppression                                  |
+| `post_process_params.has_twist`                  | boolean      | false                     | Indicates whether the model outputs twist value.                                         |
+| `densification_params.world_frame_id`            | string       | `map`                     | the world frame id to fuse multi-frame pointcloud                                        |
+| `densification_params.num_past_frames`           | int          | `1`                       | the number of past frames to fuse with the current frame                                 |
```
ktro2828, line 54: Can we replace them with `centerpoint_common.schema.json`?


```suggestion
{{ json_to_markdown("perception/autoware_lidar_centerpoint/schema/ceterpoint_common.schema.json") }}
```

ktro2828 (approved):
Overall looks good to me. I left one minor comment.

perception/autoware_lidar_centerpoint/launch/lidar_centerpoint.launch.xml
```diff
@@ -4,11 +4,7 @@
   <arg name="output/objects" default="objects"/>
   <arg name="data_path" default="$(env HOME)/autoware_data/ml_models" description="packages data and artifacts directory path"/>
   <arg name="node_name" default="lidar_centerpoint" description="node name for autoware_lidar_centerpoint_node"/>
-  <arg name="model_name" default="centerpoint_tiny" description="options: `centerpoint`, `centerpoint_tiny` or `centerpoint_short_range`"/>
-  <arg name="model_path" default="$(var data_path)/$(var node_name)"/>
-  <arg name="model_param_path" default="$(find-pkg-share autoware_lidar_centerpoint)/config/$(var model_name).param.yaml"/>
-  <arg name="ml_package_param_path" default="$(var model_path)/$(var model_name)_ml_package.param.yaml"/>
-  <arg name="class_remapper_param_path" default="$(var model_path)/detection_class_remapper.param.yaml"/>
+  <arg name="model_path" default="$(var data_path)/lidar_centerpoint/tiny" description="directory containing the model artifacts"/>
```
ktro2828, line 7: Just curious, do we have special reasons to keep `tiny` as default? I feel it's OK to use `base` as a default model.

KSeangTan (approved):
Thanks for the work, LGTM!

xmfcx:
- https://github.com/autowarefoundation/autoware_universe/pull/13087#issuecomment-5228446408

Could you add a useful error message if model path is wrong as I stated here? Sorry that it wasn't clear 🙇‍♂️

technolojin:
@xmfcx Thank you for the detailed report. Addressed in 791a19eddc5:

- Added the `std::filesystem::exists` guard before the YAML load, so a missing file now fails with `Class remapper file not found: <path>`.
- Also wrapped `rclcpp::parameter_map_from_yaml_file` in a try/catch that rethrows with the path, so a malformed file reports `Failed to parse class remapper file: <path> (<original error>)`.

Verified both by running the node:

```
what():  Class remapper file not found: /nonexistent/remapper.yaml
what():  Failed to parse class remapper file: .../malformed.yaml (Sequences can only be values and not keys in params. Error at line 1, at ./src/parse.c:822)
```

perception/autoware_lidar_centerpoint/launch/lidar_centerpoint.launch.xml
technolojin: might a historical background. a big user may wanted to set the tiny model by default.

perception/autoware_lidar_centerpoint/README.md
technolojin: I will fix it in the next time

PR #13088, opened 2026-07-29
fix(autoware_multi_object_tracker): protect exported trackers from overlap-merge deletion

technolojin:
## Description

The tracker overlap merger could delete a tracker that is currently in the output and keep one that the publish gate withholds. The object then disappears from `tracking/objects` for several frames and returns under a new UUID.

**Scenario — detector class handoff.** An object first seen by a polygon channel (clustering, camera-lidar fusion) gets a `polygon_tracker`, matures, and is published as UNKNOWN. When the DNN detector starts reporting it as `bounding_box / car`, that detection cannot associate to a polygon tracker, so a second, brand-new vehicle tracker spawns on the same object. On its second measurement the merger pairs the two: `compareByPriority` lets GENERAL_VEHICLE beat POLYGON on type alone, so the young tracker wins regardless of maturity. It passes the merge gate (`isConfident` at its own state time, zero extrapolation) but fails the publish gate (state time + 200 ms, where the seeded `initial_covariance.vel_long` still dominates the predicted position covariance). `applyMerges` then erases the mature tracker permanently, and nothing is published until the winner's velocity covariance collapses — right as the object becomes more relevant. The UUID change also resets downstream prediction history and ACC / obstacle-stop targets.

### 1. An exported tracker is never deleted for an unexportable one

`decidePair` vetoes the merge when the winner fails the publish-horizon check and the loser passes it:

```cpp
if (!ensurePublishConfident(winner, ctx) && ensurePublishConfident(loser, ctx)) {
  return PairDecision::NoMerge;
}
```

`ensurePublishConfident` is the same predicate on the same state as the publish gate in `getTrackedObjects`, so "the loser passes" is exactly "the loser is in the current output". The veto is stateless: the pair is re-discovered each prune cycle and merges once the winner matures. Conditioning on the loser keeps it correct in both directions — when neither tracker is exportable the pair still consolidates. While deferring, exactly one of the pair is exported, and `prune` runs before publishing in both paths, so no frame contains both.

Not implemented as a ranking tier: promoting publish-confidence above `compareByFreshFullMeasurement` would let a stale vehicle tracker absorb a fresh, fully-measured one, inverting an existing invariant. A pair-level veto defers instead.

### 2. Survival ranking compares aggregate channel support

`dominatesOnAnyChannel` treated a 0.4 margin on any *single* channel as dominance and scored a missing channel as 0.001, so any channel above 0.401 that the other tracker lacked counted as dominance. `channelSupport` replaces it with the sum of per-channel existence probabilities (cached in the snapshot), compared with the same 0.4 buffer: one scalar, monotone and symmetric, no sentinel. The tier sits below `compareByPriority`, so it applies only to same-priority pairs.

### 3. `prune` interval constant

The guard `< 2000 /*2ms*/` was actually 2 µs. Replaced with a named `min_prune_interval_ns` of 2 ms. No practical change in merge frequency (the publish-time call arrives with Δt = 0 and is suppressed; sensor batches are ≥ ~10 ms apart) — this fixes a misleading comment and a latent trap.

## Related links

None.

## How was this PR tested?

- before (with detection-by-tracker, rule-based-clustering)

https://github.com/user-attachments/assets/305a7620-fe60-4b58-b44c-777f3d2edb9f

- before (without detection-by-tracker, rule-based-clustering)

https://github.com/user-attachments/assets/5f195544-22af-4981-b51c-01d61fcec6cf

- after (with detection-by-tracker, rule-based-clustering)

https://github.com/user-attachments/assets/2679096c-bee2-4445-816e-6db34e78cc28

- `colcon build` / `colcon test` for the package — clean; the gtest binary runs 51 tests across 11 suites, all passing.
- Six tests added to `test_tracker_overlap_manager.cpp` (19 in the suite): the publish veto defers, applies once the winner matures, and still consolidates when neither tracker is exportable; plus symmetry and aggregate-support characterisation of the new ranking tier. Re-run against the previous logic, two of the new tests fail (not vacuous); the other 17 pass.
- Rosbag replay validation of the fix is still outstanding.

## Notes for reviewers

- The publish-gate tests pin the confidence regime with premise asserts (`isConfident` at merge time versus `std::nullopt`), so they fail loudly at setup if motion-model constants drift.
- The veto reads only the confidence predicate, not `ensureObject`; a publish-confident loser that fails export would defer harmlessly — `decidePair` returns `NoMerge` for that pair anyway.

## Interface changes

None.

## Effects on system behavior

- A tracker currently in the output is never deleted by an overlap merge in favour of one the publish gate would withhold; tracking output stays continuous across detector class handoffs (polygon → vehicle bbox), with no UUID reassignment.
- Two trackers that are both withheld from the output now consolidate instead of both surviving.
- The survival ranking's channel tier is decided by total support, so it defers to later tiers more often. Same-priority pairs only.


YoshiRi (commented):
1. This PR does not keep UUID but just remove blank output between `polygon UUID → vehicle UUID`. To keep UUID consistency, I guess we need to add abstract layer to handle output UUID but I think we should update that in another PR.
2. Now ranking system become very complicated. It would be nicer to refactor this system later.

ogrex (approved):
LGTM

technolojin:
@mitsudome-r 
by your commit https://github.com/autowarefoundation/autoware_universe/pull/13129 the spell-check dose not pass.
Can you please add `overlayfs` exception?

PR #13091, opened 2026-07-29
feat(autoware_obstacle_grid_utils): add obstacle-grid consumer helper

youtalk:
## Description

Adds `autoware_obstacle_grid_utils`, a small header-only library that the last-resort safety consumers share to query the sensing-side obstacle grid (`grid_map_msgs/GridMap`, layers `max_height`/`min_height`/`point_count`). Factoring the query logic once prevents each consumer from re-deriving it inconsistently.

It provides: a per-cell `Gate {min_point_count_cell, min_height}` (a cell qualifies iff `point_count >= min_point_count_cell && max_height >= min_height`, `NaN`-safe); `cell_qualifies()`; `cell_footprint()` and `cell_corners()` (the cell's footprint box and its four literal corners, so distance and membership queries are edge-aware rather than center-based, because cells are areas); `nearest_distance(grid, ego_polygon, gate)` (2D distance from an ego polygon to the nearest qualifying cell footprint, `+inf` when nothing qualifies inside the ROI); `nearest_cell(grid, ego_polygon, gate)` (the same edge-aware distance plus where the nearest qualifying cell is, as `std::optional<NearestCell{distance, position, index}>`, `std::nullopt` when nothing qualifies; on a distance tie the first cell reached by the grid iterator wins); `cells_in_polygon(grid, polygon, gate)` (centers of qualifying cells inside a lanelet/trajectory polygon); and `connected_components(grid, qualifying)` (8-connected labeling of the caller-supplied qualifying cell indices, returning `std::vector<CellComponent{cells, point_sum}>` in first-seen order, where `point_sum` is the sum of `point_count` over each component — it does not gate, the caller applies its own `Gate` first).

This is the first piece of the sensing-side safety path; no behavior changes in any existing node.

## Related links

**Parent Issue:**

- https://github.com/autowarefoundation/autoware_universe/issues/12972

## How was this PR tested?

`colcon test --packages-select autoware_obstacle_grid_utils` — 21 gtest cases pass with 0 failures in `autoware:universe-devel-jazzy`. The tests pin the edge-aware distance against a hand-computed independent oracle (distance to the near footprint corner, strictly less than the cell-center distance); the count and height gates (pass/reject), including rejection of a `NaN` `max_height` even when the count passes; the empty-ROI `+inf` return; `cells_in_polygon` collection on a known cell; `cell_corners` being the literal footprint corners; and `nearest_cell` agreeing with `nearest_distance`, returning `std::nullopt` when nothing qualifies, and picking the closer of two occupied cells. Eleven further cases cover `connected_components`: empty input, single-cell `point_sum`, straight-line and L-shaped components, a diagonal pair joining under 8-connectivity, a gap separating two components, border cells probing no out-of-range neighbour, a row wrap not falsely connecting, a duplicate index counted once, a `NaN` `point_count` skipped in the sum, and a merge across the circular-buffer wrap seam.

## Notes for reviewers

Header-only (`INTERFACE`-style, installed via `ament_auto_package`). Depends only on `grid_map_core` and `autoware_utils_geometry`. The grid contract (topic, layers, QoS) is produced by `autoware_obstacle_grid_extractor` (separate PR); this package only consumes a decoded `grid_map::GridMap`.

`connected_components` is circular-buffer aware: `grid_map` stores cells in a wrapping buffer, so neighbour probing resolves through the wrapped index rather than raw row/column arithmetic — hence the explicit wrap-seam and row-wrap tests.

This branch is cut on upstream `main` at `789bded9` and is purely additive: it adds one new package and depends on nothing that is not already upstream.

## Interface changes

New package; no topic/service/parameter changes to existing nodes.

## Effects on system behavior

None (new, unused-by-default helper library).

### DIFF
--- a/common/autoware_obstacle_grid_utils/CMakeLists.txt
+++ b/common/autoware_obstacle_grid_utils/CMakeLists.txt
@@ -4,6 +4,10 @@ project(autoware_obstacle_grid_utils)
 find_package(autoware_cmake REQUIRED)
 autoware_package()
 
+ament_auto_add_library(${PROJECT_NAME} SHARED
+  src/obstacle_grid_utils.cpp
+)
+
 if(BUILD_TESTING)
   ament_auto_add_gtest(test_${PROJECT_NAME}
     test/test_obstacle_grid_utils.cpp
--- a/common/autoware_obstacle_grid_utils/CMakeLists.txt
+++ b/common/autoware_obstacle_grid_utils/CMakeLists.txt
@@ -1,0 +1,13 @@
+cmake_minimum_required(VERSION 3.14)
+project(autoware_obstacle_grid_utils)
+
+find_package(autoware_cmake REQUIRED)
+autoware_package()
+
+if(BUILD_TESTING)
+  ament_auto_add_gtest(test_${PROJECT_NAME}
+    test/test_obstacle_grid_utils.cpp
+  )
+endif()
+
+ament_auto_package()
--- a/common/autoware_obstacle_grid_utils/README.md
+++ b/common/autoware_obstacle_grid_utils/README.md
@@ -1,6 +1,6 @@
 # autoware_obstacle_grid_utils
 
-Header-only consumer helpers for the obstacle grid published on
+Consumer helpers for the obstacle grid published on
 `/sensing/obstacle_segmentation/obstacle_grid` (`grid_map_msgs/msg/GridMap`, frame `base_link`,
 layers `max_height` / `min_height` / `point_count`, `NaN` = empty cell).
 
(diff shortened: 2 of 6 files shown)


interimadd (commented):
Thanks for the PR and for assigning me as a reviewer!
I have a few suggestions to improve maintainability. I'd appreciate it if you could take a look at the following points:

### Should the header be split into .hpp/.cpp?

- The header exposes internal-only helpers like cell_footprint() that no consumer actually calls externally. Making them public risks unintended external usage that then blocks future changes.
- Splitting also aligns the package with the structure of the other *_utils packages, which is easier to follow.
- With everything inlined in the header, any small change forces every dependent package to recompile.

### Can test readability be improved further?

- Structure each test with clear Arrange/Act/Assert sections.
- Avoid over-abbreviated variable names to reduce cognitive load. (e.g. distance_to_nearest instead of d)
- Keep each test case to a single assertion/behavior, and avoid going back to Act after an Assert.

youtalk:
Thank you for your review as always! Both applied on https://github.com/autowarefoundation/autoware_universe/pull/13091/commits/2fce4ef77168f266158b840191b4b7cca64c7d6e.

- Split. Implementations moved to `src/obstacle_grid_utils.cpp` and `cell_footprint()` is now internal to it — you were right that nothing outside calls it; I checked all ten pending consumer migrations. The same check turned up one more dead helper, `cells_in_polygon()`, which no consumer calls either, so I removed it rather than ship unused API. On the recompile point.
- Tests. Every case is now Arrange/Act/Assert covering one behavior, with no assert-act-assert. Names are spelled out, and the five-parameter `make_grid()` is replaced by `make_empty_grid()` + `mark_cell(grid, index, CellContent{...})` — its min_height parameter was dead weight, as no helper reads that layer.

interimadd (approved):
Thanks for addressing the suggestions!
Now the tests are easy to understand and exported APIs are kept to a minimum. LGTM.

Issue #12972, opened 2026-07-09
Consolidate duplicated pointcloud preprocessing in last-resort safety consumers onto a shared perception-independent obstacle grid


youtalk:
### Checklist

- [x] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [x] I've searched other issues and no duplicate issues were found.
- [x] I've agreed with the maintainers that I can plan this task.

### Description

About eight planning/control nodes subscribe to the no-ground obstacle-segmentation point cloud (`/perception/obstacle_segmentation/pointcloud`, `sensor_msgs/PointCloud2`) and each re-implements its own preprocessing before answering essentially the same question — "is there obstacle evidence within a region around ego or the trajectory, and how close?". At [`085b97a5`](https://github.com/autowarefoundation/autoware_universe/commit/085b97a56aa25e21f9ec8f6accc05f8860c42c85) (a recent `main` snapshot used as the reference below):

| consumer | raw-cloud preprocessing | key parameters | source |
| --- | --- | --- | --- |
| `autoware_autonomous_emergency_braking` | path-footprint `CropHull` → voxel → Euclidean clustering → cluster-height gate → convex hull | voxel 0.1/0.1/0.5 m, tolerance 0.15 m, min size 10, height ≥ 0.1 m, z ∈ [0.0, vehicle_height] | [crop](https://github.com/autowarefoundation/autoware_universe/blob/085b97a56aa25e21f9ec8f6accc05f8860c42c85/control/autoware_autonomous_emergency_braking/src/node.cpp#L954-L985), [cluster](https://github.com/autowarefoundation/autoware_universe/blob/085b97a56aa25e21f9ec8f6accc05f8860c42c85/control/autoware_autonomous_emergency_braking/src/node.cpp#L879-L920) |
| `autoware_planning_validator_rear_collision_checker` | x-passthrough → voxel → Euclidean → cluster-height gate | x ∈ [−100, +30] m, voxel 0.1/0.1/0.5 m, tolerance 0.15 m, min size **5**, height ≥ 0.1 m, z ∈ [**0.3**, vehicle_height − 1.0] | [cluster](https://github.com/autowarefoundation/autoware_universe/blob/085b97a56aa25e21f9ec8f6accc05f8860c42c85/planning/planning_validator/autoware_planning_validator_rear_collision_checker/src/rear_collision_checker.cpp#L329-L343) |
[...]

PR #13092, opened 2026-07-29
feat(autoware_obstacle_grid_extractor): add sensing-side obstacle-grid producer

youtalk:
- [x] #13091 

## Description

Adds `autoware_obstacle_grid_extractor`, the sensing-side producer for the perception-independent last-resort safety path. It is a stateless, `base_link`, per-cloud rasterizer that turns the no-ground obstacle point cloud into a single 2.5D occupancy/height grid (`grid_map_msgs/GridMap`, layers `max_height`/`min_height`/`point_count`/`low_max_height`). Processing is a single O(N) pass — for each finite, in-z-band point, compute its cell index and update the max/min height, the point count, and (for points at or below the `overhead_split` parameter) the tallest in-band height `low_max_height`. There is no KdTree, no Euclidean clustering and no convex hull. It subscribes only to the no-ground cloud and the static sensor→`base_link` extrinsic (never to objects/tracking/occupancy/map/trajectory/pose), so the last-resort path stays logically independent of the perception modeling stack. An empty cloud yields an all-`NaN` grid with a fresh stamp — the "alive, nothing detected" heartbeat. This producer replaces the ~8 duplicated crop+voxel+cluster+hull implementations the raw-cloud safety consumers each carried; the consumers migrate to it in follow-up PRs.

The `low_max_height` layer exists because a per-cell `(min_height, max_height)` pair cannot distinguish a wall (continuous mass from the ground up) from an overhead structure sharing its cell with ground returns (a gantry above ground residue) — both give the same signature. The extra layer lets a consumer put its height floor on in-band mass, so overhead-only content can never qualify a cell while tall in-band obstacles still do.

https://github.com/user-attachments/assets/6a1b0c0f-e673-41a7-86a8-55e8f029bb7b

## Related links

**Parent Issue:**

- https://github.com/autowarefoundation/autoware_universe/issues/12972

## How was this PR tested?

`colcon test --packages-select autoware_obstacle_grid_extractor` — 12 gtest cases pass with 0 failures in `autoware:universe-devel-jazzy`. Nine cover the rasterizer core: accumulation (two points in one cell), empty-cell `NaN`, narrow-obstacle ≥1 cell, crop z-band rejection and inclusivity at the bounds, points outside the ROI dropped, empty-cloud heartbeat, `low_max_height` discrimination (gantry-mixed cell vs overhead-only cell vs in-band tall wall), and non-finite-point rejection (a first-in-cell `NaN`-z point must not poison the height layers via `std::max`/`std::min` first-argument propagation while still incrementing `point_count`). Two cover the node wrapper: publishing a `base_link` grid when TF resolves, and dropping the cloud when the TF lookup fails. One is a size/coverage gate (geometry plus no-silent-drop on dense clouds; informational P99 latency).

Producer cost was additionally evaluated on real obstacle frames and a synthetic O(N) sweep: ~0.14–0.27 ms per real ~12 k-point cloud and ~2.1 ms for a ~200 k-point worst case — well under the 15 ms emergency-path budget and far below the existing `scan_ground_filter`+`concatenate` cost (~32 ms/frame). For end-to-end replay evidence the no-ground cloud was regenerated from raw lidar packets of a real drive by running the actual nebula decoders + `autoware_pointcloud_preprocessor` + `autoware_ground_segmentation`, and `extract()` was measured on the resulting real ~480 k-point clouds: mean 4.3 ms, p99 6.4 ms, max 7.0 ms over 301 frames, with 0/301 frames over the 15 ms emergency-path budget.

## Notes for reviewers

The rasterizer core is a headless library (`lib/obstacle_grid_extractor.cpp`) testable without rclcpp/TF; the node (`src/`) is a thin TF + pub/sub wrapper. QoS on the output grid is `RELIABLE KEEP_LAST(1)` (small fixed-size message on the last-resort path). Cell indexing uses `grid_map::getIndex` (the same mapping `autoware_costmap_generator::PointsToCostmap::fetchGridIndexFromPoint` wraps). Defaults: forward-biased ROI 60×40 m at 0.2 m, z-band `[-1.0, 3.0]` m (validated on real data), `overhead_split` 2.5 m (must be at least every consumer's own z-band top).

On a TF failure the node deliberately publishes nothing (not even the heartbeat), so consumers must treat a stale grid stamp as "unavailable", never as "clear". The consumer-side staleness watchdog is not part of this PR.

## Interface changes

New node. New published topic (no existing topic/parameter modified).

#### Additions and removals

|:--|:--|:--|:--|:--|
| Added | Sub | `~/input/pointcloud` (`/perception/obstacle_segmentation/pointcloud`) | `sensor_msgs/PointCloud2` | No-ground obstacle cloud (intra-process) |
| Added | Pub | `~/output/obstacle_grid` (`/sensing/obstacle_segmentation/obstacle_grid`) | `grid_map_msgs/GridMap` | 2.5D obstacle grid (max/min/in-band-max height + point count) |

## Effects on system behavior

None by itself — the node is new and not yet wired into any consumer (the launch wiring and the raw-cloud cut land in a later PR after all consumers migrate). When wired, it adds one O(N) producer (~2 ms worst case) co-located with the no-ground cloud and removes per-consumer re-clustering.

### DIFF
--- a/perception/autoware_obstacle_grid_extractor/config/obstacle_grid_extractor.param.yaml
+++ b/perception/autoware_obstacle_grid_extractor/config/obstacle_grid_extractor.param.yaml
@@ -1,0 +1,10 @@
+/**:
+  ros__parameters:
+    roi:
+      length_x: 60.0  # forward detection range [m] (far-region lanelet consumers)
+      length_y: 40.0  # lateral surround + adjacent-lane coverage [m]
+      offset_x: 20.0  # grid-center +x offset from base_link [m] (ego near the rear edge)
+    resolution: 0.2   # cell size [m]; <= existing ~0.3 m lateral jitter
+    crop:
+      z_min: -1.0     # union z-band lower bound [m]
+      z_max: 3.0      # union z-band upper bound [m]; consumers re-apply their own z gate
--- a/perception/autoware_obstacle_grid_extractor/config/obstacle_grid_extractor.param.yaml
+++ b/perception/autoware_obstacle_grid_extractor/config/obstacle_grid_extractor.param.yaml
@@ -8,3 +8,5 @@
     crop:
       z_min: -1.0     # union z-band lower bound [m]
       z_max: 3.0      # union z-band upper bound [m]; consumers re-apply their own z gate
+    overhead_split: 2.5  # low_max_height ceiling [m]; must be >= every consumer's own z-band top
+                         # (vehicle_height + margin) so in-band mass is never hidden from them
--- a/perception/autoware_obstacle_grid_extractor/config/obstacle_grid_extractor.param.yaml
+++ b/perception/autoware_obstacle_grid_extractor/config/obstacle_grid_extractor.param.yaml
@@ -1,12 +1,19 @@
 /**:
   ros__parameters:
     roi:
-      length_x: 60.0  # forward detection range [m] (far-region lanelet consumers)
-      length_y: 40.0  # lateral surround + adjacent-lane coverage [m]
-      offset_x: 20.0  # grid-center +x offset from base_link [m] (ego near the rear edge)
-    resolution: 0.2   # cell size [m]; <= existing ~0.3 m lateral jitter
+      # Union of the consumers' own configured ranges: rear_collision_checker crops x to [-100, 30],
+      # intersection_collision_checker crops x to [-50, 50] and extends the crossing lanelet by 50 m
+      # (hence |y| <= 50), and AEB / obstacle_collision_checker are unbounded forward, so +80 m is
+      # taken from the RSS stopping distance at their configured t_response / a_ego_min.
+      length_x: 180.0  # x extent [m]: with offset_x below this covers x in [-100, +80]
+      length_y: 100.0  # y extent [m]: y in [-50, +50]
+      offset_x: -10.0  # grid-center +x offset from base_link [m]
+    resolution: 0.2   # cell size [m]; <= existing ~0.3 m lateral jitter, and the 0.283 m cell
+                      # diagonal is what the consumer-parity evaluation was validated against
     crop:
-      z_min: -1.0     # union z-band lower bound [m]
-      z_max: 3.0      # union z-band upper bound [m]; consumers re-apply their own z gate
+      z_min: -2.5     # union z-band lower bound [m]; matches ground_segmentation margin_min_z so
+                      # far returns on a vertical curve survive to the consumers
+      z_max: 3.5      # union z-band upper bound [m]; matches ground_segmentation
+                      # detection_range_z_max. Consumers re-apply their own z gate
     overhead_split: 2.5  # low_max_height ceiling [m]; must be >= every consumer's own z-band top
                          # (vehicle_height + margin) so in-band mass is never hidden from them
--- a/perception/autoware_obstacle_grid_extractor/lib/obstacle_grid_extractor.cpp
+++ b/perception/autoware_obstacle_grid_extractor/lib/obstacle_grid_extractor.cpp
@@ -1,0 +1,79 @@
+// Copyright 2026 The Autoware Contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#include "autoware/obstacle_grid_extractor/obstacle_grid_extractor.hpp"
+
+#include <grid_map_ros/GridMapRosConverter.hpp>
+#include <rclcpp/time.hpp>
+
+#include <algorithm>
+#include <cmath>
+#include <limits>
+
+namespace autoware::obstacle_grid_extractor
+{
+namespace
+{
+constexpr char kMaxHeight[] = "max_height";
+constexpr char kMinHeight[] = "min_height";
+constexpr char kPointCount[] = "point_count";
+}  // namespace
+
+ObstacleGridExtractor::ObstacleGridExtractor(const ExtractorParams & params)
+: params_(params), grid_({kMaxHeight, kMinHeight, kPointCount})
+{
+  grid_.setFrameId("base_link");
+  grid_.setGeometry(
+    grid_map::Length(params_.roi_length_x, params_.roi_length_y), params_.resolution,
+    grid_map::Position(params_.roi_offset_x, 0.0));  // forward-biased rectangular ROI
+}
+
+grid_map_msgs::msg::GridMap ObstacleGridExtractor::extract(
+  const pcl::PointCloud<pcl::PointXYZ> & cloud_base_link,
+  const std_msgs::msg::Header & header) const
+{
+  const float nan = std::numeric_limits<float>::quiet_NaN();
+  grid_[kMaxHeight].setConstant(nan);
+  grid_[kMinHeight].setConstant(nan);
+  grid_[kPointCount].setConstant(nan);  // empty cell = NaN
+
+  auto & hi = grid_[kMaxHeight];
+  auto & lo = grid_[kMinHeight];
+  auto & cnt = grid_[kPointCount];
+
+  for (const auto & p : cloud_base_link) {  // O(N) single pass
+    if (p.z < params_.crop_z_min || p.z > params_.crop_z_max) {
+      continue;
+    }
+    grid_map::Index idx;
+    if (!grid_.getIndex(grid_map::Position(p.x, p.y), idx)) {
+      continue;  // outside ROI
+    }
+    float & c = cnt(idx(0), idx(1));
+    if (std::isnan(c)) {
+      c = 1.0f;
+      hi(idx(0), idx(1)) = p.z;
+      lo(idx(0), idx(1)) = p.z;
+    } else {
+      c += 1.0f;
+      hi(idx(0), idx(1)) = std::max(hi(idx(0), idx(1)), p.z);
+      lo(idx(0), idx(1)) = std::min(lo(idx(0), idx(1)), p.z);
+    }
+  }
+
+  grid_.setTimestamp(rclcpp::Time(header.stamp).nanoseconds());
+  grid_map_msgs::msg::GridMap msg = *grid_map::GridMapRosConverter::toMessage(grid_);
+  msg.header = header;  // base_link, exact cloud stamp -> all-NaN = heartbeat
+  return msg;
+}
+}  // namespace autoware::obstacle_grid_extractor
--- a/perception/autoware_obstacle_grid_extractor/lib/obstacle_grid_extractor.cpp
+++ b/perception/autoware_obstacle_grid_extractor/lib/obstacle_grid_extractor.cpp
@@ -27,10 +27,11 @@ namespace
 constexpr char kMaxHeight[] = "max_height";
 constexpr char kMinHeight[] = "min_height";
 constexpr char kPointCount[] = "point_count";
+constexpr char kLowMaxHeight[] = "low_max_height";
 }  // namespace
 
 ObstacleGridExtractor::ObstacleGridExtractor(const ExtractorParams & params)
-: params_(params), grid_({kMaxHeight, kMinHeight, kPointCount})
+: params_(params), grid_({kMaxHeight, kMinHeight, kPointCount, kLowMaxHeight})
 {
   grid_.setFrameId("base_link");
   grid_.setGeometry(
@@ -46,12 +47,17 @@ grid_map_msgs::msg::GridMap ObstacleGridExtractor::extract(
   grid_[kMaxHeight].setConstant(nan);
   grid_[kMinHeight].setConstant(nan);
   grid_[kPointCount].setConstant(nan);  // empty cell = NaN
+  grid_[kLowMaxHeight].setConstant(nan);
 
   auto & hi = grid_[kMaxHeight];
   auto & lo = grid_[kMinHeight];
   auto & cnt = grid_[kPointCount];
+  auto & low_hi = grid_[kLowMaxHeight];
 
   for (const auto & p : cloud_base_link) {  // O(N) single pass
+    if (!std::isfinite(p.x) || !std::isfinite(p.y) || !std::isfinite(p.z)) {
+      continue;  // a NaN z would otherwise pass the crop and poison the cell's height layers
+    }
     if (p.z < params_.crop_z_min || p.z > params_.crop_z_max) {
       continue;
     }
@@ -68,6 +74,10 @@ grid_map_msgs::msg::GridMap ObstacleGridExtractor::extract(
       c += 1.0f;
       hi(idx(0), idx(1)) = std::max(hi(idx(0), idx(1)), p.z);
       lo(idx(0), idx(1)) = std::min(lo(idx(0), idx(1)), p.z);
+    }
+    if (p.z <= params_.overhead_split) {
+      float & lm = low_hi(idx(0), idx(1));
+      lm = std::isnan(lm) ? p.z : std::max(lm, p.z);
     }
   }
 
--- a/perception/autoware_obstacle_grid_extractor/lib/obstacle_grid_extractor.cpp
+++ b/perception/autoware_obstacle_grid_extractor/lib/obstacle_grid_extractor.cpp
@@ -50,15 +50,15 @@ grid_map_msgs::msg::GridMap ObstacleGridExtractor::extract(
   grid_[kPointCount].setConstant(nan);  // empty cell = NaN
   grid_[kLowMaxHeight].setConstant(nan);
 
-  auto & hi = grid_[kMaxHeight];
-  auto & lo = grid_[kMinHeight];
-  auto & cnt = grid_[kPointCount];
-  auto & low_hi = grid_[kLowMaxHeight];
-
   // Guarded: an empty cloud may carry no field descriptors at all, and PointCloud2ConstIterator
   // throws when the requested field is missing. Skipping straight to the all-NaN heartbeat keeps
   // that case on the same path as a cloud whose every point is cropped away.
   if (static_cast<uint64_t>(cloud.width) * cloud.height > 0) {
+    auto & hi = grid_[kMaxHeight];
+    auto & lo = grid_[kMinHeight];
+    auto & cnt = grid_[kPointCount];
+    auto & low_hi = grid_[kLowMaxHeight];
+
     sensor_msgs::PointCloud2ConstIterator<float> it_x(cloud, "x");
     sensor_msgs::PointCloud2ConstIterator<float> it_y(cloud, "y");
     sensor_msgs::PointCloud2ConstIterator<float> it_z(cloud, "z");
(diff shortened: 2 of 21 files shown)


perception/autoware_obstacle_grid_extractor/lib/obstacle_grid_extractor.cpp
```diff
@@ -0,0 +1,89 @@
...
+      float & lm = low_hi(idx(0), idx(1));
+      lm = std::isnan(lm) ? p.z : std::max(lm, p.z);
+    }
+  }
+
+  grid_.setTimestamp(rclcpp::Time(header.stamp).nanoseconds());
+  grid_map_msgs::msg::GridMap msg = *grid_map::GridMapRosConverter::toMessage(grid_);
+  msg.header = header;  // base_link, exact cloud stamp -> all-NaN = heartbeat
+  return msg;
+}
```
interimadd, line 102: I suggest using `PointCloud2` directly instead of `pcl::PointCloud` like:
```cpp
grid_map_msgs::msg::GridMap ObstacleGridExtractor::extract(
  sensor_msgs::msg::PointCloud2 cloud
)
```

Here are a few reasons:
1. PointCloud2 messages include header information.
2. It keeps the input and output types consistent. Since the input is a ROS 2 message, this function clearly acts as a ROS 2 to ROS 2 message converter.
3. We can remove the dependency on the PCL library, which reduces conversion overhead.

Note that we don't need the PCL library to iterate over points in PointCloud2. We can use PointCloud2ConstIterator instead:
```cpp
sensor_msgs::PointCloud2ConstIterator<float> it_x(cloud, "x"), it_y(cloud, "y"), it_z(cloud, "z");
for (; it_x != it_x.end(); ++it_x, ++it_y, ++it_z) {
  float x = *it_x, y = *it_y, z = *it_z;
  // ...
}
```

If the input PointCloud2's frame_id is not `base_link`, we might need to raise an exception. Alternatively, I think it would be better to align the frame_id of `grid_map_msgs` with that of the input PointCloud2.

perception/autoware_obstacle_grid_extractor/src/obstacle_grid_extractor_node.cpp
```diff
@@ -0,0 +1,70 @@
...
+    RCLCPP_WARN_THROTTLE(get_logger(), *get_clock(), 1000, "TF to base_link failed: %s", e.what());
+    // Never publish a wrong-frame grid. Nothing (not even the all-NaN heartbeat) is published on
+    // this path, so consumers MUST treat a stale grid stamp as "unavailable", never as "clear".
+    return;
+  }
+  pcl::PointCloud<pcl::PointXYZ> cloud;
+  pcl::fromROSMsg(in_base_link, cloud);
+  std_msgs::msg::Header header = msg->header;
+  header.frame_id = "base_link";
+  pub_->publish(extractor_->extract(cloud, header));  // empty cloud -> heartbeat
```
interimadd: If `ObstacleGridExtractor::extract()` take PointCloud2 message directly, then these lines can be simplified like:
```
pub_->publish(extractor_->extract(cloud));
```

interimadd (changes requested):
Thanks for the PR!
I have a suggestion to improve the `extract()` interface. What do you think?

Also, it would be great if we could structure the test cases with clear Arrange/Act/Assert sections. This would make it easier to see what function is being tested, under what conditions, and what the expected output is.
https://github.com/autowarefoundation/autoware_universe/pull/13091#pullrequestreview-4843711539

perception/autoware_obstacle_grid_extractor/lib/obstacle_grid_extractor.cpp
youtalk: Done — `extract()` now takes `const sensor_msgs::msg::PointCloud2 &` and iterates with `PointCloud2ConstIterator`. The `Header` argument is gone and so is PCL (`libpcl-all-dev`, `pcl_conversions`, `find_package(PCL)`).

On `frame_id` I took your second option: the ROI is rasterized in the cloud's own frame and the grid inherits the input header verbatim, so the class holds no hardcoded frame. The node keeps the `base_link` guarantee by transforming with TF and dropping the cloud when that lookup fails.

perception/autoware_obstacle_grid_extractor/src/obstacle_grid_extractor_node.cpp
youtalk: Done, with one extra line:

```cpp
in_base_link.header.stamp = msg->header.stamp;
pub_->publish(extractor_->extract(in_base_link));
```

`tf2::doTransform()` replaces the whole header with the transform's, so without that line the grid would carry the TF lookup time rather than the sensing time. A node test now asserts the published stamp equals the input cloud's.

youtalk:
@interimadd Thank you for your detailed review. I've fixed them. Please review again.

interimadd (approved):
Thank you for addressing comments!
`ObstacleGridExtractor` is well decoupled with rclcpp::Node and have clear interface named `extract()`. LGTM

perception/autoware_obstacle_grid_extractor/src/obstacle_grid_extractor_node.cpp
```diff
@@ -0,0 +1,68 @@
...
+    "~/output/obstacle_grid", rclcpp::QoS(1).reliable());
+}
+
+void ObstacleGridExtractorNode::onCloud(const sensor_msgs::msg::PointCloud2::ConstSharedPtr msg)
+{
+  sensor_msgs::msg::PointCloud2 in_base_link;
+  try {
+    const auto tf = tf_buffer_.lookupTransform(
+      "base_link", msg->header.frame_id, msg->header.stamp, rclcpp::Duration::from_seconds(0.1));
+    tf2::doTransform(*msg, in_base_link, tf);  // static extrinsic only
```
technolojin, line 70: *f the given pointcloud frame is same 'base_link', tf2::doTransform could be avoided.

example https://github.com/autowarefoundation/autoware_universe/blob/cf7e9a3463e30a1e6b4cad45e4f9aef7638fe5d1/perception/autoware_compare_map_segmentation/src/voxel_based_compare_map_filter/node.cpp#L175

perception/autoware_obstacle_grid_extractor/config/obstacle_grid_extractor.param.yaml
```diff
@@ -0,0 +1,12 @@
+/**:
+  ros__parameters:
+    roi:
+      length_x: 60.0  # forward detection range [m] (far-region lanelet consumers)
+      length_y: 40.0  # lateral surround + adjacent-lane coverage [m]
+      offset_x: 20.0  # grid-center +x offset from base_link [m] (ego near the rear edge)
```
technolojin: the range seems narrow. 

* rear_collision_checker: -100 m < x < 30 m
* intersection_collision_checker: -50 m < x < 50 m
* AEB / obstacle_collision_checker: unbounded, maximum of the obstacle pointcloud range 
* Obstacle pointcloud: -100 m < x < 150 m, -70 m < y < 70 m. 

bench test and message size analysis shall be done by larger grid size.

perception/autoware_obstacle_grid_extractor/config/obstacle_grid_extractor.param.yaml
```diff
@@ -0,0 +1,12 @@
...
+    resolution: 0.2   # cell size [m]; <= existing ~0.3 m lateral jitter
+    crop:
+      z_min: -1.0     # union z-band lower bound [m]
+      z_max: 3.0      # union z-band upper bound [m]; consumers re-apply their own z gate
```
technolojin: z-margin also shall be expanded.
In case of vertical curves (starting of up-slope, edge of down-slope) this narrow, this narrow bound remove obstacles in far ranges.

In the rule based ground segmentation, the bound set 
https://github.com/autowarefoundation/autoware_launch/blob/cbfcd3faab22571b683dfe0ee9eadad70bb92b5b/autoware_launch/config/perception/obstacle_segmentation/ground_segmentation/ground_segmentation.param.yaml#L12
```yaml
        margin_max_z: 2.5  # to extend the crop box max_z from vehicle_height
        margin_min_z: -2.5 # to extend the crop box min_z from ground
```
for the initial range, and https://github.com/autowarefoundation/autoware_launch/blob/cbfcd3faab22571b683dfe0ee9eadad70bb92b5b/autoware_launch/config/perception/obstacle_segmentation/ground_segmentation/ground_segmentation.param.yaml#L28
```yaml
        detection_range_z_max: 3.5
```
for the ground segmentation (differential from estimated ground shape)

technolojin (changes requested):
The default grid size is small. I reviewed it that the grid need to be expanded.

While centerpoint has ±76.8 m , size of 0.32 m, it has 230k possible grid. Occupied voxels are 40k-60k in urban intersection area,  17% - 26% of sparsity. BUT, it is when ground points are include.
My estimation of sparsity of obstacle pointcloud will drop to 3-5%.

Converting sparse pointcloud to filled grid will cause inefficiency. More wider grid causes more inefficiency, considering far distance has more sparsity.

I would like to request additional analysis of computing cost and message size, base on consumers required distance range.

perception/autoware_obstacle_grid_extractor/src/obstacle_grid_extractor_node.cpp
youtalk: Done — the node now returns early when `header.frame_id` is already `base_link`, matching the guard in your `compare_map_segmentation` example.

Worth noting this is the production path rather than an optimization for an edge case: `concatenate_and_time_sync_node.param.yaml` sets `output_frame: base_link`, so the no-ground cloud always arrives in `base_link` and every frame was paying a full `doTransform` copy of a >100 k-point cloud to reproduce it unchanged.

One consequence for the tests: the existing "TF resolves" case published a `base_link` cloud, so after this change it no longer reaches `doTransform` at all. I added `TransformsNonBaseLinkCloudWhenTfResolves`, which publishes in a separate `sensor_frame` with a static `(+1.0, 0, +2.0)` extrinsic and asserts the point lands at the transformed position — and that the untransformed position is empty — so that branch stays covered.

perception/autoware_obstacle_grid_extractor/config/obstacle_grid_extractor.param.yaml
youtalk: Fixed — the band is now `[-2.5, 3.5]`, taken from the `ground_segmentation` values you linked (`margin_min_z` -2.5, `detection_range_z_max` 3.5).

You were right that the old band was cutting into real returns. Measured over the recorded no-ground clouds:

| bag              | below -1.0 m | above 3.0 m | still outside `[-2.5, 3.5]` |
| ---------------- | ------------ | ----------- | --------------------------- |
| `ariakeB_native` | 0.03 %       | 1.35 %      | 0.00 %                      |
| `prdjt_dense`    | 0.01 %       | 4.19 %      | 1.36 %                      |
| `iwaki_stoptest` | 8.31 %       | 0.03 %      | 0.29 %                      |

`prdjt_dense` loses 4.19 % of its returns above `z=3.0`, and `iwaki_stoptest` loses 8.31 % below `z=-1.0` — the latter being exactly the far-field sub-ground returns on a vertical curve that showed up in the consumer-parity run. Those were being discarded at the producer, before any consumer could gate on them.

This one is free: the grid is 2D, so the z-band does not affect message size at all. It recovers 1.0–2.4 pp of in-ROI returns for no bytes.

`overhead_split` stays at 2.5 m — it is the `low_max_height` ceiling rather than a crop bound, and it still has to sit at or above every consumer's own z-band top.

perception/autoware_obstacle_grid_extractor/config/obstacle_grid_extractor.param.yaml
youtalk: Expanded — the ROI is now `x ∈ [-100, +80]`, `y ∈ [-50, +50]` at the same 0.2 m cell, set from the consumer ranges you listed:

- `rear_collision_checker` crop box `x ∈ [-100, 30]` → covered.
- `intersection_collision_checker` crop box `x ∈ [-50, 50]` → covered. It has no `y` filter, but it extends the crossing lanelet by `detection_range` (50 m), so `|y| ≤ 50` is what that reach needs.
- AEB / `obstacle_collision_checker` are unbounded forward, so `+80` m is taken from the RSS stopping distance at their configured `t_response` / `a_ego_min` rather than from the sensor's reach — see the open question at the end.

Real returns reaching a cell go from 68–78 % to 98–99 %. The full analysis is in `benchmark/README.md`; the sizing part is in the reply to your top-level comment, since it is the same measurement.

youtalk:
**Edited 2026-08-05:** added the "What this means for transport" section and updated open question 2 with concrete numbers — working through the size table below showed the new default reaches wire-size parity with the dense no-ground cloud, which changes the transport story and was worth stating explicitly.

Here is the computing-cost and message-size analysis you asked for. Your sparsity estimate holds up, and it turned out to matter more than I expected — it changed the ROI, a claim in the README, and (per the new section below) the transport-level framing of this producer.

Method: the real `ObstacleGridExtractor` run offline over 2086 frames of recorded real no-ground clouds — two urban drives (125–132 k points/frame) plus one sparse scene (306 points/frame). Each geometry is timed in its own process; timing several in one process interleaves their allocations and inflates the larger ones by up to 2×. Message sizes are the actual CDR-serialized sizes, not estimates. Everything below is reproduced in `benchmark/README.md`.

## Sparsity and message size

| ROI (`x` × `y`)     | cells   | msg size | occupied cells | occupancy | in-ROI returns |
| ------------------- | ------- | -------- | -------------- | --------- | -------------- |
| `[-10, 50] × ±20` (old default) | 60 000 | 0.96 MB | 4 000–4 100 | 6.7 % | 68–78 % |
| `[-100, 50] × ±50`  | 375 000 | 6.0 MB   | 7 700–8 200    | 2.1 %     | 96–97 %        |
| `[-100, 80] × ±50` (new default) | 450 000 | 7.2 MB | 8 100–9 000 | 1.8–2.0 % | 98–99 % |
| `[-100, 50] × ±70`  | 525 000 | 8.4 MB   | 7 900–8 400    | 1.5–1.6 % | 96–98 %        |
| `[-100, 100] × ±50` | 500 000 | 8.0 MB   | 8 200–9 300    | 1.6–1.9 % | 98–99 %        |
| `[-100, 80] × ±60`  | 540 000 | 8.6 MB   | 8 200–9 200    | 1.5–1.7 % | 98–99 %        |
| `[-100, 150] × ±70` (full cloud range) | 875 000 | 14.0 MB | 8 500–9 700 | 1.0–1.1 % | 99 % |

Your 3–5 % estimate for obstacle-only occupancy is right at mid-range and it keeps falling: **1.0–1.1 % at the full cloud range**. Concretely, growing the grid **14.6×** (60 k → 875 k cells) raises the occupied-cell count only **~2.2×**.

For scale, the clouds themselves serialize (as `PointXYZIRC`, 16 B/point) to ~2.1 MB for the 132 k-point urban drives above, and ~7.7 MB for the ~480 k-point 8-lidar clouds of the end-to-end replay in the PR description. So at the full cloud range the 14 MB grid is **6.7× the urban cloud and 1.8× even the dense cloud it summarises** — the "compression" has become an expansion, which is exactly your point.

One result I did not expect: **widening `y` is a strictly worse trade than extending `x`**. `[-100, 50] × ±70` costs 17 % more bytes than `[-100, 80] × ±50` and holds *fewer* occupied cells, because obstacle returns concentrate along the road rather than out to the sides. That is why the new default reaches further forward but only to `|y| ≤ 50`.

## Computing cost

`extract()` at the new `[-100, 80] × ±50` @0.2 m default:

| recorded drive | points/frame | mean    | p99     | max     |
| -------------- | ------------ | ------- | ------- | ------- |
| urban A        | 131 751      | 2.31 ms | 3.30 ms | 3.58 ms |
| urban B        | 125 208      | 2.28 ms | 2.73 ms | 6.98 ms |
| sparse scene   | 306          | 0.95 ms | 0.99 ms | 4.33 ms |

The sparse-scene row is the one that matters: **306 points per frame still costs ~1 ms**. The same bag costs 0.06 ms at 60 k cells and 2.67 ms at 875 k cells — with essentially no points in the scene. So cost tracks grid *area*, not point count: the per-frame clear of the four layers and the conversion to the message are paid whether or not anything arrives.

That makes this PR's "O(N) producer, never the bottleneck" claim wrong at any realistic ROI, and I have corrected it to `O(N + C)` in the README with the `C` term called out as dominant. Thanks for pushing on this — the original claim was only true at the small default it was measured at.

## What this means for transport

Being honest about a consequence of the numbers above, per stream at 10 Hz:

| stream | per message | bandwidth |
| ------ | ----------- | --------- |
| no-ground cloud (dense, ~480 k pts) | ~7.7 MB | ~77 MB/s |
| grid, old default (60 k cells) | 0.96 MB | 9.6 MB/s |
| grid, new default (450 k cells) | 7.2 MB | 72 MB/s |

The old default was a genuine **1/8 wire-size compression** of the dense cloud. The new default is at **parity (0.94×)** with it — and 3.4× *larger* than the urban clouds. So at the coverage the consumers require, this producer is **not a bandwidth reduction at all**; its remaining transport-level benefit is fan-out (N consumers each pulling the raw cloud collapse to one shared fixed-size grid stream), and its main wins are CPU (one rasterization replacing N clustering pipelines) and the perception-independence of the safety path.

Two practical consequences for anyone deploying the grid across a link rather than intra-host:

- A 100 Mbps-class link carries the old grid at full 10 Hz but carries neither the dense cloud nor the new grid (7.2 MB × 10 Hz = 576 Mbps; it would degrade to below 2 Hz). Crossing a constrained link at the new default requires ≥1 GbE, and still occupies ~58 % of it.
- A RELIABLE writer whose transport high watermark is below one frame stalls in `publish()` — we have observed sub-second stalls of this class with the *old* 0.96 MB frame against a 500 kB watermark in a link-constrained rehearsal. At 7.2 MB/frame, watermark headroom has to be provisioned deliberately (e.g. CycloneDDS `WhcHigh` ≥ one frame with margin).

Neither point blocks this PR — the intended deployment is co-located with the no-ground producer, and `KEEP_LAST(1)` already bounds the queue — but the size table above is the input a deployer needs, and it feeds directly into open question 2 below.

## What changed

- ROI `x ∈ [-100, +80]`, `y ∈ [-50, +50]` — the union of the consumers' configured ranges (details in the thread on the param file). In-ROI returns go from 68–78 % to 98–99 %.
- Crop z-band `[-2.5, 3.5]`, matching `ground_segmentation` (details in that thread).
- The `base_link` short-circuit.
- README/QoS comment: the grid is fixed-size but not "small" — 7.2 MB per message, 72 MB/s at 10 Hz.

Resolution stays at 0.2 m deliberately: the consumer-parity evaluation was validated against the 0.283 m cell diagonal, and 0.3 m would take that to 0.424 m, so moving it would invalidate that evidence rather than being a free knob.

## Two open questions for you

1. **Forward extent.** `+80` m is the RSS stopping distance at AEB's configured `t_response` / `a_ego_min`, not a measured requirement. Going to the sensor's full `+150` m costs 14.0 MB instead of 7.2 MB and buys about 1 pp of returns. If there is an operating-speed envelope that needs more than 80 m forward, tell me and I will set it — I would rather take the number from you than guess.
2. **Representation.** At ~2 % occupancy the dense `GridMap` carries ~98 % empty cells, and no ROI choice fixes that — only a sparse encoding would. The numbers make the case concrete: ~8 600 occupied cells at 20 B each (cell index + 4 layers) is **~0.2 MB/frame, ~1/40 of the dense message and smaller than even the old default** — a sparse encoding is the only form in which this producer stays a wire-size compression at the coverage the consumers need, on top of removing the `O(C)` clear that dominates the CPU cost. It is a larger change than this PR (the `GridMap` product shape is what the consumer migrations are built against), so I have not touched it here. If you want it reconsidered, I think it belongs on the parent issue #12972 rather than in this PR; happy to open it there.

technolojin:
1. Forward extent. 
I agree that 150 m is too much, but it may require more than 80 m. The planning nodes tracks time-wise obstacles in planned paths, so the object information shall be provided farther than 80 m, to estimate and track leading vehicle or dropped box. 

2. Representation. 
Encoding sparse grid will very work well, if all the consumers implementation/computation cost is within range.
I also think that passing clustered object (objects with polygon shape, convex-hull) also reduce data scale enough.
It would be an interesting analysis.

PR #13106, opened 2026-07-30
feat(cuda_pointcloud_preprocessor): add input bounds diagnostics

mojomex:
## Description

Add explicit input bounds to `autoware_cuda_pointcloud_preprocessor`:

- add `max_input_point_count`
- add twist/IMU queue bounds derived from maximum expected frequency and queue window parameters
- truncate point clouds above the configured maximum to the first N points
- discard oldest twist/IMU samples when the bounded queues overflow
- report truncation and dropped input samples through diagnostics as ERROR

This PR keeps the existing dynamic CUDA buffer allocation behavior. The follow-up PRs use these bounds to move internal preprocessor GPU allocations to startup and then remove remaining runtime Thrust allocation paths.

## Stack

This is PR 1 of 5 in a stack. Every PR targets `autowarefoundation:main` (GitHub does not allow cross-repository stacks, and the branches live on a fork), so each **Files changed** tab shows the *cumulative* diff of the stack up to that PR.

**To review only one PR's own changes, use its Own diff link below.**

| # | PR | Own diff |
| --- | --- | --- |
| 1 | #13106 feat: add input bounds diagnostics **← this PR** | [Files changed](https://github.com/autowarefoundation/autoware_universe/pull/13106/files) (bottom of stack) |
| 2 | #13134 perf: allocate fixed GPU buffers at startup | [preproc-bounds-diagnostics → preproc-startup-capacity](https://github.com/mojomex/autoware_universe/compare/preproc-bounds-diagnostics...preproc-startup-capacity) |
| 3 | #13135 perf: avoid runtime thrust allocations | [preproc-startup-capacity → preproc-remove-runtime-thrust-allocs](https://github.com/mojomex/autoware_universe/compare/preproc-startup-capacity...preproc-remove-runtime-thrust-allocs) |
| 4 | #13136 refactor: clarify queue update flow | [preproc-remove-runtime-thrust-allocs → preproc-queue-update-cleanup](https://github.com/mojomex/autoware_universe/compare/preproc-remove-runtime-thrust-allocs...preproc-queue-update-cleanup) |
| 5 | #13137 test: cover capacity bounds behavior | [preproc-queue-update-cleanup → preproc-behavior-tests](https://github.com/mojomex/autoware_universe/compare/preproc-queue-update-cleanup...preproc-behavior-tests) |

Merge in stack order (1 → 5); each PR depends on the one below it. The whole stack is meant as one unit: tests are added at the top to keep the individual PRs small, so please review in stack order but test with the full stack applied (`preproc-behavior-tests`).

## CUDA Blocking Call Flamegraphs

Before/after of this **PR stack**.
Performance improvements come from mojomex/autoware_universe#12 and mojomex/autoware_universe#13.

image

image

#### Additions and removals

| Added | `max_input_point_count` | `integer` | `691200` | Maximum number of input points processed per cloud; larger inputs are reported as errors and truncated to the first N points. |
| Added | `max_twist_frequency_hz` | `double` | `200.0` | Maximum expected twist input frequency used to bound the twist queue. |
| Added | `max_imu_frequency_hz` | `double` | `200.0` | Maximum expected IMU input frequency used to bound the IMU queue. |

#### Additions and removals

| Change type | DiagnosticStatus name | Diagnostic field / status | Level | Description |
|:------------|:----------------------|:--------------------------|:------|:------------|
| Added | Node fully-qualified name from `this->get_fully_qualified_name()` | `Input point count` | OK | Reports the input point count for the current cloud. |
| Added | Node fully-qualified name from `this->get_fully_qualified_name()` | `Maximum input point count` | OK | Reports the configured maximum accepted input point count. |
| Added | Node fully-qualified name from `this->get_fully_qualified_name()` | `Truncated input point count` | ERROR when non-zero | Reports how many input points were truncated because `max_input_point_count` was exceeded. |
| Added | Node fully-qualified name from `this->get_fully_qualified_name()` | `Twist queue size` | OK | Reports the current bounded twist queue size. |
| Added | Node fully-qualified name from `this->get_fully_qualified_name()` | `Maximum twist queue size` | OK | Reports the configured maximum twist queue size. |
| Added | Node fully-qualified name from `this->get_fully_qualified_name()` | `Dropped twist count` | ERROR when non-zero | Reports how many twist messages were dropped because the bounded queue overflowed. |
| Added | Node fully-qualified name from `this->get_fully_qualified_name()` | `IMU queue size` | OK | Reports the current bounded IMU queue size. |
| Added | Node fully-qualified name from `this->get_fully_qualified_name()` | `Maximum IMU queue size` | OK | Reports the configured maximum IMU queue size. |
| Added | Node fully-qualified name from `this->get_fully_qualified_name()` | `Dropped IMU count` | ERROR when non-zero | Reports how many IMU messages were dropped because the bounded queue overflowed. |

## Effects on system behavior

- Point clouds larger than `max_input_point_count` are truncated to the first configured maximum number of points instead of being processed at full size.
- Twist and IMU queues are bounded using the configured maximum expected frequencies; when the bounds are exceeded, older/excess samples are dropped.
- Exceeding any of these input bounds is surfaced as an ERROR diagnostic.

## Validation

The below tests have been performed on the **whole PR stack** (on #13137).

Build and whole-system benchmark passed on our test bench.

Also did a side-by-side before (`/a/`) and after (`/b/`) comparison using [TIER IV INTERNAL](https://github.com/tier4/pa-tools). Reproduce: [TIER IV INTERNAL](https://drive.google.com/file/d/180QIIHHrudUi21tAXw7va2S4UK2-R90y/view?usp=sharing)

Screenshot from 2026-08-04 21-32-27

No behavior change observed. The 3 discrepancies between `a` and `b` are due to cuda_blackboard dropping messages when running 2 Autoware instances on my PC.

Unit tests that validate input bound handling are in the top of the stack, #13137.

### DIFF
--- a/sensing/autoware_cuda_pointcloud_preprocessor/config/cuda_pointcloud_preprocessor.param.yaml
+++ b/sensing/autoware_cuda_pointcloud_preprocessor/config/cuda_pointcloud_preprocessor.param.yaml
@@ -8,6 +8,9 @@
     processing_time_threshold_sec: 0.01
     timestamp_mismatch_fraction_threshold: 0.01
     enable_ring_outlier_filter: true
+    max_input_point_count: 691200  # Hesai OT128, dual return, high resolution mode
+    max_twist_frequency_hz: 200.0
+    max_imu_frequency_hz: 200.0
 
     crop_box.min_x: [0.0]
     crop_box.min_y: [0.0]
--- a/sensing/autoware_cuda_pointcloud_preprocessor/include/autoware/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor.hpp
+++ b/sensing/autoware_cuda_pointcloud_preprocessor/include/autoware/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor.hpp
@@ -30,6 +30,7 @@
 
 #include <thrust/device_vector.h>
 
+#include <cstddef>
 #include <cstdint>
 #include <deque>
 #include <memory>
@@ -56,6 +57,7 @@ public:
   void setRingOutlierFilterParameters(const RingOutlierFilterParameters & ring_outlier_parameters);
   void setRingOutlierFilterActive(const bool enable_filter);
   void setUndistortionType(const UndistortionType & undistortion_type);
+  void setMaxInputPointCount(const std::size_t max_input_point_count);
 
   void preallocateOutput();
   [[nodiscard]] ProcessingStats getProcessingStats() const { return stats_; }
@@ -80,8 +82,9 @@ private:
 
   int num_rings_{};
   int max_points_per_ring_{};
-  size_t num_raw_points_{};
-  size_t num_organized_points_{};
+  std::size_t num_raw_points_{};
+  std::size_t num_organized_points_{};
+  std::size_t max_input_point_count_{};
 
   std::vector<sensor_msgs::msg::PointField> point_fields_;
   std::unique_ptr<cuda_blackboard::CudaPointCloud2> output_pointcloud_ptr_;
--- a/sensing/autoware_cuda_pointcloud_preprocessor/include/autoware/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor_node.hpp
+++ b/sensing/autoware_cuda_pointcloud_preprocessor/include/autoware/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor_node.hpp
@@ -39,6 +39,7 @@
 #include <tf2_ros/buffer.h>
 #include <tf2_ros/transform_listener.h>
 
+#include <cstddef>
 #include <deque>
 #include <memory>
 #include <string>
@@ -73,6 +74,23 @@ CHECK_OFFSET(OutputPointType, autoware::point_types::PointXYZIRCAEDT, intensity)
 CHECK_OFFSET(OutputPointType, autoware::point_types::PointXYZIRCAEDT, return_type);
 CHECK_OFFSET(OutputPointType, autoware::point_types::PointXYZIRCAEDT, channel);
 
+struct InputBoundsParams
+{
+  std::size_t max_input_point_count{};
+  std::size_t max_twist_subscriber_queue_size{};
+  std::size_t max_imu_subscriber_queue_size{};
+  std::size_t max_twist_queue_size{};
+  std::size_t max_imu_queue_size{};
+};
+
+struct InputBoundsStatus
+{
+  std::size_t input_point_count{};
+  std::size_t truncated_point_count{};
+  std::size_t dropped_twist_count{};
+  std::size_t dropped_imu_count{};
+};
+
 class CudaPointcloudPreprocessorNode : public rclcpp::Node
 {
 public:
@@ -97,6 +115,8 @@ private:
 
   void updateTwistQueue(std::uint64_t first_point_stamp);
   void updateImuQueue(std::uint64_t first_point_stamp);
+  void boundTwistQueue();
+  void boundImuQueue();
   std::optional<geometry_msgs::msg::TransformStamped> lookupTransformToBase(
     const std::string & source_frame);
   std::unique_ptr<cuda_blackboard::CudaPointCloud2> processPointcloud(
@@ -119,6 +139,8 @@ private:
   bool use_imu_;
   double processing_time_threshold_sec_;
   double timestamp_mismatch_fraction_threshold_;
+  InputBoundsParams input_bounds_params_{};
+  InputBoundsStatus latest_input_bounds_status_{};
 
   std::deque<geometry_msgs::msg::TwistWithCovarianceStamped> twist_queue_;
   std::deque<geometry_msgs::msg::Vector3Stamped> angular_velocity_queue_;
--- a/sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor_node.cpp
+++ b/sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor_node.cpp
@@ -146,6 +146,7 @@ CudaPointcloudPreprocessorNode::CudaPointcloudPreprocessorNode(
     queue_size_from_frequency(max_twist_frequency_hz);
   input_bounds_params_.max_imu_subscriber_queue_size =
     queue_size_from_frequency(max_imu_frequency_hz);
+  // TODO(mojomex): equates to 1s of messages, should be made tighter in the future
   input_bounds_params_.max_twist_queue_size =
     static_cast<std::size_t>(std::ceil(max_twist_frequency_hz));
   input_bounds_params_.max_imu_queue_size =
--- a/sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor_node.cpp
+++ b/sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor_node.cpp
@@ -17,6 +17,7 @@
 #include "autoware/cuda_pointcloud_preprocessor/memory.hpp"
 #include "autoware/cuda_pointcloud_preprocessor/point_types.hpp"
 #include "autoware/pointcloud_preprocessor/diagnostics/crop_box_diagnostics.hpp"
+#include "autoware/pointcloud_preprocessor/diagnostics/diagnostics_base.hpp"
 #include "autoware/pointcloud_preprocessor/diagnostics/distortion_corrector_diagnostics.hpp"
 #include "autoware/pointcloud_preprocessor/diagnostics/latency_diagnostics.hpp"
 #include "autoware/pointcloud_preprocessor/diagnostics/pass_rate_diagnostics.hpp"
@@ -34,7 +35,9 @@
 #include <pcl/point_cloud.h>
 #include <pcl/point_types.h>
 
+#include <algorithm>
 #include <chrono>
+#include <cmath>
 #include <deque>
 #include <memory>
 #include <string>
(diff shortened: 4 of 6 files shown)


sensing/autoware_cuda_pointcloud_preprocessor/config/cuda_pointcloud_preprocessor.param.yaml
```diff
@@ -8,6 +8,10 @@
     processing_time_threshold_sec: 0.01
     timestamp_mismatch_fraction_threshold: 0.01
     enable_ring_outlier_filter: true
+    max_input_point_count: 300000
```
mojomex: Beefiest sensor currently in use with Autoware is Hesai OT128 in dual return + high resolution mode.

It has 691200 points per cloud, so let's set this value here and add a comment as to where the number comes from.

sensing/autoware_cuda_pointcloud_preprocessor/config/cuda_pointcloud_preprocessor.param.yaml
```diff
@@ -8,6 +8,10 @@
...
+    max_twist_frequency_hz: 1000.0
```
mojomex: The max of any project I could identify is 200.0

sensing/autoware_cuda_pointcloud_preprocessor/config/cuda_pointcloud_preprocessor.param.yaml
```diff
@@ -8,6 +8,10 @@
...
+    max_imu_frequency_hz: 1000.0
```
mojomex: The max of any project I could identify is 200.0

sensing/autoware_cuda_pointcloud_preprocessor/include/autoware/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor.hpp
```diff
@@ -80,8 +82,9 @@ class CudaPointcloudPreprocessor
 
   int num_rings_{};
   int max_points_per_ring_{};
-  size_t num_raw_points_{};
-  size_t num_organized_points_{};
+  std::size_t num_raw_points_{};
+  std::size_t num_organized_points_{};
+  std::size_t max_input_point_count_{300000};
```
mojomex: I'd like to avoid duplicating this number across places. ROS2 param is required, so we need to guarantee is set in constructor.

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor_node.cpp
```diff
@@ -62,6 +135,18 @@ CudaPointcloudPreprocessorNode::CudaPointcloudPreprocessorNode(
...
   use_imu_ = declare_parameter<bool>("use_imu");
   bool enable_ring_outlier_filter = declare_parameter<bool>("enable_ring_outlier_filter");
+  max_input_point_count_ =
+    static_cast<std::size_t>(declare_parameter<int64_t>("max_input_point_count", 300000));
+  const auto twist_imu_queue_window_sec =
+    declare_parameter<double>("twist_imu_queue_window_sec", 1.0);
+  max_twist_queue_size_ = queue_size_from_frequency(
+    declare_parameter<double>("max_twist_frequency_hz", 1000.0), twist_imu_queue_window_sec);
+  max_imu_queue_size_ = queue_size_from_frequency(
+    declare_parameter<double>("max_imu_frequency_hz", 1000.0), twist_imu_queue_window_sec);
```
mojomex: Autoware policy is to not have defaults here

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor_node.cpp
```diff
@@ -133,25 +218,22 @@ CudaPointcloudPreprocessorNode::CudaPointcloudPreprocessorNode(
...
     geometry_msgs::msg::TwistWithCovarianceStamped, autoware_utils::polling_policy::All>::
-    create_subscription(this, "~/input/twist", rclcpp::QoS(TWIST_QUEUE_SIZE));
+    create_subscription(this, "~/input/twist", rclcpp::QoS(max_twist_queue_size_));
 
   if (use_imu_) {
     imu_sub_ = autoware_utils::InterProcessPollingSubscriber<
       sensor_msgs::msg::Imu, autoware_utils::polling_policy::All>::
-      create_subscription(this, "~/input/imu", rclcpp::QoS(TWIST_QUEUE_SIZE));
+      create_subscription(this, "~/input/imu", rclcpp::QoS(max_imu_queue_size_));
   }
```
mojomex, line 239: This basically assumes that for every LiDAR message, `max_imu_queue_size_` IMU messages come (same for twist).

Our stack is currently 10Hz, so assuming 10Hz pointclouds with a `TODO(mojomex)` seems reasonalbe for now

sensing/autoware_cuda_pointcloud_preprocessor/include/autoware/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor_node.hpp
```diff
@@ -119,6 +122,13 @@ class CudaPointcloudPreprocessorNode : public rclcpp::Node
   bool use_imu_;
   double processing_time_threshold_sec_;
   double timestamp_mismatch_fraction_threshold_;
+  std::size_t max_input_point_count_{};
+  std::size_t max_twist_queue_size_{};
+  std::size_t max_imu_queue_size_{};
+  std::size_t latest_input_point_count_{};
+  std::size_t latest_truncated_point_count_{};
+  std::size_t latest_dropped_twist_count_{};
+  std::size_t latest_dropped_imu_count_{};
```
mojomex: I'd like to keep class members to a minimum, and structured, not flattened.

sensing/autoware_cuda_pointcloud_preprocessor/config/cuda_pointcloud_preprocessor.param.yaml
mojomex: cc: @KYabuuchi

mojomex:
Addressed the review comments in the updated branch:

- Set `max_input_point_count` to `691200` and documented the Hesai OT128 dual-return high-resolution basis in the default config.
- Changed twist/IMU frequency defaults to `200.0` Hz.
- Removed hardcoded constructor fallback capacity and made the ROS parameters required via `declare_parameter` without defaults.
- Reworked queue sizing to assume 10 Hz pointcloud input for now, with a `TODO(mojomex)` to derive this later.
- Re-grouped the node state into structured input-bound params/status instead of extra flattened members.

Validation: focused build passed for `autoware_cuda_pointcloud_preprocessor` and its dependencies with `BUILD_TESTING=OFF`. No benchmarks were run.

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor_node.cpp
```diff
@@ -44,6 +47,79 @@
 namespace autoware::cuda_pointcloud_preprocessor
 {
 using sensor_msgs::msg::PointCloud2;
+namespace
+{
+
+std::size_t queue_size_from_frequency(const double max_frequency_hz)
+{
+  // TODO(mojomex): derive this from the configured pointcloud rate instead of assuming 10 Hz.
```
mojomex:
```suggestion
  // TODO(mojomex): derive from a future pointcloud_frequency param instead of assuming 10 Hz.
```

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor_node.cpp
```diff
@@ -348,17 +448,23 @@ void CudaPointcloudPreprocessorNode::updateTwistQueue(std::uint64_t first_point_
     twist_queue_.begin(), twist_queue_.end(), first_point_stamp,
     [](const auto & twist, const std::uint64_t stamp) {
       // To prevent potential precision loss, use nanoseconds
-      return rclcpp::Time(twist.header.stamp).nanoseconds() < stamp;
+      return static_cast<std::uint64_t>(rclcpp::Time(twist.header.stamp).nanoseconds()) < stamp;
     });
 
   twist_queue_.erase(twist_queue_.begin(), it);
+  boundTwistQueue();
```
mojomex: This seems a bit convoluted to me.

Our invariants should be that:
* `twist_queue_` NEVER (even temporarily) exceeds `max_twist_queue_size` as that would incur a dynamic allocation
* sorting (oldest first, newest last) shall always be preserved

Also, `twist_msgs.size()` should not be directly compared with `max_twist_queue_size` as that number refers to the capacity of `twist_queue_`, not of the input.

We should define "dropped" as "message younger than the `first_point_stamp` has to be removed due to queue being at capacity`.
So we can first remove all points older than `first_point_stamp`, then compute free capacity vs. input size, and compute dropped from that.

Ideally we'd rewrite this function with these in mind, the code should become nicer.

mojomex: Same for IMU queue.

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor_node.cpp
```diff
@@ -62,6 +138,16 @@ CudaPointcloudPreprocessorNode::CudaPointcloudPreprocessorNode(
   use_3d_undistortion_ = declare_parameter<bool>("use_3d_distortion_correction");
   use_imu_ = declare_parameter<bool>("use_imu");
   bool enable_ring_outlier_filter = declare_parameter<bool>("enable_ring_outlier_filter");
+  input_bounds_params_.max_input_point_count =
+    static_cast<std::size_t>(declare_parameter<int64_t>("max_input_point_count"));
+  input_bounds_params_.max_twist_queue_size =
+    queue_size_from_frequency(declare_parameter<double>("max_twist_frequency_hz"));
+  input_bounds_params_.max_imu_queue_size =
+    queue_size_from_frequency(declare_parameter<double>("max_imu_frequency_hz"));
```
mojomex: We're conflating subscriber queue size and internal queue size.
Subscriber queue should be matched to the imu/twist and pointcloud frequency as is currently the case.
It has to have capacity for all incoming IMU/twist msgs in one pointcloud period.

The internal queue has to be bigger in some cases, as the point ages within one pointcloud might span more than one period (100ms): For sensors with per-channel angle correction (most 360deg rotational LiDARs we know of), the scan boundary is not a straight, but a jagged line, resulting in some points from the same timestamp have angles that put them in scan N, while others go into scan N-1.
See e.g. https://tier4.github.io/nebula/design/scan_cutting/

mojomex: So, the current calculation is good for the susbcriber queue, but not for the internal one. Let's put the internal one to a fixed 1s again, as it was previously.

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor_node.cpp
```diff
@@ -62,6 +138,22 @@ CudaPointcloudPreprocessorNode::CudaPointcloudPreprocessorNode(
...
+  const auto max_twist_frequency_hz = declare_parameter<double>("max_twist_frequency_hz");
+  const auto max_imu_frequency_hz = declare_parameter<double>("max_imu_frequency_hz");
+  input_bounds_params_.max_twist_subscriber_queue_size =
+    queue_size_from_frequency(max_twist_frequency_hz);
+  input_bounds_params_.max_imu_subscriber_queue_size =
+    queue_size_from_frequency(max_imu_frequency_hz);
+  input_bounds_params_.max_twist_queue_size =
+    static_cast<std::size_t>(std::ceil(max_twist_frequency_hz));
+  input_bounds_params_.max_imu_queue_size =
+    static_cast<std::size_t>(std::ceil(max_imu_frequency_hz));
```
mojomex, line 153:
```suggestion
  // TODO(mojomex): equates to 1s of messages, should be made tighter in the future
  input_bounds_params_.max_twist_queue_size =
    static_cast<std::size_t>(std::ceil(max_twist_frequency_hz));
  input_bounds_params_.max_imu_queue_size =
    static_cast<std::size_t>(std::ceil(max_imu_frequency_hz));
```

vividf (approved):
LGTM!

PR #13107, opened 2026-07-30
feat(autoware_trajectory_processor): apply agnocast_wrapper::Node to trajectory_optimizer

Koichi98:
## Description

Migrates `trajectory_optimizer_node` to `autoware::agnocast_wrapper::Node`.

Because the optimizer plugins are loaded through pluginlib, the virtual boundary `TrajectoryOptimizerPluginBase::initialize()` cannot be templated on the node type, so it now takes `autoware::agnocast_wrapper::Node *`. The two node-dependent components used by those plugins are made node-type independent instead of being duplicated:

- `autoware_path_smoother`: `EBPathSmoother` is split into a node-independent `EBParam` (declared by `declare_eb_param<NodeT>()`) and `BasicEBPathSmoother<NodeT>`, whose only node-dependent member is an injected `autoware_utils_debug::BasicDebugPublisher<NodeT>`. `using EBPathSmoother = BasicEBPathSmoother<rclcpp::Node>` keeps existing callers unchanged.
- `autoware_path_optimizer`: same treatment for `MPTOptimizer` (`declare_mpt_param<NodeT>()` + `BasicMPTOptimizer<NodeT>`, `using MPTOptimizer = BasicMPTOptimizer<rclcpp::Node>`).

Both are explicitly instantiated for `rclcpp::Node` and `autoware::agnocast_wrapper::Node`, so the `elastic_band_smoother` / `path_optimizer` nodes keep running on plain `rclcpp::Node` while the optimizer plugins use the wrapper.

Other changes in `autoware_trajectory_processor`:

- publishers/subscriptions of `trajectory_optimizer` and of the plugins that own one (velocity / mpt / temporal_mpt) go through the wrapper.
- polling subscribers use `autoware::agnocast_wrapper::polling::create_polling_subscriber()`.
- `smooth_trajectory_with_elastic_band()` is templated on the smoother type and moved to the header.
- `trajectory_optimizer.launch.xml` includes `agnocast_env.launch.xml` and sets `LD_PRELOAD` (no-op when Agnocast is disabled).

The trajectory modifier side of the package is untouched.

## Related links

None.

## How was this PR tested?

- Built with `ENABLE_AGNOCAST=0` and `ENABLE_AGNOCAST=1` (`autoware_path_smoother`, `autoware_path_optimizer`, `autoware_trajectory_processor`).
- `colcon test` on the three packages with `ENABLE_AGNOCAST=0`: 477 tests, 0 failures.
- Runtime verification with the planning simulator is not done yet; this PR stays draft until then.

## Notes for reviewers

- `TrajectoryOptimizerPluginBase::initialize()` changes its node parameter type, so out-of-tree optimizer plugins need the same change. There are no in-tree users other than `autoware_trajectory_processor`.
- `EBPathSmoother` / `MPTOptimizer` constructors now take parameters and a debug publisher instead of a node pointer. Both aliases keep the old type name, and there are no in-tree users outside these packages.
- The debug publishers are created lazily on first publish (`BasicDebugPublisher`), so `~/debug/eb_traj`, `~/debug/eb_fixed_traj` and the MPT debug topics only appear once they are published. Topic names and QoS are unchanged.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/planning/autoware_path_optimizer/CMakeLists.txt
+++ b/planning/autoware_path_optimizer/CMakeLists.txt
@@ -69,6 +69,15 @@ ament_auto_add_library(${PROJECT_NAME} SHARED
 target_include_directories(${PROJECT_NAME} SYSTEM PUBLIC ${EIGEN3_INCLUDE_DIR})
 target_link_libraries(${PROJECT_NAME} acados_interface)
 
+# The library instantiates the node-type templates for autoware::agnocast_wrapper::Node, so it
+# needs the agnocast headers when Agnocast is enabled and must agree with its consumers on
+# USE_AGNOCAST_ENABLED.
+if(DEFINED ENV{ENABLE_AGNOCAST} AND "$ENV{ENABLE_AGNOCAST}" STREQUAL "1")
+  find_package(agnocastlib REQUIRED)
+  ament_target_dependencies(${PROJECT_NAME} agnocastlib)
+endif()
+autoware_agnocast_wrapper_setup(${PROJECT_NAME})
+
 rclcpp_components_register_node(${PROJECT_NAME}
   PLUGIN "autoware::path_optimizer::PathOptimizer"
   EXECUTABLE path_optimizer_node
--- a/planning/autoware_path_optimizer/include/autoware/path_optimizer/mpt_optimizer.hpp
+++ b/planning/autoware_path_optimizer/include/autoware/path_optimizer/mpt_optimizer.hpp
@@ -30,6 +30,7 @@
 
 #include <Eigen/Core>
 #include <Eigen/Sparse>
+#include <autoware_utils_debug/debug_publisher.hpp>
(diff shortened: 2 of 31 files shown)


kobayu858:
The `TrajectoryOptimizer` class targeted by this PR was removed by #13165 (unifying `trajectory_optimizer_node` and `trajectory_modifier` into `trajectory_processor`) on 2026-08-14.
Since a rebase cannot follow that change, the work was rebuilt as #13243 (agnocast support in `autoware_path_smoother` / `autoware_path_optimizer`, merged 8/24) and #13242 (agnocast support in `autoware_trajectory_processor`, merged 8/31).
Both are now merged, so `agnocast_wrapper::Node` is already applied to this area and this PR is no longer needed.
Closing it in favor of those two. Thank you!

PR #13119, opened 2026-07-31
ci: remove stale debian packages in build-and-test

ktro2828:
## Description

As reported in https://github.com/autowarefoundation/autoware_universe/pull/13114, `build-and-test` and `build-and-test-differential` also fails.
This pull request adds a task to remove the matching Debian package for every package that is about to be built from source.

## Related links

- https://github.com/autowarefoundation/autoware_core/pull/1311
- https://github.com/autowarefoundation/autoware_core/pull/1311

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/.github/workflows/build-and-test-differential.yaml
+++ b/.github/workflows/build-and-test-differential.yaml
@@ -108,6 +108,29 @@ jobs:
         run: cp build_depends_stable.repos build_depends.repos
         shell: bash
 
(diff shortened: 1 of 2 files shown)


interimadd (approved):

youtalk (approved):
Good catch.

PR #13120, opened 2026-07-31
refactor(autoware_mission_planner_universe): decouple planner plugin from the node type

Koichi98:
## Description

The plugin is loaded through pluginlib, so a node type in the virtual boundary is part of the plugin ABI. This replaces the node pointer with a node-free `Context`:

- `initialize(const Context &)` and `set_map(const LaneletMapBin &)` are the new virtual entry points. `Context` carries the parameters interface, logger, clock, a debug-marker publish callback and the vehicle info.
- `make_context(node)` builds one from any node that provides `get_node_parameters_interface()`, `get_logger()`, `get_clock()` and `create_publisher()`. It is a non-virtual template instantiated in the caller's translation unit, so the node type stays out of the plugin.
- `initialize(node, map)` is kept so existing callers compile unchanged. Once they move to `make_context()` + `set_map()`, that overload can be removed and `make_context()` moved out of this header, leaving the interface free of node-shaped code altogether.
- `DefaultPlanner` no longer subscribes to `~/input/vector_map`. `MissionPlanner` already subscribes to it and now forwards the map via `set_map()` from `on_map()`, removing a duplicated subscription to the same topic in the same process.

This keeps one pluginlib base class whatever the host node type is, and is the prerequisite for making the plugin fully node-independent (a follow-up can replace the parameters interface with a parameter-source abstraction without touching any caller).

## Related links

**Parent Issue:**

## How was this PR tested?

- `colcon build` + `colcon test`: 82 tests, 0 failures (`DefaultPlannerTest` now initializes through `make_context()`).
- The out-of-repo caller of this interface (`autoware_static_centerline_generator` in `autoware_tools`) compiles unchanged against the new header.
- `make_context()` + `initialize()` + `set_map()` also compile with `autoware::agnocast_wrapper::Node` as the host node type.
- planning_simulator: set the ego pose, cleared the route and requested a new one via `/planning/set_waypoint_route`. The planner produced the same route as before this change (12 segments over lanelets 106..127), `/planning/mission_planning/route` was published and the route state reached `SET`.

## Notes for reviewers

`Context::publish_debug_marker` owns the `~/debug/goal_footprint` publisher (it is captured by the callback), so the publisher lives as long as the plugin keeps its context. Topic name and QoS are unchanged.

## Interface changes

- Removed: `initialize(rclcpp::Node *)`, the variant in which the plugin subscribed to `~/input/vector_map` itself. Its only user was `MissionPlanner`.
- Added: `initialize(const Context &)`, `set_map()`, `make_context()`.
- Unchanged: `initialize(node, map)`, the other virtual methods, the pluginlib base class name and the plugin lookup name.

## Effects on system behavior

None. The map now reaches the plugin through `MissionPlanner::on_map()` instead of the plugin's own subscription to the same topic.

### DIFF
--- a/planning/autoware_mission_planner_universe/include/autoware/mission_planner_universe/mission_planner_plugin.hpp
+++ b/planning/autoware_mission_planner_universe/include/autoware/mission_planner_universe/mission_planner_plugin.hpp
@@ -16,6 +16,7 @@
 #define AUTOWARE__MISSION_PLANNER_UNIVERSE__MISSION_PLANNER_PLUGIN_HPP_
 
 #include <autoware/route_handler/route_handler.hpp>
+#include <autoware_vehicle_info_utils/vehicle_info_utils.hpp>
 #include <rclcpp/rclcpp.hpp>
 
 #include <autoware_map_msgs/msg/lanelet_map_bin.hpp>
@@ -23,6 +24,7 @@
 #include <geometry_msgs/msg/pose_stamped.hpp>
 #include <visualization_msgs/msg/marker_array.hpp>
 
+#include <functional>
 #include <vector>
 
 namespace autoware::mission_planner_universe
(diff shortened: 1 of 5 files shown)


soblin (approved):

kobayu858:
New PR https://github.com/autowarefoundation/autoware_universe/pull/13283

kobayu858:
Closing this PR since it was addressed and merged in https://github.com/autowarefoundation/autoware_universe/pull/13283 instead.

PR #13127, opened 2026-08-02
fix(autoware_system_monitor): add missing dependency to edac-utils

mitsudome-r:
## Description
Dependency to edac-utils were added in https://github.com/autowarefoundation/autoware_universe/pull/3795. 
I have added edac-utils to rosdep so we should be able to install edac-utils from rosdep install now. 

## Related links

- https://github.com/autowarefoundation/autoware_universe/pull/3795
- https://github.com/ros/rosdistro/pull/51519

## How was this PR tested?

I ran rosdep install and made sure that edac-utils is installed. 

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/system/autoware_system_monitor/package.xml
+++ b/system/autoware_system_monitor/package.xml
@@ -28,6 +28,7 @@
   <depend>tier4_external_api_msgs</depend>
 
   <exec_depend>chrony</exec_depend>
+  <exec_depend>edac-utils</exec_depend>
   <exec_depend>sysstat</exec_depend>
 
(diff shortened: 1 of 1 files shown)


TetsuKawa:
Confirmed that this change enable to install edac-utils only by `rpsdep install` command.
[Screencast from 2026-08-03 13-49-44.webm](https://github.com/user-attachments/assets/f39f0ed7-aece-4ffe-ab88-e2f519952192)

TetsuKawa (approved):

PR #13129, opened 2026-08-02
feat(autoware_system_monitor): add argument to override hdd_mount_point

mitsudome-r:
## Description
The HddMonitor node monitors the disk mounted at disks.disk0.name, which defaults to / in hdd_monitor.param.yaml. When Autoware runs inside a container, however, the container's root filesystem is an overlayfs with no backing block device, so HddMonitor cannot resolve the underlying device and disk monitoring does not work.

This PR adds an hdd_mount_point launch argument to system_monitor.launch.xml that overrides disks.disk0.name for the HddMonitor composable node. It defaults to /, preserving the current behavior. In containerized deployments, it can be set to a bind mount of the host root (e.g. /host_root) so that the monitored mount point resolves to a real block device on the host.

This allows containerized deployments to point the disk monitor at the host disk without maintaining a full copy of hdd_monitor.param.yaml just to change the mount point.

## How was this PR tested?

* Launched system_monitor.launch.xml without the argument and confirmed that HddMonitor behaves as before (monitoring /).
* Launched it in a container with hdd_mount_point:=/host_root (host root bind-mounted) and confirmed that HddMonitor reports the host disk correctly.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


TetsuKawa (approved):
Apologies for not verifying the container startup behavior in the same test. However, I understand that this change won't affect the existing local execution.

PR #13131, opened 2026-08-03
feat(ptv3): reintroduce filtering points by probability

ktro2828:
## Description

This pull request reintroduces a configurable probability threshold for class-based filtering in the PTv3 pipeline, which has been removed in https://github.com/autowarefoundation/autoware_universe/pull/12821.
The main change is the addition of a `class_probability_threshold` parameter, which allows users to exclude points from the output point cloud if the predicted probability for any filtered class exceeds this threshold. The change is integrated across the configuration, schema, CUDA kernels, and test code to ensure consistent and robust filtering behavior.

## Related links

**Parent Issue:**

- Link

## Notes for reviewers

None.

## Interface changes

None.

#### Additions and removals

| Added | `filter.class_probability_threshold`   | `double` | `0.05`         | filter points if any filter class probability exceeds this value |

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_ptv3/config/ptv3.param.yaml
+++ b/perception/autoware_ptv3/config/ptv3.param.yaml
@@ -17,6 +17,7 @@
       workspace_size: 8589934592  # 8 GiB
       source_reconstruction: "full"
       filter:
+        class_probability_threshold: 0.05
         classes: ["drivable_flat", "non_drivable_flat", "noise"]
         output_format: ""
         apply_to_segmentation: false
--- a/perception/autoware_ptv3/include/autoware/ptv3/ptv3_config.hpp
+++ b/perception/autoware_ptv3/include/autoware/ptv3/ptv3_config.hpp
@@ -47,6 +47,7 @@ public:
     const std::vector<std::int64_t> & pooling_strides = {},
     const std::vector<std::int64_t> & enc_channels = {},
     const std::vector<std::int64_t> & palette = {},
+    const float filter_class_probability_threshold = {},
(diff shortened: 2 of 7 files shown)


amadeuszsz:
@codex review

amadeuszsz (changes requested):
Thank you for this PR.

I know you restored original behavior, but in my opinion old approach was bad. Segmentation and filter result with inconsistent output. For the instance, given threshold 0.1, filtered class [`drivable_flat`, `noise`] and class probabilities `vegetation=0.5`, `drivable_flat=0.2`, `noise=0.3`, filtered topic will remove this point, but segmentation topic will show it as `vegetation`. Not sure if this is what we expect.

Instead, I would look on this threshold as a configuration for maximizing class recall (and this is how we can document it in schema file). Consider same example, but now our threshold argument match the length of filtered classes, let's say [0.1, 0.25]. Then our point is classified to the best class from our list considering their diff:
- `drivable_flat` score = 0.2 - 0.1 = 0.1
- `noise` score = 0.3 - 0.25 = 0.05
Point is assigned to `drivable_flat` as got top score and its score is above 0.0.

I would remove `segmentation3d.filter.class_probability_threshold` and keep old behavior for filtering out the listed classes. Next, we add `segmentation3d.priority_classes` and `segmentation3d.priority_classes_thresholds` to replicate example behavior (these names do not sound good 😆). This happens before filtering, so filter will already see recall-maxed predictions.

This is my proposal, please feel free to share your thoughts, maybe I missed a more principled solution 🙏🏻

PR #13134, opened 2026-08-04
perf(cuda_pointcloud_preprocessor): allocate fixed GPU buffers at startup

mojomex:
## Description

Use the configured input bounds from the previous PR to allocate the CUDA pointcloud preprocessor working buffers during startup.

This PR:

- introduces `PreprocessorCapacity`
- adds `max_ring_count` and `max_points_per_ring` parameters
- allocates input, organized, transformed, mask, index, sort workspace, and twist struct buffers during construction
- removes dynamic resizing from the pointcloud callback path
- reports ring/per-ring overflow through diagnostics instead of reallocating
- keeps the existing cuda blackboard output ownership/allocation path unchanged

## Stack

This is PR 2 of 5 in a stack. Every PR targets `autowarefoundation:main` (GitHub does not allow cross-repository stacks, and the branches live on a fork), so each **Files changed** tab shows the *cumulative* diff of the stack up to that PR.

**To review only one PR's own changes, use its Own diff link below.**

| # | PR | Own diff |
| --- | --- | --- |
| 1 | #13106 feat: add input bounds diagnostics | [Files changed](https://github.com/autowarefoundation/autoware_universe/pull/13106/files) (bottom of stack) |
| 2 | #13134 perf: allocate fixed GPU buffers at startup **← this PR** | [preproc-bounds-diagnostics → preproc-startup-capacity](https://github.com/mojomex/autoware_universe/compare/preproc-bounds-diagnostics...preproc-startup-capacity) |
| 3 | #13135 perf: avoid runtime thrust allocations | [preproc-startup-capacity → preproc-remove-runtime-thrust-allocs](https://github.com/mojomex/autoware_universe/compare/preproc-startup-capacity...preproc-remove-runtime-thrust-allocs) |
| 4 | #13136 refactor: clarify queue update flow | [preproc-remove-runtime-thrust-allocs → preproc-queue-update-cleanup](https://github.com/mojomex/autoware_universe/compare/preproc-remove-runtime-thrust-allocs...preproc-queue-update-cleanup) |
| 5 | #13137 test: cover capacity bounds behavior | [preproc-queue-update-cleanup → preproc-behavior-tests](https://github.com/mojomex/autoware_universe/compare/preproc-queue-update-cleanup...preproc-behavior-tests) |

Merge in stack order (1 → 5); each PR depends on the one below it. The whole stack is meant as one unit: tests are added at the top to keep the individual PRs small, so please review in stack order but test with the full stack applied (`preproc-behavior-tests`).

[...]

PR #13135, opened 2026-08-04
perf(cuda_pointcloud_preprocessor): avoid runtime thrust allocations

mojomex:
## Description

Replace the remaining per-frame Thrust algorithm paths in the CUDA pointcloud preprocessor with explicit kernels and CUB calls that run on storage allocated once at startup, so steady-state processing no longer triggers device allocations.

This PR:

- replaces the Thrust fill paths with a fill kernel launched on the processing stream
- replaces Thrust `inclusive_scan` with CUB `DeviceScan::InclusiveSum` using startup-allocated temporary storage
- replaces the Thrust diagnostic `count` calls with CUB `DeviceReduce::Sum` over a transform iterator, accumulating the three counters into one device buffer that is read back in a single transfer
- sizes a single scratch workspace at startup to the largest requirement among the sort, the scan and the reductions
- keeps the accepted cuda blackboard output allocation behavior unchanged

## Stack

This is PR 3 of 5 in a stack. PRs 1 and 2 are merged, so this PR is now rebased directly onto `autowarefoundation:main` and its **Files changed** tab shows only its own changes.

| # | PR | Status |
| --- | --- | --- |
| 1 | #13106 feat: add input bounds diagnostics | merged |
| 2 | #13134 perf: allocate fixed GPU buffers at startup | merged |
| 3 | #13135 perf: avoid runtime thrust allocations | **this PR** |
| 4 | #13136 refactor: clarify queue update flow | draft, will be rebased onto this |
| 5 | #13137 test: cover capacity bounds behavior | draft, will be rebased onto this |

The stack is still meant as one unit: the behavior tests are added at the top (#13137) to keep the individual PRs small, so please review in stack order but test with the full stack applied.

## Validation

Rebased onto `main` and validated with:

```bash
colcon build --packages-up-to autoware_cuda_pointcloud_preprocessor --cmake-args -DCMAKE_BUILD_TYPE=Release
colcon test --packages-select autoware_cuda_pointcloud_preprocessor
```

The build passes with no new warnings in the touched files (the remaining `-Wreturn-type` warnings come from the polar voxel filter headers and are pre-existing). `colcon test` reports 102 tests, 0 failures, 30 skipped — lint only, since the behavior tests land in #13137.

### DIFF
--- a/sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor.cu
+++ b/sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor.cu
@@ -36,7 +36,9 @@
 #include <thrust/reduce.h>
 #include <thrust/scan.h>
 
+#include <algorithm>
 #include <cstdint>
+#include <stdexcept>
 
 namespace autoware::cuda_pointcloud_preprocessor
 {
@@ -140,6 +142,14 @@ void CudaPointcloudPreprocessor::setUndistortionType(const UndistortionType & un
   }
 
   undistortion_type_ = undistortion_type;
+}
+
+void CudaPointcloudPreprocessor::setMaxInputPointCount(const std::size_t max_input_point_count)
+{
+  if (max_input_point_count == 0) {
+    throw std::runtime_error("max_input_point_count must be positive");
+  }
+  max_input_point_count_ = max_input_point_count;
 }
 
 void CudaPointcloudPreprocessor::preallocateOutput()
@@ -281,8 +291,10 @@ std::unique_ptr<cuda_blackboard::CudaPointCloud2> CudaPointcloudPreprocessor::pr
   const std::uint32_t first_point_rel_stamp_nsec)
 {
   auto frame_id = input_pointcloud_msg.header.frame_id;
-  num_raw_points_ = static_cast<int>(input_pointcloud_msg.width * input_pointcloud_msg.height);
-  num_organized_points_ = num_rings_ * max_points_per_ring_;
+  const auto input_point_count =
+    static_cast<std::size_t>(input_pointcloud_msg.width) * input_pointcloud_msg.height;
+  num_raw_points_ = std::min(input_point_count, max_input_point_count_);
+  num_organized_points_ = static_cast<std::size_t>(num_rings_) * max_points_per_ring_;
 
   if (num_raw_points_ == 0) {
     output_pointcloud_ptr_->row_step = 0;
--- a/sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor.cu
+++ b/sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor.cu
@@ -45,8 +45,15 @@ namespace autoware::cuda_pointcloud_preprocessor
 
 namespace thrust_stream = cuda_utils::thrust_stream;
 
-CudaPointcloudPreprocessor::CudaPointcloudPreprocessor() : stream_(initialize_stream())
+CudaPointcloudPreprocessor::CudaPointcloudPreprocessor(const PreprocessorCapacity & capacity)
+: stream_(initialize_stream()), capacity_(capacity)
 {
+  if (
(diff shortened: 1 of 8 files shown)


sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor.cu
```diff
@@ -147,6 +206,7 @@ void CudaPointcloudPreprocessor::initializeBuffers()
   device_indices_.resize(num_organized_points_);
   device_twist_2d_structs_.resize(capacity_.max_twist_struct_count);
   device_twist_3d_structs_.resize(capacity_.max_twist_struct_count);
+  device_processing_stats_.resize(3);
```
vividf: magic number 3

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor.cu
```diff
@@ -250,8 +337,10 @@ std::unique_ptr<cuda_blackboard::CudaPointCloud2> CudaPointcloudPreprocessor::pr
   }
 
   // Reset all contents in the device vector
-  thrust_stream::fill(device_input_points_, InputPointType{}, stream_);
-  thrust_stream::fill(device_organized_points_, InputPointType{}, stream_);
+  fillDeviceVector(
```
vividf: nit: this reset looks redundant — the memcpy below overwrites [0, num_raw_points_) and organizeKernel never reads past num_raw_points_, so the 22 MB fill is never observed. (device_organized_points_ still needs its reset for the empty slots.) Fine to defer to a follow-up.

vividf (approved):
LGTM

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor.cu
```diff
@@ -392,18 +494,38 @@ std::unique_ptr<cuda_blackboard::CudaPointCloud2> CudaPointcloudPreprocessor::pr
     &max_points_per_ring_value, thrust::raw_pointer_cast(device_max_points_per_ring_.data()),
     sizeof(std::int32_t), cudaMemcpyDeviceToHost, stream_));
 
+  constexpr std::size_t crop_box_passed_stat_index = 0U;
+  constexpr std::size_t nan_stat_index = 1U;
+  constexpr std::size_t mismatch_stat_index = 2U;
+  constexpr std::size_t processing_stat_count = 3U;
```
mojomex: Make this a class-wide constant and re-use for `device_processing_stats_` initialization.

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor.cu
mojomex: Done — the layout now lives in class-scope constants (`crop_box_passed_stat_index` … `processing_stat_count`) in the header, so `initializeBuffers()` sizes the buffer with `processing_stat_count` and `process()` reuses the same names for the readback.

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor.cu
mojomex: Thanks, it partially is. As we previously had nasty errors with dirty buffers, I think this was introduced as a way to reduce the blast radius of such bugs.

In the future, we should remake the whole node with clearer buffer management (e.g. pass gsl::span<>-like objects to kernels) where it is impossible for a kernel to know of dirty locations.

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor.cu
mojomex: Done — the indices and the count are now `static constexpr` members next to `device_processing_stats_` in the header, and `initializeBuffers()` uses `processing_stat_count` for the resize.

PR #13136, opened 2026-08-04
refactor(cuda_pointcloud_preprocessor): clarify queue update flow

mojomex:
## Description

The twist and IMU queue bookkeeping was spread across the insertion callbacks and a pair of bounding helpers, so capacity was enforced *after* insertion and each queue carried its own copy of the pruning rules. Move it into `detail::` helpers in a new `queue_bounds.hpp` and give each queue a single update path.

This PR:

- adds `detail::prepare_queue_update()`, which prunes queue entries older than the current pointcloud, drops and counts incoming messages beyond the free capacity, and sorts the remainder — so capacity is enforced before insertion
- renames `twistCallback()`/`imuCallback()` to `insertTwistMessage()`/`insertImuMessage()`, which now only insert in stamp order
- removes `boundTwistQueue()`/`boundImuQueue()`, since nothing is inserted beyond capacity any more
- moves backward time jump detection into `detail::is_backward_time_jump()`

### Behavior change worth a look

Backward time jumps are now tolerated up to one second: the queue is cleared only when its oldest entry is more than a second ahead of the incoming stamp, rather than on any backward step. The old per-entry pop for queues spanning more than a second is gone, since `prepare_queue_update()` already prunes everything older than the current pointcloud. The tests in #13137 pin both rules down.

## Stack

This is PR 4 of 5 in a stack; PRs 1 and 2 are merged.

| # | PR | Status |
| --- | --- | --- |
| 1 | #13106 feat: add input bounds diagnostics | merged |
| 2 | #13134 perf: allocate fixed GPU buffers at startup | merged |
| 3 | #13135 perf: avoid runtime thrust allocations | open |
| 4 | #13136 refactor: clarify queue update flow | **this PR** |
| 5 | #13137 test: cover capacity bounds behavior | open |

The branches live on a fork, so this PR targets `autowarefoundation:main` and its **Files changed** tab shows #13135's changes as well. To see only this PR's own changes, use [preproc-remove-runtime-thrust-allocs → preproc-queue-update-cleanup](https://github.com/mojomex/autoware_universe/compare/preproc-remove-runtime-thrust-allocs...preproc-queue-update-cleanup). Merge in stack order (3 → 5).

## Validation

```bash
colcon build --packages-up-to autoware_cuda_pointcloud_preprocessor --cmake-args -DCMAKE_BUILD_TYPE=Release
colcon test --packages-select autoware_cuda_pointcloud_preprocessor
```

[...]

PR #13137, opened 2026-08-04
test(cuda_pointcloud_preprocessor): cover capacity bounds behavior

mojomex:
## Description

Add a gtest suite covering the behavior introduced by the rest of this stack. The tests are collected at the top of the stack to keep the individual PRs small.

Covered:

- `CudaPointcloudPreprocessor` accepts positive capacities and rejects non-positive ones
- points exactly at the ring and points-per-ring capacities are processed without reporting an overflow, and exceeding either reports one
- input clouds longer than `max_input_point_count` are truncated
- `detail::prepare_queue_update()` prunes, sorts, bounds and counts drops for both the twist and IMU queues, and rejects an already over-capacity queue
- `detail::is_backward_time_jump()` honors the one second threshold

### Running without a GPU

The cases that construct a `CudaPointcloudPreprocessor` allocate device memory, so they derive from `autoware::cuda_utils::CudaTest` and self-skip when no CUDA device is present — the CI containers are started without GPU access. The capacity validation and queue bound cases are pure host code and always run. Checked both ways locally: 13/13 pass with a device, and 8 pass / 5 skip / 0 failures under `CUDA_VISIBLE_DEVICES=""`.

## Stack

This is PR 5 of 5 in a stack; PRs 1 and 2 are merged.

| # | PR | Status |
| --- | --- | --- |
| 1 | #13106 feat: add input bounds diagnostics | merged |
| 2 | #13134 perf: allocate fixed GPU buffers at startup | merged |
| 3 | #13135 perf: avoid runtime thrust allocations | merged |
| 4 | #13136 refactor: clarify queue update flow | merged |
| 5 | #13137 test: cover capacity bounds behavior | **this PR** |

The branches live on a fork, so this PR targets `autowarefoundation:main` and its **Files changed** tab shows #13135's and #13136's changes as well. To see only this PR's own changes, use [preproc-queue-update-cleanup → preproc-behavior-tests](https://github.com/mojomex/autoware_universe/compare/preproc-queue-update-cleanup...preproc-behavior-tests). Merge in stack order (3 → 5).

This branch is the one to test with, since it carries the whole stack.

## Validation

```bash
colcon build --packages-up-to autoware_cuda_pointcloud_preprocessor --cmake-args -DCMAKE_BUILD_TYPE=Release
colcon test --packages-select autoware_cuda_pointcloud_preprocessor
```

Build passes with no warnings in the touched files; `colcon test` reports 122 tests, 0 errors, 0 failures.


sensing/autoware_cuda_pointcloud_preprocessor/CMakeLists.txt
```diff
@@ -174,6 +175,26 @@ install(DIRECTORY launch config
   DESTINATION share/${PROJECT_NAME}
 )
 
+if(BUILD_TESTING)
+  ament_add_gtest(test_cuda_pointcloud_preprocessor
+    test/test_cuda_pointcloud_preprocessor.cpp
+  )
+  target_link_options(test_cuda_pointcloud_preprocessor PRIVATE "LINKER:--no-as-needed")
```
mojomex: This is a build hack, please link necessary libs properly

sensing/autoware_cuda_pointcloud_preprocessor/test/test_cuda_pointcloud_preprocessor.cpp
```diff
@@ -0,0 +1,357 @@
...
+}
+
+TEST_F(CudaPointcloudPreprocessorDeviceTest, ReportsPointsPerRingOverflow)
+{
+  CudaPointcloudPreprocessor preprocessor{make_capacity(3, 1, 2)};
+  preprocessor.setRingOutlierFilterActive(false);
+  preprocessor.setUndistortionType(CudaPointcloudPreprocessor::UndistortionType::Undistortion2D);
+
+  const auto input_cloud =
+    make_input_cloud({make_point(1.0F, 0), make_point(2.0F, 1), make_point(3.0F, 2)});
```
manato: [ask] This code itself looks perfect. I found that the test still pass (i.e., ring overflow is detected) if we set the test data like the following even the points in a ring is equal to `max_points_per_ring` (=2).

``` c++
  const auto input_cloud =
    make_input_cloud({make_point(1.0F, 0), make_point(2.0F, 0), make_point(3.0F, 2)});
```

I feel like this situation would not trigger "ring overflow" detection because the handle-able number of points in a ring includes `max_points_per_ring` intuitively. Is this intended behavior? (This may be out of the scope of this PR. If it is the case, please skip to handle this comment)

manato (commented):
@mojomex
Thank you very much for the PR! (Finally, cuda_pointcloud_preprocessor has unit tests!). I left a comment but it might be out of scope of this PR. I'd appreciate your taking a look at it!

sensing/autoware_cuda_pointcloud_preprocessor/test/test_cuda_pointcloud_preprocessor.cpp
mojomex: Ah, this is a tricky one: `make_point(a, b)` has _timestamp_ `b`, not _ring_ `b`.  So in all examples in this thread, the rihgs of all points are `0`. You're right to find that confusing.

Additionally, `ring_overflow` can signify either an overflow in _points per ring_ or an overflow of _number of rings_. Both capacities are set via params. See: [stats_.ring_overflow = max_ring_value >= num_rings_ || max_points_per_ring_value >= max_points_per_ring_;](https://github.com/mojomex/autoware_universe/blob/a84dad50fc48545790a994deb4c74b6e033c6658/sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor.cu#L396). 

Will add tests for both cases explicitly.

sensing/autoware_cuda_pointcloud_preprocessor/CMakeLists.txt
mojomex: `cuda_pointcloud_preprocessor_lib` was missing cuda_blackboard and `concatenate_data` was missing `pointcloud_preprocessor_filter_base`. Both now link what they use, so the test only names `cuda_pointcloud_preprocessor_lib` and `--as-needed` stays at its default.

Addressed: 7771934a6

sensing/autoware_cuda_pointcloud_preprocessor/test/test_cuda_pointcloud_preprocessor.cpp
mojomex: Each limit now has its own at-capacity and over-capacity case, with the ring spelled out at every call site.

Addressed: 7771934a6

manato: Sorry, I completely misread the function :pray: . Thank you for making separate tests to check both ring index boundary and points per ring boundary, they gets easier to understand.

As you commented, the `ring_overflow` is calculated:

``` c++
stats_.ring_overflow = 
  max_ring_value >= num_rings_ || max_points_per_ring_value >= max_points_per_ring_;
```

, which means if 2 points are contained in a ring (i.e., `max_points_per_ring_value == 2`), I think this condition become `true` due to `>=` comparison with `max_points_per_ring_`. On the other hands, `TEST_F(CudaPointcloudPreprocessorDeviceTest, ProcessesPointsPerRingAtCapacityBoundary)` expects `EXPECT_FALSE(preprocessor.getProcessingStats().ring_overflow);`.
Would you mind checking if it is intended?

sensing/autoware_cuda_pointcloud_preprocessor/src/cuda_pointcloud_preprocessor/cuda_pointcloud_preprocessor.cu
```diff
@@ -515,8 +517,9 @@ std::unique_ptr<cuda_blackboard::CudaPointCloud2> CudaPointcloudPreprocessor::pr
     sizeof(processing_stats), cudaMemcpyDeviceToHost, stream_));
 
   CHECK_CUDA_ERROR(cudaStreamSynchronize(stream_));
-  stats_.ring_overflow =
-    max_ring_value >= num_rings_ || max_points_per_ring_value >= max_points_per_ring_;
+  // Both values are sentinels that `organizeKernel` only writes when a point does not fit, so any
+  // non-zero value means at least one point was dropped. They are not the maxima actually observed.
```
mojomex:
```suggestion
  // non-zero value means at least one point was dropped.
```

sensing/autoware_cuda_pointcloud_preprocessor/test/test_cuda_pointcloud_preprocessor.cpp
mojomex: @manato Apologies, the naming was misleading. `max_ring_value` is not the value of the highest _encountered_ ring, but that of the highest _out-of-bounds_ ring. So `max_ring_value = 0` for all pointclouds that have no offending rings, `max_ring_value >= num_rings_` for all pointclouds that have at least one ring more than configured.

I changed both to be flags (`1` if any offending ring was present, `0` if not). See 7f528ec

manato: @mojomex
Thank you very much for your confirmation. Now, your refactoring makes me clearer.
Thanks to your update, I notice that:

``` cuda
__global__ void organizeKernel(...)
{
  ...
  auto ring = input_points[idx].channel;  // <- this represents the channel INDEX

  if (ring >= initial_max_rings) {  // <- [No prob.] comparison between INDEX and the maximum NUMBER of the ring, which means overflow happens even for the case of ring == initial_max_rings
    atomicOr(output_rings_overflowed, 1);
    return;
  }

  int next_offset = atomicAdd(&ring_indexes[ring], 1);  // <- atomicAdd returns the value BEFORE addition

  if (next_offset >= initial_max_points_per_ring) { // <- [No prob.] next_offset == initial_max_points_per_ring means surely overflow is happening
    atomicOr(output_points_per_ring_overflowed, 1);
    return;
}

```

So, I've confirmed that all of what `organizeKernel` does, how `ring_overflow` is set, and how the flag is tested are valid!

manato (approved):
I appreciate handling all of my concerns! LGTM!

mojomex:
Oh, @manato is not a code owner of autoware_pointcloud_prerocessor :eyes: 
@vividf Could you have a look?

amadeuszsz (approved):

PR #13147, opened 2026-08-04
fix(autoware_pointcloud_preprocessor): fix azimuth_diff 1deg threshold math

### LINKED ISSUE #13146: RingOutlierFilterComponent `azimuth_diff` threshold is wrong
### Checklist

- [x] I've read the [contribution guidelines](https://github.com/autowarefoundation/autoware/blob/main/CONTRIBUTING.md).
- [x] I've searched other issues and no duplicate issues were found.
- [x] I'm convinced that this is not my fault but a bug.

### Description

```cpp
        azimuth_diff < 1.0 * (180.0 / M_PI)) {  // one degree
        continue;                               // Determined to be included in the same walk
```
should be

```cpp
        azimuth_diff < 1.0 * (M_PI / 180.0)) {  // one degree
        continue;                               // Determined to be included in the same walk
```

In the CUDA version, this is already correct:

```cpp
      azimuth_diff < 1.f * M_PI / 180.f) {
      // Determined to be included in the same walk
```

### Expected behavior

Azimuth diff threshold is `1 deg`

### Actual behavior

It's actually `3283 deg`

### Steps to reproduce

-

### Versions

_No response_

### Possible causes

_No response_

### Additional context

_No response_

mojomex:
## Description

Fixes #13146.

`azimuth_diff` is in radians, and a `1deg` threshold was being converted incorrectly, leading to a `3283deg` threshold instead.

This brings us closer to behavior parity with `autoware_cuda_pointcloud_preprocessor`, but not quite: CUDA ring outlier filter has an upper bound on its run length (11 points, hard-coded) while the CPU version does not. So, as seen below, point counts are still not 1:1 the same :smiling_face_with_tear: 

### Ran before (`/a/`) & after (`/b/`) fix side-by-side comparison 
Ran using [TIER IV INTERNAL](https://github.com/tier4/pa-tools). Reproduce: [TIER IV INTERNAL](https://drive.google.com/file/d/1cSFbxkT2VqFh2y5QWFmX2xx4ITXwyOC9/view?usp=sharing):

Screenshot from 2026-08-04 18-54-46

**~12k point difference.**

### Ran CUDA preprocessor (`/a/`) vs. fix (`/b/`) side-by-side comparison. 
Reproduce: [TIER IV INTERNAL](https://drive.google.com/file/d/1IZ_fe4sJGQjJXvTMDUX6kSXCvxjsW6Zd/view?usp=sharing):

Screenshot from 2026-08-04 19-11-03

**700 point difference.**

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

CPU ring outlier filter now produces correct (to-spec) results.

### DIFF
--- a/sensing/autoware_pointcloud_preprocessor/src/outlier_filter/ring_outlier_filter_node.cpp
+++ b/sensing/autoware_pointcloud_preprocessor/src/outlier_filter/ring_outlier_filter_node.cpp
@@ -154,7 +154,7 @@ void RingOutlierFilterComponent::faster_filter(
       if (
         std::max(current_distance, next_distance) <
           std::min(current_distance, next_distance) * distance_ratio_ &&
-        azimuth_diff < 1.0 * (180.0 / M_PI)) {  // one degree
+        azimuth_diff < 1.0 * (M_PI / 180.0)) {  // one degree
         continue;                               // Determined to be included in the same walk
       }
 


YoshiRi (approved):
LGTM!

badai-nguyen (approved):

PR #13148, opened 2026-08-04
feat(diffusion_planner): resolve model paths from base_model_directory

Owen-Liuyuxuan:
## Summary
- Replace per-file absolute ONNX/args paths with `model.base_model_directory` plus configurable filenames.
- Add `resolve_model_paths()` to join the base directory with filenames at load time.
- Expose `args_filename` and per-model `*_filename` parameters so different products can use names like `args.json` vs `diffusion_planner.param.json`, or `model.onnx` vs `diffusion_planner.onnx`.
- Update default YAML, JSON schema, and dynamic parameter handling to reload when the base directory or any filename changes.

## Test plan
- [ ] Not run locally — rely on CI
- [ ] Verify diffusion planner loads models when `model.base_model_directory` points to a valid bundle with default filenames
- [ ] Verify per-product filename overrides work without changing full paths

## Notes for reviewers
- `base_model_directory` is intended for dynamic per-vehicle configuration; filenames are typically fixed per product but remain overridable.
- Addresses review feedback to avoid hardcoding filenames in `resolve_model_paths()`.

## Related links
- Companion launch sync: https://github.com/autowarefoundation/autoware_launch/pull/1932

### DIFF
--- a/planning/autoware_diffusion_planner/src/diffusion_planner_core.cpp
+++ b/planning/autoware_diffusion_planner/src/diffusion_planner_core.cpp
@@ -34,6 +34,7 @@
 
 #include <algorithm>
 #include <cmath>
+#include <filesystem>
 #include <limits>
 #include <memory>
 #include <optional>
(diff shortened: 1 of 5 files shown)


planning/autoware_diffusion_planner/src/diffusion_planner_core.cpp
```diff
@@ -212,6 +213,16 @@ void DiffusionPlannerCore::update_params(const DiffusionPlannerParams & params)
...
 }
 
+void DiffusionPlannerCore::resolve_model_paths()
+{
+  const std::filesystem::path base_dir(params_.base_model_directory);
+  params_.single_step_model_path = (base_dir / "diffusion_planner.onnx").string();
+  params_.encoder_model_path = (base_dir / "diffusion_planner_encoder.onnx").string();
+  params_.decoder_model_path = (base_dir / "diffusion_planner_decoder.onnx").string();
+  params_.turn_indicator_model_path = (base_dir / "diffusion_planner_turn_indicator.onnx").string();
+  params_.args_path = (base_dir / "diffusion_planner.param.json").string();
```
Owen-Liuyuxuan: @SakodaShintaro Before making progress, I would like to check that this name is different from private deployments. Should we align this one?

Or we actually need a different parameter?

SakodaShintaro: We have to add different parameters to specify each file name.

Owen-Liuyuxuan:
- [ ] Run the sync-params workflow or run `python .github/scripts/sync_params.py planning` to update this file.

[Deleted] Companion launch-side param sync: https://github.com/autowarefoundation/autoware_launch/pull/1932

SakodaShintaro (approved):
I have conformed that psim normally works.
LGTM

SakodaShintaro:
https://github.com/autowarefoundation/autoware_launch/pull/1933

PR #13150, opened 2026-08-04
feat(autoware_carla_interface): handle gear commands

ujinf74:
## Description

The bridge never subscribed to `/control/command/gear_cmd` and reported `GearReport.DRIVE` unconditionally. Two consequences:

- The ego vehicle cannot be put into reverse. `carla.VehicleControl.reverse` was never set, so anything that has to back up does not work — freespace parking in particular.
- `/vehicle/status/gear_status` never reflected what Autoware asked for, so the reported gear could not be used to confirm the command took effect.

This PR subscribes to the gear command and:

- maps `REVERSE` and `REVERSE_2` onto `carla.VehicleControl.reverse`,
- holds the brake in `NONE`, `NEUTRAL` and `PARK` so the vehicle never pulls away in a gear it should not move in,
- reports the selected gear on `/vehicle/status/gear_status` instead of a constant.

`GearCommand` and `GearReport` share the same constant numbering, so the command value is reported back unchanged.

Until a gear command arrives the reported gear stays `DRIVE`, which is what the bridge did before this change, so an existing setup that only publishes `actuation_cmd` behaves as it did.

## Related links

**Parent Issue:**

- None

## How was this PR tested?

- Run against a live CARLA 0.9.16 server on the rebased code, publishing `actuation_cmd` with a constant 0.5 accel and switching `gear_cmd`, reading CARLA's applied control and the ego's forward speed back out:

  | gear command | CARLA control                                 | forward speed |
  | ------------ | --------------------------------------------- | ------------- |
  | `DRIVE`      | reverse=False gear=1 throttle=0.50 brake=0.00 | +12.83 m/s    |
  | `REVERSE`    | reverse=True gear=-1 throttle=0.50 brake=0.00 | -4.02 m/s     |
  | `PARK`       | reverse=False gear=1 throttle=0.00 brake=1.00 | 0.00 m/s      |
  | `DRIVE`      | reverse=False gear=1 throttle=0.50 brake=0.00 | +13.35 m/s    |
- `isort`, `black` and `flake8-ros` pre-commit hooks pass on the changed file.

## Notes for reviewers

One behaviour worth a second opinion: leaving `PARK`/`NEUTRAL` does not restore the throttle on the standing control, so the vehicle stays braked until the next `actuation_cmd` arrives. At the rate that topic is published this is a fraction of a cycle, and erring toward braked seemed the safer of the two.

#### Additions and removals

| Change type | Topic Type | Topic Name                   | Message Type                       | Description                       |
| :---------- | :--------- | :--------------------------- | :--------------------------------- | :-------------------------------- |
| Added       | Sub        | `/control/command/gear_cmd`  | `autoware_vehicle_msgs/GearCommand` | Gear selection from Autoware      |

## Effects on system behavior

- The ego vehicle can now be driven in reverse.
- `/vehicle/status/gear_status` reports the commanded gear rather than a constant `DRIVE`. Anything that relied on the constant will now see `PARK`, `NEUTRAL` or `REVERSE` when Autoware commands them.
- In `NONE`, `NEUTRAL` and `PARK` the accelerator is ignored and the brake is held.


simulator/autoware_carla_interface/src/autoware_carla_interface/carla_ros.py
```diff
@@ -704,8 +717,32 @@ def control_callback(self, in_cmd):
...
+    @staticmethod
+    def _apply_gear(control, gear):
+        """Apply the selected gear to a CARLA control command.
+
+        CARLA has no gear selector of its own for the throttle direction, so
+        reverse is a flag on the control command. In a gear the vehicle cannot
+        pull away in, the accelerator is ignored and the brake is held instead,
+        so the bridge never drives forward while reporting PARK or NEUTRAL.
+        """
+        control.reverse = gear in REVERSE_GEARS
```
hakuturu583, line 732: `_apply_gear` only flips `control.reverse`, but the current `control_callback` builds every actuation command with `manual_gear_shift=True` and `gear=1`, so this runs on a **manual-transmission** control — and on CARLA's default (PhysX) vehicle movement that makes reverse unreliable.

In `DefaultMovementComponent::ProcessControl`, manual mode selects the gear from `Control.gear`; the `reverse` flag is only consulted on the tick it *changes* (`GetVehicleControl().bReverse != Control.bReverse`). And `FlushVehicleControl` rewrites `LastAppliedControl.bReverse = (gear < 0)` after each apply. So a *sustained* REVERSE command alternates every tick between:
- the "reverse changed" branch → `SetTargetGear(-1)`, and
- the manual `else` branch → `SetTargetGear(Control.gear)` = `SetTargetGear(1)`

i.e. the transmission oscillates reverse ↔ forward-1 and the ego won't back up reliably. Selecting `gear = -1` keeps the manual branch in reverse. (Consistent with the PR note that reverse was verified on an earlier `carla_ros.py` that didn't force `manual_gear_shift=True`/`gear=1`.)

```suggestion
        control.reverse = gear in REVERSE_GEARS
        if control.reverse:
            # control_callback keeps manual_gear_shift=True with gear=1, and in
            # manual mode CARLA selects the gear from Control.gear (the reverse
            # flag only applies on the tick it changes), so reverse must set -1.
            control.gear = -1
```

PR #13151, opened 2026-08-04
feat(autoware_carla_interface): allow cameras to publish mono8

ujinf74:
## Description

Cameras are always published as `bgra8`, which is what CARLA renders. A consumer that only wants luminance — feature tracking, visual odometry, anything that calls `cvtColor` on arrival — pays four times the bytes through serialisation and the transport for three channels it then discards.

This PR adds an optional `image_encoding` to a camera's `ros_config`, accepting `bgra8` (the default, unchanged behaviour) or `mono8`. When `mono8` is selected the conversion happens once in the bridge, where the frame is already in hand as an array, instead of once per consumer and after the bytes have already crossed the transport.

Measured on recorded 960x540 frames: 2,073,672 bytes serialised as `bgra8` against 518,472 as `mono8`, a factor of 4.0.

An unrecognised `image_encoding` warns once at load time and falls back to `bgra8`, so a typo does not fail per frame inside a publish worker.

This composes with the subscriber-count check added in #12748: that skips the work when nobody is listening, this reduces the work when somebody is.

## Related links

**Parent Issue:**

- None

## How was this PR tested?

- The 4.0 factor is measured from `sensor_msgs/Image` payloads recorded to rosbag from this bridge at 960x540, one bag per encoding.
- `cv2.COLOR_BGRA2GRAY` output shape and size verified against the `bgra8` input for the same frame geometry.
- `isort`, `black`, `flake8-ros`, `markdownlint` and `prettier` pre-commit hooks pass on the changed files.
- The `mono8` path was exercised against CARLA 0.9.16 while developing this on an earlier release of the package. The version here is that work moved onto current `main`, where `carla_ros.py` has since been restructured; the rebased code has not itself been run against CARLA.

## Notes for reviewers

`import cv2` is added to `carla_ros.py`. `multi_camera_combiner_node.py` in this package already imports it and `package.xml` declares no explicit OpenCV dependency, so no dependency change is included here — flag it if you would rather one were declared.

`SUPPORTED_IMAGE_ENCODINGS` lives in `sensor_manager.py` next to the field it validates, so the loader and the dataclass cannot drift apart.

#### Additions and removals

| Change type | Parameter Name                          | Type     | Default Value | Description                                        |
| :---------- | :-------------------------------------- | :------- | :------------ | :------------------------------------------------- |
| Added       | `sensor_mappings.*.ros_config.image_encoding` | `string` | `bgra8`       | Image encoding for a camera: `bgra8` or `mono8` |

## Effects on system behavior

None by default — cameras without `image_encoding` continue to publish `bgra8`. A camera configured as `mono8` publishes single channel images, and a consumer expecting `bgra8` from that camera would need updating.


hakuturu583:
Thanks for the PR!
For future PRs that introduce new features, we’d appreciate it if you could attach a screenshot showing what the new feature is and how it works.

hakuturu583 (approved):

ujinf74:
> Thanks for the PR! For future PRs that introduce new features, we’d appreciate it if you could attach a screenshot showing what the new feature is and how it works.

Thanks for the feedback! I’ll include a screenshot demonstrating the feature and how it works in future feature PRs.

PR #13152, opened 2026-08-05
feat(trajectory_processor): unify plugin interface

maxime-clem:
## Description

Unify the modifier and optimizer plugins in the `autoware_trajectory_processor` package to use the same base plugin interface and data structures.

https://tier4.atlassian.net/browse/PDDP-241

## Related links

- https://tier4.atlassian.net/browse/PDDP-241

## How was this PR tested?

Unit tests

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


planning/autoware_trajectory_processor/src/trajectory_modifier.cpp
```diff
@@ -101,13 +100,18 @@ void TrajectoryModifier::on_traj(const CandidateTrajectories::ConstSharedPtr msg
...
+        modifier->process(trajectory.points, input_data) !=
+        trajectory_processor::plugin::ProcessingResult::Modified) {
+        continue;
+      }
       modifier->publish_planning_factor();
       const auto ns = "trajectory_" + std::to_string(trajectory_count);
       modifier->publish_debug_data(ns);
       if (!modified_plugins_str.empty()) modified_plugins_str += ", ";
-      modified_plugins_str += modifier->get_short_name();
+      modified_plugins_str += modifier->get_name();
```
mkquda: should probably keep `get_short_name()` here

planning/autoware_trajectory_processor/src/trajectory_modifier.cpp
```diff
@@ -162,20 +167,14 @@ tl::expected<plugin::InputData, std::string> TrajectoryModifier::make_input_data
...
 {
-  // Check if the plugin is already instantiated
-  auto it = std::find_if(
-    plugins_.begin(), plugins_.end(), [&](const auto & p) { return p->get_name() == name; });
-  if (it != plugins_.end()) {
-    RCLCPP_WARN(
-      this->get_logger(), "The plugin '%s' is already in the plugins list.", name.c_str());
-    return;
-  }
-
```
mkquda, line 175: in case of modifier plugins we don't want multiple instances of plugins

maxime-clem: Does it need to be enforced explicitly ?
In the next PR, I will combine the 2 nodes into one, and there will be only one loader, so if possible I want to "allow" multiple plugin instances.

planning/autoware_trajectory_processor/src/trajectory_modifier.cpp
maxime-clem: Fixed: 895a2f98a64

planning/autoware_trajectory_processor/src/trajectory_modifier.cpp
maxime-clem: @mkquda the optimizer does not check for duplicate plugins, so if it's okay I will keep the current implementation

mkquda (approved):
LGTM.

PR #13155, opened 2026-08-05
feat(autoware_ptv3)!: feed the encoder the level-0 serialization order

mojomex:
## Status quo

Every pooled level's `serialized_order` and `serialized_inverse` are handed to the encoder engine as inputs, but the input level's are not: the engine takes `serialized_code` and argsorts it in-graph. That is two `Argsort` nodes over all voxels, recomputing a ranking the preprocessing pipeline already produces to build the pooling metadata it does pass in.

## What this does

`PreprocessCuda` publishes the level-0 order and inverse it already derives (order 0 is the identity; the rest come from the one remaining sort), and the encoder and seg3d-head engines bind those instead of `serialized_code`. `serialized_code` stays a host-side buffer: it still chains the pooling stages.

Since stage 0 stops being special, the seg3d head's input rule collapses - every block stage, including the input level, gets the same `(serialized_order, serialized_inverse, grid_coord)` triple, without the `serialized_pooling_<i>_` prefix.

Requires tier4/autoware-ml#85 for the matching graph, plus a re-export.

In the ONNX graph, this replaces the circled subgraphs with inputs:

image

## Kernel changes

Added: `scatterInverseKernel`, and an `input_level_inverse_d_` buffer alongside the existing `input_level_order_d_`.

Removed: nothing on the CUDA side. The two `Argsort` nodes disappear from the engine, not from this package - they were graph nodes served by the `autoware_tensorrt_plugins` Argsort plugin.

`ptv3_trt.cpp` swaps the `serialized_code` engine input for `serialized_order` / `serialized_inverse` in both the encoder and the seg3d head, and `PreprocessCuda` gains `inputLevelSerializedOrder()` / `inputLevelSerializedInverse()` accessors.

## Before / after

image

| | before | after |
|---|---|---|
| encoder `Argsort` nodes | 2 | **0** |
| Argsort kernel time (radix sort + `fill_iota`) | ~140 us | 0 |
| encoder enqueue window | 2579.5 us | 2512.1 us |
| encoder enqueue kernel launches | 69 | 44 |
| inference stage mean | 17.691 ms | 17.650 ms |
| node peak GPU memory | 5616 MiB | 5612 MiB |

## Tests

`MatchesCpuReferenceForOnnxFacingInputs` asserts the published level-0 order and inverse match a CPU `stable_argsort` of the level-0 codes for every serialization order, with row 0 required to be the identity. Verified by mutating `scatterInverseKernel` to not invert - the test fails on `serialized_inverse` row 1.

## Stack

- PR 1, #13153 (**merged**): replaces voxelization hashing with serialized codes.
- PR 2, #13170: derives the pooling metadata with prefix scans instead of sorts; introduces the input-level order buffer this PR extends. **Must merge first** - this PR does not apply to `main` without it.
- **PR 3 (this one)**: the interface change. Must land together with tier4/autoware-ml#85 and a re-export, or engines fail to bind their inputs.

## Breaking change

The encoder engine's `serialized_code` input is replaced by `serialized_order` and `serialized_inverse`; the seg3d head's stage-0 inputs likewise. Existing engines must be re-exported.

### DIFF
--- a/perception/autoware_ptv3/include/autoware/ptv3/preprocess/preprocess_kernel.hpp
+++ b/perception/autoware_ptv3/include/autoware/ptv3/preprocess/preprocess_kernel.hpp
@@ -46,12 +46,22 @@ public:
   PreprocessCuda(const PTv3Config & config, cudaStream_t stream);
   ~PreprocessCuda();
 
+  // Crops, voxelizes and deduplicates the input cloud. The emitted voxels are ordered by their
+  // order-0 serialized code, which generateSerializedPoolingMetadata requires (see below).
   std::size_t generateFeatures(
     const void * input_data, CloudFormat input_format, unsigned int num_points,
     float * voxel_features, std::int32_t * voxel_coords, std::int64_t * voxel_hashes,
     void * compact_points, float * reconstruction_features, void * cropped_source_points,
     std::int64_t * inverse_map, std::size_t * num_cropped_points);
 
+  // Builds the per-stage pooling metadata the encoder graph consumes.
+  //
+  // PRECONDITION: `serialized_code` row 0 must be ascending, i.e. the voxels must be stored in
+  // order-0 serialization order. generateFeatures guarantees this. The whole hierarchy is then
+  // derived with prefix scans instead of sorts: right-shifting a serialized code yields its
+  // parent's code, which is monotone, so pooling preserves the ordering and every coarser level
+  // stays sorted by construction. Feeding an arbitrarily ordered level here silently produces wrong
+  // metadata.
   void generateSerializedPoolingMetadata(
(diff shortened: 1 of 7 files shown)


amadeuszsz (changes requested):
Overall LGTM, but since PR needs model reexport, we will merge it alongside new model release

vividf (approved):
LGTM!

mojomex:
@amadeuszsz Is this good to merge? Seems like your approval is required.

amadeuszsz (approved):
LGTM!

memo: this PR breaks inference, the new re-exported model is required. We confirmed latest v0.1.0 PTv3 release works correctly and therefore we would need to update artifacts in `awf/autoware`. However, since PTv3 is not used by default in OSS pipeline, I will update artifacts very soon with v0.1.1 release.

PR #13158, opened 2026-08-06
fix(autoware_tensorrt_plugins): make spconv plugins safe under CUDA graph capture

mojomex:
## Description

The PTv3 encoder engine cannot be built on DRIVE Thor (TensorRT 10.10). TensorRT captures the stream into a CUDA graph while it times tactics during engine build, and the spconv-backed plugins issue CUDA calls that are illegal while a stream is capturing. The plugins return `cudaErrorStreamCaptureUnsupported` (900) or `cudaErrorStreamCaptureInvalidated` (901), TensorRT then reports `Could not find any implementation for node {ForeignNode[...]}`, and spconv aborts the process outright (`CUB error: invalid device ordinal`, later `cuBLAS API failed`).

The same ONNX, optimization profile and plugin sources build without complaint on x86 (TensorRT 10.8), which is why this went unnoticed.

Three distinct violations are fixed:

| Location | Violation |
|:---|:---|
| `GetIndicesPairsPlugin::enqueue` | `cudaStreamSynchronize` on every call, plus a host-to-device `cudaMemcpyAsync` whose source is a stack variable — a pageable copy has synchronous behaviour and is rejected during capture |
| `GetIndicesPairsImplicitGemmPlugin::enqueue` | the same pageable host-to-device copy |
| `IndiceConvPlugin::enqueue` | constructs `SimpleExternalSpconvMatmul` per call, whose constructor calls `cublasLtCreate` |

Both counts are now published with a one-element kernel (`fill_scalar_int32`). Kernel arguments are passed by value, so no synchronization is needed and the source value's lifetime is no longer a concern. The cuBLASLt handle now lives for the lifetime of the plugin, with the allocator it references retargeted at the current bindings via `set_new_tensor_dict`.

Removing the per-call stream synchronization and the `cublasLtCreate`/`cublasLtDestroy` pair also takes work off the inference path, which runs once per frame.

Those fixes are necessary but not sufficient: further capture-illegal calls remain inside spconv itself, and fixing them one at a time does not converge. The four spconv-backed plugins therefore also detect capture and emit zeroed outputs instead of running. Plugin output shapes come from `getOutputShapes` rather than from data, so only tactic timing is affected, and these plugins expose a single tactic, so that timing is not used to choose between alternatives.

## Related links

**Parent Issue:**

- None

## How was this PR tested?

Built and run on a DRIVE Thor board and on an x86 workstation, using a standalone TensorRT builder that loads the plugin library, parses the ONNX and applies the same optimization profile as `PTv3TRT::initEncoderTrt`, so ROS is not in the loop.

- **DRIVE Thor (TensorRT 10.10.10), before:** the process aborts during builder tactic timing, inside spconv.
- **DRIVE Thor, after:** `ENGINE BUILD SUCCEEDED after 39 s`. The resulting engine deserializes and creates an execution context on the device.
- **x86 (TensorRT 10.8), before and after:** engine builds successfully, so the previously working platform does not regress.
- **Inference:** `enqueueV3` on the built engine returns `cudaSuccess` on both a cold and a warm run, and `cudaStreamIsCapturing` reports no capture around it, confirming the capture path is build-only and never taken during ordinary inference.

Verified against a submanifold PTv3 encoder graph (all `GetIndicePairs`/`GetIndicePairsImplicitGemm` nodes have `subm=1`).

## Notes for reviewers

These plugins consequently cannot be used inside an application-driven CUDA graph: capturing `enqueueV3` yields a graph of zeros rather than an error. Nothing in Autoware does that today, and this network's dynamic per-frame shapes make CUDA graphs a poor fit for it, so this is called out in a code comment and a runtime warning rather than enforced.

A variant keyed on `nvinfer1::TensorRTPhase` (emit zeros for `kBUILD`, fail for `kRUNTIME`) was prototyped and deliberately dropped. It does make runtime capture fail loudly, but TensorRT also creates some `kRUNTIME` plugin instances during the build: one of them hit the capture path, logged an error and cost that layer a tactic candidate. That trades a real build-robustness risk for protection against a scenario nothing currently exercises. Doing it properly would need an explicit "engine build in progress" flag set by `autoware_tensorrt_common` around `buildSerializedNetwork`, OR'd with the `kBUILD` check; happy to follow up separately if reviewers prefer that.

Separately, and out of scope here: on Thor the encoder still exceeds the 14.67 GiB GPU carve-out at the configured `voxels_num`. With this fix the build succeeds up to a max of about 24 000 voxels and fails above roughly 32 000, driven by the fused Myelin region rather than by plugin buffer bounds. Neither the TensorRT workspace limit nor `out_indices_num_limit_` affects it.

## Interface changes

None.

## Effects on system behavior

None for existing platforms. On DRIVE Thor the PTv3 encoder engine can be built where it previously could not.

### DIFF
--- a/perception/autoware_tensorrt_plugins/CMakeLists.txt
+++ b/perception/autoware_tensorrt_plugins/CMakeLists.txt
@@ -108,6 +108,7 @@ if(TRT_AVAIL AND CUDA_AVAIL AND SPCONV_AVAIL)
     src/unique_ops/unique.cu
     src/multi_scale_deform_attn_ops/ms_deform_attn_kernel.cu
     src/rotate_ops/rotate_kernel.cu
+    src/scalar_ops/fill_scalar.cu
     src/select_and_pad_ops/select_and_pad_kernel.cu
   )
 
--- a/perception/autoware_tensorrt_plugins/include/autoware/scalar_ops/fill_scalar.hpp
+++ b/perception/autoware_tensorrt_plugins/include/autoware/scalar_ops/fill_scalar.hpp
@@ -1,0 +1,31 @@
+// Copyright 2026 TIER IV, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef AUTOWARE__SCALAR_OPS__FILL_SCALAR_HPP_
+#define AUTOWARE__SCALAR_OPS__FILL_SCALAR_HPP_
+
+#include <cuda_runtime_api.h>
+
+#include <cstdint>
+
+/// \brief Write a single host-known ``std::int32_t`` into device memory.
+///
+/// This exists so that plugins can publish a scalar they computed on the host without issuing a
+/// host-to-device ``cudaMemcpyAsync`` from pageable memory. Such a copy has synchronous behaviour
+/// and is rejected with ``cudaErrorStreamCaptureUnsupported`` while a stream is capturing, which
+/// TensorRT does when it times tactics during engine build. A kernel launch carries its arguments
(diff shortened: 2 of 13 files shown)


mojomex:
Related: #13159 makes explicit-fp16 (strongly typed) PTv3 models work independently of this PR — strongly typed builds never execute plugin tactic timing, so they do not hit the capture issue this PR fixes. This PR remains valuable for weakly typed builds on Thor (fp32 stopgaps, other spconv models) and as a per-frame inference optimization (removes a stream sync, a pageable H2D copy per GetIndicePairsImplicitGemm call, and cuBLASLt handle churn).

PR #13159, opened 2026-08-07
feat(autoware_ptv3): allow strongly typed engine builds

mojomex:
> **Stacked on #13191** (renames the `autoware_tensorrt_common` precision value to `"strongly-typed"`). This PR's branch contains that commit; until #13191 merges, review only the `autoware_ptv3` commit.

## Description

Adds `"strongly-typed"` to the values the node's `trt_precision` parameter accepts. Tensor precisions are then taken from the model itself instead of being assigned by the TensorRT builder (via #13160, merged).

The node needs no code change: it already forwards `trt_precision` to the engine config of all three models (`encoder`, `segmentation3d`, `detection3d`), so this PR only widens the parameter's schema and documents what the values mean.

Models exported with reduced precision need this value. The weakly typed precision-assignment pass would re-process their already-reduced tensors, and on one of our embedded targets it crashes on PTv3-shaped graphs regardless of builder settings; strongly typed builds skip that pass entirely. Explicit-fp16 packages built this way are the fp16 deployment path there.

Engine I/O, node buffers and kernels stay fp32 for every model: reduced-precision exports keep fp32 tensors at the graph boundary through cast layers inside the model, which the exporter handles (see tier4/autoware-ml#89). Strong typing is independent of the model's precision — an fp32 export can also be built `"strongly-typed"`, for reproducible model-defined numerics.

Related but independent: #13158 (spconv plugin fixes) is not required for strongly typed builds, which never execute plugin tactic timing — verified by building the fp16 package on the embedded target with unfixed plugins.

## Related links

**Parent Issue:**

- None

- Prerequisite (stacked base): #13191
- Precision value introduced by: #13160 (merged)
- Exporter-side counterpart: tier4/autoware-ml#89
- Related, independent: #13158

## How was this PR tested?

Running the actual node with `build_only:=true` against ML packages derived from the PTv3 HuggingFace package (small `voxels_num` bounds):

- **fp32 package, `trt_precision: fp16`** (x86, TensorRT 10.8): unchanged weakly typed behavior — all three engines build, node exits cleanly.
- **fp32 package, `trt_precision: strongly-typed`** (x86): all three engines build strongly typed as pure fp32, confirming the value is valid independently of the model's precision.
- **fp16-internal package (fp32 graph I/O), `trt_precision: strongly-typed`**: all three engines build strongly typed and the node exits cleanly, on x86 and on the embedded target (DRIVE Thor), where the weakly typed fp16 build of the same models crashes.

(These were run against the `"as-is"` spelling of the value; #13191 renames the string only.)

## Notes for reviewers

- An earlier revision added a per-model `strongly_typed` flag to the ML package schema and refused to launch when it disagreed with `trt_precision`. That is gone: `trt_precision` alone selects the build, and which value a given export needs is documented on the parameter. Thanks @amadeuszsz for pushing on both points.
- If a reduced-precision export is deployed with `fp16`/`fp32` instead, it parses weakly typed — it will typically build on x86 but hit the platform builder bug on the embedded target, the same behavior as today's baseline.
- fp16 accuracy validation against the fp32 reference is pending; this PR is about build compatibility.

## Interface changes

None in C++ APIs.

#### Modifications

| Change type | Parameter Name  | Type     | Default Value | Description                                                                                                                                                    |
| :---------- | :-------------- | :------- | :------------ | :------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| Modified    | `trt_precision` | `string` | `fp16`        | Gains the `"strongly-typed"` value: build engines with tensor precisions taken from the model instead of assigned by the builder. `fp16`/`fp32` are unchanged. |

## Effects on system behavior

None for existing configs: the default and the existing `trt_precision` values behave as before. With `"strongly-typed"`, engines are built with reproducible, model-defined precisions, and reduced-precision packages become buildable on targets where weakly typed fp16 builds fail.


perception/autoware_ptv3/config/ptv3.param.yaml
```diff
@@ -3,6 +3,9 @@
     # plugins
     plugins_path: $(find-pkg-share autoware_tensorrt_plugins)/plugins/libautoware_tensorrt_plugins.so
     # network
+    # Precision requested for weakly typed TensorRT builds (fp16, fp32), or as-is to build
```
amadeuszsz: Similar comment as in autoware-ml. This is schema file job (one source of truth), please do not write documentation in the config file.

perception/autoware_ptv3/src/ptv3_node.cpp
```diff
@@ -49,12 +49,25 @@ PTv3Node::PTv3Node(const rclcpp::NodeOptions & options) : Node("ptv3", options)
     this->declare_parameter<std::string>("trt_precision", descriptor);
   const auto cloud_capacity = this->declare_parameter<std::int64_t>("cloud_capacity", descriptor);
 
+  // ML packages declare strongly_typed for engines that must be built with their tensor
+  // precisions taken from the model itself, which requires trt_precision "as-is".
+  const auto declare_strongly_typed = [this, &trt_precision](const std::string & section) {
```
amadeuszsz: As discussed, please handle it with one parameter.

mojomex: Done — the `strongly_typed` flag is gone from the ML package schema and from the node, and `trt_precision` is the only parameter involved. It now takes `"strongly-typed"` (renamed from `"as-is"` in #13191) alongside `fp16`/`fp32`, so the value itself says which build it selects and there is nothing left to disagree with.

With the flag removed the node needed no code change at all: it already forwarded `trt_precision` to every model's engine config, so what remains is the schema entry.

perception/autoware_ptv3/config/ptv3.param.yaml
mojomex: Removed — the config file carries no documentation now, and `trt_precision`'s description in `schema/ptv3.schema.json` is the single source of truth. It also covers which value a reduced-precision export needs, which is what the removed `strongly_typed` flag used to enforce.

amadeuszsz (approved):
LGTM

perception/autoware_ptv3/src/ptv3_node.cpp
```diff
@@ -49,12 +49,25 @@ PTv3Node::PTv3Node(const rclcpp::NodeOptions & options) : Node("ptv3", options)
...
+    const bool strongly_typed = this->declare_parameter<bool>(section + ".strongly_typed", false);
+    if (strongly_typed && trt_precision != "as-is") {
```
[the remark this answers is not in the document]
mojomex: Won't fix. Proper caching in the face of config changes is future work.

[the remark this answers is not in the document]
mojomex: Won't fix. Proper caching in the face of config changes is future work.

perception/autoware_ptv3/schema/ptv3.schema.json
```diff
@@ -254,8 +254,8 @@
         },
         "trt_precision": {
           "type": "string",
-          "description": "TensorRT inference precision.",
-          "enum": ["fp16", "fp32"]
+          "description": "How the TensorRT engines are built. \"fp16\" and \"fp32\" request a precision for a weakly typed build, letting the builder assign tensor precisions. \"strongly-typed\" takes every tensor precision from the model itself; use it for models exported with reduced precision (e.g. explicit-fp16 ONNX), whose baked-in types must not be reassigned by the builder.",
```
mojomex:
```suggestion
          "description": "How the TensorRT engines are built. \"fp16\" and \"fp32\" request a precision for a weakly typed build, letting the builder assign tensor precisions. \"strongly-typed\" takes every tensor precision from the model itself.",
```

PR #13167, opened 2026-08-10
fix(autoware_camera_streampetr): fix online offline difference

vividf:
## Description

The StreamPETR ONNX consumes RGB, but the node's preprocessing was BGR throughout: the
normalization constants were the BGR-ordered statistics and the fused kernel wrote the source
channels straight through in source order. A bgr8 camera therefore fed the model BGR planes, and
an rgb8 camera fed RGB planes normalized with BGR statistics — neither is what the model was
trained on. This PR makes the channel order explicit and correct end-to-end, and fixes a deployment
blocker found while bringing the compressed-image path up under a namespaced launch.

1. **RGB normalization + channel order.** `image_input_mean_`/`image_input_std_` are now in RGB
   order (`{123.675, 116.280, 103.530}` / `{58.395, 57.120, 57.375}`). The fused preprocess kernel
   gained a `swap_rb` flag: when the source buffer is BGR, source channel `i` is written to model
   channel `2 - i`, and mean/std are indexed by the *model* channel so they need no reordering.
   The swap is free — the kernel already scatters one write per channel plane, so there is no extra
   pass and no intermediate buffer.

2. **Input validation in `CameraDataStore`.** The incoming `sensor_msgs/Image` is now checked
   before it is uploaded: `rgb8` passes through, `bgr8` sets `swap_rb`, anything else is rejected.
   The buffer geometry is checked too — both upload paths `cudaMemcpyAsync` exactly
   `height * width * 3` densely packed bytes out of `msg.data`, so a padded row stride would shear
   the image and a truncated buffer was a host-side out-of-bounds read. Rejected frames are dropped,
   with the error throttled to 10 s since a misconfigured camera hits it on every frame.

3. **Dropped `image_transport`; both image paths are plain subscriptions now.** The compressed
   plugin is unusable under a namespace: `CompressedSubscriber::subscribeImpl` calls
   `substr(node_namespace.size())` on the *unresolved* base topic, so it throws `std::out_of_range`
   whenever `strlen(base_topic) < strlen(node_namespace)`. With `"~/input/cameraN/image"`
   (21 chars) under a namespace such as `"/perception/object_recognition/detection"` (40 chars) the
   node died at construction — reproduced against `ros-humble-compressed-image-transport 2.5.5`,
   and raw transport is unaffected (its `getTopicToSubscribe()` returns the base topic untouched).
   `is_compressed_image` now selects between `create_subscription<CompressedImage>` (decoded with
   `cv_bridge`) and `create_subscription<Image>`. Since the transport is chosen by a parameter
   rather than at run time, `image_transport` bought nothing on the raw path either, so the
   dependency is gone from the package (including a stray unused include in `network.hpp`).
   Topic names are unchanged, so the existing launch remappings still apply — including the
   `image/compressed` remaps, which already pointed at the resolved compressed topic.

4. **Ego mask fill colour is now RGB and follows the source channel order.** The ego mask (#12656)
   paints the configured colour directly into the raw camera buffer, which this PR makes explicitly
   either RGB or BGR. The parameter is renamed `ego_mask.fill_value_bgr` → **`ego_mask.fill_value_rgb`**
   so that it matches the model's input order, the `rgb8` encoding the online cameras publish, and
   the RGB-ordered mean/std in (1) — the package now has a single channel-order convention instead
   of two. The fill is reordered to the source buffer's channel order before the mask kernel runs,
   which is now a no-op on the online (`rgb8`) path and a swap on the compressed (`bgr8`) path.
   No visible change with the default black fill; it matters only for a configured non-grey colour.
   `applyEgoMask_launch()`'s parameters are renamed from BGR-specific names to channel-neutral ones
   (`image`, `fill_c0..c2`), since the kernel writes three bytes in buffer order and never had any
   channel semantics of its own. The identity function `clampFillBgr()` is removed — clamping
   already happens where the parameter is read.

5. **Diagnostics.** The "Not all camera info or image received" warning now names which camera IDs
   are missing `camera_info` vs. missing images.

## Notes for reviewers

- **This changes detection output.** The model previously received BGR, and the post-processing
  thresholds were tuned against that. `confidence_threshold` is carried over unchanged and may
  need re-tuning.
- The compressed path decodes to `bgr8` on purpose: that is what `cv::imdecode` returns, so
  `cv_bridge` hands the buffer back untouched. Asking for `rgb8` would add a full-image CPU
  `cvtColor` per camera per frame, while the GPU swap costs nothing.
- The mask fill reorder in (4) costs three byte moves per frame on the host, not per pixel — it is
  a no-op on the online path and is not on any hot path either way.
- Encodings are matched against `sensor_msgs::image_encodings` constants rather than string
  literals. Mono, Bayer, 16-bit and alpha encodings are all rejected: they have a different channel
  count or element size, so the fixed 3-bytes-per-pixel upload would misread the buffer.

## Effects on system behavior

- Detection results change for every deployment: the model now receives the channel order and
  normalization statistics it was trained with.
- Cameras publishing an encoding other than `rgb8`/`bgr8`, or a non-densely-packed buffer, are
  dropped instead of silently producing garbage.
- Deployments with a non-grey ego-mask fill on an `rgb8` camera will see the masked region change
  colour to the one actually configured.

## Interface changes

**One breaking parameter rename:** `ego_mask.fill_value_bgr` → `ego_mask.fill_value_rgb`, with the
component order reversed accordingly. A config still setting the old name is silently ignored
(rclcpp does not reject undeclared parameter overrides) and the mask falls back to the default
black, so existing configs must be updated. The default `[0, 0, 0]` is unchanged, so deployments
that never set a non-grey fill are unaffected.

ROS topics and launch arguments are otherwise unchanged. Package-internal only:
`resizeAndExtractRoi_launch()` gains a `swap_rb` argument;
`CameraDataStore::process_distorted_image()`/`process_regular_image()` (both private) gain a
`swap_rb` argument so the ego mask can order its fill colour; `applyEgoMask_launch()`'s fill
parameters are renamed to channel-neutral names; `EgoMaskParams::fill_bgr` /
`EgoMaskRoiConfig::fill_bgr` become `fill_rgb`. The `image_transport` dependency is removed from
`package.xml`.


perception/autoware_camera_streampetr/include/autoware/camera_streampetr/network/camera_data_store.hpp
```diff
@@ -85,10 +86,10 @@ class CameraDataStore
     const int camera_id, const Image::ConstSharedPtr & input_camera_image_msg) const;
   std::unique_ptr<Tensor> process_distorted_image(
     const int camera_id, const Image::ConstSharedPtr & input_camera_image_msg,
-    ImageProcessingParams & params);
+    ImageProcessingParams & params, const bool swap_rb);
```
KSeangTan, line 88: `swap_rgb` sounds more intuitive.

perception/autoware_camera_streampetr/include/autoware/camera_streampetr/network/camera_data_store.hpp
```diff
@@ -100,6 +101,11 @@ class CameraDataStore
     const int camera_id, const std::vector<std::uint8_t> & raster, const int width,
     const int height);
 
+  // Validates that the message can be fed to the model and reports whether the preprocessing
+  // kernel has to swap R/B. Returns false if the frame must be dropped.
+  bool validate_channel_order(
```
KSeangTan: `validate_color_channel_order`

perception/autoware_camera_streampetr/lib/network/camera_data_store.cpp
```diff
@@ -225,6 +250,45 @@ void CameraDataStore::update_camera_image(
...
+      logger_, *clock_, 10000 /* ms */,
+      "Camera %d publishes unsupported encoding '%s'. Only 'rgb8' and 'bgr8' can be fed to the "
+      "model; dropping frames from this camera.",
+      camera_id, encoding.c_str());
+    return false;
+  }
+
+  // Both upload paths copy height * width * 3 densely packed bytes straight out of the message,
+  // so a padded row stride would shear the image and a truncated buffer would read out of bounds.
+  const size_t row_bytes = static_cast<size_t>(input_camera_image_msg->width) * 3;
```
KSeangTan, line 266: What is the purpose of this function?

perception/autoware_camera_streampetr/lib/network/camera_data_store.cpp
```diff
@@ -453,6 +518,36 @@ bool CameraDataStore::check_if_all_camera_image_received() const
   return true;
 }
 
+std::string CameraDataStore::get_missing_camera_status() const
```
KSeangTan: What is the purpose of this function?

perception/autoware_camera_streampetr/lib/network/preprocess_kernel.cu
```diff
@@ -133,22 +139,24 @@ __global__ void resizeAndExtractRoi_kernel(
...
   int area = roi_h * roi_w;
   int out_idx = out_y * roi_w + out_x;
 
-  // Channel 0 (R)
-  output_img[camera_offset + (0 * area + out_idx)] = (sum_r - mean[0]) / std[0];
-  // Channel 1 (G)
-  output_img[camera_offset + (1 * area + out_idx)] = (sum_g - mean[1]) / std[1];
-  // Channel 2 (B)
-  output_img[camera_offset + (2 * area + out_idx)] = (sum_b - mean[2]) / std[2];
+  const int dst_c0 = swap_rb ? 2 : 0;
```
KSeangTan: TBH, i feel supporting RGB swapping in the code introduces unnecessary complexity. For example, an user provides flipped values for mean/std. IMO, we should always check if the input channels are `RGB8`. Otherwise, it raises an error.

perception/autoware_camera_streampetr/src/node.cpp
```diff
@@ -310,6 +324,23 @@ void StreamPetrNode::camera_image_callback(
...
 }
 
+void StreamPetrNode::compressed_image_callback(
+  CompressedImage::ConstSharedPtr input_compressed_image_msg, const int camera_id)
+{
+  cv_bridge::CvImagePtr cv_ptr;
+  try {
+    // bgr8 is what cv::imdecode already produces; rgb8 would cost a CPU cvtColor per frame while
+    // the preprocessing kernel swaps R/B on the GPU for free.
+    cv_ptr = cv_bridge::toCvCopy(input_compressed_image_msg, sensor_msgs::image_encodings::BGR8);
```
KSeangTan: What's the cost swapping `BGR8` to `RGB8`? Using compressed image itself costs some latency, for example, it moves compressed images to memory and decode it. So the cost is unavoidable as long as we are using compressed_image.

KSeangTan (changes requested):
Thanks for the work, please address comments accordingly.

perception/autoware_camera_streampetr/include/autoware/camera_streampetr/network/camera_data_store.hpp
vividf: https://docs.opencv.org/4.13.0/d6/d0f/group__dnn.html#:~:text=Function%20Documentation-,%E2%97%86%C2%A0,blobFromImage()%20%5B1/2%5D,-Mat%20cv%3A%3Adnn

Since we are only swapping r and b, I think this is clearer

perception/autoware_camera_streampetr/lib/network/camera_data_store.cpp
vividf: This is the entrance check for every incoming frame; it guards the layout assumption both upload paths make. `process_regular_image()` and `process_distorted_image()` both `cudaMemcpyAsync` exactly `height * width * 3` densely packed bytes out of `msg.data` without looking at `step`, but `sensor_msgs/Image` permits a padded row stride (`step > width * 3`) and does not guarantee `data` is large enough. Without this check, a padded buffer would silently shear every row of the model input (no error anywhere — just quietly degraded detections), and a truncated buffer would make that copy a **host-side out-of-bounds read**.

The encoding half establishes the 3-bytes-per-pixel assumption (and reports whether the kernel has to swap R/B); the geometry half validates how those bytes are laid out. Rejected frames are dropped with a throttled error naming the camera and the actual vs. required step/size, and the state is surfaced through `input_rejected_` in the camera status so diagnostics can distinguish "camera silent" from "camera publishing an unusable format".

I've also expanded the comment on the function so this rationale lives in the code.

perception/autoware_camera_streampetr/include/autoware/camera_streampetr/network/camera_data_store.hpp
vividf: Thanks!
Renamed, though to validate_image_message() rather than validate_color_channel_order() — the function checks two things, the encoding and the buffer geometry (step / data.size()), and a channel-order name would cover only the first. I considered splitting it in two, but the geometry check's width * 3 depends on the encoding check having just established 3 bytes per pixel, and both feed the same sticky input_rejected_ state, so one function with a name covering both seemed cleaner than two with an ordering contract between them.

perception/autoware_camera_streampetr/lib/network/camera_data_store.cpp
vividf: I wanted to add some diagnostics for the StreamPetr node. I’ve removed this for now and will create a separate PR with a more concrete diagnostic message.

vividf: https://github.com/autowarefoundation/autoware_universe/pull/13167/commits/1b205e4163d02cc1458d026eb4d52ea5fad89650

perception/autoware_camera_streampetr/lib/network/camera_data_store.cpp
vividf: https://github.com/autowarefoundation/autoware_universe/pull/13167/commits/1b205e4163d02cc1458d026eb4d52ea5fad89650

perception/autoware_camera_streampetr/src/node.cpp
vividf: On a desktop CPU (OpenCV 4.12, single-threaded), for a 1920×1080 BGR→RGB conversion:

cvtColor: ~0.2 ms/frame, ~1.3 ms total for 6 cameras (this is not that accurate, check https://github.com/autowarefoundation/autoware_universe/pull/13167#issuecomment-5292462181)
For reference, JPEG imdecode at the same resolution: ~3 ms/frame, so cvtColor adds roughly 7% on top of the decode
So you're right that it's not catastrophic relative to decode. My argument is rather that it's an avoidable per-frame CPU cost and an extra full-frame memory pass (read + write ~6 MB per camera), while the GPU-side swap is exactly zero cost — the preprocessing kernel already writes every pixel, and swapping R/B is just a change of destination channel index in the same scatter write.

perception/autoware_camera_streampetr/lib/network/preprocess_kernel.cu
vividf: The flipped mean/std scenario can't occur in this code. The mean/std are hardcoded constants loaded in the constructor ({123.675, 116.280, 103.530} / {58.395, 57.120, 57.375}, camera_data_store.cpp) — they are not exposed in any config, so there is no way for a user to provide flipped values.  We are just supporting raw (RGB) and uncompressed (BGR)

KSeangTan: IMO, the situation is similar to image color channels, we know `uncompressed_image` is always `RGB8`, and `compressed_image` is always `BGR8`. Since we know swapping channels in cv2 doesn't significantly increase overhead in your comment below, we should choose readability here rather than supporting two permutations, which i would strongly suggest to check color channels from inputs and raise an error if they are not `RGB8`.
For `compressed_image `, we should just swap their channels in cv2.

@YoshiRi @amadeuszsz @badai-nguyen @tzhong518 
Please let me know your opinion.

perception/autoware_camera_streampetr/src/node.cpp
KSeangTan: IMO, I would trade latency for `readability/simplicity` here since overhead is not significant.

vividf: Thanks for consideration. 
Just to clarify, the 1.3 ms depends on each computer CPUs. Also, as mentioned above, the GPU-side change is simply swapping the destination channel index, so it does not add extra processing. If there is a straightforward way to avoid the additional CPU-side conversion without sacrificing readability too much, I think it would still be worth optimizing.

perception/autoware_camera_streampetr/lib/network/preprocess_kernel.cu
vividf: @manato

perception/autoware_camera_streampetr/src/node.cpp
vividf: @manato

perception/autoware_camera_streampetr/lib/network/preprocess_kernel.cu
YoshiRi: [updated]
Largely agree with @KSeangTan because compressed data flow is only for offline.
It still sounds better to keep gpu code simple, but I am not sure about the actual computational burden when processing 2K-like 6 image decode.

I thought using `cv::IMREAD_COLOR_RGB` could make problem much easier but it seems this yields `4.11` and later opencv version. Currently I found humble use `4.5`, and jazzy also using `4.6` as a default.

manato (commented):
@vividf
Thank you for handling this problem! (and sorry for the delay in reviewing :pray: )
As the other reviewers suggested, I would suggest to handle color space conversion (i.e., bgr -> rgb) in early stage and use identical code in the later stage in terms of code readability and behavior consistency:

``` mermaid
flowchart TD
    A["input image"] --> B["memcpyAsync"]
    B --> C{"RGB or RGB?"}
    C -- "BGR" --> D["color space conversion on GPU"]
    C -- "RGB" --> E["Later process (identical code flow regardless of input color space)"]
    D -- "RGB" --> E
```

As far as I checked, the "color space conversion" can be put just after `cudaMemcpyAsync` of `CameraDataStore::process_distorted_image` and `CameraDataStore::process_regular_image`, with using the same CUDA stream of the memcpy.

In terms of color conversion, the typical options are:
1. using NVIDIA NPP Library (such as `nppiSwapChannels_8u_C3IR_Ctx` [[ref](https://github.com/tier4/accelerated_image_processor/blob/7a9cf33c4379cc67af3732d4a4c5a73596fcb248/src/accelerated_image_processor_compression/src/jpeg_compressor/jetson.cpp#L110-L114)]
2. writing a dedicated CUDA kernel

Option 1. is more handy and well tested code, but we need to align input/output memory type (`Npp8u` instead of `uint8_t`).
In the case this time, the objective is just swapping the channel order, so Option 2. is also reasonable candidate.

If you have any question about the above, feel free to ask me!

perception/autoware_camera_streampetr/lib/network/preprocess_kernel.cu
```diff
@@ -37,21 +37,47 @@
 
 namespace autoware::camera_streampetr
 {
+// In-place BGR -> RGB conversion: each thread swaps bytes 0 and 2 of one packed HWC pixel.
+__global__ void convertBGRToRGB_kernel(std::uint8_t * __restrict__ img, int num_pixels)
+{
+  int idx = blockIdx.x * blockDim.x + threadIdx.x;
+  if (idx >= num_pixels) return;
+
+  std::uint8_t * pixel = img + static_cast<size_t>(idx) * 3;
```
KSeangTan, line 46: Comment: why magic number of `3`.

KSeangTan (approved):
Thanks! LGTM Overall

perception/autoware_camera_streampetr/lib/network/preprocess_kernel.cu
vividf: Thanks, I will clean this up with another PR since all the other surrounding kernels used 3

PR #13169, opened 2026-08-12
perf(perception_utils): reduce NMS computation time using cache RDDR-996

ktro2828:
## Description

This pull request refactors the `IouBevNms` implementation to improve efficiency and clarity by introducing a new `CachedObject` structure, optimizing geometric calculations, and updating dependencies. The changes focus on reducing redundant computations during Non-Maximum Suppression (NMS) by caching geometric properties and refactoring related logic.

**Refactoring and performance improvements:**

* Introduced the `CachedObject` struct to cache frequently used properties (label, position, area, polygon, envelope) for each detected object, reducing repeated computation during NMS.
* Updated the main NMS loop in `IouBevNms::apply` to use `CachedObject` throughout, including sorting and output extraction, further streamlining the process. [[1]](diffhunk://#diff-052d2947b7dd0318752f1084106c92cb1781edfb3ce11b698010a134a6c1317bL51-R141) [[2]](diffhunk://#diff-052d2947b7dd0318752f1084106c92cb1781edfb3ce11b698010a134a6c1317bL85-R164)

As shown below, the computation time becomes **~195x faster** than previous implementation for 400 objects.

nms_only_profile_comparison

## Related links

**Parent Issue:**

- [TIER IV INTERNAL LINK: RDDR-996](https://tier4.atlassian.net/browse/RDDR-996)

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


perception/perception_utils/src/iou_bev_nms.cpp
```diff
@@ -30,6 +34,81 @@ namespace
...
+  cached.x = pose.position.x;
+  cached.y = pose.position.y;
+  cached.polygon = autoware_utils_geometry::to_polygon2d(object);
+  cached.area = boost::geometry::area(cached.polygon);
+  boost::geometry::envelope(cached.polygon, cached.envelope);
+
+  return cached;
+}
+
+bool isCachedTargetPairObject(
```
mojomex: The old name was confusing as well, but I'd vote for something like `isPairSubjectToNms` to make the function more self-explanatory.

perception/perception_utils/src/iou_bev_nms.cpp
```diff
@@ -30,6 +34,81 @@ namespace
...
+    (object1.label == Label::PEDESTRIAN || object2.label == Label::PEDESTRIAN)) {
+    return false;
+  }
+
+  const auto dx = object1.x - object2.x;
+  const auto dy = object1.y - object2.y;
+  return dx * dx + dy * dy <= search_distance_2d_sq;
+}
+
+double getCached2dIoU(const CachedObject & source_object, const CachedObject & target_object)
```
mojomex: The function name reads like the IoU value _itself_ would be cached, and that no computation would happen here.

`compute2dIoU` would be clearer (`cached` doesn't really need to be in the name at all imho).

perception/perception_utils/src/iou_bev_nms.cpp
```diff
@@ -30,6 +34,81 @@ namespace
...
+{
+  constexpr double kMinArea = 1.0e-6;
+  constexpr double kMinUnionArea = 0.01;
+
+  if (source_object.area < kMinArea || target_object.area < kMinArea) {
+    return 0.0;
+  }
+  if (boost::geometry::disjoint(source_object.envelope, target_object.envelope)) {
+    return 0.0;
+  }
```
mojomex, line 93: Are these guards really needed? Does `boost::geometry::intersection` not already deal with disjoint geometries?

mojomex (commented):
Nice! Apart from some naming and boilerplate comments, LGTM!

perception/perception_utils/src/iou_bev_nms.cpp
```diff
@@ -82,21 +147,21 @@ std::vector<DetectedObject> IouBevNms::apply(
...
+      const double iou = getCached2dIoU(target_object, source_object);
       max_iou = std::max(max_iou, iou);
       if (iou > params_.iou_threshold) {
         break;
       }
     }
 
     if (max_iou <= params_.iou_threshold) {
-      output_objects.emplace_back(ordered_objects.at(target_i));
+      output_objects.emplace_back(ordered_objects[target_i].object);
```
amadeuszsz, line 166: nit:
In my opinion there is no point to use c-style brackets access, better (and safer) use modern style and let compiler do its job (probably it will apply [LICM](https://xania.org/202512/13-licking-licm) in this loop).

perception/perception_utils/src/iou_bev_nms.cpp
ktro2828: Sounds good! I renamed in https://github.com/autowarefoundation/autoware_universe/pull/13169/commits/99f26233ab60833248a7626ae08cb83f45701864.

perception/perception_utils/src/iou_bev_nms.cpp
ktro2828: I renamed in https://github.com/autowarefoundation/autoware_universe/pull/13169/commits/99f26233ab60833248a7626ae08cb83f45701864.

perception/perception_utils/src/iou_bev_nms.cpp
ktro2828: These guards are not required for IoU correctness, but can improve performance by avoiding unnecessary calculations in `boost::geometry::intersection`, so I feel it's OK to keep them for now.

perception/perception_utils/src/iou_bev_nms.cpp
ktro2828: That makes sense. However, in this loop, both indices are guaranteed to be valid:

- `target_i < ordered_objects.size()`
- `source_i < target_i < ordered_objects.size()`

Therefore, using `operator[]` is safe here and is appropriate for this performance-oriented code. I would prefer `.at()` when the index comes from untrusted input or more complex logic where bounds cannot be easily guaranteed.

mojomex (approved):

amadeuszsz (approved):
LGTM

technolojin (approved):

YoshiRi (approved):
LGTM

PR #13170, opened 2026-08-12
perf(autoware_ptv3): cut PTv3 preprocessing from 13 radix sorts to 2

mojomex:
## Status quo

After #13153, voxels leave deduplication already sorted by their order-0 serialized code — but the pooling-metadata path still runs 12 radix sorts per frame: per pooling stage, one sort to group voxels by parent plus one sort per serialization order (`12 = 4 x (1 + 2)`).

Every pooling sort runs over `max_num_voxels` (default 192k) padded entries regardless of the stage's real occupancy, so per-sort cost is flat at ~70 us: the deepest stage, holding much fewer voxels, costs as much as the initial deduplication.

## What this PR does

Right-shifting a serialized code by `3 x pooling_depth` bits yields the parent voxel's code, and that relationship is monotone, so a level stored in ascending serialized code stays sorted after pooling. Two consequences:

- **Order 0 never needs sorting.** Segments are emitted in increasing parent code, so the pooled level's order-0 ranking is the identity at every level. The old code spent a full 63-bit sort per stage computing it.
- **The other orders never need sorting either.** Every order right-shifts to the same parent partition, so walking a level in order-o rank sequence visits each parent's children contiguously; compacting the run heads yields the pooled level's order-o ranking directly.

The hierarchy is therefore built from prefix scans and run compaction. The only remaining sorts are the per-order sorts at the finest level (skipped entirely for empty inputs): **13 radix sorts -> 2** across the whole preprocessing stage.

The sorted-input precondition is documented on `generateSerializedPoolingMetadata` (`@pre`) and asserted on device in `markPoolingRunsKernel`, which validates every level as it is consumed. The assert compiles out in Release/RelWithDebInfo (`-DNDEBUG` reaches the device front end; verified absent from the PTX and the built `.so`); in Debug it pinpoints an unsorted input instead of silently producing wrong metadata.

## Code changes

Removed kernels: 
* `preparePoolingSortInputKernel`
* `prepareOrderSortInputKernel`
* `kInvalidPoolingKey` padding sentinel

Added kernels: 
* `fillIdentityKernel`
* `prepareFinestLevelOrderSortKernel`
* `markOrderRunsKernel`

Reworked kernels: 
* `markPoolingRunsKernel` (reads codes directly instead of sorted keys)
* `fillPoolingStageKernel` (emits identity gather order and identity order-0)
* `fillOrderAndInverseKernel` (scan-based instead of sort-based)

Buffers: the four pooling-sort scratch buffers become three finest-level sort buffers plus a `[num_orders, max_num_voxels]` order table; the run-flag/run-id buffers remain.

## Benchmark

Measured on the full stack (this PR + #13153, which is perf-neutral on its own):

image

| | before | after |
|---|---|---|
| radix sorts per frame | 13 | **2** |
| preprocessing stage mean | 2.232 ms | **0.936 ms** |
| preprocessing kernel launches | 140 | **62** |
| radix-sort kernel time | 694 us | **116 us** |
| node GPU memory footprint | 11226 MiB | **11188 MiB** |
| total PTv3 runtime | 38.7 ms | **37.0 ms** |

## Tests

The metadata tests could not detect a swapped serialization order before this PR: the fixed-input fixture is almost entirely `y=0`, so it is nearly invariant under transposing x and y. Fixed by replacing the fixture with a level whose two orders rank the voxels differently at every level and where pooling genuinely merges at every stage (10 -> 6 -> 4), by adding `expect_orders_diverge` so the property is asserted rather than assumed, and by adding `MatchesCpuReferenceForEdgeCaseClouds`: hand-crafted layouts vs the CPU `stable_argsort` reference (all eight metadata arrays, every stage), each pinning one boundary with hand-verified level sizes — empty input, a single voxel, a full parent block, no merging at all, mixed run lengths, and a level at exactly `max_num_voxels`.

Verified by mutating `markOrderRunsKernel` to walk order 0's listing for every order: both reference tests fail (the mixed-run-lengths case catches it; its runs sit at different positions under the two orders).

## Validation

Before/after side-by-side comparison showed no visual discrepancies:

[Screencast from 2026年08月06日 20時11分32秒.webm](https://github.com/user-attachments/assets/968e0906-56b6-49fc-a055-f25a28a9ed5e)

Comparing frames numerically was tricky since TensorRT kernels seem to have floating-point nondeterminism I can't influence. Before vs. after results were within run-to-run variance of both before vs. before, and after vs. after, so it's OK.

## Stack

- PR 1, #13153 (**merged**): replaces voxelization hashing with serialized codes; establishes the sorted-order postcondition this PR relies on.
- **PR 2 (this one)**: derives the pooling metadata with prefix scans instead of sorts. Depends on PR 1.
- PR 3, #13155: replaces the encoder's `serialized_code` input with `serialized_order` / `serialized_inverse`. Breaking; depends on this stack.

### DIFF
--- a/perception/autoware_ptv3/include/autoware/ptv3/preprocess/preprocess_kernel.hpp
+++ b/perception/autoware_ptv3/include/autoware/ptv3/preprocess/preprocess_kernel.hpp
@@ -49,7 +49,8 @@ public:
   /**
    * @brief Crops, voxelizes and deduplicates the input cloud.
    *
-   * The emitted voxels are sorted by their order-0 serialized code.
+   * The emitted voxels are sorted by their order-0 serialized code, which
+   * generateSerializedPoolingMetadata requires.
    *
    * @param input_data Input cloud in `input_format` layout.
    * @param input_format Point layout of `input_data`.
@@ -79,10 +80,14 @@ public:
    *
    * @param grid_coord Grid coordinates of the input voxels, laid out [num_voxels, 3].
    * @param serialized_code Codes of the input voxels, laid out [num_orders, num_voxels].
-   * @param num_voxels Number of input voxels.
+   * @param num_voxels Number of input voxels; clamped to max_num_voxels internally.
    * @param stages Output device buffers to fill, one per pooling stage.
    * @param stage_counts Output voxel count per level, laid out [num_stages + 1]; entry 0 is the
-   * input count.
+   * (clamped) input count.
+   * @pre The input voxels are sorted by their order-0 serialized code (`serialized_code` row 0),
+   * as generateFeatures emits them. The coarser levels are derived with prefix scans that rely on
+   * this ordering; an unsorted input silently produces wrong metadata. Asserted on device in
+   * debug builds.
    */
   void generateSerializedPoolingMetadata(
     const std::int32_t * grid_coord, const std::int64_t * serialized_code, std::int64_t num_voxels,
@@ -117,12 +122,15 @@ private:
   cudaEvent_t num_cropped_points_copy_event_;
   cudaEvent_t num_unique_points_copy_event_;
 
-  autoware::cuda_utils::CudaUniquePtr<std::int64_t[]> pooling_keys_d_{nullptr};
-  autoware::cuda_utils::CudaUniquePtr<std::int64_t[]> pooling_sorted_keys_d_{nullptr};
-  autoware::cuda_utils::CudaUniquePtr<std::int64_t[]> pooling_indices_d_{nullptr};
-  autoware::cuda_utils::CudaUniquePtr<std::int64_t[]> pooling_sorted_indices_d_{nullptr};
-  autoware::cuda_utils::CudaUniquePtr<std::int64_t[]> pooling_run_flags_d_{nullptr};
-  autoware::cuda_utils::CudaUniquePtr<std::int64_t[]> pooling_run_ids_d_{nullptr};
+  // Serialization order of the finest level (the input voxels), laid out
+  // [num_orders, num_voxels]. Row 0 is the identity; the remaining rows are the only sorts left in
+  // the pooling-metadata path.
+  autoware::cuda_utils::CudaUniquePtr<std::int64_t[]> finest_level_order_d_{nullptr};
+  autoware::cuda_utils::CudaUniquePtr<std::int64_t[]> order_sort_keys_d_{nullptr};
+  autoware::cuda_utils::CudaUniquePtr<std::int64_t[]> order_sort_sorted_keys_d_{nullptr};
+  autoware::cuda_utils::CudaUniquePtr<std::int64_t[]> order_sort_indices_d_{nullptr};
+  autoware::cuda_utils::CudaUniquePtr<std::int64_t[]> run_flags_d_{nullptr};
+  autoware::cuda_utils::CudaUniquePtr<std::int64_t[]> run_ids_d_{nullptr};
   autoware::cuda_utils::CudaUniquePtr<std::uint8_t[]> pooling_workspace_d_{nullptr};
   std::size_t pooling_workspace_size_{0};
   int code_sort_end_bit_{64};
--- a/perception/autoware_ptv3/include/autoware/ptv3/preprocess/preprocess_kernel.hpp
+++ b/perception/autoware_ptv3/include/autoware/ptv3/preprocess/preprocess_kernel.hpp
@@ -131,11 +131,11 @@ private:
   autoware::cuda_utils::CudaUniquePtr<std::int64_t[]> order_sort_keys_d_{nullptr};
   // Sorted-keys output; CUB requires the buffer, nothing reads it afterwards.
   autoware::cuda_utils::CudaUniquePtr<std::int64_t[]> order_sort_sorted_keys_d_{nullptr};
-  // Sort payload, filled with 0..n-1; sorting it by the keys yields the order's ranking, written
-  // directly into the input_level_order_d_ row.
+  // Filled with 0..n-1 and sorted alongside the keys, which leaves it listing the voxel indices
+  // in ascending code order; written directly into the input_level_order_d_ row.
   autoware::cuda_utils::CudaUniquePtr<std::int64_t[]> order_sort_indices_d_{nullptr};
-  // Run-start flags over the current level: 1 where a new parent segment (or, when deriving a
-  // pooled order, a new rank run) begins, 0 elsewhere.
+  // Run-start flags: 1 where the parent voxel changes while walking a level, either in storage
+  // order (pooling) or as listed by one of its serialization orders; 0 elsewhere.
   autoware::cuda_utils::CudaUniquePtr<std::int64_t[]> run_flags_d_{nullptr};
   // Inclusive scan of run_flags_d_, numbering each element's run; run id - 1 is the pooled-level
   // slot the element scatters to.
--- a/perception/autoware_ptv3/include/autoware/ptv3/preprocess/preprocess_kernel.hpp
+++ b/perception/autoware_ptv3/include/autoware/ptv3/preprocess/preprocess_kernel.hpp
@@ -126,7 +126,8 @@ private:
   // out [num_orders, num_voxels]. Row 0 is the identity; the remaining rows are the only sorts
   // left in the pooling-metadata path.
   autoware::cuda_utils::CudaUniquePtr<std::int64_t[]> input_level_order_d_{nullptr};
-  // Keys for one of those sorts: the input level's codes in the order being ranked.
+  // Keys for one of those sorts: each input voxel's code under the serialization order being
+  // sorted.
   autoware::cuda_utils::CudaUniquePtr<std::int64_t[]> order_sort_keys_d_{nullptr};
   // Sorted-keys output; CUB requires the buffer, nothing reads it afterwards.
   autoware::cuda_utils::CudaUniquePtr<std::int64_t[]> order_sort_sorted_keys_d_{nullptr};
--- a/perception/autoware_ptv3/include/autoware/ptv3/preprocess/preprocess_kernel.hpp
+++ b/perception/autoware_ptv3/include/autoware/ptv3/preprocess/preprocess_kernel.hpp
@@ -122,23 +122,23 @@ private:
   cudaEvent_t num_cropped_points_copy_event_;
   cudaEvent_t num_unique_points_copy_event_;
 
-  // Serialization order of the input level (the deduplicated voxels generateFeatures emits), laid
-  // out [num_orders, num_voxels]. Row 0 is the identity; the remaining rows are the only sorts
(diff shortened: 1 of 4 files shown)


perception/autoware_ptv3/include/autoware/ptv3/ptv3_config.hpp
```diff
@@ -96,18 +95,32 @@ class PTv3Config
...
-    auto max_grid_size = std::max({grid_x_size_, grid_y_size_, grid_z_size_});
+
+    // The serialization depth must cover every coordinate the device-side grid mapping
+    // (floor(p / size) - floor(min_range / size), see gridCoord) can emit. That is one more than
+    // the rounded cell count above when a range boundary is not voxel-aligned: [0.5, 16.5) with
+    // unit voxels emits 0..16. Sizing the depth from the rounded count would then drop the extra
+    // coordinate's top Morton bit and silently merge its voxels with coordinate 0's. The largest
+    // coordinate is attained at the largest float below max_range, since the crop accepts
+    // min_range <= p < max_range and float division is monotonic.
+    const auto effective_grid_cells =
```
amadeuszsz: Please double check line 380 regarding cell size, might be necessary to use effective grid cells

mojomex: Confirmed! Fixed in the base PR, see https://github.com/autowarefoundation/autoware_universe/pull/13153#issuecomment-5289915444

perception/autoware_ptv3/include/autoware/ptv3/ptv3_config.hpp
```diff
@@ -96,18 +95,32 @@ class PTv3Config
...
+      [](const float min_range, const float max_range, const float size) {
+        const float min_coord = std::floor(min_range / size);
+        const float max_coord = std::floor(std::nextafter(max_range, min_range) / size);
+        return static_cast<std::int64_t>(max_coord - min_coord) + 1;
+      };
+    const auto max_effective_cells = std::max(
+      {effective_grid_cells(min_x_range_, max_x_range_, voxel_x_size_),
+       effective_grid_cells(min_y_range_, max_y_range_, voxel_y_size_),
+       effective_grid_cells(min_z_range_, max_z_range_, voxel_z_size_)});
```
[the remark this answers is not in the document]
mojomex: Confirmed! Fixed in the base PR, see https://github.com/autowarefoundation/autoware_universe/pull/13153#issuecomment-5289915444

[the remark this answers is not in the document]
mojomex: Confirmed! Fixed in the base PR, see https://github.com/autowarefoundation/autoware_universe/pull/13153#issuecomment-5289915444

perception/autoware_ptv3/lib/preprocess/preprocess_kernel.cu
```diff
@@ -67,9 +67,10 @@ PreprocessCuda::PreprocessCuda(const PTv3Config & config, cudaStream_t stream)
   thrust::device_ptr<std::uint32_t> idx_ptr(code_indices_d_.get());
   thrust::sequence(policy, idx_ptr, idx_ptr + config_.cloud_capacity_, 0);
 
-  // Serialized codes occupy 3 * serialization_depth_ bits. std::max guards
-  // serialization_depth_ == 0 (CUB requires end_bit > begin_bit). The workspace query below uses
-  // the full 64 bits, which upper-bounds the scratch needed for any narrower range.
+  // Serialized codes occupy 3 * serialization_depth_ bits, and pooling only right-shifts them, so
+  // this bound holds for every stage. std::max guards serialization_depth_ == 0 (CUB requires
+  // end_bit > begin_bit). Workspace queries below use the full 64 bits, which upper-bounds the
+  // scratch needed for any narrower range.
```
mojomex:
```suggestion
  // end_bit > begin_bit).
```

perception/autoware_ptv3/include/autoware/ptv3/preprocess/preprocess_kernel.hpp
```diff
@@ -117,12 +122,15 @@ class PreprocessCuda
...
-  autoware::cuda_utils::CudaUniquePtr<std::int64_t[]> pooling_run_ids_d_{nullptr};
+  // Serialization order of the finest level (the input voxels), laid out
+  // [num_orders, num_voxels]. Row 0 is the identity; the remaining rows are the only sorts left in
+  // the pooling-metadata path.
+  autoware::cuda_utils::CudaUniquePtr<std::int64_t[]> finest_level_order_d_{nullptr};
+  autoware::cuda_utils::CudaUniquePtr<std::int64_t[]> order_sort_keys_d_{nullptr};
+  autoware::cuda_utils::CudaUniquePtr<std::int64_t[]> order_sort_sorted_keys_d_{nullptr};
+  autoware::cuda_utils::CudaUniquePtr<std::int64_t[]> order_sort_indices_d_{nullptr};
+  autoware::cuda_utils::CudaUniquePtr<std::int64_t[]> run_flags_d_{nullptr};
+  autoware::cuda_utils::CudaUniquePtr<std::int64_t[]> run_ids_d_{nullptr};
```
mojomex: Document each of these with a 1-2 liner - what data is contained, or how is it used / why is it needed?

perception/autoware_ptv3/lib/preprocess/preprocess_kernel.cu
```diff
@@ -354,52 +357,83 @@ __global__ void computeGridCoordsAndSerializationKernel(
...
 
+/**
+ * @brief Marks the first child of every parent run.
+ *
+ * @pre The input level is ascending in order-0 code (asserted below), so the parent code
+ * (code >> 3 * pooling_depth) is non-decreasing and each parent's children are already
+ * contiguous; comparing against the previous element finds the run starts without sorting.
+ *
+ * @param serialized_code_in Input level's codes, laid out [num_orders, input_count]; only the
+ * order-0 row is read.
```
mojomex, line 411: `finest` and `input level` terminology is mixed throughout. Let's go with input level as it is more intuitive.

perception/autoware_ptv3/lib/preprocess/preprocess_kernel.cu
mojomex: Applied in 366ebd721.

perception/autoware_ptv3/include/autoware/ptv3/preprocess/preprocess_kernel.hpp
mojomex: Done in 366ebd721 — each buffer now has a 1-2 line comment.

perception/autoware_ptv3/lib/preprocess/preprocess_kernel.cu
mojomex: Standardized on "input level" in 366ebd721: `input_level_order_d_`, `prepareInputLevelOrderSortKernel`, and the related comments. The "finest to deepest" stage-ordering comments in `ptv3_trt.hpp`/`ptv3_config.hpp`/`ptv3_trt.cpp` are from the already-merged #13153, so they are left out of this PR's diff.

perception/autoware_ptv3/lib/preprocess/preprocess_kernel.cu
```diff
@@ -354,52 +356,129 @@ __global__ void computeGridCoordsAndSerializationKernel(
...
 
-  const auto key = sorted_keys[idx];
-  run_flags[idx] = key != kInvalidPoolingKey && (idx == 0 || key != sorted_keys[idx - 1]) ? 1 : 0;
+  const auto input_count = stage_counts_in[stage_index];
+  if (idx >= input_count) {
+    run_flags_out[idx] = 0;
+    return;
+  }
+
+  assert(idx == 0 || serialized_code_in[idx - 1] < serialized_code_in[idx]);
```
amadeuszsz, line 434: This check seems crucial, was it your intention to keep it only in `DEBUG`?

perception/autoware_ptv3/test/serialized_pooling_metadata_test.cpp
```diff
@@ -278,12 +329,118 @@ TEST_F(SerializedPoolingMetadataTest, DetectionGridCoord3StaysInsideBevGrid)
   }
 }
 
+// Randomized sweep against the CPU reference: the scan-based derivation assumes pooling preserves
+// the serialization order, so exercise that across many occupancy patterns.
+TEST_F(SerializedPoolingMetadataTest, MatchesCpuReferenceForRandomizedClouds)
```
amadeuszsz: Not sure if understand correctly this test, but up to 32 voxels with 64 grid size makes grid very sparse - very likely no common parent and therefore pointless test. You can make more voxels or smaller grid or make sure voxels are crowded in the close region.

amadeuszsz (changes requested):

perception/autoware_ptv3/lib/preprocess/preprocess_kernel.cu
mojomex: This `assert()` checks the function's documented `@pre`condition, i.e. that the function's contract has not been violated. This contract is [covered by unit tests](https://github.com/autowarefoundation/autoware_universe/pull/13170/changes/BASE..d1cd995789f0a6f7d601b47406c8f45adc920ba9#diff-3bbba7b55acb8fbdfacaeaada86a266071682ccbe445b70af0bc740d4850d628R259).

Even if we checked during runtime, there would be no sensible way to recover from the error (an error which in the first place only happens if we programmed it wrong), so I chose to go with assert only.

[C++ core guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#i5-state-preconditions-if-any) also recommends stating pre-conditions in code, with their `Expects()` statement boiling down to an `assert()` too.

perception/autoware_ptv3/test/serialized_pooling_metadata_test.cpp
mojomex: There's only [1 * 16 * 32](https://github.com/autowarefoundation/autoware_universe/pull/13170/changes/BASE..d1cd995789f0a6f7d601b47406c8f45adc920ba9#diff-d44eb052fdb1942d2237c61e20e61c41d66a530d8c68373bb1bca0d05d4d102dL181) voxels, and the test explicitly mitigates the case where all voxels are alone in their parent.

However, using randomness to find a case is more convoluted than just hand-crafting one, so I'll rewrite this to be easily verifiable.

mojomex: Addressed in 2cf203a - it tests an explicit list of edge cases now (empty, full parent, full grid, mixed parent occupancy, etc.). It's grown a bit but TEST_P would add even more boilerplate, so I decided to keep it as one function. Let me know if you prefer a split.

amadeuszsz (approved):
LGTM!

PR #13177, opened 2026-08-13
refactor(lidar_centerpoint): replace NMS with perception_utils::IouBevNms

ktro2828:
## Description

This pull request refactors the non-maximum suppression (NMS) logic in the `autoware_lidar_centerpoint` package to use the shared `perception_utils::IouBevNms` implementation instead of its own local NMS code. The local NMS implementation and its associated test are removed, and all references are updated to use the new shared utility.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

I compared the number of output boxes between before and after replacing NMS, and got the same result.

lidar_centerpoint_bbox_count_histograms

lidar_centerpoint_paired_bbox_counts

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


mojomex (approved):
Code LGTM! Did not check for behavioral/numeric equivalence.

This PR needs to be merged after #13180.

technolojin (approved):

PR #13179, opened 2026-08-13
refactor(camera-streampetr): replace NMS with perception_utils::IouBevNms

ktro2828:
## Description

This pull request refactors the non-maximum suppression (NMS) logic in the `autoware_camera_streampetr` package to use the shared `perception_utils::IouBevNms` implementation instead of a custom, local version. This change reduces code duplication, streamlines maintenance, and leverages a common utility across projects. The update includes removing the local NMS code, updating dependencies, and modifying related code to use the new interface.

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

I compared the number of output boxes between before and after replacing NMS, and got the same result.

camera_streampetr_bbox_count_histograms

camera_streampetr_paired_bbox_counts

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.


mojomex (approved):
Code LGTM! Did not check for behavioral/numeric equivalence.

KSeangTan (approved):
Thanks! LGTM!

PR #13181, opened 2026-08-13
feat(multi_object_tracker): tracker batch trigger on stale target stream

YoshiRi:
## Description
  Fix input batch processing when the current latency-reference input stream becomes stale.

  The multi object tracker selects the input stream with the largest latency as the target stream for batching. If that target stream stops receiving messages, the batch window could stop advancing, even when
  other input streams are still fresh. This could result in decreasing output topic Hz and linearly increasing `debug/input_latency`.

  This PR updates the channel timing optimization so that stale input streams are excluded from target stream selection. When the previous target stream becomes stale, the input manager can fail over to a
  fresh stream and continue advancing the batch window.

  The periodic timer was also updated based on review feedback:
  - It no longer calls `processObjects()`.
  - It only refreshes the channel timing optimizer.
  - Therefore, it does not introduce an unexpected publish path or change publish timing.

  `delay_compensation` remains independent of this fix. It only affects the publish/export reference time after batch processing; the stale target stream failover is handled before that in `InputManager`.

  ## Changes

  - Add freshness-aware channel timing optimization.
  - Expose `InputManager::optimizeChannelTimings(now)` for the optimizer timer and deterministic tests.
  - Rename the timer callback to `onChannelOptimizerTimer()` to clarify that it only updates channel selection.
  - Make target stream selection tests deterministic by injecting synthetic `rclcpp::Time` instead of using wall-clock sleeps.

## Related links

**Parent Issue:**

- https://star4.slack.com/archives/C076ZGRC15X/p1786086049624789?thread_ts=1786015513.168119&cid=C076ZGRC15X

## How was this PR tested?

  ```bash
  colcon build --symlink-install --cmake-args -DCMAKE_BUILD_TYPE=Release --parallel-workers 6 --packages-up-to autoware_multi_object_tracker

  /home/yoshiri/autoware/build/autoware_multi_object_tracker/test_multi_object_tracker --gtest_filter='InputManagerTargetSelection.*'
```

  Result: 3/3 passed.

```
  colcon test --packages-select autoware_multi_object_tracker --event-handlers console_direct+
```
  Result: gtest passed, 54/54 tests passed.

  Known existing issue: package-level colcon test still reports copyright failures because CONTRIBUTING.md and LICENSE are not found.

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

None.

### DIFF
--- a/perception/autoware_multi_object_tracker/src/processor/input_manager.cpp
+++ b/perception/autoware_multi_object_tracker/src/processor/input_manager.cpp
@@ -332,34 +332,57 @@ void InputManager::getObjectTimeInterval(
   }
 }
 
-void InputManager::optimizeTimings()
+bool InputManager::isStreamFresh(const InputStream & input_stream, const rclcpp::Time & now) const
 {
-  double max_latency_mean = 0.0;
-  uint selected_stream_idx = 0;
-  double selected_stream_latency_std = 0.1;
-  double selected_stream_interval = 0.1;
-  double selected_stream_interval_std = 0.01;
+  if (!input_stream.isTimeInitialized()) {
+    return false;
+  }
+
+  double latency_mean, latency_var, interval_mean, interval_var;
+  input_stream.getTimeStatistics(latency_mean, latency_var, interval_mean, interval_var);
+
+  constexpr double min_freshness_timeout = 0.3;     // [s]
+  constexpr double interval_timeout_multiplier = 3.0;
+  const double expected_interval = interval_mean > 1e-3 ? interval_mean : target_stream_interval_;
+  const double freshness_timeout =
+    std::max(min_freshness_timeout, interval_timeout_multiplier * expected_interval);
+
+  return (now - input_stream.getLatestMessageTime()).seconds() <= freshness_timeout;
+}
+
+void InputManager::optimizeTimings(const rclcpp::Time & now)
+{
+  double max_latency_mean = -1.0;
+  uint selected_stream_idx = target_stream_idx_;
+  double selected_stream_latency_std = target_stream_latency_std_;
+  double selected_stream_interval = target_stream_interval_;
+  double selected_stream_interval_std = target_stream_interval_std_;
+  bool selected_stream = false;
 
   {
     // ANALYSIS: Get the streams statistics
-    // select the stream that has the maximum latency
+    // select the fresh stream that has the maximum latency
     double latency_mean, latency_var, interval_mean, interval_var;
     for (const auto & input_stream : input_streams_) {
-      if (!input_stream->isTimeInitialized()) continue;
+      if (!isStreamFresh(*input_stream, now)) continue;
       input_stream->getTimeStatistics(latency_mean, latency_var, interval_mean, interval_var);
-      if (latency_mean > max_latency_mean) {
+      if (!selected_stream || latency_mean > max_latency_mean) {
         max_latency_mean = latency_mean;
         selected_stream_idx = input_stream->getIndex();
         selected_stream_latency_std = std::sqrt(latency_var);
         selected_stream_interval = interval_mean;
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Review by technolojin — CHANGES_REQUESTED** [review]
the new trigger onBatchTimer calls processObjects().
Instead of processObjects(), only the channel optimizer shall be called.

Risks
- unexpected publish timing : onBatchTimer shall not call publish()
- complex callback structure

Also, the test could be flaky (i.e. test can fail with normal code by chance). Please make test not depend on time functions.

File: perception/autoware_multi_object_tracker/src/processor/input_manager.cpp [hunk 1]
```diff
@@ -332,34 +345,57 @@ void InputManager::getObjectTimeInterval(
...
+  double latency_mean, latency_var, interval_mean, interval_var;
+  input_stream.getTimeStatistics(latency_mean, latency_var, interval_mean, interval_var);
+
+  constexpr double min_freshness_timeout = 0.3;  // [s]
+  constexpr double interval_timeout_multiplier = 3.0;
+  const double expected_interval = interval_mean > 1e-3 ? interval_mean : target_stream_interval_;
+  const double freshness_timeout =
+    std::max(min_freshness_timeout, interval_timeout_multiplier * expected_interval);
+
+  return (now - input_stream.getLatestMessageTime()).seconds() <= freshness_timeout;
```
technolojin: adaptive threshold can cause unexpected condition relaxing.
in other word, gradually increased interval_mean can rate-dropped channel as 'fresh'
however, interval_mean needed to be considered to compensate natural message arrival delay.

```suggestion
  constexpr double freshness_margin = 0.2;  // [s]
  const double expected_interval = interval_mean > 1e-3 ? interval_mean : target_stream_interval_;
  const double freshness_timeout = std::min(expected_interval, freshness_margin) + freshness_margin;

  const double elapsed = (now - input_stream.getLatestMessageTime()).seconds();
  return 0.0 <= elapsed && elapsed <= freshness_timeout;
```

it forces longer than `freshness_timeout_threshold * 2.0` message interval channel as not-fresh channel.

perception/autoware_multi_object_tracker/src/processor/input_manager.cpp
```diff
@@ -383,10 +419,10 @@ bool InputManager::getObjects(
   rclcpp::Time object_earliest_time;
   getObjectTimeInterval(now, object_latest_time, object_earliest_time);
 
-  // Optimize the target stream, latency, and its band
+  // Optimize the target stream, latency, and its band.
   // The result will be used for the next time, so the optimization is after getting the time
-  // interval
-  optimizeTimings();
+  // interval.
+  optimizeChannelTimings(now);
```
technolojin: I suggest to remove optimizeChannelTimings call here.  
As result, optimizeChannelTimings is only called by the new timer callback.

technolojin (changes requested):
I found risks of getObjectTimeInterval - getObjects algorithm. It may exist before this PR.

* Sudden increase of target_stream_latency_ can cause old/duplicated que object export (via getObjects). with old objects, the trackers will not update older than tracker time. As result, trackers can stale and can lost tracking objects. 
suggest to implement gradual target_stream_latency_ shift

* target_stream_idx_ may can be flipped frequently. suggest to introduce hysteresis

* even more, https://docs.autoware.org/main/design/autoware-architecture-v1/components/perception/reference_implementation/#architecture-for-object-recognition says the tracker triggered by a priority channel, not a largest latency channel. At least, may need to provide options to users.

technolojin (changes requested):
I created fixes to a PR
https://github.com/autowarefoundation/autoware_universe/pull/13259

technolojin (approved):
LGTM

PR #13190, opened 2026-08-14
feat(map_based_prediction): add vegetation path cutting

satoshi-ota:
## Description

This pull request adds vegetation-aware path cutting to the VRU prediction flow in `autoware_map_based_prediction`. Predicted paths are first shortened by fences, then shortened again when their swept object footprint crosses vegetation polygons from the lanelet map. The most important changes are:

- Adds a vegetation module that builds a vegetation-only lanelet map layer and cuts predicted paths before vegetation crossings.
- Skips crosswalk candidate paths that cross vegetation before reaching the crosswalk.
- Uses bounding-box prechecks before swept-footprint intersection tests to reduce unnecessary precise geometry checks.

## Related links

- Cherry-picked from:
  - https://github.com/tier4/autoware_universe/pull/3113
  - https://github.com/tier4/autoware_universe/pull/3221
  - https://github.com/tier4/autoware_universe/pull/3323

**Parent Issue:**

- Link

## How was this PR tested?

WIP

## Notes for reviewers

None.

## Interface changes

None.

## Effects on system behavior

VRU predicted paths may be shortened when their swept object footprint intersects vegetation areas in the lanelet map.


perception/autoware_map_based_prediction/lib/predictor_vru/vegetation.cpp
```diff
@@ -0,0 +1,205 @@
...
+      continue;
+    }
+    for (auto i = 0UL; i + 1 <= last_idx; ++i) {
+      const auto swept_polygon =
+        convexPolygonCoveringSegmentFootprints(path.at(i), path.at(i + 1), object_shape);
+      if (boost::geometry::intersects(swept_polygon, vegetation_polygon_2d)) {
+        return i;
+      }
+    }
+  }
```
technolojin: Comment 1

The polygon loop is outside the segment loop, so the returned index belongs to whichever polygon the
R-tree yields first, not to the first crossing along the path: with two polygons over the path a
crossing on segment 5 can be returned while the polygon crossed on segment 1 is never examined, and
the path stays running through a bush.

Segment 0's swept polygon also covers the object's current footprint plus one sample step (>= 0.7 m at
the velocity floor), so an object standing under a mapped tree or beside a hedge reports a crossing at
index 0 — dropping every crosswalk path at `predictor_vru.cpp:395-399` even when it walks away from the
vegetation, and truncating the straight path to one pose. A path leaving a vegetation area is not a
path entering one.

One restructure covers both, and stops the swept hulls being rebuilt per candidate polygon:

```suggestion
  const size_t last_idx = path.size() - 1;
  const auto path_bbox = buildPathFootprintBBox(path, object_shape, last_idx);
  const auto initial_footprint = autoware_utils_geometry::to_polygon2d(path.front(), object_shape);

  std::vector<Polygon2d> crossed_polygons;
  for (const auto & vegetation_polygon : vegetation_polygons) {
    const auto vegetation_polygon_2d = toPolygon2d(vegetation_polygon);
    if (!boost::geometry::intersects(path_bbox, vegetation_polygon_2d)) {
      continue;
    }
    // Object already overlapping vegetation: its path leaves the area, so no crossing is reported.
    if (boost::geometry::intersects(initial_footprint, vegetation_polygon_2d)) {
      return std::nullopt;
    }
    crossed_polygons.push_back(vegetation_polygon_2d);
  }

  for (auto i = 0UL; i + 1 <= last_idx; ++i) {
    const auto swept_polygon =
      convexPolygonCoveringSegmentFootprints(path.at(i), path.at(i + 1), object_shape);
    for (const auto & vegetation_polygon_2d : crossed_polygons) {
      if (boost::geometry::intersects(swept_polygon, vegetation_polygon_2d)) {
        return i;
      }
    }
  }
```

perception/autoware_map_based_prediction/lib/predictor_vru/vegetation.cpp
```diff
@@ -0,0 +1,205 @@
...
+    for (auto i = 0UL; i + 1 <= last_idx; ++i) {
+      const auto swept_polygon = convexPolygonCoveringSegmentFootprints(
+        predicted_path.path.at(i), predicted_path.path.at(i + 1), object_shape);
+      if (boost::geometry::intersects(swept_polygon, vegetation_polygon_2d)) {
+        return true;
+      }
+    }
+  }
+  return false;
+}
```
technolojin, line 139: Comment 2

Same function as `findVegetationCrossingIndex` with `last_idx` clamped to `arrival_index`, so it
carries the same two problems. Suggest not fixing them twice: add a `last_idx` parameter there and make
this a wrapper. `PredictedPathWithArrivalIndex` derives from `PredictedPath`, so it passes straight
through; the caller at line 199 supplies `predicted_path.path.size() - 1`.

```suggestion
bool doesPathCrossVegetation(
  const PredictedPathWithArrivalIndex & predicted_path,
  const autoware_perception_msgs::msg::Shape & object_shape,
  const lanelet::ConstPolygons3d & vegetation_polygons)
{
  if (predicted_path.path.size() < 2) {
    return false;
  }
  const size_t last_idx = std::min(predicted_path.arrival_index, predicted_path.path.size() - 1);
  return findVegetationCrossingIndex(predicted_path, object_shape, vegetation_polygons, last_idx).has_value();
}
```

perception/autoware_map_based_prediction/lib/predictor_vru/vegetation.cpp
```diff
@@ -0,0 +1,205 @@
...
+  for (const auto & pt : predicted_path.path) {
+    predicted_path_ls.emplace_back(pt.position.x, pt.position.y);
+  }
+  const auto candidates =
+    vegetation_layer_->polygonLayer.search(lanelet::geometry::boundingBox2d(predicted_path_ls));
+  const std::optional<size_t> crossing_index =
+    findVegetationCrossingIndex(predicted_path, object_shape, candidates);
+  if (crossing_index) {
+    cut_path.path.resize(std::min(*crossing_index + 1, cut_path.path.size()));
+  }
```
technolojin: Comment 4

A crossing on segment 0 leaves a single pose: zero length with a non-zero `time_step`, malformed for
consumers of `PredictedPath`. Comment 1 removes the object-starts-inside cause, but segment 0 still
reaches ~0.7 m ahead, so index 0 stays reachable.

```suggestion
  if (crossing_index) {
    cut_path.path.resize(std::clamp<size_t>(*crossing_index + 1, 2, cut_path.path.size()));
  }
```

technolojin (changes requested):
better as one follow-up PR once this and #13203 have merged:

All of these are in `lib/predictor_vru/vegetation.cpp` unless stated:

- cache the converted `Polygon2d`s and their envelopes in `buildFromMap` (`:143-160`), with
  `bg::is_valid` and a ">= 3 distinct points" check at load time
- for `CYLINDER` shapes the swept footprint is the segment buffered by `r`, so
  `bg::distance(segment, polygon) <= r` replaces the convex hulls and is more accurate than the 6-gon
  circle in `autoware_utils_geometry/boost_polygon_utils.cpp:143-159`
- `buildPathFootprintBBox` (`:73-85`) pays N footprint constructions for a filter looser than the
  centreline box it follows; how much it is worth keeping depends on the decision in the inline comment
  at `:178-179`
- skip the R-tree when the map has no vegetation — `createMap({})` at `:159` is non-null
- `lib/predictor_vru/fence.cpp` and this module rebuild the same `BasicLineString2d` from the same path;
  hoist it into the caller in `lib/predictor_vru/predictor_vru.cpp:246-250`

perception/autoware_map_based_prediction/lib/predictor_vru/vegetation.cpp
satoshi-ota: Thanks for the review.

You are completely right about the loop nesting. The returned index was tied to whichever polygon the R-tree yielded first rather than to the first crossing along the path. Fixed in 45fcb4f8b2, please take a look.

As for the case where the initial footprint already overlaps vegetation, I had not defined the intended behavior. Thinking it over, an object that is already on vegetation can move regardless of whether the vegetation is there, so not cutting the predicted path by vegetation in that case does not feel wrong to me as a specification.

So I ended up implementing exactly what you suggested.

perception/autoware_map_based_prediction/lib/predictor_vru/vegetation.cpp
satoshi-ota: Fixed in https://github.com/autowarefoundation/autoware_universe/pull/13190/commits/1a00b04a415faef37ce5b6ba7fee26609e86e6a8.

technolojin (approved):

PR #13191, opened 2026-08-14
refactor(autoware_tensorrt_common): rename precision "as-is" to "strongly-typed"

mojomex:
## Description

Renames the TensorRT precision value added in #13160 from `"as-is"` to `"strongly-typed"`.

`"as-is"` describes how the artifact is treated; the setting's actual effect is the build mode — the network is created with `NetworkDefinitionCreationFlag::kSTRONGLY_TYPED` and every tensor precision comes from the model. Sitting in an enum next to `fp32`/`fp16`/`int8`, `"as-is"` also reads as a precision, which it is not. `"strongly-typed"` says what happens and matches the accessor (`isStronglyTyped()`) and the log lines already in the code.

No functional change: only the accepted string differs.

## Related links

- Introduced the value: #13160 (merged)
- First consumer, updated to the new name: #13159

## How was this PR tested?

`colcon build` of `autoware_tensorrt_common` (x86, TensorRT 10.8). Building PTv3 engines strongly typed through #13159 with the renamed value; behavior is identical to `"as-is"` before the rename.

## Notes for reviewers

The value has no users in `main`: it was added by #13160 and its first consumer (#13159) is still open, so this rename breaks no existing configuration. Doing it now avoids a deprecation path later.

## Interface changes

`TrtCommonConfig::precision` accepts `"strongly-typed"` instead of `"as-is"`. `"fp32"`, `"fp16"` and `"int8"` are unchanged. No signature changes.

## Effects on system behavior

None. No in-tree consumer passes the old value.

### DIFF
--- a/perception/autoware_tensorrt_common/include/autoware/tensorrt_common/tensorrt_common.hpp
+++ b/perception/autoware_tensorrt_common/include/autoware/tensorrt_common/tensorrt_common.hpp
@@ -317,8 +317,8 @@ public:
   void printProfiling() const;
 
   /**
-   * @brief Whether the network is built strongly typed (precision "as-is"): tensor precisions
-   * are taken from the model itself.
+   * @brief Whether the network is built strongly typed (precision "strongly-typed"): tensor
+   * precisions are taken from the model itself.
    *
    * @return Whether the network is strongly typed.
    */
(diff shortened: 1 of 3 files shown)


amadeuszsz (approved):
LGTM, thanks

PR #13193, opened 2026-08-14
feat(autoware_carla_interface): allow explicit carla_map override in launch

hakuturu583:
## Description

Add an optional `carla_map` launch argument that, when non-empty, overrides the CARLA map/level name that is otherwise derived from the basename of `map_path`. This is useful when the CARLA level name does not match the Autoware map directory name.

Map-name resolution is factored into a `carla_map_resolved` `<let>`. When the `carla_map` argument is empty (the default), the name is derived from `map_path` exactly as before, so behavior is unchanged unless the user opts in.

This is one of a series of small enabler PRs toward CARLA 0.10.0 support, and is a launch-only change.

## How was this PR tested?

- Verified the launch XML is well-formed.
- Verified that, when `carla_map` is empty (the default), the value derived from `map_path` is identical to the previous behavior.

## Backward compatibility

- `carla_map` defaults to `""`. Unless the user opts in, the resolved result is the same as today and the `carla_map` value passed to the node is unchanged.

## Related PR

- autowarefoundation/autoware_launch#1948 — the downstream autoware_launch change that declares and forwards this argument from `e2e_simulator.launch.xml` (Draft; to be merged after this PR).


Max-Bin (approved):
Checked and LGTM

PR #13198, opened 2026-08-14
feat(autoware_carla_interface): add map_origin_x/y offset between CARLA and map frame

hakuturu583:
## Description

Add optional `map_origin_x` / `map_origin_y` parameters that translate between CARLA's local world origin and the Autoware map frame origin. The offset is applied in `carla_location_to_ros_point` (CARLA -> ROS) and its inverse `ros_pose_to_carla_transform` (ROS -> CARLA), and is passed from the ego `pose()` publisher and from `initialpose_callback`.

This is needed for CARLA levels that were authored with their own local origin instead of matching the lanelet2/PCD map frame (e.g. a real-world-georeferenced custom map). Stock CARLA towns are aligned by default, so the offsets default to 0.0.

This is one of a series of small enabler PRs toward CARLA 0.10.0 support.

## How was this PR tested?

- Confirmed `colcon build --packages-select autoware_carla_interface --symlink-install` passes in the jazzy dev container.

## Backward compatibility

- With the default 0.0 offsets the arithmetic is the identity transform, so the published pose and the applied initial pose are exactly identical to the previous behavior.

## Related PR

- autowarefoundation/autoware_launch#1948 — the downstream autoware_launch change that declares and forwards this argument from `e2e_simulator.launch.xml` (Draft; to be merged after this PR).


Max-Bin (approved):
LGTM

Max-Bin (commented):
LGTM

Max-Bin (approved):

PR #13201, opened 2026-08-14
feat(ptv3): enable to specify segmentation class remapping from config

ktro2828:
## Description

This pull request requires https://github.com/autowarefoundation/autoware_core/pull/1382.

This pull request introduces a configurable mapping from segmentation class names to consolidated `PointCloudClassification` values, making class consolidation a runtime configuration rather than a compile-time constant, [which is requested in the previous PR](https://github.com/autowarefoundation/autoware_universe/pull/13044#pullrequestreview-4844003131).

It updates the configuration, code, and documentation to support and enforce this mapping, improving flexibility and robustness in model class handling.

### Code Changes: Mapping and Lookup Logic
* Introduced the `declare_class_remap` function to declare and resolve the `class_remap` parameters in class name order, ensuring every class is mapped and reporting configuration errors. [[1]](diffhunk://#diff-67ecd565ef9571e4ca98e6a3b0d7017c6033d276921632dbb6889a701bac1f68R72-R87) [[2]](diffhunk://#diff-9925ade3c7efa1b0c2305e62dd0fd67ca9c175d32d9e5a1c97916478c0cd9d45R106-R129) [[3]](diffhunk://#diff-355588045e2b2edd516be487a13a4c98bb86a7d0e0824536e611f9667ee0ae27R99-R100)
* Added `class_id_to_classification_` to `PTv3Config`, with a new static method to build the lookup table from class names and remaps, replacing the previous hardcoded mapping logic. [[1]](diffhunk://#diff-7f6e7a336fc5c8c6f5eea6a40f661503b74c0a62c54b429579e42dfe95c26326R49) [[2]](diffhunk://#diff-7f6e7a336fc5c8c6f5eea6a40f661503b74c0a62c54b429579e42dfe95c26326R122-R123) [[3]](diffhunk://#diff-7f6e7a336fc5c8c6f5eea6a40f661503b74c0a62c54b429579e42dfe95c26326R283-R315) [[4]](diffhunk://#diff-7f6e7a336fc5c8c6f5eea6a40f661503b74c0a62c54b429579e42dfe95c26326R435-R436)
* Updated the CUDA postprocessing code to use the new lookup table from configuration, removing the old hardcoded mapping and lookup generation. [[1]](diffhunk://#diff-37a51e10daaf881a8c20e2446dd011d5be5434e34e42ee79e93db1568671a6d5L23-L128) [[2]](diffhunk://#diff-37a51e10daaf881a8c20e2446dd011d5be5434e34e42ee79e93db1568671a6d5R297-R303)

## Related links

**Parent Issue:**

- Link

## How was this PR tested?

The following pictures show the results of `/ptv3/segmentation` in before/after.

|Before|After|
|---|---|
|Screenshot from 2026-08-17 14-29-18 | Screenshot from 2026-08-17 14-24-44 |

## Notes for reviewers

None.

## Interface changes

None.

#### Additions and removals

| Added | `class_remap`   | `key: value string` | `...`         | Class remapping from ptv3 `class_names` to `PointCloudClassification` |

## Effects on system behavior

None.


perception/autoware_ptv3/test/ptv3_config_test.cpp
```diff
@@ -73,5 +90,79 @@ TEST(PTv3ConfigTest, RejectsInvalidDetectionThresholdTables)
     std::runtime_error);
 }
 
+TEST(PTv3ConfigTest, BuildsClassificationLutInClassNameOrder)
+{
+  using autoware::point_types::PointCloudClassification;
+
+  const auto config = makeSegmentationConfig(
+    {"car", "traffic_cone", "drivable_flat"},
+    {{"car", "CAR"}, {"traffic_cone", "HAZARD"}, {"drivable_flat", "FLAT_SURFACE"}});
```
mojomex: This test could silently pass even if not class name order, but remapping order was used. Please mix up the remapping order here so the test proves what it says.

perception/autoware_ptv3/include/autoware/ptv3/ptv3_config.hpp
```diff
@@ -274,6 +281,35 @@ class PTv3Config
...
+    const std::vector<std::string> & class_names,
+    const std::unordered_map<std::string, std::string> & class_remaps)
+  {
+    std::vector<std::uint8_t> lut;
+    lut.reserve(class_names.size());
+    for (const auto & class_name : class_names) {
+      const auto it = class_remaps.find(class_name);
+      if (it == class_remaps.end()) {
+        throw std::runtime_error("class_remaps has no entry for class name '" + class_name + "'.");
+      }
```
mojomex: (nit) bonus points if the error contains _all_ missing class names so the user doesn't have to fix iteratively.

perception/autoware_ptv3/CMakeLists.txt
mojomex: Since this instantiates ROS nodes now, this should be isolated to prevent interference with other running ROS tests. Please change to [`ament_add_ros_isolated_gtest`](https://github.com/ros2/ament_cmake_ros/blob/master/ament_cmake_ros/cmake/ament_add_ros_isolated_gtest.cmake).

mojomex (commented):
Basically good to go, some small testing comments only!

perception/autoware_ptv3/include/autoware/ptv3/ptv3_config.hpp
ktro2828: Sounds good. Because we check if all `class_names` are covered by `class_remap` in `declare_class_remap(...)` function, I updated that function in https://github.com/autowarefoundation/autoware_universe/pull/13201/commits/94ce2850b4fd70c4906cf49acb6b8254670604c8.

For example:

```shell
$ ros2 launch autoware_ptv3 ptv3.launch.xml
[INFO] [launch]: All log files can be found below /home/kotarouetake/.ros/log/2026-08-18-16-53-08-842902-dpc2409509-519730
[INFO] [launch]: Default logging verbosity is set to INFO
[INFO] [autoware_ptv3_node-1]: process started with pid [519733]
[autoware_ptv3_node-1] terminate called after throwing an instance of 'std::runtime_error'
[autoware_ptv3_node-1]   what():  segmentation3d.class_remap is missing entries for segmentation3d.class_names: [barrier, drivable_flat, building, static_clutter].
```

perception/autoware_ptv3/test/ptv3_config_test.cpp
ktro2828: Addressed in https://github.com/autowarefoundation/autoware_universe/pull/13201/commits/b2a53fc7d5d6087f7bfd25749c694f78392bbca8.

perception/autoware_ptv3/CMakeLists.txt
ktro2828: Fixed in https://github.com/autowarefoundation/autoware_universe/pull/13201/commits/8bcd768a55222ba9b080c86cdd2b3cf37bb08283.

perception/autoware_ptv3/config/ptv3.param.yaml
```diff
@@ -16,6 +16,23 @@
       engine_path: "$(var model_path)/$(var model_variant)_seg3d_head.engine"
       workspace_size: 8589934592  # 8 GiB
       source_reconstruction: "full"
+      # Mapping from segmentation3d.class_names to autoware::point_types::PointCloudClassification.
+      class_remap:
```
amadeuszsz: nit:
Would be nice to use same convention name as in `autoware-ml` which is `class_mapping`.

perception/autoware_ptv3/config/ptv3.param.yaml
```diff
@@ -16,6 +16,23 @@
...
+        vertical_thin: HAZARD
+        barrier: STRUCTURE
+        building: STRUCTURE
+        static_clutter: STRUCTURE
+        drivable_flat: FLAT_SURFACE
+        non_drivable_flat: FLAT_SURFACE
+        vegetation: VEGETATION
+        noise: NOISE
       filter:
         classes: ["drivable_flat", "non_drivable_flat", "noise"]
```
amadeuszsz, line 37: Imho now for filtering the points we should operate on mapped classes. Then it makes this parameter maintenance even simpler because it refers to hardcoded interface.

mojomex (approved):
Thanks, LGTM!

perception/autoware_ptv3/config/ptv3.param.yaml
ktro2828: Addressed in https://github.com/autowarefoundation/autoware_universe/pull/13201/commits/bab1645ca92a881187c401f9867db878cfb88a35.

PR #13203, opened 2026-08-14
feat(map_based_prediction): cut VRU paths jumping onto the road at road boundaries

satoshi-ota:
## Description

Adds a `RoadBoundaryModule` to the VRU predictor that trims predicted paths jumping out onto the road, cutting them at the right/left bound of `road` / `road_shoulder` lanelets. It mirrors the existing `FenceModule` (dedicated `road_boundary.cpp`, invoked only from `predictor_vru.cpp`), so it applies to VRUs only.

Applied to every VRU path, at the boundary crossing nearest the object, with these rules:

- Object already inside a road lanelet → not cut, so that a path *exiting* the road is not mistaken for a jump-out.
- Crossing inside a crosswalk / walkway → kept as a legitimate crossing, **unless** the crosswalk's signal is red (gated by `use_crosswalk_signal`). Crosswalk polygons use a 0.5 m margin for the containment test.
- Cut only when the object can stop before the boundary (per-class max deceleration); otherwise the jump-out is kept as unavoidable.

Intersection is judged on the path segments (polyline between consecutive points), not the object footprint.

The PR is split into two commits:

1. `feat(map_based_prediction): add per-class object deceleration parameters` — introduces `behavior_model.object_deceleration` and the shared `distance_to_stop_with_deceleration` helper in `utils`. A class with an entry of its own uses that value; every other class falls back to `base`, so only the classes that deviate need to be listed.
2. `feat(map_based_prediction): cut VRU paths jumping onto the road at road boundaries (#3213)` — the `RoadBoundaryModule` itself.

## Related links

**Parent Issue:**

- Cherry-pick of https://github.com/tier4/autoware_universe/pull/3213

## How was this PR tested?

Evaluated on the original PR against `feat/v0.64/e2e`:

- https://evaluation.ci.tier4.jp/evaluation/reports/b30bc52e-8670-561f-ae73-499b2ead64b4?project_id=x2_dev
- https://evaluation.ci.tier4.jp/evaluation/reports/3cab6acf-4013-541a-98cb-950a54e9814b?project_id=x2_dev

### A pedestrian moving from the sidewalk toward the roadway

image

### A pedestrian showing no sign of stopping on the sidewalk

image

### A pedestrian approaching a crosswalk with a green signal

image

### A pedestrian approaching a crosswalk with a red signal

Screenshot from 2026-07-22 13-20-10

### A pedestrian attempting to cross against a red signal

Screenshot from 2026-07-22 13-20-13

## Notes for reviewers

None.

#### Additions and removals

| Change type | Parameter Name                                  | Type     | Default Value | Description                                                          |
| :---------- | :---------------------------------------------- | :------- | :------------ | :------------------------------------------------------------------- |
| Added       | `behavior_model.object_deceleration.base`       | `double` | `-3.0`        | deceleration [m/ss] applied to classes without an entry of their own |
| Added       | `behavior_model.object_deceleration.pedestrian` | `double` | `-0.5`        | deceleration [m/ss] for pedestrian                                   |
| Added       | `behavior_model.object_deceleration.bicycle`    | `double` | `-1.0`        | deceleration [m/ss] for bicycle                                      |
| Added       | `behavior_model.object_deceleration.motorcycle` | `double` | `-2.0`        | deceleration [m/ss] for motorcycle                                   |

Only `base` is required. Every class defined in [`ObjectClassification.msg`](https://github.com/autowarefoundation/autoware_msgs/blob/main/autoware_perception_msgs/msg/ObjectClassification.msg) (`unknown` / `car` / `truck` / `bus` / `trailer` / `motorcycle` / `bicycle` / `pedestrian` / `animal` / `hazard` / `over_drivable` / `under_drivable`) can be added under the same namespace to override a single class; classes left out use `base`.

## Effects on system behavior

Predicted paths of VRUs (pedestrians / bicycles) that jump out from outside the road onto the road are now truncated at the road boundary, when the object could still decelerate to a stop before that boundary. Crossings at a crosswalk / walkway are left intact unless the crosswalk's signal is red.

This shortens some VRU predicted paths, so downstream planning modules may issue fewer stop / slow-down decisions for VRUs standing near the roadside.


technolojin (changes requested):
my mistake.

review is ongoing, and will request fixes.

perception/autoware_map_based_prediction/schema/map_based_prediction.schema.json
```diff
@@ -257,6 +257,75 @@
...
+          "properties": {
+            "object_deceleration": {
+              "type": "object",
+              "description": "deceleration [m/ss] assumed for each object class when judging whether the object can stop before a line it is predicted to cross. Only base is required; a class without an entry of its own uses the base value.",
+              "properties": {
+                "base": {
+                  "type": "number",
+                  "default": -3.0,
+                  "description": "deceleration [m/ss] applied to classes without an entry of their own"
+                },
```
technolojin, line 272:
```suggestion
                "base": {
                  "type": "number",
                  "default": -3.0,
                  "maximum": 0,
                  "description": "deceleration [m/ss] applied to classes without an entry of their own"
                },
```

perception/autoware_map_based_prediction/src/map_based_prediction_node.cpp
```diff
@@ -15,16 +15,64 @@
...
+
+utils::ObjectDecelerationParams declare_object_deceleration_params(
+  autoware::agnocast_wrapper::Node & node, const std::string & ns)
+{
+  utils::ObjectDecelerationParams params;
+  for (const auto & [label, object_label] : object_label_names) {
+    params.per_label.emplace(label, get_object_deceleration(node, ns, object_label));
+  }
+  return params;
+}
```
technolojin: The catch-all absorbs real misconfiguration.

Example: `pedestrian: -1` parses as an *integer*,
`declare_parameter<double>` throws `InvalidParameterTypeException`, and the class quietly gets `base`
(-3.0) — six times the intended magnitude, no log line. It also leaves the classes absent from the YAML
undeclared, so they never appear in `ros2 param list`.

Solution: Declaring each class with `base` as its default covers both, and a type mismatch then fails loudly at
start-up.

```suggestion
utils::ObjectDecelerationParams declare_object_deceleration_params(
  autoware::agnocast_wrapper::Node & node, const std::string & ns)
{
  const double base = get_or_declare_parameter<double>(node, ns + "base");
  utils::ObjectDecelerationParams params;
  for (const auto & [label, object_label] : object_label_names) {
    params.per_label.emplace(label, node.declare_parameter<double>(ns + object_label, base));
  }
  return params;
}
```


```
1787115323.6940541 [map_based_prediction-28]   what():  parameter 'behavior_model.object_deceleration.pedestrian' has invalid type: Wrong parameter type, parameter {behavior_model.object_deceleration.pedestrian} is of type {double}, setting it to {integer} is not allowed.
```

perception/autoware_map_based_prediction/include/autoware/map_based_prediction/predictor_vru/road_boundary.hpp
```diff
@@ -0,0 +1,73 @@
...
+  ///
+  /// A crossing that falls inside a crosswalk / walkway is normally exempt (kept), because it
+  /// represents a legitimate crossing. The exemption is overridden when @p is_crosswalk_signal_red
+  /// reports the crosswalk's signal as red: in that case the jump-out is cut like any other.
+  [[nodiscard]] std::vector<PredictedPath> cut_paths_crossing_road_boundary(
+    const autoware_perception_msgs::msg::PredictedObject & predicted_object,
+    bool object_within_road, const CrosswalkSignalRedFn & is_crosswalk_signal_red) const;
+
+private:
+  lanelet::LaneletMapUPtr road_boundary_layer_{nullptr};
```
technolojin:
```suggestion
  lanelet::LaneletMapConstUPtr road_boundary_layer_{nullptr};
```

perception/autoware_map_based_prediction/lib/predictor_vru/road_boundary.cpp
```diff
@@ -0,0 +1,216 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "autoware/map_based_prediction/predictor_vru/road_boundary.hpp"
+
+#include "autoware/map_based_prediction/utils.hpp"
+
+#include <autoware/motion_utils/trajectory/trajectory.hpp>
+#include <autoware/object_recognition_utils/object_classification.hpp>
+#include <autoware_lanelet2_extension/utility/query.hpp>
```
technolojin:
```suggestion
```

perception/autoware_map_based_prediction/lib/predictor_vru/predictor_vru.cpp
technolojin, line 21: because of this, better to add autoware_lanelet2_extension dependency in `package.xml`

perception/autoware_map_based_prediction/lib/predictor_vru/road_boundary.cpp
```diff
@@ -0,0 +1,216 @@
...
+#include <vector>
+
+namespace autoware::map_based_prediction
+{
+
+namespace
+{
+// Crosswalk polygons are expanded by this margin before the containment test so that intersection
+// points lying exactly on (or just outside) the crosswalk edge are still recognised as inside.
+constexpr double kCrosswalkContainmentMargin = 0.5;  // [m]
```
technolojin, line 43: may better to make it as a parameter under `behavior_model`

technolojin (changes requested):
Hysteresis of crossing-stop: like other units, is it better to introduce hysteresis to crossing/not-crossing decision?



For performance, better as one follow-up PR once this and #13190 have merged:

All line numbers are in `lib/predictor_vru/road_boundary.cpp` unless stated:

- `to2D(boundary).basicLineString()` at `:83` is recomputed per segment per boundary and returns by
  value (`lanelet2_core/primitives/LineString.h:345`); convert once next to `crossed_boundaries` (`:121`)
- cache the crosswalk polygons in `build_from_map` (`:146-177`) — `:66` rebuilds the ring per
  intersection point
- `:84` and `:88` compute the same sweep twice; `intersection` alone covers both
- the boundary layer built at `:154-174` carries ~23 % more linestrings than needed (see the inline
  comment there), so every query returns candidates that can never be the answer
- `lib/predictor_vru/fence.cpp`, this module, and `lib/predictor_vru/vegetation.cpp` once #13190 lands
  each rebuild the same `BasicLineString2d` from the same path; hoist it into the caller in
  `lib/predictor_vru/predictor_vru.cpp`

perception/autoware_map_based_prediction/include/autoware/map_based_prediction/predictor_vru/road_boundary.hpp
satoshi-ota: Fixed in https://github.com/autowarefoundation/autoware_universe/pull/13203/commits/c6a03e6078b89ca04f28efad5338080e933f47a2.

perception/autoware_map_based_prediction/schema/map_based_prediction.schema.json
satoshi-ota: Fixed in https://github.com/autowarefoundation/autoware_universe/pull/13203/commits/8484d63ca2e4a50719b728d83d99d40782f91663.

perception/autoware_map_based_prediction/src/map_based_prediction_node.cpp
satoshi-ota: Thanks for the suggestion. You are right, this would indeed cause side effects. Applied as suggested in 52da46b2c6.

For what it is worth, I modeled this implementation on https://github.com/autowarefoundation/autoware_universe/blob/3d6a95ca9eb0c9dfd95fbc0396f29316f46a94e6/planning/motion_velocity_planner/autoware_motion_velocity_obstacle_slow_down_module/src/parameters.hpp#L95-L108, so there are probably several other places in universe carrying the same problem that should be fixed separately from this PR.

technolojin (approved):

perception/autoware_map_based_prediction/lib/predictor_vru/road_boundary.cpp
satoshi-ota: Fixed in https://github.com/autowarefoundation/autoware_universe/pull/13203/commits/cb21e6f1480ad9b425a910b56042aad24f419a8e.

perception/autoware_map_based_prediction/lib/predictor_vru/predictor_vru.cpp
satoshi-ota: Fixed in https://github.com/autowarefoundation/autoware_universe/pull/13203/commits/cb21e6f1480ad9b425a910b56042aad24f419a8e.

perception/autoware_map_based_prediction/lib/predictor_vru/road_boundary.cpp
satoshi-ota: Thanks for pointing this out. That said, we do not expect this value to be tuned like a parameter at this point, so please let me keep it as a hard-coded variable until the need actually comes up.

